PR #471, opened 2017-10-07
Class or Object type can also be an interface

schalkms:
This adds some checks for that case.


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/util/MethodSignature.kt
```diff
@@ -0,0 +1,12 @@
+package io.gitlab.arturbosch.detekt.rules.bugs.util
+
+import org.jetbrains.kotlin.psi.KtFunction
+
+fun KtFunction.isEqualsMethod() =
+	this.name == "equals" && hasCorrectEqualsParameter()
```
vanniktech: `this.` is redundant here. Maybe that would also be a nice check although I'd assume it requires full type resolution.

schalkms: It definitely does.

Issue #455, opened 2017-10-02
Enum constant parameter flagged for MagicNumber


benwicks:
I have an enum class with an integer as a constructor parameter. One of the enum constants calls the constructor with a number that violates the `MagicNumber` rule (e.g. `3`).

I do not want to disable the `MagicNumber` rule for the whole project. How can I prevent this check from failing on this file and do so at the smallest scope possible?

I am interested in contributing but would need some direction. Thank you!

schalkms:
Thanks for reporting!
I agree. Detekt should not detect this case. 
You can of course supress it for now.

arturbosch:
Have you read about the baseline.xml file where you can add false positives? (See readme)

benwicks:
I did not know about the baseline. That could be useful! Thank you!

I see your point. The way I write the enums in question would be closer to:
```kotlin
enum class Bag(id: Int) {
    SMALL(1),
    EXTRA_LARGE(5)
}
```
So the enum constant name makes it clear what the "magic number" represents.

Is my best way to proceed to accept the `MagicNumber` rule violation and add it to my baseline?

arturbosch:
Ok, in your case I would also say no MagicNumber as `id` says everything we need to know! 

We have no symbol resolution to get to the right constructor and check then if the name of the variable `is enought` (configurable in the config?) but there is a PR open which fixes named arguments as MagicNumbers, so you could write:

```kotlin
enum class Bag(id: Int) {
	SMALL(id = 1),
	EXTRA_LARGE(id = 5)
}
```

which is kinda dumb as you are in the class itself ...

benwicks:
I see. I like the idea to configure no checking for enum classes.

Do you think it could be configurable down to the level of not checking just enum constant constructors? Or is that too fine-grained?

arturbosch:
Should be doable, will look at this 👍

Mauin:
What about a case like this? Are the numbers in this case more or less magic than in the examples above?

```
enum class Bag(id: Int, manufacturer: String, name: String, whatever: String) {
	SMALL(1, "brand", "smallbag", "something"),
	EXTRA_LARGE(5, "brand", "xlbag", "something")
}
```

benwicks:
@Mauin I would want a way for that example to _not_ be flagged for `MagicNumber`.

PR #473, opened 2017-10-07
Enum constant parameter flagged for MagicNumber

tagantroy:
Implementation for #455
@schalkms @arturbosch @benwicks please take a look

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
@@ -13,6 +13,7 @@ import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.KtAnnotationEntry
 import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtConstantExpression
+import org.jetbrains.kotlin.psi.KtEnumEntry
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtOperationReferenceExpression
 import org.jetbrains.kotlin.psi.KtPrefixExpression
@@ -39,6 +40,7 @@ class MagicNumber(config: Config = Config.empty) : Rule(config) {
 	private val ignoreAnnotation = valueOrDefault(IGNORE_ANNOTATION, false)
 	private val ignorePropertyDeclaration = valueOrDefault(IGNORE_PROPERTY_DECLARATION, false)
 	private val ignoreNamedArgument = valueOrDefault(IGNORE_NAMED_ARGUMENT, false)
+	private val ignoreEnums = valueOrDefault(IGNORE_ENUMS, false)
 
 	override fun visitConstantExpression(expression: KtConstantExpression) {
 		val parent = expression.parent
@@ -63,6 +65,7 @@ class MagicNumber(config: Config = Config.empty) : Rule(config) {
 		ignorePropertyDeclaration && parent is KtProperty && !parent.isLocal -> true
 		ignoreAnnotation && expression.isPartOf(KtAnnotationEntry::class) -> true
 		ignoreHashCodeFunction && expression.isPartOfHashCode() -> true
+		ignoreEnums && expression.isPartOf(KtEnumEntry::class) -> true
 		parent.isConstantProperty() -> true
 		ignoreNamedArgument
 				&& expression.isPartOf(KtValueArgument::class)
@@ -118,6 +121,7 @@ class MagicNumber(config: Config = Config.empty) : Rule(config) {
 		const val IGNORE_PROPERTY_DECLARATION = "ignorePropertyDeclaration"
 		const val IGNORE_ANNOTATION = "ignoreAnnotation"
 		const val IGNORE_NAMED_ARGUMENT = "ignoreNamedArgument"
+		const val IGNORE_ENUMS = "ignoreEnums"
 
 		private const val HEX_RADIX = 16
 		private const val BINARY_RADIX = 2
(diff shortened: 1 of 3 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumberSpec.kt
```diff
@@ -442,5 +442,33 @@ class MagicNumberSpec : Spek({
...
 		}
+		given("in enum constructor argument") {
+			val ktFile = compileContentForTest("""
+				enum class Bag(id: Int) {
+					SMALL(1),
+					EXTRA_LARGE(5)
+				}
+			""")
+			it("should be reported by default") {
+				assertThat(MagicNumber().lint(ktFile)).hasSize(1)
```
vanniktech: should not this be 2 since you have two enums?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
```diff
@@ -39,6 +40,7 @@ class MagicNumber(config: Config = Config.empty) : Rule(config) {
 	private val ignoreAnnotation = valueOrDefault(IGNORE_ANNOTATION, false)
 	private val ignorePropertyDeclaration = valueOrDefault(IGNORE_PROPERTY_DECLARATION, false)
 	private val ignoreNamedArgument = valueOrDefault(IGNORE_NAMED_ARGUMENT, false)
+	private val ignoreEnums = valueOrDefault(IGNORE_ENUMS, false)
```
arturbosch: Should this be true on default?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumberSpec.kt
```diff
@@ -442,5 +442,33 @@ class MagicNumberSpec : Spek({
...
+			}
+			it("numbers when 'ignoreEnums' is set to true"){
+				assertThat(MagicNumber(TestConfig(mapOf(MagicNumber.IGNORE_ENUMS to "true"))).lint(ktFile)).isEmpty()
```
arturbosch: Please add some variables to make it more clearer.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumberSpec.kt
```diff
@@ -442,5 +442,33 @@ class MagicNumberSpec : Spek({
...
+				enum class Bag(id: Int) {
+					SMALL(id = 1),
+					EXTRA_LARGE(id = 5)
+				}
+			""")
+			it("should be reported by default") {
+				assertThat(MagicNumber().lint(ktFile)).hasSize(1)
+			}
+			it("numbers when 'ignoreEnums' is set to true"){
+				assertThat(MagicNumber(TestConfig(mapOf(MagicNumber.IGNORE_ENUMS to "true"))).lint(ktFile)).isEmpty()
```
arturbosch: Here too

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumberSpec.kt
arturbosch: I think 1 is ignored on default in the config.

vanniktech: Oh yes right.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
tagantroy: @arturbosch 
but if I set it true by default it can affect behaviour after migration from previous version

Mauin: I'd leave this `false` by default. If I got this right, this would also ignore values as part of an enum like this: `EnumEntry(25, many, more, parameters)`. In that case it would be more of a magic number than if its the single enum entry argument.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
```diff
@@ -63,6 +65,7 @@ class MagicNumber(config: Config = Config.empty) : Rule(config) {
 		ignorePropertyDeclaration && parent is KtProperty && !parent.isLocal -> true
 		ignoreAnnotation && expression.isPartOf(KtAnnotationEntry::class) -> true
 		ignoreHashCodeFunction && expression.isPartOfHashCode() -> true
+		ignoreEnums && expression.isPartOf(KtEnumEntry::class) -> true
```
Mauin: Really more of a question (which didn't come up in the issue): What if the Enum entry contains lots of arguments and the "magic number" is just one of many. Should that change anything for this rule?

benwicks: In my case, I just have one potentially magic number per any given enum constant.

arturbosch: If the issue description is overridden with each constant value -> yes, thats helpful, otherwise redundant information. Looking at the code we do not override it. Will open an issue for that as this PR seems finished.

arturbosch:
Thanks for the PR!

PR #474, opened 2017-10-07
Rule:NestedClassVisibility update

schalkms:
This rule should not detect nested classes where the nesting depth is
deeper than 1. An enum class should be excluded.
The issue should be generic and not depend on the violated class name.


Mauin:
This PR is empty now? Was it merged to master manually?

Mauin:
Closing this PR as the commits are already on `master` (probably merged manually). Feel free to reopen if needed.

Issue #465, opened 2017-10-06
MagicNumber: Should methods that only return that number be flagged?


Mauin:
`override fun getTheNumber(): Int = 123`

Should `123` be flagged as a magic number? The method name already gives it some context.

schalkms:
No. I have a better idea.
This is another code smell. Methods should not return only constants. This includes **"magic numbers"**.
I would implement this after approval. :)

arturbosch:
Good idea @schalkms, go ahead ^^

PR #477, opened 2017-10-08
Fix for #465

schalkms:


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
```diff
@@ -70,16 +71,16 @@ class MagicNumber(config: Config = Config.empty) : Rule(config) {
...
+	private fun KtConstantExpression.isPartOfFunctionReturnConstant() =
+			parent is KtNamedFunction || (parent is KtReturnExpression && parent.parent.children.size == 1)
+
+
 	private fun KtConstantExpression.isPartOfHashCode(): Boolean {
 		val containingFunction = getNonStrictParentOfType(KtNamedFunction::class.java)
-		val name = containingFunction?.name
 		val returnType = containingFunction?.typeReference?.node?.text
-		return nameIsHashCode(name) && returnTypeIsInt(returnType)
+		return containingFunction?.name == "hashCode" && returnType == "Int"
```
vanniktech: Wasn't there an extension method defined for checking whether a method is a hashCode method?

schalkms: Yes, in PR #471
I'll wait for both to be merged. Thanks!

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/FunctionOnlyReturningConstant.kt
```diff
@@ -0,0 +1,46 @@
...
+class FunctionOnlyReturningConstant(config: Config = Config.empty) : Rule(config) {
+
+	override val issue = Issue(javaClass.simpleName, Severity.Style,
+			"A function that only returns a constant is misleading. " +
+					"Consider declaring a constant instead",
+			Debt.TEN_MINS)
+
+	override fun visitNamedFunction(function: KtNamedFunction) {
+		if (isConstantExpression(function.bodyExpression) || returnsConstant(function)) {
+			report(CodeSmell(issue, Entity.from(function)))
```
Mauin: Instead of the generic description above, should we give the concrete description here?

`"${function.name} only returns a constant. A function that only returns a constant is misleading. Consider declaring a constant instead."`

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
```diff
@@ -43,7 +44,7 @@ class MagicNumber(config: Config = Config.empty) : Rule(config) {
 	override fun visitConstantExpression(expression: KtConstantExpression) {
 		val parent = expression.parent
 
-		if (isIgnoredByConfig(parent, expression)) {
+		if (isIgnoredByConfig(parent, expression) || expression.isPartOfFunctionReturnConstant()) {
```
Mauin: Suppose that I have the `MagicNumber` rule enabled, but the `FunctionOnlyReturningConstant` disabled. Should the functions returning a constant then still be flagged as `MagicNumbers`?

I think I'm okay with this behavior, just want to raise the question.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/FunctionOnlyReturningConstant.kt
schalkms: In my opinion no. Other analysis tools also have a generic description which is fine.

arturbosch: We need both! The generic description is used in sonar-kotlin in the rule profiles and the concrete description is something we aim for in the long term for all rules (if possible). Getting faster from issue description to source code position imo.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
arturbosch: I'm for reporting it. Returning a constant in a function is such stupid that giving the user two issues instead of one sounds good to me :D ?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
arturbosch: Is merged! :)

detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -309,6 +309,8 @@ style:
   FunctionMaxLength:
     active: false
     maximumFunctionNameLength: 30
+  FunctionOnlyReturningConstant:
```
arturbosch: Why is the rule between `FunctionMax/-MinLength`?

schalkms: I have no explanation :)

schalkms: Should I move it?

arturbosch: Yes, please. To the end? :)

schalkms: Done.
Some kind of sorting order (alphabetically) for rules would be nice also.

arturbosch: When generating the default-config, we should definitely consider this ^^

Issue #475, opened 2017-10-07
Update rule descriptions and adapt debt minutes


schalkms:
In my opinion the rule description should follow a simple template.
This helps to make the violations more understandable for the user.

I think the debt minutes should also be updated accordingly.

Mauin:
What kind of template do you suggest? In my opinion the description should be as fine grained as possible to give the user a clear understanding of what detekt is not happy about in that piece of code. (See also #273)

schalkms:
I agree. I think there should be at least.
- What caused the violation (generic approach)
- What result the code causes (think of **!!** for example)
- How the user can fix it

I am of course open to improvements.

Mauin:
In general I really like how Android lint handles Issues and their description. 
Each issue has a short name, a long name, and a detailed description explaining the rationale of the Issue. This description explains why the issue is an issue and how it can be fixed.

Additionally every time an `Issue` is reported a `message` is attached which explains the specific case. This message refers to the specific code in most of the cases to point out what exactly is wrong.

As an example: 
- [The Android Lint Issue](https://android.googlesource.com/platform/tools/base/+/master/lint/libs/lint-checks/src/main/java/com/android/tools/lint/checks/LogDetector.java#59)
- [How the issue is reported](https://android.googlesource.com/platform/tools/base/+/master/lint/libs/lint-checks/src/main/java/com/android/tools/lint/checks/LogDetector.java#145)

schalkms:
Ok. I am more a fan of the sonarlint implementation but I can live with others as well. I like the separate message field. I don't want to mix the specific code details with the description of the rule.

arturbosch:
In another issue we discussed the detailed generic description with an example above the rule as KDoc. From this kdoc we could generate like md files and display them all together on a webpage?

The short name is basically our `RuleName`. The long name is kinda obvious if the short name is good. With the detailed kdoc, the issue description should be generic (for sonar rule profiles) and overridden if needed for a specific finding.

schalkms:
Good idea! I like it. This is already documented in #343
I would also combine them with compliant and noncompliant code snippets. However, I would write this snippets in a separate file.

arturbosch:
Yes, I think @vanniktech mentioned the compliant and noncompliant code snippets ^^ (or maybe you oO, I give you both credits ^^)

vanniktech:
I think it was him because I don't know what you're talking about :D

Mauin:
[...]

Issue #273, opened 2017-08-04
Giving CodeSmells better descriptions per case


Mauin:
Currently all `CodeSmell` findings just report the `Issue` of the given rule. While the `Issue` has a description it is usually very generic. That is okay for some rules (e.g. MaxLineLength) but many rules might benefit from better descriptions that refer to the actual case of the violation.

As an example (Using the `ModifierOrder`): `lateinit internal private val test: String` could report a `ModifierOrder` issue with a smarter description. Something along the lines of: "The modifiers should be sorted as: `private internal lateinit`". 

Other rules could also give more detailed descriptions on the specific violations. This can also be used to already give the user a better idea on how to fix the issue.

This would need some changes to the `CodeSmell` giving it an optional description parameter that overrides the default one or adding a new subclass of `Finding`.

arturbosch:
I like the idea! The issue handles the description for a code smell. As issue is a data class we can use `issue.copy(description = "my new description")` within the report function.

I also haven't figured out how to present the descriptions to the user (not using sonarqube). Printing them next to the location on a second line or only print them on demand (--descriptions)

Mauin:
I started using `issue.copy(description = "...")` in #328 and #320 to get this going. But it still feels a bit awkward and I think we should make it mandatory to give a good description with a clear message and reference to the current issue rather than copying the default issue. 

I'd prefer to report issues with:
```
report(CodeSmell(issue, Entitiy.from(argument), "This is a super clear message and has to be provided"))
```

That would force us to write case-by-case messages.

schalkms:
@arturbosch @Mauin this is also done, isn't it?

PR #480, opened 2017-10-10
Add message to CodeSmell

Mauin:
This is just a suggestion and currently won't compile as we'd have to update every usage of `CodeSmell`.

After the discussions in #475 and #273 it seemed to me like splitting an Issues description from a CodeSmell's detailed message makes sense.

With this the Issue description can act as a rationale for the Issue, what it means, the reasoning behind the rule and how it can be fixed. (e.g. "Long function names should be avoided because.......<insert long and detailed description and reasoning here>")

The CodeSmell message can then point out the specific problem directly. (e.g. "The function ${function.name} has a ${function.name.lenght} character long name which is longer than the threshold of $maxFunctionLength. Give this function a shorter name")

Let me know what you think. If you agree with this I can go through all places we report CodeSmells and give them an empty message to start with `""` and we should update them over time.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
@@ -8,6 +8,7 @@ package io.gitlab.arturbosch.detekt.api
  */
 open class CodeSmell(final override val issue: Issue,
 					 override val entity: Entity,
+					 override val message: String,
 					 override val metrics: List<Metric> = listOf(),
 					 override val references: List<Entity> = listOf()) : Finding {
 
@@ -18,7 +19,12 @@ open class CodeSmell(final override val issue: Issue,
 	override fun compactWithSignature() = compact() + " - Signature=" + entity.signature
 
 	override fun toString(): String {
-		return "CodeSmell(issue=$issue, entity=$entity, metrics=$metrics, references=$references, id='$id')"
+		return "CodeSmell(issue=$issue, " +
+				"entity=$entity, " +
+				"message=$message, " +
+				"metrics=$metrics, " +
+				"references=$references, " +
+				"id='$id')"
 	}
 }
 
@@ -27,10 +33,10 @@ open class CodeSmell(final override val issue: Issue,
  * for the existence of this rule violation.
  */
 open class CodeSmellWithReferenceAndMetric(
-		issue: Issue, entity: Entity, val reference: Entity, metric: Metric) : ThresholdedCodeSmell(
-		issue, entity, metric, references = listOf(reference)) {
+		issue: Issue, entity: Entity, private val reference: Entity, message: String, metric: Metric) : ThresholdedCodeSmell(
+		issue, entity, metric, message, references = listOf(reference)) {
 
-	override fun compact() = "$id - $metric - ref=${reference.name} - ${entity.compact()}"
+	override fun compact() = "$id - $metric - ref=${reference.name} - ${entity.compact()} - message=$message"
 }
 
 /**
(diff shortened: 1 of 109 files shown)


arturbosch:
So the `issue.description` is the generic message and the `CodeSmell.message` is the concrete? I like it. It will remove `issue.copy` :+1:

Mauin:
@arturbosch Exactly. Then I will go ahead and add the `CodeSmell.message` everywhere as well

Mauin:
I now added a `message` to every constructor of `CodeSmell` (thanks IntelliJ for doing almost all the work!!). 

In most cases this detailed message is just `""` and we should add a more detailed message later. I didn't want to go through every `CodeSmell` we report and think of a message that seemed logical to me.. 🙃 

Also I moved all cases where we did `issue.copy(description = "...")` and took that detailed message as the new `CodeSmell#message`.

arturbosch:
Argh just came by to say "hold on until my PR is through!" :D. Thats a huge PR and will conflict with mine. I refactor the `MultiRule`. But no worries we will get both things merged soon.

Mauin:
@arturbosch Whoooooops 🙃 

But the changes in here are fairly "simple" to rebase, so don't worry too much. Mostly just a lot of `message = ""` everywhere.

arturbosch:
I want to wait with this PR after RC5 release (hopefully soon).

Mauin:
@arturbosch Now that RC5 is out, should I get started on rebasing this and getting all these conflicts resolved?

arturbosch:
Yes, please. 

I don't think we should print the description of each finding. This will fill up the console and is printed after the findings path. Having the description in the output format is enough imo.

schalkms:
I agree with @arturbosch 
Detekt should not fill up the console.

Mauin:
Finally got around to rebasing this. This change currently won't print the `message` of each finding in the CLI output. It will use either the `message` in the XML output if present, otherwise it falls back to the `issue#description`.

arturbosch:
Great! Will check it out tomorrow.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
```diff
@@ -27,23 +33,27 @@ open class CodeSmell(final override val issue: Issue,
  * for the existence of this rule violation.
  */
 open class CodeSmellWithReferenceAndMetric(
-		issue: Issue, entity: Entity, val reference: Entity, metric: Metric) : ThresholdedCodeSmell(
-		issue, entity, metric, references = listOf(reference)) {
+		issue: Issue, entity: Entity, private val reference: Entity, message: String, metric: Metric) : ThresholdedCodeSmell(
+		issue, entity, metric, message, references = listOf(reference)) {
 
-	override fun compact() = "$id - $metric - ref=${reference.name} - ${entity.compact()}"
+	override fun compact() = "$id - $metric - ref=${reference.name} - ${entity.compact()} - message=$message"
```
arturbosch: This still gets printed in the console

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
```diff
@@ -27,23 +33,27 @@ open class CodeSmell(final override val issue: Issue,
...
+		message: String,
+		references: List<Entity> = emptyList()) : CodeSmell(
+		issue, entity, message, metrics = listOf(metric), references = references) {
 	val value: Int
 		get() = metric.value
 	val threshold: Int
 		get() = metric.threshold
 
-	override fun compact() = "$id - $metric - ${entity.compact()}"
+	override fun compact() = "$id - $metric - ${entity.compact()} = message=$message"
```
arturbosch: Here too

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/XmlOutputReport.kt
```diff
@@ -30,11 +30,16 @@ class XmlOutputReport : OutputReport() {
 		smells.groupBy { it.location.file }.forEach { fileName, findings ->
 			lines += "<file name=\"${fileName.toXmlString()}\">"
 			findings.forEach {
+				val message = when {
```
arturbosch: I would move that logic into the CodeSmell class and use a function like `messageOrDescription`. It feels a bit like feature envy.

Mauin: I'm hoping that at some point all `CodeSmell`s will just have a valid message so we can remove this again. But for now it makes sense to at least print whatever is there.

arturbosch:
Thank you!

PR #482, opened 2017-10-10
ignore "_" named variables for VariableNaming and VariableMinLength

### LINKED ISSUE #481: VariableNaming and VariableMinLength: Exclude simple Underscore from Rule
detekt marks an underscore `_` as an VariableNaming and VariableMinLength smell. But this is a correct feature of Kotlin. See https://kotlinlang.org/docs/reference/multi-declarations.html#underscore-for-unused-variables-since-11

Underscore should be excluded from these Rules

Mauin:
Resolves #481 

This excludes variables named `_` from the `VariableNaming` rule and the `VariableMinLength` rule.

In case of the `VariableNaming` we could also rely on having it in the Regex. What do you think?

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/naming/VariableNaming.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/naming/VariableNaming.kt
@@ -8,6 +8,7 @@ import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.rules.SubRule
 import org.jetbrains.kotlin.psi.KtVariableDeclaration
+import org.jetbrains.kotlin.resolve.calls.util.isSingleUnderscore
 
 class VariableNaming(config: Config = Config.empty) : SubRule<KtVariableDeclaration>(config) {
 	override val issue = Issue(javaClass.simpleName,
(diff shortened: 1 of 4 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/naming/VariableNaming.kt
```diff
@@ -16,6 +17,10 @@ class VariableNaming(config: Config = Config.empty) : SubRule<KtVariableDeclarat
 	private val variablePattern = Regex(valueOrDefault(VARIABLE_PATTERN, "^(_)?[a-z$][a-zA-Z$0-9]*$"))
 
 	override fun apply(element: KtVariableDeclaration) {
+		if (element.isSingleUnderscore) {
```
Mauin: Should we rely on this or allow only `_` in the default regexes we set for this rule?

schalkms: I would include it in the default regex.

vanniktech: It's probably better and it should be as easy as `_|(...)`.

Mauin: Only downside that I can see from that is that people would have to update their configs (or use the default) to see any effect from this change. But we can mention this in the release notes. I'll change it to use the Regex

vanniktech: Unless they use the `failFast` and don't override the value. 😁  But yeah you're right.

Mauin: I guess it's more of a thought if we consider that anyone would actually want to exclude `_` from both of these rules. If we say that `_` should be up to the configuration of the user for variable naming rules then it should probably also be a configurable parameter in the `MinVariableLength` rule.

arturbosch: In my opinion `_` should be excluded in the code not the config. It is a kotlin feature.

Mauin: Was thinking the same. So I'll leave it like this.

PR #483, opened 2017-10-11
don't report UselessPostfixExpressions on fields in return expressions

### LINKED ISSUE #466: UselessPostfixExpression: Postfix on field in return
detekt marks the postfix expressions in the following case as useless. However the value of the field will be updated and affects the next call of the function.

```
class Test {
    private var runningId: Long = 0

    fun getId(): Long {
        return runningId++
    }
}
```

Mauin:
Fixes #466

This would probably be even easier with full type resolution but it also works like this.

We keep a list of all fields and check each postfix expression on return types against that list.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UselessPostfixExpression.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UselessPostfixExpression.kt
@@ -7,8 +7,10 @@ import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import org.jetbrains.kotlin.psi.KtBinaryExpression
+import org.jetbrains.kotlin.psi.KtClass
 import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtPostfixExpression
+import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.KtReturnExpression
 
 class UselessPostfixExpression(config: Config = Config.empty) : Rule(config) {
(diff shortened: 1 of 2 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UselessPostfixExpression.kt
```diff
@@ -39,6 +51,12 @@ class UselessPostfixExpression(config: Config = Config.empty) : Rule(config) {
 		}
 	}
 
+	private fun KtPostfixExpression.isOnField(): Boolean {
+		return properties.map { it.name }
```
arturbosch: Store the properties as a `Set<String>` and use here a `contains` -> constant time instead of linear.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UselessPostfixExpression.kt
```diff
@@ -7,20 +7,32 @@ import io.gitlab.arturbosch.detekt.api.Issue
...
 
 class UselessPostfixExpression(config: Config = Config.empty) : Rule(config) {
 
 	override val issue: Issue = Issue("UselessPostfixExpression", Severity.Defect,
 			"The incremented or decremented value is unused. This value is replaced with the original value.")
 
+	val properties = mutableListOf<KtProperty>()
+
+	override fun visitClass(klass: KtClass) {
+		properties.addAll(klass.getProperties())
```
arturbosch: Make `properties` a var and reset the set (after comment below) after each class, just to be really safe.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UselessPostfixExpression.kt
```diff
@@ -7,20 +7,32 @@ import io.gitlab.arturbosch.detekt.api.Issue
...
+		super.visitClass(klass)
+	}
+
 	override fun visitReturnExpression(expression: KtReturnExpression) {
 		val postfixExpression = expression.returnedExpression as? KtPostfixExpression
-		if (postfixExpression != null) {
-			report(postfixExpression)
+		postfixExpression?.let {
+			if (!postfixExpression.isOnField()) {
```
arturbosch: Instead of a `?.let` then `if`, just but both conditions in the `if`.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UselessPostfixExpression.kt
arturbosch: There could be some false negatives -> a local variable or parameter has the same name as the field. Uff even recursively all scopes ...

Idea: Get the function `KtReturnStmt.getParent<KtNamedFunction>()` and find all locals inside that function. If a local has the same name as `PostFixIncr` and the position is before this -> report

Mauin: Oh, good point on that last one... I'll add some test cases for that and add handling for it. 👍

Issue #69, opened 2017-06-04
Support for Android? Failed to apply plugin [id 'com.android.application']


jaredsburrows:
I am trying to applying to my kotlin supported project here: https://github.com/jaredsburrows/android-gif-example.

## Versions:
- `com.android.tools.build:gradle:2.3.2`
- `org.jetbrains.kotlin:kotlin-gradle-plugin:1.1.2-3`
- `gradle.plugin.io.gitlab.arturbosch.detekt:detekt-gradle-plugin:1.0.0.M10.3`

## Error:
```
* What went wrong:
A problem occurred evaluating root project 'android-gif-example'.
> Failed to apply plugin [id 'com.android.application']
   > Could not create plugin of type 'AppPlugin'.
      > com.google.common.collect.Lists.newArrayListWithCapacity(I)Ljava/util/ArrayList;

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED in 16s
```

I am applying the `apply plugin: "io.gitlab.arturbosch.detekt"` plugin after both android and kotlin plugins.

jaredsburrows:
@scottkennedy has a work around for those that want to use it:

Setup custom configurations:
https://github.com/arturbosch/detekt/issues/24#issuecomment-306070140

Setup custom `JavaExec` tasks:
https://github.com/arturbosch/detekt/issues/24#issuecomment-306069600

arturbosch:
Tested, getting the same error. Hmm maybe it's due to usage of different kotlin compiler versions. Need to investigate this later ...

arturbosch:
I have the feeling that because I'm using the kotlin-compiler which also have guava etc classes, there can be conflicts. Will try to migrate to embeddable compiler.

arturbosch:
Will be fixed in M11 today later.

jaredsburrows:
@arturbosch Thank you!

@scottkennedy We should be able to use the plugin directly soon.

arturbosch:
When using the branch use `gradle detektCheck` for inspection. It's an undocumented change for M11 for now ;).

arturbosch:
Pls reopen the issue if something does not work :)

YajJackson:
Confused about why this issue was closed..

Mauin:
@YajJackson Are you seeing issues when using _Detekt_ in an Android project? In theory there should be no issues and we see many Android projects using Detekt and it's Gradle Plugin

[...]

PR #485, opened 2017-10-14
WIP: Type and symbol resolving

3flex:
This PR adds type and symbol resolving for Kotlin code that targets the JVM.

A lot of work is required before it's ready to merge, but I wanted to share it now so that people can provide early feedback.

Some things that are missing (others will surely have things to add to this list):

- [ ] Gradle integration
- [ ] Auto-setting classpath in Gradle
- [ ] Make type resolution optional in Gradle
- [ ] Tests for the new classes
- [ ] Tests for the new `UselessCallOnNotNull` rule (if the rule seems worth introducing)
- [ ] Sharing compiler errors/warnings with the user (they are not shown, so user has to know that their code compiles before running detekt, otherwise there'll be an error)
- [ ] Support for JS modules
- [ ] Support for "common" modules
- [ ] Description for `--classpath` command line option

I've seen concerns around the possible performance hit of type and symbol resolving, which seem warranted - this analysis takes time as a lot more work is being done. But this allows for much more complex analysis to be performed by detekt and will make the tool a lot more useful, so I think the tradeoff will be worth it.

To test, enable the `UselessCallOnNotNull` rule, add kotlin-stdlib to the classpath with the `-cp` or `--classpath` flag, and try running on a file like this:
```kotlin
package demo

val nextTest = mutableListOf<String>().orEmpty()
```
The new rule will resolve the `orEmpty()` call to `kotlin.collections.orEmpty` which will trigger a detekt issue because the `orEmpty()` call is never needed on a not-null collection.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
@@ -1,17 +1,20 @@
 package io.gitlab.arturbosch.detekt.api
 
 import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.resolve.BindingContext
 
 @Suppress("EmptyFunctionBlock")
 abstract class BaseRule(protected val context: Context = DefaultContext()) : DetektVisitor(), Context by context {
 
 	open val id: String = javaClass.simpleName
+	var bindingContext: BindingContext = BindingContext.EMPTY
 
 	/**
 	 * Before starting visiting kotlin elements, a check is performed if this rule should be triggered.
 	 * Pre- and post-visit-hooks are executed before/after the visiting process.
 	 */
-	fun visitFile(root: KtFile) {
+	fun visitFile(root: KtFile, bindingContext: BindingContext = BindingContext.EMPTY) {
+		this.bindingContext = bindingContext
 		if (visitCondition(root)) {
 			clearFindings()
 			preVisit(root)
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
@@ -12,6 +12,10 @@ abstract class BaseRule(protected val context: Context = DefaultContext()) : Det
 	/**
 	 * Before starting visiting kotlin elements, a check is performed if this rule should be triggered.
 	 * Pre- and post-visit-hooks are executed before/after the visiting process.
+	 * BindingContext holds the result of the semantic analysis of the source code by the Kotlin compiler. Rules that
+	 * rely on symbols and types being resolved can use the BindingContext for this analysis. Note that
+	 * BindingContext will have the value BindingContext.EMPTY unless detekt is provided the correct classpath while
+	 * analyzing the code.
 	 */
 	fun visitFile(root: KtFile, bindingContext: BindingContext = BindingContext.EMPTY) {
 		this.bindingContext = bindingContext
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Args.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Args.kt
@@ -13,6 +13,9 @@ class Args {
 			converter = ExistingPathConverter::class, description = "Input path to analyze (path/to/project).")
 	private var input: Path? = null
 
+	@Parameter(names = arrayOf("--classpath", "-cp"))
+	var classpath: String? = null
+
 	@Parameter(names = arrayOf("--filters", "-f"),
 			description = "Path filters defined through regex with separator ';' (\".*test.*\").")
 	var filters: String? = null // Using a converter for List<PathFilter> resulted in a ClassCastException
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -325,6 +325,8 @@ style:
     active: false
   OptionalReturnKeyword:
     active: false
+  UselessCallOnNotNull:
+    active: false
   OptionalUnit:
     active: false
   ProtectedMemberInFinalClass:
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/AnalysisEnvironment.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/AnalysisEnvironment.kt
@@ -6,14 +6,13 @@ import org.jetbrains.kotlin.cli.jvm.compiler.CliLightClassGenerationSupport
 import org.jetbrains.kotlin.cli.jvm.compiler.EnvironmentConfigFiles
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.jetbrains.kotlin.cli.jvm.compiler.TopDownAnalyzerFacadeForJVM
-import org.jetbrains.kotlin.cli.jvm.config.addJavaSourceRoot
 import org.jetbrains.kotlin.cli.jvm.config.addJvmClasspathRoot
 import org.jetbrains.kotlin.cli.jvm.config.addJvmClasspathRoots
 import org.jetbrains.kotlin.com.intellij.openapi.Disposable
 import org.jetbrains.kotlin.com.intellij.openapi.util.Disposer
 import org.jetbrains.kotlin.config.CommonConfigurationKeys
 import org.jetbrains.kotlin.config.CompilerConfiguration
-import org.jetbrains.kotlin.config.addKotlinSourceRoot
+import org.jetbrains.kotlin.config.addKotlinSourceRoots
 import org.jetbrains.kotlin.container.ComponentProvider
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.resolve.lazy.declarations.FileBasedDeclarationProviderFactory
@@ -52,13 +51,7 @@ class AnalysisEnvironment : Disposable {
 	}
 
 	fun addSources(list: List<String>) {
-		list.forEach {
-			configuration.addKotlinSourceRoot(it)
-			val file = File(it)
-			if (file.isDirectory || file.extension == ".java") {
-				configuration.addJavaSourceRoot(file)
-			}
-		}
+		configuration.addKotlinSourceRoots(list)
 	}
 
 	override fun dispose() {
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/AnalysisEnvironment.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/AnalysisEnvironment.kt
@@ -1,0 +1,66 @@
+package io.gitlab.arturbosch.detekt.core
+
+import org.jetbrains.kotlin.cli.common.CLIConfigurationKeys
+import org.jetbrains.kotlin.cli.common.messages.MessageCollector
+import org.jetbrains.kotlin.cli.jvm.compiler.CliLightClassGenerationSupport
+import org.jetbrains.kotlin.cli.jvm.compiler.EnvironmentConfigFiles
+import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
+import org.jetbrains.kotlin.cli.jvm.compiler.TopDownAnalyzerFacadeForJVM
+import org.jetbrains.kotlin.cli.jvm.config.addJavaSourceRoot
+import org.jetbrains.kotlin.cli.jvm.config.addJvmClasspathRoot
+import org.jetbrains.kotlin.cli.jvm.config.addJvmClasspathRoots
+import org.jetbrains.kotlin.com.intellij.openapi.Disposable
+import org.jetbrains.kotlin.com.intellij.openapi.util.Disposer
+import org.jetbrains.kotlin.config.CommonConfigurationKeys
+import org.jetbrains.kotlin.config.CompilerConfiguration
+import org.jetbrains.kotlin.config.addKotlinSourceRoot
+import org.jetbrains.kotlin.container.ComponentProvider
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.resolve.lazy.declarations.FileBasedDeclarationProviderFactory
+import java.io.File
+
+class AnalysisEnvironment : Disposable {
+	val configuration = CompilerConfiguration()
+
+	init {
+		configuration.put(CLIConfigurationKeys.MESSAGE_COLLECTOR_KEY, MessageCollector.NONE)
+		configuration.put(CommonConfigurationKeys.MODULE_NAME, "module for detekt static analysis")
+	}
+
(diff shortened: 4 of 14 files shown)


3flex:
To test, enable the `UselessCallOnNotNull` rule, add kotlin-stdlib to the classpath with the `-cp` or `--classpath` flag, and try running on a file like this:
```kotlin
package demo

val nextTest = mutableListOf<String>().orEmpty()
```
The new rule will resolve the `orEmpty()` call to `kotlin.collections.orEmpty` which will trigger a detekt issue.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
```diff
@@ -1,15 +1,19 @@
...
 abstract class BaseRule(protected val context: Context = DefaultContext()) : DetektVisitor(), Context by context {
 
+	var bindingContext: BindingContext = BindingContext.EMPTY
+
 	/**
 	 * Before starting visiting kotlin elements, a check is performed if this rule should be triggered.
 	 * Pre- and post-visit-hooks are executed before/after the visiting process.
 	 */
-	fun visitFile(root: KtFile) {
+	fun visitFile(root: KtFile, bindingContext: BindingContext = BindingContext.EMPTY) {
```
schalkms: Maybe add a comment here for future readers what **BindingContext** does.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Args.kt
```diff
@@ -13,6 +13,9 @@ class Args {
 			converter = ExistingPathConverter::class, description = "Input path to analyze (path/to/project).")
 	private var input: Path? = null
 
+	@Parameter(names = arrayOf("--classpath", "-cp"))
```
schalkms: Can this be made private?

detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -348,6 +348,8 @@ style:
     active: false
   UnnecessaryAbstractClass:
     active: false
+  UselessCallOnNotNull:
```
schalkms: nice rule 👍

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/AnalysisEnvironment.kt
```diff
@@ -0,0 +1,67 @@
...
+
+	fun addClasspaths(paths: List<File>) {
+		configuration.addJvmClasspathRoots(paths)
+	}
+
+	fun addClasspath(path: File) {
+		configuration.addJvmClasspathRoot(path)
+	}
+
+	fun addSources(list: List<String>) {
```
schalkms: `file.extension == ".kt"`?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.calls.callUtil.getType
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+import org.jetbrains.kotlin.types.TypeUtils
+
+class UselessCallOnNotNull(config: Config = Config.empty) : Rule(config) {
+	override val issue: Issue = Issue("UselessCallOnNotNull",
+			Severity.Performance,
+			"This call on not-null type may be reduced or removed.")
```
schalkms: Can you please write a more detailed description of the rule and what can be done to remove this issue? This might help users of **detekt** to fix this issue.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
```diff
@@ -0,0 +1,62 @@
...
+
+	private val uselessFqNames = mapOf("kotlin.collections.orEmpty" to deleteConversion,
+			"kotlin.text.orEmpty" to deleteConversion,
+			"kotlin.text.isNullOrEmpty" to Conversion("isEmpty"),
+			"kotlin.text.isNullOrBlank" to Conversion("isBlank"))
+
+	private val uselessNames = toShortFqNames(uselessFqNames.keys)
+
+	@Suppress("ReturnCount")
```
schalkms: Maybe try to refactor that method and remove the `@Supress`.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Args.kt
```diff
@@ -13,6 +13,9 @@ class Args {
...
+	var classpath: String? = null
```
schalkms: Can this be made private?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/AnalysisEnvironment.kt
```diff
@@ -0,0 +1,67 @@
...
+		list.forEach {
+			configuration.addKotlinSourceRoot(it)
+			val file = File(it)
+			if (file.isDirectory || file.extension == ".java") {
```
schalkms: `file.extension == ".kt"`?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
3flex: I think any other way would reduce clarity... do you have a suggestion?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Args.kt
3flex: It must be public as it's referenced in https://github.com/arturbosch/detekt/pull/485/files#diff-7fc579b3f16f6250e5ede30c123a96ca.

The other variables are referenced directly in other parts of the code as well, none can be made private without some refactoring.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/AnalysisEnvironment.kt
3flex: This code (like most in this PR) is based on the code in dokka. Taking a closer look the `addSources` function will only ever be called with a list paths to Kotlin files so I'll remove the unused code.

Some new code will need to be added to ensure analysis works in mixed Kotlin/Java codebases.

arturbosch:
Wow, thanks for the contribution! That's a big step for detekt :)
I will need some time before I can review this PR though!

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
schalkms: Yes. I would refactor the first part of this method into it's own method.

arturbosch:
My test case does not work unfortunately:
```kotlin
	@Test
	fun test() {
		val file = compileForTest(Case.BindingContext.path())
		val rule = UselessCallOnNotNull()
		val resolver = DetektResolver(
				listOf("/home/artur/.m2/repository/org/jetbrains/kotlin/kotlin-runtime/1.1.2/kotlin-runtime-1.1.2.jar"),
				listOf(File(file.getUserData(KtCompiler.RELATIVE_PATH)!!).absolutePath), emptyList(), TestConfig())
		val bindingContext = resolver.generate(listOf(file))
		rule.visitFile(file, bindingContext)

		println(rule.findings)
	}
```

3flex:
Can you try with `kotlin-stdlib` instead of `kotlin-runtime` on the classpath?

arturbosch:
Yep, sorry my mistake. It works!

Judging from a test case on a single KtFile, the performance penalty seems huge. We have to make sure if no classpaths are given the resolver + analyzer objects are not created at all. The gradle integration/classpath resolving would be great to test the benefits and performance. Do you plan to work further on this?

3flex:
Yep, still plan to work on this, just a bit slowly :) I have an idea about mitigating the performance impact, but need to discuss first...

You're likely familiar with Error Prone, which is implemented as a plugin for javac. Kotlin allows for the same thing, using [AnalysisHandlerExtension](https://github.com/JetBrains/kotlin/blob/v1.1.60/compiler/frontend.java/src/org/jetbrains/kotlin/resolve/jvm/extensions/AnalysisHandlerExtension.kt) to get `BindingTrace` (which has `getBindingContext()` method) and [ComponentRegistrar](https://github.com/JetBrains/kotlin/blob/v1.1.60/compiler/plugin-api/src/org/jetbrains/kotlin/compiler/plugin/ComponentRegistrar.kt) to register a compiler plugin.

There are many advantages:
- Creating the `BindingContext` is done as part of the Kotlin compilation process, so detekt wouldn't have to generate it again. detekt can just get the results of the compilation process. This process makes sense because compilation should succeed before detekt's analysis runs anyway.
- the APIs should be a little more stable (the compiler internals this PR currently takes advantage of have changed from version to version. `AnalysisHandlerExtension` and `ComponentRegistrar` have been more stable).
- Less config for detekt is required, since Kotlin compiler ensures all required files are available for compilation. Using the method from this PR doesn't enforce this (as you saw when using kotlin-runtime instead of kotlin-stdlib by mistake).
- The plugin can be used easily by [Gradle](https://kotlinlang.org/docs/reference/compiler-plugins.html#using-in-gradle), [Maven](https://kotlinlang.org/docs/reference/compiler-plugins.html#using-in-maven) and [CLI](https://kotlinlang.org/docs/reference/compiler-plugins.html#using-in-cli). This means it's simple to support all official compilation methods. I haven't used Ant but it looks like you can set CLI options in the build config file which would allow the same options as the CLI integration.
- Some lint checks are built in to the compiler, meaning they don't need to be maintained in detekt as well. For example https://github.com/arturbosch/detekt/issues/269 is unnecessary as compiler [already checks this](https://github.com/JetBrains/kotlin/blob/v1.1.60/compiler/frontend/src/org/jetbrains/kotlin/cfg/WhenChecker.kt), so if detekt runs as part of compilation process the compiler checks would be enforced at the same time.
- Bugs related to configuration and generation of `BindingContext` will be fixed upstream in Kotlin. Less for detekt to maintain.
- Better experience for developers - like Error Prone, it "hooks into your standard build, so all developers run it without thinking"

I have a POC which works, but there are some issues for this project:

- It doesn't look possible to use kotlin-compiler-embeddable. [ComponentRegistrar](https://github.com/JetBrains/kotlin/blob/v1.1.60/compiler/plugin-api/src/org/jetbrains/kotlin/compiler/plugin/ComponentRegistrar.kt) uses `com.intellij.mock.MockProject` which is shaded in the embeddable version, so it's not compatible. Would have to switch back to the regular compiler and ensure #69 doesn't occur again (maybe changes in https://github.com/arturbosch/detekt/pull/539 would help? If not, can explore [Using Shadow to Package Gradle Plugins](http://imperceptiblethoughts.com/shadow/#using_shadow_to_package_gradle_plugins)). I tried various options keeping embeddable compiler for the other dekekt modules only but couldn't get it to work, so unless someone else can suggest a workaround this will be required.
- Possible issues with versioning if the Kotlin plugin API changes. detekt would have to publish compatibility guidance to say which detekt version supports which Kotlin version if/when the API changes in future.
- A true fat jar would have to be used, otherwise it seems many `-Xplugin` command line options are required to add each of detekt's jars to the compiler's classpath. This shouldn't be difficult with Shadow.
- Some refactoring for the configuration tasks is required. Might need an interface for the required configuration options that detekt CLI tool, Gradle plugin and the compiler plugin can all hook into for configuration.
- Some refactoring for detekt's logging might be useful since the compiler has its own logging method that the plugin should hook into. This isn't essential though.
- Using checks that require `BindingContext` would only be enabled when using compiler plugin. This means detekt on command line is more limited than the compiler plugin, so lose feature parity.

I think this will be a much more robust implementation than what I originally proposed. Feedback is appreciated!

arturbosch:
All you wrote makes sense to me. That will be huge rework but will benefit detekt and the gradle plugin.

I haven't had time to dive deep into the kotlin compiler, so I have no idea how to write a compiler plugin.
I think the best way for me would be to create a new module with the compiler plugin approach and see how it works and extend this later into a new gradle plugin module. This way I can get the bigger picture on how to integrate the `core` and `api` modules and which parts need to change.

Please link me to your POC so I can start experimenting myself too. The linked issues to the gradle plugin rework should be on hold until we have a broader picture of this rework.

With the new `implementation` and `api` configurations which android developers are encouraged to use,  #69 should be fixed?

The cli feature loss could be a thing. If we manage to get this work with both maven and gradle, the changes are worth it. How this will effect `sonar-kotlin` is still unclear to me.

3flex:
I need to hack on my POC a little more, then I'll open a new PR for you to take a look at. I already implemented it as its own module, and there's very little code involved. The big changes will be to detekt's internals, which I think you should take the lead on!

3flex:
So [my plan](https://github.com/arturbosch/detekt/pull/485#issuecomment-344126107) won't work at this stage. Kotlin team did an AMA on Reddit, and [topic of compiler plugins came up](https://www.reddit.com/r/Kotlin/comments/7hoytl/kotlin_team_ama_ask_us_anything/dqsxu4n/) - I didn't realise they weren't supported at all for external plugins (though I did get it working on the CLI).

So that will have to come in future if Kotlin releases a stable compiler API. In the meantime, I'll focus my effort on this. Sorry for the long delay, but at least now I know not to sink further hours into something that would be poorly supported!

arturbosch:
Yeah, I read that today too :(. I'm scared that the compiler interns will change heavily in the future ... detekt 2.0 maybe :D

3flex:
I don't think it should be blocked if using this approach - detekt would use its embedded Kotlin compiler, so should be isolated from upstream changes in the compiler. detekt can update code that integrates with Kotlin internals when updating the Kotlin version. dokka uses this approach.

But you can leave blocked for now until I've rebased, updated and added Gradle integration.

arturbosch:
Sorry for misunderstanding. I put the blocked label just to know that it should not be reviewed for now. Ofc your work is highly valuable here!

3flex:
@arturbosch I've made some good progress in the last couple of days, but need help with something.

Right now `--input` only takes a single file or directory, but a Gradle source set can consist of multiple directories. I started making changes so `--input` would take a `;` delimited string of multiple paths and convert to `List<Path>` within detekt but there are several impacts and I'm not confident making the required changes. I'll open a new issue for it.

tomtau:
@3flex @arturbosch @schalkms Any update on this? I recently wanted to add a few rules to detekt, but realised they depended on type information.
Is there an equivalent of [SemanticDB](https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md) in Kotlin? Type and symbol resolution could be useful to different tools (e.g. dokka or ktlint), so having something like SemanticDB would save everyone's time by not directly relying kotlinc internals.

arturbosch:
hey @3flex yes please rebase and put your new changes to this branch so I can work on this after 1.0. Thanks you!

PR #487, opened 2017-10-14
Remove stale rules from config files

3flex:


arturbosch:
Thanks for the clean ups!

PR #488, opened 2017-10-14
README: update sample output from the tool

3flex:


README.md
```diff
@@ -62,22 +62,20 @@ It operates on the abstract syntax tree provided by the Kotlin compiler.
...
-      the baseline are printed in the console.
+      If a baseline xml file is passed in, only new code smells not in the 
+      baseline are printed in the console.
     --config, -c
       Path to the config file (path/to/config.yml).
     --config-resource, -cr
       Path to the config resource on detekt's classpath (path/to/config.yml).
     --create-baseline, -cb
-      Treats current analysis findings as a smell baseline for further detekt runs.
+      Treats current analysis findings as a smell baseline for further detekt
```
Mauin: Maybe we should change this to: "[...] for future detekt runs."

README.md
```diff
@@ -62,22 +62,20 @@ It operates on the abstract syntax tree provided by the Kotlin compiler.
 The CLI uses jcommander for argument parsing. Following parameters are shown when `--help` is entered:
 
 ```
-The following option is required: --input, -i
```
Mauin: `--input` is still required.

3flex: That doesn't show when running with the `--help` option. I was just copy-pasting the output from the tool verbatim. I'll add that to the text above the usage output.

README.md
3flex: Makes sense, I'll update it.

PR #489, opened 2017-10-15
If formatting rule

schalkms:


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CollabsibleIfStatements.kt
```diff
@@ -0,0 +1,35 @@
...
+		if (isNotElseIfOrElse(expression) && hasOneKtStatement(expression)) {
+			report(CodeSmell(issue, Entity.from(expression)))
+		}
+		super.visitIfExpression(expression)
+	}
+
+	private fun isNotElseIfOrElse(expression: KtIfExpression) =
+			expression.`else` == null && expression.parent !is KtContainerNodeForControlStructureBody
+
+	private fun hasOneKtStatement(expression: KtIfExpression): Boolean {
```
Mauin: Should this be named `hasOneKtIfExpression`?

schalkms: Thanks for your review! Yes.

PR #492, opened 2017-10-17
Restrict Postfix operators to ++ and -- - closes #491

### LINKED ISSUE #491: UselessPostfixExpression rule will detekt a wrong expression like “mInstance!!”

arturbosch:


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UselessPostfixExpressionSpec.kt
```diff
@@ -33,5 +33,14 @@ class UselessPostfixExpressionSpec : SubjectSpek<UselessPostfixExpression>({
 				}"""
 			assertThat(subject.lint(code)).hasSize(2)
 		}
+
+		it("should only detect ++ or -- postfix operator") {
+			val code = """
+				fun getInstance(): SwiftBrowserIdleTaskHelper {
+					return sInstance!!
```
Mauin: Can you also add a line `sInstance!!` here as well? Just to be sure that the `return` isn't excluding the postfix expression already (it shouldn't - just to make sure.)

PR #494, opened 2017-10-17
Allow to suppress StringLiteralDuplication on class level - closes #442

### LINKED ISSUE #442: Suppressing string duplication doesn't work
`@Suppress("all")` or `@Suppress("StringLiteralDuplication")` have no effect when put on class, and all string literal duplications are reported & contributed to the weight threshold anyway

arturbosch:
Closes #442

### DIFF
--- a/detekt-rules/src/test/resources/SuppressStringLiteralDuplication.kt
+++ b/detekt-rules/src/test/resources/SuppressStringLiteralDuplication.kt
@@ -1,0 +1,9 @@
+@file:Suppress("unused", "UNUSED_PARAMETER", "UnusedEquals")
+
+@Suppress("StringLiteralDuplication")
+class Duplication {
+	var s1 = "lorem"
+	fun f(s: String = "lorem") {
+		s1 == "lorem"
+	}
+}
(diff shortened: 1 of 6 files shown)


detekt-rules/src/test/resources/SuppressStringLiteralDuplication.kt
```diff
@@ -0,0 +1,9 @@
+@file:Suppress("unused", "UNUSED_PARAMETER", "UnusedEquals")
+
+@Suppress("StringLiteralDuplication")
```
vanniktech: just a question is it also valid to write `@Suppress("Detekt.StringLiteralDuplication")` if not I think we should since it makes it much clearer where this is coming from. 

Checkstyle allows this: `@SuppressWarnings("checkstyle:magicnumber")`
PMD allows this: `@SuppressWarnings("PMD.ShortClassName")`

vanniktech:
Probably worth to write `Fixes #442` in the description so it gets auto closed.

detekt-rules/src/test/resources/SuppressStringLiteralDuplication.kt
Mauin: I don't think thats possible right now. Created #497 to not loose track of the suggestion.

arturbosch:
I think `closes` in the title works too. Have both now :)

vanniktech:
Although it seems like you're having a white space too much there. `# 442`

PR #498, opened 2017-10-17
allow suppression with detekt prefixes in annotation

### LINKED ISSUE #497: Allow detekt suppression prefix
Taken from the comments in #494 

When suppressing a rule we should allow the String in the `Suppress` annotation to contain a `detekt` prefix.

Checkstyle allows this: @SuppressWarnings("checkstyle:magicnumber")
PMD allows this: @SuppressWarnings("PMD.ShortClassName")

So something along the lines of `@Suppress("detekt.SomeRuleName")` should still suppress the rule.

I'm happy to take a stab at this if people agree.

Mauin:
Resolves #497

This allows suppressing a rule by prefixing it with `detekt[.:]` (case is ignored).

This allows clients to mark more clearly which static analysis tool is reporting the issue that is suppressed.


Mauin:
Hopefully fixed the tests now and added a bunch of tests for the `isSuppressedBy` function.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Suppressible.kt
```diff
@@ -39,5 +39,8 @@ fun KtAnnotated.isSuppressedBy(id: String): Boolean {
 	val valid = listOf(id, "ALL", "all")
 	return annotationEntries.find { it.typeReferenceName.let { it == "Suppress" || it == "SuppressWarnings" } }
 			?.valueArguments
-			?.find { it.getArgumentExpression()?.text?.replace("\"", "") in valid } != null
+			?.map { it.getArgumentExpression()?.text }
+			?.map { it?.replace(Regex("(?i)detekt([.:])"), "") }
```
arturbosch: Extract both regex into variables for performance reasons.

Issue #467, opened 2017-10-06
MagicNumber: Exclude certain method calls from reporting MagicNumber


Mauin:
I think it should be possible to ignore certain methods/classes from reporting magic number violations.

`private val delay = TimeUnit.MINUTES.toMillis(3)`

In this example it's pretty clear what the `3` is. And it would only make it more verbose to define the `3` as another value. As the `TimeUnit` class is a Java class it also won't be possible to use a named parameter for the value. 
And adding suppression annotations all around the codebase for the same thing again and again seems counterintuitive.

WDYT?

schalkms:
I agree. I suggest to have a more conservative detection setting here. I would only report a `MagicNumber` if such a value is used more than once. This is similar to the `StringLiteralDuplication` rule.

arturbosch:
Agree. Maybe even combine these two rules? I think Sonar has the rule `MagicString` which is kinda what `StringLiteralDuplication` is. This rule could save all kind of literals and report them in the end?

lwasylkowski:
I wouldn't agree with having `MagicNumberDuplication` equivalent instead of current `MagicNumber` behaviour - even single magic number shouldn't be present in code if it's not named.

Maybe it would make sense to exclude magic numbers that appear anywhere in constants or declarations inside an object/companion object? I'd expect such `private val delay = TimeUnit.MINUTES.toMillis(3)` declaration in companion object, since it's effectively static.

arturbosch:
Good idea.

rock3r:
I'm hitting this issue over and over, may have to take a hard look at it sooner or later if there's no activity on this one. No promises tho 💃

Mauin:
Closing this as #501 was merged and released with RC5.

PR #501, opened 2017-10-19
Ignore companion object properties

rock3r:
This PR relates to the discussion on #467. It contains a few changes:
 * Allow to ignore magic numbers inside of companion object properties (ignored by default)
    * This allows to have the check point out magic numbers in any other property declaration that is not in a companion object
 * Allow to ignore magic numbers in constant declarations
   * It was the only non-configurable parameter in the check
 * Create `FindingAssertions.kt` that contain custom assertions for AssertJ to allow to check specific finding properties
   * This allows for more precise testing in that we can now easily assert what findings the check should bring up, instead of only relying on the number of findings which can hide errors
   * For now the best way to properly assert findings is to rely on their location, which contains the value too; this is checked with `hasLocationStrings()`

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/FindingsAssertions.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/FindingsAssertions.kt
@@ -1,0 +1,28 @@
+package io.gitlab.arturbosch.detekt.rules
+
+import io.gitlab.arturbosch.detekt.api.Finding
+import org.assertj.core.api.AbstractAssert
+import org.assertj.core.api.AbstractListAssert
+import org.assertj.core.internal.Objects
+
+fun assertThat(findings: List<Finding>) = FindingsAssert(findings)
+class FindingsAssert(actual: List<Finding>) :
+		AbstractListAssert<FindingsAssert, List<Finding>,
+				Finding, FindingAssert>(actual, FindingsAssert::class.java) {
+
+	override fun toAssert(value: Finding?, description: String?): FindingAssert =
+			FindingAssert(value).`as`(description)
+
+	fun hasLocationStrings(vararg expected: String) {
+		isNotNull
+		val actualLocationStrings = actual.map { it.entity.location.locationString }
+		areEqual(actualLocationStrings, expected)
+	}
+
+	private fun areEqual(actualLocationStrings: List<String>, expectedLocationStrings: Array<out String>) {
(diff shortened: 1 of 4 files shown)


rock3r:
Looks like the Travis build just failed because the VM died mid-way through compiling :P Can somebody restart it?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/FindingsAssertions.kt
```diff
@@ -0,0 +1,28 @@
...
+class FindingsAssert(actual: List<Finding>) :
+		AbstractListAssert<FindingsAssert, List<Finding>,
+				Finding, FindingAssert>(actual, FindingsAssert::class.java) {
+
+	override fun toAssert(value: Finding?, description: String?): FindingAssert =
+			FindingAssert(value).`as`(description)
+
+	fun hasLocationStrings(vararg expected: String) {
+		isNotNull
+		val actualLocationStrings = actual.map { it.entity.location.locationString }
```
arturbosch: Use `it.enitity.locationAsString` here please.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/FindingsAssertions.kt
```diff
@@ -0,0 +1,28 @@
+package io.gitlab.arturbosch.detekt.rules
+
+import io.gitlab.arturbosch.detekt.api.Finding
+import org.assertj.core.api.AbstractAssert
+import org.assertj.core.api.AbstractListAssert
+import org.assertj.core.internal.Objects
+
+fun assertThat(findings: List<Finding>) = FindingsAssert(findings)
+class FindingsAssert(actual: List<Finding>) :
```
arturbosch: This could be really useful in the future. Thanks.
Lets move this to the `detekt-test` module, so we can use it everywhere.

arturbosch:
Moving the `KtConstantExpression` extension function to top level below the `MagicNumber` class should be ok. Faster overview of the core logic function imo ^^
Thanks for the contribution!

arturbosch:
I will merge this and do the changes myself -> want to release RC5 today. Hope it's ok :)

rock3r:
Yeah I'm very happy to have the changes in RC5, I can re-enable the rule once it's out :) Thanks Artur :)

arturbosch:
It's out! :D

Issue #500, opened 2017-10-18
Override default detekt.yml


aochsner:
More of a question.. have searched around but couldn't find the answer.

Is there a way to override certain defaults instead of generating/copying the default-detekt.yml?  Say i want to turn off/disable a rule, and/or change the MaxLineLength value?  

This is good for 2 reasons:
1) Makes it easier to see where a project skews from the defaults
2) Assures that new rules are included as they are added/updated in future versions.

Thanks!  Open to giving it a shot if this doesn't exist and would take a day or so... Just get me pointed in the right direction

vanniktech:
Maybe failFast is for you

aochsner:
Maybe?  Where can I learn more about failFast to see if it fits?

vanniktech:
Here #179 and in the PRs

aochsner:
Thanks.  Yep that sounds about right.  Although not sure I could have found it on my own.  Seems like README needs a bit of an update?  Will try to give it a shot after work tonight...

vanniktech:
The wiki is still something that needs to be created and hence this might not be something thats available in the docs right now.

arturbosch:
Yeah we lack some documentation. Hopefully this will change in the future :).

I assume `failFast` helped you achieve what you wanted. If not feel free to open this issue again.

aochsner:
Yep failFast worked for me. Thanks!

PR #502, opened 2017-10-19
Update README for failFast option

aochsner:
#500


arturbosch:
Thanks for the documentation contribution :)!

PR #505, opened 2017-10-19
Remove double check in condition

vacxe:
Is this case u can simplify conditions because
1st check `isNotEmpty()`
2nd inside `if (isEmpty()) null else this[length - 1]` and u never get `null` in this condition

For me my variant is most readability.


arturbosch:
Your right. Thanks for the contribution :).

PR #507, opened 2017-10-21
Prepare rc5 release

arturbosch:


gradle.properties
```diff
@@ -1,7 +1,7 @@
-detektVersion=1.0.0.RC4-3
-detektGradleVersion=1.0.0.RC4-3
+detektVersion=1.0.0.RC5
+detektGradleVersion=1.0.0.RC5
 usedDetektGradleVersion=1.0.0.RC4-3
-usedDetektVersion=1.0.0.RC4-3
+usedDetektVersion=1.0.0.RC-3
```
Mauin: Is this correct? Shouldn't both of these get bumped to RC5 as well?

arturbosch: Ups, it must stay at RC4-3 until RC5 is released. Will update that to RC5 after the release.

PR #508, opened 2017-10-21
Use a test-pattern to also run detektCheck on test sources

arturbosch:


reports/failfast.yml
```diff
@@ -1,6 +1,25 @@
...
+    - 'comments'
+  exclude-rules:
+    - 'NamingRules'
+    - 'WildcardImport'
+    - 'MagicNumber'
+    - 'MaxLineLength'
+    - 'LateInitUsage'
+    - 'StringLiteralDuplication'
+    - 'SpreadOperator'
+    - 'TooManyFunctions'
```
vanniktech: should this be also in the default rule set?

reports/failfast.yml
```diff
@@ -1,6 +1,25 @@
...
+    - 'LargeClass'
```
vanniktech: same here

reports/failfast.yml
arturbosch: Hm, I think I should add these two findings just to the baseline. Big test classes are not good at all :D ?

Issue #171, opened 2017-07-07
Terminal Output customization


vanniktech:
Coming from #60

Some things that could be customized:

- having a clickable link to the file so it can be edited directly via the default editor
- only print name + package of file to save space

vanniktech:
Where should those be configured?

arturbosch:
The starting point would be `OutputFacade`. There I just use `println` for console printing. This issue sounds much wider. Maybe more the direction of `OutputFormat`

arturbosch:
I hope your not started with this issue. I'm working on detekt extensions (RuleSets, Processors, OutputFormat's, ConsoleReport's) which will be included in api module in the next RC version.

vanniktech:
Nope haven't. Knock yourself out.

vanniktech:
Closing as this should be done. If not I'll open a new one.

vanniktech:
Just picking this one up again. How can I achieve `having a clickable link to the file so it can be edited directly via the default editor` the part thats missing is the module name.

```diff
-src/androidTest/java/com/vanniktech/chessclock/screenshots/ScreenshotsTest.kt:33:3
+app/src/androidTest/java/com/vanniktech/chessclock/screenshots/ScreenshotsTest.kt:33:3
```

arturbosch:
So the project name is only missing? Which shell do you use so you can click on that link?

vanniktech:
iTerm 2 and I think the normal apple terminal can do this too.

PR #510, opened 2017-10-22
Prepend project name to findings path - #171

arturbosch:
@vanniktech please check this branch out and tell me if the file is clickable :)

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtCompiler.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtCompiler.kt
@@ -18,7 +18,9 @@ open class KtCompiler(val project: Path) {
 
 	fun compile(subPath: Path): KtFile {
 		require(subPath.isFile()) { "Given sub path should be a regular file!" }
-		val relativePath = if (project == subPath) subPath else project.relativize(subPath)
+		val relativePath =
+				if (project == subPath) subPath
+				else project.fileName.resolve(project.relativize(subPath))
 		val content = subPath.toFile().readText()
 		val lineSeparator = content.determineLineSeparator()
 		val normalizedContent = StringUtilRt.convertLineSeparators(content)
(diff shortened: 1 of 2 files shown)


vanniktech:
I found one more case where this does not work. It's when this plugin is applied in a single project that has no subprojects.

It can be reproduced [here](https://github.com/vanniktech/gradle-android-apk-size-plugin) when running `./gradlew detektCheck` and this patch file to trigger a warning:

```diff
diff --git a/src/main/kotlin/com/vanniktech/android/apk/size/ApkSizeTask.kt b/src/main/kotlin/com/vanniktech/android/apk/size/ApkSizeTask.kt
index 23a7d82..f245a90 100644
--- a/src/main/kotlin/com/vanniktech/android/apk/size/ApkSizeTask.kt
+++ b/src/main/kotlin/com/vanniktech/android/apk/size/ApkSizeTask.kt
@@ -23,7 +23,7 @@ open class ApkSizeTask : DefaultTask() {
 
     outputFile.parentFile.mkdirs()
     outputFile.createNewFile()
-    outputFile.outputStream().use { outputStream ->
+    outputFile.outputStream().use({ outputStream ->
       val appendableStream = PrintStream(outputStream)
       appendableStream.println("bytes,kilobytes,megabytes")
 
@@ -32,7 +32,7 @@ open class ApkSizeTask : DefaultTask() {
       val megaBytes = ApkSizeTools.convertBytesToMegaBytes(apkSize)
 
       appendableStream.println("$bytes,$kiloBytes,$megaBytes")
-    }
+    })
 
     failBuildMaxMethods(apkSize, extension.maxApkSize)
   }
```

The output will be:

```
UnnecessaryParentheses - [sizeApk] at gradle-android-apk-size-plugin/src/main/kotlin/com/vanniktech/android/apk/size/ApkSizeTask.kt:26:34
```

Which is not entirely correct since I'm already at the `gradle-android-apk-size-plugin` folder and hence the path should be `src/main/kotlin/com/vanniktech/android/apk/size/ApkSizeTask.kt:26:34` in order to be clickable.

vanniktech:
Why don't we always print ouf the full path in the console? That would solve all the problems. (Also executing `detektCheck` from a module level by using the installed gradle command line). Ktlint does also the same.

Issue #514, opened 2017-10-22
FunctionOnlyReturningConstant & Parcelable's describeContents


vanniktech:
In Android, we have the [`Parcelable`](https://developer.android.com/reference/android/os/Parcelable.html) interface, which specifies that you need to implement a `describeContents` method. In 99% of the cases this ends up looking like this `override fun describeContents() = 0`.

What do you guys think about whitelisting that specific method?

Mauin:
I had the same basic question in #465 and the `FunctionOnlyReturningConstant` rule got created as a result. But I'd still like a configuration param for the `FunctionOnlyReturningConstant` rule/`MagicNumber` rule for overriden functions that return a constant.

schalkms:
Yes. I think this rule should ignore overridden functions. Furthermore, it should provide a custom list.

scott-kennedy-ck:
I'm not sure this is good enough.

I have an `open fun blah() = true` that's triggering this warning.  The default value is `true`, but subclasses can override it.  Maybe there should also be an option for ignoring overridable functions?

scottkennedy:
I'm not sure this is good enough.

I have an `open fun blah() = true` that's triggering this warning.  The default value is `true`, but subclasses can override it.  Maybe there should also be an option for ignoring overridable functions?

schalkms:
@scottkennedy Would you mind creating a new issue?

PR #519, opened 2017-10-23
Enhancement #514

schalkms:


detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -334,6 +334,8 @@ style:
     active: false
   FunctionOnlyReturningConstant:
     active: false
+    ignoreOverriddenFunction: true
+    excludedFunctions: ''
```
vanniktech: should we default to `describeContents`?

Mauin: I feel like we shouldn't make any hard judgement here about defaults. Everyone can include the defaults we need and we can maybe provide a sane default for Android project. But as detekt is not just for Android projects, we should be as broad as possible with the defaults. WDYT?

vanniktech: Also a good point.

Issue #523, opened 2017-10-25
FunctionOnlyReturningConstant should have option for open functions


scottkennedy:
Per https://github.com/arturbosch/detekt/issues/514#issuecomment-339477803

I have an `open fun blah() = true` that's triggering this warning.  The default value is `true`, but subclasses can override it.  Maybe there should also be an option for ignoring overridable functions?

schalkms:
Good point! I would add a new option to the rule.

schalkms:
IMHO this can be closed.

PR #524, opened 2017-10-26
Implemented #523 - open function option FunctionOnlyReturningConstant

schalkms:


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtModifierList.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.jetbrains.kotlin.psi.KtModifierListOwner
+
+fun KtModifierListOwner.isPublicNotOverridden() =
+		isPublic() && !isOverridden()
+
+fun KtModifierListOwner.isOverridden() = hasModifier(KtTokens.OVERRIDE_KEYWORD)
+
+fun KtModifierListOwner.isOpen() = hasModifier(KtTokens.OPEN_KEYWORD)
+
+fun KtModifierListOwner.isPublic(): Boolean {
```
arturbosch: I think we could refactor this to `isPrivate` and where its used negate it?

schalkms: I am not quite sure if I understood you correctly.
!isPrivate() is also true for ‘protected’ and ‘internal’

Mauin: Agree that `isPublic` is not fully correct as a method name as it checks for much more

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantVisibilityModifierRule.kt
```diff
@@ -24,8 +25,7 @@ class RedundantVisibilityModifierRule(config: Config = Config.empty) : Rule(conf
 	private val classVisitor = ClassVisitor()
 	private val childrenVisitor = ChildrenVisitor()
 
-	private fun KtModifierListOwner.isExplicitlyPublicNotOverridden() = isExplicitlyPublic()
-			&& !this.hasModifier(KtTokens.OVERRIDE_KEYWORD)
+	private fun KtModifierListOwner.isExplicitlyPublicNotOverridden() = isExplicitlyPublic() && !isOverridden()
 
 	private fun KtModifierListOwner.isExplicitlyPublic() = this.hasModifier(KtTokens.PUBLIC_KEYWORD)
```
Mauin: I'd move this to the new file with extensions as well. This is the true `isPublic` method

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtModifierList.kt
arturbosch: I mean that this method effectively just tests if the modifier is not private. I think it is just used once and if we change the code on that line to !isPrivate, we can delete this method.

schalkms: `isPublic()` is used 6 times. 
The method checks if the modifier is not
- private
- protected
- internal

PR #535, opened 2017-11-07
UnnecessaryAbstractClass takes primary constructor vals into account

geralt-encore:
https://github.com/arturbosch/detekt/issues/534
Let me know if I can improve my code if it is needed since I am not familiar enough with the codebase yet


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAbstractClass.kt
```diff
@@ -44,9 +44,10 @@ class UnnecessaryAbstractClass(config: Config = Config.empty) : Rule(config) {
 						Entity.from(klass),
 						"An abstract class without an abstract member can be refactored to a concrete class."))
 			} else if (indexOfFirstAbstractMember == 0 && hasNoConcreteMemberLeft()) {
-				report(CodeSmell(issue,
-						Entity.from(klass),
-						"An abstract class without a concrete member can be refactored to an interface."))
+				if (!hasPrimaryConstructorWithConcreteProperties())
```
Mauin: This could be combined with the `else if` in the previous line. Just combine all three condition checks into a function along the lines of: `isAbstractClassThatCouldBeInterface` (or a better name that you can think of 🙃)

PR #536, opened 2017-11-08
run detekt for all defined profiles and allow different names

Mauin:
I was looking at `detekt-gradle-plugin` a bit and realized two weird things. 

1. When I named my profile anything other than `"main"` it seemed not to work correctly.

This seems to have been caused by the `var profile` in `DetektExtension` that was initialized to `"main"` but would never be overwritten with the actual profile name. So detekt would run the `_fallback_` profile which only runs the defaults but considers none of the profiles configuration options.

2. When defining multiple profiles only one would be run by detekt.

While we collect all profiles defined in the `detekt { ... }` closure of the `build.gradle` file in a `profiles` variable we never did anything with them. By adding a getter and looping through all profiles for the `DetektCheckTask`, `DetektCreateBaselineTask` and `DetektMigrateTask` all profiles should now be executed.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCheckTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCheckTask.kt
@@ -23,10 +23,12 @@ open class DetektCheckTask : DefaultTask() {
 		project.buildscript.dependencies.add(configuration.name, DefaultExternalModuleDependency(
 				"io.gitlab.arturbosch.detekt", "detekt-cli", detektExtension.version))
 
-		project.javaexec {
-			it.main = "io.gitlab.arturbosch.detekt.cli.Main"
-			it.classpath = configuration
-			it.args(detektExtension.profileArgumentsOrDefault(project))
+		detektExtension.getProfiles().forEach { profile ->
+			project.javaexec {
+				it.main = "io.gitlab.arturbosch.detekt.cli.Main"
+				it.classpath = configuration
+				it.args(detektExtension.profileArgumentsOrDefault(project, profile))
+			}
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


arturbosch:
Yep that behaviour was as planned :).
If no profile is declared or selected with `profile = xxx`, the `_fallback_` profile is used. The default profile is aimed to be the `main` profile if declared. If you have many profiles declared, you can pick any by specifying inside the `profile = xxx` variable or giving the `-Ddetekt.profile=xxx`- property to gradle.
If you have a profile `test` declared, which does not override all detekt extensions, the not overridden properties get inherited from the `main` profile :)

Kinda difficult to know and not documented at all (?!) :D

Running all profiles seems like a good idea, but I do not think all should be run everytime. Maybe an additional flag or task?

Mauin:
Huh, didn't know that was behaving as planned. Thanks for explaining the rationale behind how it should be used. But it feels a bit confusing to me.

I personally would not like editing my `build.gradle` files to change the `profile = xxx` values for every run of detekt. And somehow passing arguments to the gradle command is a nice option but also doesn't feel like something I'd like to do on every run.

If I set up a `main` and a `test` profile I'd personally expect them both to run. And if my `main` profile is called `"default"` it should still run that profile and not the `_fallback_` profile (which would be incredibly hard to even realize in many cases). 
I'd think that whenever I can enter a String in a gradle configuration I should be able to use whatever string I like. Forcing it to be "main" feels a bit weird.
Same goes for multiple profiles. If I explicitly define multiple profiles, why not run all of them? Maybe we could even define tasks for each profile.
Let's take `detektCheck` as an example if I define the above mentioned `main` and `test` profiles:
- `detektCheck` would run both (all profiles that are defined)
- `detektCheckMain` would run the `main` profile
- `detektCheckTest` would run the `test` profile

We can move the discussion to a new Issue as well and close this PR for now.

arturbosch:
Yes, please open an new issue and close this one. It would be good to discuss the usage of profiles. I introduced them when we had no option to exclude rules for test sources. My usecase for them was also to have separate config files for different project modules, eg. some rules are ok for gui code but not for library code.

Mauin:
I opened #540 to continue the discussion.

Issue #527, opened 2017-10-26
Separating positive and negative test cases for rules


schalkms:
At the moment the tests check the violation count.
However, there can be a case where the sum is correct and the findings are wrong.
Separating positive and negative test cases could help to minimize this problem.

arturbosch:
Which test cases do you have in mind? I'm not a fan of refactoring just because of refactoring. If the logic gets changed, we can change some test cases?

schalkms:
Consider the following test file example for checking the naming convention.
```kotlin
class MyClass {}//valid
class myclass {} // report
class myClass {} // report
```
Detekt's current test case checks the violation count. However, detekt does not check if the wrong **MyClass** classes are reported. There could be case where the right one gets reported.

arturbosch:
Ah now I understand. Yes, that makes sense and should be the preferred way.

schalkms:
Closed with 0d869e3

PR #541, opened 2017-11-12
Testcase refactoring - #527

schalkms:
In my opinion our test-cases are much clearer now. Furthermore, I think the test-cases would need comments on why something is reported.

Some issues were also found (please look at the history)
To summarize it
- fixed one bug where the  isHashCode() function does not check if this method is overridden
- commit #535 introduced another bug
- some duplicated tests were removed


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DuplicateCaseInWhenSpec.kt
```diff
@@ -1,14 +1,41 @@
...
+					when (1) {
+						1 -> println()
+						1 -> println()
+						else -> println()
+					}
+				}"""
+			assertThat(subject.lint(code)).hasSize(1)
+		}
+
+		it("should report duplicated label in when") {
```
arturbosch: "...should not report..."

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/EqualsAlwaysReturnsTrueOrFalse.kt
```diff
@@ -22,12 +21,6 @@ class EqualsAlwaysReturnsTrueOrFalse(config: Config = Config.empty) : Rule(confi
 					"Consider a good default implementation. " +
 					"For example this == other")
 
-	override fun visitClass(klass: KtClass) {
-		if (!klass.isInterface()) {
```
arturbosch: Ok, I wanted to have a testcase for interfaces now, but the compiler says interfaces cannot override methods from Any...

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAbstractClass.kt
```diff
@@ -43,7 +43,8 @@ class UnnecessaryAbstractClass(config: Config = Config.empty) : Rule(config) {
 				report(CodeSmell(issue,
 						Entity.from(klass),
 						"An abstract class without an abstract member can be refactored to a concrete class."))
-			} else if (isAbstractClassWithoutConcreteMembers(indexOfFirstAbstractMember)) {
+			}
+			if (indexOfFirstAbstractMember == 0 && hasNoConcreteMemberLeft() && hasNoConstructorParameter(klass)) {
```
arturbosch: I liked this refactored to `isAbstractClassWithoutConcreteMembers` for readability.

PR #546, opened 2017-11-14
Update to Kotlin 1.1.60

3flex:
This stood out:
> KT-20462 Don't create an array copy for '*(...)'

Does that mean [SpreadOperator](https://github.com/arturbosch/detekt/blob/master/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperator.kt) check isn't required anymore? detekt's description says "Using a spread operator causes a full copy of the array to be created before calling a method has a very high performance penalty."

Seems [KT-20462](https://youtrack.jetbrains.com/issue/KT-20462) fixes exactly that performance problem, so the detekt rule is redundant?


Mauin:
The benchmarks that led to the creation of the `SpreadOperator` rule are available here: https://github.com/renatoathaydes/kotlin-hidden-costs-benchmark

We can re-run the benchmarks with the new Kotlin version to verify that the performance is now better.

3flex:
I ran the benchmark, strangely I don't see any difference between running with Kotlin 1.1.3 and Kotlin 1.1.60. Probably made some mistake while I ran the benchmarks, or I'm not understanding the release notes. So I won't open a new PR/issue for any `SpreadOperator` changes.

arturbosch:
Oh, I did not read your last comment ... I opened an issue .. for the future! As I read [KT-20462](https://youtrack.jetbrains.com/issue/KT-20462) so that only using a newly created array with constants should not trigger a new array by the compiler not that having an array passed to a function and just using the spread operator on it.

PR #548, opened 2017-11-14
improve message of PackageDeclarationStyle

Mauin:
Was investigating an issue in a codebase today where a `PackageDeclarationStyle` issue was reported.

From the current issue description it was not really clear what was wrong in my code in the first place. This gives both cases this rule checks a clearer `CodeSmell#message`.

Should we rename this rule? Having it called `PackageDeclarationStyle` but it actually also checks the linebreak in between imports and the class declaration is a bit confusing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -578,7 +578,7 @@ style:
   ForbiddenImport:
     active: false
     imports: ''
-  PackageDeclarationStyle:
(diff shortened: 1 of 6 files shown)


schalkms:
I am sure that we can find a better name.
`SpacingBetweenPackageAndImports`

PR #549, opened 2017-11-14
Use newer extension syntax in example for Kotlin DSL and add in expli…

mkobit:
…citly needed cast for profile configuration

The `Action {}` cast is unforunately needed right now for plugins written in Kotlin.

issue https://github.com/gradle/kotlin-dsl/issues/522
Kotlin issue https://youtrack.jetbrains.com/issue/KT-20635

----

Ran into this in latest Gradle 4.3

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ detekt {
 ```
 
 ##### <a name="gradlekotlin">Configuration when using kotlin dsl</a>
-For gradle version >= 3.5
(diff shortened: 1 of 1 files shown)


arturbosch:
I need badly a gradle script kotlin project so I can stay up-to-date with their changes ... In the long term maintaining both kotlin and groovy gradle scripts in the way to go. So we have both ways tested ...

PR #557, opened 2017-11-17
Run Travis builds in containers

3flex:
Travis claims dramatically reduced time between commit push to GitHub and the start of a job on Travis CI with sudo disabled.


3flex:
My Travis knowledge is out of date - this now appears to be the default, and this isn't required (builds are already running in containers).

PR #562, opened 2017-11-19
remove SubRule and replace last usage with standard Rule

Mauin:
We only had one usage of `SubRule` left in the codebase and it's fairly straightforward to just replace it with a standard `Rule` so I don't see a huge reason to keep it around.


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLengthRuleSpec.kt
```diff
@@ -183,12 +171,11 @@ class MaxLineLengthRuleSpec : Spek({
...
-				Assertions.assertThat(it).hasSize(1)
-				val findingSource = it[0].location.source
-				Assertions.assertThat(findingSource.line).isEqualTo(6)
-				Assertions.assertThat(findingSource.column).isEqualTo(109)
-			}
+			rule.visit(fileContent)
+			Assertions.assertThat(rule.findings).hasSize(1)
+			val findingSource = rule.findings[0].location.source
+			Assertions.assertThat(findingSource.line).isEqualTo(6)
+			Assertions.assertThat(findingSource.column).isEqualTo(109)
```
schalkms: `Assertions` can be removed

Mauin: Good spot 👍

PR #565, opened 2017-11-21
housekeeping_rules

schalkms:


Mauin:
@schalkms iirc this was because `{something, _ -> blablabla()}` would report the `_` as an invalid named variable, even though it's okay. So we ignore underscores in the `VariableNaming` rule.

schalkms:
@Mauin 
You are right. The documentation mentions it in **Underscore for unused variables (since 1.1)**
https://kotlinlang.org/docs/reference/lambdas.html#underscore-for-unused-variables-since-11

schalkms:
I will implement a test quickly.

schalkms:
Travis-CI
```
oci runtime error: exec failed: container_linux.go:265: starting container process caused "could not create session key: disk quota exceeded"
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/NamingConventionViolationSpec.kt
```diff
@@ -53,6 +53,16 @@ class NamingConventionTest {
             """
 		)).hasSize(0)
 	}
+	@Test
+	fun underscoreForUnusedVariableInClosure() {
+		assertThat(NamingRules().lint(
+				"""
+			fun f() {
+				emptyMap<String, String>().forEach { _, v -> println(v) }
```
Mauin: I guess we do _not_ want to report this. The underscore should be okay. so the findings should be empty, right?

PR #569, opened 2017-11-23
Rework naming rules to match intellij inspections

arturbosch:


detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -294,10 +294,16 @@ style:
...
     active: true
-    variablePattern: '^(_)?[a-z$][a-zA-Z$0-9]*$'
-  ConstantNaming:
+    variablePattern: '[a-z][A-Za-z0-9]*'
+    privateVariablePattern: '(_)?[a-z][A-Za-z0-9]*'
+  ObjectPropertyNaming:
     active: true
-    constantPattern: '^([A-Z_]*|serialVersionUID)$'
+    propertyPattern: '[A-Za-z][_A-Za-z0-9]*'
+  TopLevelPropertyNaming:
```
schalkms: I like the granularity for the naming rules.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/naming/ObjectPropertyNaming.kt
```diff
@@ -10,27 +10,27 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.KtVariableDeclaration
 
-class ConstantNaming(config: Config = Config.empty) : Rule(config) {
+class ObjectPropertyNaming(config: Config = Config.empty) : Rule(config) {
 
 	override val issue = Issue(javaClass.simpleName,
 			Severity.Style,
-			"Constants names should follow the naming convention set in the projects configuraiton.",
+			"Constants inside objects should follow the naming convention set in the projects configuration.",
```
schalkms: It might be a bit unclear what the convention set looks like for the reader of this message. What do you think?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/naming/TopLevelPropertyNaming.kt
```diff
@@ -0,0 +1,71 @@
...
+	private fun handleProperty(property: KtProperty) {
+		if (property.isPrivate()) {
+			if (!property.identifierName().matches(privatePropertyPattern)) {
+				report(CodeSmell(
+						issue,
+						Entity.from(property),
+						message = "Private top level property names should match the pattern: $propertyPattern"))
+			}
+		} else {
+			if (!property.identifierName().matches(propertyPattern)) {
```
schalkms: Should detekt's rule configuration really differentiate between private and non-private top level properties?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/naming/NamingConventionCustomPatternTest.kt
```diff
@@ -0,0 +1,93 @@
...
+import io.gitlab.arturbosch.detekt.rules.style.naming.FunctionNaming
+import io.gitlab.arturbosch.detekt.rules.style.naming.NamingRules
+import io.gitlab.arturbosch.detekt.rules.style.naming.PackageNaming
+import io.gitlab.arturbosch.detekt.rules.style.naming.TopLevelPropertyNaming
+import io.gitlab.arturbosch.detekt.rules.style.naming.VariableNaming
+import io.gitlab.arturbosch.detekt.test.lint
+import org.assertj.core.api.Assertions
+import org.junit.jupiter.api.Test
+
+class NamingConventionCustomPatternTest {
```
schalkms: Splitting up the these test cases is a good idea!

detekt-rules/src/test/resources/cases/NamingConventions.kt
```diff
@@ -1,16 +1,18 @@
...
 
-// both valid
 val variable = 5
-val _variable = 5
 
-//invalid
+// valid: underscore only allowed for private vars
+private val _variable = 5
+val _variable2 = 5 // invalid
+
```
schalkms: I think all these test cases should be in separate test cases and not in one big file. This could lead to problems discussed in #527

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/naming/ObjectPropertyNaming.kt
arturbosch: Sorry, I don't get what you mean :)?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/naming/TopLevelPropertyNaming.kt
arturbosch: IntelliJ does, so I also implemented this. If having only top level declarations inside a file, it could be handy to name things with `_` ^^

detekt-rules/src/test/resources/cases/NamingConventions.kt
arturbosch: I think after this PR this file and `NamingConventionSpec` is not needed anymore, as all cases/rules have their own spec class then ^^

PR #570, opened 2017-11-23
UnsafeCast: Fix in SpacingBetweenPackageImports, turn on rule, add present to baseline

arturbosch:


arturbosch:
Absolutely, I think together with Naming rules rework and maybe the Java9 thing (optional).

schalkms:
I agree with Artur.

PR #571, opened 2017-11-24
rules-test-rc5

schalkms:
If multiple `it` test cases in the same `given` or `describe` function have the same name only one of them gets executed.
I think detekt should have a rule for this problem.


schalkms:
@arturbosch Which test case do you mean?

arturbosch:
I meant the rule from #572 which now is unclear if we will implement it.

PR #574, opened 2017-11-28
Purge NewLineAtEndOfFile check.

vanniktech:
Since it was decided that we should focus more on non-formatting related things it would be best to also get rid of this check. ktlint is already handling this successfully


Mauin:
I'm not convinced we need to remove all formatting related things from detekt right away. Projects not depending on ktling + detekt would lose coverage. Projects that have both ktlint and detekt enabled could disable this rule in the detekt config.

schalkms:
I agree with @Mauin

vanniktech:
I can also understand this, however we should fail if a file has more than one new line at the end of the file. This is currently not covered.

arturbosch:
We will remove that rule when we have a ktlint wrapper. This PR will come handy later :).

vanniktech:
That sounds like a good trade-off.

vanniktech:
What should we do with this PR? It's gotten super stale.

vanniktech:
I'm closing this since it has too many conflicts and it's easier to restart this effort.

Issue #566, opened 2017-11-22
ExceptionInInitializerError with 1.0.0.RC5-5


juergenzimmermann:
After upgrading from 1.0.0.RC5-4 to 1.0.0.RC5-5 I'm getting the following stacktrace when running "gradle detektCheck". I'm using Kotlin 1.2.0-rc-39 and Java 9.0.1.

```
> Task :detektCheck FAILED
Exception in thread "main" java.lang.ExceptionInInitializerError
        at io.gitlab.arturbosch.detekt.core.KtCompiler.<init>(KtCompiler.kt:17)
        at io.gitlab.arturbosch.detekt.core.KtTreeCompiler.<init>(KtTreeCompiler.kt:15)
        at io.gitlab.arturbosch.detekt.core.KtTreeCompiler$Companion.instance(KtTreeCompiler.kt:23)
        at io.gitlab.arturbosch.detekt.core.Detektor.run$default(Detektor.kt:25)
        at io.gitlab.arturbosch.detekt.cli.Runner.execute(Runner.kt:20)
        at io.gitlab.arturbosch.detekt.cli.Main.main(Main.kt:19)
[...]

PR #575, opened 2017-11-28
update kotlin to v1.2.0

Mauin:
https://github.com/JetBrains/kotlin/releases/tag/v1.2.0


arturbosch:
From 1.1 to 1.2 there are many changes inside the compiler.
I don't feel safe to ship it with RC5-6, will release this version first.

Mauin:
You're right, I was hoping this would help solve the issue reported in #566. But trying it locally that doesn't seem to be the case.

arturbosch:
Merged manually.

PR #578, opened 2017-11-29
README: Update description of the 'output' parameter for the Gradle plugin

### LINKED ISSUE #577: Output is a folder and not a file anymore
According to the gradle docs:
> output = "$project.projectDir/reports/detekt.xml" // If present, prints all findings into that file.

However it seems like this specifies a folder now; in which a plain text and XML report are saved.

pocmo:
Fixes #577


arturbosch:
Hey, thanks for the correction!

PR #579, opened 2017-11-29
detekt-generator: content section

schalkms:


schalkms:
@Mauin I completely forgot about that. I'll add it.

PR #581, opened 2017-11-30
Prepare RC5-6 release

arturbosch:


MIGRATION_GUIDE.md
```diff
@@ -1,5 +1,28 @@
 # Migration Guide
 
+### RC5-5 -> RC5-6
+
+- fixed a critical bug in `SpacingBetweenPackageImports`, please update if you use this rule.
```
Mauin: What was the bug? What about the issue with `UnsafeCast`?

arturbosch: Yep it was the unsafe cast inside that rule :)

Mauin: Ah yes, you're right. Now I remember 👍

PR #584, opened 2017-11-30
detekt-generator code examples

schalkms:
This PR implements a prototype for compliant and non-compliant code-examples in the documentation generator as discussed in #343.

Do you have an idea where to put the text files with the code examples? @Mauin @arturbosch


detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/CodeExampleCollector.kt
```diff
@@ -0,0 +1,22 @@
+package io.gitlab.arturbosch.detekt.generator.collection
+
+import java.io.File
+
+class CodeExampleCollector {
```
Mauin: this should probably extend `Collector` so that it can run together with all the other collectors.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/DetektCollector.kt
```diff
@@ -23,11 +25,13 @@ class DetektCollector : Collector<RuleSetPage> {
 		val rules = ruleCollector.items
 		val multiRules = multiRuleCollector.items.associateBy({ it.name }, { it.rules })
 		val ruleSets = ruleSetProviderCollector.items
+		val codeExamples = codeExampleCollector.collect()
```
Mauin: This should be part of the `collectors` List and then it will run as part of the `visit` method. `codeExampleCollector.items` will give you the collected Set.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/RuleSetPage.kt
```diff
@@ -1,9 +1,15 @@
...
 import io.gitlab.arturbosch.detekt.generator.collection.Rule
 import io.gitlab.arturbosch.detekt.generator.collection.RuleSetProvider
 
 data class RuleSetPage(
 		val ruleSet: RuleSetProvider,
-		val rules: List<Rule>
+		val rules: List<RuleCode>
+)
+
+data class RuleCode(
```
Mauin: Why don't we add the `codeExample` property to `Rule`?

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/out/Markdown.kt
```diff
@@ -42,6 +42,8 @@ inline fun MarkdownContent.referenceToHeading(reference: () -> String) =
 		"[${reference()}](#${reference().replace(' ', '-')})"
 
 inline fun MarkdownContent.code(code: () -> String) = "`${code()}`"
+inline fun MarkdownContent.kotlinCode(code: () -> String) = "```kotlin\n${code()}\n```"
```
Mauin: I guess Markdown would call this a code block?

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/CodeExampleCollector.kt
```diff
@@ -0,0 +1,22 @@
...
+
+	fun collect(): Set<CodeExample> {
+		return emptySet()
+	}
+
+	private fun collectExamples(): Set<CodeExample> {
```
Mauin: Where would the examples for compliant/non-compliant code live? Also in the KDoc of a rule? Then we could use a Visitor similar to the `RuleCollector`. Or should these be completely separate files? In that case it would be hard to link them and to remember to add/update them everytime rules get added/changed because they are so out of sight.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/CodeExampleCollector.kt
schalkms: Yeah, maybe. I don't need the `visit()` method.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/CodeExampleCollector.kt
schalkms: No. As I mentioned in the initial comment:
> Do you have an idea where to put the text files with the code examples?

I think detekt should have them in separate text files (possibly with the same name as the rule itself).

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/DetektCollector.kt
schalkms: The `CodeExampleCollector` does not need the `visit()` method.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/out/Markdown.kt
schalkms: Yes, thanks.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/RuleSetPage.kt
schalkms: The `CodeExample` needs to be associated with the corresponding rule.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/CodeExampleCollector.kt
Mauin: I fear that this "hides" them a little bit and makes very obfuscated where the samples are coming from. Also for Contributors it would not be very obvious that these samples even exist.

schalkms: I agree. However, I don't like it in the comment of the rule implementation either.

arturbosch: Didn't we say to put the code examples into the kdoc together with the description?

schalkms: I wouldn’t do that.
I think it’s better to move them in a separate file and implement a check/test if there’s a rule for the given code snippet.

arturbosch: Yeah but why? Java has also examples of the use case inside the docs. I don't get why you don't like the doc option because you do not provide arguments :/ ?

arturbosch: Having everything in one place makes sense for parsing and for looking up the source code?

schalkms: I think the rule implementation gets bloated. The code examples don’t provide any additional value to the implementation except for the generated documentation.

schalkms: @arturbosch that’s the advantage of this solution

Mauin: Hiding code examples makes the codebase harder understand and fully comprehend imo. It will be really easy to forget adding/keeping those examples up-to-date when they are so out of sight. Having everything in one place in the KDoc of the rule class at least makes it a little more visible and more likely that contributors and the people reviewing PRs remember that the examples should be updated.

schalkms: This could be avoided by a simple test as I mentioned.
Anyways, the votes are against this solution. I can of course live with that. I’ll adapt this prototype soon.

schalkms: Thanks for the open discussion and the review!

detekt-generator/documentation/potential-bugs.md
```diff
@@ -43,6 +43,32 @@ All hash-based collections depend on objects meeting the equals-contract. Two eq
...
+        return super.equals(other)
+    }
+
+    override fun hashCode(): Int {
+        return super.hashCode()
+    }
+}
+```
+
+#### Compliant Code:
```
arturbosch: I think the Compliant and Noncompliant code snippets are mixed

schalkms: Yep. I will update it.

Issue #496, opened 2017-10-17
Story: Improve documentation of detekt


arturbosch:
This is the story ticket to sum up all documentation improvements which need to be done for a GA release:

- [x] Kdoc for all rules
    - [x] complexity rule set
    - [x] style rule set
    - [x] potential bugs rule set
    - [x] comments rule set
    - [x] performance rule set
    - [x] exceptions rule set
    - [x] empty-blocks rule set
- [x] generate default config based on Kdoc and sources #189
- [x] github pages website with rule browsing, generate markdown page for each rule based on Kdoc
- [x] example extension project should have `build.gradle` and `pom.xml` #438
- [X] ~~split readme into wiki pages #192, #17~~
- [x] populate site with useful infos and how-tos from readme
- [x] finalize rule names #109

Mauin:
Some KDoc for the `performance` ruleset was done in #563

arturbosch:
Every point on the list was done. If something is unclear or not documented out users will tell us ^^.

PR #588, opened 2017-12-03
Add documentation in bugs rules

Mauin:
This adds some basic documentation in all rules of the `PotentialBugProvider`.  
I also added some more `@author` tags in the KDoc and marked the ones as `@active` that are currently enabled in the `default-detekt-config.yml`.


Mauin:
This touches on #496

PR #589, opened 2017-12-03
Add documentation to generate default config

### LINKED ISSUE #189: Generate default-detekt-config.yml according to rules
It's quite tedious to go and change the `default-detekt-config.yml` every time a new rule / property is added. I think it'd be possible to generate it. The only problem that I'm seeing is maybe the fact that both are living in separate modules.

Mauin:
This PR adds all KDoc necessary to generate the `default-detekt-config.yml`. The generated config on this branch should be equal to the config we manually maintain currently.

I didn't update all the other documentation for the rules, but only added those parameters needed to get this working.

Also had to make some changes to allow for lists as default parameters (used in `TooGenericExceptionCaught` and `TooGenericExceptionThrown`). 

Resolves #189
Touches on #496

**Note:** This branch also contains the changes from #588

### DIFF
--- a/detekt-generator/documentation/default-detekt-config.yml
+++ b/detekt-generator/documentation/default-detekt-config.yml
@@ -56,78 +56,116 @@ code-smell:
 comments:
   active: true
   CommentOverPrivateFunction:
-    active: false
+    active: true
   CommentOverPrivateProperty:
-    active: false
+    active: true
   UndocumentedPublicClass:
     active: false
+    searchInNestedClass: true
+    searchInInnerClass: true
+    searchInInnerObject: true
+    searchInInnerInterface: true
   UndocumentedPublicFunction:
     active: false
 
 complexity:
   active: true
   LongParameterList:
-    active: false
+    active: true
+    threshold: 5
   LongMethod:
-    active: false
+    active: true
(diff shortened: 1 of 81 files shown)


detekt-generator/documentation/default-detekt-config.yml
```diff
@@ -56,78 +56,116 @@ code-smell:
 comments:
   active: true
   CommentOverPrivateFunction:
-    active: false
+    active: true
   CommentOverPrivateProperty:
-    active: false
```
arturbosch: Oh I think having no comments over private functions/properties is my personal style and should not be active by default? Should I open an issue for discussions?

Mauin: I feel the same but of course it's open for discussion if this should be the default. For this PR I just took the same values that are currently set in the current `default-detekt-config.yml`

arturbosch: I will just provide a PR to turn them off. I often saw people turning off this rule or ask a question why it is that way ^^

schalkms: I agree. This should be disabled by default.

PR #590, opened 2017-12-03
Automatically run detekt-generator on detekt-rules build task

Mauin:
The Jar generated from the `detekt-generator` module now has an `--input` and `--output` argument to define both directories.

Adds `generateDocumentation` task in the `detekt-generator` module. This builds and executes the generator jar to generate the latest documentation.

I also added a rule that every time the `detekt-rules` module is built the `generateDocumentation` task will be executed to update the documentation.

### DIFF
--- a/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/Main.kt
+++ b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/Main.kt
@@ -2,22 +2,40 @@
 
 package io.gitlab.arturbosch.detekt.generator
 
-import io.gitlab.arturbosch.detekt.core.exists
-import java.nio.file.Path
-import java.nio.file.Paths
+import io.gitlab.arturbosch.detekt.cli.failWithErrorMessages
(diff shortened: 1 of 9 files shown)


detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/Main.kt
```diff
@@ -2,22 +2,40 @@
...
+}
+
+private fun validateCli(arguments: Args): List<String> {
+	val violations = ArrayList<String>()
+	with(arguments) {
+		if (Files.exists(outputPath) && outputPath.isFile()) {
+			violations += "Output file must be a directory."
+		}
+
+		if (!Files.exists(inputPath) || outputPath.isFile()) {
```
arturbosch: This must be `inputPath.isFile()` ?

arturbosch:
Looks good!

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/Main.kt
Mauin: Good catch!

PR #593, opened 2017-12-05
Replace default-detekt-config.yml with generated one

Mauin:
This is a follow up of #590 and contains the same changes atm. Let's merge #590 first. Once that is merged this one should be good to go.

This adds an additional output path to the generator and points it to the `detekt-cli/src/main/resources` directory so that the generated config replaces the currently manually maintained one.

As the generated config is sorted by the order of rules how they are specified in the RuleSetProviders the config will get reformatted in this PR.

Due to #590 the config will automatically get updated every time the `detekt-rules` module gets built.

Link to #496


Mauin:
Added a TODO to the description. I still need to rebase this once we have merged #590

arturbosch:
#590 is merged!

Mauin:
Rebased and updated the config with the generated one 🎉

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/Args.kt
```diff
@@ -29,8 +34,11 @@ class Args {
 	val inputPath: Path
 		get() = input ?: throw IllegalStateException("Input path was not initialized by jcommander!")
 
-	val outputPath: Path
-		get() = output ?: throw IllegalStateException("Output path was not initialized by jcommander!")
+	val documentationPath: Path
+		get() = documentation ?: throw IllegalStateException("Output path was not initialized by jcommander!")
```
arturbosch: You forgot to update the message of the exception

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/Args.kt
```diff
@@ -29,8 +34,11 @@ class Args {
...
+
+	val configPath: Path
+		get() = config ?: throw IllegalStateException("Output path was not initialized by jcommander!")
```
arturbosch: Here too.

arturbosch:
Thanks for the change! Great work!

PR #598, opened 2017-12-09
Add SpreadOperator to exclude-rules in test-pattern.

vanniktech:
I don't think the performance cost is really an issue when running tests and hence we should disable it by default.

Is there anything else where I would need to add this? I lost track of where things belong.


schalkms:
Are there any performance measures available? If the spread operator has no significant performance impact I am okay with that.

arturbosch:
@schalkms [here](https://github.com/arturbosch/detekt/issues/554) is the issue and a linked benchmark.
I think it is ok for test cases to exclude this rule on default so 👍 . @Mauin waiting for your approval :)

PR #599, opened 2017-12-10
ReturnCount augmented to ignore specified function names

patrickpilch:
Hi, thanks for the neat library!

I wanted to contribute due to me running into situations where the `ReturnCount` rule would flag some auto generated code like the following:
```kotlin
override fun equals(other: Any?): Boolean {
        if (this === other) return true
        if (javaClass != other?.javaClass) return false

        other as TestClass

        if (someList != other.someList) return false

        return true
    }
```
I thought that this would be a common occurrence, and I still wanted to use the `ReturnCount` rule without having to sprinkle suppression annotations all over, so I thought a function name whitelist would be useful.

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -237,6 +237,7 @@ style:
   ReturnCount:
     active: true
     max: 2
+    ignoredFunctionNames: "equals"
   ThrowsCount:
     active: true
     max: 2
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -237,7 +237,7 @@ style:
   ReturnCount:
     active: true
     max: 2
-    ignoredFunctionNames: "equals"
+    excludedFunctions: "equals"
   ThrowsCount:
     active: true
     max: 2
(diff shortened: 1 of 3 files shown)


detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -237,6 +237,7 @@ style:
   ReturnCount:
     active: true
     max: 2
+    ignoredFunctionNames: "equals"
```
vanniktech: We should probably name it `excludedFunctions` since other rules use the same name

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -19,26 +20,35 @@ import org.jetbrains.kotlin.psi.KtReturnExpression
  *
  * @configuration max - define the maximum number of return statements allowed per function
  * (default: 2)
+ * @configuration ignoredFunctionNames - define functions to be ignored by this check
```
Mauin: This needs to be updated to `excludedFunctions`

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -19,26 +20,35 @@ import org.jetbrains.kotlin.psi.KtReturnExpression
...
+ * (default: none)
```
Mauin: The default should be `(default: "equals")`. The `default-detekt-config.yml` will then be generated automatically on a build of the project.

PR #604, opened 2017-12-11
update documentation after some merges to master

Mauin:
Travis build broke on master because some documentation was missing.


vanniktech:
It might be worth adding the `always up to date` branch feature?

PR #608, opened 2017-12-13
False positive in ExceptionRaisedInUnexpectedLocation

aballano:
Fixes false positives when function names contains some of the keywords for this inspection, such as toString, hashCode, equals or finalize.

An usecase for this would be if an user would like to implement a function like:

`assertUsersNotEquals(...)`

Since it contains "equals" in the name, detekt would mark it as a violation.


aballano:
@arturbosch @vanniktech Thanks for the quick response! 🎉

PR #610, opened 2017-12-13
compliant and non-compliant code examples documentation

schalkms:
I just started with it. (work in progess)

### DIFF
--- a/detekt-generator/documentation/potential-bugs.md
+++ b/detekt-generator/documentation/potential-bugs.md
@@ -27,6 +27,25 @@ Flags duplicate case statements in when expressions.
 If a when expression contains the same case statement multiple times they should be merged. Otherwise it might be
 easy to miss one of the cases when reading the code, leading to unwanted side effects.
 
+#### Noncompliant Code:
+
+```kotlin
+when (i) {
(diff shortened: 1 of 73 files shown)


schalkms:
@vanniktech I think we already had a conversation about this topic somewhere here. In my opinion it's not worth the effort since you need to resolve and annotate the test cases which should be included in the documentation. You also need to build a generator and so on and so forth.
I also think that this information is not likely to change.

detekt-generator/documentation/potential-bugs.md
```diff
@@ -27,6 +27,25 @@ Flags duplicate case statements in when expressions.
 If a when expression contains the same case statement multiple times they should be merged. Otherwise it might be
 easy to miss one of the cases when reading the code, leading to unwanted side effects.
 
+#### Noncompliant Code:
+
+```kotlin
+when (i) {
```
Mauin: The first lines in the code blocks should be indented as well, shouldn't it?

schalkms: Good catch! I will update the generator.

schalkms:
I think this can be merged now.

PR #621, opened 2017-12-15
Added rule documentation to CONTRIBUTING.md

schalkms:


schalkms:
@vanniktech I see where you are coming from. I like the idea. The **detekt-generator** could easily check the rules documentation for compliance. However, at this point it doesn't make much sense since the documentation for the current rule set is not finished yet. After that, I would really like to have this. It would simplify the review process for new contributed rules.

vanniktech:
I'm a 👾. I just wanted to propose that we'll create an issue for this and merge this but you were faster. Nice!

Mauin:
Currently the `detekt-generator` simply ignores missing values but we can easily make it throw Exceptions that will in turn break the build 👍

PR #631, opened 2017-12-20
Allow numbers in ClassNaming.

### LINKED ISSUE #169: Default NamingConventionViolation classPattern does not allow numbers
The regex is `'[A-Z$][a-zA-Z$]*'` which does not allow digits. Is this done on purpose?

vanniktech:
Fixes #169

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -52,21 +52,6 @@ output-reports:
   #  - 'PlainOutputReport'
   #  - 'XmlOutputReport'
 
-comments:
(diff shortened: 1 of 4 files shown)


vanniktech:
I ran `build` locally and I don't have any diff changes for the config file. Is there a difference between my computer (mac) and travis (linux)? I'm also unsure why the order was swapped.

schalkms:
Why did it change the `detekt-config.yml` file?

vanniktech:
Good question. I guess the ordering is not really stable.

schalkms:
I've never had this problem before and I'm also on a Mac.

PR #635, opened 2017-12-21
Change variable min & max length to match IntelliJ.

vanniktech:
If you think it's better that way I can adjust the tests.


arturbosch:
Hmm it kinda looks to me as if the kotlin devs were to lazy to check to ignore variable names in loops :/

vanniktech:
I don't quite get what you mean with your last comment artur.

arturbosch:
In loops names like `i` (for i in ints {}) is ok. All other variables should not be called a, b, c ... i or whatself. It feels kinda hacky to set this defaults ourself :(

vanniktech:
I do think having a min value of 1 is fine. I wouldn't go down the street of separating them via loops properties.

PR #636, opened 2017-12-21
Exclude LongMethod in tests by default.

vanniktech:
I don't feel like this rule should be on for tests. Often a longer more descriptive test is better than a short comprehensive one that's not understandable.


arturbosch:
What for do you need more than tweenty lines of code in your tests? oO

vanniktech:
When setting up a few things for a particular test. I wouldn't say it's that uncommon. If we wouldn't use the file system for testing, we'd execute it almost every single time.

schalkms:
I also don't quite like this setting. I can live with a slightly higher threshold for the test package, but I wouldn't want to disable this check for test methods completely.

vanniktech:
Makes sense. So should we close this then?

schalkms:
I would say so.

PR #641, opened 2017-12-24
Rule: UnusedPrivateMember

### LINKED ISSUE #8: Rule: UnusedPrivateMember
Check if private fields, methods, parameter and locale variables are unused.

Idea: 
1. Visit KtFile and store names of fields and methods in a Map<String, Int>
2. Visit methods look for references, increment reference map
3. If field/method have same name as local or parameter, check positions.

Mauin:
Resolves #8 

This should cover unused private properties. Public properties are ignored as they might be used from other files.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
@@ -1,0 +1,80 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.DetektVisitor
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.psi.KtClassOrObject
+import org.jetbrains.kotlin.psi.KtElement
+import org.jetbrains.kotlin.psi.KtFunction
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.psi.KtReferenceExpression
+import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
+import org.jetbrains.kotlin.psi.psiUtil.isPrivate
+
+/**
+ * Reports unused private properties. If private properties are unused they should be removed. Otherwise this dead code
+ * can lead to confusion and potential bugs.
+ *
+ * @author Marvin Ramin
+ */
+class UnusedPrivateMember(config: Config = Config.empty) : Rule(config) {
+	override val issue: Issue = Issue("UnusedPrivateMember",
+			Severity.Maintainability,
+			"Private property is unused.")
+
+	override fun visitClassOrObject(classOrObject: KtClassOrObject) {
+		val visitor = UnusedPropertyVisitor()
+		classOrObject.accept(visitor)
+
+		visitor.properties.forEach {
+			report(CodeSmell(issue, Entity.from(it.value), "Private property ${it.key} is unused."))
+		}
+
+		super.visitClassOrObject(classOrObject)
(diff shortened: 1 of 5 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -0,0 +1,80 @@
...
+
+			val name = expression.text
+			if (!localProperties.contains(name) && properties.contains(name)) {
+				properties.remove(name)
+			}
+			super.visitReferenceExpression(expression)
+		}
+	}
+
+	class Test {
```
vanniktech: Hanging fruit?

Mauin: Oops, was testing with this one 🙃

arturbosch:
Cool, this is a useful rule! Do you want to support parameters too? Unused private functions you can also find with the same reference counting.

I have implemented a UnusedPrivateMembers rule for java. Here is the test case:
```
@SuppressWarnings("ALL") public class DeadCodeDummy {

	private int deadField = 5;
	private int usedField = 5;
	private String usedString = "Hello";

	private int deadMethod(int deadParameter) {
		int deadLocaleVariable = 5;
		usedMethod("" + usedField++);
		usedField = usedString.length();
		return 5;
	}

	public int usedMethod(String usedParameter) {
		usedField = usedParameter.length() + usedString.length();
		int usedLocaleVariable = 4;
		return Integer.valueOf(usedField).compareTo(usedLocaleVariable);
	}

	public void usedParameterInAssignment(String uri, long j) {
		int i = (int) j;
		String fullUrl = uri + ("/commits/new" + 6);
		System.out.println(i + fullUrl);
	}
}

Also you should check the refences in all possible conditional statements:
```
	private void run(boolean go) {
		boolean used = true;
		if(go && used) {
			String concat = "";
			switch (concat) {
				case "truetrue": run2(); break;
				default: break;
			}
		}
	}

	private void run2() {
		List<String> strings = new ArrayList<>();
		for (String s : strings) {
			System.out.println("");
		}
}
```
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
```diff
@@ -0,0 +1,261 @@
...
+
+class UnusedPrivateMemberSpec : SubjectSpek<UnusedPrivateMember>({
+	subject { UnusedPrivateMember() }
+
+	given("several classes with properties") {
+
+		it("reports an unused member") {
+			val code = """
+				class Test {
+				    private val unused = "This is not used"
```
tasomaniac: indentation seems to be off. This seems to be using spaces over tabs.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	class UnusedFunctionParameterVisitor : DetektVisitor() {
+		var parameters: MutableMap<String, KtParameter> = mutableMapOf()
+
+		override fun visitNamedFunction(function: KtNamedFunction) {
+			super.visitNamedFunction(function)
+
+			function.valueParameterList?.parameters?.forEach {
+				val name = it.name ?: throw IllegalStateException("Value parameter should have a name.")
```
tasomaniac: Couldn't you do just `!!`? That will throw for you. If the intention is to have a custom message, then I think the exception type should better to be `NullPointerException`

arturbosch: Are there any cases where a parameter has no name?
I think it is just for anonymous objects that it can be null.
How about using `safeName` or `nameAsSafeName` (dont remember how it is called)?

Mauin: TIL about `nameAsSafeName`. Will change it!

Mauin:
Great idea @arturbosch I added some extra handling to catch unused loop parameters.

Issue #643, opened 2017-12-24
[MagicNumber] Allow ignoring default values in constructor properties


arturbosch:
As mentioned in [sonar-kotlin](https://github.com/arturbosch/sonar-kotlin/issues/49#issuecomment-353542268)
the default values are always reported: `class SomeClassWithDefault(val defaultValue: Int = 10)`

Maybe we should also add an ignore property for class/object constructor properties.

Here a first test case:
```kotlin
given("in class declaration with default properties") {
    val ktFile = compileContentForTest("class SomeClassWithDefault(val defaultValue: Int = 10)")

    it("should report no finding") {
        assertThat(MagicNumber().lint(ktFile)).isEmpty()
    }
}
```

schalkms:
👌

schalkms:
Closed in 07555cf

PR #644, opened 2017-12-25
Implements #643 - MagicNumber ignores default values in ctor properties

schalkms:
Implements #643

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
@@ -18,9 +18,12 @@ import org.jetbrains.kotlin.psi.KtEnumEntry
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtObjectDeclaration
 import org.jetbrains.kotlin.psi.KtOperationReferenceExpression
+import org.jetbrains.kotlin.psi.KtParameter
 import org.jetbrains.kotlin.psi.KtPrefixExpression
+import org.jetbrains.kotlin.psi.KtPrimaryConstructor
 import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.KtReturnExpression
+import org.jetbrains.kotlin.psi.KtSecondaryConstructor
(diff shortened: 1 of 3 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
```diff
@@ -142,6 +146,11 @@ class MagicNumber(config: Config = Config.empty) : Rule(config) {
 private fun KtConstantExpression.isPartOfFunctionReturnConstant() =
 		parent is KtNamedFunction || (parent is KtReturnExpression && parent.parent.children.size == 1)
 
+private fun KtConstantExpression.isPartOfConstructor(): Boolean {
+	val isInConstructor = parent.parent.parent is KtPrimaryConstructor || parent.parent.parent is KtSecondaryConstructor
```
Mauin: Can we extract `parent.parent.parent` and give it a name?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
```diff
@@ -142,6 +146,11 @@ class MagicNumber(config: Config = Config.empty) : Rule(config) {
 private fun KtConstantExpression.isPartOfFunctionReturnConstant() =
 		parent is KtNamedFunction || (parent is KtReturnExpression && parent.parent.children.size == 1)
 
+private fun KtConstantExpression.isPartOfConstructor(): Boolean {
```
Mauin: `isPartOfConstructor` sounds like this would also ignore instances where the constructor is called with a magic number: `SomeClass(10)` which is not the case, I guess.

schalkms: Thanks for this. I’ll have a look later on today. Shouldn’t detekt already have such a test case?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
schalkms: Do you have any suggestions? 
**great-grandparent** does not improve the readability much I think.

PR #651, opened 2017-12-30
Align ModifierOrder check with official Kotlin coding conventions

3flex:


schalkms:
@vanniktech yes, it makes sense to compare the changes with the current test cases.

3flex:
@Mauin sorry I missed that! Updated.
@vanniktech @schalkms any updates required? I looked at the test cases and they don't need any changes to test the positive and negative cases.

vanniktech:
Then let's add one that verified the correct behavior?

3flex:
Do you mean in addition to https://github.com/arturbosch/detekt/blob/master/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ModifierOrderSpec.kt?

There are already some test cases there that check this rule - or am I misunderstanding?

vanniktech:
While there are some test cases they didn't test all of the modifiers. I'd just add 1-2 test cases for some of the modifiers that got swapped like internal and expect.

3flex:
I've added a couple of additional tests covering the `actual` and `varargs` keywords. Hopefully this will suffice.

3flex:
Triggering a new Travis build.

Issue #652, opened 2017-12-30
Update rules/new rules for new Kotlin Coding Conventions


Mauin:
JetBrains released new Kotlin Coding Conventions. Let's use this issue to gather rules that should be updated/created to match the conventions similar how #532 tracks the work towards matching the Android guides which are similar to the JetBrains one.

Conventions can be found here: https://kotlinlang.org/docs/reference/coding-conventions.html

Features (please update):
- [x] `MemberNameEqualsClassName` factory functions
- [x] ~Class layout rule~ (refer to https://github.com/shyiko/ktlint/issues/309)

Mauin:
One change to the `MemberNameEqualsClassName` rule: Factory Functions that create an instance of the class can have the same name as the class

https://kotlinlang.org/docs/reference/coding-conventions.html#function-names

schalkms:
I'll change the `MemberNameEqualsClassName` rule.

3flex:
The class ordering rule may (should?) be implemented in ktlint (https://github.com/shyiko/ktlint/issues/309)

MemberNameEqualsClassName rule has been updated.

Maybe this should be closed and specific rules can be suggested in individual issues as the Kotlin coding conventions develop over time?

schalkms:
> Maybe this should be closed and specific rules can be suggested in individual issues as the Kotlin coding conventions develop over time?

Fully agreed.

schalkms:
> The class ordering rule may (should?) be implemented in ktlint

This makes sense since this only a formatting issue.
I updated the initial post with a reference to the ktlint issue.
I'll close this for the mentioned reasons.

PR #653, opened 2017-12-30
Added factory function check to MemberNameEqualsClassName

schalkms:
Implements the feature mentioned in #652

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MemberNameEqualsClassName.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MemberNameEqualsClassName.kt
@@ -35,6 +35,18 @@ import org.jetbrains.kotlin.util.collectionUtils.concat
  * }
  * </noncompliant>
  *
+ * <compliant>
+ * class Manager {
+ *
+ *     companion object {
+ *         // factory functions can have the same name as the class
+ *         fun manager(): Manager {
+ *             return Manager()
+ *         }
+ *     }
+ * }
+ * </compliant>
+ *
  * @configuration ignoreOverriddenFunction - if overridden functions should be ignored (default: true)
  *
  * @author schalkms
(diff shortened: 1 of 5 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MemberNameEqualsClassName.kt
```diff
@@ -35,49 +55,61 @@ import org.jetbrains.kotlin.psi.KtObjectDeclaration
...
+		list
+				.filterIsInstance<KtNamedFunction>()
+				.filter { isFactoryMethod(it, klass) }
+				.forEach { it -> list.remove(it) }
+		return list
+	}
+
+	private fun isFactoryMethod(function: KtNamedFunction, klass: KtClass): Boolean {
+		val typeReference = function.typeReference
+		return typeReference == null && function.bodyExpression !is KtBlockExpression
```
schalkms: This ignores functions which return a single expression since detekt cannot infer the type (at the moment).
Consider the following code for example
```kotlin
open class A {
  companion object {
    // return type = A
    fun a(condition: Boolean) = if (condition) B() else C()
  }
}

class B: A()
class C: A()
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MemberNameEqualsClassName.kt
```diff
@@ -47,37 +51,49 @@ class MemberNameEqualsClassName(config: Config = Config.empty) : Rule(config) {
...
+		return functions
 	}
 
-	private fun checkCompanionObjectFunctions(klass: KtClass) {
-		klass.companionObjects.forEach {
-			checkClassOrObjectFunctions(it, klass.name, classMessage)
-		}
+	private fun getMisnamedCompanionObjectMembers(klass: KtClass): List<KtNamedDeclaration> {
+		val list = mutableListOf<KtNamedDeclaration>()
+		klass.companionObjects.forEach { list.addAll(getMisnamedMembers(it, klass.name)) }
```
arturbosch: This can be just a `flatMap`-Funktion?

arturbosch: Thanks!

PR #656, opened 2018-01-02
Run tests on Travis on Oracle JDK 8 & 9

3flex:
Hopefully will avoid any further surprises with Java 9.


3flex:
Once https://github.com/arturbosch/detekt/pull/654 is merged (with fix for https://github.com/arturbosch/detekt/issues/566), this should pass :)

I'll reopen when #654 is in and the new artifact is available in the repos.

Issue #602, opened 2017-12-11
HtmlOutputFormatTest tests assume Unix-style line endings


3flex:
An example: `assertThat(result).startsWith("<!DOCTYPE html>\n<html lang=\"en\">")`

`default-html-report-template.html` is checked out on Windows with Windows-style line endings (per my system default), so the test fails.

Recommend adding either a `.gitattributes` file explicitly specifying Unix-style line endings for these templates so they're checked out with Unix-style line endings on Windows, or changing the tests to remove the Unix-style line endings (I think the first option is best).

arturbosch:
Do you might opening a PR for this :) ?

3flex:
Adding .gitattributes or changing the test?

arturbosch:
.gitattributes please

arturbosch:
Thanks for the fix !

PR #660, opened 2018-01-03
Appveyor: Use default git config

3flex:
Appveyor uses `git config --global core.autocrlf input` by default. However this is not
the default on Windows machines, meaning some tests that pass on Appveyor may not pass when
checked out on a Windows machine if the tests are affected by a file's line endings.
This causes issues like https://github.com/arturbosch/detekt/issues/602 which could be picked up earlier.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,8 @@ environment:
   TERM: dumb
   matrix:
     - JAVA_HOME: C:\Program Files\Java\jdk1.8.0
+init:
+  - git config --global --unset core.autocrlf
 install:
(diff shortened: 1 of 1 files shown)


3flex:
I've been bitten by another issue like #602. Hoping the check fails on Appveyor this time like it does locally, so this doesn't happen again.

3flex:
@arturbosch do you want me to rebase? When I do this should pass now #661 is merged.

arturbosch:
I think it will be ok, just had to go. Will try now :)

PR #661, opened 2018-01-03
Rename *.yaml to *.yml so fixtures use expected line endings

3flex:
Fixes issue uncovered by #660.

Merge this before #660 to avoid build failures on `master` :)


arturbosch:
Just for interest ... what is the correct spelling of the yaml ending .. ?

3flex:
Officially it's *.yaml, but *.yml is apparently more commonly used (https://stackoverflow.com/questions/21059124/is-it-yaml-or-yml)

It doesn't matter to me at all, so long as `.gitattribtues` is in line with the extension being used by the YAML files in the tests to avoid issues when checking out on Windows.

If #660 is merged it will pick up those issues.

PR #665, opened 2018-01-08
Do not run EmptyClassBlock on objects of anonymous classes

egor-n:
There's no way to create an object of an anonymous class without specifying the body. For example, Gson's TypeToken:

```kotlin
val type = object : TypeToken<Map<String, String>>() {}.type
```

This empty body should not be flagged as there's no other way to write it.


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyClassBlock.kt
```diff
@@ -15,6 +16,8 @@ import org.jetbrains.kotlin.psi.KtClassOrObject
 class EmptyClassBlock(config: Config) : EmptyRule(config) {
```
arturbosch: Please also add your name to this class as stated in the contributing rules. Thanks!

PR #667, opened 2018-01-08
Do not run EmptyFunctionBlock on open functions

### LINKED ISSUE #666: EmptyFunctionBlock should not flag functions with open modifier
The title duplicates this test:
https://github.com/arturbosch/detekt/blob/994c732c5f45ebdbb50a917b272f87a9c0f5916f/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlockSpec.kt#L29-L32

The name is correct, the implementation is not but it passes. The test should read:
```kotlin
```

egor-n:
Overridable functions do not have to be `protected`.

Fixes #666.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlock.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlock.kt
@@ -22,5 +22,5 @@ class EmptyFunctionBlock(config: Config) : EmptyRule(config) {
 		}
 	}
 
-	private fun KtNamedFunction.notMeantForOverriding() = !(isOpen() && isProtected())
+	private fun KtNamedFunction.notMeantForOverriding() = !isOpen()
 }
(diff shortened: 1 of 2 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlockSpec.kt
```diff
@@ -16,19 +16,14 @@ class EmptyFunctionBlockSpec : SubjectSpek<EmptyFunctionBlock>({
 
 	describe("should not flag functions meant to be overridden") {
 
-		it("should not flag function with protected and open modifier") {
-			val findings = subject.lint("protected open fun stuff() {}")
```
egor-n: This test is a duplicate of the `"should not flag function with open modifier"` so I deleted it.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlock.kt
```diff
@@ -22,5 +22,5 @@ class EmptyFunctionBlock(config: Config) : EmptyRule(config) {
 		}
 	}
```
schalkms: Why is the isProtected() call not necessary anymore?

egor-n: Function doesn't have to be `protected` in order to be overridable. You can override this function, for example:
```kotlin
open class A {
  open fun doSomething() {
  }
}
```

schalkms: Sorry for my question 😉
You already mentioned it in the issue description.

PR #668, opened 2018-01-08
Remove inactive-by-default rules from test-pattern exclusions

egor-n:
Both `LateinitUsage` and `StringLiteralDuplication` are disabled by default. There's no point in including them in the `test-pattern` section in the `default-detekt-config`.

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -14,8 +14,6 @@ test-pattern: # Configure exclusions for test sources
     - 'WildcardImport'
     - 'MagicNumber'
     - 'MaxLineLength'
-    - 'LateinitUsage'
-    - 'StringLiteralDuplication'
     - 'SpreadOperator'
     - 'TooManyFunctions'
 
--- a/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/ConfigPrinter.kt
+++ b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/ConfigPrinter.kt
@@ -70,8 +70,6 @@ object ConfigPrinter : DocumentationPrinter<List<RuleSetPage>> {
 			    - 'WildcardImport'
 			    - 'MagicNumber'
 			    - 'MaxLineLength'
-			    - 'LateinitUsage'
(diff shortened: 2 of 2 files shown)


egor-n:
`failFast` is not enabled by default, though.

If you enable the `failFast` option in the default config, the disabled rules will still be disabled (for tests too).

vanniktech:
No they won't be disabled. Using `failFast` enables all the rules.

egor-n:
But they are explicitly disabled in the config - doesn't this take priority?

vanniktech:
Nope. All of the things are wrapped in a [`FailFastConfig`](https://github.com/arturbosch/detekt/blob/02c183bba187a12fd56f6b9c822f83e3dc7b13da/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt#L58).

egor-n:
https://github.com/arturbosch/detekt/blob/02c183bba187a12fd56f6b9c822f83e3dc7b13da/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt#L58-L64

It returns the original value if it is present - or am I misreading this?

vanniktech:
It's overriding the default one `false` with `true` though and that makes them active.

egor-n:
@vanniktech where does this happen? I seem to not understand this correctly.

I enabled `failFast` in this project's `detekt-cli/src/main/resources/default-detekt-config.yml` which supposedly should have enabled all rules from detekt? I ran `./gradlew detektCheck` and it didn't flag any new issues even.

I, then, removed the following from the config:
```yaml
  OptionalWhenBraces:
    active: false
```
and ran the `detektCheck` again and it did flag 3 instances of `OptionalWhenBraces` rule breakage.

vanniktech:
This is working as intended. If you use failFast you don't need to enable the rules anymore and your configuration file boils down to a few lines instead of 200+. Basically it could be only one line, however if you want to override certain properties then you specify them.

egor-n:
@vanniktech I don't understand your point.

> if you want to override certain properties then you specify them

You do the same if you want to disable a property 🤔

egor-n:
We cleared things up – the thing I didn't know is that detekt merges default config with the one specified in the build configuration. But, `failFast: true` does not override the `test-pattern` from the default config. This means that if we remove these 2 rules from the `test-pattern`, then everyone who relied on the default `test-pattern` will have to copy-paste the whole section to their config.

PR #669, opened 2018-01-08
Treat all compiler warnings as errors

egor-n:


3flex:
Could this be done in CI instead of for every build? I'm aware I don't follow best practice, but when prototyping I often have code sitting around that triggers compiler warnings which I'll clean up before submitting a PR. Having the build fail in these cases will be annoying.

I was going to submit this same change until I realised this :)

vanniktech:
That makes sense.

egor-n:
Makes sense. I have updated the PR to only enable this option if you pass `-PwarningsAsErrors=true` command line argument.

PR #515, opened 2017-10-22
LateInitUsage -> LateinitUsage

vanniktech:
Should fix #513.

PR #671, opened 2018-01-09
Change LateInitUsage to LateinitUsage in failfast.yml

egor-n:


egor-n:
Yeah, it was fixed in #515

PR #674, opened 2018-01-09
Implement MatchingDeclarationName rule

### LINKED ISSUE #664: MatchingDeclarationName rule to match single declaration to file name
If the file is named `Foo.kt` and there's only one declaration (object, class, enum or interface) this file name should match the declaration name.

Positive case:

**Something.kt**

```kotlin
object Something
```

Negative case:

**Something.kt**

```kotlin
object Somethin
```

Also I'd expect this check to work when there are top level variables declared:

**Something.kt**

```kotlin
const val MEANING_OF_LIFE = 42

object Something
```

What do you think?

arturbosch:
Fixes #664 

Edit: Need still fix the reporting findings but the rule should be rdy for review


arturbosch:
Yeah good point. Also, should the rule be active by default? Both jetbrains and google style guides have this rule, so I set it to active.

vanniktech:
Yup it should be active.

arturbosch:
Interfaces and Enums inherit from KtClass so I just had to add some test cases for them.
All new code smells from this rule are fixed.
I've added a new `output` profile to generate reports and get fast access to signatures of smells.
Needed to move SingleAssign to its own file and update the baseline.

PR #676, opened 2018-01-10
Add UntilInsteadOfRangeTo rule

geralt-encore:
https://github.com/arturbosch/detekt/issues/670
I am not sure about rule name and description tho.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
@@ -19,15 +19,15 @@ import org.jetbrains.kotlin.psi.KtForExpression
  * some value subtracted by 1. 'until' helps to prevent off-by-one errors.
  *
  * <noncompliant>
+ * for (i in 0 .. 10 - 1) {}
+ * </noncompliant>
  * for (i in 0 until 10 - 1) {}
  * for (i in 10 downTo 2 - 1) {}
  * for (i in 0 .. 10) {}
  * for (i in 0 .. 10 + 1) {}
  * for (i in 0 .. 10 - 2) {}
- * </noncompliant>
- *
  * <compliant>
- * for (i in 0 .. 10 - 1) {}
+
  * </compliant>
  *
  * @author Ilya Zorin
@@ -51,12 +51,14 @@ class UntilInsteadOfRangeTo(config: Config = Config.empty) : Rule(config) {
 		super.visitForExpression(expression)
 	}
 
-	@Suppress("ReturnCount")
 	private fun isUntilApplicable(range: Array<PsiElement>): Boolean {
-		if (range[1].text != "..") return false
-		val expression = range[2] as? KtBinaryExpression ?: return false
-		if (expression.operationToken != KtTokens.MINUS) return false
-		val rightExpressionValue = expression.right as? KtConstantExpression ?: return false
-		return rightExpressionValue.text == "1"
+		if (range[1].text == "..") {
+			val expression = range[2] as? KtBinaryExpression
+			if (expression?.operationToken == KtTokens.MINUS) {
+				val rightExpressionValue = expression?.right as? KtConstantExpression
+				return rightExpressionValue?.text == "1"
+			}
+		}
+		return false
 	}
 }
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
@@ -15,6 +15,8 @@ import org.jetbrains.kotlin.psi.KtForExpression
 
 /**
  * Reports calls to 'rangeTo' or '..' operator instead of calls to 'until'.
+ * 'until' is applicable in cases where upper range value is described as
+ * some value subtracted by 1. 'until' helps to prevent off by one errors.
  *
  * <noncompliant>
  * for (i in 0 until 10 - 1) {}
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
@@ -28,7 +28,6 @@ import org.jetbrains.kotlin.psi.KtForExpression
  *
  * <compliant>
  * for (i in 0 .. 10 - 1) {}
- * for (i in 0 rangeTo 10 - 1) {}
  * </compliant>
  *
  * @author Ilya Zorin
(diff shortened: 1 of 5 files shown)


geralt-encore:
Looks like I need to do something with default config generation, but I have no idea what to do. Would appreciate some help :)

Mauin:
@geralt-encore When you run `./gradlew build` on your machine locally the generator will update the documentation files and the default config with the changes. You can then simply commit them.

geralt-encore:
thanks @Mauin! Wasn't sure about that

geralt-encore:
So the problem is in return count rule. I am not sure how to properly handle it. It is possible to rework it with nested conditions, but it is arguably even worse. At least using returns like that is pretty common in Kotlin Intellij plugin codebase as far as I can tell

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
```diff
@@ -0,0 +1,60 @@
...
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.psi.KtBinaryExpression
+import org.jetbrains.kotlin.psi.KtConstantExpression
+import org.jetbrains.kotlin.psi.KtForExpression
+
+/**
+ * Reports calls to 'rangeTo' or '..' operator instead of calls to 'until'.
```
schalkms: Please add a more descriptive text here, so that others can understand this rule better. This will be included in detekt's documentation.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
```diff
@@ -0,0 +1,60 @@
...
+			report(CodeSmell(issue, Entity.from(loopRange),
+					"'rangeTo' or the '..' call can be replaced with 'until'"))
+		}
+		super.visitForExpression(expression)
+	}
+
+	private fun isUntilApplicable(range: Array<PsiElement>): Boolean {
+		if (range[1].text !in setOf("rangeTo", "..")) return false
+		val expression = range[2] as? KtBinaryExpression ?: return false
+		if (expression.operationToken != KtTokens.MINUS) return false
```
schalkms: Try to split up `isUntilApplicable`into separate methods.

geralt-encore: I tried it. The problem is since a couple of casts are required here and next values a based on previous ones - it ends up as a bunch of nested conditions which is by no means is better than this code I think.

schalkms: Ok, then you are allowed to supress this violation.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
schalkms: > 'until' helps to prevent off by one errors

I am afraid, I don’t know what you would like to express with your last sentence. Can we please rephrase it?

geralt-encore: It is about https://en.wikipedia.org/wiki/Off-by-one_error
but I forgot dashes.

schalkms: Oh right. Without dashes it's hard to recognize it.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
```diff
@@ -0,0 +1,63 @@
...
+class UntilInsteadOfRangeTo(config: Config = Config.empty) : Rule(config) {
+
+	override val issue = Issue(javaClass.simpleName,
+			Severity.Style,
+			"'rangeTo' or the '..' call can be replaced with 'until'",
+			Debt.FIVE_MINS)
+
+	private val minimumSize = 3
+
+	override fun visitForExpression(expression: KtForExpression) {
```
egor-n: It might be worth it to check other cases too, other than for-loops.

Take a look at the way the [IDEA inspection](https://github.com/JetBrains/kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceRangeToWithUntilInspection.kt) is implemented.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
```diff
@@ -0,0 +1,62 @@
...
+ * <noncompliant>
+ * for (i in 0 until 10 - 1) {}
+ * for (i in 10 downTo 2 - 1) {}
+ * for (i in 0 .. 10) {}
+ * for (i in 0 .. 10 + 1) {}
+ * for (i in 0 .. 10 - 2) {}
+ * </noncompliant>
+ *
+ * <compliant>
+ * for (i in 0 .. 10 - 1) {}
```
arturbosch: Seems like you missmatched the `compliant` and `noncompliant` code. Noncompliant is what detekt should detect.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
```diff
@@ -0,0 +1,62 @@
...
+		val loopRange = expression.loopRange
+		val range = loopRange?.children
+		if (range != null && range.size >= minimumSize && isUntilApplicable(range)) {
+			report(CodeSmell(issue, Entity.from(loopRange),
+					"'..' call can be replaced with 'until'"))
+		}
+		super.visitForExpression(expression)
+	}
+
+	@Suppress("ReturnCount")
```
arturbosch: why @schalkms requested changes, was to remove the 5 return statements.
multiple `if (this != that) return false` is much harder to read than nested structures:
```kotlin
private fun isUntilApplicable(range: Array<PsiElement>): Boolean {
	if (range[1].text == "..") {
		val expression = range[2] as? KtBinaryExpression
		if (expression?.operationToken == KtTokens.MINUS) {
			val rightExpressionValue = expression?.right as? KtConstantExpression
			return rightExpressionValue?.text == "1"
		}
	}
	return false
}
```

Please refactor this function to just use not more than two returns like in my example. Thanks!

geralt-encore: Thank you, it looks much better than option that I tried before

arturbosch:
Thanks for the changes and the whole rule!

PR #687, opened 2018-01-13
Fix MatchingDeclarationName false positives

### LINKED ISSUE #686: MatchingDeclarationName has false positives
In this case (filename `Whatever.kt`):

```kotlin
fun a()

private class Something
```

The MatchingDeclarationName rule flags the file — I suppose saying that it should be called `Something.kt` as it's the only class in there — but it shouldn't really as this class contains only a publicly visible top level function and that private class is only an implementation detail of the function. One such file for which this is reproducible is [this](https://github.com/squanchy-dev/squanchy-android/blob/develop/app/src/main/java/net/squanchy/tweets/service/TweetModelConverter.kt).

rock3r:
This PR fixes #686 in that it filters out private classes/interfaces/objects from the `MatchingDeclarationName` rule. These private classes/interfaces/objects are an implementation detail and should not determine the name of a file, whose only public member may well be a function.

In addition to the visibility filter, this PR also improves upon FindingsAssertions and uses them in the rule's negative cases tests in lieu of the more lax `assertThat(findings).hasSize(1)`.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MatchingDeclarationName.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MatchingDeclarationName.kt
@@ -1,6 +1,12 @@
 package io.gitlab.arturbosch.detekt.rules.style
 
-import io.gitlab.arturbosch.detekt.api.*
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
 import org.jetbrains.kotlin.psi.KtClassOrObject
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.psiUtil.isPrivate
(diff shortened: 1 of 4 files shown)


rock3r:
Good point, not sure why it's not using tabs. Will fix.

rock3r:
Done :)

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MatchingDeclarationName.kt
```diff
@@ -47,13 +48,14 @@ import org.jetbrains.kotlin.psi.KtFile
 class MatchingDeclarationName(config: Config = Config.empty) : Rule(config) {
 
 	override val issue: Issue = Issue(javaClass.simpleName, Severity.Style,
-			"If a source file contains only a single top-level class or object, " +
+			"If a source file contains only a single non-private top-level class or object, " +
```
Mauin: Can you add this non-private destinction also to the KDoc and re-generate the documentation with it? 🙃

rock3r: Sure, on it

rock3r: @Mauin Done :)

rock3r: While verifying this I have noticed #691 happens and incomplete documentation is generated for some rules

arturbosch:
Merged manually.

PR #689, opened 2018-01-14
Sort Rules in Config & Documentation alphabetically

### LINKED ISSUE #688: Default configuration file should be sorted alphabetically
I have just spent a _long_ time for something that should be extremely simple, that is, bumping Detekt from RC6 to RC6-2 because merging the rules was a very painful [manual merge](https://github.com/squanchy-dev/squanchy-android/pull/425).

This is because many of the rules had changed place in the default ruleset compared to RC6 (when I also did merge in the changes), and all the sections had changed place.

If the sections, and the rules within the sections, were sorted alphabetically it would be much easier to compare ruleset files and merge in any new and changed rule. I cannot use the default set as-is as some settings I have are different from the defaults.

Mauin:
Resolves #688 

I'm thinking if it makes sense having a `NamingRuleSetProvider` which groups all the class/object/property/function naming rules together. With this alphabetical sorting they are all placed in different parts of the `style` ruleset config. Before they were grouped together.

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -69,16 +69,9 @@ comments:
 
 complexity:
   active: true
-  LongParameterList:
+  ComplexCondition:
     active: true
-    threshold: 5
-    ignoreDefaultParameters: false
-  LongMethod:
-    active: true
-    threshold: 20
-  LargeClass:
-    active: true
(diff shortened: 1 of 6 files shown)


arturbosch:
Yep a naming ruleset sounds good. Would you also add the MatchingDeclarationName rule to it?

Mauin:
I'll create a 2nd PR for it, sure :

Issue #694, opened 2018-01-15
Consider adding `excludeAnnotatedClasses` to UseDataClass.


serj-lotutovici:
I found myself in need of an exclude because some classes that are more configuration that data, give these false positives. For example:

```
@Module
class MyModule(
    @get:Provides val someVal: SomeInterface, 
    @get:Provides val otherVal:SomeOtherInterface
)
```

I could just add a suppress here, but it seems wrong because an object of this class will never require an `equals` or `hashCode`.

arturbosch:
I hate annotation-driven-development :(

Mauin:
This addition does make sense for the rule, though. Especially in Android we'll see quite a few of these annotated classes which don't need to be `data class`es.

PR #695, opened 2018-01-16
Added `excludeAnnotatedClasses` to UseDataClass - #694

schalkms:


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClass.kt
```diff
@@ -29,6 +29,8 @@ import org.jetbrains.kotlin.psi.psiUtil.isPropertyParameter
  * data class DataClass(val i: Int, val i2: Int)
  * </compliant>
  *
+ * @configuration excludeAnnotatedClasses - if annotated class should be ignored (default: true)
```
serj-lotutovici: I think this configuration should allow more. Say specify the actual annotation which forces the rule to ignore the class (similar to the lateinit rule).

schalkms: This raises another question. Should `UseDataClass` have a second config option which specifies all excluded annotations?

serj-lotutovici: as a user or the tool, I would say it would be easier to specify one config, then keeping in mind that there's a multitude of them.

Mauin:
As @vanniktech mentioned: I think it makes sense having some consistency across different rules here and take a list of Annotations as an input for this rule

schalkms:
This uses the same kind of configuration as the `LateInit` rule.

PR #696, opened 2018-01-16
update kotlin to v1.2.20

Mauin:
Just saw the binaries pop up, don't see any release notes yet.


3flex:
This will do until a blog post goes up: https://github.com/JetBrains/kotlin/blob/1.2.20/ChangeLog.md

PR #703, opened 2018-01-21
KDocStyle MultiRule with EndOfSentenceFormat Rule

### LINKED ISSUE #699: Rule: KDoc's first sentence should have a proper end
Checkstyle's JavadocStyle for reference – http://checkstyle.sourceforge.net/config_javadoc.html#JavadocStyle.

KDoc's first sentence should have a proper end. The "proper end" can be defined with a regular expression. Looks like checkstyle uses `([.?!][ \t\n\r\f<])\|([.?!]$)`.

Would also be nice to check if the first sentence starts with a capital letter.

Mauin:
Resolves #699 

This PR sets up a `KDocStyle` MultiRule to check different aspects of KDoc in the future. For now I have only implemented the `EndOfSentenceFormat` check to verify the first KDoc sentence ends with proper punctuation.

Any suggestions on how to make the tests better? I kind of dislike calling the `KDocStyle` rule in the test instead of the `EndOfSentenceFormat` rule directly.

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -58,6 +58,9 @@ comments:
     active: false
   CommentOverPrivateProperty:
     active: false
+  EndOfSentenceFormat:
+    active: false
+    endOfSentenceFormat: ([.?!][ \t\n\r\f<])|([.?!]$)
   UndocumentedPublicClass:
     active: false
     searchInNestedClass: true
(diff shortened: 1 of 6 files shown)


arturbosch:
How about we name the rule `EndOfSentenceInKDoc` or `KDocEndOfSentence`?
Format does not tell me it is the documentation imo.

Mauin:
The rule already is in the `documentation` rule set. Does that offer enough clarification? I'm happy to add a reference to `KDoc` in the rule name. But I guess that is kind of obvious given the rule set.

arturbosch:
Ok, that makes sense now. I'm happy with the name if you are :).

PR #704, opened 2018-01-21
make detekt-cli tests depend on updated documentation/config

Mauin:
When changing or adding rules currently `./gradlew build` might fail as we have tests in the `detekt-cli` module which verify that the `default-detekt-config.yml` contains all rules. 

A `./gradlew build` will assemble all modules and execute the tests in no particular order. If the `detekt-cli` tests run before the `detekt-generator` has updated the `default-detekt-config.yml` the tests can fail. 

This PR adds a dependency on the `detekt-cli` tests to run after the documentation and config have been updated. Alternatively we can remove the tests that verify the config altogether as we now generate the config. But we can also keep the test around as a kind of "integration test".

### DIFF
--- a/detekt-cli/build.gradle
+++ b/detekt-cli/build.gradle
@@ -16,3 +16,5 @@ dependencies {
 	testRuntimeOnly "org.junit.platform:junit-platform-console:$junitPlatformVersion"
 	testRuntimeOnly "org.jetbrains.spek:spek-junit-platform-engine:$spekVersion"
 }
+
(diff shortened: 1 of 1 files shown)


arturbosch:
If we remove the test, undocumented new rules will result CI to pass? We should leave it in imo.

Mauin:
@arturbosch The test that sometimes fails without these changes verifies that all rules that are added to `RuleSetProvider`s are also added to the `default-detekt-config.yml`

PR #709, opened 2018-01-25
Gradle Plugin: Expose a copy of `profiles` 

### LINKED ISSUE #705: Tooling: Accessing the profiles defined by the user
We are working on a plugin named `static-analysis` to be able to integrate different tools together. 
Not released yet but this PR adds detekt support.

As you can see in the discussion below, we thought that using `.systemOrDefaultProfile()` to access to the output reports may not be the best. Especially when the user has multiple profiles. 

We wanted to access `profiles` field but it seems that it is private. `searchProfileWithName` is public but we cannot use that to access all profiles defines. 

https://github.com/novoda/gradle-static-analysis-plugin/pull/46#discussion_r163277423

What do you think about making `profiles` public? Or is there any other recommended way to access all reports generated by detekt?

tasomaniac:
Working on an external tool that integrates detekt, we thought that it would be useful to expose a read-only list of `profiles`

Instead of making the `profiles` just public, introduced a [backing property](https://kotlinlang.org/docs/reference/properties.html#backing-properties) to hold the profiles and another public `profiles` property as a defensive copy of the actual list.

Fixes #705

### DIFF
--- a/README.md
+++ b/README.md
@@ -623,6 +623,7 @@ If you contributed to detekt but your name is not in the list, please feel free 
 - [Serj Lotutovici](https://github.com/serj-lotutovici) - Rule improvement: LongParameterList
 - [Dmitry Primshyts](https://github.com/deeprim) - Rule improvement: MagicNumber
 - [Egor Neliuba](https://github.com/egor-n) - Rule improvement: EmptyFunctionBlock, EmptyClassBlock
+- [Said Tahsin Dane](https://github.com/tasomaniac/) - Gradle plugin improvements
 
 ### <a name="mentions">Mentions</a>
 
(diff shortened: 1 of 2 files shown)


tasomaniac:
@arturbosch added as an author. Thanks for pointing that out. 
If this is something the the codebase do, that can also be added to contributing guidelines.

Mauin:
@tasomaniac It's already in the `CONTRIBUTING.md` https://github.com/arturbosch/detekt/blob/master/CONTRIBUTING.md#specific-code-style-rules-we-use 🙃

tasomaniac:
@Mauin it's like Friday for this whole week for me.

PR #711, opened 2018-01-26
Added tests for naming rules

schalkms:


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNamingSpec.kt
```diff
@@ -21,7 +20,17 @@ class EnumNamingSpec : Spek({
 				}
 				"""
 			)
-			Assertions.assertThat(findings).isEmpty()
+			assertThat(findings).isEmpty()
+		}
+	}
+
+	describe("") {
```
arturbosch: I guess wanted to delete this describe and put the test case to the above describe block?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNamingSpec.kt
```diff
@@ -21,7 +20,17 @@ class EnumNamingSpec : Spek({
...
+		it("reports a violation") {
```
arturbosch: "it reports an underscore in enum name"

Issue #727, opened 2018-02-06
False positive on rule UnnecessaryAbstractClass


xgouchet:
Take for instance the following classes : 

```
abstract class Foo {
    abstract fun bar() : Int

    fun baz() {
    }
}
```

```
abstract class Oof : Foo() {
    fun spam() {
    }
}
```

The class Oof is reported as being an `UnnecessaryAbstractClass`, because it only contains concrete methods. But in fact, because it inherits from Foo, it also have an abstract, unimplemented method (`bar`).

schalkms:
Thanks for reporting!
Detekt does not have type resolution (yet). For now I think this rule should ignore classes which derive from an abstract class. What do you think? @arturbosch @Mauin

arturbosch:
Yep, it makes sense for now.
Will you take care of the fix @schalkms ?
Maybe we should open an issue and collect all the improvements to the rules which can be done after type resolution?

schalkms:
I can do that. I think the *type-resolution** tag is sufficient.

vanniktech:
Got something similar that should not be flagged:

```kotlin
abstract class Screenshots {
  @get:Rule val demoModeRule = DemoModeRule()

  @Before fun setUp() {
    Screengrab.setDefaultScreenshotStrategy(UiAutomatorScreenshotStrategy())

    tearDown()
  }

  @After fun tearDown() {
    nukeSharedPreferences()
  }

  fun screenshot(name: String) {
    Screengrab.screenshot(name, Screengrab.getDefaultScreenshotStrategy(), ConstantFileWritingScreenshotCallback(application()))
  }

  fun isUsingFastlane() = InstrumentationRegistry.getArguments().getString("isUsingFastlane")?.toBoolean() == true

  private class ConstantFileWritingScreenshotCallback(applicationContext: Context) : FileWritingScreenshotCallback(applicationContext) {
    override fun getScreenshotFile(screenshotDirectory: File, screenshotName: String) =
      File(screenshotDirectory, "$screenshotName.png")
  }
}
```

There are multiple reasons:

- using JUnit annotations that can be used downstream in the classes
- one can still use the provided methods (`screenshot` & `isUsingFastlane`)

schalkms:
@vanniktech Hmm, I'm a bit skeptical and want to raise some questions. Should detekt really include framework specific things (JUnit in your case)? I'm a bit worried that this leads to implementing different edge cases which are only relevant for a minor audience.

[...]

PR #728, opened 2018-02-06
Ignored classes with supertype entries from UnnecessaryAbstractClass

schalkms:
reference #727


detekt-rules/src/test/resources/cases/UnnecessaryAbstractClassNegative.kt
```diff
@@ -15,3 +15,9 @@ abstract class AbstractClassWithPrimaryConstructorConcretePropertyOk(val i: Int)
 // empty abstract classes should not be reported by this rule
 abstract class EmptyAbstractClass1
 abstract class EmptyAbstractClass2()
+
+// This test case should be removed when type resolution is available
```
arturbosch: Please link the issue to the comment :)

PR #729, opened 2018-02-06
More idiomatic build configuration

mkobit:
- All repositories required for building moved to the `settings.gradle`
- Only apply plugins when they are used
- Remove deplicate plugin requests by removing some `buildscript` blocks
- Compile all `KotlinCompile` tasks together with same configuration

Some of these changes will allow for easier migration to the `kotlin-dsl`.
In the `kotlin-dsl`, you want all plugins to be declared in the `plugins {}` blocks so that accessors can be generated.

The `detekt-sample-extensions` project should be considered as another example project or possibly another build to run with composite builds as it currently should reuse the same configuration as the rest of the build.


build.gradle
```diff
@@ -14,14 +11,15 @@ buildscript {
...
+	id 'com.github.johnrengelman.shadow' version '2.0.2' apply false
 	id 'org.sonarqube' version '2.6.1'
+	id 'com.gradle.plugin-publish' version '0.9.9' apply false
 }
 
 apply plugin: 'io.gitlab.arturbosch.detekt'
 
 tasks.withType(Wrapper) {
-	gradleVersion = '4.3.1'
+	gradleVersion = '4.5'
```
arturbosch: 4.5.1 just came out :)

detekt-sample-extensions/build.gradle
```diff
@@ -1,21 +1,3 @@
...
-		classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
-	}
-}
-
-repositories {
-	mavenLocal()
-	jcenter()
-}
-
-apply plugin: "kotlin"
```
arturbosch: this duplication is by intention so that people wanting to extent detekt can copy this project

mkobit: Got it - the intention makes sense but I think the implementation needs a little bit of tweaking. The main project leaks it's repositories, plugins, properties, and other pieces into it.

My thoughts are to do something like the following:

- create a `GradleBuild` (https://docs.gradle.org/current/dsl/org.gradle.api.tasks.GradleBuild.html) task in the root build
- use a composite build to include the main build so that they get substituted for the "main" dependencies

but that is a little bit complex for right now. I can revert these parts of the changes for now and try and tackle it in a different PR. WDYT?

build.gradle
```diff
@@ -33,6 +31,7 @@ allprojects {
 		mavenCentral()
 		mavenLocal()
 		maven { url 'http://dl.bintray.com/jetbrains/spek' }
+		jcenter()
```
Mauin: `jcenter()` is a superset of `mavenCentral()` so let's remove `mavenCentral()` from this one.

settings.gradle
```diff
@@ -1,3 +1,11 @@
+pluginManagement {
+	repositories {
+		gradlePluginPortal()
+		mavenCentral()
```
Mauin: Let's replace `mavenCentral()` with `jcenter()`

detekt-sample-extensions/build.gradle
arturbosch: That would be great!

PR #730, opened 2018-02-07
Allow aliases for rules to allow for suppression

### LINKED ISSUE #726: Handle @Suppress("UNCHECKED_CAST") for UnsafeCast rule
IntelliJ / Android Studio detect when a cast is unsafe and show a warning, which can be disabled using the `@Suppress("UNCHECKED_CAST")` annotation. It seems redundant to also add the `UnsafeCast` rule, ending with some code annotated with `@Suppress("UNCHECKED_CAST", "UnsafeCast")`

Mauin:
Resolves #726

This adds an `aliases` property to `Issue` which can optionally be set with a list of aliases which match similar issue IDs in other static analysis suites so that they can be suppressed with the same annotation.

I wanted to add the `aliases` to `Rule` at first but that would require a lot more changes to other classes as well so I opted to add it to the `Issue`.


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Suppressible.kt
```diff
@@ -38,8 +39,9 @@ private const val QUOTES = "\""
 /**
  * Checks if this kt element is suppressed by @Suppress or @SuppressWarnings annotations.
  */
-fun KtAnnotated.isSuppressedBy(id: String): Boolean {
-	val valid = listOf(id, "ALL", "all")
+fun KtAnnotated.isSuppressedBy(id: String, aliases: List<String>): Boolean {
+	val valid = mutableListOf(id, "ALL", "all")
```
arturbosch: Please change this and the aliases to a hashset. 
The `it in valid` is costly for many findings :)

schalkms:
@vanniktech 
Good question! +1

PR #733, opened 2018-02-08
Add MayBeConst rule

### LINKED ISSUE #731: Style rule: val constants can be declared as const
IDEA introduced this inspection in 1.2.20 ([changelog](https://github.com/JetBrains/kotlin/blob/master/ChangeLog.md#new-features-5)) in https://youtrack.jetbrains.com/issue/KT-20644

The description of the quickfix:

> This inspection reports object and top-level val that might be declared as const for better performance and Java interoperability.

Here are a few examples and a counter example:

```kotlin
val topLevel = "TopLevel"

class MyClass {
    companion object {
        val compValue = "Companion"
    }
}

object MyObject {
    val compValue = "Companion"
}

fun myFun() {
    // const not applicable here, so should not be detected
    val localValue = "Local"
}
```

Mauin:
This resolves #731 (kind of)

This rule checks the most obvious cases where `val`s can be `const val`. There are more cases where expressions are used to initialize the `val`. However with type resolution this could be easily detected if they could be constants. Without type resolution it might be possible but I'm not sure if that is time well spent.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MayBeConst.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MayBeConst.kt
@@ -10,7 +10,6 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.rules.isConstant
 import io.gitlab.arturbosch.detekt.rules.isOverridden
 import org.jetbrains.kotlin.KtNodeTypes
-import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtObjectDeclaration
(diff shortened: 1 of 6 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MayBeConst.kt
```diff
@@ -0,0 +1,76 @@
...
+
+	override fun visitProperty(property: KtProperty) {
+		super.visitProperty(property)
+
+		if (property.canBeConst()) {
+			report(CodeSmell(issue, Entity.from(property), "${property.nameAsSafeName} can be a `const val`."))
+		}
+	}
+
+	private fun KtProperty.canBeConst(): Boolean {
```
schalkms: Can we split this function into smaller ones?

Issue #734, opened 2018-02-08
Extend EmptyCatchBlock


Tolriq:
First thanks for the tool :)

It would be cool if you could extend the EmptyCatchBlock rule to support regexp on the parameter name to allow easy handling of the normal cases where we want that empty catch due to known possible error that should just be ignored as not important and without possible fix.

PMD for java have the 2 following settings
```
    <rule ref="category/java/errorprone.xml/EmptyCatchBlock">
        <properties>
            <property name="allowExceptionNameRegex" value="^ignore$" />
            <property name="allowCommentedBlocks" value="true" />
        </properties>
    </rule>
```

And most IDE understand ignore as an indicator for that use case.

schalkms:
Closed via 7bba502

PR #736, opened 2018-02-08
Added allowedExceptionNameRegex config for EmptyCatchBlock

schalkms:
reference #734


schalkms:
@vanniktech
I do not think that detekt should have default values here. I dislike the fact that are empty catch blocks in a codebase. I understand that it might be necessary sometimes, but I would not add any defaults here. I think this decision should be up to the user.

vanniktech:
Usually when an exception is named like one of these the IDE also understands this and then does not flag it.

schalkms:
Ok. What do you think? @arturbosch

arturbosch:
I would go with the intellij default which is `ignored` like @vanniktech suggested ^^

vanniktech:
Let's also include `expected`? I use that often when writing tests.

Issue #611, opened 2017-12-14
Migrate to Gradle Kotlin DSL


3flex:
It would be cool to see the `.gradle` files migrate to Kotlin DSL. detekt can check its own Gradle config for issues.

PR #739, opened 2018-02-10
Migrate very first projects to gradle kotlin dsl

epeee:
A very first step towards #611.


epeee:
Hmm, the travis build is failing while running detekt on the code base (during processing newly added kts files). Excluding kts files from the analysis (in io.gitlab.arturbosch.detekt.core.KtTreeCompiler#isKotlinFile) resolves the problem. 
Should we create a ticket for the problem and exclude kts files for now from the analysis?

Mauin:
@epeee It looks like the Travis build succeeds in building the detekt `.jar`s. But then when testing the detekt codebase something goes wrong. If you build detekt locally and run 
`java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar -i . --baseline ./reports/baseline.xml -f ".*/resources/.*, .*/build/.*" -c ./detekt-cli/src/main/resources/default-detekt-config.yml,./reports/failfast.yml` you should be able to reproduce this.

epeee:
@Mauin thx! 
I was able to track it down to the newly added kts files (see my previous comment/question).

arturbosch:
Looks like there is a bug in LargeClass rule when analyzing kts files.
Have you tracked down the script file or is it happen for every script file?

Mauin:
Yeah, seems to be a bug with Kotlin script files in the `LargeClass` complexity rule. I'm taking a look at how to fix it.

Mauin:
I created #741 which should fix these issues. I ran the `build.gradle.kts` changes against my fixes and the build then fails for some real detekt violations. But at least the rules don't crash anymore 👍

epeee:
@Mauin thx! 
I will have a look at the real detekt violations once your pr is merged.

Mauin:
@epeee The PR is merged 🙃 Once you rebase you should be able to build the project and run detekt on the codebase. Let me know if you run into any other issues.

epeee:
@Mauin thx!
I just rebased and fixed the violations.
Sadly, travisci is heading into the same problem as already fixed in #741 on master (while running ./gradlew detektCheck using the latest version 1.0.0.RC6-3).
It looks like we have to publish a new version which contains the fix and make use of it in this branch in order to get a green travis build.

Mauin:
Eek. Right. That means the fixes in #741 definitely worked but we might need a new detekt release to get the build green. @arturbosch WDYT?

epeee:
Yep, let's wait for RC7.
Do you have a rough estimate about RC7?
Looking forward to check kts files using detekt in other projects too.

arturbosch:
Well not really, but personally I want to finish RC7 before march ^^

epeee:
I just rebased and applied recent changes from gradle scripts. Therefore this one is ready for review now.

arturbosch:
Looks and works great, even in IntelliJ :)

PR #743, opened 2018-02-11
add task inputs/outputs to detektCheck task

### LINKED ISSUE #725: The "detektCheck" Gradle task is never UP-TO-DATE
I just realized that running the exact same `./gradlew detektCheck` task two times in a row without changing any source code does not show the `detektCheck` task to be `UP-TO-DATE`, as expected, but runs the task again. Is that by design? If so, I'd vote for changing that for performance reasons, and because tasks can be easily forced to rerun with the `--rerun-tasks` [Gradle command line option](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:rerun_tasks).

Mauin:
This is a very basic implementation to make the `detektCheck` `UP-TO-DATE` when running on the exact same sources. 

However in the current setup that `detekt` uses for the `detektCheck` task we define `--input` as the root directory of the project, which will have changes after running the task and thus it won't be `UP-TO-DATE`. But if `input` and `output` are set to decent values it works.

Also the implementation is a bit hacky, looking at the list of arguments. This should be improved later.

This should resolve #725

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCheckTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCheckTask.kt
@@ -1,18 +1,37 @@
 package io.gitlab.arturbosch.detekt
 
 import io.gitlab.arturbosch.detekt.extensions.DetektExtension
+import io.gitlab.arturbosch.detekt.extensions.INPUT_PARAMETER
+import io.gitlab.arturbosch.detekt.extensions.OUTPUT_PARAMETER
 import org.gradle.api.DefaultTask
 import org.gradle.api.internal.artifacts.dependencies.DefaultExternalModuleDependency
+import org.gradle.api.tasks.InputDirectory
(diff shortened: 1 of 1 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCheckTask.kt
```diff
@@ -1,18 +1,38 @@
...
+	val input: File?
+
+	@OutputDirectory
+	val output: File?
+
 	init {
 		description = "Analyze your kotlin code with detekt."
 		group = "verification"
+
+		val detektExtension = project.extensions.getByName("detekt") as DetektExtension
```
schalkms: Unsafe cast?

Mauin: All other tasks also do `as DetektExtension`

arturbosch: yeah, an object is always returned ...

PR #745, opened 2018-02-14
Removed obsolete OptionalReturnKeyword rule

### LINKED ISSUE #722: OptionalReturnKeyword incorrect behavior
```
	OptionalReturnKeyword - [process] at ./mobile/src/main/java/be/mygod/vpnhotspot/Utils.kt:47:9
```

https://github.com/Mygod/VPNHotspot/blob/474018422c2a8f38ed8d21881b3bc613e916ec76/mobile/src/main/java/be/mygod/vpnhotspot/Utils.kt#L47

schalkms:
This resolves #722


vanniktech:
I don't quite understand why we would delete it. What's the cause?

schalkms:
@vanniktech because the test code is not valid Kotlin code anymore

PR #752, opened 2018-02-18
Fix in yaml syntax

mlegy:


arturbosch:
Thanks for the fix :)!

Issue #109, opened 2017-06-22
Renaming rule names for 1.0.0


arturbosch:
Tell me your renaming suggestions!!

So far renamed:
- NoDocOverPublicClass -> UndocumentedPublicClass
- NoDocOverPublicMethod -> UndocumentedPublicFunction

Mauin:
Hasn't this already been done with the commits above?

arturbosch:
Just for two rules. I want finalize all rule names for 1.0.0 ^^

vanniktech:
Also one potential candidate for a rename is the property `values` in `ForbiddenComment`.

arturbosch:
- CommentOverPrivateMethod -> DocumentedPrivateFunction
- CommentOverPrivateClass -> DocumentedPrivateProperty

andrejp88:
`exceptions` in `TooGenericExceptionCaught` and `TooGenericExceptionThrown` confused me a bit at first (are they the names of forbidden exceptions, or are they exceptions to the rule (i.e. not forbidden)?). Maybe renaming it to `throwable` or something similar would make it clearer.

arturbosch:
No more renaming suggestions so closing this.

PR #760, opened 2018-02-28
Change exceptions property to throwable - #109

arturbosch:
@andrejp88 suggested to rename the property as it might cause confusion.

For TooGenericExceptionThrown we have:
```
 * @configuration throwable - exceptions which are too generic and should not be thrown
 * (default: - Error
 * 			 - Exception
 * 			 - NullPointerException
 *			 - Throwable
 * 			 - RuntimeException)
```

I'm +1 to remove `NullPointerException` from that list as it is very specific and not generic. What do you think?

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -176,7 +176,7 @@ exceptions:
     active: false
   TooGenericExceptionCaught:
     active: true
-    exceptions:
+    exceptionNames:
      - ArrayIndexOutOfBoundsException
      - Error
      - Exception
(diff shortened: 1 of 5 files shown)


arturbosch:
@vanniktech  your right, what do you suggest? Maybe `exceptionNames`?

vanniktech:
Yeah that might work. But won't the rule itself also need a rename?

arturbosch:
`TooGenericThrowableCaught` ? That could work too.
So just renaming to `exceptionNames` or changing the documentation to `throwable` instead of `exeception`?

vanniktech:
I'd keep everything in sync either we're talking about Throwables or Exception but not mixed up.

arturbosch:
I've changed the property to `exceptionNames`. Also removed `NullpointerException` entry ^^

PR #762, opened 2018-03-01
update gradle to v4.6

Mauin:
release notes: https://docs.gradle.org/4.6/release-notes.html?utm_source=twitter&utm_medium=social&utm_campaign=4-6-release

This only bumps the version, but we can also look into using the new JUnit 5 support.


Mauin:
Huh, good spot. I updated it now. Not quite sure what went wrong with the first run of the command.

Issue #767, opened 2018-03-04
UnnecessaryParentheses false positive when copying with a function


vanniktech:
I've got the following data class:

```kotlin
data class Generator @JvmOverloads constructor(
  val name: String = "",
  val children: (ResolvedDependency) -> Boolean = { true },
)
```

I'm getting `UnnecessaryParentheses` flagged with the following: `instanceOfGenerator.copy(children = { false })`.

I can't omit the `{`, `}` around false.

schalkms:
The `()` are reported wrong here. It has nothing to do with the curly braces.

vanniktech:
Oh right. Curlys don't matter here.

schalkms:
I'm currently working on a fix.

schalkms:
Closed via 2f832bd

PR #770, opened 2018-03-04
Fixed report of unnecessary parentheses when assigning a lambda to a val

schalkms:
reference #767

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParentheses.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParentheses.kt
@@ -15,7 +15,7 @@ import org.jetbrains.kotlin.psi.KtValueArgument
 
 /**
  * This rule reports unnecessary parentheses around expressions.
- * This unnecessary parentheses can safely be removed.
+ * These unnecessary parentheses can safely be removed.
  *
  * Added in v1.0.0.RC4
  *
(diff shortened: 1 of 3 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParentheses.kt
```diff
@@ -42,7 +43,8 @@ import org.jetbrains.kotlin.psi.KtValueArgument
 class UnnecessaryParentheses(config: Config = Config.empty) : Rule(config) {
 
 	override val issue = Issue("UnnecessaryParentheses", Severity.Style,
-			"Unnecessary parentheses don't add any value to the code and should be removed.")
+			"Unnecessary parentheses don't add any value to the code and should be removed.",
```
vanniktech: Probably even less than 5mins.

schalkms: That's the minimum detekt currently has.

docs/pages/documentation/style.md
```diff
@@ -630,7 +630,8 @@ class B : Object()
 
 ### UnnecessaryParentheses
 
-Reports unnecessary parentheses around expressions.
+This rule reports unnecessary parentheses around expressions.
+This unnecessary parentheses can safely be removed.
```
vanniktech: Should be: These parentheses

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParentheses.kt
arturbosch: when you count git add/commit and referencing an issue, it should be min. 5 minutes :D

Issue #580, opened 2017-11-29
The Gradle plugin tasks should extend SourceTask


rock3r:
Taking a note from how Gradle's own code quality plugins are done, the Detekt Gradle tasks should inherit from [`SourceTask`](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/SourceTask.html), a specialised subclass of `DefaultTask` (what it's using now). The validation tasks (e.g., `detektCheck`) should also implement `VerificationTask` and `Reporting<DetektReports>` (where `DetektReports` extends `ReportContainer<SingleFileReport>`) to align to the built-in Gradle Checkstyle/PMD/Findbugs/... plugins.

Thoughts?

3flex:
I 100% agree, and had started doing a bit of work to modernise the Gradle integration to better align with the code quality plugins. I put it on hold because I'm working on making detekt work as a Kotlin compiler plugin (so it hooks into the compilation process in a similar way to Error Prone). Making this work requires substantial changes to the Gradle integration. I should have something to share in a day or two, I was actually working on the Gradle integration part yesterday.

It probably makes sense to wait for that before changing the DSL and updating the Gradle plugin internals, because there might be other internal changes to be made to support that.

3flex:
A couple of other interesting classes: CodeQualityExtension, VerificationTask, Reporting. AbstractCodeQualityPlugin is probably useful but is in a package marked "internal" so probably shouldn't be used.

arturbosch:
Yep, agree both @rock3r @3flex . This issue is linked to #540 and #542 . Rework is planned after @3flex 's work.

arturbosch:
I do think `AbstractCodeQualityPlugin` is the way to go even if it is internal. I mean what could possible go wrong?
But I do not see the benefit implemening any of this classes yet ...
I have to override so many methods and the api is like ''I Have No Idea What I'm Doing'' ...

Mauin:
Closing this as the reworked plugin has been merged.

Issue #540, opened 2017-11-11
Gradle Plugin: Behavior of profiles


Mauin:
Spawned from the discussion in #536

Currently I feel like two things are weird in the way profiles are defined and used in the `detekt` configuration in the gradle plugin.

1. Only the profile `"main"` is run by default.
We let the user enter a free string for the profile name, but in the end the Gradle Plugin will only search for and run the `main` profile. If there is no `main` profile the fallback will be run. Profiles with different names can be run through a `profile: <profilename>` variable in the `detekt` closure or via a command line gradle flag. Both of these seem cumbersome to me. And realizing that the fallback is used might not be obvious to users.

2. Even if multiple profiles are defined only the `main` profile is run. Other profiles can only be run with the above mentioned configuration methods.

I suggest to change this behavior to run _any_ profile that was defined in the `detekt` plugin configuration regardless of the name. If no profile was defined, we still fall back to the default fallback values.
This means multiple profiles can be run at the same time with the `detektCheck` command.
Should the user want to run a single profile the configuration options to add `profile: <profilename>` to the plugin configuration would still work. As would the command line parameter.

Additionally we could generate new tasks for each profile:
`detektCheck` - runs all profiles defined in the configuration (in this example `"sample"` and `"test"`)
`detektCheckSample` - runs the profile with the name `"sample"`
`detektCheckTest` - runs the profile with the name `"test"`

3flex:
PMD, Checkstyle and FindBugs plugins for Gradle all use sourcesets to auto-generate tasks, for example:
- findbugsMain - Runs FindBugs against the production Java source files.
- findbugsTest - Runs FindBugs against the test Java source files.
- findbugsSourceSet - Runs FindBugs against the given source set’s Java source files

It would be good to see detekt move in that direction as well for consistency with other code quality tools.

So I'd suggest any auto-generation of tasks should be based on configured source sets, not on the detekt profile names. Users could then select a profile using `./gradlew detektCheck -Ddetekt.profile=override`.

The tasks would be:
- detektCheck - Runs all detekt tasks, including detektCheckMain and detektCheckTest, plus any automatically generated for other source sets.
- detektCheckMain - Runs detekt against the production Kotlin source files.
- detektCheckTest - Runs detekt against the test Kotlin source files.
- detektCheckSourceSet - Runs detekt against the given source set’s Kotlin source files.

[...]

PR #771, opened 2018-03-04
Implement base Gradle StaticAnalysis classes for Gradle Plugin

Mauin:
This is still WIP but publishing for feedback.

As described in Issues #580 and #540 this PR makes the Detekt Gradle Plugin implement the base StaticAnalysis classes `SourceTask` and `AbstractCodeQualityPlugin`

This work will result in breaking API changes of the Detekt Gradle Plugins API as the Plugin will now behave more like the PMD, Checkstyle, Findbugs and similar static analysis plugins.

This work will also result in detekt running as part of `./gradlew check` and therefore `./gradlew build` if the plugin is added to the project.

I had to make minor changes to account for the new way to have inputs based on source sets and not explicitly giving an input path in the API. With these changes the Gradle Plugin will automatically generate tasks for running detekt for each sourceSet: `detektMain`, `detektTest`, etc. Those sourceSets will be the inputs for the detekt CLI.

Due to the change how output reports are defined in the same manner as other tools like PMD and Checkstyle do in their Gradle API there are some more changes needed to detekt-cli and detekt-core to account for enabling disabling certain reports and changing their output paths with the CLI arguments.

The changes to the API are visible in the top-level `build.gradle` file in this PR. Essentially I believe this will make the concept of `profiles` go away as it is really easy to define a second detekt task with some special parameters (see the `detektFailfast` task in the `build.gradle` file).

Missing:
- [x] Update documentation when implementation is finished
- [x] ~~Decide on profile handling.~~ Removed
- [x] Implement `detektCreateBaseline` and `detektGenerateConfig` tasks.
- [x] Implement `idea` tasks.
- [x] Create `detektCheck` task which delegates to the main detekt task to provide backwards compatibility


arturbosch:
Looking forward to review this later! :) :+1:

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -1,31 +1,111 @@
...
+		get() = extension as DetektExtension
 
-		project.extensions.create(DETEKT_EXTENSION_NAME, DetektExtension::class.java)
-		project.task(mapOf(TYPE to DetektCheckTask::class.java), CHECK)
+	override fun getToolName() = DETEKT_EXTENSION_NAME
+	override fun getTaskType() = Detekt::class.java
+
+	override fun createExtension(): CodeQualityExtension {
+		val extension = project.extensions.create(DETEKT, DetektExtension::class.java, project)
+		extension.toolVersion = "1.0.0.RC6-MARVIN"
```
arturbosch: :D

arturbosch: Do you know why the build fails? (locally too)

Mauin: Because the current `build.gradle` setup is incompatible with the old Gradle Plugin. To build a working version you essentially have to comment out all the detekt plugin related things from `build.gradle`, publish everything locally under the version you want. Then comment it back in and it should then run 👍

Mauin:
As for the reports. Currently we configure them via the `config.yml`. However the checkstyle reports in the Gradle Plugin would allow us to independently enable/disable and change the path for the different XML/HTML reports. To fully comply with this we need to find some way in `-cli` to set the path for the XML/HTML report directly. We can also leave the old `output` parameters in the CLI. However then CLI and Gradle Plugin would diverge a bit in how they are used. I'm open for suggestions and improvements.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtTreeCompiler.kt
```diff
@@ -6,24 +6,30 @@ import java.nio.file.Path
...
 		fun instance(settings: ProcessingSettings) = with(settings) {
-			KtTreeCompiler(KtCompiler(), pathFilters, parallelCompilation)
+			KtTreeCompiler(KtCompiler(), pathFilters, parallelCompilation, settings.debug)
 		}
 	}
 
-	fun compile(project: Path): List<KtFile> {
-		require(Files.exists(project)) { "Given project path does not exist!" }
+	fun compile(path: Path): List<KtFile> {
+		require(Files.exists(path)) { "Given path path does not exist!" }
```
schalkms: duplicate `path`

Mauin:
Just realized that the default `CodeQualityExtension` has a `toolsVersion` field. That's also the way Checkstyle and friends set the version through the Gradle Plugin. So I also removed our custom `version` field to instead use `toolVersion`.

arturbosch:
I agree that we should be compliant with the reports feature.
How about we enable custom reports only through the configuration file? Like:
```
output-reports:
  active: true
  MyReport:
    active: true
    path: reports/myreport.txt
```
We can do this after this PR then. What do you think?

Mauin:
@arturbosch So effectively leaving the way I currently implemented it in this PR with the `--report-xml` and `--report-html` arguments and removing the handling of Reports via the config? I'll get right on that.

arturbosch:
Yes, I will be able to test everything on the WE when I have internet again.

Mauin:
Okay, feature-wise I think I have everything covered. Before updating all the documentation I'd love some feedback on how the Gradle Plugin API will look like after these changes.

See: https://github.com/arturbosch/detekt/pull/771/files#diff-c197962302397baf3a4cc36463dce5eaR41

build.gradle
```diff
@@ -34,14 +32,56 @@ allprojects {
...
 	apply plugin: 'maven-publish'
 	apply plugin: 'com.jfrog.bintray'
+	apply plugin: 'io.gitlab.arturbosch.detekt'
+
+	detekt {
+		debug = true
+		parallel = true
+		toolVersion = usedDetektGradleVersion
+
+		config = project.resources.text.fromFile("${rootProject.projectDir}/detekt-cli/src/main/resources/default-detekt-config.yml")
```
arturbosch: Is this because of foreach subproject? hmmm that looks odd :/
Can't we just write file(...)?

build.gradle
```diff
@@ -34,14 +32,56 @@ allprojects {
...
+			}
+		}
+	}
+
+	task detektFailfast(type: io.gitlab.arturbosch.detekt.Detekt) {
+		description = "Runs a failfast detekt build."
 
-	if (it.name in ['detekt-cli', 'detekt-watch-service', 'detekt-generator']) {
+		source = sourceSets.getAt("main").allSource
+		config = project.resources.text.fromFile("${rootProject.projectDir}/reports/failfast.yml")
```
arturbosch: detekt was configured to always use the failfast config together with the default one.
In my last DetektExtension changes `files(config1, config2)` was supported, is it still?

build.gradle
Mauin: `config` itself is not a `File` but a `TextResource`. I have added a commit that adds the possibility to set the config via a `File` with the `configFile` property (similar to Checkstyle)

build.gradle
Mauin: No, each `detekt` task can only have one config associated with it. What would detekt do before when it was passed multiple config files? Did we merge them and then run detekt-cli with it?

Mauin:
Just pinging @tasomaniac and @rock3r on this as well to get some feedback on the new behavior and DSL.

README.md
```diff
@@ -208,64 +182,70 @@ plugins {
...
+    plugins = "other/optional/ruleset.jar"                           // Jar file containing custom detekt rules.
+}
+```
+
+##### <a name="gradlepluginreports">Customizing Detekt reports</a>
+
+You can configure the reports detekt outputs with the following configuration in your `build.gradle` file:
+
+```groovy
+tasks.withType(io.gitlab.arturbosch.detekt.Detekt) {
```
tasomaniac: I've bitten by this kind of setup before. Gradle advocates for separating the input/setup that is seen by the user and the actual task.
This decoupling is really helpful since it allows you to be more flexible in the task level and allows Gradle to be more effective when it comes to up to date checks.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/OutputFacade.kt
```diff
@@ -34,16 +36,19 @@ class OutputFacade(private val arguments: Args,
...
+		arguments.htmlReport?.let {
+			HtmlOutputReport().write(File(it).toPath(), result)
+		}
+
+		arguments.xmlReport?.let {
+			XmlOutputReport().write(File(it).toPath(), result)
+		}
+
+		arguments.plainReport?.let {
+			XmlOutputReport().write(File(it).toPath(), result)
```
tasomaniac: Is plain report really the xml? I remember Detekt also had plain output. It was just a txt file with no structure. Just a dump of issues one by one.

Mauin: Ooops. You're right. This looks like a copy/paste error.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	 * The Detekt configuration file to use.
+	 */
+	fun setConfigFile(configFile: File) {
+		config = project.resources.text.fromFile(configFile)
+	}
+
+	fun configureForSourceSet(sourceSet: SourceSet) {
+		description = "Run detekt analysis for ${sourceSet.name} classes"
```
tasomaniac: What about `group = "verification"`?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -1,31 +1,100 @@
...
+
+	override fun createExtension(): CodeQualityExtension {
+		val extension = project.extensions.create(DETEKT, DetektExtension::class.java, project)
+		extension.toolVersion = "latest.release"
+		extension.configDir = project.rootProject.file("detekt-cli/src/main/resources/")
+		extension.config = project.resources.text.fromFile {
+			File(extension.configDir, "default-detekt-config.yml")
+		}
+
+		generateConfigTask = project.task(mapOf(org.gradle.api.Task.TASK_TYPE to DetektGenerateConfigTask::class.java), GENERATE_CONFIG) as DetektGenerateConfigTask
```
tasomaniac: I believe there is also `project.tasks.create` which is much more Kotlin friendly. There you have the task class as generics which also is the return type. You don't have to cast at the end either.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -1,31 +1,100 @@
...
+	}
+
+	override fun configureTaskDefaults(task: Detekt, baseName: String) {
+		val configuration = project.configurations.getAt(DETEKT)
+		configureDefaultDependencies(configuration)
+		configureTaskConventionMapping(task)
+		configureReportsConventionMapping(task, baseName)
+	}
+
+	private fun configureTaskConventionMapping(task: Detekt) {
```
tasomaniac: `ConventionMapping` is not really documented and also not really kotlin friendly. 
What about using new `Property` and `Provider` APIs for gradle plugins.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -2,87 +2,43 @@ package io.gitlab.arturbosch.detekt.extensions
...
-		} else {
-			ProfileExtension(name).apply {
-				ProfileStorage.add(this)
-				configuration.execute(this)
-			}
-		}
+	/**
+	 * The Detekt configuration file to use.
+	 */
+	@Internal
```
tasomaniac: `Internal` is for tasks. It may not change anything at all for an extension.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -2,87 +2,43 @@ package io.gitlab.arturbosch.detekt.extensions
...
  */
-open class DetektExtension(open var version: String = SUPPORTED_DETEKT_VERSION,
-						   open var debug: Boolean = DEFAULT_DEBUG_VALUE,
-						   open var profile: String = DEFAULT_PROFILE_NAME,
-						   open var ideaExtension: IdeaExtension = IdeaExtension()) {
-
-	fun ideaFormatArgs() = ideaExtension.formatArgs(this)
-	fun ideaInspectArgs() = ideaExtension.inspectArgs(this)
+open class DetektExtension(val project: Project) : CodeQualityExtension() {
+	open var debug: Boolean = DEFAULT_DEBUG_VALUE
```
tasomaniac: Just a question. It does not look like clear to me. What does `debug` mean here?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
rock3r: I think `ConventionMapping` is an internal API, we had issued with it in one of our plugins. Better to stick to the public `Property` API.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
rock3r: I think it's the corresponding CLI flag to print debug info. May be wrong tho.

rock3r:
CC @tobiasheine for planning evolution of the Novoda Static Analysis plufgin

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
rock3r: Not entirely sure why you want it to be internal, too, it seems a useful and safe API to have (e.g., to collate config files at the end of a build and attach them to the reports)

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
```diff
@@ -0,0 +1,86 @@
...
+		argumentList += CREATE_BASELINE_PARAMETER
+
+		invokeCli(project, classpath, argumentList.toList())
+	}
+
+	fun generateConfig(detekt: Detekt) {
+		val project = detekt.project
+		val classpath = project.configurations.getAt("detekt")
+
+		val args = mutableMapOf<String, String>(
```
rock3r: Does this need to be mutable? You only use it in a flatMap AFAICT

docs/pages/gettingstarted/groovydsl.md
```diff
@@ -1,22 +1,32 @@
...
 summary:
 ---
 
-- `detektCheck` - Runs a _detekt_ analysis. Configure the analysis inside the `detekt-closure`. By default the standard rule set is used without output report or  black- and whitelist checks.
-- `detektGenerateConfig` - Generates a default detekt config file within your projects location.
-- `detektBaseline` - Like `detektCheck`, but creates a code smell baseline. Further detekt runs will only feature new issues. 
+#### <a name="tasks">Available plugin tasks</a>
+
+The detekt Gradle plugin will generate `detekt` tasks for each of your source sets. For a basic project this will result
+in a `detektMain` task which will check all main source code of the project. The `detektTest` task will run detekt on
```
rock3r: Here too, I'd replace `all main source code` with ``all the `main` sourcesets`` (ditto for the `test` on the next line)

docs/pages/gettingstarted/groovydsl.md
```diff
@@ -1,22 +1,32 @@
...
+the test sources of the project
+
+- `detekt[SourceSet]` - Runs a _detekt_ analysis and complexity report on the given source set. Configure the analysis inside the `detekt` closure. By default the standard rule set is used without output report or black- and whitelist checks.
```
rock3r: This is the only place where `detekt` is italicised, is it a convention? If so, can we fix elsewhere? If no, can we not italicise?

docs/pages/gettingstarted/groovydsl.md
```diff
@@ -1,22 +1,32 @@
...
 - `detektIdeaFormat` - Uses a local `idea` installation to format your kotlin (and other) code according to the specified `code-style.xml`.
 - `detektIdeaInspect` Uses a local `idea` installation to run inspections on your kotlin (and other) code according to the specified `inspections.xml` profile.
 
-For new gradle versions >= 2.1:
- 
-```gradle
+Use the Groovy or Kotlin DSL of Gradle to apply the detekt Gradle Plugin. You can further configure the Plugin
+using the detekt closure as described [here](#closure).
+
+##### <a name="gradlegroovy">Configuration when using groovy dsl</a>
```
rock3r: `groovy dsl` -> `Groovy DSL`?

docs/pages/gettingstarted/groovydsl.md
```diff
@@ -54,70 +52,100 @@ buildscript {
...
-detekt {
-    version = "1.0.0.[version]"
-    defaultProfile {
-        input = file("src/main/kotlin")
-        config = file("detekt.yml")
-        filters = ".*/resources/.*,.*/tmp/.*"
-        output = file("reports")
-        outputName = "detekt-report"
-        baseline = file("reports/baseline.xml")
+##### <a name="gradlekotlin">Configuration when using kotlin dsl</a>
```
rock3r: `kotlin dsl` -> `Kotlin DSL`?

docs/pages/gettingstarted/groovydsl.md
```diff
@@ -54,70 +52,100 @@ buildscript {
...
+plugins {
+    id("io.gitlab.arturbosch.detekt").version("1.0.0.[version]")
+}
 ```
 
-### Options for detekt configuration closure
+##### <a name="gradleandroid">Configuration for Android projects</a>
+
+When using Android make sure to have detekt configured in the project level build.gradle file.
+The new preferred plugin configuration way is used, the old way is commented out.
```
rock3r: This seems to be quite confusing, could we split out the "before" and "after"?

build.gradle
arturbosch: There is the class `CompositeConfig` which uses the left config before using the right.
When we have two config files - the default config and the failfast config - detekt first searches into== the failfast config for overridden properties and fall backs to the second config file.
I would say this is a key feature we should keep.

vanniktech:
I'd like to review this. Would it be possible to have one last review round once this is no longer WIP?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
arturbosch: Yes, it prints internal detekt config state to the console (and before this changes which profiles were declared and which one used)

Mauin:
I'll go over the comments in the coming days and update everything. Thanks for all the feedback!

README.md
```diff
@@ -208,64 +182,70 @@ plugins {
...
+    config = project.resources.text.fromFile("path/to/config.yml")   // Define the detekt configuration you want to use.
+    configFile = file("path/to/config.yml")                          // Define the detekt configuration you want to use.
+    baseline = file("path/to/baseline.xml")                          // Specifying a baseline file will ignore all findings that are saved in the baseline file.
+    filters = ''                                                     // Regular expression of paths that should be excluded.
+    disableDefaultRuleSets = false                                   // Disables all default detekt rulesets and will only run detekt with custom rules defined in `plugins`.
+    plugins = "other/optional/ruleset.jar"                           // Jar file containing custom detekt rules.
+}
+```
+
+##### <a name="gradlepluginreports">Customizing Detekt reports</a>
```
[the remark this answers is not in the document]
arturbosch: Good question, I always tend to stick to detekt (also with a lower k :))

README.md
```diff
@@ -208,64 +182,70 @@ plugins {
...
-- `detektGenerateConfig` - Generates a default detekt config file into your projects location.
-- `detektBaseline` - Like `detektCheck`, but creates a code smell baseline. Further detekt runs will only feature new smells not in this list. 
+The detekt Gradle plugin will generate `detekt` tasks for each of your source sets. For a basic project this will result
+in a `detektMain` task which will check all main source code of the project. The `detektTest` task will run detekt on
+the test sources of the project
+
+- `detekt[SourceSet]` - Runs a _detekt_ analysis and complexity report on the given source set. Configure the analysis inside the `detekt` closure. By default the standard rule set is used without output report or black- and whitelist checks.
+- `detektGenerateConfig` - Generates a default detekt configuration file into your project directory.
+- `detektBaseline` - Similar to `detekt[SourceSet]`, but creates a code smell baseline. Further detekt runs will only feature new smells not in this list.
 - `detektIdeaFormat` - Uses a local `idea` installation to format your kotlin (and other) code according to the specified `code-style.xml`.
```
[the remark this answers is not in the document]
arturbosch: and other code like xml files or whatever intellij can detect

build.gradle
Mauin: I'm not really sure. For one this is an entirely different behavior from other static analysis tools. Also it makes the plug-in behave more like a black box and it is quite hard to figure out. I personally think having clear configs that run a very defined set of rules is much clearer than some merging logic in the plug-in. I'm open for discussion for sure but I personally prefer the clarity of not doing any merging/overriding in the plug-in

README.md
```diff
@@ -208,64 +182,70 @@ plugins {
...
-}
 ```
 
 #### <a name="tasks">Available plugin tasks</a>
 
-- `detektCheck` - Runs a _detekt_ analysis and complexity report. Configure the analysis inside the `detekt-closure`. By default the standard rule set is used without output report or  black- and whitelist checks.
-- `detektGenerateConfig` - Generates a default detekt config file into your projects location.
-- `detektBaseline` - Like `detektCheck`, but creates a code smell baseline. Further detekt runs will only feature new smells not in this list. 
+The detekt Gradle plugin will generate `detekt` tasks for each of your source sets. For a basic project this will result
+in a `detektMain` task which will check all main source code of the project. The `detektTest` task will run detekt on
```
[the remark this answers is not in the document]
arturbosch: Please make sure to update the website, the readme will only have quick start information etc

build.gradle
```diff
@@ -34,14 +32,56 @@ allprojects {
...
+		configFile = file("${rootProject.projectDir}/detekt-cli/src/main/resources/default-detekt-config.yml")
+		baseline = file("${rootProject.projectDir}/reports/baseline.xml")
+		filters = '.*/resources/.*, .*/build/.*'
+
+		idea {
+			path = "$USER_HOME/.idea"
+			codeStyleScheme = "$USER_HOME/.idea/idea-code-style.xml"
+			inspectionsProfile = "$USER_HOME/.idea/inspect.xml"
+			report = "project.projectDir/reports"
+			mask = "*.kt,"
```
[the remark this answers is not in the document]
arturbosch: nope I think that indicates that you can have many paths

mkobit:
FYI, if you wanted to add some tests (only briefly glanced over this) is to use the [Gradle Test Kit](https://docs.gradle.org/current/userguide/test_kit.html) which is fairly easy to use with the [Gradle Plugin Development Plugin](https://docs.gradle.org/current/userguide/java_gradle_plugin.html)

README.md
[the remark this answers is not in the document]
rock3r: Ah I see, so it's more IDEA's support. Thanks :)

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -0,0 +1,106 @@
...
+import org.gradle.api.tasks.VerificationTask
+import java.io.File
+import javax.inject.Inject
+
+/**
+ * @author Artur Bosch
+ * @author Marvin Ramin
+ */
+@CacheableTask
+open class Detekt : SourceTask(), VerificationTask, Reporting<CheckstyleReports> {
```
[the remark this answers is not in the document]
Mauin: Similar to Checkstyle, PMD and others which have their main tasks simply named after the tool.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -0,0 +1,106 @@
...
+	 * @since 4.2
+	 */
+	@Incubating
+	@Inject
+	open fun getObjectFactory(): ObjectFactory {
+		throw UnsupportedOperationException()
+	}
+
+	@OutputFiles
+	fun getOutputFiles(): Map<String, File> {
```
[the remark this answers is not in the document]
Mauin: As we're using the same outputs as Checkstyle does at the moment with these changes we only support XML/HTML. We could also add a custom Report implementation which allows for the plain report. WDYT?

docs/pages/gettingstarted/groovydsl.md
Mauin: @arturbosch Do we even care about this part of the documentation? In the end this is the same as in the paragraphs above.. Plus Android devs would (should) be used to how adding Gradle Plugins works.

Mauin:
Updated the PR with all the feedback and using the Gradle Property API.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtTreeCompiler.kt
```diff
@@ -6,24 +6,30 @@ import java.nio.file.Path
...
-	fun compile(project: Path): List<KtFile> {
-		require(Files.exists(project)) { "Given project path does not exist!" }
+	fun compile(path: Path): List<KtFile> {
+		require(Files.exists(path)) { "Given path $path does not exist!" }
 		return when {
-			project.isFile() && project.isKotlinFile() -> listOf(compiler.compile(project, project))
-			project.isDirectory() -> compileInternal(project)
-			else -> throw IllegalArgumentException("Provided project path $project is not a file/dir." +
-					" Detekt cannot work with it!")
+			path.isFile() && path.isKotlinFile() -> listOf(compiler.compile(path, path))
```
schalkms: Does `isKotlinFile()` not check for `isFile()` ?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektIdeaInspectionTask.kt
```diff
@@ -1,24 +1,33 @@
...
 	fun inspect() {
-		with(project.extensions.getByName("detekt") as DetektExtension) {
-			if (debug) println("$ideaExtension")
-			startProcess(ideaInspectArgs())
+		if (ideaExtension == null) {
+			throw GradleException("idea extension is not defined. It is required to run detekt idea tasks.")
 		}
+
+		if (detekt.debug.get()) Log.info("$ideaExtension")
+		ProcessExecutor.startProcess(ideaExtension!!.formatArgs(detekt.source.asFileTree.asPath))
```
schalkms: minor code duplication (`DetektIdeaFormatTask.kt`) ?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtTreeCompiler.kt
Mauin: No, it does not check if the file is an actual file but assumes that it is. The only thing `isKotlinFile` does is check the file ending. (`.kt` or `.kts`)

Mauin:
Currently seeing 2 issues when running the `detektMain` task:

- `detekt-api` cannot find the `Yaml` class. (`NoClassDefFoundError`)
- `detekt-gradle-plugin` throws `NoSuchMethodError` on `constructor.visibilityModifierType()` in `EmptyDefaultConstructor` rule.

Might be my local config. I'm still looking but if anyone has seen these issues before, let me know 🙃

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtTreeCompiler.kt
tasomaniac: It's out of scope here, but it would make sense to either check if it is file or to rename it to say `hasKotlinExtension`

arturbosch:
heyho, if you rebase everything, I can take a look at the two remaining issues too :).
Lets do a freeze until this is done

Mauin:
@arturbosch I'll try to quickly rebase the branch tomorrow and fix the conflicts. As for the issues themselves I still haven't gotten closer to identifying the issue yet.

markov:
@Mauin Thanks for doing this.

As for the `NoSuchMethodError` - I think it is something that changed in Kotlin between v1.2.21 and v1.2.31.

I am still using detekt RC5-6 and if I force everything in my project to use the latest kotlin by adding this to my root `build.gradle` file:

    configurations.all {
        resolutionStrategy {
            eachDependency { details ->
                switch(details.requested.group) {
                    case 'org.jetbrains.kotlin':
                        details.useVersion '1.2.31'
                        break
                }
            }
        }
    }

Then I see the same error.

rock3r:
Hey there, is there any updates on this? I understand it's been a very busy couple of months for all of us but I'd like to see this moved forward so we can properly support Detekt again in our static analysis plugin, which is currently broken. Let me know if we can assist in any way!

Mauin:
I'm still chipping away at this but have run into multiple issues on the Gradle level recently. As we switched to using the Kotlin DSL in detekt I noticed multiple issues with the Plugin rework that I had to address and also found blockers in Testing the plugin with Gradle v4.7 and am waiting for a Gradle v4.8 release to continue the work in testing the Plugin properly. I'm looking at pushing an updated version of the re-work soon(TM) but the iteration process on the Plugin is suuuper slow unfortunately...

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/PathConverters.kt
```diff
@@ -26,22 +27,22 @@ class PathConverter : IStringConverter<Path> {
 	}
 }
 
-interface CommaSeparatedStringConverter<T> : IStringConverter<List<T>> {
+interface DetektInputPathConverter<T> : IStringConverter<List<T>> {
 	val converter: IStringConverter<T>
 	override fun convert(value: String): List<T>
-			= value.splitToSequence(SEPARATOR_COMMA, SEPARATOR_SEMICOLON)
+			= value.splitToSequence(SEPARATOR_COMMA, SEPARATOR_SEMICOLON, SEPARATOR_COLON)
```
3flex: Unfortunately this won't work on Windows.

C:\path\to\project is split when it shouldn't be.

3flex:
@Mauin you mentioned some work that's partly done but not pushed yet - is that in a different branch? I'd like to help with this, particularly on the Gradle side (I started down this path some time back but got held up when I started looking at the changes required on the detekt side to make it work properly. But a lot of what I did could be migrated here since it looks like you've made the necessary changes in detekt which is fantastic!). I found it a real pain to work with initially, Gradle is a beast.

Good news on Gradle's side is there's a 4.8-rc-1 out now with Gradle TestKit support for projects using Kotlin DSL. You mentioned problems with testing, perhaps this will help?

Let me know how I can assist. It would be great to get this one over the line!

Mauin:
Pushed an update to the branch with now a Kotlin DSL compliant DSL including some Tests for it. Groovy Gradle DSL Tests are still missing. 
This also uses the Gradle v4.8-rc-1, which makes the Kotlin DSL testing possible, but we're still suffering from the classpath issue described in #913.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/PathConverters.kt
3flex: This fails on Windows, where paths are things like C:\path\to\project. Splitting on `:` gives you something like `['C', '\path\to\project']`

Mauin: Huh. What will Gradle report for the FileTree of the SourceSet? On Unix/Mac it'll report `/path/to/sourceset/file1:/path/to/sourceset/file2` and so on.

Mauin: Alright, then it looks like we'll have to make this list dynamic depending on the OS we're running on.

3flex:
This is great! I hope I didn't miss something major related to the classpath issues you told me about, because it seems I have it working with some very minor changes in build.gradle.kts (as long as we ignore the Windows path separator issue for now!):
* Remove this line: `classpath("io.gitlab.arturbosch.detekt:detekt-gradle-plugin:$usedDetektGradleVersion")`
* Add this line to `plugins{}` block: `id("io.gitlab.arturbosch.detekt") version "1.0.0.RC7-MH"` (change the version as appropriate)

This is based on [this documentation](https://github.com/gradle/kotlin-dsl/blob/master/doc/getting-started/Configuring-Plugins.md) from the Kotlin DSL project, which says:
> When using the Gradle Kotlin DSL it is heavily recommended to apply Gradle plugins declaratively using the plugins {} block. This will enable type-safe extension accessors you will use to configure plugins.

Also, because I'd pushed to mavenLocal, I had to add this in settings.gradle.kts, because the Gradle Plugin repo is the only one used by default for items in the `plugins {}` block:
```
pluginManagement {
	repositories {
		gradlePluginPortal()
		mavenLocal()
	}
}
```

See https://github.com/3flex/detekt/tree/mr/gradle_classes for working example.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/PathConverters.kt
3flex: Gradle's implementation uses java.io.File.pathSeparator which is semi-colon on Windows and colon on *nix (everything else).

@arturbosch since this PR makes breaking changes to the CLI, I suggest using only File.pathSeparator to support splitting provided paths on the CLI, thoughts?

Mauin: The separators we supported before were mainly for manual input definitions in the `build.gradle` file. Supporting full sourceSets is new with this updated Gradle Plugin, so I think we should add using the API you mentioned.

arturbosch:
Please make sure to rebase this on top of current master as we may have two commits here which I dropped on master.

build.gradle.kts
```diff
@@ -62,6 +63,27 @@ subprojects {
...
+		baseline = file("${rootProject.projectDir}/reports/baseline.xml")
+		configFile = file("${rootProject.projectDir}/detekt-cli/src/main/resources/default-detekt-config.yml")
+		toolVersion = usedDetektGradleVersion
+
+		idea(Action {
+			path = "$userHome/.idea"
+			codeStyleScheme = "$userHome/.idea/idea-code-style.xml"
+			inspectionsProfile = "$userHome/.idea/inspect.xml"
+			report = "project.projectDir/reports"
+			mask = "*.kt"
```
3flex: Since we're changing the config options - I'm not sure `mask` has any effect? It's passed to the "-m" option of the `inspect.bat`/`inspect.sh` script by the plugin, but that's [not a documented option](https://www.jetbrains.com/help/idea/running-inspections-offline.html)...

I'll have a play with this and will open a separate issue to remove it if it's not actually used, I just think it best to lock down the config API before a 1.0.0 release.

build.gradle.kts
```diff
@@ -62,6 +63,27 @@ subprojects {
...
+		filters = ".*/resources/.*, .*/build/.*"
+		baseline = file("${rootProject.projectDir}/reports/baseline.xml")
+		configFile = file("${rootProject.projectDir}/detekt-cli/src/main/resources/default-detekt-config.yml")
+		toolVersion = usedDetektGradleVersion
+
+		idea(Action {
+			path = "$userHome/.idea"
+			codeStyleScheme = "$userHome/.idea/idea-code-style.xml"
+			inspectionsProfile = "$userHome/.idea/inspect.xml"
+			report = "project.projectDir/reports"
```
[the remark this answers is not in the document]
Mauin: True. I didn't really touch the Idea config much.

build.gradle.kts
```diff
@@ -62,6 +63,27 @@ subprojects {
...
+		plugin("io.gitlab.arturbosch.detekt")
+	}
+
+	val userHome = System.getProperty("user.home")
+	val usedDetektGradleVersion: String by project
+
+	detekt {
+		debug = true
+		parallel = true
+		filters = ".*/resources/.*, .*/build/.*"
```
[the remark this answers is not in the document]
Mauin: I like this idea. Do you want to look into this? Otherwise I'll try to find some time.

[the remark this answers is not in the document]
tasomaniac: Would it make sense to create an issue about this? Or would it make sense to change it within this PR?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -2,101 +2,74 @@ package io.gitlab.arturbosch.detekt.extensions
...
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.plugins.quality.CodeQualityExtension
+import org.gradle.api.provider.Property
+import org.gradle.api.resources.TextResource
+import org.gradle.kotlin.dsl.property
+import java.io.File
 
 /**
  * @author Artur Bosch
  * @author Said Tahsin Dane
```
tasomaniac: Haha :) you can also remove me if you want.

Mauin:
I merged that latest changes from Master. Now running `gradle detektMain` passes successfully! Thanks for all the work @3flex. 
Now I need to finish looking at the tests. But they seem to fail for valid reasons now. Looks like we need to generate more temporary files in the test build.

build.gradle.kts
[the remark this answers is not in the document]
3flex: I've just tested, and because the `Detekt` task extends `SourceTask`, we got this for free! It already works, so can remove the `filters` option. Configure like so:
```
detekt { ... }

tasks.withType<Detekt> {
  exclude("**/*.kt")
}
```

In the current plugin the task doesn't extend SourceTask so doesn't make sense to fix it outside this PR.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -1,33 +1,100 @@
...
-		project.extensions.create(DETEKT_EXTENSION_NAME, DetektExtension::class.java)
-		project.task(mapOf(TYPE to DetektCheckTask::class.java), CHECK)
-		project.task(mapOf(TYPE to DetektIdeaFormatTask::class.java), IDEA_FORMAT)
-		project.task(mapOf(TYPE to DetektIdeaInspectionTask::class.java), IDEA_INSPECT)
-		project.task(mapOf(TYPE to DetektGenerateConfigTask::class.java), GENERATE_CONFIG)
-		project.task(mapOf(TYPE to DetektCreateBaselineTask::class.java), BASELINE)
+	override fun getToolName() = DETEKT_EXTENSION_NAME
+	override fun getTaskType() = Detekt::class.java
+
+	override fun configureConfiguration(configuration: Configuration) {
```
3flex: This method was only [added in Gradle 4.8](https://github.com/gradle/gradle/commit/d3bcd32c0fde3e9c0915876be57e285ec28cdfbe)... AbstractCodeQualityPlugin is a Gradle internal API, it could be dangerous to subclass it.

Mauin: Subclassing `AbstractCodeQualityPlugin` was the initial idea of this PR. That way we get the same behavior/configuration as Checkstyle/PMD and other code quality tools that are built in by default and would result in easier configurations across different code quality tools.

build.gradle.kts
```diff
@@ -56,6 +62,26 @@ subprojects {
...
+		baseline = file("${rootProject.projectDir}/reports/baseline.xml")
+		configFile = file("${rootProject.projectDir}/detekt-cli/src/main/resources/default-detekt-config.yml")
+		toolVersion = "1.0.0.RC7-MARVIN2"
+
+		idea(Action {
```
3flex: (left this comment earlier, but it got marked as outdated)

Is there a way to change this so it reads `idea {`? May not have been possible in older Kotlin DSL versions, but according to "Convenient configuration of nested extensions" section of Gradle Kotlin DSL 0.17.4 Release Notes should be possible from Gradle 4.8-rc-1.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
arturbosch: so we have to say make clear that the new plugin only works with bleeding edge gradle :)
And so will detekt 1.0 ^^

tasomaniac: Well, i think @3flex was talking about compatibility perspective. What this project uses does not mean that users are required to have Gradle 4.8. 

I believe this line would not break compatibility with older Gradle versions. 

Would be good to double check.

3flex: A few points for consideration:
1. The [Designing Gradle Plugins Guide](https://guides.gradle.org/designing-gradle-plugins/) says: "Keep in mind that internal APIs are subject to change and can easily break your plugin from one Gradle version to another."
2. The public API is defined by tasks and extensions, not the plugin itself. Thanks to implementing SourceTask, VerificationTask, Reporting<> and CodeQualityExtension, the public API is now aligned with the other code quality extensions. AbstractCodeQualityPlugin is just an implementation detail.
3. Asking users to stay on latest Gradle at all times isn't a great experience. In particular, in the Android world, support for latest Gradle version often lags. For example right now on a new Android project if you use Gradle 4.6 or higher there is an error. Users must either downgrade to 4.5, or alter a configuration ([which isn't intuitive](https://stackoverflow.com/questions/49990933/configuration-on-demand-is-not-supported-by-the-current-version-of-the-android-g)), If detekt only supports 4.8, users have to decide between updating detekt or trying to figure out the workaround for the Android error.

I completely understand that pulling this PR together has been a massive undertaking, so perhaps it's better to get it over the line once tests are ready and passing and look to switching to `Plugin<Project>` afterwards. I just think sticking with the internal API is going to bite later on, in possibly unpredictable ways.

Mauin: I've opened https://github.com/gradle/gradle/issues/5575 to check if Gradle would consider making `AbstractCodeQualityPlugin` part of the public API.

The `AbstractCodeQualityPlugin` still manages creating `detekt` tasks for each source set and some other things that we would have to re-implement if we want to have the same API and features as the Gradle-internal static analysis plugins.

I started with `AbstractCodeQualityPlugin` as we've discussed in #580 that it is probably the way to go to make detekt behave the same as all major code quality Gradle Plugins out there.
Of course this means an increased maintenance on Detekt to keep up-to-date with upstream changes in Gradle.

As for point three. Of course we should verify that we're compatible with previous Gradle versions. The issue with the Android Gradle Plugin and newer Gradle versions is known, including a workaround, and will be fixed in the next version of the Plugin according to the Google issue tracker.

Mauin: As mentioned in https://github.com/gradle/gradle/issues/5575 Gradle would prefer keeping it internal for now, and they urge us to not use `AbstractCodeQualityPlugin` for now. So we would try re-implementing some of its features and depending on how much code we end up duplicating they might reconsider making the class part of the public API.

Mauin: Also linking https://github.com/spotbugs/spotbugs-gradle-plugin/issues/29. Spotbugs is reverting their implementation of `AbstractCodeQualityPlugin` as well.

tasomaniac: In relating to the issue Spotbugs had, it may result into a forward compatibility issue in the future which can be pretty easily be fixed by just updating the library and integrating the latest version of Gradle API. But from users perspective it will still be backward compatible.

3flex:
@Mauin thoughts on how to get tests working in CI? gradle-plugin resolves the dependency at runtime using a module dependency, so the cli artifact has to be published to a repository.

I think two options make sense:
1. Publish SNAPSHOT builds of detekt, and have gradle-plugin consume them during testing (already requested, see https://github.com/arturbosch/detekt/issues/512)
2. Add a publishToMavenLocal step on detekt and add a mavenLocal repository when building the Gradle plugin

Thoughts? 2 can be done very cleanly in CI using a [Gradle init script](https://docs.gradle.org/current/userguide/init_scripts.html). If you're good with that I'll submit a PR to master; this would be useful in CI anyway so we'd know the plugin and latest detekt work well together.

sschuberth:
Wouldn't it be possible on CI to conditionally [use dependency substitution to replace the module dependency with a project dependency](https://docs.gradle.org/current/userguide/customizing_dependency_resolution_behavior.html#sec:project_to_module_substitution)?

3flex:
Maybe... this project now uses a Gradle composite build, so to detekt-gradle-plugin, detekt-cli is not part of the project so detekt-cli can't be treated as a project dependency from gradle-plugin. Does dependency substitution work if substituting with a file instead of a project dependency? Was unclear from the docs. I'll test locally to confirm since that should be the cleanest solution.

[This](https://docs.gradle.org/current/userguide/customizing_dependency_resolution_behavior.html#sec:configuration_defaults) might work too.

Mauin:
Closing this as it is succeeded by #1017

Issue #768, opened 2018-03-04
EndOfSentenceFormat does not work well with urls.


vanniktech:
Sometimes I've got comments just like

`/** http://www.graphviz.org/doc/info/shapes.html */` sometimes they're also prefixed.
`/** You can get more information in the documentation. http://www.graphviz.org/doc/info/shapes.html */`

One idea would be to split the comment using spaces and looking up whether the last part can be parsed to a url. If that's the case we don't need to flag it.

vanniktech:
Thanks. I'll close this now. When can we expect a new release?

PR #772, opened 2018-03-04
 Implemented EndOfSentenceFormat to support urls as last argument

schalkms:
reference #768

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/EndOfSentenceFormat.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/EndOfSentenceFormat.kt
@@ -24,12 +24,14 @@ class KDocStyle(config: Config = Config.empty) : MultiRule() {
 }
 
 /**
- * This rule validates the end of the first sentence of a KDoc comment. It should end with proper punctuation.
+ * This rule validates the end of the first sentence of a KDoc comment.
+ * It should end with proper punctuation or with a correct URL.
  *
  * @configuration endOfSentenceFormat - regular expression which should match the end of the first sentence in the KDoc
  * (default: ([.?!][ \t\n\r\f<])|([.?!]$))
  *
  *  @author Marvin Ramin
+ *  @author schalkms
  */
 class EndOfSentenceFormat(config: Config = Config.empty) : Rule(config) {
 
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/EndOfSentenceFormat.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/EndOfSentenceFormat.kt
@@ -2,6 +2,7 @@ package io.gitlab.arturbosch.detekt.rules.documentation
 
 import io.gitlab.arturbosch.detekt.api.CodeSmell
 import io.gitlab.arturbosch.detekt.api.Config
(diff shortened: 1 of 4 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/EndOfSentenceFormat.kt
```diff
@@ -24,42 +25,49 @@ class KDocStyle(config: Config = Config.empty) : MultiRule() {
...
 			Severity.Maintainability,
-			"The first sentence in a KDoc comment should end with correct punctuation.")
+			"The first sentence in a KDoc comment should end with correct punctuation.",
+			Debt.FIVE_MINS)
 
 	private val endOfSentenceFormat =
 			Regex(valueOrDefault(END_OF_SENTENCE_FORMAT, "([.?!][ \\t\\n\\r\\f<])|([.?!]\$)"))
 
+	// https://stackoverflow.com/questions/3809401/what-is-a-good-regular-expression-to-match-a-url
+	private val urlFormat = Regex("[-a-zA-Z0-9@:%._+~#=]{2,256}\\.[a-z]{2,6}\\b([-a-zA-Z0-9@:%_+.~#?&//=]*)")
```
vanniktech: We could leave the job to Javas URL class which throws an exception upon instantiation of an invalid url. But regex should work too

schalkms: I also thought about that. Javas URL class needs a URL with protocol definition though.
`www.google.com` would be invalid, but detekt could just append a protocol then.
Furthermore links to OS-dependent files and directories wouldn't be supported then.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/EndOfSentenceFormat.kt
```diff
@@ -24,42 +25,49 @@ class KDocStyle(config: Config = Config.empty) : MultiRule() {
...
+				report(CodeSmell(issue, Entity.from(declaration),
+						"The first sentence of this KDoc does not end with the correct punctuation."))
 			}
 		}
 	}
 
 	private fun String.startsWithHtmlTag() = startsWith("<") && contains(htmlTag)
 
+	private fun lastArgumentMatchesUrl(text: String): Boolean {
+		val arguments = text.trimEnd().split(Regex("\\s+"))
```
arturbosch: Please extract this refex and url regex into top level variables and reuse them. Regex creation and the `matches()` function are really expensive when doing them everytime.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/EndOfSentenceFormat.kt
arturbosch: I think URL is cheaper to use performance-wise (does not use a regex). But I think it does not handle all cases. 
Lets just use the url class and change it later if a user finds an issue with that?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/EndOfSentenceFormat.kt
schalkms: Thanks! In all the hurry I forgot to extract this regex to a property.

PR #776, opened 2018-03-04
Updated rule debt values

schalkms:
I just updated too high or too low debt values.


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/LateinitUsage.kt
```diff
@@ -31,12 +31,14 @@ import org.jetbrains.kotlin.psi.psiUtil.containingClass
  * @configuration ignoreOnClassesPattern - Allows you to disable the rule for a list of classes (default: "")
  *
  * @author Marvin Ramin
+ * @author Niklas Baudy
```
vanniktech: You really went through the files and added them as they edited the code? 🏅

schalkms: No. I did that when we introduced the rule documentation. :)
I just oversaw that.

schalkms: The default debt is 20.
I think `Issue` should be changed in order to not allow an empty description. Furthermore the caller should be forced to specify a debt value as you already said.

Mauin: Agree on the description part. I initially gave it a default value in the constructor because we never specified a description anywhere when we introduced this value. Now that all rules have a valid description we should remove the default parameter from the constructor to enforce it's always passed.

schalkms:
Should **detekt** output the estimated debt in the report and/or include the debt values in the documentation? What do you think?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/LateinitUsage.kt
arturbosch: I also agree on the description part. 
But I not having a default Debt (10min?) value adds so much clutter ...

arturbosch:
Thats a great idea! :+1: :+1:

schalkms:
Ok.
Can something be done to resolve the exception in the build besides waiting for a new RC? The `detekt.sample.extensions` references the old `Issue` in the old **.jar**.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/LateinitUsage.kt
arturbosch: Well, I guess its better to be explicit. So ok!

arturbosch:
@mauin commited a migration gradle task for that I think ...

Mauin:
I don't think I ever built that task for making that work. But the steps you explained should do the trick, yes.

schalkms:
Ok. I did a gradle build on the `detekt.sample.extensions` project.
I only get `detekt-sample-extensions-1.0.0.RC6-3` and `detekt-sample-extensions-1.0.0.RC6-3-sources`. What did I do wrong?

arturbosch:
│  │     └─ it should load the sample provider ✘ java.lang.IllegalArgumentException: Parameter specified as non-null is null: method io.gitlab.arturbosch.detekt.api.Issue.<init>, parameter debt

This test case is still broken :/

schalkms:
@arturbosch I rebased after the new RC. I have no idea how to fix it.

arturbosch:
Merged manually.

PR #780, opened 2018-03-08
Add trailing whitespace rule

misaelmt:
I didn't add \<compliant\> and \<noncompliant\> blocks to the KDoc because it's really easy to understand what the rule does and it's hard to see those white spaces at the end of the lines in the KDoc comments.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrailingWhitespace.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrailingWhitespace.kt
@@ -1,0 +1,34 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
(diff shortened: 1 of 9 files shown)


misaelmt:
What should I do to add this feature to auto correction? It would be nice to add it in a follow up PR.

vanniktech:
I proposed auto correction here but it didn't get approved. I'm just nagging on each rule that would benefit from it in hope that we'll get auto correcting.

schalkms:
As @vanniktech mentioned atm detekt does not have rules with auto correction.
The PR is fine. Nothing more to do here.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrailingWhitespace.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+class TrailingWhitespace(config: Config = Config.empty) : Rule(config) {
+
+	override val issue = Issue(javaClass.simpleName,
+			Severity.Style,
+			"Checks which lines end with a whitespace.",
+			Debt.FIVE_MINS)
+
+	override fun visitKtFile(file: KtFile) {
+		val lines = file.text.splitToSequence("\n")
```
arturbosch: This should be added to the `FileParsingRule`.

misaelmt:
@arturbosch I addressed your comments. Please re-review at your convenience.
IMO KtFileContent.content should be renamed to KtFileContent.lines

PR #782, opened 2018-03-09
Add no tabs rule

misaelmt:
I didn't add <compliant> and <noncompliant> blocks to the KDoc because it's really easy to understand what the rule does and it's hard to see those tabs in the KDoc comments.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/NoTabs.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/NoTabs.kt
@@ -1,0 +1,35 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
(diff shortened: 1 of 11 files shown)


egor-n:
It seems like tabs inside comments will also fail the build – should this be the case?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/NoTabs.kt
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * This rule reports if tabs are used in Kotlin files.
+ * According to
+ * [Google's Kotlin style guide](https://android.github.io/kotlin-guides/style.html#whitespace-characters)
+ * the only whitespace chars that are allowed in a source file are the line terminator sequence
+ * and the ASCII horizontal space character (0x20).
+ *
+ * @author Misa Torres
+ */
+class NoTabs(config: Config = Config.empty) : Rule(config) {
```
arturbosch: Heyho, this rule is a candidate for the `FileParsingRule` MultiRule (so we do not split every file x-times)
please take a look at `MaxLineLength` how it is implemented.
Thanks for contributing ! :)

misaelmt:
@egor-n @schalkms Google's Kotlin Style guide states that: "Aside from the line terminator sequence, the ASCII horizontal space character (0x20) is the only whitespace character that appears anywhere in a source file." I put a comment in the Rule about this. This rule is to ensure that the IDE is set to replace tabs for whitespaces. If you allow tabs in comments and disallow them in other parts then you are putting yourself in a modal state and that is always a bad idea.

misaelmt:
@arturbosch I addressed your comments. Please -re-review at your convenience.

arturbosch:
Thanks!

PR #784, opened 2018-03-09
Add 'no semicolons at the end of a line' rule

misaelmt:
A few notes:
- This PR is going to have a merge conflict in StyleGuideProvider.kt once one of my previous PR gets merged. 
- I didn't update README.md because I'm working in a couple more of rules. I'm going to update it in the last PR to mention the rules I added, and that way I avoid the merge conflict in this file.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SemicolonAtEndOfLine.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SemicolonAtEndOfLine.kt
@@ -1,0 +1,45 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.psi.KtFile
+
+/**
+ * This rule reports if a line ends with a semicolon.
+ *
+ * <noncompliant>
+ * println();
+ * // this is a comment;
+ * </noncompliant>
+ *
+ * <compliant>
+ * println()
+ * println(); println()
+ * // this is a comment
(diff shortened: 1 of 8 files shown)


arturbosch:
That's really a tough one. We have the `FileParsingRule` MultiRule so having both PR's won't hurt that much. But we should not still autocorrect rules. What do you think @Mauin @schalkms ?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SemicolonAtEndOfLine.kt
```diff
@@ -0,0 +1,45 @@
...
+ * <compliant>
+ * println()
+ * println(); println()
+ * // this is a comment
+ * println(); // how to get away this rule
+ * </compliant>
+ *
+ * @author Misa Torres
+ */
+class SemicolonAtEndOfLine(config: Config = Config.empty) : Rule(config) {
```
arturbosch: This could also be added to the `FileParsingRule`.

schalkms:
@arturbosch
I think the point here is that this rule were implemented in detekt in the past.
Regarding the auto correction I think this should be an optional command line parameter. Per default this should be disabled.
Moreover this feature should be postponed.

egor-n:
The following shouldn't be flagged:
```kotlin
enum class Example {
  A, B;

  fun someFun() {
  }
}
```

This too:
```kotlin
// I like semicolons;
```

misaelmt:
@egor-n those cases make sense. I knew about the comment case, it is even in the test cases.
Is there a better way to implement this rule instead of parsing the file or visiting every single type of KtElement?

arturbosch:
Mh, @schalkms is right, lets not go deeper into formatting again. KtLint has such a [rule](https://github.com/shyiko/ktlint/blob/master/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/NoSemicolonsRule.kt) and we can trigger an intellij formatting from detekt gradle plugin.
We will keep NoTabs and MaxLineLength etc now. There is an option in the future that we wrap ktlint's rules, so we should not implement more of this kind.
If you plan to implement more formatting rules, please consider your own rule set.
And to your question, there is a `TokenRule` which has methods like `visitSemicolon` and `visitDoubleSemicolon`.

arturbosch:
FYI https://arturbosch.github.io/detekt/extensions.html

misaelmt:
No problem It's good to know what are the priorities and how this tool is going to evolve.

PR #796, opened 2018-03-13
State the known issues for rc6-3 in the readme

arturbosch:


Mauin:
Agree with @vanniktech. We have the fixes for these known issues in master. Let's release them or keep issues for each big open until they are released. But this list in the readme will become outdated and our readme is big enough that people will also miss this.

Issue #797, opened 2018-03-13
Introduce a commit template


arturbosch:

schalkms:
+1 for me
I think the link in the **Contributing.md** describes a good template.

schalkms:
Should the build also fail if the commit message doesn’t comply to the template?

arturbosch:
a pre commit git hook for this would be cool :)

Mauin:
Pasting my comments fro #815 here:

> Not really sure this is going to give us much benefit. As @vanniktech already mentioned just using "Squash and Merge" instead of simply merging PRs we will get one well-named commit with the title of the PR (which should have good, descriptive names and we can even rename them if necessary). This commit hook will add another tripwire for people trying to quickly contribute to detekt. 
> I think we shouldn't make it harder than necessary. GitHub can take care of squashing, merging and renaming the appropriate commit for us, let's not put that burden onto people trying to contribute.

schalkms:
I think it's useful. It's just one command to copy this file to `.git/hooks`. This can be packed into the build script for instance.

PR #815, opened 2018-03-16
Commit hook

schalkms:
This is the first version. The commit hook works locally.
Should detekt also include a server side hook?
https://git-scm.com/book/uz/v2/Customizing-Git-An-Example-Git-Enforced-Policy

### DIFF
--- a/commit-msg
+++ b/commit-msg
@@ -21,15 +21,3 @@ if [ "$len" -ge 2 ]; then
     exit 1
   fi
 fi
-
-# checks if all lines of the commit message do not exceed the length 70
-for (( i=1; i<=len; i++ ))
-do
-  index=$(printf "%sp" "$i")
-  el=`sed -n $index $msg`
-  if [ "${#el}" -gt 70 ]; then
-    echo "Aborting commit. The commit message exceeds the maximum allowed line length of 70" >&2
-    exit 1
-  fi
-done
-
--- a/commit-msg
+++ b/commit-msg
@@ -1,0 +1,35 @@
+#!/bin/sh
+#
+# This hook script checks the commit log message.
+# Called by "git commit" with one argument, the name of the file that has the commit message.
+
+msg=$1
+
+# checks if the first character of the commit message is an uppercase letter
+uppercase='^[A-Z]'
+if ! head -1 "$msg" | grep -Eq "$uppercase"; then
+  echo "Aborting commit. The commit message does not start with an uppercase letter." >&2
+  exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


commit-msg
```diff
@@ -0,0 +1,35 @@
...
+msg=$1
+
+# checks if the first character of the commit message is an uppercase letter
+uppercase='^[A-Z]'
+if ! head -1 "$msg" | grep -Eq "$uppercase"; then
+  echo "Aborting commit. The commit message does not start with an uppercase letter." >&2
+  exit 1
+fi
+
+# checks if the second line of the commit message is empty
```
vanniktech: What's that one used for?

commit-msg
```diff
@@ -0,0 +1,35 @@
...
+len=`awk 'END{print NR}' $msg`
+if [ "$len" -ge 2 ]; then
+  line=`sed -n 2p $msg`
+  if [ "$line" != "" ]; then
+    echo "Aborting commit. The commit message does not have a empty second line." >&2
+    exit 1
+  fi
+fi
+
+# checks if all lines of the commit message do not exceed the length 70
```
vanniktech: Do we really want to enforce this?

schalkms:
@vanniktech 
This is just a proposal. There is a discussion in #797

Mauin:
Not really sure this is going to give us much benefit. As @vanniktech already mentioned just using "Squash and Merge" instead of simply merging PRs we will get one well-named commit with the title of the PR (which should have good, descriptive names and we can even rename them if necessary). This commit hook will add another tripwire for people trying to quickly contribute to detekt. 
I think we shouldn't make it harder than necessary. GitHub can take care of squashing, merging and renaming the appropriate commit for us, let's not put that burden onto people trying to contribute.

commit-msg
arturbosch: git commit should be written as:

the title with issue id

longer description if necessary.

I think it checks if the second line is empty so git detects it as the separator of the detailed message

commit-msg
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+#
+# This hook script checks the commit log message.
+# Called by "git commit" with one argument, the name of the file that has the commit message.
+
+msg=$1
+
+# checks if the first character of the commit message is an uppercase letter
```
arturbosch: personally I always start with a capital letter so :+1: from me

commit-msg
arturbosch: I also dont think we should enforce this. Ok for the other two

PR #817, opened 2018-03-16
NestedClassesVisibility doesn't detect violation for object

### LINKED ISSUE #814: NestedClassesVisibility doesn't detect violation for "object"
version: 1.0.0.R6-4

```kotlin
internal sealed class Alphabet {
    object A : Alphabet()
    data class B(val throwable: Throwable) : Alphabet()
}
```

Rule detects violation for `B`, but doesn't detect violation for `A`

schalkms:
Fixes #814

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/NestedClassesVisibility.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/NestedClassesVisibility.kt
@@ -9,7 +9,9 @@ import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.rules.isInternal
 import io.gitlab.arturbosch.detekt.rules.isPublic
+import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtClassOrObject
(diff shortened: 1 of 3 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/NestedClassesVisibility.kt
```diff
@@ -50,12 +52,12 @@ class NestedClassesVisibility(config: Config = Config.empty) : Rule(config) {
 
 	private fun checkDeclarations(klass: KtClass) {
 		klass.declarations
-				.filterIsInstance<KtClass>()
-				.filter { it.isPublic() && !it.isEnum() && it !is KtEnumEntry }
-				.forEach { report(CodeSmell(issue,
-						Entity.from(it),
+				.filterIsInstance<KtClassOrObject>()
+				.filter { it.isPublic() && !it.hasModifier(KtTokens.ENUM_KEYWORD) && it !is KtEnumEntry }
```
vanniktech: Can we remove the `it !is KtEnumEntry` then?

schalkms: No. It would match the entries of an enum then.

Mauin: can we split this into 3 filters? Or extract this into a well-named method>

PR #820, opened 2018-03-18
Migrate detekt-sample-extensions/build.gradle to kotlin dsl

epeee:
Another pr towards #611.

I can work on the remaining two build.gradle files later on (detekt-gradle-plugin and the rootproject one).


detekt-sample-extensions/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
+buildscript {
+	repositories {
+		mavenLocal()
+		jcenter()
+	}
+
+	val kotlinVersion by project
```
arturbosch: buildscript cant pick up the top level var?

arturbosch:
the gradle-plugin gets a rework in #771, so this has to wait for that, but later yes please do it :)

Mauin:
Even with the gradle plugin rework changing the `build.gradle` file is not the worst of the world. Should be fairly easy to rebase, so no reason from my side to not convert the `build.gradle` file to Kotlin as soon as we want 👍

detekt-sample-extensions/build.gradle.kts
epeee: Yep, it can't.

Issue #811, opened 2018-03-16
deprecation message is printed even when warningThreshold and failThreshold are not used


MyDogTom:
Version: 1.0.0.RC6-4

In my config I'm using `failFast: true` and do not specify `build` section at all. Nevertheless, I see the following message in console:

>[Deprecation] - 'warningThreshold' and 'failThreshold' properties are deprecated. Please use the new 'maxIssues' config property.

Is that intended?

vanniktech:
That's probably working as intended since the default configuration is used and there the `warningThreshold` and `failThreshold` are defined.

MyDogTom:
Are there any reason to keep deprecated properties in default configuration?

PR #831, opened 2018-03-25
Do not write deprecated build fail properties to default config file

arturbosch:
- Closes#811
- for 1.0.GA we need to remove all the code for `warningThreshold` and `failThreshold`


arturbosch:
closes #811

PR #832, opened 2018-03-25
Change Debt value for TopLevelPropertyNaming, UnusedImports & UnusedPrivateMember.

vanniktech:
I really like that the `Debt` is shown in the reports now however I feel like the estimates should be closer to how much you actually spend removing the code smell.

Ideally, the overall debt could be a realistic estimate how much time you'd need to spend in a new code base which isn't using Detekt yet.

Probably before 1.0 we should go through all of the issues and do some planning poker :D

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
@@ -48,6 +48,7 @@ data class Debt(val days: Int = 0, val hours: Int = 0, val mins: Int = 0) {
 		val TWENTY_MINS = Debt(0, 0, 20)
 		val TEN_MINS = Debt(0, 0, 10)
 		val FIVE_MINS = Debt(0, 0, 5)
+		val ONE_MIN = Debt(0, 0, 1)
 	}
 
 	override fun toString(): String {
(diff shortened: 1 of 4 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
```diff
@@ -48,6 +48,7 @@ data class Debt(val days: Int = 0, val hours: Int = 0, val mins: Int = 0) {
 		val TWENTY_MINS = Debt(0, 0, 20)
 		val TEN_MINS = Debt(0, 0, 10)
 		val FIVE_MINS = Debt(0, 0, 5)
+		val ONE_MIN = Debt(0, 0, 5)
```
arturbosch: I think you forgot to change the integer :D

vanniktech: Yes. Thanks :D

schalkms:
I think 1 minute is a bit too short for fixing an issue (opening IDE, have a second thought about the smell, committing). This might be feasible in a small project but certainly not in an enterprise level project.

vanniktech:
I don't think we should include opening the IDE and comitting the changes into the debt values.

- `TopLevelPropertyNaming` is a simple rename that can be done by the IDE
- `UnusedImport` is as simple as deleting the line. If you know your way with `sed` you can even remove thousands of those within seconds.
- `UnusedPrivateMember` Simple removal of something. Although something like 2-3 minutes might be more realistic if you want to gauge whether this should be used or not.

arturbosch:
I think we should be more conservative. 5 minutes is minimum if you consider choosing a good commit message, opening a PR, review the changes and merge them upstream.

vanniktech:
Hmm seems like I could not convince you so I'll close this and just mentally divide that debt number by something between 2 and 5. (zwuenf :D)

schalkms:
:)
Thanks for bringing up the discussion though.

PR #840, opened 2018-03-30
Fix false positives in unused private members

### LINKED ISSUE #812: UnusedPrivateMember false positives
version: 1.0.0.RC6-4
**Case 1**

```kotlin
@ConsumerScope
@Component(dependencies = arrayOf(OneComponent::class))
interface AnotherComponent {
    fun inject(myActivity: MyActivity)
}
```
Rule complains about `myActivity`.

```kotlin
interface UserPlugin {
    fun plug(application: Application)
    fun unplug()
}
```

Rule complains about `application`

**Case 2**

```kotlin
    override fun funA() {
        objectA.resolve(valA, object : MyCallback {
            override fun onResolveFailed(throwable: Throwable) {
                errorMessage.visibility = View.VISIBLE
            }
        })
    }
```

rule complains about `throwable`

**Case 3**
```kotlin 
class A {
    companion object {
        private const val MY_CONST = 42
    }

    fun a() {
        Compleatable.timer(MY_CONST.toLong(), TimeUnit.MILLISECONDS)
              .subscribe()
    }
}
```

rule complains about `MY_CONST`

PS. I have many more violations in my codebase. I just randomly picked few of them, but most likely all of them a similar.

arturbosch:
-  This should fix all false positives in  #812 
Edit: we can't turn on the rule in our code base for now due to detektCheck

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/Case.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/Case.kt
@@ -78,7 +78,8 @@ enum class Case(val file: String) {
 	TrailingWhitespacePositive("/cases/TrailingWhitespacePositive.kt"),
 	NoTabsNegative("/cases/NoTabsNegative.kt"),
 	NoTabsPositive("/cases/NoTabsPositive.kt"),
-	UnusedPrivateMembers("/cases/UnusedPrivateMembers.kt");
+	UnusedPrivateMemberPositive("/cases/UnusedPrivateMemberPositive.kt"),
+	UnusedPrivateMemberNegative("/cases/UnusedPrivateMemberNegative.kt");
 
 	fun path(): Path = Paths.get(resource(file))
 
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/Case.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/Case.kt
@@ -77,14 +77,14 @@ enum class Case(val file: String) {
 	TrailingWhitespaceNegative("/cases/TrailingWhitespaceNegative.kt"),
 	TrailingWhitespacePositive("/cases/TrailingWhitespacePositive.kt"),
 	NoTabsNegative("/cases/NoTabsNegative.kt"),
-	NoTabsPositive("/cases/NoTabsPositive.kt");
+	NoTabsPositive("/cases/NoTabsPositive.kt"),
+	UnusedPrivateMembers("/cases/UnusedPrivateMembers.kt");
 
 	fun path(): Path = Paths.get(resource(file))
 
 	fun getKtFileContent(): KtFileContent {
 		val file = compileForTest(path())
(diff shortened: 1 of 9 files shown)


Mygod:
This still doesn't fix false positives for extension private functions. For example this:

https://github.com/Mygod/VPNHotspot/blob/ed4c9a5610d447ee4e7541f45ca85fe3674661d9/mobile/src/main/java/be/mygod/vpnhotspot/RepeaterService.kt#L52-L55

Mygod:
Also private members in companion objects: https://github.com/shadowsocks/shadowsocks-android/blob/a7d7746195d687619899ac7b73fed7791eea9492/mobile/src/main/java/com/github/shadowsocks/AppManager.kt#L65

Mygod:
Also variables named `_`: https://github.com/shadowsocks/shadowsocks-android/blob/a7d7746195d687619899ac7b73fed7791eea9492/mobile/src/main/java/com/github/shadowsocks/ProfilesFragment.kt#L278

arturbosch:
Hey, thanks for the feedback. I will look into your cases.

vanniktech:
We also probably want to exclude properties that are annotated with @Rule by default since sometimes they do some work but you don't interact on them.

schalkms:
@vanniktech For now, this would make sense. It is in the VCS anyway.

arturbosch:
@schalkms this kind of rule just needs symbol solving when also considering public members, it is alright to have such a rule now. The problem was, that you can declare many additional objects in the same file and use their constants which was not considered in the initial implementation.
It's not that I need so long with this PR it's just that I do not have time :(

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/Case.kt
```diff
@@ -77,14 +77,14 @@ enum class Case(val file: String) {
 	TrailingWhitespaceNegative("/cases/TrailingWhitespaceNegative.kt"),
 	TrailingWhitespacePositive("/cases/TrailingWhitespacePositive.kt"),
 	NoTabsNegative("/cases/NoTabsNegative.kt"),
-	NoTabsPositive("/cases/NoTabsPositive.kt");
+	NoTabsPositive("/cases/NoTabsPositive.kt"),
+	UnusedPrivateMembers("/cases/UnusedPrivateMembers.kt");
```
schalkms: Should we stick to splitting **positive** and **negative** test cases also for `UnusedPrivateMembers`?

arturbosch: I will try this later.

arturbosch:
Besides detekt, spek and arrow, I need like 2-3 more projects for QA.
Feel free to name some.

schalkms:
@arturbosch the detekt compiler project and ktlint

schalkms:
@vanniktech definitely good idea! +1

arturbosch:
Ok, also added a configurable to allow unused private members matching  a regex (perconfigured with _|ignored|expected).

Jap, a list with projects using detekt would be cool to test out new rules/features and make sure nothing breaks ^^

Issue #807, opened 2018-03-16
Exception: Failed to create MD5 hash for file content


juergenzimmermann:
Using Windows 10, Java 10, Gradle 4.6, and detekt 1.0.0.RC6-4 I get the following stacktrace.

The exception's message in English is _The process cannot access the file, because another process has locked a part of the file_.

```
> gradle detektCheck --info --stacktrace
Initialized native services in: C:\Z\native
Removing 0 daemon stop events from registry
Starting a Gradle Daemon (subsequent builds will be faster)
Starting process 'Gradle build daemon'. Working directory: C:\Z\daemon\4.6 Command: C:\Zimmermann\jdk\bin\java.exe -Xmx1536M -Dfile.encoding=UTF-8 -Duser.country=DE -Duser.language=de -Duser.variant -cp C:\Zimmermann\gradle\lib\gradle-launcher-4.6.jar org.gradle.launcher.daemon.bootstrap.GradleDaemon 4.6
Successfully started process 'Gradle build daemon'
An attempt to start the daemon took 27.694 secs.
The client will now receive all logging from the daemon (pid: 14996). The daemon log file: C:\Z\daemon\4.6\daemon-14996.out.log
Starting build in new daemon [memory: 1,6 GB]
Using 4 worker leases.
Starting Build
Settings evaluated using settings file 'C:\Users\ziju0001\IdeaProjects\config\settings.gradle'.
Using local directory build cache for the root build (location = C:\Z\caches\build-cache-1, removeUnusedEntriesAfter = 7 days).
Projects loaded. Root project using build file 'C:\Users\ziju0001\IdeaProjects\config\build.gradle'.
Included projects: [root project 'config']

> Configure project :
Evaluating root project 'config' using build file 'C:\Users\ziju0001\IdeaProjects\config\build.gradle'.
Applying dependency management to configuration 'annotationProcessor' in project 'config'
Applying dependency management to configuration 'apiElements' in project 'config'
Applying dependency management to configuration 'archives' in project 'config'
Applying dependency management to configuration 'bootArchives' in project 'config'
Applying dependency management to configuration 'compile' in project 'config'
Applying dependency management to configuration 'compileClasspath' in project 'config'
Applying dependency management to configuration 'compileOnly' in project 'config'
Applying dependency management to configuration 'default' in project 'config'
Applying dependency management to configuration 'implementation' in project 'config'
Applying dependency management to configuration 'kapt' in project 'config'
[...]

PR #845, opened 2018-04-06
Add dependencies for gradle plugin (for custom rules)

olivierlemasle:
Following the feature request https://github.com/arturbosch/detekt/issues/795, this commit implements a way to configure Detekt plugins using Gradle plugin.

It makes possible to use
```groovy
dependencies {
    detekt project(':detekt-extensions')
}

detekt {
    profile('main') {
        input = projectDir
        config = "$projectDir/detekt.yml"
    }
}
```
to have project "detekt-extensions" on Detekt classpath.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCheckTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCheckTask.kt
@@ -4,7 +4,7 @@ import io.gitlab.arturbosch.detekt.extensions.DetektExtension
 import io.gitlab.arturbosch.detekt.extensions.INPUT_PARAMETER
 import io.gitlab.arturbosch.detekt.extensions.OUTPUT_PARAMETER
 import org.gradle.api.DefaultTask
-import org.gradle.api.internal.artifacts.dependencies.DefaultExternalModuleDependency
+import org.gradle.api.file.FileCollection
 import org.gradle.api.tasks.InputDirectory
 import org.gradle.api.tasks.OutputDirectory
 import org.gradle.api.tasks.TaskAction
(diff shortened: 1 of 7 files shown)


arturbosch:
Thants super cool, please add your name with an author tag on the modfied files and squash the commit :)_

olivierlemasle:
Thanks for your reviews. I've added some functional tests, the author tag and some documentation in README.

@Mauin I hope it will not be too cumbersome too merge into #771...

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCheckTask.kt
```diff
@@ -22,6 +23,8 @@ open class DetektCheckTask : DefaultTask() {
 	@OutputDirectory
 	val output: File?
 
+	private val classpath: FileCollection
```
mkobit: this should be marked as in input, probably with [`@Classpath`](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Classpath.html)

olivierlemasle: Actually, I could add the `@Classpath` annotation and make the property public, and in that case I could also remove the explicit dependency `dependsOn(classpath)` (line 43).

However, I'm not sure making this property public is a good thing, as I dont't think this should be editable. Do you suggest making `classpath` public?

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/FunctionalTest.kt
```diff
@@ -0,0 +1,133 @@
...
+ */
+internal class FunctionalTest : Spek({
+
+	describe("The Detekt Gradle plugin") {
+
+		it("uses built-in rules") {
+			val rootDir = createTempDir(prefix = "withoutCustomRules")
+			writeFiles(rootDir)
+
+			// Using a custom "project-cache-dir" to avoid a Gradle error on Windows
```
olivierlemasle: This Windows error is #807.

Issue #846, opened 2018-04-07
Set up pipeline to test new detekt rules/features on selected kotlin projects


arturbosch:
- [x] script to automatically download analysis projects
- [x] run detekt-cli on projects with arguments like `ruleset:rule[config=properties]`

arturbosch:
Should be enough to get started.

PR #847, opened 2018-04-07
Implement script to download given analysis projects

arturbosch:
- this resolves part 1 of #846


arturbosch:
Groovy has @Grab so it is easier to setup the "analysis pipeline" to test out new rules :).

Sounds good will do.

PR #849, opened 2018-04-10
Added option to exclude comments from MaxLineLength

### LINKED ISSUE #718: Exclude comments from MaxLineLength
This would be a nice feature!

rvoortman:
This will add an option to exclude comments from the MaxLineLength rule.

Fixes #718

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [ ![Download](https://api.bintray.com/packages/arturbosch/code-analysis/detekt/images/download.svg) ](https://bintray.com/arturbosch/code-analysis/detekt/_latestVersion)
 [![gradle plugin](https://img.shields.io/badge/gradle_plugin-1.0.0.RC6.4-blue.svg?style=flat-square)](https://plugins.gradle.org/plugin/io.gitlab.arturbosch.detekt)
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-40-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-41-orange.svg?style=flat-square)](#contributors)
 [![Awesome Kotlin Badge](https://kotlin.link/awesome-kotlin.svg)](https://github.com/KotlinBy/awesome-kotlin)
 
 Meet _detekt_, a static code analysis tool for the _Kotlin_ programming language.
@@ -647,6 +647,7 @@ If you contributed to detekt but your name is not in the list, please feel free 
 - [Said Tahsin Dane](https://github.com/tasomaniac/) - Gradle plugin improvements
 - [Misa Torres](https://github.com/misaelmt) - Added: TrailingWhitespace and NoTabs rules
 - [R.A. Porter](https://github.com/coyotesqrl) - Updated Readme links to RuleSets
+- [Robbin Voortman](https://github.com/rvoortman) - Rule improvement: MaxLineLength
 
 ### <a name="mentions">Mentions</a>
 
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLength.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLength.kt
@@ -38,8 +38,10 @@ data class KtFileContent(val file: KtFile, val content: Sequence<String>)
  * @configuration maxLineLength - maximum line length (default: 120)
  * @configuration excludePackageStatements - if package statements should be ignored (default: false)
  * @configuration excludeImportStatements - if import statements should be ignored (default: false)
+ * @configuration excludeCommentStatements - if comment statements should be ignored (default: false)
  *
  * @active since v1.0.0
+ * @author Robbin Voortman
  * @author Marvin Ramin
  * @author Artur Bosch
  */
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -6,7 +6,7 @@
 [ ![Download](https://api.bintray.com/packages/arturbosch/code-analysis/detekt/images/download.svg) ](https://bintray.com/arturbosch/code-analysis/detekt/_latestVersion)
 [![gradle plugin](https://img.shields.io/badge/gradle_plugin-1.0.0.RC6.4-blue.svg?style=flat-square)](https://plugins.gradle.org/plugin/io.gitlab.arturbosch.detekt)
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-40-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-41-orange.svg?style=flat-square)](#contributors)
```
vanniktech: Just curiosity. Is this number in sync with GitHubs number?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLength.kt
```diff
@@ -92,6 +97,13 @@ class MaxLineLength(config: Config = Config.empty) : Rule(config) {
 		return false
 	}
 
+	private fun containsIgnoredCommentStatement(line: String): Boolean {
+		if (excludeCommentStatements) {
+			return line.contains("//") || line.contains("/*") || line.trimStart().startsWith("*")
```
vanniktech: Aren't the first statements too simple? RegEx containing those would be excluded as well. 

trimStart() seems like of use there too

docs/pages/changelog.md
```diff
@@ -11,7 +11,7 @@ toc: true
 - NoSuchElementException when running from build.gradle.kts - [#793](https://github.com/arturbosch/detekt/issues/793)
 - No files analysed if path contains test - [#792](https://github.com/arturbosch/detekt/issues/792)
 - Does detekt include copy/paste detector ? - [#789](https://github.com/arturbosch/detekt/issues/789)
-- @Suppress("MaxLineLength") at file-level doesn't work - [#788](https://github.com/arturbosch/detekt/issues/788)
+- @Suppress("MaxLineLenth") at file-level doesn't work - [#788](https://github.com/arturbosch/detekt/issues/788)
```
vanniktech: Revert

README.md
rvoortman: I just noticed; it is not, will fix this soon 👍

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLength.kt
rvoortman: Good point.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLengthSpec.kt
```diff
@@ -18,7 +18,7 @@ class MaxLineLengthSpec : Spek({
 
 		it("should report no errors when maxLineLength is set to 200") {
 			val rule = MaxLineLength(TestConfig(mapOf("maxLineLength" to "200")))
-
+
```
vanniktech: trailing whitespace probably should not be there.

README.md
schalkms: No. It's just a hard-coded integer ;)

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLengthSpec.kt
```diff
@@ -96,6 +96,56 @@ class MaxLineLengthSpec : Spek({
 			assertThat(rule.findings).isEmpty()
 		}
 	}
+	given("a kt file with a long package name, long import statements, a long line and long comments") {
+		val code = """
+			package anIncrediblyLongAndComplexPackageNameThatProbablyShouldBeMuchShorterButForTheSakeOfTheTestItsNot
+
+			import anIncrediblyLongAndComplexImportNameThatProbablyShouldBeMuchShorterButForTheSakeOfTheTestItsNot
+
+			class Test {
```
schalkms: This test code with the package import and `class Test` is duplicated with line 151. Can we please refactor this either via a shared variable or a separate **test cases** file?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLength.kt
```diff
@@ -75,7 +79,8 @@ class MaxLineLength(config: Config = Config.empty) : Rule(config) {
 	private fun isValidLine(line: String): Boolean {
 		return (line.length <= maxLineLength
 				|| containsIgnoredPackageStatement(line)
-				|| containsIgnoredImportStatement(line))
+				|| containsIgnoredImportStatement(line)
+				|| containsIgnoredCommentStatement(line))
```
schalkms: I guess this condition is too complex. Please be sure to refactor this. We have a rule `ComplexCondition` which checks for cases like this.

README.md
vanniktech: I'm aware of that. I was still curious whether this has always been updated correctly in the past, or not.

schalkms: @vanniktech Artur specified a checklist for that. Before each new release this batch gets updated. See https://github.com/arturbosch/detekt/blob/master/CONTRIBUTING.md

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLengthSpec.kt
rvoortman: I think a separate testcase might be more descriptive.

schalkms: You are absolutely right. There should be a separate test case. I don't want copy&paste test code. It's also possible to put them in a separate test file.

rvoortman: I don't think that is needed, but I will update the code in a second and we can check if we really want that.

schalkms:
Looks good!
LGTM

Issue #844, opened 2018-04-05
Should ObjectPropertyNaming have a constantPattern?


scottkennedy:
`TopLevelPropertyNaming` has `propertyPattern` and `constantPattern`, but `ObjectPropertyNaming` only has `propertyPattern`, even though you can put constants in objects.

Should there be a way to specify a different pattern for constants than properties?

schalkms:
Good idea!

schalkms:
Having looked into this issue I can report that `propertyNaming` already handles this. Please take a look at `TopLevelPropertyNamingSpec.kt`. Anyways, in this source file we mixed up test cases for two separate rules.

schalkms:
I reopen this to wait for verification (@arturbosch @Mauin)

arturbosch:
it makes sense to have a constant and a property pattern on this one - just to be more configurable.
I think why I just used one `propertypattern` for this is because intellij also just uses that regex.

schalkms:
Wouldn't we want to have the same naming patterns as IntelliJ?

arturbosch:
Yes, leaving both configuration options on default being as intellij is enough don't you think?

schalkms:
Ok, so that both patterns have the same defaults.

schalkms:
Closed with c058187

PR #850, opened 2018-04-10
Splitted test cases for NamingRulesSpec

schalkms:
As described in #844 we mixed up test cases for two separate rules and this commit fixes it.


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/TopLevelPropertyNamingSpec.kt
```diff
@@ -41,7 +41,7 @@ class TopLevelPropertyNamingSpec : SubjectSpek<NamingRules>({
             val serialVersionUID = 42L
 		""")
 
-		val findings = subject.lint(code)
+		val findings = TopLevelPropertyNaming().lint(code)
```
arturbosch: This can be `subject`

PR #855, opened 2018-04-13
Implement detekt-formatting, a wrapper over ktlint

arturbosch:
I tried out an idea and this came out ^^ 
I would say for intern use only (for now) :)


arturbosch:
@Mauin @schalkms 
Ok, this one is huge :).
We support formatting now through ktlint with their editor config options via detekt config.
I renamed ModifierOrder and MaxLineLength so we can have both rules in formatting and style for now (user migration).
I deduplicated the jcommander logic through a dependency on detekt-cli for generator and one usage of reflection.

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/MaximumLineLength.kt
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * See https://ktlint.github.io for documentation.
+ *
+ * @configuration maxLineLength - maximum line length (default: 120)
+ *
+ * @active since v1.0.0
+ * @autoCorrect since v1.0.0
+ * @author Artur Bosch
+ */
+class MaximumLineLength(config: Config) : FormattingRule(config) {
```
schalkms: I guess detekt's `MaxLineLengthRule` can then be removed.

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleCollectorSpec.kt
```diff
@@ -143,7 +167,7 @@ class RuleCollectorSpec : SubjectSpek<RuleCollector>({
 			assertThat(items[0].active).isFalse()
 		}
 
-		it("is active if the tag is there") {
+		it("is active and auto-correctable if tags are present") {
```
schalkms: Does this test two things? If yes, we should split this test case.

detekt-formatting/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+	implementation.extendsFrom kotlinImplementation
+	testImplementation.extendsFrom kotlinTest
+	compile {
+		transitive = false
+	}
+}
+
+dependencies {
+	compileOnly project(':detekt-api')
+	compile("com.github.shyiko.ktlint:ktlint-ruleset-standard:0.21.0") {
```
Mauin: As we will probably update the ktlint version regularly we should pull it out into `gradle.properties`.

Mauin: ktlint also released v0.22.0 recently

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingProvider.kt
```diff
@@ -0,0 +1,21 @@
+package io.gitlab.arturbosch.detekt.formatting
+
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.RuleSet
+import io.gitlab.arturbosch.detekt.api.RuleSetProvider
+
+/**
+ * This rule set provides rules that address formatting issues.
+ *
+ * @configuration android - if android style guides should be preferred (default: false)
```
Mauin: These configurations here are a bit confusing as they're not set/used in the code of this provider.

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/KtLintMultiRule.kt
```diff
@@ -0,0 +1,91 @@
...
+			SpacingAroundComma(config),
+			SpacingAroundCurly(config),
+			SpacingAroundKeyword(config),
+			SpacingAroundOperators(config),
+			SpacingAroundRangeOperator(config),
+			StringTemplate(config)
+	)
+
+	override fun visit(root: KtFile) {
+		val sortedRules = activeRules.sortedBy { it.lastModifier() }
```
Mauin: Why do we need to sort the rules?

detekt-generator/build.gradle.kts
```diff
@@ -33,7 +33,8 @@ val generateDocumentation by tasks.creating {
             args = listOf(
                     "${rootProject.rootDir}/detekt-generator/build/libs/detekt-generator-$detektVersion-all.jar",
                     "--input",
-                    "${rootProject.rootDir}/detekt-rules/src/main/kotlin",
+                    "${rootProject.rootDir}/detekt-rules/src/main/kotlin" + "," +
```
Mauin: String interpolation

docs/pages/documentation/formatting.md
```diff
@@ -0,0 +1,135 @@
...
+keywords: rules, formatting
+permalink: formatting.html
+toc: true
+folder: documentation
+---
+This rule set provides rules that address formatting issues.
+
+### ChainWrapping
+
+See https://ktlint.github.io for documentation.
```
Mauin: Not sure this "documentation" really helps. Should we link directly to the documentation of the specific rule?

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingProvider.kt
arturbosch: KtLint is using a EditorConfig object which is passed to all the FileAstNode's. So basically all rules can use this property, but only 1-2 are using it right now like MaxLineLength.
Instead of introducing a property per rule I implemented it as a ruleset property (all rules have the ruleset config and can retrieve it). 
Hope it is a bit clearer now. Yes it's bad that is looks like it is not used here.

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleCollectorSpec.kt
arturbosch: Good catch, I had the same feelings about this but its better than having 90% copy paste for just one other line :/

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/MaximumLineLength.kt
arturbosch: Yep, we can look if many will use detekt-formatting and delete it for 1.0 or later ^^

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/KtLintMultiRule.kt
arturbosch: KtLint has marker interfaces thatindicate if a rule must run at the end after all other

docs/pages/documentation/formatting.md
arturbosch: That's actually possible, the website uses links liek https://ktlint.github.io/#rule-indentation

arturbosch:
I understand your concerns and the PR looks huge.
Thats because I merged the cli args logic and extended the generator for ruleset configurations.
Nevertheless we make formatting configurable and having all things in tool and in one config is a good thing I think. Ofc I would say let the users evaluate and say if it helps them. (for example the guys from arrow still use detekt M13 because of the formatting und our other checks in one tool.)

Issue #859, opened 2018-04-15
Ignore deprecated functions for TooManyFunctions rule


arturbosch:
## Expected Behavior
Deprecated functions have to stay or X vesions before they can be removed. You cannot just refactor to have less functions.
Allow to ignore deprecated functions through a property.

## Current Behavior
@Deprecated functions are counted.

## Context
As a library provider I have no chance to reduce the functions count easily

PR #867, opened 2018-04-17
Ignored deprecated functions for TooManyFunctions rule

schalkms:
Should detekt ignore deprecated functions per default or provide a config option?

reference #859


arturbosch:
We should go for a config property as always. We are highly configurable :)

Issue #866, opened 2018-04-17
Allow top level properties to be named as constants


emartynov:
Here is a pattern from the project that we use quite often:
```private val readyStates = listOf(ViewState.READY_IDLE, ViewState.READY_REFRESHING, ViewState.READY_ERROR)```

## Expected Behavior
I can not use `const` keyword but, during the run, it is technically a constant. I would expect from Detekt not to report naming issue for it.

## Current Behavior
However, I get
```TopLevelPropertyNaming - [READY_STATES] at src/main/java/...``` 

## Context
Here is another point of view that it should be treated as a constant:
https://kotlinlang.org/docs/reference/coding-conventions.html#property-names

**Comment by arturbosch** [comment]
Hm, when I reworked the naming rules, I swear intellij did not allowed just properties to be uppercase, now they do :). Thanks for reporting!

**Comment by geralt-encore** [comment]
It's more complicated than it looks on a first glance. The thing is that according to Kotlin coding conventions uppercase underscore-separated names should be used for **immutable** top-level properties. While mutable **top-level** are still named using camel case. So looks like that type solving is needed in order to be able to solve this issue properly.

**Comment by schalkms** [comment]
I tested this locally by refactoring the respective test cases. However, I was not able to reproduce this.
Are you using the latest version?

**Comment by arturbosch** [comment]
If it is a `val` it is technically immutable :) ?
@schalkms if you install the detekt-intellij-plugin and have on toplevel `val STUFF = ""` detekt even reports it :). Howeve I'm not sure if we fixed this already on master :o 
Will check it out later.

**Comment by olivierlemasle** [comment]
@arturbosch No, the Kotlin Coding Conventions specify [what they call constants](https://kotlinlang.org/docs/reference/coding-conventions.html#property-names): "properties marked with const, or top-level or object val properties with no custom get function that hold **deeply immutable data**"

Which means that
```kotlin
val mutableCollection: MutableSet<String> = HashSet()
```
should not be named `MUTABLE_COLLECTION` (their example), even if it's a `val`.

However, @emartynov's example is deeply immutable data (if I suppose `ViewState` is an `enum`) so it should have an uppercase underscore-separated name:
```kotlin
private val READY_STATES = listOf(ViewState.READY_IDLE, ViewState.READY_REFRESHING, ViewState.READY_ERROR)
```
I agree with @geralt-encore, we won't be able to follow the Kotlin Coding Conventions strictly without type resolution...

arturbosch:
[...]

PR #872, opened 2018-04-19
Adjust top level property naming to be on pair with intellij - #866

arturbosch:


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/TopLevelPropertyNamingSpec.kt
```diff
@@ -35,25 +35,26 @@ class TopLevelPropertyNamingSpec : SubjectSpek<TopLevelPropertyNaming>({
...
+		it("should report non private property using underscore") {
 			val code = compileContentForTest("""
-				val MY_NAME = "Artur"
-        	    val MYNAME = "Artur"
-        	    val MyNAME = "Artur"
 				val _nAme = "Artur"
-				private val NAME = "Artur"
 			""")
-			assertThat(subject.lint(code)).hasSize(5)
+			assertThat(subject.lint(code)).hasSize(1)
```
arturbosch: I can't come up with any other cases :/

schalkms: I don't think it's necessary to add a another test case.
Anyways, special characters like **èà ... ** are also valid from position 2 to n.

PR #885, opened 2018-05-01
Fix default value handling in composite config - Closes #881

### LINKED ISSUE #881: CompositeConfig does not allow forcing the default value
## Expected Behavior

Ideally in the "lookFirst" config, if a boolean (as an example) was set explicitly to the same value as the default (but not the same as "lookSecond"), it should take precedence.

## Current Behavior

A condition in the CompositeConfig prevents this by comparing the retrieved value against the default value. If they are matched, it defers to the secondary config for a value.

```
override fun <T : Any> valueOrDefault(key: String, default: T): T {
  val firstResult = lookFirst.valueOrDefault(key, default)
  return if (firstResult != default) firstResult else lookSecond.valueOrDefault(key, default)
}
```

Ideally this would instead be something like: `if(!lookFirst.isValueSet(key)) { ... }` so that the config will only look at the secondary case if the value is not set in the primary case.

As a specific example, consider a project that has both `detekt-config-tweaks.yml` and `default-detekt-config.yml`, and the tweaks file has only these contents:

```
style:
  NewLineAtEndOfFile:
    active: false
```

If running with `--config detekt-config-tweaks.yml,default-detekt-config.yml` the result will be `style.NewLineAtEndOfFile.active: true` when the desired output would be `active: false`. There doesn't appear to be anyway to bypass this behavior with the current implementation.

## Context

We are attempting to apply minor tweaks to the defaults in our customized file so we don't have to copy the full default file for our projects, but this has prevented that as we want to disable a rule that is on by default, and we cannot.

## Your Environment
* Version used: RC6-4
* Operating System and version: OSX

arturbosch:


vanniktech:
A little bird told me it's RC6-5 time. :shipit:

arturbosch:
I wanted to release it on sunday but realized that kotlin-dsl or whatever in gradle broke our plugin :(

PR #888, opened 2018-05-02
Disabled UndocumentedPublicFunction for non-public classes

Miha-x64:
Consider the following sample:
```kt
internal class Whatever {
    fun doSomething() = ...
}
```
Formal visibility modifier of `doSomething` is `public`, but the function effectively-internal.

This PR avoids triggering `UndocumentedPublicFunction` for such declarations.


Miha-x64:
Done. (without creating new test case)

PR #892, opened 2018-05-07
Correct test case's description

hungyanbin:
I found that some test case's descriptions are the same, and the second one with the same description would not be executed.

BTW, it is my first time making pull request. I apologize for that if I make any mistakes.


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DuplicateCaseInWhenExpressionSpec.kt
```diff
@@ -27,7 +27,7 @@ class DuplicateCaseInWhenExpressionSpec : SubjectSpek<DuplicateCaseInWhenExpress
 			assertThat(subject.lint(code)).hasSize(1)
 		}
 
-		it("reports duplicated label in when") {
+		it("does not reports duplicated label in when") {
```
schalkms: does not report..

PR #897, opened 2018-05-08
Exclude ForEachOnRange in default config for test-pattern.

vanniktech:
Let me know what you think about this.

Motiviation: Performance cost is no problem in tests.


arturbosch:
I'm totally OK with this change ^^

Have you forked from current master? Or are you using 4.7?
Gradle 4.7 broke our gradle-plugin, I reverted the changes to release a new version. For development on detekt please use 4.6.

Issue #906, opened 2018-05-16
RethrowCaughtException Usage


scottkennedy:
I noticed a `RethrowCaughtException` warning in my codebase, and while looking at https://github.com/arturbosch/detekt/blob/dfb6fb615a0a10cc7052dcac285773702115e071/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/RethrowCaughtException.kt#L22, I have a question.

Should this be compliant?

```kotlin
try {
    // ...
} catch (e: IOException) {
    logSomething(e)
    throw e
}
```

It's rethrowing the same exception, but specifically because some special handling is needed there.

I could suppress this usage, but another option could be modifying the rule to only catch cases where the catch block is just `throw e`, with no other exception handling code.

What do you think?

schalkms:
This should be compliant. I'll fix this immediately.

arturbosch:
Should this really be compliant?
I mean whats the point of catching the exception in the first place?

Just logging and doing nothing is bad exception handling imo. Smells like an architectural problem?
Maybe someone can enlighten me :) here please

schalkms:
@arturbosch I quickly did some research. Sonar also treats it that way.
reference (under **exceptions**) https://sbforge.org/sonar/rules/show/squid:S1166?modal=true&layout=false

scottkennedy:
In my case, the exception is handled further up the stack in a more generic case, but we need some special logging in this one specific place.

schalkms:
Closed with 6c57986

scottkennedy:
I'm still seeing this issue with RC8

schalkms:
Okay. Do you mind providing the source code which violates this rule?

scottkennedy:
```kotlin
fun foo() {}
fun bar() {}

fun test() {
    try {
        foo()
    } catch (e: IllegalStateException) {
        bar()
        throw e
    }
}
```

schalkms:
This code should be flagged since you are not doing anything with the caught exception beside rethrowing it.

scottkennedy:
Hmm.  I don't really see how this is any different.  I'm taking specific action here because this exception was thrown, and then I'm re-throwing it to be handled (or not) further up the stack.

The specific case I'm seeing right now is when I encounter an SSL error connecting to a server.  I'm tracking a specific SSL error event, and then rethrowing that exception to be handled the same way as any other network error.

marianosimone:
[...]

PR #907, opened 2018-05-16
Added support for special handling in RethrowCaughtException

schalkms:
reference #906

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/RethrowCaughtExceptionSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/RethrowCaughtExceptionSpec.kt
@@ -1,5 +1,6 @@
 package io.gitlab.arturbosch.detekt.rules.exceptions
 
+import io.gitlab.arturbosch.detekt.rules.Case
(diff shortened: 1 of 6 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/RethrowCaughtExceptionSpec.kt
```diff
@@ -1,43 +1,80 @@
...
+	}
+
+	given("a caught exception that is logged") {
+
+		it("should not report a logged exception") {
+			val code = """
+				fun x() {
+					try {
+					} catch (e: IllegalStateException) {
+						print("log")
```
arturbosch: According to sonar this should get reported, should't it? (https://sbforge.org/sonar/rules/show/squid:S1166?modal=true&layout=false)
```
// Noncompliant - exception is lost (only message is preserved)
try { /* ... */ } catch (Exception e) { LOGGER.info(e.getMessage()); }
```
The exception name must be used in any statement to not lose the exception?

arturbosch: I would say yes, lets how I interpret sonarqube's doc.

PR #876, opened 2018-04-23
fix compile erros in the detekt-gradle-plugin project

Mauin:
While rebasing #771 and trying to check some of the behavior on `master` I couldn't get the project to compile without these changes.

PR #913, opened 2018-05-20
Upgrade to Gradle 4.7

3flex:
I saw this was previously merged in https://github.com/arturbosch/detekt/pull/877, but somehow the changes are no longer in master.


arturbosch:
hey, yes I did the mistake to drop the commits instead of reverting them.
The new version of kotlin-dsl breaks our gradle plugin so we reverted this two commits to release new versions until @Mauin 's gradle rework is done.

3flex:
What broke in the Gradle plugin with 4.7? I had a quick look at the gradle issues raised but couldn't see any related to this, except #876 which looks like it was trying to fix an issue that didn't actually exist.

CI checks on this PR pass...

I'd just close this but I'm trying to work on #485, and using latest Gradle may help (Kotlin DSL includes kotlin-compiler-embedded and there are some changes to the way classpath is calculated in the latest Kotlin DSL versions, so being able to just use the latest Gradle version may help avoid those bugs so having this merged now so I can rebase on top instead of waiting for the Gradle rework will help me)

Alternatively I'll just merge these two commits in my branch for the moment.

3flex:
I'm still confused because detektCheck runs as part of the Travis CI check which passes on this PR... but I'll close, there must be something I'm missing (is it only when running detektCheck on other projects, and not when running on detekt itself?)

Mauin:
Just to confirm: The Gradle Plugin re-work also shows the same issue with Gradle v4.7 and due to bugs with the Gradle Test Kit will require Gradle v4.8 to be released as well.

Issue #912, opened 2018-05-20
Style rule: Prefer to over Pair


jkaan:
This is a very stylistic choice so I am sceptical that it is a good idea to add this to Detekt but I very much prefer creating pairs using this syntax:

`1 to 2` instead of `Pair(1, 2)`. 

Do you guys feel this could be added as an optional style rule?

arturbosch:
Sure, I think this is the idiomic way to create pairs in kotlin, so a rule makes sense ^^

jkaan:
I'll get started working on this one. This one is probably super easy. :)

schalkms:
Closed with c9e188b

PR #914, opened 2018-05-20
Add rule for preferring to syntax for creating pairs

jkaan:
This implements the rule talked about in #912


detekt-rules/src/test/resources/cases/PreferToOverPairSyntaxPositive.kt
```diff
@@ -0,0 +1,8 @@
+package cases
+
+@Suppress("unused", "UNUSED_VARIABLE")
+fun preferToOverPairSyntaxPositive() {
+	val pair1 = Pair(1, 2)
+
+	val pair2: Pair<Int, Int> = Pair(1, 2)
```
schalkms: Please also test this also with a nested pair
e.g. a pair element which contains another pair

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/PreferToOverPairSyntax.kt
```diff
@@ -0,0 +1,51 @@
...
+
+	override fun visitSimpleNameExpression(expression: KtSimpleNameExpression) {
+		val callReference = expression.getQualifiedElementOrCallableRef()
+		if (expression.getReferencedName() == PAIR_CONSTRUCTOR_REFERENCE_NAME &&
+				callReference is KtCallExpression) {
+			val (firstArg, secondArg) = callReference.valueArguments.map { it.text }
+
+			report(CodeSmell(issue, Entity.from(expression),
+					message = "Pair is created by using the pair constructor. " +
+							"This can replaced by $firstArg to $secondArg"))
```
Mauin: can we add backticks into this message? `` `$firstArg to $secondArg` ``

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/PreferToOverPairSyntax.kt
```diff
@@ -0,0 +1,51 @@
...
+		}
+
+		super.visitSimpleNameExpression(expression)
+	}
+
+	companion object {
+		const val PAIR_CONSTRUCTOR_REFERENCE_NAME = "Pair"
```
Mauin: Only issue I can see is codebases that have a custom `Pair` class might accidentally trigger this rule even though their `Pair` doesn't support the `to` notation. But in that case users should just turn off this rule 👍

vanniktech:
Another rule that would benefit from autocorrect.

Issue #915, opened 2018-05-20
Support more range expressions in ForEachOnRange rule (downTo, step, until etc.)


Zayankovsky:
The following code passes **ForEachOnRange** check:
```kotlin
fun loop() {
    (0 until 10).forEach {
        println(it)
    }
}
```

I doubt it is possible to cover the general case of IntRange, but it would be nice if ranges created using **downTo**, **step** and **until** keywords were supported, as they seem pretty common to me

schalkms:
Closed with cb83b36

PR #916, opened 2018-05-20
Support more range expressions in ForEachOnRange rule 

schalkms:
reference #915


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/ForEachOnRange.kt
```diff
@@ -38,28 +47,37 @@ import org.jetbrains.kotlin.psi.psiUtil.getReceiverExpression
...
+			val binaryExpression = parenthesizedExpression?.expression as? KtBinaryExpression
+			if (binaryExpression != null && isRangeOperator(binaryExpression)) {
+				report(CodeSmell(issue, Entity.from(expression), issue.description))
 			}
 		}
 	}
 
-	companion object {
-		val rangeRegex = Regex("\\(.*\\.\\..+\\)")
+	private fun isRangeOperator(binaryExpression: KtBinaryExpression): Boolean {
```
Zayankovsky: Can you please extend this to support statements like `0..10 step 2`?

Issue #854, opened 2018-04-13
Gradle plugin produces empty directory named '--config-resource'


nealsanche:
I have the following configuration:

```
plugins {
    id "io.gitlab.arturbosch.detekt" version "1.0.0.RC6-4"
}
detekt {
    version = "$detektVersion"
    profile("main") {
        input = "$projectDir"
        output = "$projectDir/reports"
        outputName = "varo"
        baseline = "$projectDir/reports/baseline.xml"
        config = "$projectDir/detekt.yml"
        filters = ".*test.*,.*/resources/.*,.*/tmp/.*,.*\\.kts"
    }
}
```
This is in my main configuration, since I am working on an Android application.

## Current Behavior

After running the `detektCheck` gradle task, I find the following directory is most often produced in my source directory:

![screenshot 2018-04-13 15 23 43](https://user-images.githubusercontent.com/5369033/38758566-c707230c-3f2e-11e8-93b1-8f4873e10c87.png)

## Your Environment
* Version used: 1.0.0.RC6-4
* Operating System and version: Mac OS X, 10.13.4 (17E199)
* Android Studio 3.1.1 and Gradle 4.6

arturbosch:
Hey, I also noticed the same problem in my IDE. Will take a look. Thanks for reporting!

TWiStErRob:
That indeed fixes this issue, but the root cause is still there: there's something wrong with argument parsing  or command line building.

PR #924, opened 2018-05-24
Build Gradle plugin using composite build

### LINKED ISSUE #920: Build Gradle plugin using a composite build
Looking to see if there's any opposition to this idea before spending more time exploring it.

The Detekt Gradle plugin is tightly coupled to the rest of the project. It creates a chicken and egg problem when making changes like the Gradle plugin rework, nicely illustrated by this example from the Gradle blog post that introduced the composite build feature:
> 1. Change the library
> 2. Publish the library to a local repository
> 3. Add the local repository to your application’s repositories
> 4. Change your application’s dependency to the new library version
> 5. Test your application
> 6. Repeat until the problem is fixed or you lose your mind

A possible solution is shifting to a composite Gradle build. The Gradle plugin would still live in this repository, but instead of having to push updates to a repository before consuming them in this project, the Gradle plugin could be built then consumed as part of the build itself. This should allow for much faster iterations and also allow the build to succeed in CI for the Gradle rework, and may surface the classpath issues faced in Gradle 4.7 during development and not after pushing a new version to the plugin portal.

See here for background: https://blog.gradle.org/introducing-composite-builds

3flex:
Pros:
* Creates a much improved experience developing the Gradle plugin.
* Easier to publish changes when there are bugs in the Gradle plugin that don't affect detekt (see #807 & #854), since the new plugin can be published without publishing a new detekt version as well

Cons:
* Have to manage releases of the Gradle plugin in addition to releases of detekt itself (though Gradle plugin releases only need to be made if there's a change that needs to be published. It's not necessary every time a new detekt version is published)

I personally feel the pros outweigh the cons. I'm sure @Mauin will too! (PS @Mauin I tested on your branch and seems to work, but had to cherry pick the commits from this PR because we last rebased our own branches at different points on master).

Closes #920


arturbosch:
How should the gradle plugin now run detekt without an explicit dependency on the cli module?
I checked out your branch, incremented the version and got htis:
```
* What went wrong:
Could not determine the dependencies of task ':detektCheck'.
> Could not resolve all task dependencies for configuration ':detekt'.
   > Could not find io.gitlab.arturbosch.detekt:detekt-cli:1.0.0.RC7-2.
     Searched in the following locations:
         https://jcenter.bintray.com/io/gitlab/arturbosch/detekt/detekt-cli/1.0.0.RC7-2/detekt-cli-1.0.0.RC7-2.pom
         https://jcenter.bintray.com/io/gitlab/arturbosch/detekt/detekt-cli/1.0.0.RC7-2/detekt-cli-1.0.0.RC7-2.jar
         file:/home/artur/.m2/repository/io/gitlab/arturbosch/detekt/detekt-cli/1.0.0.RC7-2/detekt-cli-1.0.0.RC7-2.pom
         file:/home/artur/.m2/repository/io/gitlab/arturbosch/detekt/detekt-cli/1.0.0.RC7-2/detekt-cli-1.0.0.RC7-2.jar
         http://dl.bintray.com/jetbrains/spek/io/gitlab/arturbosch/detekt/detekt-cli/1.0.0.RC7-2/detekt-cli-1.0.0.RC7-2.pom
         http://dl.bintray.com/jetbrains/spek/io/gitlab/arturbosch/detekt/detekt-cli/1.0.0.RC7-2/detekt-cli-1.0.0.RC7-2.jar
     Required by:
         project :
```

TWiStErRob:
Here's something that might help:

```gradle
if (gradle.parent != null) { // we're part of a composite build
	gradle.parent.projectsLoaded {
		// at this point the settings.gradle has been executed and probably the gradle files
		// compiled, but not executed. So we only have access to what gradle supports:
		println "Parent: ${gradle.parent}: ${gradle.parent.rootProject.properties}"
	}
	gradle.parent.projectsEvaluated {
		// at this point the outer project has been evaluated
		// and you can see the result of its build.gradle execution
		println "Parent: ${gradle.parent}: ${gradle.parent.rootProject.tasks}"
	}
	// careful, code put here will be executed before the gradle.parent callbacks
}
```

3flex:
If you're making changes to detekt and want to run the build with those changes you'd have to publish the new detekt CLI version somewhere so the Gradle plugin could consume it. Definitely a downside to this setup, but necessary due to the decoupling of the plugin from the main project.

When testing locally, you can do this:
1. Make changes to detekt
2. Update `detektVersion` in gradle.properties
3. Publish with `./gradlew publishToMavenLocal`
4. Update `usedDetektVersion` in gradle.properties to reflect new detekt version
5. Run `./gradlew detektCheck`

arturbosch:
@3flex that's exactly how I test new releases. You do not have missing dependency problems?

3flex:
I went through that sequence on my branch before posting my comment to verify and it appeared to work. The composite build doesn't change how the detekt CLI dependency is resolved when the plugin itself runs, only how the plugin is built.

These lines from your log suggest you maybe didn't publish the version to local, or had a mismatched version somewhere?
```
file:/home/artur/.m2/repository/io/gitlab/arturbosch/detekt/detekt-cli/1.0.0.RC7-2/detekt-cli-1.0.0.RC7-2.pom
file:/home/artur/.m2/repository/io/gitlab/arturbosch/detekt/detekt-cli/1.0.0.RC7-2/detekt-cli-1.0.0.RC7-2.jar
```

arturbosch:
Ah just wrong thinking ... we resolve the detekt-cli module at runtime.
If @Mauin also gives his yes, I will gradly merge this and publish a new release :)

3flex:
Just remember when you release: with this change detekt and detekt-gradle-plugin are now versioned independently, so you'll have to bump both versions before publishing :)

3flex:
@Mauin sorry no, I don't - we branched from master at different points and I couldn't cleanly rebase your branch on top. Sounds like this will be merged soon though, so best just to rebase in one shot once it's in master. Have fun!

Issue #900, opened 2018-05-11
MayBeConst false negative when concatenating strings


vanniktech:
```kotlin
private const val PREF_SUFFIX = "gtar-"
private val KEY_LAST_SHOWN = PREF_SUFFIX + "last_shown"
```

In this case `KEY_LAST_SHOWN` won't get flagged in the `MayBeConst` rule although it can and should be a const.

3flex:
I think this needs symbol resolution? Playing in IntelliJ an error is shown when PREF_SUFFIX isn't a const val, so the detekt rule would have to check that any variable in an expression is a const val. Symbol resolution is needed to check that.

arturbosch:
I would not say this is a bug, rather an improvement.

schalkms:
Closed with 2025fe5

PR #930, opened 2018-05-27
Fixed MayBeConst false negative when concatenating strings #900

schalkms:


schalkms:
I will add this test cases soon.

schalkms:
Okay.

PR #933, opened 2018-05-31
TooManyFunctions : Add option to ignore private functions

tokou:
> Too many functions inside a/an file/class/object/interface always indicate a violation of the single responsibility principle. Maybe the file/class/object/interface wants to manage to many things at once. Extract functionality which clearly belongs together.

While I agree with the general principle that a class with too many functions may violate SRP. I think that this only applies to functions that **expose functionality outside the class** as private functions just lets us organise the code in a cleaner way.

Having private functions count toward the "TooManyFunctions" threshold may deter us from extracting private functions to split big public functions and thus make code less readable.

That's why I suggest adding a _ignorePrivate_ flag to this rule.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
@@ -1,6 +1,5 @@
 package io.gitlab.arturbosch.detekt.rules.complexity
 
-import com.sun.tools.javac.code.Flags.DEPRECATED
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Debt
 import io.gitlab.arturbosch.detekt.api.Entity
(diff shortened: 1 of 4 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
```diff
@@ -65,7 +68,7 @@ class TooManyFunctions(config: Config = Config.empty) : Rule(config) {
 	}
 
 	override fun visitNamedFunction(function: KtNamedFunction) {
-		if (function.isTopLevel && ignoresDeprecatedFunction(function)) {
+		if (function.isTopLevel && ignoresDeprecatedFunction(function) && ignoresPrivateFunction(function)) {
```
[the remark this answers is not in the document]
Mauin: Sounds great!

Mauin:
The `isIgnoredFunction` now has an issue reported for `ReturnCount` :/

`ReturnCount - [isIgnoredFunction] at /home/travis/build/arturbosch/detekt/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt:125:2
 `

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
```diff
@@ -1,5 +1,6 @@
 package io.gitlab.arturbosch.detekt.rules.complexity
 
+import com.sun.tools.javac.code.Flags.DEPRECATED
```
Mauin: This import is new?

tokou: Sorry, just IDEA going wild

Issue #944, opened 2018-06-05
Execution of ForbiddenComment fails when launched via sonarqube task 


madorb:
## Expected Behavior
Invoking `detekt` via `sonarqube` completes as expected.

## Current Behavior
Exception is thrown when invoking `detekt` as part of the `sonarqube` task

## Steps to Reproduce (for bugs)
When i run detekt as `./gradlew detektCheck` it succeeds, when it's invoked as part of the overall sonarqube execution `./gradlew sonarqube` it results in a failure.

## Context
It _appears_ that the exception is thrown if/when any of my `build.gradle.kts` contain any of the `ForbiddenComment`. 

My build is a multi-module gradle project, and i have a `build.gradle.kts` in the root directory as well as multiple others in subdirectories such as `application/web/build.gradle.kts`. Having a ForbiddenComment in the root `build.gradle.kts` seems to not cause an expception to be thrown, but having one in any of the submodule build files DOES. Perhaps it's not properly handling the paths of the nested submodule build files? 

Disabling ForbiddenComment allows the task to complete as expected. 

## Your Environment
Using Gradle 2.8 (though this existed in 2.7) with the gradle kotlin-dsl and Kotlin 1.2.41 

```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':sonarqube'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:110)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.OutputDirectoryCreatingTaskExecuter.execute(OutputDirectoryCreatingTaskExecuter.java:51)
        at org.gradle.api.internal.tasks.execution.SkipCachedTaskExecuter.execute(SkipCachedTaskExecuter.java:105)
        at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:59)
        at org.gradle.api.internal.tasks.execution.ResolveTaskOutputCachingStateExecuter.execute(ResolveTaskOutputCachingStateExecuter.java:54)
        at org.gradle.api.internal.tasks.execution.ResolveBuildCacheKeyExecuter.execute(ResolveBuildCacheKeyExecuter.java:66)
[...]

Issue #793, opened 2018-03-13
NoSuchElementException when running from build.gradle.kts


cldfzn:
I migrated my build script from groovy to kotlin and afterwards detekt fails. If I disable the LargeClass rule things work again. (java 1.8.0_144)
```kotlin
plugins {
    java
    kotlin("jvm") version "1.2.30"
    id("io.gitlab.arturbosch.detekt") version "1.0.0.RC6-3"
    id("io.spring.dependency-management") version "1.0.4.RELEASE"
    id("org.springframework.boot") version "1.5.9.RELEASE"
}
```
```kotlin
detekt {
    profile("main", Action {
        config = "${project.projectDir}/detekt-config.yml"
        baseline = "${project.projectDir}/detekt-baseline.xml"
        filters = ".*test.*,.*/resources/.*"
    })
}
```
```bash
Exception in thread "main" java.util.concurrent.CompletionException: java.util.NoSuchElementException
        at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
        at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
        at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1592)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.util.NoSuchElementException
        at java.util.ArrayDeque.removeFirst(ArrayDeque.java:280)
        at java.util.ArrayDeque.pop(ArrayDeque.java:517)
        at io.gitlab.arturbosch.detekt.rules.complexity.LargeClass.addToHead(LargeClass.kt:51)
        at io.gitlab.arturbosch.detekt.rules.complexity.LargeClass.visitCallExpression(LargeClass.kt:119)
        at org.jetbrains.kotlin.psi.KtVisitorVoid.visitCallExpression(KtVisitorVoid.java:797)
        at org.jetbrains.kotlin.psi.KtVisitorVoid.visitCallExpression(KtVisitorVoid.java:21)
        at org.jetbrains.kotlin.psi.KtCallExpression.accept(KtCallExpression.java:35)
        at org.jetbrains.kotlin.psi.KtElementImpl.accept(KtElementImpl.java:50)
[...]

PR #946, opened 2018-06-09
Do not crash on rule exceptions - #793 #944

arturbosch:


schalkms:
@vanniktech good idea

arturbosch:
Good idea! -> Done
Now printing the file path, detekt and java version plus the os name and stacktrace on rule exceptions!

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DebugUtils.kt
```diff
@@ -0,0 +1,27 @@
...
+
+fun whichOS(): String = System.getProperty("os.name")
+
+fun whichJava(): String = System.getProperty("java.runtime.version")
+
+fun whichDetekt(): String {
+	for (resource in Detektor::class.java.classLoader
+			.getResources("META-INF/MANIFEST.MF")) {
+		try {
+			Manifest(resource.openStream())
```
vanniktech: Do we need to close the stream?

arturbosch: Your right, thanks!

PR #947, opened 2018-06-09
Suppress undocumented classes in dokka

arturbosch:
Some dokka warnings still spam the console:
- https://github.com/Kotlin/dokka/issues/269
- https://github.com/Kotlin/dokka/issues/229


build.gradle.kts
```diff
@@ -103,6 +103,9 @@ subprojects {
 	}
 
 	tasks.withType(DokkaTask::class.java) {
+		// suppresses undocumented classes but now dokka warnings
```
schalkms: *not*

arturbosch: Thanks!

PR #952, opened 2018-06-11
Fix #950 ExpressionBodySyntax

### LINKED ISSUE #950: ExpressionBodySyntax should allow for exclusion on line wrapping
## Expected Behavior
According to the official [Kotlin styleguide](https://android.github.io/kotlin-guides/style.html#expression-functions):

> Expression functions should not wrap to two lines. If an expression function grows to require wrapping, use a normal function body, a return declaration, and normal expression wrapping rules instead.

## Current Behavior

`ExpressionBodySyntax` doesn't take this into account, and reports violations even if there are line wraps:

```
fun doALot(foo: String): String {
    return foo
         .replace("a", "0")
         .replace("b", "1")
}
````

Slightly more complicated is the fact that, sometimes, the expression does fit in a line, but couldn't be made a expression body because it would break another rule (like max line length):

I couldn't put this in a line if my max line length is 50, for example

```
fun doAVeryLittle(foo: String) : String {
    return foo.replace("a", "0")
}
```

schalkms:


schalkms:
@vanniktech good idea!
Per default **detekt** adheres to the Kotlin style guide.

Issue #936, opened 2018-06-01
Use native junitPlatform from gradle 4.6


arturbosch:
Junit platform got deprecated in favor of https://docs.gradle.org/current/userguide/java_testing.html#using_junit5.

For now we see deprecation warnings when using version 1.2.

PR #953, opened 2018-06-12
Use native junitPlatform from gradle 4.6

epeee:
The 2nd part of  #936.
This pr makes use of native junitPlatform from gradle 4.6 in projects other than detekt-gradle-plugin project.


3flex:
Nice. I believe that junit-platform-console depends on junit-platform-launcher (and the gradle plugin depends on console) so it should be safe to remove both to tidy the build files a bit further.

https://junit.org/junit5/docs/current/user-guide/#dependency-diagram

Suggest trying it and ensuring the relevant tests are still run.

epeee:
@3flex thx for your feedback.
The reason why I did not remove junit-platform-launcher yet is that some IDE's have problems while executing tests (e.g. see https://bugs.eclipse.org/bugs/show_bug.cgi?id=526502).

arturbosch:
Thanks! Merged manually 9d3a69e3d9fb16b8ce3b8a98f42c2e4dd4050db7

PR #965, opened 2018-06-22
Fix SerialVersionUIDInSerializableClass when value's less then zero #964

### LINKED ISSUE #964: SerialVersionUIDInSerializableClass reports when const value less then true
Steps to reproduce:
1. Create new kotlin class implementing Serializable
2. Create long const with a value that is less then zero
```kotlin
const val serialVersionUID = -1L
```
3. Add RC7-2 Detekt to the project
4. Run `detektCheck`
5. Look at console report
6. Find a report about breaking the SerialVersionUIDInSerializableClass rule.

I have fix for that adding second condition, which checks if the PsiElement is KtPrefixExpression in the method SerialVersionUIDInSerializableClass::hasLongAssignment
```kotlin
private fun hasLongAssignment(property: KtProperty): Boolean {
		val assignmentText = property.children
				.singleOrNull { it is KtConstantExpression || it is KtPrefixExpression }?.text
		return assignmentText != null && assignmentText.last() == 'L'
				&& assignmentText.substring(0, assignmentText.length-1).toLongOrNull() != null
	}
```

samarjan92:
Fixes #964

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,7 @@ If you contributed to detekt but your name is not in the list, please feel free 
 - [Robbin Voortman](https://github.com/rvoortman) - Rule improvement: MaxLineLength
 - [Mike Gorunov](http://github.com/Miha-x64/) — Rule improvement: UndocumentedPublicFunction
 - [Joey Kaan](https://github.com/jkaan) - New rule: MandatoryBracesIfStatements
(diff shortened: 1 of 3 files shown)


samarjan92:
Thanks

I forgot to update the contributor badge(

schalkms:
That’s not really a problem. It’s often not in sync with the real contributers list, but it will be updated with every new release.

Issue #948, opened 2018-06-09
Rule: find unused private top level properties, constants or functions


vanniktech:
The following does not yield a UnusedMemberDeclaration warning:

```kotlin
fun foo(map: Array<IntArray>) {
  for (i in 0 until map.size) {
    for (j in 0 until map[i].size) {
      map[i][i] += 1
    }
  }
}
```

arturbosch:
maybe because it is a top level function and `UnusedMemberDeclaration` only detects members?
So this would be a feature request? :D

vanniktech:
Huh indeed.

vanniktech:
Wrapping inside a class does trigger the rule:

```kotlin
class Foo {
  fun foo(map: Array<IntArray>) {
    for (i in 0 until map.size) {
      for (j in 0 until map[i].size) {
        map[i][i] += 1
      }
    }
  }
}
```

marianosimone:
I'd love jump in and work on something like this, but I'm a bit lost: I can't even find `UnusedMemberDeclaration`

vanniktech:
https://github.com/arturbosch/detekt/blob/ad1808b899aacd842e434704aa451145317cf211/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt

schalkms:
@marianosimone contributions are always welcome 👍 
As mentioned `UnusedPrivateMember.kt` is the way to go. I would add this feature in a new rule. Some of the functionality of `UnusedPrivateMember` might even be reused for the new rule.

marianosimone:
I've been working on this, based on what `UnusedPrivateMember` does, and I wanted to get some feedback on what I'm thinking about before creating a PR:

`UnusedPrivateMember` seems to be doing quite a bit, and not everything is related to the rest. From the docs:

>  Reports unused private properties, function parameters and functions.

So far, this detects:
- Unused private top-level functions
- Unused private member functions
- Unused function parameters (only for non-top-level functions)
- Unused private properties (non-top-level)
- Unused private constants (non-top-level)
- Unused variable declaration inside functions (only for non-top-level functions)

The way I see it, detecting "Unused private top-level functions" is the odd one there: everything else only works when dealing with non-top-level.

Also, IMO, detecting "unused parameters" is out of place, as there's no concept of private vs non private here.

A proposal:
[...]

PR #968, opened 2018-06-26
UnusedPrivateMember: detect top level declarations

marianosimone:
This is one approach to allow `UnusedPrivateMember` to look at top level declarations as well.

Initial discussion in #948

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
@@ -21,8 +21,10 @@ import org.jetbrains.kotlin.psi.KtNamedDeclaration
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtOperationReferenceExpression
 import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.KtPrimaryConstructor
 import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.KtReferenceExpression
+import org.jetbrains.kotlin.psi.KtSecondaryConstructor
 import org.jetbrains.kotlin.psi.psiUtil.isPrivate
 import org.jetbrains.kotlin.psi.psiUtil.referenceExpression
 
@@ -59,7 +61,7 @@ class UnusedPrivateMember(config: Config = Config.empty) : Rule(config) {
 		root.accept(propertyVisitor)
 
 		propertyVisitor.getUnusedProperties().forEach {
-			report(CodeSmell(issue, Entity.from(it.value), "Private property ${it.key} is unused."))
+			report(CodeSmell(issue, Entity.from(it), "Private property ${it.nameAsSafeName.identifier} is unused."))
 		}
 	}
 
@@ -73,17 +75,11 @@ class UnusedPrivateMember(config: Config = Config.empty) : Rule(config) {
 
 	private class UnusedPropertyVisitor(private val allowedNames: Regex) : DetektVisitor() {
 
-		private val properties = mutableMapOf<String, KtElement>()
+		private val properties = mutableSetOf<KtNamedDeclaration>()
 		private val nameAccesses = mutableSetOf<String>()
 
-		fun getUnusedProperties(): Map<String, KtElement> {
-			for (access in nameAccesses) {
-				if (properties.isEmpty()) {
-					break
-				}
-				properties.remove(access)
-			}
-			return properties
+		fun getUnusedProperties(): List<KtNamedDeclaration> {
+			return properties.filter { it.nameAsSafeName.identifier !in nameAccesses }
 		}
 
 		override fun visitParameter(parameter: KtParameter) {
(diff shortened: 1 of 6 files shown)


marianosimone:
@schalkms this still doesn't handle detecting private classes, but wanted to get initial feedback before going down that path.

It ended up being less complicated than my initial implementation, mainly because all the detection was already in place, and just needed to change the conditions under which they happen (note that there are no new visitors or anything)

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -103,17 +107,14 @@ class UnusedPrivateMember(config: Config = Config.empty) : Rule(config) {
 		}
 
 		override fun visitProperty(property: KtProperty) {
-			if ((property.isPrivate() && property.isNonNestedMember())
+			if ((property.isPrivate() && property.isMemberOrTopLevel())
```
marianosimone: I did some testing around this, and didn't really get why it was trying to exclude `nonNestedMember`s.

I added some tests for the new functionality.

Note that, before this, something like:

```
class A {
    class B {
       private val silentlyIgnored = 1
     }
}
```
wouldn't be detected

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
```diff
@@ -177,7 +177,7 @@ class UnusedPrivateMemberSpec : SubjectSpek<UnusedPrivateMember>({
 			val code = """
 				fun RuleSetProvider.provided() = ruleSetId in defaultRuleSetIds
 
-				private val defaultRuleSetIds = listOf("comments", "complexity", "empty-blocks",
```
marianosimone: had to remove the private, as now the rule would complain about it

detekt-rules/src/test/resources/cases/UnusedPrivateMemberNegative.kt
```diff
@@ -80,6 +80,6 @@ val stuff = object : Iterator<String?> {
 	override fun hasNext(): Boolean = true
 }
 
-fun main(args: Array<String>) {
```
marianosimone: had to make this one ignored, or it would be detected

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
schalkms: I think @Mauin could help out here.

detekt-rules/src/test/resources/cases/UnusedPrivateMemberNegative.kt
schalkms: We might want to ignore the main method declaration.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -95,25 +96,33 @@ class UnusedPrivateMember(config: Config = Config.empty) : Rule(config) {
 			}
 		}
 
+		override fun visitPrimaryConstructor(constructor: KtPrimaryConstructor) {
```
marianosimone: Added this functionality (and think it's one of the most useful ones!.. me and my team usually do a bad job at removing no-longer used parameters from constructors)

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -51,34 +54,32 @@ class UnusedPrivateMember(config: Config = Config.empty) : Rule(config) {
...
 		}
 
 		super.visitClassOrObject(classOrObject)
 	}
 
-	class UnusedPropertyVisitor(private val allowedNames: Regex) : DetektVisitor() {
+	private class UnusedPropertyVisitor(private val allowedNames: Regex) : DetektVisitor() {
 
-		private val properties = mutableMapOf<String, KtElement>()
+		private val properties = mutableSetOf<KtNamedDeclaration>()
```
marianosimone: Using a map here no longer works, as the visitor could be looking at clashing names in different scopes. We can just use a set of the declarations, filter it before returning it, and just report from there

schalkms: Okay

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
schalkms: Definitely! I like this feature.

schalkms:
Is this still work in progress? If not, I’ll approve this PR then.

marianosimone:
@schalkms sorry.. no, I think this PR is good to go. I'll work on adding unused class detection in a separate one

Issue #974, opened 2018-06-28
Activate more rules for detekt self analysis


arturbosch:
- we have plenty of rules which are not used right now
- we should enable as many useful ones as possible and fix their findings
- also enable detekt-formatting

PR #975, opened 2018-06-29
Activate rules in failfast.yml

schalkms:
Reference #974


arturbosch:
Thanks for starting the issue with some candidates!
Please move the configuration to the `failfast.yml`. We do not actually want to activate more rules per default :)

schalkms:
Alright. I misunderstood that.

arturbosch:
Thanks!

PR #977, opened 2018-06-30
Fix ExpressionBodySyntax when multiline expression

### LINKED ISSUE #976: ExpressionBodySyntax: false positive with includeLineWrapping in multiline expression
For something like:

```
fun foo(bar: Int): Int {
    return when (bar) {
        0 -> 1
        1 -> 2
        else -> 3
    }
}
```

## Expected Behavior
No smell should be reported

## Current Behavior
A smell is reported

From #950

marianosimone:
Check the body to see if it's multiline, instead of the
return expression, which might be multiline, but with
none of its children having a \n

Fixes #976

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntax.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntax.kt
@@ -53,9 +53,9 @@ class ExpressionBodySyntax(config: Config = Config.empty) : Rule(config) {
 	override fun visitNamedFunction(function: KtNamedFunction) {
 		if (function.bodyExpression != null) {
 			val body = function.bodyExpression!!
-			body.singleReturnStatement()?.let { returnStmt ->
-				if (includeLineWrapping || !containsWhiteSpace(returnStmt.returnedExpression)) {
-					report(CodeSmell(issue, Entity.from(returnStmt), issue.description))
+			body.singleReturnStatement()?.run {
+				if (includeLineWrapping || !isLineWrapped(body)) {
+					report(CodeSmell(issue, Entity.from(body), issue.description))
 				}
 			}
 		}
@@ -63,17 +63,15 @@ class ExpressionBodySyntax(config: Config = Config.empty) : Rule(config) {
 
 	private fun KtExpression.singleReturnStatement(): KtReturnExpression? {
 		val statements = (this as? KtBlockExpression)?.statements
-		return statements?.size?.let {
-			if (it == 1 && statements[0] is KtReturnExpression) {
-				return statements[0] as KtReturnExpression
-			} else null
-		}
+		return statements
+				?.takeIf { it.size == 1 }
+				?.let { it[0] as? KtReturnExpression }
 	}
 
-	private fun containsWhiteSpace(expression: KtExpression?): Boolean {
-		return expression?.children?.any {
+	private fun isLineWrapped(expression: KtExpression): Boolean {
+		return expression.children.any {
 			it.text.contains('\n')
-		} == true
+		}
 	}
 
 	companion object {
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntaxSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntaxSpec.kt
@@ -51,7 +51,7 @@ class ExpressionBodySyntaxSpec : SubjectSpek<ExpressionBodySyntax>({
 					if (true) return true
 					return false
 				}
-			""")).hasSize(0)
+			""")).isEmpty()
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntax.kt
```diff
@@ -51,29 +51,26 @@ class ExpressionBodySyntax(config: Config = Config.empty) : Rule(config) {
 	private val includeLineWrapping = valueOrDefault(INCLUDE_LINE_WRAPPING, false)
 
 	override fun visitNamedFunction(function: KtNamedFunction) {
-		if (function.bodyExpression != null) {
-			val body = function.bodyExpression!!
-			body.singleReturnStatement()?.let { returnStmt ->
-				if (includeLineWrapping || !containsWhiteSpace(returnStmt.returnedExpression)) {
-					report(CodeSmell(issue, Entity.from(returnStmt), issue.description))
+		function.bodyExpression?.run body@{
```
marianosimone: I don't know how people feel about this syntax, but I much prefer it vs using `if` and `!!`

marianosimone: I don't know how people feel about the use of labels like this.. but I prefer it over `if` and `!!`

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntaxSpec.kt
```diff
@@ -51,7 +51,7 @@ class ExpressionBodySyntaxSpec : SubjectSpek<ExpressionBodySyntax>({
 					if (true) return true
 					return false
 				}
-			""")).hasSize(0)
+			""")).isEmpty()
```
marianosimone: feels more semantic to use `isEmpty` than `hasSize(0)` :man_shrugging:

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntax.kt
schalkms: It increases the complexity of the code. I prefer the previously used style.
We also have a rule for reporting labeled expressions.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntaxSpec.kt
```diff
@@ -65,7 +65,27 @@ class ExpressionBodySyntaxSpec : SubjectSpek<ExpressionBodySyntax>({
...
+	}
+
+	given("several return statements with multiline expression") {
+
+		val code = """
+			fun stuff(val arg: Int): Int {
+				return when(arg) {
+					0 -> 0
+					else -> 1
+				}
```
schalkms: Can we also test this with `return if .. else ..` ?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntax.kt
```diff
@@ -51,29 +51,26 @@ class ExpressionBodySyntax(config: Config = Config.empty) : Rule(config) {
...
 	private fun KtExpression.singleReturnStatement(): KtReturnExpression? {
 		val statements = (this as? KtBlockExpression)?.statements
-		return statements?.size?.let {
-			if (it == 1 && statements[0] is KtReturnExpression) {
-				return statements[0] as KtReturnExpression
-			} else null
-		}
+		return statements
+				?.takeIf { it.size == 1 }
+				?.let { it[0] as? KtReturnExpression }
```
schalkms: Refactored nicely

schalkms:
Thanks for the quick fix 👍

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntaxSpec.kt
marianosimone: done!

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntax.kt
marianosimone: Ok.. rolled back this (but did change the `let` for `run`... specially since we don't need the `returnStmt` any more, I think `run` is more idiomatic

marianosimone: Interestingly enough, `detecktCheck` did not complain about the labeled expressions... maybe there's something wrong with the config?

marianosimone:
changes applied

Issue #657, opened 2018-01-02
Are there any Rules for inspection a `var` could be `val`?


ChrisZou:
Intellij/AndroidStudio has a inspection that inspects if a local variable declared with `var` could be changed to declaring with `val`. I think it's very valuable. I wonder if detekt has any rule for that?

Mauin:
As far as I'm aware there is no such rule in detekt currently. It should be possible to implement one, though.

marianosimone:
This should be done, now that `VarCouldBeVal` is merged

PR #978, opened 2018-07-01
New rule: VarCouldBeVal

marianosimone:
This is a first implementation of a rule, as requested in #657 

Note that, for now, it only looks at locally-scoped variables inside functions. I want to make it more useful than this (think, private top-level variables, properties in classes, etc.), but given that this is my first rule, I wanted to create a quick PR to get some feedback and see if I'm going down the right path.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
@@ -1,0 +1,111 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.DetektVisitor
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.psi.KtBinaryExpression
+import org.jetbrains.kotlin.psi.KtNamedDeclaration
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.psi.KtUnaryExpression
+
+private val unaryAssignmentOperators = setOf(KtTokens.MINUSMINUS, KtTokens.PLUSPLUS)
+
+/**
+ * Reports var declarations (locally-scoped variables) that could be val, as they are not re-assigned.
+ * Val declarations are assign-once (read-only), which makes understanding the current state easier.
+ *
+ * <noncompliant>
+ * fun example() {
+ *     var i = 1 // violation: this variable is never re-assigned
+ *     val j = i + 1
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * fun example() {
+ *     val i = 1
+ *     val j = i + 1
+ * }
+ * </compliant>
+ *
+ * @author marianosimone
+ */
+class VarCouldBeVal(config: Config = Config.empty) : Rule(config) {
+
+	override val issue: Issue = Issue("VarCouldBeVal",
+			Severity.Maintainability,
+			"Var declaration could be val.",
+			Debt.FIVE_MINS,
+			aliases = setOf("CanBeVal"))
+
+	override fun visitNamedFunction(function: KtNamedFunction) {
+		val assignmentVisitor = AssignmentVisitor()
+		function.accept(assignmentVisitor)
+
+		assignmentVisitor.getNonReAssignedDeclarations().forEach {
+			report(CodeSmell(issue, Entity.from(it), "Variable '${it.nameAsSafeName.identifier}' could be val."))
+		}
+		super.visitNamedFunction(function)
+	}
+
+	private class AssignmentVisitor : DetektVisitor() {
+
+		private val declarations = mutableSetOf<KtNamedDeclaration>()
+		// an easy way to find declarations when walking up the contexts of an assignment
+		private val contextsByDeclarationName = mutableMapOf<String, MutableSet<PsiElement>>()
+		private val assignments = mutableMapOf<String, MutableSet<PsiElement>>()
+
+		fun getNonReAssignedDeclarations(): List<KtNamedDeclaration> {
+			return declarations.filter { declaration ->
+				assignments[declaration.nameAsSafeName.identifier]
+						?.let { declaration.parent !in it }
+						?: true
+			}
+		}
+
+		override fun visitProperty(property: KtProperty) {
+			if (property.isVar) {
+				declarations.add(property)
+				val contextsForName = contextsByDeclarationName.getOrPut(property.nameAsSafeName.identifier) { mutableSetOf() }
+				contextsForName.add(property.parent)
+			}
+			super.visitProperty(property)
+		}
+
+		override fun visitUnaryExpression(expression: KtUnaryExpression) {
+			if (expression.operationToken in unaryAssignmentOperators) {
+				expression.baseExpression?.run {
+					visitAssignment(text, expression.parent)
+				}
+			}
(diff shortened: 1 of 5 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
```diff
@@ -0,0 +1,77 @@
...
+ * </compliant>
+ *
+ * @author marianosimone
+ */
+class VarCouldBeVal(config: Config = Config.empty) : Rule(config) {
+
+	override val issue: Issue = Issue("VarCouldBeVal",
+			Severity.Maintainability,
+			"Var declaration could be val.",
+			Debt.FIVE_MINS, // TODO
```
marianosimone: not exactly sure what the right value is here

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
```diff
@@ -0,0 +1,77 @@
...
+			aliases = setOf("")) // TODO
```
marianosimone: nor here

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
arturbosch: 5 minutes is okay for a var to val change, commit, pr, merge :).
Basically 5 minutes is our lowest value we rate rules as we count all the vcs steps into it.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
arturbosch: Aliases are an optional feature for rules.
For example if intellij or some other static analysis tools has a similar rule and offers a different string to suppress such cases as we do (`RuleName`) we define aliases so the user just need to write `@Suppress("ONCE")`

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
```diff
@@ -0,0 +1,77 @@
...
+
+		override fun visitProperty(property: KtProperty) {
+			if (property.isVar) {
+				declarations.add(property)
+			}
+			super.visitProperty(property)
+		}
+
+		override fun visitBinaryExpression(expression: KtBinaryExpression) {
+			if (expression.operationToken.toString() == "EQ") {
```
arturbosch: There is more than just the EQ token.
I think you need to use `expression.operationToken in KtTokens.ALL_ASSIGNMENTS` here

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
marianosimone: Cool, thanks for the sanity check. Removed the TODO

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
marianosimone: Great! Just found `CanBeVal`, so putting that instead of the TODO

marianosimone:
Assignments in if statements are not allowed:

```
var a = 1
if (a = 2) {
    // do something
}
```

Is not valid, as `Assignments are not expressions, and only expressions are allowed in this context`

When I added some more tests (specially around shadowed variables, thanks @schalkms !!!!), I discovered that my approach was WAY too naive. Now it's a bit more complex, taking the context into account, and trying to figure out which variable is being re-assigned by walking up the parents until a matching declaration is found

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
marianosimone: Great call! Changed that and added a test case

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
```diff
@@ -0,0 +1,106 @@
...
+					var shadowed = 2
+					shadowed = 3
+				}
+			}
+			"""
+			val lint = subject.lint(code)
+
+			assertThat(lint).hasSize(1)
+			with(lint[0]) {
+				assertThat(entity.name).isEqualTo("shadowed")
```
marianosimone: I'm experimenting with this type of assertions... I saw other tests just asserting that there's 1 smell being reported, but (specially when doing TDD), something like this gives me way more confidence

marianosimone:
The `appveyor` check fail seems unrelated to my change, and having to do with a timeout :man_shrugging:

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
schalkms: I think this should be the default testing strategy for more complex rules. What do you think @arturbosch ?

arturbosch:
Yep I restarted the build and its green now, will review it next :)

arturbosch:
Ok, I just ran it on detekt itself and here are some false positives:

io/gitlab/arturbosch/detekt/generator/out/Yaml.kt
```kotlin
	fun getIndent(): String {
		var spaces = ""
		indent times {
			spaces += "  "
		}
		return spaces
	}

infix fun Int.times(function: () -> Unit) {
	var i = this
	while (i > 0) {
		function()
		i--
	}
}
```
io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
```kotlin
private fun String.trimStartingLineBreaks(): String {
	var i = 0
	while (i < this.length && (this[i] == '\n' || this[i] == '\r')) {
		i++
	}
	return this.substring(i)
}
```
detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```kotlin
	private fun countFunctionReturns(function: KtNamedFunction): Int {
		var returnsNumber = 0
		function.accept(object : DetektVisitor() {
			override fun visitKtElement(element: KtElement) {
				if (element is KtReturnExpression) {
					returnsNumber++
				}
				element.children.filter { it !is KtNamedFunction }.forEach { it.accept(this) }
			}
		})
		return returnsNumber
	}
```

Please make sure that the rule does not report any of these. 
We do also have the `scripts/get_analysis_projects.groovy` script which downloads a bunch of test projects you could run the new rule on :).

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
arturbosch: I mostly agree, when just testing code snippets as we often do like:
```kotlin
val code = """ .... """
```
its ok to use the short isEmpty or hasSize(1).
But yeah in this case and similar just using one statement is not enought to be 100% right.
We have `io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt` and more of these would greatly improve our test quality

marianosimone:
Added infix and postfix operators `++` and `--` detection. Thanks for the pointers on running my rule against both `detekt` and all the other projects.

I'm happy to tell you that there are no positives in `detekt`, and that it actually found a couple of true positives in some of the others (and no false positives!) :tada:

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
```diff
@@ -0,0 +1,125 @@
...
+					var shadowed = 2
+					shadowed = 3
+				}
+			}
+			"""
+			val lint = subject.lint(code)
+
+			assertThat(lint).hasSize(1)
+			with(lint[0].entity) {
+				assertThat(ktElement?.text).isEqualTo("var shadowed = 1")
```
marianosimone: I ended up changing this slightly, as the check was redundant. PTAL (which might give us ideas about what to add to `FindingsAssertions`)

PR #981, opened 2018-07-04
Activate rules in failfast.yml part2

schalkms:
This PR contains some fixes for the `UnusedPrivateMember` rule.


schalkms:
The following rules should be activated after a new version is released

- UnusedPrivateMember
- LabeledExpression
- NestedClassesVisibility
- CommentOverPrivateFunction
- StringLiteralDuplication

PR #986, opened 2018-07-08
Refactor UnusedImports rule

marianosimone:
1. Make tests a bit more robust (by actually verifying what the
smells are)
2. Implement the rule with an internal visitor, so there's no state
in the rule itself (which needed to be reset for each file, preventing
it for being a candidate for parallel excecution)
3. Only look at KtReferenceExpressions that are not imports, package
directives nor have children. For example:

`foo(a, b)`

actually gets visited for `foo(a(b), c)`, `foo`, `a(b)`, `a`, `b`, and
`c`, but we only care about `foo`, `a`, `b`, and `c`

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
@@ -3,16 +3,18 @@ package io.gitlab.arturbosch.detekt.rules.style
 import io.gitlab.arturbosch.detekt.api.CodeSmell
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.DetektVisitor
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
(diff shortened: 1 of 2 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
```diff
@@ -29,68 +31,71 @@ class UnusedImports(config: Config) : Rule(config) {
...
+		private val operatorSet = setOf("unaryPlus", "unaryMinus", "not", "inc", "dec", "plus", "minus", "times", "div",
+				"mod", "rangeTo", "contains", "get", "set", "invoke", "plusAssign", "minusAssign", "timesAssign", "divAssign",
+				"modAssign", "equals", "compareTo", "iterator", "getValue", "setValue")
 
-	private var imports = mutableListOf<Pair<String, KtImportDirective>>()
-	private val kotlinDocReferencesRegExp = Regex("\\[([^]]+)](?!\\[)")
+		private val kotlinDocReferencesRegExp = Regex("\\[([^]]+)](?!\\[)")
+	}
 
 	override fun visitFile(file: PsiFile?) {
```
arturbosch: I'm not sure anymore why this is a `PsiFile?` and not just `KtFile`. the parameters in visit methods cannot be null to my knownledge. Can you try to change this and remove the file?.run {...} then

marianosimone: Changed!

PR #989, opened 2018-07-09
UnusedImports: Add detection of imports in same package

### LINKED ISSUE #985: UnusedImports: False negative when importing class in same package
For the following files

```
package test

import test.A

val a = A()
```

```
package test

class A
```

## Expected Behavior
The `import test.A` should be reported, as both files share the same package

## Current Behavior
Nothing is reported

## Your Environment
* Version used: 1.0.0-RC7-3

marianosimone:
Imports in the same package are unnecessary, and can be safely removed
Fixes #985


vanniktech:
I thought about this today at work and there's one case where you need an import: qualifying a nested class when being referenced from the outer one

vanniktech:
```kotlin
package com.foo.bar

import com.foo.bar.ChooseTimeAdapter.ChooseTimeViewHolder

class ChooseTimeAdapter : RecyclerView.Adapter<ChooseTimeViewHolder>() {
  class ChooseTimeViewHolder(view: View) : RecyclerView.ViewHolder(view) {
  }
}
```

marianosimone:
@vanniktech @arturbosch : I just added a new test to make sure this is not an issue: #997

PR #991, opened 2018-07-12
Fix to work on maven multi-module project 

duck8823:
I tried using [detekt-maven-plugin](https://github.com/Ozsie/detekt-maven-plugin) is introduced by README in this repository.
But, unfortunately, it didn't work in my multi-module project.

I made a sample [here](https://github.com/duck8823/detekt-maven-sample).
The detekt task success first module, but second module occurred error `Can only specify option -i once.`.

I found that jcommander instance is defined as global variable. 
Because the jcommander instance is used multiple times in multi-module project, argument parse error occur!

In this PR, I suggest that jcommander instance define as local variable.


arturbosch:
Hey good catch! This is a leftover from a refactoring, thanks for contributin!

PR #995, opened 2018-07-13
Fix false positive lambda in constructor call - Fixes #990

### LINKED ISSUE #990: False positive UnnecessaryParentheses on constructors with lambdas
Detekt reports UnnecessaryParentheses  for the secondary constructor in the following code for class `Clazz`:

```kotlin

typealias KBiFunc<X, Y, Z> = (X, Y) -> Z

interface Func<in X : Any, out Y : Any> {
    fun apply(input: X): Y
}

interface BiFunc<in X : Any, in Y : Any, out Z : Any> {
    fun apply(
        first: X,
        second: Y
    ): Z
}

class Clazz<in X : Any, out Y : Any>(
    private val func: KBiFunc<X, Boolean, Y>
) : Func<X, Y> {

    constructor(
        func: BiFunc<X, Boolean, Y>
    ) : this(
        { first, second -> func.apply(first, second) }
    )

    override fun apply(input: X): Y = func(input, true)
}

```

I am using version `1.0.0.RC7-3`

arturbosch:
Closes #990

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParenthesesSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParenthesesSpec.kt
@@ -104,5 +104,36 @@ class UnnecessaryParenthesesSpec : SubjectSpek<UnnecessaryParentheses>({
 				"""
 			assertThat(subject.lint(code)).hasSize(0)
 		}
+
+		it("should not report lambdas within super constructor calls") {
+			val code = """
+				typealias KBiFunc<X, Y, Z> = (X, Y) -> Z
+
+				interface Func<in X : Any, out Y : Any> {
+					fun apply(input: X): Y
+				}
+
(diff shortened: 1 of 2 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParenthesesSpec.kt
```diff
@@ -104,5 +104,36 @@ class UnnecessaryParenthesesSpec : SubjectSpek<UnnecessaryParentheses>({
 				"""
 			assertThat(subject.lint(code)).hasSize(0)
 		}
+
+		it("should not report lambdas within super constructor calls") {
+			val code = """
+				typealias KBiFunc<X, Y, Z> = (X, Y) -> Z
```
schalkms: I think this whole code section should be simplified and moved to a separate test case kt file.

arturbosch: I was too lazy :(

PR #1004, opened 2018-07-17
Activate rules in failfast.yml part3

schalkms:
I'll test the rules with Artur's script and the 3rd party repositories soon.


arturbosch:
Looks good to me. Why and which rules?
We just want more rules to be integrated in the detekt CI?

schalkms:
Right. I just wanted another testing step with the 3rd party repositories. This and the previous PRs revealed some minor bugs, which were fixed as a consequence.
I don't know if you have already tested detekt's rules with the 3rd party code.

arturbosch:
Ah ok! Nope, not where `active = false`. Only for newer rules which got contributed, some of them I manually checked.

PR #1005, opened 2018-07-17
Add a missing space

scottkennedy:
Another minor issue where there's no space between two words.


vanniktech:
It's time detekt also checks for the error message. Like every other tool does it.

vanniktech:
It's time detekt also checks for the error message. Like every other tool does it.

PR #1017, opened 2018-07-25
Mr/gradle classes

marschwar:
As @Mauin does not currently have the time to finish the gradle plugin rework I looked into it. As discussed in his original pull request, using the `AbstractCodeQualityPlugin` as a base is dangerous and likely cause problems with any gradle release.

In constrast to @Mauin s solution I removed the `SourceTask` inheritance. While we technically do work with  source files, the way we use refer to them is via base directory. The SourceTask uses SourceSets and they allow the notion of includes/excludes filtering. In order to leverage that we we would need a different level of integration with the detekt core in which we pass the list of source files to the detekt process, since there is a limit doing that over cli.

There is one caveat to this PR which I am not completely sure on how to solve. For the lack a better name I gave it the version `1.0.0-GRADLE`. In order to build, it needs the new plugin version which needs the new cli. What I did locally was:

```
./gradlew clean publishToMavenLocal -x dokka -x detekt
./gradlew build -PwarningsAsErrors=true
```

As I said, I don't know how to resolve this. Releasing the cli changes first and updating the plugin second would probably be rather tedious.

I tried to follow the contribution guidelines but I am sure that I missed something so please feel free to comment.


arturbosch:
First of all thanks for the work! As this is a huge PR I will need some time to review this. I hope I get to it this weekend.

3flex:
Because I was the one who flagged usage of `AbstractCodeQualityPlugin` it would be remiss of me not to point out that usage of the `org.gradle.api.reporting.internal.*` classes should be avoided.

In this case it should be simple enough to just remove the `DetektReportsImpl` class, according to docs only the `ReportContainer<T extends Report>` interface has to be implemented by the `DetektReports` interface (which has been done), but no implementing class is required. See example at https://docs.gradle.org/current/javadoc/org/gradle/api/reporting/Reporting.html

The report-related overrides in Detekt task can be removed too.

I haven't tested this so could be wrong :/

marschwar:
Thank for for your comment @3flex. You are absolutely right, we should be avoiding the internal api. I don't think it is possible to simply remove the DetektReportsImpl and the overrides. At least all the internal Plugins such as Checkstyle , PMD, CodeNarc all implement the interface. I will look into that over the weekend.

marschwar:
I removed all references to the internal classes. The `ReportingContainer` and `Report` interfaces did seam like a good idea at first but do not provide any benefit as far as I can tell.

arturbosch:
The `DetektExtension.config` property is now a `File?`. I would say it is a crucial feature that this parameter can also be a `FileCollection` or just a `String`. It allows to use the default-config.yml but at the same time override some of the default properties in a second config file.
In the old gradle-plugin, I used `Any?` and tested it for own on these kinds `File, Path, String, FileCollection`, is this still possible?

marschwar:
If the `check` task exists in the current project, `detekt` is added as a dependency as I think that is the expected behavior. But of course it is possible to call `detekt` by itself.

As for the config I am not sure I understand what you are trying to do. At the end of the day we are calling the DetektCLI. Does that support multiple config files? I did not get the impression from the CliArgs.

```java
@Parameter(names = ["--config", "-c"],
		description = "Path to the config file (path/to/config.yml).")
var config: String? = null
```

3flex:
I believe multiple config paths can be set - the description in CliArgs is wrong. So FileCollection is the correct type to use for the config files.

arturbosch:
I had trouble merging/rebasing this and came up with #1079. Sorry for destroying authorship and commit history. I changed the author of the merge commit to @marschwar .

PR #1024, opened 2018-07-28
Add serialVersionUID to UnusedPrivateMember.allowedNames

Mygod:
https://stackoverflow.com/q/34241816/2245107

PR #1025, opened 2018-07-30
Fixed #1024 - InstanceOfCheckForException being too general

schalkms:


schalkms:
There is a weird error with the Travis CI going on. This has happened more often recently.

PR #1026, opened 2018-07-30
Refactor ObjectPropertyNaming

scottkennedy:
This adds a privatePropertyPattern and much more closely matches the
code in TopLevelPropertyNaming.  They may be able to share a base class (or be merged into one) in the future.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ObjectPropertyNaming.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ObjectPropertyNaming.kt
@@ -9,13 +9,14 @@ import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.rules.hasConstModifier
 import org.jetbrains.kotlin.psi.KtProperty
-import org.jetbrains.kotlin.psi.KtVariableDeclaration
+import org.jetbrains.kotlin.psi.psiUtil.isPrivate
 
 /**
  * Reports when property names inside objects which do not follow the specified naming convention are used.
  *
- * @configuration propertyPattern - naming pattern (default: '[A-Za-z][_A-Za-z0-9]*')
  * @configuration constantPattern - naming pattern (default: '[A-Za-z][_A-Za-z0-9]*')
+ * @configuration propertyPattern - naming pattern (default: '[A-Za-z][_A-Za-z0-9]*')
+ * @configuration privatePropertyPattern - naming pattern (default: '(_)?[A-Za-z][A-Za-z0-9]*')
  * @active since v1.0.0
  * @author Marvin Ramin
  * @author schalkms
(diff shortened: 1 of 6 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ObjectPropertyNaming.kt
```diff
@@ -27,31 +28,48 @@ class ObjectPropertyNaming(config: Config = Config.empty) : Rule(config) {
...
+	private fun handleProperty(property: KtProperty) {
+		if (property.isPrivate()) {
+			if (!property.identifierName().matches(privatePropertyPattern)) {
+				report(CodeSmell(
+						issue,
+						Entity.from(property),
+						message = "Private object property names should match the pattern: $privatePropertyPattern"))
+			}
+		} else {
+			if (!property.identifierName().matches(propertyPattern)) {
```
schalkms: else if (<<condition>>) {}

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ObjectPropertyNamingSpec.kt
```diff
@@ -87,7 +88,6 @@ class ObjectPropertyNamingSpec : SubjectSpek<ObjectPropertyNaming>({
 			val code = compileContentForTest("""
 				object O {
 					val _nAme = "Artur"
-					private val _name = "Artur"
 				}
 			""")
 			assertThat(subject.lint(code)).hasSize(2)
```
schalkms: Why does this test snippet report `2` violations?
It should only report 1 for the variable _nAme.
I suppose there's is still an error in the rule implementation.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ObjectPropertyNaming.kt
scottkennedy: I literally copied this code from `TopLevelPropertyNaming`.  I'm happy to make this change, but I'll probably make it there too to keep it consistent.

schalkms: No worries..

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ObjectPropertyNamingSpec.kt
```diff
@@ -87,10 +88,9 @@ class ObjectPropertyNamingSpec : SubjectSpek<ObjectPropertyNaming>({
 			val code = compileContentForTest("""
 				object O {
 					val _nAme = "Artur"
-					private val _name = "Artur"
 				}
 			""")
-			assertThat(subject.lint(code)).hasSize(2)
+			assertThat(subject.lint(code)).hasSize(1)
```
schalkms: @scottkennedy what did you change to only report `1` violation?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ObjectPropertyNamingSpec.kt
scottkennedy: Hmm.  I changed the test to 1, and it still passed.  I'm not convinced the test is actually running 😕

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ObjectPropertyNamingSpec.kt
scottkennedy: Because there's only one property in the object now.  I believe 1 is correct.

However, it wasn't failing the test when it was 2, and it's not failing now that it's 1, so I don't think the test is working.

schalkms: That's indeed very weird.
I checked out your branch on my machine. Running the tests in IntelliJ does work. However it doesn't when using the CLI.

Issue #1030, opened 2018-08-01
TrailingWhitespace reports incorrect line number


elihart:
## Expected Behavior
Leaving a trailing space triggers the `TrailingWhitespace` error, but it reports the wrong line and column. The correct line is only mentioned in the error message.

## Current Behavior
```
<checkstyle version="4.3">
<file name="...">
	<error line="6" column="12" severity="warning" message="Line 71 ends with a whitespace." source="detekt.TrailingWhitespace" />
</file>
</checkstyle>
```

## Steps to Reproduce (for bugs)
Add a space at the end of a comment line and run detekt

PR #1033, opened 2018-08-02
Fix incorrect line number in TrailingWhitespace

schalkms:
reference #1030

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/Junk.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/Junk.kt
@@ -1,0 +1,13 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import org.jetbrains.kotlin.com.intellij.openapi.util.TextRange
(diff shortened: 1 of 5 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/Junk.kt
```diff
@@ -0,0 +1,13 @@
+package io.gitlab.arturbosch.detekt.rules.style
```
marianosimone: is `Junk.kt` the best name for this file?

schalkms: Good point. I think we can come up with a better name :)
Kt files containing util functions in the rule package are named like this. So I sticked with this.

arturbosch: My intention with Junk was to put stuff there which do not fit somewhere else and move them later to a KtFile with similar stuff or just delete them ^^

Issue #1010, opened 2018-07-22
Referencing the outer class with a label from an inner class should be allowed


vanniktech:
In Android sometimes you have to use inner classes and want to reference the outer class. One common example are Binders and Services. 

Boiled down this looks like this:

```kotlin
class Foo {
  inner class Inner {
    fun abc() {
      val foo = this@Foo
    }
  }
}
```

There's no other way to get the instance of `Foo` from `Inner` and using an expression label seems fine, hence Detekt should not flag such code.

arturbosch:
Which rule does report this?

vanniktech:
https://github.com/arturbosch/detekt/blob/3eb867294f11a2eefd936852158d96c4263c02b4/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LabeledExpression.kt

arturbosch:
Ah ok, maybe we also need an exception part in kdoc:
```
<exception>
    [code example which is compliant by exception]
</exception>

**Comment by arturbosch** [comment]
Additional cases which need to be handled #1036

**Comment by vanniktech** [comment]
@arturbosch you mean this one, right?

```kotlin
class Foo {
  fun Bar.extensionMethod() {
    add(this@Foo)
  }
}
```

schalkms:
Closed with 91d4504

@vanniktech 
your snippet is related to #1226 so I grouped them together.

PR #1036, opened 2018-08-02
Update LabeledExpression to support outer class refs 

schalkms:
reference #1010


vanniktech:
@schalkms I've found one more case.

```kotlin
class Foo {
  fun Bar.extensionMethod() {
    add(this@Foo)
  }
}
```

vanniktech:
The same can also be applied when using `apply()` since you're in a different scope.

PR #1046, opened 2018-08-10
Add ParameterNaming rules for Constructors and Functions parameters

mickele:
I asked in the Slack channel if this was supported and @schalkms said that no, so I added 2 new rules:
- ConstructorParameterNaming
- FunctionParameterNaming

Do the default configuration look OK? Did I forget anything? :)


mickele:
@schalkms the build failed because (https://github.com/arturbosch/detekt/blob/master/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/TestPattern.kt#L19) is using `_` in the beginning of the parameter of the data class.

What do you think? Should we add this to the rule so `_` is allowed at the beginning or can I just invert the `_patterns` and `patterns` properties? Apparently that's the only place this naming pattern is used.

schalkms:
@mickele I know. It makes sense to rename this variable.
We should adhere to the official Kotlin naming guideline [here](https://kotlinlang.org/docs/reference/coding-conventions.html#property-names).
I think this rule can be enabled, since it essentially mirrors `FunctionNaming`. @arturbosch

mickele:
No problem, thanks :)

PR #1059, opened 2018-08-20
Add ForbiddenVoid rule

### LINKED ISSUE #1058: Forbid usage of Void
Detekt should warn about using [`Void`](https://docs.oracle.com/javase/7/docs/api/java/lang/Void.html) since Kotlin has [`Unit`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-unit/index.html) instead.

Although, maybe `java.lang.Void::class` should be allowed.

egor-n:
Closes #1058.

```kotlin
// Bad
runnable: () -> Void
var aVoid: Void? = null

// Good
runnable: () -> Unit
Void::class
```

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -397,6 +397,8 @@ style:
   ForbiddenImport:
     active: false
     imports: ''
+  ForbiddenVoid:
+    active: false
   FunctionOnlyReturningConstant:
     active: false
     ignoreOverridableFunction: true
--- a/reports/failfast.yml
+++ b/reports/failfast.yml
@@ -108,8 +108,6 @@ style:
   VariableNaming:
     active: true
     variablePattern: '(_)?[a-z][a-zA-Z0-9]*'
-  SpacingBetweenPackageAndImports:
-    active: true
   UnusedPrivateMember:
     active: true
     allowedNames: "(_|ignored|expected)"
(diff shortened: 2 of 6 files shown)


reports/failfast.yml
```diff
@@ -89,10 +89,6 @@ potential-bugs:
   UselessPostfixExpression:
     active: true
 
-potential-bugs:
```
egor-n: I have removed this entry as it duplicates the one on line 86.

reports/failfast.yml
```diff
@@ -112,8 +108,6 @@ style:
   VariableNaming:
     active: true
     variablePattern: '(_)?[a-z][a-zA-Z0-9]*'
-  SpacingBetweenPackageAndImports:
```
egor-n: I have removed this entry as it duplicates the one on line 108.

detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -397,6 +397,8 @@ style:
   ForbiddenImport:
     active: false
     imports: ''
+  ForbiddenVoid:
+    active: false
```
egor-n: `./gradlew build` always sets this to `false`. How do I enable it by default?

reports/failfast.yml
schalkms: Thanks! A merge seems to cause this duplicates.

detekt-cli/src/main/resources/default-detekt-config.yml
schalkms: This will be enabled after detekt v1. Currently, new contributed rules are set to `false` to have more time for testing.

Issue #1055, opened 2018-08-19
False positive with MagicNumber on default value for parameters


vanniktech:
Correct me if I'm wrong but the following method signature gets flagged with a `MagicNumber` with Detekt 1.0.0.RC8:

```kotlin
fun Context.vibrate(duration: Long = 300L)
```

I thought that default parameters won't be treated as such code smells. As for the configuration I'm using:

```yml
failFast: true

potential-bugs:
  UnsafeCast:
    active: false # We know what we're doing.
  LateinitUsage:
    active: false # We know what we're doing.

naming:
  FunctionMinLength:
    active: false
  FunctionMaxLength:
    active: false

style:
  TopLevelPropertyNaming:
    active: false
  MaxLineLength:
    active: false
  UnnecessaryParentheses:
    active: false # False positives.

complexity:
  LongParameterList:
    threshold: 8

comments:
  UndocumentedPublicClass:
    active: false
  UndocumentedPublicFunction:
    active: false
```

schalkms:
Detekt hasn't considered this case yet. `MagicNumber` allows constant declarations in constructors, but as it looks like not in functions.
We could need some help here. Would you mind taking a look at this @vanniktech ?

vanniktech:
Eventually yeah but don't get your hopes up.

PR #1061, opened 2018-08-21
Improvement #1055 MagicNumber

schalkms:
reference #1055 
@vanniktech


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumberSpec.kt
```diff
@@ -629,22 +629,22 @@ class MagicNumberSpec : Spek({
 
 	given("functions with and without braces which return values") {
 
-		it("does not report functions that always returns a constant value") {
+		it("does not report functions that always return a constant value") {
```
vanniktech: It was correct before

PR #1062, opened 2018-08-21
Make "ComplexMethod" rule also ignore "return when" if configured

### LINKED ISSUE #1054: ComplexMethod should treat "return when" as "single when" expression when ignoring
## Expected Behavior

When detekt is configured like this:

```
  ComplexMethod:
    active: true
    threshold: 10
    ignoreSingleWhenExpression: true
```
This kind of `when` should be treated as "single when expression" and ignored
```kotlin
fun someFunction(s: String): Boolean {
  return when (s) {
     "bear" -> true
     "log" -> false
     // ... 15 more branches
  }
}
```

## Current Behavior
Currently the code above will report a `ComplexMethod` rule failure, because it ignores only "non-expression "when's", like

```kotlin
fun someFunction(s: String): Unit {
  when (s) {
     "bear" -> doSomething()
     "log" -> doSomethingElse()
     // ... 15 more branches
  }
}
```

If you are OK with the proposed change I could create a PR.

dimsuz:
If configured with "ignoreSingleWhenExpression", ComplexMethod should
not only skip "when", but also "return when"

Fixes #1054

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
@@ -12,6 +12,7 @@ import io.gitlab.arturbosch.detekt.api.internal.McCabeVisitor
 import org.jetbrains.kotlin.psi.KtBlockExpression
 import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.KtReturnExpression
 import org.jetbrains.kotlin.psi.KtWhenExpression
 
 /**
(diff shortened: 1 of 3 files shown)


schalkms:
Can you please test this with single-expression functions. I don't think that the rule reports this atm.
```Kotlin
fun f(): Int {
 	return if { ... } else { ... }
}
```

dimsuz:
Do you mean that this rule should treat `if-else-if-else-if-else` as a single when? Or the opposite, always report them, irrespective to `ignoreSingleWhenExpression` flag?

schalkms:
@dimsuz Sry I have updated my comment. I meant that the rule should also report a comment if the single expression syntax is used.

schalkms:
Okay. I think we were talking at cross purposes. Adding the following snippet as another test case should report a violation. Note that I took your test code and just used the single expression syntax. Atm this case is not handled by the `ComplexMethod` rule.
```Kotlin
fun complexMethodWithSingleWhen4(i: Int) = when (i) {
	1 -> "one"
	2 -> "two"
	3 -> "three"
	else -> ""
}
```

dimsuz:
Oh, right! I got mixed up between "single expression inside the body" and "single expression function syntax" .

Ok, so I should test that your snippet above

* is causing a violation when `ignoreSingleWhenExpression` is `false` (the default)
* is not causing a violation when `ignoreSingleWhenExpression` is `true`

In other words `fun test() = when {}` should behave exactly as `fun test() { return when {} }`

I hope I got it right this time :slightly_smiling_face:.

I will also add test(s) for these cases.

dimsuz:
I have updated the code and the tests to handle a "function single expression" syntax.

In `ComplexMethodSpec` there is a spec "reports all complex methods" and ignoring is turned off it indeed reports all "single when" cases as a failures.
And when ignoring is turned on only one case without single whens is reported

dimsuz:
It turned out that your snippet above is already handled and is reported as `ComplexMethod [5/4]` when ignoring single when is turned off (though I tested it at home, on current release version of detekt, but it shouldn't have changes I guess). I didn't add a specific test for this case though.

kohnish:
Can anybody give me a guidance on what would be the best way to handle when with enum larger than 15?  

I'm learning kotlin and trying to stick with default detekt setting for now.
I am assuming detekt has a sensible default and ignoreSingleWhenExpression is disabled by default.

Issue #1064, opened 2018-08-27
Unknown exception thrown when running detekt


pavlospt:
## Expected Behavior
Expected detekt to finish without any exceptions thrown

## Current Behavior
Currently when running detekt, after enabling a rule, not sure which one, we get the following exception stacktrace:
```java
Analyzing '/path/to/the/file.kt' led to an exception.
Running detekt 'unknown' on Java '1.8.0_131-b11' on OS 'Mac OS X'.
Please create an issue and report this exception.
java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1592)
java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1582)
java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
```
The exception above is the result of enabling one of the latest rules we enabled, will probably do a `git bisect` to find which one, but I am reporting it in case you already know what is wrong so that I could resolve it sooner :D 

Before enabling that rule it was running without exxceptions being produced.

## Steps to Reproduce (for bugs)
Not sure on that one, I am still trying to figure out the exact repro flow.

## Context
We are incrementally enabling more and more detekt rules and one of those caused the above exception.

## Your Environment
* Version used: 1.0.0.RC8
* Operating System and version: macOS Sierra (10.12)
* Link to your project: It's a private project.

pavlospt:
Update: I tried to manually disable every rule that we have introduced in the branch that it stopped working and had no luck. Also tried reverting to RC07-2 and it was not resolved either. LMK if I need to provide any more info!

arturbosch:
Do you have to stau at a beta version of jdk8 ? -> '1.8.0_131-b11'.
In `detekt-core` we leverage the CompletableFuture Api from java which should be fine.
Maybe an error in beta version of jdku8. Can you upgrade and test with a recent jdk8 please?

pavlospt:
[...]

PR #1080, opened 2018-09-02
Lazy regex evaluation on Rules

pavlospt:
This PR resolves the eager Regex evaluation discussed in #1064 

1) Introduces `LazyRegex` class that evaluates a `Regex` needed inside a rule, only when it is requested and only once.
2) Makes use of `LazyRegex` in `LateinitUsage` rule.
3) Adds 2 extra test steps asserting the discussed behaviour.

@arturbosch I would be more than happy to migrate the rest of the `Regex` usages inside `Rule`s (if any) once you think this is a proper solution!


pavlospt:
@arturbosch updated with your feedback :)

arturbosch:
Thanks, feel free to add yourself to the readme and create additional PR with the more regex changes :)

pavlospt:
@arturbosch Added my name in README.md! You can merge this one and I will follow up with a new one for the rest of the Rules :)

PR #1086, opened 2018-09-05
Move Issue template & Pull Request template into .github directory.

vanniktech:
Might also be possible to move the `CONTRIBUTING.md` although I'm not sure.

Motivation: Have those files not in the root directory as they clutter.


vanniktech:
So I'll just move it into `.github` directory as well?

arturbosch:
Yes, if github will find it and suggest to new contributors why not :)

PR #1087, opened 2018-09-05
MaxLineLength: Ignore comments that end with a long url.

### LINKED ISSUE #1070: MaxLineLength should ignore link only comments by default
Given I have a very long link as a comment it should not trigger a MaxLineLength violation since there's not really much you can do if you have a long url.

vanniktech:
Fixes #1070


schalkms:
@vanniktech I would put it in the respective rule or in a separate file.

PR #1088, opened 2018-09-05
Auto deploy SNAPSHOT versions.

### LINKED ISSUE #512: Auto-deployed SNAPSHOTs
Maybe I'm missing something but it would be really beneficial to have auto-deployed snapshots to verify certain newly added behaviors without the need to create another RC.

vanniktech:
Fixes #512. Hopefully. I've got some questions.


.buildscript/deploy_snapshot.sh
```diff
@@ -0,0 +1,27 @@
...
+  echo "Skipping snapshot deployment: wrong repository. Expected '$SLUG' but was '$TRAVIS_REPO_SLUG'."
+elif [ "$TRAVIS_JDK_VERSION" != "$JDK" ]; then
+  echo "Skipping snapshot deployment: wrong JDK. Expected '$JDK' but was '$TRAVIS_JDK_VERSION'."
+elif [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+  echo "Skipping snapshot deployment: was pull request."
+elif [ "$TRAVIS_BRANCH" != "$BRANCH" ]; then
+  echo "Skipping snapshot deployment: wrong branch. Expected '$BRANCH' but was '$TRAVIS_BRANCH'."
+else
+  echo "Deploying snapshot..."
+  ./gradlew bintrayUpload
```
vanniktech: Is that the right task? You'll need to set the environment variable in Travis for the Username + password.

You can encrypt them though - https://docs.travis-ci.com/user/encrypting-files/#using-gpg

gradle.properties
```diff
@@ -1,4 +1,4 @@
-detektVersion=1.0.0.RC8
+detektVersion=1.0.0.RC9-SNAPSHOT
```
vanniktech: I hope that's the right one I changed and RC9-SNAPSHOT sounds good

arturbosch: yes its good :)

.buildscript/deploy_snapshot.sh
arturbosch: Yes, its the right one. Hm but bintray will tell me that I already have that version when uploading the same version twice

arturbosch: I think you need to add `override = true` to the bintray closure.

vanniktech: We should only do that for snapshots though, right?

arturbosch: Yeah, lets check if the vesion endswith SNAPSHOT then?

vanniktech: Yup. Done that now.

vanniktech:
@arturbosch ready for another review.

PR #1091, opened 2018-09-06
UnnecessaryAbstractClass: Support exclusion via Annotations.

### LINKED ISSUE #377: UnnecessaryAbstractClass: Relax when the class is annotated with Module
In dagger it makes sense to have an abstract class with a few abstract methods since it'll tell the Dagger Compiler how certain types can be bound.

An example is the following class:

```kotlin
@Module abstract class SomeModule {
  @Binds abstract fun provideSomeThing(mySomeThing: MySomeThing): SomeThing
}
```

We probably want to ignore those cases when `dagger.Module` is applied.

What do you think?

vanniktech:
Fixes #377

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -473,6 +473,7 @@ style:
     active: false
   UnnecessaryAbstractClass:
     active: false
+    excludeAnnotatedProperties: "dagger.Module"
   UnnecessaryInheritance:
     active: false
   UnnecessaryParentheses:
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -473,7 +473,7 @@ style:
     active: false
   UnnecessaryAbstractClass:
     active: false
-    excludeAnnotatedProperties: "dagger.Module"
+    excludeAnnotatedClasses: "dagger.Module"
   UnnecessaryInheritance:
     active: false
   UnnecessaryParentheses:
--- a/detekt-rules/build.gradle.kts
+++ b/detekt-rules/build.gradle.kts
@@ -14,6 +14,7 @@ dependencies {
 	implementation(project(":detekt-api"))
 
 	testImplementation(project(":detekt-test"))
+  testImplementation("com.google.dagger:dagger:2.17")
 	testRuntimeOnly("org.junit.platform:junit-platform-launcher:$junitPlatformVersion")
 	testRuntimeOnly("org.jetbrains.spek:spek-junit-platform-engine:$spekVersion")
 }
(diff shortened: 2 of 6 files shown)


detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -473,6 +473,7 @@ style:
     active: false
   UnnecessaryAbstractClass:
     active: false
+    excludeAnnotatedClasses: "dagger.Module"
```
schalkms: I think `excludeAnnotatedClasses` should per default be empty.
Not everyone uses this framework.

detekt-rules/build.gradle.kts
```diff
@@ -14,6 +14,7 @@ dependencies {
 	implementation(project(":detekt-api"))
 
 	testImplementation(project(":detekt-test"))
+  testImplementation("com.google.dagger:dagger:2.17")
```
schalkms: Why should detekt have a dependency to this framework? Is there any usage?

detekt-rules/src/test/resources/cases/UnnecessaryAbstractClassNegative.kt
```diff
@@ -2,6 +2,8 @@
 
 package cases
 
+import dagger.Module
```
schalkms: Ah here's the usage of dagger.
Please use an annotation in the stdlib for the test case or a custom one.

detekt-cli/src/main/resources/default-detekt-config.yml
vanniktech: In my opinion, Detekt should have good default values. In 99% of the cases Dagger modules get flagged from Detekt. If we have it as a default it won't be flagged and less configuration is needed.

detekt-rules/build.gradle.kts
vanniktech: Yup in the Case file.

detekt-rules/src/test/resources/cases/UnnecessaryAbstractClassNegative.kt
vanniktech: Check https://github.com/arturbosch/detekt/pull/1091#discussion_r215885103 first please. I can still revisit.

vanniktech:
> Please remove the dagger dependency to make this mergeable.

It's only applicable for tests so it should not be problem.

schalkms:
Anyways, adding a (potentially huge) framework just for 1 annotation test case doesn't make much sense.

vanniktech:
So if I change the annotation that we will use for testing this is good to go?

schalkms:
Yep! The `excludeAnnotatedClasses` property is discussable IMHO.

vanniktech:
Adjusted to your feedback. @Mauin what do you think about having a default for Dagger?

PR #1098, opened 2018-09-10
fix false positive in UnnecessaryParentheses

### LINKED ISSUE #1051: False positive: Unnecessary parentheses for functions with two function parameters
I have a function (inline and extension if that matters) with a signature like this:

```kt
fun test(first: (String) -> String, second: (String) -> String) {}
```

This method is called like that:

```kt
test({ something }) { something else }
```

Detekt reports:

```
detekt.UnnecessaryParentheses: Parentheses around the lambda ({ something})
                               are unnecessary and can be removed.
```

which is not the case if I have not overlooked something.

Mauin:
Fixes #1051. The 2nd lambda is not part of the `KtParameterList` but another `KtLambdaArgument` attached to the `CallExpression`.

Added a bunch of other tests for similar cases to ensure we handle this case correctly.


arturbosch:
Good to see you back working on detekt :)! Have you tested this fix on our analysis projects (get_analysis_projects.groovy script ^^)?

PR #1099, opened 2018-09-10
don't report method as complex if it returns a single objectexpression

### LINKED ISSUE #1037: ComplexMethod false positive for returning anonymous inner class
## Expected Behavior
No detection.

## Current Behavior
```
ComplexMethod - 11/10 - [f] at  ...
```

Run `ComplexMethod` on this:
```kotlin
fun f(): List<Any> {
    return object : List<Any> {
        override val size: Int get() = TODO("not implemented")

        override fun contains(element: Any): Boolean {
            TODO("not implemented")
        }

        override fun containsAll(elements: Collection<Any>): Boolean {
            TODO("not implemented")
        }

        override fun get(index: Int): Any {
            TODO("not implemented")
        }

        override fun indexOf(element: Any): Int {
            TODO("not implemented")
        }

        override fun isEmpty(): Boolean {
            TODO("not implemented")
        }

        override fun iterator(): Iterator<Any> {
            TODO("not implemented")
        }

        override fun lastIndexOf(element: Any): Int {
            TODO("not implemented")
        }

        override fun listIterator(): ListIterator<Any> {
            TODO("not implemented")
        }

        override fun listIterator(index: Int): ListIterator<Any> {
            TODO("not implemented")
        }

        override fun subList(fromIndex: Int, toIndex: Int): List<Any> {
            TODO("not implemented")
        }
    }
}
```

## Context
The method is quite simple:
```kotlin
fun f(): List<Any> {
    return object : List<Any> { ... }
}
```

## Your Environment

* Version used: 1.0.0.RC7-3
* Link to your project: private but full repro provided above

Mauin:
Fixes #1037. 

If a method only returns an object expression we don't report it as a complex method.. Not sure if this is the best way.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
@@ -9,9 +9,12 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.ThresholdRule
 import io.gitlab.arturbosch.detekt.api.ThresholdedCodeSmell
 import io.gitlab.arturbosch.detekt.api.internal.McCabeVisitor
+import io.gitlab.arturbosch.detekt.rules.asBlockExpression
+import org.jetbrains.kotlin.kotlinx.coroutines.experimental.NonCancellable.children
 import org.jetbrains.kotlin.psi.KtBlockExpression
 import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.KtObjectLiteralExpression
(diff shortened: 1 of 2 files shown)


arturbosch:
Hm, yeah I know what you mean, feels like a hack.
I'm curious if this could also fix this:
```kotlin

	override fun visitClassOrObject(classOrObject: KtClassOrObject) {
		if (classOrObject.isObjectOfAnonymousClass()) return
	}
```

Mauin:
That does not work, at least if we do it in the rule. As we have already created the `McCabeVisitor` and it will visit the function call on its own.
We could probably add the code you proposed to the `McCabeVisitor` itself. But that will also have a whole load of other side-effects, I guess.

schalkms:
To me this looks like a usage for the strategy pattern.

Mauin:
I will close this PR for now to find a better approach, possibly the suggested ones. But with more reports for the `ComplexMethod` rule (see #1131) we should look at this rule more generally as well.

PR #1100, opened 2018-09-10
Prepare changelog for RC9

arturbosch:


arturbosch:
Yep, I was just sleep yesterday, will at this at the top before the changes :)

docs/pages/changelog.md
```diff
@@ -5,6 +5,207 @@ keywords: changelog, release-notes, migration
...
+ subprojects {
+    detekt {
+        // properties
+    }
+ }
+ ```
+can be applied to the root build file.
+
+A full `detekt` configuration for multi-module gradle project could look like `detekt`'s own build file.
+Attention(!) this must be translated to groovy if you do not use `kotlin-dsl`.
```
Mauin: We can additionally also link to the pages on the website which contain these sections.

README.md
```diff
@@ -57,7 +57,7 @@ plugins {
 }
 
 detekt {
-    version = "[version]"
+    toolVersion = "[version]"
```
Mauin: Good catch!

3flex:
Will the Gradle DSL be frozen once this goes out? Or will further changes be accepted before detekt 1.0.0?

Now the Gradle plugin has been updated I'll look at updating the PR for type resolution that I started some time ago. There may be things related to the Gradle plugin that have to change to support that, which will possibly result in changes to the DSL if it doesn't support what's required e.g. automatic Gradle task generation per source set (maybe it has this, I haven't played with it yet).

vanniktech:
Can we release RC9?

arturbosch:
Its out.

PR #1110, opened 2018-09-14
update kotlin to v1.2.70

Mauin:
Release notes: https://blog.jetbrains.com/kotlin/2018/09/kotlin-1-2-70-is-out/

Currently does not seem to compile because of version conflicts of the Kotlin Plugin in the Kotlin-DSL

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -212,7 +212,7 @@ subprojects {
 		}
 	}
 
-	val kotlinVersion: String  by project
+	val kotlinVersion: String by project
 	val junitEngineVersion: String by project
 	val assertjVersion: String by project
 	val spekVersion: String by project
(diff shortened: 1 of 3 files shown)


3flex:
Kotlin 1.2.71 is out. I think the version used in the Gradle plugin will have to match what is provided by Gradle Kotlin DSL, since the plugin itself has `kotlin-dsl` as a dependency.

The rest of the project can upgrade to 1.2.71 once `kotlin-dsl` is removed as a dependency which will happen when #1151 is merged, so only the Gradle plugin needs to be held back.

When I tried making updating to 1.2.70 I came across failing tests. There's an API change in the compiler, see https://github.com/JetBrains/kotlin/commit/48017a81ee5dfb6596258f0f66ca0aa9a65efe26

I tried substituting KtClassOrObject in place of usage of KtClass in the failing code, but it didn't work. But at least this skips a couple of steps fixing things when the upgrade happens.

Mauin:
I would say lets merge #1151 first, then update everything to v1.2.71 in one go, fixing the potential compiler issues as we go.

arturbosch:
Ok #1151 merged, @Mauin can you please refactor this?

PR #1112, opened 2018-09-14
Updated contributors list

schalkms:


Mauin:
Tend to agree with @vanniktech on this one. We could keep our Readme a bit shorter and/or add more meaningful content on how detekt works/how to set it up/how to get started with contributing.

arturbosch:
Which functionality? The contributors graph? I look at the contributors graph if I want to see if a project is active and do not click on each contributor to check their commits.
I like the idea of writing the names down as a symbol of appreciation as they invested time in this project. We invest little time to add them to the readme.

PR #1119, opened 2018-09-14
Use plugin version as default Detekt version

### LINKED ISSUE #1118: Remove hardcoded default version in Detekt Gradle Plugin
## Expected Behavior
Detekt Gradle Plugin reads default version from `gradle.properties`

## Current Behavior
https://github.com/arturbosch/detekt/blob/master/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt#L100

## Context
Having the hardcoded version requires us to manually change it every release.

rock3r:
Removes hardcoded default version in Detekt Gradle Plugin. Closes #1118.

### DIFF
--- a/detekt-gradle-plugin/build.gradle.kts
+++ b/detekt-gradle-plugin/build.gradle.kts
@@ -123,6 +123,32 @@ tasks.withType(DokkaTask::class.java) {
 	outputDirectory = "$buildDir/javadoc"
 }
 
+
+val generateDefaultDetektVersionFile by tasks.creating {
+	val defaultDetektVersionFile =
+			File("$buildDir/generated/src/io/gitlab/arturbosch/detekt", "PluginVersion.kt")
+
+	outputs.file(defaultDetektVersionFile)
+
+	doFirst {
+		defaultDetektVersionFile.parentFile.mkdirs()
+		defaultDetektVersionFile.writeText("""
+			package io.gitlab.arturbosch.detekt
+
+			internal const val DEFAULT_DETEKT_VERSION = "${version}"
+			"""
+				.trimIndent()
+		)
+	}
+}
+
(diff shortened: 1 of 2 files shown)


rock3r:
CC @Mauin

.travis.yml
```diff
@@ -6,7 +6,7 @@ script:
   - ./gradlew build -PwarningsAsErrors=true
   - ./gradlew shadowJar
   - java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar --help
-  - java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar -i . --baseline ./reports/baseline.xml -f ".*/resources/.*, .*/build/.*" -c ./detekt-cli/src/main/resources/default-detekt-config.yml,./reports/failfast.yml
+  - java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar -i . --baseline ./reports/baseline.xml -f ".*/resources/.*;.*/build/.*" -c ./detekt-cli/src/main/resources/default-detekt-config.yml,./reports/failfast.yml
```
arturbosch: Why was this necessary? We support spliting with ",".

Mauin: Commas _should_ be fine. See also:
https://github.com/arturbosch/detekt/blob/eaa4de29b8a22e23dcfed15baae137147d3b563d/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/PathConverters.kt#L33

If they're not working we should create a new issue and investigate this.

rock3r: I'll open an issue then, it was not working with the comma and that was all I needed to change to make it green...

rock3r: Did some more testing and reverted to a comma, which works, but the space has to go to make it work. I'll open an issue for that.

rock3r: Opened #1125 @Mauin @arturbosch

PR #1124, opened 2018-09-15
#1116/Print report paths

### LINKED ISSUE #1116: Regression report paths are no longer printed when using the CLI
After updating to RC9 and adjusting the `--output` argument to `--report`, I no longer see that the reports are being generated.

This was requested in #801 and also implemented.

Can we add this back again?

rock3r:
This PR fixes #1116 by adding back the printing of the active reports' path when the files are written out.

It also adds some tests to `OutputFacade` to avoid future regressions (unfortunately those rely on I/O at the moment, but avoiding it would require a much larger rewrite of how `OutputReport`s are written out), and a minor cleanup on `FindingsReport` that the IDE was complaining about.


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FindingsReport.kt
```diff
@@ -16,9 +16,9 @@ class FindingsReport : ConsoleReport() {
 		val findings = detektion.findings
 		val totalDebt = DebtSumming()
 		return with(StringBuilder()) {
-			findings.forEach {
+			findings.forEach { rulesetFindings ->
```
rock3r: The IDE was complaining about the shadowed `it` in the `issuesString` assignation, so I humoured it and added an explicit name for the outer lambda.

rock3r:
@Mauin or someone with admin access to Appveyor, can you try restarting the job? It has gone green on another PR.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ReportPath.kt
```diff
@@ -12,12 +12,24 @@ import java.nio.file.Paths
...
+			if(parts.size != 2) {
+				throw IllegalArgumentException("Must consist of exactly two parts (report-id:path).")
+			}
 			val (kind, path) = parts
+			assertNotEmpty(kind, path)
 			return ReportPath(defaultMapping(kind), Paths.get(path))
 		}
 
+		private fun assertNotEmpty(kind: String, path: String) {
+			if (kind.isEmpty()) {
```
vanniktech: we could use `require(!kind.isEmpty()) { "The kind of report must not be empty" }`

should be changed in all places then though

rock3r: Happy to change.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ReportPath.kt
```diff
@@ -12,12 +12,24 @@ import java.nio.file.Paths
 data class ReportPath(val kind: String, val path: Path) {
 	companion object {
 		fun from(input: String): ReportPath {
-			val parts = input.split(":")
-			assert(parts.size == 2) { "Must consist of exactly two parts (report-id:path)." }
+			val parts = input.split(":", limit = 2)
+			if(parts.size != 2) {
```
rock3r: I have converted this to an explicit check as assertions don't run on release builds, but we always want to validate this condition.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ReportPath.kt
rock3r: Done in 8ab73ca14f7d1914b9fa67e9b2adb0c07ae0bd4a

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ReportPath.kt
```diff
@@ -12,12 +12,18 @@ import java.nio.file.Paths
 data class ReportPath(val kind: String, val path: Path) {
 	companion object {
 		fun from(input: String): ReportPath {
-			val parts = input.split(":")
-			assert(parts.size == 2) { "Must consist of exactly two parts (report-id:path)." }
+			val parts = input.split(":", limit = 2)
+			require(parts.size == 2) { "Must consist of exactly two parts (report-id:path)." }
```
rock3r: I have converted this to an explicit check as assertions don't run on release builds, but we always want to validate this condition.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/OutputFacade.kt
```diff
@@ -5,16 +5,17 @@ import io.gitlab.arturbosch.detekt.api.Detektion
...
 
 /**
  * @author Artur Bosch
  * @author Marvin Ramin
  */
 class OutputFacade(arguments: CliArgs,
 				   private val detektion: Detektion,
-				   private val settings: ProcessingSettings) {
+				   private val settings: ProcessingSettings,
+				   private val printStream: PrintStream = System.out) {
```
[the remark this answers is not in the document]
arturbosch: Please use the `errorStream` from the `ProcessingSettings` which in the unspecified case is also `System.out`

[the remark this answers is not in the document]
arturbosch: Ok, it would be `null` but you can write `?: System.out`.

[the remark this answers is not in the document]
arturbosch: Ah uh, its actually called errorPrinter for a reasons e.g. core module ...
Lets introduce a `val outPrinter: PrintStream = System.out` for the ProcessingSettings. There are actually tools which use the cli and this could be a good change.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/ReportPathTest.kt
```diff
@@ -0,0 +1,101 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.Assertions.assertThatIllegalArgumentException
+import org.jetbrains.spek.api.Spek
+import org.jetbrains.spek.api.dsl.given
+import org.jetbrains.spek.api.dsl.it
+import java.nio.file.Paths
+
+class ReportPathTest : Spek({
+
+	if (IS_WINDOWS) {
```
arturbosch: +1

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/OutputFacade.kt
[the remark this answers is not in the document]
rock3r: Hm I only moved it to be a ctor parameter tho :)

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/ReportPathTest.kt
rock3r: Not the best solution as ideally we'd want to have all tests on all platforms, but we use `Path` quite a lot and it's the only way to test for both platforms' path formats...

rock3r:
I may have hit a compiler issue...

```
Exception in thread "main" java.lang.NoSuchMethodError: io.gitlab.arturbosch.detekt.core.ProcessingSettings.<init>(Ljava/util/List;Lio/gitlab/arturbosch/detekt/api/Config;Ljava/util/List;ZZLjava/util/List;Ljava/util/concurrent/ExecutorService;Ljava/io/PrintStream;)V
	at io.gitlab.arturbosch.detekt.cli.runners.Runner.createSettings(Runner.kt:36)
	at io.gitlab.arturbosch.detekt.cli.runners.Runner.execute(Runner.kt:20)
	at io.gitlab.arturbosch.detekt.cli.Main.main(Main.kt:24)
```

But that method _is_ there (the IDE sees it and navigates to it). I'll try removing the secondary ctor from `ProcessingSettings` and seeing how it goes.

rock3r:
I think I figured out what the issue is. When `gradlew build` runs the `detekt-core:detekt` task, the plugin uses the `detekt` configuration's classpath to launch the CLI — see:

https://github.com/arturbosch/detekt/blob/815a74558bb04ac8a919fe8c7ff501641c015dad/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt#L15

The problem is that in that configuration, the current module (in this case, `detekt-core` is not grabbed from a remote repo, and is not the released version; the resolved classpath looks something like this:

```
-cp
/Users/rock3r/.gradle/caches/modules-2/files-2.1/io.gitlab.arturbosch.detekt/detekt-cli/1.0.0.RC9/38abb0174959b0758350df6dba99f77c1f28eba4/detekt-cli-1.0.0.RC9.jar
:/Users/rock3r/src/detekt/detekt-core/build/libs/detekt-core-1.0.0.RC9.jar
:/Users/rock3r/.gradle/caches/modules-2/files-2.1/io.gitlab.arturbosch.detekt/detekt-rules/1.0.0.RC9/c37168b64b009a2ebc030c55dbbf0d9f509b9bfe/detekt-rules-1.0.0.RC9.jar
:/Users/rock3r/.gradle/caches/modules-2/files-2.1/com.beust/jcommander/1.72/6375e521c1e11d6563d4f25a07ce124ccf8cd171/jcommander-1.72.jar
:/Users/rock3r/src/detekt/detekt-api/build/libs/detekt-api-1.0.0.RC9.jar
:/Users/rock3r/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-compiler-embeddable/1.2.61/8f4812486d21943cea3f170d53537e628f7eef8c/kotlin-compiler-embeddable-1.2.61.jar
:/Users/rock3r/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-reflect/1.2.61/6fd5bcfc9ffc446dd147ea006bee7ef5f0ad8ca4/kotlin-reflect-1.2.61.jar
:/Users/rock3r/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib/1.2.61/5bc44acc4b3f0d19166ae3e50454b41e8ff29335/kotlin-stdlib-1.2.61.jar
:/Users/rock3r/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-script-runtime/1.2.61/75e4bfc103b71324be6fc862ce806d5399508a14/kotlin-script-runtime-1.2.61.jar
:/Users/rock3r/.gradle/caches/modules-2/files-2.1/org.yaml/snakeyaml/1.23/ec62d74fe50689c28c0ff5b35d3aebcaa8b5be68/snakeyaml-1.23.jar
:/Users/rock3r/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib-common/1.2.61/772de03e12d932f489e41aef997d26c20a4ebee6/kotlin-stdlib-common-1.2.61.jar
:/Users/rock3r/.gradle/caches/modules-2/files-2.1/org.jetbrains/annotations/13.0/919f0dfe192fb4e063e7dacadee7f8bb9a2672a9/annotations-13.0.jar
```

Now, in this case, the `detekt-core` module contains a changed API, the ctor for `ProcessingSettings`. That's what's breaking the build: the plugin is injecting in the CLI classpath the project's current state of `:detekt-core` instead of the expected `detekt-core-1.0.0.RC9.jar` from the remote.

@Mauin this is a rather big issue as it'll happen whenever we change any method/class in the core package; any idea on how to sort this out? I don't know the Gradle plugin well enough to venture on a fix myself...

In the meantime I also noticed two other problems, that I fixed here:
 * The Gradle plugin tests were broken — at least on my machine — because we were comparing file paths without canonicalising them first (the temp input files path was `/var/...` but the actual resolved path was `/private/var`
 * We didn't notice this because the tests and static analysis for the Gradle plugin were never run on CI. Now we do run them.

arturbosch:
As this is a composite build, I would also expect that the project's `detekt-core` version is pulled and not the RC9 hmm maybe @3flex knows something.

But this could also be a good thing to fail the build if there are any breaking changes in the core module ... hmmm. @rock3r I think the break is because you changed the order of the parameters. If you leave `errorStream = null` and change `outStream` also to be default `null` it should not break.
I think it also makes sense to leave both streams nullable and only `?: System.out` them in the cli module.

rock3r:
In the meantime, pushed a fix for the conflicts introduced by #1134 and improved all the test classes touched by that PR to be more robust.

rock3r:
@arturbosch to be honest I would expect that the whole build uses the _local_ version of the gradle plugin and of all the dependencies. If we want to use the project's `detekt` classpath in the plugin, we must be using the local one instead of the remote one. Otherwise, every time we do local development, we either:
 * run into the same issue as soon as anything in `detekt-core` changes (or, for that matter, any strongly coupled breaking changes across modules used (in)directly by the plugin or CLI)
 * need to have a way that says to the build "use the local project modules instead of the remote modules" — would this be just a matter of not using a separate project for the gradle plugin but have it as a normal module?

If we did break as you suggested on any changes to the core module, we'd be unable to evolve the core module without breaking the plugin. Maybe the issue is the assumption that all plugin versions need to be compatible with all versions of the tool, but I wouldn't expect it to be. We should likely specify what versions of Detekt the plugin is compatible with (e.g., by providing an enum of supported `toolVersion`s instead of allowing free-form input).

arturbosch:
Yeah, actually me too... What was the advantage to have a composite build and no cli dependency on the gradle plugin @Mauin @3flex . As @rock3r mentioned the actual cli version can be overridden by the toolVersion flag.

3flex:
Composite build makes it much easier to modify the Gradle plugin, then run `./gradlew detekt` and have the latest version of the plugin build and then test the detekt project itself. Made iteration on the plugin much easier. No benefits beyond that.

I don't actually think the composite build is the cause of the issues here, but in case it is and to reduce confusion I can reverse the introduction of the composite build change easily. Instead, it can be built as a composite build as needed using the `--include-build` command line flag when frequent changes are being made to the Gradle module and theoretically get the same benefits. I'll try to create the necessary PR later today and merge that change in.

Gradle is powerful but hard to understand :/

rock3r:
Thanks a lot @3flex! I am a bit at a loss here, I have to admit. I am not a Gradle super expert :)

We probably want to make iterating on the plugin a little easier, as you said it is a bit tricky otherwise to do. Maybe we can figure out a way (copying from other plugins, even?) that works well for us

3flex:
I _think_ I've figured out what's happening.

The Gradle plugin is built using composite build. When it runs, it pulls the specified version of `detekt-cli` (currently RC9) from the best available location, which is a remote location.

Then detekt-cli resolves the transitive dependencies. In the failing example (`detekt-core:detekt`) detekt-cli has detekt-core as a transitive dependency. Since there's already an RC9 version of detekt-core on the classpath from the project, Gradle uses that as its dependency. detekt-core then depends on detekt-api, which is also already on the classpath, so is provided as a dependency.

So we end up with a mix of remote artifacts (detekt-cli) and local (detekt-core, detekt-api) which are not compatible, even though all have RC9 version.

@rock3r can you try overriding the dependency for the detekt configuration? Hopefully it's as easy as adding this in the subprojects.dependencies block in top-level build.gradle.kts: `detekt(project(":detekt-cli"))`

If not can you please report back what Gradle then uses as the classpath for `detekt-core:detekt` task?

3flex:
@rock3r I checked out your branch and tested locally on your branch - with the forced dependency on the project's detekt-cli artifact it works as expected.

Mauin:
@3flex Do you want to create a PR with a fix for the issue?

Mauin:
Hm.. I'd even be fine with removing the composite build and at least adding good documentation on how to build & test changes to the Gradle Plugin with the `--include-build` Gradle flags in a README.md somewhere.

However if dependency substitution works that would be great. I guess we'd replace the `usedDetektVersion` with `latest` in local builds?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/OutputReport.kt
```diff
@@ -11,11 +11,14 @@ abstract class OutputReport : Extension {
...
 
+	open val name
+		get() = this::class.simpleName
+
 	fun write(filePath: Path, detektion: Detektion) {
 		val reportData = render(detektion)
 		if (reportData != null) {
-			assert(filePath.endsWith(ending)) {
-				"The ${ending.toUpperCase()} report needs to have a file ending of type .$ending."
+			assert(filePath.fileName.toString().endsWith(ending)) {
```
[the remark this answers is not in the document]
marschwar: In my opinion it should be `required` instead of `assert` as the latter will not raise an exception unless explicitly enabled.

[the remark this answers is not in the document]
rock3r: I agree but I didn't want to go and change everything in here, especially since it's not the focus of the PR...

3flex:
@rock3r with the changes that have now been made to the Gradle plugin, if you add `detekt(project(":detekt-cli"))` to the subproject dependencies list in the top level build.gradle.kts file after rebasing, the build should hopefully pass (once #1177 is merged - or after rebasing you can remove that line yourself).

rock3r:
@vanniktech only need time to merge in things, hope this afternoon or tomorrow I'll have some.

rock3r:
@3flex looks like the Gradle plugin now doesn't have a `settings.gradle.kts` nor `gradlew` anymore, but it's still imported as composite build in the root `settings.gradle.kts`. Is this intended?

build.gradle.kts
```diff
@@ -206,6 +206,9 @@ subprojects {
 
 	dependencies {
 		implementation(kotlin("stdlib"))
+
+		detekt(project(":detekt-cli"))
```
arturbosch: why this change?

3flex:
@rock3r it's expected.

:detekt-gradle-plugin:build is run as part of the main build by virtue of it being included as a composite build, so the lines added  to the AppVeyor and Travis scripts are unnecessary. The gradlew script was removed from the plugin project as it's not necessary and means the wrapper had to be updated in two places.

Reverting the changes to AppVeyor and Travis scripts should make things green (if it passes locally)

build.gradle.kts
3flex: There's some discussion further up the thread.

This change means the detekt-cli version from the local check out will be used as the dependency for the gradle plugin when it runs.

This means all detekt dependencies across the build will come from the checked out version. Without this, gradle plugin pulls a version from the gradle plugin portal. This makes it hard to develop when changes are being made to CLI and gradle plugin as rock3r found out.

arturbosch: But does it need to be in the top level build file? I guess as it has its own configuration group it is ok?

arturbosch:
Rebased and merged this manually, thanks for the changes!

PR #1127, opened 2018-09-16
1125/Allow spaces in CLI filters param

### LINKED ISSUE #1125: Spaces around separators in CLI input filters break filters
## Expected Behavior
Passing a `-f ".*/resources/.*, .*/build/.*"` to the CLI (with a space between regexes) should exclude both resources and build folders from Detekt.

## Current Behavior
If there are spaces around the commas/semicolons, the filters aren't parsed correctly and the filtering does not work. Spotted in #1119.

## Steps to Reproduce
1. Check out revision 01bfec71c383b792fff2303e760d01b0c23754dd (or any after that)
2. Build Detekt:
    ```bash
    $ ./gradlew build shadowJar
    ```
3. Run Detekt:
    ```
    java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar -i . --baseline ./reports/baseline.xml -f ".*/resources/.*, .*/build/.*" -c ./detekt-cli/src/main/resources/default-detekt-config.yml,./reports/failfast.yml
    ```
    (notice the space after the comma in the filters)
4. Observe that Detekt picks up an issue in generated code:
    ```
    NewLineAtEndOfFile - [PluginVersion.kt] at .../detekt/detekt-gradle-plugin/build/generated/src/io/gitlab/arturbosch/detekt/PluginVersion.kt:3:55
    ```
    which should be ignored as it's in a `build` directory, which should be filtered out
5. Run Detekt again, this time without the space between the filters' regexes:
    ```bash
    java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar -i . --baseline ./reports/baseline.xml -f ".*/resources/.*,.*/build/.*" -c ./detekt-cli/src/main/resources/default-detekt-config.yml,./reports/failfast.yml
    ```

## Context
This issue was likely never noticed before because we never had generated code, but since #1119 we do.

## Your Environment
* Version used: 1.0.0.RC9
* Operating System and version: macOS 10.13.6
* Link to your project: Detekt itself :)

rock3r:
This PR fixes an issue with the filters being passed to the CLI where spaces around separators were not trimmed correctly, making some filters ineffective. A bunch of tests have been added to avoid future problems, around `PathFilter` and for `Configurations#createPathFilters()`.

This fixes #1125

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt
@@ -14,6 +14,7 @@ fun CliArgs.createPathFilters(): List<PathFilter> = filters.letIfNonEmpty {
 	split(SEPARATOR_COMMA, SEPARATOR_SEMICOLON)
 			.asSequence()
 			.map { it.trim() }
+			.filter { it.isNotEmpty() }
 			.map(::PathFilter)
 			.toList()
 }
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt
@@ -11,7 +11,11 @@ import java.nio.file.Path
  */
 
 fun CliArgs.createPathFilters(): List<PathFilter> = filters.letIfNonEmpty {
-	split(SEPARATOR_COMMA, SEPARATOR_SEMICOLON).map(::PathFilter)
+	split(SEPARATOR_COMMA, SEPARATOR_SEMICOLON)
+			.asSequence()
+			.map { it.trim() }
+			.map(::PathFilter)
+			.toList()
 }
 
 fun CliArgs.createPlugins(): List<Path> = plugins.letIfNonEmpty {
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/ConfigurationsSpec.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/ConfigurationsSpec.kt
@@ -140,6 +140,11 @@ internal class ConfigurationsSpec : Spek({
 			)
 		}
 
+		it("should ignore empty and blank filters") {
+			val filters = CliArgs().apply { filters = " ,,.*/three" }.createPathFilters()
+			assertThat(filters).containsExactly(PathFilter(".*/three"))
+		}
+
 		it("should fail on invalid filters values") {
 			assertFails { CliArgs().apply { filters = "*." }.createPathFilters() }
 			assertFails { CliArgs().apply { filters = "(ahel" }.createPathFilters() }
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/ConfigurationsSpec.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/ConfigurationsSpec.kt
@@ -1,5 +1,6 @@
 package io.gitlab.arturbosch.detekt.cli
 
+import io.gitlab.arturbosch.detekt.core.PathFilter
 import io.gitlab.arturbosch.detekt.test.resource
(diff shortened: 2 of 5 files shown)


rock3r:
The CI is failing due to the Gradle `plugin-publish` plugin JARs having disappeared from the Gradle plugin repos...

rock3r:
@vanniktech actually after checking a bit more in the code, reports cannot be chained right now. You can't do `-r "plain:/path/to/report.txt,xml:/path/to/report.xml"`, you explicitly need a `-r` per each report, so what you're asking doesn't really apply.

vanniktech:
I've tested this out and chaining them did work for me.

rock3r:
Might be some JCommander magic as we don't have explicit provisions in code to handle the case, then. You're right tho to say you can do it, I didn't read the help message for `-r` and that mentions you can chain things. Probably semicolon-separated won't work tho.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/ConfigurationsSpec.kt
```diff
@@ -78,6 +79,73 @@ internal class ConfigurationsSpec : Spek({
...
+			)
+		}
+
+		it("should load multiple semicolon-separated filters with no spaces around semicolons") {
+			val filters = CliArgs().apply { filters = ".*/one/.*;.*/two/.*;.*/three" }.createPathFilters()
+			assertThat(filters).containsExactly(
+					PathFilter(".*/one/.*"),
+					PathFilter(".*/two/.*"),
+					PathFilter(".*/three")
+			)
```
schalkms: We might want to refactor the `PathFilter()`'s in the assert clause

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt
```diff
@@ -11,7 +11,11 @@ import java.nio.file.Path
  */
 
 fun CliArgs.createPathFilters(): List<PathFilter> = filters.letIfNonEmpty {
-	split(SEPARATOR_COMMA, SEPARATOR_SEMICOLON).map(::PathFilter)
+	split(SEPARATOR_COMMA, SEPARATOR_SEMICOLON)
+			.asSequence()
+			.map { it.trim() }
```
arturbosch: I think we also need a `filter { it.isNotEmpty() }`, needed this for the `SplitPattern` if I'm not mixing things up ...

rock3r: Yeah we could while we're at it...

rock3r: I mean the current behaviour is that when creating the `Path` in the `PathFilter` it will crash, I believe. Not sure if you wanted a silent dropping of those instead?

arturbosch: If it instead crashes, that would be good I guess?

rock3r: Yeah that should be fine. We can cover with additional tests if you want?

arturbosch: That would be awesome!

rock3r: I actually implemented what you asked initially — we now filter out empty and blank filter paths in `Configurations`. `PathFilter` also now blows up if you mistakenly pass it an empty/blank path, which should avoid mistakenly specifying those as a filter. There's tests covering both behaviour changes, at any rate.

See 899dd98e60edc915ba5e1c8c57f366d7c3a43790 and 76bad42d39131f7ac7f8152cc8161d0900cfac5b

rock3r:
This failed on Appveyor because of a Gradle plugins repo read error... can someone restart the build there?

arturbosch:
Restarted :)

rock3r:
Thanks @arturbosch, green now :)

arturbosch:
Also thanks for the PR!

PR #1128, opened 2018-09-16
Upload JUnit test results to AppVeyor

rock3r:
Does what it says on the tin. This allows to troubleshoot failing tests more easily for those that don't have a Windows machine to test on...

Before | After
--- | ---
before | after


arturbosch:
Indeed!

Issue #1122, opened 2018-09-15
ObjectPropertyNaming/privatePropertyPattern not loaded from configuration


Tolriq:
```
  ObjectPropertyNaming:
    active: true
    privatePropertyPattern: '*'
```

Will still triggers errors with message when they do not match the default pattern:
ObjectPropertyNaming -> Private object property names should match the pattern: (_)?[A-Za-z][A-Za-z0-9]*

## Your Environment

* Version used: RC9

Tolriq:
The first try was to just add an _ and duplicate the public one ;)

Same result.

rock3r:
I cannot reproduce the issue; I tried adding a test for `ObjectPropertyNaming` passing a config with `privatePropertyPattern: '.*'` and it doesn't detect any private properties on an `Object` with any weird naming. Seems to be working fine for me. What's your specific case?

rock3r:
See the PR linked above for the specific test added.

Mauin:
I just noticed that in your configuration you define `propertyPattern` twice. Does removing one of those lines solve the issue?

Tolriq:
Nope fixed to:
```
  ObjectPropertyNaming:
    active: true
    constantPattern: '[A-Za-z][_A-Za-z0-9]*'
    propertyPattern: '[A-Za-z][_A-Za-z0-9]*'
    privatePropertyPattern: '(_)?[A-Za-z][_A-Za-z0-9]*'
```

And still the issue.
(Duplicate was caused by diff error when updating the conf to new generated config and I did not pay attention enough)

For test completion it's the same with

```
  ObjectPropertyNaming:
    active: true
    privatePropertyPattern: '(_)?[A-Za-z][_A-Za-z0-9]*'
```

Mauin:
Hey @Tolriq would it be possible to share your entire config.yml? Just so we can debug with the exact same values.

Do you use `failfast: true` in your config?

Tolriq:
No fail fast.

```
autoCorrect: true
failFast: false

test-pattern: # Configure exclusions for test sources
  active: true
  patterns: # Test file regexes
    - '.*/test/.*'
    - '.*Test.kt'
    - '.*Spec.kt'
  exclude-rule-sets:
    - 'comments'
  exclude-rules:
    - 'NamingRules'
    - 'WildcardImport'
    - 'MagicNumber'
    - 'MaxLineLength'
    - 'LateinitUsage'
    - 'StringLiteralDuplication'
[...]

PR #1130, opened 2018-09-16
#1122/Add tests for ObjectPropertyNaming for private val overrides

rock3r:
Following the discussions on #1122 I am adding a test to cover a supposed bug (cannot repro tho) and improve other `ObjectPropertyNaming` tests slightly.


pavlospt:
Just to give more feedback, this is also happening to us as well, after updating to RC9. We did not specify a private property pattern before, but even after specifying the same one as for const and public priperties, it would still report it. I can take a look at the rule if you think it is not reproducible! An example value is: “private val PATTERN_0 = foo” inside a companion object withe the regex pattern being: [A-Za-z][_A-Za-z0-9]*

schalkms:
@pavlospt further investigation would be very helpful in this case.
reference #1122

pavlospt:
@schalkms great, will try to dive more into it.

arturbosch:
@pavlospt further investigation would be awesome! Need to get all that regressions fixed :X :)

PR #1134, opened 2018-09-17
Fix issue when multiple input files are specified

pavlospt:
After Gradle plugin rework, there was a removal of `:` as path separator which caused inputs not being properly split by `DetektInputPathConverter`.

Commit removing `:`: https://github.com/arturbosch/detekt/commit/3b0f992bda34a25cd34bbc237d637f13d9845a1b#diff-fffb30c13416dcda612424cc7a5e3034

The above commit conflicted with the way FileCollection.asPath returns the path of specified files,
because it concatenates the absolute paths with `:`.

Updating InputArgument#toArgument to join files absolute paths with `,` fixes the issue.

### DIFF
--- a/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskKotlinDslTest.kt
+++ b/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskKotlinDslTest.kt
@@ -231,6 +231,7 @@ internal class DetektTaskKotlinDslTest : Spek({
 			val customSourceLocation = "gensrc/kotlin"
 			val detektConfig = """
 					|detekt {
+					| debug = true
 					|	input = files("$customSourceLocation")
 					|}
 				"""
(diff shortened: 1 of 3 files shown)


detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskKotlinDslTest.kt
```diff
@@ -231,6 +231,7 @@ internal class DetektTaskKotlinDslTest : Spek({
 			val customSourceLocation = "gensrc/kotlin"
 			val detektConfig = """
 					|detekt {
+					| debug = true
 					|	input = files("$customSourceLocation")
```
schalkms: [MINOR] - strange formatting

pavlospt: Indeed not sure what caused it. If it’s blocking i can get it fixed. The IDE was not showing it like that.

arturbosch: Well its a mega blocker, but would be nice to have everything formatted properly :). I think you just need to use a tab. thanks!

pavlospt: Done!

arturbosch: Thank you! :)

PR #1135, opened 2018-09-17
#1131 complex method should be less strict with simple when entries

### LINKED ISSUE #1131: ComplexMethod triggering mechanism
The following function gets flagged as complex:

```kotlin
internal fun Map<String, Any>.asBundle(): Bundle {
  val bundle = Bundle(size)

  for ((key, value) in this) {
    val transformedKey = key.asFcmId()

    when (value) {
      is Int -> bundle.putInt(transformedKey, value)
      is String -> bundle.putString(transformedKey, value)
      is Float -> bundle.putFloat(transformedKey, value)
      is Double -> bundle.putDouble(transformedKey, value)
      is Byte -> bundle.putByte(transformedKey, value)
      is Short -> bundle.putShort(transformedKey, value)
      is Long -> bundle.putLong(transformedKey, value)
      is Boolean -> bundle.putBoolean(transformedKey, value)
      else -> throw IllegalArgumentException("Unexpected type $value")
    }
  }

  return bundle
}
```

I don't think to be honest that it's complex. The reason for that is assumingly the branches from the when statement.

Can we ease the rule in this case or am I being biased?

rock3r:
This fixes #1131 by weighing simple `when` entries to `0.5` (was `1.0`) by default, and allowing users to customise the weight if they want to.

This requires the McCabe visitor to internally keep a floating-point sum of the complexities, but we still emit an integer value to keep backwards compatibility (and for simplicity).

This PR also improves `ComplexMethod`'s tests, and adds better facilities to test `Finding`s and `ThresholdedCodeSmell`s.

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -77,6 +77,7 @@ complexity:
     active: true
     threshold: 10
     ignoreSingleWhenExpression: false
+    simpleWhenEntryWeight: 0.5
   LabeledExpression:
     active: false
   LargeClass:
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
@@ -23,7 +23,7 @@ import org.jetbrains.kotlin.psi.KtWhenExpression
  * @configuration threshold - MCC threshold for a method (default: 10)
  * @configuration ignoreSingleWhenExpression - Ignores a complex method if it only contains a single when expression.
  * (default: false)
- * @configuration simpleWhenEntriesWeight - The weight used for simple (braceless) when entries. (default: 0.5)
+ * @configuration simpleWhenEntryWeight - The weight used for simple (braceless) when entries. (default: 0.5)
  *
  * @active since v1.0.0
  * @author Artur Bosch
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
@@ -23,6 +23,7 @@ import org.jetbrains.kotlin.psi.KtWhenExpression
  * @configuration threshold - MCC threshold for a method (default: 10)
  * @configuration ignoreSingleWhenExpression - Ignores a complex method if it only contains a single when expression.
  * (default: false)
+ * @configuration simpleWhenEntriesWeight - The weight used for simple (braceless) when entries. (default: 0.5)
  *
  * @active since v1.0.0
  * @author Artur Bosch
@@ -38,12 +39,13 @@ class ComplexMethod(config: Config = Config.empty,
 			Debt.TWENTY_MINS)
 
 	private val ignoreSingleWhenExpression = valueOrDefault(IGNORE_SINGLE_WHEN_EXPRESSION, false)
+	private val simpleWhenEntriesWeight = valueOrDefault(SIMPLE_WHEN_ENTRY_WEIGHT, 0.5)
 
 	override fun visitNamedFunction(function: KtNamedFunction) {
 		if (hasSingleWhenExpression(function.bodyExpression)) {
 			return
 		}
-		val visitor = McCabeVisitor()
+		val visitor = McCabeVisitor(simpleWhenEntriesWeight)
 		visitor.visitNamedFunction(function)
 		val mcc = visitor.mcc
 		if (mcc >= threshold) {
@@ -73,5 +75,6 @@ class ComplexMethod(config: Config = Config.empty,
 	companion object {
 		const val DEFAULT_ACCEPTED_METHOD_COMPLEXITY = 10
 		const val IGNORE_SINGLE_WHEN_EXPRESSION = "ignoreSingleWhenExpression"
+		const val SIMPLE_WHEN_ENTRY_WEIGHT = "simpleWhenEntryWeight"
 	}
 }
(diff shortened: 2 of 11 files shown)


rock3r:
CC @Mauin

rock3r:
Ticket opened: #1137

rock3r:
@Mauin can you re-trigger the JDK 9 build on AppVeyor? It failed with some random socket error.

detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -77,6 +77,7 @@ complexity:
     active: true
     threshold: 10
     ignoreSingleWhenExpression: false
+    simpleWhenEntryWeight: 0.5
```
schalkms: I just want to raise the question, if this should be the default value. I would argue for the value 1. Otherwise, detekt would not comply with the McCabe metric. The class `McCabeVisitor` would also need another name.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
```diff
@@ -23,6 +23,7 @@ import org.jetbrains.kotlin.psi.KtWhenExpression
  * @configuration threshold - MCC threshold for a method (default: 10)
  * @configuration ignoreSingleWhenExpression - Ignores a complex method if it only contains a single when expression.
  * (default: false)
+ * @configuration simpleWhenEntryWeight - The weight used for simple (braceless) when entries. (default: 0.5)
```
schalkms: As a user of detekt I would like want to have a bit more documentation for this config option. (range, what effect has value x)

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethodSpec.kt
```diff
@@ -30,14 +35,56 @@ class ComplexMethodSpec : Spek({
...
+					SourceLocation(42, 1)
+			)
+		}
+
+		it("does not trip for a reasonable amount of simple when entries") {
+			val config = TestConfig(mapOf(ComplexMethod.SIMPLE_WHEN_ENTRY_WEIGHT to "0.5"))
+			val subject = ComplexMethod(config)
+			val code = """
+				internal fun Map<String, Any>.asBundle(): Bundle {
+					val bundle = Bundle(size)
```
schalkms: Can we simplify this function to not have any dependencies to Android dev?
Can we also just have the when brace, since this test case should only test the weight of the when clause?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethodSpec.kt
```diff
@@ -30,14 +35,56 @@ class ComplexMethodSpec : Spek({
...
+					SourceLocation(5, 1),
+					SourceLocation(14, 1),
+					SourceLocation(24, 1),
+					SourceLocation(34, 1),
+					SourceLocation(42, 1)
+			)
+		}
+
+		it("does not trip for a reasonable amount of simple when entries") {
+			val config = TestConfig(mapOf(ComplexMethod.SIMPLE_WHEN_ENTRY_WEIGHT to "0.5"))
```
schalkms: 0.5 is the default value atm, isn't it? One could remove this statement..

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/ThresholdedCodeSmellAssert.kt
```diff
@@ -0,0 +1,41 @@
...
+fun assertThat(thresholdedCodeSmell: ThresholdedCodeSmell) = ThresholdedCodeSmellAssert(thresholdedCodeSmell)
+
+@Suppress("UnsafeCast") // False positive, see issue #1137
+fun FindingAssert.isThresholded(): ThresholdedCodeSmellAssert {
+	isNotNull
+	assert(actual is ThresholdedCodeSmell) { "The finding '$actual' is not a ThresholdedCodeSmell" }
+	return ThresholdedCodeSmellAssert(actual as ThresholdedCodeSmell?)
+}
+
+class ThresholdedCodeSmellAssert(actual: ThresholdedCodeSmell?) :
```
schalkms: Good! Much needed assert class

detekt-cli/src/main/resources/default-detekt-config.yml
vanniktech: I'm voting for 0.5 simple when entries are simple and one liners.

arturbosch:
To be honest, I do not like this change. As @schalkms mentioned, McCabe is all about "number of linearly independent paths" within a program. When using floating values this should be another rule then.

I propose to `not count` the complexity of `simple paths within when` when a config property is true. If all paths are simple, just count the whole when expressions as mccabe=1.

rock3r:
Happy to make the change if there's a consensus, I take the point.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethodSpec.kt
rock3r: There is no dependency in here, it's just the sample code taken 1:1 from the ticket. Nothing is going to depend on Android because a string contains an Android class — we don't even have imports for it :)

rock3r:
I have created a PR on my fork comparing this solution and an ignore-based solution which adheres better to the definition of MCC. Please review it here: https://github.com/rock3r/detekt/pull/2

arturbosch:
@rock3r jap that's 90% how I would want this rule to be :). The other 10% is when `if (entries.filter { it is KtBlockExpression}.size == 0) return 1` :)

arturbosch:
I guess we go with the simple=0 variant?
Can you please open a PR with your new branch?/

arturbosch:
Thank you!

rock3r:
Still WIP but have merged master in, should at least not have conflicts anymore

rock3r:
Finally managed to get the changes @arturbosch requested in.

arturbosch:
Thank you for the changes!
I rebased and merged this manually. Please use rebase when merging the master branch :)
We rebase everything to get a cleaner history ^^

PR #1144, opened 2018-09-21
Set group to Detekt task

CherryPerry:
Detekt task is in _other_ gradle task group now. Is not it better to return it back to _verification_ group?


CherryPerry:
Something like ```Launch detekt analysis with provided configuration```? I am bad in describing things.

arturbosch:
Maybe just `Runs detekt analysis on kotlin files`?

Issue #1123, opened 2018-09-15
HTML report created at the wrong place


Tolriq:
RC9 / Windows 10 / Gradle 4.10.1

```
subprojects {
    apply plugin: 'io.gitlab.arturbosch.detekt'
    detekt {
        reports {
             html.enabled = true
        }
     }
}
```

Will create the html reports as "D" at the root of each module.
But will correctly create the detekt folder at build/reports/detekt

adding a destination for example:
```
destination = file("build/reports/failfast.html")
```

Does not change the behavior and the D files are still created (but no more the folder in reports)

marschwar:
That is very weird. Could you please enable debug, rerun the task and provide the logged call of the cli. What do you mean with `"D"`? Is that the name of the html file?

```
	detekt {
		debug = true
	}
```

Tolriq:
[--input, D:\_Gradle\xxx\xxx\src\main\java, --filters, .*/resources/.*,.*/build/.*, --config, D:\_Gradle\xxx\detek
t-config.yml, --report, html:D:\_Gradle\xxx\xxx\build\reports\detekt\detekt.html, --debug]

The file is named D :)

But D is the drive now that I see the command like I guess using : as a separator is an issue. Should probably add "" around the path to avoid the issue

Pretty sure https://github.com/arturbosch/detekt/issues/1141 is the same issue but the guy have the code on C: :)

schalkms:
Thanks for investigating this, guys! This looks like a duplicate then.

Issue #1141, opened 2018-09-20
RC9 creates files called "C" with <checkstyle> tags inside


sschuberth:
## Expected Behavior

Running `./gradlew --rerun-tasks detekt` inside a Git repository with a mutli-project Gradle project should not create any (untracked) files.

## Current Behavior

After running `./gradlew --rerun-tasks detekt` and `git status` I see

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        analyzer/C
        cli/C
        downloader/C
        model/C
        reporter/C
        scanner/C
        utils-test/C
        utils/C
```

## Steps to Reproduce (for bugs)

1. git clone git clone --recurse-submodules https://github.com/heremaps/oss-review-toolkit.git
1. ./gradlew --rerun-tasks detekt
1. git status

## Context

I was trying to run `detekt` on our Kotlin project to identify code smells.

## Your Environment

* Version used: 1.0.0.RC9
* Operating System and version: Windows 10, 64-bit
* Link to your project: https://github.com/heremaps/oss-review-toolkit

Mauin:
Thanks for the report. This seems to be an issue related to our report path parsing on Windows.
Closing this as a duplicate of #1123

See #1145 for the fix

PR #1145, opened 2018-09-22
Fix issue with wrong report name

pavlospt:
When splitting a report path provided on JCommander, given a Windows path
would cause a false split, that would continue the execution despite the assert check.

This commit fixes this issue by the assumption that an absolute path on Windows
systems, will contain the current disk drive's letter.

Resolves: #1123 - #1141

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ReportPath.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ReportPath.kt
@@ -14,7 +14,8 @@ data class ReportPath(val kind: String, val path: Path) {
 		private const val NUM_OF_PARTS_UNIX = 2
 		private const val NUM_OF_PARTS_WINDOWS = 3
 		private const val REPORT_PATH_SEPARATOR = ":"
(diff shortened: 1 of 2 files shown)


schalkms:
@pavlospt I think we should check the entries of the report for equality, not just checking the size. This caused the bug in the first place. @vanniktech please correct me if you have other intentions.
```Kotlin
it("should parse multiple report entries") {
  assertThat(reports).hasSize(4)
}
```

pavlospt:
@schalkms sure, i can update the test to assert on that!

pavlospt:
@schalkms @vanniktech I enhanced `ReportsSpec` for `path` and `kind` assertions on the generated `ReportPath` entries. LMK if you want me to add anything else :)

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ReportPath.kt
```diff
@@ -11,10 +11,25 @@ import java.nio.file.Paths
  */
 data class ReportPath(val kind: String, val path: Path) {
 	companion object {
+		private const val NUM_OF_PARTS_UNIX = 2
+		private const val NUM_OF_PARTS_WINDOWS = 3
```
arturbosch: waaah, jupp I never use windows >.<

PR #1147, opened 2018-09-22
Prevent overwriting of defaultDependency detekt-cli by additional detekt dependencies

### LINKED ISSUE #1143: Running RC9 gradlew detekt crashes with detekt-formatting included
## Expected Behavior
Detekt runs fine using Gradle and reports formatter problems

## Current Behavior
Detekt crashes if detekt-formatter is included in Gradle configuration
```
Task :detekt FAILED
Error: Could not find or load main class io.gitlab.arturbosch.detekt.cli.Main

* What went wrong:
> Process 'command 'C:\java\jdk1.8.0_131\bin\java.exe'' finished with non-zero exit value 1
```
1. Configure Detekt RC9 for Gradle, detektVersion = 1.0.0.RC9
```
buildscript {
       dependencies {
        classpath "io.gitlab.arturbosch.detekt:detekt-gradle-plugin:${detektVersion}"
...
```
2. Using settings:
```
detekt {
    toolVersion = "${detektVersion}"
    input = files("src/main/kotlin","src/test/kotlin")
    config = files("detekt.yml")
    filters = ".*/resources/.*,.*/tmp/.*"
}
```
3. Include detekt-formatter RC9
```
dependencies {
    detekt("io.gitlab.arturbosch.detekt:detekt-formatting:${detektVersion}")
...
```
4. Run gradlew detekt -> detekt crashes
5. Remove detekt-formatting dependency in Gradle configuration and gradlew detekt runs fine

## Your Environment

* Version used: Detekt RC9
* Operating System and version: Win10 & Gradle 4.9
* Link to your project:

Mauin:
When including an additional dependency in the `detekt` configuration it would overwrite the `defaultdependency` we define in `DetektPlugin`. This behavior is also described in the Gradle documentation: https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/Configuration.html#defaultDependencies-org.gradle.api.Action-
> Execute the given action if the configuration has no defined dependencies when it first participates in dependency resolution.

In case of an extra `detekt` dependency the default dependency the `DetektPlugin` defines is omitted.

However we _require_ `detekt-cli` to be able to run detekt at all. To prevent the overwriting we simply define `detekt-cli` as a regular dependency in the `DetektPlugin`.

The added test fails with the `master` version of `DetektPlugin` and passes with the changes.

fixes #1143

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -82,10 +82,10 @@ class DetektPlugin : Plugin<Project> {
 	private fun determineInput(extension: DetektExtension) = extension.input.filter { it.exists() }
 
 	private fun configurePluginDependencies(project: Project, extension: DetektExtension) {
-		project.configurations.create(CONFIGURATION_DETEKT_PLUGIN) {
+		project.configurations.create(CONFIGURATION_DETEKT_INTERNAL) {
 			isVisible = false
 			isTransitive = true
-			description = "The internal $CONFIGURATION_DETEKT_PLUGIN libraries to be used for this project."
+			description = "The internal $CONFIGURATION_DETEKT_INTERNAL libraries to be used for this project."
 
 			@Suppress("USELESS_ELVIS")
 			val version = extension.toolVersion ?: DEFAULT_DETEKT_VERSION
@@ -111,4 +111,4 @@ class DetektPlugin : Plugin<Project> {
 }
 
 const val CONFIGURATION_DETEKT = "detekt"
-const val CONFIGURATION_DETEKT_PLUGIN = "detektPlugins"
+const val CONFIGURATION_DETEKT_INTERNAL = "detektInternal"
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -81,17 +81,20 @@ class DetektPlugin : Plugin<Project> {
 
 	private fun determineInput(extension: DetektExtension) = extension.input.filter { it.exists() }
 
-	private fun configurePluginDependencies(project: Project, extension: DetektExtension) =
-			project.configurations.create(DETEKT) {
-				isVisible = false
-				isTransitive = true
-				description = "The $DETEKT libraries to be used for this project."
-				defaultDependencies {
-					@Suppress("USELESS_ELVIS")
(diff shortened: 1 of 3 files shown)


3flex:
Instead of changing the detekt-cli dependency from being a default to forcing it to be applied, consider instead leaving it as a default, and creating a separate configuration for any detekt plugins that are to be added to the classpath.

This is the approach taken by the FindBugs plugin and should work well here.

There would then be two configurations in the Gradle plugin: detekt, and detektPlugins. The detekt configuration would remain the same (i.e. don't merge this PR). The new detektPlugins configuration would be combined with the detekt configuration's classpath in DetektInvoker.

Mauin:
@3flex That is something I have considered. However what would the benefit of the two configurations vs this one be?

3flex:
It's the most flexible setup and follows conventions used in other Gradle code quality plugins. People can use the `detekt` configuration when overriding the default detekt-cli dependency (like we need for #1124, unless using dependency substitution which is a hassle) but still allows adding plugins using `detektPlugins` configuration. It's a nice separation of responsibilities.

Mauin:
That makes sense. I split the configurations into `detekt` and `detektPlugins`.

Mauin:
Actually I did it the other way around. `detekt` is still the "external" configuration. The plugin uses `detektPlugins` internally. That way we don't have another breaking change to the API.

We could also go with `detekt` and `detektInternal` to make it clearer.

pavlospt:
`detektInternal` sounds better if it's meant to be only for internal usage!

Mauin:
@3flex As long as we have one "default" configuration and one "public" configuration similarly to what FindBugs does I think we should be fine. The names of which we can still discuss/adjust. I just kept `detekt` as the public configuration as for the breaking API change. 
I see your point about the current API not working correctly. At this point I don't have a strong opinion on the naming of the configurations.

arturbosch:
Mhh, I do not quite understand if this is rdy?
I've also noticed that if I use `detekt-formatting` I need also to add `detekt-cli` to the `detekt configuration`. Does this fix this issue too?

arturbosch:
Just read the issue. Will merge this now, thanks!

3flex:
@Mauin you misunderstood what the FindBugs plugin configurations do, sorry if I didn't explain properly. FindBugs doesn't have a "default" and a "public" configuration. Both "findbugs" and "findbugsPlugins" are part of the public configuration DSL. Neither is private or internal.

From the Gradle docs, the FindBugs plugin: 
> Declares a findbugs configuration which needs to be configured with the FindBugs library to be used. Additional plugins can be added to the findbugsPlugins configuration.

For detekt, the detekt configuration would be used to configure which detekt library to use (detekt-cli). Additional plugins (i.e. detekt-formatting, others) would be added with detektPlugins.

Isn't it better to do it according to Gradle conventions while version is still RC instead of publishing 1.0.0 with a DSL that isn't aligned with other Gradle plugins?

Issue #1146, opened 2018-09-22
Gradle Plugin Tests not executed with default test tasks


Mauin:
Since the Gradle Plugin is using composite builds it's test tasks are not running automatically when e.g. running `./gradlew test` or `./gradlew build`

## Expected Behavior
Gradle Plugin tests should run as part of a regular Gradle task

## Current Behavior
Gradle Plugin tests are not run

## Steps to Reproduce (for bugs)
See the following output of `./gradlew test --dry-run` (notice the absence of any mention of the gradle plugin)
```
:detekt-formatting:compileKotlin SKIPPED
:detekt-formatting:compileJava SKIPPED
:detekt-formatting:processResources SKIPPED
:detekt-formatting:classes SKIPPED
:detekt-formatting:inspectClassesForKotlinIC SKIPPED
:detekt-formatting:jar SKIPPED
:detekt-generator:compileKotlin SKIPPED
:detekt-generator:compileJava SKIPPED
:detekt-generator:processResources SKIPPED
:detekt-generator:classes SKIPPED
:detekt-generator:shadowJar SKIPPED
:detekt-generator:generateDocumentation SKIPPED
:detekt-cli:test SKIPPED
:detekt-core:compileTestKotlin SKIPPED
:detekt-core:compileTestJava SKIPPED
:detekt-core:processTestResources SKIPPED
:detekt-core:testClasses SKIPPED
:detekt-core:test SKIPPED
:detekt-formatting:compileTestKotlin SKIPPED
:detekt-formatting:compileTestJava SKIPPED
:detekt-formatting:processTestResources SKIPPED
:detekt-formatting:testClasses SKIPPED
:detekt-formatting:test SKIPPED
:detekt-generator:compileTestKotlin SKIPPED
:detekt-generator:compileTestJava SKIPPED
:detekt-generator:processTestResources SKIPPED
:detekt-generator:testClasses SKIPPED
:detekt-generator:test SKIPPED
:detekt-rules:compileTestKotlin SKIPPED
:detekt-rules:compileTestJava SKIPPED
:detekt-rules:processTestResources SKIPPED
:detekt-rules:testClasses SKIPPED
:detekt-rules:test SKIPPED
:detekt-sample-extensions:compileKotlin SKIPPED
:detekt-sample-extensions:compileJava SKIPPED
:detekt-sample-extensions:processResources SKIPPED
:detekt-sample-extensions:classes SKIPPED
[...]

PR #1148, opened 2018-09-22
Fix Gradle Plugin Tests

Mauin:
Three tests which were testing for the correct `--input` path were failing.

As the Gradle Plugin tests don't seem to run on CI at the moment (see #1146) I have a follow up PR to enable the tests on CI.


pavlospt:
I have added `private` because those tests were failing locally! @rock3r has refactored them on #1116

rock3r:
Yeah @pavlospt I had the same issue but that "fix" only works for macOS, hence why I changed those tests again myself as your solution wasn't portable enough to work for all

pavlospt:
yup @rock3r thanks for taking care of this, haven't realized it did not work on CI!

Issue #1153, opened 2018-09-23
Make `config` property in `Rule` effectively private


Mauin:
As `config` is usable in every single `Rule` class due to it being protected it is easy to use it directly, which can lead to issues similar to the one described in #1122.

`config` should be effectively private or at least very hard to use compared to the `valueOrDefault`/`valueOrNull` functions defined in `ConfigAware` to prevent similar issues in the future.

arturbosch:
I will rename the property to `ruleSetConfig`. We can't make interface properties private or internal unfortunately.

PR #1154, opened 2018-09-23
use correct instance of valueOrDefault in FormattingRule

Mauin:
Similar to #1152 I think this rule should use the `ConfigAware#valueOrDefault` method instead of accessing the `config` directly.


arturbosch:
This change is actually wrong:
```
	/**
	 * Wrapped configuration of the ruleSet this rule is in.
	 */
	val config: Config
```
Just using `valueOrDefault` will return the `subConfig` of this rule while `config.valueOrDefault` returns properties from the ruleSet itself.
Jap, as you mentioned in #1153 and #1152 maybe we should rename `config` to `ruleSetConfig`?

Mauin:
Ah, I see. Thanks @arturbosch! I just checked all accessors of the `config` property directly and this was the only usage in the rules.

PR #1151, opened 2018-09-23
Tidy the Gradle build files

3flex:
Tidies up the Gradle build files, using some of the newer syntax from Gradle Kotlin DSL, a review of some of the module dependencies, and removing some of the things added with the introduction of the composite build that I'm now aware are not required (though the composite build remains).

Less code = Less bugs (hopefully!)

PR #1161, opened 2018-09-24
Don't publish Gradle plugin to Bintray

3flex:
Doesn't seem there's any value in pushing anything related to the Gradle plugin to Bintray, or synced to Maven Central (and it doesn't look like anything's being pushed there at the moment), so makes sense to remove the config from the Gradle plugin build file.

Kept this separate to #1151 since that's about simplifying the script but not changing build behavior, where this change might affect some workflow I'm not aware of.


arturbosch:
Hm, if I'm not wrong the bintray publishing was neccesary for something .... but I do not remember why anymore ... Will leave this open for now and come to it later.

arturbosch:
For the next release I won't publish the gradle plugin to bintray and see if we need it somehow.

arturbosch:
Couldn't test this really because the gradle folks started to check the plugins before making them public and I thought it does not work because of the bintray stuff :D.
But I will try to release faster now :D.

3flex:
In that case I expect you can still use the plugin hosted on the Gradle Plugin Portal (tweak syntax for Kotlin DSL if you're using that):

```gradle
buildscript {
  repositories {
    maven {
      url "https://plugins.gradle.org/m2/"
    }
  }
  dependencies {
    classpath "io.gitlab.arturbosch.detekt:detekt-gradle-plugin:1.0.0-RC10"
  }
}

```

jayrave:
Didn't know about depending on the plugin portal itself. Thanks! All I had to do was add a new repository & a dependency

```kotlin
plugins {
    `kotlin-dsl`
}

repositories {
    maven("https://plugins.gradle.org/m2/")
    google()
    jcenter()
}

dependencies {
   compile("com.android.tools.build:gradle:3.2.1")
    compile("io.gitlab.arturbosch.detekt:detekt-gradle-plugin:1.0.0-RC10")
}
```

arturbosch:
Seems like bintray publishing is not needed :), will try for next release to delete it and publish the plugin, maybe the publishing plugin is needed for `pluginPublish`task ^^

arturbosch:
Cherry-picked into 6cfaaee721bdb44ba8613e014f98355a8884fb28, thanks!

Issue #1170, opened 2018-09-25
Choose single assertion library and test engine


3flex:
## Expected Behavior
Project uses a single assertion library and a single JUnit test engine.

## Current Behavior
Assertion libraries in use:
* kotlin-test
* AssertJ
* JUnit Jupiter

Both JUnit and Spek are used as test engines.

## Context
It's easier to contribute when a single assertion library and a single test engine is used across the project. Tests will complete more quickly on first run and in CI (fewer dependencies to pull) and may execute more quickly as well.

### ~~Assertions~~
~~Jupiter can be removed easily - it's only used in a single test (possibly by accident?). kotlin-test's assertions are very basic and would be easy to replace with AssertJ. It would be cumbersome to go the other way, as kotlin-test doesn't have a fluent API.~~

~~Propose using only AssertJ for assertions.~~ DONE!

### Test Engine
The project uses Spek much more than JUnit. However, Spek 2 is on the horizon (now at RC1). If migrating to Spek 2, majority of Spek tests would need to be modified to take advantage of improvements. JUnit is better supported both by maintainers and with external libraries and has better IntelliJ integration, though migrating to Spek 2 would be simpler.

Propose selecting either JUnit or Spek as the preferred test engine, and migrate tests accordingly (though if Spek is chosen, first migrate the JUnit tests to Spek 1, and handle Spek 2 migration separately if desired).

arturbosch:
As you said, I'd go with spek2 after migrating everything to spek1 from junit5.
I've started using spek1 2 years ago but had to switch to junit5 for a while as there were some bugs/breaking things in the intellij plugin.

PR #1174, opened 2018-09-25
Migrate all assertions to AssertJ

3flex:
Retires the "kotlin-test" dependency in favour of AssertJ, and gets rid of the single JUnit Jupiter assertion (also switched to AssertJ).

Addresses the assertions part of #1170.

### DIFF
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/console/BuildFailureReportSpec.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/console/BuildFailureReportSpec.kt
@@ -5,10 +5,12 @@ import io.gitlab.arturbosch.detekt.cli.TestDetektion
 import io.gitlab.arturbosch.detekt.cli.createFinding
 import io.gitlab.arturbosch.detekt.test.TestConfig
 import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.Assertions.assertThatExceptionOfType
+import org.assertj.core.api.Assertions.assertThatIllegalStateException
+import org.assertj.core.api.Assertions.fail
 import org.jetbrains.spek.api.dsl.describe
 import org.jetbrains.spek.api.dsl.it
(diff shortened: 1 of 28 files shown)


detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/console/BuildFailureReportSpec.kt
```diff
@@ -41,23 +42,24 @@ internal class BuildFailureReportSpec : SubjectSpek<BuildFailureReport>({
...
+				assertThatExceptionOfType(BuildFailure::class.java).isThrownBy { subject.render(detektion) }
 			}
 
 			it("should contain no stacktrace on fail") {
 				subject.init(TestConfig(mapOf("maxIssues" to "-2")))
-				val exception = assertFails { subject.render(detektion) }
-				assertThat(exception.stackTrace).isEmpty()
+				assertThatExceptionOfType(BuildFailure::class.java)
+						.isThrownBy { subject.render(detektion) }
+						.withStackTraceContaining("io.gitlab.arturbosch.detekt.cli.console.BuildFailure: Build failed with 1 weighted issues (threshold defined was -2).")
```
3flex: Should mention this as it's the only one I didn't see an obvious AssertJ method that allows testing that no stacktrace is shown, so the test is not exactly checking that the stacktrace is empty - `withStackTraceContaining` only checks the text exists in the stacktrace - it doesn't check that it's the only text.

Probably need a solution before merging.

Mauin: We could still match it similar to the kotlin-test way.

```kotlin
try { 
  subject.render(detektion)
  fail("Render should throw")
} catch (e: Exception) {
  assertThat(e).isInstanceOf(BuildFailure::class.java)
  assertThat(e.stackTrace).isEmpty()
}
```

3flex:
Tweaked that test - should be good to merge.

arturbosch:
+1 thanks!

PR #1177, opened 2018-09-28
Re-remove kotlin-reflect dependency

### LINKED ISSUE #1179: Master is broken
Running `./gradlew` on `master` fails with the following exception:

```
Script compilation error:

  Line 211:             kotlinTest("org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion")
                                                               ^ Unresolved reference: kotlinVersion

```

I suspect this is caused by a bad merge.

The fix is to change line `211` to: `kotlinTest(kotlin("reflect"))`

3flex:
Something funky happened when #1151 was merged - this fixes the build which is currently failing.

Fixes #1179 closes #1180


JLLeitschuh:
Can this be merged so that work on `master can continue?

PR #1180, opened 2018-09-28
Fix build failing to compile on master

### LINKED ISSUE #1179: Master is broken

JLLeitschuh:
Closes #1179


pavlospt:
There is a PR removing this dependency #1177 . I am not sure it should be added again :P

JLLeitschuh:
Probably correct, I was just trying to make it compile.

PR #1183, opened 2018-10-01
Allow importing just detekt-gradle-plugin into IDE

JLLeitschuh:
With the current project structure, for some reason, IntelliJ isn't able to generate accessors for the `detekt-gradle-plugin` project. This makes it very nearly impossible to work in this project.

One quick fix for this problem is to allow the `detekt-gradle-plugin` to be imported as a standalone project in IntelliJ. In order to do this you need a `gradle/wrapper/gradle-wrapper.properties` file in the project.

This PR adds such a file that is always automatically generated by the `wrapper` task.
Now IntelliJ can import the `detekt-gradle-plugin` standalone.

This PR also updates the gradle version to `4.10.2` which is the latest release of Gradle.


.gitignore
```diff
@@ -159,3 +159,6 @@ target/
 *.iml
 */out
 /docs/_site
+
+# Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
+!gradle-wrapper.jar
```
JLLeitschuh: This was being ignored and it shouldn't have been.

PR #1185, opened 2018-10-01
update kotlin to v1.2.71

Mauin:
Update Kotlin to v1.2.71

As the dependency on `kotlin-dsl` in the Gradle Plugin we cannot update the Kotlin version there right now. That feels kinda weird having two different Kotlin versions flying around the project.


Mauin:
The failing tests in `SuppressibleSpec` are failing due to the compiler issue that is fixed in #1187. Will wait until that is merged.

PR #1190, opened 2018-10-02
Change PathSensitivity to RELATIVE

### LINKED ISSUE #1178: Detekt should not use PathSensitivity.ABSOLUTE for Detekt task
Doing so makes the `Detekt` task impossible to use the build cache cross system.

The build cache will be invalid when you have a CI system populate the cache and a different developer tries to use it.

This is one examples of many:
https://github.com/arturbosch/detekt/blob/15e3c800a6b19f3a8ba1c69c81a3ca59d3e17bcd/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt#L68-L71

Mauin:
Fixes #1178

This changes the `PathSensitivity` of all Gradle Plugin Task inputs from `ABSOLUTE` to `RELATIVE` similar to what Checkstyle, PMD, etc. are also doing.

Additionally this adds some tests checking for correct `UP-TO-DATE` behavior as well as the `detekt` task using the Build Cache correctly.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -62,12 +62,12 @@ open class Detekt : DefaultTask() {
 
 	@InputFile
 	@Optional
-	@PathSensitive(PathSensitivity.ABSOLUTE)
+	@PathSensitive(PathSensitivity.RELATIVE)
 	var baseline: File? = null
 
 	@InputFiles
 	@Optional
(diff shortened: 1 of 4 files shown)


vanniktech:
Really nice tests @Mauin 👍

Mauin:
@vanniktech Thanks. Coverage of the Gradle Plugin related to the build cache/up-to-date checks could still be improved, and we could check when the Plugin gracefully recovers and/or errors. But it's a start.

CristianGM:
I was willing to try this, but I'll have to wait to the next release as there is no SNAPSHOT published.

And a question, isn't `@CacheableTask` task missing? ( or maybe I just didn't see it )

PR #1192, opened 2018-10-02
Revert "Revert "Update dependencies""

Mauin:
This reverts commit ef422e886ba77e268a56a68065ed7c0b50ed4280.

cc @3flex this seems to have broken the build on master so I reverted the original commit.


3flex:
There was a regression in shadow's JDK 10 support.

Bump to 4.0.1 and it should go green.

Mauin:
Looks like that did it. Build is green with that change. Thanks for the heads up @3flex.

arturbosch:
Thanks for the quick fixes!

Issue #1193, opened 2018-10-02
Implement lazy configuration for Gradle plugin


3flex:
## Expected Behavior
Gradle plugin works reliably.

## Current Behavior
Issues related to eager configuration being triggered at inconvenient times in Gradle's process.

## Steps to Reproduce (for bugs)
See #1171 #1181 #1132

## Context
Gradle's lazy configuration is most likely to address this.

However, before implementing, need to be aware that the Gradle Kotlin DSL and Groovy DSL will diverge due to https://github.com/gradle/kotlin-dsl/issues/1092

Because of this, would be a breaking change for those using detekt and configuring via the Gradle Kotlin DSL.

## Your Environment
N/A

Issue #1181, opened 2018-09-28
Gradle Plugin: If tasks are configured eagerly configuration from DetektExtension isn't used


JLLeitschuh:
A very simple example way of demonstrating this issue is to modify the `DetektTaskKotlinDslTest` to this. **ALL** of the Gradle Kotlin test will fail.

```
internal class DetektTaskKotlinDslTest : Spek({

	val buildGradle = """
		|import io.gitlab.arturbosch.detekt.detekt
		|
		|plugins {
		|   `java-library`
		|	id("io.gitlab.arturbosch.detekt")
		|}
		|
		|repositories {
		|	jcenter()
		|	mavenLocal()
		|}
		|tasks.all { } // Simply add this line
		""".trimMargin()
```

## Expected Behavior
If the `Detekt` or any other task is configured, the `DetectExtension` configuration that comes later shouldn't be ignored.

## Current Behavior
If I have a `build.gradle.kts` file like this:

```kotlin
tasks.all { }
detekt {
    // Whatever configuration I have in here will be completely ignored.
}
```

## Steps to Reproduce (for bugs)
See above.

## Context
Some Gradle plugins (externally & internally) still do goofy things like collapse the configuration of all of tasks by doing things like `tasks.all { }`.

If such plugins are added to a consumers project, all of their configuration of the `DetectExtension` is completely ignored.

## Solution

**This is going to require an API breaking change for your Gradle Plugin!!!!**

You're going to need to change all of your task inputs/outputs to use the new Gradle provider API:
https://docs.gradle.org/current/userguide/lazy_configuration.html

This will require changing all of the task inputs from things like this:
```kotlin
	@Input
	@Optional
	var filters: String? = null

	@InputFile
	@Optional
	@PathSensitive(PathSensitivity.ABSOLUTE)
	var baseline: File? = null
```

To this:
```kotlin
	@Input
	@Optional
	var filters: Property<String> = objectFactory.property<String>()

	@InputFile
	@Optional
	@PathSensitive(PathSensitivity.RELATIVE)
	var baseline: RefgularFileProperty = newInputFile()
```

I **HIGHLY** recommend that you make these changes before your `1.0.0` release. Failing to do so will require a major version bump of your gradle plugin as these will be API breaking changes!

## Your Environment
```
------------------------------------------------------------
Gradle 4.10.1
------------------------------------------------------------

Build time:   2018-09-12 11:33:27 UTC
[...]

Issue #1171, opened 2018-09-25
Gradle configuration not respecting configuration in RC9.2


marianosimone:
I'm trying to migrate to RC9.2 from RC8, and I have something like:

    apply plugin: 'io.gitlab.arturbosch.detekt'
    
    detekt {
        toolVersion = "1.0.0.RC9.2"
        config = files("whatever.yml")

        reports {
            xml.enabled = true
            html.enabled = false
        }
    }

This successfully adds the `detekt` task to my project, but no matter what I try setting up as `config`, I never get the expected behavior:

1. if `whatever.yml` doesn't exist, there should be a configuration error (there isn't)
2. if `whatever.yml` exists, the rules to use should be the ones defined in it (instead, the default config is used)

Also, no matter how I try to configure `reports`, I get both the `xml` and `html` reports

Environment:
- OS: Linux
- Gradle: 4.9
- yaml configuration: doesn't matter what I put it (I tried my own config, a minimal 1-rule configuration, and the one generated by detekt through the generation task)

Mauin:
Tried to repro this locally. In the case of an invalid path where the config does not exist I'm getting an error:

`Exception in thread "main" com.beust.jcommander.ParameterException: Provided path '/path/to/config.yml' does not exist!`

The 2nd issue I also cannot confirm on my end, I tried it with a custom config and it is being used correctly.

Can you provide more info: Are you using Windows/Linux/MacOS? Can you post your config?

mdelbo:
Is it possible that the 2nd issue is the same as the one I'm experiencing in #1132?

marianosimone:
Good point, @mdelbo! I ran it with `--info` now, and got the same as you are getting there:

`io.gitlab.arturbosch.detekt.cli.Main --input /myproject/src/main/java --report xml:/myproject/build/reports/detekt/detekt.xml --report html:/myproject/build/reports/detekt/detekt.html`

Note that not only `--config` is not there, but that the disabled `xml` report is.

3flex:
@mdelbo @marianosimone can you run with Gradle flag `-Dorg.gradle.internal.tasks.eager=true`?

I have a theory that this is due to lazy task configuration... if I'm right it's an easy fix.

JLLeitschuh:
This is very, very likely caused by the issue described in #1181

marianosimone:
[...]

PR #1194, opened 2018-10-03
[gradle-plugin] use lazy evaluation of properties

marschwar:
This PR attempts to solve #1193, #1181 and #1171 as they are all essentially the same thing.

I basically kept the extensions as is to avoid breaking changes and to keep groovy and kotlin dsl close together.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
@@ -1,5 +1,6 @@
 package io.gitlab.arturbosch.detekt
 
+import io.gitlab.arturbosch.detekt.extensions.DetektExtension
 import io.gitlab.arturbosch.detekt.invoke.BaselineArgument
 import io.gitlab.arturbosch.detekt.invoke.CliArgument
 import io.gitlab.arturbosch.detekt.invoke.ConfigArgument
@@ -12,7 +13,9 @@ import io.gitlab.arturbosch.detekt.invoke.InputArgument
 import io.gitlab.arturbosch.detekt.invoke.ParallelArgument
 import io.gitlab.arturbosch.detekt.invoke.PluginsArgument
 import org.gradle.api.DefaultTask
-import org.gradle.api.file.FileCollection
+import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.provider.Property
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFiles
 import org.gradle.api.tasks.Internal
@@ -23,11 +26,11 @@ import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.SkipWhenEmpty
 import org.gradle.api.tasks.TaskAction
 import org.gradle.language.base.plugins.LifecycleBasePlugin
-import java.io.File
 
 /**
  * @author Artur Bosch
  * @author Marvin Ramin
+ * @author Markus Schwarz
  */
 open class DetektCreateBaselineTask : DefaultTask() {
 
@@ -38,50 +41,50 @@ open class DetektCreateBaselineTask : DefaultTask() {
 
 	@OutputFile
 	@PathSensitive(PathSensitivity.RELATIVE)
-	var baseline: File? = null
+	var baseline: RegularFileProperty = project.layout.fileProperty()
 
 	@InputFiles
 	@PathSensitive(PathSensitivity.RELATIVE)
 	@SkipWhenEmpty
-	lateinit var input: FileCollection
+	var input: ConfigurableFileCollection = project.layout.configurableFiles()
 
 	@Input
 	@Optional
-	var filters: String? = null
+	var filters: Property<String> = project.objects.property(String::class.java)
 
 	@InputFiles
 	@Optional
 	@PathSensitive(PathSensitivity.RELATIVE)
-	var config: FileCollection? = null
+	var config: ConfigurableFileCollection = project.layout.configurableFiles()
 
 	@Input
 	@Optional
-	var plugins: String? = null
+	var plugins: Property<String> = project.objects.property(String::class.java)
 
 	@Internal
 	@Optional
-	var debugOrDefault: Boolean = false
+	var debug: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
 
 	@Internal
 	@Optional
-	var parallelOrDefault: Boolean = false
+	var parallel: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
 
 	@Internal
 	@Optional
-	var disableDefaultRuleSetsOrDefault: Boolean = false
-
+	var disableDefaultRuleSets: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
 
 	@TaskAction
 	fun baseline() {
+		val debugOrDefault = debug.getOrElse(DetektExtension.DEFAULT_DEBUG_VALUE)
 		val arguments = mutableListOf<CliArgument>(CreateBaselineArgument) +
-				BaselineArgument(baseline) +
+				BaselineArgument(baseline.get()) +
 				InputArgument(input) +
-				FiltersArgument(filters) +
+				FiltersArgument(filters.orNull) +
 				ConfigArgument(config) +
-				PluginsArgument(plugins) +
+				PluginsArgument(plugins.orNull) +
 				DebugArgument(debugOrDefault) +
-				ParallelArgument(parallelOrDefault) +
-				DisableDefaultRulesetArgument(disableDefaultRuleSetsOrDefault)
+				ParallelArgument(parallel.getOrElse(DetektExtension.DEFAULT_PARALLEL_VALUE)) +
+				DisableDefaultRulesetArgument(disableDefaultRuleSets.getOrElse(DetektExtension.DEFAULT_DISABLE_RULESETS_VALUE))
 
 		DetektInvoker.invokeCli(project, arguments.toList(), debugOrDefault)
 	}
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
@@ -25,6 +25,7 @@ import org.gradle.api.tasks.PathSensitive
 import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.SkipWhenEmpty
 import org.gradle.api.tasks.TaskAction
+import org.gradle.kotlin.dsl.property
 import org.gradle.language.base.plugins.LifecycleBasePlugin
 
 /**
@@ -50,7 +51,7 @@ open class DetektCreateBaselineTask : DefaultTask() {
 
 	@Input
 	@Optional
-	var filters: Property<String> = project.objects.property(String::class.java)
+	var filters: Property<String> = project.objects.property()
 
 	@InputFiles
 	@Optional
@@ -59,7 +60,7 @@ open class DetektCreateBaselineTask : DefaultTask() {
 
 	@Input
 	@Optional
-	var plugins: Property<String> = project.objects.property(String::class.java)
+	var plugins: Property<String> = project.objects.property()
 
 	@Internal
 	@Optional
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
@@ -1,6 +1,5 @@
 package io.gitlab.arturbosch.detekt
 
-import io.gitlab.arturbosch.detekt.extensions.DetektExtension
 import io.gitlab.arturbosch.detekt.invoke.BaselineArgument
 import io.gitlab.arturbosch.detekt.invoke.CliArgument
 import io.gitlab.arturbosch.detekt.invoke.ConfigArgument
@@ -76,17 +75,16 @@ open class DetektCreateBaselineTask : DefaultTask() {
 
 	@TaskAction
 	fun baseline() {
-		val debugOrDefault = debug.getOrElse(DetektExtension.DEFAULT_DEBUG_VALUE)
 		val arguments = mutableListOf<CliArgument>(CreateBaselineArgument) +
 				BaselineArgument(baseline.get()) +
 				InputArgument(input) +
 				FiltersArgument(filters.orNull) +
 				ConfigArgument(config) +
 				PluginsArgument(plugins.orNull) +
-				DebugArgument(debugOrDefault) +
-				ParallelArgument(parallel.getOrElse(DetektExtension.DEFAULT_PARALLEL_VALUE)) +
-				DisableDefaultRulesetArgument(disableDefaultRuleSets.getOrElse(DetektExtension.DEFAULT_DISABLE_RULESETS_VALUE))
+				DebugArgument(debug.get()) +
+				ParallelArgument(parallel.get()) +
+				DisableDefaultRulesetArgument(disableDefaultRuleSets.get())
 
-		DetektInvoker.invokeCli(project, arguments.toList(), debugOrDefault)
+		DetektInvoker.invokeCli(project, arguments.toList(), debug.get())
 	}
 }
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektIdeaInspectionTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektIdeaInspectionTask.kt
@@ -1,6 +1,5 @@
 package io.gitlab.arturbosch.detekt
 
-import io.gitlab.arturbosch.detekt.extensions.DetektExtension
 import io.gitlab.arturbosch.detekt.extensions.IdeaExtension
 import io.gitlab.arturbosch.detekt.invoke.ProcessExecutor
 import org.gradle.api.DefaultTask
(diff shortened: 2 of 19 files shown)


docs/pages/gettingstarted/groovydsl.md
```diff
@@ -123,12 +123,8 @@ task detektFailFast(type: io.gitlab.arturbosch.detekt.Detekt) {
...
    config = files("$rootDir/config.yml")
    debug = true
-   reports {
-       xml {
-           destination = file("build/reports/failfast.xml")
-       }
-       html.destination = file("build/reports/failfast.html")
-   }
+   xmlReportFile = file("build/reports/failfast.xml")       // if omitted no xml report is generated
+   htmlReportFile = file("build/reports/failfast.html")     // if omitted no html report is generated
```
3flex: This is a bit sad, after all the work that was done to align the config DSL with other code quality plugins.

There must be a way to have the existing DSL work.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/TestVersion.kt
```diff
@@ -3,4 +3,4 @@ package io.gitlab.arturbosch.detekt
 /**
  * @author Artur Bosch
  */
-const val VERSION_UNDER_TEST: String = "1.0.0-gradle-rework-beta3"
+const val VERSION_UNDER_TEST: String = "1.0.0.RC9.2"
```
3flex: Note to anyone who wants to fix this: would be best to test the plugin against the latest version in the checkout, rather than pulling it from upstream.

First, build detekt-cli, then reference that using the `detekt` configuration in Gradle in the test by referencing the detekt-cli build path.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DslBaseTest.kt
```diff
@@ -1,19 +1,35 @@
...
+					.map(VARIABLE_NAME_SOURCE::get)
+					.joinToString("")
 
-class DslBaseTest(private val buildGradleFileName: String, private val buildGradleFile: String) {
 
 	private fun getBuildFileContent(customConfig: String): String {
 		return """
     	|$buildGradleFile
+    	|
+    	|tasks.all {}
```
3flex: Should add `tasks.all {}` to the Kotlin DSL tests too just to be sure

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -15,11 +17,18 @@ import java.io.File
  * @author Markus Schwarz
  */
 open class DetektExtension(project: Project) : CodeQualityExtension() {
-	val defaultReportsDir: File = project.layout.buildDirectory.get()
+
+	val customReportsDir: File?
+		get() = reportsDir
+
+	val detektVersion: String?
+		get() = toolVersion
```
3flex: toolVersion comes from CodeQualityExtension - no reason to have an additional getter with a different name on the class?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -28,22 +37,30 @@ open class DetektExtension(project: Project) : CodeQualityExtension() {
...
+	var config: ConfigurableFileCollection = project.layout.configurableFiles()
 
-	var debug: Boolean = DEFAULT_DEBUG_VALUE
+	var debug: Boolean? = null
 
-	var parallel: Boolean = DEFAULT_PARALLEL_VALUE
+	var parallel: Boolean? = null
 
-	var disableDefaultRuleSets: Boolean = DEFAULT_DISABLE_RULESETS_VALUE
+	var disableDefaultRuleSets: Boolean? = null
```
3flex: Why change type to nullable and default to null? Aren't the existing defaults suitable?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
```diff
@@ -38,50 +41,50 @@ open class DetektCreateBaselineTask : DefaultTask() {
...
 	@InputFiles
 	@Optional
 	@PathSensitive(PathSensitivity.RELATIVE)
-	var config: FileCollection? = null
+	var config: ConfigurableFileCollection = project.layout.configurableFiles()
 
 	@Input
 	@Optional
-	var plugins: String? = null
+	var plugins: Property<String> = project.objects.property(String::class.java)
```
3flex: Since Kotlin DSL is applied to the project, can simplify to `project.objects.property()` when creating a Property

marschwar:
> Though I'm wondering why the extension has to be modified at all, if only some of the parameters are wired up as Providers? I'd have thought either all could remain as is, or all would have to be updated.

The changes to the extension are all non breaking and cosmetic in nature.

docs/pages/gettingstarted/groovydsl.md
marschwar: I feel you. The general dsl to configure the plugin is still very close to the code qualtity plugins. This change is only necessary when declaring a custom task. But I will fiddle around with it and see if I can make it work.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DslBaseTest.kt
marschwar: Both kotlin and groovy dsl tests share this helper function so that is the case already.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
marschwar: With the toolVersion beeing a platform type the intention was to make sure the compiler knows it can be `null`. The alternative is to have to remember that it can in fact be `null` and add `@Suppress("USELESS_ELVIS")`everywhere you use it. Can you think of a better way to do this?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -93,8 +88,7 @@ class DetektPlugin : Plugin<Project> {
 			isTransitive = true
 			description = "The $CONFIGURATION_DETEKT dependencies to be used for this project."
 
-			@Suppress("USELESS_ELVIS")
-			val version = extension.toolVersion ?: DEFAULT_DETEKT_VERSION
+			val version = extension.detektVersion ?: DEFAULT_DETEKT_VERSION
 			defaultDependencies {
 				add(project.dependencies.create("io.gitlab.arturbosch.detekt:detekt-cli:$version"))
 			}
```
3flex: This should be changed to:
```kotlin
defaultDependencies {
	val version = extension.toolVersion ?: DEFAULT_DETEKT_VERSION
	add(project.dependencies.create("io.gitlab.arturbosch.detekt:detekt-cli:$version"))
}
```

You can add this test to `DetektTaskKotlinDslTest`, which fails before the change, but passes afterwards:
```kotlin
it("allows configuring tool dependencies explicitly") {
	val detektConfig = """
			|detekt {
			|	toolVersion = "1.0.0.RC8"
			|}
		"""

	dslTest.writeFiles(rootDir, detektConfig)
	dslTest.writeConfig(rootDir)

	// Using a custom "project-cache-dir" to avoid a Gradle error on Windows
	val result = GradleRunner.create()
			.withProjectDir(rootDir)
			.withArguments("--project-cache-dir", createTempDir(prefix = "cache").absolutePath, "dependencies", "--configuration", "detekt")
			.withPluginClasspath()
			.build()

	assertThat(result.task(":dependencies")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
	assertThat(result.output).contains("io.gitlab.arturbosch.detekt:detekt-cli:1.0.0.RC8")
}
```

3flex: This should be changed to:
```kotlin
defaultDependencies {
	val version = extension.toolVersion ?: DEFAULT_DETEKT_VERSION
	add(project.dependencies.create("io.gitlab.arturbosch.detekt:detekt-cli:$version"))
}
```

You can add this test to `DetektTaskKotlinDslTest`, which fails before the change, but passes afterwards:
```kotlin
it("allows configuring tool dependencies explicitly") {
	val detektConfig = """
			|detekt {
			|	toolVersion = "1.0.0.RC8"
			|}
		"""

	dslTest.writeFiles(rootDir, detektConfig)
	dslTest.writeConfig(rootDir)

	// Using a custom "project-cache-dir" to avoid a Gradle error on Windows
	val result = GradleRunner.create()
			.withProjectDir(rootDir)
			.withArguments("--project-cache-dir", createTempDir(prefix = "cache").absolutePath, "dependencies", "--configuration", "detekt")
			.withPluginClasspath()
			.build()

	assertThat(result.task(":dependencies")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
	assertThat(result.output).contains("io.gitlab.arturbosch.detekt:detekt-cli:1.0.0.RC8")
}
```

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
3flex: Can also use the same technique on the Boolean properties.

docs/pages/gettingstarted/groovydsl.md
3flex: Try having `Detekt` class implement `Reporting`: https://docs.gradle.org/current/dsl/org.gradle.api.reporting.Reporting.html

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskKotlinDslTest.kt
```diff
@@ -90,9 +91,8 @@ internal class DetektTaskKotlinDslTest : Spek({
 					.withPluginClasspath()
 					.build()
 
-			assertThat(result.output).contains("number of classes: 1")
-			assertThat(result.output).contains("Ruleset: comments")
 			assertThat(result.task(":check")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
+			assertThat(result.output).contains("io.gitlab.arturbosch.detekt/detekt-cli/$customVersion")
```
3flex: This is causing the test failure - because it's checking for a path, and Windows uses backslashes.

If you use the test I suggested it will effectively check the same thing and work cross-platform.

marschwar: I guess it was late when I read your suggested test case. I like it a lot better because it just tests if the configuration is being picked up.

marschwar: The way these DSL tests are written I think has gotten out of hand. Do you agree?
I would like to clean them up and remove some of that boilerplate code. Should that be done in a separate PR?

Mauin: I think a separate PR makes sense for this.

marschwar:
Build server seems to have hick ups. Is there a way to rerun the build?

```
* What went wrong:
Error resolving plugin [id: 'com.gradle.plugin-publish', version: '0.10.0']
> Could not resolve all dependencies for configuration 'detachedConfiguration1'.
   > Could not determine artifacts for com.gradle.plugin-publish:com.gradle.plugin-publish.gradle.plugin:0.10.0
      > Could not get resource 'https://plugins.gradle.org/m2/com/gradle/plugin-publish/com.gradle.plugin-publish.gradle.plugin/0.10.0/com.gradle.plugin-publish.gradle.plugin-0.10.0.jar'.
         > Could not HEAD 'https://plugins.gradle.org/m2/com/gradle/plugin-publish/com.gradle.plugin-publish.gradle.plugin/0.10.0/com.gradle.plugin-publish.gradle.plugin-0.10.0.jar'.
            > Read timed out
```

marschwar:
Thanks for the trick.

With _current DSL_ you mean the one in RC9? I think it is possible to keep it without using the Reporting interface. 
Just out of curiosity, what are reasons why people declare a custom task? Do they need different detekt configs for local and CI?

3flex:
Yep, that's definitely one use case that's come up before (used to be that a custom profile was setup but it didn't jive well with Gradle conventions).

By "current DSL" I just mean allowing this (like other code quality plugins):
```
   reports {
       xml {
           destination = file("build/reports/failfast.xml")
       }
       html.destination = file("build/reports/failfast.html")
   }
```

docs/pages/gettingstarted/groovydsl.md
arturbosch: Why can't we use the dsl way here? They are not evaluated lazily?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
arturbosch: The idea-inspection plugin does not use gradle internal package by leveraging the CheckstyleReports: https://github.com/JetBrains/inspection-plugin/blob/master/plugin/src/main/kotlin/org/jetbrains/intellij/tasks/InspectionsTask.kt
Maybe we can have something similar to work for us?

docs/pages/gettingstarted/groovydsl.md
arturbosch: The idea-inspection plugin does not use gradle internal package by leveraging the CheckstyleReports: https://github.com/JetBrains/inspection-plugin/blob/master/plugin/src/main/kotlin/org/jetbrains/intellij/tasks/InspectionsTask.kt
Maybe we can have something similar to work for us?

marschwar: The problem here is the structure of the extension is different to the task. @3flex, should I change it back to the way it was now or wait until you have evaluated the Reporting interface?

marschwar: The problem with the Jetbrains approach is that they are also using internal classes in the [IdeaCheckstyleReports](https://github.com/JetBrains/inspection-plugin/blob/master/plugin/src/main/kotlin/org/jetbrains/intellij/IdeaCheckstyleReports.kt) to implement the ReportingContainer interface. If we decide that is no proplem that would then make things a lot easier.

3flex: Created an issue here: https://github.com/gradle/gradle/issues/7063

Not sure where this leaves things... doesn't make sense to hold up the PR just for this but the constant config DSL changes are disruptive.

Mauin: Thanks for opening this issue. I agree however that the breaking DSL changes are not that great.
Let's see what reply we get to the issue and make a decision based on that.

marschwar: Give me until tomorrow. I already started to make Sone changes to get the original DSL working again

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektIdeaInspectionTask.kt
```diff
@@ -27,20 +28,20 @@ open class DetektIdeaInspectionTask : DefaultTask() {
 	@InputFiles
 	@PathSensitive(PathSensitivity.RELATIVE)
 	@SkipWhenEmpty
-	lateinit var input: FileCollection
+	var input: ConfigurableFileCollection = project.layout.configurableFiles()
 
 	@Internal
 	@Optional
-	var debugOrDefault: Boolean = false
+	var debug: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
```
3flex: Should be possible to change to `project.objects.property()` for boolean types as well.

marschwar: When you define a custom task (in the kotlin dsl) you then have to use `debug.set(true)` instead of `debug = true` or you get `The boolean literal does not conform to the expected type Property<Boolean>`.

3flex:
Anything holding this back?

Would be great to get this merged, then #1221 could be revised and I'll pick up the resolver work on top of that branch.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
```diff
@@ -39,22 +39,23 @@ internal data class PluginsArgument(val plugins: String?) : CliArgument() {
...
-internal data class BaselineArgument(val baseline: File?) : CliArgument() {
-	override fun toArgument() = baseline?.let { listOf(BASELINE_PARAMETER, it.absolutePath) } ?: emptyList()
+internal data class BaselineArgument(val baseline: RegularFile?) : CliArgument() {
+	override fun toArgument() = baseline?.let { listOf(BASELINE_PARAMETER, it.asFile.absolutePath) } ?: emptyList()
 }
 
-internal data class XmlReportArgument(val file: File?) : CliArgument() {
-	override fun toArgument() = file?.let { listOf(REPORT_PARAMETER, "xml:${it.absolutePath}") } ?: emptyList()
+internal data class XmlReportArgument(val file: RegularFile?) : CliArgument() {
+	override fun toArgument() = file?.let { listOf(REPORT_PARAMETER, "xml:${it.asFile.absoluteFile}") } ?: emptyList()
```
Mauin: This `XmlReportArgument` is using `it.asFile.absoluteFILE` and other arguments are using `.absolutePATH` but also have a `RegularFile` argument. Can't we use the same in all cases?

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskKotlinDslTest.kt
```diff
@@ -164,9 +160,10 @@ internal class DetektTaskKotlinDslTest : Spek({
 					|
 					|	input = files("src/main/java")
 					|	config = files("${configFile.safeAbsolutePath}")
-					|	debug = true
+					|	debug.set(true)
```
Mauin: Does this still work with `debug = true`? There is a comment further up which mentions that the property in the Plugin is more verbose than it could be to enable the direct assignment?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
marschwar: You are absolutely right. Event though the outcome is the same, it should be `absolutePath` for consistency.-

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskKotlinDslTest.kt
marschwar: Good point. Turns out it does not.

PR #1196, opened 2018-10-03
Do not detect parameter name violations in overridden function

### LINKED ISSUE #1175: FunctionParameterNaming false positive
Following code is flagged with RC9.2:

```kotlin
override fun isViewFromObject(view: View, `object`: Any) = view === `object`
```

This should not be flagged since `object` is the name from the parent method and a Kotlin keyword hence the ` are required.

pawegio:
Fixes #1175

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionParameterNaming.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionParameterNaming.kt
@@ -9,6 +9,7 @@ import io.gitlab.arturbosch.detekt.api.LazyRegex
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.rules.identifierName
+import io.gitlab.arturbosch.detekt.rules.isOverridden
 import io.gitlab.arturbosch.detekt.rules.naming.util.isContainingExcludedClass
 import org.jetbrains.kotlin.psi.KtParameter
 
@@ -33,6 +34,10 @@ class FunctionParameterNaming(config: Config = Config.empty) : Rule(config) {
 
 	override fun visitParameter(parameter: KtParameter) {
 		if (parameter.isContainingExcludedClass(excludeClassPattern)) {
+			return
+		}
+
+		if (parameter.ownerFunction?.isOverridden() == true) {
 			return
 		}
 
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ParameterNamingSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ParameterNamingSpec.kt
@@ -55,6 +55,17 @@ class ParameterNamingSpec : Spek({
 			assertThat(findings).isEmpty()
 		}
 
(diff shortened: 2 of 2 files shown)


pawegio:
I can add a config `ignoreOverriddenFunctions` by default set to `true`. 
@vanniktech What do you think about it?

vanniktech:
If it's on by default, sure. But I'd wait for an opinion from others too.

arturbosch:
As @vanniktech said, the option on by default would be great.

Mauin:
@pawegio Please run a `./gradlew build` on this change. That should automatically change `detault-detekt-config.yml`. Please also commit that change so that the default config contains the new parameter already.

pawegio:
@Mauin Done, thanks.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionParameterNaming.kt
```diff
@@ -36,6 +37,10 @@ class FunctionParameterNaming(config: Config = Config.empty) : Rule(config) {
 			return
 		}
 
+		if (parameter.ownerFunction?.isOverridden() == true) {
```
marschwar: `== true` is redundant

pavlospt: @marschwar it's a nullable result, are you sure `== true` is redundant. IIRC removing `== true` will not let you compile.

pawegio: otherwise expression is nullable

marschwar: you are right. Sorry.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ParameterNamingSpec.kt
```diff
@@ -55,6 +56,29 @@ class ParameterNamingSpec : Spek({
 			assertThat(findings).isEmpty()
 		}
 
+		it("should not detect violations in overridden function by default") {
+			val findings = FunctionParameterNaming().lint(
+					"""
+					class C {
```
schalkms: `"""` formatting looks weird here

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ParameterNamingSpec.kt
```diff
@@ -55,6 +56,29 @@ class ParameterNamingSpec : Spek({
...
+				"""
+			)
+			assertThat(findings).isEmpty()
+		}
+
+		it("should detect violations in overridden function if ignoreOverriddenFunctions is false") {
+			val config = TestConfig(mapOf("ignoreOverriddenFunctions" to "false"))
+			val findings = FunctionParameterNaming(config).lint(
+					"""
+					class C {
```
schalkms: `"""` formatting looks weird here

pawegio: Totally agree, but I simply adapted to existed formatting in this spec. Anyway, I've just reformatted it.

PR #1202, opened 2018-10-07
clarify error message in verify documentation task

### LINKED ISSUE #1199: Documentation verification requires committing changes but that's not clear
I don't think the `verifyGeneratorOutput` should depend on having committed the documentation. Just chatted with @Mauin and he agrees :)

Mauin:
Fixes #1199 

It would be possible to copy all existing documentation, generate the up-to-date documentation, run a diff, then delete the copied documentation to verify this even better.

But that felt quite verbose for a fix that takes just a commit locally. Let me know what you think @rock3r


rock3r:
Better than it was before for sure :) Thanks @mauin

Issue #1158, opened 2018-09-24
In Travis CI: "0 kotlin files were analyzed."


3flex:
## Expected Behavior
detekt Gradle plugin analyzes files during CI.

## Current Behavior
detekt Gradle plugin does not analyze any files during CI.

## Steps to Reproduce (for bugs)
View the Travis build log for master branch (https://travis-ci.org/arturbosch/detekt/jobs/432263468 at time of writing). See "0 kotlin files were analyzed." multiple times in the logs.

## Context
N/A

## Your Environment
N/A - build running in CI environment

3flex:
On closer inspection, looks like this is on Travis only, not AppVeyor... may be a *nix vs Windows issue with detekt itself?

Mauin:
Just to clarify: This does not just affect `master` builds but pretty much every build on travis it seems. I cannot repro this locally.

3flex:
Simple when you know what to look for...

The filters set when Travis build runs is include `.*/build/*.`. Travis checks out the build at `/home/travis/build/arturbosch/detekt`. Since `/build/` is in the full file path, all Kotlin files under the project path are filtered out, so nothing gets scanned.

I'd say the root cause is that though the detekt is run from working directory `/home/travis/build/arturbosch/detekt`, and expectation is that it's filtering based on the relative path in that working directory it's actually filtering the absolute file path.

Mauin:
Oh. Right... You want to submit a fix for this? 
Probably simply changing it to `detekt/.*/build/.*` or similar

3flex:
Can do for Travis, yeah. But I'm also going to open a new issue to see if behaviour should change - I'm surprised that it's not just the relative path where the filter applies.

Mauin:
True, that would be the overall better fix for it. Which makes me wonder. Did _we_ change this behavior at some point? (absolute vs relative path considered for excluding paths)

Or did Travis change the directory projects get cloned into...?

3flex:
Do you want to close this in favour of a proper fix for #1212, or implement a workaround?

I'm inclined to close this since root cause is identified and an issue opened to track the issue. AppVeyor appears to run detekt just fine as part of the build so any non-compliant code will still fail CI.

Mauin:
👍 Thanks for opening the concrete issue. We should definitely fix this properly.

PR #1204, opened 2018-10-07
Revert "Re-add additional empty check for report paths"

Mauin:
This reverts commit 4c0df7cd41b7e237064d5363159108d0e8b70323.

AppVeyor is currently red. This is the first commit where it went red with the error:
`Error: Could not find or load main class io.gitlab.arturbosch.detekt.cli.Main`

Hoping this fixes it.

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ReportPath.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ReportPath.kt
@@ -8,10 +8,8 @@ import java.nio.file.Paths
 
 /**
  * @author Artur Bosch
- * @author Pavlos-Petros Tournaris
  */
 data class ReportPath(val kind: String, val path: Path) {
-
 	companion object {
 		private const val NUM_OF_PARTS_UNIX = 2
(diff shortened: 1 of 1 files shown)


3flex:
@arturbosch why was #1124 was merged when CI was failing? This error was happening on the most recent commit in that PR. Would have been best to address the problem before merging to master. Similar thing happened with #1151 merge. [When master was merged into my branch](https://github.com/arturbosch/detekt/pull/1151/commits/4f20c427d61f01e42bb81399c905fe72422a5d86) before merging back to master it added a line to `build.gradle.kts` which broke the build. Would have been better to merge master, see the build was failing, then not merge the PR back into the master branch.

Anyway looks like the actual issue on AppVeyor is that `detekt-api:detekt` job is running before detekt-cli itself has been built. I expect a proper fix will be to ensure any `detekt` tasks run after `detekt-cli:assemble` (not checked locally so sorry if this doesn't work).

Travis seems to be using a build cache which might explain #1158 and the fact the build result is different in each CI environment. Will look into this.

3flex:
Should merge #1207 in place of this.

Mauin:
As feared, this was not the issue. Closing this and looking for a real fix.

PR #1207, opened 2018-10-08
Build: build detekt-cli before running detekt tasks

### LINKED ISSUE #1208: Project dependency to override detekt-cli in composite build does not run assemble automatically
After the recent merges the build is broken on `master` (on AppVeyor) with the following error:

```
[--input, C:\projects\detekt\detekt-api\src\main\kotlin, --filters, .*/resources/.*,.*/build/.*, --config, C:\projects\detekt\detekt-cli\src\main\resources\default-detekt-config.yml,C:\projects\detekt\reports\failfast.yml, --baseline, C:\projects\detekt\reports\baseline.xml, --report, xml:C:\projects\detekt\detekt-api\build\reports\detekt\detekt.xml, --report, html:C:\projects\detekt\detekt-api\build\reports\detekt\detekt.html, --debug]
C:\projects\detekt\detekt-cli\build\libs\detekt-cli-1.0.0.RC9.2.jar
C:\projects\detekt\detekt-core\build\libs\detekt-core-1.0.0.RC9.2.jar
C:\projects\detekt\detekt-api\build\libs\detekt-api-1.0.0.RC9.2.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-compiler-embeddable\1.2.71\a79f934bfbc1c7e16638da3474feab351735f1c0\kotlin-compiler-embeddable-1.2.71.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-reflect\1.2.71\7512db3b3182753bd2e48ce8d345abbadc40fe6b\kotlin-reflect-1.2.71.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib\1.2.71\d9717625bb3c731561251f8dd2c67a1011d6764c\kotlin-stdlib-1.2.71.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\com.beust\jcommander\1.74\6b4c0f6e034ee6ce26a72e786cffe6a9c78815d1\jcommander-1.74.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-common\1.2.71\ba18ca1aa0e40eb6f1865b324af2f4cbb691c1ec\kotlin-stdlib-common-1.2.71.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.jetbrains\annotations\13.0\919f0dfe192fb4e063e7dacadee7f8bb9a2672a9\annotations-13.0.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-script-runtime\1.2.71\d5439926359caa899f41e1915b19d6d23e49750b\kotlin-script-runtime-1.2.71.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.yaml\snakeyaml\1.23\ec62d74fe50689c28c0ff5b35d3aebcaa8b5be68\snakeyaml-1.23.jar
Error: Could not find or load main class io.gitlab.arturbosch.detekt.cli.Main
```

As you can see from this log the `detekt-cli.jar` is on the classpath, however the `detekt` task cannot find or load the main class.

3flex:
detekt tasks need the latest CLI to be built before running so the CLI classes are available on the classpath.

Fixes #1208

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,6 +4,7 @@ import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 import org.gradle.api.tasks.testing.logging.TestLogEvent
 import org.gradle.util.GFileUtils
 import org.jetbrains.dokka.gradle.DokkaTask
+import io.gitlab.arturbosch.detekt.Detekt
 
 import java.util.Date
 
@@ -36,6 +37,10 @@ tasks.withType<Wrapper> {
 
 tasks.withType<Test> {
 	dependsOn(gradle.includedBuild("detekt-gradle-plugin").task(":test"))
+}
+
+tasks.withType<Detekt> {
+	dependsOn("detekt-cli:assemble")
 }
 
 val detektVersion: String by project


Mauin:
Also looking at the AppVeyor output for the failure:

```
[--input, C:\projects\detekt\detekt-api\src\main\kotlin, --filters, .*/resources/.*,.*/build/.*, --config, C:\projects\detekt\detekt-cli\src\main\resources\default-detekt-config.yml,C:\projects\detekt\reports\failfast.yml, --baseline, C:\projects\detekt\reports\baseline.xml, --report, xml:C:\projects\detekt\detekt-api\build\reports\detekt\detekt.xml, --report, html:C:\projects\detekt\detekt-api\build\reports\detekt\detekt.html, --debug]
C:\projects\detekt\detekt-cli\build\libs\detekt-cli-1.0.0.RC9.2.jar
C:\projects\detekt\detekt-core\build\libs\detekt-core-1.0.0.RC9.2.jar
C:\projects\detekt\detekt-api\build\libs\detekt-api-1.0.0.RC9.2.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-compiler-embeddable\1.2.71\a79f934bfbc1c7e16638da3474feab351735f1c0\kotlin-compiler-embeddable-1.2.71.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-reflect\1.2.71\7512db3b3182753bd2e48ce8d345abbadc40fe6b\kotlin-reflect-1.2.71.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib\1.2.71\d9717625bb3c731561251f8dd2c67a1011d6764c\kotlin-stdlib-1.2.71.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\com.beust\jcommander\1.74\6b4c0f6e034ee6ce26a72e786cffe6a9c78815d1\jcommander-1.74.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-common\1.2.71\ba18ca1aa0e40eb6f1865b324af2f4cbb691c1ec\kotlin-stdlib-common-1.2.71.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.jetbrains\annotations\13.0\919f0dfe192fb4e063e7dacadee7f8bb9a2672a9\annotations-13.0.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-script-runtime\1.2.71\d5439926359caa899f41e1915b19d6d23e49750b\kotlin-script-runtime-1.2.71.jar
C:\Users\appveyor\.gradle\caches\modules-2\files-2.1\org.yaml\snakeyaml\1.23\ec62d74fe50689c28c0ff5b35d3aebcaa8b5be68\snakeyaml-1.23.jar
Error: Could not find or load main class io.gitlab.arturbosch.detekt.cli.Main
```

We do print everything that's on the classpath and `detekt-cli-1.0.0.RC9.2.jar` is present.

Mauin:
Added more findings to #1208. While this fixes the build I'm still hesitant to merge this change. The Gradle Plugin tasks should be able to run without any build steps in the detekt repo.

3flex:
This is the root cause: https://github.com/arturbosch/detekt/pull/1124#discussion_r222044871

Doesn't affect every project using detekt, only this one, because the detekt-cli dependency for the plugin is provided by the checked out version. Obviously in that case it has to be built first, so this PR adds that explicit dependency.

For anyone else using detekt the detekt-cli dependency is pulled from an external repo by default so they're not affected and wouldn't have to make this change.

Mauin:
After some digging I figured this out. But still feel weird about this change. By defining the `detekt(project(detekt-cli))` dependency I feel like the build task should already have been added to the task dependencies. 

Also: The complexity of the whole setup regarding the Gradle Plugin and the interactions between it and the rest of the modules should really be documented in a `README.md` or somewhere. Looking for hints and documentation in multiple PRs/Issues should not be the way to go.

3flex:
Composite builds are a learning curve for us all. But they're the sanest way to develop the Gradle plugin alongside the rest of the project.

I think the key is not to merge changes to the build scrips if they're not understood, especially if they break the build.

Can you list the questions you have about the build as it stands? I'll create a PR to add comments where needed.

Mauin:
Don't get me wrong. I'm on the one hand really happy with the setup of the Gradle Plug-in and the composite build and the things it allows us to do. However the added complexity also makes it harder to debug issues like this one. 
This creates a high barrier of entry for new contributors to the Gradle Plug-in. So having some additional documentation can help lower that barrier. I will create an issue to track this. 

What I'm wondering as for this change. 
`implementation(project(name))` now behaves differently (it seems to add task dependencies to execute the build of "name") than `detekt(project(name))`.

I'm wondering if and how we can achieve the same behavior to simply build in the task dependency.

3flex:
I suspect it's a limitation of the composite build... Gradle docs say the same as you do when adding a project dependency like this, and it's what I'd expected to happy originally:
> A “lib” dependency is a special form of an execution dependency. It causes the other project to be built first and adds the jar with the classes of the other project to the classpath

That doesn't appear to be happening. It may be a Gradle bug since this is probably an edge case.

Mauin:
Looking into this deeper I will now merge this workaround and have opened https://github.com/gradle/gradle/issues/7075

We should hopefully find a way to remove the manual task dependency in the future. But for now I want to get `master` green again.

PR #1213, opened 2018-10-09
Migrate JUnit tests to Spek

### LINKED ISSUE #1170: Choose single assertion library and test engine

3flex:
Closes #1170

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -215,7 +215,6 @@ subprojects {
 		}
 	}
 
-	val junitEngineVersion: String by project
 	val assertjVersion: String by project
 	val spekVersion: String by project
 	val kotlinTest by configurations.creating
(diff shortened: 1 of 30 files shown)


pavlospt:
Just my 2 cents on this: I believe that the contribute documentation should include info on how someone could run Spek tests and debug Spek tests. For people that do not use a plugin or have their first interaction with Spek, it feels a bit difficult to get around on testing their contribution (speaking from personal experience 😂 )

3flex:
> Do we also need the Runner so that the tests work when executing ./gradlew test?

Majority of tests are already run with Spek test engine, and they execute fine with ./gradlew test. This PR just migrates the few tests that were using the JUnit engine to Spek.

Mauin:
@pavlospt Do you want to create a small issue to point out the improvements that could be made to the Documentation around Spek? Otherwise it might get lost in this PR thread.

PR #1218, opened 2018-10-09
[WIP] AppVeyor: Save Gradle cache to the AppVeyor build cache

### LINKED ISSUE #1217: AppVeyor fails downloading dependencies
This now happens quite frequently, especially as we're running a build on 3 different JDK versions. 

Is it possible to cache dependencies or tune the build in some similar manner to make the AppVeyor build more stable and thus more meaningful?

3flex:
Note that this won't save build cache in PRs by default, [according to AppVeyor docs](https://www.appveyor.com/docs/build-cache/#saveupdate-cache-before-build-finishes):
> Note: By default, saving cache is disabled in Pull Request builds. Use Save build cache in Pull Requests checkbox on the General tab of project settings if you need to save cache during PR builds. This setting is not exposed in YAML to prevent unauthorized cache modifications.

This means the build log for this PR will not show the build cache being saved.

The lock file is deleted in `after_test` since that seemed most logical. [See here for other options](https://www.appveyor.com/docs/build-configuration/#build-pipeline).

Closes #1217

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,6 +15,7 @@ environment:
 
 init:
   - git config --global --unset core.autocrlf
+  - set GRADLE_OPTS=-Dorg.gradle.daemon=false
 
 install:
   - SET PATH=%JAVA_HOME%\bin;%PATH%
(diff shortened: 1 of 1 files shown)


3flex:
Results:
```
gradlew --stop
Stopping Daemon(s)
1 Daemon stopped
del C:\Users\appveyor\.gradle\caches\modules-2\modules-2.lock
C:\Users\appveyor\.gradle\caches\modules-2\modules-2.lock
The process cannot access the file because it is being used by another process.
```

This doesn't make sense to me... What other process, other than Gradle, could be using the lock file? This means the cache will be uploaded unnecessarily.

Mauin:
On Travis we don't even run `./gradlew --stop`.

We can try 1 & 2 just to see if it makes the build more stable. Otherwise 3 sounds good to me. Debugging 4 (especially with full access to AppVeyor) sounds unrealistic.

pavlospt:
We generally run Gradle on CI with daemon disabled. Will save you from many trouble.

Mauin:
Merging this to get more green builds on PRs

Issue #1198, opened 2018-10-05
Gradle plugin: create task per source set


3flex:
## Expected Behavior
Plugin creates detekt task for each source set in a Gradle project.

## Current Behavior
Plugin creates single detekt task per Gradle project.

## Context
Required for #485 (which I will update and integrate with the Gradle plugin).

To enable type and symbol resolution when detekt runs as part of the Gradle plugin, detekt must be aware of the compilation classpath. If detekt can't access the complete classpath that's used for compilation it's not possible to resolve all types and symbols unless classpaths are setup manually on each detekt task which isn't practical for users and highly error prone.

Additionally the detekt task for a source set should by default scan the source files in the sourceset. The current default for detekt is to scan `src/main/java` and `src/main/kotlin`. This will fail for tasks which are created for other sourcesets (e.g. "test") since the source files for those would typically be a combination of `src/main/java`, `src/test/java`, `src/main/kotlin` and `src/test/kotlin`.

This also affects how reports will be configured, because they will necessarily have to be created per task. If the report filename is set using the extension (i.e. `detekt.reports.xml.destination =`) a report generated for one detekt task would be overwritten by the next detekt task in the same project because the destination file path would be the same. Instead, the report directory must be set on the extension and the destination path on the task.

It should be possible to minimise disruption by creating source set tasks like "detektMain" and "detektTest" which use the new behaviour for reporting DSL and keep the current "detekt" task as is but with a deprecation message.

Then suggest that prior to 1.0 release, update the "detekt" task so it depends on all configured detekt tasks (detektMain, detektTest etc) and remove the reporting config from the extension (with exception of reportsDir provided by CodeQualityExtension).

Mauin:
This seems similar to #540 which describes the same. Overall I agree with adding a task per sourceSet (while keeping one "global" `detekt` task). 

For the reports issue we could either suffix the report name with the sourceSet name or merge instead of overwrite the reports if multiple exist. But that would probably be quite tricky.

marschwar:
Using a suffix or change the report directory for each task could easily be done. Does that imply the same config is used for all such tasks? 

From reading #485 I did not quite get what is expected to be passed to the cli. Is it `--input src/main/java;src/main/kotlin` as well as the `--classpath`? Must the classpath contain all the compile time dependencies of the project?

Mauin:
I was looking into this earlier today and can upload a first snapshot of
the changes tomorrow.

PR #1221, opened 2018-10-10
Generate one detekt task per source set and pass classpath to CLI

Mauin:
Wanted to do this as a 2nd stage to #1220 but GitHub doesn't allow me to base the PRs on top of each other.

So here are both changes combined:
- Creating one detekt task per sourceset
- Add the `--classpath` argument to the CLI
- Path classpath to CLI for each sourceset task (not the default `detekt` task at the moment)

This is missing tests, which I will add once we're happy with the changes overall (marking the PR as blocked because of this)

#1198

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
@@ -23,6 +23,14 @@ class CliArgs : Args {
 					"These filters apply on relative paths from the project root.")
 	var filters: String? = null // Using a converter for List<PathFilter> resulted in a ClassCastException
 
+	// This is currently unused and was added for experimenting with type resolution.
+	@Suppress("detekt:UnusedPrivateMember")
+	@Parameter(names = ["--classpath", "-cp"],
+			required = false,
+			hidden = true,
+			description = "Compile Classpath of the project.")
+	private var classpath: String? = null
+
 	@Parameter(names = ["--config", "-c"],
 			description = "Path to the config file (path/to/config.yml). " +
 					"Multiple configuration files can be specified with ',' or ';' as separator.")
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -2,6 +2,7 @@ package io.gitlab.arturbosch.detekt
 
 import io.gitlab.arturbosch.detekt.extensions.DetektReports
 import io.gitlab.arturbosch.detekt.invoke.BaselineArgument
+import io.gitlab.arturbosch.detekt.invoke.ClasspathArgument
 import io.gitlab.arturbosch.detekt.invoke.CliArgument
 import io.gitlab.arturbosch.detekt.invoke.ConfigArgument
 import io.gitlab.arturbosch.detekt.invoke.DebugArgument
@@ -23,6 +24,7 @@ import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
 import org.gradle.api.reporting.ReportingExtension
 import org.gradle.api.tasks.CacheableTask
+import org.gradle.api.tasks.CompileClasspath
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFile
 import org.gradle.api.tasks.InputFiles
@@ -48,6 +50,10 @@ open class Detekt : DefaultTask() {
 	@PathSensitive(PathSensitivity.RELATIVE)
 	@SkipWhenEmpty
 	var input: ConfigurableFileCollection = project.layout.configurableFiles()
+
+	@CompileClasspath
+	@Optional
+	var classpath: ConfigurableFileCollection = project.layout.configurableFiles()
 
 	@Input
 	@Optional
@@ -124,6 +130,7 @@ open class Detekt : DefaultTask() {
 	fun check() {
 		val arguments = mutableListOf<CliArgument>() +
 				InputArgument(input) +
+				ClasspathArgument(classpath) +
 				FiltersArgument(filters.orNull) +
 				ConfigArgument(config) +
 				PluginsArgument(plugins.orNull) +
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -1,9 +1,12 @@
 package io.gitlab.arturbosch.detekt
 
 import io.gitlab.arturbosch.detekt.extensions.DetektExtension
+import org.gradle.api.Named
+import org.gradle.api.NamedDomainObjectCollection
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.file.FileCollection
+import org.gradle.api.file.SourceDirectorySet
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.SourceSetContainer
 import org.gradle.api.tasks.TaskProvider
@@ -28,7 +31,9 @@ class DetektPlugin : Plugin<Project> {
 	}
 
 	private fun createDetektTasks(project: Project, extension: DetektExtension) {
-		project.sourceSets?.map { sourceSet ->
+		val sourceSets = mutableSetOf<String>()
+		project.sourceSets?.forEach { sourceSet ->
+			sourceSets += sourceSet.name
 			val name = "$DETEKT${sourceSet.name.capitalize()}"
 			val description = "Runs detekt on the ${sourceSet.name} source set."
 			val inputProvider = project.provider { sourceSet.allSource.sourceDirectories.filter { it.exists() } }
@@ -42,6 +47,15 @@ class DetektPlugin : Plugin<Project> {
 			project.tasks.findByName(LifecycleBasePlugin.CHECK_TASK_NAME)?.dependsOn(sourceSetTask)
 		}
 
+		// TODO kotlin-multiplatform ios source sets are currently not considered
+		setOf("kotlin", "kotlin-multiplatform", "kotlin-android", "kotlin2js")
+				.forEach { pluginId ->
+					project.plugins.withId(pluginId) {
+						project.applyKotlinSourceSetDetektTasks(extension, sourceSets)
+					}
+				}
+
+
 		val detektTask =
 				createAndConfigureDetektTask(project,
 						extension,
@@ -49,6 +63,28 @@ class DetektPlugin : Plugin<Project> {
 						"Runs the default detekt task.",
 						existingInputDirectoriesProvider(project, extension))
 		project.tasks.findByName(LifecycleBasePlugin.CHECK_TASK_NAME)?.dependsOn(detektTask)
+	}
+
+	private fun Project.applyKotlinSourceSetDetektTasks(extension: DetektExtension, existingSourceSets: MutableSet<String>) {
+		project.kotlinSourceSets
+				.filter { !existingSourceSets.contains(it.name) }
+				.forEach { sourceSet ->
+					existingSourceSets += sourceSet.name
+					val name = "$DETEKT${sourceSet.name.capitalize()}"
+					val description = "Runs detekt on the kotlin ${sourceSet.name} source set."
+					val sourceDirectorySet = getKotlinSourceDirectorySetSafe(sourceSet)
+					sourceDirectorySet?.let {
+						val inputProvider = project.provider { it.sourceDirectories.filter { it.exists() } }
+						// TODO How to provide Classpath for KotlinSourceSet?
+						val sourceSetTask =
+								createAndConfigureDetektTask(project,
+										extension,
+										name,
+										description,
+										inputProvider)
+						project.tasks.findByName(LifecycleBasePlugin.CHECK_TASK_NAME)?.dependsOn(sourceSetTask)
+					}
+				}
 	}
 
 	private fun createAndConfigureDetektTask(project: Project,
@@ -131,6 +167,20 @@ class DetektPlugin : Plugin<Project> {
 
 	private val Project.sourceSets: SourceSetContainer?
 		get() = project.extensions.findByType(SourceSetContainer::class.java)
+
+	private val Project.kotlinSourceSets: NamedDomainObjectCollection<out Named>
+		get() {
+			// Access through reflection, because another project's KotlinProjectExtension might be loaded by a different class loader:
+			val kotlinExt = project.extensions.getByName("kotlin")
+			@Suppress("UNCHECKED_CAST")
+			val sourceSets = kotlinExt.javaClass.getMethod("getSourceSets").invoke(kotlinExt) as NamedDomainObjectCollection<out Named>
+			return sourceSets
+		}
+
+	private fun getKotlinSourceDirectorySetSafe(from: Any): SourceDirectorySet? {
+		val getKotlin = from.javaClass.getMethod("getKotlin")
+		return getKotlin(from) as? SourceDirectorySet
+	}
 
 	companion object {
 		private const val DETEKT = "detekt"
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -7,6 +7,8 @@ import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.file.FileCollection
 import org.gradle.api.file.SourceDirectorySet
+import org.gradle.api.plugins.JavaPluginConvention
+import org.gradle.api.plugins.JavaPluginExtension
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.SourceSetContainer
 import org.gradle.api.tasks.TaskProvider
@@ -51,6 +53,7 @@ class DetektPlugin : Plugin<Project> {
 		setOf("kotlin", "kotlin-multiplatform", "kotlin-android", "kotlin2js")
 				.forEach { pluginId ->
 					project.plugins.withId(pluginId) {
+						println("$pluginId exists")
 						project.applyKotlinSourceSetDetektTasks(extension, sourceSets)
 					}
 				}
@@ -69,6 +72,7 @@ class DetektPlugin : Plugin<Project> {
 		project.kotlinSourceSets
 				.filter { !existingSourceSets.contains(it.name) }
 				.forEach { sourceSet ->
+					println(sourceSet.name)
 					existingSourceSets += sourceSet.name
 					val name = "$DETEKT${sourceSet.name.capitalize()}"
 					val description = "Runs detekt on the kotlin ${sourceSet.name} source set."
@@ -106,7 +110,6 @@ class DetektPlugin : Plugin<Project> {
 			it.input.setFrom(project.provider { inputSources })
 			it.classpath.setFrom(project.provider { compileClasspath })
 			it.reportsDir.set(project.provider { extension.customReportsDir })
-			// TODO this does not set the report name correctly
 			it.reports = extension.reports.apply {
 				xml.setReportName(name)
 				html.setReportName(name)
@@ -163,6 +166,7 @@ class DetektPlugin : Plugin<Project> {
 			configuration.defaultDependencies { dependencySet ->
 				@Suppress("USELESS_ELVIS")
 				val version = extension.toolVersion ?: DEFAULT_DETEKT_VERSION
+				println("Using detekt $version")
 				dependencySet.add(project.dependencies.create("io.gitlab.arturbosch.detekt:detekt-cli:$version"))
 			}
 		}
@@ -170,6 +174,7 @@ class DetektPlugin : Plugin<Project> {
 
 	private val Project.sourceSets: SourceSetContainer?
 		get() = project.extensions.findByType(SourceSetContainer::class.java)
+
 
 	private val Project.kotlinSourceSets: NamedDomainObjectCollection<out Named>
 		get() {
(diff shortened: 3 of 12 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -101,6 +91,8 @@ class DetektPlugin : Plugin<Project> {
 		}
 	}
 
+	private val Project.sourceSets: SourceSetContainer? get() = project.getProperties()["sourceSets"] as? SourceSetContainer
```
marschwar: I thought the idiomatic way to find the source set is
```
private val Project.sourceSets: SourceSetContainer? get() = project.convention.findPlugin(JavaPluginConvention::class)?.sourceSets
```

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -114,4 +116,34 @@ open class Detekt : DefaultTask() {
...
 	}
+
+	companion object {
+		fun create(project: Project, 
+					extension: DetektExtension, 
+					name: String, 
+					taskDescription: String,
+					inputSources: FileCollection): TaskProvider<Detekt> {
+			return project.tasks.register(name, Detekt::class.java) {
+				description = taskDescription
```
marschwar: I don't really like the idea of a factory method on the task that as a side effect registers the task to the project. I probably would have kept a `private fun registerDetektTask(project: Project, extension: DetektExtension, sourceSet: SourceSetContainer)`.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -114,4 +116,34 @@ open class Detekt : DefaultTask() {
...
+				plugins = extension.plugins
+				input = inputSources
+				extension.reports.forEach { extReport ->
+					reports.withName(extReport.name) {
+						enabled = extReport.enabled
+						val fileSuffix = extReport.name
+						@Suppress("USELESS_ELVIS")
+						val reportsDir = extension.reportsDir ?: extension.defaultReportsDir
+						val customDestination = extReport.destination
+						destination = customDestination ?: File(reportsDir, "$name.$fileSuffix")
```
marschwar: Keeping the customDestination will result in report files being overwritten.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -114,4 +116,34 @@ open class Detekt : DefaultTask() {
...
+				parallel = extension.parallel
+				disableDefaultRuleSets = extension.disableDefaultRuleSets
+				filters = extension.filters
+				config = extension.config
+				baseline = extension.baseline
+				plugins = extension.plugins
+				input = inputSources
+				extension.reports.forEach { extReport ->
+					reports.withName(extReport.name) {
+						enabled = extReport.enabled
```
marschwar: With #1194 this structure is gone. We could add a suffix or source set marker to the `DetektReports`

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
marschwar: Also this would now create `xml.main'. I prefer `detekt.main.xml`

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
3flex: Yes - don't want all source sourcesets.

Also Gradle plugins are moving away from conventions - should be able to replace `project.convention.findPlugin(JavaPluginConvention::class.java)?.sourceSets` with `project.extensions.findByType(typeOf<SourceSetContainer>())`

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
marschwar: Ah, ok. That does make sense.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -16,35 +20,21 @@ class DetektPlugin : Plugin<Project> {
...
-					val reportsDir = extension.reportsDir ?: extension.defaultReportsDir
-					val customDestination = extReport.destination
-					destination = customDestination ?: File(reportsDir, "$DETEKT.$fileSuffix")
-				}
-			}
+	private fun createDetektTasks(project: Project, extension: DetektExtension) {
+		project.sourceSets?.map { sourceSet ->
+			val name = "$DETEKT${sourceSet.name.capitalize()}"
+			val description = "Runs detekt on the ${sourceSet.name} source set."
+			Detekt.create(project, extension, name, description, sourceSet.allSource)
```
marschwar: `sourceSet.allSource` will be expanded to a list of absolute file paths which technically is perfect because it respects the includes and excludes from the project build. But it will become very long very quickly. 

> On computers running Microsoft Windows XP or later, the maximum length of the string that you can use at the command prompt is 8191 characters. On computers running Microsoft Windows 2000 or Windows NT 4.0, the maximum length of the string that you can use at the command prompt is 2047 characters.
https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation

Is it possible to invoke the `Runner` directly, bypassing the command line?

All source directories can be obtained by `sourceSet.allSource.srcDirs.filter(File::exists)`

Mauin: Looking though some other internal Gradle code quality plugins they seem to use `getAllJava()` without transforming them to directories. But I see the issue this might cause.

marschwar: Those plugins use an ant runner instead of the command line. So they don't run into this restriction.

marschwar:
With regards to the overwriting of report files I think we are in the clear if the internal code quality plugins are the standard. I just played around with checkstyle and pmd and both blindly use whatever is configured in `destination`.
```
tasks.withType(Pmd) {
	reports {
		xml.destination = "build/custom-pmd.xml"
	}
}
```

I just now realized that both do not support the reports block in the extension itself. Do we need to do the same thing? I actually like our dsl better here.

Even though I assume this is no news to you: After ` gradle clean check` I find 1 file per source set and report.
```
./build/reports/checkstyle/test.xml
./build/reports/checkstyle/main.xml
./build/reports/checkstyle/test.html
./build/reports/checkstyle/main.html
./build/reports/pmd/test.xml
./build/reports/pmd/main.xml
./build/reports/pmd/test.html
./build/reports/pmd/main.html
```
That all seems rather easy to accomplish.

3flex:
> both do not support the reports block in the extension itself

Enabling/disabling report types is safe on the extension (though as you said the core plugins don't allow this). But setting the destination is not safe when more than one task runs in a project since reports would be overwritten. I'm guessing this is why the core plugins don't allow doing that on the extension but only on tasks.

I think ideally you would allow enabling/disabling reports on the extension (but not the destination) and allowing enabling/disabling reports, and configuring destination, on tasks. The task setting will override extension setting for the enabling/disabling case.

That way there's still nice defaults and it's easy to enable or disable report types for all tasks at once, but doesn't cause issues with overwritten reports.

marschwar:
The core plugins do have exactly that problem of reports being overwritten if `tasks.withType(Pmd) {}` is used. One would have to define a custom task in order to get this to work correctly. But I do get your point that it probably makes no sense to provide the `destination` property on the extension if that leads to reports being overwritten every time.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -56,6 +59,10 @@ open class Detekt : DefaultTask() {
 	@SkipWhenEmpty
 	lateinit var input: FileCollection
 
+	@Internal
```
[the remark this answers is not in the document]
3flex: Sorry should have said that the other annotation must replace `@Internal`. Leaving `@Internal`
> will cause the task not to be considered out-of-date when the property has changed

Which is not desired.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -22,6 +22,12 @@ class CliArgs : Args {
 			description = "Path filters defined through regex with separator ';' or ',' (\".*test.*\").")
 	var filters: String? = null // Using a converter for List<PathFilter> resulted in a ClassCastException
 
+	// This is currently unused and was added for experimenting with type resolution. 
+	@Parameter(names = ["--classpath", "-cp"],
+			required = false,
+			description = "Compile Classpath of the project.")
```
3flex: Should add `hidden = true` to the parameter annotation for now. This will hide the parameter when viewing usage.

3flex: Should add `hidden = true` to the annotation. This will [hide the parameter from usage](http://jcommander.org/#_hiding_parameters) (preferred while it's experimental).

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -56,6 +60,11 @@ open class Detekt : DefaultTask() {
 	@SkipWhenEmpty
 	lateinit var input: FileCollection
 
+	@CompileClasspath
+	@Internal
```
3flex: Sorry I should have said that `@Internal` should be replaced with the new annotation, rather than adding to it. Leaving it there
> will cause the task _not_ to be considered out-of-date when the property has changed

Which isn't what we want.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
```diff
@@ -23,6 +24,13 @@ internal object CreateBaselineArgument : CliArgument() {
 	override fun toArgument() = listOf(CREATE_BASELINE_PARAMETER)
 }
 
+internal data class ClasspathArgument(val fileCollection: FileCollection) : CliArgument() {
+	override fun toArgument(): List<String> {
+		if (fileCollection.isEmpty()) return emptyList()
+		return listOf(CLASSPATH_PARAMETER, fileCollection.joinToString(",") { it.absolutePath })
```
3flex: I'm wary of the comma as path separator - to save cross-platform issues in future detekt should really follow the Java standard for path separators.

Safest would be to use `java.io.File.pathSeparatorChar` which is designed for this purpose. [The value is set dynamically](https://docs.oracle.com/javase/9/docs/api/java/io/File.html#pathSeparatorChar) based on platform (to `:` on *nix and `;` on Windows).
```suggestion
		return listOf(CLASSPATH_PARAMETER, fileCollection.joinToString(java.io.File.pathSeparatorChar) { it.absolutePath })	
```

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
arturbosch: And/Or throw an exception if it is used.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -56,6 +60,11 @@ open class Detekt : DefaultTask() {
...
+	@Optional
+	lateinit var classpath: FileCollection
```
arturbosch: This should then be also a property like in the blocker pr?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
arturbosch: Sounds good to me, however would't this require double documentation/examples as the user must enter the right separator?

3flex: This should be transparent to the user who will either not have to specify a classpath (the plugin should be able to figure it out on its own) and if they do specify manually it would be as a file collection, so no path separator necessary.

Even though it's internal though using the appropriate path separators should minimise ongoing issues with cross-platform support.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
3flex: Better not throw an exception - either hide it, or just don't introduce it yet.

It's hidden on the CLI output (with `hidden = true`) and the comment in the code is clear that it should be unused. No risk anyone uses it accidentally.

3flex:
Have played with this, and it looks like it's generating tasks for different sourcesets for JVM targets, but not for other target types.

This makes sense since it's checking for `SourceSetContainer`s which holds a `SourceSet` collection, and a [`SourceSet` ](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/SourceSet.html) "represents a logical group of Java source and resources."

Instead might have to somehow enumerate [`KotlinSourceSet`](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin-api/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinSourceSet.kt)... or possibly enumerate all the generated KotlinCompile, Kotlin2JsCompile and KotlinCompileCommon tasks and create a corresponding detekt task for each one?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
Mauin: It now picks up only the parent directories for the source set as suggested.

Mauin:
Rebased this PR and applied some of the suggestions in the reviews. Thanks for that.

Still TODO:
- Tests for the different tasks and the way reports behave
- Look into the SourceSet issue @3flex described (any ideas?)

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektReport.kt
```diff
@@ -30,7 +33,7 @@ class DetektReport(val type: DetektReportType, private val project: Project) {
 		if (customDestination != null)
```
marschwar: As discussed we should no longer allow a custom destination in the extensions.

arturbosch: Ok, by me. I think I need to check this out and run it to fully understand what @3flex said :)

arturbosch:
I think I need to check this out and run it to fully understand what @3flex said :)

3flex:
https://github.com/JetBrains/kotlinconf-app is good to test with - it's MPP with common modules and Android, JVM and iOS targets. I expect to see a detekt task created for the common, Android and JVM modules (iOS is Swift so detekt can't help there). I think without modification this PR only works for JVM targets.

For a first cut though it might be best to merge just with JVM support only if it's too hard getting it to work with other targets. This would be OK as long as detekt can be configured somehow to work for non-JVM targets.

Mauin:
Just to update this: I'm still working on trying to get the KotlinSourceSets enumerated so that we can generate tasks for each of those source sets as well.

3flex:
> trying to get the KotlinSourceSets enumerated

Things changed a bunch for MPP in Kotlin 1.3 so hopefully you're combing through that version of the Gradle plugin... if not, then I'd recommend switching to that to avoid future rework!

This of course means we could only support Kotlin 1.3 and higher in detekt's Gradle plugin. Hopefully that's not a showstopper.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektReport.kt
Mauin: Looking into this a bit more I think we currently have three options with source set tasks and reports.

- Remove support for custom destinations (breaking change)
- Use the "source set name" as a subdirectory which we inject into the custom destination (`/path/to/report.xml` becomes `/path/to/sourceSetName/report.xml`)
- Somehow change the custom destination filename to include both the requested filename and the sourceSet name to prevent overwriting files.

Mauin:
Pushed some new changes.
- The plugin should now generate tasks for Kotlin Multiplatform, JS and other source sets as well. cc @3flex 
- Added a composite build into the Gradle Plugin tests to be able to test local changes of the CLI module in the Gradle Plugin Tests.
- As suggested by @marschwar the default `detekt` task does not overwrite reports, but source set tasks do.

All existing Plugin tests are passing for me locally. Some of the new tests are not yet passing as something seems to still not work with creating custom source sets and then generating tasks for them.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -18,28 +25,96 @@ class DetektPlugin : Plugin<Project> {
...
+				.filter { !existingSourceSets.contains(it.name) }
+				.forEach { sourceSet ->
+					println(sourceSet.name)
+					existingSourceSets += sourceSet.name
+					val name = "$DETEKT${sourceSet.name.capitalize()}"
+					val description = "Runs detekt on the kotlin ${sourceSet.name} source set."
+					val sourceDirectorySet = getKotlinSourceDirectorySetSafe(sourceSet)
+					sourceDirectorySet?.let {
+						val inputProvider = project.provider { it.sourceDirectories.filter { it.exists() } }
+						// TODO How to provide Classpath for KotlinSourceSet?
```
Mauin: @3flex KotlinSourceSets seem to currently not publish their Classpath: https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin-api/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinSourceSet.kt

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskDslTest.kt
```diff
@@ -265,6 +265,206 @@ internal class DetektTaskDslTest : Spek({
...
+
+					gradleRunner.writeKtFile("src/main/java", "SomeTestClass")
+					gradleRunner.writeKtFile("src/test/java", "SomeTestClass")
+
+					gradleRunner.runTasksAndCheckResult("detektMain", "detektTest") { result ->
+						assertThat(result.task(":detektMain")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
+						assertThat(result.task(":detektTest")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
+					}
+				}
+				it("respects custom java source sets") {
```
Mauin: This test is failing as the "custom" sourceSet is not visible in the Plugin

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DslGradleRunner.kt
```diff
@@ -21,6 +21,14 @@ class DslGradleRunner(
     	|rootProject.name = "rootDir-project"
 		|include(${projectLayout.submodules.map { "\"${it.name}\"" }.joinToString(",")})
 		|
+		|// Include original detekt dependencies as composite build
+		|includeBuild("${System.getProperty("user.dir")}/../") {
```
Mauin: The `"user.dir"` property feels a bit hacky here and will probably break if the tests are run from some other directory. Happy for suggestions on how to get the "original" directory of the gradle plugin from the tests. 

Currently they create and use a temporary project dir.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektReport.kt
arturbosch: Hm option fourth could confuse the user.
Actually a gradle user would expect to get reports inside the `build` folder, wouldn't he? Junit, jacoco and the other quality plugins do it.
This would be the easiest thing to do and would align with the expectation of gradle users.
Like pmd/checkstyle there will be a `build/pmd/main.xml` files etc.
A task to aggregate all exports can be easily written on top of this by us/or the user?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
3flex: I think the most reliable way to get the classpath is going to be to get it from the compilation task generated by the Kotlin Gradle plugin, by enumerating based on `tasks.withType<AbstractKotlinCompile>` or similar then getting the value of `compileClasspath` from the task that's created: https://github.com/JetBrains/kotlin/blob/a0399d054168e4fbedce3db57166fed1c82b4330/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/tasks/Tasks.kt#L156-L159

I'm not sure how to actually do all that from this plugin though.

3flex:
@mauin do you still plan to work on this?

I'm going to submit a PR hopefully within a week that will enable type and symbol resolving in the CLI.

Your PR will be essential for integrating that feature with Gradle.

If you're not planning to continue I'm prepared to take it on by building on the great work you've already done, but if you're going to finish it I'd be happy to leave it to you.

arturbosch:
@3flex we should also consider to move away from starting a process per detekt run on a module and soon a sourceSet. Whole ~500ms overhead here ^^

Issue #1113, opened 2018-09-14
CliArgs Documentation improvements examples


vanniktech:
The documentation for https://github.com/arturbosch/detekt/blob/master/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt isn't really that great.

Especially for the `--report`  argument as it talks about `report-id` but it's not defined what types of reports are supported.

3flex:
Is this fixed with #1225? If not what else needs updating?

vanniktech:
Yup. For some reason issues like this never get closed when the PR is merged.

PR #1225, opened 2018-10-10
CliArgs Doc Change

alistairsykes:
Issue: #1113. The documentation around the --report option didn't specify the available 'report-id''s. 

- Updated to include available report-id's
- Remove the stray '+' within the comment.
- Changed "path-to-store-report" to "path", since path has already been mention "and stores it on given 'path'". Therefore should be consistent for clarity.

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
@@ -57,7 +57,7 @@ class CliArgs : Args {
 	@Parameter(names = ["--report", "-r"],
 			description = "Generates a report for given 'report-id' and stores it on given 'path'. " +
 					"Entry should consist of: [report-id:path]. " +
-		   			"Available 'report-id' values: 'plain', 'xml', 'html' )
+		   			"Available 'report-id' values: 'txt', 'xml', 'html' )
 	private var reports: List<String>? = null
 
 	@Parameter(names = ["--disable-default-rulesets", "-dd"],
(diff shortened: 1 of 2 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -56,7 +56,8 @@ class CliArgs : Args {
 
 	@Parameter(names = ["--report", "-r"],
 			description = "Generates a report for given 'report-id' and stores it on given 'path'. " +
-					"Entry should consist of: [report-id:path-to-store-report]+")
+					"Entry should consist of: [report-id:path]. " +
+		   			"Available 'report-id' values: 'plain', 'xml', 'html' )
```
Mauin: The `plain` report has actually recently been renamed to `txt` 

https://github.com/arturbosch/detekt/blob/master/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/TxtOutputReport.kt#L11

alistairsykes: Fixed. Sorry I didn't spot this change

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -56,7 +56,8 @@ class CliArgs : Args {
 
 	@Parameter(names = ["--report", "-r"],
 			description = "Generates a report for given 'report-id' and stores it on given 'path'. " +
-					"Entry should consist of: [report-id:path-to-store-report]+")
+					"Entry should consist of: [report-id:path]. " +
+		   			"Available 'report-id' values: 'txt', 'xml', 'html' )
```
Mauin: This is missing an ending `"` which currently breaks the build.

PR #1234, opened 2018-10-11
fix test-pattern to support windows path separator

robstoll:
The current default value does not work on Windows machines or rather on file systems which use `\` as file separator.

Another possibility would be -- and that would certainly be nicer-- if one can define `/` and it works on Windows as well.

### DIFF
--- a/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/ConfigPrinter.kt
+++ b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/ConfigPrinter.kt
@@ -73,7 +73,7 @@ object ConfigPrinter : DocumentationPrinter<List<RuleSetPage>> {
 			test-pattern: # Configure exclusions for test sources
 			  active: true
 			  patterns: # Test file regexes
-			    - '.*/test/.*'
+			    - '.*(/|\\)test(/|\\).*'
 			    - '.*Test.kt'
 			    - '.*Spec.kt'
 			  exclude-rule-sets:
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -4,7 +4,7 @@ failFast: false
 test-pattern: # Configure exclusions for test sources
   active: true
   patterns: # Test file regexes
-    - '.*/test/.*'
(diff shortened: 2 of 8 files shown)


robstoll:
Build fails with:
```
> The default-detekt-config.yml is not up-to-date. Please build detekt locally to update it.
```
What is necessary next to the yml file?

Mauin:
The `default-detekt-config.yml` is generated from the sources in the `detekt-rules` module.

The parameters outside of the rule-sets themselves are currently hardcoded here: https://github.com/arturbosch/detekt/blob/master/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/ConfigPrinter.kt#L76

robstoll:
The build should have been green (JDK8 and JDK10 were fine, just JDK9 failed due to bintray problems). I cannot re-trigger jobs, so I forced pushed again. Please re-trigger single jobs in case one fails again due to bintray

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/ConfigPrinter.kt
```diff
@@ -73,7 +73,7 @@ object ConfigPrinter : DocumentationPrinter<List<RuleSetPage>> {
 			test-pattern: # Configure exclusions for test sources
 			  active: true
 			  patterns: # Test file regexes
-			    - '.*/test/.*'
+			    - '.*(/|\\)test(/|\\).*'
```
Mauin: I'm wondering, could we do a change in `detekt-core` to get either `/` or `\` but we interpret them OS independent and they would work for both cases?

robstoll: I am sure we could. The regex is used somewhere and matched against paths. So we need to bring both into a `/` from.
However, I suggest we proceed with baby steps. This pull request includes a regression and a fix. We can still improve from here.

robstoll:
AppVeyor failed again. I think it is not a problem of my change,  JDK9 and JDK10 are green. Please retrigger JDK8 or just merge the pull request. Thanks

arturbosch:
Thank you for the contribution! I will close this as @Mauin fixes the path handling in detekt itself.

PR #1246, opened 2018-10-17
Removed `MaximumLineLength`

PaulWoitaschek:
I wondered why changing this dind't do anything. Apparently it's called `MaxLineLength` and already present.

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -214,9 +214,6 @@ formatting:
     autoCorrect: true
     indentSize: 4
     continuationIndentSize: 4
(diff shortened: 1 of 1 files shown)


Mauin:
Thanks you for the PR!

`MaximumLineLength` is actually one of the rules wrapped from `ktlint` (as all rules in the `formatting` ruleset) and requires it to be set up as an additional dependency as described here: https://arturbosch.github.io/detekt/formatting.html#chainwrapping. 
`MaxLineLength` is essentially the same rule but implemented directly in detekt.

I will close this PR as both rule definitions in the `config.yml` are actually correct. I do however agree that the documentation on these "duplicated" rules and the formatting module has to be improved. Do you want to open an issue for this?

PR #1250, opened 2018-10-17
Change --filters argument to only check relative paths

### LINKED ISSUE #1212: Filters apply to absolute paths not relative paths
## Expected Behavior
Filters apply only to relative paths.

## Current Behavior
Filters apply to absolute paths.

Identified investigating #1158 (relevant build log: https://travis-ci.org/arturbosch/detekt/jobs/432263468)

## Context
The filters set when Travis build runs includes `.*/build/*.`. Travis checks out the build at `/home/travis/build/arturbosch/detekt` which is also detekt's working directory when it runs. I'd expected that it's filtering based on the relative path, but it appears to be filtering based on the absolute path. Since all Kotlin files have `/build/` in their absolute paths, everything is filtered.

## Your Environment
* Version used: master
* Operating System and version: Ubunty Trusty 14.04 (running on Travis)
* Link to your project: See #1158 and build log: https://travis-ci.org/arturbosch/detekt/jobs/432263468

Mauin:
Fixes #1212 and should therefore also allow Travis to again check files during the CI build.

This change will make the `--filters` argument only operate on the relative path to from the current directory.
This would be:
 - Possibly wherever the user currently invokes the CLI from manually
 - The root of the project/module if using the Gradle plugin

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt
@@ -15,7 +15,7 @@ fun CliArgs.createPathFilters(): List<PathFilter> = filters.letIfNonEmpty {
 			.asSequence()
 			.map { it.trim() }
 			.filter { it.isNotEmpty() }
-			.map(::PathFilter)
+			.map { filter -> PathFilter(filter) }
 			.toList()
 }
 
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
@@ -19,7 +19,8 @@ class CliArgs : Args {
 	private var input: String? = null
 
 	@Parameter(names = ["--filters", "-f"],
-			description = "Path filters defined through regex with separator ';' or ',' (\".*test.*\").")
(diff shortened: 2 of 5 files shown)


Mauin:
Just checked the green Travis build and with this change it actually analyzes files again.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt
```diff
@@ -15,7 +15,7 @@ fun CliArgs.createPathFilters(): List<PathFilter> = filters.letIfNonEmpty {
 			.asSequence()
 			.map { it.trim() }
 			.filter { it.isNotEmpty() }
-			.map(::PathFilter)
+			.map { filter -> PathFilter(filter) }
```
arturbosch: Should't the root here now be the `input` parameter?

arturbosch: Paths.get("").toAbsolutePath() could be different than the input parameter?
btw didn't know that Paths.get("") worked :)

Mauin: It could be. As mentioned in the PR description. The root in this case will be wherever you're invoking the CLI from. Alternatively we can pass a `--root` argument to the CLI. But then we start adding more and more obscure arguments only needed for special cases.

PR #1254, opened 2018-10-18
Travis: oraclejdk10 > openjdk10

### LINKED ISSUE #1251: Issue with Travis build on JDK 10
There is currently an issue on the Travis CI side which leads to the JDK 10 build to fail for reasons unrelated to detekt

See: https://travis-ci.community/t/help-with-oraclejdk10-build-on-linux/398

3flex:
Closes #1251.

Oracle :heart: license agreements. If you try and download from the URL that Travis is trying to use, you get this error:

image

Travis may come up with a solution but in case they don't it's simple enough to switch from Oracle's to OpenJDK's build for JDK 10.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: java
 jdk:
   - oraclejdk8
   - oraclejdk9
-  - oraclejdk10
+  - openjdk10
 script:
   - ./gradlew build -PwarningsAsErrors=true
   - ./gradlew shadowJar


arturbosch:
Thanks!

Mauin:
What do you think of just going for open in all three builds?

arturbosch:
Yep, lets go with everything open!

Issue #637, opened 2017-12-22
Config file paths should be recognized cross-platform


sewerk:
Detekt check should be able to recognize correct path to ignore from config file, when using of different operating systems.
Config file was created for unix:
```
test-pattern:
  active: true
  patterns:
    - '.*/test/.*'
    - '.*/androidTest/.*'
    - '.*Test.kt'
    - '.*Spec.kt'
```
But when run on Windows its reporting syntax errors for paths like `app\src\test\....`
How to configure detekt so it can be used across team with different OSs?

Mauin:
It should be possible to add both:
```
patterns:
    - '.*/test/.*'
    - '.*\test\.*'
```
I'm wondering if we should do this in detekt internally, adding a Windows config for every Unix config and the other way around.

sewerk:
Thanks, small correction, for Windows we need to escape the backslash
```
patterns:
    - '.*\\test\\.*'
```

One argument in favor - seems that other tools like Lint supports having single path config

PR #1256, opened 2018-10-18
Use PathFilter in TestPattern to make patterns OS independent

Mauin:
`PathFilter` already handles Windows vs. Unix paths. By using `PathFilter` in `TestPattern` we should be able to also apply the test patterns on both OS.

This PR currently contains all the changes from #1250 as well. Only `TestPattern` and `TestPatternTest` were changed for this PR. I'll rebase once #1250 is merged.

This should fix the issue described in #1234 in detekt itself instead of relying on the regex to be OS independent.


arturbosch:
This should also fix #637 which `was` the last open issue for 1.0 :)

PR #1267, opened 2018-10-19
Remove duplicate configurations for reports.

marschwar:
Right before #1194 was merged @JLLeitschuh commented that the functions that let the user configure reports can be expressed by a single function using an `Action`.


JLLeitschuh:
I commented a bunch of other things too but this was a big one. Hahah. I'm glad you caught that feedback.
Nice job @marschwar!

PR #1268, opened 2018-10-20
Consider variable to be this prefixed - Closes #1257

### LINKED ISSUE #1257: VarCouldBeVal false positive
I'm still trying to look into it, but initially, this minimal example shows a false positive:

```
fun createObject() = object {
	private var myVar: String? = null
	fun assign(value: String?) {
		this.myVar = value
	}
}
```

However, this one doesn't (note that I removed `this`):

```
fun createObject() = object {
	private var myVar: String? = null
	fun assign(value: String?) {
		myVar = value
	}
}
```

Nor this one (`this` is there, but there's no function body):

```
val myObject = object {
	private var myVar: String? = null
	fun assign(value: String?) {
		this.myVar = value
	}
}
```

arturbosch:


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
```diff
@@ -122,4 +122,48 @@ class VarCouldBeValSpec : SubjectSpek<VarCouldBeVal>({
...
+					private var myVar: String? = null
+					fun assign(value: String?) {
+						this.myVar = value
+					}
+				}
+			""".trimIndent()
+			assertThat(subject.lint(code)).isEmpty()
+		}
+
+		it("should report unused local variable") {
```
schalkms: *should not report* I guess

PR #1269, opened 2018-10-20
Split UnnecessaryApply tests

arturbosch:
- fixes formatting issues due to activation of detekt-formatting
- splits positive test cases

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
@@ -11,12 +11,12 @@ class UnnecessaryApplySpec : SubjectSpek<UnnecessaryApply>({
 
 	subject { UnnecessaryApply(Config.empty) }
 
-	given("unnecessary applies that can be changed to ordinary method call") {
+	given("unnecessary apply expressions that can be changed to ordinary method call") {
 
 		it("reports an apply on non-nullable type") {
 			assertThat(subject.lint("""
 				fun f() {
-					val a : Int? = null
+					val a : Int = 0
 					a.apply {
 						plus(1)
 					}
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
@@ -56,7 +56,7 @@ class UnnecessaryApplySpec : SubjectSpek<UnnecessaryApply>({
 						toString()
 					})
 				}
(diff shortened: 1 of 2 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
```diff
@@ -8,33 +8,60 @@ import org.jetbrains.spek.api.dsl.it
...
+
+		it("does report single statement in apply used as function argument") {
+			assertThat(subject.lint("""
+				fun b(i : Int?) {
+				}
+				fun f() {
+					val a : Int? = null
 					b(a.apply {
-						plus(1)
+						toString()
```
arturbosch: Actually this case is wrong, I have changed the test to use `toString()` in the apply which clearly leads to a type error for function `b`.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
```diff
@@ -8,33 +8,60 @@ import org.jetbrains.spek.api.dsl.it
...
 class UnnecessaryApplySpec : SubjectSpek<UnnecessaryApply>({
+
 	subject { UnnecessaryApply(Config.empty) }
 
-	given("some code using apply expressions extensively") {
-		it("reports unnecessary applies that can be changed to ordinary method call") {
-			val findings = subject.lint("""
-				fun b(i : Int) {
-				}
+	given("unnecessary applies that can be changed to ordinary method call") {
```
Mauin:
```suggestion
	given("unnecessary apply expressions that can be changed to ordinary method call") {
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
```diff
@@ -8,33 +8,60 @@ import org.jetbrains.spek.api.dsl.it
...
+
+		it("reports an apply on non-nullable type") {
```
Mauin: The test description says the type is non-nullable. But `a` in this test case is `Int?` and thus nullable.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
```diff
@@ -8,33 +8,60 @@ import org.jetbrains.spek.api.dsl.it
...
+			assertThat(subject.lint("""
 				fun f() {
 					val a : Int? = null
```
Mauin: I guess this should be non-nullable?

```suggestion
					val a : Int = null
```

Issue #1245, opened 2018-10-14
Rule: Prefer IntArray instead of Array<Int> to escape boxing


Mishkun:
### Prerequisites
Since Kotlin has no notion of primitive types, there are speicalized Array instances for each Java primitive:

- `IntArray`
- `DoubleArray`
- `FloatArray`
- `ShortArray`
- `ByteArray`
- `LongArray`
- `CharArray`

_From this point I will refer to the arrays of primitive types as `Array<Primitive>`._

As stated here in [docs](https://kotlinlang.org/docs/reference/basic-types.html#arrays) and [stackoverflow](https://stackoverflow.com/questions/45090808/intarray-vs-arrayint-in-kotlin), using `Array<Primitive>` leads to implicit boxing and perfomance hit.

Thus, I think we should mark `Array<Primitive>` usages to help improve performance and educate non-Java newcomers

### Actual rule

Actual rule should detect these patterns:
- [x] Using `Array<Primitive>` as a parameter of the function
- [x] Using `Array<Primitive>` as an explicit return type
- [ ] Using `arrayOf()` factory with primitive types
- [ ] All references to `Array<Primitive>` types

These can be standalone rules or one rule tweakable from config

schalkms:
@Mishkun Is there any specific reason why this rule should only detect the following cases and not all usages of `Array<Primitive>`.
- Using Array<Primitive> as a parameter of the function
- Using Array<Primitive> as an explicit return type
- Using arrayOf() factory with primitive types

schalkms:
@Mishkun I updated your initial issue post to supply a check list for future features/improvements of this rule.

Issue #852, opened 2018-04-12
Regex Performance Rule


schalkms:
It's a common mistake to initialize a Regex inside a loop or a frequently called function. Internally the compiler creates a nondeterministic finite automaton (NFA) which is then translated into a deterministic finite automaton DFA which consumes time. The compiler must recompile the pattern for each iteration or function call.
Creating a static or member variable solves this problem.

Therefore, this rule should lint uses of
```kotlin
Regex("pattern")
Pattern.compile("pattern")
```
in non-static methods excluding constructors and initializers.
Furthermore, the rule has to check if the pattern is really a constant.

schalkms:
There are other performance issues with regular expressions like intensive backtracking as well, but I think detekt should focus on the ones mentioned in my initial description.

arturbosch:
Finding Regexes inside loops should fairly easy to implement. Regex inside a method which is called often is hard, how do you know it is called often? Is 10 times from other methods often?

How about (after symbol resolving) or through control flow graphs we find methods with loops which call this method where the regex is constructed?

schalkms:
I think this is currently very much out of scope for detekt.
Profiling tools are much better at analyzing stuff like this.
Therefore, I'll close this.

PR #1270, opened 2018-10-20
Add rule for detecting arrays of primitive types in function parameters

elaydis:
Adds a rule to detect arrays of primitive types in function parameters.
See issue #1245 for more information.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/ArrayPrimitive.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/ArrayPrimitive.kt
@@ -1,0 +1,58 @@
+package io.gitlab.arturbosch.detekt.rules.performance
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.psi.KtParameter
+
+/**
+ * Using Array<Primitive> leads to implicit boxing and performance hit. Prefer using Kotlin specialized Array
+ * Instances.
+ *
+ * As stated in the Kotlin [documention](https://kotlinlang.org/docs/reference/basic-types.html#arrays) Kotlin has
+ * specialized arrays to represent primitive types without boxing overhead, such as `IntArray`, `ByteArray` and so on.
+ *
+ * <noncompliant>
+ * fun function(array: Array<Int>)
+ * </noncompliant>
+ *
+ * <compliant>
+ * fun function(array: IntArray)
+ * </compliant>
+ *
+ * @author elaydis
+ * @author inytar
+ */
+class ArrayPrimitive(config: Config = Config.empty) : Rule(config) {
+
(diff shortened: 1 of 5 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/ArrayPrimitiveSpec.kt
```diff
@@ -0,0 +1,31 @@
...
+import io.gitlab.arturbosch.detekt.test.lint
+import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.spek.api.dsl.describe
+import org.jetbrains.spek.api.dsl.it
+import org.jetbrains.spek.subject.SubjectSpek
+
+class ArrayPrimitiveSpec : SubjectSpek<ArrayPrimitive>({
+	subject { ArrayPrimitive() }
+
+	describe("function parameter") {
```
schalkms: Please make sure to test this thoroughly with more cases.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/ArrayPrimitive.kt
```diff
@@ -0,0 +1,58 @@
...
+class ArrayPrimitive(config: Config = Config.empty) : Rule(config) {
+
+	override val issue = Issue("ArrayPrimitive",
+			Severity.Performance,
+			"Using Array<Primitive> leads to implicit boxing and a performance hit",
+			Debt.FIVE_MINS)
+
+	override fun visitParameter(parameter: KtParameter) {
+		val regex = Regex("""^Array<(\w+)>$""")
+		val matchResult = regex.find(parameter.typeReference?.text as CharSequence)
```
schalkms: There's no need to use a regex here. I am pretty sure that there's a way to get the generic type parameter from the abstract syntax tree.

schalkms: Please do not initialize a regex inside a frequently called function since this could lead to performance problems. Please look at #852 for the in-depth explanation why this might not be a good idea.

elaydis: Do you have a suggestion how to get the generic type parameter from the abstract syntax tree? I couldn't find it.

schalkms: It's a bit tricky (`typeReference.typeElement.typeArgumentsAsTypes`)
I updated this rule. This does also detect cases where the array type is part of a generic type parameter as well.

schalkms: Anyways, this is a good first step into making this rule part of detekt. Thanks for that!

elaydis: Thanks for implementing this, looks very neat! We learned a lot from this.

elaydis:
We've implemented the rules for the first two cases mentioned in #1245 and feel like the third case would require a lot more thought, so we hope this can be merged as is. Maybe we will revisit this and implement a rule for the third case as well.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/ArrayPrimitive.kt
schalkms: Glad that I could help.

Issue #1261, opened 2018-10-19
detektBaseline task returns failure on success


3flex:
## Expected Behavior
detektBaseline task should not fail after successfully generating baseline file.

## Observed Behavior
Get message saying:
```
Successfully wrote smell baseline to C:\Users\username\AndroidStudioProjects\kotlinconf-app\android\app\config\detekt\baseline.xml"
```

but the Gradle build fails saying
```
> Task :android:app:detektBaseline FAILED

FAILURE: Build failed with an exception.
```

Because detekt still reports the code smells it found when generating the baseline.

## Steps to Reproduce
Run `./gradlew detektBaseline` in a project that has has some code smells.

## Context
The task successfully does what is required, but the task fails, which doesn't make sense.

## Your Environment
* Version of detekt used: master
* Operating System and version: Windows 10
* Link to your project (if it's a public repository): N/A

marschwar:
Is there any indication as to why the tasks failed? 
Could you please rerun the task with `--info` and `detekt { debug = true }`. I have not been able to reproduce the issue on Linux.
Did the baseline file already exist prior to task execution?

3flex:
This is when generating a baseline where none currently exists.

1. Add detekt to a project with existing issues and set a path for `detekt.baseline` property
2. Run `gradlew detektBaseline`

Sorry missed the reason the task fails from the message I put in the issue - it's failing on a BuildFailure exception reported by detekt:

```
Exception in thread "main" io.gitlab.arturbosch.detekt.cli.console.BuildFailure: Build failed with 319 weighted issues (threshold defined was 10).

> Task :android:app:detektBaseline FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':android:app:detektBaseline'.
```

I'd expect the task just generates a baseline file, but doesn't report on any issues - as a user I know there are issues which is why I'm generating the baseline in the first place. The task should just create the baseline and complete successfully, but not report any issues that were found.

marschwar:
The `FAILURE: Build failed with an exception.` makes me curious. Could you please rerun using `--stacktrace`.

3flex:
That's detekt itself - it returns a BuildFailure exception when it finds a number of issues over the set threshold.

The stacktrace doesn't show that, but if you run detekt directly (not using Gradle) and have it create a baseline with input that has issues it will do the same thing. It's not related to the Gradle plugin.

```
[...]

PR #1271, opened 2018-10-20
No BuildFailure for create-baseline 

marschwar:
Solves #1261 

The create-baseline task should not finish with an error if the baseline file was written correctly.

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/BuildFailureReport.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/BuildFailureReport.kt
@@ -62,7 +62,7 @@ class BuildFailureReport : ConsoleReport() {
 	}
 
 	private fun extractRuleToRuleSetIdMap(detektion: Detektion): HashMap<String, String> {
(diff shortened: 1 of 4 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/BuildFailureReport.kt
```diff
@@ -72,8 +72,8 @@ class BuildFailureReport : ConsoleReport() {
 		return weightsConfig.valueOrDefault(id,
 				if (key != null) weightsConfig.valueOrDefault(key, 1) else 1)
 	}
-}
 
-fun Int.reached(amount: Int): Boolean = !(this == 0 && amount == 0) && this != -1 && this <= amount
+	fun Int.reached(amount: Int): Boolean = !(this == 0 && amount == 0) && this != -1 && this <= amount
```
schalkms: What has changed here?

marschwar: I moved the functions inside the class to reduce its visibility.

PR #1273, opened 2018-10-21
Updated MethodOverloading message - fixed #1223

### LINKED ISSUE #1223: MethodOverloading with wrong line
## Expected Behavior
The issue should point to the method which is overloaded (e.g. 1251 in the example)

## Current Behavior
Points to 1:1

- git clone https://github.com/loewenfels/dep-graph-releaser.git
- remove `'MethodOverloading'` from exlude-rules in ./gradle/scripts/detekt.yml
- run `./gr detektCheck`
- see report in ./build/detekt

## Your Environment
* Version used: 1.0.0.RC8

schalkms:
The reported message for the MethodOverloading rule includes the method,
which is overloaded too many times.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/MethodOverloading.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/MethodOverloading.kt
@@ -58,7 +58,7 @@ class MethodOverloading(config: Config = Config.empty,
 				report(ThresholdedCodeSmell(issue,
 						Entity.from(element),
 						Metric("OVERLOAD SIZE: ", it.value, threshold),
-						message = "This method is overloaded too many times."))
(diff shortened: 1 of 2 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/MethodOverloading.kt
```diff
@@ -58,7 +58,7 @@ class MethodOverloading(config: Config = Config.empty,
 				report(ThresholdedCodeSmell(issue,
 						Entity.from(element),
 						Metric("OVERLOAD SIZE: ", it.value, threshold),
-						message = "This method is overloaded too many times."))
+						message = "The method '${it.key}' is overloaded too many times."))
```
Mauin: We could make the message even more detailed: `The method "${it.key}" is overloaded x times (limit: y)`

PR #1275, opened 2018-10-22
MagicNumber: Allow 300 by default.

vanniktech:
I would not consider 300 being a magic number. I use it very often when delaying ui actions or detecting double clicks and I think it's pretty standard. Let me know what you think.


schalkms:
I would say no. If this configuration works for you, this is fine. I think we should let the user decide.

arturbosch:
As a non-android developer I never use 300 :D. This is pretty subjective and should not be as a default value. The next guy will come and want 250 ... then we can do surveys and studies and maybe come to 277 like we a python code formatter uses max line length of 88 because they done some research ...

PR #1280, opened 2018-10-23
Readd missing `detekt-rules` dependency for cli to avoid NOP detekt run

arturbosch:
- reintroduces rules dependency
- fixes some detekt issues
- readds baseline with last two issues


detekt-cli/build.gradle.kts
```diff
@@ -13,6 +13,7 @@ configurations.testImplementation.extendsFrom(configurations.kotlinTest)
 
 dependencies {
 	implementation(project(":detekt-core"))
+	implementation(project(":detekt-rules"))
```
3flex: Use `runTimeOnly` instead of `implementation` so it's included on the runtime classpath but not the compilation classpath (where it's not required).

3flex: Also - are there missing tests that would have caught this? Definitely wouldn't have introduced that change knowing that it would drop the rules from the classpath so detekt wouldn't pick up issues when run on CLI

arturbosch: No there are no such tests (the tests have the rules dependency, everything is loaded).
An additional CI step could be to actually run detekt on one prepared file which should always contain smells.

schalkms: I am also voting for this @arturbosch 
A test here would prevent this error from happening again.

Issue #1264, opened 2018-10-19
false positive unused import


robstoll:
## Observed Behavior
```
import ch.tutteli.kbox.forEachRemaining as foreachRemaining //otherwise we use jdk8
fun foo(){
   listOf().iterator().foreachRemaining{}
}
```
Real use case including reporting:
https://sonarcloud.io/project/issues?id=robstoll_kbox&open=AWaMMX83Vfn4NtZOwYPB&resolved=false&sinceLeakPeriod=true&types=CODE_SMELL

The alias of the import is used in code. A wild guess: detekt is looking for `forEachRemaining` instead of the alias and does not find it; thus the false positive.

## Your Environment
* Version of detekt used: 1.0.0.RC8
* Link to your project: https://github.com/robstoll/kbox/blob/master/kbox-jvm/src/test/kotlin/ch/tutteli/kbox/ForEachRemainingSpec.kt

robstoll:
Yes it is detekt. Click on the issue on sonarcloud, you can see that it has the detekt label.

arturbosch:
I'm still not sure if the report is really from detekt. If sonar-cload uses an old version maybe. But as you see in the PR I needed your code and the false positive is not there :O

arturbosch:
Feel free to reopen this ticket if your problem still occurs in RC9.2 or newer.

arturbosch:
Thanks!

PR #1284, opened 2018-10-24
Add test case for reported false positive - #1264

arturbosch:


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportsSpec.kt
```diff
@@ -167,4 +167,17 @@ class UnusedImportsSpec : SubjectSpek<UnusedImports>({
 			assertThat(subject.lint(code)).isEmpty()
 		}
 	}
+
+	given("imports with aliases") {
+
+		it("should not report import as unused because the alias is used") {
+			val code = """
+				import ch.tutteli.kbox.forEachRemaining as foreachRemaining // otherwise we use jdk8
```
schalkms: Perhaps we should simplify this code a bit.

arturbosch: fair enought :)

PR #1289, opened 2018-10-26
Fix a simple typo in TooManyFunctions rule's description text

bjoernQ:
This is just a simple fix for a type in the TooManyFunctions rule's description text.


bjoernQ:
Sorry this fails on Appveyor but I'm unable to figure out why

PR #1291, opened 2018-10-26
Generate test coverage report and upload to Codecov

### LINKED ISSUE #301: Add code coverage analysis to the CI process
The idea here is to integrate a code coverage tool into detekt's CI process to measure the degree to which the source code of this repo is executed.
A tool which might be used for this is **Jacoco**.

3flex:
Closes #301 (for everything except the Gradle plugin, which is trickier than I thought, so didn't pursue it yet. See [here](https://discuss.gradle.org/t/gradle-plugins-integration-tests-code-coverage-with-jacoco-plugin/12403) for more).

Results have started coming through already as a result of this PR: https://codecov.io/gh/arturbosch/detekt

There will be some areas marked not covered, but actually are. There are some open Kotlin-related issues in JaCoCo. The report should be used as an indication only about relative test coverage, but won't be 100% accurate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ script:
   - ./gradlew verifyGeneratorOutput
 after_success:
   - .buildscript/deploy_snapshot.sh
+  - ./gradlew jacocoTestReport
+  - bash <(curl -s https://codecov.io/bash)
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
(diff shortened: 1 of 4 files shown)


3flex:
Definitely some gaps. Jacoco working on better checks for inline functions, default arguments and other Kotlin enhancements in upcoming 0.8.3 version. Will need to track tool updates as they come to make sure that's taken advantage of.

And I've just realised tool version should be specified in gradle.properties along with other dependencies... Will fix.

PR #1292, opened 2018-10-27
Typealiases for RuleId and RuleSetId

Mauin:
Currently we pass a lot of `String`s around which don't really convey what they actually include.

By introducing these typealiases for `RuleId` and `RuleSetId` it becomes a bit clearer what these types actually contain and mean imo.

I just went through some of the files where I found usages of the Id Strings and replaced them with the typealias, I probably missed some cases.


pavlospt:
Also with Kotlin 1.3 coming up and inline classes being introduced, this change is made even better!

PR #1293, opened 2018-10-28
Ignore FunctionMaxLength by default for tests.

vanniktech:
Personally, I like if function names for tests are a bit longer and expressive. Detekt itself does not have that problem since it's using Spek and the 'name' is passed as a String. Otherwise we'd have a lot of warnings reported in the Detekt codebase itself.


vanniktech:
I'm using failFast and hence it's on for me.

arturbosch:
Well, I can see the point in having this ignored in test pattern.
I'm fine with it if @Mauin and @schalkms are also fine with it.

arturbosch:
Here we go :D

vanniktech:
Thanks :)

PR #1296, opened 2018-10-30
Upgrade to Kotlin 1.3.0

3flex:


3flex:
If #1298 is merged first, hold off on merging this - it will be possible to update the Gradle plugin's Kotlin version to 1.3.0 at the same time, so will update this PR to do that after kotlin-dsl is dropped from the Gradle plugin.

arturbosch:
KK, #1298 is merged :)

PR #1298, opened 2018-10-30
Drop kotlin-dsl plugin from detekt-gradle-plugin

### LINKED ISSUE #1286: Question: drop kotlin-dsl dependency from Gradle plugin?
kotlin-dsl has a set of Kotlin dependencies which are bundled with Gradle (currently Kotlin 1.2.61). The detekt Gradle plugin is written in Kotlin, but because it has the `kotlin-dsl` plugin as a dependency, hte plugin must be written in the same version of Kotlin.

In the plugin it looks like the only kotlin-dsl feature currently used is the `property` extension function: https://github.com/gradle/kotlin-dsl/blob/v1.0-RC6/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/ObjectFactoryExtensions.kt#L52-L58

Could easily copy the extension function or just use a slightly less pleasant syntax to create a Property, and then the plugin wouldn't be held back to whichever version of Kotlin Gradle bundles.

kotlin-dsl for plugin development. Should it stay or should it go?

3flex:
Resolves #1286 

I just did a straight conversion without introducing new extension functions. They'd make the code a little cleaner, but as is I don't think it's hard to follow (and ends up being a net reduction in LOC).

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -27,16 +27,16 @@ class DetektPlugin : Plugin<Project> {
 
 	private fun createAndConfigureDetektTask(project: Project, extension: DetektExtension) {
 		val detektTask = project.tasks.register(DETEKT, Detekt::class.java) {
-			debugProp.set(project.provider { extension.debug })
-			parallelProp.set(project.provider { extension.parallel })
-			disableDefaultRuleSetsProp.set(project.provider { extension.disableDefaultRuleSets })
-			filters.set(project.provider { extension.filters })
-			config.setFrom(project.provider { extension.config })
-			baseline.set(project.layout.file(project.provider { extension.baseline }))
-			plugins.set(project.provider { extension.plugins })
-			input.setFrom(existingInputDirectoriesProvider(project, extension))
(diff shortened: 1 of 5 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -78,20 +78,20 @@ class DetektPlugin : Plugin<Project> {
 
 	private fun configurePluginDependencies(project: Project, extension: DetektExtension) {
 		project.configurations.create(CONFIGURATION_DETEKT_PLUGINS) {
-			isVisible = false
-			isTransitive = true
-			description = "The $CONFIGURATION_DETEKT_PLUGINS libraries to be used for this project."
+			it.isVisible = false
```
Mauin: For the sake of consistency should we also name this configuration?

arturbosch:
So basically everything can stay the same, but the `it` gets removed in the kotlin-dsl ;) ?

3flex:
> So basically everything can stay the same, but the it gets removed in the kotlin-dsl ;) ?

Haha yeah, not much benefit here it seems. I think the Kotlin DSL is really targeted at build consumers. You might only use the Kotlin DSL outside of a build file when you have internal plugins or include them as part of buildSrc. But seems for plugins that are published for others to use it may cause more trouble than it's worth.

PR #1302, opened 2018-11-01
Get rid of jdk classes

arturbosch:


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/baseline/XmlExtensions.kt
```diff
@@ -1,23 +1,183 @@
...
 
+fun XMLStreamWriter.comment(content: String) {
+	writeComment(content)
+}
+
 fun XMLStreamWriter.attribute(name: String, value: String) = writeAttribute(name, value)
+
+abstract class DelegatingXMLStreamWriter(writer: XMLStreamWriter) : XMLStreamWriter by writer
+
+class IndentingXMLStreamWriter(
```
Mauin: This class could be extracted.

PR #1324, opened 2018-11-09
Updated RethrowCaughtException according to #906

schalkms:
This implements the rule to be compliant with Sonars rule specification.
The specification can be found here:
https://sbforge.org/sonar/rules/show/squid:S1166?modal=true&layout=false

### DIFF
--- a/detekt-rules/src/test/resources/cases/RethrowCaughtExceptionNegative.kt
+++ b/detekt-rules/src/test/resources/cases/RethrowCaughtExceptionNegative.kt
@@ -15,4 +15,14 @@ fun rethrowCaughtExceptionNegative() {
 		print(e) // logging an exception is allowed
 		throw e
 	}
+	try {
+	} catch (e: IllegalStateException) {
+		print("log") // taking specific action before throwing the exception
+		throw e
+	}
+	try {
+	} catch (e: IllegalStateException) {
+		print(e.message) // taking specific action before throwing the exception
+		throw e
+	}
 }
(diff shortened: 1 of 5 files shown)


detekt-rules/src/test/resources/cases/RethrowCaughtExceptionNegative.kt
```diff
@@ -8,11 +8,11 @@ fun rethrowCaughtExceptionNegative() {
 	}
 	try {
 	} catch (e: IllegalStateException) {
-		print(e) // logging an exception is allowed
+		print(e) // just logging the exception is fine
 	}
 	try {
 	} catch (e: IllegalStateException) {
-		print(e) // logging an exception is allowed
+		print("log") // letting an exception propagate without doing anything with it is fine
```
marianosimone: I'd say this is not "doing anything with it". The comment would probably be clearer if it said something:

> letting an exception propagate after doing something with it is fine

schalkms: Yes, thanks. I need to update the rule documentation anyway :)

arturbosch:
Unfortunately this case is not supported:
```
// Noncompliant - exception is lost
try { /* ... */ } catch (Exception e) { throw new RuntimeException("context"); }
```

arturbosch:
Please also add this case to the negative file:
```
try { /* ... */ } catch (Exception e) { throw new RuntimeException("context", e); }
```
Its basically like the rethrow case but with one more parameter

schalkms:
> Unfortunately this case is not supported:
```
// Noncompliant - exception is lost
try { /* ... */ } catch (Exception e) { throw new RuntimeException("context"); }
```

This is handled by the `SwallowedException` rule. I didn't want to duplicate the checks.

arturbosch:
Okay, fine by me :)

PR #1325, opened 2018-11-10
Use a default baseline in gradle and add support for consolidated baseline files.

### LINKED ISSUE #1260: Gradle plugin should set default value for "baseline" property
## Expected Behavior
Onboarding experience for project with existing issues is:
1. Apply plugin in build.gradle
2. Run `gradlew detekt` to check for issues
2. Run `gradlew detektBaseline` to generate a baseline file

## Current Behavior
Onboarding experience for project with existing issues is:
1. Apply plugin in build.gradle
2. Run `gradlew detekt` to check for issues
2. Figure out how to set "baseline" property
3. Run `gradlew detektBaseline` to generate a baseline file

## Context
Sensible defaults make for a better onboarding experience. A user shouldn't have to explicitly set a configuration for the baseline file path when Gradle conventions could be followed to set a default value instead.

Conventionally this will be one of:
* $rootProject.projectDir/config/detekt/baseline.xml
* $projectDir/config/detekt/baseline.xml

The second makes more sense in this context, since there will be multiple detekt tasks (they'll be created for different modules and sourcesets) and having a single baseline would require either merging files or generating different files for different projects & sourcesets (but project and sourceset names change - if that happened and filenames were based on project name, the baseline would have to be regenerated which would be annoying to manage).

marschwar:
This PR is not yet ready but I would like to get your feedback. Eventually it is supposed to resolve #1260.

* The baseline is created by default in `$rootProject.projectDir/config/detekt/baseline.xml`
* If the file `$rootProject.projectDir/config/detekt/baseline.xml` exists, it is used as the baseline for the `detekt` task unless configured differently.
* The CLI accepts an optional `--source-set-id` parameter in order to be able to use one baseline file for multiple subprojects (and in the future source sets within those).
* The baseline file format is changed since there has to be a container for multiple baselines. Existing files in the current format can still be read.

The change missing is to actually use the sourceSetId in the gradle plugin. Could someone please tell me (or point me to some documentation) on how to use this changed cli from within the gradle subproject and how to make it work in CI?

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/baseline/BaselineHandler.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/baseline/BaselineHandler.kt
@@ -13,62 +13,41 @@ internal class BaselineHandler : DefaultHandler() {
 	private var content: String = ""
 	private var whitestamp: String? = null
 	private var blackstamp: String? = null
-	private var currentSourceSetId: String? = null
+	private var sourceSetId: String? = null
 
-	private val currentWhitelists = mutableListOf<Whitelist>()
-	private val currentWhiteIds = mutableSetOf<String>()
+	private val currentBaselines = mutableListOf<Baseline>()
+	private val sourceSetIds = mutableSetOf<String?>()
+	private val whiteIds = mutableSetOf<String>()
 	private val blackIds = mutableSetOf<String>()
 
-	internal fun createConsolidatedBaseline() = ConsolidatedBaseline(
-			Blacklist(blackIds, blackstamp ?: now()),
-			whitelistIdentifiedBy(null),
-			sourceSetWhitelistsAsMap()
-	)
+	private val now = Instant.now().toEpochMilli().toString()
 
-	private fun sourceSetWhitelistsAsMap(): Map<String, Whitelist> {
-		val sourceSets = currentWhitelists.map(Whitelist::sourceSetId)
+	internal fun createConsolidatedBaseline() = ConsolidatedBaseline(currentBaselines.toList())
 
-		val result = mutableMapOf<String, Whitelist>()
-		// by iterating over the ids we also check for duplicates
-		sourceSets.forEach {
-			it?.let { sourceSetId ->
-				val whitelist = whitelistIdentifiedBy(sourceSetId)
-				if (whitelist != null) {
-					result.put(sourceSetId, whitelist)
-				}
-			}
-		}
-
-		return result
-	}
-
-	internal fun createBaseline(sourceSetId: String?) = Baseline(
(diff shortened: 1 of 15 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/baseline/BaselineHandler.kt
```diff
@@ -7,22 +7,41 @@ import java.time.Instant
...
+internal class BaselineHandler : DefaultHandler() {
 
 	private var current: String? = null
 	private var content: String = ""
 	private var whitestamp: String? = null
 	private var blackstamp: String? = null
+	private var sourceSetId: String? = null
+
+	private val currentBaselines = mutableListOf<Baseline>()
+	private val sourceSetIds = mutableSetOf<String?>()
```
Mauin: Do we want to store nullable types in this set?

marschwar: Yes. Null is a valid value. We could also map null to a default value.

marschwar:
This should not change anything for you if you just use the CLI. The resulting baseline files should be identical.

arturbosch:
Hm, I do not think that we have to change the cli module for the gradle plugin.
The baseline feature was not intented to be complex and there should only be one baseline.xml file on top level.
I would propose that `detektBaseline` always runs for the whole project and generates a baseline file for each module in the build directory which are later combined to one baseline in the root module. The place of the last can be specified with the baseline property.
If the user has code with the same signature in two different modules its not our problem ^^.
What do you think?

marschwar:
So what you are saying is that `detektBaseline` will always run on all modules for all source sets and after that is complete, merge all the generated files into one baseline? That is a valid approach in my opinion. It could get a little awkward if the user defines different baseline files in different submodules. How would that be interpreted?

**/build.gradle**
```
detekt {
  baseline = file("config/detekt/baseline.xml")
}
```

**/foo/build.gradle**
```
detekt {
  baseline = file("$rootProject.projectDir/config/detekt/baseline.xml")
}
```

**/bar/build.gradle**
```
detekt {
  baseline = file("config/detekt-baseline.xml")
}
```
Since both `detekt` and `detektBaseline` use the baseline property to me it is not really clear how we would handle this. On the other hand I have no idea if there is even a use case for that or we can just consider it misconfiguration.

arturbosch:
Hm, yeah that is indeed bit strange with the `detekt` task.
But yeah, we just consider it misconfiguration as it was never planned to work like this :X.
Managing many baseline files is imo not feasible and we shouldn't embrace that behaviour :) and instead make clear to always use one location via `rootProject....`

AfzalivE:
4 years later. Would it be okay to consider this now? Many people use baseline per module through gradle.

PR #1326, opened 2018-11-10
Swallowed exception update

schalkms:
When this PR is merged, I suggest to add a config parameter to this rule.
There are exception types like NumberFormatException and ParseException that indicate nonexceptional outcomes. This rule should ignore unhandled exception of this type specified in the config.

### DIFF
--- a/reports/failfast.yml
+++ b/reports/failfast.yml
@@ -47,7 +47,7 @@ exceptions:
   ReturnFromFinally:
     active: true
   SwallowedException:
-    active: true
+    active: false
   ThrowingExceptionFromFinally:
     active: true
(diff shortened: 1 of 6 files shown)


reports/failfast.yml
```diff
@@ -47,7 +47,7 @@ exceptions:
   ReturnFromFinally:
     active: true
   SwallowedException:
-    active: true
+    active: false
```
Mauin: Shouldn't this be generated from the rule KDoc `@active` tag (in this case the tag would be missing as it is not active)?

schalkms: This file is the rule config which is applied to analyzing detekt itself. I deactivated it for now, since there is some work necessary as mentioned in my initial post.

Issue #1295, opened 2018-10-30
Add support for consolidated report in multi-project setup


vasanthdharmaraj:
Based on the recent change highlighted in the [change log](https://arturbosch.github.io/detekt/changelog.html), I have setup a multi-project by adding `detekt` to the subprojects like:

``` 
subprojects {
    detekt {
        // properties
    }
 }
```

The reports are generated successfully in the sub-projects. What should I do to get a consolidated report of all sub-projects? The earlier suggestions seem to be for the older `profile` based config and does not seem to work with this change.

3flex:
It's not currently possible to consolidate reports from the various subprojects into a single report.

Mauin:
It would however be possible to add support for consolidating the reports if that is a requested feature. Adding "Waiting for Feedback" to see if more requests come in for this.

vasanthdharmaraj:
Great. Thanks for the quick response.

schalkms:
This makes sense. This would be a nice improvement, I think.

abelhegedus:
Before migrating to RC9, we took advantage of the aggregated report that could be easily published as part of a Jenkins build, similarly to JUnit results or JaCoCo coverage.
Additionally, I think for a multi-project setup, it makes more sense to fail the build by taking into account the aggregated score instead of separate modules.

athkalia:
Having one report would indeed be great.

marschwar:
I have started to do some work on this topic and hope to have something to show for next week.

shareme:
a question I have is detekt using gradle's reporting stuff if so than the gradle build dashboard should consolidate things into one dash board I mention it as we can look at the code for that plugin to bootstrap on how to consolidate reports...just thinking out loud as I have the same problem as I am doing clean arch as separate modules deps to my app module

galet:
I think this is a legitimate and important enhancement. We have a project with 300+ modules, so having a single merged report would be great.

arturbosch:
Yes I agree. Please see the latest comments at #1329.
@galet 300 kotlin modules? How many minutes does detekt need to fully inspect your code base? Asking just of curiousity :)

galet:
Thanks. The project contains mostly Java modules but only a few Kotlin or mixed Java/Kotlin modules since we just started with Kotlin adoption. Nevertheless, it's still hard to find and read detekt report one by one with so many modules.

abelhegedus:
[...]

PR #1329, opened 2018-11-12
Consolidate reports in multi project setup

marschwar:
This is intended to solve #1295.

Since all subprojects are processed individually it is really only possible to consolidate on a file level.
The solution does not really seem that great to be quite honest, but I don't really see more options and would like your feedback.
I intentionally do not parse the xml to reduce the the amount of coupling to CLI as well as hopefully increase processing speed.


marschwar:
Just realized that this will only work if the parent contains sources. More to come

arturbosch:
What does happen if i run `detekt` in a sub module? All sub projects of that sub module are merged?
Or will this feature only work if run from top level?
I also think that the file level is the only option here.

marschwar:
> All sub projects of that sub module are merged?
Correct, that was the idea. Does that make any sense?

arturbosch:
Hm kind of yes, I have a project where I have like 5 modules with no sources but with rather 2-8 more modules per module. That would make sense here.
So my `parent` with no sources, should also produce an uber report :)

arturbosch:
Can you please rebase it and check the compilation errors reported by CI. Would be good to get this merged :)

marschwar:
The build failure seems to be a problem with the line endings on Win. I assume the expectations do not have `\r\n` and therefore do not match.  Does anyone have an elegant solution? Unfortunately I do not have a Win machine available and can only guess.

arturbosch:
Maybe something like `trim` or `trimEnd` can help in the tests? Thats what we often use in the Spec files...

audkar:
I will just say: tab vs spaces :fire: 

But seriously. Careful seems that you are using tabs for your code, when project uses spaces :)

arturbosch:
Yes, sorry if you missed this. We reformat the code base with KtLint and use spaces now.
Strange that there were only two check runs and our formatting does not fail the build oO

marschwar:
All right. Since it has to be rebased anyway I will have a look. I actually did try to find the right idea settings. I guess https://github.com/arturbosch/detekt/blob/master/idea-code-style.xml is no longer valid? Should it be removed then?

arturbosch:
Yes please remove it, thanks.

marschwar:
I think I fixed my local setup now. Thanks for the pointers.

With regards to this PR I am not so sure it should be merged as it will most likely collide with #1221 and I don't know if it really has that much benefit.

arturbosch:
Hm, maybe we should promote/document more the use case to have a detektAll task should run over the whole code base like we self do in https://github.com/arturbosch/detekt/blob/master/build.gradle.kts#L289 ?

marschwar:
I agree. It is obviously possible to achieve the same results without introducing even more complexity into the gradle plugin.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -140,20 +141,33 @@ open class Detekt : SourceTask() {
...
+            ParallelArgument(parallelProp.getOrElse(false)),
+            BuildUponDefaultConfigArgument(buildUponDefaultConfigProp.getOrElse(false)),
+            FailFastArgument(failFastProp.getOrElse(false)),
+            DisableDefaultRuleSetArgument(disableDefaultRuleSetsProp.getOrElse(false))
         )
 
         DetektInvoker.invokeCli(project, arguments.toList(), debugProp.getOrElse(false))
+
+        if (xmlReportTargetFile != null) {
+            val xmlReports = project.subprojects.flatMap { subproject ->
```
tasomaniac: What about more nested subprojects? Should this be recursively going over all subproject tree?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -43,6 +43,12 @@ class DetektPlugin : Plugin<Project> {
             it.setExcludes(defaultExcludes)
             it.reportsDir.set(project.provider { extension.customReportsDir })
             it.reports = extension.reports
+
+            project.subprojects.forEach { subProject ->
+                subProject.tasks.firstOrNull { t -> t is Detekt }?.let { subprojectTask ->
```
tasomaniac: What about 

```
subProject.tasks.withType(Detekt::class) { subprojectTask ->
    it.dependsOn(subprojectTask)
}
```

tasomaniac:
@arturbosch I'm not sure why running it in the whole project is better. Imagine a really big project with lots of modules. When nothing has changed, the Gradle plug-in has caching support where the tool will not even run on unchanged modules. It only runs on changed which is great.

Running it on all sources all the time is not scalable.

arturbosch:
@tasomaniac I agree however merging xml files on strange conditions (what about nested modules etc) feels overly complex and hacky. If there is more demand we should try to solve this with the capabilities of gradle?

Maybe even a `collectReportTasks`? We know the filenames in the build directory and could merge all.
But now again, for all created reports? Only from the last run?

tasomaniac:
Filenames in build directories wouldn't work because there might be stale ones from old builds if the user didn't clean. 

Closing this PR is ok. It's a tricky problem. Not easy to solve. Definitely post 1.0 problem imo. 

This can be solved with really good Gradle setup. Android Lint does this with `checkDependencies` flag.

arturbosch:
I would also prefer a gradle solution here, as it is gradle which cuts our report into pieces :)

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
marschwar: That was the idea. Since the subproject task in turn depends on its sub project tasks, the reports should be consolidated bottom up.

Issue #1317, opened 2018-11-08
ReturnCount false positive with lambdas


maragues-kolibree:
## Expected Behavior
ReturnCount does not take into account `return` inside lambdas

## Observed Behavior
If you have multiple returns inside a lambda, ReturnCount is triggered

## Steps to Reproduce
This function violates ReturnCount, and I think it's a false positive
```kotlin
fun readUsers(name: String): Flowable<User> {
        return userDao.read(name)
            .flatMap {
                if (it.isEmpty()) return@flatMap Flowable.empty<User>()

                return@flatMap Flowable.just(it[0])
            }
    }
```

I know the last return can be removed and that a flowable doesn't make sense here, but let's ignore that

## Your Environment

* Version of detekt used: 1.0.0-RC10
* Version of Gradle used (if applicable): 4.9
* Operating System and version: MacOS 10.14

arturbosch:
Agree with @schalkms .

schalkms:
Closed with ee6a252

arturbosch:
Woot, but it is not merged yet oO?

3flex:
https://github.com/arturbosch/detekt/commit/35686e849607b400307e82b55816cd50bb28cdc8 is in master... can close?

schalkms:
Closed with 35686e8

PR #1336, opened 2018-11-14
Allow to exclude labeled return statements - #1317

arturbosch:


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -43,13 +43,16 @@ import org.jetbrains.kotlin.psi.KtReturnExpression
  * (default: 2)
  * @configuration excludedFunctions - define functions to be ignored by this check
  * (default: "equals")
+ * @configuration excludeLabeled - if labeled return statements should be ignored
+ * (default: true)
```
schalkms: Should this really be true?

arturbosch: Hm, and maybe we should not just introduce ignore labeled but further the reported issue is that returns in lambdas are allowed. We should turn `excludeLabeled` to false by default and fix the labeled returns in Lambdas?

arturbosch:
Added option to exclude labaled returns returning from lambda - active by default and the option to ignore all labaled returns - off by default. Still approved :) ?

Issue #1307, opened 2018-11-03
ProjectLayout.fileProperty() method has been deprecated


juergenzimmermann:
Using Gradle _5.0-rc-1_ with detekt _1.0.0-RC10_ I get this stacktrace:
```
The ProjectLayout.fileProperty() method has been deprecated. This is scheduled to be removed in Gradle 6.0. Please use the ObjectFactory.fileProperty() method instead.
        at org.gradle.api.internal.file.DefaultProjectLayout.fileProperty(DefaultProjectLayout.java:78)
        at io.gitlab.arturbosch.detekt.DetektCreateBaselineTask.<init>(DetektCreateBaselineTask.kt:43)
        at io.gitlab.arturbosch.detekt.DetektCreateBaselineTask_Decorated.<init>(Unknown Source)
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
        at org.gradle.api.internal.DependencyInjectingInstantiator.newInstance(DependencyInjectingInstantiator.java:65)
        at org.gradle.api.internal.project.taskfactory.TaskFactory$1.call(TaskFactory.java:72)
        at org.gradle.api.internal.project.taskfactory.TaskFactory$1.call(TaskFactory.java:69)
        at org.gradle.util.GUtil.uncheckedCall(GUtil.java:459)
        at org.gradle.api.internal.AbstractTask.injectIntoNewInstance(AbstractTask.java:188)
        at org.gradle.api.internal.project.taskfactory.TaskFactory.create(TaskFactory.java:69)
        at org.gradle.api.internal.project.taskfactory.PropertyAssociationTaskFactory.create(PropertyAssociationTaskFactory.java:48)
        at org.gradle.api.internal.project.taskfactory.AnnotationProcessingTaskFactory.create(AnnotationProcessingTaskFactory.java:44)
        at org.gradle.api.internal.tasks.DefaultTaskContainer.createTask(DefaultTaskContainer.java:318)
[...]

Issue #1306, opened 2018-11-03
DefaultTask.newInputFile() method has been deprecated


juergenzimmermann:
Using Gradle _5.0-rc-1_ with detekt _1.0.0-RC10_ I get this stacktrace:
```
The DefaultTask.newInputFile() method has been deprecated. This is scheduled to be removed in Gradle 6.0. Please use the ObjectFactory.fileProperty() method instead.
        at org.gradle.api.DefaultTask.newInputFile(DefaultTask.java:66)
        at io.gitlab.arturbosch.detekt.Detekt.createNewInputFile(Detekt.kt:140)
        at io.gitlab.arturbosch.detekt.Detekt.<init>(Detekt.kt:59)
        at io.gitlab.arturbosch.detekt.Detekt_Decorated.<init>(Unknown Source)
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
        at org.gradle.api.internal.DependencyInjectingInstantiator.newInstance(DependencyInjectingInstantiator.java:65)
        at org.gradle.api.internal.project.taskfactory.TaskFactory$1.call(TaskFactory.java:72)
        at org.gradle.api.internal.project.taskfactory.TaskFactory$1.call(TaskFactory.java:69)
        at org.gradle.util.GUtil.uncheckedCall(GUtil.java:459)
        at org.gradle.api.internal.AbstractTask.injectIntoNewInstance(AbstractTask.java:188)
        at org.gradle.api.internal.project.taskfactory.TaskFactory.create(TaskFactory.java:69)
        at org.gradle.api.internal.project.taskfactory.PropertyAssociationTaskFactory.create(PropertyAssociationTaskFactory.java:48)
        at org.gradle.api.internal.project.taskfactory.AnnotationProcessingTaskFactory.create(AnnotationProcessingTaskFactory.java:44)
        at org.gradle.api.internal.tasks.DefaultTaskContainer.createTask(DefaultTaskContainer.java:318)
[...]

PR #1350, opened 2018-11-17
Update to Gradle v5

Mauin:
Updates Gradle to the latest RC of the upcoming major v5 to test how many chnages need to be made in detekt.

I'll keep this PR updated with new RC versions when they come in to track their changes.

This should obviously only be merged once Gradle v5 final is released.


Mauin:
Doesn't look that way: https://github.com/gradle/gradle/blob/v5.0.0-RC3/subprojects/core-api/src/main/java/org/gradle/api/model/ObjectFactory.java#L143

3flex:
Let's just use the deprecated methods for compatibility for the time being. Deprecation messages are annoying but it's better to keep broader compatibility. The deprecated feature will be available until gradle 6 at least according to their deprecation policy.

Mauin:
@3flex I agree. Until most people updated to Gradle v5 (and other popular Plugins do the step to only support Gradle >5) we should rely on the deprecated methods. However this will print warnings on every consumer build. 
The Kotlin Gradle Plugin also uses these methods and prints the deprecation warnings on every detekt build. Alternatively we could add a small compatibility layer which will pick the best method depending on the Gradle version used by the consumer.

detekt-sample-extensions/build.gradle.kts
```diff
@@ -5,6 +5,7 @@ val spekVersion: String by project
 
 dependencies {
 	implementation("io.gitlab.arturbosch.detekt:detekt-api:$usedDetektVersion")
+	implementation(kotlin("compiler-embeddable"))
```
Mauin: Had to add this dependency after updating to v5 RC4

Mauin:
Gradle v5 is now final: https://docs.gradle.org/5.0/release-notes.html

I updated this PR.

arturbosch:
So merging this does not break anything and we are still good to go on gradle 4.9+ ?
This does not make me happy :( ?
```
> Could not HEAD 'https://repo.jfrog.org/artifactory/libs-release-bintray/org/jetbrains/kotlin/kotlin-native-utils/1.3.0/kotlin-native-utils-1.3.0.pom?referrer'.
               > sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```

3flex:
> This does not make me happy

Pretty sure this is happening because some OpenJDK versions don't ship with a reasonable certificate authority store.

There are workarounds which require a couple extra lines in the build file, or switch back to the Oracle version instead where possible.

Or, since only 8 & 11 are currently supported, drop 9 & 10 from CI.

schalkms:
@3flex I didn't know that. Thanks for providing this information. I think we should remove it then. This also speeds up the build.

3flex:
@Mauin suggest adding "closes #1306, closes #1307" to your issue description so those issues close when this is merged.

3flex:
CI should go green if you rebase now that #1371 is merged

detekt-sample-extensions/build.gradle.kts
arturbosch: Is this dependency really necessary?
What if we change the dependency on detekt-api from implementation to compile, still needed?
detekt externsions should not need to add extra dependency on the kotlin compiler.

Mauin: `compile` dependencies should be deprecated in favor of the new `implementation` and `api`. However changing it to either of those does not change anything. I also tried changing the `compiler-embeddable` dependency in `detekt-api` to `api` which I'd assume should fix this, but it didn't.

3flex: > I'd assume should fix this, but it didn't.

I was going to suggest the same, but as I saw the comment realised why that didn't work - because in `detekt-sample-extensions` the dependency on `detekt-api` is not a project dependency, but an external dependency. So when you change `implementation` to `api` in `detekt-api` there's no effect.

That said, changing the `compiler-embeddable` dependency in `detekt-api` to `api` is still the correct thing to do.

arturbosch:
Hm, where does the compilation error came from ?

Mauin:
There were now warnings for a missing explicit dependency on `kotlin-reflect` which I have now added.

Mauin:
Same thing happened on the re-run. The `detekt-rules:test` task was executed but I don't see any tests logged in the output.

schalkms:
@Mauin I also had problems executing `detekt-rules:test` tests on my local machine. They didn't run.
The tests also don't run on the current master branch because there's only a test coverage of **33 %**.

Mauin:
That is definitely an issue then. And an issue that already existed.

It's weird because I ran the tests on my machine this morning and the
`detekt-rules` tests were successfully executed.

Issue #1359, opened 2018-11-23
Allow multiple 'it' references in UnnecessaryLet


Nycto:
## Expected Behavior
I expect this code to be considered valid by the `UnnecessaryLet` validator:

```kotlin
fun f() {
    val a : Int? = null
    a?.let { it.plus(it) }
    a?.let { foo -> foo.plus(foo) }
}
```

## Observed Behavior
The `UnnecessaryLet` validator reports a code smell, even though the parameter can't be removed -- it's referenced multiple times

## Steps to Reproduce
The code above is easy to drop into a test to confirm this is broken

## Context
This creates false positives that require a `Suppress` statement to work around

## Your Environment
* Version of detekt used: 565d86d780c15ade77b5468a9e342ba334181333
* Version of Gradle used (if applicable): 4.10.2
* Operating System and version: Linux 4.4.0-93-generic amd64
* Link to your project (if it's a public repository): Private repo

PR #1360, opened 2018-11-23
Support multiple it refs in UnnecessaryLet (#1359)

Nycto:
See the bug report here: https://github.com/arturbosch/detekt/issues/1359

This change updates the `UnnecessaryLet` rule to count the number of references in a `let` lambda for the parameter. If it finds more than one, the rule passes rather than failing.


Nycto:
Argh. Looks like IntelliJ added a wildcard import. Let me get this updated. Apologies for the noise

Nycto:
Thanks for the CR! What's the process for finalizing the merge?

PR #1365, opened 2018-11-25
Do not add +1 complexity for nested functions inside functions

### LINKED ISSUE #1037: ComplexMethod false positive for returning anonymous inner class

arturbosch:
Closes #1037

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
@@ -6,13 +6,19 @@ import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtIfExpression
 import org.jetbrains.kotlin.psi.KtLoopExpression
 import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.KtObjectLiteralExpression
 import org.jetbrains.kotlin.psi.KtTryExpression
 import org.jetbrains.kotlin.psi.KtWhenEntry
 import org.jetbrains.kotlin.psi.KtWhenExpression
 import org.jetbrains.kotlin.psi.psiUtil.getCallNameExpression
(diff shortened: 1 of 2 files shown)


Mauin:
What about local functions? imo those still should increase the complexity and are different from the case we're trying to address here.

arturbosch:
Hm, one can argue that local functions help to declutter and structurize your code, removing nestings etc removes complexity. Like an algorithm which implementation is a whole function with many little helper local functions. I would't call this code complex but rather an architectural choice to build it this way :)?

schalkms:
I also agree with @Mauin 
I think the complexity of nested functions should be taken into account. The reported issue was in regard to overridden functions from an object.

arturbosch:
Ok, I changed to skip functions inside object literals :)

arturbosch:
@schalkms @Mauin still approved :)?

PR #1367, opened 2018-11-27
Use Coroutines in detekt-core

Mauin:
Decided to try coroutines in detekt as they are now final to replace the Executors/Futures we're currently using. 
I'm by no means an expert in coroutines, so if someone has suggestions for improvements I'm happy to hear about them

Upsides:
It's more Kotlin-y

Downsides:
I ran some benchmarks and it seems like the existing solution is ~5-10% faster than this new implementation using coroutines. Not sure if this is because of my local environment, the new implementation or for other reasons.


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/ProcessingSettings.kt
```diff
@@ -21,7 +19,6 @@ data class ProcessingSettings(val inputPaths: List<Path>,
 							  val parallelCompilation: Boolean = false,
 							  val excludeDefaultRuleSets: Boolean = false,
 							  val pluginPaths: List<Path> = emptyList(),
-							  val executorService: ExecutorService = ForkJoinPool.commonPool(),
```
arturbosch: Removing this is kinda critical to other applications which include detekt as an api who want run it in their own executor.
For example an in-memory use-case where only single files are analyzed. They do not need the overhead of creating #cores of threads for the pool but rather live with a SingleThreadExecutor.

I think coroutines also allow us to specify the used executor (in the background coroutines just use the common ForkJoinPool of java by themselves).

schalkms:
This is actually very interesting.
@Mauin did you benchmark the memory usage as well? What is the plus in execution time?
I don't think coroutines will replace futures. They are just a different concept in async programming.
If this performance overhead persists, I don't think we should move to coroutines.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/ProcessingSettings.kt
3flex: > I think coroutines also allow us to specify the used executor

There's an `ExecutorService.asCoroutineDispatcher()` extension function which should allow for this.

arturbosch:
Ok, also did some benchmarks with `cli` on `detekt` with:
```
time jrun detekt-cli/build/libs/detekt-cli-1.0.0-RC11-all.jar --input /home/artur/Repos/detekt --report xml:report.xml --filters ".*/resources/.*"
```

Coroutines: ~4.17 seconds (unpatched)
Coroutines: ~3.11 seconds (patched `async(Dispatchers.Default)`)
CompletableFutures: ~2.96 seconds

To me both versions do the same:
- runBlocking -> awaitAll
- processFileAsync -> runAsync
- Dispatchers.Default -> ForkJoinPool.common()

I would not change to coroutines just to be more `kotlin` and introduce a new dependency. We just use aync processing once (ok twice when parsing KtFile's in parallel with `stream().parallel()` with an extra flag).
But I would like to have the style changes in this PR into master :)

PR #1380, opened 2018-12-08
Added KotlinScriptEngine recovery

schalkms:
If a compilation error occurs the script engine is recovered.

### DIFF
--- a/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/KotlinScriptEngine.kt
+++ b/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/KotlinScriptEngine.kt
@@ -6,26 +6,40 @@ import javax.script.ScriptEngineManager
 import javax.script.ScriptException
 
 /**
+ * The object to create the Kotlin script engine for code compilation.
+ *
  * @author schalkms
  */
 object KotlinScriptEngine {
 
-	private val engine: KotlinJsr223JvmLocalScriptEngine
+	private lateinit var engine: KotlinJsr223JvmLocalScriptEngine
 
 	init {
+		createEngine()
+	}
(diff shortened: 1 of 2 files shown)


detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/KotlinScriptEngine.kt
```diff
@@ -6,26 +6,40 @@ import javax.script.ScriptEngineManager
...
+	private lateinit var engine: KotlinJsr223JvmLocalScriptEngine
 
 	init {
-		setIdeaIoUseFallback() // To avoid error on Windows
-
-		val scriptEngineManager = ScriptEngineManager()
-		val localEngine = scriptEngineManager.getEngineByExtension("kts") as? KotlinJsr223JvmLocalScriptEngine
-		requireNotNull(localEngine) { "Kotlin script engine not supported" }
-		engine = localEngine
+		createEngine()
```
arturbosch: can you just inline this call?

schalkms: I call `createEngine()` twice. However, I can declare the function as inline.

arturbosch: I mean you can return the engine in `ceateEngine` and assign it inlined with the property and in try-catch block
 ```
private var engine: KotlinJsr223JvmLocalScriptEngine = createEngine()
....
		try {
			engine.compile(code)
		} catch (e: ScriptException) {
			engine = createEngine() // recover
			throw KotlinScriptException(e)
		}
```

PR #1384, opened 2018-12-09
Build upon default config - Closes #1248

### LINKED ISSUE #1248: Build upon the default configuration
When I supply an empty configuration file, the defaults are being run.

Now I want to suppress the `MagicNumber` rule:

```
style:
  MagicNumber:
    active: false
```

It seems that detekt now runs nothing at all. Is it possible to use detekt so it only disables this one but keeps all the other defaults?
I trust in this library to provide good defaults and only want to suppress some, without copy-pasting a 500-line configuration and updating each time you change your recommendations / add rules.

arturbosch:


arturbosch:
Yep, I just wanted to open this as soon as possible to get some feedback on the new flags. I think they are good and necessary :).
However I also think about deprecated `failFast` inside the config. This should be a explicit flag and not just an option somewhere hiding in the config?

arturbosch:
Hm, the new flags are not usable from the gradle plugin until the next version ...
So split this pr for cli and gradle?

PR #1386, opened 2018-12-09
Expose compiler-embeddable as a transitive dependency

3flex:
https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation

> Dependencies appearing in the api configurations will be transitively exposed to consumers of the library, and as such will appear on the compile classpath of consumers.

This is desired for consumers of `detekt-api`.


Mauin:
Do you want me to include this change in #1350 ?

3flex:
Sure! I was hoping to clean up the explicit dependencies on compiler-embeddable throughout the rest of the build files but got distracted. These changes at least are good to go.

PR #1389, opened 2018-12-11
Fix some issues reported by IntelliJ

3flex:
Fixed some issues caught by IntelliJ's inspections.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -3,11 +3,9 @@ package io.gitlab.arturbosch.detekt
 import io.gitlab.arturbosch.detekt.extensions.DetektExtension
 import org.gradle.api.Plugin
 import org.gradle.api.Project
-import org.gradle.api.Task
 import org.gradle.api.UnknownTaskException
 import org.gradle.api.file.FileCollection
 import org.gradle.api.provider.Provider
-import org.gradle.api.tasks.TaskProvider
 import org.gradle.language.base.plugins.LifecycleBasePlugin
 
 /**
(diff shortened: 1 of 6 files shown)


docs/_layouts/default_print.html
```diff
@@ -1,6 +1,5 @@
 <!DOCTYPE html>
 <html lang="en">
-<html>
```
schalkms: Weird <html> tag here
Is this a generated file @arturbosch ?

arturbosch: It really  looks weird. It's not generated but part of the layout shiped with the template, could remove it and see if it breaks something. I don't think so  :)

arturbosch:
Ups should have merged this maybe before, sorry ^^

docs/_layouts/default_print.html
schalkms: It would be even more weird if it does break something ;)

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -3,11 +3,9 @@ package io.gitlab.arturbosch.detekt
 import io.gitlab.arturbosch.detekt.extensions.DetektExtension
 import org.gradle.api.Plugin
 import org.gradle.api.Project
-import org.gradle.api.Task
```
Mauin: Why did detekt not fail on these unused imports?

3flex: detekt isn't run on the included build. Raised https://github.com/arturbosch/detekt/issues/1393

3flex: Also when implementing fix for #1393 I found that the unused imports rule is not enabled on the detekt run... so two reasons this wasn't found by detekt.

I've submitted a PR to run detekt on the Gradle plugin, but haven't submitted a change to the config.

Issue #1395, opened 2018-12-13
Сann't connect custom rules 


Jacks0N23:
I've read an article https://medium.com/@vanniktech/writing-your-first-detekt-rule-ee940e56428d
wrote self version of custom rule and try to connect it to my project…
I waste couple of days but nothing work

here is my detekt.gradle
![image](https://user-images.githubusercontent.com/9117565/49935690-c2db0e00-fee2-11e8-8dd0-b11864bda7bb.png)

and I apply it to my project
apply from: rootProject.file('gradle/plugin/detekt.gradle')

my custom rules gradle file
![image](https://user-images.githubusercontent.com/9117565/49925295-82bb6180-fec9-11e8-8949-5fe2b6ccc02b.png)

my files hierarchy
![image](https://user-images.githubusercontent.com/9117565/49925322-96ff5e80-fec9-11e8-90d2-f6932ada3651.png)
and in RuleSetProvider i have this line: ru.tinkoff.mlca.MLCARuleSetProvider

my custom rule set provider
![image](https://user-images.githubusercontent.com/9117565/49925382-c6ae6680-fec9-11e8-9c2c-b3f4276015ae.png)

when i run :detekt I don't see my rule in ruleset list
where is my mistake?

Mauin:
Did you enable the custom rules in your `config.yml`?

Jacks0N23:
i try to enable explicit but nothing changed
but by default it should be enabled if I understand correctly sources

Mauin:
From the discussion on Slack: It seels that `detektPlugins` does not cause the project to be a buildtime dependency. Running `./gradlew :mlca-rules:build` before `./gradlew detekt` made the custom ruleset show up. However it still seems like it wasn't invoked correctly.

arturbosch:
Ok, so now your rule set is loaded but your rule does not run?
If yes, all rules are `active=false` by default. So turning this to true should enable your rule. Can you confirm this?

arturbosch:
Should we change the default behavior of `active=false` for non standard rules?

schalkms:
@arturbosch I would leave it as it is.

PR #1403, opened 2018-12-18
Add a section about how to integrate custom extensions into detekt

arturbosch:
See #1395


docs/pages/extensions.md
```diff
@@ -171,3 +171,40 @@ All they need are implementation of the `render()`-function which takes a object
...
+```
+
+##### Pitfalls
+
+- By default all rules not marked with `@active` in their `KDoc` are disabled.
+That means your custom rules are also disabled if you have not explicitly enabled
+them in the `detekt` yaml configuration file.
+- If your extension is part of your project and you integrate it like `detektPlugins project(":my-rules"")` make sure that this
+subproject is build before `gradle detekt` is run.
+In `kotlin-dsl` you could add something like `tasks.withType<Detekt> { dependsOn(":my-rules:assemble") }` so explicitly run `detekt` only
```
schalkms: In the `kotlin-dsl`

PR #1414, opened 2018-12-30
FunctionParameterNaming: Don't trigger on an underscore parameter.

### LINKED ISSUE #1412: FunctionParameterNaming warns when using _
I do think `_` is a valid parameter name. Currently, I'm in a situation where I'm using an adapter from Moshi and I need to provide two static methods that are annotated with `@FromJson` & `@ToJson` respectively. For one of the methods, I don't care about the parameter since that operation is not supported. I have to have a paremter though.

```kotlin
@JvmStatic @ToJson fun to(@IntAlignment _: Int) = throw UnsupportedOperationException()
```

The code above will trigger `UnusedPrivateMember`. If I change it to the following:

```kotlin
@JvmStatic @ToJson fun to(@IntAlignment _: Int) = throw UnsupportedOperationException()
```

FunctionParameterNaming will yell at me. Hence, is it possible that just `_` will be allowed in the `FunctionParameterNaming` rule?

vanniktech:
Fixes #1412

### DIFF
--- a/docs/pages/documentation/naming.md
+++ b/docs/pages/documentation/naming.md
@@ -136,7 +136,7 @@ Reports function parameter names which do not follow the specified naming conven
 
 #### Configuration options:
 
-* `parameterPattern` (default: `'[a-z][A-Za-z0-9]*'`)
+* `parameterPattern` (default: `'([a-z][A-Za-z0-9]*)|_'`)
 
    naming pattern
 


vanniktech:
Regex is much better! That way everyone can still tweak it.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionParameterNaming.kt
```diff
@@ -32,7 +32,7 @@ class FunctionParameterNaming(config: Config = Config.empty) : Rule(config) {
 			"Function parameter names should follow the naming convention set in the projects configuration.",
 			debt = Debt.FIVE_MINS)
 
-	private val parameterPattern by LazyRegex(PARAMETER_PATTERN, "[a-z][A-Za-z\\d]*")
+	private val parameterPattern by LazyRegex(PARAMETER_PATTERN, "([a-z][A-Za-z0-9]*)|_")
```
vanniktech: Slight consistency fix. For an idea, how to avoid this in the future I have created #1415

PR #1417, opened 2018-12-30
Build upon default config for cli module

arturbosch:
gradle plugin changes will be adapted in #1384

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
@@ -102,4 +102,12 @@ class CliArgs : Args {
 	val reportPaths: List<ReportPath> by lazy {
 		reports?.map { ReportPath.from(it) } ?: emptyList()
 	}
+
+	companion object {
+		/**
+		 * When embedding the cli inside a tool, this closure style configuration
+		 * of the arguments should be used.
+		 */
+		operator fun invoke(init: CliArgs.() -> Unit): CliArgs = CliArgs().apply(init)
+	}
 }
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
@@ -67,6 +67,16 @@ class CliArgs : Args {
 			description = "Disables default rule sets.")
 	var disableDefaultRuleSets: Boolean = false
 
+	@Parameter(names = ["--build-upon-default-config"],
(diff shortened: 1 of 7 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -67,6 +67,16 @@ class CliArgs : Args {
 			description = "Disables default rule sets.")
 	var disableDefaultRuleSets: Boolean = false
 
+	@Parameter(names = ["--build-upon-default-config"],
+			description = "Uses the default detekt configuration as a baseline. " +
```
vanniktech: I'd try not to mix up `baseline` here since we have a baseline feature but this is not what we're talking about here.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -67,6 +67,16 @@ class CliArgs : Args {
...
+					"Allows additional provided configurations to override the defaults.")
+	var buildUponDefaultConfig: Boolean = false
+
+	@Parameter(names = ["--fail-fast"],
+			description = "Uses the default detekt configuration as a baseline but sets all rules to active! " +
```
vanniktech: Same here

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt
```diff
@@ -27,18 +27,33 @@ private fun <T> String?.letIfNonEmpty(init: String.() -> List<T>): List<T> =
...
+		declaredConfig = FailFastConfig(declaredConfig ?: initializedDefaultConfig, initializedDefaultConfig)
+	}
+
+	if (debug) println("\n$declaredConfig\n")
+	return declaredConfig ?: loadDefaultConfig()
+}
+
+private fun Config.deprecatedFailFastUsage(): Boolean {
+	val value = valueOrDefault("failFast", false)
+	LOG.printer.println("Using deprecated property 'failFast' in the yaml config. " +
```
vanniktech: Why do we do this?

arturbosch: Having a cli flag and a gradle plugin property for `failFast` and `buildUponDefaultConfig` is explicit, not need to search in the yaml file. Having the yaml file mostly for rule configuration is cleaner don't you think?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
arturbosch: Right.

PR #1423, opened 2019-01-04
Respect @Suppress in UnusedPrivateMember

andrew-arnott:


andrew-arnott:
I also added support to respect suppress for unused properties.

andrew-arnott:
@arturbosch Dang -- didn't realize it was already implemented. Either way, I hope to be contributing a bit occasionally, so it was a useful exercise working with the code base.

I will update the code to use the `defaultRuleIdAliases` and look how that is used. The name made me think that it was just aliases for the rule id itself -- I didn't expect it to have anything to do with suppress annotations.

I think some of the refactor work I did is worthwhile though. Also the additional tests cases. So, maybe we can still put these changes in once I've taken out the suppress logic that I added?

PR #1436, opened 2019-01-21
Do not report main functions args parameter in objects

arturbosch:


detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -21,7 +21,6 @@ test-pattern: # Configure exclusions for test sources
     - 'ForEachOnRange'
     - 'FunctionMaxLength'
     - 'TooGenericExceptionCaught'
-    - 'InstanceOfCheckForException'
```
vanniktech: Why was this removed?

arturbosch: by accident, when I build detekt this line gets deleted, maybe need a clean build here, don't worry the last commit will get dropped :)

PR #1446, opened 2019-02-02
Fix false positives for UnnecessaryApply rule - Closes #1305

### LINKED ISSUE #1305: False Positive UnnecessaryApply
## Expected Behavior
No `UnnecessaryApply` is reported

## Observed Behavior
`UnnecessaryApply` is reported

## Steps to Reproduce
```kt
val view = FrameLayout(context).apply {
    background = ColorDrawable(ContextCompat.getColor(context, R.color.white))
}
```

## Context
Since I'm using the return value, the `apply` isn't unnecessary.

## Your Environment
* Version of detekt used: 1.0.0-RC10
* Version of Gradle used (if applicable): 4.10.2
* Operating System and version: Various

arturbosch:


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApply.kt
```diff
@@ -40,23 +46,38 @@ class UnnecessaryApply(config: Config) : Rule(config) {
...
+			!isInsideFunctionCall() &&
+					hasOnlyOneStatement() &&
+					!isInsideAssignment() &&
+					!isReturnedFromFunction()
+}
+
+private fun KtCallExpression.hasOnlyOneStatement(): Boolean {
+	val lambdaBody = firstLambdaArg?.bodyExpression
+	if (lambdaBody.hasOnlyOneStatement()) {
+		return lambdaBody?.statements?.firstOrNull()
```
schalkms: Can't we say just `lambdaBody.statements[0]` since we know from the if check that there can only be one element in the `statements` array?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApply.kt
```diff
@@ -40,23 +46,38 @@ class UnnecessaryApply(config: Config) : Rule(config) {
...
+	private fun KtCallExpression.isViolation() =
+			!isInsideFunctionCall() &&
+					hasOnlyOneStatement() &&
+					!isInsideAssignment() &&
+					!isReturnedFromFunction()
+}
+
+private fun KtCallExpression.hasOnlyOneStatement(): Boolean {
+	val lambdaBody = firstLambdaArg?.bodyExpression
+	if (lambdaBody.hasOnlyOneStatement()) {
```
schalkms: Is there no need for the safe call operator?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApply.kt
arturbosch: Yep, but only with contracts :)

arturbosch:
Yep, I have tested it with all analysis-projects (via script), and some custom projects. Found some true positives and no false positives now :).

PR #1454, opened 2019-02-05
Exclude inline classes as they are a light form of data class - Closes  #1450

### LINKED ISSUE #1450: UseDataClass triggers on inline classes
The following Kotlin class gets a `UseDataClass` codesmell.

 ```kotlin
inline class PlayerId(val value: Long)
```

Inline classes should be excluded.

arturbosch:


vanniktech:
Thank you :heart:

PR #1455, opened 2019-02-05
Wrap four new KtLint rules

arturbosch:


arturbosch:
Thanks for the reminder, I even forgot to put them into the provider which brings me to the idea to `generate` the providers! (will open an issue)
Also the annotation rule allows two annotations on the same line (when no other source tokens are present), need to ask them if this is intentional. Nothing we can do now about it :).

arturbosch:
Seems like a KtLint rule bug.

3flex:
Before enabling new ktlint rules, check if it's enabled upstream at https://github.com/shyiko/ktlint/blob/master/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/StandardRuleSetProvider.kt

detekt can still wrap them but should disable by default in `default-detekt-config.yml`

AnnotationRule is disabled in ktlint so should probably not expect it to work properly at the moment.

arturbosch:
Thanks for the link and the suggestion, will do it.

detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -277,6 +279,8 @@ formatting:
   SpacingAroundCurly:
     active: true
     autoCorrect: true
+  SpacingAroundDot:
+    active: false
```
3flex: This one could be enabled? It's enabled by ktlint. Of these two rules only `AnnotationRule` is disabled by default in ktlint, but this one should be OK to turn on like the rest of the formatting rules.

arturbosch: It is actually a deal breaker, see https://github.com/shyiko/ktlint/issues/344 I opened just now ... :(

arturbosch:
Also the new rule `SpacingAroundDot` interferes with `ChainWrapping`:
```
❯ diff /tmp/diff1 /tmp/diff2
3c3,5
<     val anchor = owner.firstChild!!.siblings(forward = true).dropWhile { it is PsiComment || it is PsiWhiteSpace }
---
>     val anchor = owner.firstChild!!
>         .siblings(forward = true)
>         .dropWhile { it is PsiComment || it is PsiWhiteSpace }

```

detekt-cli/src/main/resources/default-detekt-config.yml
3flex: Fix was released in klint 0.31.0

3flex:
There's a new rule in 0.33: SpacingAroundUnaryOperatorsRule

Also MultiLineIfElseRule is not wrapped.

There's also a new indentation rule in the experimental ruleset which may be worth wrapping also (ImportOrderingRule from experimental ruleset is already wrapped).

arturbosch:
So this wraps four new rules. The codfactor issues are test resources. Need to investigate how to suppress them ^^.

arturbosch:
CI fails due to https://github.com/pinterest/ktlint/issues/490
Should we disable the `Annotation` rule and get this merged?

schalkms:
@arturbosch yes

arturbosch:
Well I guess KtLint formats the file differently on java8 and java9+ ...

schalkms:
Probably a bug?

PR #1462, opened 2019-02-09
Prepare RC13

arturbosch:


arturbosch:
Sure, we have more than I ever wanted to implement for a 1.0. Idk what should be missing? I want to start making minor releases ^^.

vanniktech:
There have been so many API changes in the last few releases. (Including from RC12 to 1.0.0)
I would not be confident calling this a 1.0.0 but also again I'm not a real maintainer.

What's the benefit for having a 1.0.0 instead of a RC13? Personally, I don't see one.

arturbosch:
Actually your right, things like LargeClass and LongMethod rework need a RC13 because they are active by default. Will change to RC13.

PR #1470, opened 2019-02-11
Spek 2

3flex:
Update to Spek 2. Mostly straightforward, except:

* `subject` extension was removed. Followed these instruction to convert: https://spekframework.org/migration/#subjects
* `context` and `it` are not allowed in the root scope of a test, so wrapped in `describe` where appropriate
* converting existing tests that used `given` to the new `given`, `on`, `it` style was going to be a bit of work, so in these cases changed to `describe` instead

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -24,9 +24,9 @@ tasks.withType<Wrapper> {
     distributionType = Wrapper.DistributionType.ALL
     doLast {
         /*
-		 * Copy the properties file into the detekt-gradle-plugin project.
-		 * This allows IDEs like IntelliJ to import the detekt-gradle-plugin as a standalone project.
-		 */
+         * Copy the properties file into the detekt-gradle-plugin project.
+         * This allows IDEs like IntelliJ to import the detekt-gradle-plugin as a standalone project.
+         */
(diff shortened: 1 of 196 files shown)


3flex:
Failing on `LargeClass - 821/600 - [UnusedPrivateMemberSpec] at /home/travis/build/arturbosch/detekt/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt:12:1`

Which is strange because that file is one line shorter than before, so not sure why failing in the PR but not on master?

schalkms:
Thanks for this!
That's weird. I also noticed that not all tests are executed. At least the AppVeyor [history](https://ci.appveyor.com/project/arturbosch/detekt/history) tells me that.

schalkms:
@3flex do you have any idea why the coverage is going down?
Furthermore, it seems that there are fewer tests?

3flex:
Looks good now, with no change in coverage which was the goal. I think the problem was that I'd set the same value for the description in multiple `description() {` blocks. Once I changed them so they were different the coverage improved.

May be a bug or expected behaviour in Spek 2 but at least it's resolved here.

PR #1479, opened 2019-02-13
Remove failFast from default-detekt-config.yml

BraisGabin:
`failFast` is now deprecated.


arturbosch:
Hey, thanks to getting to this. We are generating our config and documentation stuff, so just changing this in the config does not help. The right place to change this would be `detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/ConfigPrinter.kt#defaultGenericConfiguration:67`

BraisGabin:
Fixed.

PR #1481, opened 2019-02-14
Bugfix: Deprecation warning if using on Gradle >= 5.0

julianwki:
The given condition is wrong (see commit). If Gradle >=5.0 is found, it uses the deprecated method and vice versa. That leads to the message "Deprecated Gradle features were used in this build, making it incompatible with Gradle 6.0".


julianwki:
@3flex How will #1482 close this pull request?

PR #1482, opened 2019-02-14
Test Gradle plugin on multiple Gradle versions

### LINKED ISSUE #1475: RC13 doesn't work on gradlew 4.9
The error I get is:

```
* What went wrong:
A problem occurred configuring project ':x'.
> Could not create task ':x:detekt'.
   > Could not create task of type 'Detekt'.
      > org.gradle.api.model.ObjectFactory.fileProperty()Lorg/gradle/api/file/RegularFileProperty;
```

I am using detekt in an Android app. 

Based on https://developer.android.com/studio/releases/gradle-plugin the latest recommended gradlew version for Android dev is 4.10.1+.

I am using detekt RC12 with gradlew version 4.9 without issues.

To verify this bug, I kept on increasing gradlew version until I found a version which worked. 4.10.1 and 4.10.2 do not work. Gradlew 5.1 worked.

3flex:
Adds a simple test so the Gradle plugin is tested on multiple Gradle versions - the earliest and latest supported versions (4.9 and 5.2.1 currently). Will hopefully avoid regressions in Gradle support going forward.

Also adds a fix so the tests pass.

Fixes #1475 
Closes #1481


detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/MultiVersionTest.kt
```diff
@@ -0,0 +1,28 @@
...
+import io.gitlab.arturbosch.detekt.DslTestBuilder.Companion.groovy
+import io.gitlab.arturbosch.detekt.DslTestBuilder.Companion.kotlin
+import org.assertj.core.api.Assertions.assertThat
+import org.gradle.testkit.runner.TaskOutcome
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+
+object MultiVersionTest : Spek({
+
+    val testedGradleVersions = listOf("4.9", "5.2.1")
```
schalkms: Is it enough to just test these two versions?

Macarse: I would add the latest Android Gradle plugin recommended version, which is `4.10.1`

3flex: Could go either way - I wanted to keep the range small because each version has to be downloaded if not already on the machine which takes time.

I think having just the two versions will be enough to address the problem of introducing incompatible changes - maybe try with these two and if for some reason another has to be added then include it later?

Macarse: sure! 👍

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DslGradleRunner.kt
```diff
@@ -1,16 +1,17 @@
 package io.gitlab.arturbosch.detekt
 
-import java.io.File
-import java.util.UUID
 import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.GradleRunner
+import java.io.File
+import java.util.UUID
```
arturbosch: Hm KtLint's and IntelliJ default behavior for imports is not the same.
Should we deactivate KtLint's and leave detekt's unused import rule active, so contributors do not have to set up their IDE to match KtLints default?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -58,7 +57,7 @@ open class Detekt : DefaultTask() {
     @InputFile
     @Optional
     @PathSensitive(PathSensitivity.RELATIVE)
-    var baseline: RegularFileProperty = createNewInputFile()
+    var baseline: RegularFileProperty = project.layout.fileProperty()
```
arturbosch: Why not using our custom `GradleCompat#Project.fileProperty()` function?

3flex: Fixes #1475. You can try running the new multi version test without removing that custom function, but it won't pass on Gradle 4.9

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DslGradleRunner.kt
3flex: Sounds good - I'm not sure why ktlint takes a stance on import ordering anyway since official Kotlin coding conventions don't specify import ordering rules.

Won't include that change in this PR though

PR #1484, opened 2019-02-15
Remove project cache dir workaround from tests

3flex:
This was originally removed in #929 but brought back at some point.

Will see if it passes in CI, if not, will investigate.

### DIFF
--- a/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskDslTest.kt
+++ b/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskDslTest.kt
@@ -205,7 +205,7 @@ internal class DetektTaskDslTest : Spek({
 
                     val config = """
 						|detekt {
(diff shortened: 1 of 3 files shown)


3flex:
This shows that as long as the input is set correctly, and doesn't include Gradle's project cache, then the error in #807 (Failed to create MD5 hash for file content) will be avoided.

The example at https://arturbosch.github.io/detekt/gradletask.html suggests setting `input` to `$projectDir`, so if that were updated to `$projectDir/src` then that might help steer people towards the right path. Other examples already suggest using `/src` for input, and the plugin defaults to `/src/main/kotlin` and `/src/main/java`.

Setting `filters` has no effect since the problem is in how Gradle processes the files supplied as `input`, and `filters` only takes effect when detekt actually runs. Changing the task to be a `SourceTask` and using Gradle's built-in filters will help.

arturbosch:
Thanks for looking at it again. 
So this basically needs two further actions:
- improve the task documentation to prevent #807 
- look at SourceTask

PR #1488, opened 2019-02-16
Add ignoreOverridden option for TooManyFunctions rule - #1411

### LINKED ISSUE #1411: TooManyFunctions flags overridden functions
This is a bit of a dilemma. I'm subclassing an external class and I'm forced to override 15 methods. By doing this `TooManyFunctions` gets triggered. It would be nice if it weren't triggered.

Is there anything we can do?

IvanAntsiferov:
closes #1411

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -117,6 +117,7 @@ complexity:
     thresholdInEnums: 11
     ignoreDeprecated: false
     ignorePrivate: false
+    ignoreOverridden: false
 
 empty-blocks:
   active: true
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
@@ -8,6 +8,7 @@ import io.gitlab.arturbosch.detekt.api.Metric
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.ThresholdedCodeSmell
+import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.KtClass
 import org.jetbrains.kotlin.psi.KtClassOrObject
 import org.jetbrains.kotlin.psi.KtFile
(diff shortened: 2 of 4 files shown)


detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -117,6 +117,7 @@ complexity:
     thresholdInEnums: 11
     ignoreDeprecated: false
     ignorePrivate: false
+    ignoreOverridden: false
```
vanniktech: I think this should be `true` by default.

schalkms: Thanks for the contribution. Please remove this config option and just ignore enums by default. Otherwise, the code looks fine. Thanks for your contribution! :)

IvanAntsiferov: @schalkms Did you mean ignore overriden functions by default without additional rule option?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
```diff
@@ -124,6 +127,7 @@ class TooManyFunctions(config: Config = Config.empty) : Rule(config) {
     private fun isIgnoredFunction(function: KtNamedFunction): Boolean = when {
         ignoreDeprecated && function.annotationEntries.any { it.typeReference?.text == DEPRECATED } -> true
         ignorePrivate && function.isPrivate() -> true
+        ignoreOverridden && (function.modifierList?.hasModifier(KtTokens.OVERRIDE_KEYWORD) ?: false) -> true
```
arturbosch: Please use `function.isOverride()` here.

detekt-cli/src/main/resources/default-detekt-config.yml
arturbosch: I also don't understand what @schalkms wrote :). Active by default is okay for me ^^

schalkms: I am really sorry. I mixed it up with the proposal to ignore enum classes in #1489
Everything is fine.

arturbosch: Well maybe it shouldn't be true by default. Better to be strict by default and relaxed when needed.
It also aligns with the other ignore options.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
IvanAntsiferov: I did not notice that the project already has an extension function for that.

arturbosch: No problem, thanks for contributing :)!

PR #1510, opened 2019-02-24
Ensure IntelliJ is configured to use official Kotlin style

3flex:
Uses tip from https://kotlinlang.org/docs/reference/code-style-migration-guide.html#migration-to-a-new-code-style

https://github.com/arturbosch/detekt/commit/445ffb0c55b0b3af830cea363d42f1730d9642b3 reformatted the code base to use spaces instead of tabs for indentation, however there is still much code that is not formatted according to the latest Kotlin code style. This flag will ensure that new code edited in IntelliJ/Android Studio will adhere to the new style without modifying existing code.

If @arturbosch prefers I'll also add a commit that will have IntellIJ automatically migrate code to the official style.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,5 +10,7 @@ kshVersion=0.1.1
 reflectionsVersion=0.9.11
 jacocoVersion=0.8.3
 
+kotlin.code.style=official
+
 systemProp.sonar.host.url=http://localhost:9000
 systemProp.detektVersion=detektVersion


3flex:
I don't think this will conflict with any ktlint items - seems the only indentation rules they enforce is 4 space indentation (which is already applied here). This example from Kotlin website shows what most of the remaining changes are, and I don't think either detekt or ktlint has rules to check for new or old style. This is only about setting IDE default but this style won't actually be enforced by tooling as far as I'm aware.

image

So can set this flag so new code should follow the official Kotlin style and have code updated over time, or can set the flag and also do an auto-format across the code base.

I don't know if it affects comments, and from what I've seen it doesn't affect multi-line string literals either.

PR #1519, opened 2019-03-02
Update dependencies

3flex:


arturbosch:
Any idea why `detekt-rules:test` was not run in this PR ?

3flex:
The tests weren't running as Spek started failing during its test discovery phase (raised https://github.com/spekframework/spek/issues/642).

I've removed the test from detekt which allows the tests to run. Removing the test doesn't change the test coverage so may not have value anyway, but if you prefer I can hold Spek back to 2.0.0 until the bug in Spek is hopefully fixed and leave the test as is.

PR #1520, opened 2019-03-02
Single statement in nullable apply calls 

arturbosch:


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
```diff
@@ -25,16 +25,29 @@ class UnnecessaryApplySpec : Spek({
 			""")).hasSize(1)
             }
 
-            it("reports a false negative apply on nullable type") {
+            it("reports an apply on nullable type") {
                 assertThat(subject.lint("""
 				fun f() {
 					val a : Int? = null
 					// Resolution: we can't say here if plus is on 'this' or just a side effects when a is not null
```
schalkms: a side effect

PR #1523, opened 2019-03-04
Configure build-scan plugin

aalmiray:
This small PR configures the `build-scan` Gradle plugin (https://guides.gradle.org/creating-build-scans/) which yields further insight into the build, allowing the team to make informed decisions on where and when the build may be tweaked to get faster, more performant builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,14 +18,14 @@ matrix:
       osx_image: xcode10.2
 install: true
 script:
-  - ./gradlew build --build-cache -PwarningsAsErrors=true
+  - ./gradlew build --build-cache -PwarningsAsErrors=true --scan
   - ./gradlew shadowJar
   - java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar --help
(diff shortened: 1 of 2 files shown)


aalmiray:
Not really, no. The reports are for your own consumption and determine if any fixes may be needed.

3flex:
I was crossing my fingers that maybe the build scan plugin now handled composite builds, but sadly the log says:

> Build scans are not supported for this build due to a build plugin composite substitution.

So this won't actually do anything. @aalmiray as you work at Gradle perhaps you'd be able to escalate this somehow? The scans offer good insights so it would be great to have them working in this project.

aalmiray:
Would it help in the meantime to apply scans directly to the `detekt-gradle-plugin`?

3flex:
We'd get much more value being able to apply it to the root project which has a much more complex build, but applying to the plugin is isolation would still provide some insight.

arturbosch:
@3flex @aalmiray Does this mean applying the build scan to the root does not generate any report at all because of composite builds or just without the gradle plugin?

3flex:
@aalmiray I see that v2.3 of the Build Scan plugin supports plugin composite builds, and a quick local test shows the build scan being created and uploaded. Would be good to see this go in!

aalmiray:
Rebased with latest from master

PR #1532, opened 2019-03-06
Type and symbol resolving

3flex:
This PR adds type and symbol resolving for Kotlin code that targets the JVM, seeking early feedback. It's a WIP until at a minimum the following tests/features are added:

- [x] Make the functionality optional - it's slow, so want to disable by default unless a value passed to `--classpath` parameter
- [ ] Tests for the new functionality in a mixed Java/Kotlin codebase
- [x] Tests for the new `UselessCallOnNotNull` rule
- [x] Sharing compiler errors/warnings with the user (they are not currently shown, so user has to know that their code compiles before running detekt, otherwise there'll be an error)

Kotlin/JS and Kotlin/Native will not be in scope, so naturally won't look at MPP support either. Will tackle Gradle plugin integration in future but not as part of this PR.

To test, enable the `UselessCallOnNotNull` rule in the config file, add the following test file with a `kts` extension, then run with `--input path/to/testfile.kts --classpath path/to/kotlin-stdlib.jar --config path/to/config.yml`.
```kotlin
val nextTest = mutableListOf<String>().orEmpty()
```
The new rule will resolve the `orEmpty()` call to `kotlin.collections.orEmpty` which will trigger a detekt issue because the `orEmpty()` call is never needed on a not-null collection.

This is an updated version of #485, rebased and with a simplified implementation (and the intention to get it merged this time! :smile:).

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
@@ -1,6 +1,7 @@
 package io.gitlab.arturbosch.detekt.api
 
 import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.resolve.BindingContext
 
 /**
  * The type to use when referring to rule ids giving it more context then a String would.
@@ -24,12 +25,18 @@ abstract class BaseRule(
 ) : DetektVisitor(), Context by context {
 
     open val ruleId: RuleId = javaClass.simpleName
+    var bindingContext: BindingContext = BindingContext.EMPTY
 
     /**
      * Before starting visiting kotlin elements, a check is performed if this rule should be triggered.
      * Pre- and post-visit-hooks are executed before/after the visiting process.
+     * [BindingContext] holds the result of the semantic analysis of the source code by the Kotlin compiler. Rules that
+     * rely on symbols and types being resolved can use the BindingContext for this analysis. Note that detekt must
+     * receive the correct compile classpath for the code being analyzed otherwise the default value
+     * [BindingContext.EMPTY] will be used and it will not be possible for detekt to resolve types or symbols.
      */
-    fun visitFile(root: KtFile) {
+    fun visitFile(root: KtFile, bindingContext: BindingContext = BindingContext.EMPTY) {
+        this.bindingContext = bindingContext
         if (visitCondition(root)) {
             clearFindings()
             preVisit(root)
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt
@@ -23,6 +23,8 @@ fun CliArgs.createPlugins(): List<Path> = plugins.letIfNonEmpty {
     MultipleExistingPathConverter().convert(this)
 }
 
+fun CliArgs.createClasspath(): List<String> = classpath.letIfNonEmpty { split(";") }
+
 private fun <T> String?.letIfNonEmpty(init: String.() -> List<T>): List<T> =
         if (this == null || this.isEmpty()) listOf() else this.init()
 
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
@@ -1,0 +1,87 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtQualifiedExpression
+import org.jetbrains.kotlin.psi.KtSafeQualifiedExpression
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.calls.callUtil.getType
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+import org.jetbrains.kotlin.types.TypeUtils
+
+/**
+ * The Kotlin stdlib provides some functions that are designed to operate on references that may be null. These
+ * functions can also be called on non-nullable references or on collections or sequences that are known to be empty -
+ * the calls are redundant in this case and can be removed or should be changed to a call that does not check whether
+ * the value is null or not.
+ *
+ * Rule adapted from Kotlin's IntelliJ plugin: https://github.com/JetBrains/kotlin/blob/f5d0a38629e7d2e7017ee645dc4d4bee60614e93/idea/src/org/jetbrains/kotlin/idea/inspections/collections/UselessCallOnNotNullInspection.kt
+ *
+ * <noncompliant>
+ * val testList = listOf("string").orEmpty()
+ * val testList2 = listOf("string").orEmpty().map { _ }
+ * val testString = ""?.isNullOrBlank()
+ * </noncompliant>
+ *
+ * <compliant>
+ * val testList = listOf("string")
+ * val testList2 = listOf("string").map { _ }
+ * val testString = ""?.isBlank()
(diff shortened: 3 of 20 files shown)


schalkms:
Hey! Thanks for the tremendous work you spent on this one. 🥇 
I'll review it very soon.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Configurations.kt
```diff
@@ -23,6 +23,10 @@ fun CliArgs.createPlugins(): List<Path> = plugins.letIfNonEmpty {
     MultipleExistingPathConverter().convert(this)
 }
 
+fun CliArgs.createClasspath(): List<String> = classpath.letIfNonEmpty {
+    this.split(";")
```
arturbosch: nit: on the same line it would look better :)

arturbosch:
Yeah, thanks you very much for picking this up.
Looks good, will clone your branch and see how the performance is ^^

3flex:
Performance will be poor. It's effectively compiling the application (or at least all the Kotlin code) before running the rule sets.

Could probably do something creative in the intellij plugin which allows hooking into the lazy resolve capabilities built into the IDE, which is how the Kotlin plugin runs it's checks.

But to run on the CLI in a self contained tool this is the only way I'm aware of. The value of the implemented rules will have to outweigh the performance cost.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
```diff
@@ -24,12 +25,18 @@ abstract class BaseRule(
...
 
     open val ruleId: RuleId = javaClass.simpleName
+    var bindingContext: BindingContext = BindingContext.EMPTY
 
     /**
      * Before starting visiting kotlin elements, a check is performed if this rule should be triggered.
      * Pre- and post-visit-hooks are executed before/after the visiting process.
+     * BindingContext holds the result of the semantic analysis of the source code by the Kotlin compiler. Rules that
+     * rely on symbols and types being resolved can use the BindingContext for this analysis. Note that
+     * BindingContext will have the value BindingContext.EMPTY unless detekt is provided the correct classpath while
```
schalkms: is providing

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
```diff
@@ -0,0 +1,78 @@
...
+import org.jetbrains.kotlin.resolve.calls.callUtil.getType
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+import org.jetbrains.kotlin.types.TypeUtils
+
+/**
+ * TBC
+ *
+ * @author Matthew Haughton
+ */
+class UselessCallOnNotNull(config: Config = Config.empty) : Rule(config) {
```
schalkms: I'll review this rule when it's finished.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
3flex: It's trying to say that the classpath must be provided to detekt, which then provides the BindingContext to the rule.

I'll reword it though to try and make that clearer.

3flex:
I'm pretty happy where this is at - appreciate a closer review.

The main new feature since original submission is the fact that errors are shown when the input is being evaluated and the code cannot compile for some reason (and would thus not provide valid results when attempting type and symbol resolution). This will help when not all required dependencies are included on the classpath input.

Mixed Kotlin/Java inputs will also work if both `.kt` and `.java` files (or paths containing them) are provided to the `--input` flag.

Test for the new rule is there, but how to go about efficiently testing the additions in detekt-core and detekt-api are a bit beyond me. When integration with Gradle is added there'll be some good testing done from there so maybe that will be enough?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.jetbrains.kotlin.psi.KtQualifiedExpression
+import org.jetbrains.kotlin.psi.KtSafeQualifiedExpression
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.calls.callUtil.getType
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+import org.jetbrains.kotlin.types.TypeUtils
+
+/**
+ * TBC
```
schalkms: Rule documentation is missing

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @author Matthew Haughton
+ */
+class UselessCallOnNotNull(config: Config = Config.empty) : Rule(config) {
+    override val issue: Issue = Issue(
+        "UselessCallOnNotNull",
+        Severity.Performance,
+        "This call on not-null type may be reduced or removed. Some calls are intended to be called on nullable " +
```
schalkms: on a none-nullable type

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
```diff
@@ -0,0 +1,73 @@
...
+                "collection or text types (e.g. String?). When this call is used on a not-null type (e.g. String) " +
+                "it is redundant and will have no effect, so it can be removed.",
+        Debt.FIVE_MINS
+    )
+
+    // Rule adapted from the Kotlin plugin for IntelliJ:
+    // https://github.com/JetBrains/kotlin/blob/f5d0a38629e7d2e7017ee645dc4d4bee60614e93/idea/src/org/jetbrains/kotlin/idea/inspections/collections/UselessCallOnNotNullInspection.kt
```
schalkms: This should be moved to the rule documentation

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
```diff
@@ -0,0 +1,73 @@
...
+        "kotlin.text.isNullOrEmpty" to Conversion("isEmpty"),
+        "kotlin.text.isNullOrBlank" to Conversion("isBlank")
+    )
+
+    private val uselessNames = toShortFqNames(uselessFqNames.keys)
+
+    @Suppress("ReturnCount")
+    override fun visitQualifiedExpression(expression: KtQualifiedExpression) {
+        super.visitQualifiedExpression(expression)
+        if (bindingContext == BindingContext.EMPTY) return
```
schalkms: Do we need this check on every usage of the `BindingContext`?

3flex: Not strictly - the `getResolvedCall` call will fail further down with empty `BindingContext`, but checking for it sooner will save a bit of work.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
3flex: Might go with "non-null reference" as that matches Kotlin docs (https://kotlinlang.org/docs/reference/null-safety.html), what do you think?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
arturbosch: We could introduce a new rule type which extends its condition check to run only when `bindingContext != BindingContext.EMPTY`?

arturbosch:
Works great!. I've tested this on a fixed mini-project:
```
fun main() {
    val javaList = Java().list.orEmpty()
    val list = listOf(1, 2, 3).orEmpty()
}
```
and it only reports the kotlin list line :)
Could you add such a test and we are ready to merge this :)?

3flex:
Done!

I didn't add a new rule type as you suggested in your comment, but perhaps that can be added as a later enhancement. There's no interface that the base rule could implement that would ensure the `bindingContext != BindingContext.EMPTY` line was checked for every element visited which seemed like the right way to go, so because the Kotlin compiler doesn't offer that it might get out of date over time. Perhaps you were thinking another way?

arturbosch:
I was thinking of adding a rule type (or just a property for `BaseRule`) where the rule implementor specifies if this rule will only run when `classpath` is set. The `detekt-core` should then take care of running these rules ^^

PR #1541, opened 2019-03-09
Remove non popular watcher module

arturbosch:
- faster build times
- not used by others but me :)
- more energy towards intellij and sonar plugins


arturbosch:
because no bug report or dependency updates get contributed ;)

PR #1543, opened 2019-03-11
ktlint 0.31.0

3flex:


3flex:
SpacingAroundDotRule wrapper is introduced in #1455, so didn't include here.

arturbosch:
Thanks for "just" updating the version :). The two new rules are still buggy, so we won't support them for now.

PR #1547, opened 2019-03-19
Simplify Java interop

saschpe:
Part of an attempt to get it running with Spotless (https://github.com/saschpe/spotless/tree/detekt)


3flex:
I might be slightly off base here, but as I understand it:
* Spotless is a code formatter
* detekt doesn't format code (though it wraps ktlint which does)

Assuming both statements are true I think it would be better to wrap ktlint instead of detekt.

schalkms:
@3flex I think the point here is that detekt supports configuration options for the ktlint ruleset.
Here is the referenced issue from the commit: https://github.com/diffplug/spotless/issues/143

PR #1549, opened 2019-03-21
Update Gradle & dependencies

3flex:
Update Gradle and all dependencies to latest, except ktlint (#1543) and JCommander which threw an exception when updated to 1.75.


build.gradle.kts
```diff
@@ -13,12 +13,12 @@ plugins {
...
     id("org.sonarqube") version "2.7"
     id("io.gitlab.arturbosch.detekt")
-    id("org.jetbrains.dokka") version "0.9.17"
+    id("org.jetbrains.dokka") version "0.9.18"
     jacoco
 }
 
-tasks.withType<Wrapper> {
-    gradleVersion = "5.2.1"
+tasks.wrapper {
```
arturbosch: Oh nice, can the kotlin-dsl now figure this out by names, or just extension function for common tasks?

3flex: It creates synthetic accessors for task names, so `tasks.taskName`

If there's only one task that needs configuring this is the way to do it, but if you need to do all tasks of a certain type (e.g. with KotlinCompile tasks) you would stick with the current syntax.

PR #1567, opened 2019-03-28
Allow configuration to ignore magic number in ranges.

### LINKED ISSUE #1368: False-positive MagicNumber when using ranges
## Expected Behavior
A range like ` internal val SOME_RANGE = 0..7` should not be reported as MagicNumber,
as it is stored as extra value and not somewhere in the code.

## Observed Behavior
A range like the example above is reported as MagicNumber, whereas a range like `val OTHER_RANGE=0..2` is not.

## Steps to Reproduce
1) Configure a Gradle build with detekt
2) Create some Kotlin file with two ranges in it (where one needs a end with a value greater than 2)
3) Run detekt. The range with the higher end will be reported

## Context
The extraction of the range should provide a clean specification of substring ranges during the construction of a complex string.

## Your Environment
* Version of detekt used: 1.0.0-RC11
* Version of Gradle used (if applicable): 4.8
* Operating System and version: Fedora 29

marschwar:
Closes #1368 

This allows the user to ignore magic numbers in ranges. By default these occurrences are reported.

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumberSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumberSpec.kt
@@ -582,7 +582,7 @@ class MagicNumberSpec : Spek({
 			""")
 
                 it("should detect the argument") {
-                    val rule = MagicNumber(TestConfig(mapOf("ignoreNamedArgument" to "true")))
+                    val rule = MagicNumber(TestConfig(mapOf(MagicNumber.IGNORE_NAMED_ARGUMENT to "true")))
                     assertThat(rule.lint(code("53"))).hasSize(1)
                 }
             }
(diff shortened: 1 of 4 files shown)


arturbosch:
+1, feel free to suppress `LargeClass` in the Spec.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumberSpec.kt
```diff
@@ -582,7 +583,7 @@ class MagicNumberSpec : Spek({
 			""")
 
                 it("should detect the argument") {
-                    val rule = MagicNumber(TestConfig(mapOf("ignoreNamedArgument" to "true")))
+                    val rule = MagicNumber(TestConfig(mapOf(MagicNumber.IGNORE_NAMED_ARGUMENT to "true")))
```
schalkms: good

PR #1570, opened 2019-03-31
Add rules to suggest usage of check(), require() and error().

### LINKED ISSUE #1502: New rule: Use require() and check() instead of explicitly throwing exceptions
Several variations, some will be more complex to implement than others if attempting to identify all uses of (this list isn't exhaustive):
* use `error()` in place of `throw IllegalStateException`
* use `check()` in place of a conditional `IllegalStateException`
* use `checkNotNull()` in place of a null check conditional `IllegalStateException`
* use `require()` in place of a conditional `IllegalArgumentException`
* use `requireNotNull()` in place of a null check conditional `IllegalArgumentException`

To simplify implementation the rule could flag any use of any `IllegalStateException` and `IllegalArgumentException` and advise to use one of error, check, checkNotNull or require, requireNotNull respectively.

## Context
Inspired by https://www.bignerdranch.com/blog/write-better-code-using-kotlins-require-check-and-assert/

marschwar:
Closes #1502 

Since this is my first rule I am glad to get feedback and am open for alternative naming suggestions.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/ThrowExtensions.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/ThrowExtensions.kt
@@ -1,0 +1,19 @@
+package io.gitlab.arturbosch.detekt.rules
+
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtThrowExpression
+import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
+import kotlin.reflect.KClass
+
+internal fun KtThrowExpression.isIllegalStateException() =
+    isExceptionOfType(IllegalStateException::class)
+
+internal fun KtThrowExpression.isIllegalArgumentException() =
+    isExceptionOfType(IllegalArgumentException::class)
+
+internal fun <T : Exception> KtThrowExpression.isExceptionOfType(clazz: KClass<T>): Boolean {
+    return findDescendantOfType<KtCallExpression>()?.firstChild?.text == clazz.java.simpleName
+}
+
+internal val KtThrowExpression.argumentCount
+    get() = findDescendantOfType<KtCallExpression>()?.valueArgumentList?.children?.size ?: 0
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseCheckOrError.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseCheckOrError.kt
@@ -7,11 +7,11 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.rules.argumentCount
+import io.gitlab.arturbosch.detekt.rules.isIllegalStateException
 import org.jetbrains.kotlin.psi.KtBlockExpression
-import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtFunctionLiteral
 import org.jetbrains.kotlin.psi.KtThrowExpression
-import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
 
 /**
  * Kotlin provides a much more concise way to check invariants as well as pre- and post conditions than to manually throw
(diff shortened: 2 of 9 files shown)


marschwar:
I have not done it yet, but will definitely do so. Thanks for the tip. Are you worried about false positives or false negatives? I can't really think of any circumstance where it is better to throw an IllegalArgumentException or IllegalStateException directly.  But I guess I will see.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseCheckOrErrorSpec.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    describe("UseCheckOrError rule") {
+
+        it("reports if a an IllegalStateException is thrown") {
+            val code = """
+				fun x() {
+					doSomething()
+                    if (a < 0) throw IllegalStateException()
+				}"""
+            Assertions.assertThat(subject.lint(code)).hasSize(1)
```
3flex: nit: please import `org.assertj.core.api.Assertions.assertThat` so these can change to `assertThat(...)` instead of `Assertions.assertThat(...)`

schalkms:
Regarding the symbol resolution, I would say we need to go over all rules and see where we can apply it after the PR is merged.

marschwar:
I just ran a check on most of those analysis projects and there were a lot of occurrences where the rule correctly reports a violation. I did find 2 cases with possible false positives though.

Creating a lambda that throws an IllegalArgumentException I would think is an edge case. But I could try to detect this.
```
  fun unsafeRunSync(): A =
    unsafeRunTimed(Duration.INFINITE)
      .fold({ throw IllegalArgumentException("IO execution should yield a valid result") }, ::identity)
```

This I consider more relevant. Throwing an exception with a cause should never report an issue.
```
    private fun missing(): Nothing {
        if  (cause == null) {
            throw IllegalStateException(
                "Module with the Main dispatcher is missing. " +
                        "Add dependency providing the Main dispatcher, e.g. 'kotlinx-coroutines-android'"
            )
        } else {
            val message = "Module with the Main dispatcher had failed to initialize" + (errorHint?.let { ". $it" } ?: "")
            throw IllegalStateException(message, cause)
        }
    }
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseRequire.kt
```diff
@@ -42,6 +42,9 @@ class UseRequire(config: Config = Config.empty) : Rule(config) {
     }
 
     private fun KtThrowExpression.isIllegalArgumentException(): Boolean {
-        return findDescendantOfType<KtCallExpression>()?.firstChild?.text == "IllegalArgumentException"
+        val callExpression = findDescendantOfType<KtCallExpression>()
```
schalkms: Code duplication here
Can we group this together and pass the type of the exception?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseRequire.kt
```diff
@@ -42,6 +42,9 @@ class UseRequire(config: Config = Config.empty) : Rule(config) {
...
+        val argumentCount = callExpression?.valueArgumentList?.children?.size ?: 0
```
schalkms: Code duplication here
Can we group this together and pass the type of the exception?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseCheckOrError.kt
```diff
@@ -44,12 +44,15 @@ class UseCheckOrError(config: Config = Config.empty) : Rule(config) {
...
-        if (expression.isIllegalStateException()) {
+        if (expression.isIllegalStateExceptionWithoutCause()) {
             report(CodeSmell(issue, Entity.from(expression), issue.description))
         }
     }
 
-    private fun KtThrowExpression.isIllegalStateException(): Boolean {
-        return findDescendantOfType<KtCallExpression>()?.firstChild?.text == "IllegalStateException"
+    private fun KtThrowExpression.isIllegalStateExceptionWithoutCause(): Boolean {
+        val callExpression = findDescendantOfType<KtCallExpression>()
```
schalkms: Code duplication here
Can we group this together and pass the type of the exception?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/ThrowExtensions.kt
```diff
@@ -0,0 +1,19 @@
+package io.gitlab.arturbosch.detekt.rules
+
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtThrowExpression
+import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
+import kotlin.reflect.KClass
+
+internal fun KtThrowExpression.isIllegalStateException() =
```
schalkms: Nice refactoring!

PR #1572, opened 2019-04-02
Silence dokka

3flex:
An effort to remove dokka warnings from the build. Only one remaining, which looks to be a bug in dokka itself: https://github.com/Kotlin/dokka/issues/90

Note that there are no changes to any documentation produced by detekt-generator, but default values in the `@configuration` tag in KDoc for rules must now be delimited by backticks, so instead of `default: false` it must be:
```
default: `false`
```

This works better with dokka as dokka will treat anything delimited by backticks as a code span and won't try to create a link for anything inside a code span which is surrounded by `[]` (which was a problem for default values like `[a-z][A-Za-z0-9]*`)

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
@@ -8,12 +8,12 @@ import org.jetbrains.kotlin.psi.KtFile
 typealias RuleId = String
 
 /**
- * Defines the visiting mechanism for [KtFile]'s.
+ * Defines the visiting mechanism for KtFile's.
  *
  * Custom rule implementations should actually use [Rule] as base class.
  *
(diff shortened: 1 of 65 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
```diff
@@ -8,12 +8,12 @@ import org.jetbrains.kotlin.psi.KtFile
 typealias RuleId = String
 
 /**
- * Defines the visiting mechanism for [KtFile]'s.
```
schalkms: Why did you remove [KtFile] but not [MultiRule]?
How can we reference classes in comments now?

schalkms: Oh, I just saw your commit log.
```
KtFile, PsiElement and KtElement don't have any documentation published
online, so it does not make sense to link to those classes from KDoc.
```

Issue #1573, opened 2019-04-02
ForbiddenImport matches on substring, not full class


josephlbarnett:
The bug you're experiencing might have already be reported!
Please search in the [issues](https://github.com/arturbosch/detekt/issues) before creating one.

## Expected Behavior

There should be a way to forbid imports of specific classes and not match classes that are similar, but different.  

## Observed Behavior

because `ForbiddenImport` uses the `contains` method of `SplitPattern`, trying to forbid a class whose name is a substring of another class forbids both classes.  I believe calling `SplitPattern::equals` instead would fix the issue, but have not tested, and would probably break other usages (eg the wildcards in https://github.com/arturbosch/detekt/issues/1001 ).

## Steps to Reproduce

forbid `com.google.inject.Inject`, and run detekt against a class that imports `com.google.inject.Injector`

## Context

We'd like to forbid imports of guice classes that exist in javax.inject, for example `javax.inject.Inject` should be used over `com.google.inject.Inject`.  However, adding `com.google.inject.Inject`to the `ForbiddenImports` rule also forbids imports of `com.google.inject.Injector`, which is a class that has no javax.inject equivalent that we do not want to forbid.

## Your Environment

* Version of detekt used: 1.0.0-RC14
* Version of Gradle used (if applicable):
* Operating System and version:
* Link to your project (if it's a public repository):

schalkms:
Thanks for reporting and having a look into it!
Would you mind to submit a PR including your tests?

josephlbarnett:
PR submitted: https://github.com/arturbosch/detekt/pull/1575

PR #1575, opened 2019-04-03
Test case and fix for issue #1573

josephlbarnett:
Shows that forbidding net.example.R.dimen also forbids
net.example.R.dimension

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenImport.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenImport.kt
@@ -7,7 +7,6 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
-import io.gitlab.arturbosch.detekt.api.SplitPattern
 import org.jetbrains.kotlin.psi.KtImportDirective
 
 /**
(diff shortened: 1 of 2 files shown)


josephlbarnett:
This PR currently fails until a fix is implemented that still allows existing wildcard behavior

josephlbarnett:
Updated with a proposed fix.  This may be more generalizable to `SplitPattern`, but wasn't sure of possible repercussions so did not try to do so.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenImportSpec.kt
```diff
@@ -16,6 +16,7 @@ class ForbiddenImportSpec : Spek({
 
 			import com.example.R.string
 			import net.example.R.dimen
```
schalkms: The formatting here looks odd. Could you please check that?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenImport.kt
```diff
@@ -28,19 +27,32 @@ import org.jetbrains.kotlin.psi.KtImportDirective
...
-    private val forbiddenImports = SplitPattern(valueOrDefault(IMPORTS, ""))
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Style,
+        "Mark forbidden imports. A forbidden import could be an import for an unstable / experimental api" +
+                "and hence you might want to mark it as forbidden in order to get warned about the usage.",
+        Debt.TEN_MINS
+    )
+
+    private val forbiddenImports = valueOrDefault(IMPORTS, "").split(",")
```
schalkms: I think the split pattern should be preserved.
With **,** the user can split imports in the co fig file.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenImportSpec.kt
schalkms: I meant line 19.

josephlbarnett: yeah, looks like original file has tab characters in it, but my new line is all spaces.  i'll unify.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenImport.kt
josephlbarnett: not sure what you mean here? the user should still be able to have imports set to `com.whatever.Class1, com.another.Class2, com.onemore.*` and have all the intended classes forbidden?

schalkms: Sorry, I should have been more clear.
SplitPattern with **,** as a parameter also does the job here. So no extra split is necessary.

josephlbarnett: ok, is the updated commit more like what you were thinking?  the main thing is that SplitPattern.contains() matches more aggressively than we want, so need to convert globs to a regex and match on that.

schalkms: Sorry for the misunderstanding. The original one was fine except
`valueOrDefault(IMPORTS, "").split(",")`
could be
`SplitPattern(IMPORTS, ",")`
I guess.

There is no need for a separate regex.

josephlbarnett: ok, but the matching done by the existing SplitPattern methods don't do what we want:
  - `contains` is the cause of this bug report
  - `startWith` won't allow for `*.R`
  - `equals`won't allow for `*.R`, `com.example.*` or `*.R.*`
So I think we do need a separate matching mechanism, and converting the `*` globs to regexs looks the most natural to me.  Open to other suggestions though, and apologies if I'm just not understanding your suggestion correctly.

schalkms: Okay, apologies for the inconvenience caused by my wrong suggestion.
Your first solution without the regexpr is okay and then approved.
We will merge this very soon.

josephlbarnett: ok, i think that's https://github.com/arturbosch/detekt/commit/a9d5e097f902134be7ab2078536055404755f7ea -- let me know if you want me to update this branch back to that commit or you'll take care of it on your end.

schalkms: Yes please 🥇

josephlbarnett: :+1: done, thanks!

PR #1576, opened 2019-04-03
Ignore void types in overridden function signatures.

### LINKED ISSUE #1569: ForbiddenVoid should have an ignoreOverriddenFunctions setting

If I set `ForbiddenVoid` to `true`, whenever I encounter an overridden function I eventually cannot change it (or I don't know how to do it) because the `Unit` type is actually _different_ from `Void`.

## Context

I'm trying to activate the rule and use it :sunglasses:

For instance, projects like `Vert.x` where you must extend from `AbstractVerticle` and usually/potentially override `start` and `stop`, have the type `startFuture: Future<Void>` and I don't have a way to exclude those.

marschwar:
Closes #1569

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -424,6 +424,7 @@ style:
     imports: ''
   ForbiddenVoid:
     active: false
+    ignoreOverridden: false
   FunctionOnlyReturningConstant:
     active: false
     ignoreOverridableFunction: true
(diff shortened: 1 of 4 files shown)


detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -424,6 +424,7 @@ style:
     imports: ''
   ForbiddenVoid:
     active: false
+    ignoreOverridden: false
```
schalkms: This should be set to **true** in my opinion.

3flex:
Minor nitpick, but it would be good to group the tests where ignoreOverridden is true, and group the rest where it's false. You can use Spek's `context` for that purpose.

PR #1582, opened 2019-04-10
Possibility to customize logging level output

hick209:
I'm working on adding Detekt to the Linter of my company.
Right now we already have a linter in place for all the languages we work with, except for Kotlin.
Our linter calls other tools and expects a very clear output format.
I've created an extension for ConsoleOutput and was almost able to get the output I needed. The problem was that the output does not come out clean.
I've tried to suppress the extra content, but was not able to do so without hacking the codebase.

Therefore I decided to add logging levels so that it will be possible to integrate Detekt with our Linter.

To run it, I did:
`java -jar detekt-cli-nivaldo.jar --plugins custom-reporter.jar --config detekt-config.yml --log-level none`

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,7 @@ If you contributed to detekt but your name is not in the list, please feel free 
 - [Andrew Arnott](https://github.com/andrew-arnott) - UnusedPrivateMember improvement
 - [Tyler Wong](https://github.com/tylerbwong) - UnderscoresInNumericLiterals rule
 - [Daniele Conti](https://github.com/fourlastor) - ObjectPropertyNaming improvement
+- [Nivaldo Bondanca](https://github.com/hick209) - Logging improvements
 
 ### Mentions
 
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
@@ -1,6 +1,9 @@
 package io.gitlab.arturbosch.detekt.cli
 
+import com.beust.jcommander.IParameterValidator
+import com.beust.jcommander.IStringConverter
 import com.beust.jcommander.Parameter
+import com.beust.jcommander.ParameterException
 import java.nio.file.Path
 
 /**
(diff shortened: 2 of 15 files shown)


arturbosch:
Hey, as I am also keen to integrate detekt in others tools myself, I'm interested in which parts of detekt output come out not clean?

Have you tried suppressing all processors and reports available through the config file?
```
processors:
  active: true
  exclude:
  # - 'FunctionCountProcessor'
  # - 'PropertyCountProcessor'
  # - 'ClassCountProcessor'
  # - 'PackageCountProcessor'
  # - 'KtFileCountProcessor'

console-reports:
  active: true
  exclude:
  #  - 'ProjectStatisticsReport'
  #  - 'ComplexityReport'
  #  - 'NotificationReport'
  #  - 'FindingsReport'
  #  - 'BuildFailureReport'
```

You can supress even more stuff when looking at the META-INF/services:
```
io.gitlab.arturbosch.detekt.cli.DetektProgressListener
io.gitlab.arturbosch.detekt.core.processors.KtFileCountProcessor
io.gitlab.arturbosch.detekt.core.processors.PackageCountProcessor
io.gitlab.arturbosch.detekt.core.processors.ClassCountProcessor
io.gitlab.arturbosch.detekt.core.processors.FunctionCountProcessor
io.gitlab.arturbosch.detekt.core.processors.PropertyCountProcessor
io.gitlab.arturbosch.detekt.core.processors.ProjectComplexityProcessor
io.gitlab.arturbosch.detekt.core.processors.ProjectLLOCProcessor
io.gitlab.arturbosch.detekt.core.processors.ProjectCLOCProcessor
io.gitlab.arturbosch.detekt.core.processors.ProjectLOCProcessor
io.gitlab.arturbosch.detekt.core.processors.ProjectSLOCProcessor
```
Would this be sufficient for you without introducing more complexity through cli flags?

hick209:
Thanks a lot for your comment and helping pointing out how to suppress other sources.
On my local tests I can produce a clean output by suppressing those reporters, as you suggested.

My problem is I'm currently using the latest release `1.0.0-RC14` on the codebase as we need to pull it from Maven. On that version, I get the following extra lines even if I disable everything:
```
Using deprecated property 'failFast' in the yaml config. Please migrate to the new '--fail-fast' cli-flag or 'failFast' detekt extension property.
detekt finished in 1280 ms.
```

Regardless, I like better the strategy of moving everything that can be output into the config file. I'm abandoning this diff then.

In the mean time, I'm just using a bash script to call into the detekt binary and in it I pipe it to filter those undesired extra lines.

Thanks for the amazing tool you provide! I look forward it's further development and also contributing with it whenever possible.

arturbosch:
Thanks for the kind words and great that it worked out for you!
Yep, that deprecation warning was bad implemented, need to find a better deprecation strategy which should also be controlled by a output report or processor.

PR #1593, opened 2019-04-17
Added formatting information to Detektion

hick209:
This helps if any other tool that wants to use Detekt wants to control the formatting of the files or create patches.

I'm working on adding Detekt to the Linter of my company as mentioned in [this other PR](https://github.com/arturbosch/detekt/pull/1582).

The next step for us is integrating it with our formatter.
To do that we need the new formatted content info before it is applied to the file. With that we create patches for the files and can then propose those changes in our CI tool.

My first idea to accomplish this is adding that info to Detektion.
This way I generate the desired patch output with a custom reporter.

Please do feel free to suggest alternatives.

I have verified and everything still runs fine on my end and all the tests pass.


arturbosch:
Hey @hick209 , good to see your going on the limits on our extension model :)

I tried to mimic your needed feature via the current available extension classes:
```kotlin
class FormattingPrintReport : ConsoleReport() {

    override fun render(detektion: Detektion): String? = detektion.notifications
        .filterIsInstance<FormattingNotificationProvider.FormattingNotification>()
        .joinToString("\n\n") { it.content }

}

class FormattingNotificationProvider : FileProcessListener {

    data class FormattingNotification(val path: String, val content: String) : Notification {
        override val message: String = "Modification for KtFile '$path'."
    }

    override fun onFinish(files: List<KtFile>, result: Detektion) {
        files.filter { it.modificationStamp > 0 }
            .map { it.absolutePath() to it.unnormalizeContent() }
            .filter { it.first != null }
            .forEach { result.add(FormattingNotification(it.first!!, it.second)) }

    }

    private fun KtFile.unnormalizeContent(): String {
        val lineSeparator = getUserData(LINE_SEPARATOR)
        require(lineSeparator != null) {
            "No line separator entry for ktFile ${javaFileFacadeFqName.asString()}"
        }
        return StringUtilRt.convertLineSeparators(text, lineSeparator)
    }
}

val LINE_SEPARATOR: Key<String> = Key("lineSeparator")
```

Would this fulfill your requirements? I'm interessted in adding features to the extension mechanism so no modifying of the core modules are needed.
My proposed implementation however needs `LINE_SEPARATOR` and `KtFile.unnormalizeContent` to be moved from `KtFileModifier` to `detekt-api/api/internal`.

hick209:
This does meet our needs! Thank you so for taking a look into this issue and proposing this solution.
I'll update my pull request on Monday so I has the necessary changes do use this processor, if necessary otherwise I'll just drop those changes.

Once again, thanks for taking the time to show me how to accomplish this with the current tools.

hick209:
Things to be moved from `core` to `api` module:
- `LINE_SEPARATOR`
- `KtFile.unnormalizeContent`
- `KtFile.absolutePath`

This could work, but it looks like plenty of changes would be necessary, since there's a lot of code from inside the `core` module that would need to be exposed in order to make this work.

Do you think I should go ahead and make those changes?

hick209:
Tested the changes locally here and I can do it importing `detekt-core` and `org.jetbrains.kotlin:kotlin-compiler-embeddable` as a dependency to my project.

I think it would be good not to need to add the second one, so I'm going to create a PR to reflect a change that would be necessary for such. 🙂

PR #1595, opened 2019-04-18
Added auto correct information to finding

hick209:
This makes it possible for Reports to be aware of where a particular finding had the auto correct function turned on or not.

Part of the same effort as being done in [PR 1593](https://github.com/arturbosch/detekt/pull/1593). This change will help my company integrate Detekt with our tools and provide auto fixes when necessary.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
@@ -14,7 +14,7 @@ interface Finding : Compactable, HasEntity, HasMetrics {
     val issue: Issue
     val references: List<Entity>
     val message: String
-    val autoCorrect: Boolean
+    val autoCorrectEnabled: Boolean
 
     fun messageOrDescription(): String
 }
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
@@ -14,7 +14,6 @@ interface Finding : Compactable, HasEntity, HasMetrics {
     val issue: Issue
     val references: List<Entity>
     val message: String
-    val autoCorrectEnabled: Boolean
 
     fun messageOrDescription(): String
 }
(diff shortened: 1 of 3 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
```diff
@@ -14,6 +14,7 @@ interface Finding : Compactable, HasEntity, HasMetrics {
     val issue: Issue
     val references: List<Entity>
     val message: String
+    val autoCorrect: Boolean
```
3flex: It might be a bit clearer to use `wasAutoCorrected` or similar instead of `autoCorrect` to better convey the meaning of the value.

hick209: I can make that change, but we also have to keep in mind that this can be `true` even if the file was not touched, in case the `autoCorrect` is set to `false` on the top level of configuration file and set to `true` in that particular rule

3flex: `autoCorrectEnabled`?

hick209: Loved it! I'm going with that

arturbosch: Hm, lets better introduce a new class inside `detekt-api` `CorrectableCodeSmell` which has this extra property?

hick209:
@3flex, here it says you are requesting changes.
I've done the modifications you needed and marked the comment as resolved, is there something else I need to do (possible GitHub noob alert) to have the RC addressed?

3flex:
@arturbosch here you requested changes: https://github.com/arturbosch/detekt/pull/1595#discussion_r279159351

But you also approved the PR without those changes. Did you want amendments or can this be merged?

hick209:
Created the class `CorrectableCodeSmell`, let me know if there's anything else you would like to be changed.

arturbosch:
Thank you!

PR #1596, opened 2019-04-18
Allow custom report format to be defined by Gradle Plugin

### LINKED ISSUE #1583: Using custom output inside Gradle plugin
## Introduction
Hello everyone! 

First of all, thanks for the incredible work you've all put together. 

I'm using detekt for a couple of days, and I got used to it pretty quickly, so I managed to create a custom report that writes data on a file. The report class looks like the following: 

```kotlin
/**
 * Prints out a [Gitlab CI Code Quality](https://docs.gitlab.com/ee/user/project/merge_requests/code_quality.html)
 * compatible report inside a JSON file.
 * @author Riccardo Montagnin
 */
class GitlabJsonReport: OutputReport() {

    override val ending: String = "json"

    /**
     * @inheritDoc
     */
    override fun render(detektion: Detektion): String? {
        return gitlabCodeQualityReport(detektion)
    }
}
```

I'm sure this is correct because running the following command prints out the correct report: 

```bash
java -jar detekt-cli-1.0.0-RC14-all.jar \
  --input <MY-FILE-PATH> \
  --plugins detekt-gitlab-extension-0.1.jar \
  --report GitlabJsonReport:./reports/report.json \
  --config <MY-CONFIG-FILE-PATH>
```

## The problem
The only problem I encounter is the following: how do I enable the custom reporting inside the detekt gradle plugin? 

I've tried including it as follows: 

```groovy
detekt {
    debug = true
    toolVersion = "${versions.detekt}"
    input = files(
        "${project.rootDir}/main/src/main/kotlin",
        "${project.rootDir}/main/src/test/kotlin"
    )
    config = files(file("$rootDir/.detekt/detekt-config.yml"))
    filters = ".*/resources/.*,.*/build/.*"
    plugins = "${project.rootDir}/.detekt/detekt-gitlab-extension-0.1.jar"
}
```

But I have no clue on how to configure the `detekt` Gradle extension to execute my custom output. I've tried using 

```groovy
detekt {
  reports { 
    GitlabJsonReport { ... }
  }
}
```

But I obviously have an error due to the fact that the `detekt` extension does not support custom reports. How should I do it? 

Thanks for the help!

marschwar:
Closes #1583 

This introduces a breaking change. I think the `plugins` property should always have been a FileCollection instead of a simple string.
 
Still WIP - documentation needs update.


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -67,9 +70,10 @@ open class Detekt : SourceTask() {
     @PathSensitive(PathSensitivity.RELATIVE)
     var config: ConfigurableFileCollection = project.configurableFileCollection()
 
-    @Input
+    @InputFiles
     @Optional
-    var plugins: Property<String> = project.objects.property(String::class.java)
+    @PathSensitive(PathSensitivity.RELATIVE)
+    var plugins: ConfigurableFileCollection = project.configurableFileCollection()
```
3flex: This can be done without breaking changes - SpotBugs and FindBugs plugins use `pluginClasspath` as the name for the equivalent parameter, so for consistency this would be an opportunity to introduce it here:
```kotlin
    @Classpath
    @Optional
    val pluginClasspath: ConfigurableFileCollection = project.configurableFileCollection()
```

And mark existing `plugins` as `@Deprecated`. Then in `@TaskAction` check that that both parameters aren't set at the same time, and either warn or fail if so.

`pluginClasspath` should be wired up so the default value is `project.configurations.detektPlugins`

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
```diff
@@ -59,9 +59,10 @@ open class DetektCreateBaselineTask : SourceTask() {
     @PathSensitive(PathSensitivity.RELATIVE)
     var config: ConfigurableFileCollection = project.configurableFileCollection()
 
-    @Input
+    @InputFiles
     @Optional
-    var plugins: Property<String> = project.objects.property(String::class.java)
+    @PathSensitive(PathSensitivity.RELATIVE)
+    var plugins: ConfigurableFileCollection = project.configurableFileCollection()
```
3flex: Same comment as for `Detekt.kt`

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -44,7 +44,7 @@ open class DetektExtension(project: Project) : CodeQualityExtension() {
     @Deprecated("Replace with task setIncludes/setExcludes")
     var filters: String? = null
 
-    var plugins: String? = null
+    var plugins: ConfigurableFileCollection = project.configurableFileCollection()
```
3flex: This should actually be deprecated on the extension now I think about it - the correct way to set a default set of plugins is with the `detektPlugins` configuration. There's no need to have two ways to provide a default list of additional plugins.

Fortunately "extending detekt" docs already say that's the way to add plugins. By deprecating this then removing from the Kotlin DSL and Gradle DSL docs we can remove the redundant config option.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskDslTest.kt
```diff
@@ -240,6 +291,119 @@ internal class DetektTaskDslTest : Spek({
...
+
+                        val config = """
+                        |detekt {
+                        |	reports {
+                        |		custom {
+                        |           type = "foo"
+                        |           destination = file("$aDirectory")
+                        |       }
+                        |	}
+                        |}
```
3flex: Please use spaces for indentation within the new tests so the indentation is fixed up (I'm guessing this happened because there was a copy/paste from other tests, and the indentation on many of those is still based on tabs - I think the IDE uses 4 spaces as default but Github shows 8. Pretty annoying and I've run into that too).

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DslGradleRunner.kt
```diff
@@ -76,9 +76,8 @@ class DslGradleRunner(
         }
     }
 
-    fun writeProjectFile(filename: String, content: String) {
-        File(rootDir, filename).writeText(content)
-    }
+    fun writeProjectFile(filename: String, content: String) =
+        File(rootDir, filename).apply { writeText(content) }
```
3flex: Why this change?

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskDslTest.kt
```diff
@@ -240,6 +291,119 @@ internal class DetektTaskDslTest : Spek({
...
+						"""
+
+                        val gradleRunner = builder
+                            .withDetektConfig(config)
+                            .build()
+
+                        gradleRunner.runDetektTaskAndExpectFailure()
+                    }
+                    it("fails if destination of custom report is missing") {
```
3flex: Instead of failing in this case is there a default report path and filename that could be generated by the plugin instead?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/CustomDetektReport.kt
```diff
@@ -0,0 +1,28 @@
...
+import org.gradle.api.tasks.OutputFile
+import java.io.File
+
+class CustomDetektReport(private val project: Project) {
+
+    @Internal
+    val typeProperty: Property<String> = project.objects.property(String::class.java)
+    var type: String
+        get() = typeProperty.get()
+        set(value) = typeProperty.set(value)
```
3flex: Since these are new parameters it's not really necessary to rewire these getters and setters - instead just do

```kotlin
val type: Property<String> = project.objects.property(String::class.java)
```

It saves on LOC and per [Gradle's guidelines](https://docs.gradle.org/current/userguide/lazy_configuration.html#sec:lazy_configuration_faqs) for introducing lazy configuration 

> If it’s a new property, expose it as a Property or Provider using a single getter.

Just make sure the doc examples use `type.set(<value>)` for the Kotlin DSL instead of `type = <value>`

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
marschwar: great point. I will remove it from this PR and create a separate one.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
marschwar: Ok. To be honest I never used any custom plugins and never fully understood the mechanism. Thanks for the explanation.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskDslTest.kt
marschwar: Technically possible, but if someone is going through the trouble of creating a custom report type,  I don't think it is not too much to ask to provide a filename. Especially there really is know way of knowing what the content type will be.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/CustomDetektReport.kt
marschwar: I was under the impression that it is the goal to provide the same DSL for kotlin and groovy. No matter if that is actually the case I think at this point it would be inconsistent to have a different way of specifying properties for default reports and for custom reports.

```
detekt {
  reports {
    html {
      destination = file("build/reports/detekt.html")
    }
    custom {
      type.set("json")
      destination.set(file("build/reports/detekt.json"))
    }
  }
}
```

3flex: > the goal to provide the same DSL for kotlin and groovy

You can use the same syntax of `.set()` in Groovy too, so can still say that the two DSLs are consistent, but seeing it laid out like that it's clear it would be confusing for consumers of the plugin, so ignore my previous comment.

The fact that syntax has to change for Kotlin DSL when Providers are used is really unfortunate :/

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DslGradleRunner.kt
marschwar: https://kotlinlang.org/docs/reference/coding-conventions.html#function-formatting

> Prefer using an expression body for functions with the body consisting of a single expression.

3flex: Sorry, was wondering about addition of `.apply` to the expression, but I've just seen later where the return value is used in the test.

marschwar:
To avoid confusion, I will just create two new PRs.

PR #1597, opened 2019-04-19
Gradle plugin: create task per source set (JVM)

### LINKED ISSUE #1198: Gradle plugin: create task per source set

### LINKED ISSUE #1580: Using the Gradle Plugin analyzes files sequentially
## Expected Behavior
Analysis of the source files should be done in parallel

## Observed Behavior
When using the gradle plugin each source file is analyzed sequentially.

## Steps to Reproduce
```
time java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar -i detekt-rules/src/main/kotlin
```
On my machine this takes about 3.9s

Then run `./gradlew detekt-rules:detekt --rerun-tasks` to identify the input files that are effectively processed and replace `detekt-rules/src/main/kotlin` with the list of those files. For me that took about 4.5s

## Context
#1505 introduced support for source sets and determines all files that should be analyzed within the gradle plugin. The cli code historically seems to assume that an element in the input parameter list consists of multiple files and processes those files in parallel. (See `DetektFacade:34` and `Detektor:56`) 
 
I don't know if this is such a big deal but so assume it is an unintended consequence of the use of source sets and I wanted to point it out. Is there a reason the parallel processing is done in the `Detektor` instead of the `DetektFacade`?

3flex:
Closes #1198. Don't merge until tests & docs are added. Does nothing by default - users have to explicitly run the auto-generated tasks.

Try out various scenarios and help spot regressions or cases where the type resolution won't work - I've tried on the kotlinconf app and it does what it's supposed to, but I've also modified some core code that I'm not that familiar with so hopefully haven't broken anything. Tests pass.

To test:
* Check out this branch
* Add this to style block in `reports/failfast.yml`:
```yaml
  UselessCallOnNotNull:
    active: true
```
* Insert `val testList = listOf<String>().orEmpty()` somewhere in the main source set of one of the projects
* Run `./gradlew detektMain` - the above offending code should be detected by UselessCallOnNotNull rule which shows type resolution working as expected.

Limitations:
* Doesn't support JS or native projects. Type resolution capability in detekt only works on JVM currently so this isn't really an issue.
* No Android support. Will investigate, but may not include support for now.
* No multiplatform project support. I don't see this as a blocker as it's experimental in Kotlin currently.

Also closes #1580

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
@@ -1,6 +1,7 @@
 package io.gitlab.arturbosch.detekt.cli
 
 import com.beust.jcommander.Parameter
+import org.jetbrains.kotlin.config.JvmTarget
 import java.nio.file.Path
 
 /**
@@ -26,13 +27,6 @@ class CliArgs : Args {
     @Parameter(names = ["--excludes", "-ex"],
         description = "Globing patterns describing paths to exclude from the analysis.")
     var excludes: String? = null
-
-    @Parameter(
-        names = ["--classpath", "-cp"],
-        required = false,
-        description = "EXPERIMENTAL: Compile Classpath of the project."
-    )
-    var classpath: String? = null
 
     @Parameter(names = ["--config", "-c"],
         description = "Path to the config file (path/to/config.yml). " +
@@ -106,6 +100,25 @@ class CliArgs : Args {
         description = "Prints the AST for given [input] file. Must be no directory.",
         hidden = true)
     var printAst: Boolean = false
+
+    /*
+        The following @Parameters are used for type and symbol resolving. When additional parameters are required the
+        names should mirror the names found in this file (e.g. "classpath", "jvm-target"):
+        https://github.com/JetBrains/kotlin/blob/master/compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/K2JVMCompilerArguments.kt
+    */
+    @Parameter(
+        names = ["--classpath", "-cp"],
+        description = "EXPERIMENTAL: Paths where to find user class files"
+    )
+    var classpath: String? = null
+
+    @Parameter(
+        names = ["--jvm-target"],
+        converter = JvmTargetConverter::class,
+        description = "EXPERIMENTAL: Target version of the generated JVM bytecode that was generated during " +
+                "compilation and is now being used for type resolution (1.6, 1.8, 9, 10, 11 or 12)"
+    )
+    var jvmTarget: JvmTarget = JvmTarget.DEFAULT
 
     val inputPaths: List<Path> by lazy {
         MultipleExistingPathConverter().convert(input ?: System.getProperty("user.dir"))
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
@@ -32,7 +32,8 @@ class DetektFacade(
     private val inputPaths = settings.inputPaths
     private val classpath = settings.classpath
     private val pathFilters = settings.pathFilters
-    private val compilerConfiguration = createCompilerConfiguration(classpath, inputPaths)
+    private val jvmTarget = settings.jvmTarget
+    private val compilerConfiguration = createCompilerConfiguration(inputPaths, classpath, jvmTarget)
     private val environment = createKotlinCoreEnvironment(compilerConfiguration)
     private val compiler = KtTreeCompiler.instance(settings)
 
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
@@ -8,7 +8,9 @@ import io.gitlab.arturbosch.detekt.api.RuleSetProvider
 import io.gitlab.arturbosch.detekt.api.createCompilerConfiguration
 import io.gitlab.arturbosch.detekt.api.createKotlinCoreEnvironment
 import org.jetbrains.kotlin.cli.common.messages.AnalyzerWithCompilerReport
-import org.jetbrains.kotlin.cli.common.messages.MessageRenderer
+import org.jetbrains.kotlin.cli.common.messages.CompilerMessageLocation
+import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSeverity
+import org.jetbrains.kotlin.cli.common.messages.PlainTextMessageRenderer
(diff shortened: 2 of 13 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -103,6 +97,25 @@ class CliArgs : Args {
...
+    )
+    var classpath: String? = null
+
+    @Parameter(
+        names = ["--jvm-target"],
+        converter = JvmTargetConverter::class,
+        description = "EXPERIMENTAL: Target version of the generated JVM bytecode that was generated during " +
+                "compilation and is now being used for type resolution (1.6, 1.8, 9, 10, 11 or 12)"
+    )
+    var jvmTarget: JvmTarget = JvmTarget.DEFAULT
```
arturbosch: Can we here just provide the version of the running jvm?
`System.getProperty("java.vm.specification.version")=1.8`

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -5,30 +5,58 @@ import org.gradle.api.Plugin
...
         registerCreateBaselineTask(project, extension)
         registerGenerateConfigTask(project, extension)
 
         registerIdeaTasks(project, extension)
     }
 
-    private fun registerDetektTask(project: Project, extension: DetektExtension) {
+    @Suppress("UnsafeCast")
+    private fun Any.getConvention(name: String): Any? =
+        (this as HasConvention).convention.plugins[name]
```
arturbosch: You could write here `(this as? HasConvention)?.convention?.plugins[name]` as the method expectes `Any?` to return.
No suppression needed.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -60,6 +88,27 @@ class DetektPlugin : Plugin<Project> {
         checkTaskProvider?.configure { it.dependsOn(detektTaskProvider) }
     }
 
+    private fun registerDetektTask(project: Project, extension: DetektExtension, sourceSet: SourceSet) {
+        @Suppress("UnsafeCast")
+        val kotlinSourceSet = sourceSet.getConvention("kotlin") as KotlinSourceSet
```
arturbosch: I would prefer something like `as? KotlinSourceSet ?: error("We expect to operate on kotlin source sets")` here for the very unlikely case that Gradle will change something.

arturbosch:
Jap, works with detektMain (type resolution) and is 4x times slower.
Normal runs are also fixed and nearly 2x faster again ^^.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
3flex: As the default? I suppose that might be safe, but how reliable is that property for different JDK vendors, e.g. OpenJDK, Oracle JDK, Corretto... The returned string has to be one of 1.6, 1.8, 9, 10, 11 or 12 otherwise an error will be returned, which if it's something not expected won't be ideal.

I don't know enough to say either way. I think using Kotlin's current default makes sense? Or we can set it by default to 1.6 (or 1.8) so it's a static value.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
3flex: With that I get
> Only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type (Mutable)Map<String!, Any!>?

I tried a few other ways to do this, but can't see a better way than what's there - some other ways I found lead to a very ugly warning about illegal reflective access during invocations of `gradlew detektMain` so that's not really an option.

Open to suggestions but otherwise will leave as is.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
```diff
@@ -20,36 +31,61 @@ class DetektFacade(
...
-        for (current in pathsToAnalyze) {
-            val files = compiler.compile(current)
+        val filesToAnalyze = environment.getSourceFiles()
+            .filter { file -> !pathFilters.any { it.matches(Paths.get(file.virtualFilePath)) } }
+            .apply { forEach { it.addUserData(it.virtualFilePath) } }
+        val bindingContext = generateBindingContext(filesToAnalyze)
+
+        processors.forEach { it.onStart(filesToAnalyze) }
+
+        if (saveSupported) {
```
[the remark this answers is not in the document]
3flex: Before the PR, code reads that if saveSupported is true, then autocorrect is enabled, and files that require modification will be updated (`KtFileModifier(current).saveModifiedFiles(files)`). This maintains that functionality I think?

Sorry I might need more help here, am I misunderstanding the purpose of saveSupported?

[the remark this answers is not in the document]
arturbosch: Sorry for late response here.
The `if (saveSupported) {...}` must be run after all rules were run, so basically         `findings.mergeSmells(detektor.run(files, environment, classpath.isNotEmpty()))`
must run before `saveModifiedFiles`.
The `saveModifiedFiles` function basically checks if any modifications were made to the AST and save the content then.

PR #1598, opened 2019-04-19
Custom reports in gradle plugin

marschwar:
Replacement of #1596. Closes #1583 

This allows users to use their custom reports with the gradle plugin.

```
detekt {
    reports {
        custom {
            reportId = "CustomJsonReport"
            destination = file("build/reports/detekt.json")
        }
    }
}
```

PR #1599, opened 2019-04-19
Mark `plugins` config option in gradle plugin dsl as deprecated.

marschwar:
As suggested by @3flex in https://github.com/arturbosch/detekt/pull/1596#discussion_r276919771 the `plugins` option in the gradle plugin dsl should not be removed in favor of `detektPlugins`.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -69,6 +69,7 @@ open class Detekt : SourceTask() {
 
     @Input
     @Optional
+    @Deprecated("Use detektPlugins within dependencies instead")
     var plugins: Property<String> = project.objects.property(String::class.java)
 
     @Internal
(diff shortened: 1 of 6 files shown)


3flex:
Deprecating on the tasks only makes sense if the new `pluginClasspath` parameter is added as I mentioned in comments in the other PR - for now it only makes sense to deprecate on the extension.

marschwar:
Honestly I still don't get it. I don't see any point in adding a new `pluginClasspath` option when there already is an excellent way to add custom jars to the classpath. What am I missing?

3flex:
There are two problems with the existing setup:
1. Can't override the plugin classpath for specific tasks
2. Plugin classpath is not checked during Gradle's up-to-date checks, so if the JARs fed to `detektPlugins` configuration changes none of the detekt tasks will be marked out of date and rerun.

I'm happy to take this on since it's reasonably straightforward to correct, but of course feel free to tackle it if you'd like! I'm working on a few rule updates in the meantime, but definitely would be good to see this corrected.

3flex:
I realised that adding the new classpath parameters to the tasks would likely fix another issue with detekt's build as well as what I mentioned in my previous comment, so I've started working on a PR to add the new parameters. I'll have that ready in about 24 hours, but it will supersede this, so I'm going to close it for now. Thanks for bringing the issue to my attention though! The new PR will help anyone trying to use #1598 as well.

PR #1600, opened 2019-04-19
Adds support for ignoreFailures to gradle plugin

### LINKED ISSUE #1435: Gradle plugin ignores `ignoreFailures` value
```gradle
tasks {
  detekt {
    isIgnoreFailures = true
  }
}
```

marschwar:
Closes #1435 

The `ignoreFailures` property inherited from  `CodeQualityExtension` now can be used to continue the build even if there are more findings than allowed.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -44,6 +44,7 @@ import org.gradle.api.tasks.PathSensitive
 import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.SourceTask
 import org.gradle.api.tasks.TaskAction
+import org.gradle.api.tasks.VerificationTask
 import org.gradle.language.base.plugins.LifecycleBasePlugin
 import java.io.File
 
@@ -54,7 +55,7 @@ import java.io.File
  * @author Matthew Haughton
  */
 @CacheableTask
-open class Detekt : SourceTask() {
+open class Detekt : SourceTask(), VerificationTask {
 
     @Deprecated("Replace with getSource/setSource")
     var input: FileCollection
@@ -140,9 +141,12 @@ open class Detekt : SourceTask() {
         get() = failFastProp.get()
         set(value) = failFastProp.set(value)
 
+    private val ignoreFailuresProp: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
     @Input
     @Optional
-    val ignoreFailures: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
+    override fun getIgnoreFailures(): Boolean = ignoreFailuresProp.get()
+    override fun setIgnoreFailures(value: Boolean) = ignoreFailuresProp.set(value)
+    fun setIgnoreFailures(value: Provider<Boolean>) = ignoreFailuresProp.set(value)
 
     @Optional
     @Input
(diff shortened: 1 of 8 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
```diff
@@ -83,6 +83,10 @@ open class DetektCreateBaselineTask : SourceTask() {
     @Optional
     var failFast: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
 
+    @Input
+    @Optional
+    val ignoreFailures: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
```
3flex: I don't think it makes sense to support this when creating a baseline - if there are failures wouldn't you always want the task to fail since the baseline file wouldn't have been written?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -111,6 +111,10 @@ open class Detekt : SourceTask() {
         get() = failFastProp.get()
         set(value) = failFastProp.set(value)
 
+    @Input
+    @Optional
+    val ignoreFailures: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
```
3flex: The detekt task should inherit `VerificationTask` which is an interface that requires `getIgnoreFailures` and `setIgnoreFailures` properties to be set on the task, and I think should allow using `ignoreFailures = true/false` in both Kotlin and Groovy DSLs.

That can be wired up to this `Property`.

marschwar:
> Can you please also add tests showing `ignoreFailures` causes a detekt task to succeed even when there are failures over the threshold amount?
I forgot to mention that I did try that but for some reason the gradle runner seems to return exit code 1 instead of 2. I will have another look.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
marschwar: I agree. Will remove it.

Just out of curiosity. Do you think `failFast` makes any sense?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
marschwar: Good point.
As a side note: Implementing java interfaces while supporting groovy and kotlin dsl simultaneously makes everything really messy.

3flex:
> the gradle runner seems to return exit code 1 instead of 2.

I believe that's because the change introducing exit code 2 is only on master branch, and at the moment the Gradle plugin tests actually pull detekt-cli from an external repo during testing. Because exit code 2 feature isn't in any released version yet that test won't pass.

I'm going to try and improve that situation for the plugin tests, but in the meantime let's hope @arturbosch releases RC15 soon ;)

Then you can rebase and add that test.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
3flex: I believe `failFast` makes sense because it affects which rules are enabled, and the baseline should be generated based on rules that are enabled at the time the task runs.

PR #1602, opened 2019-04-21
Add MissingWhenCase rule

### LINKED ISSUE #269: MissingCases
A rule that detekts missing cases.

This could be useful when using `when` with enums or sealed classes.

[ErrorProne reference](https://github.com/google/error-prone/blob/master/docs/bugpattern/MissingCasesInEnumSwitch.md).

3flex:
Closes #269

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -387,6 +387,8 @@ potential-bugs:
     active: false
     excludeAnnotatedProperties: ""
     ignoreOnClassesPattern: ""
+  MissingWhenCase:
+    active: false
   UnconditionalJumpStatementInLoop:
     active: false
   UnreachableCode:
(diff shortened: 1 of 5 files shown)


neugaojin:
how to run with a BindingContext
I'm writing custom rules
run `./gradlew detekt`  But the `BindingContext` is BindingContext.EMPTY

3flex:
Use `./gradlew detektMain`. That experimental task should pass the correct classpath to detekt so it can generate BindingContext.

neugaojin:
Thank you for your reply
But, Is this Task added to the release version?
i found the newest version is `1.0.0-RC15`
when i access plugin 1.0.0-RC15 and run `./gradlew detektMain`
i got the error :`Task 'detektMain' not found in root project `

3flex:
It should create a task for each sourceset if you are using 1.0.0-RC15 of the Gradle plugin. Try `gradlew tasks --group="verification"`, you should have a couple of tasks with description "EXPERIMENTAL & SLOW: Run detekt analysis for main classes with type resolution".

If not, please open a new issue so we can track it there.

neugaojin:
There is no such a Task，I'll open a new issue, Thank you

neugaojin:
> Use `./gradlew detektMain`. That experimental task should pass the correct classpath to detekt so it can generate BindingContext.

Sorry，another question
how to pass `classpath` with gradle task

3flex:
You'd set the value of the `classpath` parameter on the detekt task to whatever classpath your project needs to compile.

It won't be easy without the Kotlin JVM plugin applied and using the auto generated task (which will set the classpath parameter for you) so if you're doing it manually it might be difficult and I won't be able to provide any support

PR #1603, opened 2019-04-21
Add RedundantElseInWhen rule

3flex:
Complements MissingWhenCase rule (#1602)

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/RedundantElseInWhen.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/RedundantElseInWhen.kt
@@ -1,0 +1,100 @@
+package io.gitlab.arturbosch.detekt.rules.bugs
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.builtins.KotlinBuiltIns
+import org.jetbrains.kotlin.cfg.WhenChecker
+import org.jetbrains.kotlin.codegen.kotlinType
+import org.jetbrains.kotlin.descriptors.ClassDescriptor
+import org.jetbrains.kotlin.load.kotlin.toSourceElement
+import org.jetbrains.kotlin.psi.KtWhenExpression
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.BindingContext.DECLARATION_TO_DESCRIPTOR
+import org.jetbrains.kotlin.resolve.descriptorUtil.module
+import org.jetbrains.kotlin.resolve.source.getPsi
+
+/**
+ * Turn on this rule to flag `when` expressions that contain a redundant `else` case. This occurs when it can be
+ * verified that all cases are already covered when checking cases on an enum or sealed class.
+ *
+ * Based on code from Kotlin compiler:
+ * https://github.com/JetBrains/kotlin/blob/v1.3.30/compiler/frontend/src/org/jetbrains/kotlin/cfg/ControlFlowInformationProvider.kt
+ *
+ * <noncompliant>
+ * enum class Color {
+ *     RED,
+ *     GREEN,
+ *     BLUE
+ * }
+ *
+ * fun whenOnEnumFail(c: Color) {
+ *     when(c) {
+ *         Color.BLUE -> {}
+ *         Color.GREEN -> {}
+ *         Color.RED -> {}
+ *         else -> {}
+ *     }
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * enum class Color {
+ *     RED,
+ *     GREEN,
+ *     BLUE
+ * }
+ *
+ * fun whenOnEnumCompliant(c: Color) {
+ *     when(c) {
+ *         Color.BLUE -> {}
+ *         Color.GREEN -> {}
+ *         else -> {}
+ *     }
+ * }
+ *
+ * fun whenOnEnumCompliant2(c: Color) {
+ *     when(c) {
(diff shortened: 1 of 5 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/RedundantElseInWhen.kt
```diff
@@ -0,0 +1,103 @@
...
+        Debt.TWENTY_MINS
+    )
+
+    @Suppress("ReturnCount")
+    override fun visitWhenExpression(expression: KtWhenExpression) {
+        if (bindingContext == BindingContext.EMPTY) return
+        if (expression.entries.find { it.isElse } == null) return
+        if (expression.isUsedAsExpression(bindingContext)) return
+        val subjectExpression = expression.subjectExpression ?: return
+        val subjectType = subjectExpression.kotlinType(bindingContext) ?: return
```
schalkms: It's hard to understand for me what the code from line 86-88 does and why we need it.
Maybe we can refactor this into a own method and add a descriptive name with a comment?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/RedundantElseInWhen.kt
```diff
@@ -0,0 +1,103 @@
...
+
+        if (WhenChecker.getMissingCases(expression, bindingContext).isEmpty() && !subjectType.isFlexible()) {
```
schalkms: `subjectType.isFlexible()` -> what's meant by that?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/RedundantElseInWhen.kt
```diff
@@ -0,0 +1,103 @@
...
+            val subjectClass = subjectType.constructor.declarationDescriptor as? ClassDescriptor
+            val pseudocodeDescriptor =
+                bindingContext[DECLARATION_TO_DESCRIPTOR, subjectClass?.toSourceElement?.getPsi()]
+            if (subjectClass == null ||
+                KotlinBuiltIns.isBooleanOrNullableBoolean(subjectType) ||
+                subjectClass.module == pseudocodeDescriptor?.module
```
schalkms: Could you please explain to me the last check in this if expression?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/RedundantElseInWhen.kt
```diff
@@ -0,0 +1,103 @@
...
+ *     when(c) {
+ *         Color.BLUE -> {}
+ *         Color.GREEN -> {}
+ *         Color.RED -> {}
+ *     }
+ * }
+ * </compliant>
+ *
+ * Based on code from Kotlin compiler:
+ * https://github.com/JetBrains/kotlin/blob/v1.3.30/compiler/frontend/src/org/jetbrains/kotlin/cfg/ControlFlowInformationProvider.kt
```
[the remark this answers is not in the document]
3flex: No, I intended to add it - does it need to go before the compliant/noncompliant blocks?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/RedundantElseInWhen.kt
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @author Matthew Haughton
+ */
+class RedundantElseInWhen(config: Config = Config.empty) : Rule(config) {
+
+    override val issue: Issue = Issue(
+        "RedundantElseInWhen",
+        Severity.Defect,
```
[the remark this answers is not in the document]
3flex: No, sorry, bad copy paste. Will change to `Warning` unless you think something more appropriate?

Off topic: we should document a definition for each of the Severity enum values.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/RedundantElseInWhen.kt
3flex: I'm going to review this more closely and add additional tests to make sure the rule is only active when required - looking more closely at the core this is adapted from I think there might be a couple of logical issues.

Thanks for challenging this.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/RedundantElseInWhen.kt
3flex: This is used because code in other modules not in the developer's control might unexpectedly add a new enum value (e.g. when checking cases based on enum or sealed class in an external modules) so it might be desirable to guard against that with an `else`. This was the result of discussion in KT-17497.

detekt can be more flexible here so i will introduce a "strict" config option to either enable or disable that check.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/RedundantElseInWhen.kt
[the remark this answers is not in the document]
schalkms: Yes, it does.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/RedundantElseInWhen.kt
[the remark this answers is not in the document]
schalkms: Good.
Thanks for submitting the issue about the docs right away.

3flex:
I've addressed some feedback, but unfortunately won't have the time to dedicate to the rest of it. If this isn't suitable to merge as is I'm happy for someone else to pick up my branch and continue from there.

arturbosch:
@3flex I was just waiting for the rebase ^^ thanks!
@schalkms why would you wait for 1.0 with this PR?
It's a rule which is only run when type resolution is active which is experimental and needs special flags.

schalkms:
> It's a rule which is only run when type resolution is active which is experimental and needs special flags.

You are right. I forgot about this.
LGTM then

PR #1604, opened 2019-04-22
Don't trigger SpreadOperator rule when array copy not required

### LINKED ISSUE #554: Verify if the spread operator still has performance impacts
Like mentioned in #546 an issue with the spread operator seems to be fixed [KT-20462](https://youtrack.jetbrains.com/issue/KT-20462).

We must rerun the [benchmark](https://github.com/renatoathaydes/kotlin-hidden-costs-benchmark) to verify if the `SpreadOperator` rule is obsolete.

3flex:
Closes #554

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperatorSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperatorSpec.kt
@@ -1,7 +1,10 @@
 package io.gitlab.arturbosch.detekt.rules.performance
 
+import io.gitlab.arturbosch.detekt.test.KtTestCompiler
 import io.gitlab.arturbosch.detekt.test.compileAndLint
+import io.gitlab.arturbosch.detekt.test.lintWithContext
 import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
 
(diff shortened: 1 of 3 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperatorSpec.kt
```diff
@@ -12,41 +15,177 @@ import org.spekframework.spek2.style.specification.describe
...
+            }
+            it("doesn't report when using array constructor with spread operator") {
+                val code = """
+                    fun foo(vararg xs: Int) {}
+                    val testVal = foo(xs = *intArrayOf(1))
+                """
+                val actual = subject.compileAndLint(code)
+                assertThat(actual).hasSize(1)
+                assertThat(actual.first().message).isEqualTo(
+                    "In most cases using a spread operator causes a full copy of the array to be created before " +
```
schalkms: same here

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperatorSpec.kt
```diff
@@ -12,41 +15,177 @@ import org.spekframework.spek2.style.specification.describe
...
+            it("reports when array copy required using named parameters") {
+                val code = """
+                    val xsArray = intArrayOf(1)
+                    fun foo(vararg xs: Int) {}
+                    val testVal = foo(xs = *xsArray)
+                """
+                val actual = subject.compileAndLint(code)
+                assertThat(actual).hasSize(1)
+                assertThat(actual.first().message).isEqualTo(
+                    "In most cases using a spread operator causes a full copy of the array to be created before " +
```
schalkms: Please refactor the message string

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperatorSpec.kt
```diff
@@ -12,41 +15,177 @@ import org.spekframework.spek2.style.specification.describe
...
+            it("reports when array copy required without using named parameters") {
+                val code = """
+                    val xsArray = intArrayOf(1)
+                    fun foo(vararg xs: Int) {}
+                    val testVal = foo(*xsArray)
+                """
+                val actual = subject.compileAndLint(code)
+                assertThat(actual).hasSize(1)
+                assertThat(actual.first().message).isEqualTo(
+                    "In most cases using a spread operator causes a full copy of the array to be created before " +
```
schalkms: same here

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperatorSpec.kt
```diff
@@ -12,41 +15,177 @@ import org.spekframework.spek2.style.specification.describe
...
+
+            it("doesn't report when using array constructor with spread operator when varargs parameter comes first") {
+                val code = """
+                    fun <T> asList(vararg ts: T, stringValue: String): List<Int> = listOf(1,2,3)
+                    val list = asList(-1, 0, *arrayOf(1, 2, 3), 4, stringValue = "5")
+                """
+                val actual = subject.compileAndLint(code)
+                assertThat(actual).hasSize(1)
+                assertThat(actual.first().message).isEqualTo(
+                    "In most cases using a spread operator causes a full copy of the array to be created before " +
```
schalkms: same here

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperatorSpec.kt
```diff
@@ -12,41 +15,177 @@ import org.spekframework.spek2.style.specification.describe
...
+					println(test)
+				}
+                """
+                assertThat(subject.lintWithContext(environment, code)).isEmpty()
+            }
         }
 
-        it("without vararg") {
-            val code = """
+        context("without type resolution") {
```
schalkms: Why is there a separate without type resolution?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperatorSpec.kt
```diff
@@ -12,41 +15,177 @@ import org.spekframework.spek2.style.specification.describe
...
+            it("reports when array copy required using named parameters") {
+                val code = """
+                    val xsArray = intArrayOf(1)
+                    fun foo(vararg xs: Int) {}
+                    val testVal = foo(xs = *xsArray)
+                """
+                val actual = subject.lintWithContext(environment, code)
+                assertThat(actual).hasSize(1)
+                assertThat(actual.first().message).isEqualTo(
+                    "Used in this way a spread operator causes a full copy of the array to be created before calling " +
```
schalkms: Please refactor the message string

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperatorSpec.kt
```diff
@@ -12,41 +15,177 @@ import org.spekframework.spek2.style.specification.describe
...
+            it("reports when array copy required without using named parameters") {
+                val code = """
+                    val xsArray = intArrayOf(1)
+                    fun foo(vararg xs: Int) {}
+                    val testVal = foo(*xsArray)
+                """
+                val actual = subject.lintWithContext(environment, code)
+                assertThat(actual).hasSize(1)
+                assertThat(actual.first().message).isEqualTo(
+                    "Used in this way a spread operator causes a full copy of the array to be created before calling " +
```
schalkms: same here

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperatorSpec.kt
3flex: I had doubts about this approach when building the rule, so would be good to get a second opinion.

Basically with type resolution we will know if the value passed to the varargs parameter avoids the array copy performance penalty, so that case can be whitelisted when we have access to BindingContext.

Without type resolution all we can do is check for all uses of the spread operator and say that using it *might* result in performance penalty since we can't know if it's safe or not (this is existing behaviour).

So the rule will behave differently depending on whether type resolution is active or not. I think this is confusing you and I had doubts as well so perhaps another solution is better? Don't want it to be confusing or misleading. Open to suggestions.

3flex: Just to add, the separate test cases are there to ensure the correct message is displayed - with type resolution we can be sure there's an array copy, but without type resolution we can't be sure but can simply flag usage of the spread operator.

schalkms: I'm afraid, I only partially understand your comment.
Is there a case where type resolution is not available with RC15?

3flex: The only way to enable it is by manually passing `--classpath` to detekt CLI, and it's marked (deliberately) with a scary "experimental" warning in the flag's description.

So right now it's not convenient to use type resolution, so I assume most won't. Therefore two paths within the rule and tests for the rule.

This will change with #1597 which will make it very easy to use with Gradle.

Like I said it the approach I took with this rule might not be the best, so very open to suggestion

schalkms: Okay. I think your approach is fine (given the circumstances you stated).
I would like to have a comment above these tests to avoid further confusion and to prevent people from refactoring the code duplication.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperatorSpec.kt
3flex: Sorry, not sure what do you mean here by refactor?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperatorSpec.kt
schalkms: The exact same string you used for asserting the message is also used in line number 151, 138 and 126.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperatorSpec.kt
```diff
@@ -12,41 +15,171 @@ import org.spekframework.spek2.style.specification.describe
...
+    lateinit var environment: KotlinCoreEnvironment
+
+    beforeEachTest {
+        environment = KtTestCompiler.createEnvironment()
+    }
+
+    describe("SpreadOperator rule") {
+        /** This rule has different behaviour depending on whether type resolution is enabled in detekt or not. The two
+         * `context` blocks are there to test behaviour when type resolution is enabled and type resolution is disabled
+         * as different warning messages are shown in each case.
```
schalkms: Thanks.

PR #1606, opened 2019-04-23
Replace cli filters parameter with incudes and excludes

### LINKED ISSUE #1551: Replace cli flag "filters" with "exclusions" and "inclusions"
## Expected Behavior
I can exclude everything in `**/build/**` with the exception of "**build/generated/**".

## Current Behavior
I can specify if for example the `.*/build/.*` folder should be filtered.
Having generated code inside `build` which I want to test too, needs further adjustings with filters.
 
## Context
https://github.com/arturbosch/detekt/projects/2

arturbosch:
Closes #1551


3flex:
Closes #1551?

PR #1612, opened 2019-04-24
Update dependencies

3flex:
* Kotlin 1.3.30
* Gradle 5.4
* ktlint 0.32.0


detekt-test/build.gradle.kts
```diff
@@ -4,6 +4,7 @@ dependencies {
     implementation(kotlin("script-runtime"))
     implementation(kotlin("compiler-embeddable"))
     implementation(kotlin("script-util"))
+    implementation(kotlin("scripting-compiler-embeddable"))
```
schalkms: That seemed to be the missing dependency that caused the exceptions in the scripting compiler.
Thanks for your investigations.

PR #1615, opened 2019-04-25
Publish detekt-api docs to the detekt documentation site

### LINKED ISSUE #1574: Should detekt-api be Java-friendly?
I'm looking at the Javadoc produced by dokka for detekt-api. There are a few oddities related to Java interop (e.g. because BaseRule has a top level typealias a second class BaseRuleKt is generated in Javadoc - could be fixed by using `@file:JvmName`)

Question is, as a Kotlin focused project, do we care?

If not, should we change dokka config so instead of producing Javadoc (which will need several new annotations sprinkled through the code to get a nicer output) should we instead generate Jekyll compatible markdown which documents only the Kotlin-friendly API which won't require adding these annotations? That could then be published on GitHub pages.

3flex:
Two big changes;
1. Publishes detekt-api KDoc to the documentation website
2. Stops producing Javadoc JARs - instead, generates an empty Javadoc JAR to satisfy publishing requirements, but users will have to go to the docs site to view it (or read the source). See [what Square publishes for okio](https://search.maven.org/artifact/com.squareup.okio/okio/2.2.2/jar) as an example of this practice used in the wild.

dokka will no longer run on any module other than detekt-api. This reduces build time, and doesn't provide any value since @arturbosch [has previously said](https://github.com/arturbosch/detekt/issues/1206#issuecomment-445131559):

> I consider detekt-api, public parts of detekt-cli like CliArgs, the Baseline and YamlConfig and the gradle dsl / detekt extension as public api

Therefore only detekt-api (and the rules documentation, which is already being generated) needs to be published IMHO.

Closes #1574

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -200,9 +200,8 @@ subprojects {
     }
 
     val javadocJar by tasks.creating(Jar::class) {
-        dependsOn("dokka")
+        from(tasks.javadoc)
         archiveClassifier.set("javadoc")
-        from(buildDir.resolve("javadoc"))
     }
 
     artifacts {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -2,7 +2,6 @@ import com.jfrog.bintray.gradle.BintrayExtension
 import io.gitlab.arturbosch.detekt.Detekt
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 import org.gradle.api.tasks.testing.logging.TestLogEvent
-import org.jetbrains.dokka.gradle.DokkaTask
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import java.util.Date
 
@@ -13,7 +12,7 @@ plugins {
     id("com.github.johnrengelman.shadow") version "5.0.0" apply false
     id("org.sonarqube") version "2.7"
     id("io.gitlab.arturbosch.detekt")
-    id("org.jetbrains.dokka") version "0.9.18"
+    id("org.jetbrains.dokka") version "0.9.18" apply false
     jacoco
 }
 
@@ -66,7 +65,6 @@ subprojects {
         plugin("com.jfrog.bintray")
         plugin("maven-publish")
         plugin("io.gitlab.arturbosch.detekt")
-        plugin("org.jetbrains.dokka")
     }
 
     if (project.name != "detekt-test") {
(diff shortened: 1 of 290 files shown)


schalkms:
> Stops producing Javadoc JARs - instead, generates an empty Javadoc JAR to satisfy publishing requirements, but users will have to go to the docs site to view it

I think we can live with that.

build.gradle.kts
```diff
@@ -183,14 +182,14 @@ subprojects {
...
-        // https://github.com/Kotlin/dokka/issues/294
-        enabled = JavaVersion.current().isJava8
+    if (project.name == "detekt-api") {
+        apply { plugin("org.jetbrains.dokka") }
+        tasks.withType<DokkaTask> {
+            // suppresses undocumented classes but not dokka warnings https://github.com/Kotlin/dokka/issues/90
+            reportUndocumented = false
+            outputFormat = "jekyll"
+            outputDirectory = "$rootDir/docs/pages/kdoc"
+        }
```
marschwar: Why is this not defined in `detekt-api/build.gradle.kts` instead?

3flex: I only put it in the root build file because that's where most of the other project-specific configuration is defined, but like you I think it should go in the subproject's build file. I'll move it.

3flex:
I think there's a problem with Travis - build passed on JDK 8 but there was a failure with the `install-jdk.sh` script on 11 & 12. This should be good to go.

@arturbosch do you want to check this first or happy to merge and see how it looks when live? I haven't published on GitHub Pages before so I've possibly messed something up.

arturbosch:
Looks awesome! I'm wondering if we need to checkin all kdoc files into git?
Putting them into `.gitignore` should not hinder `jekyll -b serve` to work properly in github-pages?

3flex:
I'm very familiar with Jekyll or GitHub pages but I'll look into that, unless you know what changes I'd need to make? Agree that having thousands of lines of generated code stored in git doesn't make sense.

Are you OK with empty Javadoc jars?

arturbosch:
First I wanted to write: I think you do not need to make any changes at all as long as the kdoc is present in the docs folder when github pages builds the website.
But yeah the kdoc is built by gradle and jekyll builds the `_site` which gets hosted.

We do not use any special file or commando for jekyll, so I guess github is highly integrated with jekyll which just needs a docs folder and will automatically run `jekyll build` and `jekyll serve` for us.

We could switch to a `gh-pages` branch were the generated kdoc is not ignored and checked into git.

Or we just have all kind of documentation in git, so any changes which regenerates the doc are caught when merging like it is done now.

We need to have the `default-detekt-config.yml` in git because it is top searched in our repo.
Also the kdoc of rules is tied to our documentation. So lets just live with everything is in git and in master branch. This way we have uptodate documentation without extra work on other branches ^^.

Empty Javadoc is okay for me, as long as the sources gets published (with the kdoc) everyone extending detekt will still be happy :)

arturbosch:
@3flex @schalkms if you agree, lets merge this :)

PR #1617, opened 2019-04-26
Add rule to detekt missing or invalid package declaration.

### LINKED ISSUE #744: Test for missing package line
Not sure adding a test is necessary or whether this is the best solution but I'll suggest:

Could there be a test that fails if the file does not have a _'package com.example.package'_ statement?

marschwar:
Closes #744 

This rule makes sure there is a package declared and that the package matches the file location.

> In pure Kotlin projects, the recommended directory structure is to follow the package structure with the common root package omitted (e.g. if all the code in the project is in the "org.example.kotlin" package and its subpackages, files with the "org.example.kotlin" package should be placed directly under the source root, and files in "org.example.kotlin.foo.bar" should be in the "foo/bar" subdirectory of the source root).

In order to support this use case, the `rootPackage` must be specified

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -325,6 +325,9 @@ naming:
     parameterPattern: '[a-z][A-Za-z0-9]*'
     excludeClassPattern: '$^'
     ignoreOverriddenFunctions: true
+  InvalidPackageDeclaration:
+    active: false
+    rootPackage: ''
   MatchingDeclarationName:
     active: true
   MemberNameEqualsClassName:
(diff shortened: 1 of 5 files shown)


marschwar:
Yeah! Codecov is back :fireworks:

3flex:
This looks good! I think this would be a great one to run with `scripts/get_analysis_projects.groovy` to make sure there aren't any false positives. If you could please do that and report back with results I'll approve.

marschwar:
Sorry for the delay:

```
Complexity Report:
	- 200431 lines of code (loc)
	- 142346 source lines of code (sloc)
	- 102576 logical lines of code (lloc)
	- 30991 comment lines of code (cloc)
	- 24599 McCabe complexity (mcc)
	- 0 number of total code smells
	- 21 % comment source ratio
	- 239 mcc per 1000 lloc
	- 0 code smells per 1000 lloc
```

PR #1621, opened 2019-04-28
Create a KotlinScriptEngine pool to speed up tests

schalkms:
This speeds up the execution of detekt-rules:tests.
The load for compiling code is distributed over a number of engines.

This runs all the tests for detekt's rules in ~20 seconds on my machine.

### DIFF
--- a/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/KotlinScriptEngine.kt
+++ b/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/KotlinScriptEngine.kt
@@ -1,45 +1,27 @@
 package io.gitlab.arturbosch.detekt.test
 
-import org.jetbrains.kotlin.cli.common.environment.setIdeaIoUseFallback
-import org.jetbrains.kotlin.script.jsr223.KotlinJsr223JvmLocalScriptEngine
-import javax.script.ScriptEngineManager
 import javax.script.ScriptException
 
 /**
- * The object to create the Kotlin script engine for code compilation.
+ * The object to use the Kotlin script engine for code compilation.
  *
  * @author schalkms
  */
 object KotlinScriptEngine {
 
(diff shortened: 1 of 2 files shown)


3flex:
This looks good. How long did rules tests take before this change?

Is 8 a lucky number or chosen for some other reason? Ideally that value would match # cores, I think with `Runtime.getRuntime(). availableProcessors()`

schalkms:
@3flex the problem was that I saw 4 test fixtures taking around 15 seconds in #1620
We can't create a new `KotlinScriptEngine` for every test fixtures, because we would run into an `OutOfMemoryException`. 
I suppose that we run into a cache problem with the usage of a single `KotlinScriptEngine`. Therefore, I initialized more engines.
I mean it would be logical to have as much engines as available processors, but I'm not sure if we should limit it to that.

3flex:
I just remembered that tests aren't run in parallel at the moment, so probably no more than one engine will be in use at any time, so having 8+ would be no issue.

Can you try adding this to build file and see if there's much more of a speedup? This is what's recommended by Gradle:

```kotlin
tasks.withType<Test> {
    maxParallelForks = (Runtime.getRuntime().availableProcessors() / 2).takeIf { it > 0 } ?: 1
}
```

Hopefully they'll fly!

schalkms:
@3flex I still have some questions and I quote the documentation from this [link](https://guides.gradle.org/performance/#parallel_test_execution).

> Note that if you do run the tests in parallel, you will have to ensure that they are independent, i.e. don’t share resources, be that files, databases or something else. Otherwise there is a chance that the tests will interfere with each other in random and unpredictable ways.

We access some files from multiple tests, although we only read from them.

```Kotlin
tasks.withType<Test> {
    maxParallelForks = (Runtime.getRuntime().availableProcessors() / 2).takeIf { it > 0 } ?: 1
}
```

Detekt uses `Spek`. Does it pick up our tests classes since they don't inherit from type `Test`?
We would also need to adapt the `ScriptEnginePool` to be thread-safe.

PR #1623, opened 2019-04-28
Reports no duplicated findings in empty ruleset

### LINKED ISSUE #1605: Duplicated findings are being added to `Detektion.findings`
Some issues are appearing twice in the findings list.

## Expected Behavior
Each detected issue should appear only once

## Observed Behavior
Some issues appear twice in the findings list

Issues that I noticed appears twice in output with consistency so far:
- `EmptyClassBlock`
- `EmptyFinallyBlock`

## Steps to Reproduce
Run the analysis on a file like this one:
```
class EmptyBlocks {
    interface EmptyInterface {}
    fun exceptionHandling() {
        try {
        } finally {
        }
    }
}
```
With the command:
```
java -jar detekt-cli/build/libs/detekt-cli-1.0.0-RC14-all.jar -c linter-config.yml -i EmptyBlocks.kt
```

Observe the output will contain duplicated findings.
image

## Context
I was writing unit tests for the company linter integration with Detekt.
They failed and because the of duplicated findings.

## Your Environment
* Version of detekt used:
* Operating System and version: macOS Mojave, version 10.14.4 (18E226)

schalkms:
Closes #1605

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyBlocks.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyBlocks.kt
@@ -66,17 +66,15 @@ class EmptyBlocks(val config: Config = Config.empty) : MultiRule() {
 
     override fun visitKtFile(file: KtFile) {
         emptyKtFile.runIfActive { visitFile(file) }
+        file.declarations.filterIsInstance<KtClassOrObject>().forEach {
+            emptyClassBlock.runIfActive { visitClassOrObject(it) }
+        }
         super.visitKtFile(file)
     }
 
     override fun visitCatchSection(catchClause: KtCatchClause) {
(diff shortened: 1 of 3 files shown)


schalkms:
@3flex I'm sorry. I want to clarify some things.

> Apologies, slight issue with running tests in the IDE - new test in EmptyBlocksMultiRuleSpec does flag the issue

Yes, it works as expected in my opinion.

> but the new test in EmptyClassBlockSpec passes without any other code changes, so that one doesn't flag it.

Yes, because the rule `EmptyClassBlock` itself works. The issue occurs in the `EmptyBlocksMultiRule` multi rule, where detekt determines how to call and visit the `EmptyClassBlock` rule and all the others. The logic for calling the multi rule was simply wrong.

> I think the tests will need to be altered somehow to make sure they actually pick up the unintended behaviour to avoid future regressions.

I agree. I think the multi rule for calling and visiting the rules should be tested more.

Issue #1616, opened 2019-04-26
Gradle build failed


IgorGanapolsky:
## Expected Behavior
I should be able to follow the README and run `./gradlew build shadowJar`

## Observed Behavior
Got an error: 
> FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':detekt-gradle-plugin:test'.
> There were failing tests. See the report at: file:///Users/igor.ganapolsky/workspace/git/detekt/detekt-gradle-plugin/build/reports/tests/test/index.html

## Steps to Reproduce
Follow the README

## Context
I am trying to build this library for the first time

## Your Environment

* Version of detekt used: latest master branch
* Version of Gradle used (if applicable): 5.4
* Operating System and version: Mac OS 10.13.6
* Link to your project (if it's a public repository):

3flex:
Do you have the same test output as #1611?

arturbosch:
Both run on macOS.
What Gradle version did you used?

schalkms:
I'm pretty sure that this is a duplicate.

@IgorGanapolsky 
If this is not the case please ping us here.
Please take a look at the linked issue and my comment.

3flex:
@IgorGanapolsky hopefully a fix has just been merged to master branch, would be great if you are able to checkout master and confirm? If it's not fixed please comment in #1622. Thank you!

IgorGanapolsky:
@3flex The test report in ticket #1611 states:
> baseline /var/folders/_h/3hlxwcnx7ndf76d7f5mzd0sm0000gn/T/applyPlugin2916622894074676092.tmp/detekt-baseline.xml

However, I do not have such a folder on my machine.  What does it mean?

Also, my tests are still failing (on a second Macbook Pro now).  Please see attached.
[reports.zip](https://github.com/arturbosch/detekt/files/3127547/reports.zip)

3flex:
That's a temporary folder path generated by the tests - not expected to match a path on your machine as the generated path is random.

I'll wait for confirmation (or not) on the other ticket, if that's closed though I'll reopen this as it sounds like you still have issues. Unfortunately I don't have a mac to test on so if you're able to identify the issue & fix please submit a PR. @pcl may also be able to help.

IgorGanapolsky:
I see a test failure such as
> passes multiple custom report params to cli
java.lang.AssertionError: 
Expecting:
[...]

Issue #1611, opened 2019-04-24
DetektTaskDslTest issues on macOS


pcl:
I can't build master. I get 8 errors (seemingly including some duplicates) when building `detekt-gradlew-plugin`. Attached is the report output directory.

I'm running Java 8. I also tried building the last release tag, and got the same errors.

I don't have a jetbrains.com account, so can't sign into the Slack channel. I think I was able to get in in the past; is that a new limitation?

3flex:
> I get 8 errors (seemingly including some duplicates) when building detekt-gradlew-plugin

Can you please provide the errors?

> Attached is the report output directory.

That would be good if it's available - I don't see any attachments?

> I don't have a jetbrains.com account, so can't sign into the Slack channel.

http://slack.kotlinlang.org/

pcl:
Argh. I uploaded a tgz, but only zips are allowed. Didn't notice the error window. See attached, for real this time.

Thanks for the link; might be worthwhile to change the Slack link in the readme to point to the top-level survey too, or include a link to the survey on the Slack auth screen if such a thing is customizable.

[detekt-gradle-plugin-errors.zip](https://github.com/arturbosch/detekt/files/3116465/detekt-gradle-plugin-errors.zip)

schalkms:
The problem lies in the `DetektTaskDslTest`.
This seems to be a Mac only problem.
Can you please comment out the mentioned class and try to build detekt again?

pcl:
Yep, that was it!

3flex:
@pcl have just merged your PR, can you please checkout master and build to confirm this is fixed? Thanks!

pcl:
It didn't quite work -- I was lazy and rebased my PR onto origin/master, but someone added a new test to that file over the weekend that also used `absolutePath`! Additional commit forthcoming...

schalkms:
@pcl please submit a PR with the referenced changes and add me as a reviewer. 
I checked out your branch. Everything is running fine on my Mac.

schalkms:
Closed with 2a0c46f

PR #1624, opened 2019-04-29
Issue #1611: Fix path issue with DetektTaskDslTest

pcl:
macOS puts various temporary directories into a separate mount
point, and then creates symlinks from the more-traditional
POSIX layout locations to the separate mount point. This change
uses `canonicalPath` instead of `absolutePath` to align the
expected locations with what happens during shell expansion.

This might also address #1616.


pcl:
I haven't tried this on any other platforms yet. Are there supported platforms that aren't tested in the CI pipeline?

pcl:
FYI, the Java 8 build worked, but the Java 11 and Java 12 builds failed while installing the JVM. Looks like a systemic issue.

3flex:
> Are there supported platforms that aren't tested in the CI pipeline?

It's tested in CI on Linux (via Travis) and Windows (via AppVeyor), and we'll look to introduce testing on macOS on Travis as well since there are differences on that platform that have been identified. Thanks for the quick fix!

PR #1625, opened 2019-04-29
Add CI for macOS - closes #1622

### LINKED ISSUE #1622: CI for macOS
After the build failure report in #1611 and #1616 on macOS I suggest to also include this in detekt's CI process.
Travis CI supports macOS for instance.  [reference](https://docs.travis-ci.com/user/reference/osx/)

schalkms:


3flex:
Ah, wish I'd seen this! Started working on it right after you posted the PR. I think this will have an issue, because according to Travis docs

> macOS images up to xcode9.3 can only switch up to Java 8, and images xcode9.4 and later can switch to Java 10 (if pre-installed) and later.

The default image is xcode9.4 so I don't think the Java 8 build will work with the `jdk` switcher - instead will have to use the build matrix feature. I'll submit a PR in a moment.

3flex:
See #1626

schalkms:
Discussion continued in #1626

PR #1629, opened 2019-04-29
Issue #1611: Fix path issue with DetektTaskDslTest

pcl:
... again! My previous commit was a rebase from a branch that
was just a couple days old; I didn't test after rebasing. In
the few days since I had pulled, an additional change came in
that asserted an `absolutePath` instead of a `canonicalPath`.


pcl:
@arturbosch : Here you go.

I can't add you as a reviewer; I think your repo is configured to not allow that or something.

PR #1631, opened 2019-04-30
Document build-upon-default-config

reitzig:
Flag `build-upon-default-config` has not yet been documented (in the "getting started" guides). Since the default behaviour is rather unexpected (most other tools I know have this behaviour by default), I propose adding a mention right where new users would find it.

See #1248 that introduced the flag, and also #500.


3flex:
Thanks @reitzig, good first contribution! I would only suggest changing the text in `configuration.md` to say `--build-upon-default-config` instead of just `build-upon-default-config`.

schalkms:
Thanks for providing feedback and for improving the docs!

reitzig:
Happy to help; thanks for your work and the snappy processing of this PR!

PR #1632, opened 2019-05-01
Whitelist provideDelegate operator in UnusedImports rule

### LINKED ISSUE #1608: False positive from UnusedImports on org.gradle.kotlin.dsl.provideDelegate
Hi,

I've found that when I'm writing the gradle buildSrc using kotlin dsl, and use the kotlin delegation, I would need to import `org.gradle.kotlin.dsl.provideDelegate`. But Detekt would give me an `UnusedImports` error in this case.

For example, with the following code, detekt task would fail:

```kotlin
import org.gradle.api.Plugin
import org.gradle.api.Project
import org.gradle.api.tasks.WriteProperties
import org.gradle.kotlin.dsl.getValue
import org.gradle.kotlin.dsl.provideDelegate  // got UnusedImports from this line
import org.gradle.kotlin.dsl.registering

class DumpVersionProperties : Plugin<Project> {
    override fun apply(target: Project) {
        with(target) {
            val dumpVersionProperties by tasks.registering(WriteProperties::class) {
                setProperties(mapOf("version" to "1.2.3"))
                outputFile = rootDir.resolve("version.properties")
            }

        }
    }
}
```

## Expected Behavior
It should not report `UnusedImports` if I do use it for delegation.

## Observed Behavior
Detekt task fails and report there's an `UnusedImports` error.

## Your Environment

* Version of detekt used: 1.0.0-RC14
* Version of Gradle used (if applicable): 5.4

3flex:
Fixes #1608

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportsSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportsSpec.kt
@@ -233,5 +233,19 @@ class UnusedImportsSpec : Spek({
 			""".trimIndent()
             assertThat(subject.lint(code)).isEmpty()
         }
+
+        it("should not report import of provideDelegate operator overload - #1608") {
+            val code = """
+                import org.gradle.api.tasks.WriteProperties
+                import org.gradle.kotlin.dsl.getValue
+                import org.gradle.kotlin.dsl.provideDelegate // this line specifically should not be reported
+                import org.gradle.kotlin.dsl.registering
+
+                class DumpVersionProperties {
+                    private val dumpVersionProperties by tasks.registering(WriteProperties::class)
+                }
+                """
+            assertThat(subject.lint(code)).isEmpty()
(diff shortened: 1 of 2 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportsSpec.kt
```diff
@@ -233,5 +233,27 @@ class UnusedImportsSpec : Spek({
...
+                import org.gradle.kotlin.dsl.provideDelegate // this line specifically should not be reported
+                import org.gradle.kotlin.dsl.registering
+
+                class DumpVersionProperties : Plugin<Project> {
+                    override fun apply(target: Project) {
+                        with(target) {
+                            val dumpVersionProperties by tasks.registering(WriteProperties::class) {
+                                setProperties(mapOf("version" to "1.2.3"))
+                                outputFile = rootDir.resolve("version.properties")
+                            }
```
schalkms: Could we simplify this example for this test?

PR #1637, opened 2019-05-03
Make KtFile.unnormalizeContent a public method

hick209:
An alternative to #1593.

Making `KtFile.unnormalizeContent` available will make it possible to add implement the same functionality, but without needing to add a dependency to `org.jetbrains.kotlin:kotlin-compiler-embeddable`.


hick209:
Actually, I don't need this as our codebase is already normalized.

PR #1641, opened 2019-05-07
Lazily add `detekt` task as a dependency of the `check` task

### LINKED ISSUE #1640: Detekt gradle task is not executed upon check
## Expected Behavior
The detekt task should automatically be executed when the user runs `gradle check` or `gradle build`

## Observed Behavior
The detekt task can be run by using `gradle detekt` but it is not triggered on any of the tasks mentioned above.

## Steps to Reproduce
Use any gradle project available and run `gradle clean build`

## Context
Even though there is a test for exactly this action it does not behave the same way when using the plugin. 
Knowing that this works perfectly fine with the ktlint gradle plugin, I did some digging. Since the check dependency in the both plugins are essentially configured the same way, it had to be something else. Turns out that the way the plugin is added to the project makes a difference.

NOT WORKING
```
plugins {
    id "io.gitlab.arturbosch.detekt" version "1.0.0-RC14"
}
```

WORKING
```
buildscript {
    dependencies {
        classpath "io.gitlab.arturbosch.detekt:detekt-gradle-plugin:1.0.0-RC14"
    }
}

allprojects {
}
```

If someone could please explain or provide me with pointers on what the difference is, I can try to fix it. The most annoying part is, that it is not reproducible in the tests.

3flex:
Fixes #1640 

This uses ProjectBuilder instead of GradleRunner for its tests. It's much lighter weight and runs more quickly and is useful for testing task and plugin configuration. In this case, I created a new test, which failed, then updated the plugin setup code and saw the test pass.

### DIFF
--- a/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPluginTest.kt
+++ b/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPluginTest.kt
@@ -1,0 +1,26 @@
+package io.gitlab.arturbosch.detekt
+
+import org.assertj.core.api.Assertions.assertThat
+import org.gradle.api.Task
+import org.gradle.language.base.plugins.LifecycleBasePlugin
+import org.gradle.testfixtures.ProjectBuilder
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+
+object DetektPluginTest : Spek({
+    describe("detekt plugin") {
+
+        fun Task.dependencies() = taskDependencies.getDependencies(this)
+
+        it("lazily adds detekt as a dependency of the `check` task") {
+            val project = ProjectBuilder.builder().build()
+
+            /** Ordering here is important - to prove lazily adding the dependency works the LifecycleBasePlugin must be
+             * added to the project after the detekt plugin. */
+            project.pluginManager.apply(DetektPlugin::class.java)
+            project.pluginManager.apply(LifecycleBasePlugin::class.java)
+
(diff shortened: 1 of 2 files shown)


detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPluginTest.kt
```diff
@@ -0,0 +1,26 @@
...
+    describe("detekt plugin") {
+
+        fun Task.dependencies() = taskDependencies.getDependencies(this)
+
+        it("lazily adds detekt as a dependency of the `check` task") {
+            val project = ProjectBuilder.builder().build()
+
+            /** Ordering here is important - to prove lazily adding the dependency works the LifecycleBasePlugin must be
+             * added to the project after the detekt plugin. */
+            project.pluginManager.apply(DetektPlugin::class.java)
```
marschwar: A simple block comment instead of a KDoc comment should be used here.

marschwar:
The change itself seems plausible. Well done. 

I tried to run `gradle publishToMavenLocal` to use your version, but it does not seem to be the right way to test a local version of the plugin.

3flex:
To apply a plugin using `plugins` block a composite build is probably the best way if you're familiar, but otherwise the method you described should work.

It's a bit inconvenient either way. We can merge this without your check if you're happy, as I'm confident based on the behaviour of the tests that this will solve the issue.

marschwar:
I will give it one last shot today to test it. But I am fine with merging it right away as well.

marschwar:
I can confirm that the fix works. Thanks a lot for picking it up.

PR #1646, opened 2019-05-11
Migrate to lazy searching of child nodes

arturbosch:


arturbosch:
Yep, it was a byproduct of fixing one Rule. I needed to search for a child with `collectByType.findFirst`.
Mostly we use `find` or `any` combined with this query. Also `parentOfType` was redundant.

PR #1651, opened 2019-05-14
Replace test pattern concept with rule excludes

### LINKED ISSUE #1552: Replace test-pattern with default configured exclusions in the default-detekt-config
## Expected Behavior
Use exclusions and inclusions to allow more fine-grained running configuration of rules for paths.
  
## Current Behavior
A non extensible concept allows to turn off rules for test code.

## Context
https://github.com/arturbosch/detekt/projects/2

arturbosch:
Closes #1552


3flex:
Not sure if this relates to this PR or 1648, but I think it's better to have an array rather than comma-delimited string for the `excludes` and `includes` options in the yaml config.

Makes for cleaner diffs when configs change and better reflects the data type.

arturbosch:
Yep, your right. However in some previous discussions we identified that we already use (or misuse) comma separated entries in a string for many rule properties and we wanted to stay consistent.

In theory every rule and rule set could use 1-2 (excludes/includes) of these yaml lists which would make the config file even larger.
What do you think @schalkms ?

arturbosch:
Feels so good to delete some code :)

PR #1654, opened 2019-05-20
Move autocorrect to cli

arturbosch:
The last orphant config property which is suited best to be part of the cli args and gradle properties.
A new gradle plugin property must be merged when all other modules get a RC15 release.

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
@@ -83,6 +83,13 @@ class CliArgs : Args {
             "Additional configuration files can override properties but not the 'active' one.")
     var failFast: Boolean = false
 
+    @Parameter(names = ["--auto-correct", "-ac"],
+        description = "Allow rules to auto correct code if they support it. " +
+            "The default rule sets do NOT support auto correcting and won't change any line in the users code base. " +
+            "However custom rules can be written to support auto correcting. " +
+            "The additional 'formatting' rule set, added with '--plugins', does support it and needs this flag.")
+    var autoCorrect: Boolean = false
+
     @Parameter(names = ["--debug"],
         description = "Prints extra information about configurations and extensions.")
     var debug: Boolean = false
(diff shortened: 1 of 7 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -83,6 +83,13 @@ class CliArgs : Args {
             "Additional configuration files can override properties but not the 'active' one.")
     var failFast: Boolean = false
 
+    @Parameter(names = ["--auto-correct", "-ac"],
+        description = "Allow rules to auto correct code if they support it. " +
+            "The default rule sets do NOT support auto correcting and won't change any line in the users code base. " +
+            "However custom rules can be written to support auto correcting. " +
+            "The additional 'formatting' rule set, added with '--plugins', does support it and needs this flag. ")
```
schalkms: Extra space after "."

docs/pages/gettingstarted/cli.md
```diff
@@ -43,26 +57,31 @@ Usage: detekt [options]
...
-      Apply values from config files as changes to the default configuration
-      (instead of starting from an empty configuration).
-      Default: false
+    --excludes, -ex
+      Globing patterns describing paths to exclude from the analysis.
     --fail-fast
-      Shortcut for 'build-upon-default-config' together with all available rules active 
-      and exit code 0 only when no code smells are found.
-      Additional configuration files can override properties but not the 'active' one.
-    --filters, -f
```
schalkms: Why was —filters removed?

arturbosch: It got replaced with `excludes` in cli and gradle plugin.
Or you mean that I should not edit the documentation for now until RC15 is finished? Makes sense ...

schalkms: @arturbosch I would update it and create a separate PR for now, so that you don't forget to merge it when RC15 gets released.
When do we plan to release RC15?

arturbosch: When doing #1657 I noticed that the new `KotlinEnvironment` settings which leverage symbol solving, do not allow to modify the ast. So formatting and potential custom rules are broken for now.
Need to get that fixed. Until then we can't release a new version :(

PR #1655, opened 2019-05-20
Do not use relativePath as it is not set anymore and will crash the detekt run

arturbosch:
The kotlin environment handles now the compilation of kt files and we do not add a relative path to the user data anymore.


arturbosch:
In a followup PR I will refactor the `core` and `api` module classes to get rid of all the other code which does not use the new `createCompilerConfiguration` function.

PR #1656, opened 2019-05-21
Update spek to 2.0.4

arturbosch:


3flex:
Spek now times out and fails tests that take more than 10 seconds to execute.

Suggest updating to 2.0.5 which allows setting a default timeout.

Also needs updating in the Gradle plugin (overriding timeouts will be more important here).

arturbosch:
Thanks for the info!

arturbosch:
Damn the CI runs are so slow. So the DetektCreateBaselineSpec is timing out even with 60 seconds timeout ....

PR #1657, opened 2019-05-21
Unify use of kotlin environment and make AST modification work again

arturbosch:
@3flex now that we only use the `createKotlinCoreEnvironment` endpoint and get the KtFile's from it, we should get rid of the static `PsiProject` and `PsiFactory` in `ProjectExtension.kt`.
However I run into the problem that the AST created by this method does not make the AST mutable ( see my changes `KotlinCoreEnvironment.createForProduction(...).apply { makeMutable(...) }`.
Which beaks formatting and can break custom rules ... do you know of any params we can specify in `CompilerConfiguration` maybe?

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/DetektPomModel.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/DetektPomModel.kt
@@ -1,0 +1,42 @@
+package io.gitlab.arturbosch.detekt.api.internal
+
+import org.jetbrains.kotlin.com.intellij.openapi.extensions.ExtensionPoint
+import org.jetbrains.kotlin.com.intellij.openapi.extensions.Extensions
+import org.jetbrains.kotlin.com.intellij.openapi.project.Project
+import org.jetbrains.kotlin.com.intellij.openapi.util.UserDataHolderBase
+import org.jetbrains.kotlin.com.intellij.pom.PomModel
+import org.jetbrains.kotlin.com.intellij.pom.PomModelAspect
+import org.jetbrains.kotlin.com.intellij.pom.PomTransaction
+import org.jetbrains.kotlin.com.intellij.pom.impl.PomTransactionBase
+import org.jetbrains.kotlin.com.intellij.pom.tree.TreeAspect
+import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.TreeCopyHandler
+import sun.reflect.ReflectionFactory
+
+/**
+ * Adapted from https://github.com/pinterest/ktlint/blob/master/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
+ * Licenced under the MIT licence - https://github.com/pinterest/ktlint/blob/master/LICENSE
+ */
+class DetektPomModel(project: Project) : UserDataHolderBase(), PomModel {
+
+    init {
+        val extensionPoint = "org.jetbrains.kotlin.com.intellij.treeCopyHandler"
+        val extensionClassName = TreeCopyHandler::class.java.name!!
+        arrayOf(Extensions.getArea(project), Extensions.getArea(null))
+            .filter { !it.hasExtensionPoint(extensionPoint) }
+            .forEach { it.registerExtensionPoint(extensionPoint, extensionClassName, ExtensionPoint.Kind.INTERFACE) }
+    }
+
+    override fun runTransaction(transaction: PomTransaction) {
+        (transaction as? PomTransactionBase ?: error("PomTransactionBase type expected")).run()
+    }
+
+    override fun <T : PomModelAspect?> getModelAspect(aspect: Class<T>): T? {
+        if (aspect == TreeAspect::class.java) {
+            val constructor = ReflectionFactory.getReflectionFactory()
+                .newConstructorForSerialization(aspect, Any::class.java.getDeclaredConstructor())
+            @Suppress("UNCHECKED_CAST")
+            return constructor.newInstance() as T
+        }
+        return null
+    }
+}
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/KotlinEnvironmentUtils.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/KotlinEnvironmentUtils.kt
@@ -18,6 +18,10 @@ import org.jetbrains.kotlin.config.JvmTarget
 import java.io.File
 import java.nio.file.Path
 
+/**
+ * Creates an environment instance which can be used to compile source code to KtFile's.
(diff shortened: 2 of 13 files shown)


3flex:
I haven't looked to closely yet, but you might try `KotlinCoreEnvironment.project createForProduction(...).apply`

Instead of `KotlinCoreEnvironment.createForProduction(...).apply`.

You want to make the project that is in the KotlinCoreEnvironment mutable but it looks like you're trying to make the KotlinCoreEnvironment itself mutable with won't work.

arturbosch:
Not really, the `KolinEnvironment` needs to be returned here, so getting the project here is fine  `makeMutable(project as? MockProject` :).
Hadn't a change to look at it again though.

arturbosch:
@schalkms @3flex so I found out that the `KotlinCoreEnvironment.getSourceFiles` produces `KtFile`s with `isWritable=false`. I find it hard to find out what the differences are under the hood to make it work by using `getSourceFiles`.
It seems that the `BindingContext` does not care about where the `KtFile`s are coming from, so the new rules with type solving still work and formatting now too :).

schalkms:
I'll review that tomorrow.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/DetektPomModel.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.jetbrains.kotlin.com.intellij.pom.impl.PomTransactionBase
+import org.jetbrains.kotlin.com.intellij.pom.tree.TreeAspect
+import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.TreeCopyHandler
+import sun.reflect.ReflectionFactory
+
+/**
+ * Adapted from https://github.com/pinterest/ktlint/blob/master/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
+ * Licenced under the MIT licence - https://github.com/pinterest/ktlint/blob/master/LICENSE
+ */
+class DetektPomModel(project: Project) : UserDataHolderBase(), PomModel {
```
arturbosch: This is just `makeMutable` extracted to a class, nothing changed.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/KotlinEnvironmentUtils.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.jetbrains.kotlin.com.intellij.openapi.util.Disposer
+import org.jetbrains.kotlin.com.intellij.pom.PomModel
+import org.jetbrains.kotlin.config.CommonConfigurationKeys
+import org.jetbrains.kotlin.config.CompilerConfiguration
+import org.jetbrains.kotlin.config.JVMConfigurationKeys
+import org.jetbrains.kotlin.config.JvmTarget
+import java.io.File
+import java.nio.file.Path
+
+fun createKotlinCoreEnvironment(configuration: CompilerConfiguration = CompilerConfiguration()): KotlinCoreEnvironment {
```
arturbosch: This file is just deleting static globals of `ProjectExtensions` and moving the two remaining function to `api/internal`

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
```diff
@@ -35,34 +31,27 @@ class DetektFacade(
     private val classpath = settings.classpath
     private val pathFilters = settings.pathFilters
     private val jvmTarget = settings.jvmTarget
-    private val compilerConfiguration = createCompilerConfiguration(inputPaths, classpath, jvmTarget)
-    private val environment = createKotlinCoreEnvironment(compilerConfiguration)
+    private val environment = settings.environment
```
arturbosch: The `ProcessingSettings` take now care of the kotlin environment. This way the same instance is shared between DetektFacade and KtTreeCompiler.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/KotlinEnvironmentUtils.kt
schalkms: We might want to have this information in a comment.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtCompiler.kt
```diff
@@ -1,31 +1,36 @@
...
  * @author Artur Bosch
  */
-open class KtCompiler {
+open class KtCompiler(
+    environment: KotlinCoreEnvironment = createKotlinCoreEnvironment()
+) {
 
-    protected val psiFileFactory: PsiFileFactory = PsiFileFactory.getInstance(psiProject)
+    val psiFileFactory: PsiFileFactory = PsiFileFactory.getInstance(environment.project)
+    val psiFactory: KtPsiFactory = KtPsiFactory(environment.project, false)
```
schalkms: Why do we need 2 fileFactories?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
```diff
@@ -86,23 +75,6 @@ class DetektFacade(
         }
     }
 
-    fun run(project: Path, files: List<KtFile>): Detektion = runOnFiles(project, files)
-
-    private fun runOnFiles(current: Path, files: List<KtFile>): DetektResult {
```
schalkms: It's way better than before in terms of complexity. 💯

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtCompiler.kt
schalkms: Is the psiFactory only used in `AnnotationExcluderSpec`?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
schalkms: Please let us also add this one as a comment.

arturbosch:
> Looks good!
> I have some questions.
> Please take a look at my comments.
> 
> Does this have any effects on the runtime performance of detekt?
No, everything stays the same. This even changed the parsing logic back to how it was before.

> Did you benchmark it?
Just a bit, detekt on detekt itself with java 8 needs 3.4 seconds, with java 11 its ~4 seconds.
That are also the same numbers as before if I remember correctly.
RC-13 seems to be 0.2 seconds slower :)
> Shall we add performance tests to detekt?
That would also be a good step. However CI is kinda unreliable. You have experience with performance tests on CI?

PR #1661, opened 2019-05-27
Add support for JSR test infrastructure for type resolved rules

schalkms:

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNullSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNullSpec.kt
@@ -1,7 +1,7 @@
 package io.gitlab.arturbosch.detekt.rules.style
 
 import io.gitlab.arturbosch.detekt.test.KtTestCompiler
-import io.gitlab.arturbosch.detekt.test.lintWithContext
+import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
 import org.assertj.core.api.Assertions.assertThat
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.spekframework.spek2.Spek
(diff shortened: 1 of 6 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNullSpec.kt
```diff
@@ -20,52 +20,52 @@ object UselessCallOnNotNullSpec : Spek({
...
 
         it("only reports on a Kotlin list") {
             val code = """
-                |fun main() {
-                |    val javaList = Java().list.orEmpty()
-                |    val list = listOf(1, 2, 3).orEmpty()
-                |}
+                fun main() {
+                    val javaList = Java().list.orEmpty()
+                    val list = listOf(1, 2, 3).orEmpty()
```
schalkms: I have no idea what this test should assert.
@3flex can you explain this to me?

3flex: It's testing that the rule only reports when the extensions are used on a non-nullable Kotlin list. If the `orEmpty()` extension were used on a list or some other type that is not a Kotlin list then it shouldn't report any issue.

`Java().list` is dummy code and won't resolve, so I can see why this is causing an issue with the PR, but hopefully that explanation of the purpose of the test makes sense and gives enough information to modify?

arturbosch: Yeah if it comes from a Java file the returned type is like `String!`. I think that is what the is testing.
I don't think it can be tested without extra files in the resources and with adapted KotlinEnvironment?

schalkms: I updated the test case. Please take another look at it.

PR #1664, opened 2019-05-29
Rework UnderscoresInNumericLiterals Rule

### LINKED ISSUE #1538: UnderscoresInNumericLiterals flagged for floats
```kotlin
assertThat(expressionEvaluator.evaluate("π").toDouble()).isEqualTo(3.1415926535897932385)
```

is being flagged by UnderscoresInNumericLiterals and it should not. RC14

tylerbwong:
Fixes #1538.

This patch updates the `UnderscoresInNumericLiterals` rule to only consider the left side of the decimal place when running the rule through numeric literals that are a `Float` or `Double`.


tylerbwong:
Hey @schalkms, thanks for taking a look! I believe there already is a test for a hexadecimal value at [line 182](https://github.com/arturbosch/detekt/blob/master/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiteralsSpec.kt#L182), but please let me know if you want more added 👍

schalkms:
@tylerbwong please take a look at the following comment in the referenced issue. I think these two test cases are nice to have.
https://github.com/arturbosch/detekt/issues/1538#issuecomment-476550279

PR #1673, opened 2019-06-02
 Fix MaxLineLength and suppressing issues with multiline strings

### LINKED ISSUE #1667: Unable to Suppress MaxLineLength for multiline string other than on file level
## Expected Behavior

The suppression applied to the variable in the code below, should prevent detekt from reporting the next line as exceeded the max line length.

```
class Example_Suppress_Variable {
    fun example() {
        @Suppress("MaxLineLength")
        val string = """
            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
        """.trimIndent()
    }
}
```

## Observed Behavior

The suppression above doesn't work, unless placed on the file level. I have tried placing it also on the function/class level, which didn't work too. The same for `@SuppressWarnings` alternative.

For simplicity I tested against the default detekt config, namely:

```
  MaxLineLength:
    active: true
    maxLineLength: 120
    excludePackageStatements: true
    excludeImportStatements: true
    excludeCommentStatements: false
```

## Steps to Reproduce

1. Copy-paste the code above into your project
2. Set-up detekt with the default configuration
3. Run detekt

I have also prepared a sample project demonstrating the issue, available from the link below.

## Your Environment
* Version of detekt used: 1.0.0-RC14
* Link to your project (if it's a public repository): https://github.com/Jedyny/max-line-example

cortinico:
This PR replaces/fixes the function `findAnnotatedStatementInLine`
that  is used for reporting the `MaxLineLenght` issue. Now the function is
searching  for the first `KtElement` in the parents of the affected line.

This will fix a bug that prevents suppressing `MaxLineLenght` other than
at the File level. (See 7c5aec4)

Moreover I've cleaned up the baseline file from all the `MaxLineLenght` and fixed them (See 8947d18)

Fixes #1667

### DIFF
--- a/detekt-rules/src/test/resources/cases/MaxLineLength.kt
+++ b/detekt-rules/src/test/resources/cases/MaxLineLength.kt
@@ -18,6 +18,12 @@ class MaxLineLength {
             very long multiline String that will break the MaxLineLength
         """.trimIndent()
 
+    val longMultiLineFieldWithLineBreaks =
+        """
+            This is anotehr very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very
+            very long multiline String with Line Break that will break the MaxLineLength
+        """.trimIndent()
+
     fun main() {
         val thisIsAVeryLongValName = "This is a very, very long String that will break the MaxLineLength"
 
--- a/detekt-rules/src/test/resources/cases/MaxLineLength.kt
+++ b/detekt-rules/src/test/resources/cases/MaxLineLength.kt
@@ -4,9 +4,19 @@ package cases
 class MaxLineLength {
     companion object {
         val LOREM_IPSUM = "Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua."
+
+        val A_VERY_LONG_MULTI_LINE = """
+            This is anotehr very very very very very very very very, very long multiline String that will break the MaxLineLength"
+        """.trimIndent()
     }
 
     val loremIpsumField = "Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua."
+
+    val longMultiLineField = """
+            This is anotehr very very very very very very very very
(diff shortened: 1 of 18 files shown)


reports/baseline.xml
```diff
@@ -3,19 +3,6 @@
   <Blacklist></Blacklist>
   <Whitelist>
     <ID>LargeClass:UnusedPrivateMemberSpec.kt$UnusedPrivateMemberSpec : Spek</ID>
-    <ID>MaxLineLength:BaselineFacade.kt$BaselineFacade$val blackFiltered = whiteFiltered.filterNot { finding -&gt; listings.second.ids.contains(finding.baselineId) }</ID>
-    <ID>MaxLineLength:ComplexityReportGenerator.kt$ComplexityReportGenerator.Factory$fun create(detektion: Detektion): ComplexityReportGenerator</ID>
-    <ID>MaxLineLength:Config.kt$io.gitlab.arturbosch.detekt.api.Config.kt</ID>
```
schalkms: Nice cleanup 👍

detekt-rules/src/test/resources/cases/MaxLineLength.kt
```diff
@@ -4,10 +4,20 @@ package cases
...
+
+        val A_VERY_LONG_MULTI_LINE = """
+            This is anotehr very very very very very very very very, very long multiline String that will break the MaxLineLength"
+        """.trimIndent()
     }
 
     val loremIpsumField = "Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua."
 
+    val longMultiLineField = """
+            This is anotehr very very very very very very very very
```
schalkms: Maybe we should add a test case where the beginning of the multiline string (`"""`) starts in a new line.

cortinico: Yup I'm just pushing a commit for that

PR #1674, opened 2019-06-02
Adding test for UselessCallOnNotNull and platform types

### LINKED ISSUE #1672: Add testcase for UselessCallOnNotNull verifying behaviour of List<T>! type
A testcase for calling `orEmpty` on a list from a java file.

## Context
As stated in #1661 there needs to be a test case for a kotlin file calling a java file for "UselessCallOnNotNull".

cortinico:
Adding a small test to verify that UselessCallOnNotNull is reported
correctly on variables that are using platform types (e.g. `List<Int!>`).

Fixes #1672

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNullSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNullSpec.kt
@@ -58,6 +58,12 @@ object UselessCallOnNotNullSpec : Spek({
             assertThat(subject.compileAndLintWithContext(environment, code)).hasSize(1)
         }
 
+        it("reports when calling orEmpty on a list with a platform type") {
+            // System.getenv().keys will be of type List<String!>.
+            val code = """val testSequence = System.getenv().keys.toList().orEmpty()"""
+            assertThat(subject.compileAndLintWithContext(environment, code)).hasSize(1)
+        }
+
         it("only reports on a Kotlin list") {
             val code = """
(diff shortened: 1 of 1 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNullSpec.kt
```diff
@@ -58,6 +58,12 @@ object UselessCallOnNotNullSpec : Spek({
             assertThat(subject.compileAndLintWithContext(environment, code)).hasSize(1)
         }
 
+        it("reports when calling orEmpty on a list with a platform type") {
+            // System.getenv().keys will be of type List<String!>.
```
schalkms: This actually returns a `MutableSet<String>` without the exclamation mark `<!>`.
The call also does not originate from a `Java` source file.

cortinico: > This actually returns a MutableSet<String>

Yup you're correct. I forgot a `.toList()` inside the comment.
I'm pushing a fix for the comment right now.

> without the exclamation mark <!>.

Are you sure of this? I think it's actually a plaform type since `System.getenv()` (from `java.lang.System`) will return a `MutableMap<String!, String!>` and the platform type will be carried over with the `.keys` and the `.toList()`.

> The call also does not originate from a Java source file.

If I understood the discussion correctly in #1661, the idea was to test a `List<T!>` where T is a platform type (like `String!` in this case), right? I thought that using `java.lang.System` was a fast and clean way to have methods that returns platform types.

PR #1689, opened 2019-06-12
Fix reporting logic for UseRequire and UseCheckOrErrorSpec

### LINKED ISSUE #1683: UseCheckOrError/UseRequire false positives
detekt 1.0.0-RC15

These false positives can be reported by either UseCheckOrError or UseRequire depending on the type of exception (IllegalStateException/IllegalArgumentException).

**Case 1: when**
```kotlin
override fun getColumnCount(item: IndexedItem): Int = when (item) {
    is FolderItem -> folderParams.columnCount
    is ImageItem -> imageParams.columnCount
    else -> throw IllegalArgumentException("Unexpected item ${item.javaClass}")
}
```

**Case 2: no condition around throw**
```kotlin
fun getTagIndex(tag: Tag, tags: List<Tag>): Int {
    var index = 0
    for (t in tags) {
        if (t === tag) {
            return index
        } else if (t.type == tag.type) {
            index++
        }
    }
    throw IllegalArgumentException("Tag $tag was not found in the list") // UseRequire
}
```

**Case 3: elvis operator**
```kotlin
private fun getSuperclassTypeParameter(subclass: Class<*>): Type {
    val superclass = subclass.genericSuperclass as? ParameterizedType
        ?: throw IllegalArgumentException("Super class of $subclass must be parametrized") // UseRequire
    return superclass.actualTypeArguments[0]
}
```

cortinico:
The two rules were too 'greedy' and were reporting the majority
of the Illegal[State|Argument]Exception in the analysed code.

I restricted the scope of the rule to raise a issue only if the
exception is used after a precondition, or in a when-else branch.

Fixes #1683

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/ThrowExtensions.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/ThrowExtensions.kt
@@ -21,11 +21,10 @@ internal fun <T : Exception> KtThrowExpression.isExceptionOfType(clazz: KClass<T
 internal val KtThrowExpression.argumentCount
     get() = findDescendantOfType<KtCallExpression>()?.valueArgumentList?.children?.size ?: 0
 
-internal fun KtThrowExpression.isAfterAPreCondition(): Boolean {
+internal fun KtThrowExpression.isEnclosedByConditionalStatement(): Boolean {
     return parent is KtIfExpression || parent is KtContainerNodeForControlStructureBody
 }
 
 internal fun KtThrowExpression.isInAWhenElseBranch(): Boolean {
-    val parentExpression = parent
-    return parentExpression is KtWhenEntry && parentExpression.isElse
+    return (parent as? KtWhenEntry)?.isElse ?: false
 }
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/ThrowExtensions.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/ThrowExtensions.kt
@@ -4,7 +4,6 @@ import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtContainerNodeForControlStructureBody
 import org.jetbrains.kotlin.psi.KtIfExpression
 import org.jetbrains.kotlin.psi.KtThrowExpression
-import org.jetbrains.kotlin.psi.KtWhenEntry
 import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
 import kotlin.reflect.KClass
 
@@ -23,8 +22,4 @@ internal val KtThrowExpression.argumentCount
 
 internal fun KtThrowExpression.isEnclosedByConditionalStatement(): Boolean {
     return parent is KtIfExpression || parent is KtContainerNodeForControlStructureBody
-}
-
-internal fun KtThrowExpression.isInAWhenElseBranch(): Boolean {
-    return (parent as? KtWhenEntry)?.isElse ?: false
 }
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/ThrowExtensions.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/ThrowExtensions.kt
@@ -1,7 +1,10 @@
 package io.gitlab.arturbosch.detekt.rules
 
 import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtContainerNodeForControlStructureBody
+import org.jetbrains.kotlin.psi.KtIfExpression
 import org.jetbrains.kotlin.psi.KtThrowExpression
+import org.jetbrains.kotlin.psi.KtWhenEntry
 import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
 import kotlin.reflect.KClass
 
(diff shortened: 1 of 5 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/ThrowExtensions.kt
```diff
@@ -17,3 +20,12 @@ internal fun <T : Exception> KtThrowExpression.isExceptionOfType(clazz: KClass<T
 
 internal val KtThrowExpression.argumentCount
     get() = findDescendantOfType<KtCallExpression>()?.valueArgumentList?.children?.size ?: 0
+
+internal fun KtThrowExpression.isAfterAPreCondition(): Boolean {
+    return parent is KtIfExpression || parent is KtContainerNodeForControlStructureBody
+}
+
+internal fun KtThrowExpression.isInAWhenElseBranch(): Boolean {
+    val parentExpression = parent
```
schalkms: I think this assignment can be removed.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/ThrowExtensions.kt
```diff
@@ -17,3 +20,12 @@ internal fun <T : Exception> KtThrowExpression.isExceptionOfType(clazz: KClass<T
 
 internal val KtThrowExpression.argumentCount
     get() = findDescendantOfType<KtCallExpression>()?.valueArgumentList?.children?.size ?: 0
+
+internal fun KtThrowExpression.isAfterAPreCondition(): Boolean {
```
schalkms: The name here is a bit misleading.
With `KtThrowExpression.isAfterAPreCondition` I would assume that `sibling.previous` is meant, although the code says `parent`.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/ThrowExtensions.kt
arturbosch: It does make use of smartcasts, so unfortunately no.
Maybe consider if `(parent as? KtWhenEntry).isElse ?: false` looks nicer

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/ThrowExtensions.kt
arturbosch: Consider `isEnclosedByConditionalStatement`.

3flex:
Based on the tests I don't agree with the change to `UseCheckOrError`. Both usages of `throw IllegalStateException()` in the updated tests in UseCheckOrErrorSpec can be replaced with `error()`, so detekt should report it.

cortinico:
> Both usages of throw IllegalStateException() in the updated tests in UseCheckOrErrorSpec can be replaced with error(), so detekt should report it.

@3flex Which cases are you referring to exactly?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseCheckOrErrorSpec.kt
```diff
@@ -85,14 +85,41 @@ class UseCheckOrErrorSpec : Spek({
...
-        it("reports an issue if the exception thrown as the only action in a function") {
+        it("does not report an issue if the exception thrown unconditionally") {
             val code = """fun doThrow() = throw IllegalStateException("message")"""
-            assertThat(subject.lint(code)).hasSize(1)
+            assertThat(subject.lint(code)).hasSize(0)
         }
 
-        it("reports an issue if the exception thrown as the only action in a function block") {
+        it("does not report an issue if the exception thrown unconditionally in a function block") {
             val code = """fun doThrow() { throw IllegalStateException("message") }"""
```
3flex: This should still report as it can be replaced with `fun doThrow() { error("message") }`

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseCheckOrErrorSpec.kt
```diff
@@ -85,14 +85,41 @@ class UseCheckOrErrorSpec : Spek({
...
+            it("does not report an issue if the exception is after a block") {
+                val code = """
+                    fun doSomethingOrThrow(test: Int): Int {
+                        var index = 0
+                        repeat(test){
+                            if (Math.random() == 1.0) {
+                                return it
+                            }
+                        }
+                        throw IllegalStateException("Test was too big")
```
3flex: This should still report as it can be replaced with `error("Test was too big")`

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseCheckOrErrorSpec.kt
```diff
@@ -85,14 +85,41 @@ class UseCheckOrErrorSpec : Spek({
...
+                    }""".trimIndent()
+                assertThat(subject.lint(code)).isEmpty()
+            }
+
+            it("does not report an issue if the exception is after a elvis operator") {
+                val code = """
+                    fun tryToCastOrThrow(list: List<*>) : LinkedList<*> {
+                        val subclass = list as? LinkedList
+                            ?: throw IllegalStateException("List is not a LinkedList")
```
3flex: This should still report as it can be replaced with `error("List is not a LinkedList")`

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseCheckOrErrorSpec.kt
```diff
@@ -85,14 +85,41 @@ class UseCheckOrErrorSpec : Spek({
             assertThat(subject.lint(code)).isEmpty()
         }
 
-        it("reports an issue if the exception thrown as the only action in a function") {
+        it("does not report an issue if the exception thrown unconditionally") {
             val code = """fun doThrow() = throw IllegalStateException("message")"""
```
3flex: This should still report as it can be replaced with `fun doThrow() = error("message")`

cortinico:
> Added inline comments to explain my thinking from my earlier comment, hope this makes sense?

Yup I see your point 👍 Thanks for explaining. So I've updated the `UseCheckOrError` to don't have checks for preconditions or when-else branches.

This PR should be good to go now

3flex:
Ignore the test coverage changes... The JaCoCo plugin config needs a change in config to work properly with parallel builds and build cache. I'm going to open another PR for that soon.

schalkms:
Yeah codecov is a bit buggy.
That would be nice @3flex

PR #1690, opened 2019-06-12
 Fixed bug reporting false positives with EmptyFunctionBlock

### LINKED ISSUE #1684: EmptyFunctionBlock false positive: overridden function with comment in body
detekt 1.0.0-RC15

This gets reported by EmptyFunctionBlock as violation:
```kotlin
override fun fetch() {
    // No-op
}
```

There is `ignoreOverriddenFunctions` but I don't want to ignore all overridden functions, just ones with a comment in their body.

cortinico:
This rule (EmptyFunctionBlock) had a bug in reporting functions with override and empty
blocks (with or without comments). 

I'm fixing it and adding a couple of tests to make sure it works properly.

Fixes #1684

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlock.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlock.kt
@@ -7,7 +7,10 @@ import org.jetbrains.kotlin.psi.KtNamedFunction
 
 /**
  * Reports empty functions. Empty blocks of code serve no purpose and should be removed.
- * This rule will not report functions overriding others.
+ * This rule will report all the functions with an empty body, with an exception: overriding functions with a
+ * comment in the body (e.g. a `// no-op` comment).
+ *
+ * To ignore all the override functions, please use the [ignoreOverriddenFunctions] configuration field.
  *
  * @configuration ignoreOverriddenFunctions - excludes overridden functions with an empty body (default: `false`)
  *
@@ -28,6 +31,7 @@ class EmptyFunctionBlock(config: Config) : EmptyRule(config) {
         val bodyExpression = function.bodyExpression
         if (!ignoreOverriddenFunctions) {
             if (function.isOverride()) {
+                // If this function is an override, let's ignore empty bodies with comments.
                 bodyExpression?.addFindingIfBlockExprIsEmptyAndNotCommented()
             } else {
                 bodyExpression?.addFindingIfBlockExprIsEmpty()
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlock.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlock.kt
@@ -7,12 +7,13 @@ import org.jetbrains.kotlin.psi.KtNamedFunction
 
 /**
  * Reports empty functions. Empty blocks of code serve no purpose and should be removed.
- * This rule will report all the functions with an empty body, with an exception: overriding functions with a
- * comment in the body (e.g. a `// no-op` comment).
+ * This rule will not report functions with the override modifier that have a comment as their only body contents
+ * (e.g., a // no-op comment in an unused listener function).
  *
- * To ignore all the override functions, please use the [ignoreOverriddenFunctions] configuration field.
+ * Set the [ignoreOverriddenFunctions] parameter to `true` to exclude all functions which are overriding other
+ * functions from the superclass or from an interface (i.e., functions declared with the override modifier).
  *
- * @configuration ignoreOverriddenFunctions - excludes overridden functions with an empty body (default: `false`)
+ * @configuration ignoreOverriddenFunctions - Excludes all the overridden functions (default: `false`)
  *
  * @active since v1.0.0
  * @author Artur Bosch
@@ -31,7 +32,6 @@ class EmptyFunctionBlock(config: Config) : EmptyRule(config) {
         val bodyExpression = function.bodyExpression
         if (!ignoreOverriddenFunctions) {
             if (function.isOverride()) {
-                // If this function is an override, let's ignore empty bodies with comments.
                 bodyExpression?.addFindingIfBlockExprIsEmptyAndNotCommented()
             } else {
                 bodyExpression?.addFindingIfBlockExprIsEmpty()
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlockSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlockSpec.kt
@@ -1,9 +1,10 @@
 package io.gitlab.arturbosch.detekt.rules.empty
 
 import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.test.TEST_FILENAME
 import io.gitlab.arturbosch.detekt.test.TestConfig
+import io.gitlab.arturbosch.detekt.test.assertThat
 import io.gitlab.arturbosch.detekt.test.compileAndLint
-import org.assertj.core.api.Assertions.assertThat
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
 
@@ -12,6 +13,8 @@ import org.spekframework.spek2.style.specification.describe
  * @author schalkms
  */
 class EmptyFunctionBlockSpec : Spek({
+
+    val fileName = TEST_FILENAME
 
     val subject by memoized { EmptyFunctionBlock(Config.empty) }
 
@@ -22,7 +25,7 @@ class EmptyFunctionBlockSpec : Spek({
                 class A {
                     protected fun stuff() {}
                 }"""
-            assertThat(subject.compileAndLint(code)).hasSize(1)
+            assertThat(subject.compileAndLint(code)).hasLocationStrings("'{}' at (2,27) in /$fileName")
         }
 
         it("should not flag function with open modifier") {
@@ -38,7 +41,7 @@ class EmptyFunctionBlockSpec : Spek({
 				fun a() {
 					fun b() {}
 				}"""
-            assertThat(subject.compileAndLint(code)).hasSize(1)
+            assertThat(subject.compileAndLint(code)).hasLocationStrings("'{}' at (2,10) in /$fileName")
         }
 
         context("some overridden functions") {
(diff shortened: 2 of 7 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlockSpec.kt
```diff
@@ -75,5 +75,33 @@ class EmptyFunctionBlockSpec : Spek({
...
+                        // no-op
+                    }
+
+                    override fun listenThat() {
+
+                    }
+                }
+            """.trimIndent()
+            it("should not flag overridden functions with commented body") {
+                assertThat(subject.compileAndLint(code)).hasSize(1)
```
rock3r: I'd avoid using `hasSize()` to assert things — we've been bitten by it in the past. I know that most tests still use it but that gives no guarantees (except in the case of `isEmpty()`/`hasSize(0)`), because code changes may remove the correct detection and replace it with a wrong one. In this example, it may switch to flag `listenThat()` and we'd be none the wiser.

Please use `FindingsAssertions`' own `hasLocationStrings()` and match on the location(s) detected instead.

detekt-rules/src/test/resources/cases/Empty.kt
```diff
@@ -19,10 +19,6 @@ class Empty : Runnable {
 
     }
 
-    fun emptyMethod() {
```
rock3r: Why was this removed?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlock.kt
```diff
@@ -7,9 +7,12 @@ import org.jetbrains.kotlin.psi.KtNamedFunction
 
 /**
  * Reports empty functions. Empty blocks of code serve no purpose and should be removed.
- * This rule will not report functions overriding others.
+ * This rule will report all the functions with an empty body, with an exception: overriding functions with a
```
rock3r: Maybe terser:

>This rule will not report functions with the `override` modifier that have a comment as their only body contents (e.g., a `// no-op` comment in an unused listener function)

docs/pages/documentation/empty-blocks.md
```diff
@@ -74,7 +74,10 @@ Reports empty `for` loops. Empty blocks of code serve no purpose and should be r
 ### EmptyFunctionBlock
 
 Reports empty functions. Empty blocks of code serve no purpose and should be removed.
-This rule will not report functions overriding others.
+This rule will report all the functions with an empty body, with an exception: overriding functions with a
```
rock3r: This is autogen'd from the kt class so it will change accordingly :)

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlock.kt
```diff
@@ -7,9 +7,12 @@ import org.jetbrains.kotlin.psi.KtNamedFunction
...
+ * comment in the body (e.g. a `// no-op` comment).
  *
- * @configuration ignoreOverriddenFunctions - excludes overridden functions with an empty body (default: `false`)
+ * To ignore all the override functions, please use the [ignoreOverriddenFunctions] configuration field.
+ *
+ * @configuration ignoreOverriddenFunctions - Excludes all the overridden functions (default: `false`).
```
rock3r: Don't put the dot at the end — the string is manipulated when generating the markdown and it will result in a weird entry in the doc with a superfluous space — see [further down](https://github.com/arturbosch/detekt/pull/1690/files#diff-81f6dd5347eb6b688ca081c05000c30bR90).

rock3r:
Also the CI says you have conflicts :P

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlock.kt
```diff
@@ -28,6 +31,7 @@ class EmptyFunctionBlock(config: Config) : EmptyRule(config) {
         val bodyExpression = function.bodyExpression
         if (!ignoreOverriddenFunctions) {
             if (function.isOverride()) {
+                // If this function is an override, let's ignore empty bodies with comments.
```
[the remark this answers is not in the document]
cortinico: Sort of, although the logic here is a bit intricate imho. Anyway I'm removing it, I don't have a strong opinion on this.

detekt-rules/src/test/resources/cases/Empty.kt
cortinico: Because the test:
```kotlin
it("findsEmptyFun") {
            test { EmptyFunctionBlock(Config.empty) }
}

[...]
private fun test(block: () -> Rule) {
    val rule = block()
    rule.lint(compileForTest(Case.Empty.path()))
    assertThat(rule.findings).hasSize(1)
}
```
inside EmptyCodeSpek.kt:92 was failing. The failure was triggered by having two empty functions inside the `Empty.kt` file (`override fun run()` and `fun emptyMethod()`). So either we update the `test(block: () -> Rule)` to have the number of expected issues as parameter (with maybe `1` as default value) or we just cleanup the `Empty.kt` file. I went for the latter solution.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlock.kt
cortinico: Oh damn 😅

cortinico:
> Also the CI says you have conflicts :P

Yeah I should update my master from time to time 😆

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlockSpec.kt
cortinico: Agree 👍 I've added `0560f0a15f2ad2f03eebe25f076807aa93835fe7` that contains this fix.
Although I have the strong feeling that the `hasLocationStrings()` is not properly working. Apparently failures are not raised to Spek, and tests are marked as green also if the assertion is not met. I'm going to investigate a bit further and open another PR.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlockSpec.kt
```diff
@@ -22,7 +25,7 @@ class EmptyFunctionBlockSpec : Spek({
                 class A {
                     protected fun stuff() {}
                 }"""
-            assertThat(subject.compileAndLint(code)).hasSize(1)
+            assertThat(subject.compileAndLint(code)).hasLocationStrings("'{}' at (2,27) in /$fileName")
```
schalkms: I would write it like this to also check that there is exactly 1 finding.

```Kotlin
val findings = subject.compileAndLint(code)
assertThat(findings).hasLocationStrings("...")
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlockSpec.kt
```diff
@@ -38,7 +41,7 @@ class EmptyFunctionBlockSpec : Spek({
 				fun a() {
 					fun b() {}
 				}"""
-            assertThat(subject.compileAndLint(code)).hasSize(1)
+            assertThat(subject.compileAndLint(code)).hasLocationStrings("'{}' at (2,10) in /$fileName")
```
schalkms: same here

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlockSpec.kt
```diff
@@ -72,7 +75,36 @@ class EmptyFunctionBlockSpec : Spek({
 
             it("should not flag overridden functions") {
                 val config = TestConfig(mapOf(EmptyFunctionBlock.IGNORE_OVERRIDDEN_FUNCTIONS to "true"))
-                assertThat(EmptyFunctionBlock(config).compileAndLint(code)).hasSize(1)
+                assertThat(EmptyFunctionBlock(config).compileAndLint(code)).hasLocationStrings("'{}' at (1,13) in /$fileName")
```
schalkms: same here

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlockSpec.kt
```diff
@@ -72,7 +75,36 @@ class EmptyFunctionBlockSpec : Spek({
...
+                        // no-op
+                    }
+
+                    override fun listenThat() {
+
+                    }
+                }
+            """.trimIndent()
+            it("should not flag overridden functions with commented body") {
+                assertThat(subject.compileAndLint(code))
```
schalkms: same here

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlockSpec.kt
rock3r: Hm, not sure if it's broken since I last used it tbh, worth checking!

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlockSpec.kt
rock3r: I seem to remember `hasLocationStrings()` should be checking that too? If not, probably worth creating a `hasExactlyLocationStrings()` that checks the count too, seems useful.

schalkms: `hasLocationStrings()` does not check this.
I would also prefer your solution with `hasExactlyLocationStrings()`.
Thanks for the comment @rock3r

cortinico: Agree 👍 Pushed 0ac2a8dd2df6674daabee492104a7bed4d58516b with the updated assertion logic.

cortinico:
This should be good to go now imho

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlockSpec.kt
cortinico: @rock3r Yup was broken #1706 :/

rock3r: Dang! I hope I'll have some time to fix it soon™

Issue #1694, opened 2019-06-14
Strange symbols at the beginning of each console message 


remal:
I'm using Detekt with Gradle and getting console output like this:
```
[33m[0mRuleset: complexity
[33m[0mRuleset: empty-blocks
[33m[0mRuleset: exceptions
[33m[0mRuleset: naming
```

Please remove these strange prefixes. If it's for colored output, please make "if the console supports colors" detection smarter, like it's done in Gradle, for example.

I'm running Gradle from IDEA.

## Your Environment
* Version of detekt used: Gradle plugin 1.0.0-RC15
* Version of Gradle used (if applicable): 5.4.1
* Operation system: Windows 10

marschwar:
Thank you for reporting this.

schalkms:
Thanks for reporting this bug. 
Please answer the following questions to help us reproduce the problem.
Which default system shell are you using on Windows?
Which shell does IntelliJ's terminal emulator emulate?

remal:
> Which default system shell are you using on Windows?
Basically I don't use system shell. But if I have to, I use ёсьвёю I'm not sure if I answered correctly...

> Which shell does IntelliJ's terminal emulator emulate?
I don't know.

---

As I can see in Gradle sources, Gradle uses [Ansi](https://github.com/fusesource/jansi) library to colorize the output. As I understand, the library checks if current OS is Windows and if it's so, it does some magic.

If I'm right, I suggest choosing between these options:
* Omit colors on Windows systems (you can check if `os.name` system property contains `win`)
* Use Ansi library to colorize output. Personally I don't like this solutions, as it requires to add one more dependency. But you can shade this dependency :)

remal:
@marschwar @schalkms I've submitted 2 PRs. Suggest we merge #1702 and close the other.

schalkms:
@remal big thanks 🥇

PR #1701, opened 2019-06-16
Disable colors on Windows OS

remal:
This is simple and a bit dirty fix for #1694 - remove color chars in console output for Windows OS.

IMHO this PR is **less** preferable than #1702

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/Colorizer.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/Colorizer.kt
@@ -12,7 +12,12 @@ private data class Color(private val value: Byte) {
         get() = "$ESC[${value}m"
 }
 
-private fun CharSequence.colorized(color: Color) = "${color.escapeSequence}$this${RESET.escapeSequence}"
+private val isColoredOutputSupported: Boolean = !System.getProperty("os.name", "").contains("win", true)
+private fun CharSequence.colorized(color: Color): String = if (isColoredOutputSupported) {
+    "${color.escapeSequence}$this${RESET.escapeSequence}"
+} else {
+    this.toString()
+}
 
 fun CharSequence.red() = colorized(RED)
 fun CharSequence.yellow() = colorized(YELLOW)
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/Colorizer.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/Colorizer.kt
@@ -12,12 +12,15 @@ private data class Color(private val value: Byte) {
         get() = "$ESC[${value}m"
 }
 
-private val isColoredOutputSupported: Boolean = !System.getProperty("os.name", "").contains("win", true)
+
+private val isColoredOutputSupported: Boolean = !System.getProperty("os.name", "").startsWith("Windows")
+
 private fun CharSequence.colorized(color: Color): CharSequence = if (isColoredOutputSupported) {
     "${color.escapeSequence}$this${RESET.escapeSequence}"
 } else {
     this
 }
+
 
 fun CharSequence.red() = colorized(RED)
 fun CharSequence.yellow() = colorized(YELLOW)
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/Colorizer.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/Colorizer.kt
@@ -13,16 +13,12 @@ private data class Color(private val value: Byte) {
 }
 
 private val isColoredOutputSupported: Boolean = !System.getProperty("os.name", "").startsWith("Windows")
-private fun CharSequence.colorized(color: Color): CharSequence = if (isColoredOutputSupported) {
+private fun String.colorized(color: Color) = if (isColoredOutputSupported) {
     "${color.escapeSequence}$this${RESET.escapeSequence}"
 } else {
     this
 }
 
-fun CharSequence.red(): CharSequence = colorized(RED)
-fun CharSequence.yellow(): CharSequence = colorized(YELLOW)
-fun CharSequence.decolorized(): CharSequence = this.replace(escapeSequenceRegex, "")
-
-fun String.red(): String = colorized(RED).toString()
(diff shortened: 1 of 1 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/Colorizer.kt
```diff
@@ -12,7 +12,12 @@ private data class Color(private val value: Byte) {
         get() = "$ESC[${value}m"
 }
 
-private fun CharSequence.colorized(color: Color) = "${color.escapeSequence}$this${RESET.escapeSequence}"
+private val isColoredOutputSupported: Boolean = !System.getProperty("os.name", "").contains("win", true)
```
schalkms: What happens if the OS is Dar**win**? Then this variable is true.
The check should be written with `startsWith` to be safer.
```Kotlin
!System.getProperty("os.name", "").startsWith("Windows")
```

remal:
@schalkms The Idea behind checking if the property contains `win` substring is that it's better to turn off colorizing than add colors on shell without colors support.

Actually I don't see any major drawbacks with adding Jansi dependency to `detekt-cli` module. I suppose those who don't need this library, probably, won't use `detekt-cli`.

BTW as I understand all checks you mentioned are supported by Jansi.

schalkms:
> The Idea behind checking if the property contains win substring is that it's better to turn off colorizing than add colors on a shell without colors support.

I fully agree. However, `.contains("win")` could match operating systems, which are not Windows. Thus, we need to change it to `.startsWith("Windows")` to be safer here.
[Apache System Utils](http://commons.apache.org/proper/commons-lang/javadocs/api-release/src-html/org/apache/commons/lang3/SystemUtils.html#line.1355) also does it this way.

> Actually I don't see any major drawbacks with adding Jansi dependency to `detekt-cli` module

`Jansi` uses JNI on Windows to support coloring terminal output. 
I'd say for detekt's simple usage it's not necessary to import this library.

> Also, as I understand, all checks you mentioned are supported by Jansi.

I tried it out on Windows with the following shells.

* Cygwin - supports it natively without any hacks
* [Cmder](https://cmder.net) supports cmd and unix shells - works out-of-the-box
* Cmd + Powershell - does not support coloring the terminal output without any hacks (as expected).

> BTW IDEA correctly displays colorized output by Jansi.

True, because IntelliJ only emulates the default cmd on Windows, if not configured otherwise.

marschwar:
I agree with @schalkms. I don't really see the need for adding an additional library just for this. It's far from being a core feature of detekt. 

I could live with `contains`  or `startsWith`, but in my opinion we should check for the string `windows` as it seems to be common practice.

See:
https://github.com/sonatype/plexus-interpolation/blob/16f8c521c60021fbac3295f0ed10ba5bc6cfdc9a/src/main/java/org/codehaus/plexus/interpolation/os/Os.java#L308
https://github.com/apache/commons-lang/blob/dc2953e5e41f0d5bf069349bbd7c6fe26f3362e1/src/main/java/org/apache/commons/lang3/SystemUtils.java#L1355

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/Colorizer.kt
```diff
@@ -12,7 +12,12 @@ private data class Color(private val value: Byte) {
...
+private fun CharSequence.colorized(color: Color): String = if (isColoredOutputSupported) {
+    "${color.escapeSequence}$this${RESET.escapeSequence}"
+} else {
+    this.toString()
+}
```
marschwar: Minor nitpick: In my opinion ` CharSequence.colorized(color: Color)` should itself return a CharSequence if possible.

arturbosch:
Thanks @remal @schalkms @marschwar for further investigations. So lets go with this pr then.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/Colorizer.kt
remal: fixed

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/Colorizer.kt
remal: fixed

remal:
Guys, please take a look.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/Colorizer.kt
```diff
@@ -12,9 +12,17 @@ private data class Color(private val value: Byte) {
...
-fun CharSequence.red() = colorized(RED)
-fun CharSequence.yellow() = colorized(YELLOW)
+fun CharSequence.red(): CharSequence = colorized(RED)
+fun CharSequence.yellow(): CharSequence = colorized(YELLOW)
+fun CharSequence.decolorized(): CharSequence = this.replace(escapeSequenceRegex, "")
 
-fun CharSequence.decolorized() = this.replace(escapeSequenceRegex, "")
+fun String.red(): String = colorized(RED).toString()
+fun String.yellow(): String = colorized(YELLOW).toString()
+fun String.decolorized(): String = this.replace(escapeSequenceRegex, "")
```
schalkms: Why do we have the same extension method **String** and **CharSequence**? Did I miss something?

remal: Because @marschwar asked to make `CharSequence.colorized(color: Color)` to return `CharSequence`. But there are some places where `String` is expected, so it would be more convenient to have a method that returns `String`. I suggest leaving only one variant that returns `String`. WDYT?

marschwar: Sorry to cause such a confusion. I had not actually made the change myself so I did not realize there were usages where a String is expected to be returned. My point was that the extension function should return the same type if possible. 
Please keep `fun String.colorized(color: Color): String` only. AFAIK we don't need the extension function to be on CharSequence.

remal: fixed

PR #1703, opened 2019-06-16
Update doc (report extension section)

schalkms:


docs/pages/extensions.md
```diff
@@ -166,7 +166,7 @@ class NumberOfLoopsProcessorSpec : Spek({
 _detekt_ allows you to extend the console output and to create custom output formats.
 If you want to customize the output, take a look at the `ConsoleReport` and `OutputReport` classes.
 
-All they need are implementation of the `render()`-function which takes a object with all findings and returns a string.
+All they need are an implementation of the `render()`-function which takes a object with all findings and returns a string to be printed out.
```
3flex:
```suggestion
All they need are an implementation of the `render()`-function which takes an object with all findings and returns a string to be printed out.
```

schalkms: Thanks 👍

Issue #1695, opened 2019-06-14
Please make console output more verbose


remal:
I'm using Detekt with Gradle. Detekt Gradle plugin generates beautiful HTML reports, but I don't use them. I'm looking at console output instead. Please make console output more verbose.

For example, I'm getting this message:
```
ReturnCount - [...] at ...
```

Message like this would be much more useful:
```
ReturnCount - [...] at ...
    Restrict the number of return statements in methods.
    Function ... has ... return statements which exceeds the limit of 2.
```

schalkms:
Thanks for your feedback!
I think we decided to not clutter the console output with too much information.
For this purpose we have several report formats.
Therefore, I'm against making this change.

remal:
I'm not using report files at all, as I'm using GitLab CI. So I need to print as much useful information to then console as possible. Maybe it makes sense to do it configurable...

schalkms:
But you can still persist files (in this case detekt's report files) even when using GitLab CI.

remal:
I can persist. Sure. But having verbose input whould be more convinient.

schalkms:
We will see what other users think about your suggestion.
Therefore, I put the according label on this issue.

remal:
BTW is there a programming API that will allows me to build my own Gradle plugin with verbose output? Could you please suggest a place where to read about it? If the API exists, how stable is it?

schalkms:
@remal that's certainly possible with detekt. If you want to go that way, you can do that with custom reports.
An example is located [here](https://arturbosch.github.io/detekt/extensions.html#custom-reports).
However, it's not a Gradle plugin but rather a way of extending detekt's report functionality.
If you need help just ask here or in detekt's Slack channel.

arturbosch:
@remal we would highly appreciate a more verbose `ConsoleReport` (https://github.com/arturbosch/detekt/blob/ed23c35505f382390ff1c57c855a33763723125e/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/ConsoleReport.kt#L14-L23)

Ours is here: https://github.com/arturbosch/detekt/blob/ed23c35505f382390ff1c57c855a33763723125e/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FindingsReport.kt#L10

[...]

PR #1704, opened 2019-06-17
ReportConfig: showProgress and showMessages

remal:
This PR is an implementation of #1695.

Additional config options have been added:
```yaml
console-reports:
  showProgress: false # true - show rulesets without issues, false - hide. Default value: true
  showMessages: true # true - show issue message/description, false - hide. Default value: false

output-reports:
  showProgress: false # true - show "Successfully generated ..." messages, false - hide. Default value: true
```


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FindingsReport.kt
```diff
@@ -11,32 +15,55 @@ class FindingsReport : ConsoleReport() {
 
     override val priority: Int = 40
 
-    override fun render(detektion: Detektion): String? {
-        val findings = detektion.findings
+    private lateinit var config: ConsoleReportConfig
```
arturbosch: Please use `by SingleAssign()` here which wraps lateinit and adds an additional check

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ReportConfig.kt
```diff
@@ -0,0 +1,36 @@
...
+        private const val CONSOLE_REPORTS = "console-reports"
+        private const val OUTPUT_REPORTS = "output-reports"
+    }
+}
+
+abstract class BaseReportsConfig(protected val delegate: Config) : Config by delegate {
+    val active: Boolean get() = valueOrDefault("active", true)
+    val includes: Set<String> get() = valueOrDefault("include", emptyList<String>()).toSet()
+    val excludes: Set<String> get() = valueOrDefault("exclude", emptyList<String>()).toSet()
+    val showProgress: Boolean get() = valueOrDefault("showProgress", true)
```
arturbosch: The new config property must get generated by `detekt-generator` inside `detekt/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/ConfigPrinter.kt` or one of the other classes.

arturbosch:
@schalkms you are right, it's also the behavior we discussed in the corresponding issue.
Having a `VerboseFindingsReport` should extend detekt's behavior without modifying current classes.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FindingsReport.kt
remal: fixed

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ReportConfig.kt
remal: fixed

remal:
I've fixed the problems you mentioned.

However I don't think that creating `VerboseFindingsReport` is a right choice - for me it's not clear how to make it executed instead of `FindingsReport`. `ReportLocator` is not so smart...

remal:
@schalkms I didn't get your point.... By default all reportes are active, so... Do you suggest to make `VerboseFindingsReport` not active by default and then implement a way to activate it **instead** of `FindingsReport`?

schalkms:
@remal we should have been clearer. I hope the following explanation helps to go into the right direction with this PR. If not, please report back to us.

> Do you suggest to make `VerboseFindingsReport` not active by default

Exactly. The default config then looks like this. The `VerboseFindingsReport` is excluded by default.
Users, who want a more verbose output in the terminal, can choose to exclude the *normal* `FindingsReport` and include the `VerboseFindingsReport`.

```Yaml
console-reports: 
  active: true 
  exclude:
    - 'VerboseFindingsReport'
    # - 'ProjectStatisticsReport'
    # ...
```

> and then implement a way to activate it **instead** of `FindingsReport`?

You don't need to implement anything for that except your `VerboseFindingsReport` and changing the config file as mentioned above.
That's why you don't need to touch anything in the existing classes in the cli package.
This helps us to reduce complexity. Our extension developers have the freedom to implement custom reports without touching the existing cli logic or creating a fork of detekt.

remal:
@schalkms I got your point, but I've implemented not only displaying issue message + description.

I've added a possibility to turn off displaying "Ruleset: ..." if there are no issues found in this ruleset.

Also I've added a possibility to turn off displaying "Successfully generated ... at ..." message for each out report.

My goal is to make detekt displaying only required information, but if it displays smth, I want to to have full info. What rulesets were executep and what reports were generated, IMHO, is a debug info.

schalkms:
> I got your point, but I've implemented not only displaying issue message + description.

I see. I really like this change, but we would rather have it in an own report class as mentioned in my previous post.

> I've added a possibility to turn off displaying "Ruleset: ..." if there are no issues found in this ruleset.

This information should stay, but it should say `Ruleset: style - no issues found`. I don't see any benefit of introducing a config option, which adds to the complexity of the project.

> Also I've added a possibility to turn off displaying "Successfully generated ... at ..." message for each out report.

Why would you want to remove this or even have a config option for this?
It only displays 1 line per generated report in your terminal, which is acceptable. 
Furthermore, this information is essential. It allows the user to click on the link in the terminal and open the report in the respective application. The cons outweigh the pros with this approach.

> My goal is to make detekt displaying only required information, but if it displays smth, I want to to have full info.

Both (ruleset info and report generation) should provide notifications. That's fine.

The `FindingsReport` could be improved with more verbosity. We all agreed on that. 

> What rulesets were executep and what reports were generated, IMHO, is a debug info.

I highly disagree with that. How should the user know whether all rulesets were run? We run detekt's rules on detekt's codebase and it's really good to know that everything went well when running this on the local developer machine.
When the changes create technical debt, the user wants to quickly navigate to the report with the help of the link to see where in the codebase this issue happened and why.

remal:
IMHO every message during build process should mean something. Very long build log (especially on local developer machine) is hard to read, so we have to make it as short, as possible.

Personally for me "Successfully generated ... at ..." message is meaningless. I'd like to have an option to hide it.

The same thing with "Ruleset: style - no issues found". If everything is OK, I don't need it.

Actually I don't know any static analysis tool that displays anything in the terminal if no issues were found.

arturbosch:
So lets create two reports, one more-verbose than our default one and one which can be called something like "CIReport" which has the features @remal wants?
Both are excluded by default but can be used interchangeable when needed?

schalkms:
First of all, I want to say thanks for having this discussion.

> IMHO every message during build process should mean something. Very long build log (especially on local developer machine) is hard to read, so we have to make it as short, as possible

The issue stated that you want to have a more verbose report. This issue results in an even shorter build log.
Detekt's build log is very short and concise if you compare it with other tools in the Unix world.

> Personally for me "Successfully generated ... at ..." message is meaningless. I'd like to have an option to hide it.
> 
> The same thing with "Ruleset: style - no issues found". If everything is OK, I don't need it.

That shouldn't be changed since it allows to have clickable reports in the console, which was a feature request for detekt.
You can still deactivate all reports with the following commands and have your custom **domain-specific** console report launched.

```Yaml
exclude:
    - 'ProjectStatisticsReport'
    - 'ComplexityReport'
    - 'NotificationReport'
    - 'FindingsReport'
    - 'BuildFailureReport'
    # - 'MyCustomReport' 
```

> Actually I don't know any static analysis tool that displays anything in the terminal if no issues were found.

That's a good point. By default it's ok to not display this when no issues were found.
I don't even think we need a config option.

schalkms:
Since I would hate to reject this PR, I want to summarize what this feature request should include.

1. one `VerboseFindingsReport` which displays the following information. This should be disabled by default and can be enabled in the `config.yaml` file.
```
ReturnCount - [...] at ...
    Restrict the number of return statements in methods.
    Function ... has ... return statements which exceeds the limit of 2.
```

2. conditionals around the log statements, which check the issue count. If no issues were found, the mentioned logs shouldn't be displayed.

-----------

This way we don't need any additional config options.
We adhere to detekt's philosophy of providing sensible defaults.
Hence, we don't add to the complexity of this project.

Is that acceptable for you two? @arturbosch @remal

remal:
The problem with "Successfully generated ... at ..." message is that it's not a part of any reporter. This message is displayed by `OutputFacade` class (outside of reporters). So there is no way to hide it just by implementing another report...

@schalkms I'm absolutely OK with the summary. However, I have some concerns about exclusion/inclusion configuration. For example if someone has already created his own config, by start using new version of detekt, this person will have both `VerboseFindingsReport` and`FindingsReport` reports enabled. And it would be absolutely not clear why some messages started displaying twice.

I suggest implementing some check that will warn user if both reports are enabled. WDYT?

arturbosch:
> @schalkms I'm absolutely OK with the summary. However, I have some concerns about exclusion/inclusion configuration. For example if someone has already created his own config, by start using new version of detekt, this person will have both `VerboseFindingsReport` and`FindingsReport` reports enabled. And it would be absolutely not clear why some messages started displaying twice.
> 
> I suggest implementing some check that will warn user if both reports are enabled. WDYT?

I like how you changed the reports config to have an include and an exclude. This would solve this, won't it? A used must opt in for reports.

remal:
@arturbosch But it doesn't help users to understand why they see very similar messages if both reporters are enabled.

arturbosch:
@remal if we a `DefaultFindingsReport` and a `VerboseFindingsReport`. One is included one is excluded by default, I think the user will understand that both have the same functionality but with different logging levels? Or do I miss something?

schalkms:
@arturbosch I think this is a valid assumption.
@remal I hope we didn't confuse you too much? Do you need further help?

remal:
@arturbosch @schalkms  I've created `VerboseFindingsReport`. Please take a look. Personally I think it looks more like a hack, but maybe you would like this direction...

PS: Sorry for the delay, I had a lot of work to do.

arturbosch:
@remal hey, look good to me. Also extensible for the future with the config option :).
Maybe we should consider renaming `showProgress` to something else? `printRuleSetsWithNoIssues` kinda verbose though ..., when reading progress I think about the `DetektProgressListener`

schalkms:
Please resolve the merge conflicts. After that, I'll take a look at this.
Btw, thanks for your efforts!

remal:
@arturbosch Currently `showProgress` has two purposes:
1. Hide rulesets without findings
1. Hide `Successfully generated ${report.name} at $filePath` message

I absolutely agree that something like `DetektProgressListener` is a good idea.


@schalkms I've fixed merge conflicts

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/ExtensionIsNotCompatibleWith.kt
```diff
@@ -0,0 +1,15 @@
...
+import kotlin.annotation.AnnotationTarget.CLASS
+import kotlin.reflect.KClass
+
+@Target(CLASS)
+@Inherited
+annotation class ExtensionIsNotCompatibleWith(vararg val value: KClass<out Extension>)
+
+val Class<out Extension>.notCompatibleClasses: List<Class<out Extension>>
+    get() = getAnnotation(ExtensionIsNotCompatibleWith::class.java)
+        ?.value
```
schalkms: I like this implementation.

detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -8,6 +8,7 @@ build:
 
 processors:
   active: true
+  showProgress: true
```
schalkms: I'm still heavily against this. I mean this just prints out 1 additional line per config.
I think it's not worth to add this.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/OutputFacade.kt
```diff
@@ -32,30 +34,51 @@ class OutputFacade(
...
+            }
         }
     }
 
     private fun handleConsoleReport(report: ConsoleReport, result: Detektion) {
         report.print(settings.outPrinter, result)
     }
+
+    private fun checkReportsCompatibility(reports: List<Extension>) {
+        reports.forEach { report ->
```
schalkms: This will for sure add confusion when releasing a new version of detekt.
By default every user gets this error message when not checking the change log carefully.

schalkms: Maybe a info message is enough.
Perhaps we need to find a different approach here.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ReportLocator.kt
```diff
@@ -10,47 +10,52 @@ import java.util.ServiceLoader
...
 
-    companion object {
-        private const val ACTIVE = "active"
-        private const val EXCLUDE = "exclude"
-    }
+    private fun loadOutputReports(detektLoader: URLClassLoader) =
+        if (outputActive) {
+            ServiceLoader.load(OutputReport::class.java, detektLoader)
+                .filter { outputIncludes.isEmpty() || it.id in outputIncludes }
+                .filter { it.id !in outputExcludes }
```
schalkms: Why did the filter change here?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ReportLocator.kt
```diff
@@ -10,47 +10,52 @@ import java.util.ServiceLoader
...
+    private val outputActive = outputSubConfig.active
+    private val outputIncludes = outputSubConfig.includes
+    private val outputExcludes = outputSubConfig.excludes
 
     fun load(): List<Extension> {
-        settings.debug { "console-report=$consoleActive" }
-        settings.debug { "output-report=$outputActive" }
         val detektLoader = URLClassLoader(settings.pluginUrls, javaClass.classLoader)
+
+        settings.debug { "console-report=$consoleActive" }
```
schalkms: Why did you remove the debug messages?

remal:
@arturbosch @schalkms Guys, I strongly believe that static analysis tools must be silent. Like most of them are. And like most of Linux command-line tools are. Unless a user wants some additional/debug info be displayed.

Actually I've already created my own Gradle plugin that doesn't display Detekt messages at all and add `OutputReporter` that produces a report in FindBugs's XML format (Checkstyle XML reports don't have descriptions). Also I parse this report and display issues on the console if there are any.

I'm absolutely OK if you don't agree with this points. So I suggest closing this PR without merging. `VerboseFindingsReport` doesn't solve my goals. And I don't think that it would help someone.

If you decide to make Detekt silent by default, I'll help you. But it requires some refactoring that (probably) will break backward compatibility.

schalkms:
First of all, thank you for your contribution efforts and answering our questions! We are sorry, that detekt's extension architecture doesn't completely solve all your problems.
In the end, I think you have chosen the right approach here.
Using an existing report format and having some parsing logic, which adapts the issue information to your domain-specific needs, is a nice approach.
@arturbosch I'd also favor to close this. I think we could reopen and merge this PR, if there are many more users, who have the same issue.

PR #1705, opened 2019-06-18
Rule improvement: False positive UnusedImport for componentN

DownMoney:
When using Scala/Java classes it's useful to implement `componentN` operator functions to allow [destructuring](https://kotlinlang.org/docs/reference/multi-declarations.html) an object in to individual variables.

This change adds a regex matching `component\d+` to avoid flagging the imports as unused, similar to how other operator functions are handled.


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
```diff
@@ -78,6 +79,7 @@ class UnusedImports(config: Config) : Rule(config) {
                     .filter { it.identifier()?.contains("*")?.not() == true }
                     .filter { it.identifier() != null }
                     .filter { !operatorSet.contains(it.identifier()) }
+                    .filter { !componentNRegex.matches(it.identifier()!!) }
```
schalkms: is `!!` necessary here?

arturbosch: Yep, matches is a kotlin method with a non nullable paramter ^^

PR #1707, opened 2019-06-19
Fix broken hasLocationStrings and similar broken assert

cortinico:
This PR fixes the following broken asserts:
`hasLocationStrings`
`hasExactlyLocationStrings`
`hasSourceLocations`

Now those asserts will raise a failure if the Location strings are not matched.
Moreover I changed the asserts to use `.apply` so they can be easily chained with other assertions.

### DIFF
--- a/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt
+++ b/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt
@@ -23,22 +23,26 @@ class FindingsAssert(actual: List<Finding>) :
     override fun toAssert(value: Finding?, description: String?): FindingAssert =
             FindingAssert(value).`as`(description)
 
-    fun hasLocationStrings(vararg expected: String, trimIndent: Boolean = false) {
+    fun hasLocationStrings(vararg expected: String, trimIndent: Boolean = false) = apply {
         isNotNull
         val locationStrings = actual.asSequence().map { it.locationAsString }.sorted()
-        if (trimIndent) {
-            areEqual(locationStrings.map { it.trimIndent() }.toList(), expected.map { it.trimIndent() }.sorted())
+        val (actualLocationsList, expectedLocationsList) = if (trimIndent) {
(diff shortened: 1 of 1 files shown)


detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt
```diff
@@ -23,22 +23,26 @@ class FindingsAssert(actual: List<Finding>) :
...
             FindingAssert(value).`as`(description)
 
-    fun hasLocationStrings(vararg expected: String, trimIndent: Boolean = false) {
+    fun hasLocationStrings(vararg expected: String, trimIndent: Boolean = false) = apply {
         isNotNull
         val locationStrings = actual.asSequence().map { it.locationAsString }.sorted()
-        if (trimIndent) {
-            areEqual(locationStrings.map { it.trimIndent() }.toList(), expected.map { it.trimIndent() }.sorted())
+        val (actualLocationsList, expectedLocationsList) = if (trimIndent) {
+            locationStrings.map { it.trimIndent() }.toList() to expected.map { it.trimIndent() }.sorted()
```
schalkms: Is `.toList()` necessary?

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt
```diff
@@ -23,22 +23,26 @@ class FindingsAssert(actual: List<Finding>) :
...
         } else {
-            areEqual(locationStrings.toList(), expected.toList().sorted())
+            locationStrings.toList() to expected.toList().sorted()
```
schalkms: same here

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt
arturbosch: Yep, in line 28 we make it to a sequence :).

PR #1709, opened 2019-06-19
Introduce lock-bot

schalkms:
https://github.com/apps/lock

PR #1716, opened 2019-06-20
Prevent Lockbot from commenting when locking issues

cortinico:
Lockbot is closing/blocking a lot of issues and is flooding the inboxes
of watchers of the repo. Let's turn off the commenting feature of it.

See #1709 for more context.


vanniktech:
Thanks Nicola!

PR #1718, opened 2019-06-22
Run all detekt tasks with project dependencies

3flex:
See https://github.com/arturbosch/detekt/pull/1696#issuecomment-504665839

### DIFF
--- a/detekt-sample-extensions/build.gradle.kts
+++ b/detekt-sample-extensions/build.gradle.kts
@@ -1,13 +1,16 @@
 val assertjVersion: String by project
-val usedDetektVersion: String by project
 val junitPlatformVersion: String by project
 val spekVersion: String by project
 
 dependencies {
-    implementation("io.gitlab.arturbosch.detekt:detekt-api:$usedDetektVersion")
+    // When creating a sample extesion, change this dependency to the detekt-api version you build against, e.g.
+    // io.gitlab.arturbosch.detekt:detekt-api:1.0.0-RC15
+    implementation(project(":detekt-api"))
     implementation(kotlin("compiler-embeddable"))
 
-    testImplementation("io.gitlab.arturbosch.detekt:detekt-test:$usedDetektVersion")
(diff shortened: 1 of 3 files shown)


detekt-sample-extensions/build.gradle.kts
```diff
@@ -1,13 +1,16 @@
 val assertjVersion: String by project
-val usedDetektVersion: String by project
 val junitPlatformVersion: String by project
 val spekVersion: String by project
 
 dependencies {
-    implementation("io.gitlab.arturbosch.detekt:detekt-api:$usedDetektVersion")
+    // When creating a sample extesion, change this dependency to the detekt-api version you build against, e.g.
```
schalkms: extension

detekt-sample-extensions/build.gradle.kts
```diff
@@ -1,13 +1,16 @@
...
+    // io.gitlab.arturbosch.detekt:detekt-api:1.0.0-RC15
+    implementation(project(":detekt-api"))
     implementation(kotlin("compiler-embeddable"))
 
-    testImplementation("io.gitlab.arturbosch.detekt:detekt-test:$usedDetektVersion")
+    // When creating a sample extesion, change this dependency to the detekt-test version you build against, e.g.
```
schalkms: extension

schalkms:
Do we need to update the documentation on the website as well?
I mean the section regarding the extension part.

3flex:
> Why is the test coverage falling down?

Not sure, there's nothing in the PR that should affect coverage at all... possibly Codecov configuration is incorrect or there's something to do with the caching of results that happens between CI runs? For this change only detekt-sample-extensions was rebuilt, the rest of the build would have been cached and reused, so that's probably the reason. Not sure how to fix though.

> Do we need to update the documentation on the website as well?

Good call - I just read through https://arturbosch.github.io/detekt/extensions.html to check. I didn't see anything that would need updating.

PR #1738, opened 2019-06-30
ForbiddenVoid: New option 'ignoreUsageInGenerics'

### LINKED ISSUE #1724: ForbiddenVoid and TestObserver<Void>
This is a follow-up on #1472 - I just stumbled upon the very same issue with RxJava and TestObserver of a Completable:

```kotlin
fun testMe(): TestObserver<Void> = Completable.complete().test()
```

is then also flagged with ForbiddenVoid. IMHO this rule should get some configuration, such as onlyCheckTypesInPackage which we would naturally configure to our project package path and the rule could then ignore types that reside in different (base) packages from that.

realdadfish:
Fixes #1724

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenVoidSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenVoidSpec.kt
@@ -69,8 +69,7 @@ class ForbiddenVoidSpec : Spek({
 
                     class B : A() {
                         override fun method(param: Foo<Void>) : Foo<Void> {
-                            @Suppress("ForbiddenVoid")
-                            return Foo<Void>()
+                            throw IllegalStateException()
                         }
                     }
                 """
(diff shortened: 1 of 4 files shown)


schalkms:
Please execute `gradle build` locally in order to generate the documentation.
Detekt's documentation is automatically generated from the comments in the rule class.
This helps us to have an up to date documentation.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenVoidSpec.kt
```diff
@@ -69,8 +69,7 @@ class ForbiddenVoidSpec : Spek({
 
                     class B : A() {
                         override fun method(param: Foo<Void>) : Foo<Void> {
-                            @Suppress("ForbiddenVoid")
-                            return Foo<Void>()
+                            throw IllegalStateException()
```
schalkms: Why did you change this?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenVoidSpec.kt
```diff
@@ -107,5 +106,42 @@ class ForbiddenVoidSpec : Spek({
                 assertThat(findings).hasSize(2)
             }
         }
+        describe("ignoreUsageInGenerics is enabled") {
```
schalkms: I would prefer to have the following test cases:

1. nested generic void usage (e.g. `A<B<Void>>`)
2. Dictionary (e.g. `Map<Int, Void>`)

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenVoidSpec.kt
realdadfish: It served no purpose for the particular test, but rather confused because of the additional `@Suppress("ForbiddenVoid")`.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenVoidSpec.kt
realdadfish: Will add these.

Issue #1642, opened 2019-05-07
False positive on ForbiddenVoid


mjuanes:
## Observed Behavior
False positive on ForbiddenVoid when theres a property/field/enum/class name/class or something called void.

## Steps to Reproduce
Call a class Void, o create a Void value in a enum.

## Your Environment

* Version of detekt used: RC 14, last one I think.
* Version of Gradle used (if applicable):
* Operating System and version:
* Link to your project (if it's a public repository):

schalkms:
@mjuanes This only reports when you reference properties or classes called `Void`, but not the declaration itself.
Hence, this needs type and symbol solving.
Please take a look at #1740

PR #1740, opened 2019-06-30
Add naming test for ForbiddenVoid rule

schalkms:
This adds tests to reproduce issue #1642

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenVoidSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenVoidSpec.kt
@@ -37,6 +37,20 @@ class ForbiddenVoidSpec : Spek({
             assertThat(subject.compileAndLint(code)).isEmpty()
         }
 
+        it("does not report when functions or classes are called 'Void'") {
+            val code = """
+				class Void {
+                    fun void() {}
+                }
+                
+                enum class E {
+                    Void;
+                }
+			"""
+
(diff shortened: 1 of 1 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenVoidSpec.kt
```diff
@@ -37,6 +37,20 @@ class ForbiddenVoidSpec : Spek({
             assertThat(subject.compileAndLint(code)).isEmpty()
         }
 
+        it("does not report when functions or classes are called 'Void'") {
+            val code = """
+				class Void {
+                    fun void() {}
```
arturbosch: this should be fairly easy to fix, we just check if this name is a child of a declaration?

schalkms: ? What do you want to fix? There's nothing to fix here.

schalkms:
I don't know what problem CodeFactor has. There's no trailing whitespace there.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenVoidSpec.kt
arturbosch: Ah though this is a false positive but now seeing that the declaration name does not trigger it aka the isEmpty check :)

arturbosch:
Not even in the testing code? It look wierd at github though.

PR #1742, opened 2019-06-30
Adds TXT report support on gradle plugin

pak3nuh:
Allows gradle plugin to enable and configure the TXT report.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,7 @@ If you contributed to detekt but your name is not in the list, please feel free 
 - [Daniele Conti](https://github.com/fourlastor) - ObjectPropertyNaming improvement
 - [Nicola Corti](https://github.com/cortinico) - Fixed Suppress of MaxLineLenght
 - [Michael Lotkowski](https://github.com/DownMoney) - Rule improvement: False positive UnusedImport for componentN
+- [Nuno Caro](https://github.com/Pak3nuh) - Adds TXT report support on gradle plugin
 
 ### Mentions
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -119,6 +119,7 @@ subprojects {
         reports {
             xml.enabled = true
             html.enabled = true
+            txt.enabled = true
         }
 
         idea {
@@ -329,6 +330,7 @@ val detektFormat by tasks.registering(Detekt::class) {
     reports {
         xml { enabled = false }
         html { enabled = false }
+        txt { enabled = false }
     }
 }
 
@@ -347,5 +349,6 @@ val detektAll by tasks.registering(Detekt::class) {
     reports {
         xml.enabled = false
         html.enabled = false
+        txt.enabled = false
     }
 }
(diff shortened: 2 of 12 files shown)


pak3nuh:
Let me try to clarify a bit.

The reason why I'm suggesting this is because, from my research, only the TXT report contains the issue signature needed to use on the `baseline.xml` file. Neither the console output nor the other report formats provide that information.

In our environment, we only interact with `detekt` using the gradle plugin, and I didn't saw any other way of getting those signatures.

Also, since the library provides the report, I thought it was a good addition if the plugin also supported it.

schalkms:
> The reason why I'm suggesting this is because, from my research, only the TXT report contains the issue signature needed to use on the baseline.xml file. Neither the console output nor the other report formats provide that information.

This can be done using the following flag.

> The `Codesmell_Signature` is not printed to the console but can be retrieved from the txt output file when using the `--report txt:path/to/report` cli flag.
> If you are using the gradle-plugin run the detektBaseline task to generate yourself a baseline.xml.

Source: https://arturbosch.github.io/detekt/baseline.html

pak3nuh:
If I understood correctly, what you are suggesting is that I should invoke the detektBaseline task time we want to change its content? From what I was, the whitelist section is completely rewritten.

Is that it?

schalkms:
Okay, what do you mean by issue signature then?

pak3nuh:
I meant Codesmell_Signature.

I guess I can achieve what I want invoking the baseline task, which is adding a single entry to the whitelist.

pak3nuh:
Hi,
Today I tried using the `baseline` task to add some entries to the whitelist.

It got the job done but the usability wasn't great. The file already had some entries and it forced me to diff between what was already there and what I wanted to include on the baseline.

If I had the specific Codesmell_Signature it would have been easier.

I still think the TXT report is a useful addition, or at least export the Codesmell_Signature on another report, like XML.

arturbosch:
I'm in favor of adding the txt report to the gradle plugin.
As @pak3nuh said, it is the only way to get to the code smell signature to white or black list issues without recreating the baseline.
That's actually the intention of the baseline. Generate it once and manually modify it in the future.

A workaround for now is to use:
```kotlin
detekt {
...
    cutom {
        reportId = "txt"
        destination = File(...)
    }
}
```

pak3nuh:
Hi all,

@schalkms can you share your thoughts on this, having @arturbosch comment in mind?

schalkms:
Based on our discussion I can see the benefit of this report.
I didn't have the time to review it.
I'll do that the following days.

schalkms:
Once again CodeFactor is really annoying here.

arturbosch:
@pak3nuh please rebase your branch and add the report option to @schalkms mentioned links. thanks!

pak3nuh:
I'll rebase later today.

@arturbosch Are those links generated by code or should I edit the page directly?

schalkms:
@pak3nuh Yes, you can directly edit the markdown files.

PR #1747, opened 2019-07-02
Fix cli execution doc

remal:
There is a mistake in https://arturbosch.github.io/detekt/cli.html. --jvm-target should be `1.6, 1.8, 9, 10, 11 or 12`, not `JVM_1_6, JVM_1_8, JVM_9, JVM_10, JVM_11, JVM_12`


remal:

PR #1748, opened 2019-07-02
Kotlin language version handling

remal:
This PR helps to solve an issue when a user have a Kotlin project with Kotlin version less then Detekt's Kotlin version. Without these fixes a lot of Kotlin embedded compiler warnings occur.

Also a lot of coding style issues have been fixed.


remal:

docs/pages/gettingstarted/cli.md
```diff
@@ -77,6 +77,11 @@ Usage: detekt [options]
     --input, -i
       Input paths to analyze. Multiple paths are separated by comma. If not
       specified the current working directory is used.
+    --jvm-target
+      EXPERIMENTAL: Compatibility mode for Kotlin language version X.Y, reports
+      errors for all language features that came out later.
+      Default: latest stable
```
schalkms: This one was part of another PR.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
```diff
@@ -18,6 +18,7 @@ private const val REPORT_PARAMETER = "--report"
 private const val GENERATE_CONFIG_PARAMETER = "--generate-config"
 private const val CREATE_BASELINE_PARAMETER = "--create-baseline"
 private const val CLASSPATH_PARAMETER = "--classpath"
+private const val LANGUAGE_VERSION_PARAMETER = "--language-version"
```
schalkms: This also needs an update regarding the documentation, since you introduced a new cli flag.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
```diff
@@ -42,6 +43,10 @@ internal data class ClasspathArgument(val fileCollection: FileCollection) : CliA
         fileCollection.joinToString(";") { it.absolutePath }) else emptyList()
 }
 
+internal data class LanguageVersionArgument(val languageVersion: String?) : CliArgument() {
+    override fun toArgument() = languageVersion?.let { listOf(LANGUAGE_VERSION_PARAMETER, it) } ?: emptyList()
```
schalkms: I think this also needs some tests.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -240,10 +257,29 @@ open class Detekt : SourceTask(), VerificationTask {
...
+        val urls = map { it.toURI().toURL() }
+        if (urls.isNotEmpty()) {
+            URLClassLoader(urls.toTypedArray()).use { classLoader ->
+                try {
+                    val clazz = classLoader.loadClass("kotlin.KotlinVersion")
+                    val field = clazz.getField("CURRENT")
+                    field.isAccessible = true
+                    val versionObj = field.get(null)
+                    return versionObj?.toString()
+                } catch (e: Throwable) {
```
schalkms: Are you sure that we need to catch all exceptions?
Are there any specific exceptions which could occur in the try-catch block?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -87,6 +89,14 @@ open class Detekt : SourceTask(), VerificationTask {
     @Optional
     val classpath = project.configurableFileCollection()
 
+    @Input
+    @Optional
+    internal val languageVersionProp: Property<String> = project.objects.property(String::class.javaObjectType)
```
schalkms: What does this flag now really do?
Does it just check if the Kotlin version is okay to work with or is there any other behavior?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/ProcessingSettings.kt
```diff
@@ -28,6 +29,7 @@ data class ProcessingSettings @JvmOverloads constructor(
     val excludeDefaultRuleSets: Boolean = false,
     val pluginPaths: List<Path> = emptyList(),
     val classpath: List<String> = emptyList(),
+    val languageVersion: LanguageVersion = LanguageVersion.LATEST_STABLE,
```
schalkms: Does this mean we have a problem when the Kotlin version used in detekt differs from the _latest_stable_ release provided by JetBrains or this is just a default value?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtCompiler.kt
```diff
@@ -26,9 +26,7 @@ open class KtCompiler(
         val relativePath =
             (if (root == subPath) subPath.fileName
             else root.fileName.resolve(root.relativize(subPath))).normalize()
-        val absolutePath =
-            (if (root == subPath) subPath
```
schalkms: Nice finding!🥇
This condition is an easter egg in detekt. 😉

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
```diff
@@ -29,8 +29,6 @@ class DetektFacade(
     private val saveSupported = settings.autoCorrect
     private val inputPaths = settings.inputPaths
     private val classpath = settings.classpath
-    private val pathFilters = settings.pathFilters
-    private val jvmTarget = settings.jvmTarget
```
schalkms: Are this variables unused?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -12,119 +13,166 @@ interface Args {
     var help: Boolean
 }
 
+@Suppress("MaxLineLength")
```
schalkms: Is this really necessary here?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -12,119 +13,166 @@ interface Args {
...
 class CliArgs : Args {
```
schalkms: This class is hard to review.
I hope you didn't change any meaning of the program.
This PR just does the correct formatting, right?

remal:
This PR fixes this warning:
```
warning: runtime JAR files in the classpath have the version 1.2, which is older than the API version 1.3. Consider using the runtime of version 1.3, or pass '-api-version 1.2' explicitly to restrict the available APIs to the runtime of version 1.2. You can also pass '-language-version 1.2' instead, which will restrict not only the APIs to the specified version, but also the language features
~/.gradle/wrapper/dists/gradle-4.10.3-all/81msde2dx9p4vji0mjgtvxkcb/gradle-4.10.3/lib/kotlin-reflect-1.2.61.jar: warning: runtime JAR file has version 1.2 which is older than required for API version 1.3
~/.gradle/wrapper/dists/gradle-4.10.3-all/81msde2dx9p4vji0mjgtvxkcb/gradle-4.10.3/lib/kotlin-script-runtime-1.2.61.jar: warning: runtime JAR file has version 1.2 which is older than required for API version 1.3
~/.gradle/wrapper/dists/gradle-4.10.3-all/81msde2dx9p4vji0mjgtvxkcb/gradle-4.10.3/lib/kotlin-stdlib-1.2.61.jar: warning: runtime JAR file has version 1.2 which is older than required for API version 1.3
~/.gradle/wrapper/dists/gradle-4.10.3-all/81msde2dx9p4vji0mjgtvxkcb/gradle-4.10.3/lib/kotlin-stdlib-common-1.2.61.jar: warning: runtime JAR file has version 1.2 which is older than required for API version 1.3
~/.gradle/wrapper/dists/gradle-4.10.3-all/81msde2dx9p4vji0mjgtvxkcb/gradle-4.10.3/lib/kotlin-stdlib-jdk7-1.2.61.jar: warning: runtime JAR file has version 1.2 which is older than required for API version 1.3
~/.gradle/wrapper/dists/gradle-4.10.3-all/81msde2dx9p4vji0mjgtvxkcb/gradle-4.10.3/lib/kotlin-stdlib-jdk8-1.2.61.jar: warning: runtime JAR file has version 1.2 which is older than required for API version 1.3
~/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-reflect/1.2.61/6fd5bcfc9ffc446dd147ea006bee7ef5f0ad8ca4/kotlin-reflect-1.2.61.jar: warning: runtime JAR file has version 1.2 which is older than required for API version 1.3
~/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib/1.2.61/5bc44acc4b3f0d19166ae3e50454b41e8ff29335/kotlin-stdlib-1.2.61.jar: warning: runtime JAR file has version 1.2 which is older than required for API version 1.3
~/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib-common/1.2.61/772de03e12d932f489e41aef997d26c20a4ebee6/kotlin-stdlib-common-1.2.61.jar: warning: runtime JAR file has version 1.2 which is older than required for API version 1.3
```

schalkms:
@remal you need to use the `--language-version` flag in order to fix this warning, right? Otherwise, the warning still gets displayed, which isn't a bad thing IMHO.

docs/pages/gettingstarted/cli.md
remal: It's a misprint. I meant `--language-version`. Fixed.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
remal: Please see my comment above. Documentation waas updated, but with a misprint. I've fixed it.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
remal: Yes, we can try to catch specific exceptions. But here my goal was to remain the code as stable as possible. If you don't mind, I'd like to leave as it is.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
remal: On the one hand, I agree, we need tests. But on the other hand, it would be really challenging to implement such test. Give me some time to think about it...

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
remal: You can set language version here. If it's not set, the version will be retrieved automatically. In 99% situations you don't have to set anything here. But having this property will be useful if automatically version detection fails...

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/ProcessingSettings.kt
remal: This is just a default value. Correct.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtCompiler.kt
remal: IDEA with SonarLint plugin helps a lot :)

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
remal: Yep. They are unused.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
remal: We have a comment in the file with this link:         https://github.com/JetBrains/kotlin/blob/master/compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/K2JVMCompilerArguments.kt

It can't be wrapped (it's a link), so MaxLineLength constraint fails here. Also, we can't annotate comments with `@Suppress`. I don't like the solution, but... Do you have any suggestions here? For example, we can add the issue to the project's baseline.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
remal: For the most of changed lines - yes, just reformatting. The most important change here is adding `var languageVersion: LanguageVersion = LanguageVersion.LATEST_STABLE`. If you want, I'll revert reformatting to make the PR easier to read.

remal:
@schalkms Yes, you're right. I'm trying to analyze sources of a project that depends on an older version of Kotlin (1.2). So I see this warning every time I run Detekt.

docs/pages/gettingstarted/cli.md
```diff
@@ -77,6 +77,11 @@ Usage: detekt [options]
     --input, -i
       Input paths to analyze. Multiple paths are separated by comma. If not
       specified the current working directory is used.
+    --language-version
+      EXPERIMENTAL: Compatibility mode for Kotlin language version X.Y, reports
```
schalkms: > reports  errors for all language features that came out later.

What do you mean by that? Could you please show an example to help me understand the problem?

remal:
@schalkms I have an idea to move automatically Kotlin version detection from Detekt's Gradle plugin to `detekt-cli`. It can be done by processing `--classpath` parameter. I'll try to make it tomorrow.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
schalkms: Ok, why was this not reported before?

schalkms:
@remal is this PR then obsolete?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
schalkms: > If you want, I'll revert reformatting to make the PR easier to read.

This is not necessary.
RESOLVED

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
schalkms: Accepted

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
schalkms: Accepted

schalkms: RESOLVED

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
schalkms: I guess it's because of the formatting, right?
RESOLVED

docs/pages/gettingstarted/cli.md
remal: Actually I haven't seen such warnings. I suppose we can get them if Kotlin JARs in the classpath are not aligned, for example.

remal:
@schalkms I've pushed a new commit with moving version retrieval logic.

BTW I haven't tested it in Java >= 9 environment.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/KotlinEnvironmentUtils.kt
```diff
@@ -95,18 +95,18 @@ fun createCompilerConfiguration(
 private fun Iterable<File>.getKotlinLanguageVersion(): LanguageVersion? {
```
schalkms: You might want to unit test this single method.
Then we are also sure that it works on all JVM versions with the help of CI.

remal: I've added a test. See `KotlinEnvironmentUtilsTest`.

This test does not:
1. Test the code against different versions of Kotlin
1. Test the code against different versions of JVM


PS: Asking for tests was very useful - I catched a bug :)

arturbosch:
The test seems to fail on Java 9+, any idea why?
```
io.gitlab.arturbosch.detekt.api.internal.KotlinEnvironmentUtilsTest > should match FAILED
    kotlin.KotlinNullPointerException
        at io.gitlab.arturbosch.detekt.api.internal.KotlinEnvironmentUtilsTest$1$1$1.invoke(KotlinEnvironmentUtilsTest.kt:18)
        at io.gitlab.arturbosch.detekt.api.internal.KotlinEnvironmentUtilsTest$1$1$1.invoke(KotlinEnvironmentUtilsTest.kt:9)
```

remal:
@arturbosch I'm afraid I didn't understand what you meant by "revert the continuous indent to 4 again". Could you please give a specific example? I'd appreciate it if you could make a comment to a specific line.

I've fixed the test. Please take a look. It seems that Application ClassLoader is not an instance of `URLClassLoader` in Java 9...

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -12,119 +13,166 @@ interface Args {
...
 
-    @Parameter(names = ["--report", "-r"],
+    @Parameter(
+        names = ["--report", "-r"],
         description = "Generates a report for given 'report-id' and stores it on given 'path'. " +
-            "Entry should consist of: [report-id:path]. " +
-            "Available 'report-id' values: 'txt', 'xml', 'html'. " +
-            "These can also be used in combination with each other " +
-            "e.g. '-r txt:reports/detekt.txt -r xml:reports/detekt.xml'")
+                "Entry should consist of: [report-id:path]. " +
```
arturbosch: The whole indentation in this file for example got reformatted to 8 continuous spaces instead of four.
I like the annotation parameter formatting though :).

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -12,119 +13,166 @@ interface Args {
...
         The following @Parameters are used for type and symbol resolving. When additional parameters are required the
-        names should mirror the names found in this file (e.g. "classpath", "jvm-target"):
+        names should mirror the names found in this file (e.g. "classpath", "language-version", "jvm-target"):
         https://github.com/JetBrains/kotlin/blob/master/compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/K2JVMCompilerArguments.kt
     */
     @Parameter(
         names = ["--classpath", "-cp"],
         description = "EXPERIMENTAL: Paths where to find user class files and depending jar files. " +
-            "Used for type resolution."
+                "Used for type resolution."
```
arturbosch: Like this ^^

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
remal: Take a look. Is it OK now?

remal:
@arturbosch Please take a look

PR #1749, opened 2019-07-02
FindBugs XML report

remal:
`FindBugsXmlOutputReport` has been added. Tests are included.

### DIFF
--- a/detekt-api/build.gradle.kts
+++ b/detekt-api/build.gradle.kts
@@ -1,4 +1,9 @@
 import org.jetbrains.dokka.gradle.DokkaTask
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+import java.nio.file.Files.createDirectories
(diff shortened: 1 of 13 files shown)


remal:

3flex:
Is the checkstyle XML report not sufficient?

remal:
@3flex Checkstyle reports don't have issue descriptions, but FindBugs - have

remal:
@arturbosch 
FindBugs XML report can handle classes and methods. But I haven't found a way to get classes and methods from Detekt. So I decided to output only files and file positions.

I've tested the report against my own parser, but I've **not** tested the report against these libraries. Maybe it makes sense to think about creating such test... WDYT?

arturbosch:
@remal yes please use https://github.com/tomasbjerre/violations-lib as testImplementation for detekt-api and test the parsing. They spend quite a bit on parsing Findbugs.

PR #1752, opened 2019-07-02
Don't fail when codecov test coverage decreases

schalkms:
This PR introduces a threshold in order to not fail the PR when the 
codecov test coverage decreases slightly by 0.1 %.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,3 +4,9 @@ codecov:
   notify:
     require_ci_to_pass: no
     after_n_builds: 4
+coverage:
+  status:
+    # pull-requests only
+    patch:
+      default:
+        threshold: 0.1%


.codecov.yml
```diff
@@ -1,3 +1,9 @@
 codecov:
   ci:
     - !appveyor
+coverage:
+  status:
+    # pull-requests only
+    patch:
+      default:
+        threshold: 0.1%
```
schalkms: This is of course debatable.

3flex:
I think this depends on why we have test coverage in the first place? If it's just to make sure that there are no big unexpected drops because a PR accidentally broke the test setup that's one thing, but if we're hoping to increase coverage over time I wouldn't support this since each PR should at worst not change coverage, or improve it.

But all this relies on codecov working reliably, and I think we need more time to see if the current config is stable.

So I'd propose not failing at all when codecov shows reduced coverage for now, and once we trust it's working properly, then fail either on lowered coverage or accept 0.1% or similar drops as in this PR.

3flex:
See #1754... I think my latest attempt to get Codecov working has failed, so I don't think it's worth merging this until that's resolved.

schalkms:
I'm reopening this in the hope that the issues with codecov are now resolved with #1865.

schalkms:
> if we're hoping to increase coverage over time I wouldn't support this since each PR should at worst not change coverage, or improve it

I overlooked this argument.
Agreed. Then I also think we should close this PR.

arturbosch:
I also agree. Better to enforce it and make per case exceptions while merging ^^.

PR #1758, opened 2019-07-05
Update EmptyClassBlock to skip classes with comments in the body

### LINKED ISSUE #1756: EmptyClassBlock should consider comment as "body" (via option?)
```kotlin
class Foo {
    // this is a comment about why Foo doesn't have a body (implementation detail)
}
```
the above shouldn't report `EmptyClassBlock`, IMO

## Context
Documentation like this belongs to the class, but it's not KDoc as it's not the description of a public API.
Without the class body braces the comment would be just dangling in space, and for example Move (class) refactor wouldn't consider it to be part either and would leave it where it is.

```kotlin
class Foo
// just a random comment unrelated to Foo
```

cortinico:
This is similar to #1690 but I'm fixing the `EmptyClassBlock` rule here

Fixes #1756

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyClassBlockSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyClassBlockSpec.kt
@@ -33,7 +33,7 @@ class EmptyClassBlockSpec : Spek({
         it("does not report class with multiline comments in the body") {
             val code = """
                 class SomeClass {
-                    /* 
+                    /*
                     Some comment to explain what this class is supposed to do
                     */
                 }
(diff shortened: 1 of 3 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyClassBlockSpec.kt
```diff
@@ -21,6 +21,26 @@ class EmptyClassBlockSpec : Spek({
...
+                }
+            """.trimIndent()
+            assertThat(subject.compileAndLint(code)).isEmpty()
+        }
+
+        it("does not report class with multiline comments in the body") {
+            val code = """
+                class SomeClass {
+                    /* 
+                    Some comment to explain what this class is supposed to do
```
schalkms: It's nerve-wracking that codefactor fails because of this warning.

arturbosch: Oh but there is a trailing whitespace after the `*` :D
It's right, the question is why our CI does not want it?

cortinico: I fixed it but is still failing :/

PR #1762, opened 2019-07-06
Preset ignoreFailures property with false as it is also used by Gradle

arturbosch:


marschwar:
I think `ignoreFailuresProp` should be private (as well as all the other fields ending on Prop). But that really has nothing to do with this PR.

PR #1710, opened 2019-06-19
[WIP] Use reflection to invoke detekt cli in gradle plugin - #1686

arturbosch:
Prototype to use reflection to invoke detekt.
It has the advantage that we do not need to start a new process and we do not have the limitation of too long filenames on windows.

@marschwar I've got some problems with the testcases. I'm not sure what is expected to get logged for example in `DetektTaskDslTest`. It looks like the passed arguments to the process are validated?
With this patch the normal detekt output gets printed.
```
java.lang.AssertionError: 
Expecting:
 <"The client will now receive all logging from the daemon (pid: 30403). The daemon log file: /tmp/.gradle-test-kit-abosch/test-kit-daemon/5.4.1/daemon-30403.out.log
Starting 5th build in daemon [uptime: 6.595 secs, performance: 94%, non-heap usage: 15% of 268.4 MB]
Closing daemon's stdin at end of input.
The daemon will no longer process any standard input.
Using 8 worker leases.
Starting Build
Settings evaluated using settings file '/tmp/applyPlugin7974105155632345249.tmp/settings.gradle'.
Using local directory build cache for the root build (location = /tmp/.gradle-test-kit-abosch/caches/build-cache-1, removeUnusedEntriesAfter = 7 days).
Projects loaded. Root project using build file '/tmp/applyPlugin7974105155632345249.tmp/build.gradle'.
Included projects: [root project 'rootDir-project']

> Configure project :
Evaluating root project 'rootDir-project' using build file '/tmp/applyPlugin7974105155632345249.tmp/build.gradle'.
All projects evaluated.
Selected primary task 'detekt' from project :
Tasks to be executed: [task ':detekt']
:detekt (Thread[Execution worker for ':',5,main]) started.

> Task :detekt
Build cache key for task ':detekt' is 16a72903d57a065657844231e33863a9
Task ':detekt' is not up-to-date because:
  No history is available.


1 kotlin file was analyzed.

Ruleset: comments
Ruleset: complexity
Ruleset: empty-blocks
Ruleset: exceptions
Ruleset: naming
Ruleset: performance
Ruleset: potential-bugs
Ruleset: style

Complexity Report:
	- 6 lines of code (loc)
	- 5 source lines of code (sloc)
	- 4 logical lines of code (lloc)
	- 0 comment lines of code (cloc)
	- 0 McCabe complexity (mcc)
	- 0 number of total code smells
	- 0 % comment source ratio
	- 0 mcc per 1000 lloc
[...]

PR #1769, opened 2019-07-12
Implement dry-run option for detekt gradle tasks.

marschwar:
This is sort of related to #1710 

Right now there are tests that check that the correct parameters are passed to the CLI by scraping the task output. Since the arguments are now written to a file to work around limitations on Windows, the somewhat dirty fix was to log the arguments with debug level and run gradle with the --debug option. Unfortunately this makes the output useless for actual debugging in the tests.

I opened this PR to make a suggestion for a possible solution and to gather feedback.
This dry-run option would if enabled simply print the arguments without actually invoking the CLI. In most cases this should be sufficient since we want to test the gradle task and not the CLI as well.

A different solution would be check the arguments from the `detekt.args` file.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
@@ -4,13 +4,37 @@ import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.file.FileCollection
 
-object DetektInvoker {
-    internal fun invokeCli(
-        project: Project,
+internal interface DetektInvoker {
+    fun invokeCli(
         arguments: List<CliArgument>,
         classpath: FileCollection,
         taskName: String,
         ignoreFailures: Boolean = false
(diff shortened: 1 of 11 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
```diff
@@ -4,17 +4,40 @@ import org.gradle.api.GradleException
...
+        arguments: List<CliArgument>,
+        classpath: FileCollection,
+        taskName: String,
+        ignoreFailures: Boolean = false
+    )
+
+    companion object {
+        fun create(project: Project): DetektInvoker =
+            if (project.isDryRunEnabled())
+                DryRunInvoker(project)
```
arturbosch: Please use braces when having a multiline if-statement.
Should get flagged by ktlint however we forgot to add the formatting rule set to the gradle plugin ...

marschwar:
I converted most of the DSL tests to use the `--dry-run` option. By doing so it only checks the transformation from detekt config to CLI parameters instead of checking the output of detekt.

PR #1771, opened 2019-07-19
Remove author tags

### LINKED ISSUE #1627: Drop @author tags
I hope this isn't too controversial...!

I propose that the `@author` KDoc tags be removed from the code base, for the following reasons:

* They're out of date
* Full history available in git log
* More useful to use `git blame` to find out who changed what, since the current author tags are out of date
* They're only shown in source code
* README is already used to highlight some major contributors

I definitely don't want to take away form anyone's contributions (especially from @arturbosch!) I'm just not sure the author tag is the best way to record them.

schalkms:
Closes #1627

This was done with the cli commands `sed` and `perl`

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
@@ -4,10 +4,6 @@ import com.beust.jcommander.Parameter
 import org.jetbrains.kotlin.config.JvmTarget
 import java.nio.file.Path
 
-/**
- * @author Artur Bosch
- * @author Marvin Ramin
- */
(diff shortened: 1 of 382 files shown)


cortinico:
Hey @schalkms 

As a side note I was working on a Rule to detect the `@author` tag in the KDoc comments. I haven't yet managed to send a PR out as it's not production ready yet. Maybe we shouldn't close #1627 with this change?

3flex:
@cortinico glad you could take that on!

I think this can be merged though since it closes the original issue. If you want to create a new issue for the idea for a new rule to detect forbidden KDoc tags that'd be fine.

schalkms:
@cortinico 
**ForbiddenComment** already provides this function. You can add the *author* tag to the configuration.
Thanks for your efforts!

Issue #1768, opened 2019-07-12
XML report consolidation does not work if parent project does not have sources


marschwar:
## Expected Behavior
XML reports should be consolidated for all multi module projects.

## Observed Behavior
If the parent project does not have any sources, the detekt task is being skipped with `NO_SOURCE`. This is due to the `@SkipWhenEmpty` in the `SourceSetTask`.

## Steps to Reproduce
Change the number of source files of the parent in the `DetektTaskMultiModuleConsolidationTest`

marschwar:
With #1774 being merged this can be closed.

PR #1774, opened 2019-07-20
Remove xml report consolidation.

marschwar:
I realize this is going to be a bit controversial but I wanted to get the discussion started anyway.

IMHO the report consolidation as it currently stands should be removed. Reasons for that being:

- It does not work if the root project does not have any sources (#1768).
- It does not work with source set tasks such as `detektMain`
- It makes the code inside the Detekt tasks somewhat clunky and harder to maintain.
- I believe there is a workaround by running detekt in the root project only with the root folder as the input.

I realize that this was heavily requested feature, but the way it is implemented simply introduces more problems than benefits.

Reverts #1329


marschwar:
Thank you for making me take another look. While most of the unreverted changes were refactorings in test code and formatting there was one more thing that could be removed. With this the detekt task does not have to depend on the detekt task of its child projects.

arturbosch:
I also appreciate your work and investigations on this one!
With this blocker out, the way is open for 1.0 and starting 'normal' versioning.
We can deliver this feature better in a minor 1.X update later :)

PR #1778, opened 2019-07-21
Add OWASP dependency-check

schalkms:
The dependency-check identifies the projects dependencies and reports on
known vulnerabilities.
The build fails when the CVSS score is equal or above the severity level
medium.

**Current output**:
```
> Task :dependencyCheckAnalyze
Verifying dependencies for project detekt
Checking for updates and analyzing dependencies for vulnerabilities
Generating report for project detekt
Found 0 vulnerabilities in project detekt
``

### CODE REVIEW

File: gradle.properties [hunk 1]
```diff
@@ -13,3 +13,5 @@ kotlin.code.style=official
 systemProp.sonar.host.url=http://localhost:9000
 systemProp.detektVersion=detektVersion
 systemProp.file.encoding=UTF-8
+
+severityLevel=4.0
```
arturbosch: What does this value mean? why you choose 4 :) ?

build.gradle.kts
```diff
@@ -41,6 +42,23 @@ tasks.wrapper {
     }
 }
 
+apply {
+    plugin("org.owasp.dependencycheck")
```
arturbosch: Isn't the plugins { id ... } enought?

gradle.properties
schalkms: This is documented in the build script.

```Kotlin
/*
* Specifies that the build fails when the CVSS score is equal or above the severity level medium.
* CVSS scoring explanation:
* https://nvd.nist.gov/vuln-metrics/cvss?source=post_page
*/
failBuildOnCVSS = severityLevel.toFloat()
```

build.gradle.kts
schalkms: I will try that. I think you are right. Thanks for the heads up.

schalkms:
I think it's weird that the build hangs on Windows using the JDK 1.8.

arturbosch:
I've restarted it

schalkms:
> Unable to download meta file: https://nvd.nist.gov/feeds/json/cve/1.0/nvdcve-1.0-modified.meta

They seem do have some throttling going on to prevent abuse. 
I added a property `failOnError = false` so that the build doesn't fail, when an error like this happens.

schalkms:
`failOnError = false` doesn't help.

> They seem do have some throttling going on to prevent abuse.

Looks like one needs to setup a local mirror on any kind of server, which we don't have.

arturbosch:
Thats sad, many more people reporting this behavior. There is also 5.2 out, but I don't thing this will solve these issues. So closing this?

PR #1783, opened 2019-07-26
Fix UseDataClass false positive (delegation)

### LINKED ISSUE #1585: UseDataClass false positive
A delegate is used to save data in Android shared preferences, so this class should not be a data class.

## Expected Behavior
Do not report this as a rule violation

## Observed Behavior
Reports as a rule violation

## Steps to Reproduce
```
class SharedPreferenceRepository(
    context: Context
) {
    companion object {
        private const val PREFIX = "APP"
        private const val INSTALL_UUID: String = "${PREFIX}_INSTALL_UUID"
    }

    private var _installUUID: String by SharedPreferencesExtension(context, INSTALL_UUID, "")
}
```

schalkms:
A class containing delegated properties is no data class candidate.
This closes #1585

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClass.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClass.kt
@@ -10,15 +10,15 @@ import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.SplitPattern
 import io.gitlab.arturbosch.detekt.rules.collectByType
+import io.gitlab.arturbosch.detekt.rules.doesNotExtendAnything
+import io.gitlab.arturbosch.detekt.rules.isClosedForExtension
 import io.gitlab.arturbosch.detekt.rules.isInline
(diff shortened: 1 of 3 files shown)


schalkms:
CodeFactor does not rerun when I force it. Interesting..

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClass.kt
```diff
@@ -93,12 +94,6 @@ class UseDataClass(config: Config = Config.empty) : Rule(config) {
                 klass.secondaryConstructors.all { it.isPrivate() }
     }
 
-    private fun KtClass.doesNotExtendAnything() = superTypeListEntries.isEmpty()
-
-    private fun KtClass.isClosedForExtension() = !isAbstract() && !isOpen()
-
-    private fun KtClass.extractDeclarations(): List<KtDeclaration> = body?.declarations.orEmpty()
```
arturbosch: These weren't used?

schalkms: I moved them to the utility class `KtClassOrObjectExtensions.kt`.

arturbosch: ah I see

PR #1785, opened 2019-07-26
Document meanings of rule severity levels

### LINKED ISSUE #1630: Document meanings of rule severity levels
These severities are available, but there's no description for them.

I think each should have a single sentence string value which can then be displayed on the documentation site. Looking for suggestions on what each should mean:

1. CodeSmell
2. Style
3. Warning
4. Defect
5. Minor
6. Maintainability
7. Security
8. Performance

It might also be a good time to rename, add or deprecate some of these. I feel CodeSmell and Warning are too generic, and Minor isn't clear (it's a minor what? Minor code smell? Minor defect?)

schalkms:
Closes #1630

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
@@ -26,7 +26,39 @@ data class Issue(
  * a grade which is most harmful to their projects.
  */
 enum class Severity {
-    CodeSmell, Style, Warning, Defect, Minor, Maintainability, Security, Performance
+    /**
+     * Represents clean coding violations which may lead to maintainability issues.
+     */
+    CodeSmell,
+    /**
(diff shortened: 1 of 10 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
```diff
@@ -26,7 +26,39 @@ data class Issue(
...
+     */
+    Warning,
+    /**
+     * Corresponds to coding mistakes which could lead to unwanted behavior.
+     */
+    Defect,
+    /**
+     * Represents code quality issues which only slightly impact the code quality.
+     */
+    Minor,
```
arturbosch: Lets delete `Minor`, last time I searched for usages, it was only referened once. We have a generic code smell or style severity? what do you think?

schalkms: I would leave it. We are changing the API here. We potentially break 3rd code/rules.

PR #1793, opened 2019-07-31
Initial MCC change

schalkms:
reference https://www.ndepend.com/docs/code-metrics#CC

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
@@ -3,10 +3,12 @@ package io.gitlab.arturbosch.detekt.api.internal
 import io.gitlab.arturbosch.detekt.api.DetektVisitor
 import org.jetbrains.kotlin.psi.KtBlockExpression
 import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtContinueExpression
 import org.jetbrains.kotlin.psi.KtIfExpression
 import org.jetbrains.kotlin.psi.KtLoopExpression
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtObjectLiteralExpression
+import org.jetbrains.kotlin.psi.KtOperationReferenceExpression
 import org.jetbrains.kotlin.psi.KtTryExpression
 import org.jetbrains.kotlin.psi.KtWhenEntry
 import org.jetbrains.kotlin.psi.KtWhenExpression
@@ -35,8 +37,10 @@ class McCabeVisitor(private val ignoreSimpleWhenEntries: Boolean) : DetektVisito
 
     override fun visitIfExpression(expression: KtIfExpression) {
         mcc++
-        if (expression.`else` != null) {
-            mcc++
+        val condition = expression.condition
+        if (condition != null) {
+            mcc += condition.children.count { it is KtOperationReferenceExpression
+                    && it.text == "&&" || it.text == "||" }
         }
         super.visitIfExpression(expression)
     }
@@ -58,12 +62,7 @@ class McCabeVisitor(private val ignoreSimpleWhenEntries: Boolean) : DetektVisito
     }
 
     override fun visitTryExpression(expression: KtTryExpression) {
-        mcc++
         mcc += expression.catchClauses.size
-        expression.finallyBlock?.let {
-            mcc++
-            Unit
-        }
         super.visitTryExpression(expression)
     }
 
@@ -79,6 +78,11 @@ class McCabeVisitor(private val ignoreSimpleWhenEntries: Boolean) : DetektVisito
             }
         }
         super.visitCallExpression(expression)
+    }
+
+    override fun visitContinueExpression(expression: KtContinueExpression) {
+        mcc++
+        super.visitContinueExpression(expression)
     }
 }
 
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
@@ -1,9 +1,9 @@
 package io.gitlab.arturbosch.detekt.api.internal
 
 import io.gitlab.arturbosch.detekt.api.DetektVisitor
+import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.KtBlockExpression
 import org.jetbrains.kotlin.psi.KtCallExpression
-import org.jetbrains.kotlin.psi.KtContinueExpression
 import org.jetbrains.kotlin.psi.KtIfExpression
 import org.jetbrains.kotlin.psi.KtLoopExpression
 import org.jetbrains.kotlin.psi.KtNamedFunction
@@ -12,6 +12,7 @@ import org.jetbrains.kotlin.psi.KtOperationReferenceExpression
 import org.jetbrains.kotlin.psi.KtTryExpression
 import org.jetbrains.kotlin.psi.KtWhenEntry
 import org.jetbrains.kotlin.psi.KtWhenExpression
+import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
 import org.jetbrains.kotlin.psi.psiUtil.getCallNameExpression
 import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 
@@ -21,9 +22,7 @@ import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 class McCabeVisitor(private val ignoreSimpleWhenEntries: Boolean) : DetektVisitor() {
 
     var mcc: Int = 0
-        private set(value) {
-            field = value
-        }
+        private set
 
     override fun visitNamedFunction(function: KtNamedFunction) {
         if (!isInsideObjectLiteral(function)) {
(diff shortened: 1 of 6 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
```diff
@@ -58,12 +62,7 @@ class McCabeVisitor(private val ignoreSimpleWhenEntries: Boolean) : DetektVisito
     }
 
     override fun visitTryExpression(expression: KtTryExpression) {
-        mcc++
         mcc += expression.catchClauses.size
-        expression.finallyBlock?.let {
```
arturbosch: ah good catch, try-finally are always executed so the path is linear

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
```diff
@@ -35,8 +37,10 @@ class McCabeVisitor(private val ignoreSimpleWhenEntries: Boolean) : DetektVisito
 
     override fun visitIfExpression(expression: KtIfExpression) {
         mcc++
-        if (expression.`else` != null) {
-            mcc++
+        val condition = expression.condition
+        if (condition != null) {
+            mcc += condition.children.count { it is KtOperationReferenceExpression
+                    && it.text == "&&" || it.text == "||" }
```
arturbosch: As far as I know the original cyclomatic complexity was defined without counting && and ||.
An alternative the "extended cyclomatic complexity" counts them too.
We should only count them with an extra flag to not break uses builds imo.

I can't find some good reference for the above as I do not have access to the university account anymore ... we could take a look at the original paper -> https://ieeexplore.ieee.org/document/1702388

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
```diff
@@ -26,9 +26,10 @@ import org.jetbrains.kotlin.psi.KtWhenExpression
  * Complex methods use too many of the following statements.
  * Each one of them adds one to the complexity count.
  *
- * - __Conditional statements__ - `if`, `else`, `when`
+ * - __Conditional statements__ - `if`, `else if`, `when`
  * - __Loops__ - `for`, `while`, `do-while`, `forEach`
```
arturbosch: should we move forEach to the scope functions?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
```diff
@@ -80,6 +79,11 @@ class McCabeVisitor(private val ignoreSimpleWhenEntries: Boolean) : DetektVisito
         }
         super.visitCallExpression(expression)
     }
+
+    override fun visitContinueExpression(expression: KtContinueExpression) {
+        mcc++
+        super.visitContinueExpression(expression)
```
arturbosch: Does continue introduce a new path? I would argue that continues are guarded by some conditions (if), so this shouldn't be counted?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
schalkms: True

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
schalkms: I'm not sure. I think it's also okay to have it in the `loops` section.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
schalkms: You can argue that this is basically a jump (`goto`). I think we can leave this out for now.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
schalkms: Just because Kotlin/Java abstract the branches away into a boolean expression, doesn't mean they aren't there.
I think it might be wise to rename this class to `CyclomaticComplexity` to avoid confusion.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
schalkms: In my opinion the safe navigation operator `?.` and the null coalescing operator `?:` also needs to be added. What do you think?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
arturbosch: +1, so should we leave the old implementation as is and move this changes to a new rule `CyclomaticComplexity`?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
arturbosch: I would make this configurable. Most users wont like this change, however per definition this makes sense and they add a bit of clutter.
If you inspect the bytecode there are new paths, but they do not add to mental complexity...

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/McCabeVisitor.kt
schalkms: No, I meant to rename the `McCabeVisitor` class to `CyclomaticComplexity`.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
schalkms: > If you inspect the bytecode there are new paths, but they do not add to mental complexity

I mean it's all about the code paths, right?

arturbosch: Yes, true. However when this metric was proposed no language implemented safe navigation. Most developers won't even notice the additional branches.
It's an usability-formality tradeoff.

schalkms: Yes. However, the same holds true for the `Scope Functions` part. So I have mixed feelings about leaving that out.

arturbosch: To be not too controversial I like having the scope functions part configurable and the safe navigation part.

schalkms: I disagree here. The cyclomatic complexity also determines the number of test cases required for a function. A programmer could trick the analyzer in rewriting the following code. The first example reports 2 the second example 1. Anyways, the programmer needs to write two test cases covering both branches. I know that some users may not like it, but other handle this similarly. We might want to increase the allowed complexity from 10 to 15 for instance.

```Kotlin
fun f() {
  if (foo != null) 
    foo.bar()
}
```

```Kotlin
fun f() {
  foo?.bar()
}
```

schalkms:
I'll proceed with this PR and update the obvious elements (for instance the try/catch blocks).
I'll update the *"more controversial"* elements in another PR.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
arturbosch: Ok, what do you think about chained safe navigation `foo?.bar?.toString() ?: "bla"`?
Would you count all `?` or just one because obviously only the non-null and the ?: case are interesting.

schalkms: I think I need to do further research in that regard. 
I will change the obvious things in this PR (if I have time).

- try/catch count
- if logic (with **||** and **&&**) because every other tool does it that way

We mention this in the changelog and we will be fine.
The rest is subject for research and perhaps for another PR.

amitdash291: IMO a strict check should consider all the branches in chained safe navigation, and not only the first one. For eg: Jacoco considers all branches of chained safe navigation for code coverage.

schalkms: @amitdash291 Jacoco works on the lower-level bytecode, whereas detekt uses the PSI.

detekt-core/src/test/resources/cases/ComplexClass.kt
```diff
@@ -3,11 +3,11 @@ package cases
 import org.jetbrains.kotlin.utils.sure
 
 @Suppress("unused")
-class ComplexClass {// McCabe: 56, LLOC: 20 + 20 + 4x4
+class ComplexClass {// McCabe: 44, LLOC: 20 + 20 + 4x4
```
schalkms: These test cases need to be moved to multiple unit tests.
Furthermore, we should test the `McCabeVisitor.kt` class intensively and not every usage of it.

schalkms:
I'd still merge this for the next version, since it fixes the wrong calculation.

arturbosch:
Sorry, I expect multiple issues to be raised for different releases.
I would join forces with you after 1.1 is out to get this in 1.2.

schalkms:
@arturbosch I see. However, I'd then recommend having more than 1 PR.
The change set would get too large and hard to review.

PR #1801, opened 2019-08-03
Gradle plugin: Set default path for "config" parameter

### LINKED ISSUE #1262: Default Gradle config path should be config/detekt/config.yml
## Expected Behavior
Default config path is one of:
* `$rootProject.projectDir/config/detekt/config.yml`
* `$projectDir/config/detekt/config.yml`

## Current Behavior
Config is generated at `$projectDir/default-detekt-config.yml`

## Context
Sensible defaults make for a better onboarding experience. Gradle conventions could be followed to generate the config file at a conventional path instead of in the project root.

Placing the file in `$projectDir/config/detekt/config.yml` makes most sense, so a different config is used for each Gradle project (important for example when there are Android-specific rules that apply to one project that don't apply to other projects in a multi-project build).

3flex:
This PR:
* Enhances CLI to allow creation of the default config file at a path of the user's choice on the CLI
* Uses that feature in the Gradle plugin to generate the config file at `rootDir/config/detekt/detekt.yml` by default (to match conventions in other Gradle plugins)
* Sets that path as the default config path if a file exists at that location
* Moves some of detekt's config files to match the new convention

Closes #1262


.travis.yml
```diff
@@ -21,7 +21,7 @@ script:
   - ./gradlew build --build-cache --parallel -PwarningsAsErrors=true --scan
   - ./gradlew shadowJar
   - java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar --help
-  - java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar -i . --baseline ./reports/baseline.xml -ex "**/resources/**,**/detekt*/build/**" -c ./detekt-cli/src/main/resources/default-detekt-config.yml,./reports/failfast.yml
+  - java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar -i . --baseline ./reports/baseline.xml -ex "**/resources/**,**/detekt*/build/**" -c ./detekt-cli/src/main/resources/default-detekt-config.yml,./config/detekt/detekt.yml
```
schalkms: Why did you rename our baseline file?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/ConfigExporter.kt
```diff
@@ -1,14 +1,16 @@
...
-class ConfigExporter : Executable {
+class ConfigExporter(private val arguments: CliArgs) : Executable {
 
     override fun execute() {
+        val configPath = if (arguments.config != null) arguments.config!! else DEFAULT_CONFIG
         val defaultConfig = ClasspathResourceConverter().convert(DEFAULT_CONFIG).openStream()
-        defaultConfig.copyTo(File(DEFAULT_CONFIG).outputStream())
-        println("\nSuccessfully copied $DEFAULT_CONFIG to project location.")
+        defaultConfig.copyTo(File(configPath).outputStream())
+        println("Successfully copied default config to ${File(configPath).absolutePath}")
```
schalkms: `\n` is missing

schalkms:
```
LargeClass - 832/600 - [UnusedPrivateMemberSpec] at /home/travis/build/arturbosch/detekt/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt:12:1
```

That's interesting, since you didn't touch the `UnusedPrivateMemberSpec` class. 
Something seems to be wrong here.

.travis.yml
3flex: If detekt has a default config file layout then I figured that the project itself should use the same conventions.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/ConfigExporter.kt
3flex: Does it need a leading newline? I thought it looked better in Gradle's output without it (Gradle adds its own newline before the task output is printed, so you get two empty lines before the text is printed).

I can put it back if it's an issue.

schalkms: No, it's okay.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
```diff
@@ -29,11 +32,20 @@ open class DetektGenerateConfigTask : SourceTask() {
...
+        val config = project.files("${configDir.canonicalPath}/$CONFIG_FILE")
+
         val arguments = mutableListOf(
             GenerateConfigArgument,
+            ConfigArgument(config),
             InputArgument(source)
         )
 
-        DetektInvoker.create(project).invokeCli(arguments.toList(), detektClasspath, name)
+        if (config.first().exists()) {
```
arturbosch: Can `first()` here crash when `project.files(...)` is empty or null?

3flex: Very literally yes, it would fail when project files is empty. But that won't happen as the value is set a couple lines earlier and doesn't rely on external input.

That said I'll make a new `ConfigArgument` that accepts a single file instead of a collection to clean it up.

3flex: It was a bit awkward to deal with a new `ConfigArgument` variant, so thought it's clearer to add error handling in the task itself.

To be clear this should never happen.

PR #1814, opened 2019-08-11
Tidy the build files

3flex:
Just for fun I went through the build files to look for possible improvements - this should make them a little bit clearer and performant.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -233,32 +233,32 @@ subprojects {
             groupId = this@subprojects.group as? String
             artifactId = this@subprojects.name
             version = this@subprojects.version as? String
-            pom.withXml {
-                asNode().apply {
-                    appendNode("description", "Static code analysis for Kotlin")
-                    appendNode("name", "detekt")
-                    appendNode("url", "https://arturbosch.github.io/detekt")
-
-                    val license = appendNode("licenses").appendNode("license")
-                    license.appendNode("name", "The Apache Software License, Version 2.0")
-                    license.appendNode("url", "http://www.apache.org/licenses/LICENSE-2.0.txt")
-                    license.appendNode("distribution", "repo")
-
-                    val developer = appendNode("developers").appendNode("developer")
-                    developer.appendNode("id", "Artur Bosch")
-                    developer.appendNode("name", "Artur Bosch")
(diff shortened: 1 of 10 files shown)


build.gradle.kts
```diff
@@ -77,9 +78,7 @@ allprojects {
     version = detektVersion + if (System.getProperty("snapshot")?.toBoolean() == true) "-SNAPSHOT" else ""
 
     repositories {
-        mavenLocal()
         jcenter()
-        maven(url = "https://dl.bintray.com/arturbosch/generic")
```
schalkms: Why did you remove both `maven` instructions? Is this not necessary anymore?

3flex: The build won't succeed if repositories are required that are removed. The bintray line was there because there used to be a dependency on ksh library (from memory) which is no longer used.

mavenLocal would have supported local testing of the Gradle plugin but with composite builds that's not necessary anymore.

Currently not all dependencies are available on Maven Central otherwise would use that instead of jcenter.

build.gradle.kts
```diff
@@ -139,7 +138,7 @@ subprojects {
         }
     }
 
-    tasks.withType<Test> {
+    tasks.withType<Test>().configureEach {
```
arturbosch: This does lazy configuration?

build.gradle.kts
```diff
@@ -242,32 +231,32 @@ subprojects {
...
-                    license.appendNode("url", "http://www.apache.org/licenses/LICENSE-2.0.txt")
-                    license.appendNode("distribution", "repo")
-
-                    val developer = appendNode("developers").appendNode("developer")
-                    developer.appendNode("id", "Artur Bosch")
-                    developer.appendNode("name", "Artur Bosch")
-                    developer.appendNode("email", "arturbosch@gmx.de")
-
-                    appendNode("scm").appendNode("url", "https://github.com/arturbosch/detekt")
+            pom {
```
arturbosch: Much nicer :)

build.gradle.kts
3flex: Lazy configuration and task configuration avoidance are two different things. This is task configuration avoidance (tasks are configured on demand).

https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

PR #1817, opened 2019-08-14
JCommander 1.78

3flex:
Jcommander is now on Maven Central so we can upgrade.

https://search.maven.org/artifact/com.beust/jcommander/1.75/jar


3flex:
Strange... it builds OK and the JAR works but the Gradle task is failing. I'll try and fix locally.

Issue #1797, opened 2019-08-01
Quiet mode or ability to disable all output in Gradle


TWiStErRob:
## Expected Behavior
No output, unless there are problems, even then potentially just a build failure!
Ideal output when no issues detekted: nothing
Ideal output when there are issues:
```
> Task :module:detekt
Ruleset: exceptions - 40min debt
        TooGenericExceptionCaught - [ex] at C:\dev\project\module\src\main\kotlin\net\twisterrob\project\Pipeline.kt:154:13
        TooGenericExceptionCaught - [ex] at C:\dev\project\module\src\main\kotlin\net\twisterrob\project\Listings.kt:41:12

Ruleset: style - 1h 10min debt
        MagicNumber - [provideCache] at C:\dev\project\module\src\main\kotlin\net\twisterrob\project\network\OkHttpModule.kt:31:18
        MagicNumber - [provideCacheInterceptor] at C:\dev\project\module\src\main\kotlin\net\twisterrob\project\network\OkHttpModule.kt:78:31
        MagicNumber - [provideCacheInterceptor] at C:\dev\project\module\src\main\kotlin\net\twisterrob\project\network\OkHttpModule.kt:78:61
        MagicNumber - [callFirst] at C:\dev\project\module\src\main\kotlin\net\twisterrob\project\Listing.kt:20:38
        MagicNumber - [callFirst] at C:\dev\project\module\src\main\kotlin\net\twisterrob\project\Listing.kt:20:49
        MagicNumber - [callSecond] at C:\dev\project\module\src\main\kotlin\net\twisterrob\project\Listing.kt:21:39
        MagicNumber - [callSecond] at C:\dev\project\module\src\main\kotlin\net\twisterrob\project\Listing.kt:21:50

Overall debt: 1h 50min

Successfully generated HTML report at C:\dev\project\module\build\reports\detekt\detekt.html
Successfully generated Checkstyle XML report at C:\dev\project\module\build\reports\detekt\detekt.xml
Build succeeded with 9 weighted issues (threshold defined was 10).
```
Note:
 * no runtime of detekt (build scan is for that)
 * no number of files (it's nice to see, but not on all executions)
 * no progress output with .......... (use Gradle worker API to report progress)
 * no ruleset header when the ruleset had no violations
 * report location shown only when there are things in it
 * tech debt should be configurable
 * weighted threshold report maybe

## Current Behavior
Spamming logs even with all the config I could think of to reduce noise (and no violations):
```
console-reports:
[...]

PR #1818, opened 2019-08-14
Use notifications instead of println

arturbosch:
Closes - #1797

a config like:
```
processors:
  active: true
  exclude:
    - 'DetektProgressListener'
    - 'FunctionCountProcessor'
    - 'PropertyCountProcessor'
    - 'ClassCountProcessor'
    - 'PackageCountProcessor'
    - 'KtFileCountProcessor'

console-reports:
  active: true
  exclude:
    - 'ProjectStatisticsReport'
    - 'ComplexityReport'
    - 'NotificationReport'
    - 'FindingsReport'
    - 'BuildFailureReport'
```
won't print anything.

I will write a guide on how to turn detekt silent for your CI in a separate PR.

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FindingsReport.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FindingsReport.kt
@@ -8,26 +8,27 @@ class FindingsReport : ConsoleReport() {
     override val priority: Int = 40
 
     override fun render(detektion: Detektion): String? {
-        val findings = detektion.findings
         val totalDebt = DebtSumming()
         return with(StringBuilder()) {
-            findings.forEach { rulesetFindings ->
-                val debtSumming = DebtSumming()
-                val issuesString = rulesetFindings.value.joinToString("") {
-                    debtSumming.add(it.issue.debt)
-                    it.compact().format("\t")
-                }
-                val debt = debtSumming.calculateDebt()
-                val debtString =
+            detektion.findings
+                .filter { it.value.isNotEmpty() }
+                .forEach { (ruleSetId, issues) ->
+                    val debtSumming = DebtSumming()
+                    val issuesString = issues.joinToString("") {
+                        debtSumming.add(it.issue.debt)
+                        it.compact().format("\t")
+                    }
+                    val debt = debtSumming.calculateDebt()
+                    val debtString =
                         if (debt != null) {
                             totalDebt.add(debt)
                             " - $debt debt".format()
                         } else {
                             "\n"
                         }
-                append(rulesetFindings.key.format(prefix = "Ruleset: ", suffix = debtString))
-                append(issuesString.yellow())
-            }
+                    append(ruleSetId.format(prefix = "Ruleset: ", suffix = debtString))
+                    append(issuesString.yellow())
+                }
             val debt = totalDebt.calculateDebt()
             if (debt != null) {
                 append("Overall debt: $debt".format("\n"))
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/NotificationReport.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/NotificationReport.kt
@@ -5,10 +5,10 @@ import io.gitlab.arturbosch.detekt.api.Detektion
 
 class NotificationReport : ConsoleReport() {
 
-    override val priority: Int = 50
+    // Print notifications before the build failure report but after all other reports.
+    // This allows to compute intermediate messages based on detekt results and do not rely on 'println'.
+    override val priority: Int = Int.MIN_VALUE + 1
 
-    override fun render(detektion: Detektion): String? {
-        val notifications = detektion.notifications
-        return notifications.joinToString("\n") { it.message }
-    }
+    override fun render(detektion: Detektion): String? =
+        detektion.notifications.joinToString("\n") { it.message }
 }
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/SingleRuleRunner.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/SingleRuleRunner.kt
@@ -37,6 +37,8 @@ class SingleRuleRunner(private val arguments: CliArgs) : Executable {
             ?: throw IllegalArgumentException("There was no rule set with id '$ruleSet'.")
 
         val provider = RuleProducingProvider(rule, realProvider)
+        assert(provider.instance(settings.config).rules.size == 1) { "Eager testing if rule exists failed." }
+
         val detektion = DetektFacade.create(
             settings,
             listOf(provider),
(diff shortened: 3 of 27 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/NotificationReport.kt
```diff
@@ -5,10 +5,10 @@ import io.gitlab.arturbosch.detekt.api.Detektion
 
 class NotificationReport : ConsoleReport() {
 
-    override val priority: Int = 50
+    // Print notifications before the build failure report but after all other reports.
+    // This allows to compute intermediate messages based on detekt results and do not rely on 'println'.
+    override val priority: Int = Int.MIN_VALUE + 1
```
schalkms: Sounds like **"magic"** to me.

detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -9,6 +9,7 @@ build:
 processors:
   active: true
   exclude:
+  # - 'DetektProgressListener'
```
schalkms: 👍

schalkms:
Codecov is back once again.  
I'd be interested on the circumstances. When does this integration work? To me it seems to be rather random. 🔢
I investigated this in #1810 and #1811. From the CI log it seems to me that the report gets generated successfully but it's not uploaded correctly to codecov.io.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FindingsReport.kt
```diff
@@ -8,26 +8,27 @@ class FindingsReport : ConsoleReport() {
...
-            findings.forEach { rulesetFindings ->
-                val debtSumming = DebtSumming()
-                val issuesString = rulesetFindings.value.joinToString("") {
-                    debtSumming.add(it.issue.debt)
-                    it.compact().format("\t")
-                }
-                val debt = debtSumming.calculateDebt()
-                val debtString =
+            detektion.findings
+                .filter { it.value.isNotEmpty() }
```
schalkms: I think we need to add tests here @arturbosch

arturbosch:
I've extracted the BuildFailure logic to a common file, the `BuildFailureReport` now uses it and in a much cleaner way and the runner is now responsible for throwing that exception.

Please look again at the PR.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/RunnerSpec.kt
```diff
@@ -0,0 +1,56 @@
...
+            val tmpReport = Files.createTempFile("RunnerSpec", ".txt")
+            val cliArgs = CliArgs.parse(arrayOf(
+                "--input", inputPath.toString(),
+                "--report", "txt:$tmpReport",
+                "--config-resource", "/configs/max-issues--1.yml"
+            ))
+
+            runCatching { Runner(cliArgs).execute() }
+                .onSuccess { assertThat(Files.readAllLines(tmpReport)).hasSize(1) }
+                .onFailure { fail("should not fail with a BuildFailure exception") }
```
schalkms: Weird test construct

schalkms: Does it test a successful or a failed run?

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/TestDetektion.kt
```diff
@@ -5,16 +5,21 @@ import io.gitlab.arturbosch.detekt.api.Finding
...
 import io.gitlab.arturbosch.detekt.api.ProjectMetric
 import org.jetbrains.kotlin.com.intellij.openapi.util.Key
+import org.jetbrains.kotlin.com.intellij.util.keyFMap.KeyFMap
 
 open class TestDetektion(vararg findings: Finding) : Detektion {
 
     override val metrics: Collection<ProjectMetric> = listOf()
     override val findings: Map<String, List<Finding>> = findings.groupBy { it.id }
     override val notifications: List<Notification> = listOf()
+    private var userData = KeyFMap.EMPTY_MAP
```
schalkms: What's that variable for?

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/OutputFacadeSpec.kt
```diff
@@ -50,40 +48,10 @@ internal class OutputFacadeSpec : Spek({
                 outputStream.assertThatItPrintsReportPath(HtmlOutputReport().name)
             }
         }
-
-        describe("with more findings than issues allowed") {
-            it("reports build failure for default task") {
```
schalkms: Why do we remove this test?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/SingleRuleRunner.kt
```diff
@@ -37,6 +37,8 @@ class SingleRuleRunner(private val arguments: CliArgs) : Executable {
             ?: throw IllegalArgumentException("There was no rule set with id '$ruleSet'.")
 
         val provider = RuleProducingProvider(rule, realProvider)
+        assert(provider.instance(settings.config).rules.size == 1) { "Eager testing if rule exists failed." }
```
schalkms: Sounds a bit weird. Why do we need this check here?

arturbosch:
Thanks for the feedback, I will split the changes into two PRs.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/SingleRuleRunner.kt
arturbosch: I tried to extract it to an own method but faild. Now I have an idea: `assertRuleExists` should be the name.
This verifies the given `ruleSet:ruleName` exist. When we do not assert it here, it will fail when detekt will run inside a thread pool providing strange stacktraces ^^.
This should eagerly figures out if the rule exists before trying to run and fail late.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/OutputFacadeSpec.kt
arturbosch: Because not the BuildFailureReport maanges the build failure but the `Runner` class itself.
The new `RunnerSpec` now tests this behavior.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/TestDetektion.kt
arturbosch: Like the real implementation the test implementation now can store arbitrary data. Like the processors which store metrics inside such an `userData`, we can store any data inside and cache it.
In this commit I additionally store `Detektion.getOrComputeWeightedAmountOfIssues`.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/RunnerSpec.kt
arturbosch: Your right, the catch and failure block is not necessary for this test.

Issue #1828, opened 2019-08-19
Consider a "file per test case" testing structure


patrickpilch:
In the course of developing a rule augmentation, I struggled with writing tests to my intention. I wanted to detail some of the difficulties I've faced along with my opinions, and propose a new test case structure. I'm certain there are good reasons for the existing structure that I do not have insight into due to my lack of the project's historical context, but I'm hoping that my suggestions elicit improvements that address most of them.

The current structure of maintaining one file for Positive & Negative cases respectively makes it more challenging (IMO) to write a test case that is maintainable over the longer course of an actively developed rule. My expectation of a rule test case is that it ensures a _particular message_ is reported for a _particular unit of code._ Information such as class, line, containing method, etc. are cross-cutting concerns and are out of the scope of a rule test (and instead would live within the issue reporting functionality test cases.)

So far, I see a few approaches that I consider not ideal in that regard:
### Multiple test cases in one `Case`
Some test cases run `lint` over the entirety of the Positive file, and assert the expected quantity of issues reported. This has the disadvantages of test failure traceability. Consider the scenario where I've modified a rule, and the 4th out of the 10 total positive cases is now incorrectly not reported. The test output indicates an expectation of "10" but is now "9", and spews out a list of 9 `Issue` string representations containing long file paths, line numbers, etc. that you now have to sort out. 

Placing all the cases into one file also increases the difficulty of asserting error messaging. Consider the example above, but now with also having assertions specifying that issue 4/10 should have a distinct message from the others. The reporting will now be off by one position. 
This issue will also affect you when test cases are _added_ in between existing ones (for organization's sake). Using line numbers in the assertions for a specific rule test faces all of the same issues, where merely adding a line to a large Case will result in tons of assertion changes (feels like a violation of the open/closed principle to me.)

### The `Case` enum
This enum provides a mapping of test cases to file paths along with some utility functions. But because it's an enum, it dictates a strictly one-level hierarchy in test case organization. This means that if I wanted to maintain multiple test case files, they would all have to be in the same directory and prefixed so they are grouped with related cases. This can result in some lengthy names like `SomeRuleSomeQualifierSomeSpecificInstancePositive`. 

## Proposal
I propose to use the Java packaging system as the organizational system for the `Rule`s and their test cases. If each rule has its own package, there will be a corresponding package inside the test `resources` folder that can contain multiple files and a directory structure. I think this gives the test designer more flexibility in how to group, organize and write the tests.

[...]

PR #1830, opened 2019-08-19
Migrate ComplexInterface test cases to JSR223

schalkms:
ref #1828


arturbosch:
Unfortunately we can't deactivate it there. It just tells us that it uses our config/detekt/config.yml file...
Idk what is happening there and why it is not the same as our CI detekt findings ...

PR #1833, opened 2019-08-20
Gradle 5.6

3flex:
Takes advantage of a new feature in Gradle 5.6 so all dependency versions can be declared in the same file (though not for the Gradle plugin, which would require copying the gradle.properties file around).

Also explicitly sets the DuplicateStrategy for the formatting jar creation to `INCLUDE` which is the default in Gradle 5.6 but is changing to `FAIL` in Gradle 6. This works around that issue and makes the desired behaviour explicit.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -33,6 +33,7 @@ import org.gradle.api.provider.Provider
 import org.gradle.api.reporting.ReportingExtension
 import org.gradle.api.tasks.CacheableTask
 import org.gradle.api.tasks.Classpath
+import org.gradle.api.tasks.Console
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFile
 import org.gradle.api.tasks.InputFiles
@@ -60,7 +61,7 @@ open class Detekt : SourceTask(), VerificationTask {
     @Input
     @Optional
     @Deprecated("Replace with setIncludes/setExcludes")
-    var filters: Property<String> = project.objects.property(String::class.java)
+    val filters: Property<String> = project.objects.property(String::class.java)
 
     @Classpath
     val detektClasspath = project.configurableFileCollection()
@@ -71,12 +72,12 @@ open class Detekt : SourceTask(), VerificationTask {
     @InputFile
     @Optional
     @PathSensitive(PathSensitivity.RELATIVE)
-    var baseline: RegularFileProperty = project.fileProperty()
+    val baseline: RegularFileProperty = project.fileProperty()
 
     @InputFiles
     @Optional
     @PathSensitive(PathSensitivity.RELATIVE)
-    var config: ConfigurableFileCollection = project.configurableFileCollection()
+    val config: ConfigurableFileCollection = project.configurableFileCollection()
 
     @Classpath
     @Optional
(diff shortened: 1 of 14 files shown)


3flex:
Gradle 5.6 has new validations on task property annotations. Fixing them means some changes in build files for our users (specifically for the `config` parameter which would be customised by many users).

Technically this means a major version bump. So perhaps we will stay on Gradle 5.4 to avoid triggering issues in validateTaskProperties? Or we can turn validateTaskProperties into a warning only for Gradle 5.6 then when we do a breaking change for Gradle 6 turn the validations back on?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -104,21 +105,19 @@ open class Detekt : SourceTask(), VerificationTask {
...
-    var plugins: Property<String> = project.objects.property(String::class.java)
+    val plugins: Property<String> = project.objects.property(String::class.java)
 
     @get:Internal
-    @get:Optional
     internal val debugProp: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
     var debug: Boolean
-        @Internal
+        @Input
         get() = debugProp.get()
```
marschwar: Why is it marked as @Input? Should it not be @Console?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -104,21 +105,19 @@ open class Detekt : SourceTask(), VerificationTask {
...
         set(value) = debugProp.set(value)
 
     @get:Internal
-    @get:Optional
     internal val parallelProp: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
     var parallel: Boolean
-        @Internal
+        @Console
         get() = parallelProp.get()
```
marschwar: > Attached to a task property to indicate that the property is not to be taken into account for up-to-date checking, because its value only influences the console output of the task.

How does the parallel prop influence the console output? Why is `@Internal` not correct?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektIdeaFormatTask.kt
```diff
@@ -25,7 +25,7 @@ open class DetektIdeaFormatTask : SourceTask() {
 
     @Internal
     @Optional
-    var debug: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
+    val debug: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
```
marschwar: `@Console`?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
3flex: My thinking was that if a user is changing value of debug parameter it's because they want to see debug output on next run. If it's Console the change in value would be ignored if the previous run was successful.

3flex:
This isn't going well! :/

There's an error on JDK 12 on Linux which makes no sense, and it's repeatable and happened twice in a row. It's a possible regression in Gradle but I won't know until I try and replicate in a smaller project.

The error happens when running the tests on Gradle 4.9:

```
    org.gradle.testkit.runner.InvalidRunnerConfigurationException: The build could not be executed due to a feature not being supported by the target Gradle version
        Caused by:
        org.gradle.tooling.UnsupportedVersionException: The version of Gradle you are using (Gradle distribution 'https://services.gradle.org/distributions/gradle-4.9-bin.zip') does not support the ModelBuilder API. Support for this is available in Gradle 1.2 and all later versions.
```

It's effectively saying that a feature available in all Gradle releases from 1.2 is not available in 4.9 which makes no sense.

schalkms:
@3flex okay, that’s not good. Did you report this to the Gradle guys?

settings.gradle.kts
```diff
@@ -9,3 +9,25 @@ include("detekt-api",
...
+val buildScanVersion: String by settings
+val dokkaVersion: String by settings
+val gradleVersionsPluginVersion: String by settings
+val kotlinVersion: String by settings
+val shadowVersion: String by settings
+val sonarQubeVersion: String by settings
+
+pluginManagement {
+        plugins {
+                id("com.jfrog.artifactory") version artifactoryVersion
```
arturbosch: Cool stuff!

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/MultiVersionTest.kt
```diff
@@ -9,7 +9,7 @@ import org.spekframework.spek2.style.specification.describe
 
 object MultiVersionTest : Spek({
 
-    val testedGradleVersions = listOf("4.9", "5.4")
+    val testedGradleVersions = listOf("4.9", "5.6")
```
schalkms: Shouldn't we keep this regression test for version **5.4**?

3flex: When the multi version test was implemented decision was to test on oldest and most recent supported version of Gradle that detekt supports. So the most recent version is bumped when Gradle is updated.

PR #1840, opened 2019-08-22
Do not report auto-corrected formatting issues

### LINKED ISSUE #1836: Check fails even if all issues got auto fixed
## Expected Behavior
If all found issues where automatically fixed the check/build should not fail.

## Observed Behavior
The check/build fails even if after auto fixing no issues exists.
If I run detekt a second time it passes.

## Steps to Reproduce
```
build:
  maxIssues: 1
formatting:
  active: true
  android: true
  autoCorrect: true
```
Introduce an issue which is auto correctable.
Run the detekt gradle task.

## Context
I want to use the same config/task to auto format/fix and check the result
(locally and on the ci server)

## Your Environment
* Version of detekt used: 1.0.0 (also tested with 1.0.0-RC15)
* Version of Gradle used (if applicable): Gradle 5.1.1
* Operating System and version: macOS Mojave

schalkms:
All auto-corrected issues do not add to the total issues count.
Closes #1836

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/BuildFailure.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/BuildFailure.kt
@@ -47,13 +47,18 @@ fun Detektion.getOrComputeWeightedAmountOfIssues(config: Config): Int {
 }
 
 fun Detektion.filterAutoCorrectedIssues(config: Config): Map<RuleSetId, List<Finding>> {
-    if (config.excludeCorrectable()) {
-        return findings.filter {
-            val correctableCodeSmell = it as? CorrectableCodeSmell
+    if (!config.excludeCorrectable()) {
+        return findings
+    }
+    val filteredFindings = HashMap<RuleSetId, List<Finding>>()
+    findings.forEach { (ruleSetId, findingsList) ->
+        val newFindingsList = findingsList.filter { finding ->
+            val correctableCodeSmell = finding as? CorrectableCodeSmell
             correctableCodeSmell == null || !correctableCodeSmell.autoCorrectEnabled
         }
+        filteredFindings[ruleSetId] = newFindingsList
     }
-    return findings
+    return filteredFindings
 }
 
 private fun Config.weightsConfig(): Config = subConfig(BUILD).subConfig(WEIGHTS)
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/BuildFailure.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/BuildFailure.kt
@@ -14,11 +14,11 @@ private val WEIGHTED_ISSUES_COUNT_KEY = Key.create<Int>("WEIGHTED_ISSUES_COUNT")
 private const val BUILD = "build"
 private const val WEIGHTS = "weights"
 private const val MAX_ISSUES = "maxIssues"
-private const val EXCLUDE_Correctable = "excludeCorrectable"
+private const val EXCLUDE_CORRECTABLE = "excludeCorrectable"
 
 fun Config.maxIssues(): Int = subConfig(BUILD).valueOrDefault(MAX_ISSUES, -1)
 
-fun Config.excludeCorrectable(): Boolean = subConfig(BUILD).valueOrDefault(EXCLUDE_Correctable, false)
+fun Config.excludeCorrectable(): Boolean = subConfig(BUILD).valueOrDefault(EXCLUDE_CORRECTABLE, false)
 
 fun Int.isValidAndSmallerOrEqual(amount: Int): Boolean =
     !(this == 0 && amount == 0) && this != -1 && this <= amount
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/BuildFailure.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/BuildFailure.kt
@@ -13,8 +13,11 @@ private val WEIGHTED_ISSUES_COUNT_KEY = Key.create<Int>("WEIGHTED_ISSUES_COUNT")
 private const val BUILD = "build"
 private const val WEIGHTS = "weights"
 private const val MAX_ISSUES = "maxIssues"
+private const val EXCLUDE_Correctable = "excludeCorrectable"
 
 fun Config.maxIssues(): Int = subConfig(BUILD).valueOrDefault(MAX_ISSUES, -1)
+
+fun Config.excludeCorrectable(): Boolean = subConfig(BUILD).valueOrDefault(EXCLUDE_Correctable, false)
 
 fun Int.isValidAndSmallerOrEqual(amount: Int): Boolean =
     !(this == 0 && amount == 0) && this != -1 && this <= amount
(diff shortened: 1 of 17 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
```diff
@@ -39,14 +40,21 @@ class DetektFacade(
...
             KtFileModifier().saveModifiedFiles(filesToAnalyze) {
                 notifications.add(it)
             }
         }
 
+        for ((key, value) in findings) {
+            findings[key] = value.filter {
+                val correctableCodeSmell = it as? CorrectableCodeSmell
+                correctableCodeSmell == null || !correctableCodeSmell.autoCorrectEnabled
+            }
```
schalkms: Maybe there's an easier solution for that.

schalkms: Some tests are still missing

arturbosch: Consider extracting these two files into a function. This would make the code more readable.

arturbosch:
So basically we will just print the notifications?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
schalkms: Yes, definitely! Is there a better perhaps more performant way of adding this check?

schalkms:
Should we also exclude it from the notifications part?

jonas-arkulpa:
I would argue that if it got automatically fixed I do not care about any notification / report entry.
But there might be a use case I am currently not aware of.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
arturbosch: Actually I do not like this filtering inside the core module.
Imo detekt should always report everything it finds. The reports or cli baseline are only allowed to filter the violations.

If we leave it that way some users may be interested in this notifications because detekt would be `too magic`. They could not turn them on via the report or config option.
So this should go into the `FindingsReport` then. Users interested in formatting notification can write a new report for this. 
What do you think?

schalkms: Yes, I agree. I'll try to refactor this PR in the upcoming days.

> Users interested in formatting notification can write a new report for this.

This would be a good example of 3rd party report. I think this shouldn't go into detekt at all.

schalkms:
@arturbosch I think we need to be careful here. I have two open questions:

1. If we filter the corrected formatting issues in the `FindingsReport`, it will still count to the **maxIssues** count.
2. What happens when the issues were corrected during the CI process but not locally on the developer machine, because the developer forgot to run `gradle build`. Files are reported and corrected on the CI. The commit that has been made on the developer machine doesn't include the autocorrected files. Since detekt doesn't report the corrected files there's divergence between the CI and dev machine behavior.

arturbosch:
1. This sounds like a +1 for the `excludeCorrectable` property inside `build`. That way we have a generic way to handle such a case in all extensions.
2. Normally the CI should never auto correct anything, for this the user can use `autoCorrect = $System.property.IS_CI ? false : true`. So this will need further documentation.

Does this answer your concerns?

schalkms:
@arturbosch please consider this commit as a property. This needs of course further refactoring.
Now the autocorrected issues are still counted in the `BuildFailureReport`. Hence, the build will fail, because the threshold is exceeded.

arturbosch:
> @arturbosch please consider this commit as a property. This needs of course further refactoring.
> Now the autocorrected issues are still counted in the `BuildFailureReport`. Hence, the build will fail, because the threshold is exceeded.

As a property? Sorry for not understanding your sentence. The commit looks great. A property is the right way to go

schalkms:
I meant prototype instead of property. The mentioned behavior still feels weird. Should we also exclude the fixed formatting issues from the `BuildFailureReport`?

arturbosch:
Yes this property should be used in BuildFailureReport/Runner.

schalkms:
@arturbosch I think that this in a mergeable state now. Could you please review this once again.
@sanogueralorenzo @jonas-arkulpa could you please test this out in your projects?

We should include this in v1.2.1

jonas-arkulpa:
@schalkms I can have a look next week. Are there instructions to use the detekt version of this PR?

schalkms:
@jonas-arkulpa Thanks for your help! 
- Check out this branch
- `cd detekt`
- `./gradlew build shadowJar`
- `java -jar detekt-cli/build/libs/detekt-cli-[version]-all.jar`

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/BuildFailure.kt
```diff
@@ -13,9 +14,12 @@ private val WEIGHTED_ISSUES_COUNT_KEY = Key.create<Int>("WEIGHTED_ISSUES_COUNT")
 private const val BUILD = "build"
 private const val WEIGHTS = "weights"
 private const val MAX_ISSUES = "maxIssues"
+private const val EXCLUDE_CORRECTABLE = "excludeCorrectable"
```
schalkms: This property needs to be added somewhere to the `detekt-config.yml` file.

detekt-formatting/src/test/kotlin/io/gitlab/arturbosch/detekt/formatting/AutoCorrectLevelSpec.kt
```diff
@@ -30,34 +30,6 @@ class AutoCorrectLevelSpec : Spek({
             }
         }
 
-        describe("autoCorrect: false on top level") {
-
-            val config = yamlConfig("/autocorrect/autocorrect-toplevel-false.yml")
-
-            it("should format the test file but not print to disc") {
```
arturbosch: Mh, I do not understand this removal. We do not have any test case that tests if actually a file was formatted after running detekt with autoCorrect?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/BuildFailure.kt
arturbosch: This should go into the ConfigPrinter -> https://github.com/arturbosch/detekt/blob/6c3795554f24377920664bb08833202679c8a3d4/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/ConfigPrinter.kt#L68

schalkms: Addressed this in the second to last commit. Furthermore, I rebased my branch.
I'm sorry, that this PR took me so long.

detekt-formatting/src/test/kotlin/io/gitlab/arturbosch/detekt/formatting/AutoCorrectLevelSpec.kt
schalkms: Agreed. I addressed this in the last commit. This is good to go for v1.3.1.

jonas-arkulpa:
Works like charm in the new version.

@schalkms sry I tried to test it last year and had some issues and limited time.

PR #1847, opened 2019-08-24
Gradle 5.5

3flex:
WIP - seeing if this works OK with JDK 12 on Linux on Travis

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -28,7 +28,7 @@ buildScan {
 }
 
 tasks.wrapper {
-    gradleVersion = "5.4"
+    gradleVersion = "5.5"
     distributionType = Wrapper.DistributionType.ALL
     doLast {
         /*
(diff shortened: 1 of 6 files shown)


arturbosch:
Seems like master build is now failing with Gradle 5.4 and Jdk12 with the same exception ...
```
io.gitlab.[secure].detekt.MultiVersionTest > runs on version 4.9 of Gradle FAILED

    org.gradle.testkit.runner.InvalidRunnerConfigurationException: The build could not be executed due to a feature not being supported by the target Gradle version

        Caused by:

        org.gradle.tooling.UnsupportedVersionException: The version of Gradle you are using (Gradle distribution 'https://services.gradle.org/distributions/gradle-4.9-bin.zip') does not support the ModelBuilder API. Support for this is available in Gradle 1.2 and all later versions.
```

https://travis-ci.org/arturbosch/detekt/builds/576185125

3flex:
It's very frustrating.

I've tried replicating locally with Ubuntu 18.04 and using the same release as is being used in Travis but it builds successfully. I'll close this then since if it's happening w/ Gradle 5.4 then this PR won't tell us anything.

PR #1848, opened 2019-08-24
Use Ubuntu 18.04 for JDK 11 & 12

3flex:
JDK 8 is not supported on Ubuntu 18.04 so remains on 16.04 (Xenial)

Seems to address issue on JDK 12 that was causing some previous builds to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,13 @@ language: java
 matrix:
   include:
     - os: linux
+      dist: xenial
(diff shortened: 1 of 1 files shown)


schalkms:
Thanks for investigating this strange error.

arturbosch:
Fortunately this was "just" a CI error and the release is not corrupted! Thanks!

PR #1850, opened 2019-08-25
Remove unnecessary .trimIndent() in tests

schalkms:
The .trimIndent() function is called in `BaseRule.lint()` and
`Base.compileAndLint()` anyways. Therefore, there's no need to call
.trimIndent() when creating test snippets.


3flex:
There's minor usage of `trimMargin()` too, might as well drop those at the same time? Would have to remove the `|` at the start of each quoted line.

Issue #1355, opened 2018-11-21
Grouped output


realdadfish:
Continued from https://github.com/arturbosch/detekt/issues/171 I'd love to see Detekt get some console output formatting options. Right now each executed rule lists all the violations in each file, one by one.

Now, if more than one violation in more than one file occurs, I then have to hop in between to fix all the issues in all files. So it would be super-awesome to have an option to group the output by file, and not only by violation.

arturbosch:
We could provide an additional `ConsoleReport` which sorts the findings by ruleset `and` file :).
See `FindingsReport`.

sowmyav24:
Couple of open questions for this issue is put forth in [#1852](https://github.com/arturbosch/detekt/pull/1852#issue-311662150)

sowmyav24:
We can close this issue . Merged in [#1852 ](https://github.com/arturbosch/detekt/pull/1852)

arturbosch:
Closing, there is a RuleSetBased`FindingsReport` and a `FileBasedFindingsReport`.

PR #1852, opened 2019-08-28
Group console violation reports by file

sowmyav24:
This PR addresses [#1355](https://github.com/arturbosch/detekt/issues/1355)

Open questions:
- Should we make type of report to be displayed configurable to the user ?
- The output format would be similar to the existing report and all the violations of a file appear together. Do we need to make any changes to the output format ?


sowmyav24:
Thank you for the suggestion.
Will update the PR accordingly.

sowmyav24:
> we could provide a selection of FindingsRepots and let the user choose which to use in the config file.

In the updated PR, it is a boolean value. Would it be better if it is changed it to `Enum` as it will allow the users to choose multiple console reports or would a boolean value suffice ?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/DebtSumming.kt
```diff
@@ -17,6 +19,34 @@ class DebtSumming {
         return calculate()
     }
 
+    fun calculateDebtSum(
+        issues: Map<RuleSetId, List<Finding>>,
+        totalDebt: DebtSumming
+    ): String? {
```
schalkms: Can we please put this method into a corresponding format class called `DebtSummingPrinter` and give it a better name. Only calculations are done in the `DebtSumming` class. The `calculateDebtSum()` function rather outputs text to the console. It doesn't really calculate that much.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReportSpec.kt
```diff
@@ -0,0 +1,57 @@
...
+                            createFinding(fileName = "TestFile.kt"),
+                            createFinding(fileName = "AnotherFile.kt")
+                        )
+                    ),
+                    Pair("EmptySmells", emptyList())
+                )
+            }
+            val output = subject.render(detektion)?.trimEnd()?.decolorized()
+
+            it("has the reference content") {
```
schalkms: We should simplify this test or create new ones.
The corresponding `grouped-findings-report.txt` file looks quite big.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/TestFactory.kt
```diff
@@ -12,12 +12,12 @@ import io.gitlab.arturbosch.detekt.core.ModificationNotification
 import io.gitlab.arturbosch.detekt.test.resource
 import java.nio.file.Paths
 
-fun createFinding(ruleSet: String = "TestSmell") = CodeSmell(createIssue(ruleSet), createEntity(), "TestMessage")
+fun createFinding(ruleSet: String = "TestSmell", fileName: String = "TestFile.kt") = CodeSmell(createIssue(ruleSet), createEntity(fileName), "TestMessage")
 
 fun createIssue(id: String = "TestSmell") = Issue(id, Severity.CodeSmell, "For Test", Debt.FIVE_MINS)
 
-fun createEntity() = Entity("TestEntity", "TestEntity", "S1", createLocation())
+fun createEntity(fileName: String) = Entity("TestEntity", "TestEntity", "S1", createLocation(fileName))
```
schalkms: The filename should correspond to the entity name, I guess.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/TestFactory.kt
```diff
@@ -12,12 +12,12 @@ import io.gitlab.arturbosch.detekt.core.ModificationNotification
 import io.gitlab.arturbosch.detekt.test.resource
 import java.nio.file.Paths
 
-fun createFinding(ruleSet: String = "TestSmell") = CodeSmell(createIssue(ruleSet), createEntity(), "TestMessage")
+fun createFinding(ruleSet: String = "TestSmell", fileName: String = "TestFile.kt") = CodeSmell(createIssue(ruleSet), createEntity(fileName), "TestMessage")
```
schalkms: The filename should correspond to the entity name, I guess.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReport.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    override val priority: Int = 40
+
+    override fun render(detektion: Detektion): String? {
+        val totalDebt = DebtSumming()
+        return with(StringBuilder()) {
+            val ruleSetMap = detektion.findings.filter { it.value.isNotEmpty() }
+            val distinctFileNames = ruleSetMap.values.flatten().map { it.entity.location.file }.distinct()
+            distinctFileNames
+                .forEach { fn ->
```
schalkms: `filename` instead of `fn` makes it more readable.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReport.kt
```diff
@@ -0,0 +1,30 @@
...
+                    val newRuleSetMap = HashMap<RuleSetId, List<Finding>>()
+                    ruleSetMap.forEach { (k, v) -> newRuleSetMap[k] = v.filter { it.entity.location.file == fn } }
```
schalkms: `key, value` instead of `k, v` makes it more readable.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FindingsReport.kt
```diff
@@ -10,25 +10,7 @@ class FindingsReport : ConsoleReport() {
...
-                        if (debt != null) {
-                            totalDebt.add(debt)
-                            " - $debt debt".format()
-                        } else {
-                            "\n"
-                        }
-                    append(ruleSetId.format(prefix = "Ruleset: ", suffix = debtString))
-                    append(issuesString.yellow())
-                }
+            append(totalDebt.calculateDebtSum(detektion.findings, totalDebt))
```
schalkms: This line is quite hard to understand. I think this gets better if we introduce the `DebtSummingPrinter` class.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReport.kt
```diff
@@ -0,0 +1,30 @@
...
+                    append(totalDebt.calculateDebtSum(newRuleSetMap, totalDebt))
```
schalkms: This line is quite hard to understand. I think this gets better if we introduce the `DebtSummingPrinter` class.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/DebtSumming.kt
sowmyav24: Moved it to `DebtSummingPrinter ` and renamed the method.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReportSpec.kt
sowmyav24: Since, we wanted to check `rule` and `file` based  combination, I had the test case containing multiple rules and multiple files.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/TestFactory.kt
sowmyav24: The `fileName` is used in `createLocation` and the same value is passed through findings. Hence the variable name `FileName`.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReport.kt
sowmyav24: Done !

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReport.kt
sowmyav24: Done !

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FindingsReport.kt
sowmyav24: Yes, Done !

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReport.kt
sowmyav24: Moved printing logic for reports to `DebtSummingPrinter `

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/TestFactory.kt
sowmyav24: The `fileName` is used in `createLocation` and the same value is passed through findings. Hence the variable name `FileName`.

sowmyav24:
> The report `FileBasedFindingsReport` should be added to the `default-detekt-config.yml file`.
> Perhaps a comment should be included, which states that only one of the reports can be activated at the same time (either `FindingsReport` or `FileBasedFindingsReport`)
> https://github.com/arturbosch/detekt/blob/ed77f0c9ebff73a50348c76782e6dcb1bd7d7b08/detekt-cli/src/main/resources/default-detekt-config.yml#L21-L25


When the `FileBasedFindingsReport` is added in `default-detekt-config.yml ` , the build fails.
In addition, added the `FileBasedFindingsReport` in `io.gitlab.arturbosch.detekt.api.ConsoleReport` as it is used by Service loader. Please let me know if I am mistaken here.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/DebtSummingPrinter.kt
```diff
@@ -0,0 +1,35 @@
...
+
+class DebtSummingPrinter {
+
+    fun printDebtInformation(
+        issues: Map<RuleSetId, List<Finding>>,
+        totalDebt: DebtSumming
+    ): String? {
+        with(StringBuilder()) {
+            issues
+                .filter { it.value.isNotEmpty() }
```
schalkms: We will run into merging issues here, because there's a bug fix in #1860.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReport.kt
```diff
@@ -0,0 +1,33 @@
...
+import io.gitlab.arturbosch.detekt.api.FileBasedConsoleReport
+import io.gitlab.arturbosch.detekt.api.Finding
+import io.gitlab.arturbosch.detekt.api.RuleSetId
+
+class FileBasedFindingsReport : FileBasedConsoleReport() {
+
+    override val priority: Int = 40
+
+    override fun render(detektion: Detektion): String? {
+        val totalDebt = DebtSumming()
```
schalkms: Be aware that we should `return null`, like we did in #1860, if there are no findings. Furthermore, there should be at least 1 test which asserts this behavior.
```Kotlin
val findings = detektion
             .findings
             .filter { it.value.isNotEmpty() }
if (findings.isEmpty()) {
  return null
}
```

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReportSpec.kt
schalkms: Nevertheless, this could be simplified and test only one thing at the same time.
Looking at the `grouped-findings-report.txt` the second time, I'm not even sure if this is the intent of the original poster.
Should we group the smells per file across all rulesets and leave out the **Ruleset: style** line?
CC @arturbosch

sowmyav24: Have reduced the additional ruleset set up. Now it has two ruleset violations and two files and assertion on that.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/FileBasedConsoleReport.kt
```diff
@@ -0,0 +1,9 @@
+package io.gitlab.arturbosch.detekt.api
+
+/**
+ * [FileBasedConsoleReport] is an alternate for 'FindingsReport'.
```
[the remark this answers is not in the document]
sowmyav24: Fixed it !

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/FileBasedConsoleReport.kt
```diff
@@ -0,0 +1,9 @@
...
+ * `consoleGroupByRule` should be configured as false
+ * in the 'console-reports' property of a detekt yaml config.
+ * else `FindingsReport` is the default console report
```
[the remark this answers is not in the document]
sowmyav24: Fixed it !

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/DebtSummingPrinter.kt
sowmyav24: Have filtered empty values in this method also

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReport.kt
sowmyav24: Done !

sowmyav24:
> ```
> detekt-cli/src/main/resources/META-INF/services/io.gitlab.arturbosch.detekt.api.ConsoleReport
> ```
Yes , this has been done. 
But i am unable to add `FileBasedFindingsReport ` to `default-detekt-config.yml`. This gets reset when build is run and if added, the CI fails.

schalkms:
CI fails because of other issues. (not related to this PR)

sowmyav24:
Ah okay, but the exclusion gets removed once the build runs in local. I have removed the exclusion from `default-detekt-config.yml` for the time being.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/DebtSummingPrinter.kt
sowmyav24: Resolved the merge conflicts

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FindingsReport.kt
```diff
@@ -8,37 +8,17 @@ class FindingsReport : ConsoleReport() {
     override val priority: Int = 40
 
     override fun render(detektion: Detektion): String? {
-        val findings = detektion
-            .findings
-            .filter { it.value.isNotEmpty() }
-        if (findings.isEmpty()) {
-            return null
-        }
```
schalkms: Please keep this code from L11-16. Please also copy it to the `FileBasedFindingsReport` and remove it from the `printDebtInformation()` function. Thus, the program doesn't need to create objects and call the `printDebtInformation()` function, when there are no findings.
For now, this duplication is fine.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReport.kt
schalkms: Thanks!

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/DebtSummingPrinter.kt
```diff
@@ -0,0 +1,36 @@
...
+                .also { if (it.isEmpty()) return null }
```
schalkms: Remove this check based on my other review comment.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/DebtSummingPrinter.kt
```diff
@@ -0,0 +1,35 @@
+package io.gitlab.arturbosch.detekt.cli.console
+
+import io.gitlab.arturbosch.detekt.api.Finding
+import io.gitlab.arturbosch.detekt.api.RuleSetId
+
+class DebtSummingPrinter {
+
+    fun printDebtInformation(
```
arturbosch: Seems like this could just be a function?

sowmyav24: > Sorry for the late response!
> I'm not sure why we need an additional flag and a new base class for `FileBasedFindingsReport`?
> I expect this new kind of report to be just a `ConsoleReport` which does the new ordering. It is excluded by default.
> Moreover I see the need to refactor the reports part in the config to have `includes` and `excludes` because now if a report is not excluded it means it is included.
> @schalkms @sowmyav24 please elaborate why we need a new flag for it.


If I am to understand correctly, if the additional flag is not there, there would be two console reports, one is the existing report, one is the re-ordered report and both  the reports would contain the same information but in a different format.

detekt-cli/src/test/resources/grouped-findings-report.txt
```diff
@@ -0,0 +1,11 @@
+Ruleset: AnotherTestSmell - 5min debt
+	TestSmell - [TestEntity] at AnotherFile.kt:1:1
```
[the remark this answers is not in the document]
arturbosch: Yeah, there should't be a case with two same rule names and different rule sets.
Strg+F should find the right rule and mentioning the rule set should not be needed.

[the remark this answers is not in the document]
sowmyav24: Yes, grouped by File name would be apt. Have removed the ruleset also in the output.
If ever we have two rules named similar in the different rulesets, we can add it back.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/TestFactory.kt
arturbosch: Shouldn't this be `ruleName`?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/DebtSummingPrinter.kt
```diff
@@ -0,0 +1,57 @@
+package io.gitlab.arturbosch.detekt.cli.console
```
arturbosch: Please also change the file to just `DebtSumming.kt`

detekt-cli/src/test/resources/grouped-findings-report.txt
```diff
@@ -0,0 +1,9 @@
+AnotherFile.kt - 15min debt
+	AnotherRule - [TestEntity] at AnotherFile.kt:1:1
+	NewRule - [TestEntity] at AnotherFile.kt:1:1
+	RandomRule - [TestEntity] at AnotherFile.kt:1:1
+TestFile.kt - 10min debt
+	AnotherRandomRule - [TestEntity] at TestFile.kt:1:1
+	NewRule - [TestEntity] at TestFile.kt:1:1
```
arturbosch: Looks good :+1:

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/ConfigPrinter.kt
```diff
@@ -93,6 +93,7 @@ object ConfigPrinter : DocumentationPrinter<List<RuleSetPage>> {
         #  - 'ComplexityReport'
         #  - 'NotificationReport'
         #  - 'FindingsReport'
+        #  - 'FileBasedFindingsReport'
```
arturbosch: Please remove # here. Starting with a hash is a comment in yaml. We comment out all default reports that the user knows they exist. Lines without a hash gets excluded. Thats what we want for the new report.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/DebtSummingPrinter.kt
sowmyav24: There is already a file named `DebtSumming.kt` where the calculation of debt takes place. Should we move this method there ?

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/TestFactory.kt
sowmyav24: For the tests to construct rule violations across multiple files, added a parameter fileName which is in turn used to create objects.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/DebtSummingPrinter.kt
arturbosch: Ah, yes please move it, thanks.

detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -23,6 +23,7 @@ console-reports:
   #  - 'ComplexityReport'
   #  - 'NotificationReport'
   #  - 'FindingsReport'
+  #  - 'FileBasedFindingsReport'
```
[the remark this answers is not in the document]
arturbosch: You have a compilation error in the test. An import is missing.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/ConfigPrinter.kt
sowmyav24: Done

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/DebtSummingPrinter.kt
sowmyav24: Done !

detekt-cli/src/main/resources/default-detekt-config.yml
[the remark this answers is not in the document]
sowmyav24: It passes in local and the import reported in CI is present in the file :|

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/TestFactory.kt
arturbosch: Ah I see, the `ruleSet` parameter was before your PR, just realized it should have be named `ruleName` :)

sowmyav24: Oh ! I will change it now :)

detekt-cli/src/main/resources/default-detekt-config.yml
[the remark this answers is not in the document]
arturbosch: There is a small mistake in the imports:
```
import io.gitlab.arturbosch.detekt.cli.TestDetektion
import io.gitlab.arturbosch.detekt.test.TestDetektion
```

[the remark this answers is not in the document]
sowmyav24: Realised my branch and master of forked repo was outdated and this never got reported :(

PR #1853, opened 2019-08-28
Remove unnecessary .trimMargin() in tests

schalkms:
* Removes the '|' at the start of each quoted line.
* Removes .trimMargin() where no report messages are checked
* Succeeding #1850


amitdash291:
Without the margins the line is detected as pure whitespace and results in failing checks on CodeFactor.

arturbosch:
I think this whitespace bug on codefactor is not that important to create an additional config just for codfactor.

PR #1867, opened 2019-08-30
Check for actual usage of java.lang.Void in ForbiddenVoid rule

### LINKED ISSUE #1642: False positive on ForbiddenVoid

3flex:
Rule now requires type resolution, but should not trigger false positives anymore.

Fixes #1642

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenVoidSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenVoidSpec.kt
@@ -1,16 +1,24 @@
 package io.gitlab.arturbosch.detekt.rules.style
 
 import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.test.KtTestCompiler
 import io.gitlab.arturbosch.detekt.test.TestConfig
 import io.gitlab.arturbosch.detekt.test.compileAndLint
+import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
 import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
 
 class ForbiddenVoidSpec : Spek({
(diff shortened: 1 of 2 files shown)


3flex:
These OOM errors are a bit concerning, one each on macOS and Windows while tests are running... but they are not from this PR, there were a couple recently in CI.

Will merge but something to keep an eye on.

schalkms:
> java.lang.OutOfMemoryError: GC overhead limit exceeded
>org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment.access$contentRootToVirtualFile(KotlinCoreEnvironment.kt:125)
>java.lang.OutOfMemoryError: Java heap space

The more code we `compileAndLint` with the help of Jsr223, the more memory the compile engine needs.
@3flex I investigated this issue and came to the following conclusion.
In the combination with some sort of special code, especially with the `UselessCallOnNotNull` rule tests, the compile engine throws OOM errors that are concerning.
In that case, I think this engine isn't suitable for heavy workloads like we have.
As mentioned, there is special code that utilizes the full capacity of the Jsr223 compile engine (for instance `UselessCallOnNotNull` rule tests).
Anyways, I prepared a PR in order to mitigate this problems.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenVoidSpec.kt
```diff
@@ -1,16 +1,24 @@
...
 import org.spekframework.spek2.style.specification.describe
 
 class ForbiddenVoidSpec : Spek({
     val subject by memoized { ForbiddenVoid(Config.empty) }
 
+    lateinit var environment: KotlinCoreEnvironment
+
     describe("ForbiddenVoid rule") {
+        beforeEachTest {
+            environment = KtTestCompiler.createEnvironment()
```
arturbosch: I think this line causes OOM, I think we need to dispose the kotlin environment

schalkms: You might be right. I'm currently working on a fix.

PR #1869, opened 2019-08-30
Use Gradle's logger to print debug output

### LINKED ISSUE #1855: Don't use println in Gradle plugin
`println` is used in a few places in the Gradle plugin. Instead, use Gradle's `Project.logger()`.

3flex:
Resolves #1855

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/ProcessExecutor.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/ProcessExecutor.kt
@@ -1,17 +1,18 @@
 package io.gitlab.arturbosch.detekt.invoke
 
 import io.gitlab.arturbosch.detekt.BuildFailure
+import org.gradle.api.logging.Logger
 import java.io.BufferedInputStream
 import java.io.BufferedReader
 import java.io.InputStreamReader
 
 object ProcessExecutor {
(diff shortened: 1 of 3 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/ProcessExecutor.kt
```diff
@@ -1,17 +1,18 @@
...
 
 import io.gitlab.arturbosch.detekt.BuildFailure
+import org.gradle.api.logging.Logger
 import java.io.BufferedInputStream
 import java.io.BufferedReader
 import java.io.InputStreamReader
 
 object ProcessExecutor {
-    fun startProcess(args: Array<String>, debug: Boolean = false) {
+    fun startProcess(args: Array<String>, logger: Logger) {
```
schalkms: Is the Gradle logger a globally available singleton object?
Then we could spare the extra parameter here.

3flex: Seems there is - you're right, no point for the parameter in that case.

arturbosch:
I'm sorry I merged this too fast on my mobile without paying attention of the build status. ..
I've reverted the merge. Sorry for the trouble.

schalkms:
@arturbosch 1a6327b causes problems in combination with Jsr223. The reverted commit works fine.

PR #1870, opened 2019-08-30
Dispose KotlinScriptEngine to mitigate OOM errors

schalkms:
This commit tries to mitigate OOM (java.lang.OutOfMemoryError) errors.
There is code that utilizes the capacity of the Jsr223 compile engine.
Hence, the engine now disposes state memory after each compilation step.


schalkms:
Closed because #1873 fixes this issue.

PR #1873, opened 2019-08-30
Dispose KotlinCoreEnvironment to mitigate OOM errors

schalkms:
This mitigates OOM (java.lang.OutOfMemoryError) errors.
This PR disposes the KotlinCoreEnvironment and the associated objects.
These are disposed after each test that needs type and symbol solving.

```Kotlin
// Tests are supposed to create a single project and dispose it right after use
fun createForTests()
```
**Reference**
https://github.com/JetBrains/kotlin/blob/99972a08a1324b971624bf88ae3da138dc7e10d7/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/KotlinCoreEnvironment.kt#L461

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingWhenCaseSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingWhenCaseSpec.kt
@@ -4,6 +4,8 @@ import io.gitlab.arturbosch.detekt.test.KtTestCompiler
 import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
 import org.assertj.core.api.Assertions.assertThat
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
+import org.jetbrains.kotlin.com.intellij.openapi.Disposable
+import org.jetbrains.kotlin.com.intellij.openapi.util.Disposer
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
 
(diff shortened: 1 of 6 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingWhenCaseSpec.kt
```diff
@@ -4,17 +4,22 @@ import io.gitlab.arturbosch.detekt.test.KtTestCompiler
...
+import org.jetbrains.kotlin.com.intellij.openapi.util.Disposer
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
 
 object MissingWhenCaseSpec : Spek({
 
     val subject by memoized { MissingWhenCase() }
 
-    lateinit var environment: KotlinCoreEnvironment
+    var environment: KotlinCoreEnvironment? = null
```
3flex: I wonder if there's a better way than making this nullable... there's a `createForTests` function in [KotlinCoreEnvironment](https://github.com/JetBrains/kotlin/blob/master/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/KotlinCoreEnvironment.kt) which might help.

Something to look at later.

schalkms:
We could wrap both objects in another object and only return the non-nullable reference while keeping the nullable one private or use a superclass to reduce boilerplate code.
I'll create a separate PR and merge this, so that the master branch is green again.

schalkms:
[This class](https://github.com/JetBrains/kotlin/blob/master/plugins/uast-kotlin/tests/AbstractKotlinUastTest.kt) contains an example how to dispose the object.

schalkms:
@3flex I haven't found a way to successfully implement a base class in Spek. Is that even supported? I can't access the `environment` variable.

```Kotlin
abstract class BaseSpek(root: Root.() -> Unit) : Spek({

    var environment: KotlinCoreEnvironment? = null
    lateinit var disposable: Disposable
    
    beforeEachTest {
        val pair = KtTestCompiler.createEnvironment()
        disposable = pair.first
        environment = pair.second
    }
    
    afterEachTest {
        Disposer.dispose(disposable)
        environment = null
    }
    
    root()
})
```

PR #1874, opened 2019-08-31
Use Gradle's logger to print debug output

3flex:
Reimplements #1869


3flex:
CI failures are all OOM errors.

PR #1875, opened 2019-08-31
Make all Gradle tasks cacheable

### LINKED ISSUE #1856: Most Gradle tasks not cacheable
## Expected Behavior
All Gradle tasks are cacheable.

## Observed Behavior
Only `Detekt` task is cacheable.

## Steps to Reproduce
Only `Detekt` task is annotated with `@CacheableTask` annotation.

## Context
Tasks should be cacheable if possible to take advantage of Kotlin's build cache.

Probably the create baseline & generate config tasks don't need to be cacheable since they won't be run often, but the IDEA tasks should be.

3flex:
Update annotations on tasks that were not cacheable to make them cacheable.

Using `@get:` on annotations is the style used in Gradle guides, and may in fact be required to get accurate results from `validateTaskProperties` task, so have updated existing annotations to use the `get` use-site target.

Also tweaked a couple of annotations in the Detekt task so they're correctly set on the internal & public parameters.

Fixes #1856

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
@@ -8,12 +8,14 @@ import io.gitlab.arturbosch.detekt.invoke.DetektInvoker
 import io.gitlab.arturbosch.detekt.invoke.GenerateConfigArgument
 import io.gitlab.arturbosch.detekt.invoke.InputArgument
 import org.gradle.api.file.FileCollection
+import org.gradle.api.tasks.CacheableTask
 import org.gradle.api.tasks.Classpath
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.SourceTask
 import org.gradle.api.tasks.TaskAction
 import org.gradle.language.base.plugins.LifecycleBasePlugin
 
+@CacheableTask
 open class DetektGenerateConfigTask : SourceTask() {
 
     init {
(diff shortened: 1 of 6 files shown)


3flex:
OOM errors in CI.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
```diff
@@ -8,26 +8,30 @@ import io.gitlab.arturbosch.detekt.invoke.DetektInvoker
...
     init {
         description = "Generate a detekt configuration file inside your project."
         group = LifecycleBasePlugin.VERIFICATION_GROUP
     }
 
+    // When this deprecated property is removed, also stop extending SourceTask as this task does not need any file
+    // input.
     @Deprecated("Replace with getSource/setSource")
     var input: FileCollection
         @Internal
```
schalkms: `@get:Internal` ?

3flex: There's already an `@InputFiles` annotation on `SourceTask.getSource()`. Since this class inherits from `SourceTask` there's no reason to annotate this getter with `@InputFiles` as well since the values of each are effectively identical. Also means the `@PathSensitive` annotation isn't required here.

schalkms: Thanks for the explanation.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
```diff
@@ -8,26 +8,30 @@ import io.gitlab.arturbosch.detekt.invoke.DetektInvoker
...
 
+@CacheableTask
 open class DetektGenerateConfigTask : SourceTask() {
 
     init {
         description = "Generate a detekt configuration file inside your project."
         group = LifecycleBasePlugin.VERIFICATION_GROUP
     }
 
+    // When this deprecated property is removed, also stop extending SourceTask as this task does not need any file
```
arturbosch: It should be ok to remove stuff in 1.1.0 right?

3flex: With SemVer strictly I think it would be removed in a major release since it's removing from public API...

I was thinking the deprecated options can be removed some time later if there are other breaking changes in the Gradle plugin and they can be bundled together.

PR #1879, opened 2019-08-31
Trust Kotlin compiler to identify unsafe casts

### LINKED ISSUE #1137: False positives on UnsafeCast on AppVeyor (Windows?) only
See https://github.com/arturbosch/detekt/blob/fab854abc1d1efdc6218dc8e5cf486f6fb194de2/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/ThresholdedCodeSmellAssert.kt

On *nix (my macOS machine, and on Travis' CI machines) I get no warnings.

On Windows (AppVeyor) we get these warnings:

```
Ruleset: potential-bugs - 1h debt
	UnsafeCast - [isThresholded] at C:\projects\detekt\detekt-test\src\main\kotlin\io\gitlab\arturbosch\detekt\test\ThresholdedCodeSmellAssert.kt:12:36
	UnsafeCast - [smell] at C:\projects\detekt\detekt-test\src\main\kotlin\io\gitlab\arturbosch\detekt\test\ThresholdedCodeSmellAssert.kt:26:15
	UnsafeCast - [smell] at C:\projects\detekt\detekt-test\src\main\kotlin\io\gitlab\arturbosch\detekt\test\ThresholdedCodeSmellAssert.kt:35:15
```

The warnings on Windows are incorrect, as those casts are all guarded behind an assertion. Neither the compiler nor the IDE complain about those.

## Your Environment
* Version used: RC9
* Operating System and version: macOS 10.13.6/Travis CI/AppVeyor
* Link to your project: Detekt :)

### LINKED ISSUE #1601: Make UnsafeCast less aggressive
Make UnsafeCast rule less aggressive so it only flags usage of the "unsafe" cast operator when the cast is known to fail:
```kotlin
    val something: String = 3 as String
    val somethingElse: String? = 3 as? String
```
but do not flag usage of the operator when there's not enough information for the compiler to be sure at compile time that the case will fail.

## Context
With type and symbol resolution we can flag usage of the "unsafe" cast operator when we know at compile time that the cast will fail, but ignore usage of the operator when it only *might* fail to avoid false positives.

There are some redundant usages of the "safe" cast operator within detekt that can't currently be changed to the "unsafe" operator without flagging UnsafeCast even when it would be OK, like:
```kotlin
// safe cast to KtValueArgument not necessary since type was already checked
parent is KtValueArgument && (parent as? KtValueArgument)?.isNamed() == true
```
```kotlin
// PomTransaction is a subclass of PomTransactionBase so "unsafe" cast is safe
        override fun runTransaction(transaction: PomTransaction) {
            (transaction as? PomTransactionBase)?.run()
        }
```
```kotlin
// docs say "project.version" returns a string and is never null, so safe cast not needed
override = (project.version as? String)?.endsWith("-SNAPSHOT") == true
```

Changing it to be less aggressive and match behavior of the compiler would help with this.

3flex:
UnsafeCast now trusts the Kotlin compiler to identify casts that won't succeed. This should eliminate false positives though the detection rate for actual issues will drop, as some "safe" casts that the compiler cannot always tell will fail may still cause problems.

Now requires type/symbol resolution enabled in the detekt run for the rule to work.

Fixes #1601
Fixes #1137 (I couldn't replicate the original issue by just removing the suppression lines in `ThresholdedCodeSmellAssert` and rerunning CI in AppVeyor, though this should resolve it anyway).

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnsafeCastSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnsafeCastSpec.kt
@@ -1,29 +1,53 @@
 package io.gitlab.arturbosch.detekt.rules.bugs
 
-import io.gitlab.arturbosch.detekt.test.compileAndLint
+import io.gitlab.arturbosch.detekt.test.KtTestCompiler
+import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
 import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
 
 class UnsafeCastSpec : Spek({
     val subject by memoized { UnsafeCast() }
 
+    lateinit var environment: KotlinCoreEnvironment
+
+    beforeEachTest {
+        environment = KtTestCompiler.createEnvironment()
+    }
+
     describe("check safe and unsafe casts") {
 
-        it("reports unsafe cast") {
+        it("reports cast that cannot succeed") {
+            val code = """
+                fun test(s: String) {
+                    println(s as Int)
+                }"""
+            assertThat(subject.compileAndLintWithContext(environment, code)).hasSize(1)
+        }
+
+        it("reports 'safe' cast that cannot succeed") {
+            val code = """
+                fun test(s: String) {
+                    println((s as? Int) ?: 0)
+                }"""
+            assertThat(subject.compileAndLintWithContext(environment, code)).hasSize(1)
+        }
+
+        it("does not report cast that might succeed") {
             val code = """
 				fun test(s: Any) {
 					println(s as Int)
 				}"""
-            assertThat(subject.compileAndLint(code)).hasSize(1)
+            assertThat(subject.compileAndLintWithContext(environment, code)).isEmpty()
         }
 
-        it("does not report safe cast") {
+        it("does not report 'safe' cast that might succeed") {
             val code = """
 				fun test(s: Any) {
 					println((s as? Int) ?: 0)
 				}"""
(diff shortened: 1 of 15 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
```diff
@@ -168,7 +168,7 @@ class MagicNumber(config: Config = Config.empty) : Rule(config) {
     }
 
     private fun KtConstantExpression.isNamedArgument() =
-        parent is KtValueArgument && (parent as? KtValueArgument)?.isNamed() == true && isPartOf(KtCallElement::class)
+        parent is KtValueArgument && (parent as KtValueArgument).isNamed() && isPartOf(KtCallElement::class)
```
schalkms: We could further simplify that. `parent is KtValueArgument` is unnecessary then.
```Kotlin
(parent as? KtValueArgument)?.isNamed() == true && isPartOf(KtCallElement::class)
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnsafeCastSpec.kt
```diff
@@ -1,29 +1,53 @@
...
+import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
 import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
 
 class UnsafeCastSpec : Spek({
     val subject by memoized { UnsafeCast() }
 
+    lateinit var environment: KotlinCoreEnvironment
```
schalkms: `environment` isn't disposed after each test.
Maybe we should wait for #1880 to be merged and use the new API.

3flex:
Blocked by #1880

PR #1880, opened 2019-08-31
Refactor KotlinCoreEnvironment test setup

schalkms:
This PR adds an extension function to `Root` to automatically create and
dispose the KotlinCoreEnvironment for testing purposes.
The objects in the KotlinCoreEnvironment are wrapped to simplify the
object creation and disposal.


detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/KtTestCompiler.kt
```diff
@@ -63,10 +63,23 @@ object KtTestCompiler : KtCompiler() {
...
 
     fun createPsiFactory(): KtPsiFactory = KtPsiFactory(KtTestCompiler.environment.project, false)
 }
 
+class KotlinCoreEnvironmentWrapper(
+    private var environment: KotlinCoreEnvironment?,
+    private val disposable: Disposable
+) {
+
+    fun getEnvironment() = environment!!
```
arturbosch: Lets consider changing it to `val env  get() = environment!!`. `wrapper.getEnvironment()` is quite verbose and does not need to be a function call imo.
Does the `environment` needs to be a `var` and `nullable`? The wrapper should get garbadge collection/unreferenced after each test?

schalkms: I took inspiration from the following source file in the JetBrains Kotlin repository.
The also assign `null` to the `kotlinCoreEnvironment` in the the `tearDown()` method.
I'll keep the `nullable var`, but I will refactor the function into a property as you suggested.

https://github.com/JetBrains/kotlin/blob/master/plugins/uast-kotlin/tests/AbstractKotlinUastTest.kt

PR #1883, opened 2019-09-01
Minimise usage of Kotlin reflection

3flex:
Remaining usage is in detekt-api (KClass is used in a public API, and there are some property delegates which require it).

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/IsPartOfUtils.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/IsPartOfUtils.kt
@@ -3,14 +3,13 @@ package io.gitlab.arturbosch.detekt.rules
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.psi.KtStringTemplateEntry
 import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
-import kotlin.reflect.KClass
 
 /**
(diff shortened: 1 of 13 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/IsPartOfUtils.kt
```diff
@@ -3,14 +3,13 @@ package io.gitlab.arturbosch.detekt.rules
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.psi.KtStringTemplateEntry
 import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
-import kotlin.reflect.KClass
 
 /**
  * Tests if this element is part of given PsiElement.
  */
-fun PsiElement.isPartOf(clazz: KClass<out PsiElement>): Boolean = getNonStrictParentOfType(clazz.java) != null
+inline fun <reified T : PsiElement> PsiElement.isPartOf() = getNonStrictParentOfType<T>() != null
```
arturbosch: Interesting, I remember that I had to rewrite some psi helpers because the kotlin compiler throw errors because he could not resolve all types or something ...

arturbosch:
Just out of curiosity, what was the reasoning for this changes? I see the code is now much more readable :).

3flex:
> the code is now much more readable

That was the only reason :) I hope it doesn't reintroduce any bugs.

Issue #1871, opened 2019-08-30
Indentation for compliant and non-compliant code examples is missing


schalkms:
For some rules the compliant and non-compliant code examples use tabs instead of spaces to indent lines. 
The tab character gets ignored when generating the corresponding `.md` file with the help of the `detekt-generator`.

#### Actual (with tabs)
https://arturbosch.github.io/detekt/style.html#equalsonsignatureline
https://arturbosch.github.io/detekt/style.html#explicititlambdaparameter

#### Expected (with spaces)
https://arturbosch.github.io/detekt/style.html#compliant-code-5

PR #1884, opened 2019-09-01
Add tabs check for KDoc in detekt-generator

schalkms:
This prevents the usage of tabs in KDoc to ensure correct indentation
when generating the Markdown rule documentation.
This relates to the issue described in #1871.


detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleCollectorSpec.kt
```diff
@@ -483,5 +483,21 @@ class RuleCollectorSpec : Spek({
...
+            val description = "\t"
+            val code = """
+				package foo
+
+				/**
+				 * $description
+				 */
+				class $name: Rule {
+				}
+			"""
```
3flex: The code in the string literal itself should be indented with spaces, just to make sure the only tab in code is in the $description parameter.

schalkms: Done

PR #1886, opened 2019-09-01
Only report UnsafeCallOnNullableType on actual nullable types

### LINKED ISSUE #880: UnsafeCallOnNullableType should not be reported for platform types
## Expected Behavior

detekt should not complain about this code:

    val version = UUID.randomUUID()!!

using !! here is a totally valid way to force kotlin to make version a non null type, and it's less verbose than specifying the type explicitly. It's also an autofix that IDEA suggests

## Current Behavior

    val version = UUID.randomUUID()!!

is reported as UnsafeCallOnNullableType

## Your Environment
I use detekt via codacy.

3flex:
Fixes #880 

Rule now requires type resolution.

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnsafeCallOnNullableTypeSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnsafeCallOnNullableTypeSpec.kt
@@ -1,6 +1,7 @@
 package io.gitlab.arturbosch.detekt.rules.bugs
 
-import io.gitlab.arturbosch.detekt.test.compileAndLint
+import io.gitlab.arturbosch.detekt.test.KtTestCompiler
+import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
 import org.assertj.core.api.Assertions.assertThat
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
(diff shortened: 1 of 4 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnsafeCallOnNullableTypeSpec.kt
```diff
@@ -1,37 +1,63 @@
...
+import org.jetbrains.kotlin.com.intellij.openapi.Disposable
+import org.jetbrains.kotlin.com.intellij.openapi.util.Disposer
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
 
 class UnsafeCallOnNullableTypeSpec : Spek({
     val subject by memoized { UnsafeCallOnNullableType() }
 
+    var environment: KotlinCoreEnvironment? = null
+    lateinit var disposable: Disposable
```
schalkms: Maybe we should wait for #1880 to be merged and use the new API.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnsafeCallOnNullableTypeSpec.kt
```diff
@@ -1,37 +1,63 @@
...
 				}"""
-            assertThat(subject.compileAndLint(code)).hasSize(1)
+            assertThat(subject.compileAndLintWithContext(environment!!, code)).hasSize(1)
+        }
+
+        it("does not report unsafe call on platform type") {
+            val code = """
+				import java.util.UUID
+
+                val version = UUID.randomUUID()!!"""
```
schalkms: Formatting looks weird here.

3flex:
Blocked by #1880

Issue #959, opened 2018-06-17
Rule ForbiddenComment - regexp support


AndreiZaitcev:
I think that the rule `ForbiddenComment` is not flexible enough. In my case I'd like to flag every usage of `TODO,TODO:,etc...` but if it contains a ticket number then it should pass the check.

Noncompliant code:

```
// TODO remove this code later
fun deprecatedFunction() {}
```

Compliant code:
```
// TODO DTK-2019 remove this code later
```

The main difference for me is that I can disallow usage of TODOs unless a task is tracked in any task system.

Does it make sense to have something like that in Detekt or I should create a custom rule?

vanniktech:
There's no need to prefix a ticket with TODO. Hence I'd just reference the ticket and leave the todo as is. A ticket implies that there is something missing.

AndreiZaitcev:
@vanniktech makes sense. I guess I can adjust IDEA to show such comments in the same way it shows TODOs. I'll leave this ticket open if someone has other ideas, but feel free to close it.

vanniktech:
Yes you can do that. I'm coloring links for instance:

<img width="1149" alt="screen shot 2018-06-17 at 12 18 50" src="https://user-images.githubusercontent.com/5759366/41506968-9c453b9a-7228-11e8-872f-fe5e6c0cfe78.png">

<img width="399" alt="screen shot 2018-06-17 at 12 19 08" src="https://user-images.githubusercontent.com/5759366/41506976-aa894138-7228-11e8-908d-b5645ab000ba.png">

arturbosch:
That might be a new rule, feel free to open an issue for that. It will need reference solving which is not yet used in detekt though.

mikekudzin:
Is there any way to suppress suppress `ForbiddenComment`? We have several distinct todo's we dealt 
 with and bound them to TD tickets. I got @vanniktech 's point that we do not need "todo" in a such kinda of comments. I almost agree with that. However
1. I worked on a several projects with different teams in different domains and can prove it is a quite common pattern to leave "//TODO TICKET-1234" comments when they agreed upon during a code review stage. It is convenient and has IDE support out of the box.
2. Given (1) I have to force all team members to share the same IDE config and argue with them on why we can't use an approach they used to use for years.

arturbosch:
I agree with @mikekudzin and @AndreiZaitcev here.
Next to `values` there should be `allowedPatterns` to handle ticket numbers.

PR #123, opened 2017-06-23
Improve changelog generator

rock3r:
I've tweaked the changelog generator script a bit:

 * **Parameters are now optional**, uses default values pointing to this repo and the latest milestone
    Next improvement would be named arguments so you can override only one of the defaults at a time
 * All **URLs point to human-friendly HTML pages** instead of pointing to API responses
 * Titles are `trim()`med to avoid leading and trailing spaces

#### M12

- Suppress for TooManyFunctions is not considered in detekt - [#108](https://api.github.com/repos/arturbosch/detekt/issues/108)
- Update documentation and migration guide for M12 - [#105](https://api.github.com/repos/arturbosch/detekt/issues/105)
- NoDocOverPublicClass always reported for objects - [#104](https://api.github.com/repos/arturbosch/detekt/issues/104)
...
- Baseline does not work anymore as expected due to inner restructure for the new output formats  - [#83](https://api.github.com/repos/arturbosch/detekt/issues/83)
...

See all issues at: [M12](https://api.github.com/repos/arturbosch/detekt/milestones/8)

#### M12

- Suppress for TooManyFunctions is not considered in detekt - [#108](https://github.com/arturbosch/detekt/issues/108)
- Update documentation and migration guide for M12 - [#105](https://github.com/arturbosch/detekt/issues/105)
- NoDocOverPublicClass always reported for objects - [#104](https://github.com/arturbosch/detekt/issues/104)
...
- Baseline does not work anymore as expected due to inner restructure for the new output formats - [#83](https://github.com/arturbosch/detekt/issues/83)
...

See all issues at: [M12](https://github.com/arturbosch/detekt/milestone/8)

PR #1887, opened 2019-09-02
Add configuration to allow patterns for forbidden comment

sowmyav24:
Fixes [#959](https://github.com/arturbosch/detekt/issues/959).
- Introduce configuration `allowedPatterns` to specify allowed regex (Comma separated) .
- Do not report comments that match any of the configured patterns

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
@@ -19,7 +19,8 @@ import org.jetbrains.kotlin.com.intellij.psi.PsiComment
  * </noncompliant>
  *
  * @configuration values - forbidden comment strings (default: `'TODO:,FIXME:,STOPSHIP:'`)
- * @configuration allowedPatterns - ignores comments which match any of specified regex (default: `""`)
+ * @configuration allowedPatterns - ignores comments which match any of the specified regular expressions.
+   The regular expressions are separated by `,`, like for instance `\\s*Ticket\\s*, \\s*Task\\s*`. (default: `""`)
  * @active since v1.0.0
  */
 class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
@@ -38,7 +39,7 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
             valueOrDefault(ALLOWED_PATTERNS, "")
                 .split(",")
                 .filter { it.isNotBlank() }
-                .map { r -> r.toRegex(RegexOption.IGNORE_CASE) }
+                .map { r -> r.toRegex() }
 
     override fun visitComment(comment: PsiComment) {
         super.visitComment(comment)
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
@@ -19,8 +19,8 @@ import org.jetbrains.kotlin.com.intellij.psi.PsiComment
  * </noncompliant>
  *
  * @configuration values - forbidden comment strings (default: `'TODO:,FIXME:,STOPSHIP:'`)
- * @configuration allowedPatterns - ignores comments which match any of the specified regular expressions.
-   The regular expressions are separated by `,`, like for instance `\\s*Ticket\\s*, \\s*Task\\s*`. (default: `""`)
+ * @configuration allowedPatterns - ignores comments which match the specified regular expression.
+   For example `Ticket|Task`. (default: `""`)
  * @active since v1.0.0
  */
 class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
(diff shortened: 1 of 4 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -19,7 +19,7 @@ import org.jetbrains.kotlin.com.intellij.psi.PsiComment
  * </noncompliant>
  *
  * @configuration values - forbidden comment strings (default: `'TODO:,FIXME:,STOPSHIP:'`)
- *
+ * @configuration allowedPatterns - ignores comments which match any of specified regex (default: `""`)
```
schalkms: A more descriptive comment might help our users when configuring this rule in detekt.
@arturbosch are you okay with having multiple regular expressions in **one** config option?


```suggestion
 * @configuration allowedPatterns - ignores comments which match any of the specified regular expressions. 
 The regular expressions are separated by `,`, like for instance `\\s*Ticket\\s*, \\s*Task\\s*`. (default: `""`)
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -34,11 +34,19 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
                     .split(",")
                     .filter { it.isNotBlank() }
 
+    private val allowedPatterns: List<Regex> =
+            valueOrDefault(ALLOWED_PATTERNS, "")
+                .split(",")
+                .filter { it.isNotBlank() }
+                .map { r -> r.toRegex(RegexOption.IGNORE_CASE) }
```
schalkms: Do we really want to add the **ignore_case** option for every possible regex. I don't think so.
This could be a very surprising behavior for users of detekt.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
```diff
@@ -82,5 +82,38 @@ class ForbiddenCommentSpec : Spek({
...
+            it("should not report Comment usages when all patterns are present") {
+                val comment = "// Comment Ticket:123 Task:456 comment."
+                val findings = ForbiddenComment(patternsConfig).compileAndLint(comment)
+                assertThat(findings).hasSize(0)
+            }
+
+            it("should not report Comment usages when pattern is present irrespective ofcase") {
+                val comment = "// Comment TASK - 568."
+                val findings = ForbiddenComment(patternsConfig).compileAndLint(comment)
+                assertThat(findings).hasSize(0)
```
schalkms: If we remove the **ignore_case** option, this test wil become obsolete.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
sowmyav24: Yes, I think for regex configs, we can remove this parameter. 
But if we think of the use case, where people track `TODO` with another ticket number,
They can either write as `TODO: Ticket: #123` or `TODO: TICKET: #123`,where Ticket can be project specific code. In my current project, I have seen this being used either ways.
So, if we are to ignore such comments, we can handle them in a single regex rather than two different ones.
 Or we can leave it to the user to either ignore or consider the case in their specified regex expression.
Thoughts ?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
sowmyav24: Sure, will update the description once we decide if we are going with single config option or multiple.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
3flex: > leave it to the user to either ignore or consider the case in their specified regex

I think this is the right way to go. Detekt shouldn't modify behaviour of regex provided by users.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
sowmyav24: Updated the description as suggested

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
sowmyav24: Sure, removed ignore_case, If need_be, it can be added as part of config

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
sowmyav24: Removed

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
3flex: Thanks! Would you mind also just changing "like for instance" to "for example"?

Will wait for Artur to confirm if multiple regex in one option is ok or not before merging. I've requested his review.

schalkms: I agree with @3flex 
This behavior would be very surprising for our users, since we only use a single regex as configuration options. We should be consistent here.
Furthermore, if a user configures a regex with comma `,`, this rule would break.

sowmyav24: > Furthermore, if a user configures a regex with comma `,`, this rule will break.

Yeah ! This is something that could potentially create problems. So, for a safer side, we can go with a single regex configuration and leave it to the user to support multiple expressions within it. Shall I go ahead and change it for wait for the existing code to be reviewed ?

schalkms: Yes, please change this to a single regex. This is the _detekt idiomatic_ way.
Everything else in this PR is fine. There’s no need to wait for another review.

schalkms: Thank you btw for the time to discuss this PR.

PR #1888, opened 2019-09-02
Clarify `WildcardImport` rule configuration

AlexCzar:
Add information about conflict between `formatting:` `NoWildcardImports` and `style:` `WildcardImport` rules.

There is a conflict between `formatting:` `NoWildcardImports` and `style:` `WildcardImport` , effectively the latter and its configuration do not matter when former is enabled. This PR adds a note about the fact to the documentation.

### DIFF
--- a/docs/pages/documentation/style.md
+++ b/docs/pages/documentation/style.md
@@ -1399,6 +1399,10 @@ which classes are imported and helps prevent naming conflicts.
 
 Library updates can introduce naming clashes with your own classes which might result in compilation errors.
 
+**NOTE:** This rule is effectively overridden by the `NoWildcardImports` formatting rule (a wrapped KtLint rule).
+That rule will fail the check regardless of the whitelist configured here.
+Therefore if whitelist is needed `NoWildcardImports` rule should be disabled.
+
 **Severity**: Style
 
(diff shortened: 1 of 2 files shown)


docs/pages/documentation/style.md
```diff
@@ -1410,6 +1410,11 @@ Library updates can introduce naming clashes with your own classes which might r
    Define a whitelist of package names that should be allowed to be imported
 with wildcard imports.
 
+**NOTE:** This rule conflicts with `NoWildcardImports` formatting rule.
```
cortinico: I do believe that _conflicts_ here is probably not the best term.

I would rephrase with something like:

```suggestion
**NOTE:** This rule is similar to the `NoWildcardImports` formatting rule (a wrapper of a KtLint rule).
If you need to specify a `excludeImports` configuration, please consider disabling the other rule.
```

Or something like that.

AlexCzar: I like the wrapper clarification suggestion.
Regarding conflict/similarity: this pull request was born exactly because of the conflict between these two rules and confusion surrounding it. Yes they are indeed similar, but they are also in conflict and the most restrictive one wins.

Issue #1868, opened 2019-08-30
Add option to ignore guard clauses in ReturnCount rule


3flex:
## Expected Behavior of the rule
ReturnCount rule has an option to ignore use of [guard clauses](https://www.refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html).

## Context
In some functions it is cleaner to have a series of guard clauses at the top of the function to return early than some other methods of achieving the same result.

An option to ignore guard clauses can be added to the rule. When enabled the logic could be to ignore `return`s for all statements at the beginning of a function, but only when statements are in one of the following forms:

```kotlin
if (expression) return
val x = person.name ?: return
```

These would have to be the first declarations in the function body. As soon as a statement not in that form appeared then ReturnCount would start counting instances of return.

To be debated is whether or not `require` or `check` functions could appear before guard clauses. They serve a very similar function. IMHO these functions (and variants) could appear before guard clauses as well, since it would be better for the function to throw an exception due to unexpected/incorrect state before guard clauses are checked, as guard clauses are part of normal program flow.

amitdash291:
Picking this feature up.
Targeting to support `if` condition guard clauses (need not be the first statement necessarily).
Planning to add the configuration `excludeGuardClauses` for this purpose with a default value of `false`.

3flex:
Thanks @amitdash291! I'd still like to see the guard clauses be ignored when they're the first statements (or following `require` or `check` statements) as returns from other areas of the function, even as part of a simple if statement, couldn't really be considered "guards", instead they're just part of the function's logic.

Also `ignoreGuardClauses` would be better as the config option name I think.

Looking forward to the PR!

amitdash291:
> Thanks @amitdash291! I'd still like to see the guard clauses be ignored when they're the first statements (or following `require` or `check` statements) as returns from other areas of the function, even as part of a simple if statement, couldn't really be considered "guards", instead they're just part of the function's logic.
> 
> Also `ignoreGuardClauses` would be better as the config option name I think.
> 
> Looking forward to the PR!

@3flex, thanks for the response.

Support for the Elvis operator guard clauses (eg: `val x = person.name ?: return`) is also possible. The only remaining piece of the puzzle is to make sure the clauses are the first statement of the functions. I am working on it.

amitdash291:
@3flex, the PR is ready and approved. It would be great if you could review it too :)

amitdash291:
[...]

PR #1889, opened 2019-09-02
Add option to ignore guard clauses in ReturnCount rule

amitdash291:
This PR is in relation to the issue #1868 

The following changes have been done:
 - Add option to ignore guard clauses in ReturnCount rule by using the config `excludeGuardClauses`, default value of the config is `false`
- The guard clauses include two types, namely `if condition` and `Elvis operator`
- The guard clauses are only ignored if they are the first statement in a function
- Fun fact: one of the commit fixes `ReturnCount` debt in `ReturnCount.kt` file :)

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -497,6 +497,7 @@ style:
     excludedFunctions: "equals"
     excludeLabeled: false
     excludeReturnFromLambda: true
+    excludeGuardClauses: false
   SafeCast:
     active: true
   SerialVersionUIDInSerializableClass:
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -497,7 +497,7 @@ style:
     excludedFunctions: "equals"
     excludeLabeled: false
     excludeReturnFromLambda: true
-    excludeGuardClauses: false
+    excludeGuardClauses: true
   SafeCast:
     active: true
   SerialVersionUIDInSerializableClass:
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -497,7 +497,7 @@ style:
     excludedFunctions: "equals"
     excludeLabeled: false
     excludeReturnFromLambda: true
-    excludeGuardClauses: true
+    excludeGuardClauses: false
   SafeCast:
     active: true
   SerialVersionUIDInSerializableClass:
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
@@ -54,7 +54,7 @@ import org.jetbrains.kotlin.psi.psiUtil.isFirstStatement
  * (default: `false`)
  * @configuration excludeReturnFromLambda - if labeled return from a lambda should be ignored
  * (default: `true`)
- * @configuration excludeGuardClauses - if true guard clauses at the beginning of a method will be ignored
+ * @configuration excludeGuardClauses - if true guard clauses at the beginning of a method should be ignored
  * (default: `false`)
  * @active since v1.0.0
  */
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
@@ -89,11 +89,13 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
         function.accept(object : DetektVisitor() {
             override fun visitKtElement(element: KtElement) {
                 if (element is KtReturnExpression) {
-                    if (excludeLabeled && element.labeledExpression != null) {
-                        return
-                    } else if (excludeLambdas && isNamedReturnFromLambda(element)) {
-                        return
-                    } else if (excludeGuardClauses && isGuardClause(element)) {
+                    val breakRecursion = when {
+                        excludeLabeled && element.labeledExpression != null -> true
+                        excludeLambdas && isNamedReturnFromLambda(element) -> true
+                        excludeGuardClauses && isGuardClause(element) -> true
+                        else -> false
+                    }
+                    if (breakRecursion) {
                         return
                     } else {
                         returnsNumber++
@@ -133,10 +135,12 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
     }
 
     private fun isElvisOperatorGuardClause(expression: KtReturnExpression): Boolean {
-        val ktBinaryExpression = expression.parent as? KtBinaryExpression
-        ktBinaryExpression?.let {
-            return (it.parent as? KtElement)?.isFirstStatement() ?: false
-                    && (it.operationToken as? KtSingleValueToken)?.value == "?:"
+        val elvisGuardClauseExpression = expression.parent as? KtBinaryExpression
+        elvisGuardClauseExpression?.let {
+            val elvisGuardClauseParent = it.parent as? KtElement
+            val isFirstStatement = elvisGuardClauseParent?.isFirstStatement() ?: false
+            val operationTokenValue = (it.operationToken as? KtSingleValueToken)?.value
+            return isFirstStatement && operationTokenValue == ELVIS_OPERATOR
         }
 
         return false
@@ -148,5 +152,6 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
         const val EXCLUDE_LABELED = "excludeLabeled"
         const val EXCLUDE_RETURN_FROM_LAMBDA = "excludeReturnFromLambda"
         const val EXCLUDE_GUARD_CLAUSES = "excludeGuardClauses"
+        const val ELVIS_OPERATOR = "?:"
     }
 }
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
@@ -1,9 +1,24 @@
 package io.gitlab.arturbosch.detekt.rules.style
 
-import io.gitlab.arturbosch.detekt.api.*
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.DetektVisitor
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.SplitPattern
 import io.gitlab.arturbosch.detekt.rules.parentsOfTypeUntil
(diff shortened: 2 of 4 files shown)


detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -497,6 +497,7 @@ style:
     excludedFunctions: "equals"
     excludeLabeled: false
     excludeReturnFromLambda: true
+    excludeGuardClauses: false
```
schalkms: `true` seems reasonable too.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -50,6 +54,8 @@ import org.jetbrains.kotlin.psi.KtReturnExpression
  * (default: `false`)
  * @configuration excludeReturnFromLambda - if labeled return from a lambda should be ignored
  * (default: `true`)
+ * @configuration excludeGuardClauses - if true guard clauses at the beginning of a method will be ignored
```
schalkms:
```suggestion
 * @configuration excludeGuardClauses - if true guard clauses at the beginning of a method should be ignored
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -110,10 +121,37 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
         return false
     }
 
+    private fun isGuardClause(expression: KtReturnExpression): Boolean {
+        return isIfConditionGuardClause(expression) || isElvisOperatorGuardClause(expression)
+    }
+
+    private fun isIfConditionGuardClause(expression: KtReturnExpression): Boolean {
+        val ifConditionParent = expression.parent?.parent as? KtIfExpression
```
schalkms: Is `expression.parent` not enough?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
```diff
@@ -235,8 +307,8 @@ class ReturnCountSpec : Spek({
             it("should be empty when labeled returns are de-activated") {
                 val findings = ReturnCount(
                         TestConfig(mapOf(
-                                "excludeLabeled" to "true",
-                                "excludeReturnFromLambda" to "false"
+                                ReturnCount.EXCLUDE_LABELED to "true",
+                                ReturnCount.EXCLUDE_RETURN_FROM_LAMBDA to "false"
```
schalkms: Good catch - nicely refactored!

detekt-cli/src/main/resources/default-detekt-config.yml
amitdash291: Sure, we can set it to be `true` by default.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
amitdash291: Thanks!

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
amitdash291: Interestingly `return` expression's parent is just a `then` expression. The `then` expression's parent (`return` expression's grandparent) is the `if` condition that we are looking for.

amitdash291:
@schalkms, thanks for the prompt review. I have made the required changes and replied to the comments. Please look into the changes when possible.
It would be great to see this PR getting merged :)

amitdash291:
> This looks great! I haven't reviewed in detail but tests passing is very promising.
> 
> I'd also like to see tests that ensure this sort of pattern is excluded by the config option too:
> 
> ```kotlin
> fun testFun(testVal: String?) {
>   require(testVal != null)
>   check(testVal != null)
>   if (testVal != null) return
> }
> ```
> 
> Essentially if require and check precede the guard clauses then the guard clause should not affect return count. Can you please add some tests for this? Or maybe we can leave this out for the first iteration.

I would recommend let's do the support for guard clauses followed by require and check in the next iteration, possibly as a part of a follow-up issue? I would be happy to pick it up.

I had a few questions with regards to the same, which we can discuss upon in the follow-up issue.

amitdash291:
> This looks great! I haven't reviewed in detail but tests passing is very promising.
> 
> I'd also like to see tests that ensure this sort of pattern is excluded by the config option too:
> 
> ```kotlin
> fun testFun(testVal: String?) {
>   require(testVal != null)
>   check(testVal != null)
>   if (testVal != null) return
> }
> ```
> 
> Essentially if require and check precede the guard clauses then the guard clause should not affect return count. Can you please add some tests for this? Or maybe we can leave this out for the first iteration.

Thanks for your review @3flex!

I would recommend doing the support for _guard clauses followed by require and check functions_ in the next iteration, possibly as a part of a follow-up issue? I would be happy to pick it up.

I had a few questions with regards to the same, which we can discuss upon in the follow-up issue.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -110,10 +121,37 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
...
+        ifConditionParent?.let {
+            return it.isFirstStatement() && it.`else` == null
+        }
+
+        return false
+    }
+
+    private fun isElvisOperatorGuardClause(expression: KtReturnExpression): Boolean {
+        val elvisGuardClauseExpression = expression.parent as? KtBinaryExpression
```
3flex: Instead, try this pattern:
```kotlin
expression.parent as? KtBinaryExpression ?: return false
```
Then remove the `let` block and then use `return isFirstStatement && ...` as the last statement. This will be more concise and a bit clearer.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
3flex: See my next comment regarding removal of the `let` block.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -82,9 +89,13 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
         function.accept(object : DetektVisitor() {
             override fun visitKtElement(element: KtElement) {
                 if (element is KtReturnExpression) {
-                    if (excludeLabeled && element.labeledExpression != null) {
-                        return
-                    } else if (excludeLambdas && isNamedReturnFromLambda(element)) {
+                    val breakRecursion = when {
```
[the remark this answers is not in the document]
amitdash291: The rewrite would look like the following:
```
when {
  condition1 -> return
  condition1 -> return
  condition1 -> return
  else -> returnsNumber++
}
```
It results in using 3 returns in this function which violates ReturnCount rule itself :)

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -110,10 +121,37 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
         return false
     }
 
+    private fun isGuardClause(expression: KtReturnExpression): Boolean {
```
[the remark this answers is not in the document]
amitdash291: Good suggestion, and it is done.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -110,10 +121,37 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
...
+        return false
+    }
+
+    private fun isElvisOperatorGuardClause(expression: KtReturnExpression): Boolean {
+        val elvisGuardClauseExpression = expression.parent as? KtBinaryExpression
+        elvisGuardClauseExpression?.let {
+            val elvisGuardClauseParent = it.parent as? KtElement
+            val isFirstStatement = elvisGuardClauseParent?.isFirstStatement() ?: false
+            val operationTokenValue = (it.operationToken as? KtSingleValueToken)?.value
+            return isFirstStatement && operationTokenValue == ELVIS_OPERATOR
```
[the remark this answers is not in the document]
amitdash291: Aha! `KtTokens.ELVIS` is a lifesaver, wish I had known of this constant earlier.
The change is done

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
amitdash291: This is done.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
amitdash291: The requested change is done and updated in the comments below.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -50,6 +54,8 @@ import org.jetbrains.kotlin.psi.KtReturnExpression
  * (default: `false`)
  * @configuration excludeReturnFromLambda - if labeled return from a lambda should be ignored
  * (default: `true`)
+ * @configuration excludeGuardClauses - if true guard clauses at the beginning of a method should be ignored
+ * (default: `true`)
```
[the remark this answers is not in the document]
amitdash291: This is done.

schalkms:
Some paths are not examined.
Please take a look at our test coverage report for `ReturnCount`.

https://codecov.io/gh/arturbosch/detekt/pull/1889/src/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt?before=detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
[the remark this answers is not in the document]
3flex: There may not be :/

Leave it as is, better to merge as is than have this hold it up.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
amitdash291: **Update**
Had to revert the suggestions as it was failing code coverage :(

amitdash291:
> Some paths are not examined.
> Please take a look at our test coverage report for `ReturnCount`.
> 
> https://codecov.io/gh/arturbosch/detekt/pull/1889/src/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt?before=detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt

I had to revert one of the suggested cleanups to fix this issue.

Everything else looks good, thanks @3flex and @schalkms for your detailed reviews and suggestions.

Please merge this PR when the builds are successful :)

PR #1890, opened 2019-09-02
Don't check for unnecessary parentheses around lambdas

### LINKED ISSUE #1222: false positive unnecessary parentheses in conjunction with lambdas
## Expected Behavior
I would argue that one does not put parentheses around lambdas if it is not necessary. Do not report it if lambdas are involved because in most cases it is necessary due to ambiguity problems.

## Current Behavior
A lot of issues (depending on the code of course) which cannot be solved and need to be ignored.

see project link

## Your Environment
* Version used: 1.0.0.RC8
* Link to your project: https://github.com/loewenfels/dep-graph-releaser

# Report
I am ignoring the rule thus no link to sonarcloud but at least a list for easier reproduction. Only the first and the last issue can be resolved, all others have to stay the way they are
```
C:\projects\dep-graph-releaser\api\dep-graph-releaser-api-common\src\main\kotlin\ch\loewenfels\depgraph\data\ReleasePlan.kt:102:60 
Parentheses in (it.isSubmodule && it.level == project.level) are unnecessary and can be replaced with: it.isSubmodule && it.level == project.level
C:\projects\dep-graph-releaser\atrium\src\main\kotlin\ch\tutteli\atrium\spekProjectAssertions.kt:155:57 
Parentheses around the lambda ({ isA {} }) are unnecessary and can be removed.
C:\projects\dep-graph-releaser\atrium\src\main\kotlin\ch\tutteli\atrium\spekProjectAssertions.kt:175:57 
Parentheses around the lambda ({ isA { isStateReady() property(subject::nextDevVersion).toBe(rootProjectIdAndVersions.nextDevVersion) } }) are unnecessary and can be removed.
C:\projects\dep-graph-releaser\atrium\src\main\kotlin\ch\tutteli\atrium\spekProjectAssertions.kt:288:57 
Parentheses around the lambda ( { isJenkinsMavenReleaseWaiting(project.nextDevVersion, dependency) } ) are unnecessary and can be removed.
C:\projects\dep-graph-releaser\atrium\src\main\kotlin\ch\tutteli\atrium\spekProjectAssertions.kt:313:57 
Parentheses around the lambda ( { isJenkinsUpdateDependencyWaiting(dependency) } ) are unnecessary and can be removed.
C:\projects\dep-graph-releaser\atrium\src\main\kotlin\ch\tutteli\atrium\spekProjectAssertions.kt:346:57 
Parentheses around the lambda ( { isJenkinsMavenReleaseWaiting(project.nextDevVersion, syntheticRoot) } ) are unnecessary and can be removed.
C:\projects\dep-graph-releaser\atrium\src\main\kotlin\ch\tutteli\atrium\spekProjectAssertions.kt:428:57 
Parentheses around the lambda ( { isJenkinsMultiMavenReleaseWaiting(project.nextDevVersion, dependency) } ) are unnecessary and can be removed.
C:\projects\dep-graph-releaser\atrium\src\main\kotlin\ch\tutteli\atrium\spekProjectAssertions.kt:534:54 
Parentheses around the lambda ({ contains("cyclic dependencies", dependencyBranch, *otherDependencyBranches) }) are unnecessary and can be removed.
C:\projects\dep-graph-releaser\atrium\src\main\kotlin\ch\tutteli\atrium\spekProjectAssertions.kt:546:51 
Parentheses around the lambda ({ contains("cyclic dependencies", dependencyBranch, *otherDependencyBranches) }) are unnecessary and can be removed.
C:\projects\dep-graph-releaser\dep-graph-releaser-maven\src\test\kotlin\ch\loewenfels\depgraph\maven\IntegrationSpec.kt:174:62 
Parentheses around the lambda ({ contains(pom.absolutePathAsString) }) are unnecessary and can be removed.
C:\projects\dep-graph-releaser\dep-graph-releaser-maven\src\test\kotlin\ch\loewenfels\depgraph\maven\IntegrationSpec.kt:184:62 
Parentheses around the lambda ({ contains(pom.absolutePathAsString) }) are unnecessary and can be removed.
C:\projects\dep-graph-releaser\dep-graph-releaser-runner\src\main\kotlin\ch\loewenfels\depgraph\runner\Orchestrator.kt:76:22 
Parentheses in (length - this.length) are unnecessary and can be replaced with: length - this.length`
```

3flex:
Fixes #1222 

Logic of UnnecessaryParentheses rule is now identical to the IntelliJ intention provided by the Kotlin plugin: https://github.com/JetBrains/kotlin/blob/7eac7d96dc28970762da7187884fe40817700f85/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveUnnecessaryParenthesesIntention.kt#L31-L32

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParenthesesSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParenthesesSpec.kt
@@ -36,19 +36,6 @@ class UnnecessaryParenthesesSpec : Spek({
             assertThat(subject.lint(code)).hasSize(1)
         }
 
-        it("reports unnecessary parentheses around lambdas") {
-            val code = """
-				fun function (a: (input: String) -> Unit) {
-					a.invoke("TEST")
-				}
-
-				fun test() {
-					function({ input -> println(input) })
-				}
-				"""
-            assertThat(subject.lint(code)).hasSize(1)
-        }
-
         it("doesn't report function calls containing lambdas and other parameters") {
             val code = """
 				fun function (integer: Int, a: (input: String) -> Unit) {
@@ -59,7 +46,7 @@ class UnnecessaryParenthesesSpec : Spek({
 					function(1, { input -> println(input) })
 				}
 				"""
-            assertThat(subject.lint(code)).hasSize(0)
+            assertThat(subject.lint(code)).isEmpty()
         }
 
         it("does not report unnecessary parentheses when assigning a lambda to a val") {
@@ -67,7 +54,7 @@ class UnnecessaryParenthesesSpec : Spek({
 				fun f() {
 					instance.copy(value = { false })
 				}"""
-            assertThat(subject.lint(code)).hasSize(0)
+            assertThat(subject.lint(code)).isEmpty()
         }
 
         it("does not report well behaved parentheses") {
@@ -77,7 +64,7 @@ class UnnecessaryParenthesesSpec : Spek({
 					 	println("Test")
 					}
 				}"""
-            assertThat(subject.lint(code)).hasSize(0)
+            assertThat(subject.lint(code)).isEmpty()
         }
 
         it("does not report well behaved parentheses in super constructors") {
@@ -89,7 +76,7 @@ class UnnecessaryParenthesesSpec : Spek({
 					}
 				})
 				"""
-            assertThat(subject.lint(code)).hasSize(0)
+            assertThat(subject.lint(code)).isEmpty()
         }
 
         it("does not report well behaved parentheses in constructors") {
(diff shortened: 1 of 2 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParenthesesSpec.kt
```diff
@@ -36,19 +36,6 @@ class UnnecessaryParenthesesSpec : Spek({
             assertThat(subject.lint(code)).hasSize(1)
         }
 
-        it("reports unnecessary parentheses around lambdas") {
```
schalkms: Should we rename this test to `it("*does not report* unnecessary parentheses around lambdas")`? and check if nothing gets reported?

3flex: Will do, then merge. Tested that locally and the updated rule does not flag that case so will add that back and assert findings is empty.

PR #1897, opened 2019-09-06
Fix shared variable in detekt-extension rules

schalkms:


detekt-sample-extensions/src/main/kotlin/io/gitlab/arturbosch/detekt/sample/extensions/rules/TooManyFunctionsTwo.kt
```diff
@@ -40,6 +40,7 @@ class TooManyFunctionsTwo(config: Config) : ThresholdRule(config, THRESHOLD) {
                 references = emptyList())
             )
         }
+        amount = 0
```
schalkms: Do we really need two example rules that basically do the same?

3flex: Maybe it's to demonstrate how the rule provider works when registering multiple rules? But maybe removing the second example and just adding a comment to `SampleProvider` is enough.

They don't require much maintenance though so we could just leave it.

PR #1900, opened 2019-09-07
Add RedundantExplicitType rule

### LINKED ISSUE #336: Rule: OptionalTypeDeclaration
The following code

```kotlin
private val compositeDisposable: CompositeDisposable = CompositeDisposable()
```

can be simplified by omitting `: CompositeDisposable` since the type can be interfered.

```kotlin
private val compositeDisposable = CompositeDisposable()
```

Obviously this only works when both the type of the declaration and initialization is the same. The following should not be flagged:

```kotlin
private val compositeDisposable: Disposable = CompositeDisposable()
```

since you might only want to work with the `Disposable` type which is implemented by the `CompositeDisposable`. I don't think that we need type resolution for this a simple equals check on the name should be enough.

Any thoughts?

3flex:
Closes #336 

Since this is just a port of the IntelliJ rule this uses identical logic, meaning this only checks local properties - it doesn't check for types that don't need to be declared for any other element. I think this conservative posture makes sense and there shouldn't be any false positives.

A future config option could be to enable or disable checking on top-level properties, but that's not required for now.

Blocked by #1880

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitType.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitType.kt
@@ -1,0 +1,94 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.KtNodeTypes
+import org.jetbrains.kotlin.builtins.KotlinBuiltIns
+import org.jetbrains.kotlin.descriptors.VariableDescriptor
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtConstantExpression
+import org.jetbrains.kotlin.psi.KtNameReferenceExpression
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.psi.KtStringTemplateExpression
+import org.jetbrains.kotlin.resolve.BindingContext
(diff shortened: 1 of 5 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitType.kt
```diff
@@ -0,0 +1,93 @@
...
+        "Type does not need to be stated explicitly and can be removed.",
+        Debt.FIVE_MINS
+    )
+
+    @Suppress("ReturnCount", "ComplexMethod")
+    override fun visitProperty(property: KtProperty) {
+        if (bindingContext == BindingContext.EMPTY) return
+        if (!property.isLocal) return
+        val typeReference = property.typeReference ?: return
+        val type =
```
schalkms: I see this pattern for getting the `BindingContext` quite often.
Shall we refactor this into a generic method?

schalkms:
I would also acknowledge the work (`kotlinc`) somewhere in a `license.md` to comply with the Apache 2.0 license.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitType.kt
3flex: I think that's a good idea. Can also provide a warning if context not available so user knows the rule (or parts of the rule) were not run even when the rule was enabled?

Can that be done in another PR though?

schalkms: Yes, that makes sense.
Currently, our documentation states:

```
--jvm-target
   EXPERIMENTAL: Target version of the generated JVM bytecode that was
   generated during compilation and is now being used for type resolution
```

I think it’s not experimental anymore. Furthermore, we also need to update the samples.

3flex:
> I would also acknowledge the work (kotlinc) somewhere in a license.md to comply with the Apache 2.0 license.

I'm not clear on what actually needs to be done to provide proper attribution - I know detekt and Kotlin compiler have Apache 2.0 licenses which is good, but in terms of correct attribution, is the reference in the source file to the original enough?

schalkms:
@3flex yes, I think so. I'll create an issue for that.

PR #1904, opened 2019-09-08
Rule: UseIfInsteadOfWhen

3flex:
Suggests replacing simple when `expressions` with an equivalent `if` `else` in line with this recommendation: https://kotlinlang.org/docs/reference/coding-conventions.html#if-versus-when

Ran the rule on detekt's codebase and fixed the resulting issues.

Only `when` expressions with two entries (one of them being `else` and the other a single condition) are flagged. I tried other versions that also allowed for flagging multiple conditions in a single entry but the rewritten code was no clearer, so made it more conservative. It should be quite safe to enable by default but that can be changed later.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIfInsteadOfWhen.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIfInsteadOfWhen.kt
@@ -9,7 +9,7 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import org.jetbrains.kotlin.psi.KtWhenExpression
 
 /**
- * Binary expressions are better expressed using an 'if' expression than a 'when' expression.
+ * Binary expressions are better expressed using an `if` expression than a `when` expression.
  *
  * See https://kotlinlang.org/docs/reference/coding-conventions.html#if-versus-when
  *
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIfInsteadOfWhen.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIfInsteadOfWhen.kt
@@ -1,0 +1,49 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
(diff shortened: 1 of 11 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIfInsteadOfWhen.kt
```diff
@@ -0,0 +1,49 @@
...
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.psi.KtWhenExpression
+
+/**
+ * Binary expressions are better expressed using an 'if' expression than a 'when' expression.
```
schalkms: We could use the Markdown syntax (`if`, `when`) here.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIfInsteadOfWhenSpec.kt
```diff
@@ -0,0 +1,67 @@
...
+            assertThat(subject.compileAndLint(code)).hasSize(1)
+        }
+
+        it("does not report when using one branch") {
+            val code = """
+                fun function(): Boolean? {
+                    val x = null
+                    when (x) {
+                        else -> return false
+                    }
```
schalkms: IMO this rule should also flag cases, where there's only a single `else` statement.

3flex: Did consider that, however in that case the advice of changing to an `if` doesn't make sense since there's only a single condition. Both `if` and `when` are unsuitable.

Maybe a different rule? But I think this would only happen if a dev is very inexperienced so I don't think it provides much value.

schalkms: Please leave this rule as it is. I like it.
A rule that just lints this one case is not very useful, I agree.
Resolved..

PR #1905, opened 2019-09-09
Report PreferToOverPairSyntax only for kotlin.Pair

### LINKED ISSUE #1066: PreferToOverPairSyntax false positive
Android has it's own Pair class. https://developer.android.com/reference/android/util/Pair

That one you can't construct using `1 to 2`. Hence the following code should not get flagged:

```kotlin
android.support.v4.util.Pair(Location(0.0, 0.0), 200.0)
```

3flex:
Fixes #1066 

Rule now requires type resolution.

Blocked by #1880.

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/PreferToOverPairSyntaxSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/PreferToOverPairSyntaxSpec.kt
@@ -1,13 +1,19 @@
 package io.gitlab.arturbosch.detekt.rules.style.optional
 
 import io.gitlab.arturbosch.detekt.api.Config
-import io.gitlab.arturbosch.detekt.test.compileAndLint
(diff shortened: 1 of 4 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/PreferToOverPairSyntaxSpec.kt
```diff
@@ -17,12 +30,27 @@ class PreferToOverPairSyntaxSpec : Spek({
...
         }
 
         it("does not report if it is created using the to syntax") {
             val code = "val pair = 1 to 2"
-            assertThat(subject.compileAndLint(code)).hasSize(0)
+            assertThat(subject.compileAndLintWithContext(environment!!, code)).isEmpty()
         }
+
+        it("does not report if a non-Kotlin Pair class was used") {
+            val code = """val pair1 = Pair(1, 2)
```
schalkms: `val pair1 = Pair(1, 2)` should be in its own line.

PR #1913, opened 2019-09-13
Detect deprecations

### LINKED ISSUE #267: Rule: Usages of @Deprecated functions/classes
If we had access to the compile Context we could check every function call and check if the function or its containing class has been marked as `@Deprecated`.

Without the Context we could first collect a list of everything that is marked as `@Deprecated` and then check function calls for matches.

3flex:
Closes #267 

This rule uses the diagnostic warning built into the Kotlin compiler. Earlier discussion suggested that detekt [could reimplement compiler warnings](https://github.com/arturbosch/detekt/issues/240#issuecomment-407738733), so users have control over which compiler warnings will fail the build.

Implementing rules in this way implements checks for compiler warnings with minimal effort.

One issue is that suppressions for detekt compiler warnings are separated by underscore (like "UNCHECKED_CAST") while detekt uses CamelCase.

For rules that overlay compiler checks that have names with underscores we could match the rule ID so a single suppression (`@Suppress("UNCHECKED_CAST")`) would apply to both detekt and the compiler, but then the style of the rule ID differs from the rest of the rules which have CamelCase IDs.

We should pick one option and stick to it because there are many built in warnings that would be good to implement in detekt.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/Deprecation.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/Deprecation.kt
@@ -13,30 +13,6 @@ import org.jetbrains.kotlin.resolve.BindingContext
 
 /**
  * Deprecated elements are expected to be removed in future. Alternatives should be found if possible.
- *
- * <noncompliant>
- * \@Deprecated("deprecation message")
- * abstract class Foo {
- *   abstract fun bar() : Int
- *   fun baz() {}
- * }
- *
- * abstract class Oof : Foo() {
- *   fun spam() {}
- * }
- * </noncompliant>
- *
- * <compliant>
- * abstract class Foo {
- *   abstract fun bar() : Int
- *   fun baz() {}
- * }
- *
- * abstract class Oof : Foo() {
- *   fun spam() {}
- * }
- * </compliant>
- *
  */
 class Deprecation(config: Config) : Rule(config) {
 
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/Deprecation.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/Deprecation.kt
@@ -1,0 +1,57 @@
+package io.gitlab.arturbosch.detekt.rules.bugs
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.diagnostics.Errors
+import org.jetbrains.kotlin.resolve.BindingContext
+
+/**
+ * Deprecated elements are expected to be removed in future. Alternatives should be found if possible.
+ *
+ * <noncompliant>
+ * @Deprecated("deprecation message")
+ * abstract class Foo {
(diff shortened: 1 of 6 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/Deprecation.kt
```diff
@@ -0,0 +1,57 @@
...
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.diagnostics.Errors
+import org.jetbrains.kotlin.resolve.BindingContext
+
+/**
+ * Deprecated elements are expected to be removed in future. Alternatives should be found if possible.
+ *
+ * <noncompliant>
+ * \@Deprecated("deprecation message")
```
3flex: `@Deprecated("deprecation message")` causes generateDocumentation task to throw an error.

`\@Deprecated("deprecation message")` doesn't render correctly, though it does remove the syntax highlighting in IDE so `@Deprecated` isn't highlighted, so probably that syntax should be accepted by generateDocumentation task.

arturbosch: You need to use a html escape character for `@`.
I used it somewhere ... some rule with annotations I don't remember anymore :'D

arturbosch:
Looks good and super cool that we can just wrap compiler warnings!
We do have the `aliases` feature exactly for your concern.
Take a look at https://github.com/arturbosch/detekt/blob/master/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionNaming.kt#L29

This way you can suppress detekt issues with normal IDEA/kotlinc suppressions.

3flex:
Aliases work. In this case though the detekt and Kotlin suppression is effectively the same (Deprecation/DEPRECATION), so `@Suppress("Deprecation")` suppresses both (Kotlin is case insensitive). For Kotlin rules with underscores in them the detekt and Kotlin suppression will be different: might be UncheckedCast in detekt and UNCHECKED_CAST in Kotlin for example. User needs two suppressions. This might be surprising to users since behaviour would be inconsistent.

We should either match the suppression ID that Kotlin uses when wrapping rules, or make sure it's different, and do it consistently.

arturbosch:
Maybe its too early in the morning for me to think but isn't this also handled by the aliases feature?
The rule UncheckedCast has:
```
    override val defaultRuleIdAliases: Set<String> = setOf("UNCHECKED_CAST")

    override val issue: Issue = Issue("UnsafeCast",
```
So the user only needs to suppress the kotlin compiler warning and detekt will also be suppressed?
The difficulty here will be to try out different suppression order and realize that suppressing kotlinc first will also suppress detekt ;)?

3flex:
Understood - so if the rule is the same or very similar to the kotlinc rule, add a default alias like above. Thanks, that answers my question.

I'll find the correct KDoc string as mentioned above then merge.

arturbosch:
Tested it. IntelliJ autogenerates `@Suppress("DEPRECATION")` but also `@Suppress("deprecation")` and `@Suppress("Deprecation")` works.
Detekt needs the exact name, so we would add `DEPRECATION` here.

3flex:
Using `@Deprecated` in the noncompliant sample is maybe not possible at the moment... because generateDocumentation task parses KDoc top to bottom firstly looking for a comment then looking for KDoc tags, it parses up to to the line before `@Deprecated` and treats that as the entire comment, and parses `@Deprecated` as a tag then ignores the rest.

I tried HTML-escaping with `&commat;` which works here in markdown: &commat;, but that's not rendered when generating docs... support may have to be added.

Maybe this one can go in without code samples? It's self-explanatory and I don't want to fix this in the doc generation code because it's an edge case currently.

Unless you can find the other example? I searched all KDoc in detekt-rules but cannot find any line that starts with `@` that isn't a KDoc tag.

arturbosch:
Wow I thought our generator will crash when no examples are provided.
Okay for now, could you please open an issue for the @ + KDoc + generator thingy?

PR #1914, opened 2019-09-13
Don't fail on master when test coverage decreases

schalkms:
The build on master shouldn't fail if the test coverage drops slightly.
See commit: 7187493

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,3 +4,10 @@ codecov:
   notify:
     require_ci_to_pass: no
     after_n_builds: 4
+coverage:
+  status:
+    # master branch only
+    project:
+      default:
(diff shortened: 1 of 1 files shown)


arturbosch:
What's the reasoning behind reopening this?
Due to not writing duplicated tests for `BindingContext.empty`?

schalkms:
I didn't reopen the PR. I created a new one with a different codecov configuration.
Let's say we want to merge a PR, where the code coverage slightly goes down, into master.
This happened in 7187493. The build in the feature branch failed, because the code coverage went down. However, we don't want to fail the build on master too, if we decide to merge this feature branch resulting in a slightly lower coverage.

PR #1918, opened 2019-09-15
Deprecate debug in IDEA tasks

3flex:
These parameters don't do anything in the IDEA tasks, so are deprecated and can be removed in future.


schalkms:
This seems to break the current master build.

3flex:
The error on the log is confusing... Complains about missing class "org/jetbrains/kotlin/config/JvmTarget" but only on Java 8 on Travis.

I'll revert this then let's see what happens.

3flex:
Reverting didn't fix it, see the build error on #1924.

I'm closing that and opening a new issue for the build failure.

PR #1922, opened 2019-09-16
Update the JavaExec Gradle task documentation

cortinico:
This page is using some detekt-cli flags that have been removed.
I'm cleaning it up to make sure it actually works with the
current version of Detekt.

### DIFF
--- a/docs/pages/gettingstarted/gradletask.md
+++ b/docs/pages/gettingstarted/gradletask.md
@@ -26,9 +26,8 @@ task detekt(type: JavaExec) {
 	classpath = configurations.detekt
 	def input = "$projectDir"
 	def config = "$projectDir/detekt.yml"
-	def filters = ".*/build/.*,.*/resources/.*"
-	def rulesets = ""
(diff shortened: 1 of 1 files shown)


docs/pages/gettingstarted/gradletask.md
```diff
@@ -26,9 +26,8 @@ task detekt(type: JavaExec) {
 	classpath = configurations.detekt
 	def input = "$projectDir"
 	def config = "$projectDir/detekt.yml"
-	def filters = ".*/build/.*,.*/resources/.*"
-	def rulesets = ""
-	def params = [ '-i', input, '-c', config, '-f', filters, '-r', rulesets]
+	def exclude = ".*/build/.*,.*/resources/.*"
+	def params = [ '-i', input, '-c', config, '-ex', exclude]
```
schalkms: excludes + includes -in is missing

arturbosch: The gradle plugin does not have includes and excludes.
We do not want to add them because the `SourceTask` has already `include` and `exclude`.

PR #1924, opened 2019-09-17
Revert "Deprecate debug in IDEA tasks"

3flex:
Reverts arturbosch/detekt#1918

Apparently merging the PR broke the build, so will revert to see if that helps.


3flex:
Didn't resolve build failure. Opening a new issue to track that problem.

Issue #1917, opened 2019-09-14
Add proper attribution for rules derived from Kotlin/IntelliJ source


3flex:
A couple of rules are derived or copied entirely from Apache 2.0 licensed Kotlin and its plugin for IntelliJ.

These must be properly attributed by detekt.

arturbosch:
In KDoc which will also appear on the website?

schalkms:
The same holds true for `ktlint's` rules.

cortinico:
I believe this is essentially done as every rule is linked, and we kept on adding references to rules (e.g. compiler ones). Is there anyhting missing here?

3flex:
Agreed. New issues can be opened if there are rules that were missed or some other licensing or attribution issues need resolving.

PR #1938, opened 2019-09-19
HasPlatformType rule

### LINKED ISSUE #556: Rule: Require explicit types when using platform types
I watched "[Advanced Kotlin](http://www.oredev.org/2017/sessions/retrofitting-with-retrofit-tips-on-updating-an-old-networking-layer)" talk (relevant discussion starts around 17:00 in the linked video), and have an idea for a new check for when platform types are used. Example taken from [here](https://github.com/christinalee/oredev17/blob/d4349605518f858cc65f5445e09edba10fae7ead/src/1_Types.kt#L719-L776):

If you have this in Java (with no nullability annotations):
```
import java.util.ArrayList;
import java.util.List;

public class PlatformTypes {
    public static List<String> returnsNullList() {
        return null;
    }
}
```
and you call it from Kotlin:
```
fun errorWithoutExplicitType() {
    val list = returnsNullList()
    list.size
}
```
Then `list.size` throws an NPE because `list` is `null`, but this was unknown at compile time. The recommendation is to explicitly set the type on the declaration:
```
fun errorWithExplicitType() {
    val list: List<String> = returnsNullList()
    list.size
}
```
This way the error changes from NPE on `list.size` to an IllegalStateException at `val list: List<String> = returnsNullList()` with description like "list must not be null". This helps with debugging because the exception occurs when the value is assigned to the `list` property, not when `list.size` is called, which could happen several lines down. This wastes a developer's time when debugging because they have to trace back to the location `list` was declared.

detekt should check that when a property is assigned a value, and the expression resolves to a platform type, that the property has an explicit type, to improve the experience when working with platform types.

This also requires type and symbol resolving :)

3flex:
Closes #556 

Rule is quite conservative and picks up usage in public APIs only. Future enhancement could be to add a config option to detect implicit usage of platform types anywhere they're found.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/HasPlatformType.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/HasPlatformType.kt
@@ -47,8 +47,8 @@ class HasPlatformType(config: Config) : Rule(config) {
     override fun visitKtElement(element: KtElement) {
         super.visitKtElement(element)
 
-        if (bindingContext == BindingContext.EMPTY) return
-        if (element is KtCallableDeclaration && element.hasImplicitPlatformType()) {
+        if (bindingContext != BindingContext.EMPTY && element is KtCallableDeclaration &&
+            element.hasImplicitPlatformType()) {
             report(
                 CodeSmell(
                     issue,
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/HasPlatformType.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/HasPlatformType.kt
@@ -1,0 +1,77 @@
+package io.gitlab.arturbosch.detekt.rules.bugs
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.descriptors.CallableDescriptor
+import org.jetbrains.kotlin.psi.KtCallableDeclaration
+import org.jetbrains.kotlin.psi.KtElement
+import org.jetbrains.kotlin.psi.KtFunction
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.types.isFlexible
+
+/**
+ * Platform types must be declared explicitly in public APIs to prevent unexpected errors.
+ *
+ * Based on code from Kotlin project:
+ * https://github.com/JetBrains/kotlin/blob/1.3.50/idea/src/org/jetbrains/kotlin/idea/intentions/SpecifyTypeExplicitlyIntention.kt#L86-L107
+ *
+ * <noncompliant>
+ * class Person {
+ *   fun apiCall() = System.getProperty("propertyName")
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * class Person {
+ *   fun apiCall(): String = System.getProperty("propertyName")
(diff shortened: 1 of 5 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/HasPlatformType.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.types.isFlexible
+
+/**
+ * Platform types must be declared explicitly in public APIs to prevent unexpected errors.
+ *
+ * Based on code from Kotlin project:
+ * https://github.com/JetBrains/kotlin/blob/1.3.50/idea/src/org/jetbrains/kotlin/idea/intentions/SpecifyTypeExplicitlyIntention.kt#L86-L107
```
schalkms: I'm not sure if we should really reference the source location.
Source code can be changed easily. Thus, this reference could become out of date.
It's better IMHO to include the license in our project.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/HasPlatformType.kt
```diff
@@ -0,0 +1,77 @@
...
+        Severity.Maintainability,
+        "Platform types must be declared explicitly in public APIs.",
+        Debt.FIVE_MINS
+    )
+
+    override fun visitKtElement(element: KtElement) {
+        super.visitKtElement(element)
+
+        if (bindingContext == BindingContext.EMPTY) return
+        if (element is KtCallableDeclaration && element.hasImplicitPlatformType()) {
```
schalkms: The ifs can be merged easily.
```suggestion
        if (bindingContext != BindingContext.EMPTY && element is KtCallableDeclaration && element.hasImplicitPlatformType()) {
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/HasPlatformTypeSpec.kt
```diff
@@ -0,0 +1,74 @@
+package io.gitlab.arturbosch.detekt.rules.bugs
+
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.test.KtTestCompiler
+import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
+import org.assertj.core.api.Assertions.assertThat
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+
+object HasPlatformTypeSpec : Spek({
```
schalkms: Could we please make sure that the tests cover all branches (if possible)?

```Kotlin
if (containingClassOrObject?.isLocal == true) return false

val callable = bindingContext[BindingContext.DECLARATION_TO_DESCRIPTOR, this] as? CallableDescriptor ?: return false
if (!callable.visibility.isPublicAPI) return false
return callable.returnType?.isFlexible() ?: return false
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/HasPlatformType.kt
schalkms: relates to #1917

3flex: The url is fixed to v1.3.50 of Kotlin, so the link won't break.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/HasPlatformType.kt
```diff
@@ -0,0 +1,77 @@
...
+            is KtProperty -> if (isLocal || typeReference != null) return false
+            else -> return false
+        }
+
+        if (containingClassOrObject?.isLocal == true) return false
+
+        val callable =
+            bindingContext[BindingContext.DECLARATION_TO_DESCRIPTOR, this] as? CallableDescriptor ?: return false
+        if (!callable.visibility.isPublicAPI) return false
+        return callable.returnType?.isFlexible() ?: return false
```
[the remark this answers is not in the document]
3flex: https://kotlin.github.io/kotlin-spec/#flexible-types

PR #1942, opened 2019-09-20
Extend UseDataClass with the allowVars configuration property

### LINKED ISSUE #1920: UseDataClass conflicts with DataClassShouldBeImmutable

It should be possible to write a simple `class` with a mutable property that does not violate the rules `UseDataClass` and `DataClassShouldBeImmutable`. Most likely, this means that `UseDataClass` should not be reported for classes that would violate `DataClassShouldBeImmutable` if converted to `data class`.

## Context

Consider the following code:

```kotlin
class C(
    var strings: List<String>
)
```

This produces:

> UseDataClass - [C]

So I convert it to a `data class`:

```kotlin
data class C(
    var strings: List<String>
)
```

But then I get:

> DataClassShouldBeImmutable - [strings]

I don't see a way out of this cycle, so I had to resort to using `@Suppress`.

cortinico:
I've implemented the fix suggested in #1920.
Moreover I've cleaned up `UseDataClassPositive` and `UseDataClassNegative` to new use JSR223

Fixes #1920


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
```diff
@@ -14,12 +14,168 @@ class UseDataClassSpec : Spek({
...
+            }
         }
 
-        it("does not report invalid data class candidates") {
-            assertThat(subject.lint(Case.UseDataClassNegative.path())).isEmpty()
+        describe("does report class with vars and allowVars") {
+
+            it("does not report class with mutable constructor parameter") {
+                val code = """
+                    class DataClassCandidateWithVar(var i: Int)
```
schalkms: Could be defined in 1 line

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClass.kt
```diff
@@ -72,11 +74,20 @@ class UseDataClass(config: Config = Config.empty) : Rule(config) {
...
             val containsFunctions = functions.none { !defaultFunctionNames.contains(it.name) }
             val containsPropertyOrPropertyParameters = properties.isNotEmpty() || propertyParameters.isNotEmpty()
+            val containsVars = properties.any { it.isVar } || propertyParameters.any { it.isMutable }
             val containsDelegatedProperty = properties.any { it.hasDelegate() }
 
             if (containsFunctions && containsPropertyOrPropertyParameters && !containsDelegatedProperty) {
-                report(CodeSmell(issue, Entity.from(klass), "The class ${klass.nameAsSafeName} defines no " +
-                        "functionality and only holds data. Consider converting it to a data class."))
+                if (allowVars && containsVars) {
+                    // #1920 Allows users to relax this rule to allow vars.
```
schalkms: The comment could be spared, I think.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
schalkms: Could be defined in 1 line

PR #1945, opened 2019-09-24
Use built-in collectDescendantsOfType function

3flex:


arturbosch:
Hey, thanks for the good intention of us having to maintain less code.
In this case I would suggest reverting the PR.
We often do `collectDescendantsOfType<KtThrowExpression>()?.any()` in our rules. kotlinc's version return a list and ours returns a sequence. If we consider thousands of files and hundreds of rules this might be a performance penalty.

3flex:
I hadn't considered that - but it looks like there might be more appropriate functions from the Kotlin compiler that can be used in those cases, I'll create a new PR for that.

PR #1947, opened 2019-09-25
Fix SpacingBetweenPackageAndImports issue for scripts without packages

### LINKED ISSUE #1937: Bogus SpacingBetweenPackageAndImports violation for KTS without package
## Expected Behavior

`SpacingBetweenPackageAndImports` should not be reported if no `package` directive is present.

## Observed Behavior

`SpacingBetweenPackageAndImports` is reported for a KTS (Kotlin Script) file which has no `package` directive at all.

## Steps to Reproduce

- Clone https://github.com/heremaps/oss-review-toolkit and checkout the `detekt-issue-1937` branch.
- Run `./gradlew :spdx-utils:detekt`.
- Observe:

```
21 kotlin files were analyzed.
Ruleset: style - 5min debt
   SpacingBetweenPackageAndImports - [Build_gradle] at oss-review-toolkit/spdx-utils/build.gradle.kts:12:22
```

## Context

I'm trying to let *detekt* also check my Gradle Kotlin Script build files, not only main Kotlin files.

## Your Environment

* Version of Gradle used (if applicable): 5.6.2
* Operating System and version: Ubuntu 18.04 x64
* Link to your project (if it's a public repository): https://github.com/heremaps/oss-review-toolkit/tree/detekt-issue-1937

schalkms:
Both a Kotlin file and a Kotlin script can be declared with and without
a package declaration. This commit checks if the class even contains a
package declaration for both .kt and .kts files.
Closes #1937

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SpacingBetweenPackageAndImports.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SpacingBetweenPackageAndImports.kt
@@ -44,8 +44,10 @@ class SpacingBetweenPackageAndImports(config: Config = Config.empty) : Rule(conf
             Debt.FIVE_MINS)
 
     override fun visitKtFile(file: KtFile) {
-        containsClassOrObject = file.collectDescendantsOfType<KtClassOrObject>().any() == true
-        super.visitKtFile(file)
+        if (file.packageDirective != null) {
+            containsClassOrObject = file.collectDescendantsOfType<KtClassOrObject>().any() == true
+            super.visitKtFile(file)
+        }
     }
 
     override fun visitImportList(importList: KtImportList) {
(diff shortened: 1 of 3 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SpacingBetweenPackageAndImports.kt
```diff
@@ -44,8 +44,10 @@ class SpacingBetweenPackageAndImports(config: Config = Config.empty) : Rule(conf
             Debt.FIVE_MINS)
 
     override fun visitKtFile(file: KtFile) {
-        containsClassOrObject = file.collectDescendantsOfType<KtClassOrObject>().any() == true
-        super.visitKtFile(file)
+        if (file.packageDirective != null) {
+            containsClassOrObject = file.collectDescendantsOfType<KtClassOrObject>().any() == true
+            super.visitKtFile(file)
```
sschuberth: Just asking: Is it correct to include the `super`-call into the `if`, or should it maybe be called in any case?

schalkms: Thanks for the review.
In this case it's not necessary. I don't want to visit the import list for a Kotlin script file.

sschuberth: Ok, then all looks good to me, thanks!

PR #1949, opened 2019-09-25
Improve UnusedPrivateMember when it's related with parameters

BraisGabin:
If two parameters are called the same in two different functions and both are not used now we report them as two errors.

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
@@ -333,6 +333,25 @@ class UnusedPrivateMemberSpec : Spek({
             assertThat(subject.lint(code)).hasSize(1)
         }
 
+        it("reports two parameters if they are unused and called the same in different methods") {
+            val code = """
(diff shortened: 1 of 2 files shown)


BraisGabin:
Reading the code I found that the functions has the same problem but it's far more complicated to check. I'm going to create an issue to talk about that. So this PR is only to check the parameters.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
```diff
@@ -333,6 +333,25 @@ class UnusedPrivateMemberSpec : Spek({
...
+
+            private fun usedMethod(unusedParameter: Int): Int {
+                return 5
+            }
+
+            private fun usedMethod2(unusedParameter: Int) {
+                return 5
+            }
+        }
+        """
```
schalkms: nit: 4 missing spaces (indentation)

schalkms:
> Reading the code I found that the functions has the same problem but it's far more complicated to check. 

Please elaborate on that in a separate issue, so that the community can also understand this problem.

BraisGabin:
Fixed. Yes, this is ready to merge and I'll write the issue.

PR #1042, opened 2018-08-09
Fixed #1039 - Indentation and MaximumLineLength configuration paramet…

ljazgar:
…ers are now properly passed to KTLint rules.

Names of properties of EditorConfig passed to KTLint rules are now correct.
Different formatting rules doesn't overwrite EditorConfig any more.

PR #1953, opened 2019-09-26
ForbiddenMethod Rule

cortinico:
Add a rule to identify call site of forbidden methods
specified by the user in the Detekt config.

This rule is currently not using type and symbol solving, and is just doing a text comparison to check the method invocation. I'm looking for some feedbacks or ideas on how to potentially improve it. Ideally we should use a `org.jetbrains.kotlin.resolve.calls.CallExpressionResolver` but I haven't really managed to let it work.

Fixes #1042


cortinico:
Closing as I referenced the wrong Issue ID.

PR #1960, opened 2019-09-28
Convert tabs in multiline strings to 4 spaces

schalkms:
In the past this repo used tabs instead of 4 spaces for .kt files.
Ktlint doesn't convert tabs in multiline strings to spaces.
Please take a look at the following issue.
pinterest/ktlint#575

Therefore, the following command has been used for conversion.

find ./ -iname '*.kt' -type f -exec bash -c 'expand -t 4 "$0" | sponge "$0"' {} \;

* ./ is recursively searching the given directory
* -iname is a case insensitive match .kt files
* type -f finds only regular files (no directories, binaries or symlinks)
* -exec bash -c executes the following commands in a subshell for each file
* expand -t 4 expands all tabs to 4 spaces
* sponge soaks up standard input (from expand) and writes it to the same file

Source:
https://stackoverflow.com/a/43523362


3flex:
Thanks for doing this!

PR #1962, opened 2019-09-29
Refactor

3flex:
This started as a way to address [this comment](https://github.com/arturbosch/detekt/pull/1945#issuecomment-535421650), but I got a bit carried away looking for other things to try and improve.

Hopefully this makes the code a bit more readable - should be easy enough to review by commit, except for the last commit which is a bit of a catch-all.


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunction.kt
```diff
@@ -31,7 +32,7 @@ class UndocumentedPublicFunction(config: Config = Config.empty) : Rule(config) {
         }
     }
 
-    private fun KtNamedFunction.shouldBeDocumented() = isContainingClassPublic() && isPublicNotOverridden()
+    private fun KtNamedFunction.shouldBeDocumented() = containingClass().isPublic() && isPublicNotOverridden()
 
-    private fun KtNamedFunction.isContainingClassPublic() = containingClass().let { it == null || it.isPublic }
+    private fun KtClass?.isPublic() = this == null || isPublic
```
schalkms: Don’t we already have a corresponding extension function in a file.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CollapsibleIfStatements.kt
```diff
@@ -52,8 +53,8 @@ class CollapsibleIfStatements(config: Config = Config.empty) : Rule(config) {
             expression.`else` == null && expression.parent !is KtContainerNodeForControlStructureBody
 
     private fun hasOneKtIfExpression(expression: KtIfExpression): Boolean {
-        val statements = expression.then?.children?.filterNot { it is PsiComment }
-        return statements != null && statements.size == 1 && isLoneIfExpression(statements[0])
+        val statement = expression.then?.getChildrenOfType<KtExpression>()?.singleOrNull()
```
schalkms: `PsiComment` check is missing.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SpacingBetweenPackageAndImports.kt
```diff
@@ -68,7 +66,7 @@ class SpacingBetweenPackageAndImports(config: Config = Config.empty) : Rule(conf
             element is KtPackageDirective && element.text.isNotEmpty()
 
     private fun checkKtElementsDeclaration(importList: KtImportList) {
-        val ktElements = importList.siblings(withItself = false).toList().filterIsInstance<KtElement>()
+        val ktElements = importList.siblings(withItself = false).filterIsInstance<KtElement>().toList()
```
schalkms: Can we actually spare the `toList()` statement?

3flex: Can replace with `firstOrNull()` with some minor refactoring.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CollapsibleIfStatements.kt
3flex: Getting children only of type KtExpression will already exclude PsiComment.

I've added a couple more tests to test for cases where there are other statements intermingled with the if statements to make sure they're not picked up as possibly collapsible.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunction.kt
3flex: Not exactly, we have all (or almost all) modifiers as extension properties except `isPublic` which is provided by the compiler as an extension property.

I've rewritten `shouldBeDocumented()` to make the intent clearer.

PR #1969, opened 2019-10-02
Update LongParameterList rule to match documentation

anthonykeenan:
The documentation for the `LongParameterList` rule (in https://github.com/arturbosch/detekt/blob/master/docs/pages/documentation/complexity.md) states:

"Reports functions which have more parameters than a certain threshold (default: 6)."

But it currently also reports functions which have exactly that amount of parameters.

This PR modifies the LongParameterList rule so that it follows the description in the documentation (obviously an alternative is to change the documentation to match the implementation but this seems more natural considering the description of the parameter).


anthonykeenan:
Actually - the way the threshold parameter is used here is consistent with the other rules, so it's obviously a documentation confusion. Will close this PR and raise a doc fix when I get chance.

schalkms:
@anthonykeenan we'd love to merge this doc improvement. Thank you very much.

Issue #1779, opened 2019-07-22
Html Report add code preview


ZhouGongZaiShi:
## Expected Behavior
Html Report add an error code preview, it's like the html report of lint.

## Current Behavior
Html reports only source files and line number.  
If I want to know the specific error code, I also need to search, more cumbersome.

## Context
It's like the html report of lint. 
![lint html report](https://b2.bmp.ovh/imgs/2019/07/a382bf6102e95c09.png)
Add a block of code, this way I can have a rough preview.  
I'm sure there are other developers who have similar needs.  
Thank you!

schalkms:
Thanks for your feedback!
You are actually the first one, who brought that up.

One would need to adapt the html report template and the `HtmlOutputReport` class, which handles the HTML rendering.

https://github.com/arturbosch/detekt/blob/f98ccac9aa352805621dd6a19888bb472d776cef/detekt-cli/src/main/resources/default-html-report-template.html#L1

https://github.com/arturbosch/detekt/blob/f98ccac9aa352805621dd6a19888bb472d776cef/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputReport.kt#L16

@ZhouGongZaiShi would you mind to take a stab at this feature?

BraisGabin:
We can close this issue now.

schalkms:
Closed by d31e162

PR #1975, opened 2019-10-03
Add snippet code in html report

BraisGabin:
I'm trying to solve #1779

For get this done I added `kotlinx.html` library, I hope it's ok.

This implemetation is **really** basic. It creates things like this:

Captura de pantalla 2019-10-03 a las 17 34 13

Problems that I see and I want some feedback:
- ~I'm not using any prettify library to give a bit of color to the code~ out of scope
- ~I don't know what to test exactly~
- ~I'm opening the file directly from the path... This have problems, for example for testing because the files doesn't exist. The `if (file.exist())` is a workaround, we should look for a good solution~
- ~If the error is in more than one line I just highlight the first line.~
- For some rules the snippet has no sense at all so we are wasting scroll space. We should have something to decide if we want to show the snippet or not.

On the other hand I copy&pasted some html code from the android linter. I don't know how to point that out in the code.

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputReport.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputReport.kt
@@ -87,14 +87,14 @@ class HtmlOutputReport : OutputReport() {
         }
     }
 
-    private fun FlowContent.renderFinding(it: Finding) {
+    private fun FlowContent.renderFinding(finding: Finding) {
         span("location") {
-            text("${it.file}:${it.location.source.line}:${it.location.source.column}")
+            text("${finding.file}:${finding.location.source.line}:${finding.location.source.column}")
         }
 
-        if (it.message.isNotEmpty()) {
+        if (finding.message.isNotEmpty()) {
             br()
-            span("message") { text(it.message) }
+            span("message") { text(finding.message) }
         }
     }
 }
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputReport.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputReport.kt
@@ -4,6 +4,7 @@ import io.gitlab.arturbosch.detekt.api.Detektion
 import io.gitlab.arturbosch.detekt.api.Finding
 import io.gitlab.arturbosch.detekt.api.OutputReport
 import io.gitlab.arturbosch.detekt.api.ProjectMetric
+import io.gitlab.arturbosch.detekt.api.TextLocation
 import io.gitlab.arturbosch.detekt.cli.ClasspathResourceConverter
 import kotlinx.html.CommonAttributeGroupFacadeFlowInteractiveContent
 import kotlinx.html.FlowContent
@@ -96,6 +97,12 @@ class HtmlOutputReport : OutputReport() {
             br()
             span("message") { text(finding.message) }
         }
+
+        val psiFile = finding.entity.ktElement?.containingFile
+        if (psiFile != null) {
+            val lineSequence = psiFile.text.splitToSequence('\n')
+            snippetCode(lineSequence, finding.startPosition, finding.charPosition.length())
+        }
     }
 }
 
@@ -117,3 +124,5 @@ private class SUMMARY(
     false
 ),
     CommonAttributeGroupFacadeFlowInteractiveContent
+
+private fun TextLocation.length(): Int = end - start
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputFormatTest.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputFormatTest.kt
@@ -9,10 +9,17 @@ import io.gitlab.arturbosch.detekt.api.Location
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.SourceLocation
 import io.gitlab.arturbosch.detekt.api.TextLocation
+import io.gitlab.arturbosch.detekt.cli.CliArgs
+import io.gitlab.arturbosch.detekt.cli.runners.Runner
 import io.gitlab.arturbosch.detekt.test.TestDetektion
+import io.gitlab.arturbosch.detekt.test.resource
 import org.assertj.core.api.Assertions.assertThat
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
 
 class HtmlOutputFormatTest : Spek({
 
@@ -60,6 +67,19 @@ class HtmlOutputFormatTest : Spek({
             assertThat(result).contains("<span class=\"description\">A1</span>")
             assertThat(result).contains("<span class=\"description\">A2</span>")
             assertThat(result).contains("<span class=\"description\">A3</span>")
+        }
+
+        it("integration") {
+            val result = outputFormat.render(createTestDetektionWithMultipleSmells())
+
+            val tmpReport = Files.createTempFile("HtmlOutputFormatTest", ".html")
+            Files.write(tmpReport, result.toByteArray())
+
+            try {
+                assertThat(tmpReport).hasSameContentAs(Paths.get(resource("/reports/HtmlOutputFormatTest.html")))
+            } finally {
+                Files.delete(tmpReport)
+            }
         }
     }
 })
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputFormatTest.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputFormatTest.kt
@@ -67,7 +67,7 @@ class HtmlOutputFormatTest : Spek({
             assertThat(result).contains("<span class=\"description\">A3</span>")
         }
 
-        it("integration") {
+        it("assert that the html generated is the expected") {
             val result = outputFormat.render(createTestDetektionWithMultipleSmells())
 
             val tmpReport = Files.createTempFile("HtmlOutputFormatTest", ".html")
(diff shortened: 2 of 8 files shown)


schalkms:
First all, thank you for picking this up and for formulating your ideas in detail! That's an awesome contribution.
I'll try to answer your questions in the following lines.

> I'm not using any prettify library to give a bit of color to the code

That's fine for version 1. I think you meant syntax coloring, right. There's probably a library that does syntax coloring for Kotlin.

> I don't know what to test exactly

I haven't reviewed it in detail. `FlowContent.snippetCode()` and `FlowContent.writeErrorLine()` should be internal util functions that are tested with unit tests. 
Furthermore, (almost) every branch should be covered to maintain the code coverage.
There should also be 1 integration test that tests the HTML report module.

> I'm opening the file directly from the path

The file writing process happens in the abstract base class called `OutputReport`. The derived class `HtmlOutputReport` should not implement this for testability reasons.

> If the error is in more than one line I just highlight the first line.

This should be tested thoroughly. There are many cases that could be covered by unit tests.

- Reported at the beginning of the file.
- Reported at the end of the file.
- ...

> For some rules the snippet has no sense at all so we are wasting scroll space. We should have something to decide if we want to show the snippet or not.

I'm not sure about that one. I'll ask @arturbosch to provide his expertise.

> On the other hand I copy&pasted some html code from the android linter. I don't know how to point that out in the code.

I don't quite know what you mean by that. Could you please explain that?

> This implemetation is really basic. It creates things like this:

Does your report create html files similar to the provided screenshot?

BraisGabin:
> That's fine for version 1. I think you meant syntax coloring, right. There's probably a library that does syntax coloring for Kotlin.

Ok, let's keep this out of this PR

> The file writing process happens in the abstract base class called `OutputReport`. The derived class `HtmlOutputReport` should not implement this for testability reasons.

I didn't explain myself properly. The problem is that the `Finding`s don't contain the snippet that I need (and that's ok). So I need to open the file to read the data that I need. Each `Finding` has the file path so I can do it. My point is that we should inject something to `OutputReport` to open this files. Something similar as: https://android.googlesource.com/platform/tools/base/+/master/lint/cli/src/main/java/com/android/tools/lint/HtmlReporter.java#93

What do you think?

> Does your report create html files similar to the provided screenshot?

Yes, it's a screenshot of a detekt report with this PR. And that screenshot is a bit old. Now it doesn't underline the spaces. And with #1977 it will not underline the `,` neither

BraisGabin:
CI should pass as soon as #1971 is merged

schalkms:
> CI should pass as soon as #1971 is merged

No, it won't necessarily, since CodeFactor updates detekt's dependencies on their own.
Unfortunately, you have to fix the issue by your own for now.

arturbosch:
Cool PR, thanks!
Regarding opening the file again and deciding for rules to print the location or not:
What rules do you have in might which should not print the code snippet?

For not opening the file:
A `Finding` has an `entity.ktElement` which is nullable. `KtElement.containingFile` would return the `KtFile` which has the `text` method. This could bring the desired text content.
If `entity.ktElement` is per se a `KtFile` that must be like a formatting rule or something ike MaxLineLength. I do not know by hard if KtElement can actually be null now or if we report everytime a KtElement.
However the reporting logic should handle the `null` case.

BraisGabin:
🎉 I don't need to open files anymore!

Examples of rules where the snippet is not usefull:
- TooManyFunctions
- ComplexMethod
- ComplexInterface
- LargeClass
- NestedBlockDepth
- ...

Do something with the lenght of textLocation could help but we would have some false positives. For example `ClassNaming`. The error is only the name, so the snippet is helpful but the finding show ALL the class as an error.

Maybe we should ignore this for now and try to fix the textLocation. The problem of fixig this is that we can be broking some baselines... Any idea?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputReport.kt
```diff
@@ -24,33 +38,91 @@ class HtmlOutputReport : OutputReport() {
...
+        text(
+            line.substring(
+                startIndex = errorStarts,
+                endIndex = errorEnds
+            )
+        )
+    }
+    text(line.substring(startIndex = errorEnds))
+    return errorEnds - errorStarts
+}
```
schalkms: I'd add this into a separate util file in the same package.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputFormatTest.kt
```diff
@@ -33,33 +35,191 @@ class HtmlOutputFormatTest : Spek({
...
+                // reports 1 - line with just one tab
+
+                // reports 1
+                fun myFunction() {
+                    // reports 1 - line with 1 trailing tab
+                    println("A message")
+                // reports 1
+                }
+            }
+        """.trimIndent().splitToSequence('\n')
```
schalkms: These test code tests too much. It's hard to maintain. 
I think for testing the HTML code snippet, 1 or 2 violations should be sufficient.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputFormatTest.kt
```diff
@@ -33,33 +35,191 @@ class HtmlOutputFormatTest : Spek({
...
+                <span class="lineno">   5 </span>// A comment
+                <span class="lineno">   6 </span>// reports 1
+                <span class="lineno">   7 </span><span class="error">class TrailingWhitespacePositive {</span>
+                <span class="lineno">   8 </span>    // reports 1 - line with just one tab
+                <span class="lineno">   9 </span>
+                <span class="lineno">  10 </span>    // reports 1
+                </code></pre>
+                </div>
+
+                """.trimIndent()
```
schalkms: These asserts test more too much. If we add 1 or 2 violations in the `snippet`, this and the following asserts should become easier to read and maintain.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputFormatTest.kt
```diff
@@ -33,33 +35,191 @@ class HtmlOutputFormatTest : Spek({
...
-            assertThat(result).contains("<span class=\"description\">\nA1\n</span>")
-            assertThat(result).contains("<span class=\"description\">\nA2\n</span>")
-            assertThat(result).contains("<span class=\"description\">\nA3\n</span>")
+            assertThat(result).contains("<span class=\"description\">A1</span>")
+            assertThat(result).contains("<span class=\"description\">A2</span>")
+            assertThat(result).contains("<span class=\"description\">A3</span>")
+        }
+    }
+
+    describe("HTML snippet code") {
```
schalkms: I would also move this `describe` blog into its own util test file.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputFormatTest.kt
BraisGabin: I'm not sure if I understood you. The production code writes one snippet for each finding (fidnig = violation?) so it's impossible to add multiple violations to one single snippet.

Anyway, I agree that the problem of this test is the maintainability. I changed the sample source code onces and I need to change the line numbers in all the assertions. But I don't like the tests that are in this file already. They test nearly nothing. If you, or anyone, has an idea to simplify them but not losing its "power" I would like to improve them.

schalkms: @BraisGabin As mentioned, I think 1 integration test that assert the dynamic (generated) parts of the HTML report, could be a good addition to this test group.
So the test checks character by character if the dynamic generated content matches the expected one. The expected string could be put into the `resources` folder in the `src/test` package.
Does this answer your question?

BraisGabin: I think so. So you say that this tests instead of adding all that `"""code"""` just create files in the `resources` folder and check that the output of the function is the same as the content in those files. Right?

Add add an integration test too.

schalkms: No, this code is fine here. the val code = "..." should be simplified to contain 2 issues. Then the asserts also becomes much simpler by checking only the HTML content for these 2 issues.
Th3 separate resource is just for the integration test that asserts the full HTML content.

BraisGabin: `val core = "..."` contains 0 errors right now. I have all those tests because they test different cases:

Check that the underline is correct:
1. Underline all the line
2. Underline part of the line
3. Underline multiple lines

Show the correct lines:
1. If the error is in the first line I show lines 1 to 4.
2. If the error is in the first line I show lines 1 to 5.
3. If the error is in the line 7 I show lines 4 to 10 (important to show the change between one digit and 2 digits to be sure that the alignment works, this one is checked)
4. If the error is in the penultimate line I show lines 14 to 16.
5. If the error is in the last line I show lines 13 to 16.

The case 3 is checked implicitly in the top ones.

Maybe the ones that check the lines that are showed can be simplified to just: `contains("1", "2", "3", "4")`. Would that work?

BraisGabin:
Ok, I changed the integration test. Now I'm *mocking* the `KtElement` so this test will no broke so often.

By my side this PR is ready. I have ideas for future improvements but I think that this is OK as first step.

Are the tests now ok for you?

arturbosch:
>  I don't need to open files anymore!
> 
> Examples of rules where the snippet is not usefull:
> 
> * TooManyFunctions
> * ComplexMethod
> * ComplexInterface
> * LargeClass
> * NestedBlockDepth
> * ...
> 
> I was thinking about doing something with the lenght of textLocation. If it's larger than X don't show the snippet but we would have some false positives. For example `ClassNaming`. The error is only the name, so the snippet is helpful but the finding show ALL the class as an error.
> 
> Maybe we should ignore this for now and try to fix the textLocation. The problem of fixig this is that we can break some baselines... Any idea?

Sorry, for the late reply. For findings concerning a function or class I think we would do the user a favor to print just the function and class headers.

`class A(d: D) : B, C by d` and `fun largeFunction(....)`

This should be useful?

BraisGabin:
>Sorry, for the late reply. For findings concerning a function or class I think we would do the user a favor to print just the function and class headers.
>
>class A(d: D) : B, C by d and fun largeFunction(....)
>
>This should be useful?

I agree. In that case we can move that work to other PRs fixing the `TextLocation`. So this PR is ready to be merged.

arturbosch:
I will just wait for @schalkms last review :)

BraisGabin:
I simplified the tests that were related with the number of lines. The others I insist that they are needed. They were necesary to develop the feature.

PR #1977, opened 2019-10-04
Fix EnumNaming textLocation

BraisGabin:
I'm finding some problems with the `textLocation` thanks to #1975. This is one of them.

I added two tests becase my first try to fix this bug broke that other part.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNaming.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNaming.kt
@@ -30,7 +30,7 @@ class EnumNaming(config: Config = Config.empty) : Rule(config) {
         if (!enumEntry.identifierName().matches(enumEntryPattern)) {
             report(CodeSmell(
                     issue,
-                    Entity.from(enumEntry),
+                    Entity.from(enumEntry.nameIdentifier ?: enumEntry),
                     message = "Enum entry names should match the pattern: $enumEntryPattern"))
         }
     }
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNamingSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNamingSpec.kt
@@ -27,5 +27,16 @@ class EnumNamingSpec : Spek({
                 }"""
             assertThat(NamingRules().compileAndLint(code)).hasSize(1)
         }
+
+        it("reports the correct text location in enum name") {
+            val code = """
+                enum class WorkFlow {
+                    _Default,
+                }"""
+            val findings = NamingRules().compileAndLint(code)
+            val source = findings.first().location.text
+            assertThat(source.start).isEqualTo(26)
+            assertThat(source.end).isEqualTo(34)
+        }
     }
 })
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ParameterNamingSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ParameterNamingSpec.kt
@@ -33,6 +33,16 @@ class ParameterNamingSpec : Spek({
             """
             assertThat(NamingRules().lint(code)).hasSize(5)
         }
+
+        it("should find a violation in the correct text locaction") {
+            val code = """
+                class C(val PARAM: String)
+            """
+            val findings = NamingRules().lint(code)
+            val source = findings.first().location.text
+            assertThat(source.start).isEqualTo(8)
+            assertThat(source.end).isEqualTo(25)
+        }
     }
 
     describe("parameters in a function of a class") {


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNaming.kt
```diff
@@ -30,7 +30,7 @@ class EnumNaming(config: Config = Config.empty) : Rule(config) {
         if (!enumEntry.identifierName().matches(enumEntryPattern)) {
             report(CodeSmell(
                     issue,
-                    Entity.from(enumEntry),
+                    Entity.from(enumEntry.identifyingElement!!),
```
schalkms: I don't understand why the `enumEntry` should report the wrong text location.
Could you please explain that?

schalkms: I don't understand why the `enumEntry` should report the wrong text location.
Could you please explain that?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNamingSpec.kt
```diff
@@ -27,5 +28,13 @@ class EnumNamingSpec : Spek({
                 }"""
             assertThat(NamingRules().compileAndLint(code)).hasSize(1)
         }
+
+        it("reports the correct text location in enum name") {
+            val code = """
+                enum class WorkFlow {
+                    _Default,
+                }"""
+            assertThat(NamingRules().compileAndLint(code)[0].charPosition).isEqualTo(TextLocation(26, 34))
```
schalkms: This could be made simpler by using the following code snippet.

```Kotlin
val findings = NamingRules().compileAndLint(code)
val source = findings.first.location.source
source.line
source.column
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ParameterNamingSpec.kt
```diff
@@ -33,6 +34,13 @@ class ParameterNamingSpec : Spek({
             """
             assertThat(NamingRules().lint(code)).hasSize(5)
         }
+
+        it("should find a violation in the correct text locaction") {
+            val code = """
+                class C(val PARAM: String)
+            """
+            assertThat(NamingRules().lint(code)[0].charPosition).isEqualTo(TextLocation(8, 25))
```
schalkms: Same here

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNaming.kt
BraisGabin: Sure. If I use enumEntry the selected code is `_Default,` (I'm using te test as an example). But we don't want to select the `,` for that reason I'm using `identifyingElement`. But, as I said, I don't know this API and I can't find the documentation so I can't be sure about it.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNamingSpec.kt
BraisGabin: I just find the `FindingsAssert` class. I'll add a function there and use it.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNaming.kt
schalkms: Honestly, this is weird. `originalElement` did the trick for me.

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt
```diff
@@ -56,6 +57,21 @@ class FindingsAssert(actual: List<Finding>) :
             failWithMessage("Expected source locations to be ${expectedSources.toList()} but was ${actualSources.toList()}")
         }
     }
+
+    fun hasTextLocation(vararg expected: TextLocation) = apply {
```
schalkms: The helper function is a good idea.

schalkms: However, it can be inlined with just 2 asserts on the
`source.line`
`source.column`
as we've done it in all other tests.
This refactoring with a helper function should then maybe done globally for all rules-tests.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNaming.kt
BraisGabin: With `enumEntry.originalElement` it doesn't pass the test for me (it has the same problem as `enumEntry`)

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt
BraisGabin: Done

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNaming.kt
schalkms: Yes, I also tried it out with different properties. None of them worked.
The problem with `identifyingElement` is that this API is experimental and I don't really want to have this in production code.
Furthermore, this would also potentially break detekt's [baseline](https://arturbosch.github.io/detekt/baseline.html) feature for this rule.
Unfortunately, I have to ask you to revert this line and adapt the test case. However, I'd really like to have the added tests in the productive code, since we are in the process of adding more tests like these.

I know that reporting the enum including the separator `,` feels weird, but it's not the end of the world.

BraisGabin: If the problem is the experimental we could use `nameIdentifier`. It works too and it's stable. But as I said, without the documentation I don't know the difference between them

Anyway, I checked the baseline problem and you are right. This breaks the previous generated baselines.

Related with the tests: Do you mean to add this test as "Ignored" or something like this? To note that this is wrong but we don't change it for the breaking change? Or adding a test that assume that adding the `,` is ok?

In the first case I can refactor this PR to do that. If it's the second I think that an issue to talk about how we want the rule test before doing the changes would be great. I'm open to help with that but what should we test? Number of findings? sourcePostion? textPosition?...

schalkms: > If the problem is the experimental we could use nameIdentifier. It works too and it's stable.

I'd go this route.

BraisGabin:
Ok, so now I'm not using an unstable api. And I remove the `!!` opetator. If `enumEntry.nameIdentifier` return `null` we fallback to `enumEntry`.

Anyway we have the problem of breaking the baselines...

PR #1978, opened 2019-10-04
Findings assertions

BraisGabin:
This PR adds the functions `hasTextLocation` to be used in the tests. And refactor other point of the tests to use the assertion (already created) `hasSourceLocations`.

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLengthSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLengthSpec.kt
@@ -1,10 +1,11 @@
 package io.gitlab.arturbosch.detekt.rules.style
 
+import io.gitlab.arturbosch.detekt.api.SourceLocation
 import io.gitlab.arturbosch.detekt.rules.Case
 import io.gitlab.arturbosch.detekt.test.TestConfig
+import io.gitlab.arturbosch.detekt.test.assertThat
 import io.gitlab.arturbosch.detekt.test.compileContentForTest
 import io.gitlab.arturbosch.detekt.test.compileForTest
-import org.assertj.core.api.Assertions.assertThat
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
 
@@ -183,9 +184,7 @@ class MaxLineLengthSpec : Spek({
 
                 rule.visit(fileContent)
                 assertThat(rule.findings).hasSize(1)
-                val findingSource = rule.findings[0].location.source
-                assertThat(findingSource.line).isEqualTo(6)
-                assertThat(findingSource.column).isEqualTo(17)
+                assertThat(rule.findings).hasSourceLocations(SourceLocation(6, 17))
             }
         }
     }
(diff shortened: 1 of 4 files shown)


schalkms:
Thanks for this PR. There are many more cases in the rule test package, where we check for `line` and `column`. So we could also use the new API in these cases.

I think the new function is a bit too over-engineered.
The following check should suffice.

```Kotlin
assertThat(findingSource.line).isEqualTo(x)		
assertThat(findingSource.column).isEqualTo(y)
```

BraisGabin:
> There are many more cases in the rule test package, where we check for line and column. So we could also use the new API in these cases.

I can't find those, could you point some of them? The only one that I didn't change is `TrailingWhitespaceSpec`. But this is because I created #1980.

> I think the new function is a bit too over-engineered.

It's not over-engenieered. The code is not mine, it's just a copy&paste from the function on top of this one. And if you check #1980 you will see that I'm using what this function is providing me.

schalkms:
> I can't find those, could you point some of them?

Sure.
`MaxLineLengthSpec`
`StringLiteralDuplicationSpec`

> It's not over-engenieered. The code is not mine, it's just a copy&paste from the function on top of this one. And if you check #1980 you will see that I'm using what this function is providing me.

I guess the function on the bottom is not suitable in this case, right?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLengthSpec.kt
```diff
@@ -183,9 +184,7 @@ class MaxLineLengthSpec : Spek({
 
                 rule.visit(fileContent)
                 assertThat(rule.findings).hasSize(1)
-                val findingSource = rule.findings[0].location.source
-                assertThat(findingSource.line).isEqualTo(6)
-                assertThat(findingSource.column).isEqualTo(17)
+                assertThat(rule.findings).hasSourceLocations(SourceLocation(6, 17))
```
arturbosch: I'm wondering if we should have a `hasSourceLocation(int, int)` function for the common case.
`hasSourceLocations(SourceLocation...)` sounds duplcicated. How about we accept a pair in the test code and transform it to a SourceLocation (`hasSourceLocations(6 to 17, ...)`)?

BraisGabin: Sounds good! I'm on it!

BraisGabin: I editted `hasSourceLocations` to accept Pairs... But now I'm thinking what's better:
- `hasSourceLocations(6 to 17, ...)`
- `hasSourceLocations(6..17, ...)`

About `hasSourceLocation` it's a bit strage if we are asserting over a list. I think is better to have `hasSourceLocation` than do something like: `assertThat(rule.findings.fisrt()).hasSourceLocation(SourceLocation(6, 17))`

arturbosch: Yep, right. This should be 99% of the times the case (one location).

BraisGabin:
Done

BraisGabin:
I rebased the PR to force a new CI.

Issue #1979, opened 2019-10-04
TrailingWhitespace reports incorrect line number


BraisGabin:
```kotlin
package cases

// reports 1 - line with just one space
 
// reports 1 - a comment with trailing space
// A comment 
// reports 1
class TrailingWhitespacePositive { 
	// reports 1 - line with just one tab
    
    // reports 1
	fun myFunction() { 
		// reports 1 - line with 1 trailing tab
		println("A message")	
	// reports 1
	}  
}
```

## Expected Behavior
```kotlin
package cases

// reports 1 - line with just one space
 
// reports 1 - a comment with trailing space
// A comment 
// reports 1
class TrailingWhitespacePositive { 
	// reports 1 - line with just one tab
    
    // reports 1
	fun myFunction() { 
		// reports 1 - line with 1 trailing tab
		println("A message")	
	// reports 1
	}  
}
```

This file should report 6 Trailing errors in the positions:

(4, 1), (6, 13), (8, 35), (10, 1), (12, 23) and (14, 29)

## Observed Behavior
It reports 7 and in other positions.

## Steps to Reproduce
You can check this comparison: https://github.com/arturbosch/detekt/compare/master...BraisGabin:trailing-whitespace

## Context
I found this problem developing the snippet html report. Screenshot:

<img width="651" alt="Captura de pantalla 2019-10-04 a las 23 53 41" src="https://user-images.githubusercontent.com/721244/66243351-94921d80-e704-11e9-8ed8-ebbb8e3e8385.png">

## Your Environment

* Version of detekt used: 1.0.1
* Operating System and version: Mac

PR #1980, opened 2019-10-04
TrailingWhitespace - Failing test to demostrate #1979

BraisGabin:
This PR demostrates the error reported at #1979


3flex:
Thanks for the example - since this isn't a PR that we can merge as is it would be better to link to the example branch/commit in your repo directly from the issue itself.

I'll close this and add that detail to the original PR.

PR #1984, opened 2019-10-05
Update baseline

BraisGabin:
I was trying to add something else to the baseline (unsuccesfully) and I found that this line in the baseline is not used.


3flex:
Thanks for this, it's actually still an issue in the generated file but it's not picked up using the default `detekt` task so it can be removed from the baseline. I'll submit a PR to actually fix it.

PR #1991, opened 2019-10-05
Reimplement parallelism internal logic

arturbosch:
From the changelog:
```
- `parallel=true` and `--parallel` now effect both the compilation and analysis phase. 
- Users of Gradle's `--parallel` flag are encouraged to turn off the parallelism of detekt. Or turn it on otherwise.
- detekt does not use the `ForkJoinPool.commonPool()` anymore internally. When embedding detekt feel free to pass your own `ExecutionService` to the `ProcessingSettings`.
```

Benchmarks:
detekt's parallelism is notated as p=false|true, gradles parallelism is notated as --p=false|true

Gradle Plugin run on detekt's codebase, ~40k lines:
1.1 clean buid --p=false p=false | 2.38s, 2.27s
1.2 clean build --p=false p=false | 2.30s, 2.25s

1.1 clean build --p=true p=false | 1.55s, 1.46s
1.1 clean build --p=true p=true | 1.53s, 1.47s
1.2 clean build --p=true p=false | 1.48s, 1.46s
1.2 clean build --p=true p=true | 1.46s, 1.45s

CLI on detekt's codebase, ~40k lines:
1.1 p=false | ~4.14
1.2 p=false | ~4.17
1.1 p=true | ~2.85
1.2 p=true | ~2.80

Conclusion:
There are no real performance benefits for a middle large project like detekt.
However there still could be some benefits for really large multi module Gradle projects when using Gradle's `--parallel` and using detekt's `parallel=false`.
Imo the code looks cleaner now and there is the possibility now to just use one thread.
We also do not use the CommonThreadPool and do not make him starve.


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -59,8 +59,9 @@ class CliArgs : Args {
 
     @Parameter(
         names = ["--parallel"],
-        description = "Enables parallel compilation of source files." +
-            " Should only be used if the analyzing project has more than ~200 kotlin files."
+        description = "Enables parallel compilation and analysis of source files." +
+            " Do some benchmarks first before enabling this flag." +
+            " Heuristics show performance benefits starting from 2000 lines of kotlin code."
```
schalkms:
```suggestion
            " Heuristics show performance benefits starting from 2000 lines of Kotlin code."
```

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Detektor.kt
```diff
@@ -55,4 +75,10 @@ class Detektor(
             .toList()
             .map { ruleSet -> ruleSet.id to ruleSet.accept(this, bindingContext) }
             .toMergedMap()
+
+    private fun createErrorMessage(file: KtFile, error: Throwable): String =
+        "Analyzing '${file.absolutePath()}' led to an exception.\n" +
+            "The original exception message was: ${error.localizedMessage}\n" +
+            "Running detekt '${whichDetekt()}' on Java '${whichJava()}' on OS '${whichOS()}'.\n" +
+            "If the exception message does not help, please feel free to create an issue on our github page."
```
schalkms:
```suggestion
            "If the exception message does not help, please feel free to create an issue on our GitHub page."
```

PR #1992, opened 2019-10-06
Upgrade Spek to v2.0.8

schalkms:

### DIFF
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReportSpec.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReportSpec.kt
@@ -13,7 +13,7 @@ class FileBasedFindingsReportSpec : Spek({
 
     describe("findings report") {
 
-        describe("reports the debt per file and rule set with the overall debt") {
+        context("reports the debt per file and rule set with the overall debt") {
             val expectedContent = readResource("grouped-findings-report.txt")
             val detektion = object : TestDetektion() {
                 override val findings: Map<String, List<Finding>> = mapOf(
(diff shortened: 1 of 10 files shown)


3flex:
I'm guessing the failure is due to a behaviour change in Spek meaning `described` nested in `describe` fails where it didn't before.

Changing `describe` to `context` here should fix it: https://github.com/arturbosch/detekt/blob/1c93b1e9def6054fb17c26c7655ff31ad9bcda7a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReportSpec.kt#L16

schalkms:
@3flex I'm on it. That's only a part of the problem.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/console/FileBasedFindingsReportSpec.kt
```diff
@@ -35,7 +35,7 @@ class FileBasedFindingsReportSpec : Spek({
                     Pair("EmptySmells", emptyList())
                 )
             }
-            val output = subject.render(detektion)?.trimEnd()?.decolorized()
+            val output by memoized { subject.render(detektion)?.trimEnd()?.decolorized() }
```
3flex: I think it's better to put this line in the `it()` context... the Spek docs say not to use memoized for holding the result of an action https://www.spekframework.org/core-concepts/#scope-values

So instead:
```kotlin
            it("has the reference content") {
                val output = subject.render(detektion)?.trimEnd()?.decolorized()
                assertThat(output).isEqualTo(expectedContent)
```
Will that work?

schalkms: With the updated commit with `lint`/`compileAndLint` for each and every test case, which is not really what we want.

schalkms: With the updated commit we `lint`/`compileAndLint` the code for each and every test case, which is not really what we want.
BeforeEachGroup also doesn't work.

3flex: Another option is putting all assertions into a single test case for the tests that assert multiple things for each test file.

I think what you have here now though is ok. This problem will go away when test cases are moved out of the external files and into individual test cases.

PR #1994, opened 2019-10-06
Allow detekt findings to be suppessed with rule set id - Closes #766

### LINKED ISSUE #766: @Suppress("Detekt.ruleset") feature
I have this function that I use for testing quite often:

```kotlin
inline fun <reified T> assertThrows(block: () -> Unit): AbstractThrowableAssert<*, out Throwable> {
  try {
    block()
  } catch (e: Throwable) {
    if (e is T) {
      return assertThat(e)
    } else {
      throw e
    }
  }

  throw AssertionError("Expected ${T::class.simpleName}")
}
```

Detekt is going wild on it and flagging all of these: `TooGenericExceptionCaught`, `InstanceOfCheckForException`, `RethrowCaughtException`.

Instead of `@Suppress("Detekt.TooGenericExceptionCaught", "Detekt.InstanceOfCheckForException", "Detekt.RethrowCaughtException")` - I know `Detekt` is optional but I like to see where things are coming from - it would be cool if I could just do `@Suppress("Detekt.exceptions")`.

arturbosch:


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/MultiRule.kt
```diff
@@ -20,7 +20,7 @@ abstract class MultiRule : BaseRule() {
 
     override fun postVisit(root: KtFile) {
         for (activeRule in activeRules) {
-            report(activeRule.findings, activeRule.aliases)
+            report(activeRule.findings, activeRule.aliases, activeRule.ruleId)
```
arturbosch: Todo: RuleSetId should be used.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
```diff
@@ -45,6 +46,8 @@ abstract class Rule(
      */
     open val defaultRuleIdAliases: Set<String> = emptySet()
 
+    internal val ruleSetId: RuleId? get() = (ruleSetConfig as? HierarchicalConfig)?.parent?.key
```
arturbosch: This is the real change in this PR. Now that we can track where each config property is coming from, we can use this info to get the `ruleSetId`.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Suppressions.kt
```diff
@@ -0,0 +1,55 @@
...
+private val detektSuppressionPrefixRegex = Regex("(?i)detekt([.:])")
+private const val QUOTES = "\""
+private val suppressionAnnotations = setOf("Suppress", "SuppressWarnings")
+
+/**
+ * Checks if this kt element is suppressed by @Suppress or @SuppressWarnings annotations.
+ */
+fun KtAnnotated.isSuppressedBy(id: RuleId, aliases: Set<String>, ruleSetId: RuleSetId? = null): Boolean {
+    val acceptedSuppressionIds = mutableSetOf(id, "ALL", "all", "All")
+    ruleSetId?.let { acceptedSuppressionIds.addAll(listOf(ruleSetId, "$ruleSetId.$id", "$ruleSetId:$id")) }
```
arturbosch: This is the second real change in this PR. We add the `ruleSetId` with its variations like `detekt.RuleSet.Rule` to the accepted content of a `@Suppress`

schalkms: Why did you add `$ruleSetId:$id` twice?

detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/SuppressionSpec.kt
```diff
@@ -98,6 +100,44 @@ internal class SuppressionSpec : Spek({
...
+            assertThat(findings).isNotEmpty()
+        }
+
+        fun assertCodeIsSuppressed(code: String) {
+            val findings = TestLPL(config).lint(code)
+            assertThat(findings).isEmpty()
+        }
+
+        it("suppresses by rule set id") {
+            assertCodeIsSuppressed("""@Suppress("complexity")$code""")
```
schalkms: I would add 1 negative case in order to test, whether code is not getting suppressed.

```Kotlin
"""@Suppress("bugs")$code"""
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Suppressions.kt
arturbosch: It is actually `:` and `.` because we also allow `detekt.Rule` and `detekt:Rule`

Issue #516, opened 2017-10-22
Fail early when configuration file has incorrect entries


vanniktech:
This would have caught the typo in `LateInitUsage` really early. In addition upgrading from RC's to the final version where we'll rename some of the rules would automatically notify the users to adjust their configuration files.

As a side note, comments should be ignored while validating.

schalkms:
One could take a look at the `io.gitlab.arturbosch.detekt.cli.DetektYmlConfigTest`, where we validate the yaml config in an integration test.
I think some of the logic there can be taken to implement this feature.

PR #1998, opened 2019-10-08
Validate yaml configurations by comparing their structure - #516

arturbosch:
The base idea is this function signature: `fun verifyConfig(config: Config, baseline: Config): List<Notification>`

We take the user config and the default detekt config and compare their structure.
My idea to integrate this into CLI is via a config block:
```yaml
config:
    active: false // by default? no performance tested yet, should be negligible
    excludes: user-ruleset, user-property
```
This way we can exclude user rules. Shouldn't be to much of a hassle due to rule set grouping.

The reporting mechanism tells the user that this property does not exist.
We could integrate a string distance comparison on each "sub config" level and suggest possible options. The performance should be tested here as for example the Levenstheins Distance could be quite expensive.


3flex:
There are requests to support multiple user config files (instead of limited to default detekt config and a single user config file). I think this should be considered when implementing this change.

arturbosch:
> There are requests to support multiple user config files (instead of limited to default detekt config and a single user config file). I think this should be considered when implementing this change.

Can you elaborate? Do you mean `CompositeConfig` aka when using `--config config1,config2`?
If yes, I just pushed a commit to support them.

Have even found two errors in your `detekt.yml` with this feature :):
```
Property 'formatting>NoItParamInMultilineLambda' is misspelled or does not exist.
Property 'style>VariableNaming' is misspelled or does not exist.
```

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -51,6 +51,12 @@ class CliArgs : Args {
     )
     var generateConfig: Boolean = false
 
+    @Parameter(
+        names = ["--validate-config", "-vc"],
+        description = "Validates the properties of the specified configuration (--config)."
+    )
+    var shouldValidateConfig: Boolean = false
```
3flex: Does this need to be a new CLI option? I don't see why you wouldn't validate in all cases. It would add a little time to detekt startup but as a user I'd want to know that the config had a problem as soon as possible instead of having to remember to run command again with this flag.

3flex:
> Can you elaborate?

See #2045. I believe that's not supported today? So instead of validating only two files the validation should support validation of a merged config, where that merged config could be made up of 1..n files instead of 1..2 like is supported today.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
arturbosch: I've removed the flag. By default we will always validate the config.
The performance penality is negligible.

arturbosch:
> > Can you elaborate?
> 
> See #2045. I believe that's not supported today? So instead of validating only two files the validation should support validation of a merged config, where that merged config could be made up of 1..n files instead of 1..2 like is supported today.

Yep, this is supported via the CompositeConfig which this PR also validates. #2045 sounds like a regression to me which does not need special treatment here.

To me this PR is ready to be reviewed :)

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ValidatableConfiguration.kt
```diff
@@ -0,0 +1,76 @@
...
+    when (config) {
+        is YamlConfig -> testKeys(config.properties, baseline.properties, null)
+        is ValidatableConfiguration -> notifications.addAll(config.validate(baseline, excludePatterns))
+        else -> error("Unsupported config type for validation: '${config::class}'.")
+    }
+
+    return notifications
+}
+
+internal fun doesNotExistsMessage(prop: String): Notification =
```
schalkms: nit: weird naming

schalkms:
Documentation seems to be outdated. That's the reason for the failing build.

Issue #2000, opened 2019-10-08
TrailingWhitespace rule should ignore indents in multiline string literals


elijah-pl:
The rule or changes you're looking for might have already been suggested!
Please search in the [issues](https://github.com/arturbosch/detekt/issues) before creating a new one.

## Expected Behavior of the rule
`TrailingWhitespace` rule should ignore indents in multiline string literals.
```kt
val example = """
    Should ignore indent on the next line
    
    Should ignore indent on the previous line
""".trimIndent()
```

## Context
Users should be able to format multiline string literals without breaking checkstyle rule.

elijah-pl:
I can start my work on this issue as soon as maintainers confirm that it needs to be fixed.

schalkms:
Thanks for reporting and willing to commit to detekt! 
We really appreciate any help from our community.
This issue is fixed by #1971 and included in the next release (1.1.1).
However, you could improve the multi-line test case in the `TrailingWhitespaceNegative` class.

PR #2003, opened 2019-10-08
Improved test case for resolved #1971 (TrailingWhitespace and multiline string)

elijah-pl:
Following up discussion in #2000. 
A bit more meaningful test case for already resolved issue.


arturbosch:
Thanks!

PR #2013, opened 2019-10-12
Better handling for the Suppresion of errors

BraisGabin:
I added a regresion with #1977. Luckly it was not released yet.

`enumEntry.nameIdentifier` is not a `KtElement` so it's not setted in the `Entity.ktElement`. Right now the only use of that the `KtElement` is to check if there is a `@Suppress` annotation. So if it was null, you can't `@Suppress`. With this PR I fix that.

I'm fixing more than that so, please, check if this is correct. What I'm doing is that, instead of checking if the current element is a `KtElement` I look for the nearest `KtElement` parent. This way, other warnings as `SpacingBetweenPackageAndImports` can use `@Supress` too.

Maybe I'm breaking the API for the plugins. If you think so, I can implement this without any api change.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Entity.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Entity.kt
@@ -5,6 +5,7 @@ import io.gitlab.arturbosch.detekt.api.internal.searchClass
 import io.gitlab.arturbosch.detekt.api.internal.searchName
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.psi.KtElement
+import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
 
 /**
  * Stores information about a specific code fragment.
@@ -27,7 +28,8 @@ data class Entity(
             val name = element.searchName()
             val signature = element.buildFullSignature()
             val clazz = element.searchClass()
-            return Entity(name, clazz, signature, Location.from(element, offset), element as? KtElement)
+            val ktElement = element.getNonStrictParentOfType<KtElement>()!!
+            return Entity(name, clazz, signature, Location.from(element, offset), ktElement)
         }
     }
 }
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNamingSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNamingSpec.kt
@@ -28,6 +28,14 @@ class EnumNamingSpec : Spek({
             assertThat(NamingRules().compileAndLint(code)).hasSize(1)
         }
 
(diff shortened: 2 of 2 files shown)


BraisGabin:
I can't implement something like `element.getNonStrictParentOfType<KtElement>()` because to do so I need to inline the function. But the function is recursive so it can't be inlined. Well, the truth is that, because is `tailrec`, it should be elegible to inline but it's not right now. Bug here: https://youtrack.jetbrains.com/issue/KT-27897

But I renamed the function to: `getNonStrictParentOfTypeKtElement`.

I debugged the `SpacingBetweenPackageAndImports` case and yes, it finds the `KtFile`. The thing is: What do you want me to test exactly?

Reading this: https://www.jetbrains.org/intellij/sdk/docs/basics/architectural_overview/psi_files.html the `PsiFile` is always the root of the tree. If that's true and detekt is a tool for Kotlin we can even ensure that this function is going to return always a non-null value. So, what do you think about this? (I did the change already in the code)

---

As a side effect we could change `Entity.ktElement` to not nullable. The only place where we are setting it as `null` are tests and in the `FormattingRule`. And in this last case we have the reference of the `KtFile` near, so we can use it.

BraisGabin:
🤦‍♂ I just realised that the function `getNonStrictParentOfType` already exist. I just changed to use it.

BraisGabin:
I think that the fails of this PR was a problem in the travis end.

arturbosch:
I would let the KtElement inside a `Finding` to be nullable.
A `KtFile` should really only be returned if the rule is about the file itself, for example a custom `TooBigFile` rule.
For formatting rules for example there is no need to "postprocess" the Finding via a `KtElement`.

`element?.getNonStrictParentOfType<KtElement>()` should do it (@schalkms ).

BraisGabin:
Right now `Entity.from` forces you to send a non-null element so we don't need to use `?.`.

Move `Entity.ktElement` to non-null was just an idea, not related with this PR. We can move this discussion to an issue if we want to talk more about it.

The only "problem" in this PR is this line:

```kotlin
val ktElement = element.getNonStrictParentOfType<KtElement>()!!
```

The use of `!!` is not needed but reading the Psi documentation it should never fail. This will help to move `Entity.ktElement` to not-null in the future if we want to. I don't know the "rules of the library" in this cases: we can be safe or follow the documentation and force the non-null.

arturbosch:
> Right now `Entity.from` forces you to send a non-null element so we don't need to use `?.`.
> 
> Move `Entity.ktElement` to non-null was just an idea, not related with this PR. We can move this discussion to an issue if we want to talk more about it.
> 
> The only "problem" in this PR is this line:
> 
> ```kotlin
> val ktElement = element.getNonStrictParentOfType<KtElement>()!!
> ```
> 
> The use of `!!` is not needed but reading the Psi documentation it should never fail. This will help to move `Entity.ktElement` to not-null in the future if we want to. I don't know the "rules of the library" in this cases: we can be safe or follow the documentation and force the non-null.

Ah right!
Hm the travis is realy not related.

PR #2017, opened 2019-10-12
Don't allow to use underscores and lowercases at the same time in names

BraisGabin:
There are places where we allow to use: `thisFormat` or `THIS_FORMAT` but we should never allow to use `This_Format`. This PR fix this issue in the 3 places where we allow the use of `_` in the naming: enums, object properties and top level properties.

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/TopLevelPropertyNamingSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/TopLevelPropertyNamingSpec.kt
@@ -1,8 +1,8 @@
 package io.gitlab.arturbosch.detekt.rules.naming
 
+import io.gitlab.arturbosch.detekt.test.assertThat
 import io.gitlab.arturbosch.detekt.test.compileContentForTest
 import io.gitlab.arturbosch.detekt.test.lint
-import org.assertj.core.api.Assertions.assertThat
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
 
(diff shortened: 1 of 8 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/TopLevelPropertyNamingSpec.kt
```diff
@@ -42,10 +42,9 @@ class TopLevelPropertyNamingSpec : Spek({
                 val serialVersionUID = 42L
                 val MY_NAME = "Artur"
                 val MYNAME = "Artur"
-                val MyNAME = "Artur"
                 private val NAME = "Artur"
-                val s_d_d_1 = listOf("")
                 private val INTERNAL_VERSION = "1.0.0"
+                private val _INTERNAL_VERSION = "1.0.0"
```
arturbosch: IntelliJ flags this line

BraisGabin: Yeah... It doesn't have too much sense to have a constant backing property. I just added because it was allowed and I wanted to be exhaustive.

BraisGabin:
I was following this: https://kotlinlang.org/docs/reference/coding-conventions.html#naming-rules (and I just realised that the function name rule in detekt doesn't follow that). If you prefer to follow what IntelliJ accepts I can close this PR because the actual code does that. Or I can keep the PR removing the last commit to keep the minor fixes.

arturbosch:
Ok, I will close it now as we are on pair with IntelliJ.

BraisGabin:
After closing this I opened an issue to IntelliJ to fix this issue there: https://youtrack.jetbrains.com/issue/KT-34508 They accepted it like a valid issue but no one fixed it yet. (I tried to create a PR fixing it but the IntelliJ project is a madness)

I'm reopening this PR because I think that, with the issue accepted by IntelliJ, we can fix it in our end. I'm going to create another PR fixing the problem with code like this `private val _HELLO`.

Issue #379, opened 2017-09-09
Rules: JUnit/Spek


schalkms:
I think this would be a nice new feature.

To be more precise, PMD does also have this kind of rules
reference:
http://pmd.sourceforge.net/pmd-4.3.0/rules/index.html#JUnit_Rules

Detection of unit tests
- lies in the tests package
- class name ends with Test/Spek
- class imports JUnit/Spek package
- has JUnit annotations

arturbosch:
A ruleset for testing code smells seems reasonable.

arunvelsriram:
I wrote a ktlint ruleset for linting JUnit test name. I am also planning to add more rules.

https://github.com/arunvelsriram/ktlint-ruleset-junit/

By looking at detekt readme it looks like detekt is more configurable than ktlint. So I am thinking of changing the ruleset to detekt. Do you have any suggestions on this?

schalkms:
> By looking at detekt readme it looks like detekt is more configurable than ktlint.

Yes.

> So I am thinking of changing the ruleset to detekt. Do you have any suggestions on this?

Ktlint also uses the PSI.
You can at least copy over the logic, but you would need to use the corresponding `visit_()` method. Testing the rules also works differently in detekt, but that shouldn't be an issue, since you could copy over the test snippets.
May I ask you what test related rules you want to see in detekt? @arunvelsriram 
We'd really appreciate any contributions in that regard.

arunvelsriram:
I have the following rules for JUnit in my repo:

1. test description style 
```
fun `should add two numbers`() { }
```
2. test must contain least one assertion
3. ignored tests should not cross a threshold

I will add the above in `detekt` and raise a PR.

schalkms:
I think 2 and 3 are a nice addition. 1 is IMHO too specific.
Thanks!

arunvelsriram:
we could make 1 configurable like backticks style, camelCase style or custom regex.

schalkms:
Excellent idea!

BraisGabin:
I would vote to close this. The idea to have a rule set jor JUnit seems a good idea. But we want to keep detekt platform/library agnostic so it should be done in a custom rule set (it could be a first party one, but out of this repo)

PR #2019, opened 2019-10-13
Add junit rule that reports if test has no assertion

arunvelsriram:
Related issue: #379 

I have added one of the junit rule (2nd one) mentioned in https://github.com/arturbosch/detekt/issues/379#issuecomment-539077928

Am working on adding other rules as well. Since this is where we are introducing junit rules I have added one rule only to get early feedback from the community.


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/junit/TestWithoutAssertion.kt
```diff
@@ -0,0 +1,47 @@
...
+        Severity.CodeSmell,
+        "A JUnit test should have at least one assertion",
+        Debt.TEN_MINS
+    )
+
+    override fun visitNamedFunction(function: KtNamedFunction) {
+        super.visitNamedFunction(function)
+        function.annotationEntries.find {
+            it.text == TEST_ANNOTATION
+        }?.let {
```
arturbosch: Please prefer normal if-statements for multiline bodies

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/junit/TestWithoutAssertion.kt
```diff
@@ -0,0 +1,47 @@
...
+            val body = function.bodyExpression?.text
```
arturbosch: You could introduce an early return here with `?: return`

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/junit/TestWithoutAssertion.kt
```diff
@@ -0,0 +1,47 @@
...
+            if (body?.contains(ASSERTION_PATTERN, ignoreCase = true) == false) {
```
arturbosch: The assertion pattern should be a configurable regex.
The user may use a library like `strikt` where he would use `expectThat` syntax instead of `assert`.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/junit/TestWithoutAssertion.kt
```diff
@@ -0,0 +1,47 @@
...
+                        Entity.from(function),
+                        issue.description
+                    )
+                )
+            }
+        }
+    }
+
+    companion object {
+        private const val TEST_ANNOTATION = "@Test"
```
arturbosch: Not sure about if this should be a configurable regex

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/providers/JUnitProvider.kt
```diff
@@ -0,0 +1,23 @@
...
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.RuleSet
+import io.gitlab.arturbosch.detekt.api.RuleSetProvider
+import io.gitlab.arturbosch.detekt.rules.junit.TestWithoutAssertion
+
+/**
+ * Rules in this rule set report issues related to JUnit tests
+ *
+ */
+class JUnitProvider : RuleSetProvider {
```
arturbosch: @schalkms @3flex should we prefer a `TestProvider` and have rule names like `JUnitTestWithoutAssertion`?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/junit/TestWithoutAssertionSpec.kt
```diff
@@ -0,0 +1,38 @@
+package io.gitlab.arturbosch.detekt.rules.junit
+
+import io.gitlab.arturbosch.detekt.test.lint
+import org.assertj.core.api.Assertions.assertThat
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+
+class TestWithoutAssertionSpec : Spek({
+    describe("junit test assertion") {
```
arturbosch: We should add more tests when the assertion pattern is configurable

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/providers/JUnitProvider.kt
3flex: I was thinking something similar, but I wasn't sure what we do about the case where we want an equivalent rule for Spek as well... would we have two rules, `JUnitTestWithoutAssertion` and `SpekTestWithoutAssertion`, or just one `TestWithoutAssertion` that covers both cases and can be extended to support other test engines?

3flex: Or, we have a `JUnitProvider` and a `SpekProvider` with possibly duplicated rules between them so it's easy to enable only the rulesets that apply to the project (which would typically use a single test engine)

schalkms: I'm okay with both approaches.

detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -285,6 +285,11 @@ formatting:
     active: true
     autoCorrect: true
 
+junit:
```
schalkms: Maybe we should use a more generic name here for all test related rules.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/junit/TestWithoutAssertion.kt
arunvelsriram: Done

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/junit/TestWithoutAssertion.kt
arunvelsriram: Done

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/junit/TestWithoutAssertion.kt
arunvelsriram: Done

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/junit/TestWithoutAssertion.kt
arunvelsriram: In JUnit, the annotation is `@Test` always as far as I know.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/junit/TestWithoutAssertionSpec.kt
arunvelsriram: Done

cortinico:
Just dropping my opinion here in the loop:

Quoting @schalkms 
> Maybe we should use a more generic name here for all test related rules.

I have the feeling that `junit` is definitely a bad name for this rule (multirule) as it will be easily confused with JUnit Rule (see  https://junit.org/junit4/javadoc/4.12/org/junit/Rule.html), a mechanism provided by the JUnit framework to run setup and cleanup statements before JUnit tests.

So reading the name of this PR:

> Add junit rule that reports if test has no assertion

The first impression I had was that we're adding a JUnit Rule to make sure that our Detekt tests contain an assertions and let our internal test suite fail if not.

Long story short: having `testing` is probably better?

arunvelsriram:
@cortinico agreed. `junit` is confusing for users.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/providers/JUnitProvider.kt
arunvelsriram: if we are changing the `ruleSetId` to `testing` https://github.com/arturbosch/detekt/pull/2019#issuecomment-546880937 then is it safe to assume that this provider will become generic like `TestingProvider`?.

arturbosch:
I agree with all of you that we should have a generic `testing` rule set.
As @schalkms mentioned without resolving all calls and searching for asserts this rule would only be useful inside the test case itself.

We could:
1. resolve all calls inside the file without using the heavy Kotlin compiler resolution - reasoning would be tests should be self contained with private helpers.
2. Work on computing a Call Graph and let rules use it for such cases
3. Rely on Kotlin compilers type resolution (which we should ofc target sooner than later...)

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/providers/JUnitProvider.kt
3flex: A third approach - don't have a "testing" rule set at all, but include the rules in existing rulesets. This is the approach PMD, Spotbugs and Error Prone take.

In this case it would go in the "bugs" ruleset since it's likely a bug if a test has no assertion.

schalkms:
@arunvelsriram do you have any updates on this PR? Do you need further assistance?

arturbosch:
Unfortunately I propose to close this PR.

This rule is basically a graph reachability problem.
We need to calculate a call graph.
Mark all nodes calling a `assert` or `expect` as "hasAssertion" and traverse all paths from test methods to find out if they eventually lead to a node with "hasAssertion".
We do not have a graph structure now.

Would a `ExplicitAssertion` rule be useful? Every JUnit method need to call a functions starting with name "assert" or "expect"...

schalkms:
> This rule is basically a graph reachability problem.

Agreed. This is currently out of scope what detekt can do.
I propose to close this and have this as a custom rule if you really need this for your code base.

> Would a ExplicitAssertion rule be useful? Every JUnit method need to call a functions starting with name "assert" or "expect"...

I don't think so. There are some many valid cases where you don't need that. For instance, what happens if you only verify the behavior of a unit (e.g. ´verify´).

PR #2024, opened 2019-10-16
Add ImplicitDefaultLocale rule

koral--:
Analogous to `DefaultLocale` from Android lint.

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ImplicitDefaultLocaleSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ImplicitDefaultLocaleSpec.kt
@@ -16,7 +16,7 @@ class ImplicitDefaultLocaleSpec : Spek({
                 fun x() {
                     String.format("%d", 1)
                 }"""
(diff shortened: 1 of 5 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ImplicitDefaultLocaleSpec.kt
```diff
@@ -0,0 +1,31 @@
...
+    val subject by memoized { ImplicitDefaultLocale(Config.empty) }
+
+    describe("ImplicitDefault rule") {
+
+        it("reports String.format call with template but without explicit locale") {
+            val code = """
+                fun x() {
+                    String.format("%d", 1)
+                }"""
+            assertThat(subject.compileAndLint(code).size).isEqualTo(1)
```
3flex:
```suggestion
            assertThat(subject.compileAndLint(code)).hasSize(1)
```

koral--:
> It would be even more useful if it also detected cases like
> 
> ```kotlin
> fun x() {
>     "String".format("%d", 1)
> }
> ```
> 
> But this will be a bit more work to implement.

OK, I'll try to do that in next PR.

Issue #2006, opened 2019-10-09
Exclusions for ThrowsCount


madhead:
## Expected Behavior of the rule
`ThrowsCount` should be more intelligent and allow intellectual exclusions (e.g. by thrown exception class).

## Context
Too many exit points like returns or throws is indeed code smell, but there are exceptions. Currently, `ThrowsCount` is pretty dumb. It only counts throws expressions without any context. One example is "guarding" throws used to fail fast the program.

Imagine I have a class `Effect`:

```kotlin
data class Effect(
    val id: String,
    val name: String,
    val size: Long
)
```

I store it in the NoSQL key-value storage like DynamoDB. I wrote a function to convert DynamoDB's object into Effect:

```kotlin
fun Map<String, AttributeValue>.toEffect(): Effect = Effect(
    id = this["id"]?.s() ?: throw MissingPropertyException("id"),
    name = this["name"]?.s() ?: throw MissingPropertyException("name"),
    size = this["size"]?.n()?.toLong() ?: throw MissingPropertyException("size")
)
```

Currently detekt reports an issue here (default number of throws is 2). I should either increase the default, or use `@Supress`. But both solution are not perfect: increasing the limit will skip the warnings for really complex codes and `@Supress` feels like I am cheating to make detekt happy.

Ideally, there should be a list of exclusions for `ThrowsCount`, so some exceptions used across the code base to guard and fail fast are not counted.

madhead:
Probably all the `throw`s and `return`s in a syntax like `?: throw` / `?: return` could be configured to be excluded, as it is a pretty common pattern of defensive programming style.

madhead:
I found `excludeGuardClauses` in returns rule, it is a perfect solution. Need something for throws.

BraisGabin:
Can we close this issue? The PR is merged.

madhead:
Oh, sure. I thought, it would be closed automatically…

PR #2025, opened 2019-10-16
Support guard clause exclusion for ThrowsCount rule

madhead:
PR for #2006.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ThrowsCount.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ThrowsCount.kt
@@ -54,9 +54,7 @@ class ThrowsCount(config: Config = Config.empty) : Rule(config) {
         if (!function.isOverride()) {
             val count = function
                 .collectDescendantsOfType<KtThrowExpression>()
-                .filterNot {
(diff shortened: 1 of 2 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ThrowsCount.kt
```diff
@@ -45,19 +47,31 @@ class ThrowsCount(config: Config = Config.empty) : Rule(config) {
...
+    private val excludeGuardClauses = valueOrDefault(EXCLUDE_GUARD_CLAUSES, false)
 
     override fun visitNamedFunction(function: KtNamedFunction) {
         super.visitNamedFunction(function)
         if (!function.isOverride()) {
-            val count = function.collectDescendantsOfType<KtThrowExpression>().count()
+            val count = function
+                .collectDescendantsOfType<KtThrowExpression>()
+                .filterNot {
+                    excludeGuardClauses && it.isGuardClause()
```
arturbosch: nit: can we put this line on the same as filterNot ?

madhead: Done.

PR #2027, opened 2019-10-16
Dont check WrongEqualsTypeParameter if the function is topLevel

BraisGabin:
You can have a function named `equals` as a top level function. It's strange but safe.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/WrongEqualsTypeParameter.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/WrongEqualsTypeParameter.kt
@@ -49,7 +49,7 @@ class WrongEqualsTypeParameter(config: Config = Config.empty) : Rule(config) {
     }
 
     override fun visitNamedFunction(function: KtNamedFunction) {
-        if (function.name == "equals" && !function.hasCorrectEqualsParameter()) {
+        if (function.name == "equals" && !function.isTopLevel && !function.hasCorrectEqualsParameter()) {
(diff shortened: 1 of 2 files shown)


3flex:
I think this rule should instead be changed to instead only `equals()` functions that are overrides (similar to `EqualsAlwaysReturnsTrueOrFalse` rule) instead of checking if it's a top level function.

I don't think a top-level `equals()` function would ever have to fulfil [this contract](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/equals.html).

BraisGabin:
I think that maybe can be reformuled to: "any non-top-level function called `equals` that has only one parameter should have the modifier `override`". The reason is that otherwise we are overloading the `equals` function and that's exactly what this rule is about. If you agree with this definition I can rewrite this rule.

3flex:
What I meant was, the rule would check for a function named "equals" that is an override. If those conditions are true, then the rule would check that the return type is as expected.

If the `equals` function is not an override then this rule shouldn't flag an issue.

BraisGabin:
> If the `equals` function is not an override then this rule shouldn't flag an issue.

Yes, it should. Think on this code:

```
fun main(args: Array<String>) {
    A().equals("")  // the output is "wrong"
}

private class A {

    fun equals(other: String): Boolean {
        println("wrong")
        return !super.equals(other)
    }

    override fun equals(other: Any?): Boolean {
        println("correct")
        return super.equals(other)
    }
}
```

The first `equals` function doesn't have the `override` modifier but it's overloading the `equals(Any?)`. So this can lead to very ugly bugs. This is what this rule is cheking.

3flex:
Thanks for your patience and contribution!

PR #2031, opened 2019-10-17
Update dependencies

3flex:
Because there was a dokka update, also regenerates documentation using the new version.


3flex:
Coverage changes would be thanks to upgrading JaCoCo.

PR #2051, opened 2019-10-24
Improve text location: MandatoryBracesIfStatements

BraisGabin:
I improved the text location and fix a bug when we were raising 2 findings but there was only one error.


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatements.kt
```diff
@@ -12,6 +12,7 @@ import org.jetbrains.kotlin.com.intellij.psi.PsiWhiteSpace
 import org.jetbrains.kotlin.psi.KtBlockExpression
 import org.jetbrains.kotlin.psi.KtIfExpression
 import org.jetbrains.kotlin.psi.psiUtil.siblings
+import kotlin.random.Random
```
arturbosch: This may be not related?

BraisGabin: You are right, fixed.

PR #2062, opened 2019-10-27
Improve text location: MatchingDeclarationName

BraisGabin:


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MatchingDeclarationNameSpec.kt
```diff
@@ -97,6 +104,14 @@ internal class MatchingDeclarationNameSpec : Spek({
                 assertThat(findings).hasLocationStrings("'object O' at (1,1) in /Objects.kt")
             }
 
+            it("should not pass for object declaration even with suppress on the object") {
+                val ktFile = compileContentForTest("""@Suppress("MatchingDeclarationName") object O""")
+                ktFile.name = "Objects.kt"
```
arturbosch: We could give the `compileContentForTest` function or even the `(compileAnd)Lint` functions a second parameter `filename`?

BraisGabin: I just did a little research... Why do we use the functions inside `CompileExtensions.kt`? They provide nothing... And why the file is called extensions if they are not extension functions?

I did a fast fix for this PR but I can create (later) a new PR removing that file and using `KtTestCompiler` directly. What do you think?

arturbosch: True the naming is kinda odd.
The intention was to not use the `KtTestCompiler` directly. `compileContentForTest` is more concise for tests.

PR #2070, opened 2019-10-29
Use constants for config keys in tests

amitdash291:


3flex:
I understand and appreciate this PR, but I'm against it because changing a constant in a rule can then break user config files since that won't then be detected in the tests. Using literals protects against that (at least for the way most rule tests have been written).

We could use the constants though if there were other tests checking the constant value was as expected, but that also seems a bit wrong, since we should ultimately be testing that the correct config keys affect the behaviour or the rule.

But at the moment I think this PR introduces some risk.

amitdash291:
> I understand and appreciate this PR, but I'm against it because changing a constant in a rule can then break user config files since that won't then be detected in the tests. Using literals protects against that (at least for the way most rule tests have been written).
> 
> We could use the constants though if there were other tests checking the constant value was as expected, but that also seems a bit wrong, since we should ultimately be testing that the correct config keys affect the behaviour or the rule.
> 
> But at the moment I think this PR introduces some risk.

Agreed that we generally test using literals ideally. I was aiming towards standardisation, as half of the tests are using constants declared in the rules to test them.

An agreeable approach could probably be introducing the constants within the test file for private access. What do you folks think? @3flex @schalkms @arturbosch

3flex:
I think that's a good idea, I'd approve that PR (would need another maintainer to sign off though so maybe wait for another reply). Thanks!

amitdash291:
> I think that's a good idea, I'd approve that PR (would need another maintainer to sign off though so maybe wait for another reply). Thanks!

Okay let's wait for other's replies and then I can go ahead and do the required changes here. Thanks @3flex

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NamingRules.kt
```diff
@@ -101,4 +101,8 @@ class NamingRules(config: Config = Config.empty) : MultiRule() {
 
     private fun KtVariableDeclaration.withinObjectDeclaration(): Boolean =
             this.getNonStrictParentOfType<KtObjectDeclaration>() != null
+
+    companion object {
+        const val IGNORE_OVERRIDDEN = "ignoreOverridden"
```
schalkms: I wouldn't want to add this constant to the product code just for testing reasons.

amitdash291: I will move this constant to the test file `NamingRulesSpec`, which will be handy to reuse for any future tests. I hope that sounds okay.

Issue #2046, opened 2019-10-22
Add complexity report to HTML output


kumarpratyush:
## Expected Behavior
The **complexity report** is only printed in console output. To get a count of code smells and McCabe Complexity in order to keep a tab on them, we have to push the console output to a file and parse that. This would break if the format of console reporter changes. 

## Current Behavior
Currently, the HTML output gives a summary of project metrics, but not the complexity report. It would be good to have the complexity report as well as part of some reporter.

## Context
We are trying to get more metrics out of detekt in one place. This could enable people to get these values and represent them graphically over time to keep a tab on the code quality of their project. 

**P.S. -** I have gone through the code and it is a small change. If you think this feature request is good to have, I can raise a PR for the same. Please let me know.

schalkms:
We would very much appreciate this. Please go ahead and submit a PR. Thanks for your time and contribution. If you have any questions please fell free to ask.

kumarpratyush:
Thank you for the response. I will raise the PR in some time.

schalkms:
Closed by 85c93ea

PR #2071, opened 2019-10-31
Add complexity report to html output

kumarpratyush:
This PR is with respect to issue discussed at #2046 .
Currently, Complexity report is only printed as part of console output. 
This PR attempts to push complexity report as part of the HTML output in a `<div>` container.


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputReport.kt
```diff
@@ -104,6 +115,16 @@ class HtmlOutputReport : OutputReport() {
             snippetCode(lineSequence, finding.startPosition, finding.charPosition.length())
         }
     }
+
+    private fun getComplexityMetrics(detektion: Detektion): List<String> {
+        var complexities = listOf<String>()
+        val complexityReportGenerator = ComplexityReportGenerator.create(detektion)
+        val complexityReport = complexityReportGenerator.generate()
+        return if (complexityReport.isNullOrBlank()) complexities else {
```
schalkms:
```suggestion
        return if (complexityReport.isNullOrBlank()) listOf<String>() else {
```

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputReport.kt
```diff
@@ -104,6 +115,16 @@ class HtmlOutputReport : OutputReport() {
...
+            complexities = complexityReport.split("\n")
+            return complexities.subList(1, complexities.size - 1)
```
schalkms:
```suggestion
            return complexityReport.split("\n").subList(1, complexities.size - 1)
```

The line above can be removed then. This way we spare the mutable variable.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputFormatTest.kt
```diff
@@ -34,6 +34,7 @@ class HtmlOutputFormatTest : Spek({
 
             assertThat(result).contains("<h1>detekt report</h1>")
             assertThat(result).contains("<h2>Metrics</h2>")
+            assertThat(result).contains("<h2>Complexity Report</h2>")
```
schalkms: Tests are missing for the **Metrics** and **Complexity Report** where there's actually an item in the list (`<li></li>`).

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputReport.kt
```diff
@@ -104,6 +115,16 @@ class HtmlOutputReport : OutputReport() {
             snippetCode(lineSequence, finding.startPosition, finding.charPosition.length())
         }
     }
+
+    private fun getComplexityMetrics(detektion: Detektion): List<String> {
+        var complexities = listOf<String>()
+        val complexityReportGenerator = ComplexityReportGenerator.create(detektion)
+        val complexityReport = complexityReportGenerator.generate()
```
arturbosch: Lets inline the `.genereate()`-call

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputReport.kt
kumarpratyush: Not really. `complexities` is an empty list at this point and the above line populates it. Hence `complexities.size - 1` would be incorrect. 
We can either make `complexities`  as an immutable object or change it to following : 
`return complexityReport.split("\n").subList(1, complexityReport.split("\n").size - 1)`

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputFormatTest.kt
kumarpratyush: Added these tests.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputReport.kt
schalkms: True. Then you could at least write something like this in order to move the mutable variable into the else clause.

```Kotlin
var complexities = complexityReport.split("\n")
return complexities.subList(1, complexities.size - 1)
```

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputFormatTest.kt
schalkms: 1 unit test is missing for the case when the complexity report is null or blank and thus returns an empty string list.
The rest looks good. Thank you! :-)

```Kotlin
if (complexityReport.isNullOrBlank()) listOf<String>()
```

kumarpratyush: My bad. Added the missing test. Thank you :)

schalkms:
Thanks! This is an awesome contribution.

kumarpratyush:
Nevermind. It was a pleasure 😄

Issue #1911, opened 2019-09-13
Enable more rules by default


arturbosch:
## Expected Behavior
Many rules aged together with the detekt codebase and should be ready to be enabled by default. No "highly subjective" rules should be enabled though. 

## Current Behavior
Only a subset of detekt rules are enabled by default.

## Context
1.2.0 Release

Which are the rules you want to see enabled by default, lets list and discuss them.

schalkms:
Good idea! 
We need to be careful regarding the rules that need type and symbol solving.
Furthermore, we should also move some rules into other rulesets. I highlighted them.
Here's my list:

`comments`:

- UndocumentedPublicClass
- UndocumentedPublicFunction

`complexity`:

- ComplexInterface
- MethodOverloading
- StringLiteralDuplication

all rules in the `exceptions` ruleset

`naming`:

- MemberNameEqualsClassName

`performance`:

- ArrayPrimitive

`potential-bugs`:

- EqualsAlwaysReturnsTrueOrFalse
- InvalidRange
- IteratorHasNextCallsNextMethod
- IteratorNotThrowingNoSuchElementException
- MissingWhenCase
- RedundantElseInWhen
- UnsafeCallOnNullableType
- UnsafeCast
- UselessPostfixExpression
- WrongEqualsTypeParameter

`style`:

- DataClassShouldBeImmutable
- EqualsNullCall
- ForbiddenComment
- FunctionOnlyReturningConstant
- LoopWithTooManyJumpStatements (**should be moved to the `complexity` ruleset!**)
- NestedClassesVisibility
- ProtectedMemberInFinalClass
- ReturnCount (**should be moved to the `complexity` ruleset!**)
- ThrowsCount (**should be moved to the `complexity` ruleset!**)
- UnnecessaryAbstractClass
- UnnecessaryInheritance
- UnusedImports
- UnusedPrivateClass
- UnusedPrivateMember
- UseDataClass
- UselessCallOnNotNull
- UtilityClassWithPublicConstructor

3flex:
Should #921 be rolled into this discussion?

arturbosch:
Sounds good, so closing #921 you mean? This one is pinned :)

schalkms:
Since the style package is getting bigger, I would split it up

- Design
- Readability

schalkms:
I updated my initial post after implementing #1935.

tasomaniac:
[...]

PR #2075, opened 2019-11-05
Activate more rules by default - #1911

arturbosch:


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/EqualsAlwaysReturnsTrueOrFalse.kt
```diff
@@ -35,6 +35,8 @@ import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
  *     return this == other
  * }
  * </compliant>
+ *
+ * @active since v1.2.0
```
schalkms: Is this generated automatically or validated in any way?

arturbosch: Nope, I've done it manually. The documentation generation does not take this property into account. When the rule was turned on by default or added to detekt could be a valuable information for the documentation.

Issue #1515, opened 2019-02-26
New rule: Add check for copyright header


Carrera15111:
## Expected Behavior of the rule
Please add a new Regex-check for copyright header, something similar to Checkstyle-check
`<property name="header"
            value="/**\n * Projectname\n * Copyright (c) 2019 Author. All rights reserved.\n * Created by Android Studio\n * User: Username\n */" />`

schalkms:
Good point. This rule makes sense to me.
You can do that via the AST and also check for the correct license here.
We happily accept contributions here.

hick209:
Would that be a style or documentation rule?

schalkms:
I'd say it belongs to the style rule set.

arturbosch:
I would say documentation :)

dector:
If nobody if working on this issue, I can adapt and generalize my solution from mozilla-mobile/android-components#846 .

schalkms:
@dector that would be awesome! We’d love to have this rule in detekt. Please go ahead.

dector:
@schalkms 

Is detekt config file an appropriate place to store license header? I think I like another option more: use config file to set file path for license template.

What do you think?

schalkms:
I think it’s better to store the copyright header in a simple text file. This approach is more flexible.

dector:
@schalkms I agree, thanks!

dector:
Seems that I have few questions about implementation details:

1. What is the convenient way to handle absent / wrong license template file. 

I see few options:
    - `(a)` Ignore rule (obviously, not a way to go as quality control system fail should be explicit).
    - `(b)` Use error logger (still not too obvious and I can't find something like this in other rules).
    - `(c)` Throw an exception.
    - `(d)` Report file absence as a defect.

Option `(d)` seems very interesting as it's smoothly integrates into the current reporting flow and will let to continue perform checks. However, it'll slightly increase implementation complexity as this case is clearly an exception in system working flow.

Option `(c)` is the most convenient way, as it seems. Technically, file path should be configured properly only once. But current UX for exception handling in `detekt` isn't very attractive/clear.

2. As I see a rule is created quite often (for each file?). File I/O can be quite time-consuming operation (relatively). Is there a way to cache/preload some data (e.g. license text)? Should I create an issue for this?

3. Can't find it in the project. What is the convenient way to resolve file path relatively to `detekt.yml`/rootProject location. Seems that files are resolved mostly when cli is starting?

[...]

PR #2077, opened 2019-11-05
#1515. Copyright header rule for Kotlin files.

dector:
:construction: _Draft_

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
@@ -1,6 +1,7 @@
 package io.gitlab.arturbosch.detekt.api
 
 import io.gitlab.arturbosch.detekt.api.Config.Companion.PRIMITIVES
+import java.nio.file.Path
 import java.util.LinkedList
 import kotlin.reflect.KClass
 
@@ -8,6 +9,8 @@ import kotlin.reflect.KClass
  * A configuration holds information about how to configure specific rules.
  */
 interface Config {
+
+    val location: Location get() = Location.Undefined
 
     /**
      * Tries to retrieve part of the configuration based on given key.
@@ -57,6 +60,11 @@ interface Config {
             Char::class,
             Long::class
         )
+    }
+
+    sealed class Location {
+        object Undefined : Location()
+        data class FromDirectory(val dir: Path) : Location()
     }
 }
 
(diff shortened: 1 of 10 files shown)


dector:
Added `location` property to config. Do you think it's a way to go or should I explore other solutions?

arturbosch:
I like the `location` idea.
It brings to me the idea that we can use the `Config` as a "data container", storing/caching additional information before actually performing the run (`HashMap`).
The hard part is that the additional information step is rule dependant.
So basically we need to have a new method for `Rule` like `initData(config: Config)` which runs once before the visiting process of the KtFile's.
For this we need to call `initData` in the `core` module in the main thread.
This sounds like a generic way for rules to cache data between KtFile's.

dector:
Glad that you've liked that. 

> This sounds like a generic way for rules to cache data between KtFile's.

Sounds awesome! Will explore it. Probably, I'll not have enough time in next 10 days, but will try to find short time-frames.

dector:
Sorry for the delay. :sweat_smile: 

Let me double check your suggestions:
  - Should `Config` serve a purpose of data cache for rules? I'm not sure that it's a very good idea as it'll bloat responsibilities for the class. Besides that what should be do with other config implementations (`CompositeConfig` and `Rule` itself as it's extending `ConfigAware` that is extending `Config`)?
  - Do you mean that `rule.initData()` should be invoked somewhere in `Detektor.run()` (or `runSync`/`runAsync`). If I get it right, we'll need to build rule sets once more for this. Is that a lightweight operation?

Thanks!

schalkms:
Sorry for the late answer @dector 
 - ping @arturbosch

arturbosch:
I've rebased on top of master and implemented a way to share the cache license header for this rule.
However I can't push my changes to this PR for whatever reasons...

Normally I check out the fork of a contributor, do some changes and can push them back to this branch. Maybe this time I'm not allowed to do this because it is a draft or something?

BraisGabin:
That's a permission that the author of the PR can give [when the PR is created](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)

arturbosch:
@dector please allow me to push to this branch :)

dector:
@arturbosch Hm.. "Edits from maintainers" are enabled for this PR. Let me undraft it.

arturbosch:
Unfortunately I couldn't push into this PR and created a new PR with the caching changes :(

PR #2078, opened 2019-11-07
Fix TrailingWhitespace reported position

### LINKED ISSUE #1979: TrailingWhitespace reports incorrect line number

BraisGabin:
This PR fixes #1979


schalkms:
Thanks for picking this up! 
I just scrolled quickly over the changes. At first glance, I'd say that this is a nice fix.
I'll take a closer look at this PR tomorrow.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrailingWhitespace.kt
```diff
@@ -23,23 +23,28 @@ class TrailingWhitespace(config: Config = Config.empty) : Rule(config) {
...
                 }
             }
             offset += 1 /* '\n' */
         }
     }
 
-    private fun hasTrailingWhitespace(line: String) =
-            line.isNotEmpty() && (line.last() == ' ' || line.last() == '\t')
+    private fun countTrailingWhitespace(line: String): Int {
+        return line.length - line.indexOfLast { it != ' ' && it != '\t' } - 1
```
schalkms: What happens if a whitespace is the second to last character in a line?
This might be a good test case for the negative scenario.

```Kotlin
println("Hello") ;
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrailingWhitespace.kt
```diff
@@ -23,23 +23,28 @@ class TrailingWhitespace(config: Config = Config.empty) : Rule(config) {
...
-                if (ktElement != null) {
-                    if (!ktElement.isPartOfString()) {
-                        report(CodeSmell(issue, Entity.from(ktElement), createMessage(index)))
+                if (ktElement == null || !ktElement.isPartOfString()) {
+                    val entity = Entity.from(file, offset - trailingWhitespaces).let { entity ->
+                        entity.copy(
+                            location = entity.location.copy(
+                                text = entity.location.text.copy(end = offset)
+                            )
+                        )
```
schalkms: I'm not sure why there's a need for copying the entity location and the entity text?
Could you please explain this to me?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrailingWhitespaceSpec.kt
```diff
@@ -20,10 +22,28 @@ class TrailingWhitespaceSpec : Spek({
...
+            it("should report the correct text location") {
+                assertThat(rule.findings).hasTextLocations(
+                    55 to 56,
+                    114 to 115,
+                    163 to 164,
+                    204 to 208,
+                    245 to 246,
+                    311 to 312,
+                    329 to 331
+                )
```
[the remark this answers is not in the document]
BraisGabin: I think that I can remove the other one. Because `hasSourceLocations` didn't check the end of the finding and I wrote code to set that value properly. But I agree both checks are not needed.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrailingWhitespace.kt
BraisGabin: Done

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrailingWhitespace.kt
BraisGabin: Sure, I'm using the `KtFile` as the entity for this finding. For that reason `Location.from` returns me the end of the file as the end of the finding. I need to correct that doing this. The start is set correctly because I'm using the argument `offset`.

I'm not a huge fan of this code... any better idea about how to create this finding?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrailingWhitespaceSpec.kt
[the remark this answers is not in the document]
schalkms: Please keep the other assertion. This one is so hard to maintain and not obvious when not looking over it.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrailingWhitespace.kt
schalkms: I'm not sure. Maybe @arturbosch has an idea.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrailingWhitespaceSpec.kt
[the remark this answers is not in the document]
BraisGabin: Done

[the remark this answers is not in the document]
schalkms: Sorry for nagging you once more. 
My previous post wasn't clearly formulated.
The following assertion should be removed, since we have the `hasSourceLocation` assertion.
The rest looks fine.

```Kotlin
assertThat(rule.findings).hasTextLocations(
                     55 to 56,
                     114 to 115,
                     163 to 164,
                     204 to 208,
                     245 to 246,
                     311 to 312,
                     329 to 331
                 )
```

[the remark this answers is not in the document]
BraisGabin: I can not remove it. Otherwise this code is not tested:

```kotlin
                    Entity.from(file, offset - trailingWhitespaces).let { entity ->
                        entity.copy(
                            location = entity.location.copy(
                                text = entity.location.text.copy(end = offset)
                            )
                        )
```

`hasSourceLocation` only tests the beginning of the assertion, but doesn't test the end. And, as you can see, I have code that set the end.

I understand your point and I don't like this test neither. What do you think about removing this huge test and create 7 small ones (one per finding)?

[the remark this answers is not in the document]
schalkms: Good to know.
I think this is a nice idea.
Separating the test case into smaller ones would be even better!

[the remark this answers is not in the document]
BraisGabin: Done

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrailingWhitespace.kt
arturbosch: Would the new `Entity.from(PsiElement, Location)` (#2083) help here without copying the entity?

BraisGabin: I'm not sure, Location is a difficult class to create. I mean, I'd do somehitng like:

```kotlin
val Location = Location.from(offset - trailingWhitespaces).let { location ->
  location.copy(text = location.text.copy(end = offset)
}
val entity = Entity.from(file, location)
```

I suppose that this is a bit better... But I don't like it that much neither.

schalkms: While both options are mediocre, the last snippet is better, since we can spare 1 copy statement.

PR #2080, opened 2019-11-08
Add license badge to README

schalkms:


3flex:
No issue with the PR itself, but does FOSSA actually report anything useful for this repo?

It says "FOSSA does not fully support this environment yet" when clicking the link, and it doesn't report the license for any dependencies (which I thought was the idea with FOSSA). All it can do right now is report that detekt uses Apache 2.0 license, but that's clear from the license file and also shown in top right of repo homepage on Github:
image

Issue #2014, opened 2019-10-12
Cleaning Entity


BraisGabin:
The class `Entity` has two unused properties: `name` and `className`. The first one can be private (we need it for `compat()`) but the second one can be removed completely.

With this we cane remove `HasEntity.name` and `HasEntity.inClass` too.

I supose that this is part of the public API. So probably for now we could add a `@Deprecation`.

arturbosch:
These properties can be used to post-process Findings or write custom reports.
I would't make `name` private.
We could remove the className or make it a function which will `searchClass` on demand to safe some cpu cycles though.

BraisGabin:
The problem of making it a function (or just a property with `get()`) is that, some times, we are losing the `PsiElement` (because it's not a `KtElement`) so we can't compute the `className` later. I would vote to remove.

Issue #1089, opened 2018-09-05
Cases files are super rigorous to work with


vanniktech:
It's so hard to read the unit tests. I'm opening one and need to tangle my way down on so many files without proper IDE support.

```kotlin
val path = Case.UnnecessaryAbstractClassPositive.path()
val findings = subject.lint(path)
```

I can click on `UnnecessaryAbstractClassPositive` which is fine. Then I'm seeing okay I need to go there `/cases/UnnecessaryAbstractClassPositive.kt`. This I can't click with the IDE and need to manually go down.

To make things worse there are two `UnnecessaryAbstractClassPositive.kt` classes.

```
./detekt-rules/src/test/resources/cases/UnnecessaryAbstractClassPositive.kt
./detekt-rules/build/resources/test/cases/UnnecessaryAbstractClassPositive.kt
```

Luckily they are the same though. By the time I'm at the case file I've already lost what I wanted to look up.

Even worse is the fact that a single unit test often checks everything. Some are split up between Positive and Negative although it does not make much of a difference.

Some of the Postive files then help themselves with comments like `// violation: no abstract members`. I don't know whether I can rely on them or not since sometimes they are present sometimes not and I would not expect them to be updated when tweaking a rule. Those things slip understandably in PR reviews.

What I propose instead is to have everything in one file and only one violation being tested at a time. With Android Lint this looks like this:

```kotlin
lint()
    .files(xml("res/layout/ids.xml", """
      <TextView
          xmlns:android="http://schemas.android.com/apk/res/android"
          android:paddingTop="16dp"
          android:paddingBottom="16dp"
          android:paddingStart="16dp"
          android:paddingEnd="16dp"/>""").indented())
    .issues(ISSUE_SUPERFLUOUS_PADDING_DECLARATION)
    .run()
    .expect("""
      |res/layout/ids.xml:1: Warning: Should be using padding instead. [SuperfluousPaddingDeclaration]
      |<TextView
      |^
      |0 errors, 1 warnings""".trimMargin())
```

or 

```kotlin
lint()
    .files(xml("res/layout/ids.xml", """
      <TextView
          xmlns:android="http://schemas.android.com/apk/res/android"
          android:paddingTop="8dp"
          android:paddingBottom="16dp"
          android:paddingStart="16dp"
          android:paddingEnd="16dp"/>""").indented())
    .issues(ISSUE_SUPERFLUOUS_PADDING_DECLARATION)
    .run()
    .expectClean()
```

[...]

PR #2084, opened 2019-11-13
Deprecate Location.locationString

BraisGabin:
This PR is a continuation of #2014 

This code is a bit strange: https://github.com/arturbosch/detekt/pull/2083/files#diff-3d9aab0c44e865c3fbae3f5f7a8edc8dR67 for that reason I research a bit about `Location.locationString` and it's not used in the production code. The only uses ar in the test code.

In this PR I refactor the tests to don't use that value and I add the `@Deprecate` annotation to that property.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
@@ -27,6 +27,7 @@ interface HasEntity {
     val entity: Entity
     val location: Location
         get() = entity.location
+    @Deprecated("Will be removed in the future. Use queries on 'ktElement' instead.")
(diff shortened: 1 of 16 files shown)


BraisGabin:
What do you think about a PR to talk about how to test a rule? The tests right now are SO different between them... There's no need to refactor all of them now but at least we should know how we want them.

schalkms:
@BraisGabin we have an existing issues regarding the rigorous case files (#1089).

> What do you think about a PR to talk about how to test a rule?

It might be worth to create an issue for that. Docs would also be nice. Maybe that's a thing for the contributing guide in the `Contributing.md` file.

arturbosch:
@schalkms I highly agree that we should a paragraph about the prefered way of testing detekt rules :+1:

PR #2090, opened 2019-11-16
Refactor ComplexMethod

### LINKED ISSUE #1921: Calculate MCC (McCabe Complexity) accordingly
Your feature request may already be reported!
Please search in the [issues](https://github.com/arturbosch/detekt/issues) before creating a new one.

## Expected Behavior
The cyclomatic complexity should be calculated with reference to the control flow graph of the program.

## Current Behavior
Some statements that introduce are new path are still not counted.

## Context

- [ ] Refactor `cases/ComplexClass.kt`!!
- [x] elvis operator `?:`
- [x] `&&` and `||` not only in if statements but rather in every binary-expression
- [x] jumps (`break` / `continue`)
- [x] `throw`
- [x] ~~safe call operator `?.` (can only be accurately calculated using Java byte code)~~

reference #1793

arturbosch:
Now considers
- &&, || and ?: everywhere
- break, continue (throw is explicitly not counted like in ndepend)
- functions used to replace `if` or `for` are now configurable
- threshold increased to 15 like ndepend does - also linked their page

Closes #1921


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
```diff
@@ -46,47 +51,61 @@ class ComplexMethod(
...
-                else -> false
-            }
+    private fun hasSingleWhenExpression(bodyExpression: KtExpression?): Boolean = when {
+        bodyExpression is KtBlockExpression && bodyExpression.statements.size == 1 -> {
+            val statement = bodyExpression.statements.single()
+            statement is KtWhenExpression || statement.returnsWhenExpression()
         }
-        return false
+        // the case where function-expression syntax is used: `fun test() = when { ... }`
+        bodyExpression is KtWhenExpression -> true
```
schalkms: Do we want to have a test for this case?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
```diff
@@ -46,47 +51,61 @@ class ComplexMethod(
...
-            Debt.TWENTY_MINS)
+        Severity.Maintainability,
+        "Prefer splitting up complex methods into smaller, easier to understand methods.",
+        Debt.TWENTY_MINS)
 
     private val ignoreSingleWhenExpression = valueOrDefault(IGNORE_SINGLE_WHEN_EXPRESSION, false)
     private val ignoreSimpleWhenEntries = valueOrDefault(IGNORE_SIMPLE_WHEN_ENTRIES, false)
+    private val ignoreNestingFunctions = valueOrDefault(IGNORE_NESTING_FUNCTIONS, false)
+    private val nestingFunctions =
+        CommaSeparatedPattern(valueOrDefault(NESTING_FUNCTIONS, ""), delimiters = ",;")
```
schalkms: I think this is the first config option that can have a semicolon `;` as a delimiter.
Shouldn't we also stick to the standard delimiter for this case?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
```diff
@@ -21,22 +22,26 @@ import org.jetbrains.kotlin.psi.KtWhenExpression
...
  * - __Exceptions__ - `catch`, `use`
  * - __Scope Functions__ - `let`, `run`, `with`, `apply`, and `also` ->
  *  [Reference](https://kotlinlang.org/docs/reference/scope-functions.html)
  *
- * @configuration threshold - McCabe's Cyclomatic Complexity (MCC) number for a method (default: `10`)
+ * @configuration threshold - McCabe's Cyclomatic Complexity (MCC) number for a method (default: `15`)
  * @configuration ignoreSingleWhenExpression - Ignores a complex method if it only contains a single when expression.
  * (default: `false`)
  * @configuration ignoreSimpleWhenEntries - Whether to ignore simple (braceless) when entries. (default: `false`)
+ * @configuration ignoreNestingFunctions - Whether to ignore functions which are often used instead of an 'if' or
```
schalkms:
```suggestion
 * @configuration ignoreNestingFunctions - Whether to ignore functions which are often used instead of an `if` or
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
```diff
@@ -21,22 +22,26 @@ import org.jetbrains.kotlin.psi.KtWhenExpression
...
+ * 'for' statement (default: `false`)
```
schalkms:
```suggestion
 * `for` statement (default: `false`)
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethodSpec.kt
```diff
@@ -11,20 +11,49 @@ import org.spekframework.spek2.Spek
...
 class ComplexMethodSpec : Spek({
+
     describe("ComplexMethod rule") {
 
         context("a complex method") {
 
             it("finds one complex method") {
-                val subject = ComplexMethod()
-                subject.lint(Case.ComplexMethod.path())
+                val findings = ComplexMethod(threshold = 10).lint(Case.ComplexMethod.path())
```
schalkms: This case file is still a pain in the ass. `Case.ComplexMethod.path()`

schalkms: However, this should be done in a separate PR.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/CyclomaticComplexity.kt
```diff
@@ -0,0 +1,130 @@
...
+        private set
+
+    override fun visitNamedFunction(function: KtNamedFunction) {
+        if (!isInsideObjectLiteral(function)) {
+            complexity++
+            super.visitNamedFunction(function)
+        }
+    }
+
+    private fun isInsideObjectLiteral(function: KtNamedFunction): Boolean =
```
schalkms:
```suggestion
    private fun isInsideObjectLiteral(function: KtNamedFunction) =
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
arturbosch: You are right, I thought the `SplitPattern` does support `,;` by default. Will change this, thx!

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
arturbosch: Yep we already have some in the end of the Spec file.

arturbosch:
> I particularly like the config implementation for the CyclomaticComplexity class.
> TIL that you can now use Kotlin functions inside Spec's `context()` functions. Cool!
> As I commented, the `ComplexMethod` case file is very hard to understand and maintain.
> Hopefully, we can also approach at some point in the future.
> For now, this PR looks very fine!

Thanks! I've refactored some test cases and removed that ComplexMethod case file :)

PR #2094, opened 2019-11-17
Improve naming tests

BraisGabin:
This PR is a general refactor in the naming tests. Split tests in different files, simplify some, fix typos, reduce the scope of others...

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ConstructorParameterNamingSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ConstructorParameterNamingSpec.kt
@@ -1,0 +1,44 @@
+package io.gitlab.arturbosch.detekt.rules.naming
+
+import io.gitlab.arturbosch.detekt.test.assertThat
+import io.gitlab.arturbosch.detekt.test.compileAndLint
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+
+class ConstructorParameterNamingSpec : Spek({
+
+    describe("parameters in a constructor of a class") {
+
+        it("should detect no violations") {
+            val code = """
+                class C(val param: String, private val privateParam: String)
+
+                class D {
+                    constructor(param: String) {}
+                    constructor(param: String, privateParam: String) {}
+                }
+            """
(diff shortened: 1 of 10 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ConstructorParameterNamingSpec.kt
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        it("should find some violations") {
+            val code = """
+                class C(val PARAM: String, private val PRIVATE_PARAM: String)
+
+                class C {
+                    constructor(PARAM: String) {}
+                    constructor(PARAM: String, PRIVATE_PARAM: String) {}
+                }
```
BraisGabin: How does this compile? Two classes with the same name.

BraisGabin: How does this compile? Two classes with the same name.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NamingRulesSpec.kt
```diff
@@ -1,104 +1,22 @@
...
                 data class D(val i: Int, val j: Int)
                 fun doStuff() {
                     val (_, HOLY_GRAIL) = D(5, 4)
-                    emptyMap<String, String>().forEach { _, V -> println(v) }
+                    emptyMap<String, String>().forEach { _, V -> println(V) }
                 }
             """
-            assertThat(subject.lint(code)).isEmpty()
+            assertThat(NamingRules().compileAndLint(code)).isEmpty()
         }
```
BraisGabin: I'd like to Remove NamingRulesSpec and move this to its own rule spec... But I don't know which rule is related with this test.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ConstructorParameterNamingSpec.kt
arturbosch: Good question. Maybe the type checking step will first trigger redundant names?

Issue #2096, opened 2019-11-17
Deprecate ignoreOverriddenFunction(s) in favor fo ignoreOverridden


BraisGabin:
`EmptyFunctionBlock`, `FunctionParameterNaming` and `MemberNameEqualsClassName` uses the configuration flag `ignoreOverriddenFunction(s)` but the rest of rules uses the generic `ignoreOverridden`.

I cheked the uses and I think that its safe to rename those three to `ignoreOverridden` too. I think that this change will make the API a bit more consistent.

This is clearly a breaking change but we can deprecate `ignoreOverriddenFunction` but honor it. We can show a warning too if someone uses it. And we can remove it later.

What do you think?

schalkms:
I appreciate consistency across our rules configuration.
However, in this case it doesn't bring that much to the table.
`MemberNameEqualsClassName:ignoreOverriddenFunction` should stay, since it isn't clear from the rule name what should be ignored.
If @arturbosch agrees, I think we can integrate this change, but not before v1.3.

BraisGabin:
I understand your point but... Why do we allow overriden functions and don't allow overriden properties too? Probably we should change a bit what `MemberNameEqualsClassName:ignoreOverriddenFunction` does to add the properties and then rename it to `ignoreOverridden` as the other ones.

schalkms:
> Why do we allow overriden functions and don't allow overriden properties too? Probably we should change a bit what MemberNameEqualsClassName:ignoreOverriddenFunction does to add the properties and then rename it to ignoreOverridden as the other ones.

Yeah, thanks for your input. You are completely right. This rule should also ignore overridden properties.

PR #2104, opened 2019-11-19
MemberNameEqualsClassName should ignore overridden property names too

BraisGabin:
Right now `MemberNameEqualsClassName` only allows to ignore overridden functions. This PR change that to ignore both: functions and properties.

This PR is related with this issue: #2096

In the future we should rename the flag `ignoreOverriddenFunction` to `ignoreOverridden`. But that's a breaking change. We are talking about that at #2096


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassName.kt
```diff
@@ -82,22 +80,20 @@ class MemberNameEqualsClassName(config: Config = Config.empty) : Rule(config) {
...
     }
 
-    private fun getMisnamedMembers(klassOrObject: KtClassOrObject, name: String?): List<KtNamedDeclaration> {
-        val body = klassOrObject.body ?: return emptyList()
-        val declarations = getFunctions(body).concat(body.properties)
-        return declarations?.filter { it.name?.equals(name, ignoreCase = true) == true }.orEmpty()
+    private fun getMisnamedMembers(klassOrObject: KtClassOrObject, name: String?): Sequence<KtNamedDeclaration> {
+        val body = klassOrObject.body ?: return emptySequence()
+        return (getFunctions(body) + body.properties)
+            .filter { !ignoreOverriddenFunction || !it.isOverride() }
```
3flex:
```suggestion
            .filterNot { ignoreOverriddenFunction && it.isOverride() }
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassName.kt
```diff
@@ -82,22 +80,20 @@ class MemberNameEqualsClassName(config: Config = Config.empty) : Rule(config) {
         super.visitObjectDeclaration(declaration)
     }
 
-    private fun getMisnamedMembers(klassOrObject: KtClassOrObject, name: String?): List<KtNamedDeclaration> {
-        val body = klassOrObject.body ?: return emptyList()
-        val declarations = getFunctions(body).concat(body.properties)
-        return declarations?.filter { it.name?.equals(name, ignoreCase = true) == true }.orEmpty()
+    private fun getMisnamedMembers(klassOrObject: KtClassOrObject, name: String?): Sequence<KtNamedDeclaration> {
+        val body = klassOrObject.body ?: return emptySequence()
+        return (getFunctions(body) + body.properties)
```
3flex:
```suggestion
        return (body.functions + body.properties)
```

There's a `functions` property on `KtClassBody` which might be used instead of the private `getFunctions` function?

https://github.com/JetBrains/kotlin/blob/v1.3.60/compiler/psi/src/org/jetbrains/kotlin/psi/KtClassBody.kt#L46-L47

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassNameSpec.kt
```diff
@@ -22,28 +28,125 @@ class MemberNameEqualsClassNameSpec : Spek({
...
+                val code = """
+                    class MethodNameContainer {
+                        class MethodNameEqualsNestedClassName {
+                            fun MethodNameEqualsNestedClassName() {}
+                        }
+                    }
+                """
+                assertThat(MemberNameEqualsClassName().compileAndLint(code)).hasSize(1)
+            }
+
```
3flex: Looks like this example wasn't migrated from the existing `MemberNameEqualsClassNamePositive` file:

```kotlin
class WrongFactoryClass1 {

    companion object {
        fun wrongFactoryClass1() {} // reports 1 - no return type
    }
}
```

BraisGabin: I thought that both tests was testing exactly the same thing for that reason I removed one. I added it again.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassName.kt
BraisGabin: Done

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassName.kt
BraisGabin: Done

BraisGabin:
The fail seems flaky

PR #2105, opened 2019-11-19
Don't compile test snippets

### LINKED ISSUE #2095: compileAndLint is 2.5 times slower than lint
I did a fast check and the line `KotlinScriptEngine.compile(content)` in `compileAndLint` slows down the tests a lot. In my computer the gradle task `:detekt-rules:test` takes ~82s with `compileAndLint` and ~32s with `lint` (2.5x faster).

Don't get me wrong: I like the safety that `compileAndLint` bring us. I move some test to `compileAndLint` and there was a lot of code that didn't compile. But this `compileAndLint` is not testing the production code. It's testing our tests.

For all that I said before, I think that we should run the line `KotlinScriptEngine.compile(content)` just in CI. This way we get the good of both parts. Fast tests in our end to test the production code and we ensure that the tests are correct in CI.

What do you think? This should be easy to accomplish with gradle.

BraisGabin:
Closes #2095

We need to check that our production is OK every time. But we don't need to check that our tests are OK every time too.

This change speeds up the tests about 2.5 times.

What do you think about the naming? I'm not sure...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,18 +7,22 @@ matrix:
     - os: linux
       dist: trusty
       jdk: oraclejdk8
+      env: COMPILE_TEST_SNIPPETS=true
     - os: linux
       dist: bionic
       jdk: openjdk11
+      env: COMPILE_TEST_SNIPPETS=false
     - os: linux
       dist: bionic
       jdk: openjdk12
+      env: COMPILE_TEST_SNIPPETS=false
     # JDK 8 - see https://docs.travis-ci.com/user/reference/osx/#jdk-and-macos
     - os: osx
       osx_image: xcode9.3
+      env: COMPILE_TEST_SNIPPETS=false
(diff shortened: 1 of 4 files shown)


schalkms:
That's interesting. It works for AppVeyor but not for Travis?

BraisGabin:
Here are the builds testing that if one of your snippet doesn't compile the CI complains abot it:
- [Travis](https://ci.appveyor.com/project/arturbosch/detekt/builds/28976449)
- [AppVenyor](https://ci.appveyor.com/project/arturbosch/detekt/builds/28976449): it failed in another run too... but I think that it was just flaky.

This is ready to merge if you are OK with the naming.

BoxResin:
There're some conflicts to resolve.

BraisGabin:
Rebased to fix the conflicts

PR #2106, opened 2019-11-19
UnusedImports: False positive if referenced in @throws/@exception/@sample

### LINKED ISSUE #2098: UnusedImports: False positive if referenced only in @throws/@exception clause
The bug you're experiencing might have already be reported!
Please search in the [issues](https://github.com/arturbosch/detekt/issues) before creating one.

## Expected Behavior
The UnusedImports rule should not warn if an imported class is only referencedf in a @throws or @exception clause in KDoc.

## Observed Behavior
False positive

From looking at the source code: The rule only checks for @see and for inline references.

## Your Environment

* Version of detekt used: 1.1.1

BraisGabin:
UnusedImports: Fix false positive

When referenced in `@throws`, `@exception` or `@sample` in Kdoc block tag.

Fix #2098


schalkms:
Thanks for the quick fix!

PR #2109, opened 2019-11-20
Update to Kotlin v1.3.60

schalkms:
Changelog:
https://github.com/JetBrains/kotlin/blob/1.3.60/ChangeLog.md


schalkms:
It's interesting that the code coverage went down.
There might be changes in the generated code by the newest Kotlin compiler.

PR #2113, opened 2019-11-21
Return non-nullable string in DebtSumming class

schalkms:
Since StringBuilder.toString() can't return null, both methods in the
DebtSumming class also can't return null.


schalkms:
TravisCI: Mac OS seems to have problems with JDK

arturbosch:
The build log tells us that Java 13 is used which is only supported by Gradle 6.0+

PR #2117, opened 2019-11-22
Create rule: ForbiddenPublicDataClass

### LINKED ISSUE #2115: NewRule: ForbiddenPublicDataClasses
Use data classes is a bad pattern for public APIs in libraries: https://jakewharton.com/public-api-challenges-in-kotlin/

TL;DR: If you expose data classes is too easy to create binary-incompatible bytecode in next versions of your library.

We can add a rule that checks for all the data classes and ensure that they are `private` or `internal`.
We should add a flag too to allow data classes in certain packages. And here I can think about two different options:

* boolean flag `ignoreInternalPackage` and ignore all the classes inside a package `*.internal.*`
* string flag `ignorePackage` and allow to set any packages.

I like the first a bit more because it forces the use of `internal` package. That it's kind of a standard solution. But the second gives more freedom to the user, and that's good too.

For sure, this rule should be disable by default because it only have sense for library developers. If you are creating a final application you are not going to have any binary-compatibility problem.

BraisGabin:
Closes #2115 

If you expose data classes is too easy to create binary-incompatible bytecode in next versions of your library. This rule allows the library maintainers to ensure that they avoid that problem.

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -455,6 +455,8 @@ style:
     active: false
     imports: ''
     forbiddenPatterns: ""
+  ForbiddenPublicDataClass:
+    active: false
   ForbiddenVoid:
     active: false
     ignoreOverridden: false
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -457,6 +457,7 @@ style:
     forbiddenPatterns: ""
   ForbiddenPublicDataClass:
     active: false
+    ignorePackages: '*.internal,*.internal.*'
   ForbiddenVoid:
     active: false
     ignoreOverridden: false
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenPublicDataClass.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenPublicDataClass.kt
@@ -26,6 +26,9 @@ import org.jetbrains.kotlin.psi.psiUtil.visibilityModifierTypeOrDefault
  * <compliant>
  * internal data class C(val a: String)
  * </compliant>
+ *
+ * @configuration ignorePackages - ignores classes in the specified packages. Split by commas ','
+ * (default: `'*.internal,*.internal.*'`)
  */
 class ForbiddenPublicDataClass(config: Config = Config.empty) : Rule(config) {
 
@@ -36,7 +39,22 @@ class ForbiddenPublicDataClass(config: Config = Config.empty) : Rule(config) {
         Debt.TWENTY_MINS
     )
 
+    private val ignorablePackages = valueOrDefault(IGNORE_PACKAGES, "*.internal,*.internal.*")
+        .splitToSequence(",")
+        .map { it.trim() }
+        .filter { it.isNotBlank() }
+        .map { it.replace(".", "\\.") }
+        .map { it.replace("*", ".*") }
+        .map { Regex(it) }
+        .toList()
+
     override fun visitClass(klass: KtClass) {
+        val packageName = klass.containingKtFile.packageDirective?.packageNameExpression?.text
+
+        if (packageName != null && ignorablePackages.any { it.matches(packageName) }) {
+            return
+        }
+
         val isPublicOrProtected = klass.visibilityModifierTypeOrDefault().let { visibility ->
             visibility != KtTokens.INTERNAL_KEYWORD && visibility != KtTokens.PRIVATE_KEYWORD
         }
@@ -46,5 +64,9 @@ class ForbiddenPublicDataClass(config: Config = Config.empty) : Rule(config) {
             }
             super.visitClass(klass)
         }
+    }
+
+    companion object {
+        const val IGNORE_PACKAGES = "ignorePackages"
     }
 }
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenPublicDataClass.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenPublicDataClass.kt
@@ -11,6 +11,7 @@ import io.gitlab.arturbosch.detekt.api.commaSeparatedPattern
 import io.gitlab.arturbosch.detekt.api.simplePatternToRegex
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.psiUtil.visibilityModifierTypeOrDefault
 
 /**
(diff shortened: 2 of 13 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenPublicDataClass.kt
```diff
@@ -0,0 +1,50 @@
...
+    override fun visitClass(klass: KtClass) {
+        val isPublicOrProtected = klass.visibilityModifierTypeOrDefault().let { visibility ->
+            visibility != KtTokens.INTERNAL_KEYWORD && visibility != KtTokens.PRIVATE_KEYWORD
+        }
+        if (klass.isData() && isPublicOrProtected) {
+            report(CodeSmell(issue, Entity.from(klass.identifyingElement ?: klass), ""))
+        }
+        if (isPublicOrProtected) {
+            super.visitClass(klass)
+        }
```
schalkms: You could rewrite this in order to not check `isPublicOrProtected` twice.

```Kotlin
if (isPublicOrProtected) {
    if (klass.isData()) {
       report(CodeSmell(issue, Entity.from(klass.identifyingElement ?: klass), "")) 
    }
    super.visitClass(klass)
}
```

schalkms:
I'm curious. What's missing for this rule to be checked in?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenPublicDataClass.kt
BraisGabin: You are right, fixed.

BraisGabin:
I haden't sawn your reply in the issue. I'd implement the configuration ignorePackage and this PR will be ready.

detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -455,6 +455,9 @@ style:
     active: false
     imports: ''
     forbiddenPatterns: ""
+  ForbiddenPublicDataClass:
+    active: false
+    ignorePackages: '*.internal,*.internal.*'
```
arturbosch: Isn't `.*internal` a subset of `.*internal.*`?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenPublicDataClass.kt
```diff
@@ -0,0 +1,72 @@
...
+        javaClass.simpleName,
+        Severity.Style,
+        "The data classes are bad for the binary compatibility in public APIs. Avoid to use it.",
+        Debt.TWENTY_MINS
+    )
+
+    private val ignorablePackages = valueOrDefault(IGNORE_PACKAGES, "*.internal,*.internal.*")
+        .splitToSequence(",")
+        .map { it.trim() }
+        .filter { it.isNotBlank() }
```
arturbosch: You could use our `CommaSeparatedPattern` which does the first three steps sequence steps for us and then `mapAll`.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenPublicDataClass.kt
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        val isPublicOrProtected = klass.visibilityModifierTypeOrDefault().let { visibility ->
+            visibility != KtTokens.INTERNAL_KEYWORD && visibility != KtTokens.PRIVATE_KEYWORD
+        }
+        if (isPublicOrProtected) {
+            if (klass.isData()) {
+                report(CodeSmell(issue, Entity.from(klass.nameIdentifier ?: klass), ""))
+            }
+            super.visitClass(klass)
```
arturbosch: Should't this always be visited for nested classes?

detekt-cli/src/main/resources/default-detekt-config.yml
arturbosch: Ah I see this is about the dot between packages and you replace it inside the rule.
Hm, we let users define regexes for other properties too, should we stick here also to plain regexes? `.*\\.internal\\..*` ?

arturbosch: The rule code could be `CommaSeparatedPattern(...).mapToRegex()`.

schalkms:
I agree with you @arturbosch

detekt-cli/src/main/resources/default-detekt-config.yml
BraisGabin: The regex is more flexible but it's harder to mantain for the users... I mean: This `*.internal,*.internal.*` is easier to read than this: `.*\\.internal($|\\..*)`. I see pros and cons to both options. Maybe we can support both: `ignorePackages` and `ignorePackagesRegex`. What do you think?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenPublicDataClass.kt
BraisGabin: If the class is private or internal any nested will be at least private or internal too. Even if you set `public` on it. (Correct me if I'm wrong here).

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenPublicDataClass.kt
BraisGabin: I didn't use `CommaSeparatedPattern` because it's a "heavy" class. What do you think about my solution? Is all in the new commit. Basically I use extension functions to not repeat the code over and over again. But the extension functions are really simple so you can compose them.

BraisGabin:
> As this is aimed to library authors and a multi module project could also have `api` modules like `detekt-api` should we update the `visitCondition` to be exactly like `LibraryCodeMustSpecifyReturnType`?

I copy&pasted the `visitCondition` from `LibraryCodeMustSpecifyReturnType` but, to be honest, I'm not sure what this does... Are the updates that I did in the test OK?

BraisGabin:
Should I do something else here?

schalkms:
It’s okay from my point of view. I’m waiting for Artur’s final approval.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenPublicDataClassSpec.kt
```diff
@@ -0,0 +1,147 @@
...
+            val code = """
+                package com.example.internalise
+                
+                data class C(val a: String)                
+            """
+
+            assertThat(subject.compileAndLint(code)).hasSize(1)
+        }
+
+        it("public data class inside a random should fail") {
```
schalkms: internal package

BraisGabin: If you point to this line I have no idea what are you mean. If you talk about "internalise" it's not a typo. I want to check that a package that starts with internal but it's not just internal is not treated as an exception.

schalkms: Sorry for the confusion

```suggestion
        it("public data class inside a random package should fail") {
```

schalkms:
`@Ignore("LongMethod")` for the `StyleGuideProvider` class popped up.

PR #2107, opened 2019-11-20
Fix formatting rule signatures

arturbosch:
TODO:
```
CorrectableCodeSmell(autoCorrectEnabled=true,issue=Issue(id='NoBlankLineBeforeRbrace', severity=Style, debt=5min), entity=Entity(name=parseTestResult, className=SurefireStaxHandler, signature=SurefireStaxHandler.kt$SurefireStaxHandler$ , location=Location(source=112:57, text=4775:4793, locationString=(112, 57), file=/home/artur/git/repos/sonar-kotlin/src/main/kotlin/io/gitlab/arturbosch/detekt/sonar/surefire/data/SurefireStaxHandler.kt), ktElement=BLOCK), message=Unexpected blank line(s) before "}", metrics=[], references=[], id='NoBlankLineBeforeRbrace')
java.lang.StringIndexOutOfBoundsException: begin 0, end 56, length 17
        at java.base/java.lang.String.checkBoundsBeginEnd(String.java:3319)
        at java.base/java.lang.String.substring(String.java:1874)
        at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt.writeErrorLine(HtmlUtils.kt:38)
        at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt.access$writeErrorLine(HtmlUtils.kt:1)
        at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt$snippetCode$1$1.invoke(HtmlUtils.kt:25)
        at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt$snippetCode$1$1.invoke(HtmlUtils.kt)
        at kotlinx.html.ApiKt.visit(api.kt:80)
        at kotlinx.html.Gen_tag_unionsKt.code(gen-tag-unions.kt:161)
        at kotlinx.html.Gen_tag_unionsKt.code$default(gen-tag-unions.kt:161)
        at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt$snippetCode$1.invoke(HtmlUtils.kt:13)
        at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt$snippetCode$1.invoke(HtmlUtils.kt)
        at kotlinx.html.ApiKt.visit(api.kt:80)

```

PR #2120, opened 2019-11-23
Revert change to calculate signature for formatting rules due to performance reasons

arturbosch:
Reverts #2083 and #2107

Tested on detekt's codebase with cli and command `jrun detekt-cli/build/libs/detekt-cli-1.2.0-all.jar   -in "**/src/main/kotlin/**/*.kt" -i .`
detekt run without plugins ~5s
detekt run with formatting plugin and no signature reported ~8s
detekt run with formatting plugin and signature ~12s

Formatting is slower and with the recent change it got even slower.


detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
```diff
@@ -66,22 +66,17 @@ abstract class FormattingRule(config: Config) : Rule(config) {
...
             )
-            // nodes of 'NoConsecutiveBlankLines' are dangling whitespace nodes
-            val element = getIfValidSignatureCanBeCreatedOrNull(node.psi)
-            val entity = if (element == null) {
-                Entity("", "", "", location)
-            } else {
-                Entity.from(element, location)
-            }
+            // The formatting rules report slightly wrong positions.
+            // As of 23.11.2019 we haven't debugged if our wrapping logic or KtLint report these wrong positions.
```
sschuberth: See https://github.com/arturbosch/detekt/issues/1843, it's not ktlint's fault. Would you mind updating the comment accordingly?

arturbosch: Thanks, will do ^^

arturbosch:
Build is failing for OSX due to Java 13. Why does osx 12.3 use Java 13 which is unsupported by Gradle 5.6.X ...

PR #2121, opened 2019-11-23
Use Java 11 for osx build until we update to Gradle 6

arturbosch:


arturbosch:
Somehow Java 13 is pulled in leading Gradle 5.6 to fail.

PR #2123, opened 2019-11-24
Add MagicNumber:ignoreLocalVariableDeclaration config

schalkms:
This config option wasn't correctly documented.
The generator didn't pick it up.
Hence, the default-detekt-config.yml file didn't include this option.


arturbosch:
Good catch!

PR #2137, opened 2019-11-27
Don't compile test snippets (bindingContext)

schalkms:
follow-up: #2105


schalkms:
Yes. This further speeds up detekt's CI. The Jsr223 compiles engines aren't even created at startup anymore if the environment variable `compile-snippet-tests` is set to `false`.

PR #2140, opened 2019-11-28
Document how to create a common baseline file for multi module gradle projects

### LINKED ISSUE #2100: DetektAll with baseline fails with error
I have a multi-module project and I want to create one baseline file for the entire project instead of one per module, so I'm trying to replicate the detektAll task that is created in this project's build.gradle.kts file. However, I have had to translate this task into Groovy, and it looks like this:
```
task detektAll(type: io.gitlab.arturbosch.detekt.Detekt) {
    description = 'Runs over whole code base without the starting overhead for each module.'
    config = files("$rootDir/detekt-config.yml")
    parallel = true
    input = files(projectDir)
    include '**/*.kt'
    include '**/*.kts'
    exclude '**/resources/**'
    exclude '**/build/**'
    baseline = file("$rootDir/config/detekt/baseline.xml")
    reports {
        xml {
            enabled = true
        }
        html {
            enabled = true
        }
        txt {
            enabled = true
        }
    }
}
```
However, running this fails with the error  
```
A problem was found with the configuration of task ':detektAll'.
File (my root directory)/config/detekt/baseline.xml does not exist.
```
Is there a fix for this?

arturbosch:
Closes #2100

### DIFF
--- a/docs/pages/baseline.md
+++ b/docs/pages/baseline.md
@@ -31,7 +31,68 @@ the `--report txt:path/to/report` cli flag.
 </SmellBaseline>
 ```
 
+#### Gradle
+
 If you are using the gradle-plugin run the `detektBaseline` task to generate yourself a `baseline.xml`.
+This will create one baseline file per Gradle module.
+As this might not be the desired behavior for a multi module project, think about implementing
+a custom meta baseline task:
+
+Gradle-DSL
+
+```gradle
+subprojects {
+    detekt {
+        // ...
+        baseline = file("${rootProject.projectDir}/config/baseline.xml")
+        // ...
+    }
+}
+
+task detektProjectBaseline(type: io.gitlab.arturbosch.detekt.DetektCreateBaselineTask) {
+    description = "Overrides current baseline."
+    ignoreFailures.set(true)
+    parallel.set(true)
+    buildUponDefaultConfig.set(true)
+    setSource(files(rootDir))
+    config.setFrom(files("$rootDir/config/detekt/detekt.yml"))
+    baseline.set(file("$rootDir/config/detekt/baseline.xml"))
+    include("**/*.kt")
+    include("**/*.kts")
+    exclude("**/resources/**")
+    exclude("**/build/**")
+}
+
+```
+
+Kotlin-DSL
+
+```kotlin
+subprojects {
+    detekt {
+        // ...
+        baseline = file("${rootProject.projectDir}/config/baseline.xml")
+        // ...
+    }
+}
+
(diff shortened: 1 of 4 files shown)


docs/pages/baseline.md
```diff
@@ -31,7 +31,38 @@ the `--report txt:path/to/report` cli flag.
...
+    }
+}
+
+task projectBaseline(type: io.gitlab.arturbosch.detekt.DetektCreateBaselineTask) {
+    description = "Overrides current baseline."
+
+    input = files(rootProject.projectDir)
+    config = files("$rootDir/detekt/config.yml")
+    baseline = file(new File(rootProject.projectDir, "/detekt/baseline.xml"))
+    failFast = true
```
adrianczuczka: As mentioned in https://github.com/arturbosch/detekt/issues/2100, `failFast = true` this caused my baseline task to fail. I would replace this with `ignoreFailures = true`.

docs/pages/baseline.md
```diff
@@ -31,7 +31,38 @@ the `--report txt:path/to/report` cli flag.
...
+    parallel = true
+    include '**/*.kt'
+    include '**/*.kts'
+    exclude '**/test/resources/**'
+}
+```
```
adrianczuczka: Can we move the entire task to the `subprojects` block? It did not work correctly for me without doing that.

docs/pages/baseline.md
```diff
@@ -31,7 +31,38 @@ the `--report txt:path/to/report` cli flag.
...
+        ...
+    }
+}
+
+task projectBaseline(type: io.gitlab.arturbosch.detekt.DetektCreateBaselineTask) {
+    description = "Overrides current baseline."
+
+    input = files(rootProject.projectDir)
+    config = files("$rootDir/detekt/config.yml")
+    baseline = file(new File(rootProject.projectDir, "/detekt/baseline.xml"))
```
adrianczuczka: Can we replace this with `baseline = file("${rootProject.projectDir}/detekt-baseline.xml")`? As far as I know it does the same thing and as mentioned in https://github.com/arturbosch/detekt/issues/2100, there is a weird IDE error on my laptop (which does not cause the build to fail, however) related to `new File`. Also, we use the suggested syntax for both `input` and `config`.

arturbosch:
@adrianczuczka thanks for the suggestions, I've incorporated them and added a `detektProjectBaseline` task for detekt too.

docs/pages/baseline.md
arturbosch: It should work on top level like in our build file. Inside `subprojects` every sub-project would get a baseline task hm :/

PR #2141, opened 2019-11-28
Use JDK 11 (LTS) + 13 for AppVeyor builds

schalkms:
Detekt uses Gradle 6, which supports JDK 13.
Thus, we should also run a CI build for that.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,7 +17,7 @@ environment:
       JAVA_HOME: C:\Program Files\Java\jdk11
       COMPILE_TEST_SNIPPETS: false
     -
-      JAVA_HOME: C:\Program Files\Java\jdk12
+      JAVA_HOME: C:\Program Files\Java\jdk13
       COMPILE_TEST_SNIPPETS: false
 
 init:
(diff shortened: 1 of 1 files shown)


schalkms:
blocked by #2143

appveyor.yml
```diff
@@ -14,10 +14,10 @@ environment:
       JAVA_HOME: C:\Program Files\Java\jdk1.8.0
       COMPILE_TEST_SNIPPETS: true
     -
-      JAVA_HOME: C:\Program Files\Java\jdk11
+      JAVA_HOME: C:\Program Files\Java\jdk12
```
BraisGabin: Same comment: https://github.com/arturbosch/detekt/pull/2142#discussion_r351948728

schalkms: Good point!

arturbosch:
I think this needs just a rebase from master and we are good to merge it :).

3flex:
Can we drop JDK 11? AppVeyor is the slowest CI because they don't support parallel jobs, so if we can test only on earliest & latest supported JDK like macOS on Travis that should give sufficient coverage and cut down the build time a bit.

PR #2142, opened 2019-11-28
Use JDK 11 (LTS) + 13 for Travis builds

schalkms:
Detekt uses Gradle 6, which supports JDK 13.
Thus, we should also run a CI build for that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
       env: COMPILE_TEST_SNIPPETS=false
     - os: linux
       dist: bionic
-      jdk: openjdk12
+      jdk: openjdk13
       env: COMPILE_TEST_SNIPPETS=false
     # JDK 8 - see https://docs.travis-ci.com/user/reference/osx/#jdk-and-macos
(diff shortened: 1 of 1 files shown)


schalkms:
blocked by #2143

.travis.yml
```diff
@@ -10,11 +10,11 @@ matrix:
       env: COMPILE_TEST_SNIPPETS=true
     - os: linux
       dist: bionic
-      jdk: openjdk11
+      jdk: openjdk12
```
BraisGabin: I think that we should not test java 12. I think that we should test the last version of java and the LTS versions (java 8 and 11).

If someone has problems with detekt and java 12 it should be safe to ask to update to the last version or go back to the last LTS. Mantain all the versions is insane.

3flex: Agree, 8, 11 and 13 should be the supported versions (latest + all LTS versions).

arturbosch: +1

schalkms: good point

PR #2143, opened 2019-11-28
Run Gradle 5 tests only for JDK versions < 13

schalkms:


detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/MultiVersionTest.kt
```diff
@@ -26,3 +26,23 @@ object MultiVersionTest : Spek({
...
 })
+
+private fun createGradleVersionsToTest(): List<String> {
+    val versions = mutableListOf<String>()
+    val jdkVersion = getJdkVersion()
+    if (jdkVersion < 13) {
+        versions.add("5.0")
+    }
+    versions.add("6.0.1")
+    return versions
```
BraisGabin:
```kotlin
return if (jdkVersion < 13) {
  listOf("5.0", "6.0.1")
} else {
  listOf("6.0.1")
}
```

I think that it's easier to know which versions are going to be tested like this. You could even create two consts: `GRADLE_5` and `GRADLE_6`. But I think that the strings helps more to the readability than the constants to the maintanability.

arturbosch: Agree,
```kotlin
private fun getGradleVersionsUnderTest() =
    if (jdkVersion() < 13) {
        listOf("5.0", "6.0.1")
    } else {
        listOf("6.0.1")
    }
```

Issue #1894, opened 2019-09-04
How to not execute detekt task on check gradle's phase?


gregorius2514:
Hello,

my questions:
1. How to 'turn off' attaching detekt tasks to gradles 'check' task?
2. Where find contribution informations how to add new detekt properties (any the most important common problems with it)?

context:

I have problem when I'm building my project detekt is executed (on check phase) and build fails.
I would like to have more controll on detekt gradle's task so I looking to execute detek's task in CI in separate step.

I looked for any configure property but I did find any so I try to make changes in detekt's source code to make that configurable feature.

Right now I have problem when my properties injected from build.gradle file are not visible inside DetektExcension, DetektPlugin classes (default values from DetektExcension are always getting). Does any one have similar problem?

Best regards

schalkms:
I'm afraid, I don't quite understand your problem description. Could you please explain that in more detail?

> How to 'turn off' attaching detekt tasks to gradles 'check' task?

Could you please post your of the Gradle build script that configures detekt?

> Where find contribution informations how to add new detekt properties (any the most important common problems with it)?

I'm also not sure what you mean.
The rules can be added in the `default-detekt-config.yml` file [Source](https://arturbosch.github.io/detekt/configurations.html).
All of detekt's parameters are listed [here](https://arturbosch.github.io/detekt/groovydsl.html) in the getting started guide.

> Right now I have problem when my properties injected from build.gradle file are not visible inside DetektExcension

What exactly are you trying to do? Could you also please post the relevant Gradle build script part here?

schalkms:
I'm afraid, I don't quite understand your problem description. Could you please explain that in more detail?

> How to 'turn off' attaching detekt tasks to gradles 'check' task?

Could you please post the part of your Gradle build script that configures detekt?

> Where find contribution informations how to add new detekt properties (any the most important common problems with it)?

I'm also not sure what you mean.
The rules can be added in the `default-detekt-config.yml` file ([Source](https://arturbosch.github.io/detekt/configurations.html)).
All of detekt's parameters are listed [here](https://arturbosch.github.io/detekt/groovydsl.html) in the getting started guide.

> Right now I have problem when my properties injected from build.gradle file are not visible inside DetektExcension

What exactly are you trying to do? Could you also please post the relevant Gradle build script part here?

marschwar:
> How to 'turn off' attaching detekt tasks to gradles 'check' task?

[...]

PR #2144, opened 2019-11-29
Document how to exclude detekt from the check task - #1894

arturbosch:
The groovy dsl page has more info sub sections and the kotlin dsl page points to the groovy dsl page.
So it's okay not to duplicate the info ^^

### DIFF
--- a/docs/pages/gettingstarted/groovydsl.md
+++ b/docs/pages/gettingstarted/groovydsl.md
@@ -147,6 +147,24 @@ task detektFailFast(type: io.gitlab.arturbosch.detekt.Detekt) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/pages/gettingstarted/groovydsl.md
```diff
@@ -147,6 +147,24 @@ task detektFailFast(type: io.gitlab.arturbosch.detekt.Detekt) {
 }
 ```
 
+##### <a name="check-lifecycle">Disabling detekt from the check task</a>
+
+detekt tasks by default are verification tasks. They get executed whenever the Gradle check task gets executed.
```
schalkms:
```suggestion
Detekt tasks by default are verification tasks. They get executed whenever the Gradle check task gets executed.
```

PR #2174, opened 2019-12-12
Add GlobalCoroutineUsage rule + coroutines ruleset

### LINKED ISSUE #2155: Add rule [GlobalScopeUsage] to disallow usage of GlobalScope
`GlobalScope` should be avoided

The rule detects all usages of `GlobalScope`.
```
GlobalScope.launch{ doSomethingSuspending() }
```
Instead a `CoroutineScope` should be used.

## Context

> Application code usually should use an application-defined CoroutineScope. Using async or launch on the instance of GlobalScope is highly discouraged.

- https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/index.html

- https://medium.com/@elizarov/the-reason-to-avoid-globalscope-835337445abc

xouabita:
- Add `GlobalScopeUsage` rule which reports use of `GlobalScope`.
See https://github.com/arturbosch/detekt/issues/2155 for more context
- Add `concurrency` rule set
- Add a way to pass `optionalPaths` in create environment
- Add `coroutines` as a test dependency to have coroutine types in the test

Closes #2155


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/concurrency/GlobalScopeUsage.kt
```diff
@@ -0,0 +1,54 @@
...
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.descriptors.ClassDescriptor
+import org.jetbrains.kotlin.psi.KtSimpleNameExpression
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getType
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+
+/**
+ * Report usages of GlobalScope. Usage of GlobalScope is highly discouraged by the Kotlin documentation.
+ * See https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-global-scope/
```
schalkms: For my taste we should provide a bit more context in our documentation and then reference to the original documentation at the end. The documentation should be meaningful, even if the user is offline, when executing detekt.
I would just write a short excerpt (1-2 sentences) from the original Kotlin doc.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/concurrency/GlobalScopeUsage.kt
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * <noncompliant>
+ * GlobalScope.launch { delay(1_000L) }
+ * </noncompliant>
+ *
+ * <compliant>
+ * CoroutineScope(Dispatchers.Default).launch { delay(1_000L) }
+ * </compliant>
+ */
+class GlobalScopeUsage(config: Config = Config.empty) : Rule(config) {
```
schalkms: The documentation states

> Using async or launch on the instance of GlobalScope is highly discouraged.

I think we should make the rule more conservative and flag only the two methods on the `GlobalScope` and not the  whole `GlobalScope` class per se.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/concurrency/GlobalScopeUsage.kt
```diff
@@ -0,0 +1,54 @@
...
+    companion object {
+        private const val GLOBALSCOPE = "GlobalScope"
+        private const val WANTED_TYPE = "kotlinx.coroutines.$GLOBALSCOPE"
+        private const val MESSAGE =
+            "This use of GlobalScope should be replaced by `CoroutineScope` or `coroutineScope`."
+    }
```
schalkms: style: we always put the companion object to the bottom of the rule class

detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -106,6 +106,11 @@ complexity:
     ignorePrivate: false
     ignoreOverridden: false
 
+concurrency:
```
3flex: I'm wondering if a better name for the ruleset is `coroutines`? It probably depends on whether new rules might be added that relate to other concurrency constructs that aren't coroutines (of course it's hard to say right now whether that might happen in future).

I don't want this to block merging so we could just leave it as is unless anyone else feels they strongly support `coroutines` over `concurrency`.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/concurrency/GlobalScopeUsage.kt
xouabita: Will do 👍

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/concurrency/GlobalScopeUsage.kt
xouabita: Happy to make it more conservative 👍 
The only drawback is that this case would not be covered:

```kt
fun bar(scope: CoroutineScope) = scope.launch { ... }

fun foo() {
    bar(GlobalScope)
}
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/concurrency/GlobalScopeUsage.kt
xouabita: Will do 👍

detekt-cli/src/main/resources/default-detekt-config.yml
xouabita: Happy with both :D

schalkms: I agree with that. `Coroutines` sounds better here.

xouabita: Will go with coroutines then

xouabita:
I have changed the rule to only report `GlobalScope.async` and `GlobalScope.launch`. It also does not need type resolution anymore.
Let me know what do you think 🙂

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/GlobalScopeUsage.kt
```diff
@@ -0,0 +1,53 @@
...
+ * > of GlobalScope is highly discouraged.
+ *
+ * See https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-global-scope/
+ *
+ * <noncompliant>
+ * GlobalScope.launch { delay(1_000L) }
+ * </noncompliant>
+ *
+ * <compliant>
+ * CoroutineScope(Dispatchers.Default).launch { delay(1_000L) }
```
thuytrinh: How can `CoroutineScope(Dispatchers.Default)` mitigate the drawback of `GlobalScope`?

xouabita: Will change this to:

#### Noncompliant Code:

```kotlin
fun foo() {
    GlobalScope.launch { delay(1_000L) }
}
```

#### Compliant Code:

```kotlin
val scope = CoroutineScope(Dispatchers.Default)

fun foo() {
   scope.launch { delay(1_000L) }
}

fun onDestroy() {
    scope.cancel()
}
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/providers/CoroutinesProvider.kt
```diff
@@ -0,0 +1,20 @@
+package io.gitlab.arturbosch.detekt.rules.providers
+
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.RuleSet
+import io.gitlab.arturbosch.detekt.api.RuleSetProvider
+import io.gitlab.arturbosch.detekt.rules.coroutines.GlobalScopeUsage
+
+/** The coroutines rule set analyzes code for potential coroutines problems.
```
arturbosch: Please move the documentation start to the next line.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/providers/CoroutinesProvider.kt
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @active since v1.3.0
```
arturbosch: Please update this to 1.4.0

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/GlobalScopeUsageSpec.kt
```diff
@@ -0,0 +1,69 @@
...
+                import kotlinx.coroutines.CoroutineScope
+                import kotlinx.coroutines.GlobalScope
+
+                fun bar(scope: CoroutineScope) = Unit
+
+                fun foo() {
+                    bar(GlobalScope)
+                }
+            """
+            assertThat(subject.compileAndLint(code)).hasSize(0)
```
arturbosch: Please use `isEmpty()` instead of `hasSize(0)`.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/GlobalScopeUsageSpec.kt
```diff
@@ -0,0 +1,69 @@
...
+                import kotlinx.coroutines.GlobalScope
+
+                fun bar(scope: CoroutineScope) = Unit
+
+                fun foo() {
+                    val scope = GlobalScope
+                    bar(scope)
+                }
+            """
+            assertThat(subject.compileAndLint(code)).hasSize(0)
```
arturbosch: Please use `isEmpty()` instead of `hasSize(0)`.

detekt-gradle-plugin/build.gradle.kts
```diff
@@ -133,8 +133,8 @@ artifacts {
 }
 
 detekt {
+    buildUponDefaultConfig = true
```
xouabita: This is needed because `:detekt-gradle-plugin:detekt` was failing.

```
* What went wrong:
Execution failed for task ':detekt-gradle-plugin:detekt'.
> Invalid detekt configuration file detected.
```

```
Property 'coroutines' is misspelled or does not exist.
io.gitlab.arturbosch.detekt.cli.InvalidConfig: Run failed with 1 invalid config property.
```

schalkms: Blocked by #2217

schalkms: If this issue is resolved, this line should be removed.

3flex: Submitted #2227 before I saw this - I think @xouabita has the correct solution, but we can discuss further in my PR. I don't see any other way to fix this while also enforcing validation of the config file.

xouabita: Awesome!

xouabita:
I have rebased with https://github.com/arturbosch/detekt/pull/2227

schalkms:
Cool thanks! I’m waiting for @3flex final review.

arturbosch:
1. How about `GlobalCoroutineUsage` then?
2. I guess we can limit this for async and launch for now and when users start use this rule they will report if these two fucntions are also discouraged to use :)

schalkms:
> `GlobalCoroutineUsage`

+1

> I guess we can limit this for async and launch for now and when users start use this rule they will report if these two fucntions are also discouraged to use :)

Absolutely! This is a very good start! I’d really like to merge this. Extension can be made in the future if we see a need.

xouabita:
will rename to `GlobalCoroutineUsage` then

Issue #2180, opened 2019-12-15
Gradle cache restoration issue


plastiv:
## Expected Behavior

Running gradle detekt task from gradle plugin on the same kotlin sources should reuse gradle task cache instead of executing task twice. 
## Observed Behavior

Missing html output file prevents cache entity being reused when running detekt task with gradle plugin. 

I believe this is due to missing ` @get:PathSensitive(PathSensitivity.RELATIVE)` annotation on report output files: https://github.com/arturbosch/detekt/blob/5f41ea56e1915b86271bfefdcc7b68e7650e6cab/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt#L145-L158

## Steps to Reproduce

* have a project with detekt plugin integrated, like `my-project1`
* make sure for a project local cache is enabled, remote cache is disabled
* copy project into different folder on disk, like `my-project2` to trigger different PATH issue
* remove local cache to avoid accidental artifact reuse, like `rm -rf ~/.gradle/caches`
* run `./gradlew clean detekt` at `my-project1` to seed a cache entry
* run `./gradlew clean detekt` at `my-project2`. 

Expected: detekt task `FROM-CACHE`. Actual: detekt task is executed due to `Missing html output file` property.

## Context

Detekt tasks are marked as cooperative and are participating in cache lifecycle by uploading and downloading task artifacts. yet, their cache can't be reused when running on CI with ephemeral nodes due to jobs running project at different folder paths. 

## Your Environment

* Version of detekt used: 1.2.2
* Version of Gradle used (if applicable): 6.0.1
* Operating System and version: mac and linux
* Link to your project (if it's a public repository): if you are not sure how to reproduce the issue, let me know I will try to find some time to work on a sample

plastiv:
I have managed to create minimal reproducible example. Run `./test.sh` to trigger steps-to-reproduce from above. 

* https://github.com/plastiv/detekt-cache-issue
* https://travis-ci.org/plastiv/detekt-cache-issue/builds/625774155

I have spent more time investigating the issue and currently believe that it isn't the html report output file is a culprit here. 

[...]

PR #2182, opened 2019-12-16
Fix Detekt gradle task cache restoration issue (#2180)

plastiv:
`Detekt` gradle task is marked as `@CacheableTask` but it's input source files are
marked with `PathSensitivity.ABSOLUTE` annotation from `source` property defined in
parent `SourceTask`.

Having task inputs with absolute path sensitivity effectively disables cache reusability:
https://guides.gradle.org/using-build-cache/#relocatability


3flex:
Thanks for picking that up, plus the fix!

plastiv:
@3flex is there a way to get a snapshot for gradle-plugin? I don't see it being published here https://oss.jfrog.org/artifactory/webapp/#/artifacts/browse/tree/General/oss-snapshot-local/io/gitlab/arturbosch/detekt

3flex:
Not sure, I don't have access to publish artifacts. @arturbosch do you know if anything can be done here?

@plastiv consider opening a new issue to track this, thanks!

PR #2188, opened 2019-12-20
Add printers to the command line runner

arturdryomov:
A bit of context to understand the motivation behind the change.

Me and @artem-zinnatullin are working on [the Bazel rule for Detekt](https://github.com/buildfoundation/bazel_rules_detekt). One of the features of the rule is the [persistent workers](https://blog.bazel.build/2015/12/10/java-workers.html) support.

* The thing with workers is that standard inputs and outputs are reserved for the Bazel / worker communication. Since there is no API to limit Detekt standard output usage we were forced to create [a workaround](https://github.com/buildfoundation/bazel_rules_detekt/blob/d5ceef75df618fdd22a0e4a4a87cba381daa29f6/detekt/wrapper/src/main/kotlin/io/buildfoundation/bazel/rulesdetekt/wrapper/SandboxExecutor.kt) which replaces system standard output streams with file-based ones.
* This worked fine (though a bit hacky) until we thought about [multiplexing](https://docs.bazel.build/versions/master/multiplex-worker.html). The multiplexing will change the flow, essentially we’ll be calling multiple Detekt runners in parallel. Unfortunately, the replace-system-streams workaround will not work in a multi-thread environment since it will be almost impossible to understand which runner printed which output to a singleton system stream.

If this change is accepted (and eventually published) we’ll be able to make Detekt invocations more or less hermetic. At the same time, the API change is backwards-compatible (arguments fall back to default values) and kinda makes sense abstracting from our personal needs (there is no way to change printers since `ProcessingSettings` in `Runner` are not accessible).

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Main.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Main.kt
@@ -28,7 +28,11 @@ fun main(args: Array<String>) {
     exitProcess(ExitCode.NORMAL_RUN.number)
 }
 
-fun buildRunner(args: Array<String>, outputPrinter: PrintStream = System.out, errorPrinter: PrintStream = System.err): Executable {
+fun buildRunner(
+    args: Array<String>,
+    outputPrinter: PrintStream = System.out,
+    errorPrinter: PrintStream = System.err
+): Executable {
     val arguments = parseArguments(args)
     return when {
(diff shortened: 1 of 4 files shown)


schalkms:
@arturdryomov the bazel detekt plugin is cool. Thanks for doing this! +1 star

schalkms:
@arturdryomov could you please add a unit test that tests the newly added parameters?

arturdryomov:
@schalkms, sure — expanded the tests suite a bit.

PR #2192, opened 2019-12-26
Add tests showing how to exclude custom config properties in plugins

### LINKED ISSUE #2169: Custom rules are not applied on Android
I added detekt to both my work and hobby project, both multi module Android projects. It works perfectly.
However, if I try to add custom rules, I can't get them to work. I tried to add rules that would definitely fail to no avail.
Sometimes the detekt task just quietly succeeds without any warning, sometimes it gives the following warnings for each module (The custom ruleset is called "babe"):

Property 'babe' is misspelled or does not exist.
io.gitlab.arturbosch.detekt.cli.InvalidConfig: Run failed with 1 invalid config property.

This probably means that the detekt config file is OK, but I got something wrong with the Gradle config.

My hobby project is open source, please have a look at here in the detekt branch:
https://bitbucket.org/herrbert74/companies-house

The custom rules are in the "core-detekt" module. All configs for Gradle are in the main build.gradle file under the allprojects and subprojects configurations. Some files are under the team-props folder in the project directory.

All in all I spent two days trying everything, without any luck. Any help would be greatly appreciated.

arturbosch:
Closes #2169


detekt-sample-extensions/src/test/resources/not-excluded-config.yml
```diff
@@ -0,0 +1,3 @@
+sample:
+  TooManyFunctions:
```
schalkms: I would rename this to `included-config.yml`.

detekt-sample-extensions/src/test/kotlin/io/gitlab/arturbosch/detekt/sample/extensions/SupportConfigValidationSpec.kt
```diff
@@ -0,0 +1,38 @@
...
+class SupportConfigValidationSpec : Spek({
+
+    describe("support config validation") {
+
+        val testDir = Files.createTempDirectory("detekt-sample")
+
+        it("fails when new rule set is not excluded") {
+            val args = CliArgs {
+                input = testDir.toString()
+                configResource = "not-excluded-config.yml"
```
schalkms:
```suggestion
                configResource = "included-config.yml"
```

Issue #2168, opened 2019-12-06
exit with error if unsupported property is detected


robstoll:
## Expected Behavior
If detekt detects an unsupported property, then it should:
1. report it
2. stop the analysis
3. fail the detekt task

## Observed Behavior
detekt does the first two points but not the third, leaving a CI-build green without actually analysing anythign.

## Steps to Reproduce
1. git clone https://github.com/robstoll/niok
2. checkout 1664347e738eec37210e8d4b00f58e0bee16d366
3. ./gr build detekt

fails due to 
https://github.com/robstoll/niok/blob/1664347e738eec37210e8d4b00f58e0bee16d366/gradle/scripts/detekt.yml#L1

but does not end with a failure

## Context
https://travis-ci.org/robstoll/kbox/jobs/616582947#L580

## Your Environment
* Version of detekt used: 1.2.2
* Link to your project (if it's a public repository):
https://github.com/robstoll/niok

PR #2201, opened 2019-12-27
Exclude yet unknown new build property

arturbosch:
detekt analysis on the gradle plugin is not run due to "invalid config".
This is exactly what happened to our users in #2168 and #2169 and will be fixed in 1.3.1 :).


arturbosch:
I will just open a PR removing this change asap and leave it open :)

Issue #2134, opened 2019-11-27
StringIndexOutOfBoundsException: String index out of range: 8 when generating HTML report after update to 1.2.0


afaucogney:
The bug you're experiencing might have already be reported!
Please search in the [issues](https://github.com/arturbosch/detekt/issues) before creating one.

## Expected Behavior
Trying to upgrade to Detekt 1.20. Was in 1.0.0
HTML Report generation failed 

## Observed Behavior
Works with 1.10 and 1.1.1
Work if only XML and TXT generated

## Steps to Reproduce
Use my config I cannot give public

## Context
Use detekt-formatting

## Your Environment

* Version of detekt used: 1.2.0
* Version of Gradle used (if applicable): 5.4.1
* Operating System and version: Mac
* Link to your project (if it's a public repository):

```
Overall debt: 3d 11h 15min

Complexity Report:
        - 64695 lines of code (loc)
        - 47134 source lines of code (sloc)
        - 31244 logical lines of code (lloc)
        - 7087 comment lines of code (cloc)
        - 7202 McCabe complexity (mcc)
        - 566 number of total code smells
        - 15 % comment source ratio
        - 230 mcc per 1000 lloc
        - 18 code smells per 1000 lloc

Project Statistics:
        - number of properties: 4525
        - number of functions: 3968
        - number of classes: 2019
        - number of packages: 259
        - number of kt files: 1065

java.lang.StringIndexOutOfBoundsException: String index out of range: 8
        at java.lang.String.substring(String.java:1963)
        at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt.writeErrorLine(HtmlUtils.kt:38)
        at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt.access$writeErrorLine(HtmlUtils.kt:1)
        at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt$snippetCode$1$1.invoke(HtmlUtils.kt:25)
        at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt$snippetCode$1$1.invoke(HtmlUtils.kt)
        at kotlinx.html.ApiKt.visit(api.kt:80)
        at kotlinx.html.Gen_tag_unionsKt.code(gen-tag-unions.kt:161)
        at kotlinx.html.Gen_tag_unionsKt.code$default(gen-tag-unions.kt:161)
[...]

PR #2209, opened 2019-12-28
Show code snippet erros in html reports

BraisGabin:
This PR is realted with #2134

If an exception is found in the code snippet rendering instead to be too smart, or crash, or hide the error this code shows the error to the user like this:
Captura de pantalla 2019-12-28 a las 14 45 04

The link opens the github issue form with some pre-filled information:
Captura de pantalla 2019-12-28 a las 15 08 42

And github render that information like this:
Captura de pantalla 2019-12-28 a las 15 08 48

----

Problems in this PR:
- [x] I don't know how to get the detekt version to autofill it. Any idea?
- [x] What if the issue is related with a plugin and not with detekt itself? We are saying to the users to report it here. Should we care about this? If we see a spam of issues related with plugins we can remove the link. What do you think?


BraisGabin:
`whichDetekt()` was exactly what I needed. Done, ready to review :).

PR #2233, opened 2020-01-09
Fix #2230 equals() func detection

### LINKED ISSUE #2230: WrongEqualsTypeParameter does not ignore multi-parameter equals methods
WrongEqualsTypeParameter

As stated with the documentation provided, WrongEqualsTypeParameter should only check for equals methods with one parameter. 

If function is not top level and name is equals and has one parameter that is not Any? or Kotlin.Any? rule should fail.

## Context

For mocking purposes I delegate some of standard library methods to helper classes. such as 

```Kotlin
interface StringHelper {
    @Contract("null -> true")
    fun isEmpty(input: String?): Boolean

    @JvmDefault
    @Contract("null, null -> true")
    fun equals(left: String?, right: String?, ignoreCase: Boolean = false): Boolean
}

class StringHelperImpl : StringHelper {

    override fun isEmpty(input: String?) = input.isNullOrEmpty()

    override fun equals(left: String?, right: String?, ignoreCase: Boolean): Boolean = left.equals(right, ignoreCase)

}
```

In this particular case rule fails. I believe rule should ignore equals methods with more than one parameter.

schalkms:
The equals() func check should assert that there is exact 1 parameter.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignature.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignature.kt
@@ -5,7 +5,7 @@ import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtObjectDeclaration
 
 fun KtFunction.isEqualsFunction() =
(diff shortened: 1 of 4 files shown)


schalkms:
This is good to go!

PR #2235, opened 2020-01-10
Don't resolve dependencies during project configuration

3flex:
See https://github.com/arturbosch/detekt/pull/2200#issuecomment-572827631

**Old build scan:**
image

**New build scan:**
image

Config derived from https://docs.gradle.org/current/userguide/working_with_files.html#sec:creating_uber_jar_example


3flex:
This didn't work as I thought it would, still get warning on build scan just now on a different configuration:
image

PR #2240, opened 2020-01-10
ThrowingExceptionInMain: fix #2248 and add tests

### LINKED ISSUE #2248: ThrowingExceptionInMain rule should consider main() function with no parameters
As of Kotlin 1.3, the explicit Array<String> can be omitted but will still be available in the byte code.
[Reference](https://stackoverflow.com/a/54325625)

## Expected Behavior
The rule should also flag parameterless `main()` functions that throw an exception.

```kotlin
fun main() { throw IllegalArgumentException() }
```

## Observed Behavior
The rule ignores parameterless `main()` functions.

## Your Environment

* Version of detekt used: 1.3.1
* Operating System and version: macOS 10.14.6

schalkms:
Fixes #2248

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ThrowingExceptionInMainSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ThrowingExceptionInMainSpec.kt
@@ -32,8 +32,13 @@ class ThrowingExceptionInMainSpec : Spek({
                         fun main(args: Array<String>) { throw IllegalArgumentException() }
                     }
                 }
+                
+                object O {
+                    @JvmStatic
+                    fun main(args: Array<String>) { throw IllegalArgumentException() }
+                }
             """
-            assertThat(subject.compileAndLint(code)).hasSize(2)
+            assertThat(subject.compileAndLint(code)).hasSize(3)
         }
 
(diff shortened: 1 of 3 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ThrowingExceptionInMainSpec.kt
```diff
@@ -27,5 +51,18 @@ class ThrowingExceptionInMainSpec : Spek({
...
             assertThat(subject.lint(file)).isEmpty()
         }
+
+        it("does not report a mains function with no @JvmStatic annotation inside a class") {
+            val code = """
+            class A {
+                fun main(args: Array<String>) { }
+                
+                companion object {
+                    fun main(args: Array<String>) { }
```
BraisGabin: shouldn't these functions have a `throw Exception()` inside it's body? Otherwise we don't know if they are ignored because they don't have the annotation or because they don't throw an exception.

If you can, add the `throw Exception()` to the previous test too.

schalkms: Thanks! Good catch!

BraisGabin:
The test `does not report top level main functions with a wrong signature` has the same problem. I know that you didn't change it, but it would be good to fix it too,

PR #2243, opened 2020-01-11
New ktlint rules

3flex:
* EnumEntryNameCase
* NoEmptyFirstLineInMethodBlock

Both are from the ktlint experimental rule set so not enabled by default.


3flex:
It's a little strange, `verifyGeneratorOutput` is failing but doesn't report what the errors actually are, so not sure how I can resolve it. Runs clean on my machine.

PR #2255, opened 2020-01-13
Add a new line at the end of the txt report

BraisGabin:
This PR adds a `\n` at the end of the txt report. This way you can concat all the reports easily. Or you can do something like `wc -l report.txt` to count the number of errors that you and not be off by one

## Context
This is a real minor thing but I lose 10 minutes today because of this. I was counting the errors of different branches using `wc -l report.txt`. And I was getting always one error less than expected. This PR will avoid that problem.

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/TxtOutputReport.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/TxtOutputReport.kt
@@ -10,7 +10,10 @@ class TxtOutputReport : OutputReport() {
     override val name = "plain text report"
 
     override fun render(detektion: Detektion): String {
-        val smells = detektion.findings.flatMap { it.value }
-        return smells.joinToString("\n") { it.compactWithSignature() }
+        val builder = StringBuilder()
+        detektion.findings
+            .flatMap { it.value }
+            .forEach { builder.append(it.compactWithSignature()).append("\n") }
+        return builder.toString()
     }
 }
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/TxtOutputReportTest.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/TxtOutputReportTest.kt
@@ -10,10 +10,32 @@ class TxtOutputReportTest : Spek({
 
     describe("TXT output report") {
 
-        it("render") {
+        it("render none") {
+            val report = TxtOutputReport()
+            val detektion = TestDetektion()
+            val renderedText = ""
+            assertThat(report.render(detektion)).isEqualTo(renderedText)
+        }
+
+        it("render one") {
             val report = TxtOutputReport()
             val detektion = TestDetektion(createFinding())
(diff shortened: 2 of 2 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/TxtOutputReport.kt
```diff
@@ -11,6 +11,8 @@ class TxtOutputReport : OutputReport() {
 
     override fun render(detektion: Detektion): String {
         val smells = detektion.findings.flatMap { it.value }
-        return smells.joinToString("\n") { it.compactWithSignature() }
+        return smells.joinToString("\n") { it.compactWithSignature() }.let {
```
schalkms: The `if else` here is redundant.
You can add a new line after every smell like this.

```suggestion
        return smells.joinToString("\n", postfix = "\n") { it.compactWithSignature() }
```

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/TxtOutputReportTest.kt
```diff
@@ -10,10 +10,32 @@ class TxtOutputReportTest : Spek({
...
+            val detektion = TestDetektion(
+                createFinding(ruleName = "TestSmellA"),
+                createFinding(ruleName = "TestSmellB"),
+                createFinding(ruleName = "TestSmellC"))
+            val renderedText = """
+                TestSmellA - [TestEntity] at TestFile.kt:1:1 - Signature=S1
+                TestSmellB - [TestEntity] at TestFile.kt:1:1 - Signature=S1
+                TestSmellC - [TestEntity] at TestFile.kt:1:1 - Signature=S1
+                
+            """.trimIndent()
```
schalkms: Here we have **two** new lines at the end of the file.

BraisGabin: `trimIndent` removed the second end of line. I can use `trimMargin` to make it easier to understand.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/TxtOutputReport.kt
BraisGabin: Yes, and this is what this code does... If you mean this: `smells.joinToString { it.compactWithSignature() + "\n" }` it would work too(out was my first impregnation). But we are losing the performance of string builder. That would create 2k useless strings.

schalkms: Uh sorry. By looking at the test code once more, it becomes clear what this code is intended to do.
But something like this should work without introducing a new branch.

```suggestion
val builder = StringBuilder();
smells.forEach { builder.append(it.compactWithSignature()).append('\n') }
return builder.toString()
```

BraisGabin: Agree! This is easier.

PR #2258, opened 2020-01-15
Modify default behavior to not output unless errors are found.  Adding a verbose flag which will have legacy behavior

zsmoore:
Adding clli argument and gradle plugin functionality to enable verbose logging.

On our project we only want to log to console if an error occurs for cleaner logging,

Swapped default logic to not output unless errors present or verbose flag enabled


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/OutputFacade.kt
```diff
@@ -53,6 +53,9 @@ class OutputFacade(
     }
 
     private fun handleConsoleReport(report: ConsoleReport, result: Detektion) {
-        report.print(settings.outPrinter, result)
+        if (!settings.suppressConsoleLogs ||
+            config.maxIssues().isValidAndSmallerOrEqual(result.getOrComputeWeightedAmountOfIssues(config))) {
```
schalkms: Shouldn't this be negated `!` ? You only want to print something if the max issue count in the config is exceeded when `--suppress-console-logs` is set, right?

arturbosch:
Cool, I had something similiar in mind, calling it `--silent`.
I will take another look at this PR later.
With the current version you can try https://arturbosch.github.io/detekt/howto-silent-reports.html for now.

BraisGabin:
I love this PR! I think that no output should be the default. And add `-v` if you want the verbose mode.

As a bonus we can have the quiet mode `-q`. It should print nothing at all. It just create the reports.

schalkms:
> I think that no output should be the default. And add -v if you want the verbose mode.

That's a good idea! I agree with that. Then we should call the flag introduced with this PR --verbose  = false
I hope that also fits your needs @zsmoore

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/OutputFacade.kt
zsmoore: resolving since behavior has changed

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/OutputFacade.kt
```diff
@@ -53,6 +53,9 @@ class OutputFacade(
     }
 
     private fun handleConsoleReport(report: ConsoleReport, result: Detektion) {
-        report.print(settings.outPrinter, result)
+        if (!settings.suppressConsoleLogs ||
```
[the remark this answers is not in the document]
zsmoore: with behavior change we should have all tests covered here

zsmoore:
Updated PR to default to no output unless error detekt should fail build or new flag `verbose` is enabled

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/OutputFacade.kt
```diff
@@ -53,6 +53,9 @@ class OutputFacade(
     }
 
     private fun handleConsoleReport(report: ConsoleReport, result: Detektion) {
-        report.print(settings.outPrinter, result)
+        if (settings.verbose ||
+            config.maxIssues().isValidAndSmallerOrEqual(result.getOrComputeWeightedAmountOfIssues(config))) {
```
schalkms: We only want to print something if the max issue count is exceeded, but here we just print something if the actual issue count **is below** the max, right?

schalkms:
Cool! Thanks for making these changes and bringing real value to our users.

BraisGabin:
I would add `-v` as an alias of `--verbose` (it's kind of a standard).

If I don't use the `--verbose` flag in a project without issues I get no output (expected behaviour 👍). But, if I do the same with a project with issues I get the error report (expected behavious 👍) but I'm getting this too:

> detekt finished in 17751 ms.
> Successfully generated HTML report at report.html
> Build failed with 3178 weighted issues (threshold defined was 10).
> io.gitlab.arturbosch.detekt.cli.BuildFailure: Build failed with 3178 weighted issues (threshold defined was 10).

Probably we should show the summary "Build failed with 3178 weighted issues (threshold defined was 10)." But the time and the generated reports should not be there. What do you think?

schalkms:
@BraisGabin I think we should keep it as simple as possible now.

BraisGabin:
At least the -v should be added, I think. The other points can be fixed later.

schalkms:
> At least the -v should be added

Agreed.

arturbosch:
I'm not really sure we need a `--verbose` flag.
We could achieve the same silent output when changing the `Extension`'s defaults.
I will take a further look at our extension mechanism. As we are now changing the `maxIssues` default, lets change the printing output too.

arturbosch:
@zsmoore thanks for bringing up this suggestions. We have decided to merge in the simpler solutions and opened #2267 for a major update. A more silent detekt with fail fast default will be released in 1.5.0 :).

PR #2261, opened 2020-01-16
Don't flag inherited visibility in NestedClassesVisibility

### LINKED ISSUE #1930: NestedClassesVisibility(False negative): Nested class doesn't inherit visibility from parent
From the comment,

**NonCompliant**
```kt
internal class NestedClassesVisibility {
   public class NestedPublicClass // should not be public
}
```
**Compliant**
```kt
 internal class NestedClassesVisibility {
     internal class NestedPublicClass
 }
```

In actuality, _both_ should be Non Compliant.

This is because both are equivalent as the nested class inherits its visibility from its parent class.

So effectively, the `public` class in the initial example is _also_ `internal`.

The way it's written, just marking an already "internal" class "internal" makes the violation go away.

schalkms:
The rule had an issue that it forced the use of an additional
unnecessary modifier. However, the visibility of nested classes is
inherited from the parent class.
These nested classes can't have a higher visibility than their parent.
The following example outlines the issue.

internal class A1 { // non-compliant
  class B
}
internal class A2 { // compliant
  internal class B
}

Closes #1930

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/NestedClassesVisibility.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/NestedClassesVisibility.kt
@@ -14,29 +14,31 @@ import org.jetbrains.kotlin.psi.KtClassOrObject
 import org.jetbrains.kotlin.psi.KtEnumEntry
 
 /**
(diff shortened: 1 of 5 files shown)


BraisGabin:
If we want that this rule detects the misleading modifiers we should check the private classes with public or internal classes too.

schalkms:
@BraisGabin Agreed. Please open a new issue for that.

PR #2262, opened 2020-01-17
Add the ability to override max Issues from cli and gradle plugin

zsmoore:
Motivation for this one.

If you want to use default config or build upon it but want to not allow any errors in code base, set via cli or command line.  For users who want quick setup without adding any config files.

Looks like currently if you supply no config your default threshold is 10, (not sure if this is a bug?) we want to test out using the default config but also don't want to allow any issues into our code base.


zsmoore:
No problem at all, ill close this once the other goes in.

I prefer setting the default to allow 0 anyway :D just wasn't sure if it would be accepted vs allowing the config.

arturbosch:
I will close this and open a new proposal for a 2.0 release.

PR #2263, opened 2020-01-17
Update config:maxIssues value to 0

schalkms:
A sensible default value is to have zero instead of 10 reported detekt issues. 
Continuing the discussion in #2262

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -1,5 +1,5 @@
 build:
-  maxIssues: 10
+  maxIssues: 0
   excludeCorrectable: false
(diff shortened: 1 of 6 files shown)


zsmoore:
This will have to be 1 instead of 0. I do however believe the wording is a bit weird.  If I set maxIssues to 10 i would expect allowing 10 issues and disallowing 11.  Similarly i would expect to have to set maxIssues to 0, rather than 1 in order to allow no issues.

I'm still fine pushing this through once it is updated to 1 but just a call out.

arturbosch:
@zsmoore agree `maxIssues` is misleading. I would also expect with `maxIssues=2` to allow two issues and not to fail. I will open a PR for this.

Issue #2264, opened 2020-01-17
Build failed with... is printed twice in the cli


BraisGabin:
## Expected Behavior
If I run the cli and I have issues at the end of the output I get the numer of issues onces

## Observed Behavior
I'm getting this output twice:

> Build failed with 3178 weighted issues (threshold defined was 10).
> io.gitlab.arturbosch.detekt.cli.BuildFailure: Build failed with 3178 weighted issues (threshold defined was 10).

## Steps to Reproduce
Run the cli on a project that have issues

## Context
I found this issue testing #2258 and then I saw that it was in master too.

## Your Environment
* Version of detekt used: master (1.4.0)

BraisGabin:
I checked the code and this is because the `BuildFailureReport` show this message once and then the `Runner` throws an exception with exactly the same message.

BraisGabin:
We can close this now, right?

PR #2266, opened 2020-01-19
Remove redundant build failure message - #2264

arturbosch:
Removes redundant `BuildFailureReport`


arturbosch:
Needs rebase on top of #2269

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/BuildFailureReport.kt
```diff
@@ -1,34 +0,0 @@
...
-
-    override fun render(detektion: Detektion): String? {
-        val amount = detektion.getOrComputeWeightedAmountOfIssues(config)
-        val maxIssues = config.maxIssues()
-        return when {
-            maxIssues.isValidAndSmallerOrEqual(amount) -> {
-                "Build failed with $amount weighted issues (threshold defined was $maxIssues).".red()
-            }
-            amount > 0 && maxIssues != -1 -> {
-                "Build succeeded with $amount weighted issues (threshold defined was $maxIssues).".yellow()
```
schalkms: Don't we loose the functionality provided by this LOC?

arturbosch: Yes, however I consider it redundant as the "yellow" finding messages provide the same information.

We can consider this change as testing if our users actually use and need the "issue weighting" feature. We could deprecate and remove this extra complexity later?

schalkms: Okay

PR #2269, opened 2020-01-19
Fix regression not considering baseline file when calculating build failure threshold

arturbosch:
@schalkms found after setting `maxIssues=0` and #2266 :+1:


schalkms:
👍

PR #2270, opened 2020-01-19
Update website ruby dependencies fixing potential security vulnerability

arturbosch:


arturbosch:
@schalkms good idea, unfortunately dependabot does not support `build.kts` files but I configured it to update the website Gemfile.

PR #2271, opened 2020-01-19
Sanitize gradle build scripts and suppress unstable api usages

arturbosch:


arturbosch:
> I’m not sure, whether we should suppress these issues.

Yeah maybe not, this way we will see when the api gets stable. I revert them.

> Why did you remove the **BuildFailure** exception?

It was only used in the `ProcessExecutor` which is again only used by the Idea wrapper tasks. Gradle will treat it as a build failure for us. I think it is legacy code from before the gradle rewrite to dynamic resolution.

PR #2279, opened 2020-01-19
Introduce DefaultRuleSetProvider interface marking detekt-rules providers as default

arturbosch:
This way we won't forget to add new Provider's like we did with `CoroutinesProvider`.


arturbosch:
Looks good, I had to add the new interface to the generator. No `"RuleSetProvider"` in the codebase.

PR #2282, opened 2020-01-19
Call detekt reflectively

arturbosch:
Continuation of https://github.com/arturbosch/detekt/pull/2200 with hopefully most of the classloader leaks fixed ...
Metaspace OOM issues may still occur as every detekt run loads ~50mb of classes when using `--parallel`.

Following screenshot shows that the gradle daemon can unload detekt classes though:
2020-01-19-222256_704x365_scrot

Insights from previous PR:
- hard coding the detekt version is 3x faster than our current Gradle Plugin
  - downside is that we can't dynamically set the detekt classpath (changing versions etc)
  - we can't use our plugin mechanism!
    - we may convert the extra plugin classpath to `--plugins jars...` parameters however when trying this I got unexpected crashes ...
- using the `project.ant` method like the gradle builtin quality tools fork a jvm
  - an anttask can be run without forking the JVM, but again Metaspace
  - performance was nearly the same as now with worse logging `anttask: detekt output` on every line
- Calling detekt reflectively revealed some classloader memory leaks
  - still not sure about the performance though the Metaspace issues are still present
  - I hope that with this way we can save ~200ms JVM starting overhead?
- Did not try out the worker api though reading it's docu does not show how a javaexec could be speed up
  - To my knownledge the worker api brings performance benefits for easier tasks than starting a new process?

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
@@ -1,0 +1,34 @@
+package io.gitlab.arturbosch.detekt.internal
+
+import org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport
+import org.gradle.api.file.FileCollection
+import java.net.URLClassLoader
+
+object ClassLoaderCache {
+
+    private var cached: URLClassLoader? = null
+    private var lastSeenClasspath: FileCollection? = null
+
+    fun getOrCreate(classpath: FileCollection): URLClassLoader {
+        val lastClasspath = lastSeenClasspath
+        if (lastClasspath == null) {
+            cache(classpath)
+        } else {
+            synchronized(ClassLoaderCache) {
+                if (!lastClasspath.minus(classpath).isEmpty) {
+                    DefaultGroovyMethodsSupport.closeQuietly(cached)
+                    cache(classpath)
+                }
+            }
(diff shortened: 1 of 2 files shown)


arturbosch:
I've finally managed to find a way to call detekt reflectively without an OOM error -> caching the class loader.
We safe the startup jvm process cost ~200ms plus loading all ~40mb classes :).
Need to further profile the performance.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
```diff
@@ -0,0 +1,34 @@
...
+        return cached ?: throw IllegalStateException("Detekt classloader expected.")
+    }
+
+    private fun cache(classpath: FileCollection) {
+        lastSeenClasspath = classpath
+        cached = URLClassLoader(
+            classpath.map { it.toURI().toURL() }.toTypedArray(),
+            null /* isolate detekt environment */
+        )
+    }
```
BraisGabin: To avoid race conditions we should save the `FileCollection` and `URLClassLoader` in a `Pair`. This way we can be sure that we don't mess up returning an `URLClassLoader` that doesn't mach with the `FileCollection`

arturbosch: Awesome suggestion! Ty.

arturbosch:
Links to #2035

PR #2288, opened 2020-01-24
Show the finding count in the html report

### LINKED ISSUE #2286: Show issue count in the html report
## Expected Behavior
When I open the html report I saw how many issues do I have in the project.

## Current Behavior
These information is not shown

## Context
It would be great to show the total and then for each section show the count too.

(I'll try to create a PR asap)

BraisGabin:
This PR adds the finding count like this:

Fixes #2286


detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputFormatTest.kt
```diff
@@ -124,21 +131,47 @@ private fun createTestDetektionWithMultipleSmells(): Detektion {
...
+    ),
+    ktElement = ktElement
+)
+
+private fun issue(id: String) = Issue(
+    id = id,
+    severity = Severity.CodeSmell,
+    description = "Description $id",
+    debt = Debt.TWENTY_MINS
+)
```
schalkms: For better readability I'd rename this to `createIssue` and so on.
Moving this into a separate file would also be a good idea in order to not clutter this test.

PR #2289, opened 2020-01-24
Number format in some report

BraisGabin:
Add thousand separators. Remove useless `- ` and remove space between value and %

This is before:
Captura de pantalla 2020-01-24 a las 22 59 46
and this is after this PR:
Captura de pantalla 2020-01-24 a las 23 04 14

This changes are made for the console output too. If we don't want them I can refactorice them

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/ComplexityReportGenerator.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/ComplexityReportGenerator.kt
@@ -24,7 +24,7 @@ class ComplexityReportGenerator(private val complexityMetric: ComplexityMetric) 
             "%,d comment lines of code (cloc)".format(Locale.US, complexityMetric.cloc),
             "%,d McCabe complexity (mcc)".format(Locale.US, complexityMetric.mcc),
             "%,d number of total code smells".format(Locale.US, numberOfSmells),
-            "%,d %% comment source ratio".format(Locale.US, commentSourceRatio),
+            "%,d%% comment source ratio".format(Locale.US, commentSourceRatio),
             "%,d mcc per 1,000 lloc".format(Locale.US, mccPerThousandLines),
             "%,d code smells per 1,000 lloc".format(Locale.US, smellPerThousandLines)
         )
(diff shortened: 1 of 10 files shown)


detekt-cli/src/test/resources/complexity-report-generate.txt
```diff
@@ -0,0 +1,9 @@
+1,000 lines of code (loc)
```
schalkms: Why is there a need for a separate resource file?
`complexity-report.txt` has the same content.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/ComplexityReportGenerator.kt
```diff
@@ -14,21 +15,19 @@ class ComplexityReportGenerator(private val complexityMetric: ComplexityMetric)
             ComplexityReportGenerator(ComplexityMetric(detektion))
     }
 
-    fun generate(): String? {
+    fun generate(): List<String>? {
         if (cannotGenerate()) return null
```
schalkms: An empty list makes more sense here.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/out/HtmlOutputReport.kt
```diff
@@ -117,13 +118,7 @@ class HtmlOutputReport : OutputReport() {
...
 
     private fun getComplexityMetrics(detektion: Detektion): List<String> {
-        val complexityReport = ComplexityReportGenerator.create(detektion).generate()
-        return if (complexityReport.isNullOrBlank()) {
-            emptyList()
-        } else {
-            val complexities = complexityReport.split("\n")
-            return complexities.subList(1, complexities.size - 1)
-        }
+        return ComplexityReportGenerator.create(detektion).generate() ?: emptyList()
```
schalkms: If `generate()` returns an empty list you won't need this conditional statement anymore.
Maybe this method can then be inlined as well.

detekt-cli/src/test/resources/complexity-report-generate.txt
BraisGabin: They are similar, no the same. One has a header and dashes and the other don't. But, probably, I shoukd inline this in the test code. Ill do it.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/ComplexityReportGenerator.kt
BraisGabin: I'm not sure about this. The null says: there's not report. An empty list sais: there is a report and it's empty. In the html report both are the same but in the console report don't. Because we need to add the header. We could check if the list is empty but I think that it's not that clean.

PR #2290, opened 2020-01-24
Sort html report

### LINKED ISSUE #2287: Keep the order of the issues in the html report
## Expected Behavior
If I run detekt, fix an issue and run it again I expect that the findind sections would keep the order.

## Current Behavior
Now if I fix an error this section can go down in the report

## Context
I was fixing some issues, I re run detekt and I thought that I finished fixing that kind of issues. But I hadn't, that section just moved down in the report.

(I'll try to create a PR asap)

BraisGabin:
Fixes #2287


BraisGabin:
The first commit is exactly the same. Because I needed that refactor in the tests too for this. I will rebase this or #2288 as soon as the other is merged. We can block this one if you want.

schalkms:
Rebasing is the best option, I think.

BraisGabin:
Rebased. We can unblock it.

PR #2298, opened 2020-01-28
Add coupled class rule

edsoncunha:
Adds a new rule to check whether a given class is coupled to a high number of other classes.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CoupledClass.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CoupledClass.kt
@@ -1,0 +1,83 @@
+package io.gitlab.arturbosch.detekt.rules.complexity
+
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Metric
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.ThresholdRule
+import io.gitlab.arturbosch.detekt.api.ThresholdedCodeSmell
+import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtConstructor
+import org.jetbrains.kotlin.psi.KtUserType
+
+/**
+ * Complex classes which reference too many other classes indicate that this class is handling too
+ * many things at once. Classes should follow the single-responsibility principle to also encourage implementations
+ * of this interface to not handle too many things at once.
+ *
+ * Tightly coupled classes should be split into smaller interfaces which have a clear responsibility and are easier
+ * to understand and implement.
+ *
+ * @configuration threshold - the amount of class references to trigger the rule (default: `10`)
+ */
+class CoupledClass(
+    config: Config = Config.empty,
+    threshold: Int = DEFAULT_MAX_REFERENCED_CLASSES
+) : ThresholdRule(config, threshold) {
+
+    override val issue = Issue(
+        javaClass.simpleName, Severity.Maintainability,
+        "A class is coupled to too many other classes. " +
+                "Coupled classes violate the Single Responsibility Principle. " +
+                "A class should have one responsibility. " +
+                "Split up large classes into smaller ones that are easier to understand.",
+        Debt.TWENTY_MINS
+    )
+
+    override fun visitClass(klass: KtClass) {
+        if (!klass.isData() && !klass.isEnum()) {
+            val referencedClasses = mutableListOf<String?>()
+
+            referencedClasses.addAll(fetchReferencedClasses(klass.primaryConstructor))
+            referencedClasses.addAll(fetchReferencedClasses(klass.secondaryConstructors))
+
+            val totalUniqueReferencedClasses = referencedClasses.toSet().size
+
+            if (totalUniqueReferencedClasses >= threshold) {
+                report(
+                    ThresholdedCodeSmell(
+                        issue,
+                        Entity.from(klass),
+                        Metric("SIZE: ", totalUniqueReferencedClasses, threshold),
+                        "The class ${klass.name} is coupled to too many other classes. " +
+                                "Consider splitting it up to the maximum authorized of $threshold or less."
+                    )
+                )
+            }
+        }
+
+        super.visitClass(klass)
+    }
+
+    private fun fetchReferencedClasses(secondaryConstructors: List<KtConstructor<*>>): List<String?> {
+        return secondaryConstructors.flatMap { fetchReferencedClasses(it) }
+    }
+
+    private fun fetchReferencedClasses(constructor: KtConstructor<*>?): List<String?> {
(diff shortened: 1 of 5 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CoupledClass.kt
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Complex classes which reference too many other classes indicate that this class is handling too
+ * many things at once. Classes should follow the single-responsibility principle to also encourage implementations
+ * of this interface to not handle too many things at once.
+ *
+ * Tightly coupled classes should be split into smaller interfaces which have a clear responsibility and are easier
+ * to understand and implement.
+ *
+ * @configuration threshold - the amount of class references to trigger the rule (default: `10`)
```
schalkms:
```suggestion
 * @configuration threshold - the number of class references to trigger the rule (default: `10`)
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CoupledClass.kt
```diff
@@ -0,0 +1,83 @@
...
+class CoupledClass(
+    config: Config = Config.empty,
+    threshold: Int = DEFAULT_MAX_REFERENCED_CLASSES
+) : ThresholdRule(config, threshold) {
+
+    override val issue = Issue(
+        javaClass.simpleName, Severity.Maintainability,
+        "A class is coupled to too many other classes. " +
+                "Coupled classes violate the Single Responsibility Principle. " +
+                "A class should have one responsibility. " +
```
schalkms:
```suggestion
                "A class should have exactly one responsibility. " +
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CoupledClassSpec.kt
```diff
@@ -0,0 +1,33 @@
+package io.gitlab.arturbosch.detekt.rules.complexity
+
+import io.gitlab.arturbosch.detekt.test.compileAndLint
+import org.assertj.core.api.Assertions.assertThat
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+
+class CoupledClassSpec : Spek({
```
schalkms: There is only a single test case. This doesn't cover all cases.
The test should be tested more thoroughly in order to reach detekt's quality gate.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CoupledClass.kt
```diff
@@ -0,0 +1,83 @@
...
+                "Split up large classes into smaller ones that are easier to understand.",
+        Debt.TWENTY_MINS
+    )
+
+    override fun visitClass(klass: KtClass) {
```
schalkms: `KtObjects` should also be checked.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CoupledClass.kt
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        super.visitClass(klass)
+    }
+
+    private fun fetchReferencedClasses(secondaryConstructors: List<KtConstructor<*>>): List<String?> {
+        return secondaryConstructors.flatMap { fetchReferencedClasses(it) }
+    }
+
+    private fun fetchReferencedClasses(constructor: KtConstructor<*>?): List<String?> {
```
schalkms: Does this paremter type have to be null? - `constructor: KtConstructor<*>?`

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CoupledClass.kt
```diff
@@ -0,0 +1,83 @@
...
+        if (constructor != null) {
+            return constructor.getValueParameters()
+                .filter { it.typeReference?.typeElement is KtUserType }
+                .map {
+                    (it.typeReference?.typeElement as KtUserType).referencedName
```
schalkms: Unsafe cast shouldn't be used in productive code, since it could yield a NPE.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CoupledClass.kt
edsoncunha-hotmart: Thanks, @schalkms. I didn't understand. You mean I should also override `fun 
 visitObjectDeclaration`?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CoupledClass.kt
edsoncunha-hotmart: Where is a filter on line 71, but I'll make the code more readable on the next commit.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CoupledClass.kt
schalkms: Yes
`KtClassOrObject` could also work in your case.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CoupledClass.kt
schalkms: That’s nice

arturbosch:
This pmd rule seems related, should we mention it in our description? 

https://github.com/pmd/pmd/blob/master/pmd-java/src/main/java/net/sourceforge/pmd/lang/java/rule/design/CouplingBetweenObjectsRule.java

It''s implementation also filters primitive types. We probably should do this too.

edsoncunha:
@arturbosch  I was wondering if it would be a good idea to have a separate rule for constructors. We usually use them for dependency injection. If take local variables and return types into consideration, the coupling threshold tends to bee higher. It could lead to a lot of objects getting passed through constructors, which is a smell of low cohesion.

schalkms:
> I was wondering if it would be a good idea to have a separate rule for constructors. We usually use them for dependency injection. If we take local variables and return types into consideration, the coupling threshold tends to bee higher. It could lead to a lot of objects getting passed through constructors, which is a smell of low cohesion.

Sorry for the late answer. Good idea. Please go ahead.

schalkms:
@edsoncunha since this PR become inactive, I'm wondering whether you need more assistance.
I'd be great to get this rule into detekt. It would be a very nice addition.

amitdash291:
> @edsoncunha since this PR become inactive, I'm wondering whether you need more assistance.
> I'd be great to get this rule into detekt. It would be a very nice addition.

@arturbosch @schalkms @edsoncunha
Is this still relevant? I can try to revive this if it makes sense.

schalkms:
@amitdash291 please go ahead! 🙂

PR #2315, opened 2020-02-04
Bugfix: MagicNumber with ignoreNamedArgument and a negative value

jonasfa:
Fixes https://github.com/arturbosch/detekt/issues/530

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
@@ -170,7 +170,15 @@ class MagicNumber(config: Config = Config.empty) : Rule(config) {
     }
 
     private fun KtConstantExpression.isNamedArgument(): Boolean {
-        val valueArgument = ((parent as? KtPrefixExpression)?.parent ?: parent) as? KtValueArgument
+        /**
+         * The information we need is in the enclosing [KtValueArgument]. When the number being evaluated is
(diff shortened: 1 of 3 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
```diff
@@ -169,8 +169,10 @@ class MagicNumber(config: Config = Config.empty) : Rule(config) {
                 .removeSuffix("f")
     }
 
-    private fun KtConstantExpression.isNamedArgument() =
-        (parent as? KtValueArgument)?.isNamed() == true && isPartOf<KtCallElement>()
+    private fun KtConstantExpression.isNamedArgument(): Boolean {
+        val valueArgument = ((parent as? KtPrefixExpression)?.parent ?: parent) as? KtValueArgument
```
schalkms: This is very hard to understand and maintain.
It would be good to describe what each case is doing with better naming.
Can we please refactor this expression?

schalkms:
Thanks for this bug fix as a your first contribution!

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
jonasfa: Thanks for the feedback. I've expanded the elvis operator into a `when` expression - hopefully the logic is more clear now. I also added a comment explaining why it exists, as the involved API names are very technical by nature and may not communicate the intent so well. What do you think?

jonasfa:
Rebased

PR #2319, opened 2020-02-05
use same behaviour for valueOrNull as for valueOrDefault

### LINKED ISSUE #2316: FailFastConfig.valueOrNull should return specified value for active and maxIssues
## Expected Behavior
`valueOrNull` delegates to originalConfig if value was defined

## Observed Behavior
`valueOrNull` does not delegate to originalConfig for `active` and for `maxIssues`

## Steps to Reproduce
ConfigurationsSpec -> should override active when specified -> switch to valueOrNull

## Context
I had a look at the assertions used in detekt to see if [Atrium](https://github.com/robstoll/atrium) would be a good fit. I'll create a WIP PR later on so that you can decide on your own.

## Your Environment
* Version of detekt used: master

robstoll:
- implemented valueOrDefault based on valueOrNull per default
- kept a different implementation for Test/YamlConfig

-fixes #2316

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
@@ -18,7 +18,7 @@ interface Config {
      * Retrieves a sub configuration or value based on given key. If configuration property cannot be found
      * the specified default value is returned.
      */
-    fun <T : Any> valueOrDefault(key: String, default: T): T
+    fun <T : Any> valueOrDefault(key: String, default: T): T = valueOrNull(key) ?: default
 
     /**
      * Retrieves a sub configuration or value based on given key.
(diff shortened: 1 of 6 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
```diff
@@ -86,9 +86,10 @@ internal object EmptyConfig : HierarchicalConfig {
     override fun subConfig(key: String): EmptyConfig = this
 
     @Suppress("UNCHECKED_CAST")
-    override fun <T : Any> valueOrDefault(key: String, default: T): T = if (key == "active") true as T else default
-
-    override fun <T : Any> valueOrNull(key: String): T? = null
+    override fun <T : Any> valueOrNull(key: String): T? = when(key){
+        "active" -> true as? T
```
robstoll: maybe it should also cover `maxIssue` as in FailFastConfig?

schalkms:
Thanks for tackling that! It looks good to me.

robstoll:
@schalkms I have updated the documentation. Moreover, I had to change the PR slightly: 
ConfigAware and CompositeConfig need to override valueOrDefault due to YamlConfig and TestConfig which implement an additional behaviour in valueOrDefault (LSP violation right in your face :laughing:).
I don't know the reasons behind the parsing in valueOrDefault and whether this could be moved into Config as well (so that all configs behave the same way) or not.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
arturbosch: Nah, maxIssues is now at 0 by default. The empty config just enables the rule under test.
I will move the BaseConfig and EmptyConfig to the internal package.

PR #2337, opened 2020-02-09
Add EmptyTryBlock rule

schalkms:

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyTryBlock.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyTryBlock.kt
@@ -1,0 +1,17 @@
+package io.gitlab.arturbosch.detekt.rules.empty
+
+import io.gitlab.arturbosch.detekt.api.Config
+import org.jetbrains.kotlin.psi.KtTryExpression
+
+/**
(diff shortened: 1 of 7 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyTryBlock.kt
```diff
@@ -0,0 +1,17 @@
+package io.gitlab.arturbosch.detekt.rules.empty
+
+import io.gitlab.arturbosch.detekt.api.Config
+import org.jetbrains.kotlin.psi.KtTryExpression
+
+/**
+ * Reports empty `try` blocks. Empty blocks of code serve no purpose and should be removed.
+ *
+ * @active since v1.5.1
```
3flex: Update with next release number?

schalkms: @arturbosch do we go straight to 1.6 or is another patch version coming?

BraisGabin: We should not activate any rule in a patch version.

schalkms: That’s true.

PR #2353, opened 2020-02-16
showcase use Atrium instead of AssertJ

robstoll:
This PR shows how the usage of Atrium could look like in detekt. I took CliArgsSpec not for special reasons (was the first Spec in detekt-cli) and rewrote ConfigSpec as well as they kind of relate.

Why Atrium and not AssertJ? IMO Atrium:
- has a more readable API
- has better error reporting
- is easy extendable

I have to say, I am of course biased as I am the author of Atrium. Personally, I am mainly interested in migrating detekt to Atrium as it would be a good use case to see how well I can migrate an AssertJ project to Atrium. I am not really an active contributor to detekt; I mainly report bugs. Therefore, you should of course think about if you see the value of Atrium as well or not. 

Please let me know if you are interested in Atrium, I would continue with re-writing the assertions in case you want to switch.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -53,7 +53,6 @@ tasks.withType<Detekt> {
 }
 
 val detektVersion: String by project
-val assertjVersion: String by project
 val spekVersion: String by project
 val reflectionsVersion: String by project
 val mockkVersion: String by project
@@ -302,7 +301,6 @@ subprojects {
         detekt(project(":detekt-cli"))
         detektPlugins(project(":detekt-formatting"))
 
-        testImplementation("org.assertj:assertj-core:$assertjVersion")
         testImplementation("org.spekframework.spek2:spek-dsl-jvm:$spekVersion")
         testImplementation("org.reflections:reflections:$reflectionsVersion")
         testImplementation("io.mockk:mockk:$mockkVersion")
--- a/detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/ConfigSpec.kt
+++ b/detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/ConfigSpec.kt
@@ -1,8 +1,11 @@
 package io.gitlab.arturbosch.detekt.api
 
+import ch.tutteli.atrium.api.fluent.en_GB.*
+import ch.tutteli.atrium.api.verbs.expect
+import io.gitlab.arturbosch.detekt.test.hasKeyValue
+import io.gitlab.arturbosch.detekt.test.hasNotKey
+import io.gitlab.arturbosch.detekt.test.valueOrDefault
 import io.gitlab.arturbosch.detekt.test.yamlConfig
-import org.assertj.core.api.Assertions.assertThat
-import org.assertj.core.api.Assertions.assertThatIllegalStateException
 import org.assertj.core.api.Assertions.fail
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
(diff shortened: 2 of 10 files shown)


arturbosch:
I like
- `expect` is shorter than `assertThat`
- auto completion for expect instead of static import of assertThat
- no `::class.java`

Can you post some examples of assertion outputs in comparison with assertJ ?
I'm interested how migrating `FindingsAssertions` would look like with `atrium`.

Any advantages over `strikt` (https://github.com/robfletcher/strikt) ?

3flex:
IMHO the benefits of Atrium over AssertJ aren't really worth it, particularly because of what looks like akward syntax (to my eyes) compared to AssertJ's fluent API, as well as the long list of Kotlin compiler bugs that are restricting Atrium's API as mentioned in the project README.

I'd be curious to see how things mature when Kotlin 1.4.0 comes out but I'm personally of the view that there's not enough benefit to switching at this time.

There definitely are benefits to using a Kotlin based assertion library to help tidy some syntax but it comes with worse syntax in other areas which isn't a good tradeoff.

Also if we did move to a Kotlin library the most appropriate should be selected from a group like Kluent, Strikt, HamKrest, Expekt, AssertK (and surely others). I haven't used any personally so maybe Atrium is best, I just think if we switch then an evaluation of alternatives should be done too.

schalkms:
I’m against this. I don’t think it’s worth the time and effort.
I only see readability as a minor advantage. Otherwise, detekt should aim to use a mature framework, since we really depend on the test suite strongly.

robstoll:
@arturbosch 

> Can you post some examples of assertion outputs in comparison with assertJ ?
I'm interested how migrating FindingsAssertions would look like with atrium.

Will do, also rewrite FindingsAssertions but probably not more at the moment due to the other voices, I guess you need to get an agreement here first.

<hr/>

@3flex
> particularly because of what looks like akward syntax (to my eyes) compared to AssertJ's fluent API,Could you please elaborate, I am not sure what you mean as Atrium has a fluent API as well. I guess you refer to the assertion blocks which corresponds to soft assertions in AssertJ. I prefer to see all failing assertions when something fails and that's why I am using it normally. For instance:
```
expect(config) {
  hasNotKey("one")
  hasNotKey("two")
  hasNotKey("three")
}
```
Could also be written as
```
expect(config)
  .hasNotKey("one")
  .hasNotKey("two")
  .hasNotKey("three")
```
Or did you mean something else?

> I'd be curious to see how things mature when Kotlin 1.4.0

Me too, JetBrains took Atrium as example project to fix some of the bugs. I hope I will then be able to provide a better syntax for feature assertions. But note, basically all assertion libraries using a reflective approach to feature assertions suffer from those bugs. The main difference is that I report them

> There definitely are benefits to using a Kotlin based assertion library to help tidy some syntax but it comes with worse syntax in other areas which isn't a good tradeoff.

Out of curiosity, could you elaborate `it comes with worse syntax in other areas` please?

> Also if we did move to a Kotlin library the most appropriate should be selected from a group

Makes sense, choose what suites you best. Feel free to close this PR at any time.

<hr/>

@schalkms

> I don’t think it’s worth the time and effort.

I think that shouldn't be really a concern as it is mostly my time and effort.

> Otherwise, detekt should aim to use a mature framework, since we really depend strongly on the test suite.

I am curious if you think about something in particular.

schalkms:
> I think that shouldn't be really a concern as it is mostly my time and effort.

I agree with the migration part and I’m thankful that you tackle that. However, this stuff needs to be maintained as well. This is often underestimated in the software engineering industry.
I also understand that you try to advertise your open source library. I’m grateful for anyone providing useful software to the community.

Regarding the maturity:
* Your library is maintained by only you. What if you decide to not maintain it anymore? Can anyone pick it up?
* Other libraries are more battle proved since they are simply used by more projects.

robstoll:
> However, this stuff needs to be maintained as well. This is often underestimated in the software engineering industry.

I totally agree with you. In case detekt switches to Atrium, then I would include it into a community-build https://github.com/robstoll/atrium-roadmap/issues/6
Which means we will see if we break things unintentionaly in a new version before releasing the new version. Moreover, I would likely use detekt to test if migration paths of new versions work out smoothly. Due to those two points I would assume that the maintenance effort for detekt contributors should be quite low. 
But... if you decide to use Atrium, then you probably have to learn a bit about Atrium in order to use it efficiently and that is of course something I can not take of the shoulders but I am happy if I can help out.

> Your library is maintained by only you. What if you decide to not maintain it anymore? Can anyone pick it up?

Currently I am the main contributor but @jGleitz is actively taking part in the progression of Atrium. IMO anyone can pick it up but of course with some additional effort.

> Other libraries are more battle proved since they are simply used by more projects.

That's definitely true.

arturbosch:
I respect the opinions of @schalkms and @3flex and we postpone the decision to choose an additional/replacing kotlin assertion library for after 1.4.0 of Kotlin.

My thoughts regarding assertion libraries and non mature frameworks:

We were early adopters of `spek` and the `kotlin-dsl`. Working with both was quite a pain. Slow IDE experience and broken plugins (Spek does not support IDEA 2020-EA yet). I even migrated away from spek (Junit5 is like always stable ;)) and didn't wanted to use the kotlin-dsl whenever possible.

I've used hamkrest, assertk, hamcrest, assertj, strikt and atrium in 2018/2019 and had like no big problems with them. I've settled mostly with `strikt` but would give `atrium` totally a chance for that better reporting. The biggest pain point for me to use atrium would be to have to look at `en_GB` in the imports ;).
 When I look at the GitHub repositories, all assertion libraries have just one maintainer (expect assertj). Migrating away from an assertion library would not consume that much of time imho.

arturbosch:
@robstoll thanks for showing the capabilities of atrium in this PR!

robstoll:
Thanks for your time. I'll ping once Kotlin 1.4 is out and Atrium further improved.

PR #2355, opened 2020-02-16
Include options to better meet requirements for utility files

### LINKED ISSUE #1500: Change MatchingDeclarationName to handle utility files
## Context
This is based on the discussion at the bottom of https://github.com/arturbosch/detekt/issues/841

I have a utility file.  It consists almost exclusively of top-level functions.  However, one of these returns an instance of a `data class` that is also declared in this file.  Because that is the only top-level class/object, Detekt says the file should be named after that data class, even though it's fairly insignificant.

As @arturbosch suggested, something like `allowClassesWithModifiers = "data,enum"` may be a reasonable configuration option to handle this.

arturbosch:
#1500

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -354,6 +354,7 @@ naming:
     rootPackage: ''
   MatchingDeclarationName:
     active: true
+    mustBeFirst: true
   MemberNameEqualsClassName:
     active: true
     ignoreOverridden: true
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MatchingDeclarationName.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MatchingDeclarationName.kt
@@ -43,31 +43,47 @@ import org.jetbrains.kotlin.psi.psiUtil.isPrivate
  *
  * </compliant>
  *
+ * @configuration mustBeFirst - name should only be checked if the file starts with a class or object (default: `true`)
+ *
  * @active since v1.0.0
  */
 class MatchingDeclarationName(config: Config = Config.empty) : Rule(config) {
 
-    override val issue: Issue = Issue(javaClass.simpleName, Severity.Style,
-            "If a source file contains only a single non-private top-level class or object, " +
-                    "the file name should reflect the case-sensitive name plus the .kt extension.",
-            Debt.FIVE_MINS)
+    override val issue: Issue = Issue(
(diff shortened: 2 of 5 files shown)


detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -354,6 +354,8 @@ naming:
     rootPackage: ''
   MatchingDeclarationName:
     active: true
+    mustBeFirst: true
+    ignoreDataClasses: true
```
henrikbarium: Would it be possible to extend this configuration to take a list of class types to ignore? The same problem can occur with enum types and inline types. It would be nice to configure for example:

```
ignoreClassTypes: ['data', 'enum', 'inline']
```

schalkms: It's possible. PRs are very welcome in that regard. Please note that we rely on contributions here, since such things have a low prior for the detekt team atm.

henrikbarium: I understand. I was hoping to ask a constructive question rather than to come across as ungrateful. Sorry about that.

schalkms: @henrikbarium you didn’t sound aggressive.
We always welcome ideas for new features or rules and try to prioritize them accordingly. There’s a general idea about what problems detekt should solve. This is also displayed in current PR’s and on the roadmap.
Atm we try to polish detekt’s core while keeping the quality bar high.

Ofc we provide assistance when writing rules as much as possible. Don’t hesitate to reach out and seek for help. Detekt’s community tries to answer questions either here or on the Slack platform.

arturbosch: @henrikbarium @schalkms this is a good point actually.
Because if we now introduce the `ignoreDataClasses` and extend this for other classes do we want to deprecate `ignoreDataClasses` in favor for `ignoreClassTypes` or do we want to introduce additonal properties like `ignoreEnumClasses` and `ignoreInlineClasses`?

henrikbarium: If there are three possible values I guess individual properties would be ok. (Are there more?) Any more than that, the list-based version would be nicer. Personally I would prefer the list-based one anyway, as I find it easier to read when the keywords are listed individually rather than combined into camelCasePropertyNames. I also think it is nice as it groups the behavior related to ignoring types of classes into one logical property. I would generally assume that different properties relate to different functionality.

arturbosch:
I've removed the `ignoreDataClasses` property for now.
In a later update we can take `data, inline and enum classes` into account.

PR #2364, opened 2020-02-23
Improve travis cache

BraisGabin:
Don't cache no important parts of gradle that invalidate the cache each time.
This invalidation consumes 60~80 seconds each run.

Captura de pantalla 2020-02-23 a las 11 17 08

And, because we are caching less things, the restore should be a bit faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,8 +36,7 @@ after_success:
   - bash <(curl -s https://codecov.io/bash)
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
-  - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
 cache:
   directories:
-    - $HOME/.gradle/caches/
-    - $HOME/.gradle/wrapper/
+    - $HOME/.gradle/caches/modules-2/
+    - $HOME/.gradle/wrapper/dists/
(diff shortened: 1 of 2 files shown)


BraisGabin:
Related topic: the size of the travis cache is 10GB!:
Captura de pantalla 2020-02-23 a las 11 18 18

I'll delete all the caches in travis as soon as this PR is marged. This way we force travis to start with an empty cache so it'll go back to a ~400MB. After this the restore will be much much faster (right now it's around 45 seconds).

BraisGabin:
Appvenyor maybe suffers the same problem... But its log displays nothing so I don't know.

BraisGabin:
Added appvenyor

BraisGabin:
I'm blocking this PR. I want to make some measurements with travis before merged it.

BraisGabin:
Closing this PR because it doesn't improve the travis time. It makes it worst! It's true that the restore and store of cache is faster but we are saving **A LOT** of time (7 minutes or more) because of the cached builds. We don't have really fresh builds on travis but I think that the time saving is worth it. Our tests are **really** slow I'll check that.

arturbosch:
Our Gradle tests are really slow, maybe we can skip them on non LTS JVMs?

BraisGabin:
The problem is that we run the tests 7 times and only 1 is a non LTS JVM. And it runs in parallel with other 4 LTS JVM so the improvemen would be nearly none...

Issue #2365, opened 2020-02-23
Deprecate api's exposing detekt-core implementation details


arturbosch:
## Expected Behavior
Following classes/functions exposes implementation details and should be moved to detekt-core or the inernal package:
- RuleSet#accept
- RuleSetProvider#buildRuleSet
- EmptyConfig
- YamlConfig
- BaseConfig
- BaseRule (sealed)

Following classes should be supported but moved to a utils package
- LazyRegex
- AnnotationExcluder
- SplitPattern

## Current Behavior
Above mentioned classes expose implementation details. 

## Context
Code in `detekt-api` exposes internal implementation details of `detekt-core`

Issue #2341, opened 2020-02-10
Mixing autocorrectable and non correctable rules results in obsolete issue locations for reports


omainegra:
I found an error in the html report:
- Rule: TooGenericExceptionCaught
- Detetk version: 1.5.1
- Stacktrace:
```
java.lang.StringIndexOutOfBoundsException: String index out of range: 19
	at java.lang.String.substring(String.java:1963)
	at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt.writeErrorLine(HtmlUtils.kt:51)
	at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt.access$writeErrorLine(HtmlUtils.kt:1)
	at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt$snippetCode$1$1.invoke(HtmlUtils.kt:35)
	at io.gitlab.arturbosch.detekt.cli.out.HtmlUtilsKt$snippetCode$1$1.invoke(HtmlUtils.kt)
```
- How to reproduce it: <WRITE HERE HOW TO REPRODUCE THIS ISSUE. A CODE SNIPPET IS THE BEST WAY.>

BraisGabin:
Could you provide a sample code to reproduce this issue?

omainegra:
🤦‍♂ I fix the code and now is working fine, let me check the history

omainegra:
I was able to reproduce it with the following snippet

```kotlin
enum class SomeEnum {
    VALUE1, VALUE2, VALUE3, UNKNOWN;

    companion object {
        fun from(string: String): SomeEnum =
            try { valueOf(string) }
            catch (e: Exception) { UNKNOWN }
    }
}
```

BraisGabin:
WOW Really strange behaviour. The problem is not related with the Exception but with the code itself. If I request the `psiFile.text` I get this:

```kotlin
enum class SomeEnum {
    VALUE1, VALUE2, VALUE3, UNKNOWN;

    companion object {
        fun from(string: String): SomeEnum =
            try { valueOf(string) } catch (e: Exception) { UNKNOWN }
    }
}
```

the `catch` is moved to the same line as the `try {}`.

Any idea about why this is happening? It seems an error in PSI.

schalkms:
@BraisGabin even if the Psi had a problem the HtmlReport shouldn’t crash.

BraisGabin:
It's not crashing. We fixed that already. But it shows this:
[...]

PR #2366, opened 2020-02-23
Deprecate rule set methods which expose implementation details of detekt-core

arturbosch:
Refactorings for #2365 and #2341

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSet.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSet.kt
@@ -21,12 +21,14 @@ class RuleSet(val id: RuleSetId, val rules: List<BaseRule>) {
     /**
      * Is used to determine if a given [KtFile] should be analyzed at all.
      */
+    @Deprecated("Exposes detekt-core implementation details.")
     var pathFilters: PathFilters? = null
 
     /**
      * Visits given file with all rules of this rule set, returning a list
      * of all code smell findings.
      */
+    @Deprecated("Exposes detekt-core implementation details.")
     fun accept(file: KtFile, bindingContext: BindingContext = BindingContext.EMPTY): List<Finding> =
         if (isFileIgnored(file)) {
             emptyList()
@@ -37,6 +39,7 @@ class RuleSet(val id: RuleSetId, val rules: List<BaseRule>) {
             }
         }
 
+    @Suppress("DEPRECATION")
     private fun isFileIgnored(file: KtFile) =
         pathFilters?.isIgnored(Paths.get(file.absolutePath())) == true
 }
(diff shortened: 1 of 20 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSetProvider.kt
```diff
@@ -22,6 +22,8 @@ interface RuleSetProvider {
      * Api notice: As the rule set id is not known before creating the rule set instance,
      * we must first create the rule set and then check if it is active.
      */
+    @Suppress("DEPRECATION")
+    @Deprecated("Exposes detekt-core implementation details.")
     fun buildRuleset(config: Config): RuleSet? {
```
BraisGabin: Does it?

arturbosch: A rule writer does not need to know how his rules will be created and how the active check is done and how we actually use the `includes` and `excludes` to filter if a KtFile should be analyzed. A `RuleSetProvider` just provide the rules.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSet.kt
```diff
@@ -21,12 +21,14 @@ class RuleSet(val id: RuleSetId, val rules: List<BaseRule>) {
...
      */
+    @Deprecated("Exposes detekt-core implementation details.")
     var pathFilters: PathFilters? = null
 
     /**
      * Visits given file with all rules of this rule set, returning a list
      * of all code smell findings.
      */
+    @Deprecated("Exposes detekt-core implementation details.")
     fun accept(file: KtFile, bindingContext: BindingContext = BindingContext.EMPTY): List<Finding> =
```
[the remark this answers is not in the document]
arturbosch: The body of the function exposes how detekt internals work.
The user of the api does not need to know how `detekt-core` creates instances of the rules and runs over the KtFile's.

arturbosch:
> I'm trying to understand this refactor. At the end how does we want to `RuleSet` be like?
> 
> just a class like this:
> 
> ```kotlin
> class RuleSet(val id: RuleSetId, val rules: List<BaseRule>, include: String, exclude: String) {
>   private val pathFilters: PathFilters = PathFilters.of(include, exclude)
> 
>   fun shouldAnalyzeFile(file: KtFile): Boolean = //...
> }
> ```
> 
> And nothing else?

If I would rewrite detekt now, I would consider this code for a RuleSetProvider:
```
interface RuleSetProvider {
    val ruleSetId: String

    fun rules(config: Config): List<Rule>
}
```

A rule author does not need to know that we bundle the rules inside a `RuleSet` class which has some logic like `filtering` and `isActive` etc.
He knows that there are some cli flags and config propeties for this but does not need to see how we iterate over rules, visit them and collect the findings :).

BraisGabin:
Oh, OK, so you would remove the rule set from the api all together. That have sense.

BraisGabin:
Oh! We should add information in the depreciation message about what the user should use instead. And, if possible, the regex to do it automatically.

arturbosch:
The problem is the user should not use these methods at all. There is no migration for the user.
He can test his rules one by one. Or if he really wants it he can pull in `detekt-test` and start a `Runner`.

PR #2378, opened 2020-02-29
Sort and run correctable rules first - #2341

arturbosch:


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Detektor.kt
```diff
@@ -70,12 +76,41 @@ class Detektor(
...
+        val result = HashMap<RuleSetId, MutableList<Finding>>()
+
+        fun executeRules(rules: List<BaseRule>) {
+            for (rule in rules) {
+                rule.visitFile(file, bindingContext)
+                for (finding in rule.findings) {
+                    val mappedRuleSet = idMapping[finding.id]
+                        ?: error("Mapping for '${finding.id}' expected.")
+                    result.putIfAbsent(mappedRuleSet, ArrayList())
+                    result[mappedRuleSet]?.add(finding)
```
BraisGabin:
```suggestion
                    result[mappedRuleSet]!!.add(finding)
```

This case should never happen but it's better to crash that miss some finding.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Detektor.kt
```diff
@@ -70,12 +76,41 @@ class Detektor(
...
-            .sortedBy { it.id }
-            .map { ruleSet -> ruleSet.id to ruleSet.visitFile(this, bindingContext) }
-            .toMergedMap()
+            .filter { it.shouldAnalyzeFile(file, config) }
+            .associate { it.id to it.rules }
+
+        fun isCorrectable(rule: BaseRule): Boolean = when (rule) {
+            is Rule -> rule.autoCorrect
+            is MultiRule -> rule.rules.any { it.autoCorrect }
+            else -> error("No other rule type expected.")
```
BraisGabin: Out of the scope of this PR but maybe we can refactor `BaseRule` to a `sealed` class

arturbosch: Totally thought the same, however this is a breaking change, I will open an Issue for the `future breaking changes`-milestone.

PR #2383, opened 2020-03-02
Add --version to cli

### LINKED ISSUE #2382: Add `--version` CLI option
Hello, I think more useful for many people if detekt would have a `--version` option to show the version to console.

## Expected Behavior

```shell-session
$ detekt --version
1.6.0
```

## Current Behavior

```shell-session
$ detekt --version
Was passed main parameter '--version' but no main parameter was defined in your arg class

Usage: detekt [options]
  Options:
    --auto-correct, -ac
...
```

## Context

For example, ktlint gets its version from its manifest file,

https://github.com/pinterest/ktlint/blob/9eb31ca6fbe1fd3f0fca708fafc15468218cecc5/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/KtlintVersionProvider.kt#L16-L18

so I believe that detekt can show the version in a similar way.

What do you think about this suggestion?
Thank you for reading.

BraisGabin:
Closes #2382

This PR adds the --version flag. It prints the version of the running detekt cli.

I have a problem testing this. And It's that `whichVersion()` retutns `null` in test mode. Any idea?

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinter.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinter.kt
@@ -1,0 +1,16 @@
+package io.gitlab.arturbosch.detekt.cli.runners
+
+import io.gitlab.arturbosch.detekt.core.whichDetekt
+import java.io.PrintStream
+
+class VersionPrinter(private val outputPrinter: PrintStream) : Executable {
+
+    override fun execute() {
+        val version = whichDetekt()
+        if (version != null) {
+            outputPrinter.println(version)
+        } else {
+            throw IllegalStateException("Can't find the detekt version")
+        }
+    }
+}
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinterSpec.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinterSpec.kt
@@ -1,0 +1,24 @@
+package io.gitlab.arturbosch.detekt.cli.runners
+
+import io.gitlab.arturbosch.detekt.core.Detektor
+import org.assertj.core.api.Assertions.assertThat
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+import java.io.ByteArrayOutputStream
+import java.io.PrintStream
+import java.nio.charset.Charset
+
+class VersionPrinterSpec : Spek({
+
+    describe("version printer") {
+
+        it("print the version") {
+            val byteArrayOutputStream = ByteArrayOutputStream()
+
+            VersionPrinter(PrintStream(byteArrayOutputStream)).execute()
+
+            assertThat(String(byteArrayOutputStream.toByteArray(), Charset.forName("UTF-8")))
+                .isEqualTo("1.6.0" + System.lineSeparator())
+        }
+    }
+})
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
@@ -173,6 +173,12 @@ class CliArgs : Args {
     )
     var jvmTarget: JvmTarget = JvmTarget.DEFAULT
 
+    @Parameter(
+        names = ["--version"],
+        description = "Prints the detekt CLI version."
+    )
+    var showVersion: Boolean = false
+
     val inputPaths: List<Path> by lazy {
         MultipleExistingPathConverter().convert(input ?: System.getProperty("user.dir"))
     }
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Main.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Main.kt
@@ -8,6 +8,7 @@ import io.gitlab.arturbosch.detekt.cli.runners.ConfigExporter
 import io.gitlab.arturbosch.detekt.cli.runners.Executable
 import io.gitlab.arturbosch.detekt.cli.runners.Runner
 import io.gitlab.arturbosch.detekt.cli.runners.SingleRuleRunner
+import io.gitlab.arturbosch.detekt.cli.runners.VersionPrinter
 import java.io.PrintStream
 import kotlin.system.exitProcess
 
@@ -48,6 +49,7 @@ fun buildRunner(
         }
     }
     return when {
+        arguments.showVersion -> VersionPrinter(outputPrinter)
         arguments.generateConfig -> ConfigExporter(arguments)
         arguments.runRule != null -> SingleRuleRunner(arguments)
         arguments.printAst -> AstPrinter(arguments)
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/MainSpec.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/MainSpec.kt
@@ -4,6 +4,7 @@ import io.gitlab.arturbosch.detekt.cli.runners.AstPrinter
 import io.gitlab.arturbosch.detekt.cli.runners.ConfigExporter
 import io.gitlab.arturbosch.detekt.cli.runners.Runner
 import io.gitlab.arturbosch.detekt.cli.runners.SingleRuleRunner
+import io.gitlab.arturbosch.detekt.cli.runners.VersionPrinter
 import org.assertj.core.api.Assertions.assertThat
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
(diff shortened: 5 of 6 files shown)


schalkms:
You were just a few seconds faster than me in regards to the PR for this topic.
I guess it is because of the manifest file that couldn't be loaded.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinter.kt
```diff
@@ -0,0 +1,11 @@
+package io.gitlab.arturbosch.detekt.cli.runners
+
+import io.gitlab.arturbosch.detekt.core.whichDetekt
+import java.io.PrintStream
+
+class VersionPrinter(private val outputPrinter: PrintStream) : Executable {
+
+    override fun execute() {
+        outputPrinter.println(whichDetekt() ?: "Unknown")
```
schalkms: I'd move `whichDetekt() ?: "Unknown"` into the method and adapt the other call side in `HtmlUtils.kt` as well.

arturbosch:
I think it is because of these lines:
https://github.com/arturbosch/detekt/blob/25a8555c03e680096dcf390f23b4a51eea78aaee/detekt-cli/build.gradle.kts#L18-L23
The manifest file is only available inside the jar.
Will it help if you hardcode a manifest file inside the `test/resources`?

BraisGabin:
Yes! adding a manifest in the test resources did the trick.

I edited the code too. I we can't know the version of detekt we should fail the task. We don't want to exit with a success because that can mess up the scripts of the users.

I can't test the failure case but it should never happen.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinter.kt
BraisGabin: I rewrote this code. If the version is `null` we should fail.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinter.kt
```diff
@@ -0,0 +1,16 @@
...
+
+class VersionPrinter(private val outputPrinter: PrintStream) : Executable {
+
+    override fun execute() {
+        val version = whichDetekt()
+        if (version != null) {
+            outputPrinter.println(version)
+        } else {
+            throw IllegalStateException("Can't find the detekt version")
+        }
```
schalkms: I'm against doing this. The other call site of `whichDetekt()` also doesn't throw an exception. I'd prefer to inline both `?:` operators into the mentioned function.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinter.kt
schalkms: Please see my comment underneath.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinter.kt
BraisGabin: As a script we can't exit with success if we don't show the version to the user. The user only asked us for one thing, so if we can't show it we should fail. Right now the only way to exit with an error code is to throw an exception.

I can do cooler things like creating my own exception and adding that exception in the try/catch that we have in the `main` function. But I think that this error message is good enough for such a edge case:

Captura de pantalla 2020-03-02 a las 22 16 30

schalkms: Agreed. Maybe you should add another test case, which checks if the exception is thrown.

schalkms: Agreed

BraisGabin: 😭 that's a problem. The way to test the happy path is to add a fake manifest file in `test/resources`. I didn't find a way to create/remove a file there programatically so I can fake the different states. Any idea?

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinterSpec.kt
```diff
@@ -0,0 +1,25 @@
...
+import java.io.ByteArrayOutputStream
+import java.io.PrintStream
+import java.nio.charset.Charset
+
+class VersionPrinterSpec : Spek({
+
+    describe("version printer") {
+
+        describe("print the version") {
+            Detektor::class.java.classLoader.getResource("")
```
arturbosch: Do we need this line?

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinterSpec.kt
```diff
@@ -0,0 +1,25 @@
...
+            val byteArrayOutputStream = ByteArrayOutputStream()
+
+            VersionPrinter(PrintStream(byteArrayOutputStream)).execute()
+
+            assertThat(String(byteArrayOutputStream.toByteArray(), Charset.forName("UTF-8")))
+                .isEqualTo("1.6.0" + System.lineSeparator())
```
arturbosch: Should we `trim` the detektVersion in `whichVersion`?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinter.kt
schalkms: Don't reference the manifest file for the second path through the method!?

schalkms: Don't reference the manifest file for covering the second path through the method!?

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinterSpec.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.spekframework.spek2.style.specification.describe
+import java.io.ByteArrayOutputStream
+import java.io.PrintStream
+import java.nio.charset.Charset
+
+class VersionPrinterSpec : Spek({
+
+    describe("version printer") {
+
+        describe("print the version") {
```
arturbosch: Should this be a `it` block?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinter.kt
arturbosch: I don't think we can test this with a unit test.
One way would be to introduce a script which is run on CI to test some common cases/states.
This would be another testcase next to our current two:
```
  - java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar -i . --baseline ./config/detekt/baseline.xml -ex "**/resources/**,**/detekt*/build/**" -c ./detekt-cli/src/main/resources/default-detekt-config.yml,./config/detekt/detekt.yml
```

But I think this would be out of scope for this PR.

schalkms: Per default if you don't pass a _special_ config file the exception is thrown, right?

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinterSpec.kt
BraisGabin: nope! That was me trying to create/remove resources at runtime

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinter.kt
arturbosch: What do you mean by config file? The manifest file is not a config file but is generated while running the `jar` task. For this test we have a fake manifest file which is always present, therefore we won't be able to execute the `throw` part in a test without removing the `test/resources/manifest` file, or do I miss something?

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinterSpec.kt
BraisGabin: `whichVersion` returns `"1.6.0"` as expected. I'm the one adding the `\n` using `println` instead of `print`. Otherwise we print things like this:

```
╭─brais ~/projects/detekt
╰─○ java -jar ~/projects/detekt/detekt-cli/build/libs/detekt-cli-1.6.0-all.jar --version             
1.6.0╭─brais ~/projects/detekt
╰─○ 
```
intead of
```
╭─brais ~/projects/detekt
╰─○ java -jar ~/projects/detekt/detekt-cli/build/libs/detekt-cli-1.6.0-all.jar --version             
1.6.0
╭─brais ~/projects/detekt
╰─○ 
```

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/VersionPrinter.kt
BraisGabin: The problem is that I'm not passing a config here. I'm just adding a file in the `test/resorces`. It's not related with this test. It affects to ALL the tests in detekt-cli. Luckly no test check that except this one. So I just can test the case with manifest or the case without. But I can't test both. I would like to test both cases but I don't see the way :(

PR #2387, opened 2020-03-02
Refactor getting the detekt version for readability

arturbosch:


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DebugUtils.kt
```diff
@@ -9,21 +8,11 @@ fun whichOS(): String = System.getProperty("os.name")
...
     for (resource in Detektor::class.java.classLoader.getResources("META-INF/MANIFEST.MF")) {
-        try {
-            val version = readDetektVersionInManifest(resource)
-            if (version != null) {
-                return version
-            }
-        } catch (_: IOException) {
-            // we search for the manifest with the detekt version
-        }
+        return runCatching { readVersion(resource) }.getOrNull() ?: continue
```
arturbosch: Cool, codefactor was useful here :).
This is a false positive isn't it? The `?:` is a condition.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DebugUtils.kt
```diff
@@ -9,21 +8,11 @@ fun whichOS(): String = System.getProperty("os.name")
...
     }
     return null
```
BraisGabin: We just find a false positive in our rules 😄 I'm opening the issue. 

Anyway, what do you thing about something like this?

```kotlin
    return Detektor::class.java.classLoader.getResources("META-INF/MANIFEST.MF")
        .asSequence()
        .map { resource -> runCatching { readVersion(resource) }.getOrNull() }
        .filterNotNull()
        .firstOrNull()
```

arturbosch: Oh! Poor me, just saw `asIterator` from Java 9. Will refactor. Thanks for creating the issue :)

PR #2393, opened 2020-03-03
Update to Kotlin v1.3.70

schalkms:
ChangeLog
https://github.com/JetBrains/kotlin/blob/1.3.70/ChangeLog.md


schalkms:
I have no idea what the problem here is.

arturbosch:
I've opened a PR with two commits on top of yours.
There were some deprecations in the Kotlin compiler which CI always observes but locally you need to write `-PwarningsAsErrors=true`.

PR #2397, opened 2020-03-03
Use argsfiles for CI

### LINKED ISSUE #2318: Allow Detekt CLI to take an args file.
When using Detekt in the Uber monorepo, we've run into issues with extremely long argument lists that result in "Argument list too long" exceptions in the build, for things like Classpath or source lists. We've worked around it for now by trimming down our use of absolute paths in a custom Buck genrule.

Detekt CLI should have the ability to take an args file containing this information to work with larger argument lists that exceed the shell limits.

arturbosch:
Closes #2318

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ install: true
 script:
   - ./gradlew build shadowJar --build-cache --parallel -PwarningsAsErrors=true -Pcompile-test-snippets=$COMPILE_TEST_SNIPPETS --scan
   - java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar --help
-  - java -jar ./detekt-cli/build/libs/detekt-cli-*-all.jar -i . --baseline ./config/detekt/baseline.xml -ex "**/resources/**,**/detekt*/build/**" -c ./detekt-cli/src/main/resources/default-detekt-config.yml,./config/detekt/detekt.yml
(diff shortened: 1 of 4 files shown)


3flex:
Apparently JCommander supports this natively? https://jcommander.org/#_syntax

arturbosch:
Ah nice, good to know, thx!

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Main.kt
```diff
@@ -9,13 +9,27 @@ import io.gitlab.arturbosch.detekt.cli.runners.Executable
...
+        var arguments = args
+        if (shouldUseArgumentsFile()) {
+            val argsFile = args.first().removePrefix("@")
+            arguments = File(argsFile).useLines { seq ->
+                seq.flatMap { it.splitToSequence(" ") }
+                    .filterNot { it.isBlank() }
+                    .toList()
+                    .toTypedArray()
+            }
+        }
```
schalkms: If we stick with this solution, this code should have two corresponding tests.
This could work with a refactoring step.

3flex:
Just FYI @arturbosch you need to remove the hyphen from your initial description to automatically close the linked issue.

Use
`Closes #2318`
Not
`Closes - #2318`

arturbosch:
Thanks for your comments!
JCommander really supports argsfiles out of the box so we do not need to add any code.

PR #2399, opened 2020-03-04
Discourage new Case files

### LINKED ISSUE #1089: Cases files are super rigorous to work with

3flex:
Closes #1089 

That issue will remain open until all case files are retired, which may take a long time as it's quite a thankless task - instead, add a comment asking contributors not to add new case files with a reference to the PR so the PR can be closed.

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/Case.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/Case.kt
@@ -4,6 +4,8 @@ import io.gitlab.arturbosch.detekt.test.resource
 import java.nio.file.Path
 import java.nio.file.Paths
 
+/* Do not add new elements to this file.
+   See https://github.com/arturbosch/detekt/issues/1089 */
 enum class Case(val file: String) {
     CollapsibleIfsPositive("/cases/CollapsibleIfsPositive.kt"),
     CollapsibleIfsNegative("/cases/CollapsibleIfsNegative.kt"),
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/Case.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/Case.kt
@@ -4,7 +4,7 @@ import io.gitlab.arturbosch.detekt.test.resource
 import java.nio.file.Path
 import java.nio.file.Paths
 
-/* Do not add new elements to this file.
+/* Do not add new elements to this file. Instead, use inline code snippets within the tests.
    See https://github.com/arturbosch/detekt/issues/1089 */
 enum class Case(val file: String) {
     CollapsibleIfsPositive("/cases/CollapsibleIfsPositive.kt"),


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/Case.kt
```diff
@@ -4,6 +4,8 @@ import io.gitlab.arturbosch.detekt.test.resource
 import java.nio.file.Path
 import java.nio.file.Paths
 
+/* Do not add new elements to this file.
```
arturbosch: Lets encourage the usage of inline snippets here?

schalkms: I agree with Artur.
Furthermore, we should also strive to inline the case files when we change something in a rule or in the corresponding spec file, which still links a case file.

schalkms: I also try to mention this in PR reviews.

3flex: I've reworded, hopefully the updated text is OK?

PR #2401, opened 2020-03-04
ExplicitCollectionElementAccessMethod: Don't report on nullable collection

xouabita:
Should fix https://github.com/arturbosch/detekt/issues/2391


BraisGabin:
We should add tests for this.

PR #2132, opened 2019-11-27
Deprecate ignoreOverriddenFunction/s in favor of ignoreOverridden

BraisGabin:
Closes #2096 

`EmptyFunctionBlock`, `FunctionParameterNaming` and `MemberNameEqualsClassName` uses the configuration flag `ignoreOverriddenFunction(s)` but the rest of rules uses the generic `ignoreOverridden`. This PR change this to get an API a bit more consistent.

PR #2410, opened 2020-03-05
Added validation of constructors to LongParameterList

r4zzz4k:
I've looked through existing rules, seems that nothing covers number of parameters in contructors. LongParameterList is the nearest thing, but it covers only methods.

Does this sound like a useful rule? In discussions over the Internet on argument count limit I mostly see costructors included. Checkstyle, for example, covers [both methods and parameters](https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.html), PMD [seems to agree on that](https://stackoverflow.com/q/56492868/2475207).

After adding this plugin `./gradlew detekt` starts failing on the project, as some classes (e.g. CorrectableCodeSmell, see below) have 6 or more parameters in constructors.

https://github.com/arturbosch/detekt/blob/521e96856d841ec8c115a12f2db8230ce01eda86/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt#L44-L51

I'll try to collect proper report, as right now I'm having troubles with it (#2411).

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
@@ -5,36 +5,72 @@ import io.gitlab.arturbosch.detekt.api.Debt
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Metric
+import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
-import io.gitlab.arturbosch.detekt.api.ThresholdRule
 import io.gitlab.arturbosch.detekt.api.ThresholdedCodeSmell
 import io.gitlab.arturbosch.detekt.rules.isOverride
+import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtConstructor
+import org.jetbrains.kotlin.psi.KtFunction
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtParameterList
+import org.jetbrains.kotlin.psi.KtPrimaryConstructor
+import org.jetbrains.kotlin.psi.KtSecondaryConstructor
 
 /**
- * Reports functions which have more parameters than a certain threshold (default: 6).
+ * Reports functions and constructors which have more parameters than a certain threshold.
  *
  * @configuration threshold - number of parameters required to trigger the rule (default: `6`)
+ * (deprecated: "Use `functionThreshold` and `constructorThreshold` instead")
+ * @configuration functionThreshold - number of function parameters required to trigger the rule (default: `6`)
+ * @configuration constructorThreshold - number of constructor parameters required to trigger the rule (default: `7`)
  * @configuration ignoreDefaultParameters - ignore parameters that have a default value (default: `false`)
+ * @configuration ignoreDataClasses - ignore long constructor parameters list for data classes (default: `true`)
  *
  * @active since v1.0.0
  */
 class LongParameterList(
-    config: Config = Config.empty,
-    threshold: Int = DEFAULT_THRESHOLD_PARAMETER_LENGTH
-) : ThresholdRule(config, threshold) {
+    config: Config = Config.empty
+) : Rule(config) {
 
     override val issue = Issue("LongParameterList",
             Severity.Maintainability,
-            "The more parameters a method has the more complex it is. Long parameter lists are often " +
+            "The more parameters a function has the more complex it is. Long parameter lists are often " +
                     "used to control complex algorithms and violate the Single Responsibility Principle. " +
-                    "Prefer methods with short parameter lists.",
+                    "Prefer functions with short parameter lists.",
             Debt.TWENTY_MINS)
+
+    private val functionThreshold: Int =
+        valueOrDefault(FUNCTION_THRESHOLD, valueOrDefault(THRESHOLD, DEFAULT_FUNCTION_THRESHOLD))
+
+    private val constructorThreshold: Int =
+        valueOrDefault(CONSTRUCTOR_THRESHOLD, valueOrDefault(THRESHOLD, DEFAULT_CONSTRUCTOR_THRESHOLD))
 
     private val ignoreDefaultParameters = valueOrDefault(IGNORE_DEFAULT_PARAMETERS, false)
 
+    private val ignoreDataClasses = valueOrDefault(IGNORE_DATA_CLASSES, true)
+
     override fun visitNamedFunction(function: KtNamedFunction) {
+        validateFunction(function, functionThreshold)
+    }
+
+    override fun visitPrimaryConstructor(constructor: KtPrimaryConstructor) {
+        validateConstructor(constructor)
+    }
+
+    override fun visitSecondaryConstructor(constructor: KtSecondaryConstructor) {
+        validateConstructor(constructor)
+    }
+
+    private fun validateConstructor(constructor: KtConstructor<*>) {
+        val owner = constructor.getContainingClassOrObject()
+        val isDataClass = owner is KtClass && owner.isData()
+        if (!ignoreDataClasses || !isDataClass) {
+            validateFunction(constructor, constructorThreshold)
+        }
+    }
+
+    private fun validateFunction(function: KtFunction, threshold: Int) {
         if (function.isOverride()) return
         val parameterList = function.valueParameterList
         val parameters = parameterList?.parameterCount()
@@ -57,7 +93,13 @@ class LongParameterList(
     }
 
     companion object {
+        const val THRESHOLD = "threshold"
+        const val FUNCTION_THRESHOLD = "functionThreshold"
+        const val CONSTRUCTOR_THRESHOLD = "constructorThreshold"
         const val IGNORE_DEFAULT_PARAMETERS = "ignoreDefaultParameters"
-        const val DEFAULT_THRESHOLD_PARAMETER_LENGTH = 6
+        const val IGNORE_DATA_CLASSES = "ignoreDataClasses"
+
+        const val DEFAULT_FUNCTION_THRESHOLD = 6
+        const val DEFAULT_CONSTRUCTOR_THRESHOLD = 7
     }
 }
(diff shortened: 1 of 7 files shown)


r4zzz4k:
Sure, will do.

Mar 5, 2020 20:37:07 M Schalk <notifications@github.com>:

r4zzz4k:
I'm playing with it for some time and don't like both abstract parent for two rules (pushed above) and code duplication without it.

Does adding new configuration parameter like `constructorThreshold` to existing rule sound viable? With default value like -1 or 255 meaning "don't validate constructors".

I'm open to any other suggestions, of course.

schalkms:
For me this PR looks good. Please fix the issues detected issues. Then we'll proceed with merging this PR.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongConstructorParameterList.kt
```diff
@@ -0,0 +1,43 @@
...
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Metric
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.ThresholdRule
+import io.gitlab.arturbosch.detekt.api.ThresholdedCodeSmell
+import io.gitlab.arturbosch.detekt.rules.*
+import org.jetbrains.kotlin.psi.*
```
schalkms: Please don't use the `*` suffix for imports

schalkms: Can we please don't use the `*` suffix for imports?

arturbosch:
Both approaches have advantages:

1. own rule:
- clear separation of concerns
- does not interfere with current config properties of LongParameterList

2. extending LongParameterList
- pmd and checkstyle have just one rule for both cases
- co-location of properties (constructorThreshold, functionThreshold, threshold)
- two similar rule names: `LongParameterList` and `LongConstructorParameterList`; I would prefer deprecating LongParameterList then and build extra logic for a LongFunctionParameterList.

I'm OK with both approaches but would prefer a property `constructorThreshold` for LPL with a default -1 value.

BraisGabin:
Why the -1 as default? It doesn't seem like a good default value. If the reason is to avoid "BC" I consider this a bug fix more than a new feature.

If we go to `LongParameterList` we should deprecate `threshold` and add `constructorThreshold` and `functionThreshold`. I did this before to have a reference: #2132

r4zzz4k:
Thanks for the reference! Proper deprecation and splitting property into two seems much nicer than -1.
I'll try that one out, as I'm also leaning to single rule instead of two different, but very similar ones.

Hoping to get back to that in the nearest future.

schalkms:
I don’t think it’s wise to use -1 as a default.

r4zzz4k:
I've went with splitting configuration option in two as this was the most discussed option here. I've also added option to ignore data classes as these value bags are a kinda separate thing in my opinion. So one may want to both validate and not validate them to their liking.

I guess CONTRIBUTING.md should note that update of both both documentation and `default-detekt-config.yml` can be done via `generateDocumentation` Gradle task as this is not quite obvious. But of course `build` covers that, so I'm not sure if this is actually important for contributors.

Looking forward to your opinions on the PR.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
```diff
@@ -5,36 +5,72 @@ import io.gitlab.arturbosch.detekt.api.Debt
...
+import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtConstructor
+import org.jetbrains.kotlin.psi.KtFunction
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtParameterList
+import org.jetbrains.kotlin.psi.KtPrimaryConstructor
+import org.jetbrains.kotlin.psi.KtSecondaryConstructor
 
 /**
  * Reports functions which have more parameters than a certain threshold (default: 6).
```
schalkms:
```suggestion
 * Reports functions and constructors which have more parameters than a certain threshold.
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
```diff
@@ -5,36 +5,72 @@ import io.gitlab.arturbosch.detekt.api.Debt
...
+
+    override fun visitPrimaryConstructor(constructor: KtPrimaryConstructor) {
+        validateConstructor(constructor)
+    }
+
+    override fun visitSecondaryConstructor(constructor: KtSecondaryConstructor) {
+        validateConstructor(constructor)
+    }
+
+    private fun <T : KtConstructor<T>> validateConstructor(constructor: KtConstructor<T>) {
```
schalkms: I'm sorry, I don't understand why this function needs generics.
Does the following function declaration also suffice?

```suggestion
    private fun validateConstructor(constructor: KtFunction) {
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterListSpec.kt
```diff
@@ -17,7 +17,7 @@ class LongParameterListSpec : Spek({
             assertThat(subject.compileAndLint(code)).hasSize(1)
         }
 
-        it("does not reports short parameter list") {
+        it("does not report short parameter list") {
```
schalkms: good catch :-)

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterListSpec.kt
```diff
@@ -33,5 +33,39 @@ class LongParameterListSpec : Spek({
...
+            assertThat(subject.compileAndLint(code)).hasSize(1)
+        }
+
+        it("does not report short parameter list for primary constructors") {
+            val code = "class LongCtor(a: Int, b: Int, c: Int, d: Int, e: Int, f: Int)"
+            assertThat(subject.compileAndLint(code)).isEmpty()
+        }
+
+        it("reports too long parameter list for secondary constructors") {
+            val code = "class LongCtor() { constructor(a: Int, b: Int, c: Int, d: Int, e: Int, f: Int, g: Int) }"
```
schalkms: Detekt's CI chain checks for valid Kotlin code snippets in tests. These snippets are actually compiled during a test run in order to not have invalid code spread through the code base.
The marked code here is invalid. That's why the test is failing.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterListSpec.kt
```diff
@@ -33,5 +33,39 @@ class LongParameterListSpec : Spek({
...
+            assertThat(subject.compileAndLint(code)).hasSize(1)
+        }
+
+        it("does not report short parameter list for secondary constructors") {
+            val code = "class LongCtor() { constructor(a: Int, b: Int, c: Int, d: Int, e: Int, f: Int) }"
```
schalkms: same here

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterListSpec.kt
```diff
@@ -33,5 +33,39 @@ class LongParameterListSpec : Spek({
...
+            val config = TestConfig(mapOf(LongParameterList.CONSTRUCTOR_THRESHOLD to "2"))
+            val rule = LongParameterList(config)
+            val code = "class LongCtor(a: Int, b: Int, c: Int)"
+            assertThat(rule.compileAndLint(code)).hasSize(1)
+        }
+
+        it("does not report long parameter list for constructors of data classes if asked") {
+            val config = TestConfig(mapOf(LongParameterList.IGNORE_DATA_CLASSES to "true"))
+            val rule = LongParameterList(config)
+            val code = "data class LongCtor(a: Int, b: Int, c: Int, d: Int, e: Int, f: Int, g: Int)"
```
schalkms: same here
Data classes need a `var` or `val` prefix.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
r4zzz4k: I've just taken least common denominator of two types, which happens to have generic parameter. More generic KtFunction should work, will check.

r4zzz4k: `.getContainingClassOrObject()` is a method of `KtConstructor`, so it's not available on `KtFunction`, but if star-projected type is okay, that's more than enough for this.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterListSpec.kt
r4zzz4k: Thanks for the insight! I've found `-Pcompile-test-snippets` and fixed issues :)
Level of dependency on IDE in writing correct code is a bit frightening for me 😅

schalkms: Yeah. I have to add this to the `Contributing.md` guide as well.
I know it's a bit annoying. However, there were problems with invalid code snippets in tests, which tested the wrong thing. This is just a safety net in order to keep detekt's test long term stable.

schalkms:
@r4zzz4k your implemented rule already found a code smell in detekt itself 😉
The rest builds fine. Thank you. Good job! 💪

```
Ruleset: complexity - 20min debt
4443	LongParameterList - 7/7 - [DslGradleRunner] at /home/travis/build/arturbosch/detekt/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DslGradleRunner.kt:8:22
```

r4zzz4k:
I'm not hundred percent sure about defaults, so it may need some tuning in future. Hoping we would be able to gather some valuable feedback on this in future.

Anyway, great that my contribution is useful to someone besides me. Thank you :) And thanks for your hard work on the project, it really changes the way one works on their codebase in a very positive way!

arturbosch:
Please suppress the issue for now so we can move merging this awesome PR :)

r4zzz4k:
Should I roll back rebasing or this is okay?

schalkms:
@r4zzz4k this is okay. The failing AppVeyor build is not related to your change. Everything's fine.

PR #2419, opened 2020-03-08
Remove YAML Extension Requirement

### LINKED ISSUE #2417: file does not end with .yml!
Trying to load a `.yaml` file leads to an exception with the message from the issue title. [The official YAML FAQ recommends `.yaml` over `.yml`](https://yaml.org/faq.html) and I also believe that it is not up to detekt to prescribe what the file extension has to be. This choice should be up to the project. 😊

Fleshgrinder:
Closes #2417

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/YamlConfig.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/YamlConfig.kt
@@ -42,7 +42,7 @@ class YamlConfig internal constructor(
 
         /**
          * Factory method to load a yaml configuration. Given path must exist
-         * and point to a file.
+         * and point to a readable file.
          */
         fun load(path: Path): Config =
             load(path.toFile().apply {
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/YamlConfig.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/YamlConfig.kt
@@ -5,7 +5,6 @@ import io.gitlab.arturbosch.detekt.api.internal.validateConfig
 import org.yaml.snakeyaml.Yaml
 import java.io.BufferedReader
 import java.net.URL
-import java.nio.file.Files
 import java.nio.file.Path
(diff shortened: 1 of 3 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/YamlConfig.kt
```diff
@@ -41,14 +41,11 @@ class YamlConfig internal constructor(
...
 
-        private const val YAML = ".yml"
-
         /**
-         * Factory method to load a yaml configuration. Given path must exist and end with "yml".
+         * Factory method to load a yaml configuration. Given path must exist.
          */
         fun load(path: Path): Config {
             require(Files.exists(path)) { "Unable to load config from '$path': file does not exist!" }
-            require(path.toString().endsWith(YAML)) { "Invalid config file name $path: file does not end with $YAML!" }
```
schalkms: I'd rewrite it to

```kotlin
require(path.toString().endsWith(YAML || YML)) { "Invalid config file name $path: file does not end with $YAML!" }
```

We should apply a defensive programming style here.

Fleshgrinder: File extensions have no meaning by definition, we can have a perfectly valid YAML files with no extension `.detekt` or one that provides some defaults `detekt.yaml.dist` or … it really shouldn’t be up to the project to decide. Either parsing succeeds, or not, but the extension plays no role in it because parsing may fail despite having a `.yml` extension. In fact, just because it has a `.yml` extension does not mean that the content is YAML in the first place. 😊

schalkms: True, although it's not common to do so. The yaml parser handles invalid tokens in the yaml configuration and thus throws an exception.
I really like the provided sanity checks. 👍

Fleshgrinder:
Generating documentation failed for me because it only works with Java 8. You might want to consider adding a `.java-version` file to support http://www.jenv.be/ and automate the JVM selection for you and your users.

PR #2423, opened 2020-03-09
Add Groovy DSL example for jvmTarget to homepage

schalkms:
This complies with the source code in the README.

follow-up task to #2422


docs/pages/gettingstarted/groovydsl.md
```diff
@@ -118,10 +118,7 @@ Type resolution is experimental and works only for predefined `detektMain` and `
 custom detekt task with the `classpath` and `jvmTarget` properties present.
```
cortinico: Should this line be updated as well (as the snippet is missing the `classpath` specification?

PR #2427, opened 2020-03-09
Fail if `--baseline` file not found

### LINKED ISSUE #2374: Should fail if `--baseline` file does not found?
Hi,

When I'm using detekt via CLI, I noticed a behavior that is not very intuitive.

When I accidentally pass a non-existent file to the `--baseline` option, detekt never outputs about the invalid file and succeeds.

```shell
# NG: basline.xml
# OK: baseline.xml
$ detekt --baseline basline.xml
```

I wasted time until I noticed the filename was incorrect, so I think it more useful if detekt would fail fast when passing an invalid file path to the `--baseline` option.

For example:

```shell
$ detekt --baseline basline.xml
Invalid the file path `basline.xml` is specified to the `--baseline` option!
```

What do you think about this idea?

Thank you for reading.

BraisGabin:
Close #2374

Check that the file set by `--baseline` exists and is not a folder.

### DIFF
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/MainSpec.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/MainSpec.kt
@@ -112,7 +112,7 @@ class MainSpec : Spek({
             }
         }
 
-        it("succeed with --baseline if the path exists and is a file") {
+        it("succeeds with --baseline if the path exists and is a file") {
             val out = ByteArrayOutputStream()
             val err = ByteArrayOutputStream()
 
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/MainSpec.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/MainSpec.kt
@@ -65,7 +65,7 @@ class MainSpec : Spek({
             }
         }
 
-        it("succeed --create-baseline and --baseline") {
+        it("succeeds with --create-baseline and --baseline") {
             val out = ByteArrayOutputStream()
             val err = ByteArrayOutputStream()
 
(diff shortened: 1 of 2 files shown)


BraisGabin:
I just changed the test descriptions

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/MainSpec.kt
```diff
@@ -44,4 +47,82 @@ class MainSpec : Spek({
...
+
+                buildRunner(args, PrintStream(out), PrintStream(err))
+                Assertions.fail("This should throw an exception.")
+            } catch (_: HandledArgumentViolation) {
+                assertThat(String(err.toByteArray()).trim())
+                    .isEqualTo("The path specified by --baseline should be a file '$path'.")
+            }
+        }
+
+        it("succeed with --baseline if the path exists and is a file") {
```
schalkms:
```suggestion
        it("succeeds with --baseline if the path exists and is a file") {
```

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/MainSpec.kt
```diff
@@ -44,4 +47,82 @@ class MainSpec : Spek({
...
+
+                buildRunner(args, PrintStream(out), PrintStream(err))
+                Assertions.fail("This should throw an exception.")
+            } catch (_: HandledArgumentViolation) {
+                assertThat(String(err.toByteArray()).trim())
+                    .isEqualTo("Creating a baseline.xml requires the --baseline parameter to specify a path.")
+            }
+        }
+
+        it("succeed --create-baseline and --baseline") {
```
schalkms:
```suggestion
        it("succeeds with --create-baseline and --baseline") {
```

PR #2429, opened 2020-03-10
Add License rule

arturbosch:
Additional changes on top of #2077 :

I reused the current `FileProcessListener` extension to store the license header inside the KtFile's which is basically caching :).
Instead of introducing a `Config.Location` property on the config, I went with introducing a new method for the `Extension` interface.
```kt
interface Extension {
    ....
    fun init(config: Config) {
        // implement for setup code
    }
    @OptIn(UnstableApi::class)
    fun init(context: SetupContext) {
        // implement for setup code
    }
}
``` 

I've marked the new `SetupContext` as Unstable with the new `@RequiresOptIn` api. This way we can add some more useful methods and properties later on without needing a major version.
When we decide that `SetupContext` is stable, we can remove the `fun init(config: Config)` function from the `Extension` interface as `SetupContext` provides the `Config`.
```kt
/**
 * Context providing useful processing settings to initialize extensions.
 */
@UnstableApi
interface SetupContext {
    /**
     * All config locations which where used to create [config].
     */
    val configUris: Collection<URI>

    /**
     * Configuration which is used to setup detekt.
     */
    val config: Config
}
```

`ProcessingSettings` was the ideal candidate to implement `SetupContext` as it has all the useful information from the cli.


dector:
Thanks, @arturbosch for reopening this! Should I add more tests?

arturbosch:
@dector that sounds awesome to be!
Maybe also some real world testing on your project if this rule does what you expect it to. Thx!

dector:
@arturbosch Battle-tested this rule on Mozilla Fenix app with the default template file. Works like a charm and it's so freaking fast! :heart_eyes:

arturbosch:
Thanks for testing! If you realize a testcase is missing, feel free to open a new PR.

dector:
:tada: thanks @arturbosch

PR #2432, opened 2020-03-15
Deprecate BaseRule and state that it will be make sealed - #2365

arturbosch:


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
```diff
@@ -20,60 +17,9 @@ typealias RuleId = String
...
-     * Could be overridden by subclasses to specify a behaviour which should be done after
-     * visiting kotlin elements.
-     */
-    protected open fun postVisit(root: KtFile) {
-        // nothing to do by default
-    }
-}
+@Deprecated("""
+Do not use this class directly. Use Rule or MultiRule instead.
+This class was introduced to support a common handling of the rule types mentioned.
```
schalkms:
```suggestion
This class was introduced to support a common handling of the mentioned rule types.
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/BaseRule.kt
```diff
@@ -20,60 +17,9 @@ typealias RuleId = String
...
+This class will be made sealed in a different release and you won't be able to derive from it.
```
schalkms:
```suggestion
This class will be made sealed in the next release and you won't be able to derive from it. 
```

schalkms:
Closes #2365 ?

BraisGabin:
If you move base rule to an internal package right now you will need to move it outside for 2.0 because all the subclasses of a sealed class should be in the same file.

So I think that it's better to keep baserule in the same place with the depreciation. As soon as we make it sealed it will not be able to be extended so the package will not matter.

arturbosch:
@BraisGabin @schalkms  true that. However if not using a `typealias` I need to suppress every usage of the `BaseRule` for our `-PwarningsAsError=true` strategy ...
And it must be top level so the `import` is also suppressed.
As we are now deprecating more and more stuff should I go further with the typealias + moving strategy or deactivate `warningsAsErrors`. What do you think?

BraisGabin:
OK, I understand the reason. I'm a bit concern about the binary compatibility but previous changes broke it already so it doesn't matter now. 👍🏻 For me.

arturbosch:
> OK, I understand the reason. I'm a bit concern about the binary compatibility but previous changes broke it already so it doesn't matter now. 👍🏻 For me.

Could you elaborate on what is not binary compatible anymore for 1.7?
Imo moving the classes while leaving a typealias behind is binary compatible, isn't it?

BraisGabin:
I can be wrong here. I'm not a compatibility expert or nothing near. I think that the `typealias` are just syntax sugar. But I repeat that I don't know this for sure.

PR #2434, opened 2020-03-15
Prepare 1.7.0-beta2 release

arturbosch:
Second round testing of the gradle plugin performance improvements


BraisGabin:
Are there new performance changes between beta1 and beta2? I tested beta1 in some projects and it works like a charm :)

arturbosch:
> Are there new performance changes between beta1 and beta2? I tested beta1 in some projects and it works like a charm :)

Nope, no new performance improvements just making sure we have a stable release for 1.7 :)

PR #2436, opened 2020-03-16
SwallowedException: fix false negative when using variables

### LINKED ISSUE #2049: SwallowedException false negative
## Expected Behavior
The following code block should be considered non-compliant:

    try {
        var message = e.message  // e is swallowed
        throw IllegalArgumentException(message)
    } catch (f: Exception) {
        var message = f.toString()  // f is swallowed
        throw Exception(IllegalArgumentException(message))
    }

## Observed Behavior
Code passes the SwalowedException check

## Steps to Reproduce
Use the sample code provided, or similar where exception message is stored and passed into new exception.

## Context
Trying to prevent instances of swallowed exceptions

## Your Environment

* Version of detekt used: 1.1.x
* Operating System and version: N/A

t-kameyama:
Fixes #2049

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt
@@ -11,14 +11,18 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.SplitPattern
 import io.gitlab.arturbosch.detekt.rules.ALLOWED_EXCEPTION_NAME
 import io.gitlab.arturbosch.detekt.rules.isAllowedExceptionName
+import org.jetbrains.kotlin.psi.KtBlockExpression
 import org.jetbrains.kotlin.psi.KtCatchClause
 import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
+import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtNameReferenceExpression
 import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.KtThrowExpression
 import org.jetbrains.kotlin.psi.psiUtil.anyDescendantOfType
 import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
+import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
 import org.jetbrains.kotlin.psi.psiUtil.getQualifiedExpressionForReceiverOrThis
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 
 /**
  * Exceptions should not be swallowed. This rule reports all instances where exceptions are `caught` and not correctly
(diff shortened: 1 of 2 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedExceptionSpec.kt
```diff
@@ -37,6 +37,22 @@ class SwallowedExceptionSpec : Spek({
             assertThat(subject.compileAndLint(code)).hasSize(2)
         }
 
+        it("reports swallowed exceptions only using exception strings via variables") {
```
schalkms: A test case that asserts conditional code could be beneficial.
e.g.
```Kotlin
catch (e: IllegalStateException) {
  if (...) {
    val message = e.message
    throw IllegalArgumentException(message)
  }
} 
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt
```diff
@@ -91,26 +94,36 @@ class SwallowedException(config: Config = Config.empty) : Rule(config) {
...
         val parameterName = catchClause.catchParameter?.name
+        val parameterNameReferencesInVariables = catchClause.catchBody?.children?.mapNotNull { child ->
+            val variable = child as? KtProperty ?: return@mapNotNull null
+            val variableName = variable.name ?: return@mapNotNull null
+            val parameterReference = when (val initializer = variable.initializer) {
+                is KtNameReferenceExpression -> initializer.takeIf { it.text == parameterName }
+                is KtDotQualifiedExpression -> initializer.receiverExpression.takeIf { it.text == parameterName }
+                else -> null
+            } ?: return@mapNotNull null
+            variableName to parameterReference
```
schalkms: This would be easier to maintain and understand if we refactor this into a separate method.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt
```diff
@@ -91,26 +94,36 @@ class SwallowedException(config: Config = Config.empty) : Rule(config) {
...
+        }.orEmpty().toMap()
```
schalkms: Could we return an `emptyMap()` instead of null pointer? Then we could spare the `orEmpty()` call.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt
```diff
@@ -91,26 +94,36 @@ class SwallowedException(config: Config = Config.empty) : Rule(config) {
...
+                        val referenceText = reference.text
+                        if (referenceText == parameterName) {
+                            reference.getQualifiedExpressionForReceiverOrThis()
+                        } else {
+                            parameterNameReferencesInVariables[referenceText]?.getQualifiedExpressionForReceiverOrThis()
+                        }
+                    }
+                    .orEmpty()
+               return parameterNameReferences.isNotEmpty() &&
+                       parameterNameReferences.all { it is KtDotQualifiedExpression }
```
schalkms: This would be easier to maintain and understand if we refactor this into a separate method.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedExceptionSpec.kt
```diff
@@ -37,6 +37,42 @@ class SwallowedExceptionSpec : Spek({
...
+
+        it("reports swallowed exceptions only using exception strings via variables in 'if' block") {
+            val code = """
+                fun f() {
+                    try {
+                    } catch (e: IllegalStateException) {
+                        if (true) {
+                            val message = e.message
+                            throw IllegalArgumentException(message)
+                        }
```
schalkms: What happens when the user has a construct like this? Is it reported? The exception only gets swallowed when the `condition` is true in that case.

```kotlin
catch (f: Exception) {
  if (condition) {
    val message = f.toString()
    throw Exception(IllegalArgumentException(message))
  }
  throw IllegalStateException(f)
}
```

schalkms: An additional test for this case might be worthwhile.

PR #2448, opened 2020-03-18
Simplify UnnecessaryApplySpec test cases

schalkms:
* Fixes invalid Kotlin code in test snippets
* Use compileAndLint() instead of lint()

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
@@ -2,7 +2,7 @@ package io.gitlab.arturbosch.detekt.rules.style
 
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.test.KtTestCompiler
-import io.gitlab.arturbosch.detekt.test.lint
+import io.gitlab.arturbosch.detekt.test.compileAndLint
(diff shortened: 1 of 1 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
```diff
@@ -46,33 +46,34 @@ class UnnecessaryApplySpec : Spek({
...
                         }
                     }
                 """)).hasSize(1)
             }
 
             it("does not report an apply with lambda block") {
-                assertThat(subject.lint("""
+                assertThat(subject.compileAndLint("""
                     fun f() {
                         val a : Int? = null
```
BraisGabin:
```suggestion
                        val a: Int? = null
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
```diff
@@ -46,33 +46,34 @@ class UnnecessaryApplySpec : Spek({
...
             it("does report single statement in apply used as function argument") {
-                assertThat(subject.lint("""
-                    fun b(i : Int?) {
-                    }
-                    fun f() {
+                assertThat(subject.compileAndLint("""
+                    fun b(i : Int?) {}
+
+                    fun main() {
                         val a : Int? = null
```
BraisGabin:
```suggestion
                        val a: Int? = null
```

PR #2454, opened 2020-03-18
Refactor FindingsReport + FileBasedFindingsReport

schalkms:


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Debt.kt
```diff
@@ -11,13 +11,36 @@ data class Debt(val days: Int = 0, val hours: Int = 0, val mins: Int = 0) {
...
+    operator fun plus(other: Debt): Debt {
+        return Debt(
+            this.days + other.days,
+            this.hours + other.hours,
+            this.mins + other.mins
+        )
+    }
+
+    /** Formats the debt time to a human readable format. */
+    fun formatDebtTime(): Debt {
```
arturbosch: Shouldn't the overflow validation be moved to the `toString()` method or the function should be renamed to `format` and return just a `String`?

schalkms: I wouldn't want to add the validation to the constructor.
The values would have to be mutable, which I wanted to avoid.
Maybe we can come up with a better name than `formatDebtTime` for this function.

arturbosch: Maybe we move this function to the companion object and call it `Debt.recalculateOverflow(instance)`?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/console/ReportFormatting.kt
```diff
@@ -1,5 +1,28 @@
...
+                .reduce { acc, d -> acc + d }
+            debtList.add(debt)
+            append("$key - $debt debt".format())
+            val issuesString = issues.joinToString("") {
+                it.compact().format("\t")
+            }
+            append(issuesString.yellow())
+        }
+        val overallDebt = debtList.reduce { acc, d -> acc + d }
+        append("Overall debt: $overallDebt".format("\n"))
```
arturbosch: I think the `overallDebt.formatDebtTime` (`Debt.recalculateOveflow(oveallDebt)`) call is missing here?

schalkms: This is a bug. Thanks! I shall add a test for it.
 The `toString()` method should call the `formatDebtTime()` function here.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Debt.kt
BraisGabin: I think that the `plus` operator should handle this problem. It's his responsability to create valid `Debt` instances. 

If we want to do something in the constructor is to check that `Debt` is valid. And if it's not, throw an exception. But I think that it's not worth it.

schalkms:
> If we want to do something in the constructor is to check that Debt is valid. And if it's not, throw an exception. But I think that it's not worth it.

Agreed.

schalkms:
I rebased the branch and incorporated your feedback in the last commit. This PR is now ready to merge.

PR #2459, opened 2020-03-18
Fix toString() output for Whitelist class

schalkms:

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/baseline/Whitelist.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/baseline/Whitelist.kt
@@ -5,6 +5,6 @@ data class Whitelist(
 ) : Listing<Whitelist> {
 
     override fun toString(): String {
-        return "Blacklist(ids=$ids)"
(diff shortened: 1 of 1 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/baseline/Whitelist.kt
```diff
@@ -5,6 +5,6 @@ data class Whitelist(
 ) : Listing<Whitelist> {
 
     override fun toString(): String {
-        return "Blacklist(ids=$ids)"
+        return "Whitelist(ids=$ids)"
     }
```
BraisGabin: This is a data class, can't we use the default `toString`? If the answer is no, we should add a simple test here to ensure that no one removes it in the future.

arturbosch: Good point
```kt
data class Whitelist(
    override val ids: Set<String>
) : Listing<Whitelist>

fun main() {
    println(Whitelist(setOf("1", "2")))
}
```
It actually prints the same :)

schalkms:
see #2467

PR #2462, opened 2020-03-18
Add test for BaselineHandler

schalkms:
The BaselineHandler xml parser should handle empty ID entries.

### DIFF
--- a/detekt-cli/src/test/resources/smell-baseline.xml
+++ b/detekt-cli/src/test/resources/smell-baseline.xml
@@ -6,5 +6,6 @@
     </Blacklist>
     <Whitelist>
         <ID>FeatureEnvy:Signature</ID>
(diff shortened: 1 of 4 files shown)


detekt-cli/src/test/resources/smell-baseline.xml
```diff
@@ -6,5 +6,6 @@
     </Blacklist>
     <Whitelist>
         <ID>FeatureEnvy:Signature</ID>
+        <ID></ID>
```
BraisGabin: It does, but should it? Allowing this seems like a way to allow noise in our baselines.

arturbosch: Agree it does not add any value, can we throw an `error("")` for this?

PR #2465, opened 2020-03-19
Validate boolean config values - Closes #2045

### LINKED ISSUE #2045: Invalid boolean config values are evaluated silently to false
## Expected Behavior
I can use any number of configuration files.

## Observed Behavior
If I use 3 or more configuration files, detekt always throws an error when completing a task.

## Steps to Reproduce
I use my task:
```kotlin
task<Detekt>("verifyQuality") {
    source = files(subprojects.withPlugin("kotlin").srcDirs("main")).asFileTree
    val configPath = "buildSrc/src/main/resources/detekt/config"
    config.setFrom(
        "$configPath/common.yml",
        "$configPath/comments.yml",
        "$configPath/potential_bugs.yml",
        "$configPath/performance.yml"
    )
    reports {
        html {
            enabled = true
            destination = File(analysisQualityHtmlPath)
        }
        xml.enabled = false
        txt.enabled = false
    }
}
```
And the service does not note errors in the code, but prints it to the log:
https://gist.github.com/kepocnhh/558c85a50419df1a1db400eb54786d96

## Context
This error occurs only if the configuration files are 3 or more. That is, in all other cases, the detekt works:
```
config.setFrom()
```
or
```
config.setFrom("$configPath/common.yml")
```
or
```
config.setFrom(
    "$configPath/common.yml",
    "$configPath/comments.yml"
)
```
or
```
config.setFrom(
    "$configPath/common.yml",
    "$configPath/performance.yml"
)
```
or
```
config.setFrom(
    "$configPath/potential_bugs.yml",
    "$configPath/performance.yml"
)
```
or
```
config.setFrom("$configPath/potential_bugs.yml")
```

---

common.yml is
```yaml
build:
  maxIssues: 0

processors:
  active: false
```

the rest are like that
```yaml
performance:
  active: true
  ArrayPrimitive:
    active: true
```

❕ Each configuration file is unique.

## Your Environment
java 1.8.0_221
kotlin 1.3.50
detekt 1.1.1
Gradle 5.5.1
macOS Mojave Version 10.14.6 (18G87)

arturbosch:


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/BaseConfig.kt
```diff
@@ -37,9 +37,18 @@ abstract class BaseConfig : HierarchicalConfig {
...
+            } else {
+                throw ClassCastException()
+            }
         is Double -> result.toDouble()
         is String -> result
         else -> throw ClassCastException()
     }
+
+    companion object {
+        val ALLOWED_BOOL_VALUES = setOf("true", "TRUE", "false", "FALSE")
```
arturbosch: Should we constraint this just to `true` and `false`?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/BaseConfig.kt
```diff
@@ -37,9 +37,18 @@ abstract class BaseConfig : HierarchicalConfig {
 
     protected open fun tryParseBasedOnDefault(result: String, defaultResult: Any): Any = when (defaultResult) {
         is Int -> result.toInt()
-        is Boolean -> result.toBoolean()
+        is Boolean ->
+            if (result in ALLOWED_BOOL_VALUES) {
+                result.toBoolean()
+            } else {
+                throw ClassCastException()
```
BraisGabin:
```suggestion
                throw ClassCastException("$result can not be cast to Boolean")
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/BaseConfig.kt
BraisGabin: I think so. That or support all the valid [yaml values](https://yaml.org/type/bool.html): `y|Y|yes|Yes|YES|n|N|no|No|NO|true|True|TRUE|false|False|FALSE|on|On|ON|off|Off|OFF`

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/BaseConfig.kt
arturbosch: The `ClassCastException` is caught upstream and converted to the meaningful message as seen in the testcase.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/BaseConfig.kt
arturbosch: I've restricted them to true or false.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/BaseConfig.kt
```diff
@@ -37,9 +37,18 @@ abstract class BaseConfig : HierarchicalConfig {
...
+            }
         is Double -> result.toDouble()
         is String -> result
         else -> throw ClassCastException()
     }
+
+    companion object {
+        val ALLOWED_BOOL_VALUES = setOf("true", "false")
+    }
 }
```
BraisGabin:
```suggestion
}

private val ALLOWED_BOOL_VALUES = setOf("true", "false")
```

What do you think? I realised about this when I saw this implementation detail in the documentation.

PR #2475, opened 2020-03-20
Check primary ctor in UndocumentedPublicProperty

### LINKED ISSUE #2468: UndocumentedPublicProperty doesn't check primary constructor
## Expected Behavior
If a property is declared in the primary constructor (especially in a data class), the static analysis should count an issue if the property isn't documented. 

## Observed Behavior
Only properties declared in the class body are flagged when left public and undocumented.

## Steps to Reproduce
Create a class with properties declared in the primary constructor but don't  add `@property` or KDoc reference in a comment

## Your Environment

* Operating System and version: OSX

schalkms:
Closes #2468

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
@@ -34,6 +34,11 @@ class UndocumentedPublicPropertySpec : Spek({
             assertThat(subject.compileAndLint(code)).hasSize(1)
         }
 
+        it("reports undocumented public property in a primary constructor") {
+            val code = "/* comment */ class Test(val a: Int)"
+            assertThat(subject.compileAndLint(code)).hasSize(1)
+        }
+
         it("does not report documented public property") {
             val code = """
                 /**
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
@@ -29,6 +29,11 @@ class UndocumentedPublicPropertySpec : Spek({
             assertThat(subject.compileAndLint(code)).hasSize(2)
         }
 
+        it("reports undocumented public properties in a primary constructor") {
+            val code = "class Test(val a: Int)"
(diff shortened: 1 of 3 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
```diff
@@ -90,13 +100,34 @@ class UndocumentedPublicPropertySpec : Spek({
...
+        it("does not report undocumented public properties in a primary constructor for an internal class") {
+            val code = "internal class Test(val a: Int)"
+            assertThat(subject.compileAndLint(code)).isEmpty()
+        }
+
+        it("does not report documented public properties in a primary constructor") {
+            val code = """
+                /**
+                * @a int1
+                * [b] int2
```
BraisGabin: This doesn't seem like kdoc: https://kotlinlang.org/docs/reference/kotlin-doc.html 
Shouldn't it be `@property a`?

schalkms: This syntax is used for linking. Please take a look at the section on the same page.
https://kotlinlang.org/docs/reference/kotlin-doc.html#linking-to-elements

schalkms: The sample on the top of the page also shows how to use the linking syntax.
https://kotlinlang.org/docs/reference/kotlin-doc.html#kdoc-syntax

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
```diff
@@ -90,13 +100,34 @@ class UndocumentedPublicPropertySpec : Spek({
...
-                    constructor(a: Int, b: Int) : this(a)
+                class Test() {
+                    constructor(a: Int) : this()
                 }
             """
             assertThat(subject.compileAndLint(code)).isEmpty()
         }
+
+        it("does not report undocumented non-public properties in a primary constructor") {
+            val code = "class Test(internal val a: Int)"
```
t-kameyama: Please consider following case. I think it should not be reported. 
```
class Test(a: Int)
```

schalkms: Thanks for your review and for your feedback.
I added a test case for this snippet.

Issue #2217, opened 2019-12-31
"Property is misspelled or does not exist" error for new rules in default rulesets


smyachenkov:
I am getting error "Property is misspelled or does not exist" for build of pull request with the new rule https://github.com/arturbosch/detekt/pull/2215

## Expected Behavior
CI build completed successfully.

## Observed Behavior
```
> Task :detekt-gradle-plugin:detekt FAILED
io.gitlab.arturbosch.detekt.cli.InvalidConfig: Run failed with 1 invalid config property.
Property 'style>ExplicitCollectionElementAccessMethod' is misspelled or does not exist.
```

## Steps to Reproduce
Create new rule.
Add it to default ruleset provider, for example StyleGuideProvider
Run CI build with this code.

## Context
Noticed on the Travis builds of PR with the new rule https://github.com/arturbosch/detekt/pull/2215
https://ci.appveyor.com/project/arturbosch/detekt/builds/29825544

## Your Environment
https://github.com/arturbosch/detekt/pull/2215
https://ci.appveyor.com/project/arturbosch/detekt/builds/29825544

smyachenkov:
Is it related to issue https://github.com/arturbosch/detekt/issues/2168?

schalkms:
Yes, it’s related.

PR #2478, opened 2020-03-21
Extend ComplexInterface to support exclusion of private member/functions

### LINKED ISSUE #2446: ComplexInterface - should it also count private methods?
`ComplexInterface` is triggered even if the number of overrideable functions / properties is less than treshold, but there are private methods within the interface (used by default implementation of the functions), which trigger the rule.

Should it be the case? Instinctively, I would not think they would be counted, but maybe I am wrong..

cortinico:
Following up discussion on #2446, currently `ComplexInterface` is considering also `private` members/functions. I'm adding a `includePrivateDeclarations` config field (default false) to make this behavior configurable.

Fixes #2446

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexInterface.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexInterface.kt
@@ -9,6 +9,7 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.ThresholdRule
 import io.gitlab.arturbosch.detekt.api.ThresholdedCodeSmell
 import io.gitlab.arturbosch.detekt.rules.companionObject
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.psi.KtClass
 import org.jetbrains.kotlin.psi.KtClassBody
 import org.jetbrains.kotlin.psi.KtNamedFunction
@@ -68,15 +69,16 @@ class ComplexInterface(
         return if (body != null) calculateMembers(body) else 0
     }
 
-    private fun calculateMembers(body: KtClassBody) = body.children
-        .filter {
-            if (includePrivateDeclarations) {
-                true
-            } else {
-                it is KtTypeParameterListOwner && !it.isPrivate()
-            }
-        }
-        .count { it is KtNamedFunction || it is KtProperty }
+    private fun calculateMembers(body: KtClassBody): Int {
+        fun PsiElement.considerPrivate() = includePrivateDeclarations ||
+                this is KtTypeParameterListOwner && !this.isPrivate()
+
+        fun PsiElement.isMember() = this is KtNamedFunction || this is KtProperty
+
+        return body.children
+            .filter(PsiElement::considerPrivate)
+            .count(PsiElement::isMember)
+    }
 
     companion object {
         const val INCLUDE_STATIC_DECLARATIONS = "includeStaticDeclarations"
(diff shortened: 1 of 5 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexInterfaceSpec.kt
```diff
@@ -72,7 +73,47 @@ class ComplexInterfaceSpec : Spek({
             }
 
             it("reports complex interface with includeStaticDeclarations config") {
-                val rule = ComplexInterface(config, threshold = THRESHOLD)
+                val rule = ComplexInterface(staticDeclarationsConfig, threshold = THRESHOLD)
+                assertThat(rule.compileAndLint(code)).hasSize(1)
+            }
+        }
+
+        context("private function") {
```
schalkms:
```suggestion
        context("private functions") {
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexInterfaceSpec.kt
```diff
@@ -72,7 +73,47 @@ class ComplexInterfaceSpec : Spek({
...
+                assertThat(rule.compileAndLint(code)).hasSize(1)
+            }
+        }
+
+        context("private member") {
+            val code = """
+                interface I {
+                    fun f1()
+                    fun f2()
+                    private val i1: Int
```
schalkms: This declaration (`private val i1: Int`) is not valid in an interface.

javax.script.ScriptException: Error: error: abstract property in an interface cannot be private

cortinico: Right my bad, I've added a getter to the private member so this should compile now.

cortinico:
For some reason Travis hasn't run? 
Moreover AppVeyor is failing on `:detekt-gradle-plugin:detekt` because of failed validation of config file.

This is what it looks like locally (`./gradlew --project-dir detekt-gradle-plugin detekt`):

```
> Task :detekt FAILED
Property 'complexity>ComplexInterface>includePrivateDeclarations' is misspelled or does not exist.

* What went wrong:
```

Any hints? Seems related to #2217 ?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexInterface.kt
```diff
@@ -62,10 +68,19 @@ class ComplexInterface(
         return if (body != null) calculateMembers(body) else 0
     }
 
-    private fun calculateMembers(body: KtClassBody) = body.children.count { it is KtNamedFunction || it is KtProperty }
+    private fun calculateMembers(body: KtClassBody) = body.children
+        .filter {
+            if (includePrivateDeclarations) {
```
arturbosch: Consider `includeXXX || elseExpr`.
Or maybe

```    private fun calculateMembers(body: KtClassBody): Int {
        fun PsiElement.considerPrivate(): Boolean = includePrivateDeclarations ||
            this is KtTypeParameterListOwner && !this.isPrivate()
        return body.children
            .filter { it.considerPrivate() }
            .count { it is KtNamedFunction || it is KtProperty }
    }
```

arturbosch:
Try adding a  `config>excludes` entry for now: https://github.com/arturbosch/detekt/pull/2201/files
We should document this in our contribution guide :(.
Maybe even come up with an automatic way to generate them in our `generator` module to not confuse contributors ...

PR #2485, opened 2020-03-22
Use PrinterStream in AstPrinter

BraisGabin:
This is the first of a serie of PRs related with how we manage the output.

The main ideas:


BraisGabin:
The Mac images ar so flaky >_<

PR #2487, opened 2020-03-22
Simplify tests with StringPrintStream

BraisGabin:


detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/StringPrintStream.kt
```diff
@@ -0,0 +1,15 @@
+package io.gitlab.arturbosch.detekt.test
+
+import java.io.ByteArrayOutputStream
+import java.io.PrintStream
+
+class StringPrintStream private constructor(
+    private val stream: ByteArrayOutputStream
+) : PrintStream(stream) {
```
arturbosch: All usages should probably close this stream at the end.
Lets implement AutoCloseable and close the `stream` here ?

BraisGabin:
We don't really need to close this one. It doesn't hold any resource, it's all in memory.

PR #2489, opened 2020-03-22
ProcessingSettings

BraisGabin:
This is the third of a serie of PRs related with how we manage the output.

The main ideas:


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/ProcessingSettings.kt
```diff
@@ -40,48 +39,12 @@ class ProcessingSettings @JvmOverloads constructor(
...
+    val errorPrinter: PrintStream,
     val autoCorrect: Boolean = false,
     val debug: Boolean = false,
     override val configUris: Collection<URI> = emptyList()
 ) : AutoCloseable, Closeable, SetupContext {
-    /**
-     * Single project input path constructor.
-     */
-    constructor(
-        inputPath: Path,
```
arturbosch: :+1:

Issue #2009, opened 2019-10-10
excludes patterns can match directories above the included root


mww-synopsys:
Imagine that I have a project containing a MagicNumber issue in a file at the root of my project. Say the path looks like MyProject/MagicNumber.kt. On the Detekt command line, I would scan that project by providing MyProject as the input; if I were to place that project at the root of the file system, this would look like

```
java -jar detekt-cli-<version>-all.jar --input /MyProject
```

In the above circumstance, Detekt would find the defect.

Now imagine that I happened to place the whole project in a directory called test, so the invocation looked like this:

```
java -jar detekt-cli-<version>-all.jar --input /test/MyProject
```

In this circumstance, the MagicNumber defect would not be reported. Why? The default configuration's rule for excluding test directories from being scanned for MagicNumber matches the test directory above the indicated input directory. That is, exclude patterns can match directories above the project roots as described by --input options.

I think that exclude patterns are generally used to describe paths to be excluded within the project, not those that may happen to be on a path above the project. This can cause misleading behavior if, for instance, you are testing different analysis tools to see how they perform and happen to use Detekt to scan a project you have placed under a test directory.

schalkms:
Closed by 2caede1

PR #2492, opened 2020-03-22
Relativize paths based on input path or working dir - #2009

arturbosch:


chao2zhang:
This could be closed by #3319

Issue #2484, opened 2020-03-22
Errors running detekt in the detekt project


BraisGabin:
Letely when I use `./gradlew detekt` in the detekt project itself I'm getting some flaky behaviour. Sometimes it works (it finds the issues if there are any) and sometimes fails with different errors:

```
Ignoring listeners of task graph ready event, as this build (:detekt-gradle-plugin) has already executed work.
> Task :detekt-api:detekt FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':detekt-api:detekt'.
> JAR entry default-detekt-config.yml not found in /Users/brais/projects/detekt/detekt-cli/build/libs/detekt-cli-1.7.0.jar

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':detekt-api:detekt'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:205)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:263)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:203)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:184)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:114)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:62)
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
        at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:56)
        at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:36)
[...]

PR #2493, opened 2020-03-22
Make the gradle plugin a sub project again

arturbosch:
With the current setup adding a new feature like in #2492 is cumbersome.
Until a new cli version is published this feature cannot be merged or it will fail the build.
Our current approach is good for having the newest gradle plugin changes in a `gradle detekt` run. However it slows down build time and development time.
Fast feedback can still be provided with the new `pluginManagement` closure:
```gradle
pluginManagement {
    repositories {
        maven { setUrl("https://plugins.gradle.org/m2/") }
        mavenLocal() // used to publish and test local gradle plugin changes
    }
...
}
```
All it needs is to run `publishToMavenLocal`.

I propose to change it back to a sub project. To verify newest changes we could move to following commands in our CI:
```shell script
gradle build -x detekt
gradle publishToMavenLocal
gradle detekt
```

This PR leads to:
- faster build file evaluate, no need to wait for the composite build file evaluation
- cleaner build files, less duplications
- more up-to-date tasks
- faster builds, clean build is ~30 seconds faster

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -80,6 +80,13 @@ Following warning is expected until [Jekyll](https://github.com/jekyll/jekyll/is
 
 `warning: Using the last argument as keyword parameters is deprecated (Ruby 2.7.0)`
 
+### When working on the Gradle plugin ...
+
+- Make changes to the core modules (e.g. adding a new CLI flag)
+- Run `gradle publishToMavenLocal`
+- Make changes to the Gradle plugin and add tests
+- Verify with `gradle detekt`
+
 ### Release checklist
 
 - add changes in CHANGELOG.md -> `groovy github-milestone-report.groovy arturbosch detekt [milestone-number]`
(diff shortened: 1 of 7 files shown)


BraisGabin:
Would this help with #2484?

arturbosch:
> Would this help with #2484?

Yes, this should not happen as we use a stable detekt version for self analysis.

schalkms:
@arturbosch did the test coverage noticeably go down after this commit or after the following one?

```
Setup for easy local gradle plugin changes
```

arturbosch:
@schalkms seems it was this PR -> https://codecov.io/gh/arturbosch/detekt/pull/2493/changes

The gradle plugin is now a sub project and provides coverage data. So it dropped by 3.14% :(

schalkms:
@arturbosch I also noticed this when I had a free minute. The only thing we can do is to provide separate unit tests for this sub project, I guess. Before, when it wasn't a sub-project, the analyzer went through the paths.

PR #2494, opened 2020-03-22
Try fix windows test

arturbosch:


BraisGabin:
I saw where the problem is. Can we wait to merge this a bit? I'm fixing it.

BraisGabin:
https://github.com/arturbosch/detekt/pull/2497 should fix this problem. The problem was not in the test, it was in the production code. I think.

PR #2495, opened 2020-03-22
Don't use System.out in the ProgressListeners

BraisGabin:
~This PR is not ready because I have a problem here. The `DetektProgessListener` is loaded by the ServiceLoader so I can't add the `PrintStream` as a constructor parameter. I need to get it later using `init(Config)` or `init(SetupContext)`. But neigther of them has that reference.~

~Should I add it to any of them? Or should I create a new `init` that provides both `PrintStream`?~

----

This PR is part of a serie of PRs related with how we manage the output.

The main ideas:


arturbosch:
Lets add `out` and `err` to the `SetupContext`. Maybe even move the `info, debug, error` from `ProcessingSettings` functions to it.

BraisGabin:
What do you think about `info`, `debug` and `error` as extension functions of `SetupContext`?

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/DetektProgressListener.kt
```diff
@@ -1,16 +1,37 @@
 package io.gitlab.arturbosch.detekt.generator
 
 import io.gitlab.arturbosch.detekt.api.FileProcessListener
+import io.gitlab.arturbosch.detekt.api.SetupContext
+import io.gitlab.arturbosch.detekt.api.UnstableApi
 import org.jetbrains.kotlin.psi.KtFile
+import java.io.PrintStream
 
 class DetektProgressListener : FileProcessListener {
```
arturbosch: Can we remove this copy and just use the real progress listener from cli?

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/Runner.kt
```diff
@@ -13,14 +13,14 @@ class Runner(
     private val outPrinter: PrintStream,
     private val errPrinter: PrintStream
 ) {
-    private val listeners = listOf(DetektProgressListener())
+    private val listeners = listOf(DetektProgressListener(outPrinter))
```
arturbosch: Lets remove this listener. The generator output is no realy useful.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/DetektProgressListener.kt
arturbosch: Hm, lets remove this whole class alltogether as the dots are not really useful for the generator :)

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/SetupContext.kt
```diff
@@ -16,4 +17,14 @@ interface SetupContext {
      * Configuration which is used to setup detekt.
      */
     val config: Config
+
+    /**
+     * The channel to log all the output
```
arturbosch:
```suggestion
     * The channel to log all the output.
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/SetupContext.kt
```diff
@@ -16,4 +17,14 @@ interface SetupContext {
...
+     */
+    val outPrinter: PrintStream
+
+    /**
+     * The channel to lod all the errors
```
arturbosch:
```suggestion
     * The channel to log all the errors.
```

PR #2496, opened 2020-03-22
Remove default print stream

BraisGabin:
This PR is part of a serie of PRs related with how we manage the output.

The main ideas:


detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/RunnerSpec.kt
```diff
@@ -119,7 +122,7 @@ class RunnerSpec : Spek({
                 "--config-resource", "/configs/max-issues-0.yml"
             )
 
-            Runner(cliArgs).execute()
+            Runner(cliArgs, NullPrintStream(), NullPrintStream()).execute()
```
arturbosch: Should we also create a helper for this?

BraisGabin: Here we are testing `Runner` itself. I prefer to call the real constructor in this case.

PR #2500, opened 2020-03-23
UnusedPrivateMember: fix false negative when file/class has same named functions

### LINKED ISSUE #1981: UnusedPrivateMember doesn't report the correct warning count
## Expected Behavior
```kotlin
class Test {
    private fun unusedFunction(): Int {
        return 5
    }
}

class Test2 {
    private fun unusedFunction(): Int {
        return 5
    }
}
```

In that code I expect to have two warnings because there are two unused functions

## Observed Behavior
Only one warning is raised. The reason is becase the functions are called the same.

## Steps to Reproduce
You can check this comparison: https://github.com/arturbosch/detekt/compare/master...BraisGabin:unused-private-function

## Context
I found this issue fixing a similar issue with parameters that are called the same in different functions. That issue was easier to fix that this one.

## Your Environment
* Operating System and version: Mac

t-kameyama:
Fixes #1981

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
@@ -84,15 +84,17 @@ private class UnusedFunctionVisitor(
     override fun getUnusedReports(issue: Issue): List<CodeSmell> {
         return functionDeclarations.flatMap { (name, functions) ->
             val references = functionReferences[name].orEmpty()
-            val unusedFunctions = if (functions.size > 1 && bindingContext != BindingContext.EMPTY) {
-                val referenceDescriptors = references.mapNotNull {
-                    it.getResolvedCall(bindingContext)?.resultingDescriptor
+            val unusedFunctions = when {
+                functions.size > 1 && bindingContext != BindingContext.EMPTY -> {
+                    val referenceDescriptors = references.mapNotNull {
+                        it.getResolvedCall(bindingContext)?.resultingDescriptor
+                    }
+                    functions.filter {
+                        bindingContext[BindingContext.DECLARATION_TO_DESCRIPTOR, it] !in referenceDescriptors
+                    }
                 }
-                functions.filter {
-                    bindingContext[BindingContext.DECLARATION_TO_DESCRIPTOR, it] !in referenceDescriptors
-                }
-            } else {
-                if (references.isEmpty()) functions else emptyList()
+                references.isEmpty() -> functions
+                else -> emptyList()
             }
             unusedFunctions.map {
                 CodeSmell(issue, Entity.from(it), "Private function $name is unused.")
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
@@ -15,7 +15,6 @@ import io.gitlab.arturbosch.detekt.rules.isMainFunction
 import io.gitlab.arturbosch.detekt.rules.isOpen
 import io.gitlab.arturbosch.detekt.rules.isOperator
 import io.gitlab.arturbosch.detekt.rules.isOverride
-import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtClass
 import org.jetbrains.kotlin.psi.KtFile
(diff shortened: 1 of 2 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -118,19 +130,12 @@ private class UnusedFunctionVisitor(allowedNames: Regex) : UnusedMemberVisitor(a
...
-
-        if (calledMethodName != null) {
-            callExpressions.add(calledMethodName)
-        }
+        val name = when (expression) {
+            is KtOperationReferenceExpression -> expression.getReferencedName()
+            is KtNameReferenceExpression -> expression.getReferencedName()
+            else -> null
+        } ?: return
+        functionReferences.getOrPut(name) { mutableListOf() }.add(expression)
```
schalkms: Could these statements be simplified to have a lower complexity? I'm thinking about something like this.

```suggestion
        if (expression is KtOperationReferenceExpression || expression is KtNameReferenceExpression) {
            val name = expression.getReferencedName()
            functionReferences.getOrPut(name) { mutableListOf() }.add(expression)
        }
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -73,19 +73,31 @@ private abstract class UnusedMemberVisitor(protected val allowedNames: Regex) :
...
+    private val functionDeclarations = mutableMapOf<String, MutableList<KtFunction>>()
+    private val functionReferences = mutableMapOf<String, MutableList<KtReferenceExpression>>()
 
     override fun getUnusedReports(issue: Issue): List<CodeSmell> {
-        for (call in callExpressions) {
-            if (functions.isEmpty()) {
-                break
+        return functionDeclarations.flatMap { (name, functions) ->
+            val references = functionReferences[name].orEmpty()
+            val unusedFunctions = if (functions.size > 1 && bindingContext != BindingContext.EMPTY) {
```
schalkms: `functions.size > 1`

I'm sorry, from the reading the code I don't quite understand why you check this condition.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -73,19 +73,31 @@ private abstract class UnusedMemberVisitor(protected val allowedNames: Regex) :
...
+                val referenceDescriptors = references.mapNotNull {
+                    it.getResolvedCall(bindingContext)?.resultingDescriptor
+                }
+                functions.filter {
+                    bindingContext[BindingContext.DECLARATION_TO_DESCRIPTOR, it] !in referenceDescriptors
+                }
+            } else {
+                if (references.isEmpty()) functions else emptyList()
+            }
```
schalkms:
```suggestion
            } else if (references.isEmpty()) {
                functions
            } else {
                emptyList()
            }
```

schalkms: [minor] formatting

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
t-kameyama: No. To call 'getReferencedName()', smart cast is required.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
t-kameyama: If there is no function with the same name, just checking the name is enough.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
schalkms: Thanks for the explanation! :-)
resolved...

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
schalkms: Oh yes. Good one!

PR #2507, opened 2020-03-24
Update Kotlin to 1.3.71

arturbosch:


arturbosch:
Cannot assign job to clouds/groups 'pro, pro-backup'. All clouds within the group are either failing, offline, busy or reached build minutes quota.

...

We should look into GitHub actions to replace travis&appveyor...

PR #2509, opened 2020-03-24
Ignore 3rd party libraries in docs for ESLint

michal-simon:
**Motivation:**
One of the main goals of this project is measuring code quality and visibility of technical debt for Kotlin projects. A similar approach can be taken for other languages contained in this project:

image

Despite JavaScript does not have a significant representation, it is good to keep any other analysis, such as ESLint, on top of this repository clean and transparent.

**Problem:**
We want to keep the documentation generation process as simple as possible so it was decided to store a few 3rd party dependencies directly into the repository in `docs/js` without proper dependency management such as NPM or YARN. However, that causes ~288 extra issues in ESLint analysis from vendor libraries.

image

**Solution:**
Because this is not in a critical path of this tool and the process is not changing very frequently, it's OK to not invest significant amount of work to optimize it. On the other hand, we want to keep the ESLint results valuable without false-positive results. The fastest and easiest solution to this problem is to ignore the folder with the libraries from the ESLint analysis by `.eslintignore` file.


schalkms:
@michal-simon I have a question. Does Codeac also comment on PRs if there are issues in the changed code?

michal-simon:
Thank you very much for the review and nice words @schalkms. :)

We are still doing some last QA testing on the Detekt integration before we release it publicly soon but we would love to send another PR to your section once we roll it fully out. Anyway, we would appreciate any feedback from you and the whole community.

Also, if you'd like you can add our little custom badge to this repository:

[Codeac.io](https://app.codeac.io/github/arturbosch/detekt)

We are investing heavily in tight integration with GitHub and Commenting PRs is very high on our roadmap. Is that something you guys would be interested to use in here? We are also happy to slightly adjust our priorities if that would help you.

schalkms:
We have codefactor.io in use which does something similar, but is on the first glance not so powerful as Codenarc. With codefactor.io there's the problem that you couldn't use the custom `detekt-config.yml` configuration. Furthermore, it's nice for 1st time contributors to exactly notice what has triggered a violation in the GitHub review area with the help of a comment from a bot or tool. This saves time for the human when doing a code review, which is what I see one of the goals of your project. Eventually, auto fixes are one of the most awesome things for code reviews. I'm thinking about something like this. GitHub's auto suggestion functionality is of great help here.

**issue: LineEndsWithWhitespace**
```diff
- println("Hello World") 
+ println("Hello World")
```

I hope I could provide some feedback.
I wish you good luck with your project. Keep on rocking! 💪👍

CC @arturbosch suggesting code fixes could be a nice add-on for detekt and its users/integrators. I know that we decided (with good reasoning) to not mess with user code.
But providing suggesting is a different story, I guess. What do you think?

BraisGabin:
I don't get the point of this PR. We don't run ESLint in this project and I don't see why we should run it.

Your product seems really nice btw.

> Eventually, auto fixes are one of the most awesome things for code reviews. I'm thinking about something like this. GitHub's auto suggestion functionality is of great help here.

MAN! That would be awesome!

> CC @arturbosch suggesting code fixes could be a nice add-on for detekt and its users/integrators. I know that we decided (with good reasoning) to not mess with user code.
But providing suggesting is a different story, I guess. What do you think?

We could move this to an issue if it gets long. Could this be done by a third party? If it is not, what could we change to allow it? I'm not against it but how big do we want this project?

michal-simon:
Thank you for the feedback guys, those are awesome ideas!

> I don't get the point of this PR. We don't run ESLint in this project and I don't see why we should run it.

At Codeac.io we believe that these days only few projects contain single language and our telemetry strongly supports us in it. Because of that, we decided to analyze all languages found in repository we support so far to give our users the big picture of technical debt. In other words, if a developer has JS file in the project it should be good JavaScript and we want to help them achieve it. 
To filter out false positives on this repo I wanted to help ignore the irrelevant issues and provide a clean dashboard not only for Codeac.io.

According to @schalkms's comment with auto fixing, we are definitely adding more features to help our users fix their technical debt. One example of this is going out this week called Quick Fixes. Here is a little sneak peek relevant to this pull-request:
image
Over time we want to also move into more advanced fixes as you mentioned. We need to be very careful though touching our user's code as well so I totally understand your decision. However, from the feedback we received, helping developers fix the issue is way more valuable then only finding it. I am happy to explore ways of integration of those fixes if you'd like to generate them in the future.

> With codefactor.io there's the problem that you couldn't use the custom `detekt-config.yml` configuration.

All tools we integrate have some sort of configuration files. We are committed to supporting them in their original format with minimum breaking changes. In the case of Detekt we check if there is a `detekt.yml` file in the root of the repository which we then deliver to `--config` parameter of the binary. We are just working on proper documentation for it now. Feel free to add such a file and play with the config if there is a use-case. :)

arturbosch:
Thanks for the PR Michal!

I've looked through some issues on codeac.
I liked the overview from above without any configuration.
This lead to some refactorings and ideas to improve a rule - #2545, #2542, #2541.

The continuous analysis somehow stopped. It worked only for three timepoints (23-24.03).
I do not see any trigger and would expect it continuously to run on the master branch.
The PR section is also empty.

We have a convenient `--generate-config` flag which creates a configuraton file
at `config/detekt/detekt.yml`. The Gradle plugin looks at this location if no
config was provided. Maybe codeac could also apply this convention.

What I miss is an option to turn off specific analyzers without adding files/commits
to the detekt repository or have file filter option to suppress generated code for example.
As we do not use ESLint, we do not want to suppress it's findings.
I'm afraid this is the reason I need to decline this PR.

What I noticed is that the open source plan integration needs permission to read 
all private repositories.
I understand why this was implemented however it may hinder privacy concerned users to sign up. 

@BraisGabin @schalkms suggest fixes sounds like a lot of fun to implement :).
We should totally have this in mind when designing the api for a 2.0 release!

PR #2510, opened 2020-03-25
UnusedPrivateMember: report unused overloaded operators

### LINKED ISSUE #1444: UnusedPrivateMember should consider overloaded operators
For following code `minus` should get reported but not `plus`. 
```kotlin
import java.util.Date
class Foo {
    val bla: Date = Date(System.currentTimeMillis()) + 300L
    companion object {
        private operator fun Date.plus(diff: Long): Date = Date(this.time + diff)
        private operator fun Date.minus(diff: Long): Date = Date(this.time - diff)
    }
}
```
## Context
In issue #1354 we stopped looking at operator functions as we do not have type and symbol resolution.

t-kameyama:
Fixes #1444

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
@@ -15,6 +15,8 @@ import io.gitlab.arturbosch.detekt.rules.isMainFunction
 import io.gitlab.arturbosch.detekt.rules.isOpen
 import io.gitlab.arturbosch.detekt.rules.isOperator
 import io.gitlab.arturbosch.detekt.rules.isOverride
+import org.jetbrains.kotlin.lexer.KtSingleValueToken
+import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.psi.KtClass
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.KtFunction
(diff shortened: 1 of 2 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
```diff
@@ -1002,7 +1002,21 @@ class UnusedPrivateMemberSpec : Spek({
...
+                import java.util.Date
+                class Foo {
+                    val bla: Date = Date(System.currentTimeMillis()) + 300L
+                    companion object {
+                        private operator fun Date.plus(diff: Long): Date = Date(this.time + diff)
+                        private operator fun Date.minus(diff: Long): Date = Date(this.time - diff)
+                    }
+                }
+            """
+            assertThat(subject.compileAndLintWithContext(wrapper.env, code)).hasSize(1)
```
schalkms: I think it's a good idea to assert only 1 thing in a test.
So I suggest to split this and move it to 2 separate tests.
1. plus -> hasSize(0)
2. minus -> hasSize(1)
With that in mind you know exactly which `operator fun` triggers the violation.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -103,15 +114,9 @@ private class UnusedFunctionVisitor(
...
     override fun visitNamedFunction(function: KtNamedFunction) {
-        if (!isDeclaredInsideAnInterface(function)) {
-            when {
-                function.isOperator() -> {
-                    // Resolution: we skip overloaded operators due to no symbol resolution #1444
-                }
-                function.isPrivate() -> collectFunction(function)
-            }
+        if (!isDeclaredInsideAnInterface(function) && function.isPrivate()) {
+            collectFunction(function)
```
schalkms: Nicely refactored

PR #2512, opened 2020-03-26
Setup GitHub Actions

cortinico:
This is a proposal to replace Travis + AppVeyor with Github Actions (as suggested in #2507). Having GH Actions will allow to run a matrix with [win, mac, linux] x [jdk8, jdk11, jdk13].

I've also removed Travis and AppVeyor configuration but we can keep it if we want to have them run alongside for some time.

I've adapted the `.buildscript/deploy_snapshot.sh` to work although I wasn't able to test it. I anyway think it's better to have a separate workflow that runs only on one agent + only on push to master.

Happy to get some input/feedbacks/ideas.


.buildscript/deploy_snapshot.sh
```diff
@@ -2,24 +2,26 @@
 #
 # Deploy a jar, source jar, and javadoc jar to Sonatype's snapshot repo.
 #
-# Adapted from https://coderwall.com/p/9b_lfq and
+# The script was originally adapted from https://coderwall.com/p/9b_lfq and
 # http://benlimmer.com/2013/12/26/automatically-publish-javadoc-to-gh-pages-with-travis-ci/ and
 # https://github.com/JakeWharton/RxBinding/blob/master/.buildscript/deploy_snapshot.sh
+#
+# Not is has been adapted to work with Github Actions
```
arturbosch:
```suggestion
# Now is has been adapted to work with Github Actions.
```

.buildscript/deploy_snapshot.sh
```diff
@@ -2,24 +2,26 @@
 #
 # Deploy a jar, source jar, and javadoc jar to Sonatype's snapshot repo.
 #
-# Adapted from https://coderwall.com/p/9b_lfq and
+# The script was originally adapted from https://coderwall.com/p/9b_lfq and
 # http://benlimmer.com/2013/12/26/automatically-publish-javadoc-to-gh-pages-with-travis-ci/ and
 # https://github.com/JakeWharton/RxBinding/blob/master/.buildscript/deploy_snapshot.sh
+#
+# Now is has been adapted to work with Github Actions.
```
arturbosch:
```suggestion
# Now it has been adapted to work with Github Actions.
```

.github/workflows/pre-merge.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - name: Verify Generator Output
+      run: ./gradlew verifyGeneratorOutput
+
+
+    - name: Deploy Snapshot
+      run: .buildscript/deploy_snapshot.sh
+
+
+    - name: Generate Coverage Report
+      run: ./gradlew jacocoTestReport --scan
```
arturbosch: The `--scan` triggers gradle build scan again. I think this was made by mistake previously?

arturbosch: We can savely remove it.

.github/workflows/pre-merge.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+    - name: Build detekt (WIN)
+      run: ./gradlew build installShadowDist --build-cache --parallel -PwarningsAsErrors=true -Pcompile-test-snippets=%COMPILE_TEST_SNIPPETS%
+      if: matrix.os == 'windows-latest'
+    - name: Run detekt-cli --help (WIN)
+      run: detekt-cli\build\install\detekt-cli-shadow\bin\detekt-cli --help
+      if: matrix.os == 'windows-latest'
+    - name: Run detekt-cli with argsfile (WIN)
+      run: detekt-cli\build\install\detekt-cli-shadow\bin\detekt-cli @config\detekt\argsfile_windows
+      if: matrix.os == 'windows-latest'
```
BraisGabin: I think that it's safe to remove this steps and remove the conditions on the top ones. And remove the file `config\detekt\argsfile_windows` and probably rename `config\detekt\argsfile_unix` to `config\detekt\argsfile`.

.github/workflows/pre-merge.yaml
```diff
@@ -0,0 +1,82 @@
...
+jobs:
+  gradle:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        jdk: [8, 11, 13]
+    runs-on: ${{ matrix.os }}
+    if: "!contains(github.event.head_commit.message, 'ci skip')"
+    env:
+      COMPILE_TEST_SNIPPETS: true
```
BraisGabin: If we are going to compile the code in all the cases we can remove this env. Or we can set this to true just if `jdk = 8`.

arturbosch: I would go with jdk = 8 and see the duration differences ^^

arturbosch:
> Thanks! I very much welcome this change, since I invested quite some time to re-trigger the flaky builds on TravisCI. Good job @cortinico !
> But we have to be careful here, before we remove the TravisCI and AppVeyor scripts. The current GitHub action script misses `codecov.io` code coverage reports.
> I think it can be added by using a corresponding [GitHub action](https://github.com/codecov/codecov-action).

Do you mean the last action?
```
 - name: Publish Coverage
      uses: codecov/codecov-action@v1
```

.github/workflows/pre-merge.yaml
cortinico: Actually the paths used by Unix/Windows are different:

```
// Win
detekt-cli\build\install\detekt-cli-shadow\bin\detekt-cli

// Unix
./detekt-cli/build/libs/detekt-cli-*-all.jar
```

Also the Windows build is calling the `installShadowDist` task while Unix is calling `shadowJar`. I tried to keep the same configuration from Travis/AppVeyor but I'm not aware of the reason why the approach is different between the two.

.github/workflows/pre-merge.yaml
cortinico: Moved to be true only for JDK8

cortinico:
So far I've addressed all the comments. There are still two open points to address: caching and snapshots.

> Don't we need these lines?
>   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
>   - rm -fr $HOME/.gradle/caches/*/plugin-resolution/

I was playing around with those in a06c017. Those are Travis specific configuration commands. Looks like caching is working fine on the Ubuntu worked (build time ~6 minutes). MacOS and Win worker are instead failing to cache properly. I'm still investigating what's going on.

As for Snapshots, is currently hard to test as it's configured to publish a snapshot on `push` to the `master` branch. This mirrors the former Travis setup. I'm currently concerned as the current setup will run the `deploy_snapshot.sh` script 3 times (JDK8 for win, mac, linux). Looks like this was the case also before on Travis, right? (JDK8 for mac and linux). 

Should I limit this to run only on JDK8/Linux? Or should it be a separate action?

arturbosch:
> As for the Snapshot setup, is currently hard to test as it's configured to publish a snapshot on `push` to the `master` branch. This mirrors the former Travis setup. I'm currently concerned as the current setup will run the `deploy_snapshot.sh` script 3 times (JDK8 for win, mac, linux). Looks like this was the case also before on Travis, right? (JDK8 for mac and linux).
> 
> Should I limit this to run only on JDK8/Linux? Or should it be a separate action?

Lets go with just JDK8/Linux. 
A separate action which just runs on the actual merge to master would be cool too.

.github/workflows/pre-merge.yaml
BraisGabin: I think that they are different because that's what happen when you have duplicated code. And, The bottom path isn't a valid path on windows? I think so. 

Anyway, this comment is not a stopper. We can always improve the ci later. Let's focus on cache and snapshots. Once this is merged we can go back to this topic.

cortinico: Yup let's focus on landing this PR and improve this specific step later.

schalkms:
Do we know why the `Publish Coverage` task lasts so long on MacOS?

cortinico:
The cache miss issue should be solved now. Funny story, @3flex had the same issue we were facing here: https://github.com/actions/cache/issues/144

I've moved the logic of the publish script to its own Action.

> Do we know why the Publish Coverage task lasts so long on MacOS?

I noticed that the `Publish Coverage` goes sometimes in timeout. Apparently the external GH Action is doing some retry. I could eventually switch to the good old `bash <(curl -s https://codecov.io/bash)` approach.

schalkms:
> I could eventually switch to the good old bash <(curl -s https://codecov.io/bash) approach.

That would be my preferred approach until the referenced issue is fixed. Thanks again for the awesome research work in this PR! @cortinico

schalkms:
AppVeyor is still triggered even though there's no config anymore. Therefore, the failed build on this platform can be ignored.

.github/workflows/pre-merge.yaml
```diff
@@ -0,0 +1,82 @@
...
+  gradle:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        jdk: [8, 11, 13]
+    runs-on: ${{ matrix.os }}
+    if: ${{ !contains(github.event.head_commit.message, 'ci skip') }}
+    env:
+      # We compile the test snippets only on Java 8.
+      COMPILE_TEST_SNIPPETS: ${{ matrix.jdk == 8 }}
```
schalkms: I think it's enough to only compile the test snippets on Ubuntu JDK 8. This improves the build run times.

```suggestion
      COMPILE_TEST_SNIPPETS: ${{ matrix.os == 'ubuntu-latest' && matrix.jdk == 8 }}
```

cortinico:
We should be good to go.
Although I noticed that the cache on the `macos` workers is significantly bigger (2+ Gb vs an average of 400Mb for win & linux). I don't exactly know why.

I don't think this is a blocker as for now as the overall time is anyway lower than it was before with AppVeyor/Travis so I'd love to followup on a later PR on that.

3flex:
> Funny story, @3flex had the same issue we were facing here: actions/cache#144

Because I was trying to introduce Actions for this project myself :)

Glad you worked it out, thanks! And I saw the same regarding cache size, and agree it shouldn't be a blocker - if the cache grows too large it will be pruned automatically, though we'd lose some caching benefits. Overall though I see this as a win especially as Windows builds run in parallel which wasn't possible with AppVeyor.

PR #2513, opened 2020-03-26
UnnecessaryAbstractClass: fix false positive when abstract class has abstract inherited members

### LINKED ISSUE #727: False positive on rule UnnecessaryAbstractClass

t-kameyama:
Fixes #727


t-kameyama:
I'm fixing.

PR #2517, opened 2020-03-26
Prepare 1.7.1 release

arturbosch:


build.gradle.kts
```diff
@@ -346,3 +347,21 @@ val detektProjectBaseline by tasks.registering(DetektCreateBaselineTask::class)
...
+// release section
+
+githubRelease {
+    token(project.findProperty("github.token") as? String ?: "")
+    owner.set("arturbosch")
+    repo.set("detekt")
+    overwrite.set(true)
+    dryRun.set(false)
+    body {
+        var changelog = project.file("docs/pages/changelog 1.x.x.md").readText()
```
schalkms: A hard-coded version file is not a good idea. What happens if we update to detekt 2.x.y? Do you want to keep this md file?

arturbosch: Lets keep it hardcoded for way. For 2.0 I would want to rename the old changelog.md to `changelog_rc.md` and just have `changelog` to be the actual one.

PR #2520, opened 2020-03-26
Simplify an unnecessary complex test case

schalkms:


BraisGabin:
🤔 maybe this big functions were to test that `LargeClass` doesn't flag "large functions" as "large classes". probably that's the reson to have such a complex code in this test.

schalkms:
@BraisGabin you are completely right. I updated the corresponding code snippet accordingly.

PR #2524, opened 2020-03-27
UseDataClass: fix false positive with `copy` function

### LINKED ISSUE #1040: False-positive for UseDataClass: `copy` shouldn't be considered a default function
One of the rules to check if we should use a data class is to check if the class only contains "default functions": https://github.com/arturbosch/detekt/blob/master/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClass.kt#L56

I understand why `hashCode`, `equals` and `toString` are there, but Kotlin's data class `copy` is not an overridable function.

False-positive class:
```
class Clipboard(private val type: Int) {
    fun copy(text: String) = Unit
}
```

I think `copy` could be removed from the list of default functions since it's not overridable.

What do you think?

t-kameyama:
Fixes #1040

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
@@ -1,9 +1,11 @@
 package io.gitlab.arturbosch.detekt.rules.style
 
 import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.test.KtTestCompiler
 import io.gitlab.arturbosch.detekt.test.TestConfig
 import io.gitlab.arturbosch.detekt.test.assertThat
 import io.gitlab.arturbosch.detekt.test.compileAndLint
+import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
 import io.gitlab.arturbosch.detekt.test.lint
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
@@ -11,6 +13,11 @@ import org.spekframework.spek2.style.specification.describe
 class UseDataClassSpec : Spek({
 
     val subject by memoized { UseDataClass(Config.empty) }
+
+    val wrapper by memoized(
+        factory = { KtTestCompiler.createEnvironment() },
+        destructor = { it.dispose() }
+    )
 
     describe("UseDataClass rule") {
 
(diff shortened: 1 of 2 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
```diff
@@ -137,6 +144,73 @@ class UseDataClassSpec : Spek({
...
+            it("does report with copy method") {
+                val code = """
+                    class D(val a: Int, val b: String) {
+                        fun copy(a: Int, b: String): D = D(a, b)
+                    }
+                """
+                assertThat(subject.compileAndLintWithContext(wrapper.env, code)).hasSize(1)
+            }
+
+            it("does report with copy method which has implicit return type") {
```
schalkms:
```suggestion
            it("does report with copy method which has an implicit return type") {
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
```diff
@@ -137,6 +144,73 @@ class UseDataClassSpec : Spek({
...
+            it("does not report with copy method which has different parameter types 2") {
+                val code = """
+                    class D(val a: Int, val b: String) {
+                        fun copy(a: Int, b: String?): D = D(a, b.toString())
+                    }
+                """
+                assertThat(subject.compileAndLintWithContext(wrapper.env, code)).isEmpty()
+            }
+
+            it("does not report with copy method which has different return type") {
```
schalkms:
```suggestion
            it("does not report with copy method which has a different return type") {
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
```diff
@@ -137,6 +144,73 @@ class UseDataClassSpec : Spek({
...
+            it("does not report with copy method which has no parameters") {
+                val code = """
+                    class D(val a: Int, val b: String) {
+                        fun copy(): D = D(0, "")
+                    }
+                """
+                assertThat(subject.compileAndLintWithContext(wrapper.env, code)).isEmpty()
+            }
+
+            it("does not report with copy method which has parameters more than primary constructor") {
```
schalkms:
```suggestion
            it("does not report with copy method which has more parameters than the primary constructor") {
```

Issue #2523, opened 2020-03-27
regression 1.7.1, threshhold does not exist


robstoll:
## Expected Behavior
Upgrading a patch version should always work

## Observed Behavior
```
> Task :detekt FAILED

Property 'complexity>LongParameterList>threshold' is deprecated. Use 'functionThreshold' and 'constructorThreshold' instead.

Property 'complexity>LongParameterList>threshold' is misspelled or does not exist.
```

## Steps to Reproduce
https://github.com/robstoll/niok/pull/41/files
https://travis-ci.org/github/robstoll/niok/jobs/667554736

## Your Environment
* Version of detekt used: 1.7.0

schalkms:
That's why it's dangerous to change configuration options. It's always a model change so to speak. Changing the options could always lead to breaking the user configuration, which should be avoided.

BraisGabin:
Maybe we can rethink the branching model. If we need to create a patch release we can create a new branch starting at the previus tag and cherry-pick the desired fixes. Then we merge that branch again in `master`

arturbosch:
Well, imo this property deprecation was the perfect case that it worked as expected.
The old and the new properties work; default will be the old; a deprecation message was printed.
The problem was that the config validation framework had a bug. It needed/needs some iterations to be stable.

Versions between 1.5.0-1.7.1 had this problem. Maybe other users experienced it but didn't report this.

Tolriq:
If the properties are deprecated should not the default from generateConfig use the new ones?

When I update detekt I always diff the default config to see the changes and the impact of the version. Not having deprecated properties replaced with the new ones hides such changes and makes it harder to know what to do.

robstoll:
Maybe it makes sense to have a flag, turning deprecation warnings into errors (in case that does not already exist)? This way detekt would still provide the current behaviour which might be what users want and now kind of expect.

arturbosch:
> If the properties are deprecated should not the default from generateConfig use the new ones?
> 
> When I update detekt I always diff the default config to see the changes and the impact of the version. Not having deprecated properties replaced with the new ones hides such changes and makes it harder to know what to do.

Yes, the generator uses the new properties so the default config does not have the old properties anymore. This is why the config validation failed for this "unknown" property.

> Maybe it makes sense to have a flag, turning deprecation warnings into errors (in case that does not already exist)? This way detekt would still provide the current behaviour which might be what users want and now kind of expect.

This sounds good to me and could be a config property next to `validation`.
[...]

PR #2525, opened 2020-03-27
Do not fail the build on config property warnings/deprecations - #2523

arturbosch:


arturbosch:
The codecov report is wierd. Maybe the github actions cache more stuff...

PR #2530, opened 2020-03-27
Fix report for documented data classes property

### LINKED ISSUE #2529: UndocumentedPublicProperty reported for documented data classes property
## Expected Behavior
The following code:

```kotlin
/**
 * @property prop1 The property is documented
 */
data class MyDataClass(val prop1: String)
```

should not see any `UndocumentedPublicProperty` being reported because all properties are documented.

## Observed Behavior
Running detekt reports an issue:

```
UndocumentedPublicProperty - [prop1] at /home/me/workspace/project/src/main/kotlin/Example.kt:8:24
```

## Steps to Reproduce
The code above is all you need to reproduce the issue.

## Context
I was trying to update detekt from `1.6.0` to `1.7.1` and I got this errors. But I cannot fix them, because the properties are already documented.

## Your Environment
* Version of detekt used: `1.7.1`
* Version of Gradle used: `6.3`
* Operating System and version: `linux 5.5.11`
* Link to your project: https://github.com/jcornaz/kwik/pull/141

schalkms:
A documented property should not be reported as undocumented by the UndocumentedPublicProperty rule.

This closes #2529

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
@@ -11,9 +11,7 @@ class UndocumentedPublicPropertySpec : Spek({
     describe("UndocumentedPublicProperty rule") {
 
         it("reports undocumented public property") {
-            val code = """
-                val a = 1
-            """
+            val code = "val a = 1"
             assertThat(subject.compileAndLint(code)).hasSize(1)
         }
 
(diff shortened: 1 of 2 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
```diff
@@ -125,10 +123,12 @@ class UndocumentedPublicPropertySpec : Spek({
         it("does not report documented public properties in a primary constructor") {
             val code = """
                 /**
-                * @a int1
-                * [b] int2 
+                * @property a int1
+                * [b] int2
+                * @property [c] int3
+                * @param d int4
```
BraisGabin: If `d`is just documented as parameter... we should flag it, right? I mean, we want to document the properties.

And probably it would be better to split this test in 4.

schalkms: As always, thanks for the review. Please see my comment in the referenced issue.
https://github.com/arturbosch/detekt/issues/2529#issuecomment-605294205

Splitting the test case up makes sense.

PR #2543, opened 2020-03-28
Release version 1.7.2

arturbosch:


arturbosch:
Coverage was okay because changing `gradle.properties` leads to a clean build ...

PR #2547, opened 2020-03-28
Cleanup Gradle Folders

cortinico:
After some investigation seems like the `macos` worker has several old versions of Gradle in the home (4.x, 5.x, ...). I'm adding a cleanup of the cache folders to make sure we don't cache/restore 3Gb of data.

Here a run with a proof: https://github.com/cortinico/detekt/actions/runs/65408267

The first run of this PR will be longer as I've invalidated the caches.

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -24,17 +24,25 @@ jobs:
     - name: Checkout Repo
       uses: actions/checkout@v2
 
+      # Let's cleanup the gradle cache folders to make sure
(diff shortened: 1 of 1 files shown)


BraisGabin:
I found the issue: https://github.com/actions/virtual-environments/issues/427

It seems that the fix is done and it will be released soon.

cortinico:
> I found the issue: [actions/virtual-environments#427](https://github.com/actions/virtual-environments/issues/427)
> It seems that the fix is done and it will be released soon.

Awesome, thanks for investigating @BraisGabin

PR #2554, opened 2020-03-29
Include statistics for our custom detekt tasks

arturbosch:
Also `detektFormat` finds more issues with `detekt-formatting` than the Gradle task for each sub module ... I guess the bug reports are real :D

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -269,10 +269,11 @@ subprojects {
 
 // detekt self analysis section
 
-val analysisDir = files(projectDir)
+val analysisDir = file(projectDir)
 val baselineFile = file("$rootDir/config/detekt/baseline.xml")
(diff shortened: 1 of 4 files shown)


schalkms:
> Also detektFormat finds more issues with detekt-formatting than the Gradle task for each sub module ... I guess the bug reports are real :D 

:-/

arturbosch:
> I have no idea what these changes do. And I don't know which bugs you are refering in the description. Do you mean that `detektFormat` fixes things that `detekt` doesn't flag?

Oh, I just turned off the "silent mode" on the custom tasks like `detektFormat` and `detektAll`.
Yes, `detektFormat` finds issues which just `detekt` does not report. We have some open bug tickets for detekt-formatting. I was just noticing that apparently they are real ...

PR #2556, opened 2020-03-30
UseCheckOrError/UseRequire: fix false positive with a non-String argument

### LINKED ISSUE #2514: False positive: UseCheckOrError
## Expected Behavior

This code should not be flagged

```kotlin
val lambda: (Throwable) -> Unit = {
  if (it is IllegalStateException) {
    println("something")
  } else {
    throw IllegalStateException(it)
  }
}
```

## Observed Behavior
It is flagged

## Context
Replacing `throw IllegalStateException(error)` with `error(error)` changes the behaviour so it should not be flagged.
## Your Environment

t-kameyama:
Fixes #2514

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtValueArgument.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtValueArgument.kt
@@ -1,0 +1,21 @@
+package io.gitlab.arturbosch.detekt.rules
+
+import org.jetbrains.kotlin.builtins.KotlinBuiltIns
+import org.jetbrains.kotlin.descriptors.ValueParameterDescriptor
+import org.jetbrains.kotlin.psi.KtValueArgument
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+
+fun KtValueArgument.isString(bindingContext: BindingContext): Boolean {
+    return if (bindingContext != BindingContext.EMPTY) {
+        val descriptor =
+            getArgumentExpression()?.getResolvedCall(bindingContext)?.resultingDescriptor as? ValueParameterDescriptor
+        val type = descriptor?.type
+        type != null && KotlinBuiltIns.isString(type)
+    } else {
+        true
+    }
+}
+
+fun List<KtValueArgument>.isEmptyOrSingleStringArgument(bindingContext: BindingContext): Boolean =
+    isEmpty() || singleOrNull()?.isString(bindingContext) == true
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtValueArgument.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtValueArgument.kt
@@ -2,18 +2,20 @@ package io.gitlab.arturbosch.detekt.rules
 
 import org.jetbrains.kotlin.builtins.KotlinBuiltIns
 import org.jetbrains.kotlin.descriptors.ValueParameterDescriptor
+import org.jetbrains.kotlin.psi.KtStringTemplateExpression
 import org.jetbrains.kotlin.psi.KtValueArgument
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
 
 fun KtValueArgument.isString(bindingContext: BindingContext): Boolean {
+    val argumentExpression = getArgumentExpression()
     return if (bindingContext != BindingContext.EMPTY) {
         val descriptor =
-            getArgumentExpression()?.getResolvedCall(bindingContext)?.resultingDescriptor as? ValueParameterDescriptor
+            argumentExpression?.getResolvedCall(bindingContext)?.resultingDescriptor as? ValueParameterDescriptor
         val type = descriptor?.type
         type != null && KotlinBuiltIns.isString(type)
     } else {
-        true
+        argumentExpression is KtStringTemplateExpression
     }
 }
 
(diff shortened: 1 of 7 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtValueArgument.kt
```diff
@@ -0,0 +1,21 @@
...
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+
+fun KtValueArgument.isString(bindingContext: BindingContext): Boolean {
+    return if (bindingContext != BindingContext.EMPTY) {
+        val descriptor =
+            getArgumentExpression()?.getResolvedCall(bindingContext)?.resultingDescriptor as? ValueParameterDescriptor
+        val type = descriptor?.type
+        type != null && KotlinBuiltIns.isString(type)
+    } else {
+        true
```
BraisGabin: Can we check here if the value argument is a literal string otherwise return false (or null because we don't know)?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseRequireSpec.kt
```diff
@@ -83,6 +90,15 @@ class UseRequireSpec : Spek({
             assertThat(subject.lint(code)).isEmpty()
         }
 
+        it("does not report an issue if the exception thrown has a non-String argument") {
+            val code = """
+                fun test(throwable: Throwable) {
+                    if (throwable !is NumberFormatException) throw IllegalArgumentException(throwable)
+                }
+            """.trimIndent()
```
schalkms: It's not necessary to call `.trimIndent()` again, since `lint()` already does it for us.

```suggestion
            """
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseCheckOrErrorSpec.kt
```diff
@@ -94,5 +101,17 @@ class UseCheckOrErrorSpec : Spek({
...
+
+        it("does not report an issue if the exception thrown has a non-String argument") {
+            val code = """
+                fun test(throwable: Throwable) {
+                    when(throwable) {
+                        is NumberFormatException -> println("a")
+                        else -> throw IllegalStateException(throwable)
+                    }
+                }
+            """.trimIndent()
```
schalkms: It's not necessary to call `.trimIndent()` again, since `lint()` already does it for us.

```suggestion
            """
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseCheckOrErrorSpec.kt
```diff
@@ -10,6 +12,11 @@ class UseCheckOrErrorSpec : Spek({
...
     val subject by memoized { UseCheckOrError(Config.empty) }
 
+    val wrapper by memoized(
+        factory = { KtTestCompiler.createEnvironment() },
+        destructor = { it.dispose() }
+    )
+
     describe("UseCheckOrError rule") {
 
         it("reports if a an IllegalStateException is thrown") {
```
schalkms: What happens if the user throws an exception with 2 exceptions arguments, whereby the 1st is of type `String` and the 2nd is of another type. It might be worth to add a test for this case.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseCheckOrError.kt
```diff
@@ -18,16 +19,16 @@ import org.jetbrains.kotlin.psi.KtThrowExpression
  * Prefer them instead of manually throwing an IllegalStateException.
  *
  * <noncompliant>
- * if (value == null) throw new IllegalStateException("value should not be null")
- * if (value < 0) throw new IllegalStateException("value is $value but should be at least 0")
+ * if (value == null) throw IllegalStateException("value should not be null")
+ * if (value < 0) throw IllegalStateException("value is $value but should be at least 0")
```
schalkms: good catch ^^ (Java syntax)

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtValueArgument.kt
schalkms: I agree with this. We might want to consider not allowing rules to run for the scenario when no `BindingContext` is available.
The complexity will increase steadily, if we continue to support both cases.
We might want to start issuing a warning for this case.
CC @3flex @arturbosch

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseRequireSpec.kt
```diff
@@ -83,6 +90,15 @@ class UseRequireSpec : Spek({
...
+            assertThat(subject.compileAndLintWithContext(wrapper.env, code)).isEmpty()
```
BraisGabin: Could we add a new test to check that if we don't use type solving this same code is not reported?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseRequireSpec.kt
```diff
@@ -83,6 +90,65 @@ class UseRequireSpec : Spek({
...
+        }
+
+        it("does not report if the exception thrown has a non-String literal argument") {
+            val code = """
+                fun test(throwable: Throwable) {
+                    val s = ""
+                    if (throwable !is NumberFormatException) throw IllegalArgumentException(s)
+                }
+            """
+            assertThat(subject.lint(code)).isEmpty()
```
schalkms: Please use `compileAndLint()` instead of `lint()` to make sure that the code actually compiles.

PR #2559, opened 2020-03-30
Display actual and threshold values for several metric rules

swsms:
Display actual values and threshold values for several metric rules.
The text is similar to [this one](https://github.com/arturbosch/detekt/blob/d7088791a6d2a075202570ce87f1bba8c0413c59/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt#L105).


BraisGabin:
Nice improvement

Issue #2498, opened 2020-03-22
Support yaml lists in the configuration values


BraisGabin:
This issue is a continuation of #2463. There we talked about ForbiddenImports but we should generalice that solution to the rest of lists that we have in our yaml config file. (This case was implemented already #2474).

The reason for doing this is: 
- It's easier to mantain the configuration using lists than strings
- We don't need to handle strange cases splitting strings (empty strings and similar)

The idea is to allow the use of yaml list where we are now using a string with `,`. And if all goes well remove we can remove this old way in 2.0.

This is the list of configurations that, in 1.7.0, are lists splited by `,`:

- [x] *>excludes
- [x] ComplexMethod>nestingFunctions
- [x] ExceptionRaisedInUnexpectedLocation>methodNames
- [x] ForbiddenClassName>forbiddenName
- [x] ForbiddenComment>values
- [X] ForbiddenImport>imports
- [x] ForbiddenMethodCall>methods
- [x] ForbiddenPublicDataClass>ignorePackages
- [x] FunctionOnlyReturningConstant>excludeAnnotatedFunctions
- [x] LabeledExpression>ignoredLabels
- [x] LateinitUsage>excludeAnnotatedProperties
- [x] LongParameterList>ignoreAnnotated
- [x] MagicNumber>ignoreNumbers
- [x] SwallowedException>ignoredExceptionTypes
- [x] ThrowingExceptionsWithoutMessageOrCause>exceptions
- [x] UnnecessaryAbstractClass>excludeAnnotatedClasses
- [x] UseDataClass>excludeAnnotatedClasses
- [x] WildcardImport>excludeImports

BraisGabin:
All the PRs are done now :)

arturbosch:
Great work!

PR #2570, opened 2020-04-01
Add ignoreAnnotated option to LongParameterList

### LINKED ISSUE #2563: Rule configuration request: Ignore based on annotations
```yml
LongParameterList:
  active: true
  functionThreshold: 7
  constructorThreshold: 7
  ignoreDefaultParameters: false
  ignoreDataClasses: true
  ignoreAnnotated: "javax.inject.Inject,javax.annotation.Generated"
```

## Context
Currently the `LongParameterList` rule applies to all constructors but has an opt-out for data classes. I'd like to suggest adding a new configuration to ignore based on annotations, such as `javax.inject.Inject`. This comes up often in our DI plumbing

ZacSweers:
Resolves #2563

This adds a new `ignoreAnnotated` option to `LongParameterList`, that allows for ignoring annotations within function/class/file scope.

### DIFF
--- a/detekt-rules/build.gradle.kts
+++ b/detekt-rules/build.gradle.kts
@@ -4,4 +4,7 @@ dependencies {
     implementation(project(":detekt-api"))
 
     testImplementation(project(":detekt-test"))
+
+    // For @javax.inject.Inject usage in LongParameterListSpec.kt
+    testImplementation("javax.inject:javax.inject:1")
 }
--- a/detekt-rules/build.gradle.kts
+++ b/detekt-rules/build.gradle.kts
@@ -4,7 +4,4 @@ dependencies {
     implementation(project(":detekt-api"))
 
     testImplementation(project(":detekt-test"))
-
-    // For @javax.inject.Inject usage in LongParameterListSpec.kt
-    testImplementation("javax.inject:javax.inject:1")
 }
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
@@ -80,14 +80,14 @@ class LongParameterList(
 
     private fun validateConstructor(constructor: KtConstructor<*>) {
         val owner = constructor.getContainingClassOrObject()
-        if (owner is KtClass && owner.isDataClassOrIgnored) {
+        if (owner is KtClass && owner.isDataClassOrIgnored()) {
             return
         }
         
-        fun KtClassOrObject.isDataClassOrIgnored() = isIgnored || ignoreDataClasses && owner.isData()
-        }
         validateFunction(constructor, constructorThreshold)
     }
+
+    private fun KtClass.isDataClassOrIgnored() = isIgnored() || ignoreDataClasses && isData()
 
     private fun validateFunction(function: KtFunction, threshold: Int) {
         if (function.isOverride() || function.isIgnored() || function.containingKtFile.isIgnored()) return
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
@@ -83,7 +83,6 @@ class LongParameterList(
         if (owner is KtClass && owner.isDataClassOrIgnored()) {
             return
         }
-        
         validateFunction(constructor, constructorThreshold)
     }
 
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
@@ -1,8 +1,23 @@
 package io.gitlab.arturbosch.detekt.rules.complexity
 
-import io.gitlab.arturbosch.detekt.api.*
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Metric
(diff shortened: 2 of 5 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
```diff
@@ -62,16 +72,38 @@ class LongParameterList(
...
+        }
+    }
+
+    private fun KtClass.isIgnored(): Boolean {
+        return annotationEntries.any {
+            ignoreAnnotated.contains(it.typeReference?.text)
+        }
+    }
+
+    private fun KtFile.isIgnored(): Boolean {
```
arturbosch, line 87: These three nodes have `KtAnnotated` as a common super type.
Please refactor to
```kt
    private fun KtAnnotated.isIgnored(): Boolean =
        annotationEntries.any { ignoreAnnotated.contains(it.typeReference?.text) }
```

detekt-rules/build.gradle.kts
```diff
@@ -4,4 +4,7 @@ dependencies {
     implementation(project(":detekt-api"))
 
     testImplementation(project(":detekt-test"))
+
+    // For @javax.inject.Inject usage in LongParameterListSpec.kt
+    testImplementation("javax.inject:javax.inject:1")
```
arturbosch, line 9: This dependency is actually not needed for the parser as we do not resolve any types. Please make sure the tests run without it.

ZacSweers: The CI checks fail without this, see the failure on the commit before I added it

BraisGabin: If we use `compileAndLint` we need it in the classpath. The other option is to declare that annotation in every code snipet.

ZacSweers: updated!

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
ZacSweers: Done!

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
```diff
@@ -26,6 +29,7 @@ import org.jetbrains.kotlin.psi.KtSecondaryConstructor
  * @configuration constructorThreshold - number of constructor parameters required to trigger the rule (default: `7`)
  * @configuration ignoreDefaultParameters - ignore parameters that have a default value (default: `false`)
  * @configuration ignoreDataClasses - ignore long constructor parameters list for data classes (default: `true`)
+ * @configuration ignoreAnnotated - ignore long parameters list for constructors or functions in the context of these comma-separated annotation class names (default: `''`)
```
schalkms, line 32: @BraisGabin I'm not show whether we should use the yaml list syntax for this configuration option?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
```diff
@@ -62,16 +72,26 @@ class LongParameterList(
...
         val owner = constructor.getContainingClassOrObject()
-        val isDataClass = owner is KtClass && owner.isData()
-        if (!ignoreDataClasses || !isDataClass) {
-            validateFunction(constructor, constructorThreshold)
+        if (owner is KtClass) {
+            if (ignoreDataClasses && owner.isData()) {
+                return
+            } else if (owner.isIgnored()) {
+                return
+            }
```
schalkms, line 88: I think this code can be simplified by introducing a new function.

```suggestion
        if (owner is KtClass && owner.isDataClassOrIgnored) {
            return
        }
        
        fun KtClassOrObject.isDataClassOrIgnored() = isIgnored || ignoreDataClasses && owner.isData()
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterListSpec.kt
```diff
@@ -67,5 +67,47 @@ class LongParameterListSpec : Spek({
             val code = "data class Data(val a: Int, val b: Int, val c: Int, val d: Int, val e: Int, val f: Int, val g: Int)"
             assertThat(rule.compileAndLint(code)).isEmpty()
         }
+
+        it("does not report long parameter list for constructors if file is annotated with ignored annotation") {
+            val config = TestConfig(mapOf(LongParameterList.IGNORE_ANNOTATED to "javax.annotation.Generated"))
```
schalkms, line 72: This code duplication can be removed by providing a local property inside a new `context() {}` block.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
ZacSweers: I was matching what some others that used fully qualified names do, and this allows for some wildcard matching. Unclear if using the actual yml list format would support this

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterListSpec.kt
ZacSweers: Do you mind if that's left to another PR? Seems like that's asking for refactoring to this test in general that's outside the scope of this PR

schalkms: It's okay.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
BraisGabin: I would go with string for now. For consistency. I'll change all at once. And, to give a bit of context, we are talking about #2498

PR #2583, opened 2020-04-06
ComplexMethod>nestingFunctions allow yaml list

BraisGabin:
Please, review commit by commit.

This is the first of a serie of PRs to implement #2498

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -71,7 +71,7 @@ complexity:
     ignoreSingleWhenExpression: false
     ignoreSimpleWhenEntries: false
     ignoreNestingFunctions: false
-    nestingFunctions: run,let,apply,with,also,use,forEach,isNotNull,ifNull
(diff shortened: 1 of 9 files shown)


detekt-cli/src/main/resources/default-detekt-config.yml
```diff
@@ -71,7 +71,7 @@ complexity:
     ignoreSingleWhenExpression: false
     ignoreSimpleWhenEntries: false
     ignoreNestingFunctions: false
-    nestingFunctions: run,let,apply,with,also,use,forEach,isNotNull,ifNull
+    nestingFunctions: [run, let, apply, with, also, use, forEach, isNotNull, ifNull]
```
arturbosch, line 74: Is this the short list syntax for: 
```
nestingFunctions:
  - run
  - let
  ...
```
?

BraisGabin: Yes, both works. I you prefer the other syntax as the default I can change it.

arturbosch: Lets go with this one. Saving some space and easier to refactor. Ofc the user can choose it.

PR #2587, opened 2020-04-06
*>*Annotated* allow yaml lists

BraisGabin:
This PR is one in a serie of PRs to implement #2498

~We need to merge first #2583 because I need a commit from there.~

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/AnnotationExcluder.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/AnnotationExcluder.kt
@@ -13,7 +13,7 @@ class AnnotationExcluder(
     private val excludes: SplitPattern
 ) {
 
-    private var resolvedAnnotations = root.importList
+    private val resolvedAnnotations = root.importList
             ?.imports
             ?.asSequence()
             ?.filterNot { it.isAllUnder }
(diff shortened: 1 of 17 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/AnnotationExcluder.kt
```diff
@@ -10,9 +10,14 @@ import org.jetbrains.kotlin.psi.KtFile
...
 class AnnotationExcluder(
     root: KtFile,
-    private val excludes: SplitPattern
+    private val excludes: List<String>
 ) {
 
+    @Deprecated(
+        message = "Use AnnotationExcluder(KtFile, List<String>) instead",
+        replaceWith = ReplaceWith("AnnotationExcluder(root, excludes.mapAll { it })"))
+    constructor(root: KtFile, excludes: SplitPattern) : this(root, excludes.mapAll { it })
```
arturbosch, line 19: Maybe lets leave this constructor as we now have `excludes.mapAll { it }` in four places?

BraisGabin: The `excludes.mapAll { it }` was an intermedieate step. You can check that in the final step I don't use `SplitPattern` at all. My idea is that we can, deprecate `SplitPattern` later. I think that I'm going to replace all its usages. But probably that's a decission for another PR. I will remove the `Deprecated` for now.

PR #2588, opened 2020-04-07
Fix false positive in UndocumentedPublicClass

### LINKED ISSUE #2580: UndocumentedPublicClass false positive for inner types
## Expected Behavior

Given the following source code:

```kotlin
private class MyClass {
  object MyInnerObject
}
```

I don't expect the `UndocumentedPublicClass` to be reported because `MyClass` is private.

Yes `MyInnerObject` is not *explicitly* marked private. But being an inner type of a private one, makes `MyInnerObject` *effectively* private.

## Observed Behavior
Given the snippet above and the rule `UndocumentedPublicClass` activated: 

```yaml
  UndocumentedPublicClass:
    active: true
    searchInInnerObject: true
```

A `UndocumentedPublicClass` issue is reported, saying that `MyInnerObject` is public and undocumented.

## Steps to Reproduce
The snippets above should be enough to reproduce the issue.

## Context
I added a private class with inner types and got the mentioned false-positive reported by detekt.

## Your Environment
* Version of detekt used: `1.7.4`
* Version of Gradle used: `6.3`
* Operating System and version: `Linux 5.5.13`

schalkms:
Classes that are nested, non-public and undocumented are not flagged by this rule anymore.

Closes #2580

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/Traversing.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/Traversing.kt
@@ -2,6 +2,9 @@ package io.gitlab.arturbosch.detekt.rules
 
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.psi.KtElement
+import org.jetbrains.kotlin.psi.KtNamedDeclaration
+import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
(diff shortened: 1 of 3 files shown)


BraisGabin:
You are talking about `UndocumentedPublicClass` but you edited `UndocumentedPublicProperty`.

schalkms:
@BraisGabin thanks for that. I pushed the wrong branch, since I separated the issue in `UndocumentedPublicClass` from `UndocumentedPublicProperty´.

PR #2591, opened 2020-04-07
Fix false positives in UndocumentedPublicProperty

schalkms:
Properties in classes that are nested, non-public and undocumented are not flagged by this rule anymore.

Related #2580

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
@@ -51,9 +51,25 @@ class UndocumentedPublicPropertySpec : Spek({
             val code = """
                 class Test {
                     /**
-                    *
+                    * Comment
                     */
                     val a = 1
+                }
+            """
+            assertThat(subject.compileAndLint(code)).isEmpty()
+        }
+
+        it("does not report undocumented, public and overridden property in class") {
+            val code = """
+                interface I {
+                    /**
+                     * Comment
+                     */
+                    val a: Int
+                }
+                
+                class Test : I {
+                    override val a = 1
                 }
             """
             assertThat(subject.compileAndLint(code)).isEmpty()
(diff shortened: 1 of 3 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
```diff
@@ -115,6 +131,11 @@ class UndocumentedPublicPropertySpec : Spek({
             assertThat(subject.compileAndLint(code)).isEmpty()
         }
 
+        it("does not report objects") {
+            val code = "object A {}"
+            assertThat(subject.compileAndLint(code)).isEmpty()
+        }
```
BraisGabin, line 137: This is because `UndocumentedPublicClass` reports it, right?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
```diff
@@ -51,14 +51,30 @@ class UndocumentedPublicPropertySpec : Spek({
             val code = """
                 class Test {
                     /**
-                    *
+                    * Comment
                     */
```
BraisGabin, line 55: 🤔 iteresting... should we treat empty as undocumented? Is it easy? If it's not this comment it's out of the scope of this PR anyway.

schalkms: I don't think so. FxCop also doesn't do it.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicPropertySpec.kt
schalkms: This should be removed and belongs to `UndocumentedPublicClass` as you said.

PR #2603, opened 2020-04-14
Correct documentation/recommendation of EmptyCatchBlock rule

Whathecode:
Fixes https://github.com/arturbosch/detekt/issues/2602

The current code smell recommendation for empty catch blocks is wrong:

> Empty blocks of code serve no purpose and should be removed.

Empty catch blocks cannot be removed, since that results in compilation errors.

This PR updates the documentation to give more guidance in line with how the rule is currently implemented. The recommendation seems to be to clarify the exception is intentionally ignored by naming the exception as such.

However, this also raises the question whether this should really be part of the 'empty block' ruleset. To me this is more akin to the `TooGenericExceptionCaught` rule. This is exemplified by the fact that I had to modify the `EmptyRule` class in order to be able to modify the messages. In addition, one could argue this requires a higher severity than the 'minor' severity inherited from `EmptyRule`.


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyCatchBlock.kt
```diff
@@ -7,13 +7,20 @@ import io.gitlab.arturbosch.detekt.rules.isAllowedExceptionName
 import org.jetbrains.kotlin.psi.KtCatchClause
 
 /**
- * Reports empty `catch` blocks. Empty blocks of code serve no purpose and should be removed.
+ * Reports empty `catch` blocks. Empty catch blocks indicate an exception is ignored and not handled.
```
schalkms, line 10:
```suggestion
 * Reports empty `catch` blocks. Empty catch blocks indicate that an exception is ignored and not handled.
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyCatchBlock.kt
```diff
@@ -7,13 +7,20 @@ import io.gitlab.arturbosch.detekt.rules.isAllowedExceptionName
...
+ * In case exceptions are ignored intentionally, this should be made explicit.
```
schalkms, line 11:
```suggestion
 * In case exceptions are ignored intentionally, this should be made explicit 
 * by using the specified names in the `allowedExceptionNameRegex`.
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyCatchBlock.kt
```diff
@@ -7,13 +7,20 @@ import io.gitlab.arturbosch.detekt.rules.isAllowedExceptionName
...
  *
  * @configuration allowedExceptionNameRegex - ignores exception types which match this regex
  * (default: `'^(_|(ignore|expected).*)'`)
  * @active since v1.0.0
  */
-class EmptyCatchBlock(config: Config) : EmptyRule(config = config) {
+class EmptyCatchBlock(config: Config) : EmptyRule(
+    config = config,
+    description = "Empty catch block detected. Empty catch blocks indicate an exception is ignored and not handled.",
```
schalkms, line 19:
```suggestion
    description = "Empty catch block detected. Empty catch blocks indicate that an exception is ignored 
    and not handled.",
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyCatchBlockSpec.kt
```diff
@@ -0,0 +1,35 @@
+package io.gitlab.arturbosch.detekt.rules.empty
+
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.test.assertThat
+import io.gitlab.arturbosch.detekt.test.compileAndLint
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+
+class EmptyCatchBlockSpec : Spek({
```
schalkms, line 9: We already have tests for this scenario, although they are contained in a another spec.
Hence, this class can be safely removed in order to remove the duplicated tests.

https://github.com/arturbosch/detekt/blob/0bd201cd53ef4a329f23806d4f460a45788c3263/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyCodeSpec.kt#L31-L53

PR #2606, opened 2020-04-14
Report class and object violations at the identifier

arturbosch:
Reporting them at the declaration itself will lead to highlighting the whole class/object with kdoc when post-processing the issue results. https://github.com/detekt/detekt-intellij-plugin/issues/81


pinkasey:
@arturbosch  that's great!
Would that work for `TooManyFunctions` in file as well?
For instance, I have several files like this in my codebase:
image


Also,
Can you tell when would this fix be released?

arturbosch:
Please open a new issue for TooManyFunctions. It is reported at the file level. We need to discuss if reporting it at the first line would be sufficient.

In the next days 1.8.0 will be released :)

pinkasey:
sure thing
Thanks!
https://github.com/detekt/detekt-intellij-plugin/issues/83

PR #2607, opened 2020-04-14
Generate the cli as a stand alone executable

### LINKED ISSUE #2605: Distribute cli as an executable file
Extracted from ["Getting started with the CLI"](https://arturbosch.github.io/detekt/cli.html)

> 1. `cd detekt`
> 2. `gradle build`
> 3. `java -jar detekt-cli/build/libs/detekt-cli-[version]-all.jar [parameters]*`

We are forcing to our user to clone the repo.

It would be great to have something like:
1. `curl -sSLO https://github.com/arturbosch/detekt/releases/download/[version]/detekt && chmod a+x detekt`
2. `detekt [parameters]*`

(example taken from ktlint: https://ktlint.github.io/)

To acomplish this we need to generate an executable file with ALL detekt in it.

What do you think? This will allow us to easily distribute detekt with https://brew.sh/, https://chocolatey.org/ and similars. But that's another topic.

BraisGabin:
Fixes #2605

Now if we run `./gradlew shadowJarExecutable` we can execute detekt like this: `./detekt-cli/build/run/detek`. So, if we publish this file in each new release, we can help the users how to install/update detekt.

Problem: I think that this only work on UNIX. I didn't update the documentation for this reason. I don't know how many people uses windows CMD now a days... ktlint ignores this fact completely as you can check here: https://ktlint.github.io/ should we do it too? Or we can say somthing like:

Unix: `./detekt`
Windows: `java -jar detekt`

What do you think?

### DIFF
--- a/detekt-cli/build.gradle.kts
+++ b/detekt-cli/build.gradle.kts
@@ -25,3 +25,23 @@ tasks.withType<Jar> {
         attributes(mapOf("DetektVersion" to detektVersion))
     }
 }
+
+// Implements https://github.com/brianm/really-executable-jars-maven-plugin maven plugin behaviour.
+// To check details how it works, see http://skife.org/java/unix/2011/06/20/really_executable_jars.html.
+// Extracted from https://github.com/pinterest/ktlint/blob/a86d1c76c44d0a1c1adc3f756f36d8b4cac15d32/ktlint/build.gradle#L40-L57
+tasks.register<DefaultTask>("shadowJarExecutable") {
+    description = "Creates self-executable file, that runs generated shadow jar"
+    group = "Distribution"
+
+    inputs.files(tasks.named("shadowJar"))
+    outputs.file("$buildDir/run/detekt")
+
+    doLast {
+        val execFile = outputs.files.singleFile
+        execFile.outputStream().use {
+            it.write("#!/bin/sh\n\nexec java -Xmx512m -jar \"\$0\" \"\$@\"\n\n".toByteArray())
+            it.write(inputs.files.singleFile.readBytes())
+        }
+        execFile.setExecutable(true, false)
+    }
+}
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -388,4 +388,5 @@ githubRelease {
         changelog.trim()
     }
     releaseAssets.setFrom(project(":detekt-cli").buildDir.resolve("libs/detekt-cli-${project.version}-all.jar"))
+    releaseAssets.setFrom(project(":detekt-cli").buildDir.resolve("run/detekt"))
 }


cortinico:
> So, if we publish this file in each new release, we can help the users how to install/update detekt.

Would it be useful to expose the .jar as an artifact of the Github Action run?

BraisGabin:
> Would it be useful to expose the .jar as an artifact of the Github Action run?

I don't know exctly how the github releases work. Or what "expose the .jar" means exactly. But I think that we should post this executable for each release at github. And expose all the jars of a PR and master/snapshots would be great for the early adaopters.

cortinico:
> I don't know exctly how the github releases work. Or what "expose the .jar" means exactly.

Github Actions allows to "attach" one or more artifact for every build.
Example here https://github.com/cortinico/kscript-template/actions/runs/74006424 this run has a binary artifact. 

More context here https://github.com/marketplace/actions/upload-artifact

arturbosch:
Isn't this the same thing the `application` plugin builds us? A zip containing all jars and executables for windows and unix?

Our installation instruction could look like:
```
curl detekt-cli.zip
unzip detekt-cli.zip
alias detekt = <path>/detekt-cli/bin/detekt.[sh|bat]
```

BraisGabin:
Oh! i didn't know about that zip. Yeah. This PR do something similar but not the same. That zip contains an "installation": the binaries and the libs. With this we have a stand alone executable.

I wanted to merge this to make it easy to add detekt to homebrew. If we have that file I don't care that much about this. But anyway I think that for a user that just want to download detekt and use it the stand-alone executable is more handy.

My ultimate idea is to have something like this:

How to install detekt:

- brew (MacOS):
`brew install detekt`
`detekt`
- chocolatey (Windows):
`I have no idea how you install things in chocolatey but some command here`
`detekt`
- Unix:
`curl url/version/detekt && chmod +x detekt`
`./detekt`
- Other option
`curl url/version/detekt.zip && unzip detekt.zip`
`./bin/detekt`

This PR will help with the "Unix" part. But we don't really need that one if you don't like it.

detekt-cli/build.gradle.kts
```diff
@@ -25,3 +25,23 @@ tasks.withType<Jar> {
...
+
+// Implements https://github.com/brianm/really-executable-jars-maven-plugin maven plugin behaviour.
+// To check details how it works, see http://skife.org/java/unix/2011/06/20/really_executable_jars.html.
+// Extracted from https://github.com/pinterest/ktlint/blob/a86d1c76c44d0a1c1adc3f756f36d8b4cac15d32/ktlint/build.gradle#L40-L57
+tasks.register<DefaultTask>("shadowJarExecutable") {
+    description = "Creates self-executable file, that runs generated shadow jar"
+    group = "Distribution"
+
+    inputs.files(tasks.named("shadowJar"))
+    outputs.file("$buildDir/run/detekt")
```
arturbosch, line 37: Please add this path to the `githubRelease` part so we upload this immediately in the next release :)

BraisGabin: Done. I'll create another PR updating the documentation.

ZacSweers:
favor to ask - for github releases, the last few have not had consistent naming, which prevents users from reliable downloading. Could you commit to using a standard naming scheme?

arturbosch:
@ZacSweers you mean the `v` in front of the version?
We began using the github release plugin which has this pattern by default.
I will leave it that way.

Issue #2437, opened 2020-03-16
Plugin development to detekt violations of design principles


Mkohm:
Hi!

As some of you might have seen, i am developing a plugin to detekt called detekt-hint (https://github.com/mkohm/detekt-hint). I am writing my master thesis about detekt and how one could write an extension to detekt design principle violations. The main focus is trying to detect violations of common design principles (SOLID etc.). Because this is hard, i am essentially giving the developer "hints" instead of telling the developer that a design principle is violated. Therefore i am using detekt's ruleset-weight-feature, giving detections a 0 weight, such that hints won't break the build, but instead act as weak warnings. I have a few questions:

1. I am looking for feedback and testing of the plugin and wonder if some of you would like to help me out? If so, please reach out to me! (marius.kohmann@gmail.com)
- See the repo for some samples that is already developed
- Which design principles do you think is of most importance, and that could be "enforced" by detekt rules?
- Would you use such a tool? Why, why not?
- How can i reduce the amount of false-positives to a minimum?
- Should the weak-warnings be long and include possible solutions, or should they be kept it short and concise?

2. Is such a ruleset/feature of interest to detekt when the plugin is more mature? Or would it make sense to still keep as a plugin?

3. Would featuring my plugin in a section in the readme be something of interest? Would gladly create such a PR if it is of interest.

schalkms:
Cool! 🙂👍
I like this plugin. It's always great to see in which ways the community extends detekt and how it can be customized to domain-specific needs.
I already gave it a star. 😉

> I am looking for feedback and testing of the plugin and wonder if some of you would like to help me out? If so, please reach out to me! 

Why don't you integrate your awesome tool to open source repos here on GitHub and have a huge amount of code running and testing your tool?
You could for instance add your CI tool to the CI chain in detekt's main repo.
We would really welcome such an integration. CC @arturbosch 

> Is such a ruleset/feature of interest to detekt when the plugin is more mature? Or would it make sense to still keep as a plugin?

I've laid out my opinion in #1381. We already discussed rules regarding the SOLID principles in detail. Concerning specific rules we encourage the community to submit issues. We've already merged a good number of 3rd party rules into detekt's main ruleset.

> Would featuring my plugin in a section in the readme be something of interest? Would gladly create such a PR if it is of interest.

Yes, we have a separate **integrations** section here:
https://github.com/arturbosch/detekt#mentions

Please go ahead and add your tool to this section.

Mkohm:
[...]

PR #2611, opened 2020-04-16
Integrate with detekt-hint

Mkohm:
I'll try again... Danger is having some trouble with PRs from forks (https://github.com/danger/danger/issues/1103).

As discussed in #2437, here is a PR with the integration of detekt-hint.

I would be very happy if you find any value of it - please let me know if there are any problems or required changes. If you have any feedback, please let me know!

### DIFF
--- a/docs/pages/gettingstarted/groovydsl.md
+++ b/docs/pages/gettingstarted/groovydsl.md
@@ -148,7 +148,7 @@ uses the type `Detekt`.
 task detektFailFast(type: io.gitlab.arturbosch.detekt.Detekt) {
     description = "Runs a failfast detekt build."
     source = files("src/main/java")
-    config = files("$rootDir/config.yml")
+    config.from(files("$rootDir/config.yml"))
     debug = true
     reports {
         xml {


.github/workflows/pre-merge.yaml
```diff
@@ -76,6 +76,13 @@ jobs:
     - name: Verify Generator Output
       run: ./gradlew verifyGeneratorOutput
 
+    - name: Setup Ruby for use with actions
+      uses: actions/setup-ruby@v1.0.0
+
+    - name: Run Danger
+      run: gem install bundler && bundle install && bundle update danger && bundle exec danger --verbose
```
arturbosch, line 83: This will run on every matrix entry. 
detekt-hint could be an own action so it is run just once

Gemfile
```diff
@@ -0,0 +1,7 @@
+source "https://rubygems.org"
+
+git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }
+
+gem 'danger'
+gem 'oga'
+gem "danger-kotlin_detekt", "~> 0.0.3"
```
arturbosch, line 7: Hm, the ruby part looks really generic. 
Could this be a custom GitHub action which bundles the ruby action and the whole danger thing?
We are not a ruby application and I would really like to avoid extra dependencies.

build.gradle.kts
```diff
@@ -358,6 +359,24 @@ val detektAll by tasks.registering(Detekt::class) {
     }
 }
 
+task<Detekt>("detektHint") {
```
arturbosch, line 362: This task is created eagerly. Please use `tasks.register` to make it lazy.

build.gradle.kts
```diff
@@ -358,6 +359,24 @@ val detektAll by tasks.registering(Detekt::class) {
...
+    description = "Running detekt-hint and outputs report for Danger to consume."
+    setSource(analysisDir)
+    config.setFrom(file("$rootDir/config/detekt/detekt-hint.yml"))
```
arturbosch, line 365: Lets add here a `disableDefaultRuleSets` so only detekt-hint gets executed.

config/detekt/detekt-hint.yml
```diff
@@ -0,0 +1,11 @@
+detekt-hint:
+  UseCompositionInsteadOfInheritance:
+    active: true
+    yourUniquePackageName: "io.gitlab.arturbosch"
```
arturbosch, line 4: It should be save to detekt as the fourth package.

Gemfile
Mkohm: I did'nt know, that is a really good idea! Looks like i can create a custom Github action that runs within a docker container, and put the ruby stuff and Danger file in there.

Mkohm:
> Edit: It seems that I need to register somewhere and provide a GitHub token .. ?

If you want a separate bot only for detekt, you need to add a token to the secrets in the detekt-github repository. Downside of doing that is that pull requests from forks will not be able to get comments on their PR's by Danger. To support PR's from forks (which i guess we want to) we can use the "GITHUB_TOKEN"  which is created by Github.  

However, i think the issue in Danger (https://github.com/danger/danger/issues/1103) needs to be resolved for this to work.

Mkohm:
I have addressed your comments @arturbosch, i think it works better this way! 

Unfortunately, due to the bug in danger (https://github.com/danger/danger/issues/1103), detekt-hint will not work on pr's from forks. In that case, the action will succeed (even when it fails due to the linked issue). I hope you still want to help me out by testing the tool :)

PR #2617, opened 2020-04-18
Update documentation

BraisGabin:
This PR updates two things:

- Update how to install detekt. It turns out that [detekt it's already at homebrew](https://formulae.brew.sh/formula/detekt). You can see what they do to install and run detekt here: https://github.com/Homebrew/homebrew-core/blob/master/Formula/detekt.rb
- Instead of show [version] in our code samples show the current version of detekt so the snippets are easier to copy/paste.

This PR is not ready to merge for two reasons:
- In the documentation I refer to the stand-alone executable and the last version of detekt doesn't have it. So we should wait to the next release to merge this or upload the stand-alone executable to the current version. This should be kind of easy.
- How do we do to sync the `gradle.properties>detektVersion` with `docs/_config.yml>detekt_version`? Maybe we can create a mini script and change this from `CONTRIBUTING.md`
> Increment `detektVersion` in `gradle.properties`

to something like:
> `./scripts/changeVersion X.Y.Z`

And that script will update both values at once. Or if we want to keep it simple, just add a new point in the `CONTRIBUTING.md` pointing out that we should update that too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,11 +56,12 @@ Visit [the project website](https://arturbosch.github.io/detekt/) for installati
 #### with the command-line interface
 
 ```shell script
-git clone https://github.com/arturbosch/detekt
-cd detekt
-./gradlew build shadowJar
-java -jar detekt-cli/build/libs/detekt-cli-[version]-all.jar --help
+```sh
+curl -sSLO https://github.com/arturbosch/detekt/releases/download/v[version]/detekt && chmod a+x detekt
+./detekt --help
 ```
+
+You can find [other ways to install detekt here](https://arturbosch.github.io/detekt/cli.html)
 
 #### with Gradle
 
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -109,3 +109,5 @@ description: "Meet detekt, a static code analysis tool for Kotlin."
 # needed for sitemap.xml file only
 url: http://github.com/arturbosch/detekt
 baseurl: /
+
+detekt_version: 1.7.4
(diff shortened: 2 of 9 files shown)


arturbosch:
> This PR updates two things:
> 
>     * Update how to install detekt. It turns out that [detekt it's already at homebrew](https://formulae.brew.sh/formula/detekt). You can see what they do to install and run detekt here: https://github.com/Homebrew/homebrew-core/blob/master/Formula/detekt.rb
> 
>     * Instead of show [version] in our code samples show the current version of detekt so the snippets are easier to copy/paste.
> 
> 
> This PR is not ready to merge for two reasons:
> 
>     * In the documentation I refer to the stand-alone executable and the last version of detekt doesn't have it. So we should wait to the next release to merge this or upload the stand-alone executable to the current version. This should be kind of easy.
> 
>     * How do we do to sync the `gradle.properties>detektVersion` with `docs/_config.yml>detekt_version`? Maybe we can create a mini script and change this from `CONTRIBUTING.md`

In #2621 I've created some tasks to increment the detekt version.
I will look into a task to apply the detekt version to the `docs/_config` and the `selfanalysisVersion` after a release. I hope that I can fully automate the release process soon :).
 
> 
> 
> > Increment `detektVersion` in `gradle.properties`
> 
> to something like:
> 
> > `./scripts/changeVersion X.Y.Z`
> 
> And that script will update both values at once. Or if we want to keep it simple, just add a new point in the `CONTRIBUTING.md` pointing out that we should update that too.

Great work! I will review this tomorrow.

PR #2632, opened 2020-04-24
Fix broken links in CLI related page

natiginfo:
Links were incorrect in https://detekt.github.io/detekt/cli.html.


BraisGabin:
Thanks!

PR #2642, opened 2020-04-28
Integrate with detekt-hint

Mkohm:
The previous PR got messy, so i created a new one only containing the new implementation.

I have now changed the implementation so that PRs from forks should work. Now, the only required files are the config file and the new github action that will run detekt-hint.

### DIFF
--- a/.github/workflows/detekt-hint.yaml
+++ b/.github/workflows/detekt-hint.yaml
@@ -1,9 +1,6 @@
 name: detekt hint
 
 on:
-  push:
-    branches:
-      - master
   pull_request:
     branches:
       - '*'
--- a/.github/workflows/detekt-hint.yaml
+++ b/.github/workflows/detekt-hint.yaml
@@ -1,0 +1,28 @@
+name: detekt hint
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  gradle:
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        jdk: [11]
+    runs-on: ${{ matrix.os }}
+    if: ${{ !contains(github.event.head_commit.message, 'detekt hint skip') }}
+    env:
+      JDK_VERSION:  ${{ matrix.jdk }}
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+
+      - name: Run detekt hint
+        uses: Mkohm/detekt-hint-action@master
+        env:
+          DANGER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/detekt-hint.yaml
+++ b/.github/workflows/detekt-hint.yaml
@@ -22,6 +22,7 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v2
 
+
       - name: Run detekt hint
         uses: Mkohm/detekt-hint-action@v1.1
         with:
--- a/.github/workflows/detekt-hint.yaml
+++ b/.github/workflows/detekt-hint.yaml
@@ -23,6 +23,6 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Run detekt hint
-        uses: Mkohm/detekt-hint-action@master
-        env:
-          DANGER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        uses: Mkohm/detekt-hint-action@v1.1
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/detekt-hint.yaml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ${{ matrix.os }}
+    if: ${{ !contains(github.event.head_commit.message, 'detekt hint skip') }}
+    env:
+      JDK_VERSION:  ${{ matrix.jdk }}
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+
+      - name: Run detekt hint
+        uses: Mkohm/detekt-hint-action@master
```
BraisGabin, line 26: It would be better to define a version here.

.github/workflows/detekt-hint.yaml
```diff
@@ -0,0 +1,28 @@
...
+        env:
+          DANGER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
BraisGabin, line 28: Why is this an env instead of an input?

.github/workflows/detekt-hint.yaml
Mkohm: I agree, will update!

.github/workflows/detekt-hint.yaml
Mkohm: I need the github token inside of the action. I guess i can pass the token as input, and then set the environment variable inside the action. Would that be better?

BraisGabin: Yes, this way you "hide" the implementation detail about docker.

Mkohm:
I have addressed your comments @BraisGabin, please let me know if there is anything else 👍

.github/workflows/detekt-hint.yaml
```diff
@@ -0,0 +1,26 @@
...
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  gradle:
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        jdk: [11]
```
cortinico, line 13: This can probably be simplified to `run-on: ubuntu-latest` as you don't really need a `strategy`/`matrix`.

Mkohm:
Unfortunately, a limitation related to permissions for the GITHUB_TOKEN has appeared. The integration will not work across forks. 

In desperate need of feedback on this tool; Can i create comments on pull-requests as if the integration was working? Manually posting comments with another user, just as if the tool were integrated and working automatically. This way i can see if the tool will provide any real value, and i can continue working on a solution to the above limitation.

BraisGabin:
The limitation that you point out can be fixed if you use a github app. But you will need a backend to handle all of that. And an easy way for the people to install your app in its repo. I have a prove of concept for a similar backend: https://github.com/BraisGabin/seshat

About approving/merging this PR I really don't understand the point of it so I'm waiting for the other maintainers to comment here. That wrote in the previous pr so I suppose that they will can take a better decission.

Mkohm:
> The limitation that you point out can be fixed if you use a github app. But you will need a backend to handle all of that. And an easy way for the people to install your app in its repo. I have a prove of concept for a similar backend: https://github.com/BraisGabin/seshat

Ah nice, maybe i need to have a go at doing that. Do you know how it will work for forks? When forking a repository using a certain github app, will one get a "copy" of the app with the same permissions that were requested during installation in the main repository?

BraisGabin:
The fork will not have that permissions. You will need to install the app in your repo too. Similar as how travis work. BUT if you open a PR to the upstream repo it will work.

Mkohm:
> The fork will not have that permissions. You will need to install the app in your repo too. Similar as how travis work. BUT if you open a PR to the upstream repo it will work.

Nice, that would be a good solution!

arturbosch:
@Mkohm sorry for the late response.
What exactly means "it does not work on forks"? If I or any collaborator opens a PR, does it count as a fork? Can I create a PR with these commits included and detekt-hint will post comments on the PR ?
If this will work for collaborators, I am okay to merge and test this integration ^^

Mkohm:
@arturbosch, no worries!

It means that detekt-hint will not be able to post comments on pull requests if the source-repository is a fork. 

You, with direct access to the repository is able to create pull requests to detekt with the same source and destination repository, and will be able to get comments from detekt-hint.

So in short, it will not work for collaborators, only people with direct access to the repository. It's unfortunate, but i am okay with closing this issue.

arturbosch:
@Mkohm as the integration is now only two files without modifying our build setup, I will merge this and give it a go.

PR #2643, opened 2020-04-28
UnusedPrivateMember: don't report parameters in expect/actual functions

### LINKED ISSUE #2636: UnusedPrivateMembers on expect class
## Expected Behavior
Function members **should not** be reported for `except class`

## Observed Behavior
Functions members are reported for `expect class`

## Steps to Reproduce
// _DetektTest.kt
```kotlin
expect class DetektTest  {
    fun test(i: Int)
}
```
* write `actual` declarations that **uses** the param `i`
* run Detekt on CLI and notice the following report:
`UnusedPrivateMember - [i] at ~/_DetektTest.kt:2:14`

## Context
none

## Your Environment
* Version of detekt used: **1.8.0**
* Version of Gradle used (if applicable): **6.3 stable**
* Operating System and version: **macOS 10.15.4 (19E287)**
* Link to your project (if it's a public repository): https://github.com/4face-studi0/Covid19

t-kameyama:
Fixes #2636

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
@@ -57,11 +57,20 @@ class UnusedPrivateMemberSpec : Spek({
 
     describe("expect functions") {
 
-        it("should not report parameters in expect functions") {
+        it("should not report parameters in expect class functions") {
             val code = """
                 expect class Foo  {
-                    fun test(i: Int)
+                    fun bar(i: Int)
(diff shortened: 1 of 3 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
```diff
@@ -55,6 +55,32 @@ class UnusedPrivateMemberSpec : Spek({
         }
     }
 
+    describe("expect functions") {
+
+        it("should not report parameters in expect functions") {
```
schalkms, line 60: What happens when I just write an `expect fun` not inside a class? Unused parameters should not reported for this case. Maybe we should add a test case for quality and coverage reasons.

PR #2658, opened 2020-05-07
Add style rule for mandatory braces in for loop bodies

### LINKED ISSUE #2652: Rule: Mandatory braces for single-line for loop bodies
I'd like to have the option to enforce braces around for loops with a single statement for the body. This is similar to #365, which added this for `if` statements.

Non-compliant example:
```kotlin
for (i in 0..10)
    println(i)
```

Compliant example:
```kotlin
for (i in 0..10) {
    println(i)
}
````

Compliant example 2:
```kotlin
for (i in 0..10) println(i)
```

daphil19:
This PR closes #2652


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesForLoopsSpec.kt
```diff
@@ -0,0 +1,62 @@
...
+        it("does not report with braces") {
+            val code = """
+            fun test() {
+                for (i in 0..10) {
+                    println(i)
+                }
+            }
+            """
+
+            assertThat(subject.lint(code)).isEmpty()
```
schalkms, line 22: Please use the newer API. This allows to verify the correctness of given code snippets with the help of the Kotlin compiler.

```suggestion
            assertThat(subject.compileAndLint(code)).isEmpty()
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesForLoopsSpec.kt
```diff
@@ -0,0 +1,62 @@
...
+            val code = """
+            fun test() {
+                for (i in 0..10) println(i)
+            }
+            """
+
+            assertThat(subject.lint(code)).isEmpty()
+        }
+
+        it("does report multi-line without braces") {
```
schalkms, line 35:
```suggestion
        it("reports multi-line without braces") {
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesForLoops.kt
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * <compliant>
+ * for (i in 0..10) {
+ *     println(i)
+ * }
+ *
+ * for (i in 0..10) println(i)
+ * </compliant>
+ */
+class MandatoryBracesForLoops(config: Config = Config.empty) : Rule(config) {
```
schalkms, line 36: It would make sense to also lint `while` and `do..while` loops. This adds value to this rule.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesForLoopsSpec.kt
```diff
@@ -0,0 +1,62 @@
+package io.gitlab.arturbosch.detekt.rules.style.optional
+
+import io.gitlab.arturbosch.detekt.test.lint
+import org.assertj.core.api.Assertions.assertThat
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+
+class MandatoryBracesForLoopsSpec : Spek({
```
schalkms, line 8: The spec should contain more tests with edge cases.
- Nested loops (with and without braces)
- for with if-else (with and without braces)

BraisGabin:
Can you add the test for `for (a in b) println(a)` this should not be flagged.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesForLoops.kt
daphil19: Should that be a configuration item for this rule? Or should it be added as an additional rule?

daphil19:
> Can you add the test for `for (a in b) println(a)` this should not be flagged.

I think that the test case starting at line 25 in the Spek file covers this, right?

BraisGabin:
Ups! I didn't see that one 👍🏻

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesForLoops.kt
schalkms: There's no need for a configuration option here. Why would you want to have this rule enforced only for `while` but not for `for` loops? It makes very little sense.
Thus, the rule should flag both `while` and `for` loops.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesForLoopsSpec.kt
```diff
@@ -58,5 +59,101 @@ class MandatoryBracesForLoopsSpec : Spek({
...
+            fun test() {
+                // because if statements are expressions, this code properly prints "Odd" and "Even" 
+                for (i in 0..10) 
+                    if (i % 2 == 0) {
+                        println("Odd")
+                    } else {
+                        println("Even")
+                    }
+            }
+            """
```
schalkms: What happens if you write code like this. Is the violation reported?

```Kotlin
if (i % 2 == 0)
  for (i in 0..10)
    println(i)
else {
  println("Even")
}
```

Issue #2597, opened 2020-04-12
Remove hardcoded default values from rules


BraisGabin:
## Expected Behavior
All the lines of code that we have in our production code are "meaningful".

## Observed Behavior
The second parameter in this functions is never used in production: `valueOrDefault(KEY, "value")` and it's really missleading when you are writing/editing a rule.

## Context
Working on #2498 I found some rules that doesn't have the default value hardcoded in it. And I created commits to fix that (407309756f4886e147eeee233b14d393ed655e8d and e7c0190a0889d05bd4bafa93548a0d058a980391)

But as I continue working in #2498 I see that there are even more rules that have this "issue".

I think that those values are there to make the tests easier. But we should not add any code in production to make the tests easier to write.

My proposal: change the use of `valueOrDefault(key, value)` to `valueString(key)`, `valueBoolean(key)`...

Good thing: This is not a breaking change.

arturbosch:
Those values are the same as in `default-detekt-config`.
If we remove them we need to change the hard dependency on the default-detekt-config or else a programmatically created Rule would'nt be the same as loaded with `--build-upon-default-config`.
Do I miss something?

BraisGabin:
So, for what you are saying, this is wrong, right:

https://github.com/arturbosch/detekt/blob/c5aa152cedb5bf105c2757dda8f4f5e5cf34c405/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt#L66-L77

We should have something like this:

```kotlin
private val ignoredExceptionTypes = SplitPattern(valueOrDefault(IGNORED_EXCEPTION_TYPES,  
  "InterruptedException,NumberFormatException,ParseException,MalformedURLException"))
```

I was confused here because I saw a lot of rules that have this problem. So I didn't know if the default value could be removed or we should fix the rules that doesn't have the proper default values.

arturbosch:
Ah, yes good find. Indeed these properties need the defaults from the configuration documentation.
I would classify this as a bug as users without a config file would not get any issues from these rules.

juliabirkett:
@arturbosch can you share this config documentation so I can take a look at the expected default values?

BraisGabin:
I think that artur refers about this documentation: https://detekt.github.io/detekt/comments.html

juliabirkett:
guys, I just opened my very first PR contributing to Open Source. if you please can take a look and see if it solves this issue 👍

PR #2661, opened 2020-05-08
Add default values to SwallowedException rule

juliabirkett:
There are some rules that have default values [in the documentation](https://detekt.github.io/detekt/exceptions.html) but those values are not implemented in the code.
This is a bug mapped here #2597

This PR adds default values to `SwallowedException` rule as mapped in the documentation.

---

⚠️  _This is my very first contribution to OS so sorry if I'm doing anything wrong. If this PR is OK, I'm gonna add more default values to other rules._

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt
@@ -78,10 +78,10 @@ class SwallowedException(config: Config = Config.empty) : Rule(config) {
         Debt.TWENTY_MINS)
 
     private val defaultIgnoredExceptions = listOf(
-            DEFAULT_IGNORED_EXCEPTION_TYPES.INTERRUPTED_EXCEPTION,
-            DEFAULT_IGNORED_EXCEPTION_TYPES.NUMBER_FORMAT_EXCEPTION,
-            DEFAULT_IGNORED_EXCEPTION_TYPES.PARSE_EXCEPTION,
-            DEFAULT_IGNORED_EXCEPTION_TYPES.MALFORMED_URL_EXCEPTION
+            DefaultIgnoredExceptionTypes.INTERRUPTED_EXCEPTION,
+            DefaultIgnoredExceptionTypes.NUMBER_FORMAT_EXCEPTION,
+            DefaultIgnoredExceptionTypes.PARSE_EXCEPTION,
+            DefaultIgnoredExceptionTypes.MALFORMED_URL_EXCEPTION
     ).map { it.toString() }
     private val ignoredExceptionTypes = valueOrDefaultCommaSeparated(IGNORED_EXCEPTION_TYPES, defaultIgnoredExceptions)
             .map { it.removePrefix("*").removeSuffix("*") }
@@ -167,7 +167,7 @@ class SwallowedException(config: Config = Config.empty) : Rule(config) {
         const val ALLOWED_EXCEPTION_NAME_REGEX = "allowedExceptionNameRegex"
     }
 
-    enum class DEFAULT_IGNORED_EXCEPTION_TYPES {
+    enum class DefaultIgnoredExceptionTypes {
         NUMBER_FORMAT_EXCEPTION {
             override fun toString() = "NumberFormatException"
         },
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt
@@ -77,7 +77,7 @@ class SwallowedException(config: Config = Config.empty) : Rule(config) {
         "The caught exception is swallowed. The original exception could be lost.",
         Debt.TWENTY_MINS)
 
-    private val ignoredExceptionTypes = valueOrDefaultCommaSeparated(IGNORED_EXCEPTION_TYPES, emptyList())
+    private val ignoredExceptionTypes = valueOrDefaultCommaSeparated(IGNORED_EXCEPTION_TYPES, listOf("NumberFormatException"))
         .map { it.removePrefix("*").removeSuffix("*") }
 
     private val allowedExceptionNameRegex by LazyRegex(ALLOWED_EXCEPTION_NAME_REGEX, ALLOWED_EXCEPTION_NAME)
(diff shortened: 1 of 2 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt
```diff
@@ -77,8 +77,14 @@ class SwallowedException(config: Config = Config.empty) : Rule(config) {
...
-    private val ignoredExceptionTypes = valueOrDefaultCommaSeparated(IGNORED_EXCEPTION_TYPES, emptyList())
-        .map { it.removePrefix("*").removeSuffix("*") }
+    private val defaultIgnoredExceptions = listOf(
+            DefaultIgnoredExceptionTypes.INTERRUPTED_EXCEPTION,
+            DefaultIgnoredExceptionTypes.NUMBER_FORMAT_EXCEPTION,
+            DefaultIgnoredExceptionTypes.PARSE_EXCEPTION,
+            DefaultIgnoredExceptionTypes.MALFORMED_URL_EXCEPTION
+    ).map { it.toString() }
+    private val ignoredExceptionTypes = valueOrDefaultCommaSeparated(IGNORED_EXCEPTION_TYPES, defaultIgnoredExceptions)
+            .map { it.removePrefix("*").removeSuffix("*") }
```
BraisGabin, line 87: I think that a hardcoded list of strings is cleaner here. We are not going to use the enum for anything else so just the list is more simple.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedExceptionSpec.kt
```diff
@@ -202,5 +202,55 @@ class SwallowedExceptionSpec : Spek({
...
+            """
+            assertThat(subject.compileAndLint(code)).isEmpty()
+        }
+
+        it("ignores MalformedURLException by default") {
+            val code = """
+                fun f() {
+                    try {
+                    } catch (e: MalformedURLException) {
+                        throw MyCustomException()
```
BraisGabin, line 247: Why do you use a Custom Exception instead of a generic one? like `IllegalArgumentException` or `Exception`?

juliabirkett: I used a custom exception based on the example we have [here in the documentation](https://detekt.github.io/detekt/exceptions.html), but I can change it. What do you think?

BraisGabin: That have sense... But I think that using a generic one will be cleaner because the CustomException doesn't provide too much. And we align with other tests in the same file like this one: 
https://github.com/detekt/detekt/blob/34a7628daf5aa59b0f0a4ae9ab3f4e50a825c885/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedExceptionSpec.kt#L14-L24

juliabirkett: OK, agree with you. Gonna change it!

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt
schalkms: The configuration option and not the code itself should be adapted. You might ask why?
If a detekt user doesn't want to have `NumberFormatException` as an ignored exception, the behavior can't be changed. Hence, the following lines should be touched. The code can stay as it is.

https://github.com/detekt/detekt/blob/c8890fa5133154eec4840260fd61281d8d43e919/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt#L66-L70

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedExceptionSpec.kt
```diff
@@ -202,5 +202,53 @@ class SwallowedExceptionSpec : Spek({
...
+            val code = """
+                fun f() {
+                    try {
+                    } catch (e: MalformedURLException) {
+                        throw Exception()
+                    }
+                }
+            """
+            assertThat(subject.compileAndLint(code)).isEmpty()
+        }
```
schalkms, line 252: It's not necessary to test every entry of the configuration. Adapting the config without test every member is fine. Therefore, the tests can be safely removed.

juliabirkett: Why do you think the tests are unnecessary? IMHO, if anyone changes the implementation code by removing any of the default values, the tests should be there to warn us that something has change. Furthermore, one of the reasons I write tests is because they also serve as documentation.

schalkms: What if the configuration options suddenly contains 50 exceptions per default? Would you write a unit test for each and every one? With this test you want to test, whether the specified exception in the configuration option actually gets ignored.
Regarding the documentation aspect, detekt has an automatically generated documentation on the website.

juliabirkett: If the configuration options contains 50 exceptions by default, I'd use some abstractions to be able to test all of them. This doesn't mean I'd create 50 test cases.
I don't agree in deleting the tests but I'll do it just to avoid ping-pong opinions.

arturbosch: Can we simplify this code by using dynamic tests?
I was thinking of something like:
```kotlin
arrayOf("ParseException", "...").forEach { exceptionType ->
        it("ignores $exceptionType by default") {
            val code = """
                fun f() {
                    try {
                    } catch (e: $exceptionType) {
                        throw Exception()
                    }
                }
            """
        }
}
```

juliabirkett: Sure! That's a good way to abstract the code. I thought it wasn't necessary right now because we only have 5 of them. Will do it!

arturbosch:
@juliabirkett @schalkms I've added a commit to get this into the next release soon :)

Issue #2656, opened 2020-05-07
ClassNaming rule treating Kotlin syntax ` as part of class name


hqzxzwb:
## Expected Behavior
ClassNaming rule should only check class name and should not check syntax \`

## Observed Behavior
ClassNaming rule treating Kotlin syntax \` as part of class name

## Steps to Reproduce
```Kotlin
class `$Test$`
```
The above code is warned about a ClassNaming issue with default config 
```
  ClassNaming:
    active: true
    excludes: ['**/test/**', '**/androidTest/**', '**/*.Test.kt', '**/*.Spec.kt', '**/*.Spek.kt']
    classPattern: '[A-Z$][a-zA-Z0-9$]*'
```
When I change the classPattern to ``[`A-Z$][`a-zA-Z0-9$]*``, the warning is gone.
This means detekt is treating the syntax ` as part of class name, which is probably not intended.

## Context
Class names with $ in it should pass the test, otherwise you would not write it to the default pattern. But it's not possible to have such a class name without the \` syntax.

## Your Environment

* Version of detekt used: 1.8.0
* Version of Gradle used (if applicable): 5.6.4
* Operating System and version: macOS Catalina
* Link to your project (if it's a public repository):

arturbosch:
We could ignore names with this syntax altogether or implement this as an option.
Using this syntax is quite intentional, isn't it ?

hqzxzwb:
Yes it is quite intentional. Personally I use it to create special class names when I do not want anyone to use the class but it has to be public for some other reason.

BraisGabin:
I just realised. Why do we allow classes with `$` in the name? It has little sense. The `$` is usually used by the annotation processors to create new classes. And they use that symbol to get a name that is not going to conflict with the user class names. Our tests didn't check the use of `$` neighter. And you can't use te symbol `$` if you don't use the char `. I think that it's safe to remove it.

hqzxzwb:
So remove $ from default pattern and add an option to ignore class names in \`?

BraisGabin:
I created a PR removing the `$` because right now you can't use it even with the current pattern. And I think that we don't need an option here. If you want to allow that case you can use this pattern: ``([A-Z][a-zA-Z0-9]*)|(`.*`)``

In my opinion the use of ` in the class names should not be "encoraged". If we add an option allowing that we are kind of econraging it. If you have a reason to use ` in your names you can do it with the previous pattern. What do you thing?

[...]

PR #2662, opened 2020-05-08
Remove $ as a valid char in class and method names

BraisGabin:
This issue was spotted in #2656 (but it's not completely related with that issue).

Extracted from the issue:

> I just realised. Why do we allow classes with `$` in the name? It has little sense. The `$` is usually used by the annotation processors to create new classes. And they use that symbol to get a name that is not going to conflict with the user class names. Our tests didn't check the use of `$` neighter. And you can't use te symbol `$` if you don't use the char `. I think that it's safe to remove it.

I can't even write tests for this change because I can't create a name of a class with `$` if I'm not using `. And the use of ` in class names is exactly what is being talked in #2656. Right now we don't allow it.

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -327,7 +327,7 @@ naming:
   ClassNaming:
     active: true
     excludes: ['**/test/**', '**/androidTest/**', '**/*.Test.kt', '**/*.Spec.kt', '**/*.Spek.kt']
-    classPattern: '[A-Z$][a-zA-Z0-9$]*'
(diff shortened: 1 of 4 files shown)


BraisGabin:
I didn't think of that. Yeah, I'm going to check if we have the same bug in the other ones

BraisGabin:
Updated.

BraisGabin:
> Can variables and parameters also have **``** in the name?

I think that I'm missing something. This PR is not related with the use of `. It just removes `$` as a valid char in the function and method names.

In the parameters and variables we wasn't allowing that char so there's nothing to change there.

schalkms:
Thanks for the clarification! @BraisGabin

PR #2667, opened 2020-05-09
Include Kotlin multiplatform test folders to default exclude config

### LINKED ISSUE #2608: Include Kotlin multiplatform test folders to default exclude configuration
The [default-detekt-config.yml](https://github.com/arturbosch/detekt/blob/master/detekt-cli/src/main/resources/default-detekt-config.yml) has several default 'excludes' configurations to exclude test sources.

`excludes: '**/test/**,**/androidTest/**,**/*.Test.kt,**/*.Spec.kt,**/*.Spek.kt'`

However, this does not cover test folders in Kotlin multiplatform projects, which are named `**/<platform-target>Test/**`, e.g., `**/jsTest/**` and `**/commonTest/**`. I reuse the majority of defaults, but using detekt in a multiplatform project requires me [to override all default excludes with my own](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/detekt.yml):

`excludes: '**/commonTest/**,**/jvmTest/**,**/jsTest/**,**/test/**'`

Therefore, it would be nice to either add the default multiplatform test folders to `default-detekt-config.yml`, or alternatively, to override the 'default' excludes in one place.

schalkms:
Closes #2608

### DIFF
--- a/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/TestExclusions.kt
+++ b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/TestExclusions.kt
@@ -9,7 +9,7 @@ import io.gitlab.arturbosch.detekt.generator.collection.Rule
 object TestExclusions {
 
     @Suppress("MaxLineLength")
-    const val pattern = "['**/test/**', '**/androidTest/**', '**/*.Test.kt', '**/*.Spec.kt', '**/*.Spek.kt', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']"
+    const val pattern = "['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**', '**/*.Test.kt', '**/*.Spec.kt', '**/*.Spek.kt']"
     val ruleSets = setOf("comments")
     val rules = setOf(
         "NamingRules",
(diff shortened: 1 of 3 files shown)


detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/TestExclusions.kt
```diff
@@ -8,7 +8,8 @@ import io.gitlab.arturbosch.detekt.generator.collection.Rule
  */
 object TestExclusions {
 
-    const val pattern = "['**/test/**', '**/androidTest/**', '**/*.Test.kt', '**/*.Spec.kt', '**/*.Spek.kt']"
+    @Suppress("MaxLineLength")
+    const val pattern = "['**/test/**', '**/androidTest/**', '**/*.Test.kt', '**/*.Spec.kt', '**/*.Spek.kt', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']"
```
BraisGabin, line 12: nit. What do you think about this sorting?

```suggestion
    const val pattern = "['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**', '**/*.Test.kt', '**/*.Spec.kt', '**/*.Spek.kt']"
```

schalkms: DONE

PR #2670, opened 2020-05-09
Add to do-while check to MandatoryBracesLoops rule

schalkms:


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesLoopsSpec.kt
```diff
@@ -245,4 +255,118 @@ class MandatoryBracesLoopsSpec : Spek({
...
+            """
+
+            assertThat(subject.compileAndLint(code)).isEmpty()
+        }
+
+        it("does not report nested loops on single line") {
+            val code = """		
+            fun test() {		
+                var i = 0
+                do do i += 1 while(i < 5) while (i < 5)
```
BraisGabin, line 330: 😂

schalkms: I won't have to write code like this again, hopefully... 😂

PR #2671, opened 2020-05-10
GuardCluase also matches if-with-body that contains a return

pinkasey:
@arturbosch 
I thought opening a PR was better than opening an Issue.
Although, I didn't properly test it yet - I'm new to gradle and Spek.
If you prefere, I can open an issue instead...

What I want -
to consider an `if` with a body as guard clause as well.
I've added a test-case to demonstrate.

In my company, we almost disabled `ReturnCount` because it catches guard-clauses, and then I saw guard-clauses - and was disappointed it applies only to `if`s without a body.
Our code-style convention don't allow if's without body.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
@@ -7,6 +7,7 @@ import org.jetbrains.kotlin.psi.KtIfExpression
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtReturnExpression
 import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
+import org.jetbrains.kotlin.psi.psiUtil.lastBlockStatementOrThis
 
 fun KtNamedFunction.yieldStatementsSkippingGuardClauses(): Sequence<KtExpression> = sequence {
     var firstNonGuardFound = false
@@ -24,9 +25,12 @@ fun KtNamedFunction.yieldStatementsSkippingGuardClauses(): Sequence<KtExpression
 
 fun KtExpression.isGuardClause(): Boolean {
 
-    fun KtReturnExpression.isIfConditionGuardClause(ancestorExpression: KtExpression): Boolean {
-        val ifExpr = ancestorExpression as? KtIfExpression
-        return ifExpr != null && ifExpr.`else` == null
+    fun isIfConditionGuardClause(returnExpr: KtReturnExpression): Boolean {
+        val ifExpr = this as? KtIfExpression
+            ?: return false
+
+        return ifExpr.`else` == null &&
+                returnExpr === ifExpr.then?.lastBlockStatementOrThis()
     }
 
     fun KtReturnExpression.isElvisOperatorGuardClause(): Boolean {
@@ -35,5 +39,5 @@ fun KtExpression.isGuardClause(): Boolean {
     }
 
     val returnExpr = this.findDescendantOfType<KtReturnExpression>() ?: return false
-    return returnExpr.isIfConditionGuardClause(this) || returnExpr.isElvisOperatorGuardClause()
+    return isIfConditionGuardClause(returnExpr) || returnExpr.isElvisOperatorGuardClause()
 }
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
@@ -24,8 +24,8 @@ fun KtNamedFunction.yieldStatementsSkippingGuardClauses(): Sequence<KtExpression
 
 fun KtExpression.isGuardClause(): Boolean {
 
-    fun KtReturnExpression.isIfConditionGuardClause(): Boolean {
-        val ifExpr = this.parent?.parent as? KtIfExpression
+    fun KtReturnExpression.isIfConditionGuardClause(ancestorExpression: KtExpression): Boolean {
+        val ifExpr = ancestorExpression as? KtIfExpression
         return ifExpr != null && ifExpr.`else` == null
     }
 
@@ -35,5 +35,5 @@ fun KtExpression.isGuardClause(): Boolean {
     }
 
     val returnExpr = this.findDescendantOfType<KtReturnExpression>() ?: return false
-    return returnExpr.isIfConditionGuardClause() || returnExpr.isElvisOperatorGuardClause()
+    return returnExpr.isIfConditionGuardClause(this) || returnExpr.isElvisOperatorGuardClause()
 }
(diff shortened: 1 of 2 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
```diff
@@ -29,6 +29,28 @@ class ReturnCountSpec : Spek({
             }
         }
 
+        context("a file with an if condition guard clause with body and 2 returns") {
+            val code = """
+            fun test(x: Int): Int {
+                if (x < 4) {
```
pinkasey, line 35: I want this to count as a GuardClause.
We have many of these in our code.

In fact, many consider it a bad practice to have an `if` with no body, so they'll never enjoy the benefit of the `GuardClause` feature.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
```diff
@@ -24,8 +24,8 @@ fun KtNamedFunction.yieldStatementsSkippingGuardClauses(): Sequence<KtExpression
 
 fun KtExpression.isGuardClause(): Boolean {
 
-    fun KtReturnExpression.isIfConditionGuardClause(): Boolean {
-        val ifExpr = this.parent?.parent as? KtIfExpression
+    fun KtReturnExpression.isIfConditionGuardClause(ancestorExpression: KtExpression): Boolean {
```
cortinico, line 27:
```suggestion
    fun isIfConditionGuardClause(ancestorExpression: KtExpression): Boolean {
```

This can probably be refactored to not be an extension function?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
```diff
@@ -29,6 +29,28 @@ class ReturnCountSpec : Spek({
...
+                when (x) {
+                    5 -> return 5
+                    4 -> return 4
+                }
+            }
+        """
+
+            it("should not get flagged for if condition guard clauses") {
+                val findings = ReturnCount(TestConfig(mapOf(ReturnCount.EXCLUDE_GUARD_CLAUSES to "true")))
+                    .lint(code)
```
cortinico, line 48:
```suggestion
                    .compileAndLint(code)
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
pinkasey: I actually like it that way,
and anyway I don't feel comfortable to make that change to someone else's code 🙂

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
pinkasey: @cortinico 
Are you sure?
I've copy-pasted this test from above test, and I don't see any other test in this class that uses `compileAndLint`

schalkms: `compileAndLint` is the newer API, which also asserts the correctness of the sample code.
It should be preferred over `lint` if possible.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
schalkms: +1

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
pinkasey: Awesome thanks.
Good to know.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
arturbosch: @cortinico is right here. As you are not using the receiver inside the nested method is it unused.
I would propose an additional change which checks if the return statement is actually the last statement inside the if:

```kotlin
fun KtExpression.isGuardClause(): Boolean {

    fun isIfConditionGuardClause(returnExpr: KtReturnExpression): Boolean {
        val ifExpr = this as? KtIfExpression
        return ifExpr != null &&
            ifExpr.`else` == null &&
            returnExpr === ifExpr.then?.lastBlockStatementOrThis()
    }

    fun KtReturnExpression.isElvisOperatorGuardClause(): Boolean {
        val elvisExpr = this.parent as? KtBinaryExpression
        return elvisExpr != null && elvisExpr.operationToken == KtTokens.ELVIS
    }

    val returnExpr = this.findDescendantOfType<KtReturnExpression>() ?: return false
    return isIfConditionGuardClause(returnExpr) || returnExpr.isElvisOperatorGuardClause()
}

```

pinkasey: right
I didn't notice that I no longer use the receiver 😕 
I'll change the code

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
```diff
@@ -89,7 +151,10 @@ class ReturnCountSpec : Spek({
                     if(a == null) return
                     val models = b as? Int ?: return
                     val position = c?.takeIf { it != -1 } ?: return
-                    if(b !is String) return
+                    if(b !is String) {
```
pinkasey, line 154: I thought -
why not include one with braces too

pinkasey:
tests fail because my example code (which i copy-pasted from previous it) - did not compile.
I'll fix that, and the rest of the cases while I'm at it.
please wait with review...

pinkasey:
OK
I've fixed my tests,
migrated all tests but the last one from `lint()` to `compileAndLint()`.
Hope checks pass this time.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
```diff
@@ -16,33 +16,99 @@ class ReturnCountSpec : Spek({
...
         context("a file with an ELVIS operator guard clause and 2 returns") {
             val code = """
             fun test(x: Int): Int {
                 val y = x ?: return 0
                 when (x) {
-                    5 -> return 5
+                    5 -> println("x=5")
                     4 -> return 4
                 }
+                return 6
```
pinkasey, line 105: These code-snippets didn't compile, because there wasn't a `return` in every execution path.

pinkasey:
> Thanks!

Thanks @arturbosch 
for the good comments and support!

PR #2672, opened 2020-05-10
Use configuration avoidance for withType usages

arturbosch:


arturbosch:
@runningcode thanks for pointing out the new API!

PR #2679, opened 2020-05-12
Introduce metrics module / implement cognitive complexity

arturbosch:
Unlike Cyclomatic complexity, which is a measurement for the testability of a code unit, cognitive complexity is a measurement for the maintainability of a code unit.
It tells the user how difficult a piece of code is to read and understand.

Reference used: https://www.sonarsource.com/docs/CognitiveComplexity.pdf

This is the first step towards a metric module.
Other steps are:

- Moving the processors of `detekt-core` which compute a `ProjectMetric` into detekt-metrics
- Including cognitive complexity into our complexity report
- (To discuss) Allowing to measure cognitive instead of cyclomatic complexity for the `ComplexMethod` rule (A non breaking way for 1.X.X would be to implement the switch via a config property `algorithm: cognitive`)


schalkms:
I'll review this tomorrow.

detekt-metrics/src/main/kotlin/io/github/detekt/metrics/CognitiveComplexity.kt
```diff
@@ -0,0 +1,188 @@
...
+        private fun calculateBinaryExprComplexity(usedExpr: List<BinExprHolder>) {
+            var lastOp: IElementType? = null
+            for (binExpr in usedExpr) {
+                if (lastOp == null) {
+                    complexity++
+                } else {
+                    if (lastOp != binExpr.op || binExpr.isEnclosed) {
+                        complexity++
+                    }
+                }
```
schalkms, line 172:
```suggestion
                } else if (lastOp != binExpr.op || binExpr.isEnclosed) {
                    complexity++
                }
```

detekt-metrics/src/test/kotlin/io/github/detekt/metrics/CognitiveComplexitySpec.kt
```diff
@@ -0,0 +1,231 @@
...
+            assertThat(CognitiveComplexity.calculate(code)).isEqualTo(2)
+        }
+
+        it("adds nesting for nested functions but not complexity") {
+            val code = compileContentForTest("""
+                fun main() { fun run() { if (true) {} } }
+            """)
+
+            assertThat(CognitiveComplexity.calculate(code)).isEqualTo(2)
+        }
```
schalkms, line 136: It's a bit hard to understand why these two have the mentioned complexity value.

schalkms: Maybe you can enlighten me. ;-)

detekt-metrics/src/test/kotlin/io/github/detekt/metrics/CognitiveComplexitySpec.kt
```diff
@@ -0,0 +1,231 @@
...
+                                !cond 
+                                && !(cond && cond && cond)  // +2
+                            ) {}
+                        }
+                    """)
+
+                    assertThat(CognitiveComplexity.calculate(code)).isEqualTo(3)
+                }
+
+                it("adds one for negated like operator but only once") {
```
schalkms, line 215: This has the same test description as above. If I recall it correctly, the Spek framework had problem with that in the past.

schalkms:
> (To discuss) Allowing to measure cognitive instead of cyclomatic complexity for the ComplexMethod rule (A non breaking way for 1.X.X would be to implement the switch via a config property algorithm: cognitive)

I'd leave this for now as it is and wait for feedback. This is maybe a topic for detekt > v2.

detekt-metrics/src/test/kotlin/io/github/detekt/metrics/CognitiveComplexitySpec.kt
arturbosch: Cognitive complexity adds +1 for each nesting level. Lambdas and nested function increase the nesting level but do not add complexity.
In this cases we have an `if` which would normally add +1 but due to nesting level 1, it adds +2.
hope that helps

detekt-metrics/src/test/kotlin/io/github/detekt/metrics/CognitiveComplexitySpec.kt
arturbosch: Thx! fixed.

PR #2687, opened 2020-05-14
Fix serialVersionUID false positive in UnderscoresInNumericLiteral

### LINKED ISSUE #2686: False positive UnderscoresInNumericLiterals in Serializable
## Expected Behavior
This is valid code according to docs:

    class Office(
        val id: Long,
        val address: String,
        val phone: String
    ) : Serializable {
        companion object {
            private const val serialVersionUID = -43857148126114372L
        }
    }

## Observed Behavior
UnderscoresInNumericLiterals reported for `serialVersionUID` field.

## Your Environment
* Version of detekt used: 1.9.0
* Version of Gradle used (if applicable): 6.4

schalkms:
This fixes the cases where a serialVersionUID
1. has either a negative number
2. is inside a companion object
3. or both

Closes #2686

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiteralsSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiteralsSpec.kt
@@ -231,6 +231,54 @@ class UnderscoresInNumericLiteralsSpec : Spek({
         }
     }
 
+    describe("a property named serialVersionUID in a companion object inside a serializable class") {
+
+        it("does not report a negative serialVersionUID number") {
+            val code = """
+                class Test : Serializable {
+                    companion object {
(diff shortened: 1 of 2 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiteralsSpec.kt
```diff
@@ -231,6 +231,54 @@ class UnderscoresInNumericLiteralsSpec : Spek({
...
+            assertThat(findings).hasSize(0)
+        }
+    }
+
+    describe("a property named serialVersionUID number in a serializable class") {
+
+        it("does not report a negative serialVersionUID number") {
+            val code = """
+                class Test : Serializable {
+                    private const val serialVersionUID = -43857148126114372L
```
BraisGabin, line 264: Could you use `compileAndLint`? I think that this code doesn't compile. You can't use `const` inside a `class`.

PR #2690, opened 2020-05-14
Add -v shortcut for version flag

arturbosch:
`-v` is a commonly used shortcut for --version in many cli tools.


BraisGabin:
I don't agree with this shortcut. `-v` means `verbose` in nearly all the command line apps.

Right now we don't have a `verbose` mode but there where some issues talking about it and so it's not crazy that at some point we will want to have it. So I'd keep the `-v` flag free for that reason and only use it if we implement the `verbose` mode.

PR #2695, opened 2020-05-15
Inline unreachable code

schalkms:


BraisGabin:
Checking this again... can be we losing the no-report test in this refactor. I mean in this snippet:

```

    fun throw1(p: Int): Int {
        if (p == 0) {
            throw IllegalArgumentException()
            println() // report 1 - unreachable code
        }
        throw IllegalArgumentException()
    }
```

We could be testing that the first `throw` creates unrechable code but the second don't. All at the same snippet...

schalkms:
@BraisGabin I'll create another PR that includes this scenario. I saw your comment too late.

PR #2242, opened 2020-01-10
Add rule to warn on ignored return value

bbaldino:
Addresses #2239.

I'd consider this as draft at this point.  It works for the test cases I've provided, but I'm not familiar with all of the psi element types and I worry a bit that there may be some I missed (or perhaps a better way to detect what I am detecting).

PR #2698, opened 2020-05-15
New Rule: IgnoredReturnValue

### LINKED ISSUE #209: CheckReturnValue Rule
Given a method that is annotated with `CheckReturnValue` or `CheckResult` and you call it without storing the value it should be flagged. We don't even need to resolve the imports, simply checking for the annotation name is sufficient.

Positive case:

```kotlin
interface Something {
  @CheckReturnValue fun foo(): Int
}

fun something(something: Something) {
  something.foo()
}
```

Negative case:

```kotlin
interface Something {
  @CheckReturnValue fun foo(): Int
}

fun something(something: Something): Int {
  val foo = something.foo()
  return foo + foo
}
```

### LINKED ISSUE #2239: New rule: Warn on ignored return value
The rule or changes you're looking for might have already been suggested!
Please search in the [issues](https://github.com/arturbosch/detekt/issues) before creating a new one.

If a method which returns a value is called and the return value is ignored, detekt should issue a warning.

Example:
```
fun Byte.putBit(bitPos: Int, isSet: Boolean): Byte {
    return if (isSet) {
        (this.toInt() or (0b10000000 ushr bitPos)).toByte()
    } else {
        (this.toInt() and (0b10000000 ushr bitPos).inv()).toByte()
    }
}
```
It would be easy for someone to call `myByte.putBit(1, true)` and think it would be modified in place, but actually it returns the new value.  A failure to handle this return value will almost certainly result in a bug.

## Context
In many (most?) cases, the failure to observe a value returned by a function will result in a bug, but it can be an easy thing for a developer to forget.  A warning will help catch instances where the return value should be handled.  Cases which don't fall under this condition can suppress the warning (which also serves to explicitly document that it was intended).

I plan on taking a look at implementing this rule myself when I get a chance.

cortinico:
I've cherry-picked @bbaldino commits and fixed the tests for the new rule `IgnoredReturnValue`.

I've tried this approach that checks if a `PsiElement` is "isolated" and the return value is not `Unit`. I'm not sure it's the correct one but from the tests it seems promising.

I'd be happy to receive further ideas for test cases or improvements.

Closes #209 
Closes #2239
Closes #2242

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
@@ -7,15 +7,20 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
-import org.jetbrains.kotlin.psi.KtBinaryExpression
+import org.jetbrains.kotlin.com.intellij.psi.PsiComment
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.com.intellij.psi.PsiWhiteSpace
+import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
+import org.jetbrains.kotlin.lexer.KtSingleValueToken
 import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
 import org.jetbrains.kotlin.types.typeUtil.isUnit
 
 /**
  * The Kotlin compiler gives no warning for when a function which returns a value is called but its returned
- * value is ignored.  This rule warns on instances where a function returns a value but that value is not
+ * value is ignored. This rule warns on instances where a function returns a value but that value is not
  * used in any way.
  *
  * fun returnsValue() = 42
@@ -34,7 +39,7 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
     override val issue: Issue = Issue(
         "IgnoredReturnValue",
         Severity.Defect,
-        "This call returns a value which has been ignored",
+        "This call returns a value which is ignored",
         Debt.TWENTY_MINS
     )
 
@@ -45,12 +50,45 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
         if (bindingContext == BindingContext.EMPTY) return
         val resolvedCall = expression.getResolvedCall(bindingContext) ?: return
         val returnType = resolvedCall.resultingDescriptor.returnType ?: return
+
         if (returnType.isUnit()) {
             return
         }
-        // If this is part of a binary expression, then its return value is being checked
-        if (expression.parent is KtBinaryExpression) return
 
-        report(CodeSmell(issue, Entity.from(expression), ""))
+        val elementsToInspect = mutableListOf<PsiElement>(expression)
+        if (expression.parent is KtDotQualifiedExpression) {
+            elementsToInspect += expression.parent
+        }
+
+        if (elementsToInspect.any(PsiElement::isIsolated)) {
+            report(
+                CodeSmell(
+                    issue,
+                    Entity.from(expression),
+                    message = "The call ${expression.text} is returning a value that is ignored."
+                )
+            )
+        }
     }
 }
+
+private val PsiElement.isIsolated: Boolean
+    get() =
+        true == this.prevSibling?.isAnIsolationElement && true == this.nextSibling?.isAnIsolationElement
+
+private val PsiElement?.isAnIsolationElement: Boolean
+    get() {
+        if (this is PsiWhiteSpace) {
+            return true
+        }
+        if (this is PsiComment) {
+            return true
+        }
+        if (this is LeafPsiElement && this.elementType is KtSingleValueToken) {
+            val token = (this.elementType as KtSingleValueToken)
+            if (token.value == ";") {
+                return true
+            }
+        }
+        return false
+    }
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
@@ -1,0 +1,56 @@
+package io.gitlab.arturbosch.detekt.rules.bugs
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.psi.KtBinaryExpression
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.types.typeUtil.isUnit
+
+/**
+ * The Kotlin compiler gives no warning for when a function which returns a value is called but its returned
+ * value is ignored.  This rule warns on instances where a function returns a value but that value is not
+ * used in any way.
+ *
+ * fun returnsValue() = 42
+ * fun returnsNoValue() {}
+ *
+ * <noncompliant>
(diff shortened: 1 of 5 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
```diff
@@ -0,0 +1,94 @@
...
+private val PsiElement?.isAnIsolationElement: Boolean
+    get() {
+        if (this is PsiWhiteSpace) {
+            return true
+        }
+        if (this is PsiComment) {
+            return true
+        }
+        if (this is LeafPsiElement && this.elementType is KtSingleValueToken) {
+            val token = (this.elementType as KtSingleValueToken)
```
schalkms, line 88:
```suggestion
            val token = (this.elementType as? KtSingleValueToken)
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
```diff
@@ -0,0 +1,94 @@
...
+        true == this.prevSibling?.isAnIsolationElement && true == this.nextSibling?.isAnIsolationElement
+
+private val PsiElement?.isAnIsolationElement: Boolean
+    get() {
+        if (this is PsiWhiteSpace) {
+            return true
+        }
+        if (this is PsiComment) {
+            return true
+        }
```
schalkms, line 86: nit
```suggestion
        if (this is PsiWhiteSpace || this is PsiComment) {
            return true
        }
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
```diff
@@ -0,0 +1,94 @@
...
+    override fun visitCallExpression(expression: KtCallExpression) {
+        super.visitCallExpression(expression)
+
+        if (bindingContext == BindingContext.EMPTY) return
+        val resolvedCall = expression.getResolvedCall(bindingContext) ?: return
+        val returnType = resolvedCall.resultingDescriptor.returnType ?: return
+
+        if (returnType.isUnit()) {
+            return
+        }
```
schalkms, line 56: nit: then the `@Suppress` isn't necessary anymore.

```suggestion
        if (resolvedCall.resultingDescriptor.returnType.isUnit() == true) return
```

BraisGabin:
[Quoting myself](https://github.com/detekt/detekt/issues/2239):

> The annotation `@CheckReturnValue` does that more less. There was a thread about this here: #209 and there is an implementation for [android lint and error-prone](https://github.com/ZacSweers/configurable-checkreturnvalue)

You can read #2239 for more context.

But in short: I think that this rule should use the `@CheckReturnValue` annotation. Then we can add a flag to check for all of them.

cortinico:
> But in short: I think that this rule should use the `@CheckReturnValue` annotation. Then we can add a flag to check for all of them.

Agree on this. I'll extend the rule to account only calls to functions annotated with `@CheckReturnValue`

schalkms:
> But in short: I think that this rule should use the @CheckReturnValue annotation. Then we can add a flag to check for all of them.

I would do this in a next step and merge this PR first. It makes it also easier to review.

BraisGabin:
> I would do this in a next step and merge this PR first. It makes it also easier to review.

I understand your point, but, wouldn't that delay our release? I mean, we would have no-production ready code in `master`.

schalkms:
The rule is disabled anyway.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
cortinico: Why this? This cast is already guarded

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
cortinico: Included

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
cortinico: I'd rather keep it separated to improve readability. Feel free to reopen if you think it's still needed in the following diff.

schalkms: True. Maybe the code should read `if (returnType?.isUnit() == true)`.
I just don't want to spam the codebase with `Suppress` annotations.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
schalkms: True.

cortinico:
I've posted another diff for this rule. There are still some open points that I would love to get an input from @BraisGabin @schalkms:

- Currently the annotation are checked by simple name only (`== CheckReturnValue`) against `CheckReturnValue` and `CheckResult`. Should this be configurable? (imho, yes)

- Should annotations be checked against their fully qualified name (`com.google.errorprone.annotations.CheckReturnValue` and `androidx.annotation.CheckResult`)? (imho, not necessary given how specific are those annotations).

- Should the whole annotations requirement be configurable? Like having a `restrictToAnnotatedMethods: true` parameter? (imho, yes)

- I found a false negative scenario: when the call site is the last statement of a block. Currently the rule is just ignoring this scenario. I've added a test case for that. I wasn't able to find a simple solution to this false negative, and this is probably worth

schalkms:
- Currently the annotation are checked by simple name only (`== CheckReturnValue`) against `CheckReturnValue` and `CheckResult`. Should this be configurable? (imho, yes)
=> yes, this could be done in a separate PR. This would be a nice improvement!

- Should annotations be checked against their fully qualified name (`com.google.errorprone.annotations.CheckReturnValue` and `androidx.annotation.CheckResult`)? (imho, not necessary given how specific are those annotations).
=> not necessary IMHO

- Should the whole annotations requirement be configurable? Like having a `restrictToAnnotatedMethods: true` parameter? (imho, yes)
=> yes, this could be done in a separate PR. This would be a nice improvement!

- I found a false negative scenario: when the call site is the last statement of a block. Currently the rule is just ignoring this scenario. I've added a test case for that. I wasn't able to find a simple solution to this false negative, and this is probably worth a follow-up PR
=> Agreed.

HTH 🙂

BraisGabin:
>     * Currently the annotation are checked by simple name only (`== CheckReturnValue`) against `CheckReturnValue` and `CheckResult`. Should this be configurable? (imho, yes)
>
>     * Should annotations be checked against their fully qualified name (`com.google.errorprone.annotations.CheckReturnValue` and `androidx.annotation.CheckResult`)? (imho, not necessary given how specific are those annotations).

Yes, I think that this should be configurable, and in this configuration we can support both options. The default value should be something like`['*.CheckReturnValue', '*.CheckResult']`. Then it's up to the user to use fully qualified name or not. The only thing to discuess here is which type of regex do we want here. Options that I see:

1. Full regex: Maximum flexibility, difficult to read and maintain: `['.*\.CheckReturnValue', '.*\.CheckResult']`
2. [glob](https://en.wikipedia.org/wiki/Glob_(programming)): No that flexible but gives a lot of flexibility: `['*.CheckReturnValue', '*.CheckResult']`
3. plain string: We just chech that the of the Annotation: `['CheckReturnValue', 'CheckResult']`

The problem here is that in detekt we doesn't have a defined way to do this. I'll open an issue to talk about this. In my opinion the second option is the best. It provides a lot of flexibility and it's easy to read.

>     * Should the whole annotations requirement be configurable? Like having a `restrictToAnnotatedMethods: true` parameter? (imho, yes)
Yes
>     * I found a false negative scenario: when the call site is the last statement of a block. Currently the rule is just ignoring this scenario. I've added a test case for that. I wasn't able to find a simple solution to this false negative, and this is probably worth a follow-up PR
Agree.

cortinico:
> The problem here is that in detekt we doesn't have a defined way to do this. I'll open an issue to talk about this. In my opinion the second option is the best. It provides a lot of flexibility and it's easy to read.

I'd either go for option 2 or 3. Option 2 could be achieved using this:

https://github.com/detekt/detekt/blob/57e0b91b3c57df3988d4d24ba0be411337067adb/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/SplitPattern.kt#L90-L96

This will make the rule behavior consistent with the `ForbiddenImport` rule, that seems reasonable. Option 3 is definitely the easiest. Still don't know if we're maybe offering a "too flexible" solution for a scenario that users probably will never customize.

I'd be in for option 2, I just don't want to block this PR with the issue you mentioned @BraisGabin about the glob pattern in detekt.

BraisGabin:
For me the option 2 is the best. And for sure, this doesn't need to wait for #2711. That issue is a more generic one and probably something to face for `2.0`.

For me this PR can be merged right now if you agree. The missing parts are just to add more configuration but the rule right now work as intended with the default configurations. So any other feature can be added in other PR.

cortinico:
Sure, go for that, I'll follow up with a couple of small PRs

PR #2706, opened 2020-05-18
Remove idea integration

arturbosch:
Continuation from #2704


BraisGabin:
@Mkohm this seems kind of noisy. If danger fails we don't need a comment in the PR showing the error. Is it possible to just report the code smells?

PR #2712, opened 2020-05-19
IgnoreReturnValue: config options

cortinico:
Follow-up to #2698: adding two config to the `IgnoreReturnValue` rule:

* `restrictToAnnotatedMethods: true` to toggle the analysis only on annotated methods
* `returnValueAnnotations: ["*.CheckReturnValue", "*.CheckResult"]` to provide a list of annotations to use during the inspection.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
@@ -43,7 +43,7 @@ import org.jetbrains.kotlin.types.typeUtil.isUnit
  * @configuration restrictToAnnotatedMethods - if the rule should check only annotated methods. (default: `true`)
  * @configuration returnValueAnnotations - List of glob patterns to be used as inspection annotation (default: `['*.CheckReturnValue', '*.CheckResult']`)
  */
-class IgnoredReturnValue(val config: Config = Config.empty) : Rule(config) {
+class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
 
     override val issue: Issue = Issue(
             "IgnoredReturnValue",
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
@@ -41,7 +41,7 @@ import org.jetbrains.kotlin.types.typeUtil.isUnit
  * </compliant>
  *
  * @configuration restrictToAnnotatedMethods - if the rule should check only annotated methods. (default: `true`)
- * @configuration returnValueAnnotations - List of glob patterns to be used as inspection annotation (default: `["*.CheckReturnValue", "*.CheckResult"]`)
+ * @configuration returnValueAnnotations - List of glob patterns to be used as inspection annotation (default: `['*.CheckReturnValue', '*.CheckResult']`)
  */
 class IgnoredReturnValue(val config: Config = Config.empty) : Rule(config) {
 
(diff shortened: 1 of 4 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
```diff
@@ -37,13 +39,11 @@ import org.jetbrains.kotlin.types.typeUtil.isUnit
  *     if (42 == returnsValue()) {}
  *     val x = returnsValue()
  * </compliant>
+ *
+ * @configuration restrictToAnnotatedMethods - if the rule should check only annotated methods. (default: `true`)
+ * @configuration returnValueAnnotations - List of glob patterns to be used as inspection annotation (default: `["*.CheckReturnValue", "*.CheckResult"]`)
```
BraisGabin, line 44:
```suggestion
 * @configuration returnValueAnnotations - List of glob patterns to be used as inspection annotation (default: `['*.CheckReturnValue', '*.CheckResult']`)
```

All our config file uses the single quote.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
```diff
@@ -37,13 +39,11 @@ import org.jetbrains.kotlin.types.typeUtil.isUnit
...
  */
-class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
-
-    private val requiredAnnotations = listOf(
-            "CheckReturnValue",
-            "CheckResult"
-    )
+class IgnoredReturnValue(val config: Config = Config.empty) : Rule(config) {
```
BraisGabin, line 46: Do you need `config` as a property?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValueSpec.kt
```diff
@@ -396,4 +404,92 @@ object IgnoredReturnValueSpec : Spek({
...
+                    listOfChecked("hello")
+                    return 42
+                }
+            """.trimIndent()
+            val findings = IgnoredReturnValue(config).compileAndLintWithContext(wrapper.env, code)
+            assertThat(findings).hasSize(1)
+            assertThat(findings).hasSourceLocation(8, 5)
+        }
+
+        it("does not report when a function is not annotated") {
```
schalkms, line 481:
```suggestion
        it("reports when a function is not annotated") {
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
cortinico: Nop, removing it

cortinico:
Do we know what's happening with the build here? It seems to fail on the compilation of the test snippet. I'm not able to reproduce locally. 

Moreover I see a failure in

```
ForbiddenPublicDataClassSpec > public data class inside an ignored package should pass config as string FAILED
```

That seems unrelated.

BraisGabin:
> Do we know what's happening with the build here? It seems to fail on the compilation of the test snippet. I'm not able to reproduce locally.

You should be able to reproduce it with `./gradlew :detekt-rules:test -Pcompile-test-snippets=true`. By default we don't compile the code snippets in local because it slow down the tests.

> Moreover I see a failure in
> 
> ```
> ForbiddenPublicDataClassSpec > public data class inside an ignored package should pass config as string FAILED
> ```
> 
> That seems unrelated.

🤔 Maybe that test is flaky... strange.

schalkms:
I've already experienced this problem. It could occur because of conflicting `package` declarations or classes with the same name.

PR #2714, opened 2020-05-21
add better documentation for the LongParameterList ignoreAnnotated

kenyee:
From Kotlinlang slack discussion:
https://kotlinlang.slack.com/archives/C88E12QH4/p1589836335126300

Wasn't obvious that ignoreAnnotated should include a CSV list of class names.  Updated documentation to mention that and added an example.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
@@ -32,7 +32,9 @@ import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
  * @configuration ignoreDefaultParameters - ignore parameters that have a default value (default: `false`)
  * @configuration ignoreDataClasses - ignore long constructor parameters list for data classes (default: `true`)
  * @configuration ignoreAnnotated - ignore long parameters list for constructors or functions in the context of these
- * annotation class names (default: `[]`)
+ * annotation class names (default: `[]`); (value is a comma separated string with the annotation class names);
+ * the most common case is for Dagger constructors that are annotated with @Inject which would be specified
+ * with `Inject` (quotes are required)
  *
  * @active since v1.0.0
  */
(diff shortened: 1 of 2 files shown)


kenyee:
> The docs are automatically generated from the rule description.
> The corresponding rule needs to be updated first.

Thanks...that wasn't obvious :-)

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
```diff
@@ -32,7 +32,9 @@ import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
  * @configuration ignoreDefaultParameters - ignore parameters that have a default value (default: `false`)
  * @configuration ignoreDataClasses - ignore long constructor parameters list for data classes (default: `true`)
  * @configuration ignoreAnnotated - ignore long parameters list for constructors or functions in the context of these
- * annotation class names (default: `[]`)
+ * annotation class names (default: `[]`); (value is a comma separated string with the annotation class names);
+ * the most common case is for Dagger constructors that are annotated with @Inject which would be specified
+ * with `Inject` (quotes are required)
```
BraisGabin, line 37: I think that the `` ` `` are not needed. They should not at least. And instead of use csv you can use `[Inject, Module, Suppress]` or, if you prefer `['Inject', 'Module', 'Suppress']`. Or yml even allow you to do this:

```yml
ignoreAnnotated:
  - Inject
  - Module
  - Suppress
```

Anyway, I agree that we should improve our documentation to say what's the expected value for each parameter.

kenyee: Hmm..in other parts of our config file, we do stuff like this:
    exceptions: 'IllegalArgumentException,IllegalStateException,IOException'
and
    excludes: "**/test/**,**/androidTest/**,**/*.Test.kt,**/*.Spec.kt,**/*.Spek.kt"

The docs probably need a section on specifying lists of values that this config param description could link to.  I didn't realize it was just a regular YAML parser or that you could specify strings w/o quotes in YAML :-)
Is there a preferred way of filling out these multivalue fields?

arturbosch: @BraisGabin recently added the support for the normal yaml list syntax for rule properties (which is more readable imo).
We should totally encourage users to use this syntax.

PR #2715, opened 2020-05-21
Clean up code by using builtin associateBy function

arturbosch:


arturbosch:
> Warnings
> 	
> **detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/OutputFacade.kt#L27** - Type checking is a sign of violating the Open-Closed Principle. Consider introducing an abstraction (interface) for `OutputReport`, `ConsoleReport`, with new implementations of the interface for every class.
> 
> Generated by  [Danger Kotlin](https://danger.systems) against [0d0e23e](https://github.com/detekt/detekt/commit/0d0e23e3f2560d6f74d91762f6348a9b9e44e835)

Jap, its bad design here.

Issue #2680, opened 2020-05-12
Detekt's 2.0 architecture


BraisGabin:
There are PRs and issues talking about how to arrange the code in the project for the 2.0. Some PRs moving classes and creating aliases to don't break the compatibility for "1.0" but preparing the 2.0 version.

I'm not saying that we are near to create the 2.0 version. I think that there is not a big reason right now to release a major release with breaking changes. But, anyway, we can start talking about how should be the architecture of detekt. Right now I'm mainly thinking about the gradle modules, but we can talk about other topics too.

Ideas:
- Which modules we do need? Which is the responsability for each of them?
- How do we make easy to understand what's our public API and what's our private api?

For example. Right now `:detekt-gradle-plugin` depends on `:detekt-cli`. I think that we should break that dependency. Both should be *clients* of/depend on the same module that exposes all the functionality.

Should we expose all our public api in a single module or split it? Rules/post processors/metrics?/something else?

Could we hide the implementation details from that modules? I mean, could we make does modules without depend on our "core"? This way we reduce the accidental use of our private API.

Other option to avoid this problem is to create a custom rule so the third parties can use that rule and warn them if they are using our private API.

arturbosch:
Nice idea to start collecting/discussing further architecture changes.

A `:detekt-tooling-api` could introduce interfaces/service loaders for embedding detekt.
Some ideas:
```kotlin
interface DetektCliWrapper {
    fun execute(args: Array<String>): ExitCode
}
```

```kotlin
interface RunnerProvider {
    fun load(args: Array<String>, out: Appendable, error: Appendable): Runner
}
```

```kotlin
interface DetektFacadeProvider {
    fun load(settings: ProcessingSettings, classloader: ClassLoader?): DetektFacade
}
```

```kotlin
interface DetektFacade {
    fun run(files: Collection<KtFile>, context: BindingContext): Result
}
```

schalkms:
I also feel that this issue is a good starting point for collecting ideas. Thanks for that! @BraisGabin 

> For example. Right now :detekt-gradle-plugin depends on :detekt-cli. I think that we should break that dependency. Both should be clients of/depend on the same module that exposes all the functionality.

That's the biggest pain point. For me this has the highest priority.

The rest is IMHO nice to have. As a user, I'd like to have a **long term stable** detekt API. Users don't want to change their plugins and 3rd party rules every few months.

[...]

PR #2716, opened 2020-05-21
Introduce parser and psi module

arturbosch:
First steps towards cleaner api module and a common core module for Gradle/Sonar/IntelliJ/Compiler-Plugins and Cli module - #2680.

This PR gets rid of our usage of the kotlin compiler internal stuff from our api module.
The api module only references Psi classes inherited inside the Rule/DetektVisitor class.

Nice to have: my hope by introducing simpler modules is to get rid of all the build dependencies for `detekt-test` which all modules uses as a test implementation dependency. This slows down our build because all parallel threads have to wait until detekt-test is build to start testing.


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSet.kt
```diff
@@ -1,12 +1,11 @@
 package io.gitlab.arturbosch.detekt.api
 
+import io.github.detekt.psi.absolutePath
```
BraisGabin, line 3: Should we keep this in a `internal` package?

detekt-parser/build.gradle.kts
```diff
@@ -0,0 +1,11 @@
+dependencies {
+    api(kotlin("compiler-embeddable"))
```
BraisGabin, line 2: Couldn't this be implementation?

arturbosch: Unfortunately not yet. I will see if this can be done in further PRs.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSet.kt
arturbosch: Maybe it will get hidden behind a `implementation` configuration later. I will keep this in mind.

PR #2720, opened 2020-05-22
Modularize test module

arturbosch:
This "huge" PR (mostly import changes) splits the `detekt-test` module into two modules.
`detekt-test` remains the module used for testing rules.
The new `detekt-test-utils` module is more lightweight in module dependencies.
This module specializes in utility classes and parsing kotlin files for tests.

The problem with the `detekt-test` module:
`detekt-test` depends on "all" compile tasks of other detekt-modules **but** is the requirement for all test compile tasks! This means it is not only the bottle-neck for parallel build but also triggers all tests on nearly any change (especially gradle-plugin tests with runtime of 2.5m!!!)!

This first change in test dependencies gets rid of the `detekt-test` dependency on `generator`, `gradle-plugin` and `metrics` modules.
This leads to less triggering of the gradle tests (!!!) and these modules can be executed earlier in parallel builds (20s earlier gradle plugin testing, see screenshot).

gradle-plugin-earlier-tests
  
In further non-breaking modularization effors for #2680, I want to get rid of the `detekt-test` dependency for `cli`, `core` and the `api` modules leading to less retriggering of tests.

While this change does not necessary lead to performance improvements for a clean build, it will for incremental builds. This may improve CI build times as more test tasks can be cached.

In a breaking 2.0 release we could rename `detekt-test` to `detekt-rule-testing-support` or something to make it's name more clearer.

### DIFF
--- a/detekt-parser/src/test/resources/cases/Default.kt
+++ b/detekt-parser/src/test/resources/cases/Default.kt
@@ -1,0 +1,7 @@
+package cases
+
+/**
+ * A comment
+ */
+@Suppress("Unused")
+class Default
--- a/detekt-test-utils/build.gradle.kts
+++ b/detekt-test-utils/build.gradle.kts
@@ -1,0 +1,3 @@
+dependencies {
+    implementation(project(":detekt-parser"))
+}
--- a/detekt-test-utils/build.gradle.kts
+++ b/detekt-test-utils/build.gradle.kts
@@ -1,3 +1,9 @@
 dependencies {
+    api(kotlin("stdlib-jdk8"))
+    api("org.assertj:assertj-core:${Versions.ASSERTJ}")
     implementation(project(":detekt-parser"))
+    implementation(project(":detekt-psi-utils"))
+    implementation(kotlin("script-runtime"))
+    implementation(kotlin("script-util"))
+    implementation(kotlin("scripting-compiler-embeddable"))
 }
--- a/detekt-test/build.gradle.kts
+++ b/detekt-test/build.gradle.kts
@@ -1,9 +1,4 @@
 dependencies {
-    api(kotlin("stdlib-jdk8"))
     api(project(":detekt-core"))
     api(project(":detekt-test-utils"))
-    implementation(kotlin("script-runtime"))
-    implementation(kotlin("script-util"))
-    implementation(kotlin("scripting-compiler-embeddable"))
-    implementation("org.assertj:assertj-core:${Versions.ASSERTJ}")
 }
--- a/detekt-test/build.gradle.kts
+++ b/detekt-test/build.gradle.kts
@@ -1,4 +1,4 @@
 dependencies {
-    api(project(":detekt-core"))
+    implementation(project(":detekt-core"))
     api(project(":detekt-test-utils"))
 }
--- a/detekt-test/build.gradle.kts
+++ b/detekt-test/build.gradle.kts
@@ -1,6 +1,7 @@
 dependencies {
     api(kotlin("stdlib-jdk8"))
     api(project(":detekt-core"))
+    api(project(":detekt-test-utils"))
     implementation(kotlin("script-runtime"))
     implementation(kotlin("script-util"))
     implementation(kotlin("scripting-compiler-embeddable"))
--- a/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/Resources.kt
+++ b/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/Resources.kt
@@ -3,13 +3,6 @@ package io.gitlab.arturbosch.detekt.test
 import io.gitlab.arturbosch.detekt.api.internal.YamlConfig
 import java.net.URI
 
-internal object Resources
-
-fun resource(name: String): URI {
-    val explicitName = if (name.startsWith("/")) name else "/$name"
-    val resource = Resources::class.java.getResource(explicitName)
-    requireNotNull(resource) { "Make sure the resource '$name' exists!" }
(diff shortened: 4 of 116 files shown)


detekt-parser/src/test/resources/cases/Default.kt
```diff
@@ -0,0 +1,7 @@
+package cases
+
+/**
+ * A comment
+ */
+@Suppress("Unused")
+class Default
```
BraisGabin, line 7: Shouldn't be better to inline this code?

detekt-test/build.gradle.kts
```diff
@@ -1,8 +1,4 @@
 dependencies {
-    api(kotlin("stdlib-jdk8"))
     api(project(":detekt-core"))
```
BraisGabin, line 2: Couldn't this be just `implementation`?

detekt-test/build.gradle.kts
```diff
@@ -1,8 +1,4 @@
...
-    implementation(kotlin("script-runtime"))
-    implementation(kotlin("script-util"))
-    implementation(kotlin("scripting-compiler-embeddable"))
-    implementation("org.assertj:assertj-core:${Versions.ASSERTJ}")
+    api(project(":detekt-test-utils"))
```
BraisGabin, line 3: Probably same here. If you need both you can always add both as dependencies. I think that it's better to be explicit.

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/Resources.kt
```diff
@@ -3,13 +3,6 @@ package io.gitlab.arturbosch.detekt.test
...
 
-internal object Resources
-
-fun resource(name: String): URI {
-    val explicitName = if (name.startsWith("/")) name else "/$name"
-    val resource = Resources::class.java.getResource(explicitName)
-    requireNotNull(resource) { "Make sure the resource '$name' exists!" }
-    return resource.toURI()
-}
+fun resource(name: String): URI = io.github.detekt.test.utils.resource(name)
```
BraisGabin, line 6: This function is here because you don't want to create a BC, right?

detekt-test-utils/build.gradle.kts
```diff
@@ -0,0 +1,9 @@
+dependencies {
+    api(kotlin("stdlib-jdk8"))
+    api("org.assertj:assertj-core:${Versions.ASSERTJ}")
```
BraisGabin, line 3: Why don't just use `implementation`?

detekt-parser/src/test/resources/cases/Default.kt
arturbosch: KtCompiler works on `Path` objects.

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/Resources.kt
arturbosch: Hm, no just to not need to modify all test files for now.
We never said that `detekt-test` is api or publicly supported, just mentioned it in the sample-extension build file :/.
I will migrate the `resource` usages in another PR.
We should try not to break functions like `Rule.lint` as they are more likely used.

detekt-test-utils/build.gradle.kts
arturbosch: AssetJ is our assertion library and therefore is used in every module.
We would have to declare it in 10+ modules.

detekt-test/build.gradle.kts
arturbosch: I considered this but our tests use to many dependencies from test-utils for now.
I consider hiding this dependency from the sample-extension build file vital for now.

detekt-parser/src/test/resources/cases/Default.kt
arturbosch: The empty css file is also to test KtCompiler ^^

Whathecode:
After upgrading to 1.10.0, due to this change I expect, I got the following error:

> e: ...MyRuleTest.kt: (158, 39): Cannot access 'io.github.detekt.parser.KtCompiler' which is a supertype of 'io.github.detekt.test.utils.KtTestCompiler'. Check your module classpath for missing or conflicting 

I had to add a dependency on `detekt-parser` to use `KtTestCompiler` in the test sources of my custom rules:

```
    testImplementation "io.gitlab.arturbosch.detekt:detekt-test:1.10.0"
    testImplementation "io.gitlab.arturbosch.detekt:detekt-parser:1.10.0"
```

Don't know whether this was intentional. Just thought I'd share this problem I ran into in case others encounter it.

arturbosch:
Thanks for reporting.
What is your usecase with `KtTestCompiler` ?

If its parsing content you could use `compileContentForTest` or `compileForTest` without needing the dependency.
Are you using `createEnvironment` ?

Whathecode:
I was just using `KtTestCompiler.compileFromContent`. Your recommendation worked perfectly. Thank you!

arturbosch:
@Whathecode are you using Spek as testing framework?
If yes, there is `fun Root.setupKotlinEnvironment()` which can be used as in 
https://github.com/detekt/detekt/blob/b6e88a95a11d21c0f4a9aef90455668a21de4060/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ImplicitUnitReturnTypeSpec.kt#L16

I've created #2871 to hide KtTestCompiler from users. There will be an alternative for testing rules which need type resolution.

Whathecode:
Nope, I am using `kotlin.test`.

PR #2725, opened 2020-05-24
Move processors to metrics module

arturbosch:


detekt-metrics/src/test/kotlin/io/github/detekt/metrics/processors/package.kt
```diff
@@ -0,0 +1,209 @@
+package io.github.detekt.metrics.processors
+
+val default = """
```
arturbosch, line 3: I've inlined all test files here :)

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/FileProcessorLocatorSpec.kt
```diff
@@ -41,7 +41,7 @@ class FileProcessorLocatorSpec : Spek({
 })
 
 private fun getProcessorClasses(): List<Class<out FileProcessListener>> {
-    return Reflections("io.gitlab.arturbosch.detekt.core.processors")
+    return Reflections("io.github.detekt.metrics.processors")
```
schalkms, line 44: One has to be careful with renaming stuff and the impact on the docs. Is there anything on the homepage that needs to be updated as well? A **grep** commando executed once could be beneficial.

arturbosch: Good point. We are safe here as extensions are referenced via their id:
```
interface Extension {
    /**
     * Name of the extension.
     */
    val id: String get() = javaClass.simpleName
``` 

This package moving is safe :).

PR #2729, opened 2020-05-24
Update to Gradle 6.4.1

arturbosch:


detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/MultiVersionTest.kt
```diff
@@ -31,7 +31,7 @@ private fun getGradleVersionsUnderTest() =
     if (getJdkVersion() < 13) {
         listOf("5.0", "6.2.2")
```
BraisGabin, line 32: Shouldn't we upgrade this one too?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.4.1-bin.zip
```
cortinico, line 3:
```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-6.4.1-all.zip
```

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/MultiVersionTest.kt
schalkms, line 32:
```suggestion
        listOf("5.0", "6.4.1")
```

arturbosch: It's the last Gradle version which does not support Java 13+, so we can't update this one.

arturbosch: Maybe I'm wrong here. Just tested it. Locally it works. I've merged @schalkms suggestion and see what the CI says.

PR #2736, opened 2020-05-25
Update gradle scan plugin

arturbosch:


schalkms:
@arturbosch could you ping me on Slack via pN please? I have two questions. Thanks 😊

PR #2739, opened 2020-05-27
Simplify baseline data structures

arturbosch:


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/baseline/BaselineFacade.kt
```diff
@@ -8,31 +8,24 @@ import java.nio.file.Path
...
-            if (listings != null) {
-                val whiteFiltered = smells.filterNot { finding -> listings.first.ids.contains(finding.baselineId) }
-                val blackFiltered = whiteFiltered.filterNot { finding ->
-                    listings.second.ids.contains(finding.baselineId)
-                }
-                blackFiltered
-            } else smells
+    fun filter(findings: List<Finding>): List<Finding> {
+        val whiteFiltered = findings.filterNot { finding -> baseline.whitelist.contains(finding.baselineId) }
+        return whiteFiltered.filterNot { finding -> baseline.blacklist.contains(finding.baselineId) }
```
schalkms, line 21: I’d introduce a new variable for readability reasons and then return this variable.

BraisGabin:
I'll review this tomorrow. I see something that doesn't fit but I'm not sure what it is (a bit too late for me to think properly). I mean, for sure this is better than before but I think that we could make it even better. If it blocks you dot me is OK, we can rethink it later.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/baseline/Baseline.kt
```diff
@@ -2,7 +2,9 @@ package io.gitlab.arturbosch.detekt.cli.baseline
 
 import io.gitlab.arturbosch.detekt.api.Finding
 
-data class Baseline(val blacklist: Blacklist, val whitelist: Whitelist)
+typealias FindingsIdList = Set<String>
```
BraisGabin, line 5: This is the perfect example for an `inline class`. I don't know if it's worth it though.

arturbosch: They are experimental and would need more lines of code to activate. Lets wait for 1.4 :)

arturbosch:
> I think that `Baseline` should have a function `contains` where you give it an id and it returns `true` or `false` if it's contained or not.
> 
> Then we can rebase `BaselineFacade` to `BaselineFactory` so given a file it provides us the `Baseline`.
> 
> `Baseline` should have another function that given a list of findings (or just a list of ids) it provides a new `Baseline` that mantains the Blacklist and changes the whitelist.
> 
> What do you think?
> 
> As I said, this could be done later. For that reason I approve this PR like it is.

Awesome suggestions. Applied and further simplified, thanks!

PR #2741, opened 2020-05-27
Move baseline feature to core module

arturbosch:
Needs two previous PRs to be merged first.

Towards a minimal cli and a common core module #2680


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/Runner.kt
```diff
@@ -38,7 +36,7 @@ class Runner(
             var (engineRunTime, result) = measure { facade.run() }
             settings.debug { "Running core engine took $engineRunTime ms" }
             checkBaselineCreation(result)
-            result = transformResult(result)
+            result = arguments.baseline?.let { BaselineFacade().transformResult(it, result) } ?: result
```
arturbosch, line 39: Not super nice but gets the job done for now.

PR #2742, opened 2020-05-27
Mark some well known cli functions as implicit unsupported api

arturbosch:
Not sure if necessary but may help to remind us to not break these functions in further refactorings.


arturbosch:
Cool that it is helpful :)

PR #2743, opened 2020-05-27
Simplify core facade class

arturbosch:


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/DetektFacade.kt
```diff
@@ -30,23 +28,19 @@ class DetektFacade(
...
-        val notifications = mutableListOf<Notification>()
-        val findings = HashMap<String, List<Finding>>()
-
         val filesToAnalyze = inputPaths.flatMap(compiler::compile)
         val bindingContext = generateBindingContext(filesToAnalyze)
 
         processors.forEach { it.onStart(filesToAnalyze) }
 
-        findings.mergeSmells(detektor.run(filesToAnalyze, bindingContext))
+        val findings = detektor.run(filesToAnalyze, bindingContext)
```
BraisGabin, line 23:
```suggestion
        val findings: Map<String, List<Finding>> = detektor.run(filesToAnalyze, bindingContext)
```

It's not obvious what this function returns. (I'm not even sure if it is that what it returns)

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/BindingContext.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.jetbrains.kotlin.cli.jvm.compiler.NoScopeRecordCliBindingTrace
+import org.jetbrains.kotlin.cli.jvm.compiler.TopDownAnalyzerFacadeForJVM
+import org.jetbrains.kotlin.config.languageVersionSettings
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.lazy.declarations.FileBasedDeclarationProviderFactory
+
+internal fun generateBindingContext(settings: ProcessingSettings, files: List<KtFile>): BindingContext {
+    val classpath = settings.classpath
+    val environment = settings.environment
```
BraisGabin, line 17: `ProcessingSettings` is a huge class. Here we just need two parameters, maybe we could just send those.

PR #2744, opened 2020-05-28
Fix EqualsAlwaysReturnsTrueOrFalse doc

hqzxzwb:
Use `==` in equals causes infinite recursion.


cortinico:
Thanks for the fix @hqzxzwb You need to run `./gradlew build` as the documentation will be re-generated and should be attached to the diff, otherwise the CI will fail.

PR #2750, opened 2020-05-29
Extract xml and html reports to own modules

arturbosch:
Steps to extract more common logic from the cli module #2680.

This is a non-breaking backwards compatible step as the cli still has hard dependencies on these report modules. In a next step I will move the reporting logic to the core module (OutputFacade, ReportPath etc) which is also non breaking.

In a 2.0 release we can remove the hard dependencies on reports and let the user decide for example via `detektPlugins("detekt-report-html")` to include additional reports. This will remove the hard dependency on the `jcenter` repository if the html report is not needed.

Blocked by #2749.


detekt-report-xml/src/main/kotlin/io/github/detekt/report/xml/XmlOutputReport.kt
```diff
@@ -21,7 +21,7 @@ class XmlOutputReport : OutputReport() {
         lines += "<?xml version=\"1.0\" encoding=\"utf-8\"?>"
         lines += "<checkstyle version=\"4.3\">"
 
-        smells.groupBy { it.location.file }.forEach { fileName, findings ->
+        smells.groupBy { it.location.file }.forEach { (fileName, findings) ->
```
BraisGabin, line 24: 🤔 this change is strange. Was it working before?

detekt-report-html/src/test/resources/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1 @@
+DetektVersion: 1.9.1
```
BraisGabin, line 1: 🤔 is this necessary? I imagine it's because we use the version number in the html report but it was `1.6.0`. You changed it here but there's no change in any test. There should be a problem somewhere.

BraisGabin:
Should txt report be extracted too?

detekt-report-xml/src/main/kotlin/io/github/detekt/report/xml/XmlOutputReport.kt
arturbosch: Yes, we used the forEach method from java. Intellij has an intention to convert it to kotlin's forEach version. The yellow highlighting under the forEach made me wanna get rid of it xD.
It does not need an extra import and gets resolved before the java version.

detekt-report-html/src/test/resources/META-INF/MANIFEST.MF
arturbosch: We need any version here so `whichDetekt` does not return null.
The report default is to print `unknown` when version is null.

```
val version = whichDetekt()
val header = """generated with <a href="https://arturbosch.github.io/detekt">detekt version $version</a> on """
val result = htmlReport.render(TestDetektion())
```

arturbosch:
> Should txt report be extracted too?

Done.

PR #2753, opened 2020-05-30
Add default print methods to ForbiddenMethodCall

### LINKED ISSUE #2678: New rule: NoPrintStatement
This rule finds all usages of standard `println()` or `print()` (System.out.println).
A logger or `PrintStream` facility should be used instead.

## Context
Unified logging/piping output of a (command line) program.

cortinico:
Adding all the variations of print/println from Java
to the ForbiddenMethodCall.

The rationale is to encourage the usage of a Logger or a PrintStream

Moreover this fixes the rule that was not catching all the method
calls due to overriding visitQualifiedExpression rather than visitCallExpression

Fixes #2678

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCallSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCallSpec.kt
@@ -1,6 +1,7 @@
 package io.gitlab.arturbosch.detekt.rules.style
 
 import io.github.detekt.test.utils.KtTestCompiler
+import io.gitlab.arturbosch.detekt.api.SourceLocation
 import io.gitlab.arturbosch.detekt.test.TestConfig
 import io.gitlab.arturbosch.detekt.test.assertThat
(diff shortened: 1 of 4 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCallSpec.kt
```diff
@@ -16,15 +17,19 @@ class ForbiddenMethodCallSpec : Spek({
...
     describe("ForbiddenMethodCall rule") {
 
-        it("should report nothing by default") {
+        it("should report kotlin print usages by default") {
             val code = """
-            import java.lang.System
             fun main() {
-            System.out.println("hello")
+            print("3")
+            println("4")
```
schalkms, line 24: Indentation looks weird here. It might be due to GitHub’s diff viewer?!

cortinico: Nop you were right, for some reason some tests were not indented. Fixing them (expect test failures due to assert on source location).

PR #2755, opened 2020-05-31
Introduce reporting extensions

arturbosch:
This PR introduces the concept of a `ReportingExtension`.
With this extension the user can listen to results and manipulate them.

The first two implementations are `BaselineFacade` and `OutputFacade`.
With this concept the baseline and the reporting features do not need special treatment in the `Runner` class or the Intellij or Sonarqube plugin but are just loaded from the detekt core engine and executed.

No more code like:
```kotlin
result = arguments.baseline?.let { BaselineFacade().transformResult(it, result) } ?: result	                
val (outputResultsTime) = measure { OutputFacade(arguments.reportPaths, result, settings).run() }
```

Extracting the "special @Suppress" implemention to a `ReportingExtension` is conceivable in the future.
 
This PR also introduces a concept to share data between extensions and tooling frondends (cli, gradle, sonar, intellij) via the `PropertiesAware` interface.
It allows to retrieve data in a type safe way:
```kotlin
inline fun <reified T : Any> PropertiesAware.getOrNull(key: String): T?
```
This lightweight key-value storage reduces the need for extra parameters for the `ProcessingSettings` class.

I've marked both new api interfaces as `@UnstableApi` for now.

#2680

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/PropertiesAware.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/PropertiesAware.kt
@@ -16,7 +16,7 @@ interface PropertiesAware {
     /**
      * Binds a given value with given key and stores it for later use.
      */
-    fun register(key: String, value: Any?)
+    fun register(key: String, value: Any)
 }
 
 /**
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/PropertiesAware.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/PropertiesAware.kt
@@ -1,0 +1,32 @@
+package io.gitlab.arturbosch.detekt.api
+
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
+
+/**
+ * Properties holder. Allows to store and retrieve any data.
+ */
+@UnstableApi
+interface PropertiesAware {
+
+    /**
+     * Raw properties.
+     */
+    val properties: Map<String, Any?>
+
+    /**
+     * Binds a given value with given key and stores it for later use.
+     */
+    fun register(key: String, value: Any?)
+}
+
+/**
+ * Allows to retrieve stored properties in a type safe way.
+ */
+@UnstableApi
+inline fun <reified T : Any> PropertiesAware.getOrNull(key: String): T? {
+    val value = properties[key]
+    if (value != null) {
+        return value.safeAs() ?: error("No value of type ''${T::class} for key '$key'.")
+    }
+    return null
+}
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/ReportingExtension.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/ReportingExtension.kt
@@ -1,0 +1,28 @@
+package io.gitlab.arturbosch.detekt.api
+
+/**
+ * Allows to intercept detekt's result container by listening to the initial and final state
+ * and manipulate the reported findings.
+ */
+@UnstableApi
+interface ReportingExtension : Extension {
+
+    /**
+     * Is called before any [transformFindings] calls were executed.
+     */
+    fun onRawResult(result: Detektion) {
+        // intercept for the initial results
+    }
+
+    /**
+     * Allows to transform the reported findings e.g. apply custom filtering.
+     */
+    fun transformFindings(findings: Map<RuleSetId, List<Finding>>): Map<RuleSetId, List<Finding>> = findings
+
+    /**
+     * Is called after all extensions's [transformFindings] were called.
+     */
+    fun onFinalResult(result: Detektion) {
+        // intercept for the final results
+    }
+}
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/SingleRuleRunner.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/SingleRuleRunner.kt
@@ -25,22 +25,25 @@ class SingleRuleRunner(
             arguments.runRule?.split(":")
         ) { "Unexpected empty 'runRule' argument." }
 
-        arguments.createSettings(outputPrinter, errorPrinter)
-            .use { settings ->
-                val realProvider = requireNotNull(
-                    RuleSetLocator(settings).load().find { it.ruleSetId == ruleSet }
-                ) { "There was no rule set with id '$ruleSet'." }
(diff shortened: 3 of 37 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineResultMapping.kt
```diff
@@ -0,0 +1,43 @@
...
+import io.gitlab.arturbosch.detekt.api.SetupContext
+import io.gitlab.arturbosch.detekt.api.UnstableApi
+import io.gitlab.arturbosch.detekt.api.getOrNull
+import io.gitlab.arturbosch.detekt.core.DetektResult
+import java.nio.file.Path
+
+@OptIn(UnstableApi::class)
+class BaselineResultMapping : ReportingExtension {
+
+    private var baselineFile: Path? = null
```
BraisGabin, line 15: lateinit?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/SingleRuleRunner.kt
```diff
@@ -18,43 +16,31 @@ import java.io.PrintStream
...
-                listOf(DetektProgressListener().apply { init(settings) })
-            ).run()
-        }
+                DetektFacade.create(
+                    settings,
+                    listOf(provider),
+                    listOf(DetektProgressListener().apply { init(settings) })
+                ).run()
+            }
     }
```
BraisGabin, line 44: Is this correctly fomatted?

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/SingleRuleRunnerSpec.kt
```diff
@@ -26,7 +24,7 @@ class SingleRuleRunnerSpec : Spek({
                 "--run-rule", "test:test"
             )
 
-            SingleRuleRunner(args, NullPrintStream(), NullPrintStream()).execute()
+            SingleRuleRunner(args, System.out, System.err).execute()
```
BraisGabin: This creates noisy test, right?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/OutputFacade.kt
```diff
@@ -1,36 +1,54 @@
...
-            .load()
-            .forEach(::handleOutputReport)
-        ConsoleReportLocator(settings)
-            .load()
-            .forEach(::handleConsoleReport)
+    override fun init(context: SetupContext) {
+        val reportPaths: Collection<ReportPath> =
+            context.getOrNull(DETEKT_OUTPUT_REPORT_PATHS_KEY) ?: emptyList()
+        reports = reportPaths.associateBy { it.kind }
+        settings = context as? ProcessingSettings ?: error("ProcessingSettings expected.")
```
BraisGabin, line 25: 🤔 this cast is a bit strange, isn't it?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/OutputFacade.kt
```diff
@@ -13,6 +13,8 @@ import kotlin.system.measureTimeMillis
 @OptIn(UnstableApi::class)
 class OutputFacade : ReportingExtension {
 
+    override val priority: Int = Int.MIN_VALUE
```
BraisGabin, line 16: Why is this necessary? OutputFacade only overrides `onFinalResult` and in that state no one should change the `Detektion` (for that reason I was talking about `Detektion` as an immutable class)

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/ReportingExtension.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * Allows to transform the reported findings e.g. apply custom filtering.
+     */
+    fun transformFindings(findings: Map<RuleSetId, List<Finding>>): Map<RuleSetId, List<Finding>> = findings
+
+    /**
+     * Is called after all extensions's [transformFindings] were called.
+     */
+    fun onFinalResult(result: Detektion) {
```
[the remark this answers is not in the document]
arturbosch: Yes, please feel free to write it down in the issue.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineResultMapping.kt
arturbosch: The baseline file can be null.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/SingleRuleRunner.kt
arturbosch: Yes, still I extracted the long lambda to an function to make it more clear :).

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/OutputFacade.kt
arturbosch: However `Detektion` is still mutable. So it might be that anyone may add additional `ProjectMetric`'s and expect them to be printed on the console.

With running the output facade last we will ensure that everything stored will be printed.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/OutputFacade.kt
arturbosch: Yes! I'm not sure how to handle this. It needs the plugin classloader. An extension which loads other extensions.
Feels kinda wrong.
I will convert it to a plain core class again.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/PropertiesAware.kt
```diff
@@ -0,0 +1,32 @@
...
+ * Allows to retrieve stored properties in a type safe way.
+ */
+@UnstableApi
+inline fun <reified T : Any> PropertiesAware.getOrNull(key: String): T? {
+    val value = properties[key]
+    if (value != null) {
+        return value.safeAs() ?: error("No value of type ''${T::class} for key '$key'.")
+    }
+    return null
+}
```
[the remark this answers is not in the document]
arturbosch: Changed.

PR #2763, opened 2020-06-03
Support layout property for ImportOrdering rule

arturbosch:
Looks like we should not rush in a release featuring 0.37.0

https://github.com/pinterest/ktlint/issues/761
https://github.com/pinterest/ktlint/issues/682

I plan to prepare a release candidate 1.10.0-RC1 in the next days as we have many refactorings in this one.

### DIFF
--- a/detekt-cli/src/main/resources/default-detekt-config.yml
+++ b/detekt-cli/src/main/resources/default-detekt-config.yml
@@ -238,6 +238,7 @@ formatting:
   ImportOrdering:
     active: false
     autoCorrect: true
+    layout: 'idea'
   Indentation:
     active: false
     autoCorrect: true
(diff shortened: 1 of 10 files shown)


schalkms:
> I plan to prepare a release candidate 1.10.0-RC1 in the next days as we have many refactorings in this one.

Have you tested this recent changes thoroughly? I don't think it's a good idea to let the client/user find out breaking stuff.

arturbosch:
> > I plan to prepare a release candidate 1.10.0-RC1 in the next days as we have many refactorings in this one.
> 
> Have you tested this recent changes thoroughly? I don't think it's a good idea to let the client/user find out breaking stuff.

Yes, while refactoring and will do more :).

samramez:
How can we find examples on how we're suppose to use the `layout` property?

schalkms:
@samramez yes, there is documentation available on the [homepage](https://arturbosch.github.io/detekt/formatting.html#importordering). You can configure it in the [detekt.yml](https://github.com/detekt/detekt/blob/master/detekt-core/src/main/resources/default-detekt-config.yml) file.
What exactly do you find missing?

PR #2767, opened 2020-06-04
Update spek to 2.0.11 disabling timeouts

arturbosch:


arturbosch:
Oh sorry.
Spek removed the default timeout of 10 seconds which leads to failing tests when debugging on IntelliJ.
They are getting back to it when they find a way to turn this off only on IDE.

PR #2771, opened 2020-06-04
Remove core module dependency for detekt-test

arturbosch:
Results in earlier execution of test tasks for all modules -> better parallalization.

I think we should enable caching and parallel builds by default for everyone in gradle.properties.
I'm using both for months and it is amazing :)


detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/CorrectableRulesFirstSpec.kt
```diff
@@ -38,7 +37,7 @@ class CorrectableRulesFirstSpec : Spek({
 
             val testFile = path.resolve("Test.kt")
             val detector = Detektor(
-                createProcessingSettings(testFile, yamlConfig("one-correctable-rule.yml")),
+                createProcessingSettings(testFile, yamlConfig("configs/one-correctable-rule.yml")),
```
arturbosch, line 40: Memo for next PR: remove memory leak

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleSetExtensions.kt
```diff
@@ -1,13 +1,14 @@
 package io.gitlab.arturbosch.detekt.test
 
 import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.RuleSet
 import io.gitlab.arturbosch.detekt.api.RuleSetProvider
-import io.gitlab.arturbosch.detekt.core.rules.createRuleSet
 
 /**
  * Creates a RuleSet instance of given RuleSetProvider.
```
schalkms, line 8:
```suggestion
 * Loads a [RuleSet] instance of given RuleSetProvider.
```

PR #2773, opened 2020-06-05
Run in parallel by default

arturbosch:


BraisGabin:
We can remove the parallel flag in the ci now.

PR #2774, opened 2020-06-05
Do not print passing tests on the console

arturbosch:


buildSrc/src/main/kotlin/commons.gradle.kts
```diff
@@ -48,9 +48,8 @@ subprojects {
             // set options for log level LIFECYCLE
             events = setOf(
                 TestLogEvent.FAILED,
-                TestLogEvent.PASSED,
-                TestLogEvent.SKIPPED,
-                TestLogEvent.STANDARD_OUT
```
cortinico, line 53: `TestLogEvent.STANDARD_OUT` is probably worth to keep as there are tests that are printing out stuff to STDOUT.

PR #2778, opened 2020-06-06
Rename Blacklist and Whitelist to be self explanatory

remcomokveld:
I originally started this exercise because I believe the terms blacklist and whitelist have racial connotations that are inappropriate to be used. But while trying to find new names I found that `SuppressedFalsePositive` and `TemporarySuppressedIssues` actually describe much more accurately than the previous values. 

This PR would currently be a breaking change because existing baseline files would no longer work. If desired I think it would be possible to build in backward compatibility. However, I think the fact that this is a breaking change is a good thing because it will raise more awareness about the fact that blacklist and whitelist are really terms that we should all stop using.

Edit:
Updated the PR to not be a breaking change anymore. Existing baseline files with `<Blacklst>` and `<Whitelist>` tags will still be supported, but newly generated baselines will use `SuppressedFalsePositive` and `TemporarySuppressedIssues` respectively.

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/Baseline.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/Baseline.kt
@@ -9,12 +9,12 @@ internal typealias FindingsIdList = Set<String>
 internal typealias FindingId = String
 
 internal data class Baseline(
-    val suppressedFalsePositives: FindingsIdList,
-    val temporarySuppressedIssues: FindingsIdList
+    val manuallySuppressedIssues: FindingsIdList,
+    val currentIssues: FindingsIdList
 ) {
 
     fun contains(id: FindingId): Boolean =
-        temporarySuppressedIssues.contains(id) || suppressedFalsePositives.contains(id)
+        currentIssues.contains(id) || manuallySuppressedIssues.contains(id)
 
     companion object {
 
@@ -30,8 +30,8 @@ const val DETEKT_BASELINE_PATH_KEY = "detekt.baseline.path.key"
 const val DETEKT_BASELINE_CREATION_KEY = "detekt.baseline.creation.key"
 
 internal const val SMELL_BASELINE = "SmellBaseline"
-internal const val SUPPRESSED_FALSE_POSITIVES = "SuppressedFalsePositives"
-internal const val TEMPORARY_SUPPRESSED_ISSUES = "TemporarySuppressedIssues"
+internal const val MANUALLY_SUPPRESSED_ISSUES = "ManuallySuppressedIssues"
+internal const val CURRENT_ISSUES = "CurrentIssues"
 internal const val ID = "ID"
 
 internal val Finding.baselineId: String
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/Baseline.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/Baseline.kt
@@ -8,9 +8,13 @@ import java.nio.file.Path
 internal typealias FindingsIdList = Set<String>
 internal typealias FindingId = String
 
-internal data class Baseline(val blacklist: FindingsIdList, val whitelist: FindingsIdList) {
+internal data class Baseline(
+    val suppressedFalsePositives: FindingsIdList,
+    val temporarySuppressedIssues: FindingsIdList
+) {
 
-    fun contains(id: FindingId): Boolean = whitelist.contains(id) || blacklist.contains(id)
+    fun contains(id: FindingId): Boolean =
+        temporarySuppressedIssues.contains(id) || suppressedFalsePositives.contains(id)
 
     companion object {
 
@@ -26,8 +30,8 @@ const val DETEKT_BASELINE_PATH_KEY = "detekt.baseline.path.key"
 const val DETEKT_BASELINE_CREATION_KEY = "detekt.baseline.creation.key"
 
 internal const val SMELL_BASELINE = "SmellBaseline"
-internal const val BLACKLIST = "Blacklist"
-internal const val WHITELIST = "Whitelist"
+internal const val SUPPRESSED_FALSE_POSITIVES = "SuppressedFalsePositives"
+internal const val TEMPORARY_SUPPRESSED_ISSUES = "TemporarySuppressedIssues"
 internal const val ID = "ID"
 
 internal val Finding.baselineId: String
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt
@@ -1,5 +1,6 @@
 package io.gitlab.arturbosch.detekt.core.baseline
 
+import io.gitlab.arturbosch.detekt.core.NotApiButProbablyUsedByUsers
 import org.xml.sax.Attributes
 import org.xml.sax.helpers.DefaultHandler
 
@@ -15,11 +16,12 @@ internal class BaselineHandler : DefaultHandler() {
     override fun startElement(uri: String, localName: String, qName: String, attributes: Attributes) {
         when (qName) {
             // Blacklist and Whitelist were previous XML tags. They have been replaced by more appropriate names
-            // To not break anything this will still parse those values
-            MANUALLY_SUPPRESSED_ISSUES, "Blacklist" -> {
+            // For the removal of these to not be a breaking change these have been implemented to be synonyms
+            // of [MANUALLY_SUPPRESSED_ISSUES] and [CURRENT_ISSUES].
+            MANUALLY_SUPPRESSED_ISSUES, BLACKLIST -> {
                 current = MANUALLY_SUPPRESSED_ISSUES
             }
-            CURRENT_ISSUES, "Whitelist" -> {
+            CURRENT_ISSUES, WHITELIST -> {
                 current = CURRENT_ISSUES
             }
             ID -> content = ""
@@ -42,5 +44,13 @@ internal class BaselineHandler : DefaultHandler() {
 
     override fun characters(ch: CharArray, start: Int, length: Int) {
         if (current != null) content += String(ch, start, length)
+    }
+
+    companion object {
+        @NotApiButProbablyUsedByUsers
+        private const val BLACKLIST = "Blacklist"
+
+        @NotApiButProbablyUsedByUsers
+        private const val WHITELIST = "Whitelist"
     }
 }
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt
@@ -14,10 +14,12 @@ internal class BaselineHandler : DefaultHandler() {
 
     override fun startElement(uri: String, localName: String, qName: String, attributes: Attributes) {
         when (qName) {
-            SUPPRESSED_FALSE_POSITIVES -> {
+            // Blacklist and Whitelist were previous XML tags. They have been replaced by more appropriate names
+            // To not break anything this will still parse those values
+            SUPPRESSED_FALSE_POSITIVES, "Blacklist" -> {
                 current = SUPPRESSED_FALSE_POSITIVES
             }
-            TEMPORARY_SUPPRESSED_ISSUES -> {
+            TEMPORARY_SUPPRESSED_ISSUES, "Whitelist" -> {
                 current = TEMPORARY_SUPPRESSED_ISSUES
             }
             ID -> content = ""
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt
@@ -7,18 +7,18 @@ internal class BaselineHandler : DefaultHandler() {
 
     private var current: String? = null
     private var content: String = ""
-    private val whiteIds = mutableSetOf<String>()
-    private val blackIds = mutableSetOf<String>()
+    private val temporarySuppressedIds = mutableSetOf<String>()
+    private val falsePositiveIds = mutableSetOf<String>()
 
-    internal fun createBaseline() = Baseline(blackIds, whiteIds)
+    internal fun createBaseline() = Baseline(falsePositiveIds, temporarySuppressedIds)
 
     override fun startElement(uri: String, localName: String, qName: String, attributes: Attributes) {
         when (qName) {
-            BLACKLIST -> {
-                current = BLACKLIST
+            SUPPRESSED_FALSE_POSITIVES -> {
+                current = SUPPRESSED_FALSE_POSITIVES
             }
-            WHITELIST -> {
-                current = WHITELIST
+            TEMPORARY_SUPPRESSED_ISSUES -> {
+                current = TEMPORARY_SUPPRESSED_ISSUES
             }
             ID -> content = ""
         }
@@ -29,12 +29,12 @@ internal class BaselineHandler : DefaultHandler() {
             ID -> {
                 check(content.isNotBlank()) { "The content of the ID element must not be empty" }
                 when (current) {
-                    BLACKLIST -> blackIds.add(content)
-                    WHITELIST -> whiteIds.add(content)
+                    SUPPRESSED_FALSE_POSITIVES -> falsePositiveIds.add(content)
+                    TEMPORARY_SUPPRESSED_ISSUES -> temporarySuppressedIds.add(content)
                 }
                 content = ""
             }
-            BLACKLIST, WHITELIST -> current == null
+            SUPPRESSED_FALSE_POSITIVES, TEMPORARY_SUPPRESSED_ISSUES -> current == null
         }
     }
 
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt
@@ -7,20 +7,20 @@ internal class BaselineHandler : DefaultHandler() {
 
     private var current: String? = null
     private var content: String = ""
-    private val temporarySuppressedIds = mutableSetOf<String>()
-    private val falsePositiveIds = mutableSetOf<String>()
+    private val currentIssues = mutableSetOf<String>()
+    private val manuallySuppressedIssues = mutableSetOf<String>()
 
-    internal fun createBaseline() = Baseline(falsePositiveIds, temporarySuppressedIds)
+    internal fun createBaseline() = Baseline(manuallySuppressedIssues, currentIssues)
 
     override fun startElement(uri: String, localName: String, qName: String, attributes: Attributes) {
         when (qName) {
             // Blacklist and Whitelist were previous XML tags. They have been replaced by more appropriate names
             // To not break anything this will still parse those values
-            SUPPRESSED_FALSE_POSITIVES, "Blacklist" -> {
-                current = SUPPRESSED_FALSE_POSITIVES
+            MANUALLY_SUPPRESSED_ISSUES, "Blacklist" -> {
+                current = MANUALLY_SUPPRESSED_ISSUES
             }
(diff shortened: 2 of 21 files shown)


ZacSweers:
I would encourage you to reconsider!

There is wider precedent for this, ietf has a great write-up here: https://tools.ietf.org/id/draft-knodel-terminology-00.html#rfc.section.1.2

Also:
1 - blacklist/whitelist can have racial connotations with "white good black bad" semantics
2 - they have no meaning to non-native speakers, as they are english-only colloquialisms

schalkms:
I understand the potential connotations of these words. I explicitly didn't negate that.

I said that changes like this have an impact on users, clients and 3rd party contributors with custom domain-specific detekt extensions. I think that compatibility and long term support are very important for static code analyzers in order to foster user acceptance. Don't forget the users in this process.

remcomokveld:
Thanks for the feedback. As mentioned in the OP it is quite easy to add backward compatibility. I just pushed a change that does that. 
Existing baseline files that contain the XML tags `Whitelist` and `Blacklist` are still supported and will be parsed properly while the new values are also supported and the tasks to write the baseline uses the new tags.

schalkms:
I still think the names `TemporarySuppressedIssues` and `SuppressedFalsePositives` are weird. Why is a whitelist a temporary suppressed issue and blacklist a suppressed false positive? The names would need an explanation in a separate section. I could live with allow and deny list from the user perspective.

remcomokveld:
In the docs it says this
> The intention of a whitelist is that only new code smells are printed on further analysis. The blacklist can be used to write down false positive detections (instead of suppressing them and pollute your code base).

I agree that `TemporarySuppressedIssues` might not be the right term, but the goal of those is to temporarily suppress detekt issues that are found when first adding detekt. Those issues should be considered tech debt and eventually cleaned up, hence temporary.

This actually show that blacklist and whitelist might were bad names to begin with. a whitelist and a blacklist are opposites, one is traditionally used to include things and the other to exclude. In Detekt they are functionally doing the same thing.

schalkms:
@remcomokveld thanks for pointing that out. What do you think of `allow` and `deny` list? Do you have other terms/words in your mind?

arturbosch:
Hi,

thanks for the discussion here.
My first thought was "good change for a 2.0 release" because of breaking change
but implementing this in a backwards compatible way should be safe for a minor release.

@schalkms @remcomokveld @ZacSweers the NCSC uses the terms allow and deny lists as @schalkms mentioned - https://inews.co.uk/news/government-cyber-experts-blacklist-whitelist-racism-fears-2840970

Open question:
Should we automatically migrate the baselines to the new names or if the user used the old names it should stay old until 2.0 so this does not cause an additional git commit for them?

remcomokveld:
They would both be `allow` because in both lists, if an issue is added there it will allow detekt to succeed even though the issue is there. The only difference between the two lists is communicating reason to developers why the issues are ignored.

The only place where these lists are read is here
https://github.com/detekt/detekt/blob/98b89f4f1df4c05186de588d68d17904d2861874/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/Baseline.kt#L13

schalkms:
> Should we automatically migrate the baselines to the new names or if the user used the old names it should stay old until 2.0 so this does not cause an additional git commit for them?

I'm voting for **option 2** here.

remcomokveld:
I agree with option 2 as well. The only thing that might be nice to add to this is a warning message when deprecated `<Whitelist>` and `<Blacklist>` tags are being used. That would give users more of a heads up that a breaking change will happen in 2.0 and already start migrating.

cortinico:
Thanks for the change @remcomokveld 

I also think that this is the opportunity to remove some of the confusion around the current baseline. As you already mentioned, `<Whitelist>` and `<Blacklist>` are "opposite" terms while they are both used to let detekt ignore a specific issue report.

Given that when running `detektBaseline` the former `<Whitelist>` is populated, maybe we can rename them to something like:

- `Whitelist` -> `CurrentIssues` or `DiscoveredIssue` to convey the meaning that those issue have been discovered by Detekt and stored inside the baseline. This will suggest the user that they should slowly clean the `CurrentIssue` list.

- `Blacklist` -> `ManuallySuppressedIssues` to convey the meaning that those issues have been intentionally suppressed by the user.

I would argue against using `Temporary` in any of the two as having time related terminology here is not a great fit (also there is nothing more permanent than a temporary solution).

As for the `SuppressedFalsePositive` (aka the `Whitelist`), technically you could use it to suppress a failure that is not a detekt false positive. The rule might just behave correctly but your specific code imposes you to suppress it for some reason, and I don't consider that a false positive.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt
```diff
@@ -7,18 +7,20 @@ internal class BaselineHandler : DefaultHandler() {
...
-    internal fun createBaseline() = Baseline(blackIds, whiteIds)
+    internal fun createBaseline() = Baseline(falsePositiveIds, temporarySuppressedIds)
 
     override fun startElement(uri: String, localName: String, qName: String, attributes: Attributes) {
         when (qName) {
-            BLACKLIST -> {
-                current = BLACKLIST
+            // Blacklist and Whitelist were previous XML tags. They have been replaced by more appropriate names
+            // To not break anything this will still parse those values
+            SUPPRESSED_FALSE_POSITIVES, "Blacklist" -> {
```
arturbosch, line 19: Lets store which name was used for the element so we can go with option 2 and write it back on save.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/Baseline.kt
```diff
@@ -8,9 +8,13 @@ import java.nio.file.Path
 internal typealias FindingsIdList = Set<String>
 internal typealias FindingId = String
 
-internal data class Baseline(val blacklist: FindingsIdList, val whitelist: FindingsIdList) {
+internal data class Baseline(
```
arturbosch, line 11: The baseline class will need an additional object storing the names for suppressed/temporary list sections.

This structure can later be used in `fun XMLStreamWriter.save(baseline: Baseline)` to write the original used names.

arturbosch:
> Thanks for the change @remcomokveld
> 
> I also think that this is the opportunity to remove some of the confusion around the current baseline, and be more inclusive. As you already mentioned, `<Whitelist>` and `<Blacklist>` are "opposite" terms while they are both used to let detekt ignore a specific issue report.
> 
> Given that when running `detektBaseline` the former `<Whitelist>` is populated, maybe we can rename them to something like:
> 
> * `Whitelist` -> `CurrentIssues` or `DiscoveredIssue` to convey the meaning that those issue have been discovered by Detekt and stored inside the baseline. This will suggest the user that they should slowly clean the `CurrentIssue` list.
> * `Blacklist` -> `ManuallySuppressedIssues` to convey the meaning that those issues have been intentionally suppressed by the user.
> 
> I would argue against using `Temporary` in any of the two as having time related terminology here is not a great fit (also there is nothing more permanent than a temporary solution).
> 
> As for the `SuppressedFalsePositive` (aka the `Whitelist`), technically you could use it to suppress a failure that is not a detekt false positive. The rule might just behave correctly but your specific code imposes you to suppress it for some reason, and I don't consider that a false positive.

I like all three names. Lets go with `CurrentIssues` and `ManuallySuppressedIssues`?
And when we are at renaming maybe change `SmellBaseline` to just `Baseline` (ofc can be another PR)?

BraisGabin:
> > Should we automatically migrate the baselines to the new names or if the user used the old names it should stay old until 2.0 so this does not cause an additional git commit for them?
> 
> I vote for **option 2** here.

I vote option 1. We will generate a bit of noise in git but all those users will have their baselines migrated so once they move to 2.0 it will just work.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt
remcomokveld: I'm not sure I understand why it would need to be saved. When a new baseline file is written, that would need to be checked in in any case, why would you want to write it with the old tags if you can just as well write it with the new tags?

remcomokveld: Oh, I now realise what you meant with option 1 vs option 2. In that case I'd rather go for option 1

arturbosch:
> > > Should we automatically migrate the baselines to the new names or if the user used the old names it should stay old until 2.0 so this does not cause an additional git commit for them?
> > 
> > 
> > I vote for **option 2** here.
> 
> I vote option 1. We will generate a bit of noise in git but all those users will have their baselines migrated so once they move to 2.0 it will just work.
> 
> And thanks for the PR!

Hm, well to be honest. If the user updates the detekt version he already has to create one commit.
Including baseline tag renamings should be fine.
As I now realize this will only happen for `gradle detektBaseline` not `gradle detekt` so it is okay.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt
```diff
@@ -7,18 +7,20 @@ internal class BaselineHandler : DefaultHandler() {
 
     private var current: String? = null
     private var content: String = ""
-    private val whiteIds = mutableSetOf<String>()
-    private val blackIds = mutableSetOf<String>()
+    private val temporarySuppressedIds = mutableSetOf<String>()
+    private val falsePositiveIds = mutableSetOf<String>()
```
cortinico, line 11: Please keep the name of those private fields consistent with the XML names.

```suggestion
    private val manuallySuppressedIssues = mutableSetOf<String>()
    private val currentIssues = mutableSetOf<String>()
```

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/Baseline.kt
```diff
@@ -26,8 +30,8 @@ const val DETEKT_BASELINE_PATH_KEY = "detekt.baseline.path.key"
 const val DETEKT_BASELINE_CREATION_KEY = "detekt.baseline.creation.key"
 
 internal const val SMELL_BASELINE = "SmellBaseline"
-internal const val BLACKLIST = "Blacklist"
-internal const val WHITELIST = "Whitelist"
+internal const val MANUALLY_SUPRESSED_ISSUES = "ManuallySuppressedIssues"
```
cortinico, line 33:
```suggestion
internal const val MANUALLY_SUPPRESSED_ISSUES = "ManuallySuppressedIssues"
```

cortinico: I'm not a native speaker, but the correct spelling should be `suppressed`

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt
remcomokveld: Oops, I thought I had all of them, but missed one. Nice catch!

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/Baseline.kt
remcomokveld: Yeah you're right, actual value was spelled but made a typo in the constant

remcomokveld:
@arturbosch any clue what is causing the checks to fail now? I'm not seeing what exactly went wrong

arturbosch:
> @arturbosch any clue what is causing the checks to fail now? I'm not seeing what exactly went wrong

I have not the slightest clue :(.
I've restarted the job and it still fails so it might be reproducible.
Could you add `--stacktrace` to `./gradlew verifyGeneratorOutput` in `.github/workflows/pre-merge.yaml` please?

docs/pages/baseline.md
```diff
@@ -19,14 +19,14 @@ the `--report txt:path/to/report` cli flag.
 
 ```xml
 <SmellBaseline>
-  <Blacklist>
+  <ManuallySuppressedIssues>
     <ID>CatchRuntimeException:Junk.kt$e: RuntimeException</ID>
-  </Blacklist>
-  <Whitelist>
+  </ManuallySuppressedIssues>
+  <CurrentIssues>
```
xanderblinov, line 25: It is not clear what means "CurrentIssues". Could you please find another name?

docs/pages/changelog.md
```diff
@@ -1704,8 +1704,8 @@ See all issues at: [M7](https://gitlab.com/arturbosch/detekt/milestones/7)
 ##### New features
 
 - allow to fail builds on code smell thresholds (configurable), see 'Configure build failure thresholds' in README
-- blacklist code smells which are false positives
-- generate a baseline whitelist of code smells if your project is already old and has many smells, so only new
+- suppress code smells which are false positives
+- generate a baseline of code smells if your project is already old and has many smells, so only new
```
xanderblinov, line 1708: missed world. baseline of what?

docs/pages/documentation/style.md
```diff
@@ -1652,8 +1652,8 @@ which classes are imported and helps prevent naming conflicts.
 Library updates can introduce naming clashes with your own classes which might result in compilation errors.
 
 **NOTE:** This rule is effectively overridden by the `NoWildcardImports` formatting rule (a wrapped ktlint rule).
-That rule will fail the check regardless of the whitelist configured here.
-Therefore if whitelist is needed `NoWildcardImports` rule should be disabled.
+That rule will fail the check regardless of the baseline configured here.
+Therefore if baseline is needed `NoWildcardImports` rule should be disabled.
```
xanderblinov, line 1656: incorrect replacement

dimsuz:
> I said that changes like this have an impact on users, clients and 3rd party contributors with custom domain-specific detekt extensions. I think that compatibility and long term support are very important for static code analyzers in order to foster user acceptance. Don't forget the users in this process.

I second this. If tomorrow will bring another set of words considered inappropriate in the current momentum, would this mean all project should ignore their users and just blindly go on with whatever is demanded?

Here's a [good answer](https://github.com/rubocop-hq/rubocop/issues/8091#issuecomment-640158172) from an opensource project maintainer on a similar issue.

docs/pages/baseline.md
dimsuz: I agree. The context is ambiguous here. I might have some other "current" issues, which are not in the baseline.

ZacSweers:
@dimsuz I recognize the point you're trying to make regarding churn and making invasive changes for what you feel may be trivial changes, and I appreciate your effort to engage in this honestly and from a position of pragmatism. If this were a significant breaking change, it would be valid to suggest waiting until a major release to change. If these came frequently, it could be valid to suggest being tactical about what changes are meaningful. In this specific PR however, there are no breaking changes. It is a free swing at doing something that _is_ meaningful to a nontrivial number of users that do face endemic terminology like this every day.

If social progress were as slippery of a slope as you're suggesting, we'd be in a much better place today than we are. That is not the case though, not yet. In light of that, we should aspire to embrace opportunities progress in all its forms, whenever and if ever they arrive on our doorstep.

EKnizhnik:
Blacklist and whitelist words have nothing to do with racism. Or any other social problems. It is just words. Pretty understandable from professional point of view. Without any additional meaning. 
Any change like that would be problematic at  very least. It solves absolutely no problem and makes no sense at all.

bendb-instacart:
> Pretty understandable from professional point of view.
I think it's been demonstrated in this thread that the meanings of blacklist and whitelist are _not_ clear in detekt, and that the proposed names of "ManuallySuppressedIssues" and "CurrentIssues" capture the meanings more clearly.

EKnizhnik:
> > Pretty understandable from professional point of view.
> 
> I think it's been demonstrated in this thread that the meanings of blacklist and whitelist are _not_ clear in detekt, and that the proposed names of "ManuallySuppressedIssues" and "CurrentIssues" capture the meanings more clearly.

Generally, if community decides, that these words are more understandable - no objections from my side. I’m against “these words are racists“ motivation.

docs/pages/baseline.md
remcomokveld: CurrentIssues without any context would indeed be confusing. In the context of a baseline file, however, I think that the name captures exactly what is intended. If you have a better name I'm open to suggestions though

dimsuz:
If everybody feels _that_ sensitive to words they write once and forget and it's not a major breaking change for users then at least we need to find a good replacements. As I've already noted "CurrentIssues" is not a suitable one imo. I'm currently failing to come up with a good one though, maybe someone would? 

By the way, are you sure that these are non-breaking changes? If I have a `baseline.xml` and update to a newer detekt version, it seems like old tag names won't be picked up anymore.

remcomokveld:
In the parsing logic the old and the new tags are treated as synonyms.
https://github.com/detekt/detekt/blob/d37b9d84454ad3db89f94b2ec4ffbdf444cbe378/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt#L17-L19
There was also a test added which verifies that the old baseline format still parses correctly here
https://github.com/detekt/detekt/blob/d37b9d84454ad3db89f94b2ec4ffbdf444cbe378/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFormatSpec.kt#L42-L51

jeksor:
What about blockList, allowList?

BraisGabin:
> What about blockList, allowList?

Both lists are "allowList". The only difference between them is that the values in `CurrentIssues` are discarted and generated again each time you run `--create-baseline` and the ones in `ManuallySuppressedIssues` are keeped.

schalkms:
Based on the reactions this seems to be a controversial change.
What can we do about that? Which improvements do people suggest? What does the community think?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineHandler.kt
```diff
@@ -7,18 +7,20 @@ internal class BaselineHandler : DefaultHandler() {
...
-    internal fun createBaseline() = Baseline(blackIds, whiteIds)
+    internal fun createBaseline() = Baseline(manuallySuppressedIssues, currentIssues)
 
     override fun startElement(uri: String, localName: String, qName: String, attributes: Attributes) {
         when (qName) {
-            BLACKLIST -> {
-                current = BLACKLIST
+            // Blacklist and Whitelist were previous XML tags. They have been replaced by more appropriate names
+            // To not break anything this will still parse those values
+            MANUALLY_SUPPRESSED_ISSUES, "Blacklist" -> {
```
arturbosch, line 19: Please extract the old names again to constants and mark them with the annotation `@NotApiButProbablyUsedByUsers`

docs/pages/documentation/style.md
arturbosch:
``` 
* **NOTE:** This rule has a twin implementation `NoWildcardImports` in the formatting rule set (a wrapped KtLint rule).
 * When suppressing an issue of `WildcardImport` in the baseline file, make sure to suppress the corresponding `NoWildcardImports` issue.
 *
```

arturbosch:
> Based on the reactions this seems to be a controversial change.
> What can we do about that? Which improvements do people suggest? What does the community think?

Judging by the amount of comments and emojis I see a clear trend towards changing these tags.
Imo we now have much better descriptive names as before (also when looking back on many questions I got about the difference between what to write where in the baseline file).

I want to prepare a 1.10.0-RC1 release tonight. So if most are happy with `CurrentIssues`, we can ship this PR too.

blacklistisnotracist:
@remcomokveld You can start repainting The White House because it's a symbol of white supremacy.

docs/pages/changelog.md
rooksoto: Consider:
`- generate a baseline of code smells if your project is already old and has many smells, so only new`

arturbosch:
@schalkms anything else you see missing in this PR to get it merged?

schalkms:
There's a merge conflict. Documentation has been updated as far as I can see.

docs/pages/baseline.md
blundell: I'm too late to the party, but I also think CurrentIssues is a very strange name, there was a previous comment to "avoid the concept of time", but "current" is all about time.

I think the relationship should be reinstated between the two type of ignore, that way you can infer the meaning of one by understanding the other.

Perhaps these'll be changed in the future or perhaps never again, here's my input anyway:

`BaselineSuppressedIssues`
`ManuallySuppressedIssues`

Issue #2701, opened 2020-05-17
Restrict non-boolean properties with is prefix


imanushin:
## Expected Behavior of the rule
Deny property, which starts with ```is``` (e.g. name is like ```isDefault```), however the return value isn't ```Boolean```.

E.g. the right property sample: ```isValid: Boolean```. The invalid property sample: ```isValid: String```.

## Context
According to Java (not Kotlin) rules, only boolean properties should start with ```is```. Kotlin has the following behavior: for property ```isXXX```, the getter will be ```isXXX```, not ```getIsXXX```.

For example, for ```class Temp(val isValid: String)``` resulting decompiled bytecode ([please check via this solution](https://stackoverflow.com/a/40762755/185498)) will be:
```java
public final class Temp {
   @NotNull
   private final String isValid;

   @NotNull
   public final String isValid() {
      return this.isValid;
   }

   public Temp(@NotNull String isValid) {
      Intrinsics.checkParameterIsNotNull(isValid, "isValid");
      super();
      this.isValid = isValid;
   }
}
```

From Java Language Specification, ```is``` prefix is allowed for ```Boolean``` properties only, please check chapter 8.3.2 at [the document](https://download.oracle.com/otn-pub/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/beans.101.pdf?AuthParam=1589736840_2ff06f9f65563391bd83768de1b0ab41).

**Finally: the method ```isValid```, which returns ```String```, isn't correctly named.**

## Where it can be
Here is just example, this isn't generic behavior.

Let's imagine, that you want to send json into Oracle and parse it there (to achieve bulk operations benefits).

On the interoperation layer, some classes could be used just for serialization/deserialization from Json input/output. So such values can be allowed there (just for interoperation idea). For example, [Oracle doesn't support Boolean datatype](https://www.oreilly.com/library/view/oracle-plsql-programming/0596003811/ch12.html). So if you want to send json into Oracle and deserialize it there, you should use ```NUMBER(1)``` or ```CHAR(1)``` (e.g. you have to send 0/1 or Y/N to Oracle).

For example, for Kotlin dto ```class SomeValue(val isDefault: Boolean, val type: String)```, you have to create special dto for Oracle (you can't use Boolean). So, Oracle-friendly dto will be ```class SomeValue(val isDefault: Int, val type: String)``` (```isDefault``` can be only 0 or 1).

And this class is wrong, because:
[...]

PR #2779, opened 2020-06-07
Feature/restrict is properties

imanushin:
Fix for #2701 

What I covered:
1. Property/arguments with val/var shouldn't start with "is" for non-boolean types
2. ```isengard``` property name is allowed for any type, ```isDefault``` is disallowed
3. Generic types are out of scope (to be formal, it is better to prohibit them too, however this question isn't so simple for me, need help)

This rule is enabled by default in the naming ruleset, it doesn't have any parameters.


cortinico:
Thanks for your contribution @imanushin 🎉 

The CI is red because there are a couple of detekt failures on your code (`NoWildcardImports`, etc.) plus the documentation needs to be re-generated.

If you run `./gradlew build` on your branch you'll get the same failures and the docs will be re-generated. Once your build is green locally, the CI should also be green.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/IsPropertyNaming.kt
```diff
@@ -0,0 +1,73 @@
...
+import io.gitlab.arturbosch.detekt.rules.identifierName
+import org.jetbrains.kotlin.js.descriptorUtils.getJetTypeFqName
+import org.jetbrains.kotlin.psi.KtCallableDeclaration
+import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.resolve.typeBinding.createTypeBindingForReturnType
+
+/**
+ * Reports when property names inside objects which do not follow the specified naming convention are used.
+ *
```
cortinico, line 13: Please provide an example of compliant/noncompliant code like the other rules are doing:

```
 * <noncompliant>
 * var isSlow : Int = 500
 * </noncompliant>
 *
 * <compliant>
 * var isSlow : Boolean = false
 * </compliant>
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/IsPropertyNaming.kt
```diff
@@ -0,0 +1,73 @@
...
+        "Only boolean property names can start with 'is' prefix.",
+        debt = Debt.FIVE_MINS
+    )
+
+    override fun visitParameter(parameter: KtParameter) {
+        if (parameter.hasValOrVar()) {
+            validateDeclaration(parameter)
+        }
+
+        super.visitParameter(parameter)
```
cortinico, line 33: Call to the superclass should always be first, unless you have a valid reason

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/IsPropertyNaming.kt
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    override fun visitProperty(property: KtProperty) {
+        validateDeclaration(property)
+
+        super.visitProperty(property)
```
cortinico, line 39: _same as above_

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/IsPropertyNaming.kt
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private fun validateDeclaration(declaration: KtCallableDeclaration) {
+        val name = declaration.identifierName()
+
+        if (name.startsWith("is") && name.length > 2 && !name[2].isLowerCase()) {
+            val typeName = getTypeName(declaration)
+
+            if (typeName !== null && typeName != kotlinBooleanTypeName && typeName != javaBooleanTypeName) {
```
cortinico, line 48:
```suggestion
            if (typeName != null && typeName != kotlinBooleanTypeName && typeName != javaBooleanTypeName) {
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/IsPropertyNaming.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    override fun visitProperty(property: KtProperty) {
+        validateDeclaration(property)
+
+        super.visitProperty(property)
+    }
+
+    private fun validateDeclaration(declaration: KtCallableDeclaration) {
+        val name = declaration.identifierName()
+
```
cortinico, line 44: This rule is using type and symbol resolution (you're accessing a `bindingContext` inside the `getTypeName` function).

You should add this check in your rule body:

```kotlin
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/IsPropertyNamingSpec.kt
```diff
@@ -0,0 +1,202 @@
...
+            """
+                val findings = subject.compileAndLintWithContext(env, code)
+
+                assertThat(findings).isEmpty()
+            }
+
+            it("should warn about primitive types") {
+                val code = """
+                class O {
+                    var isDefault: Int = 0
```
cortinico, line 158: Can you provide a test case where the type is inferred?

```
var isDefault = 0
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/IsPropertyNaming.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.jetbrains.kotlin.js.descriptorUtils.getJetTypeFqName
+import org.jetbrains.kotlin.psi.KtCallableDeclaration
+import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.resolve.typeBinding.createTypeBindingForReturnType
+
+/**
+ * Reports when property names inside objects which do not follow the specified naming convention are used.
+ *
+ * @active since v1.9.0
```
cortinico, line 14:
```suggestion
 * @active since v1.10.0
```

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -389,6 +389,8 @@ naming:
     constantPattern: '[A-Za-z][_A-Za-z0-9]*'
     propertyPattern: '[A-Za-z][_A-Za-z0-9]*'
     privatePropertyPattern: '(_)?[A-Za-z][_A-Za-z0-9]*'
+  IsPropertyNaming:
+    active: true
```
cortinico, line 393: I'm unsure if it was discussed somewhere that this rule is active by default

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/IsPropertyNamingSpec.kt
cortinico: Also a test for a non primitive type would be useful:

```
var isKnown = listOf(1,2,3)
```

BraisGabin:
Usually we don't activate a new rule by default. It's safer to have it disable by default and get feedback from the users that enabled it. If there are not issues we can activate it later.

arturbosch:
Please make sure to rebase on top of #2782 when the PR is merged :muscle:

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -377,6 +377,9 @@ naming:
   InvalidPackageDeclaration:
     active: false
     rootPackage: ''
```
imanushin, line 379: This file is autogenerated with "active:true" suggestion. Could you please help, is it correct?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/IsPropertyNaming.kt
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * <noncompliant>
+ * val isEnabled : Int = 500
+ * </noncompliant>
+ *
+ * <compliant>
+ * val isEnabled : Boolean = false
+ * </compliant>
+ *
+ * @active since v1.10.0
```
arturbosch, line 30: When using the `@active` tag the generator prints `active: true` for the default detekt config.
Just remove that line and everything should be finde :).

imanushin:
Finally I got green build, however the following test was removed. Could you please help, do you know why it isn't work on ubuntu linux (it perfectly works at my Windows 10 computer)?

Removed test:
```kotlin
it("should not detect Java Boolean") {
    val code = """
    class O {
        var isDefault: java.lang.Boolean = java.lang.Boolean.FALSE // replacing to "false" doesn't work too
    }
"""
    
}
```

cortinico:
> Could you please help, do you know why it isn't work on ubuntu linux (it perfectly works at my Windows 10 computer)?

I guess the failure was due to the fact that on the build instance `ubuntu-latest`+`java8` we also compile the test snippets (to make sure you're actually writing valid Kotlin code inside the snippets). We do it only there as it will slow down the CI too much.

Given that you rebase on top of master, now you see that there is a job called `compile-test-snippets` that will do exactly that.

To test the same scenario on your machine you can invoke:
```
gw test -x ":detekt-gradle-plugin:test" -Pcompile-test-snippets=true
```

That will compile the snippets and tell you what's wrong with your code.

As an alternative, you can use play.kotlinlang.org: https://pl.kotl.in/W9SFqrlPJ
Just copy your snippet there and it should be all green. In your specific case, you can replace with:

```kotlin
class O {
    var isDefault: java.lang.Boolean = java.lang.Boolean(false)
}
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/IsPropertyNaming.kt
```diff
@@ -0,0 +1,91 @@
...
+import io.gitlab.arturbosch.detekt.rules.identifierName
+import org.jetbrains.kotlin.js.descriptorUtils.getJetTypeFqName
+import org.jetbrains.kotlin.psi.KtCallableDeclaration
+import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.typeBinding.createTypeBindingForReturnType
+
+/**
+ * Reports when property with 'is' prefix doesn't boolean type.
```
cortinico, line 19:
```suggestion
 * Reports when property with 'is' prefix doesn't have a boolean type.
```

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/IsPropertyNaming.kt
```diff
@@ -0,0 +1,91 @@
...
+ * Please check the charter 8.3.2 at Java Language Specification
```
cortinico, line 20:
```suggestion
 * Please check the [chapter 8.3.2 at Java Language Specification](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.3.2)
```

imanushin:
@arturbosch , much appreciated the text fix.

I'll return tests today evening (I couldn't push from working computer, sorry). And I'm ok with merging now, I can add the tests in the separate PR.

imanushin:
Documentation was regenerated, tests were reverted.

PR #2787, opened 2020-06-11
Add automatic detekt tasks for Android Plugins

realdadfish:
The Gradle Plugin was missing automatic task generation for Android modules, so I tried adding them.

While this works in general, I still get a couple of compilation issues when running the task on different modules, mainly things that stem from code generation:

```kotlin
/DownloadApiFactory.kt:7:24: error: unresolved reference: BuildConfig
import my.module.BuildConfig
                       ^
/DownloadApiFactory.kt:28:52: error: unresolved reference: BuildConfig
        .client(downloadOkHttpFactory.createClient(BuildConfig.DEBUG, readTimeOutSeconds))
```

### DIFF
--- a/detekt-gradle-plugin/build.gradle.kts
+++ b/detekt-gradle-plugin/build.gradle.kts
@@ -3,8 +3,18 @@ plugins {
     id("com.gradle.plugin-publish") version "0.12.0"
 }
 
+repositories {
+    google()
+}
+
 dependencies {
+    val androidGradlePlugin = "com.android.tools.build:gradle:4.0.1"
     implementation(kotlin("gradle-plugin-api"))
+    compileOnly(androidGradlePlugin)
+
+    testImplementation(project(":detekt-test-utils"))
+    testImplementation(kotlin("gradle-plugin"))
+    testImplementation(androidGradlePlugin)
 }
 
 gradlePlugin {
@@ -25,7 +35,7 @@ pluginBundle {
     website = "https://detekt.github.io/detekt"
     vcsUrl = "https://github.com/detekt/detekt"
     description = "Static code analysis for Kotlin"
-    tags = listOf("kotlin", "detekt", "code-analysis", "linter", "codesmells")
+    tags = listOf("kotlin", "detekt", "code-analysis", "linter", "codesmells", "android")
 
     (plugins) {
         "detektPlugin" {
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -1,26 +1,22 @@
 package io.gitlab.arturbosch.detekt
 
 import io.gitlab.arturbosch.detekt.extensions.DetektExtension
-import io.gitlab.arturbosch.detekt.extensions.DetektReport
-import org.gradle.api.GradleException
+import io.gitlab.arturbosch.detekt.internal.DetektAndroid
+import io.gitlab.arturbosch.detekt.internal.DetektJvm
+import io.gitlab.arturbosch.detekt.internal.registerCreateBaselineTask
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.file.FileCollection
-import org.gradle.api.internal.HasConvention
-import org.gradle.api.plugins.JavaPluginConvention
 import org.gradle.api.plugins.ReportingBasePlugin
 import org.gradle.api.provider.Provider
 import org.gradle.api.reporting.ReportingExtension
-import org.gradle.api.tasks.SourceSet
 import org.gradle.language.base.plugins.LifecycleBasePlugin
-import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
-import java.io.File
 
 class DetektPlugin : Plugin<Project> {
 
     override fun apply(project: Project) {
         project.pluginManager.apply(ReportingBasePlugin::class.java)
-        val extension = project.extensions.create(DETEKT_TASK_NAME, DetektExtension::class.java)
+        val extension = project.extensions.create(DETEKT_EXTENSION, DetektExtension::class.java)
         extension.reportsDir = project.extensions.getByType(ReportingExtension::class.java).file("detekt")
 
         val defaultConfigFile =
@@ -32,25 +28,27 @@ class DetektPlugin : Plugin<Project> {
         configurePluginDependencies(project, extension)
         setTaskDefaults(project)
 
-        registerOldDetektTask(project, extension)
-        registerDetektTasks(project, extension)
-        registerCreateBaselineTask(project, extension)
-        registerGenerateConfigTask(project)
+        project.registerOldDetektTask(extension)
+        project.registerDetektJvmTasks(extension)
+        project.registerDetektAndroidTasks(extension)
+        project.registerOldCreateBaselineTask(extension)
+        project.registerGenerateConfigTask()
     }
 
-    private fun registerDetektTasks(project: Project, extension: DetektExtension) {
-        // Kotlin JVM plugin
-        project.plugins.withId("org.jetbrains.kotlin.jvm") {
-            project.afterEvaluate {
-                project.convention.getPlugin(JavaPluginConvention::class.java).sourceSets.all { sourceSet ->
-                    registerDetektTask(project, extension, sourceSet)
-                }
-            }
+    private fun Project.registerDetektJvmTasks(extension: DetektExtension) {
+        plugins.withId("org.jetbrains.kotlin.jvm") {
+            DetektJvm(this).registerDetektJvmTasks(extension)
         }
     }
 
-    private fun registerOldDetektTask(project: Project, extension: DetektExtension) {
-        val detektTaskProvider = project.tasks.register(DETEKT_TASK_NAME, Detekt::class.java) {
+    private fun Project.registerDetektAndroidTasks(extension: DetektExtension) {
+        plugins.withId("kotlin-android") {
+            DetektAndroid(this).registerDetektAndroidTasks(extension)
+        }
+    }
+
+    private fun Project.registerOldDetektTask(extension: DetektExtension) {
+        val detektTaskProvider = tasks.register(DETEKT_TASK_NAME, Detekt::class.java) {
             it.debugProp.set(project.provider { extension.debug })
             it.parallelProp.set(project.provider { extension.parallel })
             it.disableDefaultRuleSetsProp.set(project.provider { extension.disableDefaultRuleSets })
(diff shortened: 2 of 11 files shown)


realdadfish:
Hrm... I still seem to be doing something else wrong. When this task is run, my custom rule in my project fails to execute, because it cannot determine the absolute path of the incoming `KtFile`, i.e. `ktFile.getUserData(Key<String>("absolutePath"))` returns null. Any idea here?

realdadfish:
I eventually opened https://issuetracker.google.com/issues/158777988 for the `BuildConfig` issue above, in case I did not oversee anything in the existing AGP 4.0 API how to retrieve the generated source classpath into Detekt.

arturbosch:
> Hrm... I still seem to be doing something else wrong. When this task is run, my custom rule in my project fails to execute, because it cannot determine the absolute path of the incoming `KtFile`, i.e. `ktFile.getUserData(Key<String>("absolutePath"))` returns null. Any idea here?

Please try to rebase on current master.
We have moved some code for 1.10.0 which running 1.9.1 with newer extensions won't handle.
Does this fix it?

realdadfish:
Build is broken - the new `detekt-bom` artefact is apparently missing and not published. Working around that.

```
Could not parse POM /Users/thomaskeller/.m2/repository/io/gitlab/arturbosch/detekt/detekt-gradle-plugin/1.10.0-RC1-android-gradle-plugin/detekt-gradle-plugin-1.10.0-RC1-android-gradle-plugin.pom
            > Could not find io.gitlab.arturbosch.detekt:detekt-bom:1.10.0-RC1-android-gradle-plugin.
```

realdadfish:
Ok, I manually added the missing POM file for `detekt-bom` and got it working:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>io.gitlab.arturbosch.detekt</groupId>
  <artifactId>detekt-bom</artifactId>
  <version>1.10.0-RC1</version>
  <name>detekt</name>
  <packaging>pom</packaging>
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>com.beust</groupId>
        <artifactId>jcommander</artifactId>
        <version>1.78</version>
      </dependency>
      <dependency>
        <groupId>org.yaml</groupId>
        <artifactId>snakeyaml</artifactId>
        <version>1.26</version>
      </dependency>
      <dependency>
        <groupId>org.jetbrains.kotlinx</groupId>
        <artifactId>kotlinx-html-jvm</artifactId>
        <version>0.7.1</version>
      </dependency>
    </dependencies>
  </dependencyManagement>
</project>
```

realdadfish:
> Please try to rebase on current master.
> We have moved some code for 1.10.0 which running 1.9.1 with newer extensions won't handle.
> Does this fix it?

This solved the issue with my custom rule not finding the KtFile's path. Analysis runs through in general, just that I still see a lot of compilation errors beforehand (that don't break the task). I don't know how exactly Detekt's type resolution works, but I guess issues that stem from synthetic imports won't be resolvable, because the compiler plugin generating code for those doesn't place that generated code anywhere visible.

```
> Task :ui:detektDebug
/BarcodeScannerDialogFragment.kt:24:24: error: unresolved reference: synthetic
import kotlinx.android.synthetic.main.dialog_fragment_barcode_scanner.camera_container
```

arturbosch:
> Ok, I manually added the missing POM file for `detekt-bom` and got it working:
> 
> ```
> <?xml version="1.0" encoding="UTF-8"?>
> <project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
>   <modelVersion>4.0.0</modelVersion>
>   <groupId>io.gitlab.arturbosch.detekt</groupId>
>   <artifactId>detekt-bom</artifactId>
>   <version>1.10.0-RC1</version>
>   <name>detekt</name>
>   <packaging>pom</packaging>
>   <dependencyManagement>
>     <dependencies>
>       <dependency>
>         <groupId>com.beust</groupId>
>         <artifactId>jcommander</artifactId>
>         <version>1.78</version>
>       </dependency>
>       <dependency>
>         <groupId>org.yaml</groupId>
>         <artifactId>snakeyaml</artifactId>
>         <version>1.26</version>
>       </dependency>
>       <dependency>
>         <groupId>org.jetbrains.kotlinx</groupId>
>         <artifactId>kotlinx-html-jvm</artifactId>
>         <version>0.7.1</version>
>       </dependency>
>     </dependencies>
>   </dependencyManagement>
> </project>
> ```

The bom will be published with #2814

detekt-gradle-plugin/build.gradle.kts
```diff
@@ -3,9 +3,14 @@ plugins {
...
 }
 
+repositories {
+    google()
+}
+
 dependencies {
     implementation(kotlin("gradle-plugin"))
     implementation(kotlin("gradle-plugin-api"))
+    implementation("com.android.tools.build:gradle:4.0.0")
```
arturbosch, line 13: We should change all these dependencies to `compileOnly`.

The detekt plugin should dynamically test if the android plugin was applied and react to it.
The user may have a different conflicting version of this dependency.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -31,25 +41,66 @@ class DetektPlugin : Plugin<Project> {
...
+        project.registerCreateBaselineTask(extension)
+        project.registerGenerateConfigTask()
     }
 
-    private fun registerDetektTasks(project: Project, extension: DetektExtension) {
+    private fun Project.registerDetektTasks(extension: DetektExtension) {
+        // There is not a single Android plugin, but each registers an extension based on BaseExtension,
+        // so we catch them all by looking for this one
+        project.afterEvaluate {
+            val androidExtension = project.extensions.findByType(BaseExtension::class.java)
```
arturbosch, line 54: When changing to compileOnly dependency this will crash.

Please take a look how the KtLint gradle plugin solves this problem: https://github.com/JLLeitschuh/ktlint-gradle/blob/master/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt#L39

realdadfish:
> Does android need the synthetic props/files while running `detekt` or for a sourceset?

Can't think of any, no. Just the source files should be enough IMHO.

realdadfish:
I'll come back to this in a few weeks, right now it's release crunching time here and I have no more time to spent on this PR. I'll keep it open until then and then rebase against the recent master.

realdadfish:
Would be happy if somebody could re-review this. Thanks!

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/KotlinExtension.kt
```diff
@@ -4,8 +4,27 @@ package io.gitlab.arturbosch.detekt
...
 import io.gitlab.arturbosch.detekt.extensions.DetektExtension
 import org.gradle.api.Project
+import org.gradle.api.tasks.TaskProvider
 
 @Deprecated("""Either apply detekt plugin to root project or follow this advice:
 https://docs.gradle.org/current/userguide/kotlin_dsl.html#sec:multi_project_builds_applying_plugins""")
 fun Project.detekt(configure: DetektExtension.() -> Unit) =
     extensions.configure(DetektExtension::class.java, configure)
+
+fun Project.registerDetektTask(
```
arturbosch, line 14: This is shared between the plugins?
As it is public lets move it to the `internal` package.

realdadfish:
> The code looks good to me.
> Could you elaborate why we technically need two Gradle plugins to support Android?
> 
> Shouldn't `project.plugins.withId("kotlin-android") {...}` to enough to not run into android specific code?

I'm not sure if I understand the question. The original reviewer's comment was that the Android-specific things should not become a burden for non-Android projects, so no Android-specific dependencies should be in the original plugin's code. I understood that and followed along. I figured however that there was no other easy and maintainable way to query the Android Gradle Plugin's API except for referencing it, the alternative would have be to write a lot of class loading magic.

The other thing is in many projects Android modules and pure JVM modules are intermixed. My aim was to have a single plugin (the new android plugin) to handle both, without code and configuration duplication. That's why I also created those aggregation tasks, `detektMain` and `detektTest`, to make it easy to run detekt on a codebase that contained either module type.

realdadfish:
@arturbosch I opened up a separate, experimental sub PR to figure out how feasible it is to generate a baseline file with type resolution enabled, but I stumbled upon a weird issue. Could you comment there? https://github.com/realdadfish/detekt/pull/1

realdadfish:
Issue solved in a separate PR here. https://github.com/detekt/detekt/pull/2879

realdadfish:
So, the TR-enabled baseline creation is working nicely alongside checking. I will add a couple of more integration tests to check whether the appropriate tasks are actually created once I got some feedback on the implementation.

I kind of "hacked" the configuration of the baseline file and dynamically extend it's name with the sourceSet or variant name. I don't like this approach all too much; the alternative would have been to not configure a single baseline file, but to provide a directory where these files reside (configure `module/baselines`, plugin reads / writes `module/baselines/{main,test,debug,release}.xml`, while falling back to a common `module/baselines/detekt.xml` if the specific baseline file is not found).

Feedback is welcome!

arturbosch:
> Issue solved in a separate PR here. #2879

Thanks for the fix! 

It will take me some time to review this PR. I will come back to you in a few days.

realdadfish:
@arturbosch Have you had time to look at the PR yet? I'm working with a fork right now and would really like to get this "upstream-ready" :)

arturbosch:
> @arturbosch Have you had time to look at the PR yet? I'm working with a fork right now and would really like to get this "upstream-ready" :)

Sorry for the delay. Unfortunately no, not yet. I will take time for this PR this week. Thanks for the patience.

arturbosch:
> @arturbosch Have you had time to look at the PR yet? I'm working with a fork right now and would really like to get this "upstream-ready" :)

Hey, when trying to build locally I get:

```
com.android.builder.errors.EvalIssueException: SDK location not found. Define location with an ANDROID_SDK_ROOT environment variable or by setting the sdk.dir path in your project's local properties file at '/tmp/gradle17916022041907269827projectDir/local.properties'.
```

Any idea why this is happening? I can successfully build `ktlint-gradle` without needing the android sdk.

arturbosch:
> > The code looks good to me.
> > Could you elaborate why we technically need two Gradle plugins to support Android?
> > Shouldn't `project.plugins.withId("kotlin-android") {...}` to enough to not run into android specific code?
> 
> I'm not sure if I understand the question. The original reviewer's comment was that the Android-specific things should not become a burden for non-Android projects, so no Android-specific dependencies should be in the original plugin's code. I understood that and followed along. I figured however that there was no other easy and maintainable way to query the Android Gradle Plugin's API except for referencing it, the alternative would have be to write a lot of class loading magic.
> 

I meant that if you protect code which uses Android specific methods/references with `project.plugins.withId("kotlin-android") {...}` it won't crash for non-android users.
I'm sorry that I don't follow now :). Which original comment do you mean?
If we can combine the task generation for plain jvm and android users, we should definitely do this in just one plugin.
Gradle plugin tests are expensive and cost us most of the time in our build.
Having a separation through different files/functions is good enough I would say. The `ktlint-gradle` plugin proves that this is possible.

realdadfish:
> > @arturbosch Have you had time to look at the PR yet? I'm working with a fork right now and would really like to get this "upstream-ready" :)
> 
> Hey, sorry for the wait!
> When trying to build locally I get:
> 
> ```
> com.android.builder.errors.EvalIssueException: SDK location not found. Define location with an ANDROID_SDK_ROOT environment variable or by setting the sdk.dir path in your project's local properties file at '/tmp/gradle17916022041907269827projectDir/local.properties'.
> ```
> 
> Any idea why this is happening? I can successfully build `ktlint-gradle` without needing the android sdk.

Can you give me the full stacktrace? Then I can investigate why the Gradle plugin asks for the SDK at this stage. I don't know the ktlint plugin enough to answer your second question.

realdadfish:
> > I'm not sure if I understand the question. The original reviewer's comment was that the Android-specific things should not become a burden for non-Android projects, so no Android-specific dependencies should be in the original plugin's code. I understood that and followed along. I figured however that there was no other easy and maintainable way to query the Android Gradle Plugin's API except for referencing it, the alternative would have be to write a lot of class loading magic.
> 
> I meant that if you protect code which uses Android specific methods/references with `project.plugins.withId("kotlin-android") {...}` it won't crash for non-android users.

Ok, had to re-read on class-loading, because my original assumption was that if I want to be truly "compatible" I'd have to use a lot of reflection in my code, but apparently popular JVMs have some loopholes and it might be enough to move the Android-specific code in separate classes, eventually even only into a separate method (cf. https://stackoverflow.com/a/34263422/305532). I try to come up with a better solution then.

> I'm sorry that I don't follow now :). Which original comment do you mean?

It was given by schalkms on the Kotlin Slack at #detekt iirc.

> If we can combine the task generation for plain jvm and android users, we should definitely do this in just one plugin.
> Gradle plugin tests are expensive and cost us most of the time in our build.

Well, the amount of tests would stay the same if I'd merge both, but in any case one could speed up the tests in-module with parallelism (higher fork-count) eventually if this is an issue.

> Having a separation through different files/functions is good enough I would say. The `ktlint-gradle` plugin proves that this is possible.

Again, don't know the innards of ktlint-gradle, but I try my best.

realdadfish:
@arturbosch I rebased to current master and merged the implementation back into the main plugin. Let me know what you think.

realdadfish:
If this is ready for prime-time, I should probably document the configuration options and meta tasks / potential pitfalls as well as the configuration file fallback somewhere. What would be a good place to do so?

BraisGabin:
> If this is ready for prime-time, I should probably document the configuration options and meta tasks / potential pitfalls as well as the configuration file fallback somewhere. What would be a good place to do so?

I suppose that here is the best place: https://github.com/detekt/detekt/blob/master/docs/pages/gettingstarted/gradle.md if it gets big we can split it later.

realdadfish:
@arturbosch These warnings appear only for `detekt<Variant>`, yes, they won't appear on `detekt<SourceSet>` on JVM modules.

Things to note: `detektMain` and `detektTest`, as well as `detektBaselineMain` and `detektBaselineTest` tasks are proxy tasks that can be called to check and eventually create baseline files for all main and all test variants of a module. 

Baseline files are recognized by source set or variant name, while the latter takes precedence. That means if both, a `<baseline-name>-main.xml` and a `<baseline>.xml` exist in a JVM module, the first is used for `main` source sets and the latter for `test` source sets. Likewise, if a `<baseline-name>-<variant>.xml` exists, it takes precendence for the `detekt<Variant>` and `detektBaseline<Variant>` tasks before `<baseline-name>.xml` (if that exists).

realdadfish:
I've added some documentation to `gradle.md` here: https://github.com/detekt/detekt/blob/f125d511cc5ce4a537f9946e943e89e66bdd2060/docs/pages/gettingstarted/gradle.md

BraisGabin:
TL;DR: Ready to merge.

I did a quick fix to make it compile. But I'm not sure if I could break something (I know nearly nothing about gradle scriptting)

Things that I found:

* A lot of warnings like:
```
Ignoring a file detekt cannot handle: /Users/brais.gabin/projects/XXXXX/file.java
```
* If I run `./gradlew detekt` it doesn't check the flavor/buildType source sets nor the tests. (I didn't check androidTest but I suppose that it will be the same)
* I have 2 flavors so detekt finds nearly all the issues twice because nearly all the code is in `main`. This have sense... but if we find an way to remove duplicates would be great.

Neigther of those seem like blockers. I think that this is a huge improvement.

We can mark the support to Android as "experimental" so we can break things when we get some feedback from the community.

realdadfish:
> If I run ./gradlew detekt it doesn't check the flavor/buildType source sets nor the tests. (I didn't check androidTest but I suppose that it will be the same)

Yes, the `detekt` task wasn't and isn't used at all for Android - I could only add this as another "proxy" to run `detekt<AllSourceSetsOrVariants>` in the background, but frankly, I didn't want to confuse user's expectations what this task then would do (TR-enabled checking) than what the regular `detekt` task would do on a non-Android project (non-TR-enabled checking).

> I have 2 flavors so detekt finds nearly all the issues twice because nearly all the code is in main. This have sense... but if we find an way to remove duplicates would be great.

This is a hard problem, because each analysis sees basically "two different versions" of the same module. What I could think about in general however is some kind of "merge" task, similar to what Jacoco also provides for it's binary format. We'd however have to separate analysis from report generation task-wise, and I don't know how feasible this is to do right now, given the fact that the development goes towards a compiler plugin.

BraisGabin:
I have *a lot* of false positives in the rule `IgnoredReturnValue`... All of them related with RxJava3 (that have sense, because RxJava uses the `@CheckReturnValue` annotation). I don't know if this is an issue with the rule itself or the plugin. I feel strange that no one opened an issue about this if the problem is the rule.

Then I have false positives in `UnnecessarySafeCall` related with the missing classes. For example:

```kotlin
binding.address2?.visibility = if (state.address == null) GONE else VISIBLE
```

The problem is that `binding` is a generated class so detekt can't know if `address2` is or not nullable.

I really don't know if these issues are in the plugin, my setup or the rules.

realdadfish:
@BraisGabin I've recently upgraded to Rx3 (3.0.6) as well and got no additional warnings from Detekt wrt `IgnoredReturnValue`. Interesting...

Regarding databinding (this is databinding, and not synthetic accessors, right?), we don't use safe-calls on those, so we don't see those warnings there either.

BraisGabin:
> @BraisGabin I've recently upgraded to Rx3 (3.0.6) as well and got no additional warnings from Detekt wrt `IgnoredReturnValue`. Interesting...

Yes... probably a problem in my setup? I have issues in code like:

```kotlin
Observable.just("hi")
  .observeOn(main) // This line
  .subscribe()
  .addTo(compositeDisposable)
```

or

```kotlin
return if (something) {
  Observable.just("hi") // this line
} else {
  Observable.just("hello") // and this line too
}
```

> Regarding databinding (this is databinding, and not synthetic accessors, right?), we don't use safe-calls on those, so we don't see those warnings there either.

It is `ViewBinding`. With a layout for mobile/tablet with some missing views in the mobile.

But I'm wondering how detekt should behave about this errors... Manage the case that one class is not correctly solved in the rule doesn't seem rule-writter friendly... So maybe we should just don't care about it. This will be fixed with the compile plugin so add that complexity to the rules doesn't seem a good idea.

arturbosch (approved):

arturbosch:
Cool :) lets merge this then, looks good @realdadfish thanks for the contribution!

BraisGabin:
🎉🎉🎉🎉🎉 Great contribution @realdadfish

PR #2788, opened 2020-06-11
Smarter Caching on Github Actions

cortinico:
Currently our Github Actions setup relies on a static key for caching.

This means that we restore an old Gradle cache that doesn't contain the version of Gradle we need (it gets re-downloaded every time).

Given that we don't have a lockfile (should we consider it?), I'm hashing 'buildSrc/src/main/kotlin/Versions.kt' and adding it to the cache key. So the cache will be partially invalidated whenever we release a new version (or whenever we bump a library).


cortinico:
CI is red because of #2795

PR #2792, opened 2020-06-12
Introduce BoM to manage our own dependencies

BraisGabin:
Right now this [Bill of Materials (BoM)](https://docs.gradle.org/current/userguide/platforms.html) module just manage our dependencies. We could expose it so our users can align it's dependencies easier too. If we want to do that we should iterate this file a bit more.

After this change the file `buildSrc/Versions.kt` seems kind of pointless. I think that we should be able to move all those consts to better places.

### DIFF
--- a/buildSrc/src/main/kotlin/commons.gradle.kts
+++ b/buildSrc/src/main/kotlin/commons.gradle.kts
@@ -16,7 +16,7 @@ allprojects {
     }
 }
 
-subprojects {
+configure(subprojects.filter { it.name != "detekt-bom" }) {
 
     val project = this
 
--- a/buildSrc/src/main/kotlin/commons.gradle.kts
+++ b/buildSrc/src/main/kotlin/commons.gradle.kts
@@ -70,15 +70,16 @@ configure(subprojects.filter { it.name != "detekt-bom" }) {
     }
 
     dependencies {
+        implementation(platform(project(":detekt-bom")))
         compileOnly(kotlin("stdlib-jdk8"))
 
-        testImplementation("org.assertj:assertj-core:${Versions.ASSERTJ}")
-        testImplementation("org.spekframework.spek2:spek-dsl-jvm:${Versions.SPEK}")
-        testImplementation("org.reflections:reflections:${Versions.REFLECTIONS}")
-        testImplementation("io.mockk:mockk:${Versions.MOCKK}")
+        testImplementation("org.assertj:assertj-core")
+        testImplementation("org.spekframework.spek2:spek-dsl-jvm")
+        testImplementation("org.reflections:reflections")
+        testImplementation("io.mockk:mockk")
 
-        testRuntimeOnly("org.junit.platform:junit-platform-launcher:${Versions.JUNIT}")
-        testRuntimeOnly("org.spekframework.spek2:spek-runner-junit5:${Versions.SPEK}")
+        testRuntimeOnly("org.junit.platform:junit-platform-launcher")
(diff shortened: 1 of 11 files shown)


cortinico:
> After this change the file `buildSrc/Versions.kt` seems kind of pointless. I think that we should be able to move all those consts to better places.

Nice addition.
On this side, agree that we can get rid of that file. Therefore, we should update #2788 to use `detekt-bom/build.gradle.kts` as cache key.

BraisGabin:
> Therefore, we should update #2788 to use `detekt-bom/build.gradle.kts` as cache key.

It was that pr that made me think about this.

buildSrc/src/main/kotlin/commons.gradle.kts
```diff
@@ -16,7 +16,7 @@ allprojects {
     }
 }
 
-subprojects {
+configure(subprojects.filter { it.name != "detekt-bom" }) {
```
arturbosch, line 19: Should we introduce a helper in `buildSrc` for this query?

BraisGabin: should we? The code is repeated only twice and it's really clear what it does. I'm not sure if it will give us value

realdadfish:
I think this change broke at least maven-local publishing. The POM of the Gradle plugin lists `detekt-bom` in it's dependencies, but since `detekt-bom` is not published, it fails to resolve this dependency:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <!-- This module was also published with a richer model, Gradle metadata,  -->
  <!-- which should be used instead. Do not delete the following line which  -->
  <!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
  <!-- that they should prefer consuming it instead. -->
  <!-- do_not_remove: published-with-gradle-metadata -->
  <modelVersion>4.0.0</modelVersion>
  <groupId>io.gitlab.arturbosch.detekt</groupId>
  <artifactId>detekt-gradle-plugin</artifactId>
  <version>1.10.0-RC1-android-gradle-plugin</version>
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>io.gitlab.arturbosch.detekt</groupId>
        <artifactId>detekt-bom</artifactId>
        <version>1.10.0-RC1-android-gradle-plugin</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-gradle-plugin</artifactId>
      <version>1.3.72</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-gradle-plugin-api</artifactId>
      <version>1.3.72</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>com.android.tools.build</groupId>
      <artifactId>gradle</artifactId>
      <version>4.0.0</version>
      <scope>runtime</scope>
    </dependency>
  </dependencies>
</project>
```

realdadfish:
Hrm... ok, the Bintray publishing does not contain a reference to detekt-bom. Very weird: https://search.maven.org/artifact/io.gitlab.arturbosch.detekt/detekt-gradle-plugin/1.10.0-RC1/jar

PR #2798, opened 2020-06-14
Simplify the Code Coverage workflow

cortinico:
We don't need a build matrix for the codecoverage workflow, so I'm simplifying the workflow a bit.

### DIFF
--- a/.github/workflows/codecoverage.yaml
+++ b/.github/workflows/codecoverage.yaml
@@ -9,13 +9,9 @@ on:
       - '*'
 
 jobs:
-  gradle:
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        jdk: [11]
-    runs-on: ${{ matrix.os }}
+  publish-code-coverage:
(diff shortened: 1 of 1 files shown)


.github/workflows/codecoverage.yaml
```diff
@@ -26,7 +22,7 @@ jobs:
       - name: Setup Java
         uses: actions/setup-java@v1
         with:
-          java-version: ${{ matrix.jdk }}
+          java-version: 11
```
schalkms, line 25: I’m not sure. Codecov was kind of flaky. Should we use JDK8, since we had the most consistent results with it?

cortinico: I don't have much context on the codecov flakyness, so I can't say.

arturbosch: We used jdk 11 for quite some time for coverage, didn't we?

cortinico: Yeah, Let's keep an eye on it and if we notice more flakyness we can still update the Java version.

PR #2808, opened 2020-06-16
use parallel build options

imanushin:
Add parallel option for gradle builds.

Probably it can improve #2680 after more and more modules introduction.

Timing examples:
* [Without parallel - ubuntu 8 - 4m 23s](https://github.com/detekt/detekt/pull/2779/checks?check_run_id=778117131).
* [With parallel - ubuntu 8 - 1m 21s](https://github.com/detekt/detekt/pull/2808/checks?check_run_id=778157588).
* [Without parallel - macos 8 - 6m 14s](https://github.com/detekt/detekt/pull/2779/checks?check_run_id=778117167).
* [With parallel - macos 8 - 3m 38s](https://github.com/detekt/detekt/pull/2808/checks?check_run_id=778157753).

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -53,7 +53,7 @@ jobs:
 
 
     - name: Build detekt (UNIX)
-      run: ./gradlew build shadowJar -PwarningsAsErrors=true
+      run: ./gradlew build shadowJar -PwarningsAsErrors=true --parallel
       if: matrix.os == 'macos-latest' || matrix.os == 'ubuntu-latest'
     - name: Run detekt-cli --help (UNIX)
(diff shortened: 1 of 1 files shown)


arturbosch:
Interesting so `gradle.properties` does not get picked up from CI/gradle daemon?

imanushin:
@arturbosch , I don't know.

However yesterday I found [better option](https://github.com/kotest/kotest/blob/master/.github/workflows/build.yml#L97), I can make a PR with this change.

arturbosch:
> @arturbosch , I don't know.
> 
> However yesterday I found [better option](https://github.com/kotest/kotest/blob/master/.github/workflows/build.yml#L97), I can make a PR with this change.

Yes go on please: we already have `GRADLE_OPTS: -Dorg.gradle.daemon=false` in use

PR #2812, opened 2020-06-20
UnusedPrivateClass: don't report imported classes

### LINKED ISSUE #2809: False positive UnusedPrivateClass when importing private enum constants directly.
## Expected Behavior
No detection because private class is used via import.

## Observed Behavior
```
> Task :detekt FAILED
style - 5min debt
   UnusedPrivateClass - [E] at /.../com/example/Foo.kt:9:5
Overall debt: 5min
```

## Steps to Reproduce
```kotlin
package com.example
import com.example.Foo.E.C1

class Foo {

    fun f() {
        setOf(C1)
    }
    private enum class E {
        C1,
        C2,
        C3
    }
}
```

## Context
We have a private enum like this and a mapping from a list of enums to strings, i.e. 
## Your Environment

* Version of detekt used: 1.10.0-RC1
* Operating System and version: not relevant

t-kameyama:
Fixes #2809

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateClass.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateClass.kt
@@ -18,6 +18,7 @@ import org.jetbrains.kotlin.psi.KtDoubleColonExpression
 import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.KtFunctionType
+import org.jetbrains.kotlin.psi.KtImportDirective
 import org.jetbrains.kotlin.psi.KtNameReferenceExpression
 import org.jetbrains.kotlin.psi.KtNamedDeclaration
 import org.jetbrains.kotlin.psi.KtNamedFunction
@@ -63,9 +64,16 @@ class UnusedPrivateClass(config: Config = Config.empty) : Rule(config) {
 
         private val privateClasses = mutableSetOf<KtNamedDeclaration>()
         private val namedClasses = mutableSetOf<String>()
+        private val importDirectives = mutableSetOf<String>()
 
         fun getUnusedClasses(): List<KtNamedDeclaration> {
-            return privateClasses.filter { it.nameAsSafeName.identifier !in namedClasses }
+            return privateClasses.filter { !it.isUsed() }
+        }
+
+        private fun KtNamedDeclaration.isUsed(): Boolean {
+            if (nameAsSafeName.identifier in namedClasses) return true
+            val fqName = fqName?.asString()
+            return fqName != null && importDirectives.any { it.startsWith(fqName) }
         }
 
         override fun visitClass(klass: KtClass) {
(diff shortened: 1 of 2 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateClassSpec.kt
```diff
@@ -361,5 +362,27 @@ class UnusedPrivateClassSpec : Spek({
...
+                    }
+
+                    private enum class E {
+                        E1,
+                        E2,
+                        E3
+                    }
+                }
+            """
+            val findings = UnusedPrivateClass().lint(code)
```
schalkms, line 384: Is it possible to use `compileAndLint()`?

t-kameyama: If we use `compileAndLint()`, `does not report imported enum class - #2809` test fails.
https://github.com/detekt/detekt/actions/runs/142180073

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateClass.kt
```diff
@@ -63,9 +64,16 @@ class UnusedPrivateClass(config: Config = Config.empty) : Rule(config) {
...
 
         fun getUnusedClasses(): List<KtNamedDeclaration> {
-            return privateClasses.filter { it.nameAsSafeName.identifier !in namedClasses }
+            return privateClasses.filter { !it.isUsed() }
+        }
+
+        private fun KtNamedDeclaration.isUsed(): Boolean {
+            if (nameAsSafeName.identifier in namedClasses) return true
+            val fqName = fqName?.asString()
+            return fqName != null && importDirectives.any { it.startsWith(fqName) }
```
TWiStErRob, line 76: @t-kameyama @arturbosch @schalkms isn't this a counter-example for `startsWith`?
```kotlin
import foo.bar.Baz
import foo.bar.Bazinga
fun f() {
    Baz()
}
```
IRL `Bazinga` could be generated `Baz_Factory` for example.

TWiStErRob: oh, it's only about private classes..., but there must be a similar example for that too.
e.g. a private class was renamed, but the original import not removed, but detekt wouldn't find it?

arturbosch: Hm, the import must start with the fully qualified name of the private class not the other way. So this should be okay?
I'm not sure if I understand your example.

```kotlin
package foo.bar
                import foo.bar.Baz
                import foo.bar.Bazinga
                private class BazFactory
```

t-kameyama: False negative:

```kotiln
package com.example

import com.example.C.EFG.EFG1

class C {
    fun test() {
        println(EFG1)
    }

    private enum class E { // false negative
        E1
    }

    private enum class EFG {
        EFG1
    }
}
```

TWiStErRob: Yep, that's what I meant.

Also be careful with nested classes inside nested classes, they might be broken now or after a fix too.

PR #2814, opened 2020-06-20
Publish detekt-bom module

arturbosch:


BraisGabin:
🤔 If we publish the bom... should we keep ktlint in or bom? I use detekt and ktlint as different tools.

arturbosch:
>  If we publish the bom... should we keep ktlint in or bom? I use detekt and ktlint as different tools.

The question is when to use the detekt pom? Who will use it?

Users embedding detekt or write custom rules? -> KtLint should be included

Normal users of the cli or gradle plugin won't need it, right?

BraisGabin:
> Users embedding detekt or write custom rules? -> KtLint should be included

Should it? Should I align my custom rules with the format plugin?

Should we create a different bom for the format plugin? This bom should contain the detekt bom plus the format plugin dependencies (ktlint).

arturbosch:
> > Users embedding detekt or write custom rules? -> KtLint should be included
> 
> Should it? Should I align my custom rules with the format plugin?
> 
> Should we create a different bom for the format plugin? This bom should contain the detekt bom plus the format plugin dependencies (ktlint).

Your right. `detekt-formatting` is an extension and its dependencies shouldn't be included in the bom.
But as a rule authors do I really need to include the bom? Or is `api` enough?

From the point of view of the sonarqube and intellij plugin collaborator I want to have one bom with all dependencies :).
But there shouldn't be any problems with including KtLint as a transitive dependency in this cases.

calvarez-ov:
We have a nexus repository which mirrors other repositories.

In our mirror, we somehow managed to retrieve this file: 
`https://<our nexus repo>/repository/maven-public/io/gitlab/arturbosch/detekt/detekt-bom/1.10.0/detekt-bom-1.10.0.module`

With the following content:
```
You are using a deprecated resolve URL for Pentaho projects. The correct resolve URL is:

https://nexus.pentaho.org/content/groups/omni

Please refer to the official Pentaho Documentation on how to build Pentaho projects with Maven:

https://github.com/pentaho/maven-parent-poms/wiki

This is a sample settings.xml file that could be used in your local build environment:

https://github.com/pentaho/maven-parent-poms/blob/master/maven-support-files/settings.xml
```

Could this possibly be a problem of the packaging of detekt? Or a problem with our nexus repository configuration?

cortinico:
> Could this possibly be a problem of the packaging of detekt? Or a problem with our nexus repository configuration?

A mixture of the two. You nexus is considering an error page a valid artifact and storing it (maybe this should be disabled in the configuration).
On the other hand, we're failing and publish all the needed files (specifically `detekt-bom-1.10.0.module` is missing). I've opened an issue to fix this #2878. Let's follow over there.

PR #2818, opened 2020-06-22
Conform the pattern of FunctionNaming to coding conventions

chao2zhang:
Instead of the traditional camelCases, the default pattern of a function name should be PascalCase with the following rationales:

1. The page "Kotlin coding conventions" describes that factory methods can have the same name as the abstract return type.
This was actually merged and published in Oct 2017, so it has been established for over two years.
https://github.com/JetBrains/kotlin-web-site/commit/465783bf395acc5d34f089291eca594ef3d7d4ce
https://kotlinlang.org/docs/reference/coding-conventions.html#function-names

2. Jetpack Compose also uses the PascalCase in its tutorial and code examples.
https://developer.android.com/jetpack/compose/tutorial
https://codelabs.developers.google.com/codelabs/jetpack-compose-basics

3. ObjectPropertyNaming.propertyPattern also allows PascalCase, conforming to the coding convention as well that names of properties holding references to singleton objects can use the same naming style as object declarations.
https://kotlinlang.org/docs/reference/coding-conventions.html#property-names

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,7 @@ If you contributed to detekt but your name is not in the list, please feel free 
 - [David Phillips](https://github.com/daphil19) - New rule: MandatoryBracesLoops
 - [Volkan Şahin](https://github.com/volsahin) - Documentation improvement
 - [Remco Mokveld](https://github.com/remcomokveld) - Rename Blacklist/Whitelist to more meaningful names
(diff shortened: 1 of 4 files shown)


chao2zhang:
I have other consideration to add a new `factoryMethodPattern`

arturbosch:
> I have other consideration to add a new `factoryMethodPattern`

Yes, I think we need to include some checks as if the function is top level or declared on an object.
Further we can restrict this convention by adding a property like `nameMustMatchReturnType`.

PR #2819, opened 2020-06-23
Rename IsPropertyNaming to NonBooleanPropertyPrefixedWithIs?

arturbosch:
I'm not sure if this is a better name but I wanted to start a discussion if we should rename the rule.

Imo IsPropertyNaming could lead to confusion.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -391,6 +391,9 @@ naming:
   MemberNameEqualsClassName:
     active: true
     ignoreOverridden: true
+  NonBooleanPropertyPrefixedWithIs:
+    active: false
+    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
   ObjectPropertyNaming:
     active: true
     excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
--- a/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NamingRules.kt
+++ b/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NamingRules.kt
@@ -22,7 +22,7 @@ class NamingRules(config: Config = Config.empty) : MultiRule() {
     private val variableMaxNameLengthRule = VariableMaxLength(config)
     private val topLevelPropertyRule = TopLevelPropertyNaming(config)
     private val objectConstantNamingRule = ObjectPropertyNaming(config)
-//    private val isPropertyNamingRule = IsPropertyNaming(config)
+    private val nonBooleanPropertyPrefixedWithIsRule = NonBooleanPropertyPrefixedWithIs(config)
     private val packageNamingRule = PackageNaming(config)
     private val classOrObjectNamingRule = ClassNaming(config)
     private val enumEntryNamingRule = EnumNaming(config)
@@ -39,7 +39,7 @@ class NamingRules(config: Config = Config.empty) : MultiRule() {
             variableMaxNameLengthRule,
             topLevelPropertyRule,
             objectConstantNamingRule,
-//            isPropertyNamingRule,
+            nonBooleanPropertyPrefixedWithIsRule,
             packageNamingRule,
             classOrObjectNamingRule,
             enumEntryNamingRule,
@@ -86,7 +86,7 @@ class NamingRules(config: Config = Config.empty) : MultiRule() {
     private fun handleProperty(declaration: KtProperty) {
         variableMaxNameLengthRule.runIfActive { visitProperty(declaration) }
         variableMinNameLengthRule.runIfActive { visitProperty(declaration) }
-//        isPropertyNamingRule.runIfActive { visitProperty(declaration) }
+        nonBooleanPropertyPrefixedWithIsRule.runIfActive { visitProperty(declaration) }
 
         when {
             declaration.isTopLevel -> topLevelPropertyRule.runIfActive { visitProperty(declaration) }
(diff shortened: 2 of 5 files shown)


cortinico:
Imho `BooleanPropertyNaming` is a better name. I agree that `Is` in the first part of the name is not ideal as it's also a verb and it doesn't help to immediately understand the meaning of the rule.

If possible I'd try to keep the `*Naming` suffix as we already have several rules that follows this convention:

```
./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ClassNaming.kt
./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/IsPropertyNaming.kt
./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNaming.kt
./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ObjectPropertyNaming.kt
./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/TopLevelPropertyNaming.kt
./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionNaming.kt
./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/PackageNaming.kt
./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ConstructorParameterNaming.kt
./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionParameterNaming.kt
./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/VariableNaming.kt
```

arturbosch:
> Imho `BooleanPropertyNaming` is a better name. I agree that `Is` in the first part of the name is not ideal as it's also a verb and it doesn't help to immediately understand the meaning of the rule.
> 
> If possible I'd try to keep the `*Naming` suffix as we already have several rules that follows this convention:
> 
> ```
> ./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ClassNaming.kt
> ./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/IsPropertyNaming.kt
> ./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/EnumNaming.kt
> ./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ObjectPropertyNaming.kt
> ./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/TopLevelPropertyNaming.kt
> ./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionNaming.kt
> ./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/PackageNaming.kt
> ./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ConstructorParameterNaming.kt
> ./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionParameterNaming.kt
> ./src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/VariableNaming.kt
> ```

But this rule will report the inverse of BooleanPropertyNameing -> all properties which are not of type boolean and start with an `is` are forbidden.

BraisGabin:
I like your proposal more that the current one. But the problem with the name make me think... Should this be part of the `PropertyNaming` as a flag? I thought about this after reading your description in the last comment.

cortinico:
> Should this be part of the `PropertyNaming` as a flag? I thought about this after reading your description in the last comment.

Piggybacking on this. Imho we should have a `BooleanPropertyNaming` that enforces `is-` named boolean property (not active by default as it's really strict), but that has a flag also to verify non boolean properties that are wrongly named (basically what `IsPropertyNaming` is doing now).

BraisGabin:
A lot of people look at detekt as a guideline. If we add something like that they will think that name all your boolean properties with `is` is a good thing, and I don't think so. I mean, I can understand that a team take that convention but as a normal basis I don't think so.

cortinico:
> A lot of people look at detekt as a guideline

yup, agree on this, it's probably too strict and will lead to a lot of local `@Suppress` if activated.

schalkms:
I agree with @BraisGabin here. Detekt should have a controversial ruleset, which is disabled by default. This rule would be a perfect candidate for this new ruleset.

BraisGabin:
🤔 I think that we mixed some ideas here. I think that this rule is good. Naming a property `isPrivate` and define it as `Int` is bad. Because if I read `is` I will think that this is a boolean. It's the same when a class is called `val animal = dog()` you directly think that `dog` is a function but it could be actually a class.

I mean, the rule, as it is right now is not controversial.

If we force that ALL the boolean variables are prefixed with `is` that's controversial for sure. And I never saw any article encoraging that notation.

My point is: should we merge this rule with `VariableNaming`? I know, right now `VariableNaming` is just a regex. But I think that the best fit for `IsPropertyNaming` rule is to merge it in `VariableNaming` instead of raneming it.

cortinico:
> But I think that the best fit for `IsPropertyNaming` rule is to merge it in `VariableNaming` instead of raneming it.

Yup that's probably the best place to go. 
The side effect of this is that `VariableNaming` is not relying on type and symbols resolution, while the current `IsPropertyNaming` is.

So we will end up with a rule that has a flag that enables a more "advanced" naming analysis only if a BindingContext is provided. I don't know if we have already similar rules that offer "mixed" analysis and how we want to proceed on this front.

arturbosch:
Ok, I will leave this PR open for now, remove `IsPropertyNaming` for 1.10.0 from the ruleset, release 1.10 and refactor this for 1.11 or 1.10.1.

cortinico:
> Maybe we should leave this as a separate rule?

How do we feel about the `controversial` ruleset suggested by @schalkms:

> Detekt should have a controversial ruleset, which is disabled by default. This rule would be a perfect candidate for this new ruleset.

I also feel this rule could probably be moved there

arturbosch:
@cortinico  @BraisGabin  @schalkms 
Hm, having a new rule `BooleanPropertyNaming` as @cortinico mentioned should be `controversal`.
But should this one be `controversal`? I just reread the discussion and feel that this rule with the new name `NonBooleanPropertyPrefixedWithIs` could just be under the `style` rule set?
Maybe we should postpone introducing a new rule set until 2.0 ?

cortinico:
> But should this one be `controversal`? I just reread the discussion and feel that this rule with the new name `NonBooleanPropertyPrefixedWithIs` could just be under the `style` rule set?
> Maybe we should postpone introducing a new rule set until 2.0 ?

Agree here. For what concerns this specific rule, I'd +1 the new name `NonBooleanPropertyPrefixedWithIs`. The convo around `BooleanPropertyNaming` can be resumed once we land 2.x

> Edit: well naming is style, so lets leave it there?

`naming` and `style` are actually two different ruleset right? This should definitely live in the `naming` one.

PR #2823, opened 2020-06-24
Simplify pre-merge.yaml

BraisGabin:


.github/workflows/pre-merge.yaml
```diff
@@ -54,24 +54,10 @@ jobs:
 
     - name: Build detekt (UNIX)
```
cortinico, line 55:
```suggestion
    - name: Build detekt
```

small nit

BraisGabin:
I think that Windows doesn't have problems with UNIX paths. It's default is `\` but it understand `/` too.

PR #2828, opened 2020-06-27
Fix UnnecessaryLet false negatives

### LINKED ISSUE #2826: UnnecessaryLet false negatives
Right now this `a.let { print(it) }` and this `a.let { msg -> print(msg) }` are allowed but they should be flagged as well because you can replace both with `println(a)`.

IntelliJ mark them as unnecessary too.

BraisGabin:
Closes #2826

There were some false negatives in this rule:

- `a.let { 1.plus(1) } // 1.plus(1)`
- `a?.let { 1.plus(1) } // if (a != null) 1.plus(1)` This one is to avoid overuse of let when you get nothing from it.
- `a.let { println(it) } // println(it)`
- `a.let { println(it); println("hola") }`
- etc.

I know, the tests are a real mess. I'll try to refactor them.

I know that we have a script/project to run detekt over different projects to ensure that the new rule works... Where is it? I'll like to prove this on real projects.

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLet.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLet.kt
@@ -60,17 +60,17 @@ class UnnecessaryLet(config: Config) : Rule(config) {
                 report(CodeSmell(issue, Entity.from(expression), "let expression can be omitted"))
             }
         } else {
-            val count = lambdaExpr.countReferences() ?: 0
-            if (count == 0 && !expression.receiverIsUsed(bindingContext) && isNullSafeOperator) {
+            val lambdaReferenceCount = lambdaExpr.countReferences() ?: 0
+            if (lambdaReferenceCount == 0 && !expression.receiverIsUsed(bindingContext) && isNullSafeOperator) {
                 report(
                     CodeSmell(
                         issue, Entity.from(expression),
                         "let expression can be replaces with a simple if"
                     )
                 )
-            } else if (count <= 1 && !isNullSafeOperator) {
+            } else if (lambdaReferenceCount <= 1 && !isNullSafeOperator) {
                 report(CodeSmell(issue, Entity.from(expression), "let expression can be omitted"))
-            } else if (count == 1 && canBeReplacedWithCall(lambdaExpr)) {
+            } else if (lambdaReferenceCount == 1 && canBeReplacedWithCall(lambdaExpr)) {
                 report(CodeSmell(issue, Entity.from(expression), "let expression can be omitted"))
             }
         }
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLet.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLet.kt
@@ -9,6 +9,7 @@ import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.rules.IT_LITERAL
 import io.gitlab.arturbosch.detekt.rules.LET_LITERAL
+import io.gitlab.arturbosch.detekt.rules.receiverIsUsed
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.psi.KtBlockExpression
 import org.jetbrains.kotlin.psi.KtCallExpression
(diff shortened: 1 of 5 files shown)


BraisGabin:
😭 I never understand codecov. I did improved the test coverage here. Is it complainly for moving a function from one file to another?

arturbosch:
Great job!

> I know that we have a script/project to run detekt over different projects to ensure that the new rule works... Where is it? I'll like to prove this on real projects.

It's called `detekt/scripts/get_analysis_projects.groovy`.

In case you are not familiar with Groovy: 
get it via [sdkman](https://sdkman.io/), run `sdk i groovy` and then `groovy scripts/get_analysis_projects.groovy ~/git/<test_repos>`

BraisGabin:
All this examples are flagged by the new rule. I need your opinion if they are valid, false positives or they are edge cases where a `@Suppress` will have more sense:

1. `ktlint/ktlint-test/src/main/kotlin/com/pinterest/ktlint/test/DumpAST.kt:40:22`
```kotlin
lineNumberColumnLength =
    (node.lastChildLeafOrSelf().lineNumber() ?: 1)
        .let { var v = it; var c = 0; while (v > 0) { c++; v /= 10 }; c }
```

2. `ktlint/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/MultiLineIfElseRule.kt:46:14`
```kotlin
val rightBraceIndent = node.treeParent
    .prevLeaf { it is PsiWhiteSpace && it.textContains('\n') }?.text.orEmpty()
    .let { "\n${it.substringAfterLast("\n")}" }
```

3. `kotlinx/ui/kotlinx-coroutines-android/animation-app/app/src/main/java/org/jetbrains/kotlinx/animation/Animation.kt:109:23`
```kotlin
while (true) {
    val dt = time.let { old -> awaitFrame().also { time = it } - old }
    if (dt > 0.5e9) continue // don't animate through over a half second lapses
```

4. `javalin/javalin/src/test/java/io/javalin/TestRequest.kt:257:84`
```kotlin
Unirest.get("${http.origin}/html.html").basicAuth("u", "p").asString().let { assertThat(it.body).contains("HTML works") }
```

5. `detekt/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Suppressions.kt:46:16`
```kotlin
ruleSetId?.let { acceptedSuppressionIds.addAll(listOf(ruleSetId, "$ruleSetId.$id", "$ruleSetId:$id")) }
```

6. `kotlintest/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/reflection/classMatchers.kt:52:23`
```kotlin
fun KClass<*>.shouldHaveFunction(name: String, block: (KFunction<*>) -> Unit) {
  this should haveFunction(name)
  findFunction(name)?.let(block)
}
```

I think that all of them can be ommited and will improve the code. Some of them just removing the let and others moving code to a function. But in any case I think that let is a good fit there.

The only case where I'm not that sure is the 6th. Should `let` be used in this way? Should a scope function used like this? I think that you should use it with a "real lambda" and no with a "parameter".

schalkms:
Thanks for testing this new rule thoroughly.
I agree with 1-5. What should be the preferred way of coding for 6?

```kotlin
fun KClass<*>.shouldHaveFunction(name: String, block: (KFunction<*>) -> Unit) {
  this should haveFunction(name)
  findFunction(name)?.let(block)
}
```

BraisGabin:
> What should be the preferred way of coding for 6?
> 
> ```kotlin
> fun KClass<*>.shouldHaveFunction(name: String, block: (KFunction<*>) -> Unit) {
>   this should haveFunction(name)
>   findFunction(name)?.let(block)
> }
> ```

That code is the same as this one:
```kotlin
fun KClass<*>.shouldHaveFunction(name: String, block: (KFunction<*>) -> Unit) {
  this should haveFunction(name)
  findFunction(name)?.let { block(it) }
}
```

And this is legit. So I imagine that we should allow every use of `let` using the operator `?.` if it provides the lambda as parameter. But flag the ones that call `let` just with `.`.

I'll add test for this two cases and fix them 👍.

BraisGabin:
Ok, false positive fixed. Let me know what do you think now :)

BraisGabin:
Ok, false positive fixed. Let me know what do you think now :). It's ready to merge for me.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLet.kt
```diff
@@ -53,20 +53,26 @@ class UnnecessaryLet(config: Config) : Rule(config) {
...
-            report(
-                CodeSmell(
-                    issue, Entity.from(expression),
-                    "let expression can be replaces with a simple if"
+        if (lambdaExpr == null) {
+            if (!isNullSafeOperator) {
+                report(CodeSmell(issue, Entity.from(expression), "let expression can be omitted"))
+            }
+        } else {
+            val count = lambdaExpr.countReferences() ?: 0
```
schalkms, line 63: Can we rename this to lambdaReferenceCount to make it more readable?

BraisGabin: done

PR #2835, opened 2020-06-29
Fix false negative in MandatoryBracesIfStatements

BraisGabin:
This code should be flagged by `MandatoryBracesIfStatements`

```kotlin
if (true) 50
    .toString()
```

### DIFF
--- a/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatementsSpec.kt
+++ b/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatementsSpec.kt
@@ -23,6 +23,29 @@ class MandatoryBracesIfStatementsSpec : Spek({
             assertThat(findings).hasTextLocations(32 to 41)
         }
 
+        it("reports a simple if with a simgle statment in mutlple lines") {
+            val findings = subject.compileAndLint("""
+                fun f() {
+                	if (true) 50
+                        .toString()
+                }
+            """)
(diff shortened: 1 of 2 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatementsSpec.kt
```diff
@@ -23,6 +23,29 @@ class MandatoryBracesIfStatementsSpec : Spek({
             assertThat(findings).hasTextLocations(32 to 41)
         }
 
+        it("reports a simple if with a simgle statment in mutlple lines") {
```
cortinico, line 26:
```suggestion
        it("reports a simple if with a single statement in multiple lines") {
```

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatementsSpec.kt
```diff
@@ -23,6 +23,29 @@ class MandatoryBracesIfStatementsSpec : Spek({
...
+                fun f() {
+                	if (true) 50
+                        .toString()
+                }
+            """)
+
+            assertThat(findings).hasSize(1)
+        }
+
+        it("reports if-else with a single statment in mutlple lines") {
```
cortinico, line 37:
```suggestion
        it("reports if-else with a single statement in multiple lines") {
```

PR #2839, opened 2020-07-03
Add empty line rule

Vacxe:
Time to time people creating two or more empty line one by one whats actually doesn't make sense. 3-4 time per week I have to put this comment to the PRs

I really hope that it will be useful for many projects

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -146,6 +146,8 @@ empty-blocks:
     active: true
   EmptyKtFile:
     active: true
+  EmptyLine:
+    active: true
   EmptySecondaryConstructor:
     active: true
   EmptyTryBlock:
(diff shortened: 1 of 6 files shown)


BraisGabin:
Thanks for the contribution. This rule is already implemented by ktlint so I think that we should not duplicate it. I'm not sure if we have it ported in our formatting plugin. If we don't we should port it. Other option is to use ktlint and detekt without the plugin so you get all the ktlint rules directly (this is what I do).

cortinico:
Thanks for your contrib @Vacxe 

As @BraisGabin we already have a `formatting` rule that wraps a KtLint one: `NoConsecutiveBlankLines` 
https://github.com/detekt/detekt/blob/master/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/NoConsecutiveBlankLines.kt

Have you considered using it?

Vacxe:
@cortinico @BraisGabin thanks for reply. That's definitely my miss.  Thanks for the explanation.

PR #2844, opened 2020-07-04
Update getting started Gradle snippets

cortinico:
This is a follow-up from a discussion on Slack: https://kotlinlang.slack.com/archives/C88E12QH4/p1593875942190700

Seems like our getting started documentation is not properly working:
- The `plugins {}` block was placed after a `repository {}` which is not allowed by Gradle
- The Kotlin DSL syntax was invalid: `includeGroup "org.jetbrains.kotlinx"` without parentesis.
- For some reason the Android setup was applying the Detekt plugin twice (with the `apply:` and the `plugins{}` syntax).

I've fixed all of them, plus I'm providing snippets for setting up using Plugin DSL and Legacy plugin application both for Groovy and Kotlin flavors.

Issue #2146, opened 2019-11-29
Gradle plugin fails with Could not find org.jetbrains.kotlinx:kotlinx-html-jvm:0.6.12


pkubowicz:
## Expected Behavior
I can run Detekt on a multi-project build

## Observed Behavior
Build fails with:
```
> Task :core:detekt FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':core:detekt'.
> Could not resolve all files for configuration ':core:detekt'.
   > Could not find org.jetbrains.kotlinx:kotlinx-html-jvm:0.6.12.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/org/jetbrains/kotlinx/kotlinx-html-jvm/0.6.12/kotlinx-html-jvm-0.6.12.pom
     If the artifact you are trying to retrieve can be found in the repository but without metadata in 'Maven POM' format, you need to adjust the 'metadataSources { ... }' of the repository declaration.
     Required by:
         project :core > io.gitlab.arturbosch.detekt:detekt-cli:1.2.0
```

## Steps to Reproduce
Example build.gradle:
```groovy
buildscript {
    repositories {
        jcenter()
    }
}

plugins {
    id 'org.jetbrains.kotlin.jvm' version '1.3.61' apply false
    id "io.gitlab.arturbosch.detekt" version "1.2.0" apply false
}

subprojects {
    apply plugin: 'kotlin'
    apply plugin: "io.gitlab.arturbosch.detekt"

    repositories {
        mavenCentral()
    }
}
```

## Context
I can run Detekt if I downgrade it to 1.1.0.

## Your Environment

* Version of detekt used: 1.2.0
* Version of Gradle used (if applicable): 6.0.1
* Operating System and version: Kubuntu 19.10
* Link to your project (if it's a public repository):

arturbosch:
Try adding `jcenter` to the repositories and it should work.

@schalkms @BraisGabin kotlinx.html does not seem to be hosted at maven central -> https://mvnrepository.com/artifact/org.jetbrains.kotlinx/kotlinx-html-jvm?repo=jcenter 
this will lead to followup support issues which request to be compatible with maven central :/

pkubowicz:
Yes, jcenter helps.

schalkms:
I'm not sure how much effort we need for migrating back to our own **HTML DSL**.

BraisGabin:
[...]

PR #2845, opened 2020-07-04
Review groovy documentation

pedroql:
I recently had problems using detekt on my project and in part it was caused by the documentation misleading me.

After reaching out in slack this was the proposed solution and it worked, so in part I removed some parts of the samples.

### DIFF
--- a/docs/pages/gettingstarted/groovydsl.md
+++ b/docs/pages/gettingstarted/groovydsl.md
@@ -29,23 +29,12 @@ using the detekt closure as described [here](#closure).
 ##### <a name="gradlegroovy">Configuration when using Groovy DSL</a>
 
 ```groovy
-repositories {
-    jcenter()
-
-    // or
-
-    mavenCentral()
-    jcenter {
-        content {
-            // just allow to include kotlinx projects
-            // detekt needs 'kotlinx-html' for the html report
-            includeGroup "org.jetbrains.kotlinx"
-        }
-    }
-}
-
 plugins {
     id "io.gitlab.arturbosch.detekt" version "{{ site.detekt_version }}"
+}
+
+repositories {
+    jcenter()
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/pages/gettingstarted/groovydsl.md
```diff
@@ -29,24 +29,13 @@ using the detekt closure as described [here](#closure).
...
-
-    mavenCentral()
-    jcenter {
-        content {
-            // just allow to include kotlinx projects
-            // detekt needs 'kotlinx-html' for the html report
-            includeGroup "org.jetbrains.kotlinx"
-        }
-    }
-}
```
pedroql, line 45: I removed the whole block because I did a quick test and it doesn't seem that I needed this repository. Just JCenter.

docs/pages/gettingstarted/groovydsl.md
```diff
@@ -61,15 +50,10 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:{{ site.detekt_version }}'
```
pedroql, line 64: This line had one or maybe two issues:
1. the latest version of this artifact is 4.0.0 but it is being replaced with the latest version of detekt
2. I will double check but I am not sure this dependency is needed.

docs/pages/gettingstarted/groovydsl.md
```diff
@@ -61,15 +50,10 @@ buildscript {
...
         classpath "io.gitlab.arturbosch.detekt:detekt-gradle-plugin:{{ site.detekt_version }}"
     }
 
 }
-plugins {
-    id "io.gitlab.arturbosch.detekt" version "{{ site.detekt_version }}"
-}
-
```
pedroql, line 72: This example was mixing different ways to apply plugins. `apply :plugin` will suffice.

cortinico:
Thanks for the contribution @pedroql  
I already submitted a PR for fixing the issues you pointed out as well as others. Can I ask you to provide your feedback on #2844 and let me know if that clarifies your doubts?

docs/pages/gettingstarted/groovydsl.md
BraisGabin: ups! The `{{ site.detekt_version }}` is my fault, I mean, I changed it by mistake for this placeholder. It should say `[version]`.

And yes, it's not important for detekt but I think that it helps to understand where to add the other lines "below the android gradle plugin".

docs/pages/gettingstarted/groovydsl.md
BraisGabin: That was added because of #2146.

pedroql:
@cortinico Thanks! I had a look at that PR and it looks good (left a comment). I will close this one!

PR #2851, opened 2020-07-05
Exclusion patterns

### LINKED ISSUE #2581: Allow globally defined exclusion pattern
## Expected Behavior
I would like to be able to globally edit the exclusion patterns, rather than chaning them one by one for each rules.

## Current Behavior
Currently each rules has its own exclusion pattern. That leads to a lot of repetition in the config files, and it is especially annoying and error prone when it is time to change the patterns as we are forced to go through all of them one by one.

## Context
I introduced new test source sets in a project and I have to tell detekt to ignore that sourceset for a bunch of rules. Which rules? Simple, all the rules that are already excluding the test sources.

But instead of having to change one line to tell detekt what are the patterns of test sources, I have to go through each rules and edit them one by one.

The configuration styled proposed by #775 would have been perfect.

BraisGabin:
Closes #2581

This PR introduces the `exclusion-patterns`. This way you can create a list of patterns to exclude in a list of rules. This PR also change how we exclude some rules from tests to use the new `exclusion-patterns`. For more context please read #2581

I didn't implement tests yet because I'll like to know what do you think about this implementation. It's not an elegant solution but I can't think of a more elegant way without refactoring all the core. What do you think?

And what do you think about the feature and names? The yaml straucutre?


detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -31,9 +31,43 @@ console-reports:
...
+  - patterns:
+      - '**/test/**'
+      - '**/androidTest/**'
+      - '**/commonTest/**'
+      - '**/jvmTest/**'
+      - '**/jsTest/**'
+      - '**/iosTest/**'
+    rules:
+      - 'comments'
+      - 'complexity>StringLiteralDuplication'
```
cortinico, line 44: commenting here but is more like a general comment.
I like the approach, I think it will benefit users a lot. 

I just have one comment on the YAML as:

```yaml
exclusion-patterns:
  - patterns:
      - '**/test/**'
    rules:
      - 'comments'
```

and 

```yaml
exclusion-patterns:
    patterns:
      - '**/test/**'
    rules:
      - 'comments'
```
are really easy to confuse (though we support only the first).

Moreover having more than one pattern will look a bit odd:

```
exclusion-patterns:
  - patterns:
	  - '**/test/**'
      - '**/androidTest/**'
      - '**/commonTest/**'
    rules:
      - 'comments'
  - patterns:
	  - '**/jvmTest/**'
      - '**/jsTest/**'
      - '**/iosTest/**'
    rules:
      - 'potential-bugs>LateinitUsage'
      - 'style>MagicNumber'
      - 'style>WildcardImport'
```

I don't really know how we can fix this. Having `exclusion-patterns`, `patterns` and `rules` all plural nouns makes things hard to understand for the user imho.
Maybe we can think about having a `pattern-id` or `pattern-name` used by the user to name a group of `patterns`/`rules` (that can also be reused in the future to explain why a certain rule was skipped).

BraisGabin: Agree. What do you think about this:
```yaml
exclusion-patterns:
  - tests:
    patterns:
      - '**/test/**'
      - '**/androidTest/**'
      - '**/commonTest/**'
    rules:
      - 'comments'
  - mappers:
    patterns:
      - '**/jvmTest/**'
      - '**/jsTest/**'
      - '**/iosTest/**'
    rules:
      - 'potential-bugs>LateinitUsage'
      - 'style>MagicNumber'
      - 'style>WildcardImport'
```

Or even better:

```yaml
exclusion-patterns:
  - 'dont apply in tests':
    patterns:
      - '**/test/**'
      - '**/androidTest/**'
      - '**/commonTest/**'
    rules:
      - 'comments'
  - 'mappers are an special case too':
    patterns:
      - '**/jvmTest/**'
      - '**/jsTest/**'
      - '**/iosTest/**'
    rules:
      - 'potential-bugs>LateinitUsage'
      - 'style>MagicNumber'
      - 'style>WildcardImport'
```
To be more clear that they are names

cortinico: > To be more clear that they are names

Yeah definitely clearer 👍

arturbosch: Having names/descriptions is really great!

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/ConfigPrinter.kt
```diff
@@ -108,6 +102,43 @@ object ConfigPrinter : DocumentationPrinter<List<RuleSetPage>> {
...
+              - 'naming>FunctionMaxLength'
+              - 'naming>FunctionMinLength'
+              - 'naming>FunctionNaming'
+              - 'naming>FunctionParameterNaming'
+              - 'naming>ObjectPropertyNaming'
+              - 'naming>PackageNaming'
+              - 'naming>TopLevelPropertyNaming'
+              - 'naming>VariableMaxLength'
+              - 'naming>VariableMinLength'
+              - 'naming>VariableNaming'
```
arturbosch, line 134: It seems like these are all rules from the naming rule set, should we just exclude `naming` here?
I think `MatchingDeclarationName` is missing here.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/YamlConfig.kt
```diff
@@ -67,13 +67,69 @@ class YamlConfig internal constructor(
...
+    exclusionPatterns: List<Map<String, List<String>>>,
+    ruleSetProperties: Map<String, Any>,
+    ruleSetName: String
+): Map<String, Any> {
+    var editedRuleSetProperties = ruleSetProperties
+    exclusionPatterns.forEach { exclusionPattern ->
+        val patterns = exclusionPattern["patterns"]
+        val rules = exclusionPattern["rules"]
+        if (!patterns.isNullOrEmpty() && !rules.isNullOrEmpty()) {
+            editedRuleSetProperties = fillExcludes(patterns, rules, ruleSetName, editedRuleSetProperties)
```
arturbosch, line 108: This loop looks like a reduce function.
Can we use a `ruleSetProperties.reduce` here and the reduction as an inner function here?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/YamlConfig.kt
```diff
@@ -67,13 +67,69 @@ class YamlConfig internal constructor(
...
+    var editedRuleSetProperties = ruleSetProperties
+    if (rules.contains(ruleSetName)) {
+        editedRuleSetProperties = ruleSetProperties.toMutableMap().apply {
+            set("excludes", patterns + (get("excludes") as? List<String> ?: emptyList()))
+        }
+    }
+    return editedRuleSetProperties.mapValues { (ruleName, ruleConfigAny) ->
+        val ruleConfig = (ruleConfigAny as? Map<String, Any>)?.toMutableMap()
+        if (rules.contains("$ruleSetName>$ruleName") && ruleConfig != null) {
+            ruleConfig["excludes"] = patterns + (ruleConfig["excludes"] as? List<String> ?: emptyList())
```
arturbosch, line 129: Can `(ruleConfig["excludes"] as? List<String> ?: emptyList())` be an inner function?
I think this code fragment is the same as above.

BraisGabin:
> I would like in the future to move out the knownledge about filters from the `Rule` and `RuleSet` classes so the filtering is done in core.

Agree! I'll finish this PR and then I'll try to move all this logic away. Any ideas for this?

BraisGabin:
>_< writing the test of this I just realised that this code doesn't work. You need to have the rule defined in the yaml so it can add the `exclude`. I'm going to close this PR and wait for #2861 to be merged. Then I'll try to move all the exclusion engine to core. And then I'll try again to implement this. I'm going to copy the yaml idea to the issue so we don't forget about them.

samuelneff:
Note for anyone else that wants this (I haven't looked at the other options yet). I think you can get some similar functionality by using YAML Alias Nodes.

https://yaml.org/spec/1.2.0/#id2585330

PR #2852, opened 2020-07-05
Fix typo ne documentation

BraisGabin:
As pointed out in #2845 by @pedroql the version of Android gradle plugin is not ralated with the detekt version.


BraisGabin:
This was fixed already by @cortinico

PR #2862, opened 2020-07-12
Cleanup code

amitdash291:
This PR achieves the following:
 - Use safe access `?.` instead of null check
 - Reuse `GuardClauses.kt` common code for `ThrowsCount` and remove duplicate code.
 - Add test case for excluding guard clauses in `ThrowsCountSpek`
 - Make the extension methods `internal` (in `GuardClauses.kt`) to avoid scope pollution

### DIFF
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
@@ -30,12 +30,12 @@ fun KtExpression.isGuardClause(): Boolean {
             ?: return false
 
         return ifExpr.`else` == null &&
-                returnExpr === ifExpr.then?.lastBlockStatementOrThis()
+                returnExpr == ifExpr.then?.lastBlockStatementOrThis()
     }
 
     fun KtReturnExpression.isElvisOperatorGuardClause(): Boolean {
         val elvisExpr = this.parent as? KtBinaryExpression
-        return elvisExpr != null && elvisExpr.operationToken == KtTokens.ELVIS
+        return elvisExpr?.operationToken == KtTokens.ELVIS
     }
 
     val returnExpr = this.findDescendantOfType<KtReturnExpression>() ?: return false
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
@@ -1,7 +1,11 @@
 package io.gitlab.arturbosch.detekt.rules
 
 import org.jetbrains.kotlin.lexer.KtTokens
-import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.KtBinaryExpression
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtIfExpression
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.KtReturnExpression
 import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
 import org.jetbrains.kotlin.psi.psiUtil.lastBlockStatementOrThis
 
--- a/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
+++ b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
@@ -1,11 +1,7 @@
 package io.gitlab.arturbosch.detekt.rules
 
 import org.jetbrains.kotlin.lexer.KtTokens
-import org.jetbrains.kotlin.psi.KtBinaryExpression
-import org.jetbrains.kotlin.psi.KtExpression
-import org.jetbrains.kotlin.psi.KtIfExpression
-import org.jetbrains.kotlin.psi.KtNamedFunction
-import org.jetbrains.kotlin.psi.KtReturnExpression
+import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
 import org.jetbrains.kotlin.psi.psiUtil.lastBlockStatementOrThis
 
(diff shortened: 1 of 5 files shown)


detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
```diff
@@ -77,17 +77,11 @@ class ReturnCountSpec : Spek({
...
         """
 
-            it("should get flagged for too-complicated guard clause, with EXCLUDE_GUARD_CLAUSES on") {
+            it("should get flagged for if statement with multiple returns, with EXCLUDE_GUARD_CLAUSES on") {
                 val findings = ReturnCount(TestConfig(mapOf(ReturnCount.EXCLUDE_GUARD_CLAUSES to "true")))
                     .compileAndLint(code)
                 assertThat(findings).hasSize(1)
             }
-
-            it("should get flagged for too-complicated guard clause, with EXCLUDE_GUARD_CLAUSES off") {
```
arturbosch, line 86: Could you explain why this testcase got removed?

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
```diff
@@ -30,12 +30,12 @@ fun KtExpression.isGuardClause(): Boolean {
...
 
         return ifExpr.`else` == null &&
-                returnExpr === ifExpr.then?.lastBlockStatementOrThis()
+                returnExpr == ifExpr.then?.lastBlockStatementOrThis()
     }
 
     fun KtReturnExpression.isElvisOperatorGuardClause(): Boolean {
         val elvisExpr = this.parent as? KtBinaryExpression
-        return elvisExpr != null && elvisExpr.operationToken == KtTokens.ELVIS
+        return elvisExpr?.operationToken == KtTokens.ELVIS
```
arturbosch, line 38: +1

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
```diff
@@ -30,12 +30,12 @@ fun KtExpression.isGuardClause(): Boolean {
             ?: return false
 
         return ifExpr.`else` == null &&
-                returnExpr === ifExpr.then?.lastBlockStatementOrThis()
+                returnExpr == ifExpr.then?.lastBlockStatementOrThis()
```
arturbosch, line 33: `===` is used here on purpose.
The return expr is the same node found as the last statement in the if block.
Or am I missing something?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
amitdash291: @arturbosch this test case is redundant, given it expected to be flagged even with the` EXCLUDE_GUARD_CLAUSES` flag's value being true, it will surely be flagged when the flag's value is false.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
amitdash291: thanks!

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
amitdash291: I see, so we need to match the reference rather than the structure of the node. @arturbosch if you confirm I will revert this change.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
arturbosch: Ah now I see. Please rename the testcase to `reports a too-complicated if statement for being a guard clause`

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
arturbosch: Yes, please revert this line.

amitdash291: This is reverted now.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
amitdash291: Renamed the test case as advised above

amitdash291:
@arturbosch I have done the suggested changes. I have also found and cleaned up code duplication in ThrowsCount. Please review the same and let me know of any suggestions / clarifications.

detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/GuardClauses.kt
```diff
@@ -33,11 +33,11 @@ fun KtExpression.isGuardClause(): Boolean {
...
-    fun KtReturnExpression.isElvisOperatorGuardClause(): Boolean {
-        val elvisExpr = this.parent as? KtBinaryExpression
-        return elvisExpr != null && elvisExpr.operationToken == KtTokens.ELVIS
-    }
-
     val returnExpr = this.findDescendantOfType<KtReturnExpression>() ?: return false
     return isIfConditionGuardClause(returnExpr) || returnExpr.isElvisOperatorGuardClause()
 }
+
+fun KtExpression.isElvisOperatorGuardClause(): Boolean {
```
BraisGabin, line 40: Could this be `internal`? I'm a bit worried about the scope polution... I'm going to open an issue to see if we can create some rule for this cases.

amitdash291: Sounds like a good idea. I would recommend it to be tackled in a separate issues as you suggested.

BraisGabin: Sure, I just created #2863. But, any way, can we make this function `internal`?

amitdash291: @BraisGabin I have made this extension function (and the other one in the file) `internal`
cc @arturbosch

PR #2864, opened 2020-07-13
Add new rule UseEmptyCounterpart

### LINKED ISSUE #2850: Add a rule to flag places where `emptyList` (and similar)  can be used
Kotlin provides the functions `emptyArray`, `emptyList`, `emptyMap`, `emptySequence`, and `emptySet` to improve the readability of instantiating an empty "collection".

The below can also be applied to the other "empty" variations.

### Noncompliant Code

```kotlin
val strs = listOf<String>()
```

### Compliant Code

```kotlin
val strs = emptyList<String>()
```

veyndan:
Closes #2850.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpart.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpart.kt
@@ -1,0 +1,59 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.psi.KtCallExpression
+
+/**
+ * Instantiation of an object's "empty" state should use the object's "empty" initializer for clarity purposes.
+ *
+ * <noncompliant>
+ * arrayOf()
+ * listOf()
+ * mapOf()
+ * sequenceOf()
+ * setOf()
+ * </noncompliant>
+ *
+ * <compliant>
+ * emptyArray()
+ * emptyList()
+ * emptyMap()
+ * emptySequence()
+ * emptySet()
+ * </compliant>
+ */
+class UseEmptyCounterpart(config: Config) : Rule(config) {
+
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Style,
+        """Instantiation of an object's "empty" state should use the object's "empty" initializer""",
+        Debt.FIVE_MINS
+    )
+
+    override fun visitCallExpression(expression: KtCallExpression) {
+        super.visitCallExpression(expression)
+
+        val calleeExpression = expression.calleeExpression ?: return
+        val emptyCounterpart = exprsWithEmptyCounterparts[calleeExpression.text]
+
+        if (emptyCounterpart != null && expression.valueArguments.isEmpty()) {
+            val message = "${calleeExpression.text} can be replaced with $emptyCounterpart"
+            report(CodeSmell(issue, Entity.from(expression), message))
+        }
+    }
+}
+
+private val exprsWithEmptyCounterparts = mapOf(
+    "arrayOf" to "emptyArray",
+    "listOf" to "emptyList",
+    "mapOf" to "emptyMap",
+    "sequenceOf" to "emptySequence",
+    "setOf" to "emptySet"
+)
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpart.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpart.kt
@@ -7,7 +7,10 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
 
 /**
  * Instantiation of an object's "empty" state should use the object's "empty" initializer for clarity purposes.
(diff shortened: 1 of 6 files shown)


detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpart.kt
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * <compliant>
+ * emptyArray()
+ * emptyList()
+ * emptyMap()
+ * emptySequence()
+ * emptySet()
+ * </compliant>
+ */
+class UseEmptyCounterpart(config: Config) : Rule(config) {
```
veyndan, line 31: I originally was going to go with "UseEmptyCollection", but decided it would be confusing considering this rule applies to "collections" of elements, but they don't all implement the `Collection` interface.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpartSpec.kt
```diff
@@ -0,0 +1,78 @@
...
+}
+
+private data class ArgExpr(
+    private val name: String,
+    @Language("kotlin") val noArgExample: String,
+    @Language("kotlin") val argExample: String
+) {
+
+    override fun toString(): String = name
+}
```
schalkms, line 78: This seems to be a bit overkill for the simple use cases you want to test.
I think we are better off to put the cases into 1 simple code snippet string like so.

```kotlin
arrayOf()
listOf()
mapOf()
sequenceOf()
setOf()
```

What do you think?

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpartSpec.kt
```diff
@@ -0,0 +1,78 @@
...
+            }
+
+            it("does not report $emptyInstantiation") {
+                val findings = rule.compileAndLint(emptyInstantiation.example)
+                assertThat(findings).isEmpty()
+            }
+
+            it("does not report $argInstantiation with arguments") {
+                val findings = rule.compileAndLint(argInstantiation.argExample)
+                assertThat(findings).isEmpty()
```
schalkms, line 55: For future regressions it's a good idea to test some more negative cases.

What happens if the user names a variable or a function `setOf`? An intensive test suite can prevent errors in the future. Maybe you can come up with more test cases like this.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpartSpec.kt
veyndan: Makes sense. I originally wanted a test case for each expression, but it probably isn't necessary.

Fixed in https://github.com/detekt/detekt/pull/2864/commits/6fd552c698c99d6956caad946463424a1c8fe51e.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpartSpec.kt
veyndan: I added your recommended test in https://github.com/detekt/detekt/pull/2864/commits/9ba2c351429dfc6ae2ca1a4c432d8727a83a386c but have found that it does not pass. This is because I only compare the text of the function instead of the fully qualified name.

I've done some poking around, but can't find the API to do this. Do you have any pointers?

BraisGabin:
After #2865 we should move tis rule to the correct module: `:detekt-rules-style`

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpartSpec.kt
schalkms: You would need to use type and symbol solving for this. `BindingContext` is the keyword here.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpartSpec.kt
```diff
@@ -0,0 +1,67 @@
...
+class UseEmptyCounterpartSpec : Spek({
+    setupKotlinEnvironment()
+
+    val rule by memoized { UseEmptyCounterpart(Config.empty) }
+
+    describe("UseEmptyCounterpart rule") {
+
+        it("reports no-arg instantiation") {
+            val code = """
+                val array = arrayOf<Any>()
```
cortinico, line 19: nit: maybe you can add a test where the type parameter is inferred:

```
val array : Array<Any> = arrayOf()
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpart.kt
```diff
@@ -0,0 +1,59 @@
...
+        if (emptyCounterpart != null && expression.valueArguments.isEmpty()) {
+            val message = "${calleeExpression.text} can be replaced with $emptyCounterpart"
+            report(CodeSmell(issue, Entity.from(expression), message))
+        }
+    }
+}
+
+private val exprsWithEmptyCounterparts = mapOf(
+    "arrayOf" to "emptyArray",
+    "listOf" to "emptyList",
```
cortinico, line 55: I think it should be safe to add also:
```
    "listOfNotNull" to "emptyList",
```

Though there is a type restriction (`listOfNotNull<Int?>` and `emptyList<Int>` are of different type), but I think it should be safe anyway to flag the usage of a `listOfNotNull()` with no parameter.

veyndan: I like this idea, though I have an alternative proposal that would flag incorrect usages of `listOfNotNull()` and `listOfNotNull(elem, …)`.

If we extend the rule `UselessCallOnNotNull` to also flag invocations of `listOfNotNull` with all non-nullable arguments, then a no-arg invocation of `listOfNotNull` would also be flagged. The user can then replace `listOfNotNull` with `listOf` for all these incorrect usages. Then `UseEmptyCounterpart` would flag all the no-arg `listOf` invocations as a result of this.

WDYT?

cortinico: > If we extend the rule UselessCallOnNotNull to also flag invocations of listOfNotNull with all non-nullable arguments

If possible I'll keep this out of the scope of this PR.

>then a no-arg invocation of listOfNotNull would also be flagged. The user can then replace listOfNotNull with listOf for all these incorrect usages. Then UseEmptyCounterpart would flag all the no-arg listOf invocations as a result of this.

Ideally we want to provide immediate feedback with a potential solution. Having the user go through `listOfNotNull()` -> `listOf` -> `emptyList()` and two Detekt runs seems cumbersome to me.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpart.kt
```diff
@@ -0,0 +1,59 @@
...
+        Debt.FIVE_MINS
+    )
+
+    override fun visitCallExpression(expression: KtCallExpression) {
+        super.visitCallExpression(expression)
+
+        val calleeExpression = expression.calleeExpression ?: return
+        val emptyCounterpart = exprsWithEmptyCounterparts[calleeExpression.text]
+
+        if (emptyCounterpart != null && expression.valueArguments.isEmpty()) {
```
amitdash291, line 46: Instead of `emptyCounterpart != null` probably guard clause can be used like the following:
```
val emptyCounterpart = exprsWithEmptyCounterparts[calleeExpression.text] ?: return
```

arturbosch:
Hey @veyndan , can you take a look why a testcase is crashing?
We would like your contribution in the next release :).

veyndan:
Hey @arturbosch, the build is currently failing because of https://github.com/detekt/detekt/pull/2864#discussion_r453753049. I've been swamped for the past couple of weeks, but I should have some time this coming weekend to try and fix the tests.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpart.kt
veyndan: Good point! Fixed in 920ecd0.

detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpartSpec.kt
veyndan: Thanks, that quickly put me on the right track.

Fixed in 920ecd0.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpart.kt
veyndan: Yeah that makes sense.

Fixed in https://github.com/detekt/detekt/pull/2864/commits/23b7b81d4641d04aff08a7500cb2aade03e29cd4.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseEmptyCounterpartSpec.kt
veyndan: Fixed in https://github.com/detekt/detekt/pull/2864/commits/8587d4f184fccf32821ef91ca449816b44bcf9a1.

PR #2865, opened 2020-07-13
Split rules module into a module per rule set

arturbosch:
This significantly improves build and test times.

I think we now have a perfectly parallel build xD

2020-07-13-144610_1686x320_scrot


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/PathFilters.kt
```diff
@@ -43,7 +43,7 @@ fun Config.createPathFilters(): PathFilters? {
     return PathFilters.of(includes, excludes)
 }
 
-private fun Config.valueOrDefaultCommaSeparated(
+fun Config.valueOrDefaultCommaSeparated(
```
arturbosch, line 46: This one was a duplicate in `detekt-rules`.
We use this api/internal implementation now in the rule sets.

Imo we should make this function part of the public api as we want to support yaml lists.

arturbosch:
```
        Caused by:
        java.lang.IllegalArgumentException: Kotlin script engine not supported
            at io.github.detekt.test.utils.KotlinScriptEnginePool.createEngine(KotlinScriptEnginePool.kt:37)
            ... 10 more
```

sounds like a missing dependecy?

detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinScriptEnginePool.kt
```diff
@@ -31,10 +31,7 @@ object KotlinScriptEnginePool {
 
     private fun createEngine(): KotlinJsr223JvmLocalScriptEngine {
         setIdeaIoUseFallback() // To avoid error on Windows
-
-        val scriptEngineManager = ScriptEngineManager()
-        val engine = scriptEngineManager.getEngineByExtension("kts") as? KotlinJsr223JvmLocalScriptEngine
-        requireNotNull(engine) { "Kotlin script engine not supported" }
-        return engine
+        val engine = KotlinJsr223JvmLocalScriptEngineFactory().scriptEngine as? KotlinJsr223JvmLocalScriptEngine
+        return requireNotNull(engine) { "Kotlin script engine not supported" }
```
schalkms, line 35: Good one!

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/PathFilters.kt
BraisGabin: 🤔 do we want to encorage the support of yaml lists AND comma separated strings? I don't think so, I mean, I think that it's better to only support yaml lists.

detekt-rules-style/build.gradle.kts
```diff
@@ -0,0 +1,6 @@
+dependencies {
+    compileOnly(project(":detekt-api"))
+    compileOnly(project(":detekt-metrics"))
```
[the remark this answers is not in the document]
arturbosch: `api` is meant to be compileOnly for rules.
The metrics one is used for lines of code. Imo it is okay, not as bad as depending on core.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/Case.kt
```diff
@@ -0,0 +1,24 @@
...
+    MaxLineLengthWithLongComments("/MaxLineLengthWithLongComments.kt"),
+    UtilityClassesPositive("/UtilityClassesPositive.kt"),
+    UtilityClassesNegative("/UtilityClassesNegative.kt"),
+    NoTabsNegative("/NoTabsNegative.kt"),
+    NoTabsPositive("/NoTabsPositive.kt"),
+    UnusedPrivateMemberPositive("/UnusedPrivateMemberPositive.kt"),
+    UnusedPrivateMemberNegative("/UnusedPrivateMemberNegative.kt");
+
+    fun path(): Path = resourceAsPath(file)
+}
```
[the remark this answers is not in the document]
arturbosch: >  Good one. I'm not sure it's just an idea that I had. Is it better `:detekt-rules-style` or `:detekt-rules:style`? I mean, all the rule modules as submodules of `:detekt-rules`

If we would develop in a monorepo with different projects I would also favor `detekt-rules:style` to have a clearer project structure.
However Gradle's `project.subprojects` returns just the modules one level deeper. Therefore our build scripts would need to be more complex and test if a module is a "meta-module" which just groups other modules and does not need plugins like kotlin and stuff for test.

Lets keep every module top level for now.

cortinico:
Awesome change @arturbosch 🎉

PR #2875, opened 2020-07-15
Simplify internal parsing to KtFile's

arturbosch:


detekt-parser/src/test/kotlin/io/github/detekt/parser/KtCompilerTest.kt
```diff
@@ -32,10 +34,18 @@ class KtCompilerTest : Spek({
...
+            val ktFile = ktCompiler.compile(cssPath, cssPath.resolve("test.css"))
+
+            val errors = mutableListOf<PsiErrorElement>()
+            ktFile.accept(object : KtTreeVisitorVoid() {
+                override fun visitErrorElement(element: PsiErrorElement) {
+                    errors.add(element)
+                }
+            })
+
+            assertThat(errors).isNotEmpty()
```
schalkms, line 48: Is this a behavioral change now, since you had to adapt the test case?

arturbosch: Hm, yes it is. Not sure how to handle this.
In `KtTreeCompiler` we throw an exception if the file ending is not `kt` or `kts`.

In the old implementation the PsiFactory tried to use `Language.Any` and because no language was registered was `.css` it didn't tried to parse anything and returned null which we checked with `error(...)`.
The `KtPsiFactory` always tries to parse Kotlin and returns a `PsiFile` with errors.

@schalkms @BraisGabin @cortinico should we throw an error on wrong file extension like when parsing directories?

arturbosch: Ok, we ignored it when parsing many files:
https://github.com/detekt/detekt/blob/5a0fe624e571d20e1619b1e42360c2b1f513a2ce/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtTreeCompiler.kt#L31

PR #2877, opened 2020-07-15
Update Gradle to 6.6-RC1

arturbosch:
Does this Gradle work for us with Java 11 ?


arturbosch:
The new dokka seems to be released pretty soon with Kotlin 1.4-M3 ....

PR #2879, opened 2020-07-16
Check string arguments with TR enabled

realdadfish:
Check simple string arguments even when type resolution is enabled.
This prevented the `UseRequire` and `UseCheckOrError` checks find issues with TR enabled.


realdadfish:
I'm not sure if / where this has to be described to get picked up by the changelog, or is this done automatically? Or should this just be merged?

cortinico:
> I'm not sure if / where this has to be described to get picked up by the changelog, or is this done automatically? Or should this just be merged?

It will be picked up automatically once we merge it 👍

Issue #2841, opened 2020-07-03
Java 11: Provided configuration file is invalid: Structure must be from type Map<String,Any>!


Tammie47:
I'm struggling with setting up detekt in my project. If I use Java 8 as JVM, detekt works without a problem but we use Java 11 in our project. In version 1.10.0 this error occurs every time we execute ./gradlew detekt:

```
./gradlew detekt --stacktrace
> Task :api:detekt FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':api:detekt'.
> Provided configuration file is invalid: Structure must be from type Map<String,Any>!

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':api:detekt'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$3.accept(ExecuteActionsTaskExecuter.java:166)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$3.accept(ExecuteActionsTaskExecuter.java:163)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:191)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:156)
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:62)
        at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:108)
        at org.gradle.api.internal.tasks.execution.ResolveBeforeExecutionOutputsTaskExecuter.execute(ResolveBeforeExecutionOutputsTaskExecuter.java:67)
        at org.gradle.api.internal.tasks.execution.ResolveAfterPreviousExecutionStateTaskExecuter.execute(ResolveAfterPreviousExecutionStateTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:94)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
[...]

PR #2884, opened 2020-07-20
Preserve original exception when loading an invalid config

arturbosch:
#2841


arturbosch:
> Should we add a test case for it?

Yes! Thanks, also forgot to add the document files

Issue #2883, opened 2020-07-20
Sign artifacts and publish checksums


arturbosch:
## Expected Behavior
We should use the Gradle signing plugin to sign our artifacts with gpg and generate and publish checksums next to our artifacts.
This will increase reliability for our users.

## Current Behavior
We let Bintray autosign our artifacts and do not publish any checksums.

## Context
https://docs.gradle.org/current/userguide/signing_plugin.html
https://github.com/gradle/gradle-checksum

```
From 57894cd67ff5fffec4b2a8c6c15b9beeb5dc267f Mon Sep 17 00:00:00 2001
From: Artur Bosch <arturbosch@gmx.de>
Date: Mon, 20 Jul 2020 12:06:57 +0200
Subject: [PATCH] Publish sha256 checksums

---
 buildSrc/src/main/kotlin/packaging.gradle.kts | 14 ++++++++++++++
 1 file changed, 14 insertions(+)

diff --git a/buildSrc/src/main/kotlin/packaging.gradle.kts b/buildSrc/src/main/kotlin/packaging.gradle.kts
index b07f38a04..b5629ee74 100644
--- a/buildSrc/src/main/kotlin/packaging.gradle.kts
+++ b/buildSrc/src/main/kotlin/packaging.gradle.kts
@@ -145,3 +145,17 @@ configure(subprojects.filter { it.name != "detekt-bom" }) {
         }
     }
 }
+
+subprojects {
+
+    tasks.withType<Zip>().configureEach {
+        doLast {
+            ant.withGroovyBuilder {
+                "checksum"(
+                    "file" to this@configureEach.archiveFile.get().asFile,
+                    "algorithm" to "SHA-256"
+                )
+            }
+        }
+    }
+}
--
2.27.0
```

cortinico:
I don't think we need to provide both and we should be fine with just signing.
Signing should already give proof of identity to our users.

Checksums can be calculated by the repository imho (also we would need to calculate both md5 and sha1 if we decided to do so).

schalkms:
I agree with Nico here.

cortinico:
Turns out with `maven-publish` MD5 is computed locally, so we get it for free 🎉  https://github.com/detekt/detekt/pull/2885

arturbosch:
> Turns out with `maven-publish` MD5 is computed locally, so we get it for free  #2885

And the gradle metadata contains all remaining hashes ^^.
[...]

PR #2885, opened 2020-07-20
Publish to Bintray using maven-publish

### LINKED ISSUE #2878: Detekt Deployment is missing Gradle Module files
## Expected Behavior
One of:
- Detekt `.module` files are available on Bintray/MavenCentral.
- Detekt's `pom` file shouldn't contain the Gradle Module marker: `<!--  do_not_remove: published-with-gradle-metadata  -->`

## Observed Behavior
Currently we have published `pom` files that contains the `<!--  do_not_remove: published-with-gradle-metadata  -->` marker for all the modules ([see here](https://repo1.maven.org/maven2/io/gitlab/arturbosch/detekt/detekt-core/1.10.0/detekt-core-1.10.0.pom)).

## Steps to Reproduce
Module files are missing for all the modules ([you can check it here](https://repo1.maven.org/maven2/io/gitlab/arturbosch/detekt/)).

The effective impact of this is probably what was reported in this comment https://github.com/detekt/detekt/pull/2814#issuecomment-659239216

## Context
I did some investigation and seems like the issue is that the gradle-bintray-plugin is failing to publish `.module` files: https://github.com/bintray/gradle-bintray-plugin/issues/229

Given that we are applying both `gradle-bintray-plugin` and `maven-publish` (for publishing on artifactory), my suggestion is that we drop the `gradle-bintray-plugin` as we use only `maven-publish` for publishing (to hit bintray/artifactory).

cortinico:
Moving Detekt to use only `maven-publish` for publishing and removing `gradle-bintray-plugin`.

I think this simplifies overall our setup having a couple of benefits:
- We publish the `.module` files.
- We publish the `.md5` files computed locally (partial fix for #2883)

There a couple of drawbacks:

1. Unfortunately due to https://github.com/gradle/gradle/issues/11412 I had to `systemProp.org.gradle.internal.publish.checksums.insecure=true`. That would have allowed us to publish also locally computed SHA512 for free. This unfortunately _confuses_ Bintray that we're releasing a new version for every module, that is really annoying.

2. @arturbosch you now need to manually start the Maven Central sync. If needed we can have a Gradle task that hits an HTTP endpoint to start the sync. Otherwise this has to be done manually from the Bintray UI.

Here a sample publishing:
https://bintray.com/cortinico/maven/detekt-sample-publishing/

Fixes #2878

### DIFF
--- a/buildSrc/src/main/kotlin/packaging.gradle.kts
+++ b/buildSrc/src/main/kotlin/packaging.gradle.kts
@@ -1,5 +1,4 @@
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
-import com.jfrog.bintray.gradle.BintrayExtension
 import groovy.lang.GroovyObject
 import org.jfrog.gradle.plugin.artifactory.dsl.ArtifactoryPluginConvention
 import org.jfrog.gradle.plugin.artifactory.dsl.PublisherConfig
@@ -9,7 +8,6 @@ plugins {
     `java-library` apply false // is applied in commons; make configurations available in this script
     `maven-publish` apply false
     id("com.jfrog.artifactory") apply false
-    id("com.jfrog.bintray") apply false
 }
 
 project(":detekt-cli") {
@@ -27,7 +25,6 @@ subprojects {
 
     apply {
         plugin("maven-publish")
-        plugin("com.jfrog.bintray")
         plugin("com.jfrog.artifactory")
     }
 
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/kotlin/packaging.gradle.kts
```diff
@@ -27,49 +25,26 @@ subprojects {
...
-            })
-        })
-    }
+    val versionName = project.version as? String
 
     publishing {
+        repositories {
+            maven {
+                name = "bintray"
+                url = uri("https://api.bintray.com/maven/arturbosch/code-analysis/detekt/;publish=1;override=1")
```
cortinico, line 41: We could set `publish=0` to prevent publishing of a new version from Bintray. Then @arturbosch you would have to manually publish the artifacts from Bintray (this allows to verify the deployment before publishing, maybe it's useful?)

cortinico:
@arturbosch was this working correctly for 1.11.0-RC?

arturbosch:
Yep, so many files but works greatly! Thanks.

PR #2886, opened 2020-07-22
OptionalUnit: don't report it if Unit is used as an expression

### LINKED ISSUE #2452: False positive in OptionalUnit rule
## Expected Behavior
```kotlin
val <T> T.exhaustive: T
    get() = this

enum class E { A, B }

fun a(value: E) {
    when (value) {
        E.A -> println("hi")
        E.B -> {
            if (System.currentTimeMillis() % 2 == 0L) {
                println("hi")
            }
            Unit
        }
    }.exhaustive
}
```

This code is correct. I need to add Unit at the end of `B` because `if` is not an expression so `exhaustive` fails.

## Observed Behavior
This code is flagged

## Context
I want to force exhaustive `when`s so I need to do things like this. Maybe I'm missing an easier and more elegant way to do it, if so, please tell me.

## Your Environment

t-kameyama:
Fixes #2452

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/OptionalUnit.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/OptionalUnit.kt
@@ -8,11 +8,21 @@ import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.rules.isOverride
+import org.jetbrains.kotlin.cfg.WhenChecker
 import org.jetbrains.kotlin.psi.KtBlockExpression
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/OptionalUnit.kt
```diff
@@ -57,15 +67,34 @@ class OptionalUnit(config: Config = Config.empty) : Rule(config) {
     }
 
     override fun visitBlockExpression(expression: KtBlockExpression) {
-        expression.statements
-                .filter { it is KtNameReferenceExpression && it.text == UNIT }
+        val statements = expression.statements
+        val lastStatement = statements.lastOrNull() ?: return
+        statements
```
BraisGabin, line 72: Couls these if/return be replaced with an if/else?

t-kameyama: I've replaced these with 'when' expression.

PR #2896, opened 2020-07-26
Implement a proof of concept using the Gradle worker api

arturbosch:
I've tried to use the worker api to mitigate our classloader caching problems when changing detekt artifacts.

However the worker api has it's own shortcomings. It does for example not allow to cache classloaders and brings us back the jvm startup overhead.
This will bring us back to the old javaexec-implementation - performance-wise.

https://github.com/gradle/gradle/issues/8221
https://github.com/gradle/gradle/issues/9196

Imo this is also the reason why Gradle internal quality plugins haven't yet migrated to the worker api:
https://github.com/gradle/gradle/issues/9196

The quality plugins use an `IsolatedAntBuilder` which uses internally  a `ClassPathToClassLoaderCache`.

I open and close this PR for documentation purposes.
We will need to stay on our custom classloader cache and call the cli module reflectively for the best performance.


cortinico:
> > Imo this is also the reason why Gradle internal quality plugins haven't yet migrated to the worker api
> 
> By now [Checkstyle, PMD, and CodeNarc are all now implemented via the worker API](https://github.com/gradle/gradle/issues/9196#issuecomment-1338382177). Does that open new opportunities for Detekt?

This just got merged:
- https://github.com/detekt/detekt/pull/4128

PR #2898, opened 2020-07-27
Use KtFile.name instead of storing an absolute path key

arturbosch:


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Signatures.kt
```diff
@@ -20,21 +22,43 @@ internal fun PsiElement.searchClass(): String {
...
+ * KtCompiler wrongly used Path.filename as the name for a KtFile instead of the whole path.
+ * This resulted into the question "How do we get the absolute path from a KtFile?".
+ * Fixing this problem, we do not need KtFile.absolutePath anymore.
+ *
+ * Fixing the filename will change all baseline signatures.
+ * Therefore we patch the signature here to restore the old behavior.
+ *
+ * Fixing the baseline will need a new major release - #2680.
+ */
 internal fun PsiElement.buildFullSignature(): String {
```
cortinico, line 43: Should we add a test for this extension function?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Signatures.kt
```diff
@@ -1,16 +1,18 @@
...
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.psiUtil.endOffset
 import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 import org.jetbrains.kotlin.psi.psiUtil.parents
 import org.jetbrains.kotlin.psi.psiUtil.startOffset
 
-private val signatureRegex = Regex("\\s(\\s|\t)+")
+private val multipleWhitespaces = Regex("\\s(\\s|\t)+")
```
cortinico, line 15:
```suggestion
private val multipleWhitespaces = Regex("\\s{2,}")
```
This should suffice and is also more aligned to the name of the variable.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Signatures.kt
```diff
@@ -20,21 +22,43 @@ internal fun PsiElement.searchClass(): String {
     val classElement = this.getNonStrictParentOfType<KtClassOrObject>()
     var className = classElement?.name
     if (className != null && className == "Companion") {
-        classElement?.parent?.getNonStrictParentOfType<KtClassOrObject>()?.name?.let {
-            className = "$it.$className"
+        val parentClassName = classElement?.getStrictParentOfType<KtClassOrObject>()?.name
+        if (parentClassName != null) {
+            className = "$parentClassName.$className"
```
BraisGabin, line 27: 🤔 couldn't this lead to things like `"asdf$null"`?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Signatures.kt
arturbosch: This is tested through `Entity.from` in `EntitySpec` ^^

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Signatures.kt
arturbosch: This block is protected with a `class != null` if above :)

PR #2900, opened 2020-07-27
Add bindingContext function to FileProcessListener - #2872

### LINKED ISSUE #2872: Support for BindingContext in FileProcessListener
## Current Behavior
Implementations of `Rule` have access to the `BindingContext`, granting them the ability to perform complex analysis within the scope of the entire compilation unit. On the other hand, the `FileProcessListener` (and `AbstractProcessor`, etc) APIs do not provide any way to access the `BindingContext`, limiting their ability to resolve types, functions, and so forth.

## Expected Behavior
In order to enable the same level of analysis for implementations of `FileProcessListener`, the API should provide them a `BindingContext` against which references can be resolved.

## Context
I am in the process of implementing "occurrences counting" metrics for our codebase such as "number of invocations of function A" and "number of constructor calls to classes which implement interface X". Complex queries such as this require a `BindingContext`, but the provided mechanism for making such queries, `AbstractProjectMetricProcessor`, does not provide it.

arturbosch:
Closes #2872


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/FileProcessListener.kt
```diff
@@ -13,25 +18,79 @@ interface FileProcessListener : Extension {
      * Use this to gather some additional information for the real onProcess function.
      * This calculation should be lightweight as this method is called from the main thread.
      */
-    fun onStart(files: List<KtFile>) {}
+    @Deprecated(
+        "Use alternative with a binding context.",
+        ReplaceWith("onStart(files, bindingContext)")
+    )
+    fun onStart(files: List<KtFile>) {
```
cortinico, line 25: nit: is there a specific reason why there is a new line between the parentheses here?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/FileProcessListener.kt
```diff
@@ -1,10 +1,15 @@
+@file:Suppress("DEPRECATION")
```
BraisGabin, line 1: Maybe inline this Suppress? This way, when we remove the old versions we will remove the Suppress too for sure. To ensure that we don't forget to remove this.

schalkms:
We should discuss when we want to have settings for type and symbol solving as a mandatory step. This would reduce the complexity of detekt.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/FileProcessListener.kt
arturbosch: IntelliJ auto formatting ... idk if they updated it or we never formatted it accordingly

arturbosch:
> We should discuss when we want to have settings for type and symbol solving as a mandatory step. This would reduce the complexity of detekt.

This could be quite hard for users with generated code to configure properly. I will soon invest more time in the compiler plugin where we get resolution for free.

PR #2903, opened 2020-07-28
Add NullableToStringCall rule

### LINKED ISSUE #2901: New rule: toString over a nullable value
```
val a: Any?
val key: String = a.toString() // this line should be flagged
val key2: String = "$a" // same here
```

This code compiles, and have a defined behaviour. But, usually, it is a potential bug. The idea is to have a rule that detects the use of `toString()` (explicit or implicit) and flag it.

## Context
It's really strange that you really want to print `"null"` when you have a `null` value. If you are working with UI you probably want `""`, `"-"` or something else. And the same works for any other piece of code for example this one: https://github.com/detekt/detekt/pull/2898/files#diff-6576486bfd0ee34b1da2a5967fba6cf4R27

t-kameyama:
Fixes #2901


detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullableToStringCall.kt
```diff
@@ -0,0 +1,90 @@
...
+        when {
+            qualified != null -> {
+                if (qualified.descriptor()?.fqNameOrNull() != FqName("kotlin.toString")) return
+            }
+            expression.parent is KtStringTemplateEntry -> {
+                val languageVersionSettings = languageVersionSettings ?: return
+                val descriptor = expression.descriptor() ?: return
+                val originalType = descriptor.returnType ?.takeIf { it.isNullable() } ?: return
+                @Suppress("DEPRECATION")
+                val dataFlowValueFactory = DataFlowValueFactoryImpl(languageVersionSettings)
```
cortinico, line 70: This class is deprecated and has a clear deprecation message on top:
https://github.com/JetBrains/kotlin/blob/687d13a320b4a5697aedca3ed31f6568a40deafa/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/smartcasts/DataFlowValueFactoryImpl.kt#L28

I don't think we want to directly depend on this class.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullableToStringCall.kt
```diff
@@ -0,0 +1,90 @@
...
+ * }
+ *
+ * fun bar(a: Any?): String {
+ *     return "$a"
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * fun foo(a: Any?): String {
+ *     return a?.toString ?: "-"
```
cortinico, line 38:
```suggestion
 *     return a?.toString() ?: "-"
```

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullableToStringCallSpec.kt
```diff
@@ -0,0 +1,97 @@
...
+    describe("NullableToString rule") {
+        it("reports when a nullable toString is explicitly called") {
+            val code = """
+                fun test(a: Any?) {
+                    println(a.toString())
+                }
+            """
+            val actual = subject.compileAndLintWithContext(env, code)
+            Assertions.assertThat(actual).hasSize(1)
+            Assertions.assertThat(actual.first().issue.id).isEqualTo("NullableToStringCall")
```
cortinico, line 25: You can remove all the `Assertions.assertThat(actual.first().issue.id)...` as your `subject` contains only  `NullableToStringCall`

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullableToStringCallSpec.kt
```diff
@@ -0,0 +1,97 @@
...
+            Assertions.assertThat(actual.first().message).isEqualTo("This call '\${a}' may return the string \"null\".")
+        }
+
+        it("reports when a nullable toString is implicitly called in a raw string template") {
+            val code = """
+                fun test(a: Any?) {
+                    println(${'"'}${'"'}${'"'}${'$'}a${'"'}${'"'}${'"'})
```
cortinico, line 56: It feels like this test is a bit too complicated. I'm having a hard time finding a real-world scenario where such code would be used.

BraisGabin: This code just test `println("""$a""")` so `${'"'}""${'$'}a""${'"'}` should be enought.

BraisGabin:
Why do you need `LanguageVersionSettings`? If you want to extend the detekt API we will need to move it to another PR and add some explanation for the change.

> Ideally the string template handling should be added afterwards. Potentially this can also be a separate rule as the name `NullableToStringCall` becomes confusing. Something like `NullableStringInTemplate` or something along this line.

Why do you think that this should be two different rules? I mean, it's the same error. When you do this: `"hello $any"` you are doing an implicit call of `toString()`. `NullableToStringCall` doesn't seem a bad name thinking of this fact.

cortinico:
> Why do you think that this should be two different rules? I mean, it's the same error. When you do this: `"hello $any"` you are doing an implicit call of `toString()`. `NullableToStringCall` doesn't seem a bad name thinking of this fact.

I don't have a strong opinion on having two separate rules actually. On the other hand I do believe the explicit/implicit `toString()` call should be distinguished (if is not through a separate rule, a config option will also work).

t-kameyama:
> Why do you need LanguageVersionSettings?

We don't want the following cases to be false positives. 
```kotlin
fun test(a: Any?) {
    if (a != null) {
        println("$a") // 'a' is cast to 'String'
    }
}
```

To do so, we need to know if the variables have been smart cast. To know if it has been cast, we need the 'languageVersion' as follows.
https://github.com/t-kameyama/detekt/blob/16d7dd4dfbecfbc2b3ae0bb15cb1f3f1714cb7d5/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullableToStringCall.kt#L69-L75

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullableToStringCall.kt
t-kameyama: We can get a container with the following function and get an instance of `DataFlowValueFactory` from it.
https://github.com/JetBrains/kotlin/blob/017f640f26e29f1b1d0909d89e812aca9173691c/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/TopDownAnalyzerFacadeForJVM.kt#L130

However, I'm not using a container here because we only need an instance of `DataFlowValueFactory`.

t-kameyama:
> Ideally the string template handling should be added afterwards. Potentially this can also be a separate rule as the name `NullableToStringCall` becomes confusing. Something like `NullableStringInTemplate` or something along this line.

I don't think it's necessary to separate the two because I think it's the same error.

cortinico:
> If you want to extend the detekt API we will need to move it to another PR and add some explanation for the change.

@t-kameyama please follow what @BraisGabin already suggested.

arturbosch:
@t-kameyama thanks for the rule!
I agree with @cortinico that we should prefer a simpler version of this rule for now as we want to go into a different direction with the rule api (I wrote a longer comment in https://github.com/detekt/detekt/pull/2929#issuecomment-669341879).

BraisGabin:
@t-kameyama please refactor this PR so we can merge it :). And thanks again for this great contribution!

t-kameyama:
PTAL

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullableToStringCall.kt
```diff
@@ -0,0 +1,89 @@
...
+ * fun foo(a: Any?): String {
+ *     return a?.toString() ?: "-"
+ * }
+ *
+ * fun bar(a: Any?): String {
+ *     return "${a ?: "-"}"
+ * }
+ * </compliant>
+ */
+class NullableToStringCall(config: Config = Config.empty) : Rule(config) {
```
arturbosch, line 44: Please add an additional `@since 1.11.0` tag here

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullableToStringCallSpec.kt
```diff
@@ -0,0 +1,93 @@
...
+    describe("NullableToString rule") {
+        it("reports when a nullable toString is explicitly called") {
+            val code = """
+                fun test(a: Any?) {
+                    println(a.toString())
+                }
+            """
+            val actual = subject.compileAndLintWithContext(env, code)
+            Assertions.assertThat(actual).hasSize(1)
+            Assertions.assertThat(actual.first().message).isEqualTo("This call 'a.toString()' may return the string \"null\".")
```
arturbosch, line 25: Please use an import for `Assertions` here.

Issue #2904, opened 2020-07-29
UseDataClass does not report for classes that implement interfaces


zsmoore:
## Expected Behavior
UseDataClass should report for classes that implement interfaces without methods

## Observed Behavior
UseDataClass does not report for classes that implement interfaces

## Steps to Reproduce
```
interface Entity

data class(i: Int): Entity
```
Does not report

## Context
In our app we have a class called ViewData to represent simple data classes in Java before we started using Kotlin.
Shifting to Kotlin we would like to have these classes be data classes however detekt does not report on these since the classes implement an interface

As per https://kotlinlang.org/docs/reference/data-classes.html data classes are permitted to implement interfaces

cortinico:
Sound like a valid addition. Thank you very much for taking the time to working on it 🙏 

> UseDataClass should report for classes that implement interfaces without methods

Why only for interfaces without methods? In your PR seems like you're ignoring this specific condition. Was that intentional.

nit: your example should be fixed as the correct one should be:
```
interface Entity

class(i: Int): Entity
```

zsmoore:
Ah yes thanks I accidentally added the data to the example.

For my PR, looking at the existing code it should handle not considering classes that have methods inside them so allowing interface implementers to hit the existing code seemed to do the trick for me which is why I did not add any code to check what the interface was defined to do.

schalkms:
I don't think the user should use a data class here. The rule should flag simple cases, where a data class is obvious to use, but not every corner case. The rule was designed with simplicity in mind.

arturbosch:
> I don't think the user should use a data class here. The rule should flag simple cases, where a data class is obvious to use, but not every corner case. The rule was designed with simplicity in mind.

Maybe I was to fast merging the PR without reading the issue ... I try to be better!
I agree that we should not add too many exceptions as this rule should really just find simple cases like @schalkms mentioned.

Imo this change is still okay because the rule checks if any new functions appear in this class through interfaces we won't report it. I will add a check with delegating interfaces.

zsmoore:
Ahh I missed considering delegation in this one that is my bad.
I agree we should not flag delegation however, I would argue that keeping the rule for simple interfaces is a big benefit and not an edge case.

schalkms:
Good points were brought up here. Thank you for the productive discussion here! I think this rule works smooth now. 🙂

YashNagayach95:
Is there a way to ssupress a rule when my class is implementing a specific interface?

[...]

PR #2905, opened 2020-07-29
Fix UseDataClass to accept classes that implement interfaces

zsmoore:
Addresses issue #2904 

Root cause is that our doesNotExtendAnything() considers interfaces.

There is not a very easy way to fix this, the fix I have done to detect interfaces is 
- Look at the UserType
- Go to the containing KT File
- Find the declaration with the referenced name
- See if that declaration is a KTClass and an interface

From what I searched PSI does not give us many tools to get class from superTypeEntry

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
@@ -90,7 +90,7 @@ class UseDataClassSpec : Spek({
             it("does not report a candidate with an interface that has methods") {
                 val code = """
                     interface SomeInterface {
-                        fun foo()
+                        fun foo(): Int
                     }
                     
                     class NotDataClassBecauseItsImplementingInterfaceWithMethods(val i : Int): SomeInterface {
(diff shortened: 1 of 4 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
```diff
@@ -140,6 +154,27 @@ class UseDataClassSpec : Spek({
...
+            it("does report a candidate class with a simple interface extension") {
+                val code = """
+                    interface SimpleInterface
+                    class DataClass(val i: Int): SimpleInterface
+                """.trimIndent()
+
+                assertThat(subject.compileAndLint(code)).hasSize(1)
+            }
+
+            it("does report a candidate class with an interface extension that overrides vals") {
```
cortinico, line 167: Can I ask you to please write two other tests:

```kotlin
interface SimpleInterface {
    val i: Int
}

data class DataClass(override val i: Int): SimpleInterface
```

This should not report anything as you're correctly already using a data class.
And this other test:

```kotlin
interface SimpleInterface {
    val i: Int
}

open class BaseClass(open val j: Int)

class DataClass(override val i: Int, override val j: Int): SimpleInterface, BaseClass(j)
```

This should also report nothing as you're extending from another class (and an interface).

zsmoore: Added additional tests and looks like they pass.  Great test suggestions thanks!

zsmoore:
Also as a note added myself to the readme for this one + 
https://github.com/detekt/detekt/pull/2258
https://github.com/detekt/detekt/pull/2262
https://github.com/detekt/detekt/pull/2263

PR #2908, opened 2020-07-30
Skip delegating interfaces for UseDataClass

arturbosch:
Delegating to interfaces is a form of complexity, lets not allow it - #2904

### DIFF
--- a/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtClassOrObjectExtensions.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtClassOrObjectExtensions.kt
@@ -1,9 +1,0 @@
-package io.gitlab.arturbosch.detekt.rules
-
-import org.jetbrains.kotlin.psi.KtClass
-import org.jetbrains.kotlin.psi.KtDeclaration
-import org.jetbrains.kotlin.psi.psiUtil.isAbstract
-
-fun KtClass.isClosedForExtension() = !isAbstract() && !isOpen()
-
-fun KtClass.extractDeclarations(): List<KtDeclaration> = body?.declarations.orEmpty()
--- a/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtClassOrObjectExtensions.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtClassOrObjectExtensions.kt
@@ -8,7 +8,7 @@ import org.jetbrains.kotlin.psi.psiUtil.isAbstract
 fun KtClass.isClosedForExtension() = !isAbstract() && !isOpen()
 
 fun KtClass.onlyExtendsInterfaces(): Boolean {
-    return superTypeListEntries.all { it.isInterface() }
+    return superTypeListEntries.all { it.isInterface() && " by " !in it.text }
 }
 
 private fun KtSuperTypeListEntry.isInterface(): Boolean {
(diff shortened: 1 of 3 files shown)


zsmoore:
Thank you for adding this check!

detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtClassOrObjectExtensions.kt
```diff
@@ -2,18 +2,8 @@ package io.gitlab.arturbosch.detekt.rules
 
 import org.jetbrains.kotlin.psi.KtClass
 import org.jetbrains.kotlin.psi.KtDeclaration
-import org.jetbrains.kotlin.psi.KtSuperTypeListEntry
 import org.jetbrains.kotlin.psi.psiUtil.isAbstract
 
 fun KtClass.isClosedForExtension() = !isAbstract() && !isOpen()
```
zsmoore, line 7: It looks like both of these other extension functions are also only used in UseDataClass if we want to also move them there

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
```diff
@@ -126,6 +126,16 @@ class UseDataClassSpec : Spek({
...
             }
+
+            it("does not report a class with delegating interface") {
+                val code = """
+                    interface I
+                    class B() : I
+                    class A(val b: B) : I by b
+                """.trimIndent()
+
+                assertThat(subject.compileAndLint(code)).isEmpty()
```
zsmoore, line 137: is it possible in the testing code to feed multiple files vs a single one? I was under the impression that this would check interfaces defined outside of the current file but do not have a way to confirm that it looks like.

schalkms: What do you mean by that?
Can you please share how you want the API to look like?

zsmoore: Yeah definitely.  If you look at how the android-lint api works when we want to test our lint rules we do
```
lint().files( 
                java(FILE_TEXT_1), 
                java(FILE_TEXT_2), 
                java(FILE_TEXT_2), 
```
or replace the java call with Kotlin.

This allows us to check to make sure the rules work across multiple files and then check for errors in the whole set.

schalkms: Thanks for expressing your ideas. I like it. 👍 
We could definitely do this.
However, we decided to not use source code files for testing, because these test case files (`FILE_TEXT_*` in your case) are super rigorous to work with. We had a lot of test case files in this repository. Many of them were testing multiple scenarios. Hence, it was not so easy to debug the failing case.

reference #1089

With these inline test code snippets it's easy to see, what each unit test asserts.
Anyway, an API like the following is possible, but you could do the same with string concatenation.

```kotlin
subject.compileAndLint(code1).compileAndLint(code2)
```

> I was under the impression that this would check interfaces defined outside of the current file

`compileAndLint()` uses the Kotlin scripting API to assert if the given text is actually valid Kotlin code.

detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtClassOrObjectExtensions.kt
schalkms: thanks - done

Issue #2869, opened 2020-07-13
Document Console and Output Reports


schalkms:
This should be documented on the following web page and in the source code.
https://detekt.github.io/detekt/configurations.html#console-and-output-reports

Successor of #2833

sean-abbott:
using detekt 1.10 (` id("io.gitlab.arturbosch.detekt").version("1.10.0")`), and including the config stanza:
```
output-reports:
  active: true
  exclude:
  #  - 'HtmlOutputReport'
  #  - 'TxtOutputReport'
  #  - 'XmlOutputReport'
```

I get the following error:
```
$ ./gradlew detekt

> Task :jobs:detekt FAILED
Property 'output-reports' is misspelled or does not exist.

FAILURE: Build failed with an exception.
```

Also, the documentation shows this configuration is available, but the default config file linked from the documentation does not have it.

Finally, running without this stanza commented yields empty xml reports:
```
$ cat lib/build/reports/detekt/detekt.html
...
<h2>Complexity Report</h2>

<div>
  <ul>
    <li>224 lines of code (loc)</li>
    <li>169 source lines of code (sloc)</li>
    <li>130 logical lines of code (lloc)</li>
    <li>23 comment lines of code (cloc)</li>
    <li>32 cyclomatic complexity (mcc)</li>
    <li>9 cognitive complexity</li>
    <li>0 number of total code smells</li>
    <li>13% comment source ratio</li>
    <li>246 mcc per 1,000 lloc</li>
    <li>0 code smells per 1,000 lloc</li>
  </ul>
</div>

$ cat lib/build/reports/detekt/detekt.xml
<?xml version="1.0" encoding="utf-8"?>
<checkstyle version="4.3">
</checkstyle
```

schalkms:
@sean-abbott I moved your comment to a new issue.

sean-abbott:
Thank you, sorry for misfiling...I'm new to detekt so I wasn't sure if this was user error or not.

schalkms:
No worries..
This is issue here is rather a reminder to update the docs accordingly. May I ask you to submit your complete config in the other issue.

PR #2911, opened 2020-07-31
Document Output Reports code

schalkms:
as noted in #2869

### DIFF
--- a/detekt-report-txt/src/main/kotlin/io/github/detekt/report/txt/TxtOutputReport.kt
+++ b/detekt-report-txt/src/main/kotlin/io/github/detekt/report/txt/TxtOutputReport.kt
@@ -3,6 +3,10 @@ package io.github.detekt.report.txt
 import io.gitlab.arturbosch.detekt.api.Detektion
 import io.gitlab.arturbosch.detekt.api.OutputReport
 
+/**
+ * Contains rule violations in a plain text report similar to a log file.
+ * See: https://detekt.github.io/detekt/configurations.html#output-reports
+ */
 class TxtOutputReport : OutputReport() {
 
     override val ending: String = "txt"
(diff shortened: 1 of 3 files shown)


detekt-report-txt/src/main/kotlin/io/github/detekt/report/txt/TxtOutputReport.kt
```diff
@@ -3,6 +3,10 @@ package io.github.detekt.report.txt
 import io.gitlab.arturbosch.detekt.api.Detektion
 import io.gitlab.arturbosch.detekt.api.OutputReport
 
+/**
+ * Contains rule violations in a plain text report similar to a log file.
```
cortinico, line 7:
```suggestion
 * Contains rule violations and metrics in a plain text report similar to a log file.
```

detekt-report-xml/src/main/kotlin/io/github/detekt/report/xml/XmlOutputReport.kt
```diff
@@ -6,7 +6,8 @@ import io.gitlab.arturbosch.detekt.api.OutputReport
 import io.gitlab.arturbosch.detekt.api.Severity
 
 /**
- * Generates an XML report following the structure of a Checkstyle report.
+ * Contains rule violations in an XML format. The report follows the structure of a Checkstyle report.
```
cortinico, line 9:
```suggestion
 * Contains rule violations and metrics in an XML format. The report follows the structure of a Checkstyle report.
```

detekt-report-txt/src/main/kotlin/io/github/detekt/report/txt/TxtOutputReport.kt
schalkms: This report only contains rule violations and no metrics. 😉
That's the reason for documenting the reports now.

detekt-report-xml/src/main/kotlin/io/github/detekt/report/xml/XmlOutputReport.kt
schalkms: This report only contains rule violations and no metrics. 😉
That's the reason for documenting the reports now.

PR #2915, opened 2020-07-31
Add Github Action: Detekt All to readme

schalkms:


schalkms:
It’s really beyond to me why the test coverage increases by just adding a line in a readme file. 🧐

arturbosch:
> It’s really beyond to me why the test coverage increases by just adding a line in a readme file.

I've observed that the coverage report is from the last PR changes. Like 1 commit behind or something ..

PR #2920, opened 2020-08-01
Add informative message to UselessCallOnNotNull report

veyndan:


cortinico:
> I think adding some tests would be beneficial.

Yup, @veyndan you should be able to assert on the `findings.message` inside your tests

veyndan:
@schalkms @cortinico I expanded the previous tests to include verification of the error message. I'm not too sure why the code coverage tool is failing though…

cortinico:
> @schalkms @cortinico I expanded the previous tests to include verification of the error message. I'm not too sure why the code coverage tool is failing though…

The failure is related to a drop in codecov coverage for your diff. You can disregard this has it has been really flaky recently (also the overall coverage is increasing).

Issue #2902, opened 2020-07-27
The tests are slow


BraisGabin:
The tests take around 25 seconds to start running an example:

![Grabación de pantalla 2020-07-27 a las 22 24 51](https://user-images.githubusercontent.com/721244/88588883-8a0c9000-d058-11ea-8eed-78c650e479a9.gif)

It's not a complete record but you get the idea. Why does it take more than 20 seconds to test `:detekt-report-xml`?

cortinico:
> The tests take around 25 seconds to start running an example:

Same. I thought it was a misconfiguration of mine, but apparently not.
@arturbosch was doing profiling so maybe has more insights on this

schalkms:
The same holds true on my machine.

arturbosch:
I've noticed the same recently.
Even running `:detekt-report-xml:test` without any `detekt-test` dependencies/accidentally creating `KtCompiler` the test discovery is slow.

```
 g/r/detekt/detekt-report-xml{} $ gradle cleanTest test -x :detekt-gradle-plugin:test -Dorg.gradle.caching=false -Dorg.gradle.unsafe.watch-fs=false -Dorg.gradle.parallel=false       4s 776ms
```

arturbosch:
I've migrated the report-xml and report-txt module tests to junit jupiter test engine and the test discovery got reduced by 2 seconds (https://github.com/detekt/detekt/pull/new/debug-tests):

With `Spek`:

https://scans.gradle.com/s/mdxkbujmuzqw4/tests

![2020-07-30-112828_992x416_scrot](https://user-images.githubusercontent.com/20924106/88906587-e7226480-d257-11ea-8e66-391dc75b925f.png)

With Junit Jupiter: 

https://scans.gradle.com/s/udwjw43v5ravw/tests

![2020-07-30-112936_848x399_scrot](https://user-images.githubusercontent.com/20924106/88906691-09b47d80-d258-11ea-9934-2a463fc4ffb8.png)

arturbosch:
#2921 introduces many improvements.

The big penalty for us still is and will be: kotlin environment first inits (2s) + mockk first inits (2s) in many modules (multiplier) building in parallel (slowing down each other).

BraisGabin:
> #2921 introduces many improvements.
> 
> The big penalty for us still is and will be: kotlin environment first inits (2s) + mockk first inits (2s) in many modules (multiplier) building in parallel (slowing down each other).

[...]

PR #2921, opened 2020-08-01
Test performance improvements

arturbosch:
Findings:
- recent spek discovery improvements already help a bit (3s spek vs 1s jupiter -> 2s spek vs 1s jupiter for report-xml module)
- there are still some concurrent running things we don't understand
  - the more parallelization we use the more the test discovery rises
    - using a maximum of 4 workers help a lot
- using Gradle test fixtures improving the test discovery the most! **Edit: not true, they start instantly just in IntelliJ**
  - ~module report-txt without the `detekt-test` dependency starts *instantly*!!!~
  - ~does this mean spek tries to discover to much? still running through all jars/the whole classpath??~
  - ~have we setup the KtTestCompiler and ScriptEngine wrongly??~
     - ~both classes are only used in methods and Kotlin `object`s guarantees us singletons~
     - ~I've tried to remove the `object` declarations and using custom `lazy` inits -> no improvements~

References:
- Closes https://github.com/detekt/detekt/issues/2902
- https://github.com/spekframework/spek/pull/907
- https://github.com/spekframework/spek/pull/905


arturbosch:
Even with test fixtures the test discovery is noticeable when using `--parallel` (threads are starving)

2020-08-01-210022_444x411_scrot

schalkms:
There are 2 hard problems in computer science: threads, naming things, and off-by-1 errors.

arturbosch:
Did some profiling and spek is fast enough. Classgraph single-threaded works too.
I suspect 5 coroutine threads x #cores to be the reason (spek seems to be concurrent by default).
Junit Jupiter runs the tests sequentially (https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution).
Which makes sense when using build tool parallelization.

raniejade:
Hey! Out of curiosity, what does `detekt-test` do?

raniejade:
One thing weird I found is when running `detekt-core:test` as itself, discovery is fast around 5 seconds, but when you run all the tests together - discovery goes up to 15s!

raniejade:
Changing this made the discrepancy smaller from 5s (alone) -> 7s (all tests)

Screenshot from 2020-08-02 12-16-37

It was declared like this before:

```kotlin
        val finding = mockk<Finding>()
        every { finding.id }.returns("LongParameterList")
        every { finding.signature }.returns("Signature")

        val result = TestDetektion(finding)
```

So it was building the mock during discovery phase.

raniejade:
One improvement I could suggest is avoid doing heavy stuff in the group scopes, use a combination of fixtures (beforeEachTest, beforeEachGroup, etc ..) and memoized to delay calculation values used during the execution phase.

raniejade:
I've also made a change to `setupKotlinEnvironment` to delay calculation of the environment during the execution phase.

```kotlin
fun Root.setupKotlinEnvironment() {
    val wrapper by memoized(CachingMode.SCOPE, { createEnvironment() }, { it.dispose() } )

    @Suppress("UNUSED_VARIABLE") // name is used for delegation
    val env: KotlinCoreEnvironment by memoized(CachingMode.EACH_GROUP) { wrapper.env }
}

```

raniejade:
After a couple of testing from this branch, the biggest difference I can see is when making sure nothing heavy is calculated in the discovery phase. I've tried making cg scanning and execution sequential but the difference is very negligible 1-2 seconds.

[1] I made changes to your test (2m 9.625s): https://scans.gradle.com/s/hvit4fbjj54j6/tests?collapse-all

vs

[2] No changes (2m 37.258s): https://scans.gradle.com/s/k3cwnp74ersju/tests?collapse-all

From what I can see the individual test timings in Gradle's scan doesn't account for discovery time - but just execution time.

arturbosch:
@raniejade thank you very much for taking the time to help us out!
I've incorporated your suggestions.

I also removed the coroutine usage but the time between starting the test task and executing the tests is nearly the same (1-2s as you mentioned). 

So the only thing we can do is to spop introducing new modules which need the kotlin compiler or mockk :/.

raniejade:
> So the only thing we can do is to stop introducing new modules which need the kotlin compiler or mockk :/.

I think if we can just make sure that no heavy initialization happens in the discovery phase, will help improve discovery. There are probably other areas in your tests that are doing some initialization in the discovery phase, fixing them might help improve test speeds.

schalkms:
Awesome stuff!!

arturbosch:
> > So the only thing we can do is to stop introducing new modules which need the kotlin compiler or mockk :/.
> 
> I think if we can just make sure that no heavy initialization happens in the discovery phase, will help improve discovery. There are probably other areas in your tests that are doing some initialization in the discovery phase, fixing them might help improve test speeds.

Yes, it is definitely the way to move heavy setup code from discovery to the execution phase.
However in my build scans I couldn't see any performance improvements from this.
~20-30seconds fluctuations can be normal for a fresh and a hot gradle daemon.
Reducing the parallelization was the only factor which helped in incremental build time.

PR #2922, opened 2020-08-02
Roll back to non snapshot spek version

arturbosch:
Waiting for a new release - https://github.com/detekt/detekt/issues/2902


arturbosch:
I've rolled back to 2.0.12 of Spek to fix #2976 .
Build should be 10-20 seconds slower now.

PR #2929, opened 2020-08-05
Add LanguageVersionSettings and DataFlowValueFactory to BaseRule

t-kameyama:
We need to detect smart cast types in #2903.

In order to detect the smart cast types, `LanguageVersionSettings` and `DataFlowValueFactory` are required as follows:
https://github.com/JetBrains/kotlin/blob/040204e36e8d55395cdbbda2c8145dbb68bb0f00/idea/src/org/jetbrains/kotlin/idea/codeInsight/KotlinExpressionTypeProvider.kt#L116-L119

So we should add them to `BaseRule` like `BindingContext`.


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/BaseRule.kt
```diff
@@ -23,9 +27,16 @@ abstract class BaseRule(
      * receive the correct compile classpath for the code being analyzed otherwise the default value
      * BindingContext.EMPTY will be used and it will not be possible for detekt to resolve types or symbols.
```
schalkms, line 28: Maybe the documentation of the functions should be updated to comply with the new parameters and implementation.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/BaseRule.kt
```diff
@@ -23,9 +27,16 @@ abstract class BaseRule(
...
      */
-    fun visitFile(root: KtFile, bindingContext: BindingContext = BindingContext.EMPTY) {
+    fun visitFile(
+        root: KtFile,
+        bindingContext: BindingContext = BindingContext.EMPTY,
+        languageVersionSettings: LanguageVersionSettings? = null,
+        dataFlowValueFactory: DataFlowValueFactory? = null
```
schalkms, line 34: Maybe we should group the 3 objects concerning type and symbol solving into a new object. As a result the parameter list would be smaller.

arturbosch:
@t-kameyama thanks for the contribution and showing us a new way of writing more complex rules with smart casting.

I'm afraid to tell you that I have to decline this change. I really do not want to make our current rule api and design more complex. It is not the direction we want to go with the Rule and BaseRule classes. There are already things like the context, excludes and aliases which we regret to have added to it.

In #2680 we discuss further breaking api changes and list design improvements we want to make for 2.0.
Especially for the Rule api e.g. https://github.com/detekt/detekt/issues/2680#issuecomment-668825206 the information in this PR will be helpful and gonna make sure that the new Rule api will provide access to more constructs like the LanguageVersionSettings.

schalkms:
I don’t fully agree with declining this change.
The API could introduce a new object containing all objects required for writing more complex rules.

BraisGabin:
Or we can just allow extending the current rule api following its "spirit" and once we have the new api we can port it.

arturbosch:
Hm, convinced.
I like @schalkms suggestion using an object containing additional useful analysis classes.
We should leave `BindingContext` as a property though to not update every other rule needing the BindingContext.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/BaseRule.kt
arturbosch: Lets go with an object for the new two and leave the BindingContext extra so we do not need to update all rules?

t-kameyama:
> Are there also tests available?

#2903 has tests that requires the detection of smart casting.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/BaseRule.kt
t-kameyama: Introduced 'CompilerResources' class.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -29,11 +32,15 @@ internal class Analyzer(
...
         bindingContext: BindingContext = BindingContext.EMPTY
     ): Map<RuleSetId, List<Finding>> {
+        val languageVersionSettings = settings.environment.configuration.languageVersionSettings
+        @Suppress("DEPRECATION")
+        val dataFlowValueFactory = DataFlowValueFactoryImpl(languageVersionSettings)
+        val compilerResources = CompilerResources(languageVersionSettings, dataFlowValueFactory)
         val findingsPerFile: FindingsResult =
             if (settings.spec.executionSpec.parallelAnalysis) {
-                runAsync(ktFiles, bindingContext)
+                runAsync(ktFiles, bindingContext, compilerResources)
```
schalkms, line 41: Should **bindingContext** also be moved to **compilerResources**? @arturbosch

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/BaseRule.kt
arturbosch: lets better hide the information for now :)

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
arturbosch: Ah, yes on the core side we definitely should do this. On the rule side I would like them to be separated as we rey heavily on the BindingContext.
Lets do this in another PR.

arturbosch:
I've restarted the CI and will merge it then.

PR #2931, opened 2020-08-05
Upgrade to new dokka

arturbosch:
This checks if we finally can upgrade Gradle with the new dokka engine.
KtLint is not compatible with Kotlin 1.4-rc so I had to turn off `detekt` and `formatting:test` tasks


BraisGabin:
The regression in the Compiler and ktlint not working on kotlin 1.4 are known issues right? I mean, both projects have an issue to track it.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.6-rc-5-all.zip
```
cortinico, line 3:
```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-6.6-rc-6-all.zip
```

arturbosch:
> The regression in the Compiler and ktlint not working on kotlin 1.4 are known issues right? I mean, both projects have an issue to track it.

Yes, KtLint tests against dev versions as of recently. The compiler bug is actually not a bug but a fix, the receiver is actually not needed anymore.

arturbosch:
Being unable to upgrade the Kotlin version and having to turn off analysis because of a dependency, rises the question if we should move the formatting module outside the main repo again ...

arturbosch:
```
 * What went wrong:
    An exception occurred applying plugin request [id: 'org.jetbrains.kotlin.jvm']
    > Failed to apply plugin [id 'org.jetbrains.kotlin.jvm']
       > The current Gradle version 5.0 is not compatible with the Kotlin Gradle plugin. Please use Gradle 5.3 or newer, or the previous version of the Kotlin plugin.
```

We need to update to 5.3 too.

arturbosch:
dokka changes its sorting for 

https://github.com/detekt/detekt/blob/c6ff7112b0b33afdf62e043075d67508c9e1e5b9/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Context.kt#L13-L45

so our verifyDocumentation will always fail.

I've opened https://github.com/Kotlin/dokka/issues/1333 which is a major blocker for using the new dokka and therefore updating Gradle ...

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/MultiVersionTest.kt
```diff
@@ -29,9 +29,9 @@ object MultiVersionTest : Spek({
 
 private fun getGradleVersionsUnderTest() =
     if (getJdkVersion() < 13) {
-        listOf("5.0", "6.4.1")
+        listOf("5.3", "6.6")
```
BraisGabin, line 32: We should update our documentation too, right pointing the minimum version.

arturbosch:
I've extracted the Kotlin and Gradle upgrade to https://github.com/detekt/detekt/pull/2981.
This PR needs to wait until a stable dokka gets released.

arturbosch:
New version fixed our problem. Huge diff but works as expected.

BraisGabin (approved):

PR #2932, opened 2020-08-05
Flag listOfNotNull if all arguments are non-nullable

### LINKED ISSUE #2916: Flag listOfNotNull if all arguments are non-nullable
Based on the discussion from https://github.com/detekt/detekt/pull/2864#discussion_r453944852.

We could extend the rule `UselessCallOnNotNull` to also flag invocations of `listOfNotNull` when all the functions arguments are non-nullable.

## Noncompliant Code

```kotlin
listOfNotNull("string")
```

## Compliant Code

```kotlin
listOf("string")
```

veyndan:
Closes #2916.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
@@ -8,8 +8,10 @@ import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtQualifiedExpression
 import org.jetbrains.kotlin.psi.KtSafeQualifiedExpression
+import org.jetbrains.kotlin.psi.ValueArgument
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
 import org.jetbrains.kotlin.resolve.calls.callUtil.getType
@@ -27,12 +29,14 @@ import org.jetbrains.kotlin.types.isNullable
  * <noncompliant>
  * val testList = listOf("string").orEmpty()
  * val testList2 = listOf("string").orEmpty().map { _ }
+ * val testList3 = listOfNotNull("string")
  * val testString = ""?.isNullOrBlank()
  * </noncompliant>
  *
  * <compliant>
  * val testList = listOf("string")
  * val testList2 = listOf("string").map { }
+ * val testList3 = listOf("string")
  * val testString = ""?.isBlank()
  * </compliant>
  *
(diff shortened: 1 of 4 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
```diff
@@ -78,5 +82,22 @@ class UselessCallOnNotNull(config: Config = Config.empty) : Rule(config) {
...
 
+    @Suppress("ReturnCount")
+    override fun visitCallExpression(expression: KtCallExpression) {
+        super.visitCallExpression(expression)
+
+        val resolvedCall = expression.getResolvedCall(bindingContext) ?: return
+        val fqName = resolvedCall.resultingDescriptor.fqNameOrNull() ?: return
+        if (fqName != FqName("kotlin.collections.listOfNotNull")) return
+
+        val varargs = resolvedCall.valueArguments.entries.single().value.arguments
```
schalkms, line 93: This could crash here if entries is empty and thus has no entry.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNullSpec.kt
```diff
@@ -127,5 +127,32 @@ object UselessCallOnNotNullSpec : Spek({
             assertThat(findings).hasSize(1)
             assertThat(findings[0].message).isEqualTo("Remove redundant call to orEmpty")
         }
+
+        it("reports when calling listOfNotNull on all non-nullable arguments") {
+            val code = """
+                val strings = listOfNotNull("string")
```
schalkms, line 133: What happens if you just call **listOfNotNull()** without a parameter?

veyndan: Added a test to check for this in https://github.com/detekt/detekt/pull/2932/commits/b8a911fbb9404619bc34abc4fe2e9e66ac364b73.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
veyndan: I can't see a case when entries would ever be empty. The two function definitions of `listOfNotNull` are:

- `fun <T : Any> listOfNotNull(element: T?): List<T>`
- `fun <T : Any> listOfNotNull(vararg elements: T?): List<T>`

Both of these only have a single argument entry.

In the case of `listOfNotNull()`, `resolvedCall.valueArguments.entries.single()` will resolve to `vararg elements: T?`, and then `resolvedCall.valueArguments.entries.single().value.arguments` will be empty. This can be verified in the newly added test in https://github.com/detekt/detekt/commit/b8a911fbb9404619bc34abc4fe2e9e66ac364b73 as the test for `listOfNotNull` passes without any further changes to the source code.

schalkms: @veyndan do you have any ideas in which case **entries** could be empty?

schalkms: Thanks for the detailed explanation! Resolved..

veyndan: Nope, I think I correctly showed above that `entries` can never be null, but if you have any opposing views then I'd be happy to hear them 😄

schalkms: @veyndan you were too fast. I understood it and will merge it in a few seconds 😉

veyndan: Ah, I thought I was going crazy because I saw a different comment 😂  Awesome, thanks!

PR #2934, opened 2020-08-06
Use code tabs for groovy/kotlin dsl

### LINKED ISSUE #2846: Merge gradle groovy dsl and kotlin dsl documentation in the same page
https://detekt.github.io/detekt/groovydsl.html and https://detekt.github.io/detekt/kotlindsl.html are really similar documentation. The only difference that we should have here is the code snippets. But if you check there are more (little) differences. 

What do you think about using tabs [as gradle does in its documentation](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block)? I'm not sure if Jekyll will allow us to do that.

BraisGabin:
Close #2846 

Do we want Groovy as default DSL? or Kotlin?

There were some snippets that we doesn't have in both DSL so I added the tabs too but with a `// TODO`.

But we have a big stopper:
> GitHub Pages cannot build sites using unsupported plugins. If you want to use unsupported plugins, generate your site locally and then push your site's static files to GitHub.

Any idea to workaround this? Add a action to build and upload the web for each merge?


cortinico:
> Any idea to workaround this? Add a action to build and upload the web for each merge?

Yeah that shouldn't be that hard: https://github.com/marketplace/actions/deploy-to-github-pages
We can just publish the `docs/_site` folder

docs/pages/extensions.md
```diff
@@ -214,11 +214,20 @@ For example `detekt` itself provides a wrapper over [ktlint](https://github.com/
...
+{% codetab Groovy %}
 ```groovy
 dependencies {
     detektPlugins "io.gitlab.arturbosch.detekt:detekt-formatting:{{ site.detekt_version }}"
 }
 ```
+{% endcodetab %}
+{% codetab Kotlin %}
+```kotlin
+// TODO
```
cortinico, line 227:
```suggestion
dependencies {
    detektPlugins("io.gitlab.arturbosch.detekt:detekt-formatting:{{ site.detekt_version }}")
}
```

docs/pages/gettingstarted/gradle.md
```diff
@@ -180,14 +304,30 @@ tasks.withType(io.gitlab.arturbosch.detekt.Detekt) {
...
+{% endcodetab %}
+{% endcodetabs %}
 
 ##### <a name="customdetekttask">Defining custom detekt task</a>
 
 Custom tasks for alternative configurations or different source sets can be defined by creating a custom task that
 uses the type `Detekt`.
 
-###### Groovy DSL
+###### DSL
```
cortinico, line 327: This should probably go?

arturbosch:
> Thanks for tackling this issue.
> I think Kotlin should be the default.
> I’m afraid however, the mentioned showstopper is a big deal here.
> For now, we could list both code snippets underneath each other. Remember it’s just this single page. Moreover, just little effort is required to do this.

I agree with @schalkms here on having a simpler setup. If using the github action is simple enough -> okay :). If not lets write them beneath each other and have that automatically published.

BraisGabin:
Ok, I think that I have something:

Captura de pantalla 2020-08-10 a las 20 37 09

This is hosting the page locally using jekyll. The thing is that if I try to use other plugin (or a custom one) the build fails. But with this one it works... So, maybe, this plugin in enabled by github pages. I searched in the documentation but I couldn't confirm this. I don't know how to test it for real... So, what do you think about merging it? If it doesn't work we can always revert the commit from master. Any other idea?

PR #2954, opened 2020-08-09
Implement custom rule to check spek test discovery performance issues

arturbosch:
This PR introduces a new module `custom-checks` with for now one rule `SpekTestDiscovery`.
This rule finds variable declarations in spek tests where `memoized` would help to improve the performance.

All previous PR's improved the discovery time by ~200ms per module (500-1000ms in extreme cases like metrics and rules-style modules (parsing Kotlin files)).
In conclusion nearly every object we use to setup tests has to be considered as "heavy".
Exceptions are template strings and paths. I've excluded them from the rule.

Test frameworks like `Spek` which not only use reflection to find the tests but also need to instantiate a class and run the constructor closure are errorprone to such issues if one forgets to use `memoized` or similar concepts.

### DIFF
--- a/buildSrc/src/main/kotlin/detekt.gradle.kts
+++ b/buildSrc/src/main/kotlin/detekt.gradle.kts
@@ -41,6 +41,7 @@ allprojects {
 
     dependencies {
         detekt(project(":detekt-cli"))
+        detektPlugins(project(":custom-checks"))
         detektPlugins(project(":detekt-formatting"))
     }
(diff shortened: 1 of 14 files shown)


arturbosch:
Rule works and all our tests are free of expensive constructs.
The rule needs type resolution so it will only run on `detektTest` etc

I will setup a CI job with type resolution next. For this other rule issues need to be fixed.

Issue #2943, opened 2020-08-09
Suggest rule: Class cannot be publish.


VovaStelmashchuk:
I suggest add rule: Class cannot be publish.

I need the rule in the following case: Inside library and i want to sure all my class are internal or private, exclude API classes. We have this rule in custom company rule set. We want to contribute the rule to origin project, like an optional rule. Maybe someone need this rule too.

Are rule valid to be in the origin project?

We already discussed issue in KtLint https://github.com/pinterest/ktlint/issues/795
I migrate my project to Detekt at the moment.

arturbosch:
Hi @VovaStelmashchuk ,
this sounds like a really good rule for library authors, thanks for the suggestion.
You can take a look at the `LibraryCodeMustSpecifyReturnType` rule how we achieved to only run the rule if specific filters are active.
This is how we expect the users to configure it:
https://github.com/detekt/detekt/blob/66c9afeac2f5c2adf743b09d80b88ad3c1338301/config/detekt/detekt.yml#L119 

Looking forward to your PR!

PR #2959, opened 2020-08-10
Add new rule: LibraryEntitiesCannotBePublic

VovaStelmashchuk:
Close issue #2943


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/LibraryEntitiesCannotBePublicTest.kt
```diff
@@ -0,0 +1,53 @@
...
+            it("should not report a class") {
+                assertThat(subject.compileAndLint("""
+                    internal class A
+                """)).hasSize(0)
+            }
+
+            it("should not report a typealias") {
+                assertThat(subject.compileAndLint("""
+                    internal typealias A = List<String>
+                """)).hasSize(0)
```
arturbosch, line 48: nit: we always use `isEmpty` is such a case

VovaStelmashchuk: Fixed.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/LibraryEntitiesCannotBePublic.kt
```diff
@@ -0,0 +1,48 @@
...
+ * // code from a library
+ * class A
+ * </noncompliant>
+ *
+ * <compliant>
+ * // code from a library
+ * internal class A
+ * </compliant>
+ */
+class LibraryEntitiesCannotBePublic(ruleSetConfig: Config = Config.empty) : Rule(ruleSetConfig) {
```
arturbosch, line 22: LibraryEntitiesShouldNotBePublic?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/LibraryEntitiesCannotBePublic.kt
```diff
@@ -0,0 +1,48 @@
...
+ * <noncompliant>
+ * // code from a library
+ * class A
+ * </noncompliant>
+ *
+ * <compliant>
+ * // code from a library
+ * internal class A
+ * </compliant>
+ */
```
arturbosch, line 21: Please add a `@since 1.11.0` tag here.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/LibraryEntitiesCannotBePublic.kt
```diff
@@ -0,0 +1,48 @@
...
+class LibraryEntitiesCannotBePublic(ruleSetConfig: Config = Config.empty) : Rule(ruleSetConfig) {
+
+    override fun visitCondition(root: KtFile): Boolean = super.visitCondition(root) && filters != null
+
+    override val issue: Issue = Issue(
+        "ClassCannotBePublic",
```
arturbosch, line 27: Lets use `javaClass.simpleName` here to be consistent with the rule name

VovaStelmashchuk:
> Thanks for the contribution.
> What is your opinion on extending this rule for top level functions?

We don't use top level function in our case. But I think it great idea for detekt. Do you want to see this in the current PR?

arturbosch:
> > Thanks for the contribution.
> > What is your opinion on extending this rule for top level functions?
> 
> We don't use top level function in our case. But I think it great idea for detekt. Do you want to see this in the current PR?

Sure, should be just a small change :), thanks!

VovaStelmashchuk:
@arturbosch I added check for top level function. All checks are green now.

arturbosch:
> @arturbosch I added check for top level function. All checks are green now. I recommend to use squash for merge the PR.

Thanks! We squash every PR :).

PR #2960, opened 2020-08-10
Revert "Use code tabs for groovy/kotlin dsl (#2934)"

BraisGabin:
Ok... It doesn't work. We can't use code-tabs. Reverts #2934


schalkms:
I'd say we should stick with the simple solution. Having the Groovy and Koltin example code underneath each other is also fine IMHO.

PR #2967, opened 2020-08-13
Release 1.11.0

arturbosch:
WIP: sync to jcenter not ready

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -14,7 +14,7 @@ repositories {
 
 object Plugins {
     const val KOTLIN = "1.3.72"
-    const val DETEKT = "1.11.0-RC2"
+    const val DETEKT = "1.11.0"
     const val GITHUB_RELEASE = "2.2.12"
     const val ARTIFACTORY = "4.15.1"
     const val SHADOW = "5.2.0"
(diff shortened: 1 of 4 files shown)


arturbosch:
the jcenter sync somehow broke the release ... not all files of 1.11 were synced
https://jcenter.bintray.com/io/gitlab/arturbosch/detekt/detekt-core/1.11.0/
other modules even do not have a 1.11 folder.

maybe a 1.11.1 re-release is needed.

BraisGabin:
Why don't we just publish in maven central? It will break nothing because jcenter mirrors maven central.

At least for me maven central seems more truefull. It's a bit more difficult to first setup but then it just work and enforce you sign the jar and all those things.

arturbosch:
> Why don't we just publish in maven central? It will break nothing because jcenter mirrors maven central.
> 
> At least for me maven central seems more truefull. It's a bit more difficult to first setup but then it just work and enforce you sign the jar and all those things.

It is a convenience. Bintray/Jcenter is much faster/instantly downloadable. I can start asap on releasing sonar-kotlin or intellij plugin. It syncs to maven central.
Ofc I had to register at sonartype so theoretically we can start publishing directly to maven central.
However this is the first major issue with bintray for us for now. Lets forgive it this time :).

PR #2968, opened 2020-08-14
UnnecessaryLet: fix false positive when let is used for destructuring

### LINKED ISSUE #2966: UnnecessaryLet false positive when let is used for destructuring
## Expected Behavior

I don't expect `UnnecessaryLet` to be reported for the following code:

```
val y = x.let { (a, b) -> a + b }
```

That is because, in this case, `let` is necessary for destructuring. If I really, want to get rid of `let` I have to destructure in a `val (a, b) = x` assigment. And to not let `a` and `b` leak in the scope, I'd have to replace `val y = x.let { (a, b) -> a + b}` by `val y = run { val (a, b) = x ; a + b }`. Not really an improvement in my opinion.

## Observed Behavior

An `UnnecessaryLet` is reported for the code above.

## Steps to Reproduce

1. Enable `UnnecessaryLet` rule
2. Write:

```kotlin
fun foo(x: Pair<Int, Int>): Int = x.let { (a, b) -> a + b }
```

3. Run detekt

## Your Environment

* Version of detekt used: 1.11.0-RC2

t-kameyama:
Fixes #2966

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLet.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLet.kt
@@ -115,7 +115,6 @@ private fun PsiElement.countVarRefs(varName: String): Int =
 
 private fun KtLambdaExpression.countReferences(): Int {
     val bodyExpression = bodyExpression ?: return 0
-    val firstParameter = firstParameter
     val destructuringDeclaration = firstParameter?.destructuringDeclaration
     return destructuringDeclaration?.entries?.sumBy { bodyExpression.countVarRefs(it.nameAsSafeName.asString()) }
         ?: bodyExpression.countVarRefs(firstParameter?.nameAsSafeName?.asString() ?: IT_LITERAL)
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLet.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLet.kt
@@ -60,7 +60,7 @@ class UnnecessaryLet(config: Config) : Rule(config) {
                 report(CodeSmell(issue, Entity.from(expression), "let expression can be omitted"))
             }
         } else {
-            val lambdaReferenceCount = lambdaExpr.countReferences() ?: 0
+            val lambdaReferenceCount = lambdaExpr.countReferences()
             if (lambdaReferenceCount == 0 && !expression.receiverIsUsed(bindingContext) && isNullSafeOperator) {
                 report(
                     CodeSmell(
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLet.kt
```diff
@@ -110,6 +113,10 @@ private fun KtBlockExpression.hasOnlyOneStatement() = this.children.size == 1
 private fun PsiElement.countVarRefs(varName: String): Int =
     children.sumBy { it.countVarRefs(varName) + if (it.textMatches(varName) && it !is ValueArgument) 1 else 0 }
 
-private fun KtLambdaExpression.countReferences(): Int? {
-    return bodyExpression?.countVarRefs(firstParameter?.text ?: IT_LITERAL)
+private fun KtLambdaExpression.countReferences(): Int {
+    val bodyExpression = bodyExpression ?: return 0
+    val firstParameter = firstParameter
```
schalkms, line 118: I think this one can be removed safely.

schalkms: `val firstParameter = firstParameter`

schalkms:
👏

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLet.kt
```diff
@@ -110,6 +113,10 @@ private fun KtBlockExpression.hasOnlyOneStatement() = this.children.size == 1
...
+    val destructuringDeclaration = firstParameter?.destructuringDeclaration
+    return destructuringDeclaration?.entries?.sumBy { bodyExpression.countVarRefs(it.nameAsSafeName.asString()) }
```
cortinico, line 120: Can you divide this expression in 2? Having multiple occurrencies of the elvis operator on the same line makes the code harder to read.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLetSpec.kt
```diff
@@ -191,6 +191,96 @@ class UnnecessaryLetSpec : Spek({
...
+                
+                fun test(foo: Foo) {
+                    foo.let { (a, _) -> a + a }
+                }
+            """
+                val findings = subject.compileAndLint(content)
+                assertThat(findings).isEmpty()
+            }
+
+            it("does not report when destructuring declarations are used more than once 3") {
```
cortinico, line 220: Please don't use 2,3,... in test names but try to be a bit more specific.
Given that you have a `context("destructuring declarations")` on top you can probably simplify to 

```suggestion
            it("does not report when destructuring with types") {
```

and so on.

Issue #2938, opened 2020-08-07
False positive in UnnecessaryApply


BraisGabin:
## Expected Behavior
If we use the return value of `apply`, the `UnnecessaryApply` rule should never flag that usage.

## Observed Behavior
`UnnecessaryApply` flags some of those.

## Steps to Reproduce
```kotlin
val a = listOf(mutableListOf(""))
    .map { it.apply { add("") } }
```

## Context
Reported by zmunm in slack: https://kotlinlang.slack.com/archives/C88E12QH4/p1596599789239000

I really don't like this code because it goes agains immutability. But it's legit code so we should not flag it in this rule.

## Your Environment
* Version of detekt used: master, 1.11.0-RC1

zsmoore:
I'll jump into this one that we are also seeing this when doing a 

object = data?.let { data -> new Object().apply { setData(data) } }
swapped to also temporarily to avoid checkstyle error

BraisGabin:
Yes, I think this is the same case. You are using the return of `apply` as the return of `let`. And `UnnecessaryApply` doesn't check that.

cortinico:
> If we use the return value of `apply`, the `UnnecessaryApply` rule should never flag that usage.

Turns out we already have the logic to do this, but only if type resolution is active. I'm restricting `UnnecessaryApply` to run only when type resolution is active to avoid this false positive here #2963

PR #2991, opened 2020-08-20
Fix false positive for UnnecessaryLet with disabled type resolution

### LINKED ISSUE #2987: False positive in UnnecessaryLet
## Expected Behavior
`UnnecessaryLet` takes in account if the we use the return value of `let` or not.

## Observed Behavior
`UnnecessaryLet` flags some of those.

## Steps to Reproduce
```kotlin
val a = listOf<List<String>?>(listOf(""))
    .map { list -> list?.let { it + it } }
```

## Context
This is the same problem that #2938 but with `UnnecessaryLet` instead of `UnnecessaryApply`

## Your Environment
* Version of detekt used: 1.11.1

cortinico:
The fix is the same we applied for https://github.com/detekt/detekt/issues/2938 (restricting the rule to type resolution only).

Fixes #2987


BraisGabin (commented):
Could you add this example to the tests?

```kotlin
val a = listOf<List<String>?>(listOf(""))
    .map { list -> list?.let { it + it } }
```

To ensure that in the future we just remove the if thinking "it's not necessary"

schalkms (approved):

arturbosch (approved):

BraisGabin (approved):

PR #2992, opened 2020-08-20
Fix support for Gradle configuration cache

3flex:
https://github.com/detekt/detekt/pull/2813 missed one usage of `Project` at execution time.

This PR:
* Makes `Detekt` task compatible with the configuration cache
* Adds tests
* Bumps Gradle to 6.6 to support the new tests


detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/ConfigurationCacheTest.kt
```diff
@@ -0,0 +1,61 @@
...
+                .dryRun()
+                .build()
+
+            // First run primes the cache
+            gradleRunner.runTasks("--configuration-cache", "detektGenerateConfig")
+
+            // Second run reuses the cache
+            val result = gradleRunner.runTasks("--configuration-cache", "detektGenerateConfig")
+
+            assertThat(result.output).contains("Reusing configuration cache.")
```
schalkms, line 58: Maybe we can refactor those 3 usages into 1 method in order to avoid code duplication.

arturbosch:
Oh, I started to do the same in https://github.com/detekt/detekt/pull/2944
But your tests look great :).
You will need to rebase this when the 1.4 branch is merged to have Gradle 6.6 and a running dokka on Java 11

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/ConfigurationCacheTest.kt
3flex: I could, but my preference is to keep tests self-contained, showing all steps within the test and not in extra methods. I think with just the three tests the duplication is OK in this case.

PR #2993, opened 2020-08-20
Introduce @requiresTypeResolution to KDoc for rules

cortinico:
Given that we received several reports of false negatives for users with type resolution disabled, I believe we should document this better.

I'm adding a `@requiresTypeResolution` KDoc annotation to rules that are strictly requiring Type and Symbol solving to execute. This tag will propagate also to the Markdown documentation.

### DIFF
--- a/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/RuleSetPagePrinter.kt
+++ b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/RuleSetPagePrinter.kt
@@ -42,7 +42,7 @@ object RuleSetPagePrinter : DocumentationPrinter<RuleSetPage> {
 
             if (rule.requiresTypeResolution) {
                 paragraph {
-                    bold { "Requires Type and Symbol solving" }
+                    bold { "Requires Type and Symbol Solving" }
                 }
             }
 
--- a/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/RuleSetPagePrinter.kt
+++ b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/RuleSetPagePrinter.kt
@@ -40,6 +40,12 @@ object RuleSetPagePrinter : DocumentationPrinter<RuleSetPage> {
                 paragraph { "TODO: Specify description" }
             }
 
+            if (rule.requiresTypeResolution) {
+                paragraph {
+                    bold { "Requires Type and Symbol solving" }
+                }
+            }
+
             if (rule.severity.isNotEmpty()) {
                 paragraph {
                     "${bold { "Severity" }}: ${rule.severity}"
--- a/detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/util/RuleSetPageCreator.kt
+++ b/detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/util/RuleSetPageCreator.kt
@@ -12,13 +12,46 @@ internal fun createRuleSetPage(): RuleSetPage {
 }
 
 internal fun createRules(): List<Rule> {
-    val rule1 = Rule("WildcardImport", "a wildcard import", "import foo.*", "import foo.bar", true, "Defect",
-            "10min", "alias1, alias2", "", listOf(
-                    Configuration("conf1", "a config option", "foo", null),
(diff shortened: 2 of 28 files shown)


arturbosch (approved):
Nice one!

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/util/RuleSetPageCreator.kt
```diff
@@ -12,13 +12,46 @@ internal fun createRuleSetPage(): RuleSetPage {
...
+        nonCompliantCodeExample = "fun stuff(): Unit {}",
+        compliantCodeExample = "fun stuff() {}",
+        active = true,
+        severity = "",
+        debt = "5m",
+        aliases = null,
+        parent = "",
+        configuration = emptyList(),
+        autoCorrect = true,
+        requiresTypeResolution = true
```
schalkms, line 54: I think there’s also a test to adapt.

docs/pages/documentation/style.md
```diff
@@ -882,6 +886,8 @@ This rule detects the usage of the Pair constructor to create pairs of values.
 
 Using <value1> to <value2> is preferred.
 
+**Requires Type and Symbol solving**
```
schalkms, line 889: Should **Solving** be written this way?

schalkms (approved):
Good idea, I like it!

docs/pages/documentation/style.md
BraisGabin: I always wonder why, in English, all the "things" are capitalized. In Spanish is completely different. I know, there is not an actual reason, it's just like that. But it's curious :)

BraisGabin (approved):
👏🏻👏🏻👏🏻 Great addition!

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/util/RuleSetPageCreator.kt
cortinico: It's already included. The [RuleSet.md](https://github.com/detekt/detekt/pull/2993/files#diff-45d56ba9b636fe6243083be500ea29a1) file is used to verify that the output of this class is consistent

docs/pages/documentation/style.md
cortinico: Yup you're right. Let' update it to [title case](https://en.wikipedia.org/wiki/Letter_case#Title_case) 👍

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/RuleSetPagePrinter.kt
```diff
@@ -40,6 +40,12 @@ object RuleSetPagePrinter : DocumentationPrinter<RuleSetPage> {
                 paragraph { "TODO: Specify description" }
             }
 
+            if (rule.requiresTypeResolution) {
+                paragraph {
+                    bold { "Requires Type and Symbol Solving" }
```
realdadfish, line 45: Nit: We have this spelled as "Type resolution" at https://detekt.github.io/detekt/gradle.html#using-type-resolution and as "Type and Symbol solving" - albeit with no contents (!) - here: https://detekt.github.io/detekt/type-and-symbol-solving.html

Should we use one term and eventually fill the latter page with some contents (or alternatively copy things over from the first link)?

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/RedundantSuspendModifier.kt
```diff
@@ -49,8 +49,9 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
  *   println("string")
  * }
  * </compliant>
+ *
+ * @requiresTypeResolution
```
realdadfish, line 53: Nit: Should we add another tag to denote that a specific rule's implementation is _enhanced_ or _changed_ by type resolution? I think of `SpreadOperator` and a few other rules. I remember to have seen issues where a rule worked perfectly fine w/o type resolution and failed in specific situations with type resolution.

BraisGabin: > a rule worked perfectly fine

What do you mean with that? Fine as no issues detected or fine as no false postives/negatives? I mean, any rule, with type solving should work better than without. If that's not the case, it's a bug that we should fix.

PR #3001, opened 2020-08-21
Workaround config

BraisGabin:
Right now we are using `FailFastConfig` and `DisabledAutoCorrectConfig` to modify the configuration of all detekt and not just the configuration related with the rules itself.

This is kind of complex because we have all those configurations mixed. We should think about this for 2.0.

I need this refactor to implement `--all-rules`. Because I want to enable all the rules, but I don't want to enable any PostProcessor or Report. And all of them checks the same key: `"active"`.

I really don't like too much this code but I don't want to change any current behaviour.

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
@@ -33,7 +33,7 @@ internal interface Lifecycle {
 
     fun analyze(): Detektion {
         measure(Phase.ValidateConfig) { checkConfiguration(settings) }
-        val filesToAnalyze = measure(Phase.Parsing) { parsingStrategy.invoke(spec, settings) }
+        val filesToAnalyze = measure(Phase.Parsing) { parsingStrategy.invoke(settings) }
         val bindingContext = measure(Phase.Binding) { bindingProvider.invoke(filesToAnalyze) }
         val (processors, ruleSets) = measure(Phase.LoadingExtensions) {
             processorsProvider.invoke() to ruleSetsProvider.invoke()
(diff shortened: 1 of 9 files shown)


arturbosch (commented):
I like this!

I was thinking if this location is the only one which needs to get wrapped for fail fast and co (instead of creating a new settings objects which potentially can lead to opening two threadpool/classloaders/environments)?

https://github.com/detekt/detekt/blob/7328075ae6adbb064637c5eb417e5a6f5d750c06/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt#L27

BraisGabin:
I think that we don't need to wrap it any where else.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
```diff
@@ -57,13 +60,36 @@ internal interface Lifecycle {
...
         { generateBindingContext(settings.environment, settings.classpath, it) },
     override val processorsProvider: () -> List<FileProcessListener> =
         { FileProcessorLocator(settings).load() },
     override val ruleSetsProvider: () -> List<RuleSetProvider> =
-        { spec.rulesSpec.runPolicy.createRuleProviders(settings) }
+        { settings.createRuleProviders() }
 ) : Lifecycle
+
+private fun workaroundConfiguration(settings: ProcessingSettings): ProcessingSettings {
+    return ProcessingSettings(settings.spec, settings.spec.workaroundConfiguration(settings.config))
```
arturbosch, line 74: I mean that `settings.spec.workaroundConfiguration(settings.config)` could just be in `Analyzer` and we do not need to create a new `ProcessingSettings` at all. Or do I miss something?

arturbosch: Replacing this line

https://github.com/detekt/detekt/blob/7328075ae6adbb064637c5eb417e5a6f5d750c06/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt#L27

to `private val config: Config = settings.spec.workaroundConfiguration(settings.config)`

BraisGabin: Oh! Yes, you are completely right. It have more sense to do it in the `Analyzer`. Done.

arturbosch (approved):

Issue #2647, opened 2020-04-30
Merge (rather than replace) multiple configs for ForbiddenImport rule


kkoser:
## Expected Behavior of the rule
Currently Detekt supports specifying multiple config files, which will be merged together at runtime with priority in reverse of the order they are specified. If a later file defines a given rule, that definition will supersede any previous definitions of that rule. In general, this is a great feature. However, it does not work well with ForbiddenImports - the reason being that if you wanted to build upon the previous forbidden imports, it will overwrite those previous rules and they will not longer be applied. It would be great if instead of overwriting, it merged the config for ForbiddenImports, so that you keep any previously defined disallowed imports along with ones in the newer config files. I realize this is a behavior change, so wanted to start a discussion around it!

For example, if in one config I have:
```
ForbiddenImport:
    active: true
    imports:
      - 'android.util.Log'
```

and in another I have:
```
ForbiddenImport:
    active: true
    imports:
      - 'android.util.ArrayMap'
```

and I apply both of these config files, only one of those imports would be disallowed, not both.

## Context
We want to create compounding ForbiddenImport rules for different pieces of code - basically a standard set of rules that are best practices, and then more detailed rules applied to subsets of code (usually in separate gradle modules).

chao2zhang:
I am more inclined to extend the ForbiddenImport rule to accept a list of `Pair(val fileNamePattern:String, val forbiddenPattern: String`. rather than a single `forbiddenPatterns`.

schalkms:
Thanks for the feedback and the great ideas!

I think the inheritance mode could lead to surprising behavior. What happens if you declare conflicting configuration options? This inheritance model can only work for yaml lists.

BraisGabin:
I agree that it should be handy but I can imagine use cases for the other way around. I mean, I have an import that I don't allow in general but I want to allow it in one particular module...

BraisGabin:
> I am more inclined to extend the ForbiddenImport rule to accept a list of `Pair(val fileNamePattern:String, val forbiddenPattern: String`. rather than a single `forbiddenPatterns`.

I think that this is a different feature request so you can open a new issue to track that one.

kkoser:
Yea that's a good point - this would likely also introduce the concept of exclusions along with inclusions (similar to what Java checkstyle allows: https://checkstyle.sourceforge.io/config_imports.html#ImportControl). 

[...]

Issue #2267, opened 2020-01-19
Align cli flags and defaults with other analysis tools


arturbosch:
## Expected Behavior
`--build-upon-default-config` should be `true` by default.
new flag `--all-rules` which activates all rules
remove `--fail-fast`
introduce new flag `--max-issues` to quickly manipulate the number of allowed issues without having a config file.

## Current Behavior
`--fail-fast` does three things: `--build-upon-default-config` + `maxIssues=0` + activates all rules

## Context
Align behavior wiht tools like PMD and ESLint

BraisGabin:
Are we sure that `--fail-fast` always sets the `maxIssues` to `0`? The code doesn't do that:
https://github.com/detekt/detekt/blob/c1ebc037f9f3bb161f39157f2d71a964ee24527c/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/FailFastConfig.kt#L15
and even the tests doesn't assert that:
https://github.com/detekt/detekt/blob/c1ebc037f9f3bb161f39157f2d71a964ee24527c/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/WorkaroundConfigurationKtSpec.kt#L68

I vote to remove the support to `--fail-fast`. It's buggy and no one complains about it. And it really work in really special situations because nearly always we end using the defaul configuration as the base configuration and because of this line:
https://github.com/detekt/detekt/blob/c1ebc037f9f3bb161f39157f2d71a964ee24527c/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/FailFastConfig.kt#L14

We just run the rules that are active by default.

arturbosch:
Your right, it set the default `maxIssues = 0` when the old threshold was 10, configs can overwrite it.
Now it's just `--build-upon-default-config + activates all rules`.

@BraisGabin have you started with this issue? If not, I want to look at this next.

BraisGabin:
You can do it. All my work is the wip pr that I want to close. I'll try to focus on the dsl.

arturbosch:
> You can do it. All my work is the wip pr that I want to close. I'll try to focus on the dsl.

Ah, rediscovered your PR. A good starting point.

chao2zhang:
The remaining tasks are:
- Enable `--build-upon-default-config` by default
- Remove `--fail-fast`

BraisGabin:
Probably this two should be done for 2.0, right?

[...]

Issue #1528, opened 2019-03-05
failFast configuration


vanniktech:
What's the conclusion?

I'd like to avoid having to maintain different options in my Gradle Plugin.

Is only allowing it to be configured via the CLI the way to go now? Personally, I liked that I was able to configure this via the YML file.

arturbosch:
Personally I didn't liked it in the config file. For example `failFast` was rather cryptic.

I think the cli is the right place for such boolean based flags next to existing `--disable-default-rulesets`, `--debug`, `--parallel`, '--create-baseline`.

You can create tasks which are independent of any configuration file (detektFailFast with `failFast` gradle property or detektAutoFormat with `autoCorrect` and `disableDefaultRuleSets` gradle properties).

Stuff like test-pattern and reports should be in the config file because they are highly configurable (partly custom reports can only be configured through the config when using the gradle plugin).

I am for migrating the top level autoCorrect property to the cli and leave these flags in the cli.

schalkms:
I agree with Artur here. In my opinion this should also be a command line option.
I also think we should be better at communicating the change to our users.
With version 1 coming closer I think that breaking changes should be avoided if they are not really necessary.

vanniktech:
Alright. Thanks :)

mikezx6r:
I'd still like an easy way to have all rules turned on, and then the remainder of the config file adjusts/disables those I'm not interested in. Is there currently an option like that in the config file?

I believe if I have an empty config file, then I am using the default, correct? But the default doesn't have all rules turned on?

I believe that's what failFast does, but I want that to be the default for all my projects w/o having to define all the rules in a file.

Even if it was some form of option like all-rules-enabled, I'd like it to be in the config so it doesn't matter how detekt is run, I get the same result.

arturbosch:
@mikezx6r 
Have you tried using `--fail-fast` from the cli or `failFast` in the `detekt gradle extension`?
```
detekt {
     ...
    failFast = true
    ....
} 
Its the same as the removed failFast property in the yaml file.

mikezx6r:
As soon as failFast was deprecated in config, I removed it and added the failFast in Gradle. 

Please correct me if I'm wrong. If I have some exclusions and testPatterns in my config but no failFast in the config file and run Detekt without setting the failFast flag, what rules are used?

I believe the answer is whatever is defined in the default config, correct?

[...]

PR #3002, opened 2020-08-22
[RFC] Config DSL

BraisGabin:
First ideas about how to implement our own config DSL.

```kotlin
rules {
    ruleSet("comments") {
        active = false
        undocumentedPublicClass {
            active = true
            searchInInnerClass = false
        }
    }
}
```

I didn't implement all the things that we want to cover yet. But I'll like to know what do you think about this DSL and the way it's implemented (this is my first real DSL)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/dsl/Config.kts
```diff
@@ -0,0 +1,11 @@
+package io.gitlab.arturbosch.detekt.api.dsl
+
+rules {
+    ruleSet("comments") {
```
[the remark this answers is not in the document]
arturbosch: What about something like `+LongMethod(prop1, prop2 ....)`
Would this be possible with default imports to `import io.github.detekt.rules.stye.*` etc?
A user could add his own imports with kotlin-scipting annotations and we get that dependency added to our engine for free?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/dsl/Dsl.kt
```diff
@@ -0,0 +1,88 @@
...
+    internal val list: MutableList<RuleSet> = mutableListOf()
+
+    fun ruleSet(name: String, builder: RuleSetBuilder.() -> Unit) {
+        list.add(RuleSetBuilder().apply(builder).build(name))
+    }
+}
+
+@ConfigDsl
+class RuleSetBuilder(
+    var active: Boolean = false,
```
arturbosch, line 29: Does the dsl need `active`? Or would it be the same as commenting out the line?
Everything not inside the dsl could not `active = false`.

We could layer the configuration dsl and allow to extend other configs:
```kt
extend(DefaultConfig)

rules {
  ... // declare only new rules (and overwrites?)
}
```

arturbosch:
I think we should write the rule builder uppercase like `UndocumentedPublicClass` to match the rule class name. e.g. generate the dsl based on the rule classes like @schalkms said.

schalkms:
Just to add to the comments, I want to express the feeling that the DSL should be focused on simplicity at first.
After that, feedback can be gathered. What do detekt users think about the DSL. Users of detekt have good knowledge of detekt, so they can most likely overcome the one or the other comfort function.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/dsl/Dsl.kt
BraisGabin: 🤔 if the user adds `extend(DefaultConfig)` how could the user disable the default rules if we don't have the `active`? I would like to remove the `active` too but I think that we need it.

arturbosch: hm, rules have an unique id so redeclaring them in a custom config would overwrite them like we do in yaml?

arturbosch:
> First ideas about how to implement our own config DSL.
> 
> ```kotlin
> rules {
>     ruleSet("comments") {
>         active = false
>         undocumentedPublicClass {
>             active = true
>             searchInInnerClass = false
>         }
>     }
> }
> ```
> 
> I didn't implement all the things that we want to cover yet. But I'll like to know what do you think about this DSL and the way it's implemented (this is my first real DSL)

Just an idea if we should go with different name for `ruleSet`:

```kt
rules {
  groupTo("comments") {
    undocumentedPublicClass {
      searchInInnerClass = false
    }
  }
}
```

Some thoughts about further dsl features:

```kt
rules {
  groupTo("comments") {
    val rule = +UndocumentedPublicClass {
      searchInInnerClass = false
      aliases = setOf(...)
      includes = setOf(...)
      excludes = setOf(...)
    }

    +rule.copy("api-UndocumentedPublicClass") {
       includes = setOf("**/api/*.kt")
     }
  }
}
```

Note: aliases, includes, excludes etc should be cached in map in core.

BraisGabin:
I have no idea about what the prefix `+` does in the DSL. Could you link some documentation about it? And, probably, for that reason I don't get what do you mean with: `+rule.copy("api-UndocumentedPublicClass")` What should that do?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/dsl/Dsl.kt
BraisGabin: > hm, rules have an unique id so redeclaring them in a custom config would overwrite them like we do in yaml?

We could do that. My idea (not implemented yet) is that, if you have two different configurations we run the first and then if you redefine one rule we provide the previous instance so you can decide if you want to overwrite, merge (as requested in #2647), or do whatever you want.

This way you can merge like this: `imports = this.imports + listOf("foo.bar")`.

But, about the `active`, how would a rule that was added in a previous configuration file be disabled if we don't have the `active` flag?

arturbosch:
> I have no idea about what the prefix `+` does in the DSL. Could you link some documentation about it? And, probably, for that reason I don't get what do you mean with: `+rule.copy("api-UndocumentedPublicClass")` What should that do?

Sorry for the confusion!
In a Kotlin you can use nested extension operator functions to come up with a "groovy" dsl:
```kt
class GroupBuilder {
    operator fun String.unaryPlus() {
        println("")
    }
}

inline fun ProcessingSpecBuilder.group(init: GroupBuilder.() -> Unit) {
    GroupBuilder().apply(init)
}

fun main() {
    ProcessingSpec {
        group {
            +"Hello World"
        }
    }
}
```

The same way one could write a dsl where `+Rule { ...init... }` would define a rule and add it inside the container/builder.

BraisGabin:
Oh, ok, so your idea is to use `+UndocumentedPublicApi` instead of `UndocumentedPublicApi` so we don't need to implement:

```
fun RuleSetBuilder.undocumentedPublicClass(init: UndocumentedPublicClassBuilder.() -> Unit) {
    addRule(UndocumentedPublicClassBuilder().apply(init))
}
```

For each rule, right? I think that, for the user perspective, the `+` seems a bit hacky. And I don't know how well it will work with autocompletation. If you know that all the rules are configured using an extension function you can always type `this.` and ask for autocompletation and find it there.

And I still don't get what this code would do:
```kotlin
+rule.copy("api-UndocumentedPublicClass")
```

arturbosch:
> Oh, ok, so your idea is to use `+UndocumentedPublicApi` instead of `UndocumentedPublicApi` so we don't need to implement:
> 
> ```
> fun RuleSetBuilder.undocumentedPublicClass(init: UndocumentedPublicClassBuilder.() -> Unit) {
>     addRule(UndocumentedPublicClassBuilder().apply(init))
> }
> ```
> 
> For each rule, right? I think that, for the user perspective, the `+` seems a bit hacky. And I don't know how well it will work with autocompletation. If you know that all the rules are configured using an extension function you can always type `this.` and ask for autocompletation and find it there.

True, that. Hm I expect the user to copy paste a "default config dsl" first and modify it like they do with the yaml xD. I have bad experiences with the Kotlin scripting part. We will see how it will improve. It would be cool if the user can drop his
detekt extension as an import and get autocompletion for free in idea ^^.

> 
> And I still don't get what this code would do:
> 
> ```kotlin
> +rule.copy("api-UndocumentedPublicClass")
> ```

In 1.X we restrict ourselves by having just one configured rule of a kind and this rule lives in a specific rule set.
We could get rid of the rule set concept and allow flexibility in creating rule groups and mix and match the rules in them.

```kt
rules {
  group("api") {
    ... configure rules to run over the public api
  }

  group("CI") {
    // configure costly rules for CI only
  }

  group("hook") {
    // fast commit hook setup
  }
}
```

It would make sense to reference one rule instance in different groups/ruleSets and maybe `Rule.copy` to re-configure some thresholds.

Hope the abpve makes sense ^^

BraisGabin:
Man, this is going to be a HUGE change.

Ok. Now I moved some code to it's correct "place" and I created how I think that the Analyzer should work.

It now recives a `List<RuleConfiguration>`. `RuleConfiguration` knows instantiate the rule and know the include/exclude information so it can decide if the rule should be executed or not. In this list we don't have any disabled rule, they where filtered before (not implemented).

I realised that we should keep some track of the `RuleSet` names because we will need them for the reports (html report, for example).

3flex:
I'm late to the conversation, but if the old configuration method is being thrown out, this is a good opportunity to design the way configuration works from the ground up. From what I understand the main driver of this change is to get auto-complete in the IDE? This could also be achieved with JSON schema for YAML which can be achieved with zero config in IntelliJ. I can open another issue for that.

There are so many other possible areas of improvement for the config that I would want to see before a DSL.

I played with Ruby in the past and the Rubocop linter is extremely popular, with a very similar feature set to detekt. Looking at the [configuration docs](https://docs.rubocop.org/rubocop/configuration.html) for that I see so many benefits of the approach they took (using YAML!) that solve some problems raised by detekt users.

Features I'd want from the new configurator (some already mentioned above in this thread) based on inspiration from Rubocop:
* [Including/excluding](https://docs.rubocop.org/rubocop/configuration.html#includingexcluding-files) files for all rules (#2581, https://github.com/detekt/detekt-compiler-plugin/issues/10)
* [Enable or disable all rules by default](https://docs.rubocop.org/rubocop/configuration.html#enabled) (https://github.com/detekt/detekt/issues/1528#issuecomment-470327188)
* [Explicit inheritance rules](https://docs.rubocop.org/rubocop/configuration.html#inheriting-from-another-configuration-file-in-the-project). This makes things like multi-project configs in Gradle very simple - each project can have its own config which inherits from a root config, and this can be set within the config file and doesn't require any CLI flags or Gradle config (though that could also be implemented, as long as the order of inheritance were clearly defined)
* All config would always inherit from detekt's default config. Where this is not desired, user can just disable all rules by default within the config file then build their config from there (https://github.com/detekt/detekt/issues/2267#issue-551890033)
* [Merging of hashes](https://docs.rubocop.org/rubocop/configuration.html#inheritance-of-hashes-vs-other-types) when configs are inherited, and user's [choice of merging or overriding arrays](https://docs.rubocop.org/rubocop/configuration.html#merging-arrays-using-inherit_mode) when configs inherited (https://github.com/detekt/detekt/issues/2647)
* Walking the file tree to find a config file, and also checking user's home directories for a config in case there is no project-specific config
* CurrentIssues/ManuallySuppressedIssues file is replaced with [just another config that's inherited from](https://docs.rubocop.org/rubocop/configuration.html#automatically-generated-configuration)
* Choose a [custom severity for an issue](https://docs.rubocop.org/rubocop/configuration.html#severity) - not all issues are as important to all users (https://github.com/detekt/detekt/issues/533#issuecomment-343896803)
* Setting compatibility for rules [based on Kotlin version](https://docs.rubocop.org/rubocop/configuration.html#setting-the-target-ruby-version) - e.g. if project targets Kotlin 1.3, a rule to add trailing commas for parameter lists etc. should not be suggested since it requires 1.4.

Not everything Rubocop does with its config can be literally copied for detekt as there are some differences in behaviour to consider for Java/Kotlin projects and Gradle, but majority of its config functionality could be implemented in detekt.

I've always found detekt's config setup generally confusing. I think this comes from incomplete documentation as well as config behaviour that's been built iteratively over time to fix individual issues or pain points without considering the overall picture. Detekt 2.0 is a good opportunity IMHO to rethink some of this and come up with something really robust that shouldn't need so much change over time.

The DSL is cool though :)

schalkms:
Cool, thanks for your input @3flex.
Let me elaborate on your comments.

> From what I understand the main driver of this change is to get auto-complete in the IDE?

It's one of the drivers, but not the main one.
Configuration mistakes should be caught as early as possible in my opinion. Assistance either via syntax checking or semantic checking during compile comes in handy in that regard.

> This could also be achieved with JSON schema for YAML which can be achieved with zero config in IntelliJ.

This would be a low hanging fruit. This already improves the main issues when editing the configuration file. 

> I've always found detekt's config setup generally confusing. I think this comes from incomplete documentation as well as config behaviour that's been built iteratively over time to fix individual issues or pain points without considering the overall picture. Detekt 2.0 is a good opportunity IMHO to rethink some of this and come up with something really robust that shouldn't need so much change over time.

Completely agree.

schalkms:
> Features I'd want from the new configurator (some already mentioned above in this thread) based on inspiration from Rubocop:

I think we should bring those great ideas into individual issues. Apart from that, I think a priority comes in handy to tackle this feature requests.

Regardless of that, we should decide on how to proceed with the DSL development.
Maybe we should postpone this for now to detekt v3.0 and further development this on a separate feature branch or repo. Consequently, commits from the main branch need to be merged to this feature branch.

PR #3014, opened 2020-08-25
Run Gradle Android tests conditionally when sdk path is defined

arturbosch:
#2787

### DIFF
--- a/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPluginTest.kt
+++ b/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPluginTest.kt
@@ -6,13 +6,13 @@ import com.android.build.gradle.LibraryExtension
 import com.android.build.gradle.TestExtension
 import io.gitlab.arturbosch.detekt.extensions.DetektExtension
 import org.assertj.core.api.Assertions.assertThat
-import org.gradle.api.Action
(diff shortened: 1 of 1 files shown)


detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPluginTest.kt
```diff
@@ -30,7 +30,10 @@ object DetektPluginTest : Spek({
         }
     }
 
-    describe("detekt plugin - Android") {
+    describe(
+        "detekt plugin - Android",
+        skip = if (System.getenv("ANDROID_SDK_ROOT") != null) Skip.No else Skip.Yes("No android sdk.")
```
BraisGabin, line 35: I was going to say that "normal" environment is `ANDROID_HOME` but I just found that they [deprecated that environment](https://developer.android.com/studio/command-line/variables#envar). A thing that I learned today. And a thing that I need to change in all my environments too...

BraisGabin (approved):

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPluginTest.kt
arturbosch: It's the one got reported when the build crashed for me locally xD.
One day I will code some Android too 8)

PR #3015, opened 2020-08-25
Run detekt with type resolution analysis on CI

arturbosch:


detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt
```diff
@@ -74,13 +68,12 @@ class FindingsAssert(actual: List<Finding>) :
                 failWithMessage("Expected ${expected.size} findings but was 0")
             }
         }
-        val code = finding!!.entity.ktElement?.containingKtFile?.text
-        if (code == null) {
-            failWithMessage("Expected ${expected.size} findings but was 0")
+        val code = requireNotNull(finding?.entity?.ktElement?.containingKtFile?.text) {
+            "Finding expected to provide an KtElement."
```
arturbosch, line 72:
```suggestion
            "Finding expected to provide a KtElement."
```

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/rules/MultiRuleSpec.kt
```diff
@@ -83,7 +83,7 @@ private class TestMultiRule(config: Config) : MultiRule() {
     }
 }
 
-private abstract class AbstractRule(config: Config) : Rule(config) {
+private open class AbstractRule(config: Config) : Rule(config) {
```
BraisGabin, line 86: 🤔 isn't this a false positive? I mean, we don't want any instance of `AbstractRule` so mark it as `abstract` seems like a good idea.

detekt-rules-empty/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyRule.kt
```diff
@@ -14,6 +14,7 @@ import org.jetbrains.kotlin.psi.KtExpression
 /**
  * Rule to detect empty blocks of code.
  */
+@Suppress("detekt.UnnecessaryAbstractClass") // we really do not want instances of this class
```
BraisGabin, line 17: Maybe we should rethink this rule...

.github/workflows/detekt-with-type-resolution.yaml
```diff
@@ -0,0 +1,37 @@
...
+        restore-keys: |
+          cache-gradle-
+
+    - name: Setup Java
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11
+
+    - name: Run analysis
+      run: ./gradlew detektMain detektTest --build-cache --parallel
```
BraisGabin, line 37: We can change now the pre merge workflow so it doesn't runs detekt each time, right?

BraisGabin (approved):

detekt-rules-empty/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyRule.kt
arturbosch: Well the only real purpose of this class is to share some logic, which could be done via composition.
It's methods also do two things `addFindingIfBlockExprIsEmpty` check and add xD

.github/workflows/detekt-with-type-resolution.yaml
arturbosch: If we do this than only in the pre-merge.yml so it is run locally.
But do we safe so much time? detekt without type resolution is fast.

arturbosch: Hm, but yes. Just one CI job for quality is a nicer separation.

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/rules/MultiRuleSpec.kt
arturbosch: The rule could additionally check if any inherited members used in it's members.
Or a simpler `extends Something` check for not requiring type resolution.

.github/workflows/detekt-with-type-resolution.yaml
BraisGabin: Yes, the idea is that. if intellij doesn't remove a unused import I don't want all the checks in red.

PR #3025, opened 2020-08-27
Fix RedundantSuspendModifier message

szymonkozak:
Message was copy-pasted from other rule and is invalid.


BraisGabin (approved):

schalkms (approved):
Good catch! Thank you very much 🥇

Issue #1589, opened 2019-04-15
Whitelist not applied


mirecalu:
I'm trying to add some rules into the Whitelist but it has no effect on the report file.
There is an example of `UndocumentedPublicClass` in `MenuEntryView.kt` that should be white listed.

```kotlin
//detekt.gradle
detekt {
    config = files("$rootDir/default-detekt-config.yml")
    filters = ".*build.*,.*/resources/.*,.*/tmp/.*"
    baseline = file("$rootDir/detekt-baseline.xml")
}
```

```xml
<!--detekt-baseline.xml-->
<?xml version="1.0" ?>
<SmellBaseline>
    <Whitelist>
        <ID>
            UndocumentedPublicClass:MenuEntryView.kt$com.foo.bar.menu.MenuEntryView.kt
        </ID>
    </Whitelist>
</SmellBaseline>
```
I don't manage to see what I'm missing into the documentation

arturbosch:
Hm can you try to inline the the ID like `<ID>[issue]</ID>`?
Maybe its the whitespace.

mirecalu:
Yes, It didnt work either

marschwar:
Can you please run `./gradlew detektBaseline` to create a baseline file. I did just that and received

```
<?xml version="1.0" ?>
<SmellBaseline>
  <Blacklist timestamp="1555415294513"></Blacklist>
  <Whitelist timestamp="1555415294513">
    <ID>UndocumentedPublicClass:MenuEntryView.kt$MenuEntryView</ID>
  </Whitelist>
</SmellBaseline>
```

As you can see, the simple class name instead of the fully qualified name is used.

mirecalu:
When I run `./gradlew detektBaseline` I only have theses two empty list (even if I have plenty of errors in the HTML report when I run `./gradlew detekt`)
```xml
<?xml version="1.0" ?>
<SmellBaseline>
  <Blacklist timestamp="1555432176090"></Blacklist>
  <Whitelist timestamp="1555432176857"></Whitelist>
</SmellBaseline>
```

I also tried to add this line ` <ID>UndocumentedPublicClass:MenuEntryView.kt$MenuEntryView</ID>`, it doesnt work either

marschwar:
Which version of detekt are you using?

mirecalu:
Okey I figured it out ! 
I have two modules in my AndroidProject : 
- app : which has several findings in detekt
- cleanModule ; which has NO finding

[...]

PR #3027, opened 2020-08-27
Don't create a baseline file if there are no issues

### LINKED ISSUE #3024: Should the Baseline task generate a baseline file when no issues are recorded?
Personally I think it should not, because the mere existance of a baseline file - even if empty - suggests that it is ok for developers to add issues to it, either manually or by calling the respective baseline task.

What do you think?

realdadfish:
An attempt to fix #3024 

Open points:
- Is the `Appendable` the right way to output things to console? I noticed `Detektion` also has an `add(notification)` API, but its hard to get a reference to `Detektion` into a `ReportingExtension`.
- Is it ok to tell people to get rid of their existing baselines if no issues are found? Projects might share baseline files (Detekt does!), as multiple modules have no issues, but the baseline they reference is non-empty. Personally I think each (sub)module should have it's own baseline file, to avoid eventual file name / signature clashes, even if these should be very, very rare.
- I don't see how baseline updates can actually work in a shared-baseline-file world. The `BaselineFacade` always overwrites `currentIssues` with the newly reported ones, so if I understand this correctly a baseline creation / update in some module will always overwrite any previous baseline updates, no?
- Lastly, I know, a test is still missing, would have added one if the class would have already be tested, but this one isn't yet :)

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFacade.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFacade.kt
@@ -23,5 +23,5 @@ class BaselineFacade {
         BaselineFormat().write(baseline, baselineFile)
     }
 
-    fun baselineExists(baseline: Path) = baseline.exists() && baseline.isFile()
+    internal fun baselineExists(baseline: Path) = baseline.exists() && baseline.isFile()
 }
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFacade.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFacade.kt
@@ -23,5 +23,5 @@ class BaselineFacade {
         BaselineFormat().write(baseline, baselineFile)
     }
 
-    private fun baselineExists(baseline: Path) = baseline.exists() && baseline.isFile()
+    fun baselineExists(baseline: Path) = baseline.exists() && baseline.isFile()
 }
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineResultMapping.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineResultMapping.kt
@@ -39,12 +39,9 @@ class BaselineResultMapping : ReportingExtension {
             if (flatten.isNotEmpty()) {
                 facade.createOrUpdate(baselinePath, flatten)
             } else {
-                output?.appendLine("No issues found, baseline file will not be created / updated.")
+                val action = if (facade.baselineExists(baselinePath)) "updated" else "created"
+                output?.appendLine("No issues found, baseline file will not be $action.")
             }
-        }
-
-        if (flatten.isEmpty() && facade.baselineExists(baselinePath)) {
-            output?.appendLine("No issues found, you can safely delete the existing baseline file at $baselinePath.")
         }
 
         return facade.transformResult(baselinePath, DetektResult(this)).findings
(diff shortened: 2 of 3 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFacade.kt
```diff
@@ -23,5 +23,5 @@ class BaselineFacade {
         BaselineFormat().write(baseline, baselineFile)
     }
 
-    private fun baselineExists(baseline: Path) = baseline.exists() && baseline.isFile()
+    fun baselineExists(baseline: Path) = baseline.exists() && baseline.isFile()
```
arturbosch, line 26:
```suggestion
    internal fun baselineExists(baseline: Path) = baseline.exists() && baseline.isFile()
```

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineResultMapping.kt
```diff
@@ -14,30 +14,39 @@ class BaselineResultMapping : ReportingExtension {
 
     private var baselineFile: Path? = null
     private var createBaseline: Boolean = false
+    private var output: Appendable? = null
 
     override fun init(context: SetupContext) {
         baselineFile = context.getOrNull(DETEKT_BASELINE_PATH_KEY)
         createBaseline = context.getOrNull(DETEKT_BASELINE_CREATION_KEY) ?: false
+        output = context.outputChannel
```
arturbosch, line 22: Yep, the output channel is the right way to output messages.
`Detektion.add(Notification)` was there before extensions could access the `SetupContext`.

arturbosch:
> An attempt to fix #3024
> 
> Open points:
> 
> * Is the `Appendable` the right way to output things to console? I noticed `Detektion` also has an `add(notification)` API, but its hard to get a reference to `Detektion` into a `ReportingExtension`.
Yep ^^
> * Is it ok to tell people to get rid of their existing baselines if no issues are found? Projects might share baseline files (Detekt does!), as multiple modules have no issues, but the baseline they reference is non-empty. Personally I think each (sub)module should have it's own baseline file, to avoid eventual file name / signature clashes, even if these should be very, very rare.

You are right, lets not report this case.
Agree on a baseline file per module with #1589 to get a merged top level version.
In our repo we use the workaround to have one task `detektProjectBaseline` to create a baseline for the whole project and each module uses it read-only.

> * I don't see how baseline updates can actually work in a shared-baseline-file world. The `BaselineFacade` always overwrites `currentIssues` with the newly reported ones, so if I understand this correctly a baseline creation / update in some module will always overwrite any previous baseline updates, no?
You are right - #1589.

> * Lastly, I know, a test is still missing, would have added one if the class would have already be tested, but this one isn't yet :)
You can take a look at

https://github.com/detekt/detekt/blob/c45791768df8fbae19be04f2b0fe69e9f83b7aa7/detekt-rules-documentation/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/AbsentOrWrongFileLicenseSpec.kt#L63-L72

on how to stub the SetupContext and use a `StringBuilder` to assert the new output.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineResultMapping.kt
```diff
@@ -14,30 +14,39 @@ class BaselineResultMapping : ReportingExtension {
...
+        val facade = BaselineFacade()
+        val flatten = this.flatMap { it.value }
 
-            if (createBaseline) {
-                val flatten = findings.flatMap { it.value }
+        if (createBaseline) {
+            if (flatten.isNotEmpty()) {
                 facade.createOrUpdate(baselinePath, flatten)
+            } else {
+                output?.appendLine("No issues found, baseline file will not be created / updated.")
```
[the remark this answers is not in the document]
realdadfish: The file might actually exist, this is checked inside the `BaselineFacade`. To make the message more specific here we'd have to check for the file's existance.

BraisGabin (approved):
Sorry for all this time. I know that I was against this PR but I changed my mind after reading #3344. If you misconfigure the baseline file you will notice it because detekt will fail. So we don't need to have the same behaviour for baseline and configuration.

We need to fix other thing: right now detekt fails if you have the baseline configured but the file doesn't exist. But that could be done in other PR.

BraisGabin:
I think that we could merge this as soon as #3429 is merged.

BraisGabin:
This was merged in #3450

realdadfish:
Thanks for taking care!

PR #3030, opened 2020-08-28
Improve documentation

BraisGabin:
Right now our gradle documentation is really verbose It repeats the same block of code 5 times so it's difficult to find what you are looking for.

I do think that jcenter is not the best repository and qw should be avoid it when possible. But that's the user choice and our documentation is not the place where we should promote that.

### DIFF
--- a/docs/pages/gettingstarted/gradle.md
+++ b/docs/pages/gettingstarted/gradle.md
@@ -175,6 +175,8 @@ repositories {
 }
 ```
 
+For more information about how to configure the repositories read [about the repositories](#repositories)
+
 ##### <a name="closure">Options for detekt configuration closure</a>
 
 ###### Groovy DSL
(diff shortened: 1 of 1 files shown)


BraisGabin:
We need to do something with `codecov` it's failing all the time.

schalkms (approved):

docs/pages/gettingstarted/gradle.md
```diff
@@ -68,18 +68,7 @@ plugins {
...
-    jcenter()
-
-    // or
-
-    mavenCentral()
-    jcenter {
-        content {
-            // just allow to include kotlinx projects
-            // detekt needs 'kotlinx-html' for the html report
-            includeGroup "org.jetbrains.kotlinx"
```
arturbosch, line 80: hm, I would really like to leave one of these for reference on.

cortinico: Agree. Maybe we can refactor this into a `## Repositories` section and have it listed only once?

BraisGabin: I implemented the @cortinico suggestion. What do you think?

arturbosch (approved):

Issue #2994, opened 2020-08-20
Rules with mixed behavior based on Type resolution


cortinico:
We do have several rules that have a mixed behavior (i.e. perform more checks or more detailed analysis) based on wether Type resolution is enabled or not.

This is a non-exhaustive list (current list can be found by looking for `@Suppress("ViolatesTypeResolutionRequirements")`):

- [x] `ImplicitDefaultLocale` #6344
- [x] `MapGetWithNotNullAssertionOperator` #6345 
- [x] `InstanceOfCheckForException` #6346 
- [x] `MemberNameEqualsClassName` #6347 
- [x] `SpreadOperator` #6351
- [x] `UnusedPrivateMember` #6356
- [x] `UnnecessaryAbstractClass` #5829 
- [x] `UseDataClass` #6355 
- [x] `UseRequire` #6358
- [x] `OptionalUnit` #6359
- [x] `UnusedImport` #6360
- [ ] `LongParameterList`
- [x] `LateinitUsage` #6379
- [x] `FunctionOnlyReturningConstant` #6380
- [x] `UseCheckOrError` #6361

While those rules are providing value also for users that don't have Type resolution enabled, they also lead to false negative reports. Generally, having the behavior of a rule changing based on the configuration of Detekt is not ideal in my opinion.

How should we treat them?

Imho we should guard all of them and make them required to run only with Type resolution enabled (i.e. adding a `if (bindingContext == EMPTY) return` in their body). 
On the other hand, users might complaining that those rules "stopped working" if they don't have type resolution enabled.

EDIT: Updated list

arturbosch:
I think we should support them at least till 2.0 and later rely on the compiler plugin to get first class support for resolution.

BraisGabin:
Agree, I think that we should keep them for now. In my actual project I don't have type solving because it's android with custom build variants and flavors. And, to be honest, I don't know how to enable it properly.

We don't even have a proper documentation. We said that it's an experimental feature, but we already know that it's the way we want for detekt. 


#2787 looks promising. If, after that, we update our documentation to enable type solving by default and then, give the option to disable it we can improve the reports that our users get and decrease the number of issues we get for user that don't have the type solving enabled.

cortinico:
As a rule of thumb I think we should refrain from adding rules that have mixed behavior. 
And I agree on keeping them around till Detekt 2.x and then promote them to type resolution only.

cortinico:
Thanks for the heads up. That will be available:

[...]

PR #3033, opened 2020-08-29
UnusedImports rule: fix false negative when same name identifiers are imported and used

### LINKED ISSUE #3020: Unused import not detected
## Expected Behavior

An unused import should be detected.

## Observed Behavior

The unused import is not detected.

## Steps to Reproduce

When using [clikt](https://github.com/ajalt/clikt), import

```
import com.github.ajalt.clikt.parameters.arguments.convert
import com.github.ajalt.clikt.parameters.options.convert
```

but use only `com.github.ajalt.clikt.parameters.arguments.convert`, then `com.github.ajalt.clikt.parameters.options.convert` is not detected as an unused import.

## Context

See https://github.com/oss-review-toolkit/ort/blob/376555874ab7e46a95e894a2702e15b94326ee85/helper-cli/src/main/kotlin/commands/FormatRepositoryConfigurationCommand.kt#L24-L25

## Your Environment

* Link to your project (if it's a public repository): https://github.com/oss-review-toolkit/ort

t-kameyama:
Fixes #3020

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
@@ -18,11 +18,16 @@ import org.jetbrains.kotlin.psi.KtImportList
 import org.jetbrains.kotlin.psi.KtPackageDirective
 import org.jetbrains.kotlin.psi.KtReferenceExpression
 import org.jetbrains.kotlin.psi.psiUtil.getChildrenOfType
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
 
 /**
  * This rule reports unused imports. Unused imports are dead code and should be removed.
  * Exempt from this rule are imports resulting from references to elements within KDoc and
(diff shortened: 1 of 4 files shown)


schalkms (approved):

arturbosch (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
```diff
@@ -53,17 +58,31 @@ class UnusedImports(config: Config) : Rule(config) {
...
 
-            fun KtImportDirective.isNotUsed() =
-                    aliasName !in namedReferences && identifier() !in namedReferences
+            @Suppress("ReturnCount")
+            fun KtImportDirective.isNotUsed(): Boolean {
+                val namedReferencesAsString = namedReferences.map { it.text.trim('`') }
+                if (aliasName in (namedReferencesInKDoc + namedReferencesAsString)) return false
+                val identifier = identifier()
+                if (identifier in namedReferencesInKDoc) return false
+                return if (bindingContext == BindingContext.EMPTY) {
```
cortinico, line 77: This is yet another rule where we're adding a mixed behavior based on Type Resolution: #2994 

Ideally we should either:
- Restrict this rule to work on TR only or 
- Remove the `@requiresTypeResolution` in the documentation as it's an overstatement
- Revert this change and find a different solution that doesn't require TR.

PR #3060, opened 2020-09-08
Prepare 1.13.0 release

arturbosch:


BraisGabin (approved):

arturbosch:
I'm not sure if jcenter just needs again some time after failing to upload so many files or if https://github.com/detekt/detekt/pull/3006 broke uploading some artifacts to the gradle plugin portal.

Need to investigate how to exclude publishing `md5` hashes. One hash should be enough (sha1).

schalkms (approved):
Agree @arturbosch

arturbosch:
I republished the Gradle plugin.
Jcenter sync still needs some time due to too many failing uploads :/

arturbosch:
Releases is up and in the releases repository. Should be up on maven central in the next minutes/hours.
CI will fail until then.

buildSrc/src/main/kotlin/packaging.gradle.kts
```diff
@@ -8,6 +8,29 @@ plugins {
...
+}
+
+val sonatypeUsername: String? = findProperty("sonatypeUsername")
+    ?.toString()
+    ?: System.getenv("MAVEN_CENTRAL_USER")
+val sonatypePassword: String? = findProperty("sonatypePassword")
+    ?.toString()
+    ?: System.getenv("MAVEN_CENTRAL_PW")
+
+// still needed for artifactory snapshot publishing
```
cortinico, line 21: How about we move to use to the Nexus Snapshot repository? So we can remove `id("com.jfrog.artifactory")` and we don't have to handle double credentials 👌

arturbosch: Yep, that would be good. I saved it for another PR ^^.
Feel free to pick it up if you have time. I will have more time starting next week I think.

PR #3064, opened 2020-09-10
MagicNumber rule: don't check Character, Boolean and null

t-kameyama:


schalkms (commented):
Thanks for the improvement!
Does this belong to any issue?
Can we please verify the additional paths with unit tests? Thanks!

t-kameyama:
> Does this belong to any issue?

No.

> Can we please verify the additional paths with unit tests? Thanks!

https://github.com/detekt/detekt/blob/dd1a995f55b8d8fc37ff9dfd141351f97933b135/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumberSpec.kt#L308-L328

https://github.com/detekt/detekt/blob/dd1a995f55b8d8fc37ff9dfd141351f97933b135/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumberSpec.kt#L525-L532

schalkms (approved):
So this adds a precondition to this rule in order to avoid executing unnecessary rule logic statements. Good! I hope that I got it right.

PR #3075, opened 2020-09-15
Update KtLint to 0.39.0

cortinico:
Bumping KtLint to the latest stable. 
I also noticed that there are two experimental rules on KtLint side that we should wrap: `AnnotationSpacingRule` and `ArgumentListWrappingRule`.
Should we do it in a separate PR?


schalkms (approved):

BraisGabin (approved):

3flex:
I'll raise issues for the two new wrappings.

PR #3088, opened 2020-09-18
New rule: ClassOrdering

3flex:
A rule to check the contents of a class match this ordering: https://kotlinlang.org/docs/reference/coding-conventions.html#class-layout

I've enabled this for detekt, and you can see in that second commit the changes that were required to comply with the new rule. I think the changes are mostly not controversial, but for those classes where they are a suppression is probably the right way to go, or perhaps changing some extension properties into extension functions.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
@@ -66,11 +66,11 @@ class ClassOrdering(config: Config = Config.empty) : Rule(config) {
     }
 
     override fun visitClassBody(classBody: KtClassBody) {
+        super.visitClassBody(classBody)
+
         if (!classBody.declarations.isInOrder(lengthComparator)) {
             report(CodeSmell(issue, Entity.from(classBody), issue.description))
         }
-
-        super.visitClassBody(classBody)
     }
 
     @Suppress("MagicNumber")
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
@@ -16,7 +16,8 @@ import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.KtSecondaryConstructor
 
 /**
- * This rule ensures class contents are ordered as follows as recommended by the Kotlin Coding Conventions:
+ * This rule ensures class contents are ordered as follows as recommended by the Kotlin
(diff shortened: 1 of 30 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
```diff
@@ -0,0 +1,91 @@
...
+import org.jetbrains.kotlin.psi.KtClassBody
+import org.jetbrains.kotlin.psi.KtClassInitializer
+import org.jetbrains.kotlin.psi.KtDeclaration
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.KtObjectDeclaration
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.psi.KtSecondaryConstructor
+
+/**
+ * This rule ensures class contents are ordered as follows as recommended by the Kotlin Coding Conventions:
```
cortinico, line 19: Maybe add the link here?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
```diff
@@ -0,0 +1,91 @@
...
+        if (orderPriority(str1) == null || orderPriority(str2) == null) return@Comparator 0
+        compareValues(orderPriority(str1), orderPriority(str2))
+    }
+
+    override fun visitClassBody(classBody: KtClassBody) {
+        if (!classBody.declarations.isInOrder(lengthComparator)) {
+            report(CodeSmell(issue, Entity.from(classBody), issue.description))
+        }
+
+        super.visitClassBody(classBody)
```
cortinico, line 72: I believe the super call should be done as first statement

cortinico (approved):
Great rule 👌  I'm just unsure if we want to enabled it by default as it has the risk of raising a lot of warnings for our users. However the logic is really straightforward and aligned with the Kotlin guidelines.

3flex:
> I'm just unsure if we want to enabled it by default

Sorry I wasn't clear. I've enabled it for detekt scans of the detekt project itself, but it's not enabled by default for users - it's `active: false` in default-detekt-config.yml. I thought the same thing as you, it could potentially lead to a lot of warnings on larger projects.

I also think that in almost all cases any violations it finds will lead to a more logical class layout so I think there's value in enabling by default. It's quick to cut/paste code to fix issues it finds. But for now it's not enabled by default.

BraisGabin (approved):
I like this one a lot! Myabe we could add some configuration later but, let's see what the community think when they test it :)

cortinico:
> Sorry I wasn't clear. I've enabled it for detekt scans of the detekt project itself, but it's not enabled by default for users - it's `active: false` in default-detekt-config.yml. I thought the same thing as you, it could potentially lead to a lot of warnings on larger projects.

Thanks for the clarification 👌 Looks good to me 🚀

schalkms (approved):

PR #3089, opened 2020-09-19
New rule: ReplaceSafeCallChainWithRun

3flex:
I saw this neat trick while browsing Kotlin compiler source.

Instead of having a long chain of safe calls, when many of those safe calls are redundant (because the type returned by the preceding call is non-nullable), that chain of safe calls can be surrounded with `run {}` and the redundant safe calls can be removed.

This rule only checks from the end of a chain and works backwards, so it won't recommend inserting `run` blocks in the middle of a safe call chain which I think could be very confusing to read, especially since it could mean multiple `run` blocks in a very long chain.

The rule also checks for any opportunity to replace a chain by surrounding with `run`, so any redundant safe calls at the end of a chain will be reported, even if there's only one. I personally think this is fine because it still reduces complexity which I see as being the main benefit of this rule.

I'm interested in feedback on this one!

### DIFF
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ReplaceSafeCallChainWithRun.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ReplaceSafeCallChainWithRun.kt
@@ -7,8 +7,6 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
-import org.jetbrains.kotlin.com.intellij.psi.PsiElement
-import org.jetbrains.kotlin.psi.KtElement
 import org.jetbrains.kotlin.psi.KtSafeQualifiedExpression
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
@@ -58,22 +56,13 @@ class ReplaceSafeCallChainWithRun(config: Config = Config.empty) : Rule(config) 
 
         var counter = 0
 
-        var check: PsiElement = expression
-
-        while (check.firstChild is KtSafeQualifiedExpression) {
-            if (
-                (expression.firstChild as KtElement)
-                    .getResolvedCall(bindingContext)
-                    ?.resultingDescriptor
-                    ?.returnType
-                    ?.isNullable() == true
-            ) {
-                break
-            }
-
-            check = check.firstChild
+        var receiver = expression.receiverExpression
+        while (receiver is KtSafeQualifiedExpression) {
+            if (receiver.getResolvedCall(bindingContext)?.resultingDescriptor?.returnType?.isNullable() == true) break
             counter++
+            receiver = receiver.receiverExpression
         }
+
         if (counter >= 1) report(
             CodeSmell(issue, Entity.from(expression), issue.description)
         )
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ReplaceSafeCallChainWithRun.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ReplaceSafeCallChainWithRun.kt
@@ -18,20 +18,25 @@ import org.jetbrains.kotlin.types.isNullable
  * Chains of safe calls on non-nullable types are redundant and can be removed by enclosing the redundant safe calls in
  * a `run {}` block. This improves code coverage and reduces cyclomatic complexity as redundant null checks are removed.
  *
+ * This rule only checks from the end of a chain and works backwards, so it won't recommend inserting run blocks in the
+ * middle of a safe call chain as that is likely to make the code more difficult to understand.
+ *
+ * The rule will check for every opportunity to replace a safe call when it sits at the end of a chain, even if there's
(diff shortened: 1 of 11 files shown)


BraisGabin (approved):
Wow, neat trick. Those endless safe chains are always a problem for me. They are hard to understand what can really be null. Even for one is worth it. 👏🏻👏🏻 Nice one.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ReplaceSafeCallChainWithRun.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.jetbrains.kotlin.psi.KtElement
+import org.jetbrains.kotlin.psi.KtSafeQualifiedExpression
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.types.isNullable
+
+/**
+ * Chains of safe calls on non-nullable types are redundant and can be removed by enclosing the redundant safe calls in
+ * a `run {}` block. This improves code coverage and reduces cyclomatic complexity as redundant null checks are removed.
+ *
```
schalkms, line 20: Maybe we should add your good detailed facts regarding the implementation also to the description.
It might be helpful to understand why some chains are reported from the user perspective.

```
Instead of having a long chain of safe calls, when many of those safe calls are redundant (because the type returned by the preceding call is non-nullable), that chain of safe calls can be surrounded with run {} and the redundant safe calls can be removed.

This rule only checks from the end of a chain and works backwards, so it won't recommend inserting run blocks in the middle of a safe call chain which I think could be very confusing to read, especially since it could mean multiple run blocks in a very long chain.

The rule also checks for any opportunity to replace a chain by surrounding with run, so any redundant safe calls at the end of a chain will be reported, even if there's only one. I personally think this is fine because it still reduces complexity which I see as being the main benefit of this rule.
```

schalkms (approved):
I agree. This rule is a nice addition. Great idea! 👍 🥇

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ReplaceSafeCallChainWithRun.kt
```diff
@@ -0,0 +1,76 @@
...
+        need to run checks on this one */
+        if (expression.parent is KtSafeQualifiedExpression) return
+
+        var counter = 0
+
+        var check: PsiElement = expression
+
+        while (check.firstChild is KtSafeQualifiedExpression) {
+            if (
+                (expression.firstChild as KtElement)
```
t-kameyama, line 60:
```suggestion
                (check.firstChild as KtElement)
```

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ReplaceSafeCallChainWithRun.kt
```diff
@@ -0,0 +1,76 @@
...
+                    ?.resultingDescriptor
+                    ?.returnType
+                    ?.isNullable() == true
+            ) {
+                break
+            }
+
+            check = check.firstChild
+            counter++
+        }
```
t-kameyama, line 71: IMHO
```kotlin
var receiver = expression.receiverExpression
while (receiver is KtSafeQualifiedExpression) {
    if (receiver.getResolvedCall(bindingContext)?.resultingDescriptor?.returnType?.isNullable() == true) break
    counter++
    receiver = receiver.receiverExpression
}
```

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ReplaceSafeCallChainWithRun.kt
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Chains of safe calls on non-nullable types are redundant and can be removed by enclosing the redundant safe calls in
+ * a `run {}` block. This improves code coverage and reduces cyclomatic complexity as redundant null checks are removed.
+ *
+ * <noncompliant>
+ * val x = classOrObject.body?.declarations
+ *         ?.filterIsInstance<KtNamedFunction>()
+ *         ?.filter { !isIgnoredFunction(it) }
+ *         ?.size ?: 0
```
cortinico, line 25:
```suggestion
 * val x = System.getenv()
 *             ?.getValue("HOME")
 *             ?.toLowerCase()
 *             ?.split("/") ?: emptyList()
```

It's a easier example and does not involve detekt specific types/functions.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ReplaceSafeCallChainWithRun.kt
```diff
@@ -0,0 +1,76 @@
...
+ * </noncompliant>
+ *
+ * <compliant>
+ * val x = classOrObject.body?.run {
+ *         declarations
+ *             .filterIsInstance<KtNamedFunction>()
+ *             .filter { !isIgnoredFunction(it) }
+ *             .size
+ *     } ?: 0
```
cortinico, line 34:
```suggestion
 * val x = getenv()?.run {
 *     getValue("HOME")
 *         .toLowerCase()
 *         .split("/")
 * } ?: emptyList()
```

cortinico (approved):
Great addition 👍

PR #3091, opened 2020-09-19
Enable parallel spek test execution

schalkms:
https://github.com/spekframework/spek/releases/tag/2.0.13


arturbosch:
Have you experienced any benefits when using this together with Gradle parallel?

schalkms:
@arturbosch Unfortunately, as it seems like, there's no benefit for detekt in this case. Also, there are failing tests in the parallel test environment. Consequently, I'll have to close this.

PR #3092, opened 2020-09-19
Update RedundantVisibilityModifierRule to find redundant internal modifiers

3flex:
There were some redundant `internal` modifiers identified by the IDE that weren't picked up by detekt's own rule, so I updated the rule to find this type of usage.

I also enabled the rule for detekt self-scans and fixed the reported issues.

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantVisibilityModifierRuleSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantVisibilityModifierRuleSpec.kt
@@ -127,5 +127,27 @@ class RedundantVisibilityModifierRuleSpec : Spek({
             """
             assertThat(subject.compileAndLint(code)).isEmpty()
         }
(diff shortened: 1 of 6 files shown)


schalkms (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantVisibilityModifierRuleSpec.kt
```diff
@@ -127,5 +127,27 @@ class RedundantVisibilityModifierRuleSpec : Spek({
             """
             assertThat(subject.compileAndLint(code)).isEmpty()
         }
+
+        it("reports internal modifier on nested class in private object") {
+            val code = """
+                private object Xml10EscapeSymbolsInitializer {
```
cortinico, line 133:
```suggestion
                private object HiddenObject {
```

nit: let's try to keep the tested code agnostic.

cortinico (approved):

arturbosch (approved):

PR #3094, opened 2020-09-21
Fix false positive (+= overload) in UnusedPrivateMember

fourlastor:
Hi! I noticed that detekt emits a false positive when checking for unused members if the function is an operator function, and if it's used via an assignment operation (`+=` in the test case).

I know that the issue likely lies in `io.gitlab.arturbosch.detekt.rules.style.UnusedFunctionVisitor#getUnusedReports`:
```kotlin
val operatorToken = OperatorConventions.getOperationSymbolForName(Name.identifier(name))
```

This will return the token `+` as the name of the function is `plus`, so `+=` will get skipped and appear as unused.

I'm not too familiar with detekt internals so I'm not sure how to solve it, I'm open to ideas/suggestions though/happy to let someone else take over the PR.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
@@ -97,7 +97,11 @@ private class UnusedFunctionVisitor(
                         val operatorToken = OperatorConventions.getOperationSymbolForName(Name.identifier(name))
                         val operatorValue = (operatorToken as? KtSingleValueToken)?.value
                         when (operatorToken) {
-                            KtTokens.PLUS, KtTokens.MINUS, KtTokens.MUL, KtTokens.DIV, KtTokens.PERC -> operatorValue?.let {
+                            KtTokens.PLUS,
+                            KtTokens.MINUS,
+                            KtTokens.MUL,
+                            KtTokens.DIV,
+                            KtTokens.PERC -> operatorValue?.let {
                                 functionReferences[it].orEmpty() + functionReferences["$it="].orEmpty()
                             }.orEmpty()
                             else -> operatorValue?.let { functionReferences[it] }.orEmpty()
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
@@ -17,6 +17,8 @@ import io.gitlab.arturbosch.detekt.rules.isOpen
 import io.gitlab.arturbosch.detekt.rules.isOperator
 import io.gitlab.arturbosch.detekt.rules.isOverride
 import org.jetbrains.kotlin.lexer.KtSingleValueToken
+import org.jetbrains.kotlin.lexer.KtToken
+import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.psi.KtClass
 import org.jetbrains.kotlin.psi.KtFile
@@ -95,7 +97,12 @@ private class UnusedFunctionVisitor(
                     val referencesViaOperator = if (isOperator) {
                         val operatorToken = OperatorConventions.getOperationSymbolForName(Name.identifier(name))
                         val operatorValue = (operatorToken as? KtSingleValueToken)?.value
-                        operatorValue?.let { functionReferences[it] }.orEmpty()
+                        when (operatorToken) {
+                            KtTokens.PLUS, KtTokens.MINUS, KtTokens.MUL, KtTokens.DIV, KtTokens.PERC -> operatorValue?.let {
+                                functionReferences[it].orEmpty() + functionReferences["$it="].orEmpty()
+                            }.orEmpty()
+                            else -> operatorValue?.let { functionReferences[it] }.orEmpty()
+                        }
                     } else {
                         emptyList()
                     }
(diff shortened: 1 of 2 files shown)


schalkms (commented):
Thanks for reporting this scenario to detekt and contributing a test case. 👍 
You caught the right line in the source code that prevents this code from not being reported.

The following section in the Kotlin doc describes how this is handled internally by the compiler.
https://kotlinlang.org/docs/reference/operator-overloading.html#assignments

So one has to be careful here. I think there is a simple (stupid) solution here. If there is an operator fun `plus` available, also add `plusAssign` to the corresponding reference list. This should be done for `+= -= *= /= %=`.
Overridden `plus` and `plusAssign` operator function pairs can't coexist as the following statement in the doc mentions.

> If the corresponding binary function (i.e. plus() for plusAssign()) is available too, report error (ambiguity)

@t-kameyama do you have better ideas on how to solve this? Do you know about an existing Kotlin function that already solves this issue?

fourlastor:
@schalkms thanks! that was really helpful

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -95,7 +96,12 @@ private class UnusedFunctionVisitor(
                     val referencesViaOperator = if (isOperator) {
                         val operatorToken = OperatorConventions.getOperationSymbolForName(Name.identifier(name))
                         val operatorValue = (operatorToken as? KtSingleValueToken)?.value
-                        operatorValue?.let { functionReferences[it] }.orEmpty()
+                        when (operatorToken) {
+                            KtTokens.PLUS, KtTokens.MINUS, KtTokens.MUL, KtTokens.DIV, KtTokens.PERC -> operatorValue?.let {
+                                functionReferences[it].orEmpty() + functionReferences["$it="].orEmpty()
+                            }.orEmpty()
+                            else -> operatorValue?.let { functionReferences[it] }.orEmpty()
```
schalkms, line 103: The code in the whole function could probably be simplified. However, I'd see this out of scope for this PR. Consequently, this refactoring step should go into a new PR.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
```diff
@@ -1054,6 +1054,20 @@ class UnusedPrivateMemberSpec : Spek({
...
 
+        it("does not report used plus operator in methods with +=") {
+            val code = """
+                import java.util.Date
+                class Foo {
+                    var number: Float? = null
+                    fun foo() {
+                        number += 3f
+                    }
+                    private operator fun Float?.plus(other: Float?) = if (this == null && other == null) null else this ?: 0f + (other ?: 0f)
```
schalkms, line 1065: Detekt strives to keep the test code as simple as possible in order to make it more readable and maintainable.
The test snippets should be as simple as possible and test only the rule under test.

```suggestion
                class Test {
                    fun f() {
                        var number: Int = 0
                        number += 1
                        number -= 1
                        number *= 1
                        number /= 1
                        number %= 1
                    }
                    private operator fun Int.plus(other: Int) = 1
                    private operator fun Int.minus(other: Int) = 2
                    private operator fun Int.times(other: Int) = 3
                    private operator fun Int.div(other: Int) = 4
                    private operator fun Int.rem(other: Int) = 5
                    private operator fun Int.mod(other: Int) = 6
                }
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -95,7 +96,12 @@ private class UnusedFunctionVisitor(
                     val referencesViaOperator = if (isOperator) {
                         val operatorToken = OperatorConventions.getOperationSymbolForName(Name.identifier(name))
                         val operatorValue = (operatorToken as? KtSingleValueToken)?.value
-                        operatorValue?.let { functionReferences[it] }.orEmpty()
+                        when (operatorToken) {
+                            KtTokens.PLUS, KtTokens.MINUS, KtTokens.MUL, KtTokens.DIV, KtTokens.PERC -> operatorValue?.let {
```
schalkms, line 100: Detekt has a dog-fooding process in action, which basically means that we run detekt's rules over it's own codebase.
It detekt line 100 as too long (`MaxLineLength`). Can we please wrap the line in a subsequent commit in order to merge this awesome PR?

schalkms (commented):
Looks good! Thanks for providing the fix in an additional commit! 👍 
Please see my suggestion on how to improve the test code a little bit.

fourlastor:
Thank you for the suggestion! I'll address these tomorrow

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
fourlastor: b974091 that's a good point. I had to keep the nullable type, as kotlin would otherwise give precedence to the operator function defined in `Int`. I also didn't add `mod` as it would be effectively unused (`rem` would be used instead). I preferred `rem` over `mod` as the latter is deprecated.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
fourlastor: 57e9911 I should have looked at the CI output 😅

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
fourlastor: 27678ef I did some cleanup

schalkms (approved):
Thanks for incorporating the feedback and comments. Well done 👍

Issue #3066, opened 2020-09-11
Add ability to pass rootclasspaths to KtTestCompiler


tyvsmith:
## Expected Behavior
KtTestCompiler can take additional optional paths for loading on the root classpath for testing scenarios.

## Current Behavior
Currently, [KtTestCompiler](https://github.com/detekt/detekt/blob/a51719fc33a2b4d84f1515068c2ed3538a326c67/detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KtTestCompiler.kt#L37) allows for a test environment without specifying additional existing classes that may be needed on the root classpath. 

## Context
We have this change internally used for our testing environment when dealing with existing files both on the main classpath and in testing (we share EP and detekt stubs in some cases). I can upstream this if it'll be widely valuable.

Proposed Change:

```kotlin
internal fun createEnvironment(additionalPaths: List<File> = listOf()): KotlinCoreEnvironmentWrapper {
  val configuration = CompilerConfiguration()
  configuration.put(CommonConfigurationKeys.MODULE_NAME, "test_module")
  configuration.put(CLIConfigurationKeys.MESSAGE_COLLECTOR_KEY, MessageCollector.NONE)

  // Get the runtime location of stdlib jar and pass to the compiler so it's available to generate
  // the
  // BindingContext for rules under test.
  val path = File(CharRange::class.java.protectionDomain.codeSource.location.path)
  configuration.addJvmClasspathRoot(path)
  
  configuration.addJvmClasspathRoots(additionalPaths)

  val parentDisposable = Disposer.newDisposable()
  val kotlinCoreEnvironment =
      KotlinCoreEnvironment.createForTests(
          parentDisposable, configuration, EnvironmentConfigFiles.JVM_CONFIG_FILES)
  return KotlinCoreEnvironmentWrapper(kotlinCoreEnvironment, parentDisposable)
}
```

Example usage:

```kotlin
    val EXTRA_CLASSPATH_FILES =
        listOf(File(TestBannedMethods::class.java.protectionDomain.codeSource.location.path))

    val ENVIRONMENT = createEnvironment(EXTRA_CLASSPATH_FILES)
```

schalkms:
Thanks for providing a good suggestion!
I think it’s fine to merge this function upwards, since only test environment code gets added.
Let’s wait for @arturbosch feedback.

schalkms:
Btw sorry for the late reply. This good idea got missed somehow.

arturbosch:
[...]

PR #3101, opened 2020-09-25
Add additional classpaths to KtTestCompiler

tyvsmith:
This allows additional jvm root classpaths to be provided to the compiler for test. Useful for custom rules that need shared access to prod resources.

Example
```
    val EXTRA_CLASSPATH_FILES =
        listOf(File(ImportantProdClass::class.java.protectionDomain.codeSource.location.path))

    val ENVIRONMENT = createEnvironment(EXTRA_CLASSPATH_FILES)

```

Solves #3066


schalkms (approved):
Thanks for submitting this nice utility function to detekt. This can be helpful for other 3rd party contributors too.
Please take a look at the following issue reported by detekt, which should be fixed. After that, we are able to merge. 💪

`NoTrailingSpaces - [] at D:\a\detekt\detekt\detekt-test-utils\src\main\kotlin\io\github\detekt\test\utils\KotlinCoreEnvironmentWrapper.kt:28:3 `

PR #3107, opened 2020-09-26
Fix kotlin-reflect problem for idea plugin

arturbosch:
For reference: https://github.com/detekt/detekt-intellij-plugin/issues/100


schalkms (approved):

schalkms:
Is there any way we can think of to prevent such issues happening again in the future?

arturbosch:
> Is there any way we can think of to prevent such issues happening again in the future?

To be honest I don't know and I don't think so.
Seems to be related to callable reference improvements - https://kotlinlang.org/docs/reference/whatsnew14.html#callable-reference-improvements.
Some of these issues can be fixed by migrating away from Spek and Strikt and every other Kotlin library using `kotlin-reflect`.
The release candidate of Gradle seems to still use 1.3, maybe this will fix everything magically.

Reverting kotlin-reflect does not help - https://stackoverflow.com/questions/63474372/kotlin-delegate-property-causes-a-preview-rendering-error-in-android-studio.

cortinico (approved):

BraisGabin (approved):

PR #3109, opened 2020-09-27
Improve description of AbsentOrWrongFileLicense rule

### LINKED ISSUE #3100: How to define the license header?
I want to mandate a license header to all `kt` files.
The [AbsentOrWrongFileLicense](https://detekt.github.io/detekt/comments.html#absentorwrongfilelicense) rule allows me to do so with the `licenseTemplateFile` (default: `license.template`).

However, I could not find any further information about how to define the license template.
I need some flexibility for the date and the name.
For example, I want to create a license template that looks like this:
```
Copyright (c) [year] [name] (https://www.levelrin.com)
```
The above template should accept any year and name.

How to define the rules for the license header?

Thanks

schalkms:
The new description states how the rule checks Kotlin source files
against the given license header.
Closes #3100

### DIFF
--- a/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/AbsentOrWrongFileLicense.kt
+++ b/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/AbsentOrWrongFileLicense.kt
@@ -10,7 +10,9 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import org.jetbrains.kotlin.psi.KtFile
 
 /**
- * This rule will report every Kotlin source file which doesn't have required license header.
+ * This rule will report every Kotlin source file which doesn't have the required license header.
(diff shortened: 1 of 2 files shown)


detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/AbsentOrWrongFileLicense.kt
```diff
@@ -10,7 +10,9 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import org.jetbrains.kotlin.psi.KtFile
 
 /**
- * This rule will report every Kotlin source file which doesn't have required license header.
+ * This rule will report every Kotlin source file which doesn't have the required license header.
+ * Each Kotlin source file is checked, whether the header starts with the read text from the passed file in the
```
arturbosch, line 14:
```suggestion
 * Each Kotlin source file is checked whether the header starts with the read text from the passed file in the
```

Not sure if the comma is needs to be removed here.

arturbosch (approved):

PR #3112, opened 2020-09-27
Refactoring: Use indices syntax in loop

schalkms:


arturbosch (approved):

BraisGabin:
Do we have a rule for this?

schalkms:
@BraisGabin No, but there’s an IntelliJ inspection available.

Issue #3065, opened 2020-09-11
1.13.0 changes to detektGenerateConfig


Tolriq:
## Expected Behavior
Now detektGenerateConfig tries to create the configuration on the configured path.
This change is strange because:
- If the file exist the task refuse to create / update the file as already present
- If the file does not exist the task does not run either as the file is missing

detektGenerateConfig ignores detekt.config paramenter - #2565

The task should not take in account this parameter and eventually there should be a new parameter that allows to define default configuration location.

Currently we can no more generate default configuration and see what new options have been added / changed.

arturbosch:
I couldn't reproduce this.
The file gets generated for me.
A problem exist with multi project gradle builds where only the first subproject generates the file but all sub projects reference it.

Tolriq:
Yes the second issue was due to 1.13 publishing problem.

The problem of using the configured production path and not a default configuration path is that once you are in production you can't generate a new config to see what have changed between releases to update the production configuration.

Having a different file allows to keep both in git, see the diff apply what we want to production file. Now since the production file exist you can't generate a new file. If we delete the production file we can generate the new default but you can't diff since the diff is irrelevant to detekt version update but takes in account previous production file.

There should be 2 different files for default and production configuration using the same config value for both brings a lot of issues without really solving anything.

arturbosch:
You can create a new task based on our task for the diff usecase:

```kt
import io.gitlab.arturbosch.detekt.DetektGenerateConfigTask

val createDetektConfigForDiff by tasks.registering(DetektGenerateConfigTask::class) {
    description = ""
    config.setFrom(buildDir.resolve("detekt-diff.yaml"))

    doFirst {
      // optionally delete the old config diff file first 
    }
}
```

Does this work for you?

Tolriq:
Yes works perfectly, did not thought it would be so easy 👍 

Maybe worth some kind of recipes for others as this sounds like a normal need to know what changes between version and easily update config;

arturbosch:
> Yes works perfectly, did not thought it would be so easy  Thanks.
> 
> Maybe worth some kind of recipes for others as this sounds like a normal need to know what changes between version and easily update config;

Thanks for the suggestion! 
Will post it under `https://detekt.github.io/detekt/news`.

PR #3114, opened 2020-09-27
Write recipe to compare user and default config - #3065

arturbosch:


docs/_posts/2020-09-27-additional-diff-config-task.md
```diff
@@ -0,0 +1,39 @@
...
+
+```kt
+detekt {
+  ...
+  config = files(...)
+  ...
+}
+```
+
+When the file on this location already exists, your configuration won't be overwritten, and the task is a noop.
```
schalkms, line 20:
```suggestion
When this file path already exists, your configuration won't be overwritten, and the task is a noop.
```

schalkms (approved):
Thanks for writing ✍️ this article. It’s a very helpful reference for potential questions concerning this topic.

cortinico (approved):

Tolriq:
Actually what I do is not diff updated config with prod config but diff default config with previous default config.

This give a precise and easy to follow list of added checks / changed default values that can be applied to prod after.

Diffing default config with prod configuration that may have dozens of already known changes makes this task harder.

BraisGabin:
> Actually what I do is not diff updated config with prod config but diff default config with previous default config.
> 
> This give a precise and easy to follow list of added checks / changed default values that can be applied to prod after.
> 
> Diffing default config with prod configuration that may have dozens of already known changes makes this task harder.

My workflow is similar at yours. And I have an idea but I must say that I don't use it yet:

Create a patch file between the default config and my actual config. This way, the next time that I update detekt I can take the new default and apply the patch to it. And then I can diff my old config with the merge between my customization and the new default.

PR #3127, opened 2020-10-04
[RFC] New Rule API

BraisGabin:
This is a prove of concept for a new API for our rules.

The base idea is that a rule is just this:

```
interface NewRule : (KtFile, BindingContext, CompilerResources) -> List<Finding> {

    override fun invoke(file: KtFile, binding: BindingContext, resources: CompilerResources): List<Finding>
}
```

You don't even need to implement `KtTreeVisitor` if you don't need it for your rule.

This interface is really abstract so it force a lot of boilerplate for nearly all the use-cases. For this reason I implemented `BaseNewRule` that provides (more less) the same functionality as `BaseRule` without the boilerplate.

The core code isn't production ready. It don't even work. I know it, I just want to show the api and prove that it's possible to mix both apis in core.

In the last commit you can see another idea to simplify the rules a bit more. They expose it's information just once. And then, in core, we melt that information with the issues to create the `Finding`. This way we can remove `CodeSmell` from our public API and just expose the simplier `data class Asdf(Entity, message: String)` (name pending)

Notes:

- I only moved the coroutines rules because is our smallest rule set. But as you can see the BaseRule doesn't know about configuration so all the configurations should be injected to the rule. Probably those rules will need a factory to reduce the code inside the `RuleCollectionProvider`. Later, those Factories will be removed/refactored by #3002.
- I think that we can publish this API in detekt 1.0. We can support both APIs for a while. The new API should be used with an opt-in annotation. Once we like our new API we can deprecate the old and remove it in 2.0.
- I don't like the names that I'm using. Any idea in this regard is more than welcome.
- I know that I have #3002 open. But I don't see how to continue with it after we address this.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Finding.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Finding.kt
@@ -1,0 +1,77 @@
+package io.gitlab.arturbosch.detekt.api.v2
+
+import io.github.detekt.psi.FilePath
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.SeverityLevel
+import org.jetbrains.kotlin.psi.KtElement
+import java.nio.file.Path
+
+interface Finding {
+    val id: String
+    val message: String
+    val location: Location
+    val severityLevel: SeverityLevel
+    val debt: Debt
+    val entity: Entity
+    val correctable: Boolean
+    val rule: RuleInfo
+}
+
+interface Entity {
+    val id: String
+    val ktElement: KtElement
+}
+
+interface RuleInfo {
+    val id: String
+    val ruleSetId: String
+    val description: String
+}
+
+interface Location {
+    val source: SourceLocation
+    val text: TextLocation
+    val filePath: FilePath
+}
+
+interface FilePath {
+    val absolute: Path
+    val relative: Path?
+    val base: Path?
+}
+
+interface SourceLocation {
+    val line: Int
+    val column: Int
+
+    companion object {
+        operator fun invoke(line: Int, column: Int): SourceLocation {
+            return Impl(line, column)
+        }
+    }
+
+    private data class Impl(
+        override val line: Int,
+        override val column: Int
+    ) : SourceLocation {
+        override fun toString(): String = "$line:$column"
+    }
+}
+
+interface TextLocation {
+    val start: Int
+    val end: Int
+
+    companion object {
+        operator fun invoke(start: Int, end: Int): TextLocation {
+            return Impl(start, end)
+        }
+    }
+
+    private data class Impl(
+        override val start: Int,
+        override val end: Int
+    ) : TextLocation {
+        override fun toString(): String = "$start:$end"
+    }
+}
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Finding.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Finding.kt
@@ -32,6 +32,9 @@ interface SourceLocation {
     val line: Int
     val column: Int
 
+    operator fun component1(): Int = line
+    operator fun component2(): Int = column
+
     companion object {
         operator fun invoke(line: Int, column: Int): SourceLocation {
             return Impl(line, column)
@@ -49,6 +52,9 @@ interface SourceLocation {
 interface TextLocation {
     val start: Int
     val end: Int
+
+    operator fun component1(): Int = start
+    operator fun component2(): Int = end
 
     companion object {
         operator fun invoke(start: Int, end: Int): TextLocation {
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Finding.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Finding.kt
@@ -4,7 +4,6 @@ import io.github.detekt.psi.FilePath
 import io.gitlab.arturbosch.detekt.api.Debt
 import io.gitlab.arturbosch.detekt.api.SeverityLevel
 import org.jetbrains.kotlin.psi.KtElement
-import java.nio.file.Path
 
 interface Finding {
     val id: String
@@ -12,14 +11,9 @@ interface Finding {
     val location: Location
     val severityLevel: SeverityLevel
     val debt: Debt
-    val entity: Entity
-    val correctable: Boolean
-    val rule: RuleInfo
-}
-
-interface Entity {
-    val id: String
     val ktElement: KtElement
+    val autoCorrectable: Boolean
+    val rule: RuleInfo
 }
 
 interface RuleInfo {
@@ -31,13 +25,7 @@ interface RuleInfo {
 interface Location {
     val source: SourceLocation
     val text: TextLocation
-    val filePath: FilePath
-}
-
-interface FilePath {
-    val absolute: Path
-    val relative: Path?
-    val base: Path?
+    val filePath: FilePath // TODO The api should not depend on io.github.detekt.psi.FilePath
 }
 
 interface SourceLocation {
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
@@ -1,7 +1,9 @@
 package io.gitlab.arturbosch.detekt.api.v2
 
+import io.gitlab.arturbosch.detekt.api.Debt
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.SeverityLevel
 import org.jetbrains.kotlin.psi.KtFile
 
 // This will be a sealed interface in 1.5
@@ -23,7 +25,7 @@ interface TypeSolvingRule : Rule, (KtFile, ResolvedContext) -> List<NewIssue> {
 interface NewIssue {
     val entity: Entity
     val message: String
-    // This should be here but the code is not ready yet
-    //val severity: SeverityLevel
-    //val debt: Debt
+    val severity: SeverityLevel
+    val debt: Debt
+    val autoCorrectable: Boolean
 }
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
@@ -1,0 +1,31 @@
+package io.gitlab.arturbosch.detekt.api.v2
+
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.internal.CompilerResources
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.resolve.BindingContext
+
+// This will be a sealed interface in 1.5
+interface Rule {
+    // This should not be here but the code is not ready yet to remove it
+    val issue: Issue
+}
+
+interface PlainRule : Rule, (KtFile) -> List<NewIssue> {
+
+    override fun invoke(file: KtFile): List<NewIssue>
+}
+
+interface TypeSolvingRule : Rule, (KtFile, BindingContext, CompilerResources) -> List<NewIssue> {
+
+    override fun invoke(file: KtFile, binding: BindingContext, resources: CompilerResources): List<NewIssue>
+}
+
+interface NewIssue {
+    val entity: Entity
+    val message: String
+    // This should be here but the code is not ready yet
+    //val severity: SeverityLevel
+    //val debt: Debt
+}
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
@@ -22,7 +22,7 @@ interface TypeSolvingRule : Rule, (KtFile, ResolvedContext) -> List<Issue> {
 }
 
 interface Issue {
-    val entity: Entity
+    val entity: Entity // TODO We should rethink this class. I'm using the old because I don't want to address this yet
     val message: String
     val severity: SeverityLevel
     val debt: Debt
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
@@ -11,17 +11,17 @@ interface Rule {
     val description: String
 }
 
-interface PlainRule : Rule, (KtFile) -> List<NewIssue> {
+interface PlainRule : Rule, (KtFile) -> List<Issue> {
 
-    override fun invoke(file: KtFile): List<NewIssue>
+    override fun invoke(file: KtFile): List<Issue>
 }
 
-interface TypeSolvingRule : Rule, (KtFile, ResolvedContext) -> List<NewIssue> {
+interface TypeSolvingRule : Rule, (KtFile, ResolvedContext) -> List<Issue> {
 
-    override fun invoke(file: KtFile, resolvedContext: ResolvedContext): List<NewIssue>
+    override fun invoke(file: KtFile, resolvedContext: ResolvedContext): List<Issue>
 }
 
-interface NewIssue {
+interface Issue {
     val entity: Entity
     val message: String
     val severity: SeverityLevel
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
@@ -2,9 +2,7 @@ package io.gitlab.arturbosch.detekt.api.v2
 
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
-import io.gitlab.arturbosch.detekt.api.internal.CompilerResources
 import org.jetbrains.kotlin.psi.KtFile
-import org.jetbrains.kotlin.resolve.BindingContext
 
 // This will be a sealed interface in 1.5
 interface Rule {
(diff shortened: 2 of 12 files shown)


BraisGabin:
How can I get some comments here? Do you want me to finish the PR completely first? Do you prefer me to close this PR because we don't want to change the Rule API right now? If you don't like anything of this PR, please, just tell it and I'll close it :)

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/NewRule.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import io.gitlab.arturbosch.detekt.api.internal.CompilerResources
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.resolve.BindingContext
+
+interface NewRule : (KtFile, BindingContext, CompilerResources) -> List<NewIssue> {
+
+    val issue: Issue
+
+    override fun invoke(file: KtFile, binding: BindingContext, resources: CompilerResources): List<NewIssue>
```
arturbosch, line 11: I like it's simplicity, being a function makes it clearer that a Rule should not mutate any state.
We do have `SetupContext` which we can pass here to be able to access `Config` and `Logging`.

arturbosch: Having a Rule interface not being a visitor is actually really good to get rid of the difference between a `Rule` and a `MultiRule`.
This just returns a list of issues.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/NewRule.kt
```diff
@@ -0,0 +1,17 @@
...
+}
+
+data class NewIssue(
```
arturbosch, line 14: What were your motivations for going with a simpler `Issue` class here?

On the one hand I created the original `Finding` interface with the intent to allow to report more complex findings with additional entities like locations in code where the control flow must be taken to lead to a NPE or similar :).
On the other hand we do not use any of this flexibility.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleCollectionProvider.kt
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * A rule collection provider, as the name states, is responsible for creating rule collection.
+ *
+ * When writing own rule set providers make sure to register them according the ServiceLoader documentation.
+ * http://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html
+ */
+interface RuleCollectionProvider {
+
+    // TODO Can we remove this?
+    val ruleSetId: String
```
arturbosch, line 12: This is a good question.
If we really move everything which is rule related in our current API like filtering and suppression and rule set ids outside the `Rule` class and make it optional/post processing step and let the user choose rule set names on his own, this can go, right?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -111,26 +117,50 @@ internal class Analyzer(
...
-        executeRules(otherRules)
+        executeRules(correctableRules.map(::BaseRuleToNewRule) + correctableRules2)
+        executeRules(otherRules.map(::BaseRuleToNewRule) + otherRules2)
 
         return result
     }
 }
 
+private class BaseRuleToNewRule(private val rule: BaseRule) : NewRule {
+    override val issue: Issue
```
arturbosch, line 150: Hm, we have defined an `Issue` to "represent a problem in the codebase".
What if a `Rule` does not need to find only one kind of issues?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleCollectionProvider.kt
```diff
@@ -0,0 +1,20 @@
+package io.gitlab.arturbosch.detekt.api
+
+/**
+ * A rule collection provider, as the name states, is responsible for creating rule collection.
+ *
+ * When writing own rule set providers make sure to register them according the ServiceLoader documentation.
+ * http://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html
+ */
+interface RuleCollectionProvider {
```
arturbosch, line 9: Ideally a `RuleProvider` should be an `Extension`?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/NewRule.kt
```diff
@@ -0,0 +1,17 @@
+package io.gitlab.arturbosch.detekt.api
+
+import io.gitlab.arturbosch.detekt.api.internal.CompilerResources
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.resolve.BindingContext
+
+interface NewRule : (KtFile, BindingContext, CompilerResources) -> List<NewIssue> {
```
cortinico, line 7: I really like the functional interface that this PR is introducing. Should we consider making the rule interface a `fun interface`? That would make easier to create inline rules (I can't think of a use case for that just yet, but could be helpful for testing or so).

cortinico (commented):
Dropping my thoughts here as I haven't review this yet. +1 for the RFC 👍 

> * But as you can see the BaseRule doesn't know about configuration so all the configurations should be injected to the rule.

I was also wondering if the configuration should be added as one of the parameter to the `invoke()` function. Ideally all the required information should be available as a parameter once the rule is running

As for the deprecation cycle, I would just include this in Detekt 2.x as a breaking change. I don't think that supporting both interfaces really add much value (we could still do it if it's not too much of a pain).

BraisGabin:
Thanks for reviewing it!

> I was also wondering if the configuration should be added as one of the parameter to the `invoke()` function. Ideally all the required information should be available as a parameter once the rule is running

The problem with that is that we lose type safety. The current configuration is just a Map<String, Any> and we hope to have the values we want there at compilation time. If instead of that we force it to be injected you can use a map if you want, but you un one argument per configuration so you know that you get the desired types.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/NewRule.kt
BraisGabin: Good idea!

cortinico:
> The problem with that is that we lose type safety. The current configuration is just a Map<String, Any> and we hope to have the values we want there at compilation time. If instead of that we force it to be injected you can use a map if you want, but you un one argument per configuration so you know that you get the desired types.

Yup fair point. Although, type safety on the configuration is a nice-to-have (as you said our status quo is just a `Map<String, Any>`) I think it's valuable to have it in the upcoming Rules API

BraisGabin:
I didn't update this PR for long time because I had a problem and I think that, finally, I found a solution for it:

- A lot of users say that detekt is fast and they value it more than other tools (android lint) for that reason.
- We know that as soon as we enable type solving by default we will lose a lot of that performance. But we want to do this because we want better rules.

So I was thinking of a way to allow both worlds to live together: fast but more basic rules and slower but more and better rules. The problem is that, right now, the core can't know if a rule needs or not the `BindingContext` so we had little options here. My solution: `sealed interfaces` (coming in 1.5, but we don't need to wait for them):

```kotlin
sealed interface NewRule {
    val issue: Issue
}

interface PlainRule : NewRule, (KtFile) -> List<NewIssue> {

    override fun invoke(file: KtFile): List<NewIssue>
}

interface TypeSolvingRule : NewRule, (KtFile, BindingContext, CompilerResources) -> List<NewIssue> {

    override fun invoke(file: KtFile, binding: BindingContext, resources: CompilerResources): List<NewIssue>
}
```

This allow us to remove the annotation in or kdoc too. Now the type of the rule will tell you if it needs Type Solving or not.

With this we can do a smart thing in the core: We instantiate all the active rules and if all of the active rules are `PlainRule` we don't need to build a `BindingContext` so 🎉 both worlds can work toghether.

And the good part is that the `sealed interface`s allow us to do this without anything hacky! But until they arrive we will need to do some hacky things.

What do you think?

chao2zhang:
> * A lot of users say that detekt is fast and they value it more than other tools (android lint) for that reason.
> * We know that as soon as we enable type solving by default we will lose a lot of that performance. But we want to do this because we want better rules.

We can reimplement our checks in the detekt-compiler-plugin so that we do not perform kotlin compilation pass + detekt check pass.

> With this we can do a smart thing in the core: We instantiate all the active rules and if all of the active rules are `PlainRule` we don't need to build a `BindingContext` so 🎉 both worlds can work together.

🎉  I believe this is also optimization for auto correctible rules.

cortinico:
> What do you think?

That sounds like a nice improvement.

If we wish to draft something, `sealed interfaces` are actually available in `1.4.30-RC` with `-language-version 1.5`. See https://github.com/JetBrains/kotlin/releases/tag/v1.4.30-RC

I like the separation between the two words, as it really allows to address needs from two different type of customers. I'm wondering how the user-facing API will look like. I think it will be annoying for users to have to check the source code to understand if a rule is a `PlainRule` or a `TypeSolvingRule`. Perhaps they should be grouped into two distinct rule sets? 🤔

BraisGabin:
> As for the deprecation cycle, I would just include this in Detekt 2.x as a breaking change. I don't think that supporting both interfaces really add much value (we could still do it if it's not too much of a pain).

I just follow your advice and now all feels much, much better.

You can see the new `Analyzer`. My idea is to make the detekt core reactive. This way we can make it run in parallel more easily.

We should rethink a bit the `FileProcessListener`. Could we remove this callback?

```kotlin
fun onStart(files: List<KtFile>, binding: BindingContext, resources: CompilerResources)
```

It forces us to read ALL the files before we can start running any rule. I think that we can remove it because if you want to get all the rules you can use `onProcess`.

Any feedback about this new implementation of the core is more than welcome. Next things that I want to work on:
- How to implement the excludes, filters and suppression. (I think that the baseline supression should be handled down stream)
- We have `Issue`s, `Finding`s and `DetektResult`s. I want to review all that part and see if it could be simplified.

Other ideas:
- Implement a logger so the rules and the core could display more information if the users wants to. In the clients we will have a flag to define the level of log that the users wants: error, warning (default), verbose, debug, info.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
```diff
@@ -0,0 +1,31 @@
...
+}
+
+interface PlainRule : Rule, (KtFile) -> List<NewIssue> {
+
+    override fun invoke(file: KtFile): List<NewIssue>
+}
+
+interface TypeSolvingRule : Rule, (KtFile, BindingContext, CompilerResources) -> List<NewIssue> {
+
+    override fun invoke(file: KtFile, binding: BindingContext, resources: CompilerResources): List<NewIssue>
```
chao2zhang, line 22: Do you think we can combine `BindingContext` with `CompilerResources` into a single model? (I would suggest the name `ResolvedContext`) This would allow us to simplify the interfaces and avoid future breaking changes if there are more environment details to be added.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/RuleCollectionProvider.kt
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * A rule collection provider, as the name states, is responsible for creating rule collection.
+ *
+ * When writing own rule set providers make sure to register them according the ServiceLoader documentation.
+ * http://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html
+ */
+interface RuleCollectionProvider {
+
+    // TODO Can we remove this?
```
chao2zhang, line 13: We probably want to keep an identifier of the rule collection / rule set, thus I think it is reasonable to keep it.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/v2/Analyzer.kt
```diff
@@ -0,0 +1,184 @@
...
+    }
+}
+
+@FlowPreview
+private fun myAnalyze(
+    rules: Flow<Rule>,
+    files: Flow<KtFile>,
+    fileProcessListeners: Flow<FileProcessListener>,
+    bindingContext: Deferred<BindingContext>,
+    compilerResources: Deferred<CompilerResources>,
```
chao2zhang, line 60: Really like these Deferred usage to delay BindingContext or CompilerResources until they are needed :D No need to handcraft logic.

One thing I would call out is that we should probably prompt a warning when detekt plain is run but some type resolving rules are active.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/v2/Utils.kt
```diff
@@ -0,0 +1,26 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onCompletion
+import kotlinx.coroutines.flow.shareIn
+import kotlinx.coroutines.flow.takeWhile
+
+suspend fun <T> Flow<T>.reusable(
```
chao2zhang, line 12: Could you elaborate on why we need the extra `map, onCompletion, takeWhile and map`.  Would a simple `shareIn()` fits our need?

I have been scratching my head and I don't know enough about exception handling in flow.

BraisGabin: This is not related with exception handling at all. The problem is that `shareIn` returns a `SharedFlow` and the problem is that a [`SharedFlow` never ends](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-shared-flow/). So I need to workaround this with a `takeWhile` and emit some special item to know when to end it. I don't have too much experience with `Flow` so I don't know what's the better way to implement this.

chao2zhang: Well I am studying Flow like you (Since the project I am working on recently started to prototype flow). Now I seem to understand the intention here since we are emitting a failure inside `onCompletion{}` to stop the shared flow. I assume we are following the doc `flow-truncating operators like Flow.take and Flow.takeWhile can be used on a shared flow to turn it into a completing one.`

BraisGabin: Didn't know about that, I'll look at that documentation. I'm not very proud of this implementation. It works but it's odd.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/v2/Analyzer.kt
BraisGabin: Agree we need to find a way to warn about this.

BraisGabin:
Ok, new iteration!

This time I implemented the `Finding` interface. It's far from perfect and I need to work the report part of the process to check if I'm missing somethig here.

The basic idea here is that the rules returns a list of `Issue`s. And then the core "hydrates" those `Issue`s to convert them into `Finding`s. The `Finding`s are the instances that are sended to the file processors and the reports.

`Issue` -> easy to instantiate (so easier code in rules)
`Finding` -> easy to use and extract information from it (so easier code in post processors and report generators)

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
```diff
@@ -0,0 +1,30 @@
...
+    override fun invoke(file: KtFile): List<Issue>
+}
+
+interface TypeSolvingRule : Rule, (KtFile, ResolvedContext) -> List<Issue> {
+
+    override fun invoke(file: KtFile, resolvedContext: ResolvedContext): List<Issue>
+}
+
+interface Issue {
+    val entity: Entity // TODO We should rethink this class. I'm using the old because I don't want to address this yet
```
chao2zhang, line 25: `Entity` is contextual and makes sense as a property of `Finding`. What do you think?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/v2/Filter.kt
```diff
@@ -0,0 +1,10 @@
+package io.gitlab.arturbosch.detekt.core.v2
+
+import io.gitlab.arturbosch.detekt.api.v2.Issue
+import java.nio.file.Path
+
+interface Filter {
```
chao2zhang, line 6: What's the purpose of `Filter`?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Finding.kt
```diff
@@ -0,0 +1,71 @@
...
+    val autoCorrectable: Boolean
+    val rule: RuleInfo
+}
+
+interface RuleInfo {
+    val id: String
+    val ruleSetId: String
+    val description: String
+}
+
```
chao2zhang, line 24: I am wondering about the necessity of this interface. It seems to be to keep a `Rule` property in `Finding`, and we do not need the extra `ruleSetId` as we should be able to query the ruleSetId given the ruleId externally.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/v2/Filter.kt
BraisGabin: Core use it for two things related with filtering. For each rule we have its filter. You can check in the analyzer that we receive a Flow of `Pair<Rule, Filter>`. It does two kind of filters:
- Given a path know if we should run the rule or not (this way we can implment the exluce/include logic)
- Given an issue raised by the rule know if we should suppress it or not: Right now it will only implement the `@Suppress` logic. But we can extend this to implement the generic suppression that we talk in some other issues. Suppress all the functions with specific name, or with an annotation, or the functions inside a class with an specific name... This interface allow us to expand the filtering.

Maybe this is not the best name, any ideas are more than welcome.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
BraisGabin: I think that entity have far too much information on it. I think that we can refine it a bit so it's easier to report and then split the different parts of it in the Finding instead of having all in one. For example, the position of a finding I think that should be directly in the finding. And there is no need to know that it is inside the entity.

I mean, `finding.position` is better than `finding.entity.position`.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Finding.kt
BraisGabin: I'm not sure here. I'll remove it in the next iteration. If we end up needing it we can add it again.

BraisGabin:
Closing this in favor of: #3651

PR #3132, opened 2020-10-06
Support sarif as a report type - #3045

arturbosch:
New optional report format which can be plugged via `--plugins detekt-report-sarif.jar`:

Output preview via the VSCode extension:
2020-10-06-153635_927x499_scrot
2020-10-06-153629_932x514_scrot

CI will fail due to not a published version of sarif4j. We need to wait until #3045 gets an answer from the sarif authors.

### DIFF
--- a/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/RuleDescriptors.kt
+++ b/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/RuleDescriptors.kt
@@ -36,9 +36,6 @@ fun MultiRule.toDescriptors(ruleSetId: RuleSetId): List<ReportingDescriptor> =
 fun Rule.toDescriptor(ruleSetId: RuleSetId): ReportingDescriptor = descriptor {
     id = "detekt.$ruleSetId.$ruleId"
     name = "$ruleSetId>$ruleId"
-    shortDescription = MultiformatMessageString().apply {
-        text = issue.description.substringBefore(System.lineSeparator())
-    }
-    fullDescription = MultiformatMessageString().apply { text = issue.description }
+    shortDescription = MultiformatMessageString().apply { text = issue.description }
     helpUri = URI.create("https://detekt.github.io/detekt/${ruleSetId.toLowerCase()}.html#${ruleId.toLowerCase()}")
 }
--- a/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/RuleDescriptors.kt
+++ b/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/RuleDescriptors.kt
@@ -1,0 +1,44 @@
+package io.github.detekt.report.sarif
+
+import io.github.detekt.sarif4j.MultiformatMessageString
+import io.github.detekt.sarif4j.ReportingDescriptor
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.MultiRule
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.RuleSetId
+import io.gitlab.arturbosch.detekt.api.RuleSetProvider
+import java.net.URI
+import java.util.ServiceLoader
+
+fun ruleDescriptors(config: Config): HashMap<String, ReportingDescriptor> {
+    val sets = ServiceLoader.load(RuleSetProvider::class.java)
+        .map { it.instance(config.subConfig(it.ruleSetId)) }
+    val descriptors = HashMap<String, ReportingDescriptor>()
+    for (ruleSet in sets) {
+        for (rule in ruleSet.rules) {
+            if (rule is MultiRule) {
+                descriptors.putAll(rule.toDescriptors(ruleSet.id).associateBy { it.name })
+            } else {
+                assert(rule is Rule)
+                val descriptor = (rule as Rule).toDescriptor(ruleSet.id)
+                descriptors[descriptor.name] = descriptor
+            }
+        }
+    }
+    return descriptors
+}
+
+fun descriptor(init: ReportingDescriptor.() -> Unit) = ReportingDescriptor().apply(init)
+
+fun MultiRule.toDescriptors(ruleSetId: RuleSetId): List<ReportingDescriptor> =
+    this.rules.map { it.toDescriptor(ruleSetId) }
+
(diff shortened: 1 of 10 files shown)


detekt-report-sarif/src/test/kotlin/io/github/detekt/report/sarif/SarifOutputReportSpec.kt
```diff
@@ -0,0 +1,38 @@
...
+
+        it("renders multiple issues") {
+            val result = TestDetektion(
+                createFinding(ruleName = "TestSmellA"),
+                createFinding(ruleName = "TestSmellB"),
+                createFinding(ruleName = "TestSmellC")
+            )
+
+            val report = SarifOutputReport().apply { init(EmptySetupContext()) }
+            val json = JsonPath.from(report.render(result))
```
BraisGabin, line 24: Shouldn't be better to save the expected json in a file and then assert that they are the same as json?

BraisGabin (approved):

schalkms:
I just skimmed through the code change list. Hence, I only have some general questions.
Is there a roadmap for the Java support? I'm not sure whether reinventing the wheel by implementing the SARIF support in Kotlin is the right approach here?

arturbosch:
> I just skimmed through the code change list. Hence, I only have some general questions.
> Is there a roadmap for the Java support? I'm not sure whether reinventing the wheel by implementing the SARIF support in Kotlin is the right approach here?

When the official MS java support will be out, we can easily migrate to it.
We do not reinvent any wheels here, just generating POJOs :).
Please also see my comment in https://github.com/detekt/detekt/issues/3045#issuecomment-711062052.

schalkms (approved):
Good to know. Thanks for the explanation. I really appreciate your work for supporting this feature. For me this PR looks fine.

detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifDsl.kt
```diff
@@ -0,0 +1,49 @@
...
+fun tool(init: Tool.() -> Unit): Tool = Tool().apply(init)
+
+fun component(init: ToolComponent.() -> Unit): ToolComponent = ToolComponent().apply(init)
+
+fun SarifSchema210.withDetektRun(config: Config, init: Run.() -> Unit) {
+    runs.add(
+        Run().apply {
+            tool = tool {
+                driver = component {
+                    guid = "022ca8c2-f6a2-4c95-b107-bb72c43263f3"
```
cortinico, line 34: is this randomly generated?

detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/RuleDescriptors.kt
```diff
@@ -0,0 +1,44 @@
...
+    val descriptors = HashMap<String, ReportingDescriptor>()
+    for (ruleSet in sets) {
+        for (rule in ruleSet.rules) {
+            if (rule is MultiRule) {
+                descriptors.putAll(rule.toDescriptors(ruleSet.id).associateBy { it.name })
+            } else {
+                assert(rule is Rule)
+                val descriptor = (rule as Rule).toDescriptor(ruleSet.id)
+                descriptors[descriptor.name] = descriptor
+            }
```
cortinico, line 25: Maybe this can be refactored with a `when(rule) {}`?

michaelcfanning:
@arturbosch, just checking in on this, as there's strong interest internally at Microsoft for consuming this new report format. Any idea when it might be merged/available?

arturbosch:
> @arturbosch, just checking in on this, as there's strong interest internally at Microsoft for consuming this new report format. Any idea when it might be merged/available?

Hey @michaelcfanning , thanks for interest and the ping.
I've worked on this further locally. I have one last question regarding default -1 values in the scheme which should not appear in the report. I will keep you updated in #3132 later this week.
There are so many good rule improvements in the main branch so I really should get up to speed to finish this and release 1.15.0 xD.

michaelcfanning:
@arturbosch, glad to help w/any SARIF advice. The large # of optional properties/defaults is definitely a concern depending on your JSON mechanism. The C# SARIF SDK uses a library (Newtonsoft JSON.NET) that automatically elides/hydrates these on save/open. We autogenerated the C# OM that has appropriate attributes for this by writing a [custom tool, 'JsonSchema.ToDotNet'](https://github.com/microsoft/jschema/tree/master/src) that parses the SARIF JSON schema. We could consider a similar trick in your scenario, if your JSON serializer supports similar features.

I took Larry's suggestion and have opened an issue to write a validator that flags all unnecessarily rendered default values. There's a lot of surface here to examine...

Something to keep in mind, though: SARIF which contains this data is valid. You might consider hard-coding eliminating obvious things for now, while considering a more systematic/comprehensive solution.

detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifDsl.kt
arturbosch: Yes, I generated it randomly once with `UUID.getXXX()`.
As I've understand it, the guid needs to be the same between all detekt 1.x.y runs.
Correct me if I'm wrong here @michaelcfanning

detekt-report-sarif/src/test/kotlin/io/github/detekt/report/sarif/SarifOutputReportSpec.kt
arturbosch: Done, thanks!

arturbosch:
> @arturbosch, glad to help w/any SARIF advice. The large # of optional properties/defaults is definitely a concern depending on your JSON mechanism. The C# SARIF SDK uses a library (Newtonsoft JSON.NET) that automatically elides/hydrates these on save/open. We autogenerated the C# OM that has appropriate attributes for this by writing a [custom tool, 'JsonSchema.ToDotNet'](https://github.com/microsoft/jschema/tree/master/src) that parses the SARIF JSON schema. We could consider a similar trick in your scenario, if your JSON serializer supports similar features.
> 
> I took Larry's suggestion and have opened an issue to write a validator that flags all unnecessarily rendered default values. There's a lot of surface here to examine...
> 
> Something to keep in mind, though: SARIF which contains this data is valid. You might consider hard-coding eliminating obvious things for now, while considering a more systematic/comprehensive solution.

@michaelcfanning thanks for the advise.
I've switcher to Jackson which allowed to preserve the order of properties and eliminate default values from the report.

PR #3139, opened 2020-10-09
UnusedPrivateMember: fix false positive with getValue/setValue operator functions

### LINKED ISSUE #3128: False positive UnusedPrivateMember for getValue, setValue operator functions
## Expected Behavior
Private property delegation operators are marked as used when they are used.

## Observed Behavior
Property delegation operator `by` (`getValue`, `setValue`) has false positives for `UnusedPrivateMember`.

## Steps to Reproduce
I've created a unit test that fits into `UnusedPrivateMemberSpec` that triggers the false positive. Looking through the other tests, the public property should be enough to be considered used.
```
        it("Does not report used getValue, setValue") {
            val code = """
                class Test {
                    var delegated by "Hello"
                }
                
                private operator fun String.getValue(test: Test, prop: KProperty<*>): String {
                    return "working"
                }
                
                private operator fun String.setValue(test: Test, prop: KProperty<*>, value: String) {
                    error("setValue")
                }
            """.trimIndent()

            assertThat(subject.compileAndLintWithContext(env, code)).hasSize(0)
        }
```

## Context
I originally wanted to create delegate methods for `ThreadLocal` scoped to within my current file so I could use it like a regular variable, but after updating detekt it's now being flagged as an unused member.

It is not being flagged in 1.5.1,  b190e11a

Bisect shows it was introduced here with the unused operator overload detection: 38c54c80d67b92

## Your Environment
* Version of detekt used: 1.14.1,  782fb75b49
* Version of Gradle used (if applicable): gradlew in repo (Gradle 6.7-rc-2)
* Operating System and version: Artic Linux 5.8.10-artix1-1

t-kameyama:
Fixes #3128

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
@@ -40,6 +40,7 @@ import org.jetbrains.kotlin.psi.psiUtil.isPrivate
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
 import org.jetbrains.kotlin.types.expressions.OperatorConventions
+import org.jetbrains.kotlin.util.OperatorNameConventions
 
 /**
  * Reports unused private properties, function parameters and functions.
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -86,8 +89,12 @@ private class UnusedFunctionVisitor(
 
     private val functionDeclarations = mutableMapOf<String, MutableList<KtFunction>>()
     private val functionReferences = mutableMapOf<String, MutableList<KtReferenceExpression>>()
+    private val propertyDelegates = mutableListOf<KtPropertyDelegate>()
 
     override fun getUnusedReports(issue: Issue): List<CodeSmell> {
+        val propertyDelegateResultingDescriptors by lazy {
```
BraisGabin, line 95: If you use lazy, use the none policy. You don't need concurrent safity here.

BraisGabin (approved):

schalkms (approved):

Issue #3141, opened 2020-10-10
ClassOrdering only reports first misorder, not all misorders


hbmartin:
## Expected Behavior
ClassOrdering should report all misorderings in a class

## Observed Behavior
ClassOrdering only reports the first misordering in a class

## Steps to Reproduce
This can be easily observed by changing a test in `ClassOrderingSpec` to have multiple misorders, but still pass the assertions that the list of findings only has 1 item.

## Context
As a user, I expect that fixing all errors from running detekt will result in the next run succeeding. This may not be the case now, as fixing a `ClassOrdering` violation may reveal further violations in a subsequent run.

## Your Environment

* Version of detekt used:
* Version of Gradle used (if applicable):
* Operating System and version:
* Link to your project (if it's a public repository):

@BraisGabin please assign to me

schalkms:
Closed by a714f54

PR #3138, opened 2020-10-09
ClassOrdering reports a message describing the misorder

hbmartin:
* Currently the `ClassOrdering` rule reports its own description as a message. This PR modifies the message to report the specific entities identified as out of order.
* It also slightly refactors the class organization, but requires no test changes, only test changes exist to check message
* Additional tests (for test coverage sake) describe existing behavior
* Future PR could improve this by reporting all misordered pairs, not simply the first one: https://github.com/detekt/detekt/issues/3141
* Follow up to: https://github.com/detekt/detekt/pull/3088

PR #3142, opened 2020-10-11
ClassOrdering rule reports a list of errors

hbmartin:
* Addresses https://github.com/detekt/detekt/issues/3141 where only the first misorder in a class is reported
* Also report entity where error occurred, not just parent class
* Both of these changes might break existing baselines? - not sure how to address this
* Follow up to https://github.com/detekt/detekt/pull/3138

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
@@ -68,22 +68,27 @@ class ClassOrdering(config: Config = Config.empty) : Rule(config) {
     override fun visitClassBody(classBody: KtClassBody) {
         super.visitClassBody(classBody)
 
-        comparator.findOutOfOrder(classBody.declarations)?.let {
-            report(CodeSmell(issue, Entity.from(classBody), generateMessage(it)))
+        comparator.findOutOfOrder(classBody.declarations).letNotEmpty { misorders ->
+            report(
+                misorders.map {
+                    CodeSmell(
+                        issue = issue,
+                        entity = Entity.from(it.first),
+                        message = "${it.first.description} should not come before ${it.second.description}",
+                        references = listOf(Entity.from(classBody))
+                    )
+                }
+            )
         }
-    }
-
-    private fun generateMessage(misordered: Pair<KtDeclaration, KtDeclaration>): String {
-        return "${misordered.first.description} should not come before ${misordered.second.description}"
     }
 }
 
 private fun Comparator<KtDeclaration>.findOutOfOrder(
     declarations: List<KtDeclaration>
-): Pair<KtDeclaration, KtDeclaration>? {
-    declarations.zipWithNext { a, b -> if (compare(a, b) > 0) return Pair(a, b) }
-    return null
-}
+): List<Pair<KtDeclaration, KtDeclaration>> =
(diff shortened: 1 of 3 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
```diff
@@ -68,22 +68,27 @@ class ClassOrdering(config: Config = Config.empty) : Rule(config) {
     override fun visitClassBody(classBody: KtClassBody) {
         super.visitClassBody(classBody)
 
-        comparator.findOutOfOrder(classBody.declarations)?.let {
-            report(CodeSmell(issue, Entity.from(classBody), generateMessage(it)))
+        comparator.findOutOfOrder(classBody.declarations).letNotEmpty { misorders ->
```
schalkms, line 71: I think a `if (comparator.findOutOfOrder(classBody.declarations).isNotEmpty)` would be simpler and easier to read.
What do you think?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
```diff
@@ -68,22 +68,27 @@ class ClassOrdering(config: Config = Config.empty) : Rule(config) {
...
+            report(
+                misorders.map {
+                    CodeSmell(
+                        issue = issue,
+                        entity = Entity.from(it.first),
+                        message = "${it.first.description} should not come before ${it.second.description}",
+                        references = listOf(Entity.from(classBody))
+                    )
```
schalkms, line 79: An assert statement that checks the reported message would be great in order to ensure the correct behavior.

schalkms (commented):
Thanks for addressing this issue and further improving this rule!
Please see my attached comments in regards to two minor improvements.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
hbmartin: @schalkms went back and forth on this, so I'll take your lead: 384666887da9e33395fb8999cecfba296823af71

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
hbmartin: @schalkms good point, tested each error message in order: 384666887da9e33395fb8999cecfba296823af71

schalkms (approved):
Good one!

BraisGabin:
I'm not sure if this is the best way to report all the errors. I mean, we can be reporting more than the actual movements that you need to do. This is the same problem as find the minimum insertions to sort an array ([related](https://www.geeksforgeeks.org/minimum-insertions-sort-array/) [links](https://stackoverflow.com/a/20394497/842697))

I'm not against merging this because it's better than the current implementation but I think that we should keep an issue to track this and implement it to report the proper number of issues.

hbmartin:
@BraisGabin made an issue to track your comment: https://github.com/detekt/detekt/issues/3144 , please assign to me 😃

BraisGabin (approved):

PR #3149, opened 2020-10-17
Remove obsolete api documentation entries

arturbosch:
dokka deleted them.


arturbosch:
Seems like changing the jdk changes the generated documentation ...

schalkms (approved):
That’s interesting. Why does another configured JDK version produce a different output?

BraisGabin (approved):

arturbosch:
Our CI runs with jdk14. Using 14 locally does not change the docs. 
Using jdk11 does.
Should we consider building and verifying the docs with Java11 / always use LTS ?
Which java version do you guys use?
It is indeed a small inconvenience for our contributors.
However no one yet complained about it.

BraisGabin:
I use JDK8. I had this problem in my last 2 PRs but I didn't commit those changes because I readed this PR first.

I couldn't find any related issue in dokka. Should we open one?

arturbosch:
> I use JDK8. I had this problem in my last 2 PRs but I didn't commit those changes because I readed this PR first.
> 
> I couldn't find any related issue in dokka. Should we open one?

Yes, I've opened https://github.com/Kotlin/dokka/issues/1577 for this ,thanks!

PR #3150, opened 2020-10-18
Const naming

BraisGabin:
This is just a reopen for #2017 (I don't know why I can't reopen it). Extracted from there: 

> There are places where we allow to use: thisFormat or THIS_FORMAT but we should never allow to use This_Format. This PR fix this issue in the 3 places where we allow the use of _ in the naming: enums, object properties and top level properties.

Why am I reopening this?

After @arturbosch closed that PR I opened this issue to IntelliJ: https://youtrack.jetbrains.com/issue/KT-34508 It was accepted as a valid issue but after one year it's not fixed.

I think that we should rethink about merging this.

~We have another related issue: `private val _HELLO` should not be valid and we are allowing it. I'm going to create another PR fixing it.~

### DIFF
--- a/detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ObjectPropertyNamingSpec.kt
+++ b/detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ObjectPropertyNamingSpec.kt
@@ -149,13 +149,26 @@ class ObjectPropertyNamingSpec : Spek({
             assertThat(subject.compileAndLint(code)).isEmpty()
         }
 
-        it("should detect private variables not complying to the naming rules") {
+        it("should detect constants not complying to the naming rules") {
+            val code = """
+                object O {
+                    val MyNAME = "Artur"
+                    val n_a_m_e = "Artur"
+                }
+            """
+            assertThat(subject.compileAndLint(code)).hasSize(2)
+        }
+
+        it("should detect private constants not complying to the naming rules") {
             val code = """
                 object O {
                     private val __NAME = "Artur"
+                    private val MyNAME = "Artur"
+                    private val n_a_m_e = "Artur"
+                    private val _MY_NAME = "Artur"
                 }
             """
-            assertThat(subject.compileAndLint(code)).hasSize(1)
+            assertThat(subject.compileAndLint(code)).hasSize(3)
         }
     }
 
@@ -216,7 +229,6 @@ abstract class NamingSnippet(private val isPrivate: Boolean, private val isConst
     val negative = """
                     ${visibility()}${const()}val MY_NAME_8 = "Artur"
                     ${visibility()}${const()}val MYNAME = "Artur"
-                    ${visibility()}${const()}val MyNAME = "Artur"
                     ${visibility()}${const()}val name = "Artur"
                     ${visibility()}${const()}val nAme = "Artur"
                     ${visibility()}${const()}val serialVersionUID = 42L"""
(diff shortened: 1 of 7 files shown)


detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ObjectPropertyNamingSpec.kt
```diff
@@ -149,13 +149,26 @@ class ObjectPropertyNamingSpec : Spek({
...
+            assertThat(subject.compileAndLint(code)).hasSize(2)
+        }
+
+        it("should detect private constants not complying to the naming rules") {
             val code = """
                 object O {
                     private val __NAME = "Artur"
+                    private val MyNAME = "Artur"
+                    private val n_a_m_e = "Artur"
+                    private val _MY_NAME = "Artur"
```
arturbosch, line 168: Hm, as long as IntelliJ does not change it's own regex I would not like to change this.
All this cases are okay in IntelliJ when I paste them and would cause a lot of issues for our users

arturbosch:
> This is just a reopen for #2017 (I don't know why I can't reopen it). Extracted from there:

Stale issues without comments get locked by our bot integration.

> 
> > There are places where we allow to use: thisFormat or THIS_FORMAT but we should never allow to use This_Format. This PR fix this issue in the 3 places where we allow the use of _ in the naming: enums, object properties and top level properties.
> 
> Why am I reopening this?
> 
> After @arturbosch closed that PR I opened this issue to IntelliJ: https://youtrack.jetbrains.com/issue/KT-34508 It was accepted as a valid issue but after one year it's not fixed.

I would really like to have IntelliJ change its default before doing such a change which will lead to new issues for our users.
> 
> I think that we should rethink about merging this.
> 
> ~We have another related issue: `private val _HELLO` should not be valid and we are allowing it. I'm going to create another PR fixing it.~

BraisGabin:
Right now our regex allow names like this: `private val OneProperty: String`. And that's clearly wrong and jetbrains agree with it. I tried to fix that in IntelliJ but I can't understand that project and how to run it's tests.

I think that our users expect that if someone in their teams adds a property with a name like the one in my example detekt will complain. I don't think that flag more code to our users is bad thing. They realay on detekt to find issues in its code.

I understand that it will be better if IntelliJ implements it too and for that reason I wait nearly a year to reopen this issue. But if they don't fix it I don't see why we should keep this issue in our rule too.

But, this change should be merged for a minor release and not a patch.

BraisGabin:
Could we take a decision about this update @cortinico @schalkms what do you think about this? I think that it could be merged for the 1.15 relase if you agree. Or close the topic if disagree.

schalkms:
In my opinion detekt should be compliant to IntelliJs rules.

cortinico:
> In my opinion detekt should be compliant to IntelliJs rules.

Agree.

I've 👍 -d your YouTrack ticket. Other than that, I believe we should try to align as much as possible to commonly used RegExes (like the one used in IntelliJ) unless we have a strong argument/case to change one. So far it seems this is a minor concern, and users that are interested in altering the inspection behavior to adhere to the official docs, can still update the Regex.

I believe if we end up waiting for JetBrains, we could still leave a note in the `*Naming` rules with a link to this issue or a mention of the alternate regex.

BraisGabin:
> I believe if we end up waiting for JetBrains, we could still leave a note in the `*Naming` rules with a link to this issue or a mention of the alternate regex.

I added those comments but they look kind of strange there so I think that we should close this and if we get a notificacion that the issue was fixed we will update it. Maybe I try again to fix it in the intellij code but I have no idea how to run the test there.

PR #3154, opened 2020-10-18
Fix two eager Regex creations in rules which may confuse users if this rules are inactive but evaluated

arturbosch:


schalkms (approved):

schalkms:
@arturbosch do you have samples for that?

> which may lead to confusing error messages on wrong config properties

arturbosch:
> @arturbosch do you have samples for that?
> 
> > which may lead to confusing error messages on wrong config properties

It prevents cases like mentioned in https://github.com/detekt/detekt/issues/1064

PR #3157, opened 2020-10-19
Do not report vararg arguments which are just passed to a vararg parameter 

### LINKED ISSUE #3145: False positive performance SpreadOperator in case of pass-through vararg
## Expected Behavior
When using the spread operator to pass on a vararg array from a vararg function implementation to a vararg function call, the spread operator should not be reported.

## Observed Behavior
The spread operator passing a vararg array from a vararg function implementation to a vararg function call is reported by Detekt as a performance issue because of the spread operator.

This is wrong because, in the case of passing vararg to vararg, the spread operator does not (anymore since Kotlin 1.1.60, 1.2-Beta2) create a new array and thus is not a performance issue. Refer to the following ticket on Kotlin: https://youtrack.jetbrains.com/issue/KT-20462

## Steps to Reproduce
Call one vararg method from another using the spread operator.
A popular example is to call `runApplication(args)` from `main(vararg args: String)` in Spring applications.

The following code should be accepted as correct by detekt:
```
package com.example

import org.springframework.boot.autoconfigure.SpringBootApplication
import org.springframework.boot.runApplication
import org.springframework.context.ApplicationContext

@SpringBootApplication
class ExampleApplication

lateinit var appContext: ApplicationContext
    private set

fun main(vararg args: String) {
    appContext = runApplication<ExampleApplication>(*args) // ⇐ Point of Interest
}
```

## Context
The impact is low.
The situation of passing through vararg is relatively rare.
It is easy to deal with the situation by using a `@SuppressWarnings("SpreadOperator")` annotation on the corresponding method.

## Your Environment

* Version of detekt used: 1.14.1
* Operating System and version: Kubuntu 20.04 LTS

arturbosch:
Closes #3145

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -139,6 +139,8 @@ style:
     active: true
   RedundantVisibilityModifierRule:
     active: true
+  ReturnCount:
+    excludeGuardClauses: true
   SpacingBetweenPackageAndImports:
     active: true
   UnnecessaryAbstractClass:
--- a/detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperator.kt
+++ b/detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/SpreadOperator.kt
@@ -9,8 +9,10 @@ import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import org.jetbrains.kotlin.descriptors.ConstructorDescriptor
 import org.jetbrains.kotlin.descriptors.ParameterDescriptor
+import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtValueArgument
 import org.jetbrains.kotlin.psi.KtValueArgumentList
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.CompileTimeConstantUtils
 import org.jetbrains.kotlin.resolve.DescriptorUtils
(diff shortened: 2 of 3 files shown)


BraisGabin (approved):

schalkms (commented):
In general I approve this PR. However, I wanted to raise the question, whether we want to support this rule without having type resolution enabled?

arturbosch:
> In general I approve this PR. However, I wanted to raise the question, whether we want to support this rule without having type resolution enabled?

As long as our compiler plugin is not 1.0 and the changes are easy, I would say yes :)

chris-tipton:
It appears this exception to the `SpreadOperator` rule still requires type resolution. Is that accurate?

cortinico:
> It appears this exception to the `SpreadOperator` rule still requires type resolution. Is that accurate?

Yes it's correct.
Moreover I'd like to point out the `still requires` you mentioned. In the future Detekt will evolve to have type resolution enabled by default. More and more rules will evolve to support type resolution (we  will not remove it).

chris-tipton:
> > It appears this exception to the `SpreadOperator` rule still requires type resolution. Is that accurate?
> 
> Yes it's correct.
> Moreover I'd like to point out the `still requires` you mentioned. In the future Detekt will evolve to have type resolution enabled by default. More and more rules will evolve to support type resolution (we will not remove it).

Yeah, I understand that is part of the plan for a 2.0 release. I just wanted to make sure that I wasn't missing anything based on the above comments about supporting this rule without type resolution. Our project currently does not use type resolution, so for the time being I am manually suppressing the rule for cases like this.

Thanks for your response!

cortinico:
> I just wanted to make sure that I wasn't missing anything based on the above comments about supporting this rule without type resolution

This is one of the rule with "mixed behavior" (see more on #2994). Our take is that we will keep on supporting them for the time being, but will slowly transition to offer more and more Type Resolution capabilities.

In other words: this rule could be converted to by a Type Resolution-only rule as part of our ongoing migration.

Issue #3170, opened 2020-10-25
False negative in IgnoredReturnValue


BraisGabin:
## Expected Behavior
A use of an annotated method with `@CheckReturnValue` at the end of an if expression should be flagged by `IgnoredReturnValue` if the output of the if is not used.

Same for the `when`.
## Steps to Reproduce
```kotlin
package test

import kotlin.random.Random

@CheckReturnValue
fun returnsInt() = 42

if (Random.nextBoolean()) {
    println("hello")
} else {
    returnsInt()
}
```

## Your Environment
* Version of detekt used: master (1.14.2)

PR #3174, opened 2020-10-26
Improves in IgnoredReturnValue

BraisGabin:
Last night I was working on #3170. I couldn't find the way to fix it but I did some refactors around the code/test that make it the code more maintenable and create the test that should be changed to show the error.

What do you think about merging the comment with `FIXME`? We have the error right now, this comment only points out it.


schalkms (commented):
I'm fine with it, since the modifications are an improvement (especially the unnecessary smart cast).

cortinico:
> What do you think about merging the comment with `FIXME`? We have the error right now, this comment only points out it.

Fine for me 👍  It makes clearer that we have that scenario to address.

cortinico (approved):

PR #3175, opened 2020-10-26
Suppress RedundantVisibilityModifierRule if explicit API mode enabled

### LINKED ISSUE #3125: Suppress RedundantVisibilityModifierRule if explicit API mode enabled

If explicit API mode is enabled:
```kotlin
kotlin {
    explicitApi()
}
```
Detekt should suppress `RedundantVisibilityModifierRule` for `public` modifier.

## Context

[Explicit API mode](https://kotlinlang.org/docs/reference/whatsnew14.html#explicit-api-mode-for-library-authors) was added in Kotlin 1.4
It prevents libraries' authors from making APIs public unintentionally. In this mode, you should explicitly specify the visibility modifier even if it is `public`.

schalkms:
Explicit API mode was added in Kotlin 1.4
It prevents libraries' authors from making APIs public unintentionally.
In this mode, the visibility modifier should be defined explicitly even if it is public.
See: https://kotlinlang.org/docs/reference/whatsnew14.html#explicit-api-mode-for-library-authors

Closes #3125

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantVisibilityModifierRule.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantVisibilityModifierRule.kt
@@ -24,6 +24,11 @@ import org.jetbrains.kotlin.psi.psiUtil.isPrivate
 
 /**
  * This rule checks for redundant visibility modifiers.
+
+ * One exemption is the
+ * [explicit API mode](https://kotlinlang.org/docs/reference/whatsnew14.html#explicit-api-mode-for-library-authors)
+ * In this mode, the visibility modifier should be defined explicitly even if it is public.
+ * Hence, the rule ignores the visibility modifiers in explicit API mode.
  *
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantVisibilityModifierRule.kt
```diff
@@ -53,11 +55,21 @@ class RedundantVisibilityModifierRule(config: Config = Config.empty) : Rule(conf
 
     private fun KtModifierListOwner.isExplicitlyPublic() = this.hasModifier(KtTokens.PUBLIC_KEYWORD)
 
+    /**
+     * Explicit API mode was added in Kotlin 1.4
```
cortinico, line 59: Maybe add a note in the KDoc on top of the file so it gets propagated in the generated documentation?

cortinico (approved):

chao2zhang:
Thank you for making this change. I meant to report a new issue but found that this is already addressed!

Issue #3177, opened 2020-10-26
 False Positive PreferToOverPairSyntax 


antoniorosario:
## Expected Behavior

A warning is not reported when a Pair is created by a function that uses the to syntax.

## Observed Behavior

A warning is reported when a Pair is created by a function that uses the to syntax.

## Steps to Reproduce

```kotlin
val pair = createPair()
fun createPair() = 1 to 2
```

Adding this test  
```kotlin
 it("does not report if it is created using a function that uses the to syntax") {
     val code = """
         val pair = createPair()
         fun createPair() = 1 to 2
     """

    assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
}
```
to [PreferToPairSyntaxSpec](https://github.com/detekt/detekt/blob/master/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/PreferToOverPairSyntaxSpec.kt) should result in a failing test.

## Context
I'm using a function to create a Pair in unit tests to avoid manually creating the same Pair in all places that this object is required.

## Your Environment

* Version of detekt used: 1.14.1
* Version of Gradle used (if applicable): 6.7
* Operating System and version: macOS Catalina 10.15.5
* Link to your project (if it's a public repository): N/A

PR #3178, opened 2020-10-27
Fix false positive when `to` is used to create a pair within a function

sowmyav24:
Fixes [#3177](https://github.com/detekt/detekt/issues/3177)

- Added check to ensure `Pair` is used within the callExpression before reporting
- Added test to create pair using `Pair` and `to` within a function


sowmyav24:
The code coverage for the class,  although it fails in CI, when run locally gives 100%
coverage

BraisGabin (approved):

BraisGabin (approved):

cortinico (approved):

schalkms (approved):

PR #3179, opened 2020-10-27
Fix false negative in IgnoredReturnValue

### LINKED ISSUE #3170: False negative in IgnoredReturnValue

t-kameyama:
Fixes #3170

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
@@ -9,19 +9,12 @@ import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.internal.valueOrDefaultCommaSeparated
 import io.gitlab.arturbosch.detekt.api.simplePatternToRegex
-import org.jetbrains.kotlin.com.intellij.psi.PsiComment
-import org.jetbrains.kotlin.com.intellij.psi.PsiElement
-import org.jetbrains.kotlin.com.intellij.psi.PsiWhiteSpace
-import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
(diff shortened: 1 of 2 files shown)


detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
```diff
@@ -72,8 +65,12 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
     @Suppress("ReturnCount")
     override fun visitCallExpression(expression: KtCallExpression) {
         super.visitCallExpression(expression)
-
         if (bindingContext == BindingContext.EMPTY) return
+
+        val parentQualifiedExpression = expression.getStrictParentOfType<KtQualifiedExpression>()
+        if (parentQualifiedExpression != null && parentQualifiedExpression.selectorExpression != expression) return
+        if (expression.isUsedAsExpression(bindingContext)) return
```
BraisGabin, line 72: 🤔 what if the statment is not the last of the if? I mean:

```kotlin
val a = if (random()) {
  2
} else {
  returnsInt()
  3
}
```

Is this handled correctly?

t-kameyama: Yes 5c2d3e3, but the following case is not flagged. I will fix.
```kotlin
if (Random.nextBoolean()) {
    1
} else {
    returnsInt()
    2
}.plus(1)
```

t-kameyama: Fixed in 217901b

BraisGabin (approved):
👏🏻👏🏻

schalkms (approved):

PR #3181, opened 2020-10-28
Improve PreferToOverPairSyntax

t-kameyama:
- Use `visitCallExpression` instead of `visitSimpleNameExpression`
- Use `getResolvedCall(bindingContext)` instead of `getType(bindingContext)`


schalkms (approved):
Thanks for simplifying the code! Looks good!

Issue #3131, opened 2020-10-06
UnusedImports false positive for enums in annotation attributes (with type resolution)


vladimirfx:
## Expected Behavior
When enum constants used in annotation attributes the enum import should not marked as unused.

## Steps to Reproduce
In such code `BatchFetch` import statement marked as unused:
```
...
import org.eclipse.persistence.annotations.FetchType
...
@Entity
class Sale(
    @Id
    val id: Long,
    @ManyToOne(fetch = FetchType.LAZY, optional = false)
    var client: Client,
...
)
```

## Context

We use Gradle plugin with type resolution mode (detektMain task). Bug is introduced after 1.12.0 (witch works as expected). 

## Your Environment
* Version of detekt used: 1.14.1
* Version of Gradle used (if applicable): 6.6.1
* Operating System and version: MacOS 10.15.7

chriskn:
This issue occurs from version 1.13.0 up. A fix would be highly appreciated

cortinico:
> In such code `BatchFetch` import statement marked as unused:

It seems to me that `BatchFetch` is actually unused in your code.
Can you provide a reproducible example that is not specific to any framework?

vladimirfx:
> > In such code `BatchFetch` import statement marked as unused:
> 
> It seems to me that `BatchFetch` is actually unused in your code.
> Can you provide a reproducible example that is not specific to any framework?

Sorry - my mistake. It is fragment from real project and BatchFetch used and reported too. I've edit sample code. Any enum used in annotation in way `EnumName.CONSTANT` reported as unused import.

vladimirfx:
I've try detekt 1.15.0-RC1 - issue fixed

PR #3188, opened 2020-10-31
Check for static imports in unused imports rule

sowmyav24:
Fixes [#3131 ](https://github.com/detekt/detekt/issues/3131)

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportsSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportsSpec.kt
@@ -555,9 +555,9 @@ class UnusedImportsSpec : Spek({
 
         it("does not report static import") {
             val mainFile = """
-                import x.y.z.Foo
+                import x.y.z.FetchType
                 
-                val x = Foo.LAZY
+                val x = FetchType.LAZY
             """
             val additionalFile = """
                 package x.y.z
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportsSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportsSpec.kt
@@ -554,16 +554,19 @@ class UnusedImportsSpec : Spek({
         }
 
         it("does not report static import") {
-            val main = """
-                package com.example
-
-                import org.eclipse.persistence.annotations.FetchType
-
-                class Test {
-                    var fetch = FetchType.LAZY
+            val mainFile = """
+                import x.y.z.Foo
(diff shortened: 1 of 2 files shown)


BraisGabin (approved):

sowmyav24:
When I run the task `./gradlew :detekt-rules-style:test` in local it passes. In CI it seems to fails for test-snippet compilation.

BraisGabin:
Use `./gradlew :detekt-rules-style:test -Pcompile-test-snippets=true`

Compile all the snippets each time that we run the tests slow down the test and is kind of meta testing (testing the tests). For that reason we have the compilation disabled by default and we just compile them in CI.

`import org.eclipse.persistence.annotations.FetchType` is not in the classpath so it fails. `compileAndLintWithContext` allows multiple `Strings` that are like different files. This way you can create your own enum in other package and then use it in your test.

cortinico (approved):

schalkms (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportsSpec.kt
```diff
@@ -552,5 +552,21 @@ class UnusedImportsSpec : Spek({
...
+                import x.y.z.Foo
+                
+                val x = Foo.LAZY
+            """
+            val additionalFile = """
+                package x.y.z
+                
+                enum class FetchType {
+                    LAZY
+                }
```
BraisGabin, line 567: Why does this compile? the enum is called `FetchType` but the import is for `Foo` 🤔

BraisGabin (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportsSpec.kt
sowmyav24: I wrote the tests with Foo and later renamed them before committing  to mimic the issue reported. However, your point of compilation seems to be valid 😮

BraisGabin: @schalkms do you have any idea about why does this happen? This doesn't stop the PR but I will like to know why does this happen

schalkms: @BraisGabin Yes, I do. Please take a look at the following lines and closely check line number 68.

https://github.com/detekt/detekt/blob/da6259a15c986d248799cc5f2db308689e4fc68e/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt#L63-L72

arturbosch (approved):

Issue #3183, opened 2020-10-29
Internal exception should fail the gradle task


chao2zhang:
<!-- The bug you're experiencing might have already be reported! -->
<!-- Please search in the [issues](https://github.com/detekt/detekt/issues) before creating one. -->

## Expected Behavior
Internal exception during the analysis phase should fail the gradle task

## Observed Behavior
When the YAML config has a type mismatch, the exception stack trace is printed but the gradle task was still successful.

## Context

We found this issue when we updated the custom YAML config to a new version but still used an old detekt version.
I would also imagine this issue remains when we update detekt version but did not update the custom YAML.
The current issue ended up leaving us not actually running detekt for over a week.

## Steps to Reproduce

Sample project https://github.com/chao2zhang/DetektSilentError

Assume we have the following config:
```
config:
  validation: true

style:
  MaxLineLength:
    active: true
    maxLineLength: 'abc'
    excludePackageStatements: true
    excludeImportStatements: true
    excludeCommentStatements: false
```
Apparently, maxLineLength should be a number, and I expect the detekt task to fail because the config is incorrect.
When running `./gradlew detekt`, although the errors are printed out, the gradle task turns out to be successful.

```
~/DetektSilentError(main) » ./gradlew detekt --rerun-tasks                                                                                                                                                                                                                                          cazhang@cazhang-mn4

> Task :detekt
Analyzing '/Users/cazhang/DetektSilentError/src/main/kotlin/A.kt' led to an exception.
The original exception message was: java.lang.String cannot be cast to java.lang.Number
Running detekt '1.14.2' on Java '1.8.0_212-b10' on OS 'Mac OS X'.
If the exception message does not help, please feel free to create an issue on our GitHub page.
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Number
        at io.gitlab.arturbosch.detekt.rules.style.MaxLineLength.<init>(MaxLineLength.kt:36)
        at io.gitlab.arturbosch.detekt.rules.style.FileParsingRule.<init>(FileParsingRule.kt:9)
        at io.gitlab.arturbosch.detekt.rules.style.StyleGuideProvider.instance(StyleGuideProvider.kt:32)
        at io.gitlab.arturbosch.detekt.core.Analyzer$analyze$4.invoke(Analyzer.kt:106)
        at io.gitlab.arturbosch.detekt.core.Analyzer$analyze$4.invoke(Analyzer.kt:25)
[...]

PR #3193, opened 2020-11-02
Throw error instead of logging as error in analysis phase

chao2zhang:
This PR addresses #3183

A few changes worth calling out:
- Move a few public elements from `Junk.kt` to private elements `Analyzer.kt`
- Add a new primary constructor for UnexpectedError which accepts the message
- Rename a few places where the name `Detektor` appears to `Analyzer`.  #2861 renamed the class from `Detektor` to `Analyzer` but did not update all references like variable names.

With this PR, users will see the following output in the console:
```
~/DetektSilentError(main*) » ./gradlew detekt                                                                                                                                     1 ↵ cazhang@cazhang-mn4
> Task :detekt FAILED

* What went wrong:
> io.github.detekt.tooling.api.AnalysisError: Analyzing /Users/cazhang/DetektSilentError/src/main/kotlin/A.kt led to an exception. 
  The original exception message was: java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Number
  Running detekt '1.14.2' on Java '1.8.0_212-b10' on OS 'Mac OS X'
  If the exception message does not help, please feel free to create an issue on our GitHub page.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 749ms
1 actionable task: 1 executed
---------------------------------
```

There is one more usage of `LoggingAware.error()`, although tangential to this PR, it would be great to remove `LoggingAware.error()` because logging as an error is not treating the severity correctly. When an exception happens, we should surface the error back to the user rather than logging and telling the user that the command completed successfully.

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
@@ -1,7 +1,6 @@
 package io.gitlab.arturbosch.detekt.core
 
 import io.github.detekt.psi.absolutePath
-import io.github.detekt.tooling.api.UnexpectedError
 import io.github.detekt.tooling.api.spec.ProcessingSpec
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.FileProcessListener
@@ -70,7 +69,7 @@ internal class Analyzer(
         ktFiles.map { file ->
             processors.forEach { it.onProcess(file, bindingContext) }
             val findings = runCatching { analyze(file, bindingContext, compilerResources) }
-                .onFailure { throw UnexpectedError(message = createErrorMessage(file, it), cause = it) }
+                .onFailure { throwIllegalStateException(file, it) }
                 .getOrDefault(emptyMap())
             processors.forEach { it.onProcessComplete(file, findings, bindingContext) }
             findings
@@ -88,9 +87,7 @@ internal class Analyzer(
                 val findings = analyze(file, bindingContext, compilerResources)
                 processors.forEach { it.onProcessComplete(file, findings, bindingContext) }
                 findings
-            }.recover {
-                throw UnexpectedError(message = createErrorMessage(file, it), cause = it)
-            }
+            }.recover { throwIllegalStateException(file, it) }
         }
         return awaitAll(tasks).filterNotNull()
     }
@@ -144,12 +141,15 @@ private fun MutableMap<String, List<Finding>>.mergeSmells(other: Map<String, Lis
     }
 }
 
-private fun createErrorMessage(file: KtFile, error: Throwable): String = """
+private fun throwIllegalStateException(file: KtFile, error: Throwable): Nothing {
+    val message = """
     Analyzing ${file.absolutePath()} led to an exception. 
     The original exception message was: ${error.localizedMessage}
     Running detekt '${whichDetekt() ?: "unknown"}' on Java '${whichJava()}' on OS '${whichOS()}'
     If the exception message does not help, please feel free to create an issue on our GitHub page.
-""".trimIndent()
+    """.trimIndent()
+    throw IllegalStateException(message, error)
+}
 
 internal fun ProcessingSpec.workaroundConfiguration(config: Config): Config = with(configSpec) {
     var declaredConfig: Config? = when {
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
@@ -1,5 +1,7 @@
 package io.gitlab.arturbosch.detekt.core
 
+import io.github.detekt.psi.absolutePath
+import io.github.detekt.tooling.api.UnexpectedError
 import io.github.detekt.tooling.api.spec.ProcessingSpec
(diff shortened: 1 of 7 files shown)


detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/DetektError.kt
```diff
@@ -9,4 +9,6 @@ class MaxIssuesReached(message: String) : DetektError(message)
 
 class InvalidConfig(message: String) : DetektError(message)
 
-class UnexpectedError(override val cause: Throwable) : DetektError(null, cause)
+class UnexpectedError(message: String?, override val cause: Throwable) : DetektError(message, cause) {
+    constructor(cause: Throwable) : this(null, cause)
```
schalkms, line 13: Why is the explicit constructor call now necessary?

schalkms (approved):
Thanks for this PR. This one looks good. I think the mentioned improvements should be tackled in a subsequent PR. What do you think? May I ask you to please create an issue containing this improvement for tracking purposes?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -82,8 +89,7 @@ internal class Analyzer(
                 processors.forEach { it.onProcessComplete(file, findings, bindingContext) }
                 findings
             }.recover {
-                settings.error(createErrorMessage(file, it), it)
-                emptyMap()
+                throw UnexpectedError(message = createErrorMessage(file, it), cause = it)
```
arturbosch, line 92: Throwing in a parallel environment won't immediately stop detekt but would throw up to #cores errors.

Imo a cleaner implementation would be:
1. Log the errors as they are
2. Introduce a mutable boolean property `hasErrors`; it does not need to be thread safe as we only care about it changing to `true`
3. Check before returning in

https://github.com/detekt/detekt/blob/923b3b5b00a3cfb090c13e2f53020cc7ba373dae/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt#L32

if `hasErrors` is `true` and throw a `IllegalStateException` with a message that detekt analysis finished with errors and please see the log.

What do you think?

chao2zhang: I don't think that's necessarily cleaner: `checkConfiguration()` throws the exception the same way as is in this PR. And we don't necessarily need to print out the stacktrace unless the user specifies `--stacktrace` when running gradle.

I attached the console output in the PR description. Please take a look

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/DetektError.kt
chao2zhang: Because we have callers that invoke `UnexpectedError(throwable)` without a message. Here I changed the constructor taking one parameter to secondary and defined the primary constructor to take two parameters.

chao2zhang:
> Thanks for this PR. This one looks good. I think the mentioned improvements should be tackled in a subsequent PR. What do you think? May I ask you to please create an issue containing this improvement for tracking purposes?

See #3183. This PR is already linked to the issue aforementioned.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
arturbosch: Can you try again with 2+ files and `--parallel` and post the console output please?

If we go further with throwing an Error, lets not throw an `UnexpectedError`. It is meant to be a wrapper for the tooling API. Please throw an `IllegalStateException` here, it will automatically be wrapped in an `UnexpectedError`.
Further this error does not have a message by design; we could pass the wrapped error message but lets not introduce a field for it.

```kt

class UnexpectedError(override val cause: Throwable) : DetektError(null, cause)

```

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -132,6 +138,19 @@ private fun <T, U, R> Sequence<Pair<T, U>>.mapLeft(transform: (T, U) -> R): Sequ
     return this.map { (first, second) -> transform(first, second) to second }
 }
 
+private fun MutableMap<String, List<Finding>>.mergeSmells(other: Map<String, List<Finding>>) {
+    for ((key, findings) in other.entries) {
+        merge(key, findings) { f1, f2 -> f1.plus(f2) }
+    }
+}
+
+private fun createErrorMessage(file: KtFile, error: Throwable): String = """
```
arturbosch, line 147: How about we build and throw the error here instead of two times `throw UnexpectedError(message = createErrorMessage(file, it), cause = it)` ?

chao2zhang: 👍

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/DetektError.kt
schalkms: Thank you for the explanation!

arturbosch (approved):

Issue #3189, opened 2020-10-31
MissingWhenCase false negative with nulls


BraisGabin:
## Expected Behavior
`MissingWhenCase` should check for null case if the enum or sealed can be `null` 

## Observed Behavior
It doesn't check that.

## Steps to Reproduce
```kotlin
fun whenOnEnumPass(c: Color?) {
    when(c) {
        Color.BLUE -> {}
        Color.GREEN -> {}
        Color.RED -> {}
    }
}
```

## Your Environment
* Version of detekt used: `master` (`1.14.2`)

schalkms:
Closed due to the arguments made in the referenced PR.
- https://github.com/detekt/detekt/pull/3194#issuecomment-720625803
- https://github.com/detekt/detekt/pull/3194#issuecomment-720635060

BraisGabin:
Should we create a tag "require smart cast" or something similar to track this issues? I mean, even if we close it to find them if we eventually implement it. Or if we want ro evaluate if it's worth it to implement it.

schalkms:
An available smart cast detection alone won't solve the problem here. You'll need more sophisticated analysis than that.

BraisGabin:
🤔 why not?

schalkms:
I didn't know that we already have dataflow analysis as part of the `bindingContext`. Great. Thanks for bringing this up.
This issue can be reopened.

https://github.com/detekt/detekt/blob/b1269b58dc704c096ed402f91c9962942989a18a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullableToStringCall.kt#L66-L78

schalkms:
By the way, this has nothing to do with smart casts per se. The enabler here is data flow analysis as part of the Kotlin compiler. With the help of data flow analysis we can detect whether null checks or the necessary casts have been made before the `when` clause.

PR #3194, opened 2020-11-02
Check for presence of null case in MissingWhenCase rule

sowmyav24:
Fixes: [#3189 ](https://github.com/detekt/detekt/issues/3189)

- `WhenChecker` does not return `null` while fetching missing cases.
- Used the method `containsNullCase` to verify if null was defined by user.

### DIFF
--- a/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingWhenCaseSpec.kt
+++ b/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingWhenCaseSpec.kt
@@ -81,6 +81,26 @@ object MissingWhenCaseSpec : Spek({
                 assertThat(actual.first().message).isEqualTo("When expression is missing cases: null. Either add missing cases or a default `else` case.")
             }
 
+            it("does not report missing null case in `when` expression when it is handled outside of `when`") {
+                val code = """
+                enum class Color {
+                    RED,
+                    GREEN,
+                    BLUE
+                }
+
+                fun whenNulLCheckEnum(c: Color?) {
+                    if(c == null) return
+                    when(c) {
+                        Color.BLUE -> {}
+                        Color.GREEN -> {}
+                        Color.RED -> {}
+                    }
+                }
+                """
+                assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
+            }
+
             it("does not report when `when` expression used as statement and all cases are covered") {
                 val code = """
                 enum class Color {
@@ -171,6 +191,26 @@ object MissingWhenCaseSpec : Spek({
                 assertThat(actual).hasSize(1)
                 assertThat(actual.first().issue.id).isEqualTo("MissingWhenCase")
                 assertThat(actual.first().message).isEqualTo("When expression is missing cases: VariantC, null. Either add missing cases or a default `else` case.")
+            }
+
+            it("does not report missing null case in `when` expression when it is handled outside of `when`") {
+                val code = """
+                sealed class Variant {
+                        object VariantA : Variant()
+                        class VariantB : Variant()
+                        object VariantC : Variant()
+                    }
+
+                    fun whenOnEnumFail(v: Variant?) {
+                        if(v == null) return
+                        when(v) {
+                            is Variant.VariantA -> {}
+                            is Variant.VariantB -> {}
+                            is Variant.VariantC -> {}
+                        }
+                    }
+                """
+                assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
             }
 
             it("does not report when `when` expression used as statement and all cases are covered") {
--- a/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingWhenCaseSpec.kt
+++ b/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingWhenCaseSpec.kt
@@ -38,6 +38,49 @@ object MissingWhenCaseSpec : Spek({
                 assertThat(actual.first().message).isEqualTo("When expression is missing cases: RED. Either add missing cases or a default `else` case.")
             }
 
+            it("reports when `when` expression used as statement and not all cases including null are not covered") {
+                val code = """
+                enum class Color {
+                    RED,
+                    GREEN,
+                    BLUE
+                }
+
+                fun whenOnEnumFail(c: Color?) {
+                    when(c) {
+                        Color.BLUE -> {}
+                        Color.GREEN -> {}
+                    }
+                }
+                """
+                val actual = subject.compileAndLintWithContext(env, code)
+                assertThat(actual).hasSize(1)
+                assertThat(actual.first().issue.id).isEqualTo("MissingWhenCase")
+                assertThat(actual.first().message).isEqualTo("When expression is missing cases: RED, null. Either add missing cases or a default `else` case.")
+            }
+
+            it("reports when `when` expression used as statement and null case is not covered") {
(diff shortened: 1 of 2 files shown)


BraisGabin (dismissed):

t-kameyama:
False positive:
```kotlin
enum class Color {
    RED,
    GREEN,
    BLUE
}
fun whenOnEnumFail(c: Color?) {
    if (c == null) return
    when (c) {
        Color.RED -> {}
        Color.BLUE -> {}
        Color.GREEN -> {}
    }
}
```

sowmyav24:
> False positive:
> 
> ```kotlin
> enum class Color {
>     RED,
>     GREEN,
>     BLUE
> }
> fun whenOnEnumFail(c: Color?) {
>     if (c == null) return
>     when (c) {
>         Color.RED -> {}
>         Color.BLUE -> {}
>         Color.GREEN -> {}
>     }
> }
> ```

@BraisGabin Do you think we can check if null check has been done prior to `when` block execution ?

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingWhenCaseSpec.kt
```diff
@@ -38,6 +38,49 @@ object MissingWhenCaseSpec : Spek({
                 assertThat(actual.first().message).isEqualTo("When expression is missing cases: RED. Either add missing cases or a default `else` case.")
             }
 
+            it("reports when `when` expression used as statement and not all cases including null are not covered") {
```
schalkms, line 41: Double negation:

```suggestion
            it("reports when `when` expression used as statement and not all cases including null are covered") {
```

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingWhenCaseSpec.kt
```diff
@@ -87,6 +130,49 @@ object MissingWhenCaseSpec : Spek({
...
+                        }
+                    }
+                """
+                val actual = subject.compileAndLintWithContext(env, code)
+                assertThat(actual).hasSize(1)
+                assertThat(actual.first().issue.id).isEqualTo("MissingWhenCase")
+                assertThat(actual.first().message).isEqualTo("When expression is missing cases: null. Either add missing cases or a default `else` case.")
+            }
+
+            it("reports when `when` expression used as statement and not all cases including null are not covered") {
```
schalkms, line 155: Double negation:

```suggestion
            it("reports when `when` expression used as statement and not all cases including null are covered") {
```

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingWhenCaseSpec.kt
sowmyav24: Done 👍

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingWhenCaseSpec.kt
sowmyav24: Done 👍

BraisGabin:
> > False positive:
> > ```kotlin
> > enum class Color {
> >     RED,
> >     GREEN,
> >     BLUE
> > }
> > fun whenOnEnumFail(c: Color?) {
> >     if (c == null) return
> >     when (c) {
> >         Color.RED -> {}
> >         Color.BLUE -> {}
> >         Color.GREEN -> {}
> >     }
> > }
> > ```
> 
> @BraisGabin Do you think we can check if null check has been done prior to `when` block execution ?

🤔 Can we know about the smart casting? This case is trivial and we could say "you should add the `null` inside the `when`". But I can imagine more complex code where you can't just add it in the `when`. So we should relay in the smart cast. If we can't do that I think that it's better to have a false-negative than a false-positive.

schalkms:
> Can we know about the smart casting? This case is trivial and we could say "you should add the null inside the when". But I can imagine more complex code where you can't just add it in the when. So we should relay in the smart cast. If we can't do that I think that it's better to have a false-negative than a false-positive.

After reading the good comment from @t-kameyama I agree with @BraisGabin here. We should leave the null check out and stick with the false-negative. I think there are many more false-positives that look similar to the mentioned case, but are not the same. To detect those cases, you would need to use data-flow analysis to get it completely right. Currently, detekt does not have it and I'm unsure, whether detekt will have it in the near future. My tendency would be no.

sowmyav24:
> To detect those cases, you would need to use data-flow analysis to get it completely right

I agree with this. To avoid `false positives`, I think we should go ahead and close this PR and the issue.

schalkms:
Okay, I'll close it.
I'm sorry for you @sowmyav24 that your hard work on this PR isn't rewarded by a merge.

schalkms:
Okay, I'll close it.
I'm sorry @sowmyav24 that your hard work on this PR isn't rewarded by a merge.

sowmyav24:
> I'm sorry @sowmyav24 that your hard work on this PR isn't rewarded by a merge.

It is good we caught the issue before merging :)

t-kameyama:
We can detect smart casting. 

See:
https://github.com/detekt/detekt/blob/b1269b58dc704c096ed402f91c9962942989a18a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullableToStringCall.kt#L66-L78

https://github.com/detekt/detekt/blob/282c038c2e9fdc462b73952c65d1688b15fd35da/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullableToStringCallSpec.kt#L73-L87

BraisGabin:
@schalkms you said that smart cast was not enough for this issue, could you explain why? Otherwise we can implement this.

schalkms:
I sent my comments. Please check the referenced issue. Thanks to @t-kameyama for mentioning the necessary data flow analysis code snippet in order to achieve this.

BraisGabin:
@sowmyav24 if you want, you can continue with this PR. Thanks @t-kameyama !

sowmyav24:
I will update the PR 👍

sowmyav24:
In order to fetch the data flow information, we have to traverse and visit the complete block of function/expression. Currently `MissingWhenCase` takes into account only `visitWhenExpression`. Does anyone see any downside to that ?

t-kameyama:
I think we should check `WhenChecker.getMissingCases(expression, bindingContext)`.

1. 
```
enum class Color {
    RED,
    GREEN,
    BLUE
}
fun whenOnEnumFail(c: Color?) {
    when (c) {
        Color.RED -> {}
        Color.BLUE -> {}
        Color.GREEN -> {}
    }
}
```
`WhenChecker.getMissingCases(expression, bindingContext)` is not empty.

2.
```
enum class Color {
    RED,
    GREEN,
    BLUE
}
fun whenOnEnumFail(c: Color?) {
    if (c == null) return
    when (c) {
        Color.RED -> {}
        Color.BLUE -> {}
        Color.GREEN -> {}
    }
}
```
`WhenChecker.getMissingCases(expression, bindingContext)` is empty.

sowmyav24:
@t-kameyama Thank you for this input . I did not realise `getMissingCases` ensures it is taken care of.

sowmyav24:
@BraisGabin @schalkms Please take a look at the latest commit changes. I have incorporated the suggestion given by @t-kameyama

BraisGabin (approved):

schalkms (approved):

PR #3202, opened 2020-11-07
Remove Suppress annotation from UnusedImports

schalkms:

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
@@ -84,13 +84,16 @@ class UnusedImports(config: Config) : Rule(config) {
             super.visitPackageDirective(directive)
         }
 
-        @Suppress("UnsafeCallOnNullableType")
         override fun visitImportList(importList: KtImportList) {
-            imports = importList.imports.asSequence().filter { it.isValidImport }
-                .filter { it.identifier()?.contains("*")?.not() == true }
-                .filter { it.identifier() != null }
-                .filter { !operatorSet.contains(it.identifier()) }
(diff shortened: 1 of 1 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
```diff
@@ -84,13 +84,16 @@ class UnusedImports(config: Config) : Rule(config) {
...
-                .filter { !componentNRegex.matches(it.identifier()!!) }.toList()
+            imports = importList.imports.asSequence()
+                .filter { it.isValidImport }
+                .filter {
+                    val identifier = it.identifier()
+                    identifier?.contains("*")?.not() == true &&
+                            !operatorSet.contains(identifier) &&
+                            !componentNRegex.matches(identifier)
+                }
+                .toList()
```
BraisGabin, line 96: Does this improve readability? It removes an unsafe call but I think that it is not worth it

BraisGabin: But, I like to move the `.toList()` and the first `filter` to their own line

cortinico: > Does this improve readability? 

I believe it does as the bigger `filter{}` is all logically related to filtering out identifiers.

cortinico (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
schalkms: I agree with Nico. Furthermore, we get rid of the **!!**.

PR #3209, opened 2020-11-07
Remove Suppress annotation from ArrayPrimitive

schalkms:


detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/ArrayPrimitive.kt
```diff
@@ -49,18 +49,20 @@ class ArrayPrimitive(config: Config = Config.empty) : Rule(config) {
         Debt.FIVE_MINS
     )
 
-    @Suppress("ReturnCount")
     override fun visitCallExpression(expression: KtCallExpression) {
         super.visitCallExpression(expression)
-        if (bindingContext == BindingContext.EMPTY) return
+        expression.checkArrayPrimitive()
```
BraisGabin, line 54: I imagine that you need to move this to an extension function just because `ReturnCount` doesn't count as checks the returns behind the super call, right? Shouldn't we change `ReturnCount` instead?

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/ArrayPrimitive.kt
```diff
@@ -49,18 +49,20 @@ class ArrayPrimitive(config: Config = Config.empty) : Rule(config) {
...
+    }
 
-        if (expression.calleeExpression?.text !in factoryMethodNames) return
-        val descriptor = expression.getResolvedCall(bindingContext)?.resultingDescriptor ?: return
+    private fun KtCallExpression.checkArrayPrimitive() {
+        if (bindingContext == BindingContext.EMPTY) return
```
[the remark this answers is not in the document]
schalkms: I'll refactor it once more to make the intention clear.

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/ArrayPrimitive.kt
schalkms: I wouldn't do that just because detekt heavily uses the Visitor pattern.

cortinico (approved):

BraisGabin (approved):

PR #3210, opened 2020-11-07
Refactor HasPlatformType rule

schalkms:


detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/HasPlatformType.kt
```diff
@@ -61,19 +61,21 @@ class HasPlatformType(config: Config) : Rule(config) {
...
 
-        val callable =
-            bindingContext[BindingContext.DECLARATION_TO_DESCRIPTOR, this] as? CallableDescriptor ?: return false
-        if (!callable.visibility.isPublicAPI) return false
-        return callable.returnType?.isFlexible() ?: return false
+        return when (this) {
+            is KtFunction -> !isLocal && !hasDeclaredReturnType() && isPlatFormType()
+            is KtProperty -> !isLocal && typeReference == null && isPlatFormType()
+            else -> false
+        }
```
BraisGabin, line 79:
```suggestion
        return when (this) {
            isLocal -> false
            !isPlatformType() -> false
            this is KtFunction -> !hasDeclaredReturnType()
            this is KtProperty -> typeReference == null
            else -> false
        }
```

BraisGabin (approved):

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/HasPlatformType.kt
schalkms: That doesn't work, because the type `KtCallableDeclaration` does not have `isLocal` nor `hasDeclaredReturnType()`.

PR #3220, opened 2020-11-09
Fix false negatives in UnreachableCode rule

### LINKED ISSUE #3219: False negatives in UnreachableCode rule
1.
```kotlin
fun test(b: Boolean): Int {
    if (b) {
        return 1
    } else {
        return 2
    }
    return 0
}
```

2.
```kotlin
enum class E { A, B }

fun test(e: E): Int {
    when (e) {
        E.A -> return 1
        E.B -> return 2
    }
    return 0
}
```

3.
```kotlin
fun test(): Int {
    try {
        return 1
    } catch (e: Exception) {
        throw e
    }
    return 0
}
```

t-kameyama:
Fixes #3219


BraisGabin (approved):

schalkms (changes requested):
Thanks for submitting this PR! The rule is now more sophisticated than ever before with the help of type and symbol solving. 🙂
Consequently, this change is breaking. We would need to add the require annotation to the rule comment.
Furthermore, the check for an empty **bindingContext** is missing in my opinion.

schalkms (approved):

PR #3224, opened 2020-11-11
Simplify EqualsOnSignatureLine rule

t-kameyama:


BraisGabin (approved):

schalkms (commented):
Thanks for this PR!
I’m not sure whether those changes really simplify the rule. Could you please elaborate why? The codecov report shows an increase in complexity of a very small margin.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/EqualsOnSignatureLine.kt
```diff
@@ -37,22 +38,13 @@ class EqualsOnSignatureLine(config: Config = Config.empty) : Rule(config) {
     override val issue = Issue(javaClass.simpleName, Severity.Style, MESSAGE, Debt.FIVE_MINS)
 
     override fun visitNamedFunction(function: KtNamedFunction) {
-        val equals = function.equalsToken ?: return
-
-        // Get the tokens after the parameter list (but not including)
-        val afterParams = function.valueParameterList?.siblings(forward = true, withItself = false) ?: return
-
-        // Collect tokens until we find the equals sign
-        val untilEquals = afterParams.takeWhile { it !== equals }.toList()
```
t-kameyama, line 46: `toList()` is useless. It should be processed as Sequence.

t-kameyama (commented):
I think it's simpler to go back from the equals token.

schalkms:
Thanks for the explanation. Let’s see what others think about the change.

PR #3226, opened 2020-11-11
Remove unused resources from the website

cortinico:
I noticed that we do have a lot of resources inside the `docs/` folder that are unused and that gets published on the website. E.g. https://detekt.github.io/detekt/docker-compose.yml

I suspect we're not using them so I'm cleaning up the folder with all the stuff we don't need.


BraisGabin (approved):
I'm thinking... Now we can use any jekyll plugin, right? So we can have tabs for the different gradle dsls so our doc is a bit less verbose.

cortinico:
> I'm thinking... Now we can use any jekyll plugin, right? So we can have tabs for the different gradle dsls so our doc is a bit less verbose.

True. However while looking at the website I realised we used this theme, that already has support for that tabbed layouts: https://idratherbewriting.com/documentation-theme-jekyll/mydoc_navtabs.html

BraisGabin:
Good to know. But, do we want html in our documentation? The other option is this: https://github.com/detekt/detekt/pull/2934/files#diff-3213dd73753b7a82d708ad93393cc0bd16cc1d406339419751c28948b8b78f37

schalkms (approved):

cortinico:
> t html in our documentation? The other option is this: [#2934 (files)](https://github.com/detekt/detekt/pull/2934/files#diff-3213dd73753b7a82d708ad93393cc0bd16cc1d406339419751c28948b8b78f37)

Yeah `jekyll-code-tabs` is probably better than raw HTML

PR #3199, opened 2020-11-04
Setup the website publishing pipeline

cortinico:
This PR sets up a workflow for building the KDoc/Documentation and the website and publish it to Github Pages. I've also removed all the generated files that were creating a lot of noise in the history (see #3160).

The generated website will be published on the `gh-pages` branch as a commit created by GH Actions.

@arturbosch If we decide to merge this PR, you should update the GH Pages settings of the repo to publish from the `gh-pages` branch (I believe now is set to publish from the `docs/` folder of the `master` branch).

I've also updated the `verify-documentation` job of the pre merge workflow to verify only the Generated Detekt Config file (as it's pointless now to verify the generated documentation).

PR #3227, opened 2020-11-11
Fix broken website redirects

cortinico:
It seems like after #3199 (or even before?) we broke redirects.

https://detekt.github.io/detekt/kotlindsl.html should redirect to https://detekt.github.io/detekt/gradle.html. See: 

https://github.com/detekt/detekt/blob/cd659ce8737fb177caf140f46f73a1a86b22be56/docs/pages/gettingstarted/gradle.md#L7-L9

This is broken right now, I'm fixing it.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -111,7 +111,7 @@ description: "Meet detekt, a static code analysis tool for Kotlin."
 # the description is used in the feed.xml file
 
 # needed for sitemap.xml file only
-url: https://detekt.github.io/detekt/
-baseurl: /detekt/
+url: https://detekt.github.io
+baseurl: /detekt
 
 detekt_version: 1.14.2


BraisGabin (approved):
We should check the site map too after this mege. It should improve a bit the SEO if it's correct. I remember that a fix it once but, probably, it was broken too now.

schalkms (approved):

cortinico:
@arturbosch Seems like the documentation website is not updated 🤔 

The deployment to `gh-pages` works file:
https://github.com/detekt/detekt/blob/ba0ede8011ba590a37eabf623a2a50587a55a224/redirects.json#L1

But GH is failing to publish the website as the `/docs` folder is missing:
https://github.com/detekt/detekt/runs/1390077842

You should be able to update this in the settings of the repo (deploying pages from `root` of `gh-pages` and not from the `docs/` folder).

cortinico:
> @arturbosch Seems like the documentation website is not updated 🤔

Now it works correctly 👌

Issue #3197, opened 2020-11-02
Return value in finally block try-catch-finally expression


rwqwr:
Detekt doesn't check unaffected finally block in try-catch-finally expression on result value.
See: https://kotlinlang.org/docs/reference/exceptions.html#try-is-an-expression
## Noncompliant Code
```kotlin
val expression = try {
    "try"
} catch (e: Exception) {
    "exception"
} finally {
    "finally"
}
```
## Compliant Code
```kotlin
val expression = try {
    "try"
} catch (e: Exception) {
    "exception"
}
```

schalkms:
Thanks for submitting this issue. I suppose **ReturnFromFinally.kt** does not report this snippet.

rwqwr:
You are right. ReturnFromFinally.kt does not report this snippet.
Can I implement this case?

cortinico:
> Can I implement this case?

Sure feel free. I'm assigning this issue to you 👍

PR #3228, opened 2020-11-12
Rule improvement: ReturnFromFinally

rwqwr:
Fixes https://github.com/detekt/detekt/issues/3197

### DIFF
--- a/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt
+++ b/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt
@@ -20,8 +20,7 @@ import org.jetbrains.kotlin.resolve.calls.callUtil.getType
 /**
  * Reports all `return` statements in `finally` blocks.
  * Using `return` statements in `finally` blocks can discard and hide exceptions that are thrown in the `try` block.
- * Also report unaffected return value in `finally` block if it used as expression.
- * See: https://kotlinlang.org/docs/reference/exceptions.html#try-is-an-expression
+ * Furthermore, this rule reports values from `finally` blocks, if the corresponding `try` is used as an expression.
  *
  * <noncompliant>
  * fun foo() {
--- a/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt
+++ b/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt
@@ -10,12 +10,18 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import org.jetbrains.kotlin.psi.KtFinallySection
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtReturnExpression
+import org.jetbrains.kotlin.psi.KtTryExpression
 import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
 import org.jetbrains.kotlin.psi.psiUtil.parents
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.bindingContextUtil.isUsedAsExpression
+import org.jetbrains.kotlin.resolve.calls.callUtil.getType
 
 /**
  * Reports all `return` statements in `finally` blocks.
  * Using `return` statements in `finally` blocks can discard and hide exceptions that are thrown in the `try` block.
+ * Also report unaffected return value in `finally` block if it used as expression.
+ * See: https://kotlinlang.org/docs/reference/exceptions.html#try-is-an-expression
  *
  * <noncompliant>
  * fun foo() {
@@ -25,9 +31,12 @@ import org.jetbrains.kotlin.psi.psiUtil.parents
  *         return // prevents MyException from being propagated
  *     }
  * }
+ *
+ * val a: String = try { "s" } catch (e: Exception) { "e" } finally { "f" }
  * </noncompliant>
  *
  * @configuration ignoreLabeled - ignores labeled return statements (default: `false`)
+ * @requiresTypeResolution
  */
 class ReturnFromFinally(config: Config = Config.empty) : Rule(config) {
 
@@ -43,6 +52,30 @@ class ReturnFromFinally(config: Config = Config.empty) : Rule(config) {
             .collectDescendantsOfType<KtReturnExpression> { isNotInInnerFunction(it, innerFunctions) &&
                     canFilterLabeledExpression(it) }
             .forEach { report(CodeSmell(issue, Entity.from(it), issue.description)) }
+    }
+
+    override fun visitTryExpression(expression: KtTryExpression) {
+        super.visitTryExpression(expression)
+        if (bindingContext == BindingContext.EMPTY) return
+
+        val finallyBlock = expression.finallyBlock
+
+        if (expression.isUsedAsExpression(bindingContext) && finallyBlock != null) {
+            val finallyExpression = finallyBlock.finalExpression
+            if (finallyExpression.statements.isEmpty()) return
+
+            val finallyReturnType = finallyBlock.finalExpression.getType(bindingContext)
+
+            if (finallyReturnType == expression.getType(bindingContext)) {
+                report(
+                    CodeSmell(
+                        issue = issue,
+                        entity = Entity.Companion.from(finallyBlock),
+                        message = issue.description
+                    )
+                )
+            }
+        }
     }
 
     private fun isNotInInnerFunction(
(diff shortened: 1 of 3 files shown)


schalkms (changes requested):
Thanks for tackling this issue! 🙂
It's a really nice 1st contribution to detekt. 

Please note, that there's already an existing rule. The following rule should be adapted to handle this case. There's no need for a new rule. Otherwise, this is very confusing for users of detekt.
Please take a look at the following class.
https://github.com/detekt/detekt/blob/master/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt

Please kindly let me know, if you have any questions.

rwqwr:
Thank you for your response! 🙂
I fixed this moment and implemented this case in `ReturnFromFinally` rule.

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt
```diff
@@ -10,12 +10,18 @@ import io.gitlab.arturbosch.detekt.api.Severity
...
 import org.jetbrains.kotlin.psi.psiUtil.parents
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.bindingContextUtil.isUsedAsExpression
+import org.jetbrains.kotlin.resolve.calls.callUtil.getType
 
 /**
  * Reports all `return` statements in `finally` blocks.
  * Using `return` statements in `finally` blocks can discard and hide exceptions that are thrown in the `try` block.
+ * Also report unaffected return value in `finally` block if it used as expression.
+ * See: https://kotlinlang.org/docs/reference/exceptions.html#try-is-an-expression
```
schalkms, line 24: This might be better understandable to first time users of this rule.

```suggestion
 * Reports all `return` statements in `finally` blocks.
 * Using `return` statements in `finally` blocks can discard and hide exceptions that are thrown in the `try` block.
 * 
 * Furthermore, this rule reports values from `finally` blocks, if the corresponding `try` is used as an expression.
```

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt
```diff
@@ -45,6 +54,30 @@ class ReturnFromFinally(config: Config = Config.empty) : Rule(config) {
             .forEach { report(CodeSmell(issue, Entity.from(it), issue.description)) }
     }
 
+    override fun visitTryExpression(expression: KtTryExpression) {
+        super.visitTryExpression(expression)
+        if (bindingContext == BindingContext.EMPTY) return
+
+        val finallyBlock = expression.finallyBlock
+
+        if (expression.isUsedAsExpression(bindingContext) && finallyBlock != null) {
```
schalkms, line 63: This statement can be switched, since the usage of `bindingContext` in this case could be more expensive.

```suggestion
        if (finallyBlock != null && expression.isUsedAsExpression(bindingContext)) {
```

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt
```diff
@@ -45,6 +54,30 @@ class ReturnFromFinally(config: Config = Config.empty) : Rule(config) {
             .forEach { report(CodeSmell(issue, Entity.from(it), issue.description)) }
     }
```
schalkms, line 55: Maybe the logic from `visitFinallySection()` can
* be merged together with `visitTryExpression()`
* use the `bindingContext` capability as well.

This would reduce the complexity of this rule.

schalkms (commented):
Thanks for incorporating the changes! This PR looks good.
The mentioned refactoring can also be done in a separate PR, if you want.

rwqwr:
I refactored code and changed the implementation of `visitFinallySection()` with the `bindingContext` capability, but left the check inside it. It seemed to me more correct.

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt
```diff
@@ -66,16 +69,25 @@ class ReturnFromFinally(config: Config = Config.empty) : Rule(config) {
...
+    private fun isReturnFromTargetFunction(
+        blockExpression: KtBlockExpression,
+        returnStmts: KtReturnExpression
+    ): Boolean {
+        val targetFunction = returnStmts.getTargetFunction(bindingContext)
+            ?: return false
+
+        val targetFunctionBodyExpressionStatements = targetFunction
+            .blockExpressionsOrSingle()
+            .toList()
```
schalkms, line 87: Is this mandatory?

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt
```diff
@@ -66,16 +69,25 @@ class ReturnFromFinally(config: Config = Config.empty) : Rule(config) {
...
 
-    private fun isNotInInnerFunction(
-        returnStmts: KtReturnExpression,
-        childFunctions: Collection<KtNamedFunction>
-    ): Boolean = !returnStmts.parents.any { childFunctions.contains(it) }
+    private fun isReturnFromTargetFunction(
+        blockExpression: KtBlockExpression,
+        returnStmts: KtReturnExpression
+    ): Boolean {
+        val targetFunction = returnStmts.getTargetFunction(bindingContext)
```
schalkms, line 82: I guess we also need to check for an empty binding context here.

detekt-rules-exceptions/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinallySpec.kt
```diff
@@ -19,16 +18,29 @@ class ReturnFromFinallySpec : Spek({
...
+                
+                fun test() {
                 try {
                 } finally {
                     return
                 }
-            }
+                return
+                }
+                }
```
schalkms, line 39: Unfortunately, the formatting seems to be off here.

schalkms:
Thanks! We also need to add a `@requireTypeResolution` annotation to the rule comment, since type and symbol solving needs to be configured.

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt
```diff
@@ -25,9 +31,12 @@ import org.jetbrains.kotlin.psi.psiUtil.parents
  *         return // prevents MyException from being propagated
  *     }
  * }
+ *
+ * val a: String = try { "s" } catch (e: Exception) { "e" } finally { "f" }
  * </noncompliant>
  *
  * @configuration ignoreLabeled - ignores labeled return statements (default: `false`)
+ * @requiresTypeResolution
```
rwqwr, line 41: > We also need to add a `@requireTypeResolution` annotation to the rule comment, since type and symbol solving needs to be configured.

Do you mean this annotation?

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt
rwqwr: I added check to `visitFinallySection()`
Is it ok?

detekt-rules-exceptions/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinallySpec.kt
rwqwr: Sorry, it is my mistake :)

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt
rwqwr: `blockExpressionsOrSingle()` extension return `Sequence<KtElement>` and `isInsideOf` take `Iterable<PsiElement>` as argument.

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ReturnFromFinally.kt
schalkms: Exactly

schalkms (approved):

PR #3229, opened 2020-11-12
Enable more rules by default

### LINKED ISSUE #2838: Enable more rules by default
The same idea as #1911 9 months later :P. All the rules that we add, we add them disabled by default. So we need to enable by default at some point. Any way we should be carefull here as artur say in the linked issue:

> No "highly subjective" rules should be enabled

I extracted a list of rules that I have enabled by default in my projects and I think that they are "safe":

- VarCouldBeVal
- UseRequire
- UseCheckOrError
- UnusedPrivateMember
- UnnecessaryLet
- UnnecessaryApply
- UnderscoresInNumericLiterals
- SerialVersionUIDInSerializableClass
- RedundantVisibilityModifierRule
- UnsafeCast
- UnnecessarySafeCall
- UnnecessaryNotNullOperator
- ImplicitDefaultLocale
- IgnoredReturnValue
- InvalidPackageDeclaration
- ThrowingExceptionsWithoutMessageOrCause
- SwallowedException
- RethrowCaughtException
- ExceptionRaisedInUnexpectedLocation

I'm not saying that we should enable all of these nor any of the others. This is just a first draft so we have a starting point to decide.

What do you think about these ones? Should we add some other?

schalkms:
This is a first draft for enabling more detekt rules by default.
This way users can benefit from a good default config.

Closes #2838


detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -468,11 +468,11 @@ potential-bugs:
   HasPlatformType:
     active: false
   IgnoredReturnValue:
-    active: false
+    active: true
```
cortinico, line 471: I'm unsure about this. We had several false positives for this rules, specifically from RxJava users.

cortinico (commented):
We should also recall to annotate with `@active since ...`

detekt-core/src/main/resources/default-detekt-config.yml
arturbosch: Agree here

arturbosch:
The added rules look good to me :+1:
Lets wait for 1.16.0 for this PR, okay?

We have also some more rules enabled in our analysis which are not active by default:
https://github.com/detekt/detekt/issues/2838#issuecomment-726365903

schalkms:
> Lets wait for 1.16.0 for this PR, okay?

That's okay for me. I'll also check the ones mentioned in your last comment.

BraisGabin (commented):
As @cortinico said we need to add the `@active`. Then we could just regenerate the file and the values should be changed automatically.

schalkms:
> we need to add the `@active`

Yep, this is the 2nd step. The 1st step should be the draft of the activated rules.
I’ll add the tag to the rules in the next few days. Fortunately, I had the pleasure to co-create and shape the detekt-verify as well as the -generator back then. 🙂

> typos

Regarding the typos, it’s probably not a good idea to edit the config on the go with a smartphone web browser. Well, we luckily have a good CI system and carefully observing reviewers for those reasons.

Cheers and have a nice weekend,
Markus

BraisGabin:
What do you think about enable the coroutine ones? I think that all are quite safe. The more controversial is `GlobalCoroutineUsage` and the documentation of Kotlin always say "please don't use Global" so it should be safe.

So my list to be enabled is:

- [ ] `GlobalCoroutineUsage`
- [ ] `RedundantSuspendModifier`
- [ ] `SuspendFunWithFlowReturnType`
- [ ] `InvalidPackageDeclaration`
- [ ] `HasPlatformType`
- [ ] `MapGetWithNotNullAssertionOperator`
- [ ] `ForbiddenMethodCall`
- [ ] `ForbiddenVoid`
- [ ] `UnderscoresInNumericLiterals`
- [ ] `UnnecessaryAnnotationUseSiteTarget`
- [ ] `UnnecessaryLet`
- [ ] `UseArrayLiteralsInAnnotations`
- [ ] `UseCheckOrError`
- [ ] `UseIfEmptyOrIfBlank`
- [ ] `UseRequire`
- [ ] `UseRequireNotNull` We should probably merge this one with `UseRequire`. `UseRequire` already handles `check` and `checkNotNull` so there's not a good reason to have this one splitted

Not so sure:

- [ ] `MandatoryBracesIfStatements`
- [ ] `MandatoryBracesLoops`
- [ ] `NullableToStringCall` 🤔 this one is too new but this issue creates bugs so often...

schalkms:
Thanks for your work, @BraisGabin.
I had reasons for not enabling all mentioned rules. Hence, I'm not okay with enabling the following rules.

- `UnderscoresInNumericLiterals` => this one is really a matter of choice. I don't think all agree with using underscores for numeric literals. We shouldn't enable this by default.
- `UnnecessaryLet` => this rule is not battle-tested IMHO. So there might be many false-positives.
- `UseCheckOrError`, `UseRequire` & `UseRequireNotNull`=> we don't even use this throughout detekt's code base consistently. It's a matter of taste which style you use.
- The rules in the **not sure list** should be disabled.

BraisGabin:
> Thanks for your work, @BraisGabin.
> I had reasons for not enabling all mentioned rules. Hence, I'm not okay with enabling the following rules.
> 
> - `UnderscoresInNumericLiterals` => this one is really a matter of choice. I don't think all agree with using underscores for numeric literals. We shouldn't enable this by default.
> - `UnnecessaryLet` => this rule is not battle-tested IMHO. So there might be many false-positives.
> - `UseCheckOrError`, `UseRequire` & `UseRequireNotNull`=> we don't even use this throughout detekt's code base consistently. It's a matter of taste which style you use.
> - The rules in the **not sure list** should be disabled.

I agree with you. We should not enable the rules you list.

cortinico:
> * `UnnecessaryLet` => this rule is not battle-tested IMHO. So there might be many false-positives.

Agree on this

> * ForbiddenMethodCall

I think this should not be enabled as it's a rule users have to configure, with the list of Methods they want to restrict.

BraisGabin:
> I think this should not be enabled as it's a rule users have to configure, with the list of Methods they want to restrict.

We have two default values there for println. And I think that we should forbid that for default.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SerialVersionUIDInSerializableClass.kt
```diff
@@ -37,6 +37,8 @@ import org.jetbrains.kotlin.psi.KtProperty
  *     }
  * }
  * </compliant>
+ *
+ * @active since v1.16.0
```
chao2zhang, line 41: Note that Android context, `Parcelable` is preferred over `Serializable`. This rule is almost a no-op for Android code.

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/PrintStackTrace.kt
```diff
@@ -43,6 +43,8 @@ import org.jetbrains.kotlin.psi.psiUtil.getReceiverExpression
  *     }
  * }
  * </compliant>
+ *
+ * @active since v1.16.0
```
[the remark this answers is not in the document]
BraisGabin: Every rule should have a descripction so any user can know why he/she should enable/disable it. If you think that in this case (or any other) the description is not good enough you can create an issue. Or, even better, if you know how to improve it, create a PR.

But I don't think that we should add any documentation about why a rule is or is not active by default.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SerialVersionUIDInSerializableClass.kt
BraisGabin: I agree that in Android we should not use Serializable, but, if we have, we should use the SerialVersionUID. So I think that it's safe to have this enabled.

We talked some times about an Android rule set. A good rule for there would be to flag any use of `Serializable`.

BraisGabin: I just create a discussion to talk about it: #3303

arturbosch (approved):
@chao2zhang kudos for resolving the conflicts +1

schalkms:
@chao2zhang thanks for pushing this over the finish line.

PR #3230, opened 2020-11-12
Remove already activated rules from detekt.yml

schalkms:
Those rules are already activated in the default-detekt-config.yml.


schalkms:
based on the idea here
https://github.com/detekt/detekt/issues/2838#issuecomment-726365903

arturbosch (approved):

PR #3232, opened 2020-11-15
Using lint- and compileWithContext is perfectly fine

schalkms:
Hence, I removed the comment.

### DIFF
--- a/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt
+++ b/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt
@@ -38,11 +38,6 @@ fun BaseRule.lint(path: Path): List<Finding> {
     return findingsAfterVisit(ktFile)
 }
 
-/**
- * Compare with [compileAndLintWithContext], this does not use [KotlinScriptEngine]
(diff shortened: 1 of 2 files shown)


detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt
```diff
@@ -38,11 +38,6 @@ fun BaseRule.lint(path: Path): List<Finding> {
     return findingsAfterVisit(ktFile)
 }
 
-/**
- * Compare with [compileAndLintWithContext], this does not use [KotlinScriptEngine]
- * to compile. Please use this sparingly: A typical usecase is when KotlinScript
- * failed to compile even when the Kotlin code in [content] is legal.
- */
 fun BaseRule.lintWithContext(
```
schalkms, line 41: This function is now only used once by `compileAndLintWithContext()`.

BraisGabin:
Don't we want to use compile when possible?

schalkms:
> Don't we want to use compile when possible?

That's not possible. If top level are declared with visibility modifiers the Kotlin script engine cannot compile it.

BraisGabin (approved):
Oh! I though this was a comment in the lint extensions. 🤐

PR #3235, opened 2020-11-15
Compile test snippets if context is passed

### LINKED ISSUE #3191: compileAndLint doesn't compile code if additionalContents are passed
The compileAndLint() function in test code does not work correctly if `additionalContents` are passed.
The passed code should be compiled no matter if `additionalContents` are present.
Please take a look at the following investigation below.

------

@BraisGabin Yes, I do. Please take a look at the following lines and closely check line number 68.

https://github.com/detekt/detekt/blob/da6259a15c986d248799cc5f2db308689e4fc68e/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt#L63-L72

_Originally posted by @schalkms in https://github.com/detekt/detekt/pull/3188#discussion_r515520959_

schalkms:
The compileAndLint() function in test code does not work correctly,
if additionalContents are passed.
The code should be compiled no matter if additionalContents are present.

The fix changes the following in RuleExtensions.kt:
  * Only 1 content can be passed to compileAndLintWithContext()
  * Only 1 content is compiled
  * lintWithContext() stays as it is

Test snippets in UnusedImportsSpec.kt are not compiled,
since the Kotlin script engine can't resolve user declared imports.

Test snippets were adapted in IgnoredReturnValueSpec.kt accordingly:
* Compile all test snippets
* Declare the annotation class only at necessary places
* Don't use conflicting package names to satisfy the Kotlin script engine
* Refactor and simplify test snippets

Fixes #3191

### DIFF
--- a/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValueSpec.kt
+++ b/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValueSpec.kt
@@ -19,93 +19,81 @@ object IgnoredReturnValueSpec : Spek({
 
         it("does not report when a function which returns a value is called and the return is ignored") {
             val code = """
-                package test
-
                 fun foo() {
                     listOf("hello")
                 }
             """
-            val findings = subject.compileAndLintWithContext(env, code, checkReturnValueAnnotationCode)
+            val findings = subject.compileAndLintWithContext(env, code)
             assertThat(findings).isEmpty()
         }
 
         it("does not report when a function which returns a value is called before a valid return") {
             val code = """
-                package test
-
                 fun foo() : Int {
                     listOf("hello")
                     return 42
                 }
             """
-            val findings = subject.compileAndLintWithContext(env, code, checkReturnValueAnnotationCode)
+            val findings = subject.compileAndLintWithContext(env, code)
             assertThat(findings).isEmpty()
         }
 
         it("does not report when a function which returns a value is called in chain and the return is ignored") {
             val code = """
-                package test
-
                 fun foo() {
                     listOf("hello").isEmpty().not()
                 }
             """
-            val findings = subject.compileAndLintWithContext(env, code, checkReturnValueAnnotationCode)
+            val findings = subject.compileAndLintWithContext(env, code)
             assertThat(findings).isEmpty()
         }
 
         it("does not report when a function which returns a value is called before a semicolon") {
             val code = """
-                package test
-
                 fun foo() {
                     listOf("hello");println("foo")
                 }
             """
-            val findings = subject.compileAndLintWithContext(env, code, checkReturnValueAnnotationCode)
+            val findings = subject.compileAndLintWithContext(env, code)
             assertThat(findings).isEmpty()
         }
 
         it("does not report when a function which returns a value is called after a semicolon") {
             val code = """
-                package test
-
                 fun foo() {
                     println("foo");listOf("hello")
                 }
             """
-            val findings = subject.compileAndLintWithContext(env, code, checkReturnValueAnnotationCode)
+            val findings = subject.compileAndLintWithContext(env, code)
             assertThat(findings).isEmpty()
         }
 
         it("does not report when a function which returns a value is called between comments") {
             val code = """
-                package test
-
                 fun foo() {
                     listOf("hello")//foo
                 }
             """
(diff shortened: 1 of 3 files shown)


detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt
```diff
@@ -60,14 +60,13 @@ fun BaseRule.lintWithContext(
 }
 
 fun BaseRule.compileAndLintWithContext(
-    environment: KotlinCoreEnvironment,
-    @Language("kotlin") content: String,
-    @Language("kotlin") vararg additionalContents: String,
```
arturbosch, line 65: Removing this property will break code from users of `detekt-test`.
This property was contributed by someone from uber.
Should we print a warning on the console when `additionalContents` are passed here?
Calling this function with `additionalContents` is basically just `lintWithContext` right?

schalkms: > Removing this property will break code from users of detekt-test.

I'd happily acknowledge this. 
Why? - Because we don't break anything. Please take a look at my comment underneath.
If `additionalContents` is passed, we don't compile anything here.
Hence, the purpose of the method becomes invalid.

> This property was contributed by someone from uber.

chao2zhang contributed it thankfully, as a result of not being able to compile `private` top level function with the script engine. This hindered the execution of the following test snippet for the `LibraryCodeMustSpecifyReturnType` rule.

https://github.com/detekt/detekt/blob/415ca77591ac2cb9a2ce31ded8de0d3d1721754a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/LibraryCodeMustSpecifyReturnTypeSpec.kt#L99-L104

> Should we print a warning on the console when additionalContents are passed here?

No, because the function is not released in any detekt version as of now. It was contributed on 23.10.2020.
Therefore, we don't break any user code. Please take a look at the following commit.
2705f090797a8212c23d883cffb9ba154fa7727c

> Calling this function with additionalContents is basically just lintWithContext right?

Yes, which is a bit weird. The function name suggests that code is being compiled. This is unfortunately not the case.

arturbosch: Please take another look at your linked commit:
chao2zhang contributed `fun BaseRule.lintWithContext` as a new function.
The diff obscures this by changing `compileAndLintWithContext` to `lintWithContext` and adds "the new function".
So this turns out to be a breaking change, right?

schalkms: True. Yes, this is breaking. In this case, we fix a bug and so I'm not concerned about the changing the public API, which would only affect user **test** code. Furthermore, this can easily be fixed by calling `lintWithContext` instead of `compileAndLintWithContext`. With this change no behavioral change is done.

arturbosch: Ok lets do this. I tag it with migration tag. :+1:

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValueSpec.kt
```diff
@@ -643,33 +665,25 @@ object IgnoredReturnValueSpec : Spek({
...
+            val findings = subject.compileAndLintWithContext(env, code)
             assertThat(findings).hasSize(1)
-            assertThat(findings).hasSourceLocation(6, 5)
+            assertThat(findings).hasSourceLocation(4, 5)
             assertThat(findings[0]).hasMessage("The call listOfChecked is returning a value that is ignored.")
         }
     }
 })
-
-private const val checkReturnValueAnnotationCode = """
```
arturbosch, line 671: Lets keep one test with an additional source file?

arturbosch (approved):

BraisGabin (approved):

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValueSpec.kt
schalkms: good point, thx 🙂
DONE in 68937a2

cortinico (approved):

PR #3237, opened 2020-11-16
Exclude detekt:LargeClass rule in test sources

schalkms:


BraisGabin (approved):
I have this change in all my projects too. Would it have sense to add this change to the default config?

schalkms:
> I have this change in all my projects too. Would it have sense to add this change to the default config?

I think this heavily depends on the project setup and the used test framework. In general I would wait for more requests and feedback from the community. If they share the same opinion, we can for sure go this way. At the moment, I’d leave it as it is, since it’s a sensible default.

PR #3247, opened 2020-11-24
Introduce --all-rules flag

arturbosch:
Part one to replace and deprecate `--fail-fast` in the long run - #2267

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/AllRulesConfig.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/AllRulesConfig.kt
@@ -14,7 +14,6 @@ internal data class AllRulesConfig(
     override fun <T : Any> valueOrDefault(key: String, default: T): T {
         return when (key) {
             "active" -> originalConfig.valueOrDefault(key, true) as T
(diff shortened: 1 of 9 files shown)


BraisGabin (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/AllRulesConfig.kt
```diff
@@ -3,24 +3,24 @@ package io.gitlab.arturbosch.detekt.core.config
 import io.gitlab.arturbosch.detekt.api.Config
 
 @Suppress("UNCHECKED_CAST")
-data class FailFastConfig(private val originalConfig: Config, private val defaultConfig: Config) :
```
cortinico, line 6: nit: Should this be considered a breaking change (not worth a major bump but a mention in the release notes)?

cortinico (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/AllRulesConfig.kt
arturbosch: Yes, thanks. I marked https://github.com/detekt/detekt/pull/3163 as `migration` instead.

PR #3253, opened 2020-11-28
Introduce --all-rules flag

arturbosch:
Part one to replace and deprecate `--fail-fast` in the long run - #2267

@BraisGabin @cortinico  I accidentally merged #3247 and needed to revert it. So this is the same as #3247.
Now really for 1.16.0.


cortinico (approved):

BraisGabin (approved):

chao2zhang:
I got confused when reading the code and documentation around `failFast` as they are at odds. Thank you for publishing this to wipe off such confusion!

PR #3256, opened 2020-11-29
Add DestructuringDeclarationWithTooManyEntries rule

3flex:
This rule checks for too many destructuring entries when using destructuring declarations. I've seen IntelliJ in the past recommend using destructuring on data classes with IMHO far too many fields, making the suggested code harder to understand. It's also [suggested by Kotlin's team](https://www.reddit.com/r/Kotlin/comments/ji9z19/kotlin_team_ama_2_ask_us_anything/ga5abq3/) that destructuring be used mostly for things like key/value pairs with only a very small number of elements.

The default maximum of 3 was chosen because it allows easy use of Kotlin stdlib `Pair` and `Triple`, and 4 seemed too many.

There weren't any violations in detekt's codebase when I was testing :)

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/DestructuringDeclarationWithTooManyEntriesSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/DestructuringDeclarationWithTooManyEntriesSpec.kt
@@ -1,0 +1,80 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.test.TestConfig
+import io.gitlab.arturbosch.detekt.test.lint
+import org.assertj.core.api.Assertions.assertThat
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+
+class DestructuringDeclarationWithTooManyEntriesSpec : Spek({
+    val subject by memoized { DestructuringDeclarationWithTooManyEntries() }
+
+    describe("DestructuringDeclarationWithTooManyEntries rule") {
+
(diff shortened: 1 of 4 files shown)


BraisGabin:
I love it! We clearly need rules about destructuring declarations. Could you extend this issue for destructuring in lambda paramenters? 

```kotlin
hello { (_, name) ->
  println("Hello $name")
}
```

These parenthesis are nearly invisible for some people because in java that those parenthesis do nothing.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/DestructuringDeclarationWithTooManyEntriesSpec.kt
```diff
@@ -0,0 +1,80 @@
...
+                    println(x)
+                    println(y)
+                    
+                    val (a, b, c) = Triple(1, 2, 3)
+                    println(a)
+                    println(b)
+                    println(c)
+                }
+            """.trimIndent()
+                assertThat(subject.lint(code)).isEmpty()
```
cortinico, line 29: Please replace with `compileAndLint` or `compileAndLintWithContext`

cortinico (approved):
Great addition 🚀

schalkms (commented):
I like this new rule and would like to have it in the next release. What’s missing to get this merged?

3flex:
@BraisGabin 

> Could you extend this issue for destructuring in lambda paramenters?

The rule picks these up since the compiler treats destructuring declarations in lambdas the same way as when using `val (x, y) = ...`. I've added a couple of tests though to show this.

This should be ready to merge now.

BraisGabin (approved):

PR #3258, opened 2020-11-29
Add ExitOutsideMain rule

3flex:
A port of [Error Prone's SystemExitOutsideMain rule](http://errorprone.info/bugpattern/SystemExitOutsideMain).

The rule also flags usage of `System.exit` or `exitProcess` that are in `main` but in a nested function. It's debatable whether it makes sense to flag usage in the nested function since it's technically in the `main` function and maybe not really an issue. I'm looking for feedback on that. If the logic stays as originally submitted then I'll try and come up with a better name for the rule.

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ExitOutsideMain.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ExitOutsideMain.kt
@@ -1,0 +1,70 @@
+package io.gitlab.arturbosch.detekt.rules.bugs
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.rules.isMainFunction
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+
+/**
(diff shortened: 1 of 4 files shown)


detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ExitOutsideMain.kt
```diff
@@ -0,0 +1,70 @@
...
+ *     exitProcess(2)
+ *   } else {
+ *     exitProcess(0)
+ *   }
+ * }
+ * </compliant>
+ *
+ * @requiresTypeResolution
+ */
+
```
cortinico, line 47: This white line is unnecessary

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ExitOutsideMain.kt
```diff
@@ -0,0 +1,70 @@
...
+    override fun visitCallExpression(expression: KtCallExpression) {
+        if (context == BindingContext.EMPTY) return
+
+        if (expression.getStrictParentOfType<KtNamedFunction>()?.isMainFunction() == true) return
+        val fqName = expression.getResolvedCall(bindingContext)?.resultingDescriptor?.fqNameOrNull() ?: return
+
+        when (fqName.asString()) {
+            "kotlin.system.exitProcess" -> { report(CodeSmell(issue, Entity.from(expression), issue.description)) }
+            "java.lang.System.exit" -> { report(CodeSmell(issue, Entity.from(expression), issue.description)) }
```
cortinico, line 65:
```suggestion
            "kotlin.system.exitProcess", "java.lang.System.exit" -> { report(CodeSmell(issue, Entity.from(expression), issue.description)) }
```

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ExitOutsideMain.kt
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        super.visitCallExpression(expression)
```
cortinico, line 68: This should be moved as first statement

cortinico (approved):

schalkms (commented):
This new rule looks good in my opinion. Is there any chance that the next detekt version includes this rule too?

BraisGabin (approved):
I agree that nested functions should not be flagged. But it seems like a really edge case to me. I'm OK with this or changing the rule to allow those cases, up to you.

3flex:
Thanks for the feedback and sorry for the turnaround time. This should be good to go. I've left the logic as is for the nested functions as I agree it's very much an edge case and IMHO not something you'd want to allow anyway.

PR #3259, opened 2020-11-29
Add MultilineLambdaItParameter rule

adamkobor:
This new rule checks if there is any multiline lambda with a parameter called `it` (whether it's an implicit or an explicit one, it doesn't matter).

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
@@ -13,34 +13,49 @@ import org.jetbrains.kotlin.psi.KtLambdaExpression
 /**
  * Lambda expressions are are very useful in a lot of cases and they often include very small chunks of
  * code using only one parameter. In this cases Kotlin can supply the implicit `it` parameter
- * to make code more concise. However, when you are dealing with multiline lambdas, you might end up with a code
- * that is hard to read if you don't specify the parameter name explicitly.
+ * to make code more concise. However, when you are dealing with lambdas that contain multiple statements,
+ * you might end up with a code that is hard to read if you don't specify the parameter name explicitly.
  *
  * <noncompliant>
  * val digits = 1234.let {
+ *   println(it)
  *   listOf(it)
  * }
  *
  * val digits = 1234.let { it ->
+ *   println(it)
  *   listOf(it)
  * }
  *
  * val flat = listOf(listOf(1), listOf(2)).mapIndexed { index, it ->
+ *   println(it)
  *   it + index
  * }
  * </noncompliant>
  *
  * <compliant>
  * val digits = 1234.let { explicitParameterName ->
+ *   println(explicitParameterName)
  *   listOf(explicitParameterName)
  * }
+ *
  * val lambda = { item: Int, that: String ->
+ *   println(item)
  *   item.toString() + that
  * }
  *
  * val digits = 1234.let { listOf(it) }
+ * val digits = 1234.let {
+ *   listOf(it)
+ * }
  * val digits = 1234.let { it -> listOf(it) }
+ * val digits = 1234.let { it ->
+ *   listOf(it)
+ * }
  * val digits = 1234.let { explicit -> listOf(explicit) }
+ * val digits = 1234.let { explicit ->
+ *   listOf(explicit)
+ * }
  * </compliant>
  */
 class MultilineLambdaItParameter(val config: Config) : Rule(config) {
@@ -52,12 +67,13 @@ class MultilineLambdaItParameter(val config: Config) : Rule(config) {
     override fun visitLambdaExpression(lambdaExpression: KtLambdaExpression) {
         super.visitLambdaExpression(lambdaExpression)
         val parameterNames = lambdaExpression.valueParameters.map { it.name }
-        val isMultiline = lambdaExpression.text.contains("\n")
-        if (isMultiline && (parameterNames.isEmpty() || IT_LITERAL in parameterNames)) {
+        val statementCount = lambdaExpression.bodyExpression?.statements?.size ?: 0
+        if (statementCount > 1 && (parameterNames.isEmpty() || IT_LITERAL in parameterNames)) {
             report(
                 CodeSmell(
                     issue, Entity.from(lambdaExpression),
-                    "The parameter name in a multiline lambda should not be an implicit/explicit `it`. Consider giving your parameter a name"
+                    "The parameter name in a multiline lambda should not be an implicit/explicit `it`. " +
+                            "Consider giving your parameter a name"
                 )
             )
         }
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
@@ -57,7 +57,7 @@ class MultilineLambdaItParameter(val config: Config) : Rule(config) {
             report(
                 CodeSmell(
                     issue, Entity.from(lambdaExpression),
-                    "The parameter name in a multiline lambda should not be an implicit/explicit `it`"
+                    "The parameter name in a multiline lambda should not be an implicit/explicit `it`. Consider giving your parameter a name"
                 )
             )
         }
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
@@ -1,0 +1,65 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.rules.IT_LITERAL
+import org.jetbrains.kotlin.psi.KtLambdaExpression
+
+/**
+ * Lambda expressions are are very useful in a lot of cases and they often include very small chunks of
+ * code using only one parameter. In this cases Kotlin can supply the implicit `it` parameter
+ * to make code more concise. However, when you are dealing with multiline lambdas, you might end up with a code
+ * that is hard to read if you don't specify the parameter name explicitly.
+ *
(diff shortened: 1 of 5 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
```diff
@@ -0,0 +1,65 @@
...
+ *   listOf(it)
+ * }
+ *
+ * val digits = 1234.let { it ->
+ *   listOf(it)
+ * }
+ *
+ * val flat = listOf(listOf(1), listOf(2)).mapIndexed { index, it ->
+ *   it + index
+ * }
```
cortinico, line 30: I believe this rule is a bit too strict.
I think we should report only lambda that contains actually more than one line (and not checking for `\n`). 

The reason is that I think detekt should treat:

```
val digits = 1234.let {
    listOf(it)
}
```

and

```
val digits = 1234.let { listOf(it) }
```

equivalent. Otherwise this becomes more a rule for the formatter (like `ktlint`).

I think we should rather report lambdas like this one:

```
val digits = 1234.let {
	println(it)
    listOf(it)
}
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    override fun visitLambdaExpression(lambdaExpression: KtLambdaExpression) {
+        super.visitLambdaExpression(lambdaExpression)
+        val parameterNames = lambdaExpression.valueParameters.map { it.name }
+        val isMultiline = lambdaExpression.text.contains("\n")
+        if (isMultiline && (parameterNames.isEmpty() || IT_LITERAL in parameterNames)) {
+            report(
+                CodeSmell(
+                    issue, Entity.from(lambdaExpression),
+                    "The parameter name in a multiline lambda should not be an implicit/explicit `it`"
```
cortinico, line 60:
```suggestion
                    "The parameter name in a multiline lambda should not be an implicit/explicit `it`. Consider giving your parameter a name"
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameterSpec.kt
```diff
@@ -0,0 +1,91 @@
...
+    describe("MultilineLambdaItParameter rule") {
+        context("single parameter, multiline lambda") {
+            it("reports when parameter name is implicit `it`") {
+                val findings = subject.compileAndLint(
+                    """
+                fun f() {
+                    val digits = 1234.let { 
+                        listOf(it)
+                    }
+                }""")
```
cortinico, line 21: I'd suggest you extract the source code to:

```suggestion
            val code = """
                fun f() {
                    val digits = 1234.let { 
                        listOf(it)
                    }
                }
            """
```

And use `code` to invoke `compileAndLint` afterwards. It makes the test easier to read.

cortinico (commented):
Thanks for your contribution @adamkobor.
I'd love to hear other members feedback on the idea behind your rule. I'm also wondering if we should merge the behavior inside the `ExplicitItLambdaParameter` rule and treat this as a config option.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameterSpec.kt
adamkobor: Makes sense, will fix it 👍

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
adamkobor: Fair enough, I agree that reporting lambdas that are wrapped into multiple lines but only contains a single statement is a bit too strict.

> Thanks for your contribution @adamkobor.
> I'd love to hear other members feedback on the idea behind your rule. I'm also wondering if we should merge the behavior inside the `ExplicitItLambdaParameter` rule and treat this as a config option.

Thanks! Initially, I was thinking about the same (extending `ExplicitItLambdaParameter` somehow), but it seems like a quite different rule for me, speaking of what it wants to enforce. Even if we could merge the two, it would be very hard to find a good name for the new rule IMO :) 
However, I'm open to any kind of suggestion about this if you have a better idea 🙌

adamkobor:
> Thanks for your contribution @adamkobor.
> I'd love to hear other members feedback on the idea behind your rule. I'm also wondering if we should merge the behavior inside the `ExplicitItLambdaParameter` rule and treat this as a config option.

Thanks! Initially, I was thinking about the same (extending `ExplicitItLambdaParameter` somehow), but it seems like a quite different rule for me, speaking of what it wants to enforce. Even if we could merge the two, it would be very hard to find a good name for the new rule IMO :) 
However, I'm open to any kind of suggestion about this if you have a better idea 🙌

BraisGabin:
Nice rule!

I think that they are different rules. We need new buckets to define rules because this one helps readability but the other just removes redundant code. 

I would keep this one as a different rule.

And should we care if the code defines the parameter as `it`? It's just an open question. I have no idea which one is the bette option.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
```diff
@@ -0,0 +1,81 @@
...
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.rules.IT_LITERAL
+import org.jetbrains.kotlin.psi.KtLambdaExpression
+
+/**
+ * Lambda expressions are are very useful in a lot of cases and they often include very small chunks of
```
chao2zhang, line 14:
```suggestion
 * Lambda expressions are very useful in a lot of cases and they often include very small chunks of
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
```diff
@@ -0,0 +1,81 @@
...
+ * code using only one parameter. In this cases Kotlin can supply the implicit `it` parameter
+ * to make code more concise. However, when you are dealing with lambdas that contain multiple statements,
+ * you might end up with a code that is hard to read if you don't specify the parameter name explicitly.
+ *
```
chao2zhang, line 18: My team also found it not readable to replace `it` parameters with poor names that are too abstract. Such as `data`, `obj` etc. I think should call it out to replace `it` with descriptive and readable names.

I would also recommend replacing the issue message by `Multiline lambdas should use a descriptive name rather than it as the parameter name`

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
```diff
@@ -0,0 +1,81 @@
...
+ * }
+ *
+ * val digits = 1234.let { listOf(it) }
+ * val digits = 1234.let {
+ *   listOf(it)
+ * }
+ * val digits = 1234.let { it -> listOf(it) }
+ * val digits = 1234.let { it ->
+ *   listOf(it)
+ * }
```
chao2zhang, line 54: These are actually code smells, right? https://github.com/detekt/detekt/blob/master/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExplicitItLambdaParameter.kt

chao2zhang: `IT_LITERAL in parameterNames` is already checked in the rule mentioned above, so we don't need to `double report`

chao2zhang (changes requested):

adamkobor:
Sorry for the late answer, I will look into this in the coming days, but we have family issues right now at home with our newborn, so my time is quite limited :)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
adamkobor: Thanks, I totally agree that these rules actually have some overlap, but I think their perspective are quite different. Therefore, I wouldn't remove this overlap, because it's possible that someone is using only one of them. What do you think?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
adamkobor: Makes sense, I've updated the doc block and the issue message as well.

cortinico:
> but we have family issues right now at home with our newborn, so my time is quite limited :)

Sure take your time :) If you need some help just drop us a message

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
chao2zhang: People may enable both rules, and they may get double issues...

That could be our expectation since abusing `it` is lowering the readability significantly.
I am okay with keeping it as is.

adamkobor:
> > but we have family issues right now at home with our newborn, so my time is quite limited :)
> 
> Sure take your time :) If you need some help just drop us a message

Thanks @cortinico, I already made the adjustments you and @chao2zhang requested, so if you have some time, please take a look at it :)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameterSpec.kt
```diff
@@ -0,0 +1,134 @@
...
+        context("single parameter, multiline lambda with a single statement") {
+            it("does not report when parameter name is an implicit `it`") {
+                val code = """
+                fun f() {
+                    val digits = 1234.let { 
+                        listOf(it)
+                    }
+                }"""
+                val findings = subject.compileAndLint(code)
+                assertThat(findings).hasSize(0)
```
cortinico, line 58:
```suggestion
                assertThat(findings).isEmpty()
```

adamkobor:
> Other than minor comments, LGTM 👍

Thanks, I also applied your suggestions 🙌

chao2zhang (approved):

cortinico (approved):

PR #3262, opened 2020-12-01
CI with Java 15

3flex:
14 is EOL

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIfEmptyOrIfBlank.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIfEmptyOrIfBlank.kt
@@ -127,7 +127,8 @@ class UseIfEmptyOrIfBlank(config: Config = Config.empty) : Rule(config) {
             Replacement(FqName("kotlin.collections.Collection.isEmpty"), ifEmpty),
             Replacement(FqName("kotlin.text.isNotBlank"), ifBlank, negativeCondition = true),
             Replacement(FqName("kotlin.text.isNotEmpty"), ifEmpty, negativeCondition = true),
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIfEmptyOrIfBlank.kt
```diff
@@ -127,7 +127,8 @@ class UseIfEmptyOrIfBlank(config: Config = Config.empty) : Rule(config) {
             Replacement(FqName("kotlin.collections.Collection.isEmpty"), ifEmpty),
             Replacement(FqName("kotlin.text.isNotBlank"), ifBlank, negativeCondition = true),
             Replacement(FqName("kotlin.text.isNotEmpty"), ifEmpty, negativeCondition = true),
-            Replacement(FqName("kotlin.collections.isNotEmpty"), ifEmpty, negativeCondition = true)
+            Replacement(FqName("kotlin.collections.isNotEmpty"), ifEmpty, negativeCondition = true),
+            Replacement(FqName("kotlin.String.isEmpty"), ifEmpty)
```
BraisGabin, line 131: Could you move this to another pr? And add a test so we don't have the same error again.

BraisGabin (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIfEmptyOrIfBlank.kt
3flex: I don't think that makes sense because I had to add this to get the build passing with Java 15 in the first place. So that new line is already effectively being tested by CI on Java 15.

PR #3275, opened 2020-12-11
Fix ForbiddenComment rule not checking for KDoc

### LINKED ISSUE #3273: ForbiddenComments don't report TODO: in KDoc
## Expected Behavior

`TODO:` in KDoc should be reported.

## Observed Behavior

`TODO:` in KDoc not be reported.

## Steps to Reproduce

```kotlin
/**
 * TODO: this is not reported
 */
object Test {

    fun testFunc() {
        /**
         * TODO: this is not reported, too.
         */
    }
}
```

`detekt -r xml:result.xml -i Test.kt`

Nothing is reported in result.xml

## Your Environment

* Version of detekt used:  1.14.2 CLI
* Operating System and version: OSX 10.15.4

schalkms:
ForbiddenComment now analyzes KDoc elements for forbidden comments.
KDoc sections don't inherit from the PsiComment type.
Hence, this rule needs to also gather and analyze KDocSection types.

Fixes #3273

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -84,6 +84,7 @@ style:
   ForbiddenComment:
     active: true
     values: ['TODO:', 'FIXME:', 'STOPSHIP:', '@author']
+    excludes: ['**/detekt-rules-style/**/ForbiddenComment.kt']
   LibraryCodeMustSpecifyReturnType:
     active: true
     excludes: ['**/*.kt']
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
@@ -1,15 +1,12 @@
 package io.gitlab.arturbosch.detekt.rules.style
 
-import io.gitlab.arturbosch.detekt.api.CodeSmell
-import io.gitlab.arturbosch.detekt.api.Config
-import io.gitlab.arturbosch.detekt.api.Debt
-import io.gitlab.arturbosch.detekt.api.Entity
-import io.gitlab.arturbosch.detekt.api.Issue
-import io.gitlab.arturbosch.detekt.api.LazyRegex
-import io.gitlab.arturbosch.detekt.api.Rule
-import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.*
 import io.gitlab.arturbosch.detekt.api.internal.valueOrDefaultCommaSeparated
 import org.jetbrains.kotlin.com.intellij.psi.PsiComment
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.kdoc.psi.impl.KDocSection
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
 
 /**
  * This rule allows to set a list of comments which are forbidden in the codebase and should only be used during
@@ -31,7 +28,7 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
 
     override val issue = Issue(javaClass.simpleName,
             Severity.Style,
-            "Flags a forbidden comment. Defaults values are TODO:, FIXME: or STOPSHIP:",
+            "Flags a forbidden comment.",
             Debt.TEN_MINS)
 
     private val values: List<String> = valueOrDefaultCommaSeparated(VALUES, listOf("TODO:", "FIXME:", "STOPSHIP:"))
@@ -40,15 +37,29 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
 
     override fun visitComment(comment: PsiComment) {
         super.visitComment(comment)
-
         val text = comment.text
+        checkForbiddenComment(text, comment)
+    }
 
+    override fun visitKtFile(file: KtFile) {
+        super.visitKtFile(file)
+        file.collectDescendantsOfType<KDocSection>().forEach { comment ->
+            val text = comment.getContent()
+            checkForbiddenComment(text, comment)
+        }
+    }
+
+    private fun checkForbiddenComment(text: String, comment: PsiElement) {
         if (allowedPatterns.pattern.isNotEmpty() && allowedPatterns.containsMatchIn(text)) return
 
         values.forEach {
             if (text.contains(it, ignoreCase = true)) {
-                report(CodeSmell(issue, Entity.from(comment), "This comment contains text that has been " +
-                        "defined as forbidden in detekt."))
+                report(
+                    CodeSmell(
+                        issue, Entity.from(comment), "This comment contains text that has been " +
+                                "defined as forbidden in detekt."
+                    )
+                )
             }
         }
     }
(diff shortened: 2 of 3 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -40,15 +37,29 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
...
     override fun visitComment(comment: PsiComment) {
         super.visitComment(comment)
-
         val text = comment.text
+        checkForbiddenComment(text, comment)
+    }
+
+    override fun visitKtFile(file: KtFile) {
+        super.visitKtFile(file)
+        file.collectDescendantsOfType<KDocSection>().forEach { comment ->
```
arturbosch, line 46: Are there `visit` methods for `KDocSection`?
If yes would this be cleaner than writing a multiline lambda?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -40,15 +37,29 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
...
+    private fun checkForbiddenComment(text: String, comment: PsiElement) {
         if (allowedPatterns.pattern.isNotEmpty() && allowedPatterns.containsMatchIn(text)) return
 
         values.forEach {
             if (text.contains(it, ignoreCase = true)) {
-                report(CodeSmell(issue, Entity.from(comment), "This comment contains text that has been " +
-                        "defined as forbidden in detekt."))
+                report(
+                    CodeSmell(
+                        issue, Entity.from(comment), "This comment contains text that has been " +
```
arturbosch, line 59: Would splitting a parameter per line remove the need for the string concatenation?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -1,15 +1,12 @@
...
 
-import io.gitlab.arturbosch.detekt.api.CodeSmell
-import io.gitlab.arturbosch.detekt.api.Config
-import io.gitlab.arturbosch.detekt.api.Debt
-import io.gitlab.arturbosch.detekt.api.Entity
-import io.gitlab.arturbosch.detekt.api.Issue
-import io.gitlab.arturbosch.detekt.api.LazyRegex
-import io.gitlab.arturbosch.detekt.api.Rule
-import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.*
```
arturbosch: This line may lead to CI failure :)

schalkms: Yes, a fix is in the queue.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -40,15 +44,29 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
...
+    private fun checkForbiddenComment(text: String, comment: PsiElement) {
         if (allowedPatterns.pattern.isNotEmpty() && allowedPatterns.containsMatchIn(text)) return
 
         values.forEach {
             if (text.contains(it, ignoreCase = true)) {
-                report(CodeSmell(issue, Entity.from(comment), "This comment contains text that has been " +
-                        "defined as forbidden in detekt."))
+                report(
+                    CodeSmell(
+                        issue, Entity.from(comment), "This comment contains text that has been " +
```
arturbosch, line 66:
```suggestion
                        issue, Entity.from(comment), "This comment contains '$it' that has been " +
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
schalkms: This is actually the default IntelliJ format.
I used the IntelliJ extract function option.
If you like, I can revert back to the old format.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
arturbosch: We can actually be specific here. This would help when just reading the reports.

arturbosch (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
schalkms: This would then report the whole comment. I'm not sure whether we want to print the whole text. Detekt already outputs the flagged entity.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
arturbosch: Only a nit here. Feel free to leave it that way.
IntelliJ should allow both formats I think. I would prefer a strict mode with an idempotent result :).

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
schalkms: Nope, I couldn't find a separate `visit` method.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
arturbosch: `it` would be in that block `TODO:` for example.
Do you mean it outputs it in the html report?
The console report would just print `"ForbiddenComment - This comment contains text that has been defined as forbidden in detekt."  right?

schalkms: Yes, thanks for the explanation. I'll change it. That makes sense.

config/detekt/detekt.yml
```diff
@@ -84,6 +84,7 @@ style:
   ForbiddenComment:
     active: true
     values: ['TODO:', 'FIXME:', 'STOPSHIP:', '@author']
+    excludes: ['**/detekt-rules-style/**/ForbiddenComment.kt']
```
BraisGabin, line 87: 😂 This is funny we need to exclude the rule from itself.

BraisGabin (approved):

config/detekt/detekt.yml
schalkms: The `Suppress` annotation won't work for comments in this case. That's the consequence here.

PR #3285, opened 2020-12-14
UnconditionalJumpStatementInLoop: don't report a return after a conditional jump

### LINKED ISSUE #3280: False positive "Unconditional loop jump"
Given the following code:
```
fun f(): Int {
    for (i in 0 until 10) {
      val a = i * i
      if (a < 27) continue
      return a
    }
}
```

## Expected Behavior
It should pass the check

## Observed Behavior
It reports "unconditional loop jump" because of `return` statement

## Steps to Reproduce
-

## Context
-

## Your Environment
Used detekt from codefactor.io, see https://www.codefactor.io/repository/github/kotlin/kotlin-jupyter/source/notebook-api/jupyter-lib/kotlin-jupyter-api/src/main/kotlin/org/jetbrains/kotlin/jupyter/api/KotlinKernelVersion.kt#L13

t-kameyama:
Fixes #3280

### DIFF
--- a/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnconditionalJumpStatementInLoopSpec.kt
+++ b/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnconditionalJumpStatementInLoopSpec.kt
@@ -49,11 +49,23 @@ class UnconditionalJumpStatementInLoopSpec : Spek({
             val findings = subject.lint("""
                 fun f(): Int {
                     for (i in 0 until 10) {
-                      val a = i * i
-                      if (a < 27) continue
-                      return a
+                        val a = i * i
+                        if (a < 27) continue
+                        return a
                     }
-                }                
+                    return 0
+                }
+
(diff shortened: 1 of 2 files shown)


schalkms (approved):

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnconditionalJumpStatementInLoopSpec.kt
```diff
@@ -44,5 +44,19 @@ class UnconditionalJumpStatementInLoopSpec : Spek({
 
             assertThat(findings).hasSize(1)
         }
+
+        it("does not report a return after a conditional jump") {
+            val findings = subject.lint("""
+                fun f(): Int {
+                    for (i in 0 until 10) {
+                      val a = i * i
+                      if (a < 27) continue
```
schalkms, line 53: Can we add a `when` expression in order to cover this path?

```Kotlin
private fun PsiElement.isConditionalJumpStatement(): Boolean =
         (this is KtIfExpression || this is KtWhenExpression)
```

PR #3290, opened 2020-12-16
NamedArguments: fix false positive with java method call

### LINKED ISSUE #3289: NamedArguments with java code false positive
## Expected Behavior

This code should not raise any `NamedArguments` issue:
```kotlin
import org.joda.time.DateTime

DateTime(2020, 3, 13, 14, 0, 0)
```

## Observed Behavior
It does

## Context
If you use a java function in kotlin you can't use `NamedArguments` so there is no way to fix this.

## Your Environment
* Version of detekt used: 1.15.0-RC1

t-kameyama:
Fixes #3289

### DIFF
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArguments.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArguments.kt
@@ -8,6 +8,8 @@ import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.ThresholdRule
 import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
 
 /**
  * Reports function invocations which have more parameters than a certain threshold and are all not named.
@@ -25,6 +27,7 @@ import org.jetbrains.kotlin.psi.KtCallExpression
  * </compliant>
  *
  * @configuration threshold - number of parameters that triggers this inspection (default: `3`)
+ * @requiresTypeResolution
  */
 class NamedArguments(
     config: Config = Config.empty,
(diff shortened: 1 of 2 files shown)


chao2zhang (approved):

BraisGabin:
Are we OK about removing features of detekt without type solving? Or do we want to keep the mixed rules? I prefer no mix.

BraisGabin (approved):

cortinico (approved):

cortinico:
> Are we OK about removing features of detekt without type solving? Or do we want to keep the mixed rules? I prefer no mix.

I'm fine with this change. I believe we should push users to use Type Resolution + restrict the rules with mixed behavior (#2994). This rule is going to become a Type Resolution only so it 👍  on my end.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArguments.kt
```diff
@@ -38,8 +41,12 @@ class NamedArguments(
     )
 
     override fun visitCallExpression(expression: KtCallExpression) {
+        if (bindingContext == BindingContext.EMPTY) return
         val valueArguments = expression.valueArguments
-        if (valueArguments.size > threshold && valueArguments.any { !it.isNamed() }) {
+        if (valueArguments.size > threshold &&
+            valueArguments.any { !it.isNamed() } &&
+            expression.getResolvedCall(bindingContext)?.candidateDescriptor?.hasStableParameterNames() == true
```
schalkms, line 48: That's a nice solution. Thank you for that.
Do you know by chance, whether there exists an API to detect Java method calls?

detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArgumentsSpec.kt
```diff
@@ -107,7 +111,19 @@ class NamedArgumentsSpec : Spek({
                 
                 val obj = C(1, 2)
             """
-            val findings = namedArguments.compileAndLint(code)
+            val findings = namedArguments.compileAndLintWithContext(env, code)
+            assertThat(findings).hasSize(0)
+        }
+
+        it("java method invocation") {
```
schalkms, line 118:
```suggestion
        it("java method invocation should not be flagged") {
```

schalkms (approved):

schalkms:
I agree. This shouldn't be a mixed rule.
The goal is to stay away from that, since type and symbol solving should become stable with detekt v2.0.0.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArguments.kt
t-kameyama: `expression.getResolvedCall(bindingContext)?.candidateDescriptor is JavaMethodDescriptor`

schalkms: cool, thanks

Issue #3274, opened 2020-12-10
Configurable severity per rule


chao2zhang:
This topic was previously discussed in #3129 and #533, but I would like to make a proposal with new information.

Some well-known static analysis tools allow the users to configure severity on rule or ruleset:

-  [Checkstyle](https://checkstyle.sourceforge.io/config.html#Severity) allows configuring the severity of each rule to ERROR, WARNING, or IGNORE
- [Android Lint](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-master-dev:lint/libs/lint-api/src/main/java/com/android/tools/lint/detector/api/Severity.kt) allows configuring the severity of each rule to Fatal, Error, Warning, Information, Ignore

This configurable severity level allows the tool itself to control the output and the exit code of the process. In addition, it also cooperates with other tools:
- [Github checks](https://docs.github.com/en/free-pro-team@latest/rest/reference/code-scanning#upload-a-sarif-file) may use [SARIF](https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Ref493404972) output, which has four levels: error, warning, note, none. For example, Error means blocking the PR from being merged. Warning is informational.
- Danger as mentioned in #3129, also defines behavior based on the severity level
- As mentioned in #533, we also want to control the output as well as the result for detekt based on severity level
- At my company, we also have internal tooling to treat these severity levels
    - For example, ERROR means a PR cannot introduce any new violations with ERROR
    - WARNING means that a PR cannot increase the total number of WARNING in the codebase unless it is approved by a specific group of people

Now to support all the usecases, we need to be able to configure the severity per rule in the XML output as well as SARIF output. So far, the only configurable option I can see is `config.warningsAsErrors`, which is all or nothing. And the severity mapping for [XmlOutputReport](https://github.com/detekt/detekt/blob/4ce376e282bfff176d9b5b141462e623aab91660/detekt-report-xml/src/main/kotlin/io/github/detekt/report/xml/XmlOutputReport.kt#L18) and [SarifOutputReport](https://github.com/detekt/detekt/blob/415ca77591ac2cb9a2ce31ded8de0d3d1721754a/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt#L49)  are not flexible enough.

In my opinion, this feature should be implemented before we close #533 and #3045

chao2zhang:
[...]

PR #3292, opened 2020-12-17
Standardize "active" constant

chao2zhang:
While working on #3274, I noticed that I cannot find the scope where I need to add `severity` as a universal config key.
So I created this PR to use Config.ACTIVE_KEY as the root. Please keep in mind that there are still "active" as the string literals not touched in this PR. Those appearances are primarily test code and website generation code.


schalkms (approved):
Thanks for the refactoring. Looks good to me. 🙂

cortinico (approved):

BraisGabin (approved):

PR #3298, opened 2020-12-19
General housekeeping

chao2zhang:
Please let me know if this type of PR is preferred. I tend to move minor refactoring into a separate PR when making fairly large changes.

- Refactor more config keys as key companion object
- Convert some KDocs from using # into using []
- Remove self-recursive function in Context
- Fix correct dokka task name in CONTRIBUTING.md

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
@@ -56,7 +56,6 @@ interface Config {
 
         const val ACTIVE_KEY: String = "active"
         const val AUTO_CORRECT_KEY: String = "autoCorrect"
-        const val MAX_ISSUES_KEY: String = "maxIssues"
         const val EXCLUDES_KEY: String = "excludes"
         const val INCLUDES_KEY: String = "includes"
         const val CONFIG_SEPARATOR: String = ">"
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
@@ -55,6 +55,8 @@ interface Config {
         val empty: Config = EmptyConfig
 
         const val ACTIVE_KEY: String = "active"
+        const val AUTO_CORRECT_KEY: String = "autoCorrect"
+        const val MAX_ISSUES_KEY: String = "maxIssues"
         const val EXCLUDES_KEY: String = "excludes"
         const val INCLUDES_KEY: String = "includes"
         const val CONFIG_SEPARATOR: String = ">"
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Context.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Context.kt
@@ -24,21 +24,20 @@ interface Context {
     fun report(finding: Finding, aliases: Set<String> = emptySet(), ruleSetId: RuleSetId? = null)
 
     /**
+     * Same as [report] but reports a list of [findings].
+     */
+    fun report(findings: List<Finding>,
+               aliases: Set<String> = emptySet(),
+               ruleSetId: RuleSetId? = null
+    ) {
+        findings.forEach { report(it, aliases, ruleSetId) }
+    }
+
+    /**
      * Clears previous findings.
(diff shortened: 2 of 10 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Context.kt
```diff
@@ -21,16 +21,7 @@ interface Context {
...
-        report(finding, aliases, null)
-    }
-
-    /**
-     * Same as [report] but reports a list of findings.
-     */
-    fun report(findings: List<Finding>, aliases: Set<String> = emptySet(), ruleSetId: RuleSetId? = null) {
-        report(findings, aliases, null)
-    }
+    fun report(finding: Finding, aliases: Set<String> = emptySet(), ruleSetId: RuleSetId? = null)
```
BraisGabin, line 24: This is the api module, we can't remove a function from an interface, it's a breaking change. We could deprecate it but I think that for 2.0 we will remove this class completely so I think that we should keep it as it is. This way we don't force to our users to change it twice.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
```diff
@@ -55,6 +55,8 @@ interface Config {
         val empty: Config = EmptyConfig
 
         const val ACTIVE_KEY: String = "active"
+        const val AUTO_CORRECT_KEY: String = "autoCorrect"
+        const val MAX_ISSUES_KEY: String = "maxIssues"
```
BraisGabin, line 59: Should MAX_ISSUES_KEY here? It's not related with out public API an it's not used in this module either. I would keep it in `IssueExtension.kt`

AUTO_CORRECT_KEY is good.

BraisGabin (commented):
Sure! This PRs are more than welcome. Please check my comments.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Context.kt
chao2zhang: I moved it to an extension method, strictly it is not backward compatible but I would be surprised if people are overriding it. I reverted my change on the removal here.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
chao2zhang: I agree that we should think more about the necessity of adding public APIs.
However, I regard `default-detekt-config.yml` as a public API and all the keys defined there should be exposed in the programming interface. Therefore I still believe moving it here has the benefit of standardizing (MAX_ISSUE_KEY is also used by two files).

Let me know if you think otherwise.

schalkms (approved):

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
BraisGabin: I don't see any usage of `MAX_ISSUE_KEY` in `:detekt-api`. I only see usages in `:detekt-core`. Tell me if I'm loosing something here.

Sure, all the content in `default-detekt-config.yml` is public API. But you should not relay on that value because we could change it. Something like #2778.

I agree with you in the other comment that probably no one is using that or it will use this variable. But as soon as we open it we should support it so it's better to keep it as minimal as possible.

chao2zhang: Reverted this change and moved the constant back to IssueExtension.kt

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/IssueExtension.kt
```diff
@@ -11,7 +11,7 @@ import org.jetbrains.kotlin.com.intellij.openapi.util.Key
 
 private val WEIGHTED_ISSUES_COUNT_KEY = Key.create<Int>("WEIGHTED_ISSUES_COUNT")
 private const val WEIGHTS = "weights"
-internal const val MAX_ISSUES = "maxIssues"
+const val MAX_ISSUES_KEY: String = "maxIssues"
```
BraisGabin, line 14:
```suggestion
internal const val MAX_ISSUES_KEY: String = "maxIssues"
```

BraisGabin (approved):

PR #3299, opened 2020-12-19
Magic number extension functions

BraisGabin:
We should allow things like:
```kotlin
500.dp()
5.days
```
Because they are not magic numbers, they have units.

More about this in #3283

This PR adds a new flag to MagicNumber to allow this. Open questions:

- Should we enable this flag by default?
- Is `ignoreExtensionFunctions` a good name?


schalkms (approved):

BraisGabin:
I changed the PR a bit: I enabled `ignoreExtensionFunctions` by default (last commit)

PR #3300, opened 2020-12-19
Move the questions to discussions

BraisGabin:
The idea is to move the questions about detekt outside our issue tracker and make the disccusion section more relevant.

This will appear when a user clicks "create an issue":

Captura de pantalla 2020-12-19 a las 12 47 50


schalkms (approved):
Thanks! Should we also move existing issues, which are discussions?

BraisGabin:
> Thanks! Should we also move existing issues, which are discussions?

Can we do that? If so, I think that we should. This way our issue tracker would be more actionable.

schalkms:
@BraisGabin yes, it's possible. I haven't tried it though.

image

cortinico (approved):

PR #3306, opened 2020-12-19
[POC] Support rule severity overwriting via config properties

arturbosch:
#3274


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
```diff
@@ -13,6 +13,9 @@ interface Finding : Compactable, HasEntity, HasMetrics {
     val references: List<Entity>
     val message: String
 
+    val severity: String
+        get() = issue.severity.toString().toLowerCase()
+
```
chao2zhang, line 18: I think we would like to achieve the following:
- Each rule author should be able to define a default severity per issue
- Users can override the severity through config.yml

My local draft accomplishes the second one, but the first one is a backwards incompatible change and is not yet completed.

PR #3309, opened 2020-12-20
Don't mix kotlin 1.3 and 1.4

BraisGabin:
`com.android.tools.build:gradle:4.0.1` uses kotlin-reflect. And it uses the version 1.3.72. We should not mix kotlin-reflect 1.3 and kotlin-reflect 1.4.

With this constraint we force kotlin-reflect to update.

We should remove this constraint as soon as we update to a version of `com.android.tools.build:gradle` that uses kotlin 1.4

This change does little change until we update to gradle 6.8 because we still have the warning but it's a step in the right direction.


chao2zhang:
AGP 4.2 is the first version using Kotlin 1.4, which just enters beta02. The expectation for the final release would be April 2021.

arturbosch (approved):

PR #3310, opened 2020-12-20
Support configurable severity per ruleset/rule in XML and Sarif output

chao2zhang:
Summary
=======
This PR implements configuring the severity in the XML and Sarif output, which is the first step to implement #3274. The implementation is also different from #3306, because the severity in the configuration can only be a simplified version and does not allow custom strings.

Implementation
===========
Adding the `severity` as a property of `Issue` does not work: Each rule initializes the `issue` first then uses the `issue.id` to initialize `ruleConfig`. If we want to use `ruleConfig` to parse `severity`, it has to be done after `issue` is created.
We can break such cycling dependency by refactoring, but I would prefer doing so in the next PR where we can remove the existing `Severity` and replace it with the new `SeverityLevel`.

The choice here is to add the `severity` as a property of `Finding`, which is appended at the end to mitigate incompatibility. However, since `Finding` and its children are not data class, I need to create `copyWithSeverity` to override the given severity.

Followup
===========
I did not deprecate or remove `Severity`, `maxIssues`, `warningsAsErrors` in this PR. If this is a good time to do so, I would be happy to do those in a follow-up PR.

Test
====
In addition to the unit test, I also performed an integration test in a different repo. Users can configure severity at ruleset level and rule level.
```
style:
  severity: error
  MagicNumber:
    severity: ignore
```

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
@@ -15,17 +15,12 @@ open class CodeSmell(
     override val entity: Entity,
     override val message: String,
     override val metrics: List<Metric> = listOf(),
-    override val references: List<Entity> = listOf(),
-    override val severity: SeverityLevel = SeverityLevel.WARNING
+    override val references: List<Entity> = listOf()
 ) : Finding {
 
-    constructor(
-        issue: Issue,
-        entity: Entity,
-        message: String,
-        metrics: List<Metric> = listOf(),
-        references: List<Entity> = listOf()
-    ) : this(issue, entity, message, metrics, references, SeverityLevel.WARNING)
+    internal var internalSeverity: SeverityLevel = SeverityLevel.WARNING
+    override val severity
+        get() = internalSeverity
 
     override val id: String = issue.id
 
@@ -44,20 +39,6 @@ open class CodeSmell(
     }
 
     override fun messageOrDescription(): String = if (message.isEmpty()) issue.description else message
-
-    /**
-     * Create a new copy of [CodeSmell] with specified [severity].
-     */
-    override fun copyWithSeverity(severity: SeverityLevel): Finding {
-        return CodeSmell(
-            issue = this.issue,
-            entity = this.entity,
-            message = this.message,
-            metrics = this.metrics,
-            references = this.references,
-            severity = severity
-        )
-    }
 }
 
 /**
@@ -65,32 +46,20 @@ open class CodeSmell(
  *
  * @see CodeSmell
  */
-@Suppress("LongParameterList")
 open class CorrectableCodeSmell(
     issue: Issue,
     entity: Entity,
     message: String,
     metrics: List<Metric> = listOf(),
     references: List<Entity> = listOf(),
-    val autoCorrectEnabled: Boolean,
-    severity: SeverityLevel = SeverityLevel.WARNING
+    val autoCorrectEnabled: Boolean
 ) : CodeSmell(
     issue = issue,
     entity = entity,
     message = message,
     metrics = metrics,
-    references = references,
-    severity = severity
+    references = references
 ) {
-
-    constructor(
-        issue: Issue,
-        entity: Entity,
-        message: String,
-        metrics: List<Metric> = listOf(),
-        references: List<Entity> = listOf(),
-        autoCorrectEnabled: Boolean
-    ) : this(issue, entity, message, metrics, references, autoCorrectEnabled, SeverityLevel.WARNING)
 
     override fun toString(): String {
         return "CorrectableCodeSmell(" +
@@ -102,18 +71,6 @@ open class CorrectableCodeSmell(
             "references=$references, " +
             "severity=$severity, " +
             "id='$id')"
-    }
-
-    override fun copyWithSeverity(severity: SeverityLevel): Finding {
-        return CorrectableCodeSmell(
-            issue = this.issue,
-            entity = this.entity,
-            message = this.message,
-            metrics = this.metrics,
-            references = this.references,
-            autoCorrectEnabled = this.autoCorrectEnabled,
-            severity = severity
-        )
     }
 }
 
@@ -129,23 +86,13 @@ open class ThresholdedCodeSmell(
     val metric: Metric,
     message: String,
     references: List<Entity> = emptyList(),
-    severity: SeverityLevel = SeverityLevel.WARNING
 ) : CodeSmell(
     issue = issue,
     entity = entity,
     message = message,
     metrics = listOf(metric),
     references = references,
-    severity = severity
 ) {
-
-    constructor(
-        issue: Issue,
-        entity: Entity,
-        metric: Metric,
-        message: String,
-        references: List<Entity> = listOf()
-    ) : this(issue, entity, metric, message, references, SeverityLevel.WARNING)
 
     val value: Int
         get() = metric.value
@@ -155,15 +102,4 @@ open class ThresholdedCodeSmell(
     override fun compact(): String = "$id - $metric - ${entity.compact()}"
 
     override fun messageOrDescription(): String = if (message.isEmpty()) issue.description else message
-
-    override fun copyWithSeverity(severity: SeverityLevel): Finding {
-        return ThresholdedCodeSmell(
-            issue = this.issue,
-            entity = this.entity,
-            metric = this.metric,
-            message = this.message,
-            references = this.references,
-            severity = severity
-        )
-    }
 }
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
@@ -54,13 +54,12 @@ open class CorrectableCodeSmell(
     references: List<Entity> = listOf(),
     val autoCorrectEnabled: Boolean
 ) : CodeSmell(
-    issue = issue,
-    entity = entity,
-    message = message,
-    metrics = metrics,
-    references = references
+    issue,
+    entity,
+    message,
+    metrics,
+    references
 ) {
-
     override fun toString(): String {
         return "CorrectableCodeSmell(" +
             "autoCorrectEnabled=$autoCorrectEnabled," +
@@ -85,13 +84,9 @@ open class ThresholdedCodeSmell(
     entity: Entity,
     val metric: Metric,
     message: String,
-    references: List<Entity> = emptyList(),
+    references: List<Entity> = emptyList()
 ) : CodeSmell(
-    issue = issue,
-    entity = entity,
-    message = message,
-    metrics = listOf(metric),
-    references = references,
+    issue, entity, message, metrics = listOf(metric), references = references
 ) {
 
     val value: Int
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
@@ -15,7 +15,8 @@ open class CodeSmell(
     override val entity: Entity,
     override val message: String,
     override val metrics: List<Metric> = listOf(),
-    override val references: List<Entity> = listOf()
+    override val references: List<Entity> = listOf(),
+    override val severity: SeverityLevel = SeverityLevel.WARNING
 ) : Finding {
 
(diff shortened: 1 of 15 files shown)


chao2zhang:
I also realized that the severity, for now, is only in XML + Sarif. We are not differentiating severities in HTML or console output. If this is a good idea, please react to this comment so I will open a new issue for supporting them.

1. For txt, the output should include the severity level. 
2. For console, we should color-code each finding in addition to showing the severity.
3. For HTML, we should both display the severity and use a different background.

BraisGabin:
I agree with all except html. I don't think that we should change all the background. We could use a label and only change that label color.

schalkms (approved):
Looks good!
Anyhow, a refactoring would be great in separate PR in order to implement the severity levels properly and keep the architecture tidy. Then there's no need for copying stuff around and no risk to introduce cyclic dependencies with future refactorings.

> We can break such cycling dependency by refactoring, but I would prefer doing so in the next PR where we can remove the existing Severity and replace it with the new SeverityLevel.

schalkms:
> If this is a good idea, please react to this comment so I will open a new issue for supporting them.
> 1. For txt, the output should include the severity level.
> 2. For console, we should color-code each finding in addition to showing the severity.
> 3. For HTML, we should both display the severity and use a different background.

+1 great
I agree with this suggestion. I wouldn't want to build it overly complicated, so simplicity is the key here.

chao2zhang:
> Looks good!
> Anyhow, a refactoring would be great in separate PR in order to implement the severity levels properly and keep the architecture tidy. Then there's no need for copying stuff around and no risk to introduce cyclic dependencies with future refactorings.
> 
> > We can break such cycling dependency by refactoring, but I would prefer doing so in the next PR where we can remove the existing Severity and replace it with the new SeverityLevel.

#3127 prototyped a new Rule API, which is much cleaner, but also not backward compatible. In order to keep the severity within every `Issue`, it will be a breaking change and I would prefer to wait till Detekt 2.0 for making such changes.

I have seen Detekt 2.0 mentioned in a few issues. Do we know when and how we are planning this? I imagine there will be a separate 2.0 branch where we can merge all breaking changes.

schalkms:
> #3127 prototyped a new Rule API, which is much cleaner, but also not backward compatible. In order to keep the severity within every Issue, it will be a breaking change and I would prefer to wait till Detekt 2.0 for making such changes.
>
> I have seen Detekt 2.0 mentioned in a few issues. Do we know when and how we are planning this? I imagine there will be a separate 2.0 branch where we can merge all breaking changes.

At the moment those PRs are tagged in the **milestone** field with the corresponding detekt version.
Nevertheless, with the number of those PRs growing I suggest to create a new v2.0 branch for simplicity reasons.

BraisGabin:
We talk a lot about 2.0 but we don't have the list of things that we want to change. So I'm not sure about creating a new branch. #3127 pretends to be retrocompatible. I mean, the idea is to deprecate the old api and keep it until 2.0.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
```diff
@@ -64,18 +65,34 @@ abstract class Rule(
...
+    /**
+     * Compute severity in the priority order:
+     * - Severity of the rule
+     * - Severity of the parent ruleset
+     * - Default severity: Error
+     */
+    private fun computeSeverity(): SeverityLevel {
+        val configValue: String = valueOrNull(SEVERITY_KEY)
+            ?: ruleSetConfig.valueOrDefault(SEVERITY_KEY, "error")
+        return enumValueOf(configValue.toUpperCase())
```
arturbosch, line 77: All `Finding` implementation use `Warning` as default which should be backwards compatible (checkstyle format is consumed by jenkins warnings plugin). Defaulting to error here may break CI for users of the jenkins plugin.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
```diff
@@ -12,11 +12,17 @@ interface Finding : Compactable, HasEntity, HasMetrics {
...
 
     /**
      * Explanation why this finding was raised.
      */
     fun messageOrDescription(): String
+
+    /**
+     * Copy function to assign a new severity.
+     */
+    fun copyWithSeverity(severity: SeverityLevel): Finding
```
arturbosch, line 25: Having a default implementation which returns self would make this backwards compatible with 1.x?

arturbosch (changes requested):
Implementation looks good!
I think this could be backwards compatible for 1.x, please see my two comments

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
chao2zhang: Good point

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
chao2zhang: Great catch. I updated the default severity in `TestFactory.createFinding()` but forgot to update it here.

arturbosch (approved):

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
```diff
@@ -78,9 +110,15 @@ open class ThresholdedCodeSmell(
     entity: Entity,
     val metric: Metric,
     message: String,
-    references: List<Entity> = emptyList()
+    references: List<Entity> = emptyList(),
+    severity: SeverityLevel = SeverityLevel.WARNING
```
BraisGabin, line 114: I do little about code compatibility, but this will break binary compatibility, right? I mean, we are going to force to all our users to recompile their plugins and upload a new version.

If we want to fix it I think that we should create a secundary constructor that have exactly the same signature as the previous one. The binary compatibility in Kotlin is messy... Maybe we could definde some number of minor release where we provide binary compatibility. This way we could deprecate this secundary constructor and give it a version or time when it will be removed. We can even create a custom rule to check this.

BraisGabin (commented):
Great PR! I'm just a bit concerned about the compatibility.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
arturbosch: Good catch! We also used a secondary constructor for `Entity` and co.

chao2zhang: - Adding a default argument maintains the source compatibility, but breaks the binary compatibility
  [Maintaining compatibility in Kotlin libraries](https://zsmb.co/maintaining-compatibility-in-kotlin-libraries/)
- We could use [Kotlin binary compatibility](https://github.com/Kotlin/binary-compatibility-validator) to validate API breaking changes.
- We could also enable [explicit API mode](https://kotlinlang.org/docs/reference/whatsnew14.html#explicit-api-mode-for-library-authors) to make sure we are not accidentally exposing API.
- In this particular case, let me fix as you suggested.

BraisGabin: @chao2zhang, could you add your ideas in #3315? I just created it so we can automatice this. It is impossible to catch these breaking changes just by eye.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
```diff
@@ -78,11 +128,25 @@ open class ThresholdedCodeSmell(
...
-    issue, entity, message, metrics = listOf(metric), references = references
+    issue = issue,
+    entity = entity,
+    message = message,
+    metrics = listOf(metric),
+    references = references,
+    severity = severity
 ) {
 
+    constructor(
```
BraisGabin, line 142:
```suggestion
    @Deprecated("This constructor is provided for binary compatibility, use the primary constructor instead")
    constructor(
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
```diff
@@ -78,11 +128,25 @@ open class ThresholdedCodeSmell(
     entity: Entity,
     val metric: Metric,
     message: String,
-    references: List<Entity> = emptyList()
+    references: List<Entity> = emptyList(),
+    severity: SeverityLevel = SeverityLevel.WARNING
```
BraisGabin, line 132: Do we want a default value now that we don't need it for source compatibility?

chao2zhang: The concern here is that I do not expect rule authors to override the severity here.

Here is another idea: Rather than exposing `Finding.copyWithSeverity()` in the interface, we could create the severity as a `var` with `internal set()`, this way only `Rule` is able to mutate it and we further minimize the exposed APIs.

chao2zhang: Check out the new diff I published.

BraisGabin (approved):

chao2zhang:
This PR also depends on #3320

detekt-report-sarif/src/test/kotlin/io/github/detekt/report/sarif/SarifOutputReportSpec.kt
```diff
@@ -19,9 +24,10 @@ class SarifOutputReportSpec : Spek({
 
         it("renders multiple issues") {
             val result = TestDetektion(
-                createFinding(ruleName = "TestSmellA"),
-                createFinding(ruleName = "TestSmellB"),
-                createFinding(ruleName = "TestSmellC")
+                createFinding(ruleName = "TestSmellA", severity = SeverityLevel.ERROR),
+                createFinding(ruleName = "TestSmellB", severity = SeverityLevel.WARNING),
+                createFinding(ruleName = "TestSmellC", severity = SeverityLevel.INFO),
+                createFinding(ruleName = "TestSmellD", severity = SeverityLevel.IGNORE)
```
BraisGabin, line 30: 🤔 does this have sense? Should we ever have an issue with ignore severity? I mean, to our users we provide the option to set severity as ignore. But, inside the core should we have that severity? Shouldn't we map that severity as `active = false` and don't even have it as a valid value in `SeverityLevel`?

schalkms: Good idea 💡

chao2zhang: It is a good question.  When breaking change is allowed, I meant to see if we should all migrate `active = false` to `severity = ignore`.

We can postpone that discussion later and I will remove SeverityLevel.IGNORE in this PR to not incur confusion.

schalkms: > We can postpone that discussion later and I will remove SeverityLevel.IGNORE in this PR to not incur confusion.

+1

BraisGabin: Do we need to pospone it? I mean, right now we can make that `Rule.active` is the same as `yml.active && yml.severity != "ignore"`. And we can remove `SeverityLevel.IGNORE` as a value of the enum.

chao2zhang: I think about determining `severity == "ignore"` when calculating active, but the more I think, the more edge cases we could introduce. There can be a combination of cases:
- `active` is false, and `severity` is not set either => We should not run the rule
- `active` is false, but `severity` is set => We should not run the rule, but should we raise an error?
- `active` is true, and `severity` is not set => We should run the rule, with default level = warning
- `active` is true, and `severity` is set => We should run the rule, unless the default level is ignore?

Now in the future for Detekt 2.0, I would prefer using one config key to configure the severity and enabled state. Having two causes more confusion in my opinion.

BraisGabin: I agree, we should remove `active` from the config file in 2.0.

Issue #3279, opened 2020-12-11
Activated rules don't do anything without an includes filter


schalkms:
This happens for the following rules
- LibraryCodeMustSpecifyReturnType
- LibraryEntitiesShouldNotBePublic

This behavior is confusing. So I suggest to update the rule doc and the config as follows.

```diff
LibraryCodeMustSpecifyReturnType:
-  active: true
+  active: false
+ includes: ['**/main/**']
LibraryEntitiesShouldNotBePublic:
  active: false
+ includes: ['**/main/**']
```

Context:
Please see #3215 where this scenario came to attention.

I would like to get a general consensus before submitting a PR. Thank you!

schalkms:
Please note that this rules only make sense for `main` but not for `test` code.

arturbosch:
Truly the documentation does not point out that `include` needs to be used for these rules.

Both rules work as intended imo. We can't say what library code is and what not. So the developer has to specifically opt-in by providing the paths which this rule will check.

I suggest updating the docs and provide an example configuration.

schalkms:
@arturbosch will do. That's a good idea. Anyways, both rules are great and should be activated by default IMHO.

chao2zhang:
> This happens for the following rules
> 
> * LibraryCodeMustSpecifyReturnType
> * LibraryEntitiesShouldNotBePublic
> 
> This behavior is confusing. So I suggest to update the rule doc and the config as follows.
> 
> ```diff
> LibraryCodeMustSpecifyReturnType:
> -  active: true
> +  active: false
> + includes: ['**/main/**']
> LibraryEntitiesShouldNotBePublic:
>   active: false
> + includes: ['**/main/**']
> ```
> 
> Context:
> Please see #3215 where this scenario came to attention.
> 
> I would like to get a general consensus before submitting a PR. Thank you!

`ForbiddenPublicDataClass` also falls into this group.

> @arturbosch will do. That's a good idea. Anyways, both rules are great and should be activated by default IMHO.

Other than library modules, we may have app modules or repos consuming detekt. Because we do not have data on how much of our clients are library modules, so I would not recommend enabling this by default.

schalkms:
> Other than library modules, we may have app modules or repos consuming detekt. Because we do not have data on how much of our clients are library modules, so I would not recommend enabling this by default.

If there's no includes filter configured the rule doesn't yield any violations.

BraisGabin:
[...]

PR #3312, opened 2020-12-20
Add KDoc for convoluted PathFilters.isIgnored

chao2zhang:
#3279 describes the scenario that rules are not doing anything if neither `includes` or `excludes` are specified.

When reading the code, it comes to me that `PathFilters.isIgnored()` is a bit convoluted and probably the implementation is not working as expected. Therefore I publish this PR to describe the current behavior. And in case we update the behavior later, we should be able to correct the doc and help the reviewers to understand the status quo.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/PathFilters.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/PathFilters.kt
@@ -10,6 +10,16 @@ class PathFilters internal constructor(
     private val excludes: Set<PathMatcher>?
 ) {
 
+    /**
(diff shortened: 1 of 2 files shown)


schalkms (approved):
Good one! Thank you.
Could we also add this phenomenal comment to the test description in `PathFiltersSpec.kt`? I think this will be beneficial for future readers of the code to have a test for each case.

Bonus (test) case: have the same path included as well as excluded

chao2zhang:
> Good one! Thank you.
> Could we also add this phenomenal comment to the test description in `PathFiltersSpec.kt`? I think this will be beneficial for future readers of the code to have a test for each case.
> 
> Bonus (test) case: have the same path included as well as excluded

I added the test case as another form of documentation

BraisGabin (approved):

Issue #853, opened 2018-04-13
Add config option to choose between absolute and relative paths when reporting


arturbosch:
## Expected Behavior
A finding should match into one console line.
For this, let the user choose if he want the findings to be reporting with relative or absolute paths.

## Current Behavior
When reporting the absolute path, findings are often broken into two lines and it looks awful.
This comes due to changes to the reporting api to have the paths clickable.

## Context
It is hard to read the findings now.

RiccardoM:
Another use case for this would be allowing custom output reports to print either absolute or relative (to the project root folder) path. 

For example, if I want to integrate Detekt with [Gitlab code quality system](https://docs.gitlab.com/ee/user/project/merge_requests/code_quality.html) I have to output the paths as relative to the project root folder. Currently I cannot do this becasue all the paths are printed as absolute paths.

arturbosch:
It could be done by a postprocessor if we provide the `basePath` from the tooling api to the `SetupContext` in api ^^.
I think we should:
1. drop processor, report configurations from the config
2. Have a configuration option to include and exclude extensions
Similiar to https://github.com/detekt/detekt/blob/5a0fe624e571d20e1619b1e42360c2b1f513a2ce/detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/spec/ExtensionsSpec.kt#L22-L25
3. Support `deactivatedByDefault` behavior for extensions
4. Implement this as a deactivated extension lying around waiting for users to be used.

chao2zhang:
We already have a draft #2492. But as mentioned in #3045, I will pick this up. Please update the status to in progress in Projects - Roadmap.

BraisGabin:
@chao2zhang we could close this now, right?

PR #3313, opened 2020-12-21
Fix reports not propagated to detekt task with type resolution

chao2zhang:
Configuring detekt gradle extension as below does not actually generate SARIF reports when running `detekt<SourceSet>` like `:detektMain`. The problem here is that we are merely not passing the `reports` in `DetektExtension` to `reports` of `Detekt` as a gradle task.

```
detekt {
    reports {
        sarif.enabled = true
    }
}
```

In addition to integration testing, I also manually verified through a sample project that the above is working.

### DIFF
--- a/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPluginTest.kt
+++ b/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPluginTest.kt
@@ -28,11 +28,33 @@ object DetektPluginTest : Spek({
 
             assertThat(project.tasks.getAt("check").dependencies()).contains("detekt")
         }
+
+        it("configures detekt task correctly") {
+            val project = ProjectBuilder.builder().build()
+
+            with(project.pluginManager) {
+                apply(DetektPlugin::class.java)
+                apply(LifecycleBasePlugin::class.java)
+                apply("org.jetbrains.kotlin.jvm")
+            }
+
+            project.configureExtension<DetektExtension> {
+                reports {
+                    it.sarif {
+                        enabled = true
+                    }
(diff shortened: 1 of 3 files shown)


detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPluginTest.kt
```diff
@@ -28,11 +28,33 @@ object DetektPluginTest : Spek({
 
             assertThat(project.tasks.getAt("check").dependencies()).contains("detekt")
         }
+
+        it("configures detekt task correctly") {
```
schalkms, line 32:
```suggestion
        it("configures detekt task with sarif report correctly") {
```

schalkms (approved):

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPluginTest.kt
chao2zhang: Sarif report is only one piece of configuration. I believe we will and should verify more configurations of task here

schalkms: alright

chao2zhang:
This PR is blocking me from testing the solution for #853. When possible, please merge this pull request.

schalkms:
@chao2zhang good to know. This PR will be merged asap if at least two maintainers review and approve it. Hence, I send out an review invitation.

BraisGabin (approved):
Nice catch! And it's great to increase our test base 🙂

Issue #2693, opened 2020-05-15
Index Out Of Bounds Exception without relatable error message


HartmannS:
## Expected Behavior
`./gradlew check` should not run into exception, it should produce meaningful error message

## Observed Behavior
```
Analyzing '/usr/app/src/main/kotlin/simpleTensorFlow/domain/layer/CrossEntropyLayer.kt' led to an exception.
The original exception message was: java.lang.IndexOutOfBoundsException: Wrong offset: 1750. Should be in range: [0, 1749]
Running detekt '1.8.0' on Java '13-ea+32' on OS 'Linux'.
If the exception message does not help, please feel free to create an issue on our GitHub page.
java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314)
java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:319)
java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1773)
java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
java.base/java.lang.Thread.run(Thread.java:830)
org.jetbrains.kotlin.com.intellij.openapi.editor.impl.LineSet.findLineIndex(LineSet.java:178)
org.jetbrains.kotlin.com.intellij.openapi.editor.impl.DocumentImpl.getLineNumber(DocumentImpl.java:991)
org.jetbrains.kotlin.diagnostics.PsiDiagnosticUtils.offsetToLineAndColumn(PsiDiagnosticUtils.java:73)
org.jetbrains.kotlin.diagnostics.DiagnosticUtils.getLineAndColumnInPsiFile(DiagnosticUtils.java:84)
io.gitlab.arturbosch.detekt.rules.LinesOfCodeKt.line(LinesOfCode.kt:47)
io.gitlab.arturbosch.detekt.rules.LinesOfCodeKt$linesOfCode$1.invoke(LinesOfCode.kt:43)
io.gitlab.arturbosch.detekt.rules.LinesOfCodeKt$linesOfCode$1.invoke(LinesOfCode.kt)
kotlin.sequences.TransformingSequence$iterator$1.next(Sequences.kt:172)
kotlin.sequences.DistinctIterator.computeNext(Sequences.kt:524)
[...]

PR #3317, opened 2020-12-24
Switch order of correctable rules and other rules

chao2zhang:
Interestingly, the previous code caused a crash in (#2693). The problem is that we run Ktlint/CommentSpacing before running LongMethod. Because CommentSpacing mutates the `ktfile.text`, but does not update
`Ktfile.viewerProvider.document`, resulting in an inconsistent state and eventually crash (https://github.com/detekt/detekt/issues/2693#issuecomment-750395108)

Note that this commit is merely a short-term fix. In the long term, we should be more careful in rule running order, while leveraging parallelism to optimize the runtime.

Test
====
I verified that the correct violations `CommentSpacing` are reported in the sample project (https://github.com/andrey-bolduzev/detekt_issue_2693) after the change. Before this change, it will crash with `java.lang.IndexOutOfBoundsException: Wrong offset: 834. Should be in range: [0, 833]`

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
@@ -124,8 +124,8 @@ internal class Analyzer(
             }
         }
 
-        executeRules(correctableRules)
         executeRules(otherRules)
+        executeRules(correctableRules)
 
         return result
     }


chao2zhang:
Fun fact: I almost spent my day debugging this. I initially thought this is a Kotlin compiler bug and started by refactoring the code to extract `PsiWhitespace`, then scratched my head on why `PsiWhitespace` is by default not provided in the parsed `KtFile`. Luckily I was able to debug through the stack and find the root cause.

BraisGabin:
We can't merge this because it will break the html report and the position of a lot of issues. If you first run the no correctable and then the correctable ones all the issues that you found before could have change its position.

We need to find other way to fix this.

If I understand it correctly this error happens because some ktlint rules uses a value that doesn't update after change the KtFile, right? Could we create a PR (or open an issue first) to them?

BraisGabin (changes requested):
We can't merge this because it will break the html report and the position of a lot of issues. If you first run the no correctable and then the correctable ones all the issues that you found before could have change its position.

We need to find other way to fix this.

If I understand it correctly this error happens because some ktlint rules uses a value that doesn't update after change the KtFile, right? Could we create a PR (or open an issue first) to them?

chao2zhang:
That's a good point. Let me summarize the dilemma:
- Run non-correctable rules before running correctable ones: The position reported in the non-correctable rules may be outdated when correctable ones were mutating the `KtFile` content
- Run correctable rules before running non-correctable ones: Ktlint rules mutates the ASTNode directly. Therefore there can be stale references that were not updated. In our case, `LinesOfCode` depends on `DiagnosticUtils.getLineAndColumnInPsiFile`, which internally uses the class `Document` to find line and column in the original value before correction. In fact `DiagnosticUtils` is eventually invoked by many of our rules.

There are several ways to fix them:
1. Keep with option 1 - If the `KtFile` content is mutated, we announce that locations might be offsetted.
2. Keep with option 1 - Find a smart way to handle `ASTNode` modification.
3. Keep with option 2 - Find a smart way to update all the stale references: Although this effort might be difficult to be exhaustive, and we need to keep maintaining them when we upgrade kotlin versions.
4. Keep with option 2 - Find a smart way to refactor off methods using the stale references.

There is a collection of tangential issues related to this:
a. KtlintMultiRule is regarded as either all auto-correctible or non-autocorrectible. But in fact, it mixes auto-correctible with non-autocorrectible rules. Thus we need to split KtlintMultiRule into two sets, at least.
b. KtLint rules have certain modifiers to impose orders as `MaxLineLengthRule` runs as the `Last` rule. Some long lines may not actually exceed the line length limit after mutation from other rules. I am pretty confident that this issue exists for Detekt as well.
c. Parallelism is my perception of the intention of creating `MultiRule`, but I believe we could do much better here to figure out the strategy across the board. The argument here would be a `Rule` can report `Multiple` issues.

Overall autocorrection is a difficult problem. I am wondering if Ktlint owners can shed some lights (The ordering marker interface was added in https://github.com/pinterest/ktlint/issues/101 and modified in https://github.com/pinterest/ktlint/issues/158)

BraisGabin:
Other option is to run the correctable ones first and, if there is any change, regenerate the KtFile. This one would be slower.

chao2zhang:
I may not have the privilege, but is it possible to convert this PR to a discussion? If not, let's create a new discussion and close this PR

BraisGabin:
> I may not have the privilege, but is it possible to convert this PR to a discussion? If not, let's create a new discussion and close this PR

It seems that we can't convert PRs in conversations. But I think that this is better suited in an issue. This is a bug.

PR #3319, opened 2020-12-25
Support relative output paths

chao2zhang:
This addresses #2492 and a part of #3045. 

Out of scope items
============
- Actually integrate with SARIF in Github actions
- Change the default value for the base path to be the project root
- Move SARIF support from experimental to formal
- Update docs about enabling SARIF support + Github actions
- Update CodeSmell messages where the absolute path is used

Implementation
===========
The implementation was done by putting the relative path and base path as user data of `KtFile`.
`OutputReport`s read them through the paths stored in `Location`.

Test
===========
In addition to adding tests in Spek, I also perform manual tests with a local repository.


schalkms (commented):
I quickly skimmed through the code. The PR looks good to me.
Thanks for picking this up. Relative paths are a highly requested detekt feature.
I'm contemplating about adding more end-to-end tests in this area of detekt, since you also mentioned that you've done manual tests (thanks btw). For regression purposes it definitely makes sense. Could some of those manual tests be automated?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/PathFilters.kt
```diff
@@ -28,6 +33,8 @@ class PathFilters internal constructor(
         return isIncluded()?.not() ?: isExcluded() ?: true
     }
 
+    fun isIgnored(ktFile: KtFile): Boolean = isIgnored(ktFile.absolutePath())
```
schalkms, line 36: Maybe we could reference the nice documentation of the called function `fun isIgnored(path: Path): Boolean` for an easier overview.

detekt-report-txt/src/test/kotlin/io/github/detekt/report/txt/TxtOutputReportSpec.kt
```diff
@@ -38,5 +39,20 @@ class TxtOutputReportSpec : Spek({
             """.trimIndent()
             assertThat(report.render(detektion)).isEqualTo(renderedText)
         }
+
+        it("render from relative paths") {
```
schalkms, line 43:
```suggestion
        it("renders from relative paths") {
```

detekt-report-xml/src/test/kotlin/io/github/detekt/report/xml/XmlOutputFormatSpec.kt
```diff
@@ -35,15 +38,25 @@ class XmlOutputFormatSpec : Spek({
         it("renderEmpty") {
```
schalkms, line 38:
```suggestion
        it("renders empty XML") {
```

detekt-report-xml/src/test/kotlin/io/github/detekt/report/xml/XmlOutputFormatSpec.kt
```diff
@@ -35,15 +38,25 @@ class XmlOutputFormatSpec : Spek({
...
             val result = outputFormat.render(TestDetektion())
 
-            assertThat(result).isEqualTo("<?xml version=\"1.0\" encoding=\"utf-8\"?>\n<checkstyle version=\"4.3\">\n</checkstyle>")
+            assertThat(result).isEqualTo("""
+                <?xml version="1.0" encoding="utf-8"?>
+                <checkstyle version="4.3">
+                </checkstyle>""".trimIndent())
         }
 
         it("renderOneForSingleFile") {
```
schalkms, line 47:
```suggestion
        it("renders one reported issue in single XML file") {
```

detekt-report-xml/src/test/kotlin/io/github/detekt/report/xml/XmlOutputFormatSpec.kt
schalkms: Those descriptions are from the past when the tests were written in JUnit. I think that a refactoring is overdue here.

detekt-report-xml/src/test/kotlin/io/github/detekt/report/xml/XmlOutputFormatSpec.kt
```diff
@@ -35,15 +38,25 @@ class XmlOutputFormatSpec : Spek({
...
+                <?xml version="1.0" encoding="utf-8"?>
+                <checkstyle version="4.3">
+                <file name="src/main/com/sample/Sample1.kt">
+                $TAB<error line="11" column="1" severity="warning" message="" source="detekt.id_a" />
+                </file>
+                </checkstyle>
+            """.trimIndent())
         }
 
         it("renderTwoForSingleFile") {
```
schalkms, line 62:
```suggestion
        it("renders two reported issues in single XML file") {
```

detekt-report-xml/src/test/kotlin/io/github/detekt/report/xml/XmlOutputFormatSpec.kt
```diff
@@ -52,7 +65,14 @@ class XmlOutputFormatSpec : Spek({
...
+                <?xml version="1.0" encoding="utf-8"?>
+                <checkstyle version="4.3">
+                <file name="src/main/com/sample/Sample1.kt">
+                $TAB<error line="11" column="1" severity="warning" message="" source="detekt.id_a" />
+                $TAB<error line="11" column="1" severity="warning" message="" source="detekt.id_b" />
+                </file>
+                </checkstyle>""".trimIndent())
         }
 
         it("renderOneForMultipleFiles") {
```
schalkms, line 78: here too

detekt-report-xml/src/test/kotlin/io/github/detekt/report/xml/XmlOutputFormatSpec.kt
```diff
@@ -61,7 +81,42 @@ class XmlOutputFormatSpec : Spek({
...
+                <file name="src/main/com/sample/Sample1.kt">
+                $TAB<error line="11" column="1" severity="warning" message="" source="detekt.id_a" />
+                </file>
+                <file name="src/main/com/sample/Sample2.kt">
+                $TAB<error line="22" column="2" severity="warning" message="" source="detekt.id_a" />
+                </file>
+                </checkstyle>""".trimIndent())
+        }
+
+        it("renderRelativePaths") {
```
schalkms, line 96: here too

detekt-report-xml/src/test/kotlin/io/github/detekt/report/xml/XmlOutputFormatSpec.kt
```diff
@@ -61,7 +81,42 @@ class XmlOutputFormatSpec : Spek({
...
+                <file name="Sample1.kt">
+                $TAB<error line="1" column="1" severity="warning" message="TestMessage" source="detekt.id_a" />
+                </file>
+                <file name="Sample2.kt">
+                $TAB<error line="1" column="1" severity="warning" message="TestMessage" source="detekt.id_b" />
+                </file>
+                </checkstyle>""".trimIndent())
         }
 
         it("renderTwoForMultipleFiles") {
```
schalkms, line 122: here too

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -85,6 +85,13 @@ class CliArgs {
     )
     private var reports: List<String>? = null
 
+    @Parameter(
+        names = ["--working-dir", "-wd"],
+        description = "Specify a working directory as the base path for file paths in output and reports.",
+        converter = PathConverter::class
+    )
+    var workingDir: Path? = null
```
schalkms, line 93: I guess some user documentation would be helpful for users in order to convert their CLI or Gradle config to use the new feature.

chao2zhang:
> I'm contemplating about adding more integration tests in this area of detekt

This PR definitely added some integration tests in `DetektTaskDslTest`. 

However, if you indicate end-to-end tests, I only see `DetektTaskIntegrationTest` in the codebase. And there is no integration test written at the moment.

In my mind, such end-to-end tests should be a separate module or repo that applies detekt gradle plugin, runs against some kotlin code with violations, and validates the error can be reported. This does not exist today but I definitely can help sparkle the discussion

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/PathFilters.kt
chao2zhang: Good call.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
chao2zhang: As the PR description, I would update docs about enabling SARIF support + Github actions in a later PR altogether, after we have successfully set them up. Please let me know otherwise.

schalkms:
> However, if you indicate end-to-end tests, I only see DetektTaskIntegrationTest in the codebase. And there is no integration test written at the moment.
> 
> In my mind, such end-to-end tests should be a separate module or repo that applies detekt gradle plugin, runs against some kotlin code with violations, and validates the error can be reported. This does not exist today but I definitely can help sparkle the discussion

Yes, I definitely meant end-to-end tests. Thanks for the hint. End-to-end tests are a separate topic and as you mentioned outside of the scope for this PR.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
schalkms: I should have been more clear. I meant the documentation about the flags and Gradle input properties, not the SARIF support. Those flags can be seen [here](https://detekt.github.io/detekt/cli.html).

chao2zhang: Oops I thought this is auto-generated. Let me add it there.

schalkms: Yes, that's the cli part. Thanks! I think the Gradle part of the documentation is also helpful. This belongs to the main use case for this feature.

https://detekt.github.io/detekt/gradle.html#options-for-detekt-configuration-closure

schalkms (approved):

BraisGabin (commented):
🤔 I didn't have time to tests it but if you add the `workingDir` the console report keeps reporting the absolute path. Right?

And, should we wait to merge #3320 before merging this? To ensure that we don't break compatibity.

chao2zhang:
> 🤔 I didn't have time to tests it but if you add the `workingDir` the console report keeps reporting the absolute path. Right?
> 
> And, should we wait to merge #3320 before merging this? To ensure that we don't break compatibity.

I think #3320 is ready to go. I will handle the merge once that one is in.

docs/pages/gettingstarted/gradle.md
```diff
@@ -197,6 +197,7 @@ detekt {
     ignoredBuildTypes = ["release"]                       // Android: Don't create tasks for the specified build types (e.g. "release")
     ignoredFlavors = ["production"]                       // Android: Don't create tasks for the specified build flavor (e.g. "production")
     ignoredVariants = ["productionRelease"]               // Android: Don't create tasks for the specified build variants (e.g. "productionRelease")
+    workingDir = layout.projectDirectory                  // Specify a working directory as the base path for file paths in output and reports. If not, all file paths reported will be absolute file path.
```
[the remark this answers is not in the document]
chao2zhang: I have considered the possibility of whether we want to have the flexibility to specify any path or switch between project/absolute.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/AstPrinter.kt
```diff
@@ -19,7 +19,7 @@ class AstPrinter(
             "Input path $input must be a kotlin file and not a directory."
         }
 
-        val ktFile = KtCompiler().compile(input, input)
+        val ktFile = KtCompiler().compile(null, input)
```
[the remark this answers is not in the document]
chao2zhang: Is AstPrinter considered as output? It sounds like yes, please confirm :)

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -63,6 +63,7 @@ class DetektPlugin : Plugin<Project> {
             it.reportsDir.set(project.provider { extension.customReportsDir })
             it.reports = extension.reports
             it.ignoreFailuresProp.set(project.provider { extension.ignoreFailures })
+            it.workingDir.set(extension.workingDir)
```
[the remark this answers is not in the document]
chao2zhang: I already added the statement in `SharedTasks.registerCreateBaselineTask`.

detekt-report-xml/src/test/kotlin/io/github/detekt/report/xml/XmlOutputFormatSpec.kt
```diff
@@ -79,7 +134,18 @@ class XmlOutputFormatSpec : Spek({
...
+                <checkstyle version="4.3">
+                <file name="src/main/com/sample/Sample1.kt">
+                $TAB<error line="11" column="1" severity="warning" message="" source="detekt.id_a" />
+                $TAB<error line="11" column="1" severity="warning" message="" source="detekt.id_b" />
+                </file>
+                <file name="src/main/com/sample/Sample2.kt">
+                $TAB<error line="22" column="2" severity="warning" message="" source="detekt.id_a" />
+                $TAB<error line="22" column="2" severity="warning" message="" source="detekt.id_b" />
+                </file>
+                </checkstyle>""".trimIndent())
```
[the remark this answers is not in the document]
chao2zhang: Yes let me just extract refactoring to a separate PR. I did the same in #3310

detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/KtFiles.kt
```diff
@@ -21,8 +21,30 @@ fun PsiFile.fileNameWithoutSuffix(): String {
...
-fun PsiFile.relativePath(): Path {
-    val value = getUserData(RELATIVE_PATH)
-    checkNotNull(value) { "KtFile '$name' expected to have an relative path." }
-    return Paths.get(value)
+/**
+ * A file path that represents both relative path or absolute path.
+ * If representing an absolute path, [basePath] is null and [path] represents the full path.
+ * If representing a relative path, [basePath] is the base path while [path] represents the path relative to [basePath].
+ */
+data class FilePath constructor(val path: String, val basePath: String? = null) {
```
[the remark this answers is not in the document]
chao2zhang: Good suggestion. I think the code could be more readable with extra properties.

detekt-report-xml/src/test/kotlin/io/github/detekt/report/xml/XmlOutputFormatSpec.kt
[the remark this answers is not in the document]
chao2zhang: Created #3325. I will later rebase both #3319 and #3310.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/AstPrinter.kt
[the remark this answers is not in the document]
BraisGabin: Yes, this command prints a file's psi tree.

docs/pages/gettingstarted/gradle.md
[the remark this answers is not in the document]
BraisGabin: I think that all our reports should use relative paths (except console) because you can share them with other people and it have no sense to have the absolute path there. Does this have sense? The problem that I see here is... could this be considered a breaking change? I think it's not. And, if the user really want the absolute path can set this value to `null`.

I think that we should set as default value the root of each project.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/AstPrinter.kt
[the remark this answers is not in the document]
chao2zhang: On second thought, this seems to conflict with https://github.com/detekt/detekt/pull/3319#discussion_r549045142. AstPrinter also outputs to console.

detekt-parser/src/test/kotlin/io/github/detekt/parser/KtCompilerTest.kt
```diff
@@ -23,7 +23,7 @@ class KtCompilerTest : Spek({
 
             assertThat(ktFile.getUserData(LINE_SEPARATOR)).isEqualTo(System.lineSeparator())
             assertThat(ktFile.getUserData(RELATIVE_PATH))
-                .isEqualTo(path.fileName.resolve("Default.kt").toString())
+                .isEqualTo("Default.kt")
```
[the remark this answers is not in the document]
chao2zhang: Added verification for BASE_PATH

docs/pages/gettingstarted/gradle.md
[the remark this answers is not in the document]
chao2zhang: Changed the default to be the project root path for XML, HTML, and SARIF.

chao2zhang:
Apologize for the numerous force-push. I was debugging a CI failure that is not reproducible on local.

chao2zhang:
There is a new problem that I encountered in this PR.

A) In https://github.com/detekt/detekt/pull/3319#discussion_r549045142, we are suggested to set the *default* relative path base to the project source dir
B) The implementation goes that the detekt gradle plugin will generate the extra args to invoke `DetektCli`, which means `Main.kt` need to be able to recognize the new parameter `--report-base-path`.
C) However, the tests using `GradleRunner/gradle-testkit` are actually not dependent on the local versions of detekt, but actually performs resolution through all the repositories.
D) This explains why I was able to pass the test locally because I ran `./gradlew publishToMavenLocal` before `./gradlew :detekt-gradle-plugin:test`. My maven local contains my change while when we run the pre-merge check on CI, publishToMavenLocal was not run.

So to unblock this PR, we can fix any one of the steps listed here:
A) By default,  we still remain the old absolute path behavior. The new gradle extension for `reportBasePath` remains empty if there is no provided value.
B) We create a PR to add `--report-base-path` parameter to `CliArgs`, release the next Detekt version to maven/jcenter, then merge in this PR.
C) I have been juggling with [testkit with classpath injection] (https://docs.gradle.org/current/userguide/test_kit.html#sub:test-kit-classpath-injection), but still cannot get it to work with the current setup. The most hopeful try is simply adding `implementation(project(":detekt-cli"))` to dependencies.
We should at least separate the `GradleRunner/gradle-testkit` tests into at least a separate `Test` task, more to be discussed in #3324
D) Run `publishToMavenLocal` in pre-merge-checks so that `GradleRunner/gradle-testkit` tests will run against the jar from maven local.

Among all the options, 
A) is a reasonable choice.
C) #3324 is the correct approach but will block this PR for quite some time.
D) feels a short-term patch if #3324 is not resolved.
B) is our last straw if none of A/C/D are agreed and implemented.

Apologize for the long response. I am seeking thoughts from all the maintainers, especially @BraisGabin since A) is per his request.

BraisGabin:
Thanks for the reasearch and the detailed report!

Agree with the option A). If the default value adds to much complexity we can merge this witout it and open an issue so we can find the best option to add it. What do you think?

We can add it in the next release if that make the things easier. We could even use a `RC` version.

chao2zhang:
> If the default value adds to much complexity we can merge this witout it and open an issue so we can find the best option to add it. What do you think?

It sounds like we are on the same page. What I probably need to do next is to prototype #3324. Since that would be impacting future changes and testing for detekt-gradle-plugin.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -85,6 +85,14 @@ class CliArgs {
     )
     private var reports: List<String>? = null
 
+    @Parameter(
+        names = ["--report-base-path", "-rbp"],
+        description = "Specifies a directory as the base path for all file paths in the formatted reports. " +
+            "File paths in console output and txt report are not affected and remain as absolute paths.",
+        converter = PathConverter::class
+    )
+    var reportBasePath: Path? = null
```
BraisGabin, line 94: Event that we just use the base-path for the reports right now, should we bind this flag to the reports? I'm thinking in `--base-path` instead of `--report-base-path`.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskDslTest.kt
```diff
@@ -236,7 +236,7 @@ internal object DetektTaskDslTest : Spek({
                     }
 
                     it("no report param is set") {
-                        assertThat(result.output).doesNotContain("--report")
+                        assertThat(result.output).doesNotContain("--report ")
```
BraisGabin, line 239: And we will not need this change if we do the other.

BraisGabin (approved):
This is good to go for me. My last concern (sorry the PR is big and each time I review it I realiced different things) is the cli flag name: `--report-base-path`. Right now we use the relative paths just for reporting but this feature is not just about that so, maybe, we could rename it to something more generic as `--base-path`. What do you think? Any way, this is not a blocking topic.

chao2zhang:
> This is good to go for me. My last concern (sorry the PR is big and each time I review it I realiced different things) is the cli flag name: `--report-base-path`. Right now we use the relative paths just for reporting but this feature is not just about that so, maybe, we could rename it to something more generic as `--base-path`. What do you think? Any way, this is not a blocking topic.

I don't know other use cases of base path. The input paths can be both absolute and relative based on `user.dir`, and internally we are always using absolute paths. If we want to expand this idea, we could choose to add a different option specifically.

detekt-api/api/detekt-api.api
```diff
@@ -315,17 +315,22 @@ public final class io/gitlab/arturbosch/detekt/api/LazyRegex : kotlin/properties
...
 	public final fun component2 ()Lio/gitlab/arturbosch/detekt/api/TextLocation;
 	public final fun component3 ()Ljava/lang/String;
-	public final fun copy (Lio/gitlab/arturbosch/detekt/api/SourceLocation;Lio/gitlab/arturbosch/detekt/api/TextLocation;Ljava/lang/String;)Lio/gitlab/arturbosch/detekt/api/Location;
-	public static synthetic fun copy$default (Lio/gitlab/arturbosch/detekt/api/Location;Lio/gitlab/arturbosch/detekt/api/SourceLocation;Lio/gitlab/arturbosch/detekt/api/TextLocation;Ljava/lang/String;ILjava/lang/Object;)Lio/gitlab/arturbosch/detekt/api/Location;
+	public final fun component4 ()Lio/github/detekt/psi/FilePath;
+	public final fun copy (Lio/gitlab/arturbosch/detekt/api/SourceLocation;Lio/gitlab/arturbosch/detekt/api/TextLocation;Ljava/lang/String;Lio/github/detekt/psi/FilePath;)Lio/gitlab/arturbosch/detekt/api/Location;
+	public static synthetic fun copy$default (Lio/gitlab/arturbosch/detekt/api/Location;Lio/gitlab/arturbosch/detekt/api/SourceLocation;Lio/gitlab/arturbosch/detekt/api/TextLocation;Ljava/lang/String;Lio/github/detekt/psi/FilePath;ILjava/lang/Object;)Lio/gitlab/arturbosch/detekt/api/Location;
```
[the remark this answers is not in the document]
chao2zhang: Indeed. The learning from [JakeWharton's Blog](https://jakewharton.com/public-api-challenges-in-kotlin/#copy-functions) is that the data class' APIs always change.

BraisGabin:
> > This is good to go for me. My last concern (sorry the PR is big and each time I review it I realiced different things) is the cli flag name: `--report-base-path`. Right now we use the relative paths just for reporting but this feature is not just about that so, maybe, we could rename it to something more generic as `--base-path`. What do you think? Any way, this is not a blocking topic.
> 
> I don't know other use cases of base path. The input paths can be both absolute and relative based on `user.dir`, and internally we are always using absolute paths. If we want to expand this idea, we could choose to add a different option specifically.

Posted in our slack channel: https://kotlinlang.slack.com/archives/C88E12QH4/p1609854726084000?thread_ts=1609854726.084000&cid=C88E12QH4

A user asking exactly for this feature and it's not related with the reports. So, probably, we should rename the flag.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
chao2zhang: Fixed.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskDslTest.kt
chao2zhang: Fixed.

sschuberth:
> A user asking exactly for this feature and it's not related with the reports. So, probably, we should rename the flag.

That user was me 😀 To clarify, I need to write a custom rule that checks for a project's naming conventions based on the relative path of source files, where the base is the Gradle root project directory. So my assumption would be that after this PR is merged, a call to the `PsiFile.relativePath()` extension function from the `io.github.detekt.psi` package would return exactly that relative path when using the detekt plugin for Gradle. Right now, a call to `PsiFile.relativePath()` just returns the file name (and extension) without any path.

Is my assumption / expectation about the goal of this PR correct?

BraisGabin (approved):

BraisGabin:
Yes, they are. You will need one extra line of configuration in your gradle because of this https://github.com/detekt/detekt/pull/3319#issuecomment-753389072 that we decided that is out of the scope of this PR (but it will be implemented later). But you would be able to get the relative path using a extension function over the KtFile.

PR #3320, opened 2020-12-25
Apply binary compatibility plugin to Detekt

chao2zhang:
This will help catch API changes like discussed in https://github.com/detekt/detekt/pull/3310#discussion_r546868711
There is also a temporary workaround to exclude rootProject from compability check. See https://github.com/Kotlin/binary-compatibility-validator/issues/32

When running `./gradlew build`, if there is any change in :detekt-api, the task will fail with the following message
```
~/detekt(api*) » ./gradlew build  
> Task :detekt-api:apiCheck FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':detekt-api:apiCheck'.
> API check failed for project detekt-api.
  --- /Users/cazhang/detekt/detekt-api/api/detekt-api.api
  +++ /Users/cazhang/detekt/detekt-api/build/api/detekt-api.api
  @@ -315,17 +315,20 @@
   
   public final class io/gitlab/arturbosch/detekt/api/Location : io/gitlab/arturbosch/detekt/api/Compactable {
        public static final field Companion Lio/gitlab/arturbosch/detekt/api/Location$Companion;
  -     public fun <init> (Lio/gitlab/arturbosch/detekt/api/SourceLocation;Lio/gitlab/arturbosch/detekt/api/TextLocation;Ljava/lang/String;)V
  +     public fun <init> (Lio/gitlab/arturbosch/detekt/api/SourceLocation;Lio/gitlab/arturbosch/detekt/api/TextLocation;Ljava/lang/String;Ljava/io/File;)V
  +     public synthetic fun <init> (Lio/gitlab/arturbosch/detekt/api/SourceLocation;Lio/gitlab/arturbosch/detekt/api/TextLocation;Ljava/lang/String;Ljava/io/File;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
        public fun <init> (Lio/gitlab/arturbosch/detekt/api/SourceLocation;Lio/gitlab/arturbosch/detekt/api/TextLocation;Ljava/lang/String;Ljava/lang/String;)V
        public fun compact ()Ljava/lang/String;
        public fun compactWithSignature ()Ljava/lang/String;
        public final fun component1 ()Lio/gitlab/arturbosch/detekt/api/SourceLocation;
        public final fun component2 ()Lio/gitlab/arturbosch/detekt/api/TextLocation;
        public final fun component3 ()Ljava/lang/String;
[...]

PR #3322, opened 2020-12-26
Simplify our buildSrc

BraisGabin:
The idea is to simplify how we configure our own gradle.

Right now we force that all our modules have a specific configuration and with this new approuch we provide a plugin that configures the module as we want and is the module that opts-in this configuration.

An advantage is that we don't need to add exceptions for `:detekt-bom`. It just don't use this module.

The idea is that this is just the first step. Once we do this I'll like to use the plugin https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin to improve how we manage our dependencies because we add a lot by default and we probably don't need that much. And, probably, we could remove all the code in `commons.gradle.kts` at some point.

This will help with the integration with the plugin `binary-compatibility-validator` too. This last one was the trigger to start this refactor.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,6 +10,10 @@ plugins {
     id("binary-compatibility-validator")
 }
 
+repositories {
+    jcenter()
+}
+
 buildScan {
     termsOfServiceUrl = "https://gradle.com/terms-of-service"
     termsOfServiceAgree = "yes"
--- a/buildSrc/src/main/kotlin/module.gradle.kts
+++ b/buildSrc/src/main/kotlin/module.gradle.kts
@@ -1,0 +1,103 @@
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+import org.gradle.api.tasks.testing.logging.TestLogEvent
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    kotlin("jvm")
+    `maven-publish`
+    jacoco
+}
+
+repositories {
+    jcenter()
+}
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+// bundle detekt's version for all jars to use it at runtime
+tasks.withType<Jar>().configureEach {
+    manifest {
+        attributes(mapOf("DetektVersion" to Versions.DETEKT))
+    }
+}
+
+jacoco.toolVersion = Versions.JACOCO
+
+tasks.withType<Test>().configureEach {
+    useJUnitPlatform()
+    systemProperty("spek2.jvm.cg.scan.concurrency", 1) // use one thread for classpath scanning
+    systemProperty("spek2.execution.test.timeout", 0) // disable test timeout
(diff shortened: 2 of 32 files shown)


buildSrc/src/main/kotlin/module.gradle.kts
```diff
@@ -0,0 +1,104 @@
...
+    archives(sourcesJar)
+    archives(javadocJar)
+}
+
+publishing {
+    publications.named<MavenPublication>(DETEKT_PUBLICATION) {
+        from(components["java"])
+        artifact(sourcesJar.get())
+        artifact(javadocJar.get())
+        if (project.name == "detekt-cli") {
```
chao2zhang, line 100: Could we refactor this condition to `detekt-cli`?

buildSrc/src/main/kotlin/packaging.gradle.kts
```diff
@@ -95,32 +82,3 @@ subprojects {
         logger.info("Signing Disabled as the PGP key was not found")
     }
 }
-
-configure(subprojects.filter { it.name != "detekt-bom" }) {
```
chao2zhang, line 99: This looks great!

buildSrc/src/main/kotlin/module.gradle.kts
BraisGabin: Yes, I agree that this should be removed too. But I'd prefer to do it in other PR.

schalkms (approved):
Great stuff 👏

buildSrc/src/main/kotlin/module.gradle.kts
BraisGabin: Maybe we can remove this if directly: #3329

chao2zhang:
What's the recommended merging order between this PR and https://github.com/detekt/detekt/pull/3320? I am open to merge in this PR first then see if the hack in  #3320 would be fixed

BraisGabin:
I think that this is a big change and we should wait for other maintainer to approve it. I know that @arturbosch worked a lot refactoring our gradle scripts (they were a real mess all in one file) so I think that we should wait a bit to merge this.

chao2zhang:
> I think that this is a big change and we should wait for other maintainer to approve it. I know that @arturbosch worked a lot refactoring our gradle scripts (they were a real mess all in one file) so I think that we should wait a bit to merge this.

Is it a good time to merge #3320 consequently? It creates merge conflicts for a few PRs already.

build.gradle.kts
```diff
@@ -4,11 +4,15 @@ plugins {
     releasing
     detekt
     id("org.jetbrains.dokka") apply false
-    id("com.github.johnrengelman.shadow") apply false
+    id("com.github.johnrengelman.shadow") version "5.2.0" apply false
```
arturbosch, line 7: This introduces a new place to remember where a version needs to be changed.
Is this change necessary?

arturbosch (approved):
Looks good!

build.gradle.kts
BraisGabin: I don't remember why I did that. Probably because we don't need the shadow pugin in buildSrc now. But I moved it to the buildSrc again 👍

PR #3325, opened 2020-12-26
Refactor XmlOutputFormatSpec

chao2zhang:
Test content is crowded into a single line, making it difficult to read and update.


BraisGabin (approved):
Thanks! There was not need to do this in this case. It was just a suggestion for future PRs. But I really appreciate it :)

Issue #3327, opened 2020-12-27
Refactor detekt-gradle-plugin off com.android.tools.build:gradle


chao2zhang:
Android Gradle plugin enters a new phase of aligning versions with Gradle [Official announcement](https://developer.android.com/studio/releases/gradle-plugin#versioning-update), this does have an impact on plugin authors. We should not depend on com.android.tools.build:gradle but com.android.tools.build:gradle-api. As Google plans to deprecate com.android.tools.build:gradle in 7.0 and remove it completely in 8.0 (This is according to the conversation between my colleague and the technical lead of Android Developer Tools, so there could be falsehood).

What I am proposing to do for detekt on Android projects:
- Experiment with gradle-api:4.1.1. As I was told that gradle-api:4.1/4.2 is evolving and but gradle-api:4.1 should work for the basic use case.
- Create one Detekt task per SourceSet, instead of per variant + test variant. This should not impact the behavior in terms of the generated gradle tasks but should significantly reduce the code complexity in our project.

I am open to any feedback or correction in the information I received.

BraisGabin:
> * Experiment with gradle-api:4.1.1. As I was told that gradle-api:4.1/4.2 is evolving and but gradle-api:4.1 should work for the basic use case.

I couldn't find anything about this `gradle-api` dependency. But for sure that we can test it if that allow us to be ready to any future change. But we should wait until we get some official documentation. Then we can start merging our "research".

> * Create one Detekt task per SourceSet, instead of per variant + test variant. This should not impact the behavior in terms of the generated gradle tasks but should significantly reduce the code complexity in our project.

I agree with you that have two tasks own per production code and other for test code is not handy. And this is not a problem just with out android compatibility. We have the same problem with pure kotlin projects too.

I think that we should rethink all our gradle tasks. My current idea is to have 4 principal tasks for 2.0:
- `detektCheck` - runs detekt with type solving. It never reformat code. It will be like running `detektMain` and `detektTest` but with a single report. For android modules it will run detekt for each build variant. For android we should create a task for each build variant too (`detektCheckDebug`, `detektCheckFlavourDebug`...)
- `detektFix` - runs the rules that are correctable. It never fails nor creates any report. It just fixes all the issues that it can. This could even help with #3250
- `detektBaseline` - generates the baseline
- `detektConfiguration` - generates the configuration

I think that gradle should not support checks without type solving for 2.0.

chao2zhang:
[...]

PR #3328, opened 2020-12-27
Update Android Gradle Plugin to 4.1.1

chao2zhang:
This is a prerequisite for https://github.com/detekt/detekt/issues/3327.

### DIFF
--- a/detekt-gradle-plugin/build.gradle.kts
+++ b/detekt-gradle-plugin/build.gradle.kts
@@ -8,7 +8,7 @@ repositories {
 }
 
 dependencies {
-    val androidGradlePlugin = "com.android.tools.build:gradle:4.0.1"
+    val androidGradlePlugin = "com.android.tools.build:gradle:4.1.1"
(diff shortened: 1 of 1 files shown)


BraisGabin:
This is always a question that I ask myself: should we use the lastest version of AGP? Or should we use the lowest version that we support? And, if so, which versions should we support?

schalkms (approved):

BraisGabin (approved):

chao2zhang:
> This is always a question that I ask myself: should we use the lastest version of AGP? Or should we use the lowest version that we support? And, if so, which versions should we support?

AGP eventually will be coupled with Gradle, so the policy should be applicable to both.
Eventually, I regard `:detekt-gradle-plugin` relatively independent from other artifacts, since it is a plugin project. 
- Plugin project does not leak the dependencies to consuming clients.
- People don't need to rely on detekt-gradle-plugin in order to latest version of detekt.

PR #3330, opened 2020-12-27
The output of updateVersion should not depend on the OS that executes it

BraisGabin:


schalkms (commented):
May I ask what’s the reason for the change?

BraisGabin:
There is not a real reason here. I was working on #3322 and I saw this. It's strange that the file that we crate with the version number depend on the OS of the person that runs it. Right?

I mean, if Artur moves from Unix to Windows or vice versa the file should not change it's end of line.

schalkms (approved):

PR #3332, opened 2020-12-27
Fix failing website deployment on master

### LINKED ISSUE #3321: Website deployment fails on master branch
The `Deploy Website / build-website` task fails with the following error since commit c6a971f.

```
Gemfile.lock It is likely that you need to grant write permissions for that path.
```

schalkms:
The reason was the following:
"Gemfile.lock It is likely that you need to grant write permissions for that path."
Consequently, write permission is given for the Gemfile.lock file.

Fixes #3321

### DIFF
--- a/.github/workflows/deploy-website.yaml
+++ b/.github/workflows/deploy-website.yaml
@@ -82,6 +82,8 @@ jobs:
         mkdir -p vendor/bundle/
         chmod -R 777 _site/
         chmod -R 777 vendor/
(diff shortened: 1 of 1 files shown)


BraisGabin (approved):
This is strange, why did it work before 🤔?

schalkms:
Quite frankly, I don’t have a valid explanation here. However, I didn’t dig into it further being happy that it works™️ again.

PR #3333, opened 2020-12-28
Refactor tests in detekt-gradle-plugin

chao2zhang:
1. Moved data classes, DslGradleRunner, and DslTestBuilder under `testkit`
2. Updated because reason in build.gradle.kts
3. Inlined and updated the formatting version under test.


detekt-gradle-plugin/build.gradle.kts
```diff
@@ -18,7 +18,7 @@ dependencies {
 
     constraints {
         implementation("org.jetbrains.kotlin:kotlin-reflect:1.4.0") {
-            because("com.android.tools.build:gradle:4.0.1 depends on Kotlin 1.3.72 and we should not mix 1.3 and 1.4")
+            because("$androidGradlePlugin depends on Kotlin 1.3.72 and we should not mix 1.3 and 1.4")
```
BraisGabin, line 21: 🤔 how can we be sure that we remove this constraint as soon as we update to `com.android.tools.build:gradle:4.2.0`. Or at least have a hint that we should remove it.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskDslTest.kt
```diff
@@ -637,3 +639,5 @@ internal object DetektTaskDslTest : Spek({
         }
     }
 })
+
+const val VERSION_UNDER_TEST: String = "1.15.0"
```
BraisGabin, line 643: Can this be private?

BraisGabin (approved):
Nice!

It will help if, the next time, you move files in one commit and refactor the file in another. This way we can review it commit by commit and see the actual changes in that file. Otherwise git says "this file was removed and this other was added".

Don't change it now! There's no need.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektTaskDslTest.kt
chao2zhang: Good catch

detekt-gradle-plugin/build.gradle.kts
chao2zhang: https://dl.google.com/android/maven2/com/android/tools/build/gradle/4.2.0-beta02/gradle-4.2.0-beta02.pom. I can add a message stating that this line should be removed once we are on AGP 4.2

BraisGabin (approved):

schalkms (approved):

PR #3334, opened 2020-12-29
[RFC] Register Detekt tasks based on source set for Android

chao2zhang:
Changes:
- This addressed #3327 by registering Detekt tasks based on source set for Android. This brings the implementation of `DetektAndroid` closer to `DetektJvm` (Although, `AndroidSourceSet` is different from Gradle's `SourceSet`, see [Documentation](https://developer.android.com/reference/tools/gradle-api/4.1/com/android/build/api/dsl/AndroidSourceSet).
- Now detekt-gradle-plugin only depends on `gradle-api`.  Although `AndroidSourceSet` and `CommonExtension` are @Incubating, they are not changed in AGP 4.2/7.0.
- Removed `detektMain` and `detektTest` and replaced with `detektAll`, because Main/Test does not really fit the standard source set groups: `main`, `test` and `androidTest`.
- Removed `sally` from `flavorSetup()` to limit the LoC of test code.

Missing items in this PR:
- [ ] Testing Groovy DSL on `sourceSetFilter`
- [ ] Testing in real Android projects

### DIFF
--- a/detekt-gradle-plugin/build.gradle.kts
+++ b/detekt-gradle-plugin/build.gradle.kts
@@ -8,13 +8,13 @@ repositories {
 }
 
 dependencies {
-    val androidGradlePlugin = "com.android.tools.build:gradle:4.1.1"
+    val agpVersion = "4.1.1"
     implementation(kotlin("gradle-plugin-api"))
-    compileOnly(androidGradlePlugin)
+    compileOnly("com.android.tools.build:gradle-api:$agpVersion")
 
     testImplementation(project(":detekt-test-utils"))
     testImplementation(kotlin("gradle-plugin"))
-    testImplementation(androidGradlePlugin)
+    testImplementation("com.android.tools.build:gradle:$agpVersion")
 
     constraints {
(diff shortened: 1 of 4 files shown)


BraisGabin:
I couldn't review the code yet. But reading the description this is a huge break change, isn't it? I mean, I agree that this is the direction that we should go but it breaks the curren behaviour of the plugin.

Would it be possible to keep detektMain/detektTest and rename `detektAll` to `detektCheck`? We have examples using `detektAll` so I assume that it could conflict with custom rules defined by our users (for example, nearly all my projects)

chao2zhang:
detekt-android is still experimental (at least the documentation from the code), so I believe we are open to making changes. 
- This split also does not make sense because on Android we usually have three groups: main, test, and androidTest. We are using different commands to compile and run them. And if people really want to keep the old behavior, they can create custom gradle tasks themselves. However, if we strongly believe that behavioral change is breaking, I can always add them back.
- Yes I can rename them to `detektCheck`.

BraisGabin:
> detekt-android is still experimental (at least the documentation from the code)

Then I'm ok breaking that compatibility. Let's see what other maintainers think.

PR #3335, opened 2020-12-30
Add new rule SleepInsteadOfDelay

severn-everett:
Added new SleepInsteadOfDelay rule for checking for whether `Thread.sleep()` is being used in suspend functions and coroutines.


detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SleepInsteadOfDelaySpec.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    describe("SleepInsteadOfDelay rule") {
+        it("should report Thread.sleep() in suspend functions") {
+            val code = """
+                suspend fun foo() {
+                    Thread.sleep(1000L)
+                }
+            """
+            Assertions.assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
+        }
```
BraisGabin, line 26: Could you add a test for: 

```
                suspend fun foo() {
                    delay(1000L)
                }
```

To show that theres not any issue?

BraisGabin (commented):
It looks good :) just one more test and it's good to go for my end.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SleepInsteadOfDelay.kt
```diff
@@ -0,0 +1,108 @@
...
+ *     Thread.sleep(1_000L)
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * suspend fun foo() {
+ *     delay(1_000L)
+ * }
+ * </compliant>
+ */
```
chao2zhang, line 41: Could we add `@requiresTypeResolution`? Because type resolution is still an experimental feature: See [doc](https://detekt.github.io/detekt/type-resolution.html#is-my-rule-using-type-resolution)

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SleepInsteadOfDelay.kt
```diff
@@ -0,0 +1,108 @@
...
+        }
+        super.visitNamedFunction(function)
+    }
+
+    override fun visitQualifiedExpression(expression: KtQualifiedExpression) {
+        val fqName = expression.getResolvedCall(bindingContext)
+            ?.resultingDescriptor
+            ?.fqNameOrNull()
+            ?.asString()
+        if (fqName == LAUNCH_COROUTINE_NAME) {
```
chao2zhang, line 74: I think we should also check `async` here. This could be in a follow-up PR.

chao2zhang (approved):

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SleepInsteadOfDelay.kt
```diff
@@ -0,0 +1,108 @@
...
+        Debt.FIVE_MINS
+    )
+
+    override fun postVisit(root: KtFile) {
+        super.postVisit(root)
+        sleepImported = false
+    }
+
+    override fun visitImportDirective(importDirective: KtImportDirective) {
+        sleepImported = importDirective.importedFqName?.asString() in IMPORT_PATHS
```
cortinico, line 58: Is `sleepImported` needed at all? You're checking the fully qualified name of the `Thread.sleep` function below (line 92) so it should be irrelevant if sleep was imported or not.

severn-everett: `sleepImported` is there to avoid having to search through all `KtCallExpression` descendents, given that there will only be dot-qualified expressions of `Thread.sleep()` if `sleepImported` is false.

severn-everett:
Suggestions for additional tests to increase code coverage would be welcome.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SleepInsteadOfDelay.kt
cortinico: > to avoid having to search through all KtCallExpression descendents

Are you trying to avoid this for performance reasons?

severn-everett:
Yes, basically. If Thread.sleep isn't imported, then there wouldn't be any
non dot-qualified expressions of the function, correct? In that case, the
rule could skip parsing through expressions that it wouldn't have an
expectation of finding the sleep() function.

Op wo 6 jan. 2021 21:44 schreef Nicola Corti <notifications@github.com>:

cortinico:
> Yes, basically. If Thread.sleep isn't imported, then there wouldn't be any non dot-qualified expressions of the function, correct? In that case, the rule could skip parsing through expressions that it wouldn't have an expectation of finding the sleep() function.

I would say that in this case the readability outweighs the performance benefits. It was not immediately clear to me why you were interested in the checking if `sleep()` was imported or not. 

I would simplify the `checkDescendants` function with 

```kotlin
    private fun PsiElement.checkDescendants(message: String) {
        forEachDescendantOfType<KtCallExpression> { it.verifyExpression(message) }
        forEachDescendantOfType<KtDotQualifiedExpression> { it.verifyExpression(message) }
    }
```

You're anyway running `verifyExpression` for all the `KtDotQualifiedExpression`, I don't see the benefit of optionally skipping the `KtCallExpression` occurrences only. You can then remove all the `postVisit` and `visitImportDirective` code as well as the specific cases in the tests.

cortinico (approved):
LGTM

BraisGabin (approved):

schalkms (approved):

mhernand40:
It seems that although `1.16.0` has not been released yet, this new rule has been published in https://detekt.github.io/detekt/coroutines.html#sleepinsteadofdelay.

For me, this led to
```
  	- Property 'coroutines>SleepInsteadOfDelay' is misspelled or does not exist.
```

When I unknowingly tried to enable this with version `1.15.0`.

cortinico:
> It seems that although `1.16.0` has not been released yet, this new rule has been published in [detekt.github.io/detekt/coroutines.html#sleepinsteadofdelay](https://detekt.github.io/detekt/coroutines.html#sleepinsteadofdelay).

That's expected as the documentation website follows the status of the current `master`. If you want to use `SleepInsteadOfDelay` you can consider using our snapshot version https://detekt.github.io/detekt/snapshots.html or wait for 1.16.0 to be released.

PR #3336, opened 2020-12-30
recover binary compatibility with 1.15.0

BraisGabin:
I move the project to 1.15.0 and runned the task `apiDump` and I found that we already had a binary compatibility problem. This code just restores that compatibility so 1.16.0 will be binary compatible with 1.15.0

### DIFF
--- a/detekt-api/api/detekt-api.api
+++ b/detekt-api/api/detekt-api.api
@@ -121,6 +121,7 @@ public abstract interface class io/gitlab/arturbosch/detekt/api/Context {
 }
 
 public final class io/gitlab/arturbosch/detekt/api/Context$DefaultImpls {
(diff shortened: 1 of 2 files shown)


chao2zhang:
I removed that default implementation because the default implementation is self-recursive and no one should be using the buggy default implementation.

If the goal is to ensure absolute binary compatibility, then this PR would make sense.

chao2zhang:
On second thought, we could maintain the binary compatibility by implementing the function with an empty body..

BraisGabin:
> On second thought, we could maintain the binary compatibility by implementing the function with an empty body.

I like that! Done.

chao2zhang (approved):

cortinico (approved):

schalkms (approved):

Issue #2560, opened 2020-03-30
Suggesting auto fixes for violated parts to fix reports issued by rules


schalkms:
Suggesting code fixes for violated parts in the code base could be a nice add-on for detekt and its users/integrators.
I know that we decided (with good reasoning) to not mess with user code. Anyway, this is a different story, since this feature doesn't mess with source code without any user intervention.

It's nice for 1st time contributors to exactly notice, what has triggered a violation and how to fix the issue.
Since this allows to fix issues automatically with the help of GitHubs auto suggestion feature, for instance, this could lead to new plugins or use cases for detekt's users and integrators.
Auto fixes are one of the most awesome things for code reviews and saves time for the human reviewer.

For instance, the following rule could be a good starting point.
**issue: LineEndsWithWhitespace**
```diff
- println("Hello World") 
+ println("Hello World")
```

BraisGabin:
Thinking about detekt, which is the difference between suggesting auto fixes and auto fix the code? I mean, which is the different in the output of detekt? Instead of change the code create `diff` files that you could apply to the code if you like them?

After reading your suggestion in the other PR I hacked around a bit and I wrote this: https://github.com/BraisGabin/seshat it is just a prove of concept.

This is a backend that receives a diff file and transforms it to comments in a PR as code sugestions. And it works: https://github.com/BraisGabin/ktlint-bot-playground/pull/1 (the comment was writted by the code in seshat)

The idea:
- Run your code formatter/fixer in your CI
- `git diff > information.diff`
- Upload the `information.diff` to a github app
- The github app parses the file and comment in the PR the changes done by the code formatter/fixer

The good part of this is that it works for any tool you want and any language. This is just a prove of concept I don't know if I'll continue with it.

cortinico:
> This is just a prove of concept I don't know if I'll continue with it.

I really like the idea. I was also looking for something like that and I haven't found much around. Happy to see that there is development on this side 👍 Happy to support making this a production ready tool.

schalkms:
To bring this in a production ready state I think detekt has to assert the reported locations in the code by our rules more thoroughly.  I once formulated this in issue #2087.
@cortinico thanks btw for your awesome support. 🙂

BraisGabin:
> An incubation in the detekt group could be an option to make this tool more visible and attract some contributors similar to the `detekt-compiler-plugin`.

[...]

PR #3339, opened 2021-01-01
Add one minute Debt

VovaStelmashchuk:
I have a suggest about add one minute Debt, because a lot of rules are a one minute fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ If you contributed to detekt but your name is not in the list, please feel free 
 - [Remco Mokveld](https://github.com/remcomokveld) - Rename Blacklist/Whitelist to more meaningful names
 - [Zachary Moore](https://github.com/zsmoore) - Rule, cli, gradle plugin, and config improvements
 - [Veyndan Stuart](https://github.com/veyndan) - New rule: UseEmptyCounterpart; Rule improvement: UselessCallOnNotNull
-- [Parimatch Tech](https://github.com/parimatchtech) - New rule: LibraryEntitiesShouldNotBePublic
+- [Parimatch Tech](https://github.com/parimatch-tech) - New rule: LibraryEntitiesShouldNotBePublic, add Debt.ONE_MIN
 - [Chao Zhang](https://github.com/chao2zhang) - SARIF report format; Rule improvements
 - [Marcelo Hernandez](https://github.com/mhernand40) - New rule: SuspendFunWithFlowReturnType
 - [Harold Martin](https://github.com/hbmartin) - Rule improvement: ClassOrdering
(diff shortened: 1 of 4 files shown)


chao2zhang:
We might need to perform a general swipe of all rules to see if they are also 1 min debt. I often laughed at myself why autoCorrectible detekt violations would take >2hours of my time to fix.

VovaStelmashchuk:
You right. Maybe we need to add flag to Debt, like a canBeFixByAutoCorrect ?

chao2zhang:
Sadly none of our detekt native rules are autoCorrectible. Only those imported from ktlint are.
We have #2560 to track the progress on implementing an autoCorrectible rule, so let's not divert this PR too much.

BraisGabin:
I think that there is not issue that can be fixed in 1 minute:

- Check the report
- Create branch
- Fix issue
- Create PR

The problem is that you can fix 60 issues in 5 minutes. Because you need nerly the same time to fix one unused import that to fix 60. I don't think that any of the standard rules should use 1 minute.

VovaStelmashchuk:
I usually fix a detekt issues in feature branch

BraisGabin:
Would it work for you if you could change the time in the yaml configuration?

VovaStelmashchuk:
Good point. Do detekt has with feature, or we need to implement this?

BraisGabin:
That feature is not implemented. But before implementing it, could you open an issue explaining your use case and why you need this feature?

This feature would add complexity to the project and I'm not sure how much value it could bring. This way other people can say it's point of view.

chao2zhang:
> I think that there is not issue that can be fixed in 1 minute:
> 
> * Check the report
> * Create branch
> * Fix issue
> * Create PR
> 
> The problem is that you can fix 60 issues in 5 minutes. Because you need nerly the same time to fix one unused import that to fix 60. I don't think that any of the standard rules should use 1 minute.

This statement makes sense. I would still like to point out that using the sum of debt of each individual finding is confusing. When we publish a PR, we fix multiple violations altogether, where each finding would take maybe 1 minute. The overhead time of creating a review may take up to 5 minutes. A more reasonable approach is to calculate the overhead (5 minutes) + the time to fix individual finding (1 minute / each), so that the total debt time is not way off the actual time spent on fixing those issues.

BraisGabin:
We need to take some decission here. I think that this conversation is interesting and we could rethink this feature a bit. I personally use it as a metric to see how my project is going so a more realistic value would be great.

What do you think about closing this and moving it to a conversation/issue? So we can find a formula that everyone (or most of us) agree with and then implement it. We can create another issue to track the feature to cusomize the time for issue if you consider it worth it.

BraisGabin:
Closing this. If a custom rule write wants it's rules to have 1 minute debt there's nothing in our API stopping them.

If someone have some ideas about how to improve the calculations related with the debt time that detekt provides, please open a discussion so we can talk about it. I do think that we need to improve that but, right now, I have no idea how.

Issue #3340, opened 2021-01-01
Add UnnecessaryFilter


VovaStelmashchuk:
## Expected Behavior of the rule
```kotlin
val x = listOf(1, 2, 3)
      .filter { it > 1 }
      .size
```
No erros

## Context
Everethink should be simple.
Add create a PR only with tests, I haven't idea how to create implementation at the moment.

BraisGabin:
I like the idea. We could check it for `Sequence`s too. And thanks for implement the tests!

But I must say that I didn't understand your PR until I read the tests you provided. Could you elaborate a bit more this issue? This way other person can pick it easily and implement it.

VovaStelmashchuk:
In same case we need to calculate count of items in collection by predicate. Sometimes people use `filter{}.size` or `filter {}.count()`. The correct way to use kotlin function count {}. Because filter create a new instanse of collection.

chao2zhang:
Close via #3341

PR #3341, opened 2021-01-01
Add UnnecessaryFilter rule

VovaStelmashchuk:
Issue: https://github.com/detekt/detekt/issues/3340


VovaStelmashchuk:
I don't have idea about implementation at the moment.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilterSpec.kt
```diff
@@ -0,0 +1,72 @@
...
+
+        it("None item") {
+            val code = """
+                val x = listOf(1, 2, 3)
+                    .filter { it > 2 }
+                    .isEmpty()
+            """
+
+            val findings = subject.compileAndLint(code)
+            assertThat(findings).isEmpty()
```
BraisGabin, line 47: Shouldn't this report one issue? This could be replaced by `none { it > 2 }`

VovaStelmashchuk: Fixed.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
```diff
@@ -0,0 +1,41 @@
...
+ *      .isEmpty()
+ * </noncompliant>
+ *
+ * <compliant>
+ * val x = listOf(1, 2, 3)
+ *      .count { it > 2 }
+ * }
+ *
+ * val x = listOf(1, 2, 3)
+ *      .none { it > 1 }
```
cortinico, line 27: Please add a `@requiresTypeResolution` here

VovaStelmashchuk:
Implementation ready. Can someone review my changes?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
chao2zhang: @cortinico maybe I missed something, but why do we need a `@requiresTypeResolution`?

VovaStelmashchuk: https://detekt.github.io/detekt/type-resolution.html#is-my-rule-using-type-resolution
I agree with @chao2zhang According to this doc we don't need this annotation. I will remove this.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * </compliant>
+ */
+class UnnecessaryFilter(config: Config = Config.empty) : Rule(config) {
+    override val issue: Issue = Issue("UnnecessaryFilter", Severity.Style,
+        "UnnecessaryFilter",
```
chao2zhang, line 40:
```suggestion
        "filter() with other collection operations may be simplified.",
```

chao2zhang (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
```diff
@@ -0,0 +1,70 @@
...
+
+        val calleeExpression = expression.calleeExpression
+        if (calleeExpression?.text != "filter") return
+
+        expression.checkNextLeaf("size")
+        expression.checkNextLeaf("count")
+        expression.checkNextLeaf(
+            "isEmpty", "any")
+        expression.checkNextLeaf(
+            "isNotEmpty", "none")
```
schalkms, line 54: This will yield false-positives, if you have property accessors like this. It's not checked, whether the calling function is actually a high order function named `filter` with a lambda expression in it.

```
call().filter.size
```

```suggestion
        expression.checkNextLeaf("isEmpty", "any")
        expression.checkNextLeaf("isNotEmpty", "none")
```

schalkms (changes requested):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
```diff
@@ -0,0 +1,70 @@
...
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtLambdaExpression
+import org.jetbrains.kotlin.psi.psiUtil.nextLeaf
+import org.jetbrains.kotlin.psi.unpackFunctionLiteral
+
+/**
+ * Unnecessary filter add complexity to the code and accomplish noting. They should be removed.
```
cortinico, line 16:
```suggestion
 * Unnecessary filters add complexity to the code and accomplish noting. They should be removed.
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
cortinico: > whether the calling function is actually a high order function named filter

nit: To be 100% exact we should also verify that the calling method is the "filter" from the Kotlin stdlib (but that will require type resolution)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
```diff
@@ -0,0 +1,70 @@
...
+        expression.checkNextLeaf(
+            "isNotEmpty", "none")
+    }
+
+    private fun KtCallExpression.checkNextLeaf(leafName: String, correctOperator: String = leafName) {
+        val hasNextLeaf = this.nextLeaf { it.text == leafName } != null
+
+        if (hasNextLeaf) {
+            report(CodeSmell(issue, Entity.from(this),
+                "$this can be replace by $correctOperator { ${this.lambda()?.text}}"))
```
cortinico, line 62:
```suggestion
                "$this can be replaced by $correctOperator { ${this.lambda()?.text}}"))
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    private fun KtCallExpression.lambda(): KtLambdaExpression? {
+        val argument = lambdaArguments.singleOrNull() ?: valueArguments.singleOrNull()
+        return argument?.getArgumentExpression()?.unpackFunctionLiteral()
+    }
```
cortinico, line 69: I'm unsure about this. While is a nice to have, it's used only for "suggestion" and it exposes  us to further breakage in the future shall the API be changed on the Kotlin side.

I believe we should either remove it, or if we wish to keep it I'd like to ask you to asset the report message in your tests.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
chao2zhang: Ah that's why you were suggesting adding requiresTypeResolution in the first place.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
VovaStelmashchuk: Test added.

VovaStelmashchuk:
I see codecov report. I can add some test for increase test coverage. But i cannot see reason for add test only for increase test coverage. The additional test will not add real benefits.

cortinico (approved):

cortinico:
> I see codecov report. I can add some test for increase test coverage. But i cannot see reason for add test only for increase test coverage. The additional test will not add real benefits.

You're fine like that 👍 Codecov coverage report is not a hard requirement for merging. In this case you're at 76.00% of your patch that is totally fine

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
```diff
@@ -0,0 +1,87 @@
...
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtLambdaExpression
+import org.jetbrains.kotlin.psi.psiUtil.nextLeaf
+import org.jetbrains.kotlin.psi.unpackFunctionLiteral
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+
+/**
+ * Unnecessary filters add complexity to the code and accomplish noting. They should be removed.
```
schalkms, line 20:
```suggestion
 * Unnecessary filters add complexity to the code and accomplish nothing. They should be removed.
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
```diff
@@ -0,0 +1,87 @@
...
+        val calleeExpression = expression.calleeExpression
+        if (calleeExpression?.text != "filter") return
+
+        val resolvedCall = expression.getResolvedCall(bindingContext) ?: return
+        if (resolvedCall.resultingDescriptor.fqNameOrNull() !in filterFqNames) return
+
+        expression.checkNextLeaf("size")
+        expression.checkNextLeaf("count")
+        expression.checkNextLeaf("isEmpty", "any")
+        expression.checkNextLeaf("isNotEmpty", "none")
```
schalkms, line 63: Those calls could also be checked by: `resultingDescriptor.fqNameOrNull()`

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
```diff
@@ -0,0 +1,99 @@
...
+        val shortName = leafName.shortName().toString()
+        val nextLeaf = this.nextLeaf { it.text == shortName }?.parent as? KtElement ?: return
+        val resolvedCall = nextLeaf.getResolvedCall(bindingContext) ?: return
+
+        if (resolvedCall.resultingDescriptor.fqNameOrNull() != leafName) return
+
+        report(
+            CodeSmell(
+                issue, Entity.from(this),
+                "${this.text} can be replaced by ${correctOperator ?: shortName} ${this.lambda()?.text}"
```
chao2zhang, line 81:
```suggestion
                "'${this.text}' can be replaced by '${correctOperator ?: shortName} ${this.lambda()?.text}'"
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilterSpec.kt
```diff
@@ -0,0 +1,159 @@
...
+        it("Filter with size") {
+            val code = """
+                val x = listOf(1, 2, 3)
+                    .filter { it > 1 }
+                    .size
+            """
+
+            val findings = subject.compileAndLintWithContext(env, code)
+            assertThat(findings).hasSize(1)
+            assertThat(findings[0]).hasMessage("filter { it > 1 } can be replaced by size { it > 1 }")
```
chao2zhang, line 25:
```suggestion
            assertThat(findings[0]).hasMessage("'filter { it > 1 }' can be replaced by 'size { it > 1 }'")
```

chao2zhang (approved):

chao2zhang:
Thank you so much for the contribution and for being patient in the back-and-forth requested changes.

chao2zhang:
For other maintainers in this PR (@cortinico @BraisGabin @schalkms): We entered 1.16-RC but is it reasonable to add new rules with active = false by default?
I can think of arguments for both sides:
- We want to only allow bug fixes in RC to speed up the full release
- We want people to try out experimental new rules early so we can enable them by default faster

cortinico:
> * We want to only allow bug fixes in RC to speed up the full release
> * We want people to try out experimental new rules early so we can enable them by default faster

Historically we've been pretty slow at enabling rules by default so I don't mind adding a disabled rule in a RC. Also if we have rules that are ready to go, like this one or #3413, I'll try to squeeze them into `1.16.0` to let them reach our users sooner (so we can get a faster feedback loop).

That being said, I'm in for merging this

BraisGabin:
:shipit:! The RC are more to make sure that some changes in the core or clients doesn't break. We did add rules between RC previusly.

PR #3354, opened 2021-01-06
Improve gradle plugin

BraisGabin:
Align how we register all our tasks in our plugin.

And I renamed "oldDetekt" to "vanillaDetekt". The old is kind of missleading because is the only non-experimental way that we expose.


chao2zhang:
What do you think about the name `typeNotResolvedDetekt` over `vanillaDetekt`?

cortinico (approved):
Nice cleanup :)

cortinico:
> What do you think about the name `typeNotResolvedDetekt` over `vanillaDetekt`?

I'd be more for `plainDetekt`

BraisGabin:
Sorry for the delay, renamed to Plain.

chao2zhang (approved):

arturbosch (approved):

Issue #3324, opened 2020-12-26
Setup the correct pipeline for integration test of detekt gradle plugin


schalkms:
@chao2zhang and me had a discussion about adding more end-to-end like tests to detekt's code base. This is definitely beneficial in terms of code quality.
Hence, the quote from PR #3319 is included in this issue. It contains a good idea, which might be followed. Of course this is an opened discussion. So please feel free to submit your ideas inside this issue.

> In my mind, such end-to-end tests should be a separate module or repo that applies detekt gradle plugin, runs against some kotlin code with violations, and validates the error can be reported. This does not exist today but I definitely can help sparkle the discussion

BraisGabin:
Other option (no exclusive): checkout some opensource projects that already use detekt and run them.

Problem: we need to be carefull with false positives if they break their build.
Pros: we test agains real code.

Some options: Firefox & LeakCanary.

schalkms:
> Problem: we need to be carefull with false positives if they break their build.
> Pros: we test agains real code.

I guess this is hard to automate, since the tests directly depend on 3rd party code. If 3rd party gets changed, those tests have to be adapted in the detekt repository. For testing purposes it's enough to have a few files containing some issues and folders to test detekt end-to-end.
For rules testing your approach is fine though. This repo already contains a script to verify rule behavior on 3rd party GitHub code. However, this topic here is strictly *not* about testing rules.

chao2zhang:
Such integration tests need to keep some code with violation of detekt rules. I doubt a 3rd party repo would satisfy our need: Well-maintained repos tend to have good hygiene to remove detekt violations quickly.

chao2zhang:
My statement was based on out-dated information. Please allow me to correct the statement here. In addition to the common unit testing, this project embraces two ways to test our Gradle plugins:
- Using ProjectBuilder from gradle-api, we can assert on specific Task this way.
- Using GradleRunner from gradle-testkit, we can perform assertions based on the input and output of Gradle.

The end-to-end would probably use GradleRunner, but the current test does not seem to have high coverage. For example:
- `DslGradleRunner.ktFileContent` only generates Kotlin code violating MagicNumber.
- AutoCorrect is not verified.

End-to-end would be exceeding beneficial as we are introducing tooling and lower-level changes.

chao2zhang:
Problem
======
One of the problems that I encountered when prototype #3327:
- Because the integration tests using gradle-testkit are all under `detekt-gradle-plugin/src/test`
[...]

PR #3356, opened 2021-01-06
Refactor integration test for detekt-gradle-plugin

chao2zhang:
This addresses #3324 by creating a different sourceSet/configuration to run integration tests using `Gradle Testkit`.
Note that `DetektPluginTest.kt` remains in the test set, as it is testing against `Gradle API`, not `Gradle Testkit`.

Gradle documentation on test kit: https://docs.gradle.org/current/userguide/test_kit.html


chao2zhang:
The Github action checks run detekt through CLI: java -jar ./detekt-cli/build/run/detekt "@./config/detekt/argsfile"
where as `./gradlew detekt` passed successfully for me locally.

There seems to be an issue that the standard gradle plugin does not respect all source sets..

schalkms (approved):

PR #3359, opened 2021-01-08
Integrate SARIF report with Github code scanning

### LINKED ISSUE #3324: Setup the correct pipeline for integration test of detekt gradle plugin

chao2zhang:
This should wrap up basic support for #3045. This PR integrates SARIF output with Github code scanning by adding a new GitHub workflow using detekt without type resolution.
- Fix a bug in KtCompiler.kt that we need to yield the absolute path of the base path.
- Use `%SRCROOT%` as uriBaseId. In [SARIF support for coding scanning](https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/sarif-support-for-code-scanning), `%SRCROOT%` was used as an example, which seems to be the only way that works with Github Action. If we pass the base path based on the local directory, like `"uriBaseId": "/Users/chaozhang/detekt"`, it will not be consumed by Github.
- In the screenshot, we are showing duplicate messages. This is because not each `CodeSmell` defines its own message. It will fallback to issue description if no message is specified, which is exactly the case in the manual test.

Follow up
=====
- Once this PR is merged, I will add a specific page in doc on integrating with Github actions.
- Type resolution is still experimental that I see errors in the log. Running detekt **with** type resolution will also generate one SARIF per source set and require Github work flow to upload multiple SARIF files, instead of one file. Given these complications, I am only using the SARIF output for detekt run **without** type resolution in this PR.
- Once there is a new detekt version released, I can change the option to enable the SARIF output by default + to enable relative path by default given the project root. (If we resolve #3324, we no longer need to wait for new detekt versions to be released).

Test
=====
See [the PR on my own fork of detekt](https://github.com/chao2zhang/detekt/pull/4), you can find GitHub codescanning like the following:
image

### DIFF
--- a/.github/workflows/detekt-with-type-resolution.yaml
+++ b/.github/workflows/detekt-with-type-resolution.yaml
@@ -33,11 +33,11 @@ jobs:
         with:
           java-version: 11
 
-      - name: Build detekt and skip checks
-        run: ./gradlew build moveJarForIntegrationTest shadowJarExecutable -x check
+      - name: Package executable jar
+        run: ./gradlew jar moveJarForIntegrationTest shadowJarExecutable
 
       - name: Run detekt-cli with argsfile, override the exit code
-        run: java -jar ./detekt-cli/build/run/detekt "@./config/detekt/argsfile" || true
+        run: ./detekt-cli/build/run/detekt "@./config/detekt/argsfile" || true
 
       - name: Upload SARIF to Github using the upload-sarif action
         uses: github/codeql-action/upload-sarif@v1
--- a/.github/workflows/detekt-with-type-resolution.yaml
+++ b/.github/workflows/detekt-with-type-resolution.yaml
@@ -9,6 +9,42 @@ on:
       - '*'
 
 jobs:
+  without-type-resolution:
+    runs-on: ubuntu-latest
+    if: ${{ !contains(github.event.head_commit.message, 'ci skip') }}
+    env:
+      GRADLE_OPTS: -Dorg.gradle.daemon=false
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+
+      - name: Cache Gradle Folders
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches/
+            ~/.gradle/wrapper/
+          key: cache-gradle-${{ hashFiles('detekt-bom/build.gradle.kts') }}
+          restore-keys: |
+            cache-gradle-
+
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+
+      - name: Build detekt and skip checks
+        run: ./gradlew build moveJarForIntegrationTest shadowJarExecutable -x check
+
+      - name: Run detekt-cli with argsfile, override the exit code
+        run: java -jar ./detekt-cli/build/run/detekt "@./config/detekt/argsfile" || true
+
+      - name: Upload SARIF to Github using the upload-sarif action
+        uses: github/codeql-action/upload-sarif@v1
+        with:
+          sarif_file: build/detekt-report.sarif
+
+
   gradle:
     runs-on: ubuntu-latest
     if: ${{ !contains(github.event.head_commit.message, 'ci skip') }}
--- a/config/detekt/argsfile
+++ b/config/detekt/argsfile
@@ -17,3 +17,5 @@ txt:./build/detekt-report.txt
 sarif:./build/detekt-report.sarif
 -p
 ./build/detekt-formatting.jar
+-bp
+.
--- a/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
+++ b/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
@@ -5,6 +5,7 @@ import io.github.detekt.sarif4j.ArtifactLocation
 import io.github.detekt.sarif4j.JacksonSarifWriter
 import io.github.detekt.sarif4j.Location
 import io.github.detekt.sarif4j.Message
+import io.github.detekt.sarif4j.OriginalUriBaseIds
 import io.github.detekt.sarif4j.PhysicalLocation
 import io.github.detekt.sarif4j.Region
 import io.github.detekt.sarif4j.Result
@@ -17,6 +18,11 @@ import io.gitlab.arturbosch.detekt.api.SetupContext
 import io.gitlab.arturbosch.detekt.api.SeverityLevel
 import io.gitlab.arturbosch.detekt.api.SingleAssign
 import io.gitlab.arturbosch.detekt.api.UnstableApi
+import io.gitlab.arturbosch.detekt.api.getOrNull
+import java.nio.file.Path
+
+const val DETEKT_OUTPUT_REPORT_BASE_PATH_KEY = "detekt.output.report.base.path"
+const val SARIF_SRCROOT_PROPERTY = "%SRCROOT%"
 
 class SarifOutputReport : OutputReport() {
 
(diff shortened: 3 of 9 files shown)


config/detekt/argsfile
```diff
@@ -17,3 +17,5 @@ txt:./build/detekt-report.txt
 sarif:./build/detekt-report.sarif
 -p
 ./build/detekt-formatting.jar
+-bp
+.
```
BraisGabin, line 21: This seems like a good default value for the cli, right?

.github/workflows/detekt-with-type-resolution.yaml
```diff
@@ -9,6 +9,42 @@ on:
...
+          restore-keys: |
+            cache-gradle-
+
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+
+      - name: Build detekt and skip checks
+        run: ./gradlew build moveJarForIntegrationTest shadowJarExecutable -x check
```
BraisGabin, line 37: Should not be enough to run `shadowJarExecutable`?

.github/workflows/detekt-with-type-resolution.yaml
```diff
@@ -9,6 +9,42 @@ on:
...
+
+      - name: Run detekt-cli with argsfile, override the exit code
+        run: java -jar ./detekt-cli/build/run/detekt "@./config/detekt/argsfile" || true
```
BraisGabin, line 40: You are using `shadowJarExecutable`. So you could run this like 
```suggestion
        run: ./detekt-cli/build/run/detekt "@./config/detekt/argsfile" || true
```

detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
```diff
@@ -64,7 +64,7 @@ private fun Finding.toIssue(ruleSetId: RuleSetId): SarifIssue = result {
                 artifactLocation = ArtifactLocation().apply {
                     if (location.filePath.basePath != null && location.filePath.relativePath != null) {
                         uri = location.filePath.relativePath?.toUnifiedString()
-                        uriBaseId = location.filePath.basePath?.toFile()?.toURI()?.toString()
+                        uriBaseId = "%SRCROOT%"
```
BraisGabin, line 67: The reason to implement sarif was the github integration but I think that we should follow the standard and no what github needs. We could ask in #3045 there was some Github and Microsoft employees offering help for this https://github.com/detekt/detekt/issues/3045#issuecomment-697019858

If there's no other option I would prefer to create a bash script to adapt the standard to what github needs (it should be easy to do with `sed`)

config/detekt/argsfile
chao2zhang: I am assuming either
1. When passing `-bp` without arguments, use the user.dir as the base path
2. When not passing `-bp`, change the default behavior to output relative paths

I like 2, but not sure if it will be too breaking.

detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
chao2zhang: [SARIF format](http://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/sarif-v2.0-csprd01.html) 3.3.3, `uriBaseId` should be present if the `uri` represents  a relative path. `The uriBaseId property can be any string; it does not need to have any particular syntax or follow any particular naming convention. In particular, it does not need to designate a machine environment variable or similar value, although it might. The SARIF producer and any SARIF consumers need to agree on the meanings of any values for the uriBaseId property that appear in the log file.`

What Github needs seems to be a more correct version of SARIF. I can definitely ask people from #3045 if I encountered further issues. For now I have found a correct prototype from my colleague that does not include duplicate message. I will convert this PR to draft and start updating it.

.github/workflows/detekt-with-type-resolution.yaml
chao2zhang: Good point. This is merely copy pasta, as we have found there is also a lot of housekeeping work to simplify our github workflows.

chao2zhang: If we run only the task `shadowJarExecutable`, detek-formatting won't be packaged into the jar.

```
➜  detekt git:(master) ✗ ./detekt-cli/build/run/detekt "@./config/detekt/argsfile"
Exception in thread "main" com.beust.jcommander.ParameterException: Provided path './build/detekt-formatting.jar' does not exist!
```

chao2zhang: If we run `build` and `shadowJarExecutable`, the same exception will be thrown.

chao2zhang: Looks like the minimal set of gradle task is `jar moveJarForIntegrationTest shadowJarExecutable`

.github/workflows/detekt-with-type-resolution.yaml
chao2zhang: Good suggestion. I will pick it up in my own commit bundled with the previous comment.

config/detekt/argsfile
chao2zhang: 3. Adding an option in the detekt config file to specify the behavior.

detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
chao2zhang: OriginalUriBaseIds is now populated reflecting the SARIF producer's local path.

.github/workflows/detekt-with-type-resolution.yaml
BraisGabin: That seems like a missconfiguration of `moveJarForIntegrationTest  ` but that's out of the scope of this PR.

config/detekt/argsfile
BraisGabin: I was thinking that if you don't set `-bp` the working dir is used by default. This way all the reports will use relative paths.  I think that all of them should, those files should be sharables so it have little sense to have a full path. (Again this is out of scope of the PR but just ideas about how we could improve this new feature)

BraisGabin (approved):

config/detekt/argsfile
chao2zhang: Do you think it is a good idea to open an issue to document all things to follow up? This PR is one of the working stream.

chao2zhang: https://github.com/detekt/detekt/issues/3360 was created to track the follow-up issues. There are many good suggestions like this so I think we need at least a single place to document where they should be.

schalkms (approved):
👏

schalkms:
I don't quite understand the following comment. Could you please elaborate?
What happens with **enabled** type resolution?
What happens with **disabled** type resolution?

> Type resolution is still experimental that I see errors in the log. Running detekt without type resolution will also end up uploading multiple SARIF files, instead of one file. I will investigate and change the type resolution when it is ready in the future.

chao2zhang:
> I don't quite understand the following comment. Could you please elaborate?
> What happens with **enabled** type resolution?
> What happens with **disabled** type resolution?
> 
> > Type resolution is still experimental that I see errors in the log. Running detekt without type resolution will also end up uploading multiple SARIF files, instead of one file. I will investigate and change the type resolution when it is ready in the future.

Running detekt with *enabled* type resolution on detekt yields compiler warnings:
```
                                                         ^
//Users/cazhang/detekt/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFormatSpec.kt:56:45: error: cannot access 'Baseline': it is internal in 'io.gitlab.arturbosch.detekt.core.baseline'
            val savedBaseline by memoized { Baseline(setOf("4", "2", "2"), setOf("1", "2", "3")) }
                                            ^
//Users/cazhang/detekt/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFormatSpec.kt:61:30: error: cannot access 'BaselineFormat': it is internal in 'io.gitlab.arturbosch.detekt.core.baseline'
                val format = BaselineFormat()
                             ^
```
Running detekt with *enabled* type resolution will generate one file per source set(I need to correct my original PR summary), hence we need to upload multiple SARIF output to Github, which has not been battled-tested and is not on the official documentation yet.

Therefore, the github workflow I added in this PR uploads the SARIF output from detekt with *disabled*.

schalkms:
Thank you for the remark! I understand the current limitations.

Issue #266, opened 2017-08-01
Rule NoDocOverPublicMethod works on public methods from internal classes


emcegp:
Is this planned behaviour? I think public methods from internal classes should be excluded from this rule. The same for internal classes themselves...

arturbosch:
Could you check this behaviour in a newer version? `NoDocOverPublicMethod` is now `UndocumentedFunction` etc, thank you!

emcegp:
I have something like this:
```internal interface HeaderContract {

    interface View {
        fun showLogo()
        fun hideLogo()
        fun showLogin()
        fun hideLogin()
        fun showJoin()
        fun hideJoin()
    }

    interface Presenter {
        fun didTapSearch(headerState: HeaderState)
    }
}```
and still getting: `UndocumentedPublicClass` and `UndocumentedPublicFunction`

**Comment by arturbosch** [comment]
How Kotlin handles the visibility of inner interfaces when the outer class is internal? Can you still use the interfaces from outside the package? If no then it is a detekt bug.

This is the new configurations for comments:
```
  UndocumentedPublicClass:
    active: false
    searchInNestedClass: true
    searchInInnerClass: true
    searchInInnerInterface: true
  UndocumentedPublicFunction:
    active: false
``` 

If you turn off searchInInnerClass, the documentation check would not appear.

emcegp:
Thanks - it did its job!

PR #3370, opened 2021-01-12
GlobalClassLoaderCache: move call to getFiles out of synchronized block

### LINKED ISSUE #3047: Running multiple Detekt tasks concurrently may cause deadlock
## Context
In my Android app, I have configured separate Detekt tasks for each variant, including test variants, with the full dependency classpath for each (e.g. `detektDebug`, `detektDebugAndroidTest`, etc).

## Expected Behavior
I am trying to run multiple at the same time, e.g. `./gradlew detektDebug detektDebugAndroidTest`, which Gradle (correctly) infers can happen concurrently.

## Observed Behavior
Running multiple Detekt tasks concurrently leads to a deadlock in certain situations. The problem appears to be with [`ClassLoaderCache`](https://github.com/detekt/detekt/blob/c08de812c11c9a84d48ef0ee2e659fa27c8f77b5/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt#L1): `FileCollection.files` resolves the included files, but also (unexpectedly to me) blocks until the dependencies are satisfied if they haven't been already. By calling `files` while being `synchronized` on `ClassLoaderCache`, other `Detekt` invocations become blocked waiting on the monitor, which results in the first `Detekt` never having its dependencies satisfied, leading to a deadlock.

## Steps to Reproduce
I don't have a code sample at the moment (but can provide if necessary) since the configuration and setup would be non-trivial to extract, but the gist is:
- Register a `Detekt` task with a `classpath` containing the classpath of the project's main source set / variant
- Register a `Detekt` task with a `classpath` containing the classpath the project's test source set / variant (I _think_ this is important, based on the fact that the test Detekt task will depend on the classpath of the main source set / variant)
- Invoke Gradle with both tasks as goals (and sufficient cores/workers to run multiple concurrently)

A thread dump for when I have observed this occurring can be found [here](https://gist.github.com/damianw/f96c6db70155974065bc48ec8e0aeb72):
- [`"Execution worker for ':' Thread 3" #266`](https://gist.github.com/damianw/f96c6db70155974065bc48ec8e0aeb72#file-dump-txt-L338-L499) is blocked on `FileCollection.files`, [having locked `ClassLoaderCache`](https://gist.github.com/damianw/f96c6db70155974065bc48ec8e0aeb72#file-dump-txt-L393)
- [`"Execution worker for ':' Thread 11" #274`](https://gist.github.com/damianw/f96c6db70155974065bc48ec8e0aeb72#file-dump-txt-L101-L210) is blocked, [waiting to lock `ClassLoaderCache`](https://gist.github.com/damianw/f96c6db70155974065bc48ec8e0aeb72#file-dump-txt-L102-L104)

If I were to provide a suggestion, I think that moving the call to `FileCollection.files` outside of the `synchronized` block would be the simplest fix.

## Your Environment
* Version of Gradle used (if applicable): 6.5
* Operating System and version: Mac OS 10.15.6

damianw:
Fixes #3047

This should fix the deadlock issue from #3047 by moving the call to `FileCollection.getFiles` out of the `synchronized` block in `GlobalClassLoaderCache`. I am having a harder time reproducing the issue these days, but we are seeing `Detekt` tasks hanging in our CI jobs ~1% of the time, which I suspect is caused by #3047. Hopefully we see those disappear after this change.

I added an (admittedly contrived) test to verify that the cache doesn't call `FileCollection.getFiles` while synchronized on itself, which failed prior to this change but now passes. I also verified that I was able to run Detekt without issue on my own project.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
@@ -2,6 +2,7 @@ package io.gitlab.arturbosch.detekt.internal
 
 import org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport
 import org.gradle.api.file.FileCollection
+import java.io.File
 import java.net.URLClassLoader
 
 interface ClassLoaderCache {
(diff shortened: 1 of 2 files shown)


cortinico:
I guess @arturbosch is the best candidate to review this.

I just have one question:

> by moving the call to FileCollection.getFiles out of the synchronized block in GlobalClassLoaderCache

What makes you think that `.getFiles` is the culprit?

damianw:
Thanks for the quick reply!

> What makes you think that `.getFiles` is the culprit?

This is based on the thread dump that I provided in #3047:
> A thread dump for when I have observed this occurring can be found [here](https://gist.github.com/damianw/f96c6db70155974065bc48ec8e0aeb72):
> - [`"Execution worker for ':' Thread 3" #266`](https://gist.github.com/damianw/f96c6db70155974065bc48ec8e0aeb72#file-dump-txt-L338-L499) is blocked on `FileCollection.files`, [having locked `ClassLoaderCache`](https://gist.github.com/damianw/f96c6db70155974065bc48ec8e0aeb72#file-dump-txt-L393)
> - [`"Execution worker for ':' Thread 11" #274`](https://gist.github.com/damianw/f96c6db70155974065bc48ec8e0aeb72#file-dump-txt-L101-L210) is blocked, [waiting to lock `ClassLoaderCache`](https://gist.github.com/damianw/f96c6db70155974065bc48ec8e0aeb72#file-dump-txt-L102-L104)

arturbosch (approved):
Hey, thanks for investigating this hard issue.
Code looks good to me.
Yes, the other detekt invocations wait for the first one, but I still don't understand why `getFiles` blocks/waits and who blocked it?
`DefaultResourceLockCoordinationService.withStateLock` and `DefaultWorkerLeaseService.acquireLocks` really look suspicious.
I'm okay with merging this and see in a RC release if this helps and/or does not introduce other issues.

chao2zhang:
@damianw Did this PR help you with CI reliability? We may have an incoming PR #3799 that is impacting the behavior again.

sschuberth:
Actually, I don't think #3799 has an impact on this PR, as IIUC the main point of this PR was to not do `val classpathFiles = classpath.files` in a `synchronized` block, which still is the case with #3799. Also, if #3799 had a negative impact on the changes in this PR, I'd also expect the test from this PR to break.

damianw:
I haven't seen this issue recently so as far as I am aware it seems to have been fixed (although the intermittent nature doesn't rule out other confounding variables). 😅 

And yes, that's my conclusion as well @sschuberth.

PR #3380, opened 2021-01-15
Fix EmptyConfig making all rules active in production

### LINKED ISSUE #3378: Not failing or not using baseline
## Expected Behavior
detekt should run using the baseline and fail when issues are found if fastFail is set to true.

## Observed Behavior
If I don't set fastFail, then detekt runs and will not fail when issues are found. If I set fastFail = true, then detekt doesn't seem to use the baseline file, or somehow find new issues

## Steps to Reproduce
In my buildSrc module, I include 'io.gitlab.arturbosch.detekt:detekt-gradle-plugin:1.15.0' in the dependencies. Then in a custom plugin class that all my project modules use, I have 

```
pluginManager.apply(DetektPlugin::class.java)
tasks.withType(Detekt::class.java).configureEach {
    it.jvmTarget = "1.8"
    it.config.from(rootProject.file("detekt-config.yml"))
    it.reports.apply {
        html.enabled = false
        xml.enabled = false
    }
    val file = File(rootProject.projectDir, "detekt-baseline.xml")
    it.baseline.set(file)
}
```
Next, I run detekt. 

## Context
I'm trying to include the detekt task in my CI for pull request, etc, but if the task doesn't fail, we won't know there are issues.

## Your Environment
* Version of detekt used: 1.15.0
* Version of Gradle used (if applicable): 6.5
* Version of Gradle plugin: 4.0.1
* Operating System and version: Mac OS, catalina

### LINKED ISSUE #3379: Empty custom config enables rules disabled by default
## Expected Behavior

When I create an empty custom config file and enable it though `detekt { config ... }`, it should work as if the custom config wasn't enabled.

## Observed Behavior
My Gradle config is as follows:

```
detekt {
    buildUponDefaultConfig = true
    config = files("$projectDir/detekt-custom-config.yml")
}
```

and the config file is empty. When I run Detekt, I'm getting

```

* What went wrong:
Execution failed for task ':backend:detekt'.
> Rule 'AbsentOrWrongFileLicense': License template file not found at `/home/piotr/repos/teams/backend/license.template`.
  Create file license header file or check your running path.
```

I customized one rule that I needed and it now seems to work correctly.

## Steps to Reproduce

See above.

## Context

This issue caused confusion as the behavior is counter-intuitive, or if it's somehow by design, it's not documented well enough.

## Your Environment

* Version of detekt used: 1.15.0
* Operating System and version: Ubuntu 20.04
* Link to your project (if it's a public repository): (private repo - easy to reproduce given above description)

chao2zhang:
Fixes #3379

The root cause is that we are returning `Config.empty` when the custom config file is an empty yaml. This is problematic because the `Config.empty` by default enables all rules.
We fixes this by returning a `Config(emptyMap)` instead of an `EmptyConfig`. This conforms to the comment "An empty configuration with no properties. This config should only be used in **test** cases. Always returns the default value except when 'active' is queried, it returns true."

I also updated the default-detekt-config.yml to list all FileProcessorListeners, as discovered when debugging #3379.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
@@ -50,7 +50,7 @@ interface Config {
         /**
          * An empty configuration with no properties.
          * This config should only be used in test cases.
-         * Always returns the default value except when 'active' is queried, it returns true .
+         * Always returns the default value except when 'active' is queried, it returns true.
          */
         val empty: Config = EmptyConfig
 
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/YamlConfig.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/YamlConfig.kt
@@ -72,7 +72,7 @@ class YamlConfig internal constructor(
                 Yaml().loadAs(it, Map::class.java) as Map<*, *>?
             }.getOrElse { throw Config.InvalidConfigurationError(it) }
             if (map == null) {
-                Config.empty
+                YamlConfig(emptyMap())
             } else {
                 YamlConfig(map as Map<String, Any>)
             }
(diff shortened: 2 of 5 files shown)


cortinico (approved):
Nice work 👌  I guess the CI is unhappy as you have to update the `default-detekt-config` file.

schalkms (approved):

PR #3385, opened 2021-01-17
Make gradle plugin defaults consider multiplatform sourcesets

gustavkarlsson:
By default, the gradle plugin only considers `src/main/java` and `src/main/kotlin` when looking for source directories. This means multiplatform projects must add custom configuration to work.

This PR adds multiplatform directories by using a `FileTree` with patterns to find all applicable directories

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
@@ -23,11 +23,11 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
     val reports = DetektReports()
 
     var input: ConfigurableFileCollection =
-        objects.fileCollection().from(DEFAULT_SRC_DIR_JAVA, DEFAULT_SRC_DIR_KOTLIN)
+        objects.fileCollection().from(
+            objects.fileTree().from("src").include("main/kotlin/", "*Main/kotlin/", "main/java/", "*Main/java/")
+        )
 
     var baseline: File? = null
-
-    var basePath: String? = null
 
     var config: ConfigurableFileCollection = objects.fileCollection()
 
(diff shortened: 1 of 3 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -23,12 +23,12 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
     val reports = DetektReports()
 
     var input: ConfigurableFileCollection =
-        objects.fileCollection().from(DEFAULT_SRC_DIR_JAVA, DEFAULT_SRC_DIR_KOTLIN)
+        objects.fileCollection().from(
+            objects.fileTree().from("src").include("main/kotlin/", "*Main/kotlin/", "main/java/", "*Main/java/")
```
cortinico, line 27: Thanks for taking the time for working on this PR.

I believe we should instead take a similar approach to what we do for the Android plugin: check if the `multiplatform` Gradle plugin is incuded, if yes create tasks for every source sets.

Using `*Main` as a default for the `input` property exposes us to potential bugs and doesn't really address the full multiplatform support.

schalkms: Thanks for this PR! I agree with Nico. I think the mentioned long-sighted approach is a better fit here.

gustavkarlsson: Fair enough! 🙂

cortinico:
Hej @gustavkarlsson
What's the status of this PR? Are you actively working on it?
If not, I'd like to pick-up multiplatform support for detekt Gradle plugin, 
but I don't want to step on your shoes 👌

gustavkarlsson:
@cortinico I'm not working on it. I honestly don't know enough gradle. More than happy to see someone else do it 😄

cortinico:
> @cortinico I'm not working on it. I honestly don't know enough gradle. More than happy to see someone else do it 😄

#3453 should add KMM support to detekt 👍

Issue #2340, opened 2020-02-10
Hyperlink to error inside Android studio


ahmadmssm:
I can not find a way to make the reported issues hyperlink that opens the line with the issue inside a class

schalkms:
Do you use the command line report of detekt?
Do you use the command line emulator in Android Studio?
What OS and what terminal do you use?
The links in the command line report are clickable (see #171 for details).

3flex:
I have the same issue. Running detekt directly from IntelliJ's Gradle integration (on Windows) so it's not specific to Android Studio, but IntelliJ itself.

ahmadmssm:
@schalkms Thanks for the quick reply, 
1. I use this command ` gradle detekt`, and I write it in Android studio's built-in terminal.
2. I use the run green button to launch the Emulator (I think it is something that is not related to the emulator).
3. I Use MAC OSC Catalina 10.15.3 and use the native mac terminal.

schalkms:
Okay, did you try it with zsh in the native Mac terminal or with iTerm? By default Mac OS used the bash, but if I remember it correctly they switched to zsh in one of the recent OS versions.

ahmadmssm:
@schalkms Yes, that's right, MAC OS switched to zsh starting from Catalina, I guess.
I'm using the native terminal and not iTerm or iTerm2.

schalkms:
Does it work in the _native_ terminal?

ahmadmssm:
No

arturbosch:
Can we do something about this in detekt?
I guess the terminal emulator/intellij should find absolute paths and make them clickable?

schalkms:
This was already implemented in #171 but somehow broke with the new path resolving API.
The printed path needs to have a special format, so that the terminal understands and interprets it as a hyperlink.

PR #3386, opened 2021-01-18
Fix hyperlink for elements of the type 'KtFile'

biazmoreira:
Issue: https://github.com/detekt/detekt/issues/2340

**Description:**

IntelliJ (and consequently Android Studio), considers the following implementation when enabling hyperlink in its terminal emulator: [AbstractFileHyperlinkFilter](https://github.com/JetBrains/intellij-community/blob/master/platform/execution-impl/src/com/intellij/execution/filters/AbstractFileHyperlinkFilter.java)

I noticed that there were two issues happening related to hyperlinks not being enabled in the output of detect:
1. Files outside the considered project scope by IntelliJ (/build, etc) were not hyperlinking because IntelliJ uses the Filter above, and consequently, the file must be inside the considered project scope by IntelliJ in order to be Highlighted. 
2. And output with two file paths in the same line -- this usually happens for elements of type KtFile, classes and functions seem to hyperlink just fine. I didn’t find a reason as to why this happens, but I think it’s out of the scope of Detekt to fix it -- other tools have the same problem. 

Example of the point 2 before the changes I applied in this PR:
Screenshot 2021-01-17 at 21 42 43

------

**Fix:**
A fix for point 2 is to simplify the Entity name in the case of Files by getting the file name without the absolute path since the path is also written along with the output. 

Regarding other terminal emulators, it really depends on the terminal emulator you're using to specify the way of hyperlinking: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda. 
This only tackles the IntelliJ Terminal Emulator.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Signatures.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Signatures.kt
@@ -18,6 +18,16 @@ internal fun PsiElement.searchName(): String {
     return this.namedUnwrappedElement?.name?.formatElementName() ?: "<UnknownName>"
 }
 
+/*
+ * PsiElements of type KtFile use the absolute path as their names. This resulted in
+ * having two absolute paths being printed along with the output: one for the name
+ * and one for the location. Consequently, IntelliJ was not hyperlinking them because
+ * it only hyperlinks one path file per line.
+ *
+ * This gets the file name only, instead of the entire absolute path, and takes it as the element name.
+ *
+ * Example: KtFile with name /full/path/to/Test.kt will have its name formatted to be simply Test.kt
+ */
 private fun String.formatElementName(): String =
     if (contains(File.separatorChar)) substringAfterLast(File.separatorChar)
     else this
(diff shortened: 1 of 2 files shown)


detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/EntitySpec.kt
```diff
@@ -14,9 +14,10 @@ import java.nio.file.Paths
 class EntitySpec : Spek({
 
     describe("entity signatures") {
-
+        val path = Paths.get("/full/path/to/Test.kt")
         val code by memoized(CachingMode.SCOPE) {
-            compileContentForTest("""
+            compileContentForTest(
+                """
```
schalkms, line 20: minor formatting improvement

```suggestion
            compileContentForTest("""
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Signatures.kt
```diff
@@ -10,13 +10,18 @@ import org.jetbrains.kotlin.psi.psiUtil.endOffset
...
 private val multipleWhitespaces = Regex("\\s{2,}")
 
 internal fun PsiElement.searchName(): String {
-    return this.namedUnwrappedElement?.name ?: "<UnknownName>"
+    return this.namedUnwrappedElement?.name?.formatElementName() ?: "<UnknownName>"
 }
 
+private fun String.formatElementName(): String =
+    if (contains(File.separatorChar)) substringAfterLast(File.separatorChar)
+    else this
```
schalkms, line 23: I think this function should really have an explanation on top of it. Future readers will be grateful for any hint on why separator chars are filtered out here.

schalkms (commented):
Thanks for the detailed investigation! This PR looks very good! Finally, this issue gets closed. I like it.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Signatures.kt
biazmoreira: Done :)

BraisGabin (approved):
Thanks! Really good investigation!

schalkms (approved):

PR #3390, opened 2021-01-19
Ignore actual functions in FunctionOnlyReturningConstant (#3388)

### LINKED ISSUE #3388: FunctionOnlyReturningConstant should ignore actual functions
By default, `actual` functions in multiplatform projects should be ignored when evaluating `FunctionOnlyReturningConstant`.

There should be a configuration option `ignoreActualFunction`, which would allow changing this behaviour.

#### commonMain
```kotlin
expect class Foo {
    fun foo(param: Any): Boolean
}
```
#### jvmMain
```kotlin
actual class Foo {
    actual fun foo(param: Any): Boolean = checkSomeCondition(param)
}
```
#### jsMain
```kotlin
actual class Foo {
    actual fun foo(param: Any): Boolean = false    // <-- FunctionOnlyReturningConstant should be ignored
}
```

## Context
Similar to `ignoreOverridableFunction`.

Implementing an `actual` function, is conceptually similar to overriding a function from an interface/abstract class.
It makes sense for such a function to provide complex logic on some platforms, and only return a constant value on others.

sczerwinski:
Closes #3388

By default, `actual` functions in multiplatform projects should be ignored when evaluating `FunctionOnlyReturningConstant`.

There should be a configuration option `ignoreActualFunction`, which would allow changing this behaviour.

#### commonMain
```kotlin
expect class Foo {
    fun foo(param: Any): Boolean
}
```
#### jvmMain
```kotlin
actual class Foo {
    actual fun foo(param: Any): Boolean = checkSomeCondition(param)
}
```
#### jsMain
```kotlin
actual class Foo {
    actual fun foo(param: Any): Boolean = false    // <-- FunctionOnlyReturningConstant should be ignored
}
```

## Context
Similar to `ignoreOverridableFunction`.

Implementing an `actual` function, is conceptually similar to overriding a function from an interface/abstract class.
It makes sense for such a function to provide complex logic on some platforms, and only return a constant value on others.

### DIFF
--- a/.github/workflows/codecoverage.yaml
+++ b/.github/workflows/codecoverage.yaml
@@ -6,7 +6,7 @@ on:
       - master
   pull_request:
     branches:
-      - '*'
+      - '**'
 
 jobs:
   publish-code-coverage:
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/FunctionOnlyReturningConstantSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/FunctionOnlyReturningConstantSpec.kt
@@ -29,7 +29,7 @@ class FunctionOnlyReturningConstantSpec : Spek({
             actual class ActualFunctionReturningConstant {
                 actual fun f() = 1
             }
-        """.trimIndent()
+        """
 
         it("does not report actual functions which return constants") {
             assertThat(subject.lint(actualFunctionCode)).isEmpty()
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/FunctionOnlyReturningConstantSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/FunctionOnlyReturningConstantSpec.kt
@@ -25,6 +25,22 @@ class FunctionOnlyReturningConstantSpec : Spek({
             assertThat(rule.lint(path)).hasSize(9)
         }
 
+        val actualFunctionCode = """
+            actual class ActualFunctionReturningConstant {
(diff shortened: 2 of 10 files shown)


.github/workflows/codecoverage.yaml
```diff
@@ -6,7 +6,7 @@ on:
       - master
   pull_request:
     branches:
-      - '*'
+      - '**'
```
chao2zhang, line 9: Thank you for fixing the glob pattern in github workflows.

chao2zhang (approved):

.github/workflows/codecoverage.yaml
sczerwinski: Happy to help. The difference between `*` and `**` is well hidden in GitHub Actions documentation. Took me some time to figure out the first time.

chao2zhang: I learned [Filter Pattern Cheat Sheet](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) by reviewing this PR 🙏

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/FunctionOnlyReturningConstantSpec.kt
```diff
@@ -25,6 +25,22 @@ class FunctionOnlyReturningConstantSpec : Spek({
             assertThat(rule.lint(path)).hasSize(9)
         }
 
+        val actualFunctionCode = """
+            actual class ActualFunctionReturningConstant {
+                actual fun f() = 1
+            }
+        """.trimIndent()
```
schalkms, line 32:
```suggestion
        """
```

schalkms (approved):
Thanks!

PR #3391, opened 2021-01-19
Introduce --max-issues flag for cli - #2267

arturbosch:
Part 2 of replacing the --fail-fast flag

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
@@ -25,7 +25,12 @@ internal fun CliArgs.createSpec(output: Appendable, error: Appendable): Processi
             autoCorrect = args.autoCorrect
             @Suppress("DEPRECATION")
             activateAllRules = args.failFast || args.allRules
-            maxIssuePolicy = RulesSpec.MaxIssuePolicy.NonSpecified // not yet supported; prefer to read from config
+            maxIssuePolicy = when (val count = args.maxIssues) {
+                null -> RulesSpec.MaxIssuePolicy.NonSpecified // prefer to read from config
+                0 -> RulesSpec.MaxIssuePolicy.NoneAllowed
+                in -1 downTo Int.MIN_VALUE -> RulesSpec.MaxIssuePolicy.AllowAny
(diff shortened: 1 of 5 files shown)


schalkms (commented):
I guess the documentation should also get some love with an update concerning the new flag. Otherwise, this pr looks nice.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
```diff
@@ -25,7 +25,12 @@ internal fun CliArgs.createSpec(output: Appendable, error: Appendable): Processi
             autoCorrect = args.autoCorrect
             @Suppress("DEPRECATION")
             activateAllRules = args.failFast || args.allRules
-            maxIssuePolicy = RulesSpec.MaxIssuePolicy.NonSpecified // not yet supported; prefer to read from config
+            maxIssuePolicy = when (val count = args.maxIssues) {
+                null -> RulesSpec.MaxIssuePolicy.NonSpecified // prefer to read from config
+                0 -> RulesSpec.MaxIssuePolicy.NoneAllowed
+                in -1 downTo Int.MIN_VALUE -> RulesSpec.MaxIssuePolicy.AllowAny
+                else -> RulesSpec.MaxIssuePolicy.AllowAmount(count)
```
cortinico, line 32: nit: are you able to add a test case for those two scenarios as well? Perhaps in the `RunnerSpec.kt` file as you did for the `0` case?

cortinico (approved):
LGTM, just a minor comment

docs/pages/gettingstarted/cli.md
```diff
@@ -82,6 +82,10 @@ Usage: detekt [options]
       Default: false
     --excludes, -ex
       Globing patterns describing paths to exclude from the analysis.
+    --max-issues
+      DEPRECATED: please use '--build-upon-default-config' together with '--all-rules'.
```
BraisGabin, line 86: I think that this first line should not be here.

chao2zhang: Ah I meant to add this line to `--fail-fast`. This is a good catch

chao2zhang: Let me revert the changes in `cli.md` since we will be auto-generating `cli-options.md` #3399

BraisGabin (approved):

PR #3393, opened 2021-01-20
Update documentations on snapshots

cortinico:
I'm updating the documentation around using SNAPSHOTS as it was lacking.

Moreover I'm updating the SNAPSHOT name to be `master-SNAPSHOT`. If we don't like it we could use the upcoming version (but that forces us to keep two versions number updated).

### DIFF
--- a/docs/pages/snapshots.md
+++ b/docs/pages/snapshots.md
@@ -3,10 +3,10 @@ title: "Using Snapshots"
 keywords: snapshot releases 
 sidebar: 
 permalink: snapshots.html
-summary: This page explains you how you can setup snapshots for your Detekt build to test the latest unreleased features. 
+summary: This page explains how you can setup snapshots for your Detekt build to test the latest unreleased features. 
 ---
 
-This page explains you how you can use our **latest snapshots**, to test upcoming unreleased features.
+This page explains how you can use our **latest snapshots** to test upcoming unreleased features.
 
 ## Where to download snapshots
 
--- a/docs/pages/snapshots.md
+++ b/docs/pages/snapshots.md
@@ -1,0 +1,82 @@
+---
+title: "Using Snapshots"
+keywords: snapshot releases 
+sidebar: 
+permalink: snapshots.html
+summary: This page explains you how you can setup snapshots for your Detekt build to test the latest unreleased features. 
+---
+
+This page explains you how you can use our **latest snapshots**, to test upcoming unreleased features.
+
+## Where to download snapshots
+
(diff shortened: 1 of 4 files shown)


BraisGabin (approved):
👏🏻👏🏻👏🏻👏🏻 Nice one! And I like the `master-SNAPSHOT` name.

docs/pages/snapshots.md
```diff
@@ -0,0 +1,82 @@
...
+    repositories {
+        // Your other repos here.
+        maven {
+            url = uri("https://oss.sonatype.org/content/repositories/snapshots/")
+        }
+    }
+}
+```
+
+Please note that the extra `resolutionStrategy{}` block is needed as we don't publish a Gradle Plugin marker for our snapshots.
```
chao2zhang, line 82: I came to read https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_markers. I am wondering why we don't publish a Gradle Plugin marker in our build?

chao2zhang (approved):
I have one question if you may know the answer to.

docs/pages/snapshots.md
cortinico: The reason is here: https://github.com/detekt/detekt/blob/1164c34f1ce1bd4f8f4a6df57d7599798f1ff06d/detekt-gradle-plugin/build.gradle.kts#L58-L60

I'm unsure of what was the exact issue with `isAutomatedPublishing` and the invalid signature, but we can definitely look into it.

docs/pages/snapshots.md
```diff
@@ -0,0 +1,82 @@
+---
+title: "Using Snapshots"
+keywords: snapshot releases 
+sidebar: 
+permalink: snapshots.html
+summary: This page explains you how you can setup snapshots for your Detekt build to test the latest unreleased features. 
+---
+
+This page explains you how you can use our **latest snapshots**, to test upcoming unreleased features.
```
schalkms, line 9:
```suggestion
This page explains how you can use our **latest snapshots** to test upcoming unreleased features.
```

docs/pages/snapshots.md
```diff
@@ -0,0 +1,82 @@
+---
+title: "Using Snapshots"
+keywords: snapshot releases 
+sidebar: 
+permalink: snapshots.html
+summary: This page explains you how you can setup snapshots for your Detekt build to test the latest unreleased features.
```
schalkms, line 6:
```suggestion
summary: This page explains how you can setup snapshots for your Detekt build to test the latest unreleased features. 
```

schalkms (approved):

PR #3394, opened 2021-01-21
Merge XML outputs into a consolidated report

chao2zhang:
Addresses #1295

Here are my own judgment calls that needs review:
- Create a new gradle plugin for the root project. Because merging only makes sense for a multi-project setup. Configuring in any subproject will create dependencies across subprojects, which is not a good gradle practise.
- Not depending on XML parsing library because we can safely assume that all the generated output is managed by detekt.
- Not automatically apply the merge when running `./gradlew detekt` or `./gradlew detektMain`, this is because of incremental cacheability of gradle that does not guarantee each detekt task will be actually run / and or each individual gradle task may fail due to detected issues. The expectation for our users is to run `./gradlew detekt mergeXmlReport --continue` or `./gradlew detekt --continue; ./gradlew mergeXmlReport`

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektProjectPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektProjectPlugin.kt
@@ -1,10 +1,5 @@
 package io.gitlab.arturbosch.detekt
 
-import com.android.build.gradle.internal.tasks.factory.dependsOn
-import io.gitlab.arturbosch.detekt.report.XmlOutputMergeTask
-import org.gradle.api.Plugin
-import org.gradle.api.Project
-import org.gradle.api.file.ConfigurableFileCollection
 import io.gitlab.arturbosch.detekt.report.XmlOutputMergeTask
 import org.gradle.api.Plugin
 import org.gradle.api.Project
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektProjectPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektProjectPlugin.kt
@@ -1,0 +1,45 @@
+package io.gitlab.arturbosch.detekt
+
+import com.android.build.gradle.internal.tasks.factory.dependsOn
+import io.gitlab.arturbosch.detekt.report.XmlOutputMergeTask
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.plugins.ReportingBasePlugin
+import org.gradle.api.reporting.ReportingExtension
+
+class DetektProjectPlugin : Plugin<Project> {
+
+    override fun apply(project: Project) {
+        require(project.isRootProject()) {
+            "DetektProjectPlugin should be applied to rootProject only."
+        }
+        project.pluginManager.apply(ReportingBasePlugin::class.java)
+        project.registerMergeReportTask()
+    }
+
+    private fun Project.registerMergeReportTask() {
+        val xmlOutputMergeTask = tasks.register(MERGE_XML_REPORT, XmlOutputMergeTask::class.java) { task ->
+            task.description = "Merge XML report for all individual detekt tasks in this project"
+            task.output.set(
+                project.extensions.getByType(ReportingExtension::class.java).file(MERGE_XML_FILE_RELATIVE_PATH)
+            )
+        }
+        // Delay the configuration so that detektTask.reports.xml.destination is always set
+        // when configuring the input of merge task.
+        project.afterEvaluate { project ->
+            project.subprojects { subproject ->
+                subproject.tasks.withType(Detekt::class.java).configureEach { detektTask ->
+                    xmlOutputMergeTask.configure { it.input.from(detektTask.reports.xml.destination) }
+                }
+            }
+        }
+    }
+
+    private fun Project.isRootProject() = path == ":"
+
+    companion object {
+        private const val MERGE_XML_REPORT = "mergeXmlReport"
+        private const val MERGE_XML_FILE_RELATIVE_PATH = "detekt/merged.xml"
+    }
+}
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektProjectPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektProjectPlugin.kt
@@ -5,6 +5,9 @@ import io.gitlab.arturbosch.detekt.report.XmlOutputMergeTask
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.file.ConfigurableFileCollection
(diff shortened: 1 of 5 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/XmlOutputMerger.kt
```diff
@@ -0,0 +1,40 @@
...
+            return null
+        }
+        val lines = input.readLines()
+        if (lines.size < MIN_LINES) return null
+        return if (lines.first().trim() == XML_PROLOG &&
+            lines[1].trim() == CHECKSTYLE_OPEN_TAG &&
+            lines.last().trim() == CHECKSTYLE_CLOSE_TAG) {
+            lines.subList(2, lines.size - 1).joinToString(separator = System.lineSeparator())
+        } else {
+            null
```
schalkms, line 37: I guess it makes more sense to return the empty string here. Thus, the extra null check on the call side can be spared.

```suggestion
            ""
```

schalkms (commented):
Thank you very much for picking up this feature request! 🙂
This is a highly requested feature for a long time. Please let me review this PR from a user perspective. Since this PR is a draft, I guess it needs more E2E testing and some refactoring.

> - Create a new gradle plugin for the root project. Because merging only makes sense for a multi-project setup. Configuring in any subproject will create dependencies across subprojects, which is not a good gradle practise.

Agreed.

> - Not depending on XML parsing library because we can safely assume that all the generated output is managed by detekt.

I'm not sure about this. Could we get encoding issues and problems with special characters and so on?

> - Not automatically apply the merge when running `./gradlew detekt` or `./gradlew detektMain`, this is because of incremental cacheability of gradle that does not guarantee each detekt task will be actually run / and or each individual gradle task may fail due to detected issues. The expectation for our users is to run `./gradlew detekt mergeXmlReport --continue` or `./gradlew detekt --continue; ./gradlew mergeXmlReport`

Agreed. This should be a separate task / flag.

chao2zhang:
> I'm not sure about this. Could we get encoding issues and problems with special characters and so on?

I have updated this PR by using Java XML API. There are some nuances that ending up with more configurations, and the output is not ideal either: attributes are sorted that column is shown before line.

chao2zhang:
> This is a highly requested feature for a long time. Please let me review this PR from a user perspective. Since this PR is a draft, I guess it needs more E2E testing and some refactoring.

The only missing part is documentation. If the code looks good, I can add documentation in the same PR or follow-up PR.
I have done e2e testing in https://github.com/chao2zhang/RequireNamedArgument/tree/testdetekt

schalkms:
Okay, thanks for the detailed explanation.The codecov report shows a decrease in the code coverage. Is this only because of the newly added Gradle tasks?

chao2zhang:
> Okay, thanks for the detailed explanation.The codecov report shows a decrease in the code coverage. Is this only because of the newly added Gradle tasks?

Good catch.
- codecov report won't be able to report gradle integration test coverage.
- I can add integration test in this PR if possible.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektProjectPlugin.kt
```diff
@@ -0,0 +1,43 @@
...
+        }
+        project.pluginManager.apply(ReportingBasePlugin::class.java)
+        project.registerMergeReportTask()
+    }
+
+    private fun Project.registerMergeReportTask() {
+        val xmlOutputMergeTask = tasks.register(MERGE_XML_REPORT, XmlOutputMergeTask::class.java) { task ->
+            task.description = "Merge XML report for all individual detekt tasks in this project"
+            task.output.set(
+                project.extensions.getByType(ReportingExtension::class.java).file(MERGE_XML_FILE_RELATIVE_PATH)
```
cortinico, line 23: Commenting here but it's more a design feedback. I believe this task should interact with the other classes for reporting we already have.

Specifically what happens if the user has defined those fields from the `DetektExtension`:
https://github.com/detekt/detekt/blob/b0041fea94bfa6335c7772111f5d4f7a418d705a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt#L20-L23

cortinico (commented):
I'm wondering if adding a second plugin is the right approach here. 

We followed the single-plugin approach once we added Android support and I belive this could be implemented in a similar way (if the `detekt` plugin is added to the root project, add the `merge` task as well).

I'm afraid that having two Gradle plugins will add more complexity (double artifacts, double plugin markers, users confusions about which plugin should be applied where, etc.) than benefits.

cortinico:
As a side note to my comments: great work @chao2zhang 🎉  This was a highly-requested feature, thanks for tackling it :)

chao2zhang:
> I'm afraid that having two Gradle plugins will add more complexity (double artifacts, double plugin markers, users confusions about which plugin should be applied where, etc.) than benefits.

I started with developing in the same plugin, and it didn't take long for me to realize that we need to deal with an exponential level of complexity.
- Single module project: If using the same plugin, this should be easier. However, for the single module project, merging the XML output does not seem a problem necessary to solve.
- Multi-module with detekt on both root project and subprojects: If using the same plugin, we should not expect any Detekt task to root module because of the issues reported in #3389, only subprojects should have Detekt task and the rootproject should have MergeTask only. Configuring DetektExtension on the root project is also erroneous.

The decision of creating a different gradle plugin is favoring a cleaner multi-module setup.

cortinico:
> * we should not expect any Detekt task to root module because of the issues reported in #3389, only subprojects should have Detekt task and the rootproject should have MergeTask only.

Yup you're right, thanks for clarifying this 👌

BraisGabin:
I think more or less like @cortinico here. Add an extra plugin seems to add too much complexity. Ideas:

- I know that in gradle you can add a plugin in the root and it adds taks in subprojects. Can we do the opposite? Add a task in the parent?
- Maybe, instead of providing this functionality in our own plugin we could create tasks and add them in our documentation. So anyone can copy&paste.

Right now I think that the second point is the best one. My reason is that `detektMergeXmlReport` doesn't have a deterministic output. It depends in your previous commands. And we can't make it deterministic because we have too much detekt tasks and we don't know which one the user wants to run.

If we implement #3364 this problem goes away. We could make `detektMergeXmlReport` dependOn `detektCheck`.

And each time that I think about it I'm more sure. detekt should output SARIF and then we should have other tasks to create the other reports using that file. And for this I think that just a task in the documentation is the best option, because we are not sure if we will keep that task in the future.

chao2zhang:
> * I know that in gradle you can add a plugin in the root and it adds taks in subprojects. Can we do the opposite? Add a task in the parent?

We can, but that does comprise the correctness:
- We only need to add it once for a given subproject, so we need to perform a check to see if this is already added to the root project.
- It is unclear which `DetektExtension` would be used to configure the behavior of the root project plugin either. There is one `detekt` extension in each subproject.
- This may also cause conflict with the plugins already at the root project.



> * Maybe, instead of providing this functionality in our own plugin we could create tasks and add them in our documentation. So anyone can copy&paste.

Yes, we can expose Gradle tasks themselves and add examples. This is a good idea.

chao2zhang:
I am going to close this PR and publish a new one to expose Gradle Tasks to merge XML + SARIF with provided groovy/kts scripts.

PR #3399, opened 2021-01-23
Auto generate CLI options in docs

chao2zhang:
There may be some text discrepancies between the existing `cli.md` and the annotation in `CliArgs.kt`.


docs/_data/sidebars/home_sidebar.yml
```diff
@@ -38,6 +38,9 @@ entries:
     - title: with the Command Line Interface
       url: /cli.html
       output: web
+    - title: Command Line Interface Options
+      url: /cli-options.html
+      output: web
```
BraisGabin, line 43: Should this be in getting started? I think that it fits better in user documentation.

BraisGabin (approved):
I like this! Is there a way in jekyll to include one markdown inside another? It could be great to keep this in the same document I think.

Anyway, we can merge this because the important thing here is to have a reliable documentation.

docs/_data/sidebars/home_sidebar.yml
chao2zhang: I have successfully used "include" in jekyll. So I will revert this change.

chao2zhang:
> I like this! Is there a way in jekyll to include one markdown inside another? It could be great to keep this in the same document I think.
> 
> Anyway, we can merge this because the important thing here is to have a reliable documentation.

The only limitation is that we cannot include files in the parent path `../` according to https://jekyllrb.com/docs/includes/#including-files-relative-to-another-file

BraisGabin (approved):
👏🏻👏🏻👏🏻

PR #3413, opened 2021-01-27
New rule: DontDowncastCollectionTypes

cortinico:
I'm proposing a new rule: `DontDowncastCollectionTypes`

Down-casting immutable types from kotlin.collections should be discouraged.
The result of the downcast is platform specific and can lead to unexpected crashes.
Prefer to use instead the `toMutable<Type>()` functions.

### Non-compliant
```kotlin
val list : List<Int> = getAList()
if (list is MutableList) {
    list.add(42)
}

(list as MutableList).add(42)
```

### Compliant 
```kotlin
val list : List<Int> = getAList()
list.toMutableList().add(42)
```

Happy to discuss if we want to adapt it to check casts against Java concrete collection types (e.g. `ArrayList`, etc.).

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -467,6 +467,8 @@ potential-bugs:
   active: true
   Deprecation:
     active: false
+  DontDowncastCollectionTypes:
+    active: false
   DuplicateCaseInWhenExpression:
     active: true
   EqualsAlwaysReturnsTrueOrFalse:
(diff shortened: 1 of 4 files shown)


cortinico:
This rule is inspired from Effective Kotlin: Item 1 - Limit Mutability (credits to @MarcinMoskala)

BraisGabin:
Great rule! I think that it would be good to think of ArrayList at least because it exists in kotlin (with a type alias).

And I'm not sure about the name. But I can't think of anyone better.

schalkms (approved):
Great rule!

cortinico:
> Great rule! I think that it would be good to think of ArrayList at least because it exists in kotlin (with a type alias).

I've added those 5 type aliases to the mutable types to consider: https://github.com/JetBrains/kotlin/blob/46b7a774b558001c136be225cf4367fa09ba1aee/libraries/stdlib/jvm/src/kotlin/collections/TypeAliases.kt#L13-L17

BraisGabin (approved):

PR #3419, opened 2021-01-27
Fix False Positive on `UnnecessarySafeCall`

### LINKED ISSUE #3409: False Positive UnnecessarySafeCalls re: Android Views (1.15.0)
## Expected Behavior
Elvis operators `?` for safe access of Android `View`s must not be falsely reported as unnecessary.

## Observed Behavior
False positive `UnnecessarySafeCall`s were reported when using `?` operator on nullable Android `View`s. For example:

``` Kotlin
viewBinding.viewStubProxy.viewStub?.layoutResource = R.layout.some_layout
```
Where `viewStub` refers to the following Java getter (return value marked as `@Nullable`):

``` Java
    @Nullable
    public ViewStub getViewStub() {
        return mViewStub;
    }
```

## Your Environment
* Version of detekt used: `1.15.0`
* Version of Gradle used (if applicable): `6.7.1`
* Version of Kotlin used: `1.3.72`
* Operating System and version: `macOS Catalina 10.15.7`

### LINKED ISSUE #3414: False positive for `UnnecessarySafeCall` on map accessor in kotlinx.serialization `JsonObject`.
## Expected Behavior
Accessing a map using the square braces syntax, followed by a safe accessor (`?.`) should not give a `UnnecessarySafeCall` warning. But, in the case of [kotlinx.serialization's `JsonObject`](https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-json/kotlinx-serialization-json/kotlinx.serialization.json/-json-object/index.html), it does.

## Observed Behavior/Steps to Reproduce
```
val json = Json { }
@Serializable
class World( val answer: Int )
val serialized = json.encodeToString( World( 42 ) )
val jsonElement = json.parseToJsonElement( serialized ).jsonObject
val answer = jsonElement[ World::answer.name ]?.jsonPrimitive?.content
```

This gives a warning on the safe accessor, whereas it should not.

P.s.[ the code samples of `UnnecessarySafeCall`](https://detekt.github.io/detekt/potential-bugs.html) seem incorrect. I think `someValue` was intended to be `a`.

## Your Environment
* Version of detekt used: 1.16.0-RC1
* Version of Gradle used (if applicable): 6.6
* Link to your project (if it's a public repository): https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.protocols.core/src/commonTest/kotlin/dk/cachet/carp/protocols/domain/devices/AltBeaconTest.kt#L52

cortinico:
This is a fix for the various false positives on `UnnecessarySafeCall`.

After some debugging the problems seems to be with misconfigured class-paths (e.g. https://github.com/cph-cachet/carp.core-kotlin/pull/223). If the classpath doesn't contain a third-party types, but still has a safe call operator, the Kotlin Compiler raises an `Errors.UNNECESSARY_SAFE_CALL` detection.

Inside the detection the compiler adds information on the non-nullable type that is the receiver of the safe call (in the `.a` field). In this patch I'm try to see if the compiler failed to resolve the type => If so, we can skip the report.

Fixes #3409 
Fixes #3414

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
@@ -8,11 +8,13 @@ import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
+import org.jetbrains.kotlin.diagnostics.DiagnosticWithParameters1
 import org.jetbrains.kotlin.diagnostics.Errors
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.KtSafeQualifiedExpression
 import org.jetbrains.kotlin.psi.psiUtil.getChildOfType
 import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.types.ErrorType
 
 /**
  * Reports unnecessary safe call operators (`?.`) that can be removed by the user.
@@ -39,6 +41,7 @@ class UnnecessarySafeCall(config: Config = Config.empty) : Rule(config) {
         Debt.FIVE_MINS
     )
 
+    @Suppress("ReturnCount")
     override fun visitSafeQualifiedExpression(expression: KtSafeQualifiedExpression) {
         super.visitSafeQualifiedExpression(expression)
 
@@ -49,8 +52,18 @@ class UnnecessarySafeCall(config: Config = Config.empty) : Rule(config) {
             return
         }
 
-        val compilerReports = bindingContext.diagnostics.forElement(safeAccessElement)
-        if (compilerReports.any { it.factory == Errors.UNNECESSARY_SAFE_CALL }) {
+        val compilerReport = bindingContext
+            .diagnostics
+            .forElement(safeAccessElement)
+            .firstOrNull { it.factory == Errors.UNNECESSARY_SAFE_CALL }
+
+        if (compilerReport != null) {
+            // For external types, if they're not included in the classpath, we still get an Errors.UNNECESSARY_SAFE_CALL.
+            // This causes false positives if our users are misconfiguring detekt with Type Resolution.
+            // Here we try to check if the compiler reports failed to resolve the nullable type.
+            if (compilerReport is DiagnosticWithParameters1<*, *> && compilerReport.a is ErrorType) {
+                return
+            }
             report(
                 CodeSmell(
                     issue, Entity.from(expression), "${expression.text} contains an unnecessary " +
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
@@ -61,6 +61,11 @@ class UnnecessarySafeCall(config: Config = Config.empty) : Rule(config) {
             // For external types, if they're not included in the classpath, we still get an Errors.UNNECESSARY_SAFE_CALL.
             // This causes false positives if our users are misconfiguring detekt with Type Resolution.
             // Here we try to check if the compiler reports failed to resolve the nullable type.
+
+            // More reference on where the compiler is attaching this information is here:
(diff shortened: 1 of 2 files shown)


BraisGabin (commented):
Is there any way to create a unit test?

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
```diff
@@ -49,8 +52,18 @@ class UnnecessarySafeCall(config: Config = Config.empty) : Rule(config) {
...
+            .forElement(safeAccessElement)
+            .firstOrNull { it.factory == Errors.UNNECESSARY_SAFE_CALL }
+
+        if (compilerReport != null) {
+            // For external types, if they're not included in the classpath, we still get an Errors.UNNECESSARY_SAFE_CALL.
+            // This causes false positives if our users are misconfiguring detekt with Type Resolution.
+            // Here we try to check if the compiler reports failed to resolve the nullable type.
+            if (compilerReport is DiagnosticWithParameters1<*, *> && compilerReport.a is ErrorType) {
+                return
+            }
```
schalkms, line 66: For me this is still strange. Why is this popping up only after updating to the newest compiler version?

schalkms (approved):

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
```diff
@@ -49,8 +52,18 @@ class UnnecessarySafeCall(config: Config = Config.empty) : Rule(config) {
             return
         }
 
-        val compilerReports = bindingContext.diagnostics.forElement(safeAccessElement)
-        if (compilerReports.any { it.factory == Errors.UNNECESSARY_SAFE_CALL }) {
+        val compilerReport = bindingContext
+            .diagnostics
+            .forElement(safeAccessElement)
+            .firstOrNull { it.factory == Errors.UNNECESSARY_SAFE_CALL }
```
chao2zhang, line 58: Should we still keep `any` instead of `firstOrNull`? I think the code depends on the order for the diagnostics, which seems too fragile.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
chao2zhang: We should be able to find the exact `DiagnosticWithParameters1 ` definition of the errors being reported and add this to the documentation.

cortinico: > For me this is still strange. Why is this popping up only after updating to the newest compiler version?

I believe this is related to us enabling the rule by default:
https://github.com/detekt/detekt/pull/3419

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
cortinico: > Should we still keep `any` instead of `firstOrNull`?

I could change it to a `forEach` as we need to get access to the `compilerReport`

cortinico:
> Is there any way to create a unit test?

Found a test case for it 👌  The only drawback is that we can't compile the test snippet as the import will break the compilation.

chao2zhang (approved):

BraisGabin (approved):

cortinico:
Turns out the Kotlin compiler was actually wrong here. I've pushed a fix upstream https://github.com/JetBrains/kotlin/pull/4072 so we can revert our patch as soon as Kotlin 1.5 is out 🥳

BraisGabin:
How can we be sure that we don't forget? Create the pr with a blocked label and merge it as soon as we move to kotlin 1.5?

schalkms:
> Turns out the Kotlin compiler was actually wrong here. I've pushed a fix upstream JetBrains/kotlin#4072 so we can revert our patch as soon as Kotlin 1.5 is out 🥳

That's really interesting, so this suspicion was right.
https://github.com/detekt/detekt/issues/3409#issuecomment-767824578

> How can we be sure that we don't forget? Create the pr with a blocked label and merge it as soon as we move to kotlin 1.5?

I'd suggest this approach.

cortinico:
> How can we be sure that we don't forget? Create the pr with a blocked label and merge it as soon as we move to kotlin 1.5?

Created here https://github.com/detekt/detekt/pull/3439

Issue #2692, opened 2020-05-14
Detekt's 2.0 cli


BraisGabin:
The same as #2680 but for talk about breaking changes that we could like to have in the cli once Dekte 2.0 arrives. The same as before, there's nothing wrong in 1.0. We don't need a breaking change in the inmediate future but it's good to talk about ideas.

My ideas:

I think that we should simplify our config file. It should only contain the configuration of the rules. And maybe the `config` about the exclusion... But I'm not sure about that last part. 

The other things could be moved out. For example `maxIssues` by default is 0. To change it we could use a flag like `--max-issues=10` or `--max-issues=-1`

`console-reports` doesn't seems like a good thing. If they are reports we should allow the user to have them in other formats to process them later. And if they are not "really" reports we should hide them behind a `verbose` mode.

And the processors that we support right now are two types: show progress and show counts. I think that the counts should be displayed in some reports and print them in the `versbose` mode. And the progress the same, behind the `verbose` mode or drop it. Or maybe we could rethink it and print the name of the file that we are processing. Somthing like `tar` works with `-v`.

And now that I'm thinking it... probably we should move the configuration file outside the `cli` because we want to use the same configuration file for the `gradle-plugin`. Same thing for the `baseline`.

arturbosch:
https://github.com/detekt/detekt/issues/2680#issuecomment-657709706

I agree with moving out the build and processor, console report stuff out of the config.
Introducing a `--max-issues` flag would be non-breaking and will be supported with the tooling-api (https://github.com/detekt/detekt/pull/2861/files#diff-2aa2bc358229c132fb7a02ff5a800b3cR12).

The defaut config and the baseline feature is already part of core :).

arturbosch:
from https://github.com/detekt/detekt/issues/2267#issuecomment-766433748

remove --fail-fast
activate --build-upon-default-config by default (remove flag then)

PR #3431, opened 2021-01-31
Add allRules and deprecate failFast in gradle tasks

chao2zhang:
This is a follow-up of #3391 and should be merged as early as possible to ensure a smoother transition to the new Detekt 2.0 CLI #2692

### DIFF
--- a/detekt-gradle-plugin/src/intTest/kotlin/io/gitlab/arturbosch/detekt/DetektTaskDslTest.kt
+++ b/detekt-gradle-plugin/src/intTest/kotlin/io/gitlab/arturbosch/detekt/DetektTaskDslTest.kt
@@ -397,7 +397,7 @@ internal object DetektTaskDslTest : Spek({
                         |    debug = true
                         |    parallel = true
                         |    disableDefaultRuleSets = true
-                        |    failFast = true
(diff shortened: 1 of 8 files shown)


detekt-gradle-plugin/src/intTest/kotlin/io/gitlab/arturbosch/detekt/DetektTaskDslTest.kt
```diff
@@ -397,7 +397,7 @@ internal object DetektTaskDslTest : Spek({
                         |    debug = true
                         |    parallel = true
                         |    disableDefaultRuleSets = true
-                        |    failFast = true
+                        |    allRules = true
```
BraisGabin, line 400: should we remove the failFast tests? Can't we keep it so we don't break it until we remove it?

chao2zhang: We could keep both.

BraisGabin (approved):

cortinico (approved):

schalkms (approved):

PR #3439, opened 2021-02-02
Remove UnnecessarySafeCall safeguard against ErrorType

cortinico:
Opening this to recall to remove this extra patch introduced in #3419
A fix was pushed upstream in the Kotlin compiler and should land in Kotlin 1.5

As soon as we bump the `compiler-embeddable` version we can also merge this PR (as the tests should be green)

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
@@ -10,13 +10,11 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
 import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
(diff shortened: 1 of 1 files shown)


BraisGabin (approved):

BraisGabin:
We should be able to unblock this now, right?

cortinico:
Seems like the change didn't make to `1.5.0`:
https://github.com/JetBrains/kotlin/blob/e73e039e43d740b890eaa4d7256362eeaf6ff1f7/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/CallExpressionResolver.kt#L542-L544

But is on `master`:
https://github.com/JetBrains/kotlin/blob/23faacd21b84af32921ecad785c8d30f1b930c34/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/CallExpressionResolver.kt#L545-L547

So we will probably have to wait a bit more

chao2zhang:
Kotlin compiler fix is included in [1.5.20-M1](https://github.com/JetBrains/kotlin/blob/v1.5.20-M1/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/CallExpressionResolver.kt), so we should be unblocked by then.

chao2zhang:
Ooops sorry I misclicked the button.

3flex:
I checked out this branch, rebased on main & ran a full build, so this just needs a rebase to check the CI passes on all platforms then it can be merged for 1.18.1 or 1.19.0.

cortinico:
> so this just needs a rebase to check the CI passes on all platforms

Just rebased 👍

Issue #3436, opened 2021-02-01
Exceptions - Add rule that prohibits an object (Kotlin singleton) from extending any type of Throwable


mhernand40:
## Expected Behavior of the rule
Kotlin `objects` should _NOT_ be allowed to extend any type of `Throwable`.

**Noncompliant Code:**
```Kotlin
object FooException : RuntimeException()
```

**Compliant Code:**
```Kotlin
class FooException : RuntimeException()
```

## Context
`objects` in Kotlin are native singletons and remain in memory for the duration of the process. `Throwables`/`Exceptions` on the other hand should typically be instantiated on a case-by-case basis and thrown accordingly. `Throwables` are stateful and come packed with lots of information in their stack traces and may even reference another `Throwable` as the underlying `cause` and/or multiple `Throwables` in a mutable `List` of `suppressedExceptions`. By making an `object` extend a `Throwable` type, it automatically leads to stateful, thread unsafe singletons and it encourages reuse of global and stateful exceptions which can be thrown and referenced from various pleases from different threads. In my opinion, this would be a code smell.

cortinico:
> In my opinion, this would be a code smell.

I generally agree with this rule. However do you have some more concrete examples? Like in which scenario you feel that someone would use a object to represent an `Exception`?

I just wanted to make sure that, if we add a rule to detekt, it serves some real-word scenarios

mhernand40:
> > In my opinion, this would be a code smell.
> 
> I generally agree with this rule. However do you have some more concrete examples? Like in which scenario you feel that someone would use a object to represent an `Exception`?
> 
> I just wanted to make sure that, if we add a rule to detekt, it serves some real-word scenarios

In the code base I currently work out of, I have seen several developer make this mistake and we currently have a decent number of violations, ~20. For some reason I am seeing that devs tend to favor `object` over `class`, which may be a concern in and of itself, but I find it even more problematic when it comes to custom exceptions. Some hypothetical examples that are similar to code I have seen are as follows:
```Kotlin
object BanException : RuntimeException(…)
object AuthException : RuntimeException(…)
```
as well as
```Kotlin
sealed class DomainSpecificException : IllegalStateException() {
    data class FooException(val prop1: Prop1, val prop2: Prop2) : DomainSpecificException
    object BarException : DomainSpecificException("Bar exception") // <-- Violation
    object BazException : DomainSpecificException("Baz exception") // <-- Violation
}
```

[...]

PR #3443, opened 2021-02-04
Add new ObjectExtendsThrowable rule

mhernand40:
An Exceptions rule that reports when an `object` extend any type of `Throwable`. Although using `object` may look innocent at times, having one that extends a `Throwable`/`Exception` inadvertently introduces a global singleton exception instance. Any custom exception should be made instantiable by being defined as a regular `class`.

Addresses https://github.com/detekt/detekt/issues/3436.

### DIFF
--- a/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
+++ b/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
@@ -53,20 +53,17 @@ class ObjectExtendsThrowable(config: Config = Config.empty) : Rule(config) {
     )
 
     override fun visitObjectDeclaration(declaration: KtObjectDeclaration) {
-        declaration
-            .takeUnless { bindingContext == BindingContext.EMPTY }
-            ?.takeUnless { it.isObjectLiteral() }
-            ?.takeIf { it.isSubtypeOfThrowable() }
-            ?.also {
-                report(
-                    CodeSmell(
-                        issue = issue,
-                        entity = Entity.from(element = it),
-                        message = issue.description
-                    )
-                )
-            }
         super.visitObjectDeclaration(declaration)
+        if (bindingContext == BindingContext.EMPTY) return
+        if (!declaration.isObjectLiteral() && declaration.isSubtypeOfThrowable()) {
+            report(
+                CodeSmell(
+                    issue = issue,
+                    entity = Entity.from(element = declaration),
+                    message = issue.description
+                )
+            )
+        }
     }
 
     private fun KtObjectDeclaration.isSubtypeOfThrowable(): Boolean {
--- a/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
+++ b/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
@@ -1,0 +1,79 @@
+package io.gitlab.arturbosch.detekt.rules.exceptions
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.psi.KtObjectDeclaration
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.types.typeUtil.isNotNullThrowable
+import org.jetbrains.kotlin.types.typeUtil.supertypes
+
+/**
+ * This rule reports all `objects` including `companion objects` that extend any type of
+ * `Throwable`. Throwable instances are not intended for reuse as they are stateful and contain
+ * mutable information about a specific exception or error. Hence, global singleton `Throwables`
+ * should be avoided.
+ *
+ * See https://kotlinlang.org/docs/reference/object-declarations.html#object-declarations
+ * See https://kotlinlang.org/docs/reference/object-declarations.html#companion-objects
+ *
+ * <noncompliant>
+ * object InvalidCredentialsException : Throwable()
+ *
+ * object BanException : Exception()
+ *
+ * object AuthException : RuntimeException()
+ * </noncompliant>
+ *
+ * <compliant>
+ * class InvalidCredentialsException : Throwable()
+ *
+ * class BanException : Exception()
+ *
+ * class AuthException : RuntimeException()
+ * </compliant>
+ *
+ * @requiresTypeResolution
+ */
+class ObjectExtendsThrowable(config: Config = Config.empty) : Rule(config) {
+    override val issue = Issue(
+        id = "ObjectExtendsThrowable",
+        severity = Severity.CodeSmell,
+        description = "An `object` should not extend and type of Throwable. Throwables are " +
+                "stateful and should be instantiated only when needed for when a specific error " +
+                "occurs. An `object`, being a singleton, that extends any type of Throwable " +
+                "consequently introduces a global singleton exception whose instance may be " +
+                "inadvertently reused from multiple places, thus introducing shared mutable " +
+                "state.",
+        debt = Debt.TEN_MINS
+    )
+
+    override fun visitObjectDeclaration(declaration: KtObjectDeclaration) {
+        declaration
+            .takeUnless { bindingContext == BindingContext.EMPTY }
+            ?.takeUnless { it.isObjectLiteral() }
+            ?.takeIf { it.isSubtypeOfThrowable() }
+            ?.also {
+                report(
+                    CodeSmell(
+                        issue = issue,
+                        entity = Entity.from(element = it),
+                        message = issue.description
+                    )
+                )
+            }
+        super.visitObjectDeclaration(declaration)
+    }
+
+    private fun KtObjectDeclaration.isSubtypeOfThrowable(): Boolean {
+        return bindingContext[BindingContext.CLASS, this]
+            ?.defaultType
+            ?.supertypes()
+            .orEmpty()
+            .any { it.isNotNullThrowable() }
+    }
+}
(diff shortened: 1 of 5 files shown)


detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
```diff
@@ -0,0 +1,79 @@
...
+                "inadvertently reused from multiple places, thus introducing shared mutable " +
+                "state.",
+        debt = Debt.TEN_MINS
+    )
+
+    override fun visitObjectDeclaration(declaration: KtObjectDeclaration) {
+        declaration
+            .takeUnless { bindingContext == BindingContext.EMPTY }
+            ?.takeUnless { it.isObjectLiteral() }
+            ?.takeIf { it.isSubtypeOfThrowable() }
```
chao2zhang, line 59: Do you think a simple `if` with boolean operators will be clearer? Consecutive `takeUnless` makes it quite difficult to read.

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
```diff
@@ -0,0 +1,79 @@
...
+                    )
+                )
+            }
+        super.visitObjectDeclaration(declaration)
+    }
+
+    private fun KtObjectDeclaration.isSubtypeOfThrowable(): Boolean {
+        return bindingContext[BindingContext.CLASS, this]
+            ?.defaultType
+            ?.supertypes()
```
chao2zhang, line 75: What do you think if we should report `object SomeException : Throwable?`?

detekt-rules-exceptions/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowableSpec.kt
```diff
@@ -0,0 +1,127 @@
...
+
+    describe("ObjectExtendsThrowable rule") {
+
+        it("reports objects that extend Throwable") {
+            val code = """
+            object BanException : Throwable()
+            object AuthException : RuntimeException()
+            object ReportedException : Exception()
+            object FatalException : Error()
+            object ObjectCustomException : CustomException("singleton custom exception")
```
chao2zhang, line 24: Thank you for the exhaustive test cases!

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
mhernand40: Sure, I can address that. I actually have a question. When overriding `visitObjectDeclaration`, is it important that we call `super.visitObjectDeclaration()`? I see other examples that seem to always call it towards the end of the function rather than at the very top, which is why I tried to avoid doing any sort of early `returns`.

chao2zhang: It probably does not matter since each rule inherits `KtTreeVisitorVoid`. So I would probably lean towards stay consistent with the existing code patterns.

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
```diff
@@ -0,0 +1,79 @@
...
+        declaration
+            .takeUnless { bindingContext == BindingContext.EMPTY }
+            ?.takeUnless { it.isObjectLiteral() }
+            ?.takeIf { it.isSubtypeOfThrowable() }
+            ?.also {
+                report(
+                    CodeSmell(
+                        issue = issue,
+                        entity = Entity.from(element = it),
+                        message = issue.description
```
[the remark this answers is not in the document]
mhernand40: Thanks for the suggestion. I was trying to maintain consistency with some of the existing exception rules, such as:
* [InstanceOfCheckForException](https://github.com/detekt/detekt/blob/master/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/InstanceOfCheckForException.kt#L57)
* [NotImplementedDeclaration](https://github.com/detekt/detekt/blob/master/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/NotImplementedDeclaration.kt#L42)
* [RethrowCaughtException](https://github.com/detekt/detekt/blob/master/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/RethrowCaughtException.kt#L62)
* [ThrowingExceptionInMain](https://github.com/detekt/detekt/blob/master/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ThrowingExceptionInMain.kt#L33)

These are just a few that reference `issue.description` when setting the `message` parameter.

Are we satisfied with the current issue description or do we feel that even the issue description could use some shortening? If we make it shorter, do we still want a separate `CodeSmell` `message`? Thanks again.

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
mhernand40: This would not compile. The `Throwable` being extended must have its constructor invoked since it is a `class`; not an `interface`. And by invoking the constructor, we are inferring a non-null `Throwable`.

Screen Shot 2021-02-04 at 11 28 46 AM

Screen Shot 2021-02-04 at 11 29 18 AM

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
```diff
@@ -0,0 +1,79 @@
...
+                "occurs. An `object`, being a singleton, that extends any type of Throwable " +
+                "consequently introduces a global singleton exception whose instance may be " +
+                "inadvertently reused from multiple places, thus introducing shared mutable " +
+                "state.",
+        debt = Debt.TEN_MINS
+    )
+
+    override fun visitObjectDeclaration(declaration: KtObjectDeclaration) {
+        declaration
+            .takeUnless { bindingContext == BindingContext.EMPTY }
```
cortinico, line 57: Please replace with:
```
```

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
cortinico: Please make sure you call `super.visitObjectDeclaration(declaration)` as the first statement of your function. 

It won't have any direct effect, but that's implicitly breaking the visitor patter (and we should fix the other occurrences where this happens, but that's outside the scope of this PR).

detekt-rules-exceptions/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowableSpec.kt
```diff
@@ -0,0 +1,127 @@
...
+
+            sealed class DomainException : RuntimeException() {
+                data class Exception1(val prop1: String, val prop2: Boolean) : DomainException()
+                class Exception2 : DomainException()
+                object Exception3 : DomainException()
+            }
+
+            open class CustomException(message: String) : RuntimeException(message)
+            """
+            assertThat(subject.compileAndLintWithContext(env, code)).hasSize(6)
```
cortinico, line 34: As rule of thumb, this should be split in 3 tests as you're testing multiple scenarios together. If the test fails for some reason, it will be harder for the maintainer to understand the exact reason.

If you wish to keep bigger tests, please assert on the source location (with `.hasSourceLocation`) so at least the test is more exhaustive.

cortinico (approved):
Great work 👍 Left a couple of nits

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
mhernand40: 8e09600

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
mhernand40: 8e09600

detekt-rules-exceptions/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowableSpec.kt
mhernand40: 144daa3

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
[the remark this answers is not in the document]
mhernand40: I have not forgotten about this. Just awaiting further feedback. 🙂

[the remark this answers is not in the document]
BraisGabin: I do agree with chao here. One is the desciprtion of the rule and the oher is the description of one particular smell. I think that a description like: "XXX should be a class instead of an object because it extends from YYY" is enought for the message.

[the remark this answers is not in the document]
BraisGabin: or "from a throwable class"... or something like this. I'm not too good with this things :P

BraisGabin (approved):

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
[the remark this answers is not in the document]
mhernand40: Thanks everyone. Let me know if this works b6197eb.

cortinico (approved):

Issue #3449, opened 2021-02-06
False positives for `MultilineLambdaItParameter`


chao2zhang:
## Expected Behavior
No issues detected for the code below.

## Observed Behavior
Current a violation of `MultilineLambdaItParameter` is issued for the code below. Because the lambda does not have a parameter at all.

## Steps to Reproduce
Run detekt with active `MultilineLambdaItParameter`.

## Context
```
    with(sharedPreferences.edit()) {
        if (me != null) {
            putString(KEY_ME, JSONObjectGenerator.toJSONObject(me).toString())
        } else {
            remove(KEY_ME)
        }
        apply()
    }
```

## Your Environment

* Version of detekt used: 1.16.0-RC1
* Operating System and version: macOS

PR #3451, opened 2021-02-07
Fix false positives for MultilineLambdaItParameter.kt

chao2zhang:
This addresses #3449.

Notes:
- This rule now requires type-resolution to run in full mode. Without type-resolution, this rule would check only if `it` is used as explicit parameters.
- I have updated the tests to cover the scenario with type-resolution and without type-resolution.


BraisGabin:
Do we want more mixed rules? This one is kind of new. Shouldn't we just turn it in a pure type solving rule? The problem with mixed rules are the user expectations. Detekt seems buggy because of the mixed ones.

chao2zhang:
> Do we want more mixed rules? This one is kind of new. Shouldn't we just turn it in a pure type solving rule? The problem with mixed rules are the user expectations. Detekt seems buggy because of the mixed ones.

I remember reading some discussions but couldn't remember the resolution. I can change it to a full type-resolving rule.

cortinico:
> I can change it to a full type-resolving rule.

+1. We discussed about it in #2994

BraisGabin (approved):

cortinico (approved):

PR #3478, opened 2021-02-16
Add UnreachableCatchBlock rule

### LINKED ISSUE #305: Detect unreachable catch blocks
Code like that below compiles fine but only first catch block can be ever reached:
```kotlin
try {
	//whatever
} catch (t: Throwable) {

} catch (e: Exception) {

} catch (e: Exception) {

}
```
I can try to make PR for that but want to ask first what do you think. Should `UnreachableCode` rule cover such cases or should it be a separate one?

t-kameyama:
Fixes #305

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnreachableCatchBlock.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnreachableCatchBlock.kt
@@ -17,15 +17,32 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.isSubclassOf
 
 /**
  * Reports unreachable catch blocks.
+ * Catch blocks can be unreachable if the exception has already been caught in the block above.
  *
  * <noncompliant>
  * fun test() {
  *     try {
+ *         foo()
  *     } catch (t: Throwable) {
- *     } catch (e: Exception) { // unreachable
+ *         bar()
+ *     } catch (e: Exception) {
+ *         // Unreachable
+ *         baz()
  *     }
  * }
  * </noncompliant>
+ *
+ * <compliant>
+ * fun test() {
+ *     try {
+ *         foo()
+ *     } catch (e: Exception) {
+ *         baz()
+ *     } catch (t: Throwable) {
+ *         bar()
+ *     }
+ * }
+ * </compliant>
  *
  * @requiresTypeResolution
  */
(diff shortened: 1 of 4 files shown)


detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnreachableCatchBlock.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.jetbrains.kotlin.psi.KtCatchClause
+import org.jetbrains.kotlin.psi.KtTryExpression
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.descriptorUtil.isSubclassOf
+
+/**
+ * Reports unreachable catch blocks.
+ *
+ * <noncompliant>
```
schalkms, line 21: It might be good to add an extended description and compliant code example in order to have a better understand of the rule from the user point of view.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnreachableCatchBlock.kt
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+    @Suppress("ReturnCount")
+    override fun visitCatchSection(catchClause: KtCatchClause) {
+        super.visitCatchSection(catchClause)
+        if (bindingContext == BindingContext.EMPTY) return
+
+        val tryExpression = catchClause.getStrictParentOfType<KtTryExpression>() ?: return
+        val prevCatchClauses = tryExpression.catchClauses.takeWhile { it != catchClause }
+        if (prevCatchClauses.isEmpty()) return
```
schalkms, line 47: Is this `if empty` check even necessary considering the `any` statement underneath?

schalkms (approved):

schalkms:
I like this rule 👍 
Thank you!

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnreachableCatchBlock.kt
t-kameyama: I'd like to avoid unnecessary `catchClassDescriptor()` call in the following case.
```
try {
} catch (e: Exception) {
}
```

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnreachableCatchBlock.kt
t-kameyama: 0cb50f316dc9b9b32d50175d08dd37f84553f18c

BraisGabin (approved):
nice one!

PR #3486, opened 2021-02-20
Allow using regular expressions when defining license header templates

serso:
Problem: different projects might impose different rules on license headers.
It's not uncommon, for example, that a year in the license header is the year when the source
file was created. This leads to a situation when different source files have different license
headers (at least some parts of them differ) making it impossible to use a single static
license header template to match against.

This change makes it possible to define license header templates as regexps.
An example of such template is:
  \/\/ Copyright 20[0-9]{2} Artur Bosch & Contributors
which would match both
  \\ Copyright 2020 Artur Bosch & Contributors
and
  \\ Copyright 2021 Artur Bosch & Contributors

The way `AbsentOrWrongFileLicense` works now is controlled
by the new configuration option `licenseTemplateIsRegex`:
when `licenseTemplateIsRegex` is false the template is considered to be static
and the old matching rules apply.
If `licenseTemplateIsRegex` is set to true the new regexp based logic kicks in.

To avoid breakage of existing projects, the added configuration option
`licenseTemplateIsRegex` is disabled by default.

This commit implements a feature requested here #3100.

For more information, see the [CONTRIBUTING guide](https://github.com/detekt/detekt/blob/master/.github/CONTRIBUTING.md).

### DIFF
--- a/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/LicenceHeaderLoaderExtension.kt
+++ b/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/LicenceHeaderLoaderExtension.kt
@@ -6,7 +6,9 @@ import io.gitlab.arturbosch.detekt.api.SetupContext
 import io.gitlab.arturbosch.detekt.api.SingleAssign
 import io.gitlab.arturbosch.detekt.api.UnstableApi
 import io.gitlab.arturbosch.detekt.rules.documentation.AbsentOrWrongFileLicense.Companion.DEFAULT_LICENSE_TEMPLATE_FILE
+import io.gitlab.arturbosch.detekt.rules.documentation.AbsentOrWrongFileLicense.Companion.DEFAULT_LICENSE_TEMPLATE_IS_REGEX
 import io.gitlab.arturbosch.detekt.rules.documentation.AbsentOrWrongFileLicense.Companion.PARAM_LICENSE_TEMPLATE_FILE
+import io.gitlab.arturbosch.detekt.rules.documentation.AbsentOrWrongFileLicense.Companion.PARAM_LICENSE_TEMPLATE_IS_REGEX
 import io.gitlab.arturbosch.detekt.rules.documentation.AbsentOrWrongFileLicense.Companion.RULE_NAME
 import org.jetbrains.kotlin.com.intellij.openapi.util.Key
 import org.jetbrains.kotlin.com.intellij.openapi.util.text.StringUtilRt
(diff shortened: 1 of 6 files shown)


BraisGabin (approved):

serso:
@BraisGabin @arturbosch can you please advice me what I should do with the test coverage? Codecov marks red the lines that I haven't touched. Or maybe I'm reading its report wrong?

chao2zhang:
I would say let's ignore the `codecov/path` failure. Since the `codecov/project` is increased, I am doubtful if the `codecov/patch` not hitting 80% is mathematically correct.

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/LicenceHeaderLoaderExtension.kt
```diff
@@ -56,24 +62,35 @@ class LicenceHeaderLoaderExtension : FileProcessListener {
...
         }
 
-        fun cacheLicence(dir: Path) {
+        fun cacheLicence(dir: Path, isRegexTemplate: Boolean) {
             val licenceHeader = loadLicence(dir)
+            if (!isRegexTemplate) {
+                for (file in files) {
+                    file.putUserData(LICENCE_KEY, licenceHeader)
+                }
+                return
```
schalkms, line 71: With this return the code path underneath, which previously was executed, isn't executed anymore, if `isRegexTemplate` is `false`. Is there are reason, why this isn't necessary anymore?

schalkms (commented):
Thanks for contributing this rule! Great job! 😀
I have 1 remaining question. I think this ready to be merged.

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/LicenceHeaderLoaderExtension.kt
serso: > Is there are reason, why this isn't necessary anymore?

Are you referring to `if (configPath != null && shouldRuleRun())...`? If yes then this code still runs as it is located in the enclosing (outer) function:
```
fun onStart(...) {
  fun shouldRuleRun() { ... }
  fun getPathTemplate() { ... }
  fun cacheLicence() { 
    // Code here returns earlier if `isRegexTemplate = false`
  }
  
  // Method body
  if (configPath != null && shouldRuleRun()) { ... }
}
```

I'm not a fan of the nested (inner) functions myself but I decided to leave it as it was before assuming it's a common practice in this project.

serso: This is how the code looked before:
```kotlin
        fun cacheLicence(dir: Path) {
            val licenceHeader = loadLicence(dir)
            for (file in files) {
                file.putUserData(LICENCE_KEY, licenceHeader)
            }
        }
```

This is how the code looks now:
```kotlin
        fun cacheLicence(dir: Path, isRegexTemplate: Boolean) {
            val licenceHeader = loadLicence(dir)
            if (!isRegexTemplate) {
                for (file in files) {
                    file.putUserData(LICENCE_KEY, licenceHeader)
                }
                return
            }
            val licenseHeaderRegex = licenceHeader.toRegex(RegexOption.MULTILINE)
            for (file in files) {
                file.putUserData(LICENCE_REGEX_KEY, licenseHeaderRegex)
            }
        }
```

schalkms: Sorry, my bad. I overlooked this in the diff viewer. Everything's fine.
RESOLVED

schalkms (approved):

PR #3487, opened 2021-02-20
Refactor gradle integration test

chao2zhang:
This is mostly meant to reorganize the integration tests as I am adding integration tests for Detekt Android.
- Changed `detektConfig` in `Submodule` to `buildFileContent` as I am going to apply different android build logic for each submodule.
- Remove `ProjectLayout` as a data class and it feels inefficient to create a new `ProjectLayout` each time when calling `withSubmodule`


detekt-gradle-plugin/src/intTest/kotlin/io/gitlab/arturbosch/detekt/testkit/ProjectLayout.kt
```diff
@@ -1,29 +1,29 @@
...
+class ProjectLayout(
     val numberOfSourceFilesInRootPerSourceDir: Int,
     val numberOfCodeSmellsInRootPerSourceDir: Int = 0,
-    val submodules: List<Submodule> = emptyList(),
     val srcDirs: List<String> = listOf("src/main/java")
 ) {
 
-    fun withSubmodule(
+    private val mutableSubmodules: MutableList<Submodule> = mutableListOf()
+    val submodules: List<Submodule>
```
schalkms, line 10: Is this efficient to store both the mutable and immutable version in a member variable? Can we just store the mutable version and return it publicly as immutable?

schalkms (approved):

detekt-gradle-plugin/src/intTest/kotlin/io/gitlab/arturbosch/detekt/testkit/ProjectLayout.kt
chao2zhang: This is the exact way to implement backing fields https://kotlinlang.org/docs/properties.html#backing-properties And in the decompiled bytecode there is only one field
```
public final class ProjectLayout {
   private final List mutableSubmodules;
   private final int numberOfSourceFilesInRootPerSourceDir;
   private final int numberOfCodeSmellsInRootPerSourceDir;
   @NotNull
   private final List srcDirs;

   @NotNull
   public final List getSubmodules() {
      return this.mutableSubmodules;
   }
```

PR #3453, opened 2021-02-07
Add support to Kotlin Multiplatform Projects

cortinico:
This PR adds support to generate detekt tasks for Kotlin Multiplatform projects. Specifically, tasks for `jvm` and `android` targets have Type Resolution enabled, while `js` and `native` don't (as I still haven't found a way to properly invoke the compiler with the correct classpath/dependency).

For a KMM project like this one:
```
kotlin {
    android()
    ios()
    jvm("jvmDesktop")
    jvm("jvmBackend")
    js()
}
```

The following tasks will be added to the graph:

```
detektAndroidDebug - EXPERIMENTAL: Run detekt analysis for target android and source set debug with type resolution.
detektAndroidDebugAndroidTest - EXPERIMENTAL: Run detekt analysis for target android and source set debugAndroidTest with type resolution.
detektAndroidDebugUnitTest - EXPERIMENTAL: Run detekt analysis for target android and source set debugUnitTest with type resolution.
detektAndroidRelease - EXPERIMENTAL: Run detekt analysis for target android and source set release with type resolution.
detektAndroidReleaseUnitTest - EXPERIMENTAL: Run detekt analysis for target android and source set releaseUnitTest with type resolution.
detektIosArm64Main - Run detekt analysis for target iosArm64 and source set main
detektIosArm64Test - Run detekt analysis for target iosArm64 and source set test
detektIosX64Main - Run detekt analysis for target iosX64 and source set main
detektIosX64Test - Run detekt analysis for target iosX64 and source set test
detektJsMain - Run detekt analysis for target js and source set main
detektJsTest - Run detekt analysis for target js and source set test
detektJvmBackendMain - EXPERIMENTAL: Run detekt analysis for target jvmBackend and source set main with type resolution.
detektJvmBackendTest - EXPERIMENTAL: Run detekt analysis for target jvmBackend and source set test with type resolution.
detektJvmDesktopMain - EXPERIMENTAL: Run detekt analysis for target jvmDesktop and source set main with type resolution.
detektJvmDesktopTest - EXPERIMENTAL: Run detekt analysis for target jvmDesktop and source set test with type resolution.
```

Alongside `detektBaseline*` tasks for each of them.

I still haven't managed to write documentation for this, since I'd like to kickoff the discussion to understand if this is going in the right direction.

Also with the KMM ecosystem growing quickly, I believe this could be a great addition to the ecosystem.

Related to #3414 
Closes #3385

PR #3489, opened 2021-02-21
Refactor Gradle integration tests

chao2zhang:
This PR serves two purposes:
- Move all Gradle integration tests to use the Gradle testkit API. There is no more unit test depending on Gradle internal API itself. 
  - One benefit is less maintenance over the future Gradle upgrade
  - Another benefit is that after #3356, IntelliJ is not really working with the new `intTest` source set and we cannot launch those tests in IDE. IDEA-218815 is an issue that's applicable to a lot of people in the recent versions. So I am planning to remove the `intTest` and merge it back to `test`.
- Set up testing infrastructure for #3327 and I did verify that the RFC in #3334 is actually not working :)

### Note
While this mostly maintains the test coverage, I did reduce the following test cases:
- Test for `com.android.test` is removed. Since there are many other plugin's like `com.android.dynamic-feature`, `com.android.lint` and `com.android.reporting`, it is not necessary to test all different android plugins. As long as the base plugin `com.android.build.gradle.api.AndroidBasePlugin` is applied, we do not need to worry too much about other subplugins.
- Removed one flavor dimension `sally` from `name`. This does not provide too much value in my opinion while contributes significantly the lines of code in tests.


BraisGabin (approved):
👏🏻👏🏻 Nice refactor. I had no idea that detektMain was running the android tasks too. Nor the ignored configuration. I have work to do Monday morning at work 😅.

BraisGabin (approved):

cortinico (approved):
LGTM. 
Just one note, should we merge #3453 before this?

chao2zhang:
> LGTM.
> Just one note, should we merge #3453 before this?

Good call. Now we have one remaining non-testkit gradle unit test

PR #3497, opened 2021-02-24
New rule: disallow to cast to nullable type

### LINKED ISSUE #3479: New rule: disallow to cast to nullable type
noncompliant:
```kotlin
fun foo(a: Any?) {
    val x: String? = a as String?
    // If 'a' is not String, ClassCastException will be thrown.
}
```

compliant:
```kotlin
fun foo(a: Any?) {
    val x: String? = a as? String
}
```

t-kameyama:
Fixes #3479

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastToNullableType.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastToNullableType.kt
@@ -1,0 +1,56 @@
+package io.gitlab.arturbosch.detekt.rules.bugs
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.psi.KtBinaryExpressionWithTypeRHS
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.types.typeUtil.makeNotNullable
+
+/**
+ * Disallow to cast to nullable types. There are cases where `as String?` is misused as the safe cast (`as? String`),
+ * so if you want to prevent those cases, turn on this rule.
+ *
+ * <noncompliant>
+ * fun foo(a: Any?) {
+ *     val x: String? = a as String? // If 'a' is not String, ClassCastException will be thrown.
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * fun foo(a: Any?) {
+ *     val x: String? = a as? String
+ * }
+ * </compliant>
+ *
+ * @requiresTypeResolution
+ */
+class CastToNullableType(config: Config = Config.empty) : Rule(config) {
+    override val issue: Issue = Issue(
+        javaClass.simpleName,
+        Severity.Defect,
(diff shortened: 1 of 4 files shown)


detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastToNullableTypeSpec.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    describe("CastToNullableTypeSpec rule") {
+        it("casting to nullable types") {
+            val code = """
+                fun foo(a: Any?) {
+                    val x: String? = a as String?
+                } 
+            """
+            val findings = subject.compileAndLintWithContext(env, code)
+            Assertions.assertThat(findings).hasSize(1)
```
schalkms, line 24:
```suggestion
            assertThat(findings).hasSize(1)
```

schalkms (approved):

BraisGabin:
Sorry, but I still don't see the value of this rule. And why is OK if the initial type is not nullable. I'd like to know what the other maintainers think about it.

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastToNullableTypeSpec.kt
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        it("safe casting") {
+            val code = """
+                fun foo(a: Any?) {
+                    val x: String? = a as? String
+                } 
+            """
+            val findings = subject.compileAndLintWithContext(env, code)
+            assertThat(findings).hasSize(0)
```
cortinico, line 35:
```suggestion
            assertThat(findings).isEmpty()
```

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastToNullableTypeSpec.kt
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        it("type checking") {
+            val code = """
+                fun foo(a: Any?) {
+                    val x = a is String?
+                } 
+            """
+            val findings = subject.compileAndLintWithContext(env, code)
+            assertThat(findings).hasSize(0)
```
cortinico, line 45:
```suggestion
            assertThat(findings).isEmpty()
```

cortinico (approved):
I think the rule is valuable. To be honest I never really wrote a cast to a nullable type (`as String?`) and I've always used the `as?` variant, but I understand that it could be useful to detect them.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastToNullableType.kt
```diff
@@ -0,0 +1,56 @@
...
+    override fun visitBinaryWithTypeRHSExpression(expression: KtBinaryExpressionWithTypeRHS) {
+        super.visitBinaryWithTypeRHSExpression(expression)
+        if (bindingContext == BindingContext.EMPTY) return
+
+        val operationReference = expression.operationReference
+        if (operationReference.getReferencedNameElementType() != KtTokens.AS_KEYWORD) return
+
+        val typeReference = expression.right ?: return
+        val type = bindingContext[BindingContext.TYPE, expression.right] ?: return
+        if (!type.isMarkedNullable) return
```
chao2zhang, line 51: Do you think we can replace the check here with `expression.right.typeElement is KtNullableType` to avoid type resolution?

t-kameyama: Good catch!

chao2zhang (approved):

chao2zhang (approved):

PR #3499, opened 2021-02-25
Add UnusedUnaryOperator rule

### LINKED ISSUE #312: Rule: Possible Semicolon Inference Issues
How about to add a new rule for next the case?

```
//First Part
val var1 = 1 + 2 + 3 + 4
println(var1)

//Second Part
val var2 = 1 + 2
              + 3 + 4
println(var2)

//Result
//10
//3
```

t-kameyama:
Fixes #312


detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnusedUnaryOperatorSpec.kt
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        it("overloaded unary operator") {
+            val code = """
+                data class Foo(val x: Int)
+                operator fun Foo.plus(other: Foo) = Foo(this.x + other.x)
+                operator fun Foo.unaryMinus() = Foo(-x)
+                fun test() {
+                    val p = Foo(1) + Foo(2)
+                        - Foo(3)
```
BraisGabin, line 113: This case could have false negatives, but that's better than false postives for sure.

BraisGabin (approved):

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnusedUnaryOperator.kt
```diff
@@ -0,0 +1,72 @@
...
+    )
+
+    @Suppress("ReturnCount")
+    override fun visitPrefixExpression(expression: KtPrefixExpression) {
+        super.visitPrefixExpression(expression)
+        if (bindingContext == BindingContext.EMPTY) return
+
+        if (expression.baseExpression == null) return
+        val operationToken = expression.operationToken
+        if (operationToken != KtTokens.PLUS && operationToken != KtTokens.MINUS) return
```
cortinico, line 55: Is there a reason why `.not()` (the `!`) unary operator is not considered in this rule?

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnusedUnaryOperatorSpec.kt
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        it("used as annotation value argument") {
+            val code = """
+                annotation class Ann(val x: Int)
+                @Ann(x = -1)
+                val y = 2
+            """
+            val findings = subject.compileAndLintWithContext(env, code)
+            assertThat(findings).hasSize(0)
```
cortinico, line 103:
```suggestion
            assertThat(findings).isEmpty()
```

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnusedUnaryOperator.kt
t-kameyama: `!` cannot be used as a binary operator.

chao2zhang:
I talked with my colleagus about this rule and found out some interesting discussions:
- https://github.com/android/kotlin-guides/issues/51
- https://github.com/pinterest/ktlint/issues/163

With some manual testing
```
val d = 1
    + 1
``` would fail because of `unary-op-spacing`  `+ 1` should be `+1`.

But even we change it to `+1`
```
val d = 1
    +1
```  still failed because of `indent` `Unexpected indentation (8) (should be 4)`

I believe this new rule is already covered by the combination of `unary-op-spacing` and `indent`, which does not require type resolution. Therefore my proposal would be not to add duplicate rules.

**Comment by t-kameyama** [comment]
ktlint reports nothing in the following case:
```kotlin
fun test() {
    val x = 1 + 2
    +3
}
```

chao2zhang:
> ktlint reports nothing in the following case:
> 
> ```kotlin
> fun test() {
>     val x = 1 + 2
>     +3
> }
> ```

In this scenario, `+3` is clearly a separate line instead of the continuation of the previous line.

t-kameyama:
I think this problem is easier to understand with one detekt rule than with a combination of two ktlint rules, but if you don't need this rule, close this PR.

chao2zhang:
@cortinico @BraisGabin What do you guys think? If both of you support this rule, I am fine with merging this in.

cortinico:
> What do you guys think? If both of you support this rule, I am fine with merging this in.

I don't have a strong opinion honestly. I would lean towards merging this. Mostly because is disabled by default and because _explicit is better than implicit_, we offer a rule that covers this exact scenario rather than "suggesting" a combination of two.

BraisGabin:
I think that this is a code smell problem, not a formatting problem. So I vote to merge it.

It would be nice to have "experimental rules". Rules that we are not sure if we want to keep or not because they can have far too much false positive or they could be too subjective... But I don't think that this rule should wait for us to implement something like that.

chao2zhang (approved):
Thank you for the great test coverage. I played around with the code a little bit and found that those test cases were quite exhaustive for false positives.

Issue #3498, opened 2021-02-25
ForbiddenMethodCall rule distinct different signatures


DcortezMeleth:
I have checked for similar feature request but couldn't have found any.

## Expected Behaviour of the rule
I would expect ForbiddenMethodCall to allow for specifying not only method name but a whole signature, so we can write it config like below to report `LocalDate.now()` calls:
```
ForbiddenMethodCall:
  active: true
  methods: ['java.time.LocalDate.now']
```
and  use config as below to report `LocalDate.now(clock)` calls:
```
ForbiddenMethodCall:
  active: true
  methods: ['java.time.LocalDate.now(java.time.Clock)']
```

Of course this just an example of how I see that feature implemented, but if you feel that notation will be more readable or easier for parse go for it.

## Context
In some use cases it is important to distinct methods with the same name but different signatures.
In our case it would be distinction between `LocalDate#now` and `LocalDate#now(Clock)`.

This distinction is required as we are working with time domain, so usages of parameterless calls are undesired because they make code much harder to test and may make test unstable (passing on one day and not passing on the next one). On the other hand calls with clock as a param in all cases use Clock instance injected by DI framework, so they work fine in our case.

cortinico:
I think this is doable. Perhaps we should:
- Consider all the overloads if you specify `java.time.LocalDate.now`
- Check also the parameters if you specify the parameters list: `java.time.LocalDate.now(java.time.Clock)`

This would mean that specifying `java.time.LocalDate.now` and `java.time.LocalDate.now()` would have a different semantic.

DcortezMeleth:
Sound good to me, as this way if you decide do deprecate all methods with the same name, you will still have a compact config, but when it is necessary you can go for more verbose notation giving you more granularity.

cortinico:
Would you be up to propose a PR for this new feature @DcortezMeleth ?

DcortezMeleth:
I could give it a try :)

cortinico:
Let us know if you get stuck or you need further help 🙏

DcortezMeleth:
I have create PR with simple implementation. It would be grate if you could take a look at it and tell if this is a good approach as I didn't have previous experience with reflection in Kotlin and may be using wrong mechanisms.

I'm aware that docs should be updated and maybe some more tests added, but didn't won't to invest additional time not being sure if logic or api won't change.

cortinico:
Thanks for the PR, it's already great so far

[...]

PR #3505, opened 2021-02-28
Adding support for full method signatures in ForbiddenMethodCall

DcortezMeleth:
This PR is for now a proof of concept for [#3498](https://github.com/detekt/detekt/issues/3498).

Namely it's adding support for full method signatures (like `java.time.LocalDate.now(java.time.Clock)` in ForbiddenMethodCall rule.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
@@ -8,6 +8,8 @@ import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.internal.valueOrDefaultCommaSeparated
+import io.gitlab.arturbosch.detekt.rules.extractMethodNameAndParams
+import org.jetbrains.kotlin.js.descriptorUtils.getJetTypeFqName
 import org.jetbrains.kotlin.psi.KtBinaryExpression
 import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtExpression
@@ -29,7 +31,10 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
  * }
  * </noncompliant>
  *
- * @configuration methods - Comma separated list of fully qualified method signatures which are forbidden
+ * @configuration methods - Comma separated list of fully qualified method signatures which are forbidden.
+ * Methods can be defined without full signature (i.e. `java.time.LocalDate.now`) which will report calls of all methods
+ * with this name or with full signature (i.e. `java.time.LocalDate(java.time.Clock)`) which would report only call
+ * with this concrete signature.
  *  (default: `['kotlin.io.println', 'kotlin.io.print']`)
  *
  * @requiresTypeResolution
@@ -45,6 +50,7 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
     )
 
     private val forbiddenMethods = valueOrDefaultCommaSeparated(METHODS, DEFAULT_METHODS)
+        .map { extractMethodNameAndParams(it) }
 
     override fun visitCallExpression(expression: KtCallExpression) {
         super.visitCallExpression(expression)
@@ -70,14 +76,29 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
         if (bindingContext == BindingContext.EMPTY) return
 
         val resolvedCall = expression.getResolvedCall(bindingContext) ?: return
-        val fqName = resolvedCall.resultingDescriptor.fqNameOrNull()?.asString()
+        val methodName = resolvedCall.resultingDescriptor.fqNameOrNull()?.asString()
+        val encounteredParamTypes = resolvedCall.resultingDescriptor.valueParameters
+            .map { it.type.getJetTypeFqName(false) }
 
-        if (fqName != null && fqName in forbiddenMethods) {
-            report(
-                CodeSmell(
-                    issue, Entity.from(expression), "The method $fqName has been forbidden in the Detekt config."
-                )
-            )
+        if (methodName != null) {
+            forbiddenMethods
+                .filter { methodName == it.first }
+                .forEach {
+                    val expectedParamTypes = it.second
+                    val noParamsProvided = expectedParamTypes == null
+                    val paramsMatch = expectedParamTypes == encounteredParamTypes
+
+                    if (noParamsProvided || paramsMatch) {
+                        report(
+                            CodeSmell(
+                                issue,
+                                Entity.from(expression),
+                                "The method ${it.first}(${expectedParamTypes?.joinToString() ?: ""}) " +
+                                        "has been forbidden in the Detekt config."
+                            )
+                        )
+                    }
+                }
         }
     }
 
--- a/detekt-psi-utils/api/detekt-psi-utils.api
+++ b/detekt-psi-utils/api/detekt-psi-utils.api
@@ -104,6 +104,7 @@ public final class io/gitlab/arturbosch/detekt/rules/KtValueArgumentKt {
 }
 
 public final class io/gitlab/arturbosch/detekt/rules/MethodSignatureKt {
+	public static final fun extractMethodNameAndParams (Ljava/lang/String;)Lkotlin/Pair;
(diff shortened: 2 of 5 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
```diff
@@ -45,6 +46,14 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
     )
 
     private val forbiddenMethods = valueOrDefaultCommaSeparated(METHODS, DEFAULT_METHODS)
+        .map { methodSignature ->
+            val groups = METHOD_PARAM_REGEX.find(methodSignature)?.groups
+            val methodName = groups?.get(METHOD_GROUP_NAME)?.value
+            val params = groups?.get(METHOD_PARAM_GROUP_NAME)?.value?.split(",")
+                ?.map { it.trim() }
+                ?.filter { it.isNotBlank() }
```
cortinico, line 54: Please replace with something like:

```suggestion
            val groups = methodSignature.split('(',')')
            val methodName = groups[0].trim()
            val params = if (groups.size <= 1) {
                null
            } else {
                groups[1].split(",")
                    .map(String::trim)
                    .filter(String::isNotBlank)
            }
```

Regexes are hard to maintain in the long run, especially for shared codebases like Detekt.

cortinico: If you wish, you can also create a util function inside the `detekt-psi-utils` package and write a unit test (like a `.parseFqNameAndParams()` or so).

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
```diff
@@ -71,13 +80,24 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
...
-            report(
-                CodeSmell(
-                    issue, Entity.from(expression), "The method $fqName has been forbidden in the Detekt config."
-                )
-            )
+        if (fqName != null) {
+            forbiddenMethods
+                .filter { fqName == it.first }
+                .forEach {
+                    if (it.second == null || it.second == methodParameterFqNames) {
```
cortinico, line 90: To tackle the `java.time.LocalDate.of(Int, Int, Int)`, you should make sure you run a param-by-param comparison of the two lists and check both the fully qualified and the simple name:

```kotlin
val configParams = it.second
val match = methodParameterFqNames.foldIndexed(true) { index: Int, acc: Boolean, param: String ->
    acc && (param == configParams[index] || param.split('.').last() == configParams[index])
}
```

cortinico (commented):
Great work 💪  We need a bit of fine tuning and I think we should be able to merge this

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
BraisGabin: Probably, because it's not that easy because some functions can be called `` `like, this`(Int)``.

And, if we want to move out this code,  it should be placed in tooling utils instead of psi utils, right? This code is to parse the configuration. It's not related with psi.

cortinico: > it should be placed in tooling utils instead of psi utils, right?

I checked and `detekt-psi-utils` is already imported inside the `-rules` modules as an `api` dependency. The `detekt-tooling` is not and I don't think it's worth to add a module dependency for a function.

Not to be done in this PR, but I think we should rename `detekt-psi-utils` to `detekt-rules-utils` (given that inside there is also not PSI related utils) or create a separate `detekt-rules-utils` just for those functions.

DcortezMeleth: Good idea. My first thought was to use regex and I was so fixed on this approach that I missed simpler one.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
DcortezMeleth: I see one issue here. Won't we create inconsistency between how method should be defined (only fully qualified) and params (fully qualified or just class name)?
Other than that it seems ok, as it will allow you to shorten definition and only use qualified names when false-positives would be encountered otherwise.

DcortezMeleth:
I have updated PR following your instructions. 
I didn't extract methods params parsing to util as I wasn't sure if it will be useful anywhere else but if you feel that we should go for it just tell me where is the best place to put it inside util module.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
cortinico: > I see one issue here. Won't we create inconsistency between how method should be defined (only fully qualified) and params (fully qualified or just class name)?
> Other than that it seems ok, as it will allow you to shorten definition and only use qualified names when false-positives would be encountered otherwise.

You're right. I think we could then revert to always be explicit. Maybe add a note example in the doc so users are aware on how to us it properly.

cortinico:
> I wasn't sure if it will be useful anywhere else but if you feel that we should go for it just tell me

Having the function extracted was more for testing reason. It's easier to write a couple of unit tests in a separate module for just that function. Anyway that's not a hard requirement for merging this PR.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
DcortezMeleth: Done.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
```diff
@@ -8,6 +8,8 @@ import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.internal.valueOrDefaultCommaSeparated
+import io.gitlab.arturbosch.detekt.rules.extractMethodNameAndParams
+import org.jetbrains.kotlin.js.descriptorUtils.getJetTypeFqName
```
chao2zhang, line 12: Not necessarily an issue for this PR, because `getJetTypeFqName` is used at several places. However, the package path looks very weird because `js` appears.

chao2zhang:
What do we think about methods with default arguments? If `fun foo(bar: Int, baz: Int = 5)` is in the configuration and we encounter `foo(5)`, should we also report that?

(This PR looks good in general, this might be an edge case that we can follow up)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
DcortezMeleth: As I mentioned it is my first meeting with kotlin reflection and it was the solution I have found but if you have some idea how to achieve it with other method I will be happy to try that out.

DcortezMeleth:
As for the methods with default parameters I think we should report them all, because otherwise if we would like to forbid method with more than one default parameter, we would have to cover or possible combinations in configuration.
Also typically we wan't to discourage usage of a method and not the certain way of calling it, and even if we did, I feel that this should be done be a separate rule then.

I have added test case with default parameter. It worked without changes in the rule logic.

chao2zhang (approved):

DcortezMeleth:
Thanks for approval. If I can do anything else to help you with this PR please let me know!

cortinico (approved):

schalkms (commented):
If @BraisGabin approves, I think we can merge this.
Thanks for this awesome PR. It's a nice improvement of this rule.

BraisGabin (approved):

BraisGabin:
Great work!

DcortezMeleth:
No problem. I'm happy that I could have helped.

PR #3507, opened 2021-03-01
Fix SARIF validation failure

chao2zhang:
When trying the relative output of SARIF report, [the official validation tool](https://sarifweb.azurewebsites.net/Validation) yields the following failure:
WX20210228-232942


BraisGabin (approved):
It seems like SARIF babysits a lot.

schalkms (approved):

Issue #3490, opened 2021-02-21
Unable to load class 'com.android.build.gradle.BaseExtension'


volsahin:
## Steps to Reproduce

I intended to upgrade detekt version from 1.9.1 to latest 1.16.0-RC1 in android project and build failed.
The project uses classic groovy gradle files. It is nice to mention that i apply detekt as a seperate gradle script.

Here is the current detekt.gradle script

```
buildscript {
    ext.detektVersion = '1.9.1'
    repositories {
        jcenter()
    }
    dependencies {
        classpath "io.gitlab.arturbosch.detekt:detekt-gradle-plugin:$detektVersion"
    }
}

apply plugin: io.gitlab.arturbosch.detekt.DetektPlugin

detekt {
    // some options
}
```
And that is how it's applied into app
`apply from: '../scripts/detekt.gradle'`

As i see 1.11.1 is the latest version which can be built without a problem. Starting 12.0.0 to the latest version, build fails.
Gradle plugin versions are the latest one.

## Stacktrace

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':app'.
> Failed to notify project evaluation listener.
   > com/android/build/gradle/BaseExtension

* Exception is:
org.gradle.api.ProjectConfigurationException: A problem occurred configuring project ':app'.
Caused by: org.gradle.internal.event.ListenerNotificationException: Failed to notify project evaluation listener.
Caused by: java.lang.NoClassDefFoundError: com/android/build/gradle/BaseExtension
	at io.gitlab.arturbosch.detekt.internal.DetektAndroid$registerDetektAndroidTasks$1.execute(DetektAndroid.kt:59)
        at io.gitlab.arturbosch.detekt.internal.DetektAndroid$registerDetektAndroidTasks$1.execute(DetektAndroid.kt:21)
Caused by: java.lang.ClassNotFoundException: com.android.build.gradle.BaseExtension
        ... 139 more

* Get more help at https://help.gradle.org

BUILD FAILED in 435ms
```

Also #3476 #3260 these issues experienced similar exceptions as far as i can see.

## Environment
* Version of detekt used: 1.16.0-RC1
* Version of Gradle used (if applicable): 4.1.2
* Version of Kotlin: 1.4.21

chao2zhang:
In your root project, did you apply both detekt plugin and `kotlin-android` plugin? I imagine this is an issue where we accidentally apply detekt to the root project.

cortinico:
[...]

Issue #3476, opened 2021-02-16
Failed to apply detekt(v1.16.0-RC1) plugin in subproject


niqiu1683:
## Expected Behavior

Everything performs well under older detekt version, such as 1.15.0, however I can't apply detekt plugin in subproject under version 1.16.0-RC1.

root build.gradle file:.
```gradle
plugins {
 // id("io.gitlab.arturbosch.detekt").version("1.15.0")  // older version is ok
    id("io.gitlab.arturbosch.detekt").version("1.16.0-RC1")
}

subprojects {
    apply plugin: "io.gitlab.arturbosch.detekt"    // this line has a problem
    tasks.detekt.jvmTarget="1.8"
    detekt {
        config=files("$rootDir/checkstyle/detekt_config_codecc_1.0.yml")
        reports {
            html.enabled=true
        }
    }
    tasks.matching {
        it.name.startsWith("detekt")
    }.configureEach {
        outputs.upToDateWhen { false }
    }
}
```

## Observed Behavior

There is a problem occurred when I sync gradle, show as below:

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':app'.
> Failed to notify project evaluation listener.
   > com/android/build/gradle/internal/core/InternalBaseVariant

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.ProjectConfigurationException: A problem occurred configuring project ':app'.
	at org.gradle.configuration.project.LifecycleProjectEvaluator.wrapException(LifecycleProjectEvaluator.java:80)
	at org.gradle.configuration.project.LifecycleProjectEvaluator.addConfigurationFailure(LifecycleProjectEvaluator.java:73)
	at org.gradle.configuration.project.LifecycleProjectEvaluator.access$600(LifecycleProjectEvaluator.java:53)
	at org.gradle.configuration.project.LifecycleProjectEvaluator$NotifyAfterEvaluate.run(LifecycleProjectEvaluator.java:199)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:402)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:394)
[...]

PR #3511, opened 2021-03-03
Allow opt-out configuring detekt android and multiplatform

chao2zhang:
Detekt Android can worsen the user experience in case of
- Mismatched android gradle plugin versions: #3476
- Apply script plugins: #3490

For those "experimental" gradle plugin setups, we should provide a way to allow users to opt-out to unblock users from upgrading.

### DIFF
--- a/detekt-gradle-plugin/build.gradle.kts
+++ b/detekt-gradle-plugin/build.gradle.kts
@@ -17,6 +17,7 @@ dependencies {
     compileOnly(kotlin("gradle-plugin"))
 
     testImplementation(project(":detekt-test-utils"))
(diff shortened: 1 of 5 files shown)


cortinico (approved):

BraisGabin (approved):
I'm ok with this PR but I must admit that I have no idea how this help our users or why this creates issues

chao2zhang:
> I'm ok with this PR but I must admit that I have no idea how this help our users or why this creates issues

This can help users to upgrade detekt versions while giving us more breathing room to debug Detekt Android/Multiplatform related issues.

PR #3515, opened 2021-03-04
Fix not being able to override `insertFinalNewLine`

### LINKED ISSUE #3513: FinalNewline with insertFinalNewLine=false always failing
We switched from **detekt + ktlint** (jlleitschuh gradle plugin) to **detekt + detekt formatting plugin**.

I adopted all our code-style configurations and the formatting rules are working fine except the FinalNewline rule. We don't allow final-newlines. So we have `insert_final_newline=false` set in our `.editorconfig`. Ktlint [checks this configuration](https://github.com/pinterest/ktlint/blob/b5b8e40986e192b550e385481b0759bcbfba1850/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/FinalNewlineRule.kt#L58) and runs as expected (warns if a final-newline exists, does not warn if it is absent)

To get the same result I configured detekt like this:
```
formatting:
  FinalNewline:
    active: true
    insertFinalNewLine: false
```

## Expected Behavior
When `insertFinalNewLine: false` is set, detekt formatting should not warn about files with no final-newline (behave like ktlint with `insert_final_newline=false`)

## Observed Behavior
Detekt reports FinalNewline errors if a file has no final-newline, even though `insertFinalNewLine: false` is set

## Your Environment
* Version of detekt used: 1.16.0-RC3
* Version of Gradle used (if applicable): 6.8.3
* Operating System and version: macOS 11.2.2

chao2zhang:
This fixes the type in the constant to match `@configuration`

Resolve #3513

### DIFF
--- a/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/FinalNewline.kt
+++ b/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/FinalNewline.kt
@@ -34,4 +34,4 @@ class FinalNewline(config: Config) : FormattingRule(config) {
     )
 }
 
-const val INSERT_FINAL_NEWLINE = "insertFinalNewline"
+const val INSERT_FINAL_NEWLINE = "insertFinalNewLine"


schalkms (approved):

cortinico (approved):

BraisGabin:
Good catch 👏👏👏

BraisGabin (approved):

Issue #3360, opened 2021-01-08
Continuous improvement on SARIF support


chao2zhang:
This issue lists all the improvements after we built a basic SARIF support #3045
- [x] Enable SARIF output by default in Gradle+CLI. See https://github.com/detekt/detekt/pull/3268
- [x] Enable relative path by default. See https://github.com/detekt/detekt/pull/3359#discussion_r554181042
- [x] Add documentation and samples for relative path, severity, and Github integration for SARIF
- [x] Add integration tests with GradleKit to verify SARIF, relative path, and configurable severity once a version is published. The publishing constraint can be lifted once https://github.com/detekt/detekt/issues/3324 is completed.
- [x] Fix absolute path usage in the message of CodeSmell -> This is fixed by #3386
- [x] In Detekt 2.0, remove the deprecated Severity -> This is addressed by #6399 
- [x] In Detekt 2.0, refactor the implementation of SeverityLevel so it can be configured per issue. https://github.com/detekt/detekt/pull/3310#pullrequestreview-556036609
- [ ] Plan to improve the CorrectSmell message so that we do not fall back to the issue description.
- [x] Merge SARIF support from all modules to generate a single top-level report

schalkms:
Good summary!

madlymad:
Maybe some "SARIF" validation is also required. I noticed a case where an complete empty `.kt` file was part of the repository and the created SARIF report was invalid

BraisGabin:
> Maybe some "SARIF validation" is also required in the unit testing level. I noticed a case where a completely empty `.kt` file was part of the repository and the created SARIF report was invalid.

If that's the case, please open an issue with a way to reproduce it. That's a bug and we should fix it. It's not an improvement.

BraisGabin:
Can we close this one now? I think that sarif is in a good state right now. If there are things to do I think that we should have separated issues for it. Also I don't fully understand what we need to do to finish some of those points that are still missing. Do you recall them @chao2zhang?

chao2zhang:
It's been a while since I work any project involving Kotlin 😅 . I have verified the check list in the original summary, only one issue remaining `Plan to improve the CorrectSmell message so that we do not fall back to the issue description.` but I couldn't find reference of `CorrectSmell` in the codebase now.

I would also vote for closing this issue.

PR #3522, opened 2021-03-07
Merge SARIF reports

chao2zhang:
This PR addresses the last item in #3360

## Notes
1. Although Gradle repackages Jackson dependency internally, its version seems outdated and I encountered two parsing issues. Therefore, I am adding the latest Jackson dependency explicitly.

2. `-Dorg.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m` becomes necessary. It turns out that `DetektSarifReportMergeTest.kt` runs out of Metaspace as OutOfMemoryError with Java 11 + Ubuntu/Windows. Hence I am bumping the Xmx and MaxMetaspaceSize slightly for test to pass.


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/SarifReportMergeTask.kt
```diff
@@ -0,0 +1,29 @@
...
+    abstract val input: ConfigurableFileCollection
+
+    @get:OutputFile
+    abstract val output: RegularFileProperty
+
+    @TaskAction
+    fun merge() {
+        logger.info(input.files.joinToString(separator = "\n") { it.absolutePath })
+        SarifReportMerger.merge(input.files, output.get().asFile)
+        logger.lifecycle("Merged SARIF output to ${output.get().asFile.absolutePath}")
```
cortinico, line 27: This basically `XmlReportMergeTask` modulo the `SarifReportMerger`.

Could we refactor this a bit? Say define a `ReportMerger` Interface + a `ReportMergeTask` that takes an `enum` as input property?

We could probably have other mergers (HTML?) so it's better we clean it up a bit before we start exposing this 🤔 ?

docs/pages/reporting.md
```diff
@@ -77,11 +77,13 @@ The machine-readable report formats support report merging.
 Detekt Gradle plugin is not opinionated in how merging is set up and respects each project's build logic, especially 
 the merging makes most sense in a multi-module project. In this spirit, only Gradle tasks are provided.
 
-At the moment, only merging XML is supported. You can refer to the sample script below and 
-run `./gradlew detekt xmlReportMerge --continue` to execute detekt tasks and merge their reports.
+At the moment, merging XML and SARIF are supported.
```
cortinico, line 80:
```suggestion
At the moment, only merging XML and SARIF are supported. 
```

cortinico (approved):
LGTM 👍 I've left a minor comment for a refactoring. But we can also skip it if we wish

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/SarifReportMergeTask.kt
chao2zhang: Good idea. We could potentially avoid duplication in docs as well.

PR #3543, opened 2021-03-11
Enable SARIF output format by default

chao2zhang:
Given that we have 1.16.0 release, this PR flips the default flag to true for SARIF output format. See #3360

(I also removed the dependency for `:detekt-report-sarif` to `:detekt-tooling`.


schalkms (approved):

BraisGabin (approved):
Should we wait a bit to merge this one? It seems that 1.16.0 have some issues so maybe we create 1.16.1 and we don't want this in that version.

chao2zhang:
Sounds good. This PR is tagged with 1.17 so we can definitely wait a little bit.

Issue #3523, opened 2021-03-07
SwallowedException: Multi-line code block renders wrong in docs


sschuberth:
This

https://github.com/detekt/detekt/blob/2fbb7c99e171648bc458e7fb25909881528de52b/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/SwallowedException.kt#L67-L70

[renders](https://detekt.github.io/detekt/exceptions.html#swallowedexception) like

![image](https://user-images.githubusercontent.com/349154/110237736-d3518600-7f3d-11eb-9cf4-23893aaded9e.png)

I'm not quite sure how to fix this in the best way, though.

sschuberth:
Hmm, I started looking into this, but some magic that I don't understand seems to be going on. When running

    ./gradlew :detekt-generator:generateDocumentation --build-cache --parallel

the generated `docs/pages/documentation/exceptions.md` contains double-backticks like

```
#### Configuration options:

* ``ignoredExceptionTypes`` (default: ``- InterruptedException
    - NumberFormatException
    - ParseException
    - MalformedURLException``)

   exception types which should be ignored by this rule
```

Is that intended? As it seems unnecessary, and all double-backticks should be single backticks.

schalkms:
I think this is a bug in the `RuleSetPagePrinter.kt`. Multi-line code configurations are not supported, as it seems like.

https://github.com/detekt/detekt/blob/master/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/RuleSetPagePrinter.kt

sschuberth:
> As it seems unnecessary, and all double-backticks should be single backticks.

That particular bug seems to stem from

https://github.com/detekt/detekt/blob/2fbb7c99e171648bc458e7fb25909881528de52b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/out/Markdown.kt#L42

But I agree there seem to be multiple bugs. I'll look into them.

schalkms:
> But I agree there seem to be multiple bugs. I'll look into them.

Thank you! 🙂

BraisGabin:
The double quote is intended. We have some code blocks that contains a single `` ` `` for example ``fun `hola`()`` if we use single quote this happen: `fun `hola`()`

sschuberth:
@BraisGabin please have a look at https://github.com/detekt/detekt/pull/3545.

[...]

PR #3545, opened 2021-03-12
Markdown: Reintroduce double-backticks for inline code rendering

sschuberth:
This partly reverts 9cf9f08 and add a code comment to explain why the
double-backticks were in fact intended, also see

https://github.com/detekt/detekt/issues/3523#issuecomment-793532727

### DIFF
--- a/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/out/Markdown.kt
+++ b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/out/Markdown.kt
@@ -39,7 +39,9 @@ inline fun MarkdownContent.orderedList(sectionList: () -> List<String>) {
     }
 }
 
-inline fun MarkdownContent.code(code: () -> String) = "`${code()}`"
+// Note: Use double-backticks here to be able to render code that itself contains backticks.
+inline fun MarkdownContent.code(code: () -> String) = "``${code()}``"
+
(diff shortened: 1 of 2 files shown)


BraisGabin (approved):

schalkms (commented):
Wouldn't it make more sense to remove the occurrences that contain back-ticks in order to avoid this special case here? What do you think? @BraisGabin

BraisGabin:
How would you avoid this? https://detekt.github.io/detekt/naming.html#functionnaming

What's the problem with the double tick?

cortinico (approved):

schalkms:
@cortinico 

One can avoid this scenario by putting three back-ticks around it **'```'**.

BraisGabin:
> @BraisGabin one can avoid this scenario by putting three back-ticks around it **'```'**.

That would make us to change the documentation and now that we are introducing the `@Configuration` annotation to extract the information directly from the code we can't do this. I still don't get what's wrong with using double quote.

schalkms:
The official Markdown guide specifies the double back-tick as a valid way to escape a code snippet [1]. Hence, this is fine for me.

[1] https://www.markdownguide.org/basic-syntax/#escaping-backticks

schalkms:
The test suite is unfortunately not green for this PR.

sschuberth:
> The test suite is unfortunately not green for this PR.

Should be fixed now.

PR #3548, opened 2021-03-13
docs: Improve the reporting examples for merging reports

sschuberth:
In particular, use finalizedBy() so the merge task is also run if the
detekt task fails.

### DIFF
--- a/docs/pages/reporting.md
+++ b/docs/pages/reporting.md
@@ -17,11 +17,11 @@ Similar to the console output, each line of the txt output represents a finding 
 finding signature to help edit [baseline files](gettingstarted/gradle.md).
 
 ```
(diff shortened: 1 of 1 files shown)


docs/pages/reporting.md
```diff
@@ -106,7 +107,7 @@ subprojects {
 
 ```kotlin
 val reportMerge by tasks.registering(io.gitlab.arturbosch.detekt.report.ReportMergeTask::class) { 
-  output.set(project.layout.buildDirectory.file("reports/detekt/merge.xml")) // or file("reports/detekt/merge.sarif")
+  output.set(rootProject.buildDir.resolve("reports/detekt/merge.xml")) // or "reports/detekt/merge.sarif"
```
chao2zhang, line 110: Good catch

chao2zhang (approved):

docs/pages/reporting.md
sschuberth: Actually, the original syntax worked, but the new one is a bit shorter and more "Kotlin-ish".

schalkms (approved):

PR #3550, opened 2021-03-13
Adds a ForbiddenSingleExpressionSyntax rule

### LINKED ISSUE #1675: Disallow single expression syntax when the function returns Unit
OK
```
fun doSth() = getName()

fun getName() = "Igor"
```

Fail
```
fun doSth() = println("ABC")

fun println(): Unit {
    // ...
}
```

The main reason behind it is that it's not obvious that no value will be returned in 2nd example

marschwar:
closes #1675 

While trying to find a reference for the rule documentation, I realized that there does not seem to be agreement about this in the community. I guess this will never become a rule that is turned on be default. 

While I personally like the rule, I am wondering if detekt should even be so opinionated about things that are not part of the official coding conventions or guidelines.

And then I may have formatting issues (again). Opening any other rule class causes the ktlint plugin to report an error such as `Unexpected indentation (20) (should be 12) (indent)`. Any help will be greatly appreciated.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSingleExpressionFunction.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSingleExpressionFunction.kt
@@ -1,0 +1,78 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getType
+import org.jetbrains.kotlin.types.typeUtil.isUnit
+import org.jetbrains.kotlin.types.typeUtil.makeNotNullable
+
+/**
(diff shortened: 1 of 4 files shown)


BraisGabin:
> While I personally like the rule, I am wondering if detekt should even be so opinionated about things that are not part of the official coding conventions or guidelines.

We talked about this a lot of times in different issues. We need a controversial ruleset. But I also like this rule a lot. And I know other devs that don't like the overuse of this feature either. 

> And then I may have formatting issues (again). Opening any other rule class causes the ktlint plugin to report an error such as `Unexpected indentation (20) (should be 12) (indent)`. Any help will be greatly appreciated.

We can open a new issue about this topix. I think that we should enable ktlint in this project. I have the same issue autoformatting some rules that it changes far too much lines.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSingleExpressionFunction.kt
```diff
@@ -0,0 +1,79 @@
...
+ */
+class ForbiddenSingleExpressionFunction(config: Config = Config.empty) : Rule(config) {
+    override val issue: Issue = Issue(
+        "ForbiddenSingleExpressionFunction",
+        Severity.Style,
+        "Avoid using single-expression syntax for functions that return `Unit` or `Unit?`.",
+        Debt.FIVE_MINS
+    )
+
+    @Suppress("ReturnCount")
```
schalkms, line 48: This suppression shouldn't be necessary, since the guard clause having a `return` statement are excluded.

schalkms (approved):

schalkms:
> We need a controversial ruleset.

I suggested this in a few PRs as well. +1 from my side.

marschwar:
How about I at least add a `@controversial` to the kdoc?

BraisGabin:
Great idea!!! Yes! We can set the rules in any ruleset but we can mark them as controversial! I think that we can make a new issue to implement this idea and merge this directly. The annotation is a better solution and it is not a breaking change so we can even mark old rules. Win-win!

marschwar:
I think this is obsolete as there is already a `ImplicitUnitReturnType` rule that checks the exact same case.

PR #3552, opened 2021-03-14
add generateDocumentation task to contribution guide

marschwar:
As mentionend in #3549 the contribution guide seems to be outdated. Running `gradlew build` after adding a new rule will fail without running `gradlew generateDocumentation` first.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -51,12 +51,11 @@ automatically. The format of the KDoc should be as follows:
        active: false
        name: whatever should be the default
     ```
-
-- ... do not forget to test the new rule and/or add tests for any changes made to a rule.
-Run detekt on itself and other kotlin projects with the `--run-rule RuleSet:RuleId` option to test your rule in isolation.
-Make use of the `scripts/get_analysis_projects.groovy` script to automatically establish a set of analysis projects.
-- ... do not forget to run `./gradlew build`. This will execute tests locally and update the `default-detekt.config.yml`
-as well as add the new/changed rules to the documentation.
+- ... run `./gradlew generateDocumentation` to add your rule and its config options to the `default-detekt-config.yml`.
+- ... do not forget to run `./gradlew build`. This will execute tests locally.
+- ... do not forget to test the new rule and/or add tests for any changes made to a rule. Run detekt on itself and other
+  kotlin projects with the `--run-rule RuleSet:RuleId` option to test your rule in isolation. Make use of
+  the `scripts/get_analysis_projects.groovy` script to automatically establish a set of analysis projects.
 - To print the AST of sources you can pass the `--print-ast` flag to the CLI which will print each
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -51,12 +51,10 @@ automatically. The format of the KDoc should be as follows:
        active: false
        name: whatever should be the default
     ```
-
+- ... run `./gradlew generateDocumentation` to add your rule and its config options to the `default-detekt-config.yml`.
 - ... do not forget to test the new rule and/or add tests for any changes made to a rule.
 Run detekt on itself and other kotlin projects with the `--run-rule RuleSet:RuleId` option to test your rule in isolation.
 Make use of the `scripts/get_analysis_projects.groovy` script to automatically establish a set of analysis projects.
-- ... do not forget to run `./gradlew build`. This will execute tests locally and update the `default-detekt.config.yml`
```
cortinico, line 58: I would suggest we keep this. The `./gradlew build` will take care of running your new rule against the whole detekt codebase, plus is the command that the CI is running to verify every PR

marschwar: Will it really do that? If the rule is not active by default (which most new rules are), then it is not used. Or am I missing something? Running `gradlew build` should be a given. Maybe the second sentence on the page should be changed from

> Use gradle build -x dokkaJekyll to build the source but exclude documentation jar generating to save time.

to

Make sure to run `./gradlew build` before opening a pull request. This will execute all tests and run all detekt rules against the entire code base. You can use `./gradlew build -x dokkaJekyll` to exclude documentation jar generation to save time.

But I am also ok with putting the sentence back in ;)

BraisGabin: 🤔 I think that the previous point talk about the tests so I think that we don't need this point. Maybe we can recommend to run `./gradlew test` before summiting any PR. But there are some PR's that just adds the failing test. Those never get a green but they are very valuable PRs too.

As a side note: If we need to advise to use `build -x dokkaJekyll` seems like `build` is doing far too much things. Don't you think?

BraisGabin (approved):

.github/CONTRIBUTING.md
cortinico: > Will it really do that? If the rule is not active by default (which most new rules are), then it is not used.

You're right. The Detekt run on the detekt project is done using the default config. Still, running the `test` task is a subset of what the CI will run.

If your code fails because of a `MagicNumber` violation, you won't see it locally if you run only `gradle test`.

I'd say we should keep `gradle build` to give confidence to contributors that whatever they run locally, is as near as possible to what the CI will run. 

> As a side note: If we need to advise to use build -x dokkaJekyll seems like build is doing far too much things. Don't you think?

Yeah that's another problem we should probably look into.

cortinico (approved):

PR #3553, opened 2021-03-14
New Rule: DoubleMutabilityForCollection

cortinico:
I'm adding a new rule that detects if you're declaring a `var` with also a Mutable type like:

```
var myList = mutableListOf(1,2,3)
```

Instead you should prefer to:

```
val myList = mutableListOf(1,2,3)
// OR
var myList = listOf(1,2,3)
```

Having "double mutability" is considered an anti-pattern as the `myList` variable can both be reassigned and `.clear()`-ed.

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DoubleMutabilityForCollection.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DoubleMutabilityForCollection.kt
@@ -1,0 +1,79 @@
+package io.gitlab.arturbosch.detekt.rules.bugs
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.descriptors.VariableDescriptor
+import org.jetbrains.kotlin.js.descriptorUtils.getJetTypeFqName
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.resolve.BindingContext
+
+/**
+ * Using `var` when declaring a mutable collection leads to double mutability. Consider instead
+ * declaring your variable with `val` or switching your declaration to use an immutable type.
+ *
+ * <noncompliant>
+ * var myList = mutableListOf(1,2,3)
+ * var mySet = mutableSetOf(1,2,3)
+ * var myMap = mutableMapOf("answer" to 42)
+ * </noncompliant>
+ *
+ * <compliant>
+ * // Use val
+ * val myList = mutableListOf(1,2,3)
+ * val mySet = mutableSetOf(1,2,3)
+ * val myMap = mutableMapOf("answer" to 42)
+ *
+ * // Use immutable types
+ * var myList = listOf(1,2,3)
+ * var mySet = setOf(1,2,3)
+ * var myMap = mapOf("answer" to 42)
+ * </compliant>
+ *
+ * @requiresTypeResolution
+ */
+class DoubleMutabilityForCollection(config: Config = Config.empty) : Rule(config) {
+
+    override val issue: Issue = Issue(
+        "DoubleMutabilityInCollection",
+        Severity.CodeSmell,
+        "Using var with mutable collections leads to double mutability. " +
+                "Consider using val or immutable collection types.",
+        Debt.FIVE_MINS
+    )
+
+    override fun visitProperty(property: KtProperty) {
+        super.visitProperty(property)
+
+        if (bindingContext == BindingContext.EMPTY) return
+
+        val type =
+            (bindingContext[BindingContext.DECLARATION_TO_DESCRIPTOR, property] as? VariableDescriptor)?.type ?: return
+        val standardType = type.getJetTypeFqName(false)
+        if (property.isVar && standardType in mutableTypes) {
+            report(CodeSmell(
+                issue,
+                Entity.from(property),
+                "Property ${property.name} is declared as `var` with a mutable type $standardType. " +
+                        "Consider using `val` or an immutable collection type"
+            ))
+        }
+    }
+
+    companion object {
+        val mutableTypes = setOf(
+            "kotlin.collections.MutableList",
+            "kotlin.collections.MutableMap",
+            "kotlin.collections.MutableSet",
+            "java.util.ArrayList",
+            "java.util.LinkedHashSet",
+            "java.util.HashSet",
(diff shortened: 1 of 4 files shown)


detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DoubleMutabilityInCollectionSpec.kt
```diff
@@ -0,0 +1,226 @@
...
+                }
+                """
+                val result = subject.compileAndLintWithContext(env, code)
+                assertThat(result).isEmpty()
+            }
+
+            it("does not detect var declaration with immutable map") {
+                val code = """
+                fun main() {
+                    val myMap = mapOf("answer" to 42)
```
BraisGabin, line 218: ❤️

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DoubleMutabilityForCollection.kt
```diff
@@ -0,0 +1,79 @@
...
+
+        val type =
+            (bindingContext[BindingContext.DECLARATION_TO_DESCRIPTOR, property] as? VariableDescriptor)?.type ?: return
+        val standardType = type.getJetTypeFqName(false)
+        if (property.isVar && standardType in mutableTypes) {
+            report(CodeSmell(
+                issue,
+                Entity.from(property),
+                "Property ${property.name} is declared as `var` with a mutable type $standardType. " +
+                        "Consider using `val` or an immutable collection type"
```
BraisGabin, line 62: 🤔 A variable inside a function can be called "property"? Shouldn't we call it variable?

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DoubleMutabilityInCollectionSpec.kt
```diff
@@ -0,0 +1,226 @@
...
+
+    describe("DoubleMutabilityInCollection rule") {
+
+        describe("valid cases") {
+
+            it("detects var declaration with mutable list") {
+                val code = """
+                fun main() {
+                    var myList = mutableListOf(1,2,3)
+                }
```
BraisGabin, line 25: All your tests check the case inside a function. Could you add some tests to check class properties? And as a top level properties?

BraisGabin (approved):
👍

Once we have a real support for severity I'd like to add a rule to detect all the usages of `var`. There are very few cases where it uses make sense.

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DoubleMutabilityInCollectionSpec.kt
cortinico: Yup included

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DoubleMutabilityForCollection.kt
```diff
@@ -0,0 +1,79 @@
...
+    )
+
+    override fun visitProperty(property: KtProperty) {
+        super.visitProperty(property)
+
+        if (bindingContext == BindingContext.EMPTY) return
+
+        val type =
+            (bindingContext[BindingContext.DECLARATION_TO_DESCRIPTOR, property] as? VariableDescriptor)?.type ?: return
+        val standardType = type.getJetTypeFqName(false)
```
t-kameyama, line 56:
```suggestion
       val standardType = type.constructor.declarationDescriptor?.fqNameOrNull()?.asString()
```
`getJetTypeFqName` function is in the `js` package, so I think you shouldn't use it.

BraisGabin: 🤔 should we add `org.jetbrains.kotlin.js.*` as a forbidden import to avoid this? It's not the first time that I see this kind of error.

cortinico: Sadly other rules are also using this util: https://github.com/detekt/detekt/search?q=getJetTypeFqName

I wasn't able to find an equivalent in the common compiler package.
Given that this function is using nothing js specific, I suggest we copy it over to the `-psi-utils`:

https://github.com/JetBrains/kotlin/blob/29b23e79f32791e456a5b4a453277f0f0b3e984d/js/js.frontend/src/org/jetbrains/kotlin/js/descriptorUtils/descriptorUtils.kt#L22-L41

chao2zhang: This is a good catch @t-kameyama and I have already raised the same concern in other PRs. There is no replacement for this function in the Kotlin source code.

I don't think we want to block this PR because it is using `getJetTypeFqName()`, I can help create a new issue to address this.

chao2zhang: https://github.com/detekt/detekt/issues/3605

chao2zhang (approved):
Let's keep in mind that we need to annotate this rule later with the new @RequiresTypeResolution #3579

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DoubleMutabilityForCollection.kt
cortinico: I would opt for merging this rule and following up on #3605 once solved.

3flex: If this is merged I'll address it in #3613

BraisGabin:
> Let's keep in mind that we need to annotate this rule later with the new @RequiresTypeResolution #3579

#3579 is merged already so I think that we should do this change and we can merge this. The `js` problem can be faced at #3613

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DoubleMutabilityForCollection.kt
chao2zhang: We need to update this PR since #3613 is merged.

chao2zhang: Never mind, I can submit another PR addressing this.

Issue #3554, opened 2021-03-15
SARIF report should only include active rules for RuleDescriptors


chao2zhang:
## Expected Behavior
SARIF report should only include active rules

## Observed Behavior
SARIF report includes all rules on the classpath  

## Context
This should help reduce the output size as well as confusion to readers of SARIF.

## Your Environment
* Version of detekt used: 1.16.0

BraisGabin:
Could you add some reference to the documentation?

BraisGabin:
@chao2zhang was this issue solved by #3556? What should be done here to close this?

chao2zhang:
This issue requires only printing resultDescriptors of rules that are active/enabled. I can take another look this week.

TWiStErRob:
Hmm, interesting, I just wanted to ask for the opposite. Don't understand why a sarif file needs to contain all rules that exist inside Detekt. For me, an empty report means there were no issues.
detekt.txt 0 bytes
detekt.xml 79 bytes (`<checkstyle />`)
detekt.html 2200 bytes (CSS + metrics)
detekt.sarif 76k (all rule definitions)

Now, imagine I have a project with 200 modules, each running Detekt. That's 15MB per CI build. Now let's have 30 developers, pushing 5-10 times a day, that's 3GBs worth of "empty" reports a day in artifacts.

So, would it be possible to only include rules that are Detekted in the report, or is that going against SARIF?

chao2zhang:
I believe it should
- Definitely not include all rules because of the waste of space
- Include enabled rules because those were active in that detekt run

If the size is a concern for you to adopt SARIF format, we can definitely add some flexibility of configuring how to generate lightweight SARIF or not.

TWiStErRob:
It's not an issue yet (we're not actively using Sarif yet), and thinking more about it, I think it might not be in the future, because we can easily remove a 200x multiplier:
```kotlin
tasks.register<io.gitlab.arturbosch.detekt.report.ReportMergeTask>("detektReportMergeSarif") {
	output.set(rootProject.buildDir.resolve("reports/detekt/merge.sarif"))
}
allprojects { project ->
	val detektReportMergeSarif = rootProject.tasks.named<ReportMergeTask>("detektReportMergeSarif")
	project.tasks.withType<Detekt> {
		finalizedBy(detektReportMergeSarif)
		detektReportMergeSarif.configure { input.from(this@withType.sarifReportFile) }
	}
}
```

[...]

PR #3555, opened 2021-03-15
Refactor Analyzer so that RuleSetProvider.instance is only called once

chao2zhang:
## Background
I WTF'ed when debugging https://github.com/detekt/detekt/pull/3515. Also, my colleague @kkoser WTF'ed when working on a custom detekt rule. 

## Solution
The problem here is that we are not only calling RuleSetProvider.instance() twice for each rule, but the first time to generate `IdMapping` is even wrong because the top-level config instead of `config.subConfig(it.ruleSetId)` is passed to each ruleSet.

To fix this, I have refactored the code to share `Sequence<>` to make sure `RuleSetProvider.instance()`.
(Technically, RuleSetProvider.instance() is called again in `:detekt-output-sarif/RuleDescriptors.kt`, but I am going to address this in #3554)

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
@@ -17,7 +17,6 @@ import io.gitlab.arturbosch.detekt.api.internal.whichOS
 import io.gitlab.arturbosch.detekt.core.config.DefaultConfig
 import io.gitlab.arturbosch.detekt.core.config.DisabledAutoCorrectConfig
 import io.gitlab.arturbosch.detekt.core.config.AllRulesConfig
-import io.gitlab.arturbosch.detekt.core.rules.IdMapping
 import io.gitlab.arturbosch.detekt.core.rules.associateRuleIdsToRuleSetIds
 import io.gitlab.arturbosch.detekt.core.rules.isActive
 import io.gitlab.arturbosch.detekt.core.rules.shouldAnalyzeFile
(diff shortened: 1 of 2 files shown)


BraisGabin:
> The problem here is that we are not only calling RuleSetProvider.instance() twice for each rule, but the first time to generate `IdMapping` is even wrong because the top-level config instead of `config.subConfig(it.ruleSetId)` is passed to each ruleSet.

Twice for each rule? We call it once per file and one more time when we instantiate the analyzer, right? And in this last one we don't care about the configuration because we just want to extract the ids. I imagine that this last part is not true for what you want to do with SARIF, right?

All those calls to `RuleSetProvider.instance()` are gone in the new analyzer. But the fact that you need to instantiate a rule just to get it's id is something that I don't like too much. Luckly the rules are light to instantiate and in the new analyzer they are instantiated just once.

BraisGabin (approved):

chao2zhang:
Agreed.

> All those calls to `RuleSetProvider.instance()` are gone in the new analyzer.
Could we push the new analyzer to a separate branch on master and start collaboration?

chao2zhang:
> Twice for each rule? We call it once per file and one more time when we instantiate the analyzer, right? And in this last one we don't care about the configuration because we just want to extract the ids. I imagine that this last part is not true for what you want to do with SARIF, right?

To make sure the idea is communicated properly:
- The first call is wrong https://github.com/detekt/detekt/pull/3555/files#diff-6149700bc8437bfd660774c9282c1051c9de2e5f7ea07711393cc17ca5944edeL38
- The second call is correct https://github.com/detekt/detekt/pull/3555/files#diff-6149700bc8437bfd660774c9282c1051c9de2e5f7ea07711393cc17ca5944edeL109
- The third call is inside SARIF report but that's out of the scope for this PR

schalkms (approved):

cortinico (approved):

chao2zhang:
@schalkms I am reverting this since this is for 1.17 release. We are probably waiting for Kotlin 1.4.32 so that we can release 1.16.1

schalkms:
@schalkms sorry again, my bad. Can you please reopen this changeset and tag it with a `blocked` label. Then we can prevent this things from happening in the future.

PR #3556, opened 2021-03-15
Fix rules not appearing in the sarif output

chao2zhang:
This partially addresses #3554. The problem is that all the detekt extensions are loaded by plugin loader, which also loads `SarifOutputReport`. However, `ServiceLoader.load(Class<>)` by default will with the current thread context's loader, which is the main detekt loader and will not load anything.

I would recommend merging this now instead of waiting for the next major release.

## Before detekt.sarif
```
{
  "$schema" : "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
  "version" : "2.1.0",
  "runs" : [ {
    "tool" : {
      "driver" : {
        "guid" : "022ca8c2-f6a2-4c95-b107-bb72c43263f3",
        "name" : "detekt",
        "organization" : "detekt",
        "fullName" : "detekt",
        "version" : "1.16.0",
        "semanticVersion" : "1.16.0",
        "downloadUri" : "https://github.com/detekt/detekt/releases/download/v1.16.0/detekt",
        "informationUri" : "https://detekt.github.io/detekt",
        "rules" : [ ],
        "language" : "en"
      }
    },
    "results" : [ ]
  } ]
}
```

## After detekt.sarif - SARIF validation passed
```{
  "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
  "version": "2.1.0",
  "runs": [
    {
      "results": [
      ],
      "tool": {
        "driver": {
          "downloadUri": "https://github.com/detekt/detekt/releases/download/v1.16.0/detekt",
          "fullName": "detekt",
          "guid": "022ca8c2-f6a2-4c95-b107-bb72c43263f3",
          "informationUri": "https://detekt.github.io/detekt",
          "language": "en",
          "name": "detekt",
          "organization": "detekt",
          "rules": [
            {
              "helpUri": "https://detekt.github.io/detekt/style.html#classordering",
              "id": "detekt.style.ClassOrdering",
              "name": "ClassOrdering",
              "shortDescription": {
                "text": "Class contents should be in this order: Property declarations/initializer blocks; secondary constructors; method declarations then companion objects."
              }
            },
            {
              "helpUri": "https://detekt.github.io/detekt/style.html#collapsibleifstatements",
              "id": "detekt.style.CollapsibleIfStatements",
              "name": "CollapsibleIfStatements",
              "shortDescription": {
[...]

PR #3557, opened 2021-03-15
Switch SARIF report off jackson

chao2zhang:
Size of detekt-report-sarif jar: 2.2MB -> 15KB

## Noticeable change
I was not able to find a JSONSchema -> Kotlin data class annotated with Kotlinx-serialization tool that preserves the order.
- https://github.com/wuseal/JsonToKotlinClass - I tried by it is crashing for my JSON schema. See https://github.com/wuseal/JsonToKotlinClass/issues/333
- https://quicktype.io/ This does allow unstandard `propertyOrder` but requires changing the JSON schema.

In the end, object JSON Schema should expect to have a unordered set of properties according to https://tools.ietf.org/html/draft-wright-json-schema-00#section-4.2, but as mentioned in https://github.com/detekt/detekt/issues/3045#issuecomment-711071231, it is still recommended to preserve order for streaming parsing consumers.

## TODO
Port #3556


BraisGabin (approved):

cortinico (approved):
I have little context on this part of the codebase, but the change looks sane to me 👍

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/report/SarifReportMergerSpec.kt
```diff
@@ -15,7 +15,8 @@ internal class SarifReportMergerSpec : Spek({
             SarifReportMerger.merge(listOf(input1, input2), output)
 
             val expectedOutput = resourceAsFile("output.sarif.json")
-            assertThat(output.readText()).isEqualTo(expectedOutput.readText())
+            assertThat(output.readText().replace("\\s".toRegex(), ""))
+                .isEqualTo(expectedOutput.readText().replace("\\s".toRegex(), ""))
```
marschwar, line 19: This seems a litte strange. Would `assertThat(output.readText()).isEqualToIgnoringWhitespace(expectedOutput.readText())` maybe also do the trick?

chao2zhang: That's a good point. I would try this one (Hopefully also convert other test cases to this one.)

Issue #3532, opened 2021-03-08
ClassOrdering false positive - inner/local extension var/val order


zayo:
Hi,

in our project, we tend to help ourselves with lot of private extension methods that serves the better readability.

For example
```kotlin
class Helper(private val intent: Intent) {
    private val isConnected: Boolean
        get() = !intent.isNoConnectivity

    public fun getPendingIntent(): PendingIntent {
          return intent.toPendingIntent()
    }
    
    private val Intent.isNoConnectivity: Boolean get() =
        getBooleanExtra(ConnectivityManager.EXTRA_NO_CONNECTIVITY, false)
}
```

For us, this (Intent.isNoConnectivity) is just idiomatic way of private methods that would otherwise mess up the code flow.

However, the new rule **ClassOrdering** forces these extensions to be ordered/mixed with private properties of target class. I believe it's a false positive, and every extension var/val should act more like method, taking in account only public/private at most, or be ignored completely.

* Version of detekt used: 1.16.0-RC3

chao2zhang:
Linking to the ktlint discussion (https://github.com/pinterest/ktlint/issues/309) on the difficulty of formatting tools to orchestrate class layout.

This request seems to go beyond the existing [Kotlin style guide](https://kotlinlang.org/docs/coding-conventions.html#class-layout), would you mind submitting a PR to detail your proposed changes to the guidelines? I am a little bit hesitant to modify the existing implementation of `ClassOrdering`.

cortinico:
> However, the new rule ClassOrdering forces these extensions to be ordered/mixed with private properties of target class

I don't see how extension fun/val should be considered an exception. You could also have a `private fun` that is logically coupled to a `public` fun/val and you want to keep it locally near.

The rationale behind this rule is to make sure your public API is ordered and appears on top of the class and not spread between `private` methods.

zayo:
In fact, the extension properties does not exist. You cannot modify the receiver object, nor append any field, as doc says.
In other words, the extension property inside class serves purpose of public/private fun, which means, they should be treated as public/private fun.

But they are expected to be sorted as private fields, which the are not.
I've edited original post, as this is not true extension funs, they are treated as normal funs. So only problematic part is the extension vals/vars

BraisGabin:
[...]

PR #3559, opened 2021-03-15
Fix [ClassOrdering] to treat extensions properties as functions.

zayo:
Proposal of fixing the #3532 issue.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
@@ -109,7 +109,7 @@ private val KtDeclaration.printableDeclaration: String
 @Suppress("MagicNumber")
 private val KtDeclaration.priority: Int?
     get() = when (this) {
-        is KtProperty -> 0
+        is KtProperty -> if (receiverTypeReference == null) 0 else 2 // Treat extension property as function.
         is KtClassInitializer -> 0
         is KtSecondaryConstructor -> 1
         is KtNamedFunction -> 2
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrderingSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrderingSpec.kt
@@ -194,7 +194,7 @@ class ClassOrderingSpec : Spek({
                 }
             """
 
-            assertThat(subject.compileAndLint(code)).hasSize(0)
+            assertThat(subject.compileAndLint(code)).isEmpty()
         }
 
         it("does report all issues in a class with multiple misorderings") {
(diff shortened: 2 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
```diff
@@ -109,7 +109,7 @@ private val KtDeclaration.printableDeclaration: String
 @Suppress("MagicNumber")
 private val KtDeclaration.priority: Int?
     get() = when (this) {
-        is KtProperty -> 0
+        is KtProperty -> if (receiverTypeReference == null) 0 else 2 // Treat extension property as function.
```
chao2zhang, line 112: I am worried about the intended change that we are going against the Kotlin Documentation. Would you mind adding this behavior change behind a config?

chao2zhang (commented):
Thank you for the contribution. Code looks great!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
schalkms: > I am worried about the intended change that we are going against the Kotlin Documentation. 

I think we shouldn't merge this due to this fact. This creates problems in the long run.

chao2zhang: Kotlin Documentation does accept change, although it make take a longer and a formal cycle to do it - https://github.com/JetBrains/kotlin-web-site/edit/master/docs/topics/coding-conventions.md

In parallel, we can add something as `@controversial`.

zayo: I was writing the change (it suddenly stopped building for no reason, I don't have that patience). Then I realised I don't want to do it.

By adding a flag allowing this explicitly we would say, that the original buggy rule  is actually correct, which is not, I presume.
There is no single test case for any extension, so I doubt this is intentional and it was thing through.

Also you still try to tell me this is wrong, but I never saw any argument for that (Depends whether we call this a bug fix, or changing of solution). I don't see any contradiction to the doc.
As I posted earlier in the issue: https://kotlinlang.org/docs/extensions.html#companion-object-extensions says that extension properties cannot have a backing field (so not a property - does not hold any state), acts like method. @JvmField impossible too. So I see this pretty much clearly. Putting here a flag to explicitly enable it would be the mistake for me, and it's something I don't trust to be signed under.

The only change I would consider is to not treat extensions as function, and make them to be independent of order. In fact, it translates to some kind of inner class, which are not required to be ordered.
What do you think guys?

chao2zhang: I agree with the reasoning of #3532 and this PR, it is just that we have two conflicting principles:
- Detekt should not go against Kotlin documentation.
- Detekt should improve the code quality.

If you would like to merge in this PR immediately, my suggestion would be to add a configuration.
If you would like to avoid adding configurations, my suggestion would be to publish a PR to edit `https://github.com/JetBrains/kotlin-web-site/edit/master/docs/topics/coding-conventions.md` to explicitly state that extension property should be in `Method declarations`. Once that is merged, we can approve this PR.

zayo: In fact I don't see this going against official doc. And not forcing extension properties to be mixed with properties sounds like an improvement to me.

But yes, I understand this is based on just presumption, not explicitely pointed fact. I'm not in a hurry. So I would choose to make this clear in doc and update detekt accordingly. I will take a look at the doc.

This PR anyway supposed to be just the proposal asked in the original issue :)

schalkms: I agree with this recommendation.

> * If you would like to avoid adding configurations, my suggestion would be to publish a PR to edit `https://github.com/JetBrains/kotlin-web-site/edit/master/docs/topics/coding-conventions.md` to explicitly state that extension property should be in Method declarations. Once that is merged, we can approve this PR.

cortinico:
What's the status of this PR?

schalkms:
> What's the status of this PR?

Please take a look at the discussion above.
https://github.com/detekt/detekt/pull/3559#pullrequestreview-612420252

I guess we should close this PR and reconsider it whenever the official Kotlin documentation gets updated in that regard. As a side note, I already closed the linked issue with the same argument.

PR #3564, opened 2021-03-17
UnnecessaryApply: fix false positive when lambda has multiple member references

### LINKED ISSUE #3561: Unnecessary Apply rule being applied for cases when apply is needed
Hi,

I found discovered an issue regarding adding the UnnecessaryApply rule. 

It appears that this rule correctly covers the most basic cases where an apply block is used for only one statement such as the following:

`textview.apply { visibility = View.VISIBLE }`.

However, I noticed this is also flagging cases like the following, where the apply actually is needed multiple times but is only used in that manner in a nested block:

```
textview.apply {
    post {
        visibility = View.VISIBLE
        text = "Hello world!"
    }
}
```

The block above is flagged under this rule but the apply is actually being used inside the post block.

Looking at the commits where this rule was added: https://github.com/detekt/detekt/pull/1229, the rule seems to be only checking if the body of the outer lambda has one statement, but doesn't inspect anything nested within that statement

t-kameyama:
Fixes #3561


cortinico (approved):

BraisGabin:
Can we wait a bit to merge this, I'm not so sure about this issue. It seems like a true positive to me.

BraisGabin:
What do you think about what I said in the issue?

t-kameyama:
```kotlin
class Foo {
    fun bar(f: () -> Unit) {}
    var baz = 1
}

fun test(foo: Foo) {
    // case 1
    foo.apply {
        println(baz)
        println(baz)
    }

    // case 2
    foo.apply {
        bar {
            println(baz)
            println(baz)
        }
    }
}
```

Since it's not reported in case 1, I don't think it should be reported in case 2 as well. I understand that this rule does not flag when lambda has multiple member references.

BraisGabin:
Maybe we should update it and make it flag the first case too. Promoting `with` or `let`

t-kameyama:
> Maybe we should update it and make it flag the first case too. Promoting `with` or `let`

I don't think so. The first case can be replaced with `let`, but it's not an **unnecessary** `apply`.

BraisGabin (approved):
> Since it's not flagged in case 1, I don't think it should be flagged in case 2 as well. I understand that this rule does not flag when lambda has multiple member references.

I agree with you here. The problem that I'm talking about is already in the rule so this PR just make it more consistent so we should merge this. I think that we need to rethink a bit this rule but that's another topic 👍.

PR #3569, opened 2021-03-17
Show information from @active in the website

### LINKED ISSUE #3568: Add more info about rules to website
It would be great if the website would display the following info about the rules:

 * Is it on by default
 * Which version it was introduced in

Right now, the first info is only found in the default detekt configuration file on the repo, and the latter is not available easily anywhere. It's a bit of a pain.

cortinico:
Fixes #3568

The information from the `@active` KDoc annotation can be easily added to the website to make easier for our user to understand if a rule is active or not.

I've also cleaned up a bit the `@since` annotation as it was not really used.


detekt-generator/src/test/resources/RuleSet.md
```diff
@@ -36,10 +38,14 @@ import foo.bar
 
 equals null
 
+**Active**: No
+
 ### NoUnitKeyword
 
 removes :Unit
 
+**Active**: Yes - Since v1.16.0
```
BraisGabin, line 47: Should we make it more clear that this is a default value? I mean, that this rule is active by default.

BraisGabin (approved):

detekt-generator/src/test/resources/RuleSet.md
cortinico: How about the label it as "Active by default"?

BraisGabin: Add it just for the ones that are active by default? I like it

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullableToStringCall.kt
```diff
@@ -47,7 +47,6 @@ import org.jetbrains.kotlin.types.isNullable
  * }
  * </compliant>
  *
- * @since 1.11.0
```
schalkms, line 50: I guess those removed `active` tags are all wrong, right?

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
```diff
@@ -27,6 +27,7 @@ internal class RuleVisitor : DetektVisitor() {
     private var compliant = ""
     private var name = ""
     private var active = false
+    private var activeSince: String? = null
```
schalkms, line 30: Wouldn't it be better to introduce a `SemanticVersion` class like in [C#](https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.semanticversion?view=powershellsdk-7.0.0) instead of using a nullable string. We could then also spare the necessary null-checks.

cortinico: Given that this is just for visualization, a `String` should be enough no?

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullableToStringCall.kt
cortinico: Yes I cross checked with the config file

detekt-generator/src/test/resources/RuleSet.md
cortinico: I've added it for both actually. I think it's better to be explicit rather than implicit, or is there a reason why we want to hide it for disabled rules?

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
schalkms: Yep 👍 - okay for now

schalkms (approved):

detekt-generator/src/test/resources/RuleSet.md
cortinico: Going to merge this as it is, and we can still re-discuss this if needed.

Issue #3567, opened 2021-03-17
Weird warning recommending to use undocumented options


lwasyl:
## Expected Behavior
Build doesn't emit warnings or the warnings are **_easily_** actionable

## Observed Behavior
I see the following warning:
> 'failFast' is deprecated. Please use 'buildOnDefaultConfig' together with 'allRules'.

and it's not clear what to do about it.

## Steps to Reproduce
<details>
<summary>Detekt Gradle Plugin configuration:</summary>

```kotlin
configure<DetektExtension> {
    failFast = true // Enables all rules by default
    reports {
        xml.enabled = false
        html.enabled = false
    }
    input = files("src/")
    config = rootProject.files("quality/detekt.yml")
}
```
</details>

<details>
<summary>Detekt configuration file</summary>

```yaml
build:
  maxIssues: 0

processors:
  active: false

console-reports:
  active: true
  exclude:
    - 'ProjectStatisticsReport'
    - 'ComplexityReport'
    - 'NotificationReport'

comments:
  active: false

complexity:
  active: false

empty-blocks:
  active: true

exceptions:
  active: true
  NotImplementedDeclaration:
    active: false
  RethrowCaughtException:
    active: false
  ReturnFromFinally:
    active: false
  SwallowedException:
    active: false
  ThrowingExceptionsWithoutMessageOrCause:
    active: false
  ThrowingNewInstanceOfSameException:
    active: false
  TooGenericExceptionCaught:
    active: false
  TooGenericExceptionThrown:
    active: true
    exceptionNames:
      - Error
      - Exception
      - Throwable
      - RuntimeException

naming:
  active: true
  ForbiddenClassName:
    active: true
    forbiddenName: [ "^.*Impl$" ]
  FunctionMaxLength:
    active: false

performance:
  active: true
  SpreadOperator:
    active: false

potential-bugs:
  active: true
  CastToNullableType:
    active: false
  LateinitUsage:
    active: false
  UnsafeCast:
    active: false
  EqualsAlwaysReturnsTrueOrFalse:
    active: true

style:
  active: true
  CollapsibleIfStatements:
    active: false
  ClassOrdering:
    active: false
  DataClassContainsFunctions:
    active: false
  DataClassShouldBeImmutable:
    active: false
  DestructuringDeclarationWithTooManyEntries:
    excludes: [ '**/test/**' ]
  ExpressionBodySyntax:
    active: true
    includeLineWrapping: true
  ForbiddenComment:
    active: false
[...]

PR #3572, opened 2021-03-18
fix error message -> buildUponDefaultConfig instead of buildOnDefaultConfig

marandaneto:
when migrating detekt version, I came thru this error and the field name is likely wrong.


cortinico (approved):
> when migrating detekt version, I came thru this error and the field name is likely wrong.

Thanks for fixing this (related to #3567).
Can you just split the string as the line is getting to long (that's why the CI is failing).

marandaneto:
done, just a quick heads up, `detektFormat` ran but didn't format anything, so I did manually.

schalkms (approved):

chao2zhang (approved):

PR #3574, opened 2021-03-18
Support Gradle config cache in detekt's build

3flex:
A few improvements to eventually allow the Gradle configuration cache to be used when building detekt.

There are changes required in some Gradle plugins used by the detekt build before this will work:
* Kotlin 1.4.30 or higher
* Gradle 7.0 (or 6.8.4 which should include a backported fix for an issue)
* https://github.com/Kotlin/dokka/issues/1217 to be fixed
* https://github.com/Kotlin/binary-compatibility-validator/pull/50 to be merged & released

### DIFF
--- a/buildSrc/src/main/kotlin/module.gradle.kts
+++ b/buildSrc/src/main/kotlin/module.gradle.kts
@@ -6,6 +6,7 @@ plugins {
     kotlin("jvm")
     `maven-publish`
     jacoco
+    id("detekt")
 }
 
 // bundle detekt's version for all jars to use it at runtime
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.2-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.3-all.zip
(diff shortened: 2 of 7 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.2-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.3-bin.zip
```
chao2zhang, line 3: I assume we want `-all` so that we can check the gradle code easily.

buildSrc/src/main/kotlin/module.gradle.kts
```diff
@@ -6,6 +6,7 @@ plugins {
     kotlin("jvm")
     `maven-publish`
     jacoco
+    id("detekt")
```
chao2zhang, line 9: This is great!

chao2zhang (commented):
Did you mean to turn on `org.gradle.unsafe.configuration-cache=true`? Or is this PR purely to fix issues to be prepared for the work?

3flex:
It's just to fix issues in the build for now. As I mentioned there are third party Gradle plugin updates that are required first, but once they've made fixes and detekt updates those dependencies it should be OK to enable.

chao2zhang (approved):

cortinico (approved):

BraisGabin:
@3flex should we have an issue to track the things that we need to do?

PR #3578, opened 2021-03-19
Revert "Refactor Analyzer so that RuleSetProvider.instance is only called once"

chao2zhang:
Reverts detekt/detekt#3555


chao2zhang:
The original PR is intended for 1.17+

schalkms (commented):
Sorry my bad, I overlooked that.

Issue #3562, opened 2021-03-15
Use actual Kotlin annotation class to provide metadata to rules


chao2zhang:
We currently have `@configuration`, `@active`, and `@requiresTypeResolution` inside comment to provide metadata, and we are about to add `@controvertial` according to https://github.com/detekt/detekt/issues/3560.

Annotating comments to annotate rules is acceptable, but to further improve the quality of these metadata, we should use Kotlin annotation class to enforce information, such as `@Active(since = "1.16.0")` or `@Configuration(type = String::class)`. 

The benefits:
- Improve code review quality since we do not need to ask contributors to annotate according to a given format
- Facilitate custom detekt rules to check implementations, such as enforcing `@AutoCorrectible` rules to invoke mutation methods. (Technically `@requiresTypeResolution` should be gone in https://github.com/detekt/detekt/pull/3127/)
- Allow clients, such as gradle plugins, to configure rules based on the annotation (This PR is probably part of https://github.com/detekt/detekt/pull/3002)

schalkms:
That's a good idea. The generated stays the same. The rules as well as the generator have to be changed.

cortinico:
+1 for me. Does it also makes sense to expose them as part of our public API? 
Users that interact with out tools could filter or get more insights on rule metadata directly from the code.

chao2zhang:
> Does it also makes sense to expose them as part of our public API?
Definitely.

marschwar:
I also really like the idea. Are you already working on it? Were you also thinking about generating code from those annotations? For example it should be fairly simple to generate a type safe configuration class from the annotations as well. For example:

```kotlin
@Rule(
    since = "1.1.0",
    activeByDefault = true,
    activeByDefaultSince = "1.15.0",
    config = Config(
        ints = [
            IntConfig("threshold", defaultValue = DEFAULT_DUPLICATION),
        ],
        strings = [
            StringConfig("ignoreStringsRegex", defaultValue = IGNORE_STRINGS_REGEX_DEFAULT),
        ],
        booleans = [
            BooleanConfig("ignoreAnnotation", defaultValue = true),
            BooleanConfig("excludeStringsWithLessThan5Characters", defaultValue = true),
        ]
    )
)
class StringLiteralDuplication() ...
```

could be used to generate 

```kotlin
data class StringLiteralDuplicationConfig(
    val threshold: Int = 3,
    val ignoreStringsRegex: String = "$^",
    val ignoreAnnotation: Boolean = true,
[...]

Issue #3577, opened 2021-03-19
Detect type resolving rules run without type resolution classpath set


chao2zhang:
This is raised from https://github.com/detekt/detekt/issues/3537#issuecomment-795846800. 
1. If the type resolution classpath is not set, but there are active type resolving rules, we should print out warnings to remind users.
2. We can also skip type resolving rules if type resolution classpath is not set.
3. We probably need to wait for #3562 to be implemented.

cortinico:
> We can also skip type resolving rules if type resolution classpath is not set.

We're basically already doing it, no?

chao2zhang:
What I was looking for `Rule.isActive()` should be false for those type resolving rules without classpath set. In SARIF output, we should show the active rules run.

cortinico:
> 1. we should print out warnings to remind users.

I'm unsure about this. We might have users that are fine having rules enabled only on `detektMain` and not on `detekt`. Perhaps the warning should be configurable?

I'm also wondering what was the original use case of this feature request.

cortinico:
> 1. If the type resolution classpath is not set, but there are active type resolving rules, we should print out warnings to remind users.

Following up here: we could add a warning when we do the _if bindingContext is empty_ check that instead of just returning it's printing something like "Rule xyz was skipped because Type Resolution is not available".

eygraber:
> I'm unsure about this. We might have users that are fine having rules enabled only on detektMain and not on detekt. Perhaps the warning should be configurable?

Yes please :smile: 

My detekt runs just got very noisy with 1.22.0-RC1. I have one config file that contains all the enabled rules for my project, but I sometimes run detekt without type resolution.

BraisGabin:
Could you open this as a new issue? You can link there this issue to haver better context.

eygraber:
https://github.com/detekt/detekt/issues/5334

PR #3579, opened 2021-03-19
Replace @requiresTypeResolution kdoc tag with @RequiresTypeResolution annotation

marschwar:
This PR belongs to #3562 and replaces the @requiresTypeResolution kdoc tag with a @RequiresTypeResolution annotation.

There are a few things that I would like to point out/discuss:
* I still have some formatting issues. That is why there is one commit that just reformats the code according to the ktlint plugin
* I added the annotation to the detekt API which might be a bit premature. Maybe it should be placed in `internal` until it is stable?
* I started out adding `@UnstableApi` to the annotation but that would all rules to require to add `@OptIn(UnstableApi::class)` which does not feel right either.
* Adding a `@Suppress("TooManyFunctions")` to the `RuleVisitor` does not feel great, but I thinkt the code reads much better with those extension functions. Suggestions are welcome.

### DIFF
--- a/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
+++ b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
@@ -2,6 +2,7 @@ package io.gitlab.arturbosch.detekt.generator.collection
 
 import io.gitlab.arturbosch.detekt.api.Debt
 import io.gitlab.arturbosch.detekt.api.DetektVisitor
+import io.gitlab.arturbosch.detekt.api.RequiresTypeResolution
 import io.gitlab.arturbosch.detekt.api.ThresholdRule
 import io.gitlab.arturbosch.detekt.formatting.FormattingRule
 import io.gitlab.arturbosch.detekt.generator.collection.exception.InvalidAliasesDeclaration
@@ -17,7 +18,9 @@ import org.jetbrains.kotlin.psi.KtValueArgument
 import org.jetbrains.kotlin.psi.psiUtil.containingClass
 import org.jetbrains.kotlin.psi.psiUtil.getSuperNames
 import java.lang.reflect.Modifier
+import kotlin.reflect.KClass
 
+@Suppress("TooManyFunctions")
 internal class RuleVisitor : DetektVisitor() {
 
     val containsRule
@@ -90,7 +93,7 @@ internal class RuleVisitor : DetektVisitor() {
             throw InvalidDocumentationException("KDoc for rule $name must not contain tabs")
         }
 
-        active = classOrObject.kDocSection()?.findTagByName(TAG_ACTIVE) != null
+        active = classOrObject.hasKDocTag(TAG_ACTIVE)
         val activeTagContent = classOrObject.kDocSection()
             ?.findTagByName(TAG_ACTIVE)
             ?.getContent()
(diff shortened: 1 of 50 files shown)


detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
```diff
@@ -17,7 +18,9 @@ import org.jetbrains.kotlin.psi.KtValueArgument
 import org.jetbrains.kotlin.psi.psiUtil.containingClass
 import org.jetbrains.kotlin.psi.psiUtil.getSuperNames
 import java.lang.reflect.Modifier
+import kotlin.reflect.KClass
 
+@Suppress("TooManyFunctions")
```
chao2zhang, line 23: Let's move the methods of `KtClassOrObject.requiresTypeResolution()` and `KtClassOrObject.isAnnotatedWith()` to `detekt-api`. I think they should be exposed and can be reused.

Hopefully we won't require suppressing `TooManyFunctions`

chao2zhang (approved):

marschwar:
I removed the annotation from the exposed api by moving it to the internal package. Also the retention is changed to SOURCE so it is not available at runtime. This way we should be fairly flexible when it comes to changing the annotation(s) without breaking peoples code.

cortinico (approved):
> I still have some formatting issues. That is why there is one commit that just reformats the code according to the ktlint plugin

I wonder why this is happening 🤔 

> I started out adding @UnstableApi to the annotation but that would all rules to require to add @OptIn(UnstableApi::class) which does not feel right either.

I think `@UnstableApi` is a bit of an overkill. Having the annotation inside `.internal` is sufficient. Also I don't know how many users will start relying on this annotation right after, so I won't worry too much about protecting this API.

chao2zhang (approved):
@cortinico I was wondering if this should be in external API. The problem is that we should also allow custom rules to be annotated correctly, #3577 is a good example to fully skip type resolution rules.

Regardless, this PR looks good as it stands for now. I don't mind if we "escalate" these apis later.

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleCollectorSpec.kt
```diff
@@ -203,6 +203,32 @@ class RuleCollectorSpec : Spek({
...
+                /**
+                 * description
+                 */
+                @io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
+                class SomeRandomClass : Rule
+            """
+            val items = subject.run(code)
+            assertThat(items[0].requiresTypeResolution).isTrue()
+        }
+
```
BraisGabin, line 231: Are we lacking a test to check that this value is false if we don't have the annotation?

BraisGabin (approved):

cortinico:
> @cortinico I was wondering if this should be in external API. The problem is that we should also allow custom rules to be annotated correctly

I believe yes, this should be an external API. Something to remember for the future 👍

chao2zhang (approved):

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleCollectorSpec.kt
chao2zhang: I can follow up with the test. This PR has been open for a while so I would merge this in as we are adding more annotations.

PR #3581, opened 2021-03-19
Format test code for RedundantSuspendModifierSpec

schalkms:


3flex (approved):
It's odd that coverage improved, though not a bad thing!

PR #3582, opened 2021-03-20
Replace @since KDoc tag with @SinceDetekt

marschwar:
This also belongs to #3562 and replaces the rarely used `@since` KDoc tag with a `@SinceDetekt` annotation. Name and structure are "inspired by" `@kotlin.SinceKotlin`.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/SinceDetekt.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/SinceDetekt.kt
@@ -1,0 +1,11 @@
+package io.gitlab.arturbosch.detekt.api.internal
+
+/**
+ * Specifies the first version of Detekt where a declaration has appeared.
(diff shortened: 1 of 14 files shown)


cortinico:
Can I ask you to rebase @marschwar?
Specifically the `@since` annotation was addressed in #3569. It was a stale annotation that clashed with the `@active` annotation we already have, so I've cleaned it up.

marschwar:
What do you mean by stale? If the `@since` tag should not be used, then I can just delete this PR. I was thinking there should be a `@SinceDetekt` and a `@ActiveByDefault` annotation. But maybe there is only the need for the latter. Let me know what you think.

cortinico:
> What do you mean by stale?

We had a couple of occurrences and we were not really using it (e.g. in 1.16.0 we enabled several rules but never really annotated those with `@since`).

I belive we should have an `@ActiveByDefault` with a `since : String` parameter.

marschwar:
Ok. I'll close this one.

PR #3592, opened 2021-03-22
Replace `@active` KDoc tag with `@ActiveByDefault` annotation

marschwar:
Following up on #3562 this replaces the `@active¸ kdoc tag with an `@ActiveByDefault` annotation.


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ActiveByDefault.kt
```diff
@@ -0,0 +1,15 @@
...
+ * is active by default.
+ */
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.SOURCE)
+annotation class ActiveByDefault(
+    /**
+     * The Detekt version since when the rule is active by default. This can be left blank
+     * if activation version is not known.
+     */
+    val since: String = ""
```
BraisGabin, line 14: What do you think about having it null by default?

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingProvider.kt
```diff
@@ -16,8 +17,8 @@ import io.gitlab.arturbosch.detekt.api.RuleSetProvider
  *
  * @configuration android - if android style guides should be preferred (default: `false`)
  * @configuration autoCorrect - if rules should auto correct style violation (default: `true`)
- * @active since v1.0.0
  */
+@ActiveByDefault("v1.0.0")
```
BraisGabin, line 21: What do you think about `1.0.0` isntead of `v1.0.0`. If then we want to add the `v` in the documentation we can do it.

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleCollectorSpec.kt
```diff
@@ -86,66 +86,68 @@ class RuleCollectorSpec : Spek({
...
         }
 
-        it("is active tag present with since") {
+        it("is active by default with blank since") {
             val code = """
                 /**
                  * description
-                 * @active whatever
                  */
+                @ActiveByDefault("  ")
```
BraisGabin, line 120: This should fail, don't you think?

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleCollectorSpec.kt
```diff
@@ -86,66 +86,68 @@ class RuleCollectorSpec : Spek({
...
         }
 
-        it("is active tag present with since") {
+        it("is active by default with since") {
             val code = """
                 /**
                  * description
-                 * @active since v1.2.3
                  */
+                @ActiveByDefault("v1.2.3")
```
BraisGabin, line 107: Would it be valuable to have a regex to ensure that we don't add a wrong value?

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
```diff
@@ -37,9 +38,8 @@ import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
  * </compliant>
  *
  * @configuration threshold - the number of conditions which will trigger the rule (default: `4`)
- *
- * @active since v1.0.0
  */
+@ActiveByDefault("v1.0.0")
```
cortinico, line 42: Commenting here but valid for the whole PR:
Can you use named parameters? That would make it clear that the string param is the the _since_ version.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ActiveByDefault.kt
marschwar: annotation may not have a nullable parameter. But I do agree, that we should not provide a default value (now that I know that all active tags actually have versions).

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingProvider.kt
marschwar: Great point

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
marschwar: Will do

BraisGabin (approved):

marschwar:
Just wondering - do we need to still support the kdoc tag as well in order to work for custom rules that have not made the switch?

cortinico:
> Just wondering - do we need to still support the kdoc tag as well in order to work for custom rules that have not made the switch?

To my knowledge, not: the `detekt-generator` is the only part of the code where we effectively read the KDoc annotations and we don't generated docs for external rules.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ActiveByDefault.kt
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Annotated [io.gitlab.arturbosch.detekt.api.Rule] or [io.gitlab.arturbosch.detekt.api.RuleSetProvider]
+ * is active by default.
+ */
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.SOURCE)
+annotation class ActiveByDefault(
+    /**
+     *  The Detekt version the rule was actived by default in the following format: <major>.<minor> or <major>.<minor>,
```
chao2zhang, line 11:
```suggestion
     *  The Detekt version the rule was actived by default in the following format: <major>.<minor> or <major>.<minor>.<patch>,
```

chao2zhang (approved):

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ActiveByDefault.kt
marschwar: Actually it was supposed to be `<major>.<minor>.<patch>` ony

PR #3595, opened 2021-03-22
Support factory exemption for generic classes

### LINKED ISSUE #3591: MemberNameEqualsClassName factory function exemption doesn't work on generics
## Expected Behavior
The `MemberNameEqualsClassName` rule is documented as `Factory functions that create an instance of the class are exempt from this rule.`  Would expect this exemption to apply even if the class in question has generics.

## Observed Behavior
A generic class with a generic factory function is reported as violating the rule.

## Steps to Reproduce
This code triggers the rule:
```kotlin
data class FieldInfo<T>(
    val alias: String,
    val extractor: (String) -> T
) {
    companion object {
        // this complains about MemberNameEqualsClassName
        fun <T> fieldInfo(
            alias: String,
            extractor: (String) -> T
        ): FieldInfo<T> {
            return FieldInfo(alias, extractor)
        }
    }
}
```
Removing the generic does not trigger the rule:
```kotlin
data class FieldInfo(
    val alias: String,
    val extractor: (String) -> String
) {
    companion object {
        // this does not complain about MemberNameEqualsClassName
        fun fieldInfo(
            alias: String,
            extractor: (String) -> String
        ): FieldInfo {
            return FieldInfo(alias, extractor)
        }
    }
}
```

## Context
The above example is a trivial reproduction, in our real world use case we have inline factory functions with reified type parameters that provide default `extractor` implementations. (eg `inline fun <reified T> fieldInfo(alias: String) = FieldInfo(alias, defaultImpl())` where `defaultImpl` is an inline function that returns different implementations based on T::class)

## Your Environment

* Version of detekt used:  1.16.0
* Version of Gradle used (if applicable): n/a, maven/antrun config here: https://github.com/trib3/leakycauldron/blob/main/parent-pom/pom.xml#L1469-L1505
* Operating System and version: ubuntu 21.04
* Link to your project (if it's a public repository): n/a, but public maven parent pom linked above

josephlbarnett:
Compare the referenced type name instead of the fully
qualified generic type name when possible.

Fixes #3591

### DIFF
--- a/detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassNameSpec.kt
+++ b/detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassNameSpec.kt
@@ -237,6 +237,22 @@ class MemberNameEqualsClassNameSpec : Spek({
                 assertThat(MemberNameEqualsClassName().compileAndLintWithContext(env, code)).isEmpty()
             }
 
+            it("doesn't report a generic factory function") {
+                val code = """
+                    data class GenericClass<T>(val wrapped: T) {
+                        companion object {
+                            fun <T> genericClass(wrapped: T): GenericClass<T> {
+                                return GenericClass(wrapped)
+                            }
+                            fun genericClass(): GenericClass<String> {
+                                return GenericClass("wrapped")
+                            }
+                        }
+                    }
+                """
+                assertThat(MemberNameEqualsClassName().compileAndLintWithContext(env, code)).isEmpty()
+            }
+
             it("doesn't report a body-less factory function") {
                 val code = """
                     open class A {
--- a/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassName.kt
+++ b/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassName.kt
@@ -17,6 +17,7 @@ import org.jetbrains.kotlin.psi.KtDeclaration
 import org.jetbrains.kotlin.psi.KtNamedDeclaration
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtObjectDeclaration
+import org.jetbrains.kotlin.psi.KtUserType
 import org.jetbrains.kotlin.resolve.BindingContext
 
 /**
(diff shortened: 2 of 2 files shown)


josephlbarnett:
There may be a better way to do this, as I don't fully understand the kotlin compiler psi APIs, but this works for the test case included and falls back to the previous implementation for other cases.

cortinico (approved):

detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassNameSpec.kt
```diff
@@ -237,6 +237,19 @@ class MemberNameEqualsClassNameSpec : Spek({
                 assertThat(MemberNameEqualsClassName().compileAndLintWithContext(env, code)).isEmpty()
             }
 
+            it("doesn't report a generic factory function") {
+                val code = """
+                    data class GenericClass<T>(val wrapped: T) {
+                        companion object {
+                            fun <T> genericClass(wrapped: T): GenericClass<T> {
```
BraisGabin, line 244: In this test is not clear if the problem was the generic in the class or the generic in the function. Could you simplify it? Or, if they are both, create two tests.

BraisGabin (approved):

detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassNameSpec.kt
josephlbarnett: Added a non-generic factory function to the test, does that help?

BraisGabin: Which is the problem exactly this code:

```kotlin
                    data class GenericClass(val wrapped: Any) {
                        companion object {
                            fun <T> genericClass(wrapped: T): GenericClass<T> {
                                return GenericClass(wrapped)
                            }
                        }
                    }
```

or this:


```kotlin
                    data class GenericClass<T>(val wrapped: T) {
                        companion object {
                            fun genericClass(wrapped: Any): GenericClass<Any> {
                                return GenericClass(wrapped)
                            }
                        }
                    }
```

josephlbarnett: the top one is closer to real world usage, but i suppose the bottom one is the minimal example?  both fail without the change and pass with it, fwiw.

BraisGabin (approved):

chao2zhang (approved):

Issue #3528, opened 2021-03-08
suggest Single Abstract Method


zmunm:
## Expected Behavior of the rule

SAM conversion is possible, but I would like to find out where I am creating an anonymous class.

java interface
```java
public interface Foo {
    void bar();
}
```

Written
```kotlin
val listener = object : Foo {
    override fun bar() {
         ...
    }
}
```

What I want

```kotlin
val listener = Foo {
    ...
}
```

## Context

https://kotlinlang.org/docs/java-interop.html#sam-conversions

chao2zhang:
Hello @zmunm, this is a good idea since IntelliJ/Android Studio has the inspection to convert to the SAM syntax "Redundant SAM constructor - This inspection reports SAM constructor usages which can be replaced with lambdas". If you would like to contribute, please let us know.

zmunm:
It may take a while, but I want to try :)

cortinico:
I'm assigning this to you @zmunm 👍 Feel free to let us know if you need further support.

zmunm:
I have met some edge cases and have had a hard time. 😞 

TL; DR
Because of the default implementation of the interface, it is difficult to check if it is a SAM by looking at only the file.
Very close example is [here](https://github.com/detekt/detekt/blob/master/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/TopLevelAutoCorrectSpec.kt)

---
Long description

There is a syntax that looks like SAM. But not always.
```kotlin
object: Sam {
    override fun foo() {
    }
}
```


non compliant java interface
```java
interface Sam {
    default void foo() {}
}
```


non compliant kotlin interface
```kotlin
interface Sam {
    fun foo() {}
}
```


compliant java interface
```java
interface Sam {
    void foo();
    default void bar() {}
}
```


compliant kotlin interface
```kotlin
fun interface Sam {
    fun foo()
    fun bar() {}
}
```

Is there a good way to check this?

cortinico:
> Is there a good way to check this?

I'm unsure as I never faced such scenario.
Ideally you want to access the `Sam` interface during the inspection and check if it's a SAM interface or not 🤔

zmunm:
@cortinico 
It seems that I can try to check the package and import by iterating through the `Class.forName`.
However, I doubt that using reflect is acceptable for detekt.

+++
There were a few more cases, and I think it's hard to see without reflect.

1. overloading
```kotlin
interface A {
    fun foo(a: Boolean = true)
[...]

PR #3599, opened 2021-03-23
New Rule: ObjectLiteralToLambda

zmunm:
apply #3528 

I hope there isn't much I've missed.

Thank you @cortinico


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambda.kt
```diff
@@ -0,0 +1,74 @@
...
+ *     }
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * Foo {
+ * }
+ * </compliant>
+ *
+ * @active since v1.17.0
```
BraisGabin, line 37: Don't activate it by default please.

zmunm: Thank you. Are there any criteria to set the rule to be enabled by default?

BraisGabin: We never enable a rule by default the first time we introduced it because it can have false positives and we will impact to a lot of users. We prefer to enable it later if we think that its a rule that nearly all our user base will like to have it. But we don't have any defined rule here.

BraisGabin (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambda.kt
```diff
@@ -0,0 +1,74 @@
...
+ *     }
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * Foo {
+ * }
+ * </compliant>
+ */
+class ObjectLiteralToLambda(config: Config = Config.empty) : Rule(config) {
```
cortinico, line 37: You should add the `@RequiresTypeResolution` annotation here.

cortinico (commented):
Nice test coverage 👍

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambda.kt
```diff
@@ -0,0 +1,76 @@
...
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Style,
+        "Report object literals that can be changed to lambdas.",
+        Debt.FIVE_MINS
+    )
+
+    private val KotlinType.isSamInterface
```
chao2zhang, line 47:
```suggestion
    private val KotlinType.couldBeSamInterface
```

According to its [original documentation](
https://github.com/JetBrains/kotlin/blob/master/core/descriptors/src/org/jetbrains/kotlin/descriptors/ClassDescriptor.java#L112), `isDefinitelyNotSamInterface` may return falsely false

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambda.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.jetbrains.kotlin.psi.KtObjectLiteralExpression
+import org.jetbrains.kotlin.psi.KtSuperTypeEntry
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.types.KotlinType
+
+/**
+ * An anonymous object that does nothing other than the implementation of a single method
+ * can be used as a lambda.
+ *
+ * See https://kotlinlang.org/docs/fun-interfaces.html
```
chao2zhang, line 23: Would this rule detect findings when the consuming code is Kotlin 1.3 or 1.3-?

chao2zhang: (It may be good at least to call it out in the documentation)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambdaSpec.kt
```diff
@@ -0,0 +1,347 @@
...
+
+class ObjectLiteralToLambdaSpec : Spek({
+    setupKotlinEnvironment()
+
+    val env: KotlinCoreEnvironment by memoized()
+    val subject by memoized { ObjectLiteralToLambda() }
+
+    describe("ObjectLiteralToLambda rule") {
+
+        context("report convertible expression") {
```
chao2zhang, line 20: Would you mind adding a test case like the following? Note: This is from real production code that I found is not possible to use SAM Interface. This can be a follow-up PR since this may require more research or changes.
```
fun <T> LiveData<Resource<T>>.observeUntilFinished(observer: Observer<Resource<T>>? = null) {
    // Can't use SAM we are using implicit `this` receiver
    observeForever(object : Observer<Resource<T>> {
        override fun onChanged(resource: Resource<T>?) {
            observer?.onChanged(resource)
            if (resource?.status == Status.SUCCESS || resource?.status == Status.ERROR) {
                removeObserver(this)
            }
        }
    })
}
```

chao2zhang:
Actually, I also find that converting to SAM may indicate anonymous object converted to singleton:  https://github.com/JetBrains/kotlin/commit/99a6bdeec7f052b180d859f59cbc5829ab6e9034

I guess this is an edge case but we should at least call it out. (We can soon have @Controverial annotation)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambdaSpec.kt
zmunm: I've found some edge cases related to this, more research is needed, and code structure may change as well.
As of now, I cannot confirm only positive cases, so I would like to do more research.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambda.kt
marschwar: It is not merged yet.

cortinico: Yup sorry, so let's sync and make sure that @marschwar we merge your PR first

chao2zhang: `@RequiresTypeResolution` is now merged https://github.com/detekt/detekt/pull/3579

zmunm: done. I was mistaken for KDoc  comment :sweat_smile:

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambdaSpec.kt
zmunm: done

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambda.kt
zmunm: done

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambda.kt
zmunm: I wandering looking for a way to test. Can you give me a hint? @chao2zhang

zmunm:
In the process of adding the case, there have been many modifications to the code and JetBrains/kotlin@99a6bde is still under research.
So I changed PR to a draft and I'll request again as soon as it is resolved. Thank you

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambdaSpec.kt
```diff
@@ -0,0 +1,509 @@
...
+    
+                fun newObject() = object : Sam {
+                    override fun foo() {
+                    }
+                }
+    
+                fun lambda() = Sam {}
+
+                val a = newObject() === newObject() // false
+                val b = lambda() === lambda() // true
```
zmunm, line 500: I made a reproduction for JetBrains/kotlin@99a6bde

And I've found a bit past [issue](https://stackoverflow.com/questions/47025233/android-lifecycle-library-cannot-add-the-same-observer-with-different-lifecycle) that's occurring for similar reasons.

This is obviously confusing, but I haven't found a good solution to check this @chao2zhang

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambda.kt
chao2zhang: I believe this requires us to set up a sample project and set its kotlin version to 1.3.*

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambdaSpec.kt
chao2zhang: 👍  Just to confirm, did you refer to the code https://github.com/JetBrains/kotlin/blob/29b23e79f32791e456a5b4a453277f0f0b3e984d/idea/src/org/jetbrains/kotlin/idea/intentions/ObjectLiteralToLambdaIntention.kt#L83 to exclude this-reference?

zmunm: Yes. I used that here.
https://github.com/detekt/detekt/pull/3599/files#diff-40f82782290e1472fb9c77f556f94156050d6f123b9eaa2e4da8edd62155854aR65

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambda.kt
zmunm: When running detektMain with java.util.Runnable in 1.3.72, it was confirmed that it is detected as a test case.

BraisGabin: Couldn't we get that information from `CompilerResources`?

About how to test that `if`... we don't have any way to make it right now. But maybe we can create an issue to fix that and accept that `if` untested for now.

BraisGabin (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambda.kt
zmunm: Everything is new, so I'm grasping it. :loading: 
Please let me know if you have any misunderstandings.

From what I checked, in `CompileResource`, We can only see that the version is 1.3, and it seems that We can't actually compile with 1.3. Is that right?

What we want to check is the latter, and in java-interop there is nothing we can do with `LanguageVersion.KOTLIN_1_3` because the code we need to detect is exactly the same.

When I tested `fun interface` in 1.3, there is nothing else to do because it fails in the :compileKotlin task.

If I make an `issue`, I think I will have to rewrite the case in the module implementing 1.3, but I wonder if this falls within the scope of the project's support.

chao2zhang: The bottom line is that we do not crash when compiling Kotlin 1.3.
What's better is that even if the code does not compile, detekt findings should be easy to understand.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambdaSpec.kt
```diff
@@ -0,0 +1,509 @@
...
+                val code = """
+                fun interface Sam {
+                    fun foo()
+                }
+                val a = object : Sam {
+                    override fun foo() {
+                    }
+                }
+                """
+                subject.compileAndLint(code).assert().isEmpty()
```
chao2zhang, line 156:
```suggestion
                subject.compileAndLintWithContext(env, code).assert().isEmpty()
```
We should use compileAndLintWithContext() instead of compileAndLint() here, because the latter runs without type resolution and would degrade to no-op.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambdaSpec.kt
```diff
@@ -0,0 +1,509 @@
...
+                    }
+                }
+                """
+                subject.compileAndLintWithContext(env, code)
+                    .assert()
+                    .hasSize(1)
+                    .hasSourceLocations(SourceLocation(7, 9))
+            }
+
+            it("recursive call") {
```
chao2zhang, line 462: Nice!

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambdaSpec.kt
chao2zhang: No worries. The test code is good documentation itself.

chao2zhang (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambdaSpec.kt
```diff
@@ -0,0 +1,509 @@
...
+                val code = """
+                fun interface Sam {
+                    fun foo()
+                }
+                val a = object : Sam {
+                    override fun foo() {
+                    }
+                }
+                """
+                subject.compileAndLintWithContext(env, code).assert().isEmpty()
```
chao2zhang, line 156:
```suggestion
                subject.compileAndLint(code).assert().isEmpty()
```

My bad, I didn't realize that this test is for non-typeresolution.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambdaSpec.kt
```diff
@@ -0,0 +1,509 @@
...
+                """
+                subject.compileAndLintWithContext(env, code)
+                    .assert()
+                    .hasSize(1)
+                    .hasSourceLocations(SourceLocation(4, 9))
+            }
+        }
+
+        context("is not correct implement") {
+            it("without context") {
```
chao2zhang, line 146:
```suggestion
            it("without type resolution") {
```

chao2zhang (approved):

zmunm:
@chao2zhang Please change the linked issue to #3528

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambda.kt
chao2zhang: Currently, the default configured `detekt` task does not depend on `compileKotlin`. If the error message is confusing, we should consider adding a `LanguageVersion.KOTLIN_1_3` version check, otherwise, this PR should be good to go.

chao2zhang:
Should Detekt support primarily the current and the immediate previous feature release? 
As Kotlin 1.5 is approaching, adding checks for Kotlin 1.4 language features should be okay to merge and we can loose our compatibility requirement for Kotlin 1.3-.

zmunm:
Is there something else I need to add for this?

chao2zhang:
Kotlin 1.5 has arrived already (Not in Detekt yet), we could say that Detekt at maximum will be supporting 2 major releases of Kotlin.

PR #3604, opened 2021-03-24
Add --stacktrace to help triage CI flakiness

chao2zhang:
There seem to be a recent spike of CI random failures such as
https://github.com/detekt/detekt/pull/3595/checks?check_run_id=2178012720

I am adding `--stacktrace` to all gradle commands in CI to yield more debugging information in the CI failure.
(Some gradle commands already come with `--stacktrace`)
Meanwhile, I am also removing `--parallel` since we already have it set to true in `gradle.properties`.

### DIFF
--- a/.github/workflows/deploy-snapshot.yaml
+++ b/.github/workflows/deploy-snapshot.yaml
@@ -31,7 +31,7 @@ jobs:
         java-version: 8
 
     - name: Build detekt
-      run: ./gradlew build --build-cache --parallel
+      run: ./gradlew build --build-cache --stacktrace
 
     - name: Deploy Snapshot
       env:
--- a/.github/workflows/deploy-website.yaml
+++ b/.github/workflows/deploy-website.yaml
@@ -31,7 +31,7 @@ jobs:
           java-version: 11
 
       - name: Build Detekt Documentation
-        run: ./gradlew :detekt-generator:generateDocumentation --build-cache --parallel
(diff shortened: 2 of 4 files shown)


BraisGabin (approved):
Do we need this? Don't we upload all our builds to gradle scan?

chao2zhang:
I don't think we have build scan enabled in CI. Do you suggest adding `--scan` for gradle commands in CI?

BraisGabin:
I though that we had it by default. But we don't:

https://github.com/detekt/detekt/blob/4a89a25c82d35874571270fa1ff32751f2f268a1/build.gradle.kts#L14-L17

We just enabled the permission

chao2zhang:
I think what you are suggesting is https://docs.gradle.com/enterprise/gradle-plugin/#publishing_every_build
```
gradleEnterprise {
    buildScan {
        publishOnFailure()
    }
}
```

However, I don't think we should have that block in the main repo - `Be careful not to commit agreement to the terms of service into a project that may be built by others.` We can conditionally apply the service agreement when running on CI. https://docs.gradle.com/enterprise/gradle-plugin/#connecting_to_scans_gradle_com
```
gradleEnterprise {
    buildScan {
        termsOfServiceUrl = "https://gradle.com/terms-of-service"
        termsOfServiceAgree = "yes"
    }
}
```

chao2zhang:
In short, there is still value by adding `--stacktrace` itself, because gradle build scans may get purged later.

Issue #3514, opened 2021-03-04
NPE in DetektPomModel, probably race condition, 1.16.0-RC3


davidburstrom:
## Observed Behavior

Got the following stack trace about 50% of invocations of a closed-source tool. Happens on 1.16.0 RC2 and RC3, but not with RC1:

```
io.github.detekt.tooling.api.UnexpectedError
        at io.gitlab.arturbosch.detekt.core.tooling.AnalysisFacade$runAnalysis$1.invoke(AnalysisFacade.kt:50)
        at io.gitlab.arturbosch.detekt.core.tooling.AnalysisFacade$runAnalysis$1.invoke(AnalysisFacade.kt:21)
        at io.gitlab.arturbosch.detekt.core.tooling.ProcessingSpecSettingsBridgeKt.withSettings(ProcessingSpecSettingsBridge.kt:26)
        at io.gitlab.arturbosch.detekt.core.tooling.AnalysisFacade.runAnalysis$detekt_core(AnalysisFacade.kt:42)
        at io.gitlab.arturbosch.detekt.core.tooling.AnalysisFacade.run(AnalysisFacade.kt:25)
        ...
Caused by: java.lang.NullPointerException: Extensions.getRootArea() must not be null
        at io.github.detekt.parser.DetektPomModel.<init>(DetektPomModel.kt:24)
        at io.github.detekt.parser.KotlinEnvironmentUtilsKt.createKotlinCoreEnvironment(KotlinEnvironmentUtils.kt:56)
        at io.gitlab.arturbosch.detekt.core.settings.EnvironmentFacade$environment$2.invoke(EnvironmentAware.kt:38)
        at io.gitlab.arturbosch.detekt.core.settings.EnvironmentFacade$environment$2.invoke(EnvironmentAware.kt:22)
        at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
        at io.gitlab.arturbosch.detekt.core.settings.EnvironmentFacade.getEnvironment(EnvironmentAware.kt)
        at io.gitlab.arturbosch.detekt.core.ProcessingSettings.getEnvironment(ProcessingSettings.kt)
        at io.gitlab.arturbosch.detekt.core.KtTreeCompiler.<init>(KtTreeCompiler.kt:14)
        at io.gitlab.arturbosch.detekt.core.tooling.ParsingStrategyKt$inputPathsToKtFiles$1.invoke(ParsingStrategy.kt:24)
[...]

PR #3606, opened 2021-03-25
Update to Kotlin 1.4.32

3flex:
It's not officially released yet (though the JARs were published to Maven Central) so there's no changelog to confirm [KT-45007](https://youtrack.jetbrains.com/issue/KT-45007) fix was included, but the issue on YouTrack states it's fixed in this version so it should be safe to update.

See https://github.com/detekt/detekt/issues/3514#issuecomment-791612636

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -13,7 +13,7 @@ repositories {
 }
 
 object Plugins {
-    const val KOTLIN = "1.4.21"
+    const val KOTLIN = "1.4.32"
     const val DETEKT = "1.16.0"
     const val GITHUB_RELEASE = "2.2.12"
     const val SHADOW = "6.1.0"


chao2zhang (approved):
I was talking to @cortinico the other day about Kotlin 1.4.32 published to `mavenCentral()` without release notes.

chao2zhang:
> but the issue on YouTrack states it's fixed in this version so it should be safe to update.

I verified `kotlin-compiler-1.4.32-sources.jar`, the `KotlinCoreEnviroment.kt` does not include the change in this commit https://github.com/JetBrains/kotlin/commit/99a6bdeec7f052b180d859f59cbc5829ab6e9034
Let's wait for the official announcement of 1.4.32.

3flex:
That sounds right then. The fix for the issue is https://github.com/JetBrains/kotlin/commit/e7e17cdaa9a63684db38d1e6804864fdf1956d84 which is a revert of the commit you've linked to.

chao2zhang:
My bad. I didn't closely read the comments in KT-45007. So `kotlin-compiler-1.4.32-sources.jar` should have the reverted version:
```
                // Disposing of the environment is unsafe in production then parallel builds are enabled, but turning it off universally
                // breaks a lot of tests, therefore it is disabled for production and enabled for tests
                if (System.getProperty(KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY).toBooleanLenient() != true) {
                    // JPS may run many instances of the compiler in parallel (there's an option for compiling independent modules in parallel in IntelliJ)
                    // All projects share the same ApplicationEnvironment, and when the last project is disposed, the ApplicationEnvironment is disposed as well
                    Disposer.register(parentDisposable, Disposable {
                        synchronized(APPLICATION_LOCK) {
                            if (--ourProjectCount <= 0) {
                                disposeApplicationEnvironment()
                            }
                        }
                    })
                }
```

cortinico:
> Edit: We can still wait for a few hours/days for the official release & release notes.

+1 for this

3flex:
https://github.com/JetBrains/kotlin/releases/tag/v1.4.32

KT-45007 confirmed fixed.

BraisGabin:
https://kotlinlang.org/docs/releases.html#release-details 1.4.32 release 🎉

BraisGabin (approved):

PR #3608, opened 2021-03-26
RFC - Initial idea for @Configuration annotation

marschwar:
This is what I have got so far. I had not realized that there were so many custom checks on rules and rule providers.

### What has been done
* I created an annotation that can be applied to value parameters 
* I picked a random rule (`ComplexMethod`) to illustrate what it would look like 
* The rule is configured by exposing all config options in the constructor binding them to immutable fields.
* The mapping between the config and those fields is done in a factory method
* In the RuleSetProvider the factory method is used

### What would come next
* Extract the documentation and default conifg from the annotation instead of the kdoc
* Create an annotation processor to generate a `[RuleName]Factory` similar to the factory method
* Disable  some custom rule checks for rules that have the annotations

### And still after that
* Infer the default property value from the annotated parameter for documentation and default config 
* Adapt the pattern to more rules
* ...

Since this is quite a change, I would like your opinions on whether this is worth the effort and if this going in the right direction or what else you had in mind when you saw the issue come up.

### DIFF
--- a/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
+++ b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
@@ -7,6 +7,8 @@ import io.gitlab.arturbosch.detekt.generator.collection.exception.InvalidDocumen
 import io.gitlab.arturbosch.detekt.rules.isOverride
 import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtClassOrObject
+import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
+import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.KtStringTemplateExpression
@@ -101,10 +103,22 @@ class RuleSetProviderVisitor : DetektVisitor() {
             val ruleArgumentNames = (ruleListExpression as? KtCallExpression)
                 ?.valueArguments
                 ?.map { it.getArgumentExpression() }
-                ?.mapNotNull { it?.referenceExpression()?.text }
+                ?.mapNotNull(this::guessRuleNameOrNull)
                 ?: emptyList()
 
             ruleNames.addAll(ruleArgumentNames)
+        }
+    }
+
+    private fun guessRuleNameOrNull(expressionOrNull: KtExpression?): String? {
+        return expressionOrNull?.let {
+            when (it) {
+                // constructor
+                is KtCallExpression -> it.referenceExpression()?.text
+                // factory method
+                is KtDotQualifiedExpression -> it.receiverExpression.text
+                else -> null
+            }
         }
     }
 }
--- a/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
+++ b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
@@ -102,7 +102,7 @@ class RuleSetProviderVisitor : DetektVisitor() {
 
             val ruleArgumentNames = (ruleListExpression as? KtCallExpression)
                 ?.valueArguments
-                ?.map { it.getArgumentExpression() }
+                ?.mapNotNull { it.getArgumentExpression() }
                 ?.mapNotNull(this::guessRuleNameOrNull)
                 ?: emptyList()
 
@@ -110,15 +110,12 @@ class RuleSetProviderVisitor : DetektVisitor() {
         }
     }
 
-    private fun guessRuleNameOrNull(expressionOrNull: KtExpression?): String? {
-        return expressionOrNull?.let {
-            when (it) {
-                // constructor
-                is KtCallExpression -> it.referenceExpression()?.text
-                // factory method
-                is KtDotQualifiedExpression -> it.receiverExpression.text
-                else -> null
-            }
+    private fun guessRuleNameOrNull(expression: KtExpression): String? =
+        when (expression) {
+            // constructor
+            is KtCallExpression -> expression.referenceExpression()?.text
+            // factory method
+            is KtDotQualifiedExpression -> expression.receiverExpression.text
+            else -> null
         }
-    }
 }
--- a/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
+++ b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
@@ -14,6 +14,7 @@ import io.gitlab.arturbosch.detekt.rules.empty.EmptyRule
 import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtClass
 import org.jetbrains.kotlin.psi.KtClassOrObject
+import org.jetbrains.kotlin.psi.KtPrimaryConstructor
 import org.jetbrains.kotlin.psi.KtSuperTypeList
 import org.jetbrains.kotlin.psi.KtValueArgument
 import org.jetbrains.kotlin.psi.psiUtil.containingClass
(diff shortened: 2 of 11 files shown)


detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
```diff
@@ -90,17 +94,29 @@ class RuleSetProviderVisitor : DetektVisitor() {
...
+                .map { it.getArgumentExpression() }
+                .firstOrNull { it?.referenceExpression()?.text == "listOf" }
                 ?: throw InvalidDocumentationException("RuleSetProvider $name doesn't provide list of rules.")
 
             val ruleArgumentNames = (ruleListExpression as? KtCallExpression)
-                    ?.valueArguments
-                    ?.map { it.getArgumentExpression() }
-                    ?.mapNotNull { it?.referenceExpression()?.text }
+                ?.valueArguments
+                ?.map { it.getArgumentExpression() }
```
BraisGabin, line 103:
```suggestion
                ?.mapNotNull { it.getArgumentExpression() }
```

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
```diff
@@ -90,17 +94,29 @@ class RuleSetProviderVisitor : DetektVisitor() {
...
+                ?.mapNotNull(this::guessRuleNameOrNull)
                 ?: emptyList()
 
             ruleNames.addAll(ruleArgumentNames)
         }
     }
+
+    private fun guessRuleNameOrNull(expressionOrNull: KtExpression?): String? {
```
BraisGabin, line 111:
```suggestion
    private fun guessRuleNameOrNull(expressionOrNull: KtExpression): String? {
```
There's not reason to accept a `null` value here, right?

BraisGabin (commented):
This looks so great! Inject the configuration through the constructor is a great step forward.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
```diff
@@ -49,19 +50,35 @@ import org.jetbrains.kotlin.psi.KtWhenExpression
  */
 class ComplexMethod(
     config: Config = Config.empty,
-    threshold: Int = DEFAULT_THRESHOLD_METHOD_COMPLEXITY
+
+    @Configuration("McCabe's Cyclomatic Complexity (MCC) number for a method. (default: `15`)")
+    threshold: Int = DEFAULT_THRESHOLD_METHOD_COMPLEXITY,
+
+    @Configuration("Ignores a complex method if it only contains a single when expression. (default: `false`)")
+    val ignoreSingleWhenExpression: Boolean = false,
```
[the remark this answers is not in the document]
marschwar: You mean instead of specifying it in the annotation? Yes, that is what I meant with
> Infer the default property value from the annotated parameter for documentation and default config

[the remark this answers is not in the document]
BraisGabin: I didn't read the description 😅

BraisGabin:
> * The rule is configured by exposing all config options in the constructor binding them to immutable fields. 
> * The mapping between the config and those fields is done in a factory method
> * In the RuleSetProvider the factory method is used

I really like this idea. We could think a way to support "new types" as `Regex` and `Globbing`. So in the documentation is easier to document what's expected.

> * Create an annotation processor to generate a `[RuleName]Factory` similar to the factory method

Is it worth it? Would it be easier to write the first version without it? Maybe we can do it later ot maybe it's easier to implement the annotation processor than implement the factories manually. What do you think?

> * Disable  some custom rule checks for rules that have the annotations

I don't understand what do you mean with this.

> * Infer the default property value from the annotated parameter for documentation and default config

That would be huge. We had/have a lot of misalignment in this topic.

----

In general I think that this aligns well with all the ideas that we have around 2.0

marschwar:
Thank you for taking the time to look into this PR.

>> Create an annotation processor to generate a [RuleName]Factory similar to the factory method

> Is it worth it? Would it be easier to write the first version without it? Maybe we can do it later ot maybe it's easier to implement the annotation processor than implement the factories manually. What do you think?

That is actually a great point. The factory method does work for now and is not more code than binding the parameters to fields like it is currently done

>> Disable some custom rule checks for rules that have the annotations

> I don't understand what do you mean with this.
Maybe this will not be necessary after all. I will check.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
```diff
@@ -104,9 +110,31 @@ class ComplexMethod(
...
         const val DEFAULT_THRESHOLD_METHOD_COMPLEXITY = 15
+
+        // Still to do: Disable custom rule check for annotated rules
+        //  io.gitlab.arturbosch.detekt.core.ConfigAssert#checkOptions
         const val IGNORE_SINGLE_WHEN_EXPRESSION = "ignoreSingleWhenExpression"
         const val IGNORE_SIMPLE_WHEN_ENTRIES = "ignoreSimpleWhenEntries"
         const val IGNORE_NESTING_FUNCTIONS = "ignoreNestingFunctions"
         const val NESTING_FUNCTIONS = "nestingFunctions"
+
+        // Code similar to this will be generated in the next iteration by an annotation processor
```
chao2zhang, line 121: 👏

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
```diff
@@ -36,20 +37,31 @@ import org.jetbrains.kotlin.psi.KtWhenExpression
...
- * `for` statement (default: `false`)
- * @configuration nestingFunctions - Comma separated list of function names which add complexity
- * (default: `[run, let, apply, with, also, use, forEach, isNotNull, ifNull]`)
  */
 @ActiveByDefault(since = "1.0.0")
 class ComplexMethod(
     config: Config = Config.empty,
-    threshold: Int = DEFAULT_THRESHOLD_METHOD_COMPLEXITY
+
+    @Configuration("McCabe's Cyclomatic Complexity (MCC) number for a method. (default: `15`)")
```
chao2zhang, line 45: We should not need to specify `(default: `15`)` in the description. The default should be auto generated.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
```diff
@@ -35,14 +36,21 @@ internal class RuleVisitor : DetektVisitor() {
...
     private val classesMap = mutableMapOf<String, Boolean>()
 
     fun getRule(): Rule {
         if (description.isEmpty()) {
             throw InvalidDocumentationException("Rule $name is missing a description in its KDoc.")
         }
 
+        if (configurationByAnnotation.isNotEmpty() && configurationByKdoc.isNotEmpty()) {
+            throw InvalidDocumentationException(
+                "Rule $name is using both annotations and kdoc to define configuration parameter."
```
chao2zhang, line 50: 👍

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
```diff
@@ -36,20 +37,31 @@ import org.jetbrains.kotlin.psi.KtWhenExpression
...
+        "Whether to ignore functions which are often used instead of an `if` or `for` statement. " +
+            "(default: `false`)"
+    )
+    val ignoreNestingFunctions: Boolean = false,
+
+    @Configuration(
+        "Comma separated list of function names which add complexity. " +
+            "(default: `[run, let, apply, with, also, use, forEach, isNotNull, ifNull]`)"
+    )
+    val nestingFunctions: Set<String> = DEFAULT_NESTING_FUNCTIONS
```
chao2zhang, line 64: What's your inclination between `@Configuration` on constructor parameters and `@Configuration` on properties?
My instinct tells me that keeping them on properties can avoid creating factory methods.

marschwar:
Thank you @BraisGabin, @cortinico  and @chao2zhang for providing feedback. 

> My instinct tells me that keeping them on properties can avoid creating factory methods.

This made me rethink the entire thing and I created a config property delegate to avoid factory method boilerplate code. The latest version looks like this.

```Kotlin
class ComplexMethod(config: Config = Config.empty) : Rule(config) {

    @Configuration("McCabe's Cyclomatic Complexity (MCC) number for a method.")
    private val threshold: Int by config(DEFAULT_THRESHOLD_METHOD_COMPLEXITY)

    @Configuration("Ignores a complex method if it only contains a single when expression.")
    private val ignoreSingleWhenExpression: Boolean by config(false)
...
}
``` 

I will close this for now while I am working on automatically inferring the correct default value for the config parameter.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
marschwar: I know. I am on it.

BraisGabin:
For 2.0 all the configuration will work with factories. That's not a stopper and we should find the best solution for 1.0. But if it works for both will be even better. But I insist, we should find the best solution for 1.0 now.

PR #3611, opened 2021-03-27
Decouple Gradle projects

3flex:
This reduces the number of subprojects that have to be recompiled when there's a change to one of the rules, saving quite a bit of time when building the project.

The most interesting change is in the detekt-cli build file since the shadow jar config has changed. The new config can be verified by executing the resulting jar with `--debug` and confirming that the bundled rulesets have been registered.

### DIFF
--- a/detekt-cli/build.gradle.kts
+++ b/detekt-cli/build.gradle.kts
@@ -1,5 +1,3 @@
-import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
-
 plugins {
     id("com.github.johnrengelman.shadow")
     module
@@ -10,9 +8,7 @@ application {
     mainClassName = "io.gitlab.arturbosch.detekt.cli.Main"
 }
 
-tasks.withType<ShadowJar>().configureEach {
-    mergeServiceFiles()
-}
+val bundledRules by configurations.creating
 
 dependencies {
     implementation(kotlin("stdlib-jdk8"))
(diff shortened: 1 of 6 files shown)


detekt-cli/build.gradle.kts
```diff
@@ -10,29 +8,26 @@ application {
...
     implementation("com.beust:jcommander")
     implementation(project(":detekt-tooling"))
     implementation(project(":detekt-parser"))
     runtimeOnly(project(":detekt-core"))
-    runtimeOnly(project(":detekt-rules"))
 
     testImplementation(project(":detekt-test"))
-    testImplementation(project(":detekt-rules"))
+
+    bundledRules(project(":detekt-rules"))
```
chao2zhang, line 21: If we remove this `bundleRules` configuration, what would happen?

chao2zhang (approved):

detekt-cli/build.gradle.kts
3flex: The rule JARs wouldn't be bundled with the JAR created by shadow for detekt-cli. They'd instead have to be passed in manually using the `--plugins` flag to detekt-cli.

detekt-core/build.gradle.kts
```diff
@@ -14,7 +14,7 @@ dependencies {
     implementation(project(":detekt-report-xml"))
     implementation(project(":detekt-report-sarif"))
 
-    testImplementation(project(":detekt-rules"))
+    testRuntimeOnly(project(":detekt-rules"))
```
BraisGabin, line 17: It seems like we have at least one unit test that do far too much. Otherwise we should not need this, right?

BraisGabin (approved):
Great boost to the build time 👏🏻👏🏻

detekt-cli/build.gradle.kts
chao2zhang: Thanks for the explanation. Let's keep the `bundleRules` configuration then.

schalkms (approved):

detekt-core/build.gradle.kts
3flex: Yeah there are a couple of places detekt-rules is used when it maybe shouldn't be.

I'll open a new issue for that because it would help further with build times when iterating on rules.

PR #3612, opened 2021-03-27
Simplify Gradle config

3flex:
Uses [withJavadocJar](https://docs.gradle.org/current/javadoc/org/gradle/api/plugins/JavaPluginExtension.html#withJavadocJar--) and [withSourcesJar](https://docs.gradle.org/current/javadoc/org/gradle/api/plugins/JavaPluginExtension.html#withSourcesJar--) which are functionally equivalent to the config that was removed.

The new config also avoids configuring the javadoc/sourcesjar tasks during Gradle's configuration phase.

Tested by publishing to mavenLocal on my machine. The javadoc jars are empty anyway with the current config, and sources are published as expected.

### DIFF
--- a/buildSrc/src/main/kotlin/module.gradle.kts
+++ b/buildSrc/src/main/kotlin/module.gradle.kts
@@ -72,26 +72,13 @@ dependencies {
     testRuntimeOnly("org.spekframework.spek2:spek-runner-junit5")
 }
 
-val sourcesJar by tasks.registering(Jar::class) {
-    dependsOn(tasks.classes)
-    archiveClassifier.set("sources")
-    from(sourceSets.main.get().allSource)
-}
-
-val javadocJar by tasks.registering(Jar::class) {
-    from(tasks.javadoc)
-    archiveClassifier.set("javadoc")
-}
-
-artifacts {
-    archives(sourcesJar)
-    archives(javadocJar)
+java {
+    withSourcesJar()
+    withJavadocJar()
 }
 
 publishing {
     publications.named<MavenPublication>(DETEKT_PUBLICATION) {
         from(components["java"])
-        artifact(sourcesJar.get())
-        artifact(javadocJar.get())
     }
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/module.gradle.kts
```diff
@@ -45,8 +45,6 @@ tasks.withType<Test>().configureEach {
 }
 
 tasks.withType<KotlinCompile>().configureEach {
-    sourceCompatibility = JavaVersion.VERSION_1_8.toString()
-    targetCompatibility = JavaVersion.VERSION_1_8.toString()
```
chao2zhang, line 49: Were `sourceCompatibility` and `targetCompatibility` the default value as well?

3flex: These flags are only needed when there's Java source (to set the source compatibility and the version to use in the generated class files). But the Kotlin compiler doesn't need them because there's no Java source in this project.

Even if there was, these flags would need to be set on the JavaCompile tasks instead because the Kotlin compiler doesn't compile Java source.

chao2zhang: Thanks for the explanation! I found https://youtrack.jetbrains.com/issue/KT-32805 also helpful.

chao2zhang (approved):

BraisGabin (approved):
> The javadoc jars are empty anyway with the current config

Should we create a new issue to track this?

schalkms (approved):

3flex:
> Should we create a new issue to track this?

I don't think so. IIRC the javadoc jars are only created to satisfy a Maven Central requirement to provide them on upload, even if they're empty. The public Kotlin API is documented on the detekt website and since this is a Kotlin-only project I don't think it makes sense to publish Javadoc which would cover the Java flavour of the API.

PR #3615, opened 2021-03-27
Apply more formatting rules to our code

### LINKED ISSUE #3593: Reformat the project
I remember that Arthur talked about this time ago and we decide not to do it but maybe now is the time. #3579 has a lot of changes that provides 0 information about what the PR really does. The problem is that the old code isn't properly formatted.

I vote to activate the formatting plugin of detekt or add ktlint to the project directly so we ensure that the format in all the code is consistent. This will create a big PR changing a lot of lines but we know that those changes doesn't affect the behaviour so they are safe.

BraisGabin:
Fixes #3593

I enabled more formatting rules for this project and run detekt with autocorrect. Then I run ktlint 0.41.0 with the `--experimental` argument and push the changes that doesn't conflict with the current rules. And finally I fixed some minor style things manually.

I can't enable the rules:
- `ArgumentListWrapping` because it have conflicts with `Indentation`
- `AnnotationOnSeparateLine` because it have false positives.

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/LibraryEntitiesShouldNotBePublicTest.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/LibraryEntitiesShouldNotBePublicTest.kt
@@ -13,9 +13,12 @@ internal class LibraryEntitiesShouldNotBePublicTest : Spek({
         it("should not report without explicit filters set") {
             val subject = LibraryEntitiesShouldNotBePublic(TestConfig(Config.EXCLUDES_KEY to "**"))
             assertThat(
-                subject.compileAndLint("""
+                subject.compileAndLint(
+                    """
                     class A 
-            """)).isEmpty()
+            """
+                )
+            ).isEmpty()
         }
 
         val subject by memoized {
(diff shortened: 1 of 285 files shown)


chao2zhang:
> * `ArgumentListWrapping` because it have conflicts with `Indentation`

This is what I have observed as well in other projects

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/dsl/ConfigSpecBuilder.kt
```diff
@@ -9,6 +9,7 @@ class ConfigSpecBuilder : Builder<ConfigSpec> {
 
     var shouldValidateBeforeAnalysis: Boolean = true
     var knownPatterns: Collection<String> = emptyList()
+
```
chao2zhang, line 12: Is this empty line addition done by autoCorrection from formatting rules? This is a little bit surprising if this is true.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/LibraryEntitiesShouldNotBePublicTest.kt
```diff
@@ -24,62 +27,94 @@ internal class LibraryEntitiesShouldNotBePublicTest : Spek({
 
         describe("positive cases") {
             it("should report a class") {
-                assertThat(subject.compileAndLint("""
+                assertThat(
+                    subject.compileAndLint(
+                        """
                     class A
-                """)).hasSize(1)
+                """
```
chao2zhang, line 34: The multi-line string looks worse in my opinion. But we can always manually fix these as a follow-up

chao2zhang (approved):

schalkms (approved):

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/dsl/ConfigSpecBuilder.kt
BraisGabin: Yes, it is. I just did it again and this is a funny case... I needed to run detekt twice in this file to get the formatting rules happy... this seems like a bug.

PR #3618, opened 2021-03-27
Improve tests

BraisGabin:
This is a boy scout PR. I'm just preparing the tests to add some more meat on them. I think that reviewing it commit by commit will help to understand it.


chao2zhang (approved):
Thanks for adding test coverage for type resolution and `detektTest`

PR #3620, opened 2021-03-27
Remove useless config `continuationIndentSize`

### LINKED ISSUE #3533: Indentation rule clashes with ArgumentListWrapping
Hi,

I found out that both [Indentation] and [ArgumentListWrapping] is doing the same thing.
Following code:
```kotlin
    private val remoteUrl: String
        get() = context.getString(R.string.faq_url,
            context.getString(R.string.action_id),
            context.getString(R.string.element_id),
            context.getString(R.string.product_id),
            language)
```
is reported by both. I would expect that Indentation is silent, as this is continuation indentation of 4.

I've also tried to put everything on separate line and indent by 8 spaces. Which was again reported by both, while I would expect this to be reported only from Indentation, as the wrap was fulfilled.

If I'm not mistaken, the ArgumentListWrapping is just duplication with less options, that could be dropped. Or making both rules taking a smaller scope as they are likely doing the same thing.

* Version of detekt used: 1.16.0-RC3

chao2zhang:
This fixes #3533 by not exposing useless configs.

Although this PR reduces confusion for our users. This is still a breaking change though, and I am open to delaying this until the next major breaking version.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -277,7 +277,6 @@ formatting:
     active: false
     autoCorrect: true
     indentSize: 4
-    continuationIndentSize: 4
   MaximumLineLength:
     active: true
     maxLineLength: 120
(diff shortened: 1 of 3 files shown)


schalkms (approved):

schalkms:
This could go into 1.17.0 IMHO.

BraisGabin:
This will break all the configurations files with that value. We had code to deprecate configuration. We could deprecate it and remove it in 3 versions.

chao2zhang:
> We could deprecate it and remove it in 3 versions.

It is good to know the policy here. Did we have an example of removing configs like this before?

schalkms:
Since the configuration isn't doing anything, the option could be removed earlier IMHO. Anyways, I don't think there's a strong policy regarding deprecations in detekt. I'm fine with either way here though.

PR #3624, opened 2021-03-29
Upgrade to ktlint 0.41.0

chao2zhang:
## Main change in wrapping rules
- https://github.com/pinterest/ktlint/pull/1093 refactored and made EditorConfigProperties more publicly visible. We can also remove our hacks in the wrapping rules.
- Some new configuration properties are also added to catch up with Ktlint reading configurations.

## Formatting fixes
- HtmlOutputReport.kt - This seemingly unrelated change is probably due to changes for `Indentation` rule.
- @file:Suppress("SpacingBetweenDeclarationsWithAnnotations") in KotlinEnvironmentTestSetup.kt - This is reported as https://github.com/pinterest/ktlint/issues/1125

## Note
We now have warnings in our `ImportOrderingRule` like the following:
```
[WARNING] `idea` is deprecated! Please use `*,java.**,javax.**,kotlin.**,^` instead to ensure that the Kotlin IDE plugin recognizes the value
[WARNING] `idea` is deprecated! Please use `*,java.**,javax.**,kotlin.**,^` instead to ensure that the Kotlin IDE plugin recognizes the value
```
We probably need a way to notify our consumers to migrate the values like `idea` or `ascii` to the recommended regexs.


detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/ImportOrdering.kt
```diff
@@ -1,18 +1,20 @@
...
 /**
  * See <a href="https://ktlint.github.io">ktlint-website</a> for documentation.
  *
  * For defining custom import layout patterns see: https://github.com/pinterest/ktlint/blob/cdf871b6f015359f9a6f02e15ef1b85a6c442437/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt
  *
- * @configuration layout - the import ordering layout; use 'ascii', 'idea' or define a custom one (default: `'idea'`)
+ * @configuration layout - the import ordering layout; use 'ascii', 'idea' or define a custom one.
+ * See the documentation on the Ktlint ImportOrderingRule for customized details.
+ * https://github.com/pinterest/ktlint/blob/master/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt
+ * (default: `'idea'`)
```
BraisGabin, line 17: We shouldn't set a deprecated value in a configuration, right?

detekt-test-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KotlinEnvironmentTestSetup.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("SpacingBetweenDeclarationsWithAnnotations")
```
BraisGabin, line 1: I'd add a comment to this annotation refering to the known issue. This way, eventually, we eill remove this workaround

BraisGabin (commented):
> ## Note
> 
> We now have warnings in our `ImportOrderingRule` like the following:
> 
> ```
> [WARNING] `idea` is deprecated! Please use `*,java.**,javax.**,kotlin.**,^` instead to ensure that the Kotlin IDE plugin recognizes the value
> [WARNING] `idea` is deprecated! Please use `*,java.**,javax.**,kotlin.**,^` instead to ensure that the Kotlin IDE plugin recognizes the value
> ```
> 
> We probably need a way to notify our consumers to migrate the values like `idea` or `ascii` to the recommended regexs.

This is a runtime warnning that our users can see, right? I think that we should just change our default value and that's it. If they have the `idea` value in their configuration they will see the warning and they will fix it. And if they don't have it they will use the correct default value.

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/ArgumentListWrapping.kt
```diff
@@ -2,15 +2,37 @@ package io.gitlab.arturbosch.detekt.formatting.wrappers
...
+import io.gitlab.arturbosch.detekt.formatting.DEFAULT_INDENT
 import io.gitlab.arturbosch.detekt.formatting.FormattingRule
+import io.gitlab.arturbosch.detekt.formatting.INDENT_SIZE_KEY
+import io.gitlab.arturbosch.detekt.formatting.MAX_LINE_LENGTH_KEY
 
 /**
  * See <a href="https://ktlint.github.io">ktlint-website</a> for documentation.
  *
+ * @configuration indentSize - indentation size (default: `4`)
+ * @configuration maxLineLength - maximum line length (default: `120`)
```
BraisGabin, line 16: Out of scope but should we have this configurations at rule set level? Do we want to allow different values between rules?

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/ImportOrdering.kt
chao2zhang: We shouldn't. Although changing the default value is a backward-incompatible change in my opinion.

detekt-test-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KotlinEnvironmentTestSetup.kt
chao2zhang: Good idea.

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/ArgumentListWrapping.kt
chao2zhang: For ktlint these values are consistent across rules, for detekt we seem to add flexibility to allow different rules.

When implementing this PR, I thought about this but also worried that this may already be exploited by our users. 
We can keep this in mind when ktlint exits `FeatureInAlphaState` and eventually moves to `EditorConfigProperty`

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/ImportOrdering.kt
BraisGabin: Why is that? `*,java.**,javax.**,kotlin.**,^` shoudl be the same as 'idea', right?

chao2zhang: You are right. The equivalence is maintained.

BraisGabin (approved):

PR #3630, opened 2021-03-29
Feature/support config resource http

blagerweij:
Added support for loading config from a URL resource.


3flex:
It's surely low risk but since security hasn't been a consideration  I don't know whether supporting http resources is a good idea - https only would be the preference.

Also just wondering if you need this in the CLI or if we can just offer it in the Gradle plugin?

BraisGabin:
I like the idea but, could you provide a bit of context? Why do you want this feature? Which is your use case?

And I kind of agree with @3flex. Should we support http? Shouldn't it be better to just support https?

chao2zhang:
@blagerweij I am closing this PR. If the suggested workaround does not work for you, please don't hesitate to open new issue or reopen this PR.

PR #3621, opened 2021-03-28
Remove deprecated configurations and deprecate new ones

chao2zhang:
- Remove deprecated configurations that were deprecated at least 3 minor versions ago (https://github.com/detekt/detekt/pull/3620#issuecomment-808807405)
- Add new deprecation for `formatting>Indentation>continuationIndentSize` (https://github.com/detekt/detekt/pull/3620)

PR #3637, opened 2021-03-30
Rule Configuration using annotations

marschwar:
This PR relates to #3562 and makes it possible to configure a Rule using a `@Configuration` annotation on properties in conjunction with a config delegate.

### Advantages
* The default value for a config parameter is detected automatically
* The name for a config parameter is automatically the same as the property
* There is no need to call `valueOrDefault` as this is done by the delegate automatically

### Limitations / Points to discuss
* Right now this is only implemented on one rule `ComplexMethod` to illustrate what it would look like.
* Currently only String, Boolean, Int, Long and List<String> are supported as configuration properties. Which types did I miss?
* The default value can only be inferred when it is definied directly in the delegate expression or as a constant in the rules companion object. (see #3628)
* The formatting of lists of strings in the `default-detekt-config.yml` is inconsistent (`[run, let, apply]` vs. `['**/test/**', '**/androidTest/**']`) so I opted for the version with quotes. Is that a problem?
* I disabled the checks in ConfigAssert for rules that use `@Configuration` as they are not needed IMO.


BraisGabin:
> This PR relates to #3562 and makes it possible to configure a Rule using a `@Configuration` annotation on properties in conjunction with a config delegate.
> ### Advantages
> 
>     * The default value for a config parameter is detected automatically
> 
>     * The name for a config parameter is automatically the same as the property
> 
>     * There is no need to call `valueOrDefault` as this is done by the delegate automatically
> 
> 
> ### Limitations / Points to discuss
> 
> 
> * Currently only String, Boolean, Int, Long and List are supported as configuration properties. Which types did I miss?

`Regex`? It's the same as `String` but maybe we could use it to make more clear what type of String we expect there. But this is out of scope of this PR. It could be done as a follow up.

> * The default value can only be inferred when it is definied directly in the delegate expression or as a constant in the rules companion object. (see #3628)

I don't know how to help you here. Are there a lot of these cases? Could we simplify them to use always values in the same file?

> * The formatting of lists of strings in the `default-detekt-config.yml` is inconsistent (`[run, let, apply]` vs. `['**/test/**', '**/androidTest/**']`) so I opted for the version with quotes. Is that a problem?

This is not a problem at all. I was wondering about other configurations where we use this syntax:

```yaml
    exceptionNames:
     - ArrayIndexOutOfBoundsException
     - Error
     - Exception
     - IllegalMonitorStateException
     - NullPointerException
     - IndexOutOfBoundsException
     - RuntimeException
     - Throwable
```

Could it be possible to use both? I mean, the one-line syntax in this example would be really difficult to maintain and works really bad with `git`.

But again this don't need to be faced in this PR if it makes it more complex.

BraisGabin:
> List format: Maybe it always should be the line-per-item syntax for all list parameters?

I'm not sure about this either. We have some configurations that will probably never change (`[run, let, apply]`) and they looks better in just one line... But this should not be a blocker. We can fix this later.

marschwar:
As far as I know you had not commented on this PR but rather #3608. I created an entirely new PR. I probably should have referenced it in the description

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/ConfigurationCollector.kt
```diff
@@ -0,0 +1,165 @@
...
+                "Type of '$name' is not supported. " +
+                    "For properties annotated with @Configuration use one of$SUPPORTED_TYPES."
+            }
+        }
+        if (!isInitializedWithConfigDelegate()) {
+            invalidDocumentation { "'$name' is not using the '$DELEGATE_NAME' delegate" }
+        }
+
+        val propertyName: String = checkNotNull(name)
+        val deprecationMessage = firstAnnotationParameterOrNull(Deprecated::class)
```
chao2zhang, line 84: This is great - https://github.com/detekt/detekt/pull/3621 is a use case for this.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/ConfigurationCollector.kt
```diff
@@ -0,0 +1,165 @@
...
+            get() = parent as KtProperty
+
+        private val KtProperty.declaredTypeOrNull: String?
+            get() = typeReference?.text
+
+        private val KtElement.listDeclarationOrNull: KtCallExpression?
+            get() = findDescendantOfType { it.isListDeclaration() }
+
+        private val KtElement.listDeclaration: KtCallExpression
+            get() = checkNotNull(listDeclarationOrNull)
```
chao2zhang, line 139: I would recommend using `extension method` such as `getListDeclarationOrNull` over `extension property` as the implementation is computation heavy and describes a behavior, not state.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/DocumentationCollector.kt
```diff
@@ -0,0 +1,83 @@
+package io.gitlab.arturbosch.detekt.generator.collection
+
+import io.gitlab.arturbosch.detekt.generator.collection.exception.InvalidCodeExampleDocumentationException
+import org.jetbrains.kotlin.psi.KtClassOrObject
+
+class DocumentationCollector {
```
chao2zhang, line 6: Great refactor to separate the responsibility!

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/DocumentationCollector.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    fun setClass(classOrObject: KtClassOrObject) {
+        name = classOrObject.name?.trim() ?: ""
+        classOrObject.kDocSection()
+            ?.getContent()
+            ?.trim()
+            ?.replace("@@", "@")
+            ?.also { comment ->
+                extractRuleDocumentation(comment)
+            }
```
chao2zhang, line 24:
```suggestion
            ?.let(::extractRuleDocumentation)
```
`also` is confusing because we are not preserving the evaluated value of the previous expression.

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleCollectorSpec.kt
```diff
@@ -151,114 +151,419 @@ class RuleCollectorSpec : Spek({
...
+                            @Configuration("description")
+                            private val config: String by config("[A-Z$]")
+                        }                        
+                    """
+                    val items = subject.run(code)
+                    assertThat(items[0].configuration).hasSize(1)
+                    assertThat(items[0].configuration[0].name).isEqualTo("config")
+                    assertThat(items[0].configuration[0].description).isEqualTo("description")
+                    assertThat(items[0].configuration[0].defaultValue).isEqualTo("'[A-Z$]'")
+                    assertThat(items[0].configuration[0].deprecated).isNull()
```
chao2zhang, line 182:
```suggestion
                    val expectedConfiguration = Configuration(
                        name = "config",
                        description = "description",
                        defaultValue = "'[A-Z$]'"
                        deprecated = null
                    )
                    assertThat(items[0].configuration[0]).isEqualTo(expectedConfiguration)
```

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
```diff
@@ -36,21 +36,24 @@ import org.jetbrains.kotlin.psi.KtWhenExpression
...
  */
 @ActiveByDefault(since = "1.0.0")
-class ComplexMethod(
-    config: Config = Config.empty,
-    threshold: Int = DEFAULT_THRESHOLD_METHOD_COMPLEXITY
-) : ThresholdRule(config, threshold) {
+class ComplexMethod(config: Config = Config.empty) : Rule(config) {
+
+    @Configuration("McCabe's Cyclomatic Complexity (MCC) number for a method.")
+    private val threshold: Int by config(DEFAULT_THRESHOLD_METHOD_COMPLEXITY)
```
chao2zhang, line 44: I would personally inline 15.. I guess that's violated by the MagicNumber rule.

chao2zhang (approved):
PR looks great to me!

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexMethod.kt
marschwar: I am torn. I really think it is redundant since it is clearly the default value and does not need extra naming. But on the other hand I dislike us excessively suppressing our own rules.

.github/CONTRIBUTING.md
```diff
@@ -14,56 +14,71 @@
...
-    function correctly by default. 
-    The default value will be taken as is for the configuration option and pasted into the `default-detekt-config.yml`.
-    
-    A `@configuration` tag as described above will translate to a rule entry in the `default-detekt-config.yml`:
-    ```yml
-    SomeRule:
-       active: false
-       name: whatever should be the default
-    ```
+- ... add the [correct KDoc](#contents-and-structure-of-a-rules-kdoc) and [annotations](#rule-annotations) to your `Rule` class. This is used to generate wiki pages and the `default-detekt-config.yml` automatically.
```
BraisGabin, line 17:
```suggestion
- ... add the [correct KDoc](#contents-and-structure-of-a-rules-kdoc) and [annotations](#rule-annotations) to your `Rule` class. This is used to generate documentation pages and the `default-detekt-config.yml` automatically.
```

BraisGabin (approved):
And after this we need to make the migration and then remove the old code. Are you traking the remaining taask somewhere? I could help you with this more "mechanical" work.

BraisGabin:
And, of course, this is a huge improvement!

BraisGabin:
> What would be a good format for this? Create an issue per task or one for all? 

That's your call. Both seem fine to me.

chao2zhang (approved):

PR #3641, opened 2021-03-31
fix for KtNamedFunction.isMainFunction()

krzykrucz:
Fixes the issue that `vararg` variation of `main` function was not taken into account in `KtNamedFunction#isMainFunction`.

As a result fixes also rules using this function: `ThrowingExceptionInMain`, `ExitOutsideMain`.


BraisGabin (approved):

BraisGabin:
Good catch!

chao2zhang (approved):

Issue #3643, opened 2021-04-01
Rename MayBeConst to MayBeConstant for consistency


ZacSweers:
Detekt's rule for checking properties that could be `const` uses the name `MayBeConst`. The IDE inspection from Kotlin however uses `MayBeConstant`, which means suppressing the warning requires two redundant suppressions

![image](https://user-images.githubusercontent.com/1361086/113336726-d0577480-92f4-11eb-86f8-db832e55545c.png)

I'd like to propose matching the Kotlin plugin's name so they can just be consolidated to one

BraisGabin:
I see three steps here:

- For now you can add an `aliases: 'MayBeConstant'` in the yaml configuration. below the `MayBeConst` rule.
- Then we can add that alias as by default in detekt. So no one needs to add that alias after the next version of detekt
- Wait for 2.0 so we can rename it (this is a breaking change so we can't do it now)

Thanks for the report! These "quality of life" things are usually no reported and they are as important as the other ones.

ZacSweers:
Thanks! Didn't know about the alias option 👍

PR #3644, opened 2021-04-01
Add MayBeConstant as alias for MayBeConst

BraisGabin:
Workaround for #3643 meanwhile we can't change the name of the rule

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MayBeConst.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MayBeConst.kt
@@ -45,6 +45,8 @@ class MayBeConst(config: Config = Config.empty) : Rule(config) {
         Debt.FIVE_MINS
     )
 
+    override val defaultRuleIdAliases = setOf("MayBeConstant")
+
     private val binaryTokens = hashSetOf<KtSingleValueToken>(
         KtTokens.PLUS,
         KtTokens.MINUS,


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MayBeConst.kt
```diff
@@ -45,6 +45,8 @@ class MayBeConst(config: Config = Config.empty) : Rule(config) {
         Debt.FIVE_MINS
     )
 
+    override val defaultRuleIdAliases = setOf("MayBeConstant")
```
chao2zhang, line 48: If we are using `MayBeConstant` as the name, what do you think if we change the class name to `MayBeConstant` and have `MayBeConst` to be the alias?

chao2zhang: Aliases are actually only intended for suppressions, not for configs :(

chao2zhang (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MayBeConst.kt
BraisGabin: Correct, for that reason we can change it now. We need to wait for 2.0 (if we really want to change it, I do)

schalkms (approved):

PR #3645, opened 2021-04-02
Fail the tests when we use deprecated code in our gradle plugin

BraisGabin:
What do you think about adding this to our tests? We should not use deprecated code in our plugin to avoid compatibility problems with future versions of gradle.

I must say that I don't know gradle enough to fix the current warnings so if someone knows how please feel free to go for it.

This is more an issue than a PR but I created it like this so you can see the output of the tests. What do you think, should we have this enable by default? Could this create too much trouble because we need to give backword and forward compatibility?

### DIFF
--- a/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
+++ b/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
@@ -113,6 +113,7 @@ class DslGradleRunner @Suppress("LongParameterList") constructor(
             if (dryRun) {
                 add("-Pdetekt-dry-run=true")
             }
+            add("--warning-mode=fail")
             addAll(tasks.toList())
         }
 


schalkms:
I like the idea. It makes sense to get rid of deprecated code as soon as possible.

cortinico (commented):
> I must say that I don't know gradle enough to fix the current warnings so if someone knows how please feel free to go for it.

I generally like the idea. However, I'm unsure if this is realistic 🤔

BraisGabin:
What do you mean with realistic? If we want to support the new versions we need to remove those deprecations, right? And, right now, we don't even know which apis we are using that are deprecated.

I imagine that this could slow us down updating to new versions of gradle but that should not be a big deal. We don't need to use the last version of detekt all the time.

3flex:
> If we want to support the new versions we need to remove those deprecations, right?

It depends. Sometimes things are deprecated in Gradle and marked for removal in the next major release, then that doesn't happen for one reason or another and the removal is deferred until the next major release: https://docs.gradle.org/current/userguide/feature_lifecycle.html#sec:deprecated

> Once deprecated, it may be removed in the next major release. Deprecated features may be supported across major releases, but this is not guaranteed.

I think we can be pragmatic and replace deprecated options & features when we can as long as we can keep supporting our minimum version (which should align with Kotlin's minimum supported version), but if that means building workarounds just to avoid the deprecation warning then I say we just live with the warning until we're forced to act when the deprecated code is actually removed from Gradle.

BraisGabin:
> I think we can be pragmatic and replace deprecated options & features when we can as long as we can keep supporting our minimum version (which should align with Kotlin's minimum supported version), but if that means building workarounds just to avoid the deprecation warning then I say we just live with the warning until we're forced to act when the deprecated code is actually removed from Gradle.

I agree and disagree. I don't like the workarounds. Usually they just make the code less readable and we don't remove them as soon as they are not needed so in the future no one knows why that's like that.

But the problem is that those deprecation are displayed to our users and that's not nice... But I think that at this moment the first point is more important than the second.

I hope kotlin updates their minimum gradle version faster so we don't need to support that much versions. I'm closing this.

PR #2981, opened 2020-08-18
Support Kotlin 1.4

arturbosch:

PR #3646, opened 2021-04-02
Warnings as errors

BraisGabin:
This PR removes all the current warnigns in the project and enables `warningsAsErrors` in CI as before #2981. It was removed for some problems with Kotlin 1.4 that are fixed now so it's safe to add it again.


detekt-parser/src/main/kotlin/io/github/detekt/parser/DetektPomModel.kt
```diff
@@ -21,8 +21,8 @@ class DetektPomModel(project: Project) : UserDataHolderBase(), PomModel {
     init {
         val extension = "org.jetbrains.kotlin.com.intellij.treeCopyHandler"
         val extensionClass = TreeCopyHandler::class.java.name
-        synchronized(Extensions.getRootArea()) {
-            arrayOf(project.extensionArea, Extensions.getRootArea())
+        synchronized(@Suppress("DEPRECATION") Extensions.getRootArea()) {
+            arrayOf(project.extensionArea, @Suppress("DEPRECATION") Extensions.getRootArea())
```
BraisGabin, line 25: I have no idea what this code does and I can't find any alternative to this. The other suppressions are safe because we are suppressing our own deprecations and that's fine. But this case is different. What do you think? The problem is that if we don't fix this we can't enable `warningsAsErrors` again.

Other option would be to disable `warningsAsErrors` just for this module.

What do you think?

chao2zhang: https://github.com/pinterest/ktlint/blob/235a414be565a49568c26f1b30d396d347efbd0f/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/internal/KotlinPsiFileFactory.kt#L111

The code here seems to enable AST mutations.

BraisGabin: :confused: so ktlint still have that deprecated code... Any idea about how to fix this deprecation properly? How should we proceed here?

chao2zhang: The replacement can be found here https://github.com/JetBrains/intellij-community/blob/master/platform/extensions/src/com/intellij/openapi/extensions/Extensions.java#L34. Would you mind trying it and verify the AST mutation still works?

chao2zhang (approved):
We can delay `Extensions.getRootArea()` to a follow-PR. The urgency is to avoid adding new violations as warnings.

BraisGabin:
I created #3683 so we don't forget about it.

PR #3647, opened 2021-04-02
Drop support to Gradle 5

BraisGabin:
Drop support to gradle 5.

5.0 was release at [16/Apr/2019](https://gradle.org/releases/). This moves the support 6.0. It was release at [08/Nov/2019](https://gradle.org/releases/). Maybe we coud be even more aggresive here but I think that 6.0 is a safe commitment.

As a reference the Gradle Android Plugin set the [minimum gradle version to 6.1 at April 2020]( https://developer.android.com/studio/releases/gradle-plugin#4-0-0). And the last version of the AGP requires Gradle 6.5

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ You can find [other ways to install detekt here](https://detekt.github.io/detekt
 
 #### with Gradle
 
-Gradle 5.4+ is required:
+Gradle 6.0+ is required:
 
 ```kotlin
 buildscript {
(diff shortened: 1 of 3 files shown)


chao2zhang (approved):

3flex (changes requested):
Kotlin's current minimum supported Gradle version is 5.4. Why not keep the versions aligned? Increasing the minimum now seems arbitrary.

That said, I believe Kotlin 1.5 will bump the minimum supported version to 6.1 so when that's released and detekt updates to 1.5 I'd support an update to 6.1.

BraisGabin:
Good reason! Yes, let's keep us aligned with the Kotlin Gradle Plugin.

BraisGabin:
> That said, I believe Kotlin 1.5 will bump the minimum supported version to 6.1 so when that's released and detekt updates to 1.5 I'd support an update to 6.1.

I can not find anywhere about this bump of the minimum version. Do this happen?

3flex:
It did: https://kotlinlang.org/docs/gradle.html#plugin-and-versions

Issue #3102, opened 2020-09-25
Does detekt check the test code?


levelrin:
### Motivation

I've noticed that the `detekt` Gradle task does not check the test code.
I want to analyze my test code as well.

### Folder Structure

This is my folder structure (following the maven convention):
```
src
  - main <- source code under this is analyzed
    - kotlin
      - com.levelrin
        - Main.kt
  - test <- source code under this is NOT analyzed
    - kotlin
      - com.levelrin
        - MainTest.kt
```

### Build Script

This is my `build.gradle`:
```groovy
plugins {
    id 'org.jetbrains.kotlin.js' version '1.4.10'
    id 'io.gitlab.arturbosch.detekt' version '1.13.1'
}

repositories {
    mavenCentral()
    jcenter()
}

dependencies {
    testImplementation 'org.jetbrains.kotlin:kotlin-test-js'
}

kotlin {
    js {
        browser {
            binaries.executable()
            webpackTask {
                cssSupport.enabled = true
            }
            runTask {
                cssSupport.enabled = true
            }
            testTask {
                useKarma {
                    useChromeHeadless()
                    webpackConfig.cssSupport.enabled = true
                }
            }
        }
    }
}

detekt {
    failFast = true // fail build on any finding
    //buildUponDefaultConfig = true // preconfigure defaults
    config = files("$projectDir/config/detekt.yml")
    reports {
        html.enabled = true // observe findings in your browser with structure and code snippets
        xml.enabled = true // checkstyle like format mainly for integrations like Jenkins
        txt.enabled = true // similar to the console output, contains issue signature to manually edit baseline files
    }
}

```

### Question

Does `detekt` support analysis for the test code?

schalkms:
Yes, your impression is correct. The **detekt** task by default doesn’t run on test code.
The detekt website including it’s docs describe the **detektTest**, which you most likely want. Please see [here](https://detekt.github.io/detekt/gradle.html#available-plugin-tasks) for more information.

PR #3649, opened 2021-04-03
Add src/test/java and src/test/kotlin by default in PlainDetekt

BraisGabin:
We aren't adding the source sets `src/test/java` and `src/test/kotlin` in the PlainDetekt. It have little sense that we don't add those by default because our default configuration take those sources into account.


chao2zhang:
> It have little sense that we don't add those by default because our default configuration take those sources into account.

Agree 👍

schalkms (approved):

chao2zhang (approved):

cortinico:
This is something we want to mention specifically in the changelog right? I expect users that are using the default config to get a lot of violation for their tests: https://github.com/detekt/detekt/issues/3102#issuecomment-699093488

BraisGabin:
Agree, this is not housekeeping

PR #3650, opened 2021-04-03
Merge JaCoCo coverage reports the "right" way

3flex:
Advantages over the current config are that there's no [cross project configuration](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins_vs_cross_configuration) using `subprojects` (which is discouraged), and that the generated HTML report now renders properly from the merged JaCoCo task outputs.

Although more advanced/technical than typical Gradle build scripts this technique is actually the recommended way to [share outputs between projects](https://docs.gradle.org/current/userguide/cross_project_publications.html).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -2,8 +2,6 @@ import io.gitlab.arturbosch.detekt.Detekt
 import io.gitlab.arturbosch.detekt.DetektCreateBaselineTask
 
 plugins {
-    kotlin("jvm") apply false
-    jacoco
     packaging
     releasing
     detekt
@@ -14,33 +12,6 @@ plugins {
 allprojects {
     group = "io.gitlab.arturbosch.detekt"
     version = Versions.currentOrSnapshot()
-}
-
-jacoco.toolVersion = libs.versions.jacoco.get()
-
-tasks {
-    jacocoTestReport {
-        executionData.setFrom(fileTree(project.rootDir.absolutePath).include("**/build/jacoco/*.exec"))
-
-        val examplesOrTestUtils = setOf(
-            "detekt-bom",
-            "detekt-test",
-            "detekt-test-utils",
-            "detekt-sample-extensions"
-        )
-
-        subprojects
-            .filterNot { it.name in examplesOrTestUtils }
-            .forEach {
-                this@jacocoTestReport.sourceSets(it.sourceSets.main.get())
-                this@jacocoTestReport.dependsOn(it.tasks.test)
-            }
-
-        reports {
-            xml.isEnabled = true
-            xml.destination = file("$buildDir/reports/jacoco/report.xml")
-        }
-    }
 }
 
 val analysisDir = file(projectDir)
--- a/.github/workflows/codecoverage.yaml
+++ b/.github/workflows/codecoverage.yaml
@@ -25,7 +25,7 @@ jobs:
           java-version: 11
 
       - name: Generate Coverage Report
-        run: ./gradlew jacocoTestReport --no-build-cache -x :detekt-gradle-plugin:test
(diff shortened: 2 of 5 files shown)


build.gradle.kts
```diff
@@ -18,28 +18,81 @@ allprojects {
 
 jacoco.toolVersion = Versions.JACOCO
 
-tasks {
-    jacocoTestReport {
-        executionData.setFrom(fileTree(project.rootDir.absolutePath).include("**/build/jacoco/*.exec"))
+dependencies {
+    implementation(project("custom-checks"))
```
chao2zhang, line 22: Using the `implementation` configuration feels weird: Would that create unnecessary tasks by the plugins applied in this build.gradle?

I am wondering if we could use a different configuration, or move the jacoco to a separate module.

chao2zhang (commented):
> Advantages over the current config are that there's no [cross project configuration](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins_vs_cross_configuration) using `subprojects` (which is discouraged), and that the generated HTML report now renders properly.

To paraphrase: If module B's test code covers module A's code, then module A's code would be counted as "covered" in jacoco. Is that correct?

> Coverage report on my machine seems slightly improved with this config as well.

Would you mind sharing the jacoco output? (I assume our detekt-gradle-plugin code cannot be correctly calculated still)

build.gradle.kts
3flex: Moving jacoco generation to a separate project/module makes sense to me. I believe a different configuration would work instead of `implementation` too. I can make those changes.

> Would that create unnecessary tasks by the plugins applied in this build.gradle?

I don't believe so. But if there were unnecessary tasks, and they're not created during the configuration phase, there'll be no penalty to build performance. This is a `jacocoTestReport` [build scan](https://scans.gradle.com/s/4i3hjg2j737ca/performance/configuration) and a `codeCoverageReport` [build scan](https://scans.gradle.com/s/r2abguor2sxem/performance/configuration) for comparison. The only tasks created immediately or during configuration are from nexus-staging, binary-compatibility-validator, gradle.plugin-publish, github-release and sonarqube plugins.

The benefit of using configurations like this is it's more declarative so manually setting up task dependencies can be avoided, and only the tasks that need to be run are run. The disadvantage is the boilerplate.

3flex:
> To paraphrase: If module B's test code covers module A's code, then module A's code would be counted as "covered" in jacoco. Is that correct?

I believe that's already happening - this change is just about getting the data from each subproject's JaCoCo output, source files and compiled classes in the "right" way. So instead of the merge task reaching into directories manually, the relevant files are packaged by the new configurations (like `coverageDataElements`) in each project then that configuration is consumed by the new configurations in the root Gradle project. This helps decouple the builds.

> Would you mind sharing the jacoco output?

I'm not sure if the line coverage differential in the JaCoCo HTML report run on my machine is actually consequential - now the report is picked up by Codecov you can see there's no difference in coverage reported when using the new task.

build.gradle.kts
chao2zhang: Can the following block of build script be removed
```

        subprojects
            .filterNot { it.name in examplesOrTestUtils }
            .forEach {
                this@jacocoTestReport.sourceSets(it.sourceSets.main.get())
                this@jacocoTestReport.dependsOn(it.tasks.test)
            }
```

I don't know jacoco enough to realize the actual benefit. The boilderplate might be fine since we do not add/change gradle modules often.

BraisGabin (approved):

cortinico (approved):

PR #3651, opened 2021-04-04
[RFC] New Rule API

BraisGabin:
This is a prove of concept for a new API for our rules.

The base idea is that a rule is just this:

```
interface NewRule : (KtFile) -> List<Finding> {

    override fun invoke(file: KtFile): List<Finding>
}
```

You don't even need to implement `KtTreeVisitor` if you don't need it for your rule.

This interface is really abstract so it force a lot of boilerplate for nearly all the use-cases. For this reason I implemented `BaseNewRule` that provides (more less) the same functionality as `BaseRule` without the boilerplate.

My focus right now is in the core module and how I need to adapt it to use the new API. The idea is to use Flow and suspend functions to have as much parallelism as possible. This change is HUGE because it force me to reimplement a lot of the core code but I think that it's worth it.

Other key points of this refactor is to make the core more composable. You can see this composability in the file `Runner.kt`. I need multiple pieces to run the analyzer and I get them by parameter (even the analyzer) and then that function orchestrates all of those parameters. This should improve the maintenability because it will be easier to unit test each piece.

Next steps to solve:
- How to get a `Filer` instance for each Rule.
- Implement the baseline
- Implement the reports
- The current FileProcessingListener API lost some feature. This should be solved
- I'm using the old code to get the `KtFile`s and the `BindingContext`. And that code blocks. So it should be refactored.
- We need to expend the new Rule API to allow to perform analysis with all the files of the project. More about this here: https://github.com/detekt/detekt/discussions/3610#discussioncomment-539533

Note: I created this to have the branch in the main repo but it's the same idea as #3127 (You can read it to get more context).


BraisGabin:
There is a big change in the API between this PR and the previous one: I removed `PlainRule` and `TypeSolvingRule`. The idea was cool but It wasn't the best solution. `ResolvedContext` can be injected in the rules so the analyzer doesn't need to know about this class.

Now this responsability is moved to the `CollectionRuleProvider`. If it has a `BindingContext` it can instantiate all the rules if it has not it could decide what to do. We know that we don't want mixed rules but this API allow our users to implement them if they want to do it.

BraisGabin:
This PR is getting bigger and bigger and I'm not sure how to split it.

If you see the api it's small and I'd like to keep it like this. We can create helpers on top of it to make it easier to implement the more general cases. But the api v2 should allow to implement any rule or extension.

The problem of this change is that core should be rewritten nearly from scratch. And for this reason ths PR is getting bigger and bigger.

My next step is to focus in the settings. `:detekt-core` should expose an `interface` to define how it wants the settings and then it should be the responsability of each client (cli, gradle, maven, whatever) to provide an instance of that.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Finding.kt
```diff
@@ -0,0 +1,71 @@
...
+
+interface Location {
+    val source: SourceLocation
+    val text: TextLocation
+    val filePath: FilePath // TODO The api should not depend on io.github.detekt.psi.FilePath
+}
+
+interface SourceLocation {
+    val line: Int
+    val column: Int
```
chao2zhang, line 33: I had a recent finding from checkstyle that some violations do not need to have a column: For example, MaxLineLength is reported on a single line: We could specify a default value `UNDEFINED` for column if not specified.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Finding.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    operator fun component1(): Int = line
+    operator fun component2(): Int = column
+
+    companion object {
+        operator fun invoke(line: Int, column: Int): SourceLocation {
+            return Impl(line, column)
+        }
+    }
```
chao2zhang, line 42: Why don't we write factory method?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/v2/providers/FileProcessListenersProvider.kt
```diff
@@ -0,0 +1,46 @@
...
+) : FileProcessListenersProvider {
+
+    constructor(
+        settings: ProcessingSettings
+    ) : this(
+        flow {
+            emitAll(
+                ServiceLoader.load(CollectionFileProcessListenerProvider::class.java, settings.pluginLoader).asFlow()
+            )
+        },
```
chao2zhang, line 36:
```suggestion
            ServiceLoader.load(CollectionFileProcessListenerProvider::class.java, settings.pluginLoader).asFlow(),

```

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/v2/providers/ReportingModifiersProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+) : ReportingModifiersProvider {
+
+    constructor(
+        settings: ProcessingSettings
+    ) : this(
+        flow {
+            emitAll(
+                ServiceLoader.load(CollectionReportingModifierProvider::class.java, settings.pluginLoader).asFlow()
+            )
+        },
```
chao2zhang, line 34:
```suggestion
        ServiceLoader.load(CollectionReportingModifierProvider::class.java, settings.pluginLoader).asFlow()
```

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/v2/providers/OutputReportersProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+) : OutputReportersProvider {
+
+    constructor(
+        settings: ProcessingSettings
+    ) : this(
+        flow {
+            emitAll(
+                ServiceLoader.load(CollectionOutputReporterProvider::class.java, settings.pluginLoader).asFlow()
+            )
+        },
```
chao2zhang, line 33:
```suggestion
        ServiceLoader.load(CollectionOutputReporterProvider::class.java, settings.pluginLoader).asFlow(),
```

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/v2/providers/FileProcessListenersProvider.kt
```diff
@@ -0,0 +1,46 @@
...
+        settings,
+    )
+
+    override suspend fun get(resolvedContext: Deferred<ResolvedContext>): List<FileProcessListener> {
+        return collectionFileProcessListenerProviders
+            .flatMapMerge { collectionProvider -> collectionProvider.get(setupContext, resolvedContext) }
```
chao2zhang, line 42: I realized that `.flatMapMerge` actually allows us to parallelize operations which was not possible before.

But on the other hand, why do we return `List<FileProcessListener>` instead of `Flow<FileProcessListener>` - We could still order them by priority but the actual `FileProcessListener.onProcess()` can be parallelized.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/v2/Analyzer.kt
```diff
@@ -0,0 +1,70 @@
...
+            }
+        }
+        .flatMapMerge { file ->
+            flow {
+                val path = file.absolutePath()
+                val findings = rules
+                    .filter { (_, filter) -> filter.filter(path) }
+                    .flatMapMerge { (rule, filter) ->
+                        rule.invoke(file)
+                            .filter { filter.filter(it) }
```
chao2zhang, line 49: Could you elaborate on what this step does? Is there a case that we want to run a Rule but let the filter findings?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
```diff
@@ -0,0 +1,21 @@
...
+
+interface Rule : (KtFile) -> List<Issue>{
+    val id: String
+    val description: String
+
+    override fun invoke(file: KtFile): List<Issue>
+}
+
+interface Issue {
+    val entity: Entity // TODO We should rethink this class. I'm using the old because I don't want to address this yet
```
chao2zhang, line 16: Actually the existing v1 `Issue` does not point to the entity, we probably need a descriptor or metadata of `Rule` (https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317556)

My impression is that a Detekt `Issue` should be the metadata that is unique per rule while a Detekt `Finding` contains the actual violations and should refer to `Entity` instead.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/v2/providers/FileProcessListenersProvider.kt
BraisGabin: Your code is eagerly creating the instances. I want it to be lazy to get agvantage of multithread.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/v2/Analyzer.kt
BraisGabin: Sure, this is something that detekt doesn't have right now. You know that we get a los of issues talking about how to supress some rules in some special cases (Dagger, JUnit, Sprint, etc.). We usually refuse those issues and say that they should write a custom rule changing the original one.

The idea behind this is to have a more powerful api regarding suppression. You can see that each rule has it's own Filter and it can be implemented in any way.

Using this API we could implement things like:

```yaml
  UnsafeCallOnNullableType:
    active: true
    excludeIfFunctionName: [...] 
    excludeIfClassName: [...]
    excludeIfImplementsInterface: [...]
    excludeIfTopFunction: [...]
    excludeIfAnnotation: [...]
```

And every rule could have all of those. No more custom flags to supress some issues is some rules. In this PR I don't want to implement this yet. I'm just implementing an API and core that allow us to create things like this. And I don't like the names and how the yaml looks, they are just to illustrate the idea. We will see how to implement it in the future.

If we end up writing our setting using a Kotlin DSL we could even allow the user to write their custom filters using lambdas, and that would be SO good.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt
chao2zhang: Let me try with a different way: `Issue` is the metadata and only contains information for each `Rule`, `Finding` is contextual, it includes necessary information for any detected findings in user's codebase. 

Therefore, I don't think `Entity` should be a property of `Issue`. `Entity` is not a property of `Issue` V1 either.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/v2/Analyzer.kt
chao2zhang: 👍  This looks like a good generalization overall. A line of comment should be sufficient enough.

BraisGabin:
I want to change the responsabilities of `core`. `core` should not be responsable to know how to load the the `KtFile`s, `Rule`s nor any other instance. It just receive providers and it call them in the correct other to satisfy the dependencies.

This make that each client should do that work instead. My idea is to create a module that provides a common implementation so any client can use it. But this will allow other clients to be smart about how to provide each thing dependency.

With this I want to avoid the usage of `detekt-cli` in `detekt-gradle-plugin` too. The idea is that all that code that is between `core` and `cli` related with the configuration should be moved to a different module so any client can use those default implementations if they want.

`core` should exposes this function:

```
suspend fun run(
    filesProvider: KtFilesProvider,
    resolvedContextProvider: ResolvedContextProvider,
    ruleProvider: RulesProvider,
    fileProcessListenersProvider: FileProcessListenersProvider,
    reportingModifiersProvider: ReportingModifiersProvider,
    consoleReportersProvider: ConsoleReportersProvider,
    outputReportersProvider: OutputReportersProvider,
): Detektion
```

------------

Right now `core` does the analysys, postAnalysis and reporting. I'm not sure if the reporting should be part of the core at all. If we want to wrok with unified reports that only have sense in the `gradle` client. But it doesn't have sense for `cli` so that's a hint that, maybe, even the reporting should be moved out of core. That way `core` will just do the `analysis` and `postAnalysis` and return a list of all the issues.

PR #3655, opened 2021-04-05
Noisy gradle

### LINKED ISSUE #3019: detektMain is noisy "Ignoring a file detekt cannot handle"
## Expected Behavior
This kind of messages doesn't appear
```
> Task :common:core:detektDebugUnitTest
Ignoring a file detekt cannot handle: /Users/brais.gabin/projects/XXXXX/Class.java
```
## Current Behavior
If I run `./gradlew detekt` this messages never appear. But if I run `./gradlew detektMain` or `./gradlew detektTest` they do appear.

## Context
Now that we have support to Android variants I'm moving my detekt configuration to use `detektMain` instead of `detekt` so I can use Type resolutions

BraisGabin:
Fix #3019

The JVM and the Android tasks are sending non-kotlin files as input to the client. This was configured properly in Plain detekt but not in the others. This PR fixes that.

> Ignoring a file detekt cannot handle: /Users/brais.gabin/ranom/path/AClass.java

_This PR was originally opened here #3648 but I messed up with a merge and it end up closed automatically._

### DIFF
--- a/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
+++ b/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
@@ -97,6 +97,8 @@ class DslGradleRunner @Suppress("LongParameterList") constructor(
         writeKtFile(File(rootDir, srcDir), className)
     }
 
+    fun Submodule.projectFile(path: String): File = File(moduleRoot, path).canonicalFile
+
     private fun writeKtFile(dir: File, className: String, withCodeSmell: Boolean = false) {
         dir.mkdirs()
         File(dir, "$className.kt").writeText(ktFileContent(className, withCodeSmell))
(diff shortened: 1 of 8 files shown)


detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
```diff
@@ -157,3 +159,21 @@ class DslGradleRunner @Suppress("LongParameterList") constructor(
...
+fun DslGradleRunner.createJavaClass(name: String) {
+    projectFile("${projectLayout.srcDirs.first { it.contains("main") }}/$name.java")
+        .apply { createNewFile() }
+        .writeText("public class $name {}")
+    projectFile("${projectLayout.srcDirs.first { it.contains("test") }}/${name}Test.java")
+        .apply { createNewFile() }
+        .writeText("public class ${name}Test {}")
+}
+
+fun DslGradleRunner.createJavaClass(submodule: Submodule, name: String) {
```
chao2zhang, line 172: For these extension methods - Do we think it would be better to modularize as:
`fun DslGradleRunner.findMainSrcDir()`
`fun DslGradleRunner.findTestSrcDir()`
`fun Submodule.findMainSrcDir()`
`fun Submodule.findTestSrcDir()`
and
`fun writeJavaFile()`

It might be confusing to have `createJavaClass()` write to both main and test.

chao2zhang (approved):
Thank you for fixing this!
(I left a comment but feel free to drop, the DslGradleRunner is a bit messy when I read the code even before this PR)

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
BraisGabin: I agree... I think that we should not have a difference between module and submodule. Both hae the same features. This DSL needs a refactor but I'm not sure if it's worth it.

BraisGabin: I can't implement this:

`fun Submodule.findMainSrcDir()`
`fun Submodule.findTestSrcDir()`

Because a `Submodule` doesn't have the information about `DslGradleRunner` and I need the `DslGradleRunner` to get the base path.

I'd vote to merge this and if you think that we should rethinkg the `DslGradleRunner` we can open an issue and see how can we refactor it.

chao2zhang: We probably just need housekeeping work on `DslGradleRunner`

Issue #3423, opened 2021-01-29
Show detailed error messages


chao2zhang:
## Expected Behavior
When a final newline is reported, the error message will be shown in the console:
```
/Users/cazhang/detekt/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt:1:1: File must end with a newline (\n)
```

## Current Behavior
The error message is shown in the console
```
formatting - 5min debt
        FinalNewline - [] at /Users/cazhang/detekt/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt:1:1
```

## Context
I have received reports from other devs complaining they don't know what FinalNewLine is (Probably related to the line and column, but this is a fix for upstream(ktlint). However, I found that ktlint shows detailed error messages which help people to understand the issue.

schalkms:
I'm not sure whether this should be done for the console report. The console report could become quite verbose.

chao2zhang:
We can provide an option to specify the verbosity of output. The issue description and message description are both shown in the SARIF report output (See https://github.com/chao2zhang/detekt/pull/4/files) and most of the rules are currently displaying the same message as the issue description

schalkms:
Yes, that’s true. I always root for having the console and XML report to display concise information. For more detailed information there’s the HTML or SARIF report available.

schalkms:
@cortinico & @BraisGabin what's your take on this? I think we should either close this issue or tag it as improvement depending on the approved decision.

cortinico:
> I think we should either close this issue or tag it as improvement

Potentially this

I'm curious to know what's the message we receive from ktlint here:
https://github.com/detekt/detekt/blob/f5b6b04b0822b670b206015c6cf91706e73467b1/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt#L89

Isn't that `File must end with a newline (\n)`?

cortinico:
> @cortinico & @BraisGabin what's your take on this? I think we should either close this issue or tag it as improvement depending on the approved decision.

Closing as the agreement seems to be leaning towards that direction. Users can still enable other type of reports + some of the rules are reporting long messages (which is good for HTML report and so on), but will end up polluting the console.

PR #3656, opened 2021-04-05
Print top 5 smells when failing on threshold

pinkasey:
When detekt build fails due to threshold (see `maxIssues`), the concluding error message does not help you find what got you over the threshold. Instead, it sends you to go over the log, and look for issues found.
This has two problems:
1. The build log can be very long, and it may take time to go over it
2. I case you have many existing issues with small weight, and one new issue with big weight - it becomes a search for a needle in a hay-stack.

#### Background
In my team, adoption of Detekt is slow. There are some that embrace Detekt, some that are indifferent, and some who would rather see it removed. The main issue is the question "what rules should we adopt?", combined with the fact that when we added Detekt we already had thousands of warnings.
So after tweaking and disabling some rules, and adding some custom ones, we now have ±500.
But we still don't want to fail the build over any minor issue that is yet in dissagreement.
So the solution was to pick a few that we consider **critical** (all of them custom rules that reveal a certain bug), **give them 
weight `10001`, and set the threshold to `10000`**.
#### Problem
Now the build fails only on these "critical" issues, but when it does - it's hard to tell by the build's log.
Running the plugin in IntelliJ helps, but it's another step to take, and you'd have to make sure you're on the exact commit etc.

#### Solution
What this commit aims to do, is change the print in case of threshold-failure, to something like this:
```
Build failed with 300 weighted issues. Top 5:
  10001      ExpectWithoutToBe   /src/test/foo/BarTest.kt:56
  1          MaximumLineLength   /src/main/foo/Bar.kt:145
  1          MaximumLineLength   /src/main/foo/Qux.kt:172
  1          ComplexCondition    /src/main/foo/Quux.kt:172
  1          LongMethod          /src/main/foo/Quuz.kt:172
```
Now the user knows immediately exactly what broke the build

I didn't know about [custom reports](https://detekt.github.io/detekt/extensions.html#custom-reports) API when I wrote this,
but I think this is a reasonable default format for this message - I think everyone can benefit from it. I'm guessing there are other Detekt users that use the threshold mechanism in a similar way.

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/IssueExtension.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/IssueExtension.kt
@@ -1,10 +1,6 @@
 package io.gitlab.arturbosch.detekt.core.config
 
-import io.gitlab.arturbosch.detekt.api.Config
-import io.gitlab.arturbosch.detekt.api.Detektion
-import io.gitlab.arturbosch.detekt.api.Finding
-import io.gitlab.arturbosch.detekt.api.RuleId
-import io.gitlab.arturbosch.detekt.api.RuleSetId
+import io.gitlab.arturbosch.detekt.api.*
 import io.gitlab.arturbosch.detekt.core.reporting.BUILD
 import io.gitlab.arturbosch.detekt.core.reporting.filterAutoCorrectedIssues
 import org.jetbrains.kotlin.com.intellij.openapi.util.Key
@@ -18,8 +14,13 @@ internal fun Detektion.getOrComputeWeightedAmountOfIssues(config: Config): Int {
     if (maybeAmount != null) {
         return maybeAmount
     }
+    val smellsWithWeights = getIssuesWithWeights(config)
+    val amount = smellsWithWeights.sumBy { it.second }
+    this.addData(WEIGHTED_ISSUES_COUNT_KEY, amount)
+    return amount
+}
 
-    val smells = filterAutoCorrectedIssues(config).flatMap { it.value }
+internal fun Detektion.getIssuesWithWeights(config: Config): List<Pair<Finding, Int>> {
     val ruleToRuleSetId = extractRuleToRuleSetIdMap(this)
     val weightsConfig = config.subConfig(BUILD).subConfig(WEIGHTS)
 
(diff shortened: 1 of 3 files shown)


BraisGabin:
Could you explain the reason of this PR? Do you know the [custom reports](https://detekt.github.io/detekt/extensions.html#custom-reports) API? Isn't it enough for your use case? What do you miss in it?

pinkasey:
> Could you explain the reason of this PR? Do you know the [custom reports](https://detekt.github.io/detekt/extensions.html#custom-reports) API? Isn't it enough for your use case? What do you miss in it?

Thanks!
I didn't know about custom reports API when I wrote this,
but I think this is a reasonable default format for this message - I think everyone can benefit from it. I'm guessing there are other Detekt users that use the threshold mechanism in a similar way.

pinkasey:
(I still need to fix and add tests)

pinkasey:
@BraisGabin -
Does the idea of this PR sound reasonable?
Would it be accepted, once I fix everything and add tests?

I want to know if to invest in this.
If you give me a general OK, I'll fix and add tests, and ask for review when done.

BraisGabin:
I think that we don't need this new report. Right now the html report shows the number of fails for each rule. It's not sorted by that number but we could add some javascript to the report to get that behaviour. Somthing like:

Sort by:
- set/rule (alphabetic)
- rule (number of issues)
- files (this would be out of the scope of this issue but I think that it could be useful too)

I see the point to have a report like yours in some projects but I don't think it would be used that much to add it in the main project.

Kind of related with #3668 (not the same feature but they are similar)

I'd like to know the opinion of other maintainers too.

roiwexler:
this report style would be great!

chao2zhang:
I would vote for using a custom extension to configure the report - Every team would like to customize the console output like https://github.com/detekt/detekt/issues/3423.

If your codebase has too many violations, I would suggest disabling those rules with extensive violations in the beginning, then enabling them slowly. This process is quite a chore, so my team set the Detekt upgrading policy by every time we need to upgrade Android Gradle Plugin.

chao2zhang:
Additionally, top 5 smells do not work really well with per-module / per-variant detekt, as there can be several `Detekt` gradle tasks that failed and we would print top 5 smells multiple times in the console.

BraisGabin:
Thanks for the proposal but for the reasons explained above I'm going to close this. I think that this is a perfect example for a custom report. If you create it, please let us know so we can link it so other people with the same needs can find it.

Issue #3415, opened 2021-01-27
False positive on `UnusedPrivateMember` for actual/expected inline method parameters.


Whathecode:
## Expected Behavior
Expected inline methods in a Kotlin multiplatform project should not give `UnusedPrivateMember` warning for parameters.

## Observed Behavior/Steps to Reproduce
```
internal expect object Reflection
{
    internal inline fun <reified T> extendsType( klass: KClass<*> ): Boolean
}
```

This gives a `UnusedPrivateMember` warning for the `klass` parameter.

## Your Environment

* Version of detekt used: 1.16.0-RC1
* Version of Gradle used (if applicable): 6.6
* Operating System and version: Windows 10
* Link to your project (if it's a public repository): https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/reflect/ReflectionAvailable.kt#L32

cortinico:
Thanks for the report. The `expect object` scenario is indeed not covered by this rule. Fix is in #3417

Whathecode:
@schalkms I just updated to 1.16.0-RC2 This is now fixed for the `expect` definition, but still gives the same warning on my actual implementation.

```
internal actual object Reflection
{
    internal actual inline fun <reified T> extendsType( klass: KClass<*> ): Boolean =
        throw UnsupportedOperationException()
}
```

cortinico:
While working on this I noticed that we have a misalignment between `expect` and `actual` and it seems to be intentional: https://github.com/detekt/detekt/pull/2643#discussion_r416774303

@schalkms is that still relevant?

schalkms:
I don't think it's intentional. The fix incorporated for this issue should include the check for the `actual` definition. There seems to be a misalignment between the `actual` and `expect` in the rule implementation.

https://github.com/detekt/detekt/blob/c5f65a9c8ed029f0382674371c070ec0f3071b28/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt#L204

cortinico:
> There seems to be a misalignment between the `actual` and `expect` in the rule implementation.

Yup that's exactly why I asked. In your comment here: https://github.com/detekt/detekt/pull/2643#discussion_r416774303 you mentioned:

> Actual classes are an implementation concern. Thus, unused parameters should be reported.

[...]

PR #3669, opened 2021-04-10
Ignore actual members in UnusedPrivateMember

sczerwinski:
The general rule of thumb should be:
- if a rule is ignored for `abstract` members, it should be ignored for `expect` members,
- if a rule is ignored for **overridden** members, it should be ignored for `actual` members.

See also:
- #3388
- #3390

Related to #3415

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
@@ -11,6 +11,7 @@ import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
 import io.gitlab.arturbosch.detekt.rules.isAbstract
+import io.gitlab.arturbosch.detekt.rules.isActual
 import io.gitlab.arturbosch.detekt.rules.isExpect
 import io.gitlab.arturbosch.detekt.rules.isExternal
 import io.gitlab.arturbosch.detekt.rules.isMainFunction
@@ -255,7 +256,8 @@ private class UnusedParameterVisitor(allowedNames: Regex) : UnusedMemberVisitor(
     private fun KtNamedFunction.isRelevant() = !isAllowedToHaveUnusedParameters()
 
     private fun KtNamedFunction.isAllowedToHaveUnusedParameters() =
-        isAbstract() || isOpen() || isOverride() || isOperator() || isMainFunction() || isExternal() || isExpect()
+        isAbstract() || isOpen() || isOverride() || isOperator() || isMainFunction() || isExternal() ||
+            isExpect() || isActual()
 }
 
 private class UnusedPropertyVisitor(allowedNames: Regex) : UnusedMemberVisitor(allowedNames) {
@@ -292,7 +294,10 @@ private class UnusedPropertyVisitor(allowedNames: Regex) : UnusedMemberVisitor(a
     override fun visitPrimaryConstructor(constructor: KtPrimaryConstructor) {
         super.visitPrimaryConstructor(constructor)
         constructor.valueParameters
-            .filter { (it.isPrivate() || !it.hasValOrVar()) && it.containingClassOrObject?.isExpect() == false }
+            .filter {
+                (it.isPrivate() || (!it.hasValOrVar() && !constructor.isActual())) &&
+                    it.containingClassOrObject?.isExpect() == false
+            }
             .forEach { maybeAddUnusedProperty(it) }
     }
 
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -204,7 +205,7 @@ private class UnusedParameterVisitor(allowedNames: Regex) : UnusedMemberVisitor(
     }
 
     override fun visitClassOrObject(klassOrObject: KtClassOrObject) {
-        if (klassOrObject.isExpect()) return
+        if (klassOrObject.isExpect() || klassOrObject.isActual()) return
```
chao2zhang, line 208: How is this related to the rule of thumb `if a rule is ignored for overridden members, it should be ignored for actual members.`?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -292,7 +294,11 @@ private class UnusedPropertyVisitor(allowedNames: Regex) : UnusedMemberVisitor(a
     override fun visitPrimaryConstructor(constructor: KtPrimaryConstructor) {
         super.visitPrimaryConstructor(constructor)
         constructor.valueParameters
-            .filter { (it.isPrivate() || !it.hasValOrVar()) && it.containingClassOrObject?.isExpect() == false }
+            .filter {
+                (it.isPrivate() || !it.hasValOrVar()) &&
+                    it.containingClassOrObject?.isExpect() == false &&
+                    it.containingClassOrObject?.isActual() == false
```
chao2zhang, line 300: How is this change related to the rule of thumb `if a rule is ignored for overridden members, it should be ignored for actual members.`?

sczerwinski: There is no such thing as abstract or overridden constructors, but there can be `expect` or `actual` constructors, and those should be treated the same way as `expect`/`actual` functions (line 259-260).

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
sczerwinski: This line is actually my mistake (I went with `isExpect()` one step too far). Actual classes or objects should be visited. Just their actual members should be ignored.

I'll update the PR.

sczerwinski:
@chao2zhang, Sorry for the mistake. I've fixed the PR, rebased and squashed for better readability.

Also an example of an actual constructor, that should be ignored:
**commonMain:**
```kotlin
expect class Foo(bar: Bar) {
    expect fun doSomething()
}
```
**jvmMain:**
```kotlin
actual class Foo actual constructor(private val bar: Bar) {
    actual fun doSomething() {
        bar.doSomething()
    }
}
```
**jsMain:**
```kotlin
actual class Foo actual constructor(bar: Bar) {
    actual fun doSomething() {
        // Do something with JS that doesn't require bar
    }
}
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
sczerwinski: I added more test cases and changed this snippet. My initial condition was excluding too many cases.

chao2zhang: Thanks for the detailed explanation and additional test cases.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -292,7 +294,10 @@ private class UnusedPropertyVisitor(allowedNames: Regex) : UnusedMemberVisitor(a
     override fun visitPrimaryConstructor(constructor: KtPrimaryConstructor) {
         super.visitPrimaryConstructor(constructor)
         constructor.valueParameters
-            .filter { (it.isPrivate() || !it.hasValOrVar()) && it.containingClassOrObject?.isExpect() == false }
+            .filter {
+                (it.isPrivate() || (!it.hasValOrVar() && !constructor.isActual())) &&
+                    it.containingClassOrObject?.isExpect() == false
```
chao2zhang, line 299: I am actually planning to separate UnusedPrivateParameter from UnusedPrivateMember, so the logic here should be simplified.

chao2zhang (approved):

BraisGabin (approved):

Issue #3670, opened 2021-04-10
Use @Configuration annotation instead of kdoc in all rules


marschwar:
While the annotations `@ActiveByDefault` and `@RequiresTypeResolution` from #3562 have already been applied to all rules, the `@Configuration` introduced by #3637 has yet to be used in all rules. 
I propose to adapt the rules one module at a time and use this issue to keep track of the progress.  

- [x] detekt-rules-complexity (#3768 )
- [X] detekt-rules-coroutines (nothing to do)
- [X] detekt-rules-documentation
- [X] detekt-rules-empty
- [X] detekt-rules-errorprone
- [x] detekt-rules-exceptions (#3798 )
- [X] detekt-rules-naming
- [X] detekt-rules-performance (nothing to do)
- [x] detekt-rules-style (#3774 )
- [X] custom-checks
- [X] detekt-formatting
- [X] ~~replace annotations in ruleset providers~~ (#3816 )

**Once all rules have been adapted:**
- [x] Remove the checks in `ConfigAssert#checkOptions`
- [x] Remove the `ThresholdRule` ( #3891)
- [X] Remove support for `@configuration` kdoc tag in `io.gitlab.arturbosch.detekt.generator.collection.RuleVisitor` ~~and `RuleSetProviderVisitor`~~.
- [x] Remove `LazyRegex` ( #3891)

marschwar:
I am working on `rules-complexity` ATM

BraisGabin:
I was working on `detekt-rules-exceptions` and I had some problems with the rule `SwallowedException` (I had these problems in other rules too but this example illustrates the problem).

Current kdoc with the configuration:
```kotlin
/*
 * @configuration ignoredExceptionTypes - exception types which should be ignored (both in the catch clause and body)
 * (default: `- InterruptedException
 *            - NumberFormatException
 *            - ParseException
 *            - MalformedURLException`)
 * @configuration allowedExceptionNameRegex - ignores too generic exception types which match this regex
 * (default: `'_|(ignore|expected).*'`)
 */
```

and the properties are:
```kotlin
    private val ignoredExceptionTypes = valueOrDefaultCommaSeparated(IGNORED_EXCEPTION_TYPES, defaultIgnoredExceptions)
        .map { it.removePrefix("*").removeSuffix("*") }

    private val allowedExceptionNameRegex by LazyRegex(ALLOWED_EXCEPTION_NAME_REGEX, ALLOWED_EXCEPTION_NAME)
```

[...]

PR #3675, opened 2021-04-11
Add support for fallback property

marschwar:
This PR relates to #3670 and adds support for annotated properties that have been renamed. Example:

**without annotation**
```kotlin
    private val functionThreshold: Int =
        valueOrDefault(FUNCTION_THRESHOLD, valueOrDefault(THRESHOLD, DEFAULT_FUNCTION_THRESHOLD))
```

**with annotation**
```kotlin
    private val functionThreshold: Int  by fallbackConfig(
        fallbackProperty = "threshold", 
        defaultValue = DEFAULT_FUNCTION_THRESHOLD
    )
```

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ConfigProperty.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ConfigProperty.kt
@@ -7,10 +7,10 @@ import kotlin.reflect.KProperty
 fun <T : Any> config(defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
     SimpleConfigProperty(defaultValue)
 
-fun <T : Any> fallbackConfig(fallbackProperty: String, defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
+fun <T : Any> configWithFallback(fallbackProperty: String, defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
     FallbackConfigProperty(fallbackProperty, defaultValue)
 
-fun listConfig(defaultValue: List<String>): ReadOnlyProperty<ConfigAware, List<String>> =
+fun configList(defaultValue: List<String>): ReadOnlyProperty<ConfigAware, List<String>> =
     ListConfigProperty(defaultValue)
 
 private class SimpleConfigProperty<T : Any>(private val defaultValue: T) : ReadOnlyProperty<ConfigAware, T> {
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ConfigProperty.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ConfigProperty.kt
@@ -13,12 +13,22 @@ fun <T : Any> configWithFallback(fallbackPropertyName: String, defaultValue: T):
 private fun <T : Any> getValueOrDefault(configAware: ConfigAware, propertyName: String, defaultValue: T): T {
     @Suppress("UNCHECKED_CAST")
     return when (defaultValue) {
-        is List<*> -> configAware.valueOrDefaultCommaSeparated(propertyName, defaultValue as List<String>) as T
+        is List<*> -> {
+            if (defaultValue.all { it is String }) {
+                val defaultValueAsListOfStrings = defaultValue as List<String>
+                configAware.valueOrDefaultCommaSeparated(propertyName, defaultValueAsListOfStrings) as T
+            } else {
+                error("Only lists of strings are supported. '$propertyName' is invalid. ")
+            }
+        }
         is String,
         is Boolean,
         is Int,
         is Long -> configAware.valueOrDefault(propertyName, defaultValue)
-        else -> error("${defaultValue.javaClass} is not supported as ")
+        else -> error(
+            "${defaultValue.javaClass} is not supported for delegated config property '$propertyName'. " +
+                "Use one of String, Boolean, Int, Long or List<String> instead."
+        )
     }
 }
 
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ConfigProperty.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ConfigProperty.kt
@@ -10,12 +10,21 @@ fun <T : Any> config(defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
 fun <T : Any> configWithFallback(fallbackPropertyName: String, defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
     FallbackConfigProperty(fallbackPropertyName, defaultValue)
 
-fun configList(defaultValue: List<String>): ReadOnlyProperty<ConfigAware, List<String>> =
-    ListConfigProperty(defaultValue)
+private fun <T : Any> getValueOrDefault(configAware: ConfigAware, propertyName: String, defaultValue: T): T {
+    @Suppress("UNCHECKED_CAST")
+    return when (defaultValue) {
+        is List<*> -> configAware.valueOrDefaultCommaSeparated(propertyName, defaultValue as List<String>) as T
+        is String,
+        is Boolean,
+        is Int,
+        is Long -> configAware.valueOrDefault(propertyName, defaultValue)
+        else -> error("${defaultValue.javaClass} is not supported as ")
+    }
+}
 
 private class SimpleConfigProperty<T : Any>(private val defaultValue: T) : ReadOnlyProperty<ConfigAware, T> {
     override fun getValue(thisRef: ConfigAware, property: KProperty<*>): T {
-        return thisRef.valueOrDefault(property.name, defaultValue)
+        return getValueOrDefault(thisRef, property.name, defaultValue)
     }
 }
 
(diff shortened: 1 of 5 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ConfigProperty.kt
```diff
@@ -4,21 +4,30 @@ import io.gitlab.arturbosch.detekt.api.ConfigAware
...
-fun config(defaultValue: Boolean): ReadOnlyProperty<ConfigAware, Boolean> = simpleConfig(defaultValue)
-fun config(defaultValue: List<String>): ReadOnlyProperty<ConfigAware, List<String>> = ListConfigProperty(defaultValue)
+fun config(defaultValue: List<String>): ReadOnlyProperty<ConfigAware, List<String>> =
+    ListConfigProperty(defaultValue)
 
-private fun <T : Any> simpleConfig(defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
+fun <T : Any> config(defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
     SimpleConfigProperty(defaultValue)
 
+fun <T : Any> config(fallbackProperty: String, defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
```
chao2zhang, line 13: I would recommend using separate names: `fallbackConfig`, `listConfig` etc.

chao2zhang (approved):

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ConfigProperty.kt
marschwar: Agreed. I was thinking about it and for some reason thought it would make it harder to determine the default value. But I think the opposite is true.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ConfigProperty.kt
```diff
@@ -4,21 +4,30 @@ import io.gitlab.arturbosch.detekt.api.ConfigAware
...
-fun config(defaultValue: Int): ReadOnlyProperty<ConfigAware, Int> = simpleConfig(defaultValue)
-fun config(defaultValue: Long): ReadOnlyProperty<ConfigAware, Long> = simpleConfig(defaultValue)
-fun config(defaultValue: Boolean): ReadOnlyProperty<ConfigAware, Boolean> = simpleConfig(defaultValue)
-fun config(defaultValue: List<String>): ReadOnlyProperty<ConfigAware, List<String>> = ListConfigProperty(defaultValue)
-
-private fun <T : Any> simpleConfig(defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
+fun <T : Any> config(defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
     SimpleConfigProperty(defaultValue)
 
+fun <T : Any> fallbackConfig(fallbackProperty: String, defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
```
chao2zhang, line 10: A wild idea: Could we use `ReadOnlyProperty<ConfigAware, T>` as the type of `fallbackProperty`? This can help us guarantee that there is no typo to be made as `String`.

marschwar: I played around with this idea, but I don't see how that would be possible. The concern of having a typo is very valid IMO so I added a check in the config collector.

cortinico:
> ```kotlin
>     private val functionThreshold: Int  by fallbackConfig(
>         fallbackProperty = "threshold", 
>         defaultValue = DEFAULT_FUNCTION_THRESHOLD
>     )
> ```

I came a bit late to the party but I have a couple of (maybe dumb) questions here:

1. Could we implement this as an property for `config` with default value (either `null` or `""`):

```
fun config(fallback: String = "", defaultValue: String): ReadOnlyProperty<ConfigAware, String> = simpleConfig(fallback, defaultValue)
```

2. If 1. is not an option, could we call the delegate function `configWithFallback`? I think it's going to be more discoverable (especially for autocompletion and similar) and it brings more context on what the delegate is actually doing.

marschwar:
I like the suggestion from @chao2zhang to give the each config delegate its own name so we don't just have the same function overloaded multiple times. 

When it comes to naming I am open for anything. I understand the reasoning and like your suggestion. What would you call the delegate that expects a List<String> to be configured? 
```
config            -> config
fallbackConfig    -> configWithFallback
listConfig        -> ??? configOfList ???
```

cortinico:
> What would you call the delegate that expects a List to be configured?

How about `configList`?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ConfigProperty.kt
```diff
@@ -4,21 +4,30 @@ import io.gitlab.arturbosch.detekt.api.ConfigAware
...
-fun config(defaultValue: Long): ReadOnlyProperty<ConfigAware, Long> = simpleConfig(defaultValue)
-fun config(defaultValue: Boolean): ReadOnlyProperty<ConfigAware, Boolean> = simpleConfig(defaultValue)
-fun config(defaultValue: List<String>): ReadOnlyProperty<ConfigAware, List<String>> = ListConfigProperty(defaultValue)
-
-private fun <T : Any> simpleConfig(defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
+fun <T : Any> config(defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
     SimpleConfigProperty(defaultValue)
 
+fun <T : Any> configWithFallback(fallbackProperty: String, defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
+    FallbackConfigProperty(fallbackProperty, defaultValue)
```
chao2zhang, line 11:
```suggestion
fun <T : Any> configWithFallback(fallbackPropertyName: String, defaultValue: T): ReadOnlyProperty<ConfigAware, T> =
    FallbackConfigProperty(fallbackPropertyName, defaultValue)
```

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/ConfigurationCollector.kt
```diff
@@ -107,14 +116,49 @@ class ConfigurationCollector {
...
+        if (arguments.size == 1) {
+            return checkNotNull(arguments[0].getArgumentExpression())
+        }
+        val defaultArgument = arguments
+            .find { it.getArgumentName()?.text == DEFAULT_VALUE_ARGUMENT_NAME }
+            ?: arguments.last()
+        return checkNotNull(defaultArgument.getArgumentExpression())
+    }
+
+    private object ConfigWithFallbackSupport {
```
chao2zhang, line 137: Today I learned: We can define a non-companion object inside a class.

chao2zhang (approved):

BraisGabin:
Why do we need `configList`? Isn't enought to use `config`? And, if we need `configWithFallback` don't we need `configListWithFallback` too? I'd vote for an unified delegate `config`.

marschwar:
I agree that the differenciation between `config` and `configList` is a bit awkward. Let me see if there is something that can be done.

marschwar:
If this is something we can agree on, I would clean up and add more tests.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ConfigProperty.kt
```diff
@@ -4,23 +4,35 @@ import io.gitlab.arturbosch.detekt.api.ConfigAware
...
+
+private fun <T : Any> getValueOrDefault(configAware: ConfigAware, propertyName: String, defaultValue: T): T {
+    @Suppress("UNCHECKED_CAST")
+    return when (defaultValue) {
+        is List<*> -> configAware.valueOrDefaultCommaSeparated(propertyName, defaultValue as List<String>) as T
+        is String,
+        is Boolean,
+        is Int,
+        is Long -> configAware.valueOrDefault(propertyName, defaultValue)
+        else -> error("${defaultValue.javaClass} is not supported as ")
```
BraisGabin, line 21: Maybe using the code from here: https://github.com/detekt/detekt/pull/3676#issuecomment-819605352 we could make this more type safe.

BraisGabin:
> If this is something we can agree on, I would clean up and add more tests.

I think that the API looks nice :)

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/ConfigProperty.kt
marschwar: The problem here is, that `fun <T : Any> config(defaultValue: T): ReadOnlyProperty<ConfigAware, T>` removes most of the type safety since a rule author could write `val prop: Regex by config(Regex("aaa"))` which would compile but fail at runtime. With the code from this PR it is more explicit and we can provide a specific error message if the provided type is not supported.

BraisGabin: If we want that type of safetiness (I agree with it). We could add a config file for each type. This way we ensure even more the type safety.

BraisGabin (approved):

BraisGabin:
This looks fine now. We can iterate later if we want.

marschwar:
Is there anything I can do to help getting this merged?

PR #3687, opened 2021-04-19
Deprecate configuration of reports in detekt Gradle extension

3flex:
Reports need to be configured on tasks directly, not by using the detekt extension.

This is to avoid setting the same destination path for multiple detekt reports. An example of this can be seen in the detektTest task of [this build scan](https://scans.gradle.com/s/2ktgllhdfoeje/timeline?details=4ef5ngfx5snfi) where the html report is written to the same path as the report generated by the detektMain task because both tasks get the destination path from the extension.

I will add a new section to the Gradle config doc and link to it from the deprecation message on the extension once the PR has been reviewed.

Fixes: #3926
Fixes: #4035
Fixes: #4044
Edit by @cortinico - Added fix issue reference

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
@@ -23,7 +23,7 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
         get() = reportsDir
 
     @Deprecated("Customise the reports on the Detekt task(s) instead.", level = DeprecationLevel.WARNING)
-    val reports = objects.newInstance(DetektReports::class.java)
+    val reports: DetektReports = objects.newInstance(DetektReports::class.java)
 
     @Deprecated(message = "Please use the source property instead.", replaceWith = ReplaceWith("source"))
     var input: ConfigurableFileCollection
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
@@ -18,6 +18,7 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
             isIgnoreFailures = value
         }
 
+    @Deprecated("Use reportsDir which is equivalent", ReplaceWith("reportsDir"))
     val customReportsDir: File?
         get() = reportsDir
 
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
@@ -21,6 +21,7 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
     val customReportsDir: File?
         get() = reportsDir
 
+    @Deprecated("Customise the reports on the Detekt task(s) instead.", level = DeprecationLevel.WARNING)
     val reports = DetektReports()
 
     @Deprecated(message = "Please use the source property instead.", replaceWith = ReplaceWith("source"))
@@ -78,6 +79,8 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
      */
     var ignoredFlavors: List<String> = emptyList()
 
+    @Suppress("DeprecatedCallableAddReplaceWith", "DEPRECATION")
+    @Deprecated("Customise the reports on the Detekt task(s) instead.", level = DeprecationLevel.WARNING)
     fun reports(configure: Action<DetektReports>) {
         configure.execute(reports)
     }
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
@@ -23,7 +23,7 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
         get() = reportsDir
 
     @Deprecated("Customise the reports on the Detekt task(s) instead.", level = DeprecationLevel.WARNING)
-    val reports = DetektReports()
+    val reports = objects.newInstance(DetektReports::class.java)
 
     @Deprecated(message = "Please use the source property instead.", replaceWith = ReplaceWith("source"))
     var input: ConfigurableFileCollection
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektMultiplatform.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektMultiplatform.kt
@@ -116,7 +116,6 @@ internal class DetektMultiplatform(private val project: Project) {
             }?.let { baselineFile ->
                 baseline.set(layout.file(provider { baselineFile }))
             }
-            reports = extension.reports
             reports.xml.setDefaultIfUnset(File(extension.reportsDir, compilation.name + ".xml"))
             reports.html.setDefaultIfUnset(File(extension.reportsDir, compilation.name + ".html"))
             reports.txt.setDefaultIfUnset(File(extension.reportsDir, compilation.name + ".txt"))
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektMultiplatform.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektMultiplatform.kt
@@ -91,6 +91,7 @@ internal class DetektMultiplatform(private val project: Project) {
         }
     }
 
+    @Suppress("LongMethod")
     private fun Project.registerMultiplatformTasksForNonAndroidTarget(
         compilation: KotlinCompilation<KotlinCommonOptions>,
         target: KotlinTarget,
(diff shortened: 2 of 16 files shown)


docs/pages/gettingstarted/gradle.md
```diff
@@ -238,34 +238,6 @@ detekt {
...
-        sarif {
-            enabled = true                                
-            destination = file("build/reports/detekt.sarif")
-        }
-        custom {
-            // The simple class name of your custom report.
-            reportId = "CustomJsonReport"                   
-            destination = file("build/reports/detekt.json")
-        }
-    }
```
BraisGabin, line 268: How do I enable/disable some reports now? And should we talk about `reportsDir` here? Or is it a default in gradle so we don't need to?

3flex: All report config is done on the task, so you would enable/disable reports either on the individual task level or using `withType`. That's the doc section I need to write.

`reportsDir` is used in Gradle for other code quality plugins so we'd now follow that convention. I can mention that in the doc and remove references to the deprecated option (I didn't search to see if there are any yet)

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektMultiplatform.kt
```diff
@@ -55,11 +56,34 @@ internal class DetektMultiplatform(private val project: Project) {
...
+                                }
+                            )
+                        )
+                        reports.sarif.outputLocation.convention(
+                            layout.projectDirectory.file(
+                                providers.provider {
+                                    File(extension.reportsDir, compilation.name + ".sarif").absolutePath
+                                }
+                            )
+                        )
```
BraisGabin, line 86: would it be possible to extract this to a function/extension function so we should need to pass the name and it will configure all the reports? We have a lot of duplicated code between Multiplatform, JVM and Android (not only this)

BraisGabin: And, should we change Plain too?

3flex: Plain config is OK because it defers to the default report path set in the task if the report destination (or now, the outputLocation) is not set:

https://github.com/detekt/detekt/blob/da8676d396cd29534c1be643cb2a505d4470fd11/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt#L270-L273

There's a fair bit of refactoring that can be done in the plugin, lots of which is easier once we're on Gradle 6.1. I'd prefer to wait until then before fixing up the repetition.

BraisGabin (approved):

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -79,6 +81,8 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
      */
     var ignoredFlavors: List<String> = emptyList()
 
+    @Suppress("DeprecatedCallableAddReplaceWith", "DEPRECATION")
+    @Deprecated("Customise the reports on the Detekt task(s) instead.", level = DeprecationLevel.WARNING)
     fun reports(configure: Action<DetektReports>) = configure.execute(reports)
```
chao2zhang, line 86: In the DetektExtension, will it make sense to keep the enabling flags for each report type? My assumptions are:
- Fewer users are configuring the report output location
- More users are configuring whether to keep txt/xml/html/sarif.

3flex: I understand your point, but in Gradle reports are configured on tasks, not on the extension. That's what's done for CodeNarc, Checkstyle and PMD extensions, and that intention is made explicit in the docs:

> Things that produce reports (typically tasks) expose a report container that contains Report objects for each possible report that they can produce.

[reference](https://docs.gradle.org/current/javadoc/org/gradle/api/reporting/ReportContainer.html)

We can't implement ReportContainer directly but in keeping with that convention (and slightly simplifying the plugin) I still think it should be removed.

3flex:
Any other concerns on this before merging? I'd only suggest waiting until after 1.17 release now some RCs have been published.

BraisGabin:
Agree, we should wait until 1.17.0 is released.

3flex:
LGTM?

BraisGabin (approved):

cortinico (commented):
Can we please extend a bit on the documentation for this PR?

Specifically we're deprecating a property that is suggested to use in our README and in the homepage of the website: https://detekt.github.io/detekt/

Potentially also a migration page should be added as well.

cortinico:
Should we push this to `1.19.0`?

3flex:
Yeah sorry I haven't had time to dedicate to this recently. Happy to push it back.

I'll need to rebase and still write some docs.

3flex:
Rebased & docs updated. I believe I've updated all sections where the deprecated configuration was used, except the old "News" posts.

Migration instructions were added to the changelog.

BraisGabin (approved):
I'd vote to split this PR in two. One with code and chancellor and the other with the doc. Otherwise the users will see a "too updated" documentation untilnwe release 1.19

3flex:
Time for versioned docs? :)

I've split them up - doc updates in #4064. I've left the changelog & migration update in this PR because the changelog is versioned so it's clear that it's related to unreleased changes.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -78,6 +80,8 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
      */
     var ignoredFlavors: List<String> = emptyList()
 
+    @Suppress("DeprecatedCallableAddReplaceWith", "DEPRECATION")
+    @Deprecated("Customise the reports on the Detekt task(s) instead.", level = DeprecationLevel.WARNING)
```
cortinico, line 84: Commenting here but this is valid for the whole PR.

I believe we could take a less disruptive apporach here. Specifically I would not deprecate `reports{}` but just the destination paths.

1. Top level `reports{}` will still remain as a way to toggle reports types (like enable/disable xml reports for all the tasks).
2. We should raise a warning if a destination is set inside `reports{}` as this is basically causing reports to be overridden.
3. By default reports should use the task name as default name (like `detektMain.xml`).
4. We should invite users to customize reports destination at the task level (I also think this is sort of an edge case. Majority of users are fine with the default file location).

3flex: > 1. Top level reports{} will still remain as a way to toggle reports types (like enable/disable xml reports for all the tasks).

I disagree, for two reasons: this isn't how any of the Gradle code quality plugins work, nor Spotbugs & ktlint. It's also more work to maintain. For example, deprecating the output path on the extension, but not the tasks, means duplicating the `DetektReports` class so there's one with the deprecated output path that's used in the extension and another without the reprecated output path deprecated that's used in tasks.

Also all reports are enabled by default, so unless someone is particularly compelled to disable reports of a certain type this wouldn't be used much anyway, and there's a documented way to disable all reports globally if desired using the same method as the other Gradle plugins I mentioned.

Also I've just realised that in this PR I have not linked custom tasks declared on the extension to the tasks created by the plugin... but again, another thing to link up (I will add this), and missed both by me and reviewers because the report code is currently complex. Much of the complexity can go away once `reports` is removed from the extension.

> 2. We should raise a warning if a destination is set inside reports{} as this is basically causing reports to be overridden.

Yes and no... if destination is set in `reports {}` in the extension, then that's generally the default for all tasks. However with changes in this PR, the tasks registered by this plugin will set an appropriate output path and ignore the value set in the extension. So the report output will not be overridden for any task created by the plugin, so the warning isn't necessary.

This is only an issue if someone creates multiple detekt tasks manually without specifying a report output path for active reports. I can look at adding a warning for this case.

> 3. By default reports should use the task name as default name (like detektMain.xml).

They use the variant, sourceset or compilation name depending on whether it's Android, JVM or MPP respectively, which is the same convention as used to name the tasks (just without the "detekt" prefix) so the report file names will be unique and follow the usual Gradle conventions for report naming. The exception is the Plain task which has output file name of `detekt.xml` for XML output for example.

> 4. We should invite users to customize reports destination at the task level (I also think this is sort of an edge case. Majority of users are fine with the default file location).

Agreed and this is what this PR aims to achieve, along with the migration guide and updated docs.

cortinico: You convinced me @3flex. Thanks for claryfing everything 🙏 

Some minor comments on what you wrote:

> so unless someone is particularly compelled to disable reports of a certain type

I honestly don't know how much this is going to be a problem. I believe we're fine, but if someone wanted to disable say the `.sarif` report, now we're making it a bit harder for them to do. Not a major problem afterall.

> the tasks registered by this plugin will set an appropriate output path and ignore the value set in the extension

I still believe we should warn in some form, as we're effectively ignoring a value the user provided. So if you have a `report { html { destination = "xyz" } }` where `"xyz"` is not the default URL for that specific report type, we should `logger.warn` or even break the build, just to inform the user that we're ignoring what they specified. In other words, we're both deprecating and introducing a behavioral change. I believe that just the IDE warning on `report{}` is not sufficient.

> The exception is the Plain task which has output file name of detekt.xml for XML output for example.

How about they all use the task name instead? This will make the report name more predictable.

> Agreed and this is what this PR aims to achieve, along with the migration guide and updated docs.

🚀 Great work with driving this very much needed PR

3flex: I've added a warning that will show in the build if an output location/destination is set on the `detekt {}` extension and the Gradle plugin also sets up a task. This is simpler than checking expected vs actual path of the output location and I think the warning is descriptive.

> How about they all use the task name instead? This will make the report name more predictable.

I prefer the way it is now - and I didn't actually change the existing naming convention in this PR. Maybe that can be opened on another issue for discussion?

This should be ready to go otherwise. Thanks for the feedback!

cortinico (approved):

BraisGabin:
Is there something missing here? Can we merge this? This is a big change fot 1.19.0 so I think that as soon as we merge it as better. Someone using the SNAPSHOT can find some issue or we can find some issue while developing the tool itself.

cortinico:
☝️ Agree. I believe we're good to go for merging this

PR #3692, opened 2021-04-22
Add extra rules to gitattributes

### LINKED ISSUE #3659: Specify somewhere that core.autocrlf = true is required on Windows for working on detekt itself
## Expected Behavior

Cloning and running `gradle build` works on Windows

## Observed Behavior

[This test](https://github.com/detekt/detekt/blob/9f894f56b9d8d91cfae994c626ead0b4e0628a8a/detekt-parser/src/test/kotlin/io/github/detekt/parser/KtCompilerTest.kt#L22-L30) fails due to the assertion line 25 which expects the system's line ending, i.e. `\r\n` on Windows. However, in order to have `\r\n` line endings on clone, you need to set the `core.autocrlf` Git option to `true`, otherwise the test fails because the file uses `\n` line endings.

## Steps to Reproduce

* Set `core.autocrlf` to `false` or `input` on Windows
* Clone the detekt repository
* Run `gradle build`

## Context

This makes the onboarding developers (e.g. me) confusing, as a seemingly random test fails for what appears to be no reason. This degrades the developer experience.

A fix could be to tell Windows user to use the following command in the README and CONTRIBUTING files:

```
git config --global core.autocrlf true
```

## Your Environment

* Version of detekt used: Main branch
* Version of Gradle used (if applicable): Wrapper version from detekt (6.8.3)

3flex:
Fixes #3659

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,9 +3,9 @@
 *.bat           text eol=crlf
 
 *.css           diff=css
-*.htm           diff=html
-*.html          diff=html
-*.md            diff=markdown
+*.htm           eol=lf diff=html
+*.html          eol=lf diff=html
+*.md            eol=lf diff=markdown
 
 *.jar           binary
 *.png           binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,9 +1,17 @@
-# Always use crlf line endings Gradle .bat file
-gradlew.bat text eol=crlf
+* text eol=auto
 
-# Mark Gradle wrapper as a binary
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,9 +1,17 @@
-# Always use crlf line endings Gradle .bat file
-gradlew.bat text eol=crlf
+* text eol=auto
```
3flex, line 1: This effectively sets `core.autocrlf` in the repo by default.

schalkms (approved):

BraisGabin (approved):

PR #3696, opened 2021-04-23
 MultilineLambdaItParameter: don't report when lambda has no implicit parameter references

### LINKED ISSUE #3570: MultilineLambdaItParameter should not be triggered when parameter is not used at all
When lambda body does not refer to implicit `it` parameter at all, the rule should not be triggered.

## Context
Spring transaction template provides an `execute` extension method where lambda parameter is `TransactionStatus`. In most cases the parameter is not used at all, e.g.:

```kotlin
tx.execute {
    val entity = entityManager.find(Entity::class.java, 1)
    // more code within transaction
    // lambda parameter is never used but the rule is triggered
}
```

t-kameyama:
Fixes #3570


BraisGabin (approved):

chao2zhang (approved):

chao2zhang:
Code coverage is reporting test coverage miss on https://app.codecov.io/gh/detekt/detekt/compare/3696/diff#D1L118. What scenario will it take to fail the condition `getResolvedCall(bindingContext)?.resultingDescriptor == implicitParameter`, is it by shadowing the `it` name with another local variable?

t-kameyama:
Yes, I fixed the test. https://github.com/detekt/detekt/pull/3696/commits/01104e17b7199976acee51f5c44adc7c185cf007

chao2zhang (approved):

PR #3698, opened 2021-04-23
Build & test on Java 16

3flex:


3flex:
Blocked until next Gradle Shadow plugin release.

cortinico (approved):

cortinico:
@3flex where you able to understand what are the build failures? It seems we have failing tests plus `generateDocumentation` is failing

3flex:
I believe it's now blocked thanks to an out of date transitive dependency of mockk and also Android Gradle Plugin incompatibility with Java 16.

Mockk will be fixed soon enough I expect but I can manually add the required dependency version to fix it in the meantime.

I can't find compatibility info for AGP. They probably only officially support the JDK version shipped with Android Studio (currently 11?) So Android tests should probably be skipped on other JDK versions.

I have to check the generateDocumentation task.

3flex:
All issues resolved, but still blocked until we update to Kotlin 1.5.0

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektAndroidTest.kt
```diff
@@ -11,7 +11,9 @@ import org.spekframework.spek2.style.specification.describe
 object DetektAndroidTest : Spek({
     describe(
         "When applying detekt in an Android project",
-        skip = if (isAndroidSdkInstalled()) Skip.No else Skip.Yes("No android sdk.")
+        skip = if (!isAndroidSdkInstalled()) Skip.Yes("No android sdk.")
+        else if (getJdkVersion() >= 16) Skip.Yes("Android 4.1.2 & 4.2.0 don't run on JDK 16 or higher")
+        else Skip.No
```
cortinico, line 16: Should we extract this to a function (especially as it's used across two test now)?

gradle/libs.versions.toml
```diff
@@ -44,6 +44,7 @@ sarif4k = "io.github.detekt.sarif4k:sarif4k:0.0.1"
 assertj = "org.assertj:assertj-core:3.19.0"
 reflections = "org.reflections:reflections:0.9.12"
 mockk = "io.mockk:mockk:1.11.0"
+byteBuddy = "net.bytebuddy:byte-buddy:1.11.0" # remove this explicity dependency when mockk is updated next
```
cortinico, line 47: Why do we need this?

gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 dokka = "1.4.32"
 jacoco = "0.8.7"
-kotlin = "1.4.32"
+kotlin = "1.5.0"
```
cortinico, line 4: Can we remove this bump as we have a separate PR for this?

cortinico (approved):

PR #3703, opened 2021-04-26
(Try to) improve CI build reliability

3flex:
These changes should all help, or at least not make things any worse, by:
* reducing work done (by dropping one of the multi-Gradle version tests)
* ~~reducing memory use (by not allocation 2GB heap for each Gradle integration test - 512MB is the default, and unless there are OOM errors there shouldn't be any need to increase this)~~
* reducing CPU contention during heavy load during the build (by reducing max worker count from 4 to # of CPUs, which is 2 on Windows & Linux CI runners and 3 on macOS)

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -23,16 +23,6 @@ jobs:
     steps:
     - name: Checkout Repo
       uses: actions/checkout@v2
-
-      # Let's cleanup the gradle cache folders to make sure
-      # we don't accidentally cache stale files.
-    - name: Cleanup Gradle Folders
-      shell: bash
-      run: |
-        rm -rf ~/.gradle/caches/ && \
-        rm -rf ~/.gradle/wrapper/
-
-
     - name: Cache Gradle Folders
       uses: actions/cache@v2
       with:
@@ -44,22 +34,17 @@ jobs:
           cache-gradle-${{ matrix.os }}-${{ matrix.jdk }}-
           cache-gradle-${{ matrix.os }}-
           cache-gradle-
-
-
     - name: Setup Java
       uses: actions/setup-java@v2
       with:
         java-version: ${{ matrix.jdk }}
         distribution: 'adopt'
-
-
     - name: Build detekt
       run: ./gradlew build moveJarForIntegrationTest -x detekt
     - name: Run detekt-cli --help
       run: java -jar ./build/detekt-cli-all.jar --help
     - name: Run detekt-cli with argsfile
       run: java -jar ./build/detekt-cli-all.jar "@./config/detekt/argsfile"
-
 
   verify-generated-config-file:
     if: ${{ !contains(github.event.head_commit.message, 'ci skip') }}
@@ -88,7 +73,6 @@ jobs:
         distribution: 'adopt'
     - name: Verify Generated Detekt Config File
       run: ./gradlew verifyGeneratorOutput
-
 
   compile-test-snippets:
     if: ${{ !contains(github.event.head_commit.message, 'ci skip') }}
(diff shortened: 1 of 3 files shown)


detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
```diff
@@ -117,7 +117,6 @@ class DslGradleRunner @Suppress("LongParameterList") constructor(
             add("--stacktrace")
             add("--info")
             add("--build-cache")
-            add("-Dorg.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m")
```
chao2zhang, line 120: This line was added to increase the heapsize and metaspace size, which should reduce flakiness for me locally and according to my enterprise development experience.

Is this against your past experience?

3flex: The Gradle daemon shouldn't need that much allocated for the simple integration test builds. If it does, there's likely another problem that increasing the memory is just masking. Perhaps older Kotlin versions were causing an issue?

If the build is not stable after this change because of out of memory errors due to heap or metaspace then let's revisit. Or do you have logs that point to this being an issue?

.github/workflows/pre-merge.yaml
```diff
@@ -23,16 +23,6 @@ jobs:
...
     - name: Checkout Repo
       uses: actions/checkout@v2
-
-      # Let's cleanup the gradle cache folders to make sure
-      # we don't accidentally cache stale files.
-    - name: Cleanup Gradle Folders
-      shell: bash
-      run: |
-        rm -rf ~/.gradle/caches/ && \
-        rm -rf ~/.gradle/wrapper/
```
cortinico, line 33: Those lines were originally added as the CI was flaky on MacOS due to Gradle caches.
I believe they could be removed right now, or at least we could try, and re-add them if we keep on saying weirdness happening on the MacOS cache.

3flex: I referenced that comment/issue in the commit message - as you said the workaround isn't needed anymore.

https://github.com/detekt/detekt/pull/2547#issuecomment-605611809

https://github.com/actions/virtual-environments/issues/427#issuecomment-623858143

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
chao2zhang: We could test flakiness after merging this PR. For me it was highly reproducible, but I do not have any traces back to my original experiences.

.github/workflows/pre-merge.yaml
cortinico: Awesome, let's remove it then 👍

cortinico (approved):

BraisGabin (approved):

3flex:
I think we should hold off merging because it doesn't actually make the build more reliable, since it's still failing on the Gradle plugin merge reports test.

https://github.com/detekt/detekt/issues/3705#issuecomment-828047662

Or I can skip the tests on Java 11 for now... At least then the build will hopefully stay green, but it's just hiding the problem. I can't reproduce locally though so I don't know how to resolve.

cortinico:
> Or I can skip the tests on Java 11 for now... At least then the build will hopefully stay green, but it's just hiding the problem. I can't reproduce locally though so I don't know how to resolve.

Can we split it in smaller chunks? E.g. the GH Actions Cache fix could go alone imho

3flex:
There doesn't seem to be a direct link between the memory usage and the specific test failure that kept coming up with this change, but it seems to consistently fail otherwise so I've rolled that back. The other changes will still help reduce flakiness.

PR #3710, opened 2021-04-28
Publish build scans for all CI builds

3flex:


settings.gradle.kts
```diff
@@ -33,13 +33,18 @@ include(
...
+    id("com.gradle.enterprise") version "3.6.1"
 }
 
 gradleEnterprise {
+    val isCiBuild = System.getenv("CI") != null
+
     buildScan {
         termsOfServiceUrl = "https://gradle.com/terms-of-service"
-        termsOfServiceAgree = "yes"
+        if (isCiBuild) {
```
chao2zhang, line 44: Nice! I had wanted to do an if check here since we should not sign an agreement for others building our project.

chao2zhang (approved):

PR #3718, opened 2021-04-29
Update Kotlin to 1.5

eygraber:
Looks like if `kotlinOptions.apiVersion` is not set to `"1.5"` the IDE won't resolve any of the new APIs. I'm assuming that's only needed until the IDE plugin is released.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -9,6 +9,9 @@ plugins {
     detekt
     id("com.github.ben-manes.versions")
     id("org.sonarqube")
+
+    // remove once Kotlin 1.5.10 is released (https://youtrack.jetbrains.com/issue/KT-46368)
+    id("dev.zacsweers.kgp-150-leak-patcher") version "1.1.0"
 }
 
 allprojects {
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 dokka = "1.4.32"
 jacoco = "0.8.7"
-kotlin = "1.4.32"
+kotlin = "1.5.0"
 ktlint = "0.41.0"
 spek = "2.0.15"
 
@@ -25,7 +25,7 @@ kotlin-scriptingCompilerEmbeddable = { module = "org.jetbrains.kotlin:kotlin-scr
 kotlin-stdlibJdk8 = { module = "org.jetbrains.kotlin:kotlin-stdlib-jdk8", version.ref = "kotlin" }
 
 kotlinx-html = "org.jetbrains.kotlinx:kotlinx-html-jvm:0.7.3"
-kotlinx-coroutines = "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.1"
+kotlinx-coroutines = "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.0"
 
 android-gradlePlugin = "com.android.tools.build:gradle:4.2.0"
 
--- a/buildSrc/src/main/kotlin/module.gradle.kts
+++ b/buildSrc/src/main/kotlin/module.gradle.kts
@@ -55,7 +55,12 @@ tasks.withType<KotlinCompile>().configureEach {
             "-Xopt-in=kotlin.RequiresOptIn"
         )
         // Usage: <code>./gradlew build -PwarningsAsErrors=true</code>.
-        allWarningsAsErrors = project.findProperty("warningsAsErrors") == "true" || System.getenv("CI") == "true"
+        // Note: currently there are warnings for detekt-gradle-plugin that seemingly can't be fixed
+        //       until Gradle releases an update (https://github.com/gradle/gradle/issues/16345)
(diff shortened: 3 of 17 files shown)


eygraber:
[Blocked by Jacoco 0.8.7](https://github.com/jacoco/jacoco/issues/1155)

chao2zhang:
Thanks! Although Kotlin 1.5 is not yet publicly announced, I would suggest holding this PR for a little longer until we have Detekt 1.17 out. It is very likely that Kotlin 1.5 is going to introduce new language and tooling issues.

3flex:
Thumbs up if we should hold back on upgrade because of this (memory leak in Kotlin daemon which will be fixed in 1.5.10): https://youtrack.jetbrains.com/issue/KT-46368

Thumbs down if we should ignore this issue.

It won't affect CI but may be frustrating for contributors.

3flex:
There's a memory leak in Kotlin's Gradle Plugin in 1.5.0, but there's a plugin available that patches this issue until a proper fix is released in 1.5.10.

I suggest this plugin be applied with this PR (or once this is merged): https://github.com/ZacSweers/kgp-150-leak-patcher

The problem won't affect CI but contributors might (or definitely will?) end up with poor daemon/build performance over time.

eygraber:
I can apply that plugin here. 

Another option I usually use when there's a leak like this is to get into the habit of killing the daemon occasionally.

eygraber:
For the `:detekt-psi-utils:apiCheck` failure, do I need to run `:detekt-psi-utils:apiDump` or is there another way to address that?

chao2zhang:
Yes, the intention is to run `:detekt-psi-utils:apiDump` so that we can easily tell which methods are added/changed/removed through PR.

BraisGabin:
It seems that you need to run `:detekt-api:apiCheck` too

eygraber:
@BraisGabin check out the last commit that I just made. It seems like the only way to fix [this Gradle/Kotlin issue](https://youtrack.jetbrains.com/issue/KT-41142) where Gradle 7.0 is using Kotlin 1.4.31, and that causes a warning because detekt is using Kotlin 1.5.0.

There might be a better way to address that, but that's what I have for now :disappointed:

BraisGabin:
Could you add a comment pointing the issue near this workaround? This way we will know when it's safe to remove it.

eygraber:
Yup. I found an issue filed on Gradle that might be better to point to https://github.com/gradle/gradle/issues/16345

eygraber:
Although, now I'm thinking it might be better to keep the gradle plugin on Kotlin 1.4.31 until Gradle supports 1.5 but that seemed to break some things when I tested it before. I'll see if I can get that working.

eygraber:
I went back to disabling `allWarningsAsErrors`, and added a note explaining why.

BraisGabin (approved):

build.gradle.kts
```diff
@@ -9,6 +9,9 @@ plugins {
     detekt
     id("com.github.ben-manes.versions")
     id("org.sonarqube")
+
+    // remove once Kotlin 1.5.10 is released (https://youtrack.jetbrains.com/issue/KT-46368)
+    id("dev.zacsweers.kgp-150-leak-patcher") version "1.0.1"
```
BraisGabin, line 14: The version 1.1.0 was released. (not blockig we can update it in another PR)

eygraber:
@BraisGabin should I write tests to cover the changes that codecov is complaining about?

BraisGabin:
> @BraisGabin should I write tests to cover the changes that codecov is complaining about?

There's not need

build.gradle.kts
eygraber: Updated that in the latest push

gradle/libs.versions.toml
```diff
@@ -25,7 +25,7 @@ kotlin-scriptingCompilerEmbeddable = { module = "org.jetbrains.kotlin:kotlin-scr
 kotlin-stdlibJdk8 = { module = "org.jetbrains.kotlin:kotlin-stdlib-jdk8", version.ref = "kotlin" }
 
 kotlinx-html = "org.jetbrains.kotlinx:kotlinx-html-jvm:0.7.3"
-kotlinx-coroutines = "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.1"
+kotlinx-coroutines = "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.0-RC"
```
cortinico, line 28:
```suggestion
kotlinx-coroutines = "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.0"
```

Was released some days ago 👍

cortinico (approved):
Looks good on my end 👍 Thanks for taking the time to do this.

I'm ok with having the Gradle plugin on 1.5.x.
Ideally by the time we release 1.18.0, Kotlin 1.5.10 will also be out

gradle/libs.versions.toml
eygraber: Done

chao2zhang (approved):

PR #3719, opened 2021-04-29
Fix modifier order value class

### LINKED ISSUE #3717: False positive on ModifierOrder for value classes in Kotlin 1.5
```
@JvmInline
private value class Foo(val bar: Int)
```

## Expected Behavior
Should not fail the ModifierOrder check

## Observed Behavior
It does

## Your Environment
* Version of detekt used: 1.16.0

eygraber:
On top of https://github.com/detekt/detekt/pull/3718

Fixes #3717

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ModifierOrder.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ModifierOrder.kt
@@ -34,6 +34,7 @@ import org.jetbrains.kotlin.lexer.KtTokens.PUBLIC_KEYWORD
 import org.jetbrains.kotlin.lexer.KtTokens.SEALED_KEYWORD
 import org.jetbrains.kotlin.lexer.KtTokens.SUSPEND_KEYWORD
 import org.jetbrains.kotlin.lexer.KtTokens.TAILREC_KEYWORD
+import org.jetbrains.kotlin.lexer.KtTokens.VALUE_KEYWORD
 import org.jetbrains.kotlin.lexer.KtTokens.VARARG_KEYWORD
 import org.jetbrains.kotlin.psi.KtModifierList
 import org.jetbrains.kotlin.psi.psiUtil.allChildren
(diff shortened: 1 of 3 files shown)


cortinico (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ModifierOrder.kt
```diff
@@ -75,6 +76,7 @@ class ModifierOrder(config: Config = Config.empty) : Rule(config) {
         ENUM_KEYWORD, ANNOTATION_KEYWORD, FUN_KEYWORD,
         COMPANION_KEYWORD,
         INLINE_KEYWORD,
+        VALUE_KEYWORD,
```
chao2zhang, line 79: It seems reasonable to put `value` here because `value class` is `inline class`. However, would you mind creating a PR for Kotlin coding convention [here](https://kotlinlang.org/docs/coding-conventions.html#modifiers-order) since it does not include the `value` modifier yet?

chao2zhang (approved):

BraisGabin (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ModifierOrder.kt
BraisGabin: I just created that PR: https://github.com/JetBrains/kotlin-web-site/pull/2300

Should we wait for it to be merged before merging this? Or can we merge it and revert it in case it is rejected?

chao2zhang: I am inclined to merge it now since this will impact the correctness when running Detekt on Kotlin 1.5 code. See https://github.com/detekt/detekt/pull/3810/files#r637354618

netroy:
@cortinico @chao2zhang can we **please** have a new release that'd include this fix?

PR #3723, opened 2021-04-30
GitHub Actions cache fixes

3flex:
Summary of changes:
* Runs Gradle tasks with gradle-build-action@v2. This handles running Gradle tasks and caching various outputs. v2 is in beta, but once it's stable it will be supported by Gradle and recommended as the canonical way to run Gradle tasks on GitHub Actions.
* Enable Gradle's configuration cache for jobs that support it, which will reduce the job's execution time

Blocked until [gradle-build-action](https://github.com/gradle/gradle-build-action) v2 goes into production release as the configuration syntax could potentially change.

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -22,33 +22,28 @@ jobs:
     steps:
     - name: Checkout Repo
       uses: actions/checkout@v2
-    - name: Cache Gradle Folders
-      uses: actions/cache@v2
-      with:
-        path: |
-          ~/.gradle/caches/
-          ~/.gradle/wrapper/
-        key: cache-gradle-${{ matrix.os }}-${{ matrix.jdk }}-${{ hashFiles('gradle/libs.versions.toml') }}
-        restore-keys: |
-          cache-gradle-${{ matrix.os }}-${{ matrix.jdk }}-
-          cache-gradle-${{ matrix.os }}-
-          cache-gradle-
     - name: Setup Java
       uses: actions/setup-java@v2
       with:
         java-version: ${{ matrix.jdk }}
         distribution: 'temurin'
     - name: Build detekt
-      run: ./gradlew build -x detekt
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: build -x detekt
     - uses: actions/upload-artifact@v2
       with:
         name: heap-dump
         path: '**.hprof'
         if-no-files-found: ignore
     - name: Run detekt-cli --help
-      run: ./gradlew :detekt-cli:runWithHelpFlag
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: :detekt-cli:runWithHelpFlag
     - name: Run detekt-cli with argsfile
-      run: ./gradlew :detekt-cli:runWithArgsFile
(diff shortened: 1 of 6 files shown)


.github/workflows/pre-merge.yaml
```diff
@@ -19,101 +19,37 @@ jobs:
...
-      uses: actions/setup-java@v2
-      with:
-        java-version: 11
-        distribution: 'adopt'
     - name: Verify Generated Detekt Config File
+      if: ${{ matrix.jdk == '11' && matrix.os == 'ubuntu-latest' }}
       run: ./gradlew verifyGeneratorOutput --stacktrace
-
-
-  compile-test-snippets:
```
cortinico, line 93: Why you removed those two jobs?

3flex: They're not removed, they're now just steps in the `gradle` job. This way they don't create their own copies of `~/gradle/caches`, and there's just a single cache that can be used by all steps in the job. Total cache size per repository is 5GB so removing this duplication is useful.

But I'm going to move the compile-test-snippets step back into its own job so it can run in parallel again, because in a full rebuild running sequentially will take too long.

cortinico: > But I'm going to move the compile-test-snippets step back into its own job so it can run in parallel again,

Yeah that was exactly my point. I'd rather prioritize them running in parallel rather than reducing the cache size. We can also let those steps reuse the same cache key as the others

3flex: I'm still working on this, it's just a draft PR. I'll comment when it's ready for review.

3flex:
I'll come back to this later.

BraisGabin:
Sorry! I just removed the branch
And the mobile app doesn't have undo.

BraisGabin (approved):
Nice one!

chao2zhang (approved):
It seems that we are selectively enabling configuration cache for some flow. Do we want to do it the reverse way? (`org.gradle.unsafe.configuration-cache=true` in gradle.properties and use `--no-configuration-cache` when it does not work or has high risks.

3flex:
> org.gradle.unsafe.configuration-cache=true in gradle.properties

If we did this it would apply to all builds, including local builds. I think this will be annoying for contributors if that flag has to be specified for any builds that configure problematic tasks to avoid configuration cache errors from being displayed.

cortinico:
> Still waiting on the final release of Gradle build action V2.

Till we wait for that, could we bump `actions/setup-java` only?

3flex:
🎉 

https://github.com/gradle/gradle-build-action/releases/tag/v2.0.0

3flex:
An issue I've experienced with this implementation: https://github.com/gradle/gradle-build-action/issues/120

Hopefully this is caught by the action in future. I don't think it's worth rolling back; it's not typical to have several jobs running concurrently. We might also try only caching on main branch: https://github.com/gradle/gradle-build-action/blob/v2.0.0/README.md#only-write-to-the-cache-from-the-default-branch

This would avoid many uploads occurring at once thanks to multiple PRs being opened quickly one after the other, but it would still occur when multiple PRs are merged quickly one after the other on main branch.

PR #3724, opened 2021-04-30
Remove --stacktrace now that we have scan

BraisGabin:
This PR reverts #3604 now that we have #3710

`--stacktrace` is quite noisy and it makes difficult to find which test failed.


schalkms (approved):

chao2zhang (commented):
Ah I didn't publish this review

PR #3726, opened 2021-04-30
We have a logo :)

BraisGabin:
I added the svg to the project inside the folder media (should we upload them in other place instead?)

I added changed the favicon (thanks @JavierSegoviaCordoba). I used the purple one. as I can read here svg is supported in the last version of all major browsers so I think it's better to not use `.ico`.

I added the logo in the index of our home page. I used the purple with the name.

I didn't add the logo in the readme because the version with "detekt" on ti doesn't work well on dark mode.

I didn't updated the social media image (thanks @G00fY2) nor the organization image because I'm not 100% sure if we all agree about the purple one. To be honest I like them both. The purple reminds me more to kotlin but the orange feels more like "warning".

And, once we decide if we prefer purple or organe should I remove the other color from `/media`?

And, of course! Thanks again @payathedo for the logos! :)


BraisGabin:
And, of course! Thanks again @payathedo for the logos! :)

schalkms (approved):
I think we should unify the appearance of detekt and use the same logo wherever possible.

docs/index.md
```diff
@@ -6,6 +6,7 @@ permalink: index.html
 summary:
 ---
 
+![detekt logo](images/logo.svg "detekt logo")
```
chao2zhang, line 9: When I switched to view rich diff, the detekt logo appears as not loaded - Is this just because we have not merged in this PR yet?

chao2zhang (approved):
We should highlight this in our changelog, right?

docs/index.md
chao2zhang: Actually, we could add @payathedo to our contributors list

cortinico:
It's ok to store them in the repo as they're SVG 👌 

> I didn't add the logo in the readme because the version with "detekt" on ti doesn't work well on dark mode.

Maybe we can create a logo with some background to make it render good in both themes?

> I didn't updated the social media image (thanks @G00fY2) nor the organization image because I'm not 100% sure if we all agree about the purple one. To be honest I like them both. The purple reminds me more to kotlin but the orange feels more like "warning".
> And, once we decide if we prefer purple or organe should I remove the other color from `/media`?

+1 for Purple

> And, of course! Thanks again @payathedo for the logos! :)

<3

> We should highlight this in our changelog, right?

Yup, I believe this can be on the spotlight for 1.17.0 🎉

G00fY2:
> > I didn't add the logo in the readme because the version with "detekt" on ti doesn't work well on dark mode.
> 
> Maybe we can create a logo with some background to make it render good in both themes?
I think we would need a background layer (thats how most of the repo logos are designed) or we could choose a color that works on both themes ([coil](https://github.com/coil-kt/coil) is a good example).

Maybe in the future it is also possible to dynamically react to the current theme: [Support theme context for images in light vs dark mode](https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981/53)

docs/index.md
cortinico: > Actually, we could add @payathedo to our contributors list

Indeed 🚀

chao2zhang (approved):

docs/index.md
BraisGabin: > When I switched to view rich diff, the detekt logo appears as not loaded - Is this just because we have not merged in this PR yet?

This link is to be used by jekyll so it should work once it is "compiled"

BraisGabin:
> I think we should unify the appearance of detekt and use the same logo wherever possible.

I removed the orange ones 👍

PR #3727, opened 2021-05-01
Prepare Detekt 1.17.0-RC3

cortinico:
Baking a new Release Candidate 🎉 
Please try and let us know if everything works fine

PR #3575, opened 2021-03-19
Gradle tweaks

3flex:
Just rearranging a few things to make the build config a little more idiomatic, as well as fix a deprecation warning and fix a couple of issues in the README examples.

PR #3732, opened 2021-05-02
Specify Java language level in module plugin

### LINKED ISSUE #3731: Gradle plugin no longer supporting Java 1.8 with 1.17.0-RC1, by accident?
## Expected Behavior

Maybe it is a deliberate change to drop Java 1.8 support, but nothing was mentioned in the release notes about it.

## Observed Behavior

Trying to resolve Gradle plugin 1.17.0-RC1 with JDK 1.8 yields:

```
   > Could not resolve io.gitlab.arturbosch.detekt:detekt-gradle-plugin:1.17.0-RC1.
     Required by:
         project :
      > No matching variant of io.gitlab.arturbosch.detekt:detekt-gradle-plugin:1.17.0-RC1 was found. The consumer was configured to find a runtime of a library compatible with Java 8, packaged as a jar, and its dependencies declared externally, as well as attribute 'org.gradle.plugin.api-version' with value '7.0' but:
          - Variant 'apiElements' capability io.gitlab.arturbosch.detekt:detekt-gradle-plugin:1.17.0-RC1 declares a library, packaged as a jar, and its dependencies declared externally:
              - Incompatible because this component declares an API of a component compatible with Java 11 and the consumer needed a runtime of a component compatible with Java 8
              - Other compatible attribute:
                  - Doesn't say anything about org.gradle.plugin.api-version (required '7.0')
```

## Your Environment
* Version of detekt used: 1.17.0-RC1
* Operating System and version: macOS 11.2.3

cortinico:
From https://github.com/detekt/detekt/pull/3727#issuecomment-830733869

We need to specify the Java `sourceCompatibility` and `targetCompatibility` inside the `module` plugin.
Sadly the -RC1 was build with JDK11 and and this makes it unusable for users on JRE 8

Fixes #3731

### DIFF
--- a/buildSrc/src/main/kotlin/module.gradle.kts
+++ b/buildSrc/src/main/kotlin/module.gradle.kts
@@ -73,6 +73,8 @@ dependencies {
 java {
     withSourcesJar()
     withJavadocJar()
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
 }
 
 publishing {


buildSrc/src/main/kotlin/module.gradle.kts
```diff
@@ -73,6 +73,8 @@ dependencies {
 java {
     withSourcesJar()
     withJavadocJar()
+    sourceCompatibility = JavaVersion.VERSION_1_8
```
3flex, line 76: Shouldn't need `sourceCompatibility` set here, since there's no Java source in the project.

buildSrc/src/main/kotlin/module.gradle.kts
```diff
@@ -73,6 +73,8 @@ dependencies {
...
+    targetCompatibility = JavaVersion.VERSION_1_8
```
3flex, line 77: Dropping `targetCompatibility` in #3575 breaking things here is unfortunate, and obviously not expected :(. Kotlin apparently ignores the `jvmTarget` option on the compilation tasks, or at least that value doesn't filter down to the generated JARs.

[JetBrains says](https://youtrack.jetbrains.com/issue/KT-45335#focus=Comments-27-4765781.0-0) the proper way to set the value is with this:
```kotlin
configurations["runtimeElements"].attributes {
    attribute(TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE, 8)
}
```

But I agree setting `targetCompatibility` here is easier to understand.

buildSrc/src/main/kotlin/module.gradle.kts
cortinico: I believe you're right, although I wasn't able to find any reference that `sourceCompatibility` is using only for compilation of Java file and I can't exclude other side effects. I won't mind keeping both.

buildSrc/src/main/kotlin/module.gradle.kts
cortinico: Yeah the suggested approach feel really fragile. I've ⭐ the YouTrack issue. Ideally this could be addressed by the KGP in some form.

chao2zhang: This is probably the behavior configured in Gradle: https://github.com/gradle/gradle/blob/v7.0.0/subprojects/plugins/src/main/java/org/gradle/api/plugins/jvm/internal/DefaultJvmPluginServices.java#L211 - The target compatibility is used to configure the `org.gradle.jvm.version` attribute

chao2zhang (approved):
This is probably worth paying attention to in future Gradle upgrades.

chao2zhang:
This PR is likely not necessary when we bump to Kotlin 1.5
https://blog.jetbrains.com/kotlin/2021/05/kotlin-1-5-0-released/

`New default JVM target: 1.8. The 1.6 target is deprecated.`

PR #3740, opened 2021-05-05
Ignore gradle plugin in codecov

BraisGabin:
The code coverage reports from codecov about our gradle plugin aren't correct. For that reason it's better to ignore them so we don't get that noise in our PRs.

Following this documentation: https://docs.codecov.io/docs/ignoring-paths

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -10,3 +10,6 @@ coverage:
         target: 80%
         # Only post a patch status to pull requests
         only_pulls: true
+
+ignore:
+  - "detekt-gradle-plugin/"
--- a/.github/workflows/codecoverage.yaml
+++ b/.github/workflows/codecoverage.yaml
@@ -25,7 +25,7 @@ jobs:
           java-version: 11
 
       - name: Generate Coverage Report
-        run: ./gradlew jacocoTestReport --no-build-cache
+        run: ./gradlew jacocoTestReport --no-build-cache -x :detekt-gradle-plugin:test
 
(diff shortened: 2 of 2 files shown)


3flex:
Does filtering out the project in the root build.gradle.kts have the same effect?

There are already some subprojects that are ignored that way, might add well use the same method if possible.

chao2zhang (approved):

chao2zhang:
> Does filtering out the project in the root build.gradle.kts have the same effect?

Good point. The build.gradle.kts configures jacoco test report while codecov.yml configures the code coverage CI.
If we want to report code coverage for detekt-gradle-plugin separately, I believe modifying codecov.yml as shown in this PR would be a reasonable choice.

schalkms:
> > Does filtering out the project in the root build.gradle.kts have the same effect?

> Good point. The build.gradle.kts configures jacoco test report while codecov.yml configures the code coverage CI.
If we want to report code coverage for detekt-gradle-plugin separately, I believe modifying codecov.yml as shown in this PR would be a reasonable choice.

Yes, this would be the preferred way.

schalkms (approved):

3flex:
`:detekt-gradle-plugin:test` task should also be skipped on the code coverage CI job which will speed it up. It doesn't currently use the build cache and the Gradle test task takes 12 minutes (only part of run time is in parallel with other tasks).

https://scans.gradle.com/s/dt73yepgj2gxm/timeline

PR #3741, opened 2021-05-05
Enable Gradle's version catalogs

### LINKED ISSUE #3688: Use Gradle 7.0 version catalog
We are managing dependencies in an inconsistent way that 
- Most versions are in https://github.com/detekt/detekt/blob/main/buildSrc/build.gradle.kts
- Some versions are in https://github.com/detekt/detekt/blob/main/detekt-gradle-plugin/build.gradle.kts
- We also use Gradle BoM as introduced in https://github.com/detekt/detekt/pull/2792

Gradle 7 comes with an [opinionated way](https://docs.gradle.org/7.0/release-notes.html) to centralize dependency versions. 
Although this is experimental, it is still recommended over BoM platform https://docs.gradle.org/7.0/userguide/platforms.html#sub:platforms-vs-catalog. We should be able to centralize all dependencies in a single place.

3flex:
Closes #3688

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -16,7 +16,7 @@ allprojects {
     version = Versions.currentOrSnapshot()
 }
 
-jacoco.toolVersion = Versions.JACOCO
+jacoco.toolVersion = libs.versions.jacoco.get()
 
 tasks {
     jacocoTestReport {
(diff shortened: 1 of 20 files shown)


chao2zhang (approved):

schalkms (approved):
👍

PR #3748, opened 2021-05-06
Report redundant explicit imports

arhont375:
Fixes 3699

Check for redundant explicit imports according to the [kotlin docs](https://kotlinlang.org/docs/packages.html#default-imports).

1. Introduce new class in Style rules
2. Register this class
3. No tests yet

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitImport.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitImport.kt
@@ -1,0 +1,55 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.psi.KtImportDirective
+
+/**
+ * This rule checks for list of default imports.
+ * See [kotlin documentation](https://kotlinlang.org/docs/packages.html#imports).
+ */
+class RedundantExplicitImport(config: Config = Config.empty) : Rule(config) {
+
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Style,
+        "Mark default import. Default imported don't need to be specified explicitly.",
+        Debt.FIVE_MINS
+    )
+
+    override fun visitImportDirective(importDirective: KtImportDirective) {
+        super.visitImportDirective(importDirective)
+
+        val import = importDirective.importedFqName?.asString() ?: ""
+        if (defaultImports.contains(import.dropLastWhile { it != '.' })) {
+            report(
+                CodeSmell(
+                    issue,
+                    Entity.from(importDirective),
+                    "The import " +
+                            "$import is default import and don't need to be imported explicitly."
+                )
+            )
+        }
+    }
+}
+
+// As of Kotlin 1.5
+val defaultImports = setOf(
+    "kotlin.",
+    "kotlin.annotation.",
+    "kotlin.collections.",
+    "kotlin.comparisons.",
+    "kotlin.io.",
+    "kotlin.ranges.",
+    "kotlin.sequences.",
+    "kotlin.text.",
+    // JVM specific
+    "java.lang.",
+    "kotlin.jvm.",
+)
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitImport.kt
```diff
@@ -0,0 +1,55 @@
...
+    "kotlin.annotation.",
+    "kotlin.collections.",
+    "kotlin.comparisons.",
+    "kotlin.io.",
+    "kotlin.ranges.",
+    "kotlin.sequences.",
+    "kotlin.text.",
+    // JVM specific
+    "java.lang.",
+    "kotlin.jvm.",
```
arhont375, line 54: I there any way to know for which platform Detekt is running?

arhont375:
Don't know how to make sure rule is enabled/disabled by default

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitImport.kt
```diff
@@ -0,0 +1,59 @@
...
+)
```
BraisGabin, line 59: I think that this rule isn't this easy. For example `import kotlin.Result` needs to be imported if a class named `Result` is defined in the same package.

This code doesn't compile:

```kotlin
fun foo() {
  val a: Result<Unit> = runCatching { println("bar") }
}

class Result
```

This does:

```kotlin
import kotlin.Result // we need this

fun foo() {
  val a: Result<Unit> = runCatching { println("bar") }
}

class Result
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitImport.kt
chao2zhang: We don't support inferring the Kotlin platform, this will need to be a new entry for platform in `CliArgs` and the gradle plugin will infer the platform from the plugins applied.

I think inferring the platform is a nice-to-have for this PR though, because we can list possible redundant imports for all platforms.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitImport.kt
chao2zhang: I think running Detekt always comes with the assumption that the code can be compiled successfully. Checking the existence of `import kotlin.Result` should be sufficient enough.

BraisGabin: But the second snippet compiles and creates a false positive in this rule, isn't it?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitImport.kt
BraisGabin: Good point, we should create an issue to track this feature, right?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitImport.kt
arhont375: Indeed, let me try different things and see how it will handle them.

Also as shown in the test for some types from the range of  default imports as `java.lang.*` Kotlin have it's own type that shadow that type, for example:
```
import java.lang.Exception

fun a() = throw Exception()
```
In this case this Rule will report `java.lang.Exception` but we must import it as otherwise `kotlin.Exception` would be used. Not sure how better to handle this situation here :thinking: 

How much information Detekt have about surrounding classes/packages?

chao2zhang: I wouldn't say it is a false positive. We shouldn't clash with stdlib classes, because we then need to use `import my.example.Result` when we want to use the customized `Result` class

schalkms (commented):
What's the status of this PR? What's left to do in order to merge this into detekt?
I think this rule is really great. I can't wait to use it in my projects. 😃

arhont375:
> What's the status of this PR? What's left to do in order to merge this into detekt?
> I think this rule is really great. I can't wait to use it in my projects.

Currently I didn't have much time to return to this. 
As it was commented by maintainers needs better consideration of corner cases here. I may try to return to this in nearest weeks or it can be taken by someone else with more time

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitImport.kt
BraisGabin: We have a forbidden class name rule, don't we? We could add as default values those names. But that's not the point of this rule.

Issue #3743, opened 2021-05-05
detekt-cli 1.17.0-RC2 is published with multiple variants


3flex:
## Expected Behavior
Single variant with `"org.gradle.usage": "java-runtime"` (apiElements) is published in Gradle metadata.

## Observed Behavior
Two variants with `"org.gradle.usage": "java-runtime"` (apiElements, shadowRuntimeElements) are published in Gradle metadata.

## Steps to Reproduce
Compare .module file in https://repo1.maven.org/maven2/io/gitlab/arturbosch/detekt/detekt-cli/1.16.0/ & https://repo1.maven.org/maven2/io/gitlab/arturbosch/detekt/detekt-cli/1.17.0-RC2/

Diff: https://gist.github.com/3flex/c82cf481f4d45c0bb38ee04dd769bcb8

## Context
May make consuming detekt-cli in Gradle projects more difficult than it should be, see https://github.com/detekt/detekt/pull/3738

chao2zhang:
It looks like the new runtime variants were introduced in https://github.com/detekt/detekt/pull/3574

3flex:
Thanks, that's pointed me in the right direction.

I'll have a PR soon.

PR #3738, opened 2021-05-04
Set the org.gradle.dependency.bundling attribute to external

cortinico:
This fixes the failures on the PR https://github.com/detekt/detekt/pull/3727/ such as:

```
    > Could not resolve all files for configuration ':child1:detekt'.
       > Could not resolve io.gitlab.arturbosch.detekt:detekt-cli:1.17.0-RC2.
         Required by:
             project :child1
          > Cannot choose between the following variants of io.gitlab.arturbosch.detekt:detekt-cli:1.17.0-RC2:
              - runtimeElements
              - shadowRuntimeElements
            All of them match the consumer attributes:
              - Variant 'runtimeElements' capability io.gitlab.arturbosch.detekt:detekt-cli:1.17.0-RC2:
                  - Unmatched attributes:
                      - Provides org.gradle.category 'library' but the consumer didn't ask for it
                      - Provides org.gradle.dependency.bundling 'external' but the consumer didn't ask for it
                      - Provides org.gradle.jvm.version '8' but the consumer didn't ask for it
                      - Provides org.gradle.libraryelements 'jar' but the consumer didn't ask for it
                      - Provides org.gradle.status 'release' but the consumer didn't ask for it
                      - Provides org.gradle.usage 'java-runtime' but the consumer didn't ask for it
                      - Provides org.jetbrains.kotlin.localToProject 'public' but the consumer didn't ask for it
                      - Provides org.jetbrains.kotlin.platform.type 'jvm' but the consumer didn't ask for it
              - Variant 'shadowRuntimeElements' capability io.gitlab.arturbosch.detekt:detekt-cli:1.17.0-RC2:
                  - Unmatched attributes:
                      - Provides org.gradle.category 'library' but the consumer didn't ask for it
                      - Provides org.gradle.dependency.bundling 'shadowed' but the consumer didn't ask for it
                      - Provides org.gradle.libraryelements 'jar' but the consumer didn't ask for it
                      - Provides org.gradle.status 'release' but the consumer didn't ask for it
                      - Provides org.gradle.usage 'java-runtime' but the consumer didn't ask for it
                      -
```

PR #3750, opened 2021-05-07
Revert "Set the org.gradle.dependency.bundling attribute to external"

3flex:
Reverts detekt/detekt#3738

See #3743 & #3747. The workaround won't be necessary now we're back to publishing a single variant with the attribute `"org.gradle.usage": "java-runtime"`


chao2zhang (approved):

cortinico:
Thanks for taking care of this @3flex 🙏

PR #3761, opened 2021-05-11
Fix kdoc link on blog navigation

zmunm:
https://detekt.github.io/detekt/pages/kdoc/detekt-api/index.html
The link created by dokkaJekyll is not working due to the wrong location.


schalkms:
Thank you!

PR #3762, opened 2021-05-12
Invalidate GitHub Actions cache when dependencies change

3flex:
With https://github.com/detekt/detekt/pull/3741 the majority of Gradle dependencies moved from the BOM to `libs.versions.toml`. This means one of the inputs to the GitHub Actions cache keys would very rarely change, meaning the cache wouldn't be invalidated in some circumstances.

This also removes the restore keys. The reason is when there's a cache hit on one of the restore keys the cache isn't updated - this means that in the worst case a cache hit may happen on e.g. `cache-gradle-` which could be from any OS, JDK version or an old set of dependencies, which means the cached build cache and dependencies would not be usable in the Gradle build. This change should mean more frequent cache hits in Actions as well as Gradle's build cache.

### DIFF
--- a/.github/workflows/deploy-snapshot.yaml
+++ b/.github/workflows/deploy-snapshot.yaml
@@ -21,7 +21,7 @@ jobs:
         path: |
           ~/.gradle/caches/
           ~/.gradle/wrapper/
-        key: cache-gradle-${{ hashFiles('detekt-bom/build.gradle.kts') }}
+        key: cache-gradle-${{ hashFiles('gradle/libs.versions.toml') }}
         restore-keys: |
           cache-gradle-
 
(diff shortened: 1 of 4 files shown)


schalkms (approved):

cortinico:
> This also removes the restore keys. The reason is when there's a cache hit on one of the restore keys the cache isn't updated

Are you sure about this @3flex ?

The documentation says the opposite:
https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action

> When key doesn't match an existing cache, it's called a cache miss, and a new cache is created if the job completes successfully. When a cache miss occurs, the action searches for alternate keys called restore-keys.

3flex:
I must be going crazy... I don't know how I got the wrong idea about that. I could have sworn I saw the behaviour I described in builds on our repo and an issue opened on actions/cache repo suggesting the same.

But looking at some recent builds (before this PR was merged) I can see that it does in fact have the behaviour described in the documentation. I'll submit a PR to revert that commit.

PR #3768, opened 2021-05-12
Use annotations to configure rules in rules-complexity

marschwar:
This belongs to #3670 and replaces all configuration kdoc tags in `rules-complexity` with annotations. There are a few decisions to make:

* Named arguments vs. positional arguments
* Method references vs. lambda 

**simple properties**
```kotlin
private val prop: Boolean by config(defaultValue = false)
// or
private val prop: Boolean by config(false)
```

**fallback properties**
```kotlin
private val prop: Int by configWithFallback(fallbackPropertyName = "threshold", defaultValue = 1)
// or 
private val prop: Int by configWithFallback("threshold", 1)
````

**transformed properties**
```kotlin
private val prop: Regex by config("$^", String::toRegex)
// or
private val prop: Regex by config("$^") { it.toRegex() }
```

I am fairly ambivalent about this and I don't know if that is something we even have to agree upon. I am curious about your thoughts.

### DIFF
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
@@ -5,10 +5,12 @@ import io.gitlab.arturbosch.detekt.api.Debt
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Metric
+import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
-import io.gitlab.arturbosch.detekt.api.ThresholdRule
 import io.gitlab.arturbosch.detekt.api.ThresholdedCodeSmell
 import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import io.gitlab.arturbosch.detekt.api.internal.config
 import org.jetbrains.kotlin.psi.KtBinaryExpression
 import org.jetbrains.kotlin.psi.KtDoWhileExpression
 import org.jetbrains.kotlin.psi.KtExpression
@@ -36,14 +38,11 @@ import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
  *
  * fun hasCorrectEnding() = return !str.endsWith("foo") && !str.endsWith("bar") && !str.endsWith("_")
  * </compliant>
- *
- * @configuration threshold - the number of conditions which will trigger the rule (default: `4`)
  */
 @ActiveByDefault(since = "1.0.0")
 class ComplexCondition(
-    config: Config = Config.empty,
-    threshold: Int = DEFAULT_CONDITIONS_COUNT
-) : ThresholdRule(config, threshold) {
+    config: Config = Config.empty
+) : Rule(config) {
 
     override val issue = Issue(
         "ComplexCondition",
@@ -51,6 +50,9 @@ class ComplexCondition(
         "Complex conditions should be simplified and extracted into well-named methods if necessary.",
         Debt.TWENTY_MINS
     )
+
+    @Configuration("the number of conditions which will trigger the rule")
+    private val threshold: Int by config(DEFAULT_CONDITIONS_COUNT)
 
     override fun visitIfExpression(expression: KtIfExpression) {
         val condition = expression.condition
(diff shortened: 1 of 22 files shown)


detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArguments.kt
```diff
@@ -28,22 +30,20 @@ import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
...
  * sum(a = 1, b = 2, c = 3, d = 4)
  * </compliant>
- *
- * @configuration threshold - number of parameters that triggers this inspection (default: `3`)
  */
 @RequiresTypeResolution
-class NamedArguments(
-    config: Config = Config.empty,
-    threshold: Int = DEFAULT_FUNCTION_THRESHOLD
-) : ThresholdRule(config, threshold) {
```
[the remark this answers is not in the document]
marschwar: There is a TODO Checkbox in #3670 to remove it alltogether once all rules are updated.

schalkms (commented):
Thanks for pushing this initiative forward. This is really improves detekt's code base. No misuse of the comments anymore.

I have a few open questions.
- I guess the `generator` module also needs an update in order to output the correct website sources. Am I right?
- Do we need any other documentation? I'm thinking about the `Contributing.md` in order to not mislead new contributors or users. Detekt's docs become updated quite often. It's often forgotten when introducing changes.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArguments.kt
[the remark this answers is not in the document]
chao2zhang: Nice. Thank you for thinking ahead

marschwar:
The `generator` module and the contribution guide have already been updated.

schalkms:
> The `generator` module and the contribution guide have already been updated. 

Cool. I missed this, unfortunately.

schalkms (approved):

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
```diff
@@ -110,6 +112,6 @@ class ComplexCondition(
     }
 
     companion object {
-        const val DEFAULT_CONDITIONS_COUNT = 4
+        private const val DEFAULT_CONDITIONS_COUNT = 4
```
BraisGabin, line 115: Probably out of scope but I think that inline this values will improve the readability of this code.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArguments.kt
```diff
@@ -28,22 +30,20 @@ import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
...
+class NamedArguments(config: Config = Config.empty) : Rule(config) {
 
     override val issue = Issue(
         "NamedArguments",
         Severity.Maintainability,
-        "Function invocation with more than $threshold parameters must all be named",
+        "Parameters of function invocation must all be named",
```
BraisGabin, line 40: Can't we use the threshold? The description that you set here is not correct.

BraisGabin (approved):

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArguments.kt
marschwar: That is a great catch. Unfortunately that is not possible. When `threshold` is evaluated, the ruleId is used to lookup the config key. In order to get the ruleId, the issue is created. -> infinite loop

How about: _"Prefer named arguments for functions to improve readability."_  This is semi-consistent with rules like ComplexInterface or ComplexCondition where only the general rule is described without mentioning the exact threshold.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
marschwar: Great point. I was afraid to trigger magic number violations but it seems to work

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArguments.kt
chao2zhang: This is actually a bug today: Even if the `threshold` in configuration is 4, the `issue.description` will always be `Function invocation with more than 3 parameters must all be named` because the existing description is reading value from the `threshold` constructor argument.

chao2zhang (approved):

PR #3769, opened 2021-05-12
Use annotations to configure rules in rules-naming

marschwar:
This belongs to #3670 and replaces all configuration kdoc tags in `rules-naming` with annotations.

### DIFF
--- a/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ForbiddenClassName.kt
+++ b/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ForbiddenClassName.kt
@@ -7,15 +7,14 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
-import io.gitlab.arturbosch.detekt.api.internal.valueOrDefaultCommaSeparated
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import io.gitlab.arturbosch.detekt.api.internal.config
 import org.jetbrains.kotlin.psi.KtClassOrObject
 
 /**
  * Reports class names which are forbidden per configuration.
  * By default this rule does not report any classes.
  * Examples for forbidden names might be too generic class names like `...Manager`.
- *
- * @configuration forbiddenName - forbidden class names (default: `[]`)
  */
 class ForbiddenClassName(config: Config = Config.empty) : Rule(config) {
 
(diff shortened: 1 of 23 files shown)


detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ForbiddenClassName.kt
```diff
@@ -7,15 +7,14 @@ import io.gitlab.arturbosch.detekt.api.Entity
...
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import io.gitlab.arturbosch.detekt.api.internal.config
 import org.jetbrains.kotlin.psi.KtClassOrObject
 
 /**
  * Reports class names which are forbidden per configuration.
  * By default this rule does not report any classes.
  * Examples for forbidden names might be too generic class names like `...Manager`.
- *
- * @configuration forbiddenName - forbidden class names (default: `[]`)
```
chao2zhang, line 18: Good catch. It feels like a semantically correct configuration name is `forbiddenNames` with `s`, but that would be out of scope for this PR.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/VariableNaming.kt
```diff
@@ -72,8 +75,6 @@ class VariableNaming(config: Config = Config.empty) : Rule(config) {
 
     companion object {
         const val VARIABLE_PATTERN = "variablePattern"
-        const val PRIVATE_VARIABLE_PATTERN = "privateVariablePattern"
         const val EXCLUDE_CLASS_PATTERN = "excludeClassPattern"
```
chao2zhang, line 78: Do we still need these constants?

chao2zhang (approved):

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/VariableNaming.kt
marschwar: Yes, they are used in `NamingConventionCustomPatternTest` which tests multiple rules at the same time and I think it would lead to confusion if they were inlined there. This way you at least have a chance to see where they are coming from.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ForbiddenClassName.kt
marschwar: Yes, very unfortunate

schalkms (approved):

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/VariableNaming.kt
BraisGabin: We should simplify split that test in multiple ones. But that's completely out of the scope of this PR.

BraisGabin (approved):

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ForbiddenClassName.kt
```diff
@@ -25,12 +24,15 @@ class ForbiddenClassName(config: Config = Config.empty) : Rule(config) {
         "Forbidden class name as per configuration detected.",
         Debt.FIVE_MINS
     )
-    private val forbiddenNames = valueOrDefaultCommaSeparated(FORBIDDEN_NAME, emptyList())
-        .map { it.removePrefix("*").removeSuffix("*") }
+
+    @Configuration("forbidden class names")
+    private val forbiddenName: List<String> by config(listOf<String>()) { names ->
+        names.map { it.removeSurrounding("*") }
```
chao2zhang: As discovered in #3772, `removeSurrounding("*")` is not equivalent to `removePrefix("*").removeSuffix("*")`

chao2zhang:
@BraisGabin Let's use https://github.com/detekt/detekt/discussions/3784 to document configuration naming / deprecations

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ForbiddenClassName.kt
chao2zhang: https://github.com/detekt/detekt/discussions/3784 Already created a discussion.

chao2zhang (approved):

PR #3774, opened 2021-05-13
Use @Configuration in detekt-rules-style

marschwar:
This belongs to #3670 and replaces all configuration kdoc tags in detekt-rules-style.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -681,6 +681,7 @@ style:
   ThrowsCount:
     active: true
     max: 2
+    excludeGuardClauses: false
   TrailingWhitespace:
     active: false
   UnderscoresInNumericLiterals:
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
@@ -52,14 +52,12 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
             "(i.e. `java.time.LocalDate(java.time.Clock)`) which would report only call " +
             "with this concrete signature."
     )
-    private val methods: List<String> by config(
+    private val methods: List<Pair<String, List<String>?>> by config(
         listOf(
             "kotlin.io.println",
             "kotlin.io.print"
         )
-    )
-
-    private val forbiddenMethods = methods.map { extractMethodNameAndParams(it) }
+    ) { it.map(::extractMethodNameAndParams) }
 
     override fun visitCallExpression(expression: KtCallExpression) {
         super.visitCallExpression(expression)
@@ -90,7 +88,7 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
             .map { it.type.fqNameOrNull()?.asString() }
 
         if (methodName != null) {
-            forbiddenMethods
+            methods
                 .filter { methodName == it.first }
                 .forEach {
                     val expectedParamTypes = it.second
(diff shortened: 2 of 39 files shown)


detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -681,6 +681,7 @@ style:
   ThrowsCount:
     active: true
     max: 2
+    excludeGuardClauses: false
```
marschwar, line 684: This seems to have been an undocumented feature

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumber.kt
```diff
@@ -101,21 +82,40 @@ class MagicNumber(config: Config = Config.empty) : Rule(config) {
...
+    private val ignoreConstantDeclaration: Boolean by config(true)
+
+    @Configuration("whether magic numbers in companion object declarations should be ignored")
+    private val ignoreCompanionObjectPropertyDeclaration: Boolean by config(true)
+
+    @Configuration("whether magic numbers in annotations should be ignored")
+    private val ignoreAnnotation: Boolean by config(false)
+
+    @Configuration("whether magic numbers in named arguments should be ignored")
+    private val ignoreNamedArgument: Boolean by config(true)
```
marschwar, line 109: This previously was false by default but the documentation had true as the default. I opted to change it as the generated default config yaml was generated from the documentation.

schalkms (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
```diff
@@ -50,8 +45,21 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
...
+            "with this concrete signature."
+    )
+    private val methods: List<String> by config(
+        listOf(
+            "kotlin.io.println",
+            "kotlin.io.print"
+        )
+    )
+
+    private val forbiddenMethods = methods.map { extractMethodNameAndParams(it) }
```
BraisGabin, line 62: This one is to use a better name in the rule, right?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLength.kt
```diff
@@ -92,10 +93,6 @@ class MaxLineLength(config: Config = Config.empty) : Rule(config) {
     }
 
     companion object {
-        const val MAX_LINE_LENGTH = "maxLineLength"
         const val DEFAULT_IDEA_LINE_LENGTH = 120
```
BraisGabin, line 96: Can't this be private? or removed?

BraisGabin (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
marschwar: To be honest, primarily I did not like the type of the field in combination with its name. 
```kotlin
    private val methods: List<Pair<String, List<String>?>> by config...
```
I can change it for consistency.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/FunctionOnlyReturningConstant.kt
```diff
@@ -51,18 +46,24 @@ class FunctionOnlyReturningConstant(config: Config = Config.empty) : Rule(config
...
-    )
-        .map { it.removePrefix("*").removeSuffix("*") }
+    @Configuration("if overriden functions should be ignored")
+    private val ignoreOverridableFunction: Boolean by config(true)
+
+    @Configuration("if actual functions should be ignored")
+    private val ignoreActualFunction: Boolean by config(true)
+
+    @Configuration("excluded functions")
+    private val excludedFunctions: SplitPattern by config("describeContents") { SplitPattern(it) }
```
marschwar, line 56: Why does this default to "describeContents"? Is this a misconfiguration?
Should we improve the documentation while we are at it?

marschwar: I looked at the original commit and this was clearly not intended to have `'describeContents`' as the default.

PR #3782, opened 2021-05-16
Checkout files with LF line endings on Windows

3flex:
This ensures the line endings of files in git's working directory are LF on Windows, which matches behaviour on Linux & macOS.

This helps with Gradle's up-to-date checks - for example on Windows checkout the default-detekt-config.yml file is currently checked out with CRLF line endings. During the initial build the file will be rewritten by the generateDocumentation task with LF line endings, so on the next build any tasks (and their successors) that use the config file as an input will be out of date, causing many tasks to run again unnecessarily.

Two test assertions were dependent on the OS the test is running under, so those have been updated to run on files that will be checked out with CRLF and LF line endings regardless of the OS.

### DIFF
--- a/detekt-parser/src/test/resources/cases/.gitattributes
+++ b/detekt-parser/src/test/resources/cases/.gitattributes
@@ -1,0 +1,2 @@
+DefaultLf.kt   text eol=lf
+DefaultCrLf.kt text eol=crlf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,11 +1,11 @@
-* text eol=auto
+* text eol=lf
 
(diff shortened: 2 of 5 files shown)


detekt-parser/src/test/resources/cases/.gitattributes
```diff
@@ -0,0 +1,2 @@
+DefaultLf.kt   text eol=lf
+DefaultCrLf.kt text eol=crlf
```
BraisGabin, line 2: <3

BraisGabin (approved):

schalkms (approved):

PR #3790, opened 2021-05-17
Avoid overlapping outputs

3flex:
javadocJar & publishPluginJar output to the same file path, as did
sourcesJar & publishPluginJavaDocsJar, preventing Gradle's up-to-date
checks from working properly.


schalkms (approved):

BraisGabin (approved):

cortinico:
When releasing `1.18.0-RC1` I noticed that this PR is blocking the release process.
Specifically the sources.jar and the javadoc.jar are missing for `detekt-gradle-plugin`.

Maven Central is not allowing the whole bundle to be published if those artifacts are missing.

Should we revert this change?

schalkms:
> Should we revert this change?

I don't see an alternative.

3flex:
> I don't see an alternative.

I'd argue there's no reason to publish the plugin artifact to Maven Central anyway, since it's already published to the Gradle Plugin Portal.

That's probably why the two publishing methods provided by Gradle don't work well together in this case (overlapping outputs) since it's not a typical setup.

cortinico:
> I'd argue there's no reason to publish the plugin artifact to Maven Central anyway, since it's already published to the Gradle Plugin Portal.

I'm unsure about the historical reason why `detekt-gradle-plugin` is on Central as well.

One thing to point out is that `detekt-gradle-plugin` relies on `detekt-bom` and a lot of other deps that are published on Central. So it sounds reasonable to me that we publish all of our artifacts to me on Central + we offer `detekt-gradle-plugin` as a convenience for Gradle users.

PR #3792, opened 2021-05-17
Revert "Noisy gradle (#3655)"

### LINKED ISSUE #3791: `UnusedImports` marks all Java imports as unused
## Expected Behavior
Not unused imports from Java files in Kotlin files shouldn't be reported as part of `UnusedImports`.

## Observed Behavior
After upgrading to detekt 0.17.0, all Java imports are marked as unused.

## Steps to Reproduce
1. Create a Java file
2. Create a Kotlin file
(the files can be in the same gradle module)
3. Import and use some Java dependency in the Kotlin file

## Your Environment
* Version of detekt used: 0.17.0
* Relevant part of the config applied to every gradle module in the project:

```groovy
detekt {
    parallel = true
    failFast = false
    autoCorrect = !project.hasProperty("ci")
    buildUponDefaultConfig = true
    config = project.parent.files(".static/config/default-detekt-config.yml")
    reports {
        html.enabled = true
        xml.enabled = false
        txt.enabled = false
        sarif.enabled = false
    }
}
```

BraisGabin:
I don't know exactly why but we should provide the java files as input to the cli to make it work properly. I'll check exactyl why and I'll find another way to fix the issue that #3655 wanted to fix.

Closes #3791

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -106,6 +106,8 @@ class DetektPlugin : Plugin<Project> {
         const val BASELINE_TASK_NAME = "detektBaseline"
         const val DETEKT_EXTENSION = "detekt"
         private const val GENERATE_CONFIG = "detektGenerateConfig"
+        internal val defaultExcludes = listOf("build/")
+        internal val defaultIncludes = listOf("**/*.kt", "**/*.kts")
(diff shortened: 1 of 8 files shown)


chao2zhang (approved):

cortinico:
We discussed this internally, but for the sake of transparency:

> I don't know exactly why but we should provide the java files as input to the cli to make it work properly. I'll check exactyl why and I'll find another way to fix the issue that #3655 wanted to fix.

Not passing the `.java` had the side effect of causing unresolved types for Java classes. This caused false positives in rules that relied on type resolution. Therefore we're reverting this change and we will tackle the noise logging in a separate PR.

This is most likely going to be shipped in a patch release.

cortinico (approved):

PR #3795, opened 2021-05-18
New rule: BooleanPropertyNaming

AlinaRakhimova:
I added a new rule to check the naming of boolean variables to improve the readability of the code.

For example, 

`val enabled = true`

Instead of should prefer to 

`val isEnabled = true`


schalkms (commented):
Hello Alina,
I don't know if you are aware of it, but there exists already a rule, which works essentially the same as your rule. The controversial rule is called `NonBooleanPropertyPrefixedWithIs`. Hence, detekt disables this rule by default.

I don't think detekt needs a second rule even if minor differences exist. Therefore, I have to unfortunately declare this as *won't merge*. Still, you can use this rule in a custom rule set as you like. See [here](https://detekt.github.io/detekt/extensions.html#custom-rulesets) for more details.

https://github.com/detekt/detekt/blob/main/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NonBooleanPropertyPrefixedWithIs.kt

shumidub:
But, this is different rules. NonBooleanPropertyPrefixedWithIs - for NOT boolean property and don't check boolean properties naming.

BooleanPropertyPrefixedWithAllowedWords - for booleans.

cortinico:
Thanks for taking the time to add a new rule 🙏

I believe we could consider this rule, but it falls a bit in the `controversial` area. 
Specifically I don't agree that:

```
val enabled = true
```
is less readable than:
```
val isEnabled = true
```
but that's debatable, so I can understand how developers might want to enforce some prefixes on their `Boolean` property.

I would like to get other maintainers opinion on this. 

Some things that needs to be addressed/discussed:
- The rule name should probably be tweaked. Ideally `BooleanPropertyPrefix` or `BooleanPropertyNaming`.
- We already have a similar rule that does this, is `VariableNaming`. Your rule is basically doing the same behavior but just for `Boolean` property. I'm unsure if this should be a separate rule or we should extend `VariableNaming` (that will impose type resolution on this rule, making it a "mixed rule" that we try to avoid #2994).
- Do we have some reference from the official Kotlin guidelines where the boolean variable naming is suggested? If so, that should be linked in the documentation.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWords.kt
```diff
@@ -0,0 +1,99 @@
...
+import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.typeBinding.createTypeBindingForReturnType
+
+/**
+ * Reports when a boolean property doesn't have one of the following prefixes:
+ * is/has/should/need/noNeed/was/are/may/can/had/for/with.
+ *
+ * @requiresTypeResolution
```
cortinico, line 23: You should use the `@RequiresTypeResolution` annotation instead.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWords.kt
```diff
@@ -0,0 +1,99 @@
...
+import org.jetbrains.kotlin.psi.KtCallableDeclaration
+import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.typeBinding.createTypeBindingForReturnType
+
+/**
+ * Reports when a boolean property doesn't have one of the following prefixes:
+ * is/has/should/need/noNeed/was/are/may/can/had/for/with.
+ *
```
cortinico, line 22: `<compliant>` and `<noncompliant>` examples are missing

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWords.kt
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        if (parameter.hasValOrVar()) {
+            validateDeclaration(parameter)
+        }
+    }
+
+    override fun visitProperty(property: KtProperty) {
+        super.visitProperty(property)
+        if (property.isConstant()) return
```
cortinico, line 53: Why are you skipping constants?

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWords.kt
```diff
@@ -0,0 +1,99 @@
...
+        javaClass.simpleName, Severity.CodeSmell,
+        description,
+        Debt.FIVE_MINS
+    )
+
+    override fun visitParameter(parameter: KtParameter) {
+        super.visitParameter(parameter)
+        val name = parameter.identifierName()
+
+        if (name in listOf(VALUE_CONDITION, VALUE_PREDICATE, VALUE_IT, UNUSED_PARAM_NAME)) {
```
cortinico, line 42: Why are you filtering for `condition` and `predicate`?

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWords.kt
```diff
@@ -0,0 +1,99 @@
...
+ * is/has/should/need/noNeed/was/are/may/can/had/for/with.
+ *
+ * @requiresTypeResolution
+ */
+class BooleanPropertyPrefixedWithAllowedWords(config: Config = Config.empty) : Rule(config) {
+
+    private val description = "Boolean property name should starts with " +
+        "'is/has/should/need/noNeed/was/are/may/can/had/for/with' prefix."
+
+    private val regex = Regex("^(is|has|should|need|noNeed|was|are|may|can|had|for|with)")
```
cortinico, line 30: The list of prefixes should be configurable

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWords.kt
```diff
@@ -0,0 +1,99 @@
...
+            issue,
+            Entity.from(declaration),
+            message = "$description Actual type of $name: $typeName"
+        )
+    }
+
+    private fun getTypeName(parameter: KtCallableDeclaration): String? {
+        return parameter.createTypeBindingForReturnType(bindingContext)
+            ?.type
+            ?.getJetTypeFqName(false)
```
cortinico, line 88: You should not use `getJetTypeFqName` but `fqNameOrNull`: https://github.com/detekt/detekt/pull/3613

detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWordsSpec.kt
```diff
@@ -0,0 +1,181 @@
...
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+
+class BooleanPropertyPrefixedWithAllowedWordsSpec : Spek({
+    setupKotlinEnvironment()
+
+    val subject by memoized { BooleanPropertyPrefixedWithAllowedWords() }
+    val env: KotlinCoreEnvironment by memoized()
+
+    describe("IsPropertyNaming rule") {
```
cortinico, line 16:
```suggestion
    describe("BooleanPropertyPrefixedWithAllowedWords rule") {
```

detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWordsSpec.kt
```diff
@@ -0,0 +1,181 @@
...
+
+        context("argument declarations") {
+            it("should warn about Kotlin Boolean") {
+                val code = """data class O (var default: Boolean)"""
```
cortinico, line 20: `O` is an unlucky char for a class name. It's hard to read if it's a capital O or a zero 0. Maybe change with `Data` or whatever class name you prefer :)

schalkms:
> I believe we could consider this rule, but it falls a bit in the controversial area

The question is whether more controversial rules shall exist in detekt considering the increasing maintenance effort in this project.

> We already have a similar rule that does this, is VariableNaming. Your rule is basically doing the same behavior but just for Boolean property. I'm unsure if this should be a separate rule or we should extend VariableNaming (that will impose type resolution on this rule, making it a "mixed rule" that we try to avoid #2994).

I agree with this and forgot to also reference the `VariableNaming` rule.

> Do we have some reference from the official Kotlin guidelines where the boolean variable naming is suggested? If so, that should be linked in the documentation.

I don't think there exists one. This constitutes more of a common rule in many projects.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWords.kt
AlinaRakhimova: I think we can give the ability to name the parameters "condition" and "predicate". For example, it can be useful for creating common extension functions with a boolean parameter.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWords.kt
AlinaRakhimova: I thought constants could use their own naming conventions. Do you think I should remove this condition?

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWords.kt
cortinico: I believe there should not be any hardcoded "exception", to follow the least astonishment principle (https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

Why "condition" and "predicate" should be an exception, and other synonym should not (e.g. "criteria", "rule", "guard", "expression", ...)

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWords.kt
cortinico: > Do you think I should remove this condition?

Not necessarily :) I'm just challenging some of the rule assumptions to understand the use cases.

> I thought constants could use their own naming conventions.

Why this? Do you have some examples?

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWords.kt
AlinaRakhimova: You're right.
Then I'll make the list of prefixes configurable and remove all of these conditions. The user will be able to customize the rule through the configuration

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyPrefixedWithAllowedWords.kt
AlinaRakhimova: Actually, I have no real examples. It seems to me that logical constants are rarely used. Then I will remove this condition.

BraisGabin:
> Do we have some reference from the official Kotlin guidelines where the boolean variable naming is suggested? If so, that should be linked in the documentation.

I would not add this rule if there is not a source or reference to back it. I think that, as detekt, we should not be that opinated. #3150 was closed even with references. I understand the rule and maybe I would use it in my project too, but I don't think we should have it in our default rulesets.

AlinaRakhimova:
If the user does not want to use this rule, he will not enable it in the config file

AlinaRakhimova:
> > Do we have some reference from the official Kotlin guidelines where the boolean variable naming is suggested? If so, that should be linked in the documentation.
> 
> I would not add this rule if there is not a source or reference to back it. I think that, as detekt, we should not be that opinated. #3150 was closed even with references. I understand the rule and maybe I would use it in my project too, but I don't think we should have it in our default rulesets.

If the user does not want to use this rule, he will not enable it in the config file

shumidub:
The rule is very useful, it improves the readability and clarity of the code. It will be great if it will be available as an optional rule.

chao2zhang:
With regards to references:
1. Java Bean requires boolean property to be prefixed with `is` (https://www.oracle.com/java/technologies/javase/javabeans-spec.html)
In addition, for boolean properties, we allow a getter method to match the pattern:
`public boolean is<PropertyName>();`
This “is<PropertyName>” method may be provided instead of a “get<PropertyName>” method, or it may be provided in addition to a “get<PropertyName>” method.
In either case, if the “is<PropertyName>” method is present for a boolean property then we will
use the “is<PropertyName>” method to read the property value.
An example boolean property might be:
```
   public boolean isMarsupial();
   public void setMarsupial(boolean m);
```

2. Kotlin-Java interop (https://kotlinlang.org/docs/java-interop.html#getters-and-setters)
Boolean accessor methods (where the name of the getter starts with is and the name of the setter starts with set) are represented as properties which have the same name as the getter method.

3. Android lint implemented checks requiring specific bean style for Java methods `Accessor methods require a ‘get’ prefix or for boolean-returning methods an ‘is’ prefix can be used.`
https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-master-dev:lint/libs/lint-checks/src/main/java/com/android/tools/lint/checks/InteroperabilityDetector.kt;l=141?q=Interoperability%20lint

The above references apply to Java code, yet we do not have a Kotlin coding convention encouraging or discouraging the naming pattern for a boolean property. My vote would be yes as this rule is rather easy to maintain and we could document these references in the KDoc so that we keep this rule disabled in the future.

chao2zhang:
Other thoughts? We probably want to reach a consensus on this PR before the next release.

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -363,6 +363,10 @@ formatting:
 
 naming:
   active: true
+  BooleanPropertyNaming:
+    active: false
+    allowedPrefixes: '^(is|has|should|need|noNeed|was|are|may|can|had|for|with)'
```
cortinico, line 368:
```suggestion
    allowedPrefixes: '^(is|has|are)'
```

I believe the default value is a bit too long? 🤔 Or at least is really opinionated. I would opt for a shorter list if possible.

cortinico (approved):
I believe we could add this rule and have it disabled 👍 It's not as controversial as other and the maintenance cost is not as high.

detekt-core/src/main/resources/default-detekt-config.yml
shumidub: maybe  '^(is|has|are|should)' ? 🤔

but the current long list is not bad either

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNaming.kt
```diff
@@ -0,0 +1,101 @@
...
+import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.typeBinding.createTypeBindingForReturnType
+
+/**
+ * Reports when a boolean property doesn't have one of the following prefixes:
+ * is/has/should/need/noNeed/was/are/may/can/had/for/with.
+ *
+ * * <noncompliant>
```
marschwar, line 24:
```suggestion
 * <noncompliant>
```

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNaming.kt
```diff
@@ -0,0 +1,101 @@
...
+ * val progressBar: Boolean = true
+ * </noncompliant>
+ *
+ * <compliant>
+ * val hasProgressBar: Boolean = true
+ * </compliant>
+ *
+ * @configuration allowedPrefixes - naming pattern (default: `'^(is|has|should|need|noNeed|was|are|may|can|had|for|with)'`)
```
marschwar, line 32: Please remove the kdoc configuration in favour of the `@Configuration` annotation.  e.g.

```kotlin
@Configuration("Description of this configuration option")
private val allowedPrefixes: Regex by config("^(is|has)'", String::toRegex)
```

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNaming.kt
```diff
@@ -0,0 +1,101 @@
...
+ */
+@RequiresTypeResolution
+class BooleanPropertyNaming(config: Config = Config.empty) : Rule(config) {
+
+    private val allowedPrefixes by LazyRegex(
```
marschwar, line 37: This should either be called `allowedPattern` (or something similar or it should be a list of Strings
```kotlin 
@Configuration("Description of this configuration option")
private val allowedPattern: Regex by config("^(is|has)'", String::toRegex)
// or
@Configuration("Description of this configuration option")
private val allowedPrefixes: List<String> by config(listOf("is", "has"))
```

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NamingRules.kt
```diff
@@ -32,6 +32,7 @@ class NamingRules(config: Config = Config.empty) : MultiRule() {
     private val forbiddenClassNameRule = ForbiddenClassName(config)
     private val constructorParameterNamingRule = ConstructorParameterNaming(config)
     private val functionParameterNamingRule = FunctionParameterNaming(config)
+    private val booleanPropertyNamingRule = BooleanPropertyNaming(config)
```
marschwar, line 35: I could be wrong on this, but please double check if `visitParameter()` is ever called on your rule. Maybe someone else can suggest whether this rule should be part of the MutlRule or should be set up separately.

detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNamingSpec.kt
```diff
@@ -0,0 +1,142 @@
...
+                        var hasDefault: Boolean = true
+                    }
+                    """
+                val findings = subject.compileAndLintWithContext(env, code)
+
+                assertThat(findings).isEmpty()
+            }
+        }
+    }
+})
```
marschwar, line 142: Maybe add an additional test case that checks if the configuration option is picked up correctly?

shumidub (approved):

marschwar (changes requested):
Please use the annotation based rule configurations instead of kdoc.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NamingRules.kt
AlinaRakhimova: Hi!
`visitParameter()` is called on this rule
I added my rule to this class as well as `NonBooleanPropertyPrefixedWithIs`

detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNamingSpec.kt
AlinaRakhimova: Test was added

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -401,6 +401,10 @@ formatting:
 
 naming:
   active: true
+  BooleanPropertyNaming:
+    active: false
+    allowedPrefixes: '^(is|has|are)'
```
marschwar, line 406: Please rerun `./gradlew :detekt-generator:generateDocumentation` for the renaming of the property to take effect.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNaming.kt
```diff
@@ -0,0 +1,101 @@
...
+ * </compliant>
+ */
+@RequiresTypeResolution
+class BooleanPropertyNaming(config: Config = Config.empty) : Rule(config) {
+
+    @Configuration("naming pattern")
+    private val allowedPattern by LazyRegex(
+        ALLOWED_PREFIXES,
+        "^(is|has|are)"
+    )
```
marschwar, line 40:
```suggestion
    private val allowedPattern: Regex by config("^(is|has|are)", String::toRegex)
```
`LazyRegex` will be marked deprecated soon.

marschwar: I think this confusing to the user. We should either allow users to specify regular expression patterns that match the entire boolean property or have a list of prefixes. Having a pattern that matches only parts of the name to me feels rather unusual.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNaming.kt
```diff
@@ -0,0 +1,101 @@
...
+import io.gitlab.arturbosch.detekt.rules.identifierName
+import org.jetbrains.kotlin.psi.KtCallableDeclaration
+import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.typeBinding.createTypeBindingForReturnType
+
+/**
+ * Reports when a boolean property doesn't have one of the following prefixes:
+ * is|has|are.
```
marschwar, line 23: I think it is _"dangerous"_ to put the defaults into the rule description. This will eventually deviate from the truth.

AlinaRakhimova:
> Please use the annotation based rule configurations instead of kdoc.

I made it

cortinico:
Is there anything missing to merge this?

AlinaRakhimova:
> Is there anything missing to merge this?

Hi! All discussions were resolved

cortinico:
> Hi! All discussions were resolved

Can I ask you to address the CI failures?

AlinaRakhimova:
> > Hi! All discussions were resolved
> 
> Can I ask you to address the CI failures?

I'll try to do it

marschwar:
I think the code itself is fine (once it compiles again :smiley: - The config delegate was moved to a different package.). 

I have an issue with its inconsistent documentation. I know that it started out to focus on variable name prefixes, but when the rule was renamed to `BooleanPropertyNaming` all mentions of _"prefix"_ should have been removed from it as well. The users will not look at the code but rather its documentation so please be precise with what the rule does and how it can be configured.

Thank you for your contribution. I think it's really close. Sorry for the delay.

marschwar (approved):

cortinico:
Thanks for the contrib @AlinaRakhimova, sorry it took a bit longer than usual 🙏

Issue #3671, opened 2021-04-10
Use dynamic list format for default-detekt-config.yml


marschwar:
## Expected Behavior
The list of default strings that are inferred from the `config` delegate should dynamically choose the appropriate format for the yaml file. 

**Suggestion**:  

When the list fits on one line (less than xx characters), the one-line format is used. 

```yaml
ForbiddenComment:
  values: ['TODO:', 'FIXME:', 'STOPSHIP:', '@author']
``` 
For longer lists, the multi-line format is used.

```yaml
TooGenericExceptionCaught:
  exceptionNames:
   - ArrayIndexOutOfBoundsException
   - Error
   - Exception
   - IllegalMonitorStateException
   - NullPointerException
   - IndexOutOfBoundsException
   - RuntimeException
   - Throwable
```

## Current Behavior
Currently, the single-line format is always generated.

## Context
This is a follow up to #3637.
No matter which format is used, the rule is executed with the correct values. This is rather a matter of readability and maintainability.

3flex:
It should be multi line in all cases IMHO. Config change diffs are clearer that way.

PR #3798, opened 2021-05-19
Use annotations to configure rules in detekt-rules-exceptions

marschwar:
This belongs to #3670 and replaces all configuration kdoc tags with annotations.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -186,7 +186,7 @@ exceptions:
   active: true
   ExceptionRaisedInUnexpectedLocation:
     active: true
-    methodNames: [toString, hashCode, equals, finalize]
+    methodNames: ['toString', 'hashCode', 'equals', 'finalize']
   InstanceOfCheckForException:
     active: false
     excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
@@ -203,11 +203,7 @@ exceptions:
     ignoreLabeled: false
   SwallowedException:
     active: true
-    ignoredExceptionTypes:
-     - InterruptedException
-     - NumberFormatException
-     - ParseException
-     - MalformedURLException
+    ignoredExceptionTypes: ['NumberFormatException', 'InterruptedException', 'ParseException', 'MalformedURLException']
     allowedExceptionNameRegex: '_|(ignore|expected).*'
   ThrowingExceptionFromFinally:
     active: true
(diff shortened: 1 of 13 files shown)


detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -203,11 +203,7 @@ exceptions:
     ignoreLabeled: false
   SwallowedException:
     active: true
-    ignoredExceptionTypes:
-     - InterruptedException
-     - NumberFormatException
-     - ParseException
-     - MalformedURLException
+    ignoredExceptionTypes: ['NumberFormatException', 'InterruptedException', 'ParseException', 'MalformedURLException']
```
marschwar, line 206: I realize this is a problem for git diffs. It is already adressed in #3671

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -216,32 +212,17 @@ exceptions:
   ThrowingExceptionsWithoutMessageOrCause:
     active: true
     excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
-    exceptions:
-     - IllegalArgumentException
-     - IllegalStateException
-     - IOException
+    exceptions: ['ArrayIndexOutOfBoundsException', 'Error', 'Exception', 'IllegalMonitorStateException', 'NullPointerException', 'IndexOutOfBoundsException', 'RuntimeException', 'Throwable']
```
marschwar, line 215: There previously was a disconnect between the documented and the actual default. Which one should we keep?

schalkms (approved):

detekt-core/src/main/resources/default-detekt-config.yml
BraisGabin: I think that these list is better.

detekt-core/src/main/resources/default-detekt-config.yml
BraisGabin: I agree with @3flex in that issue, use multiline in all of them.

BraisGabin (approved):

detekt-core/src/main/resources/default-detekt-config.yml
cortinico: > There previously was a disconnect between the documented and the actual default.

As a rule of thumb: `code >> documentation`. Always prefer what the code says as docs could get out of date

detekt-core/src/main/resources/default-detekt-config.yml
cortinico: Are you following up to multiline in a separate PR?

cortinico (approved):

detekt-core/src/main/resources/default-detekt-config.yml
marschwar: Yes I am working on a fix for the mutliline issue.

Issue #3248, opened 2020-11-25
Detekt parallel crash


VladislavSumin:
## Observed Behavior
 Detekt crashes when executed in a project with multiple modules and enabled org.gradle.parallel with java.lang.ClassNotFoundException: kotlin.ExceptionsKt

The problem is reproducible with 100 percent probability.
I found similar closed issue #2629 

<details>
  <summary>Detekt configuration</summary>
  
 
  ```kotlin
plugins {
    id("io.gitlab.arturbosch.detekt").version("1.15.0-RC1")
}

subprojects {

    apply { plugin("io.gitlab.arturbosch.detekt") }
    detekt {
        failFast = true
        buildUponDefaultConfig = true
        baseline = file("${rootProject.projectDir}/static_analysis/config/detekt/baseline.xml")
        config = files("${rootProject.projectDir}/static_analysis/config/detekt/config.yml")
        autoCorrect = true
        ignoreFailures = false
        parallel = true

        reports {
            html.enabled = true
            xml.enabled = true
            txt.enabled = true
        }
    }

    tasks {
        withType<io.gitlab.arturbosch.detekt.Detekt> {
            this.jvmTarget = "1.8"
            description = "Overrides current baseline."

            setSource(files(rootDir))
            include("**/*.kt")
            include("**/*.kts")
            exclude("**/resources/**")
            exclude("**/build/**")
        }
    }
}
  ```
  
</details>

<details>
  <summary>Full stack trace</summary>
  
[2020-11-25 12:43:19.016616] FAILURE: Build failed with an exception.
[2020-11-25 12:43:19.016624] 
[2020-11-25 12:43:19.016630] * What went wrong:
[2020-11-25 12:43:19.016635] Execution failed for task ':core:detekt'.
[2020-11-25 12:43:19.016640] > kotlin/ExceptionsKt
[2020-11-25 12:43:19.016645] 
[2020-11-25 12:43:19.016650] * Try:
[2020-11-25 12:43:19.016655] Run with --info or --debug option to get more log output. Run with --scan to get full insights.
[2020-11-25 12:43:19.016660] 
[2020-11-25 12:43:19.016665] * Exception is:
[2020-11-25 12:43:19.016670] org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':core:detekt'.
[...]

PR #3799, opened 2021-05-19
plugin: Simpler classLoader caching (#3248)

Dominick1993:
Previous caching mechanism didn't work properly when multiple projects
were running detekt task in parallel and differed in classpath. Due to
the broken locking mechanism the classpath files would not match the
content in the classloader.

The new implementation uses ConcurrentHashMap which handles locking by
itself and allows us to store classloader for any combination of files.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
@@ -1,9 +1,8 @@
 package io.gitlab.arturbosch.detekt.internal
 
-import org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport
 import org.gradle.api.file.FileCollection
-import java.io.File
 import java.net.URLClassLoader
+import java.util.concurrent.ConcurrentHashMap
 
 fun interface ClassLoaderCache {
 
@@ -12,40 +11,18 @@ fun interface ClassLoaderCache {
 
 internal class DefaultClassLoaderCache : ClassLoaderCache {
 
-    private var loaderAndClasspathFiles: Pair<URLClassLoader, Set<File>>? = null
+    private val classpathFilesHashWithLoaders = ConcurrentHashMap<Int, URLClassLoader>()
 
     override fun getOrCreate(classpath: FileCollection): URLClassLoader {
         val classpathFiles = classpath.files
-        synchronized(this) {
-            val lastLoader = loaderAndClasspathFiles?.first
-            val lastClasspathFiles = loaderAndClasspathFiles?.second
-
-            if (lastClasspathFiles == null) {
-                cache(classpathFiles)
-            } else if (hasClasspathChanged(lastClasspathFiles, classpathFiles)) {
-                DefaultGroovyMethodsSupport.closeQuietly(lastLoader)
-                cache(classpathFiles)
-            }
-
-            return loaderAndClasspathFiles?.first ?: error("Cached or newly created detekt classloader expected.")
+        val classpathHashCode = HashSet(classpathFiles).hashCode()
+        return classpathFilesHashWithLoaders.getOrPut(classpathHashCode) {
+            URLClassLoader(
+                classpathFiles.map { it.toURI().toURL() }.toTypedArray(),
+                null /* isolate detekt environment */
+            )
         }
-    }
-
-    private fun cache(classpathFiles: Set<File>) {
-        loaderAndClasspathFiles = URLClassLoader(
-            classpathFiles.map { it.toURI().toURL() }.toTypedArray(),
-            null /* isolate detekt environment */
-        ) to classpathFiles
     }
 }
 
 object GlobalClassLoaderCache : ClassLoaderCache by DefaultClassLoaderCache()
-
-internal fun hasClasspathChanged(lastClasspathFiles: Set<File>, currentClasspathFiles: Set<File>): Boolean {
-    if (lastClasspathFiles.size != currentClasspathFiles.size) {
-        return true
-    }
-    return lastClasspathFiles.sorted()
-        .zip(currentClasspathFiles.sorted())
-        .any { (last, current) -> last != current || last.lastModified() != current.lastModified() }
-}
--- a/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCacheSpec.kt
+++ b/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCacheSpec.kt
@@ -11,53 +11,40 @@ import java.util.concurrent.CountDownLatch
 import java.util.concurrent.Executors
 import java.util.concurrent.TimeUnit
 import java.util.function.Supplier
-import kotlin.random.Random
 
 internal class ClassLoaderCacheSpec : Spek({
 
     describe("classpath changes") {
 
-        it("passes for same files") {
-            val changed = hasClasspathChanged(
-                setOf(FixedDateFile("/a/b/c")),
-                setOf(FixedDateFile("/a/b/c"))
-            )
+        it("same classloader is returned for the same files") {
+            val cache = DefaultClassLoaderCache()
+            val initialClassLoader = cache.getOrCreate(TestFileCollection(File("a/b/c")))
+            val secondClassLoader = cache.getOrCreate(TestFileCollection(File("a/b/c")))
 
-            assertThat(changed).isFalse()
+            assertThat(initialClassLoader === secondClassLoader)
         }
 
-        it("reports for different file count") {
-            val changed = hasClasspathChanged(
-                setOf(DifferentDateFile("/a/b/c"), DifferentDateFile("/c/b/a")),
-                setOf(DifferentDateFile("/a/b/c"))
-            )
+        it("different classloaders are returned for different files") {
+            val cache = DefaultClassLoaderCache()
+            val firstClassLoader = cache.getOrCreate(TestFileCollection(File("a/b/c")))
+            val secondClassLoader = cache.getOrCreate(TestFileCollection(File("c/b/a")))
 
-            assertThat(changed).isTrue()
+            assertThat(firstClassLoader !== secondClassLoader)
         }
 
-        it("reports different files") {
-            val changed = hasClasspathChanged(
-                setOf(DifferentDateFile("/c/b/a")),
-                setOf(DifferentDateFile("/a/b/c"))
-            )
+        it("same classloader for the same files in different order") {
+            val cache = DefaultClassLoaderCache()
+            val firstClassLoader = cache.getOrCreate(TestFileCollection(File("a/b/c"), File("d/e/f")))
+            val secondClassLoader = cache.getOrCreate(TestFileCollection(File("d/e/f"), File("a/b/c")))
 
-            assertThat(changed).isTrue()
-        }
-
-        it("concurrent blocking file resolution does not deadlock") {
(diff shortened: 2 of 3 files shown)


chao2zhang (commented):
Thank you for looking into the long-standing issue - I am still curious about how the existing implementation caused crash, because the existing implementation is inefficient in waiting for lock release but should still be correct - Return a URLClassloader for the given classpath files.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
```diff
@@ -12,30 +12,18 @@ fun interface ClassLoaderCache {
...
+            )
         }
-    }
-
-    private fun cache(classpathFiles: Set<File>) {
-        loaderAndClasspathFiles = URLClassLoader(
-            classpathFiles.map { it.toURI().toURL() }.toTypedArray(),
-            null /* isolate detekt environment */
-        ) to classpathFiles
+        return classpathFilesWithLoaders[classpathFiles]!!
```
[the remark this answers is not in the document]
sschuberth: Or `getOrPut`? As its implementation has an interesting comment:

```
public inline fun <K, V> ConcurrentMap<K, V>.getOrPut(key: K, defaultValue: () -> V): V {
    // Do not use computeIfAbsent on JVM8 as it would change locking behavior
    return this.get(key)
            ?: defaultValue().let { default -> this.putIfAbsent(key, default) ?: default }

}
```

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
```diff
@@ -12,30 +12,18 @@ fun interface ClassLoaderCache {
 
 internal class DefaultClassLoaderCache : ClassLoaderCache {
 
-    private var loaderAndClasspathFiles: Pair<URLClassLoader, Set<File>>? = null
+    private var classpathFilesWithLoaders: ConcurrentHashMap<Set<File>, URLClassLoader> = ConcurrentHashMap()
```
BraisGabin, line 15: This can be `val` now, right?

BraisGabin (commented):
Should we cache all the URLClassLoader instead of just the last one? Aren't we creating a memory leak here?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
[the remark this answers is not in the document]
Dominick1993: We can certainly use `getOrPut()` with lambda, I essentially reimplemented it here. Thanks for the suggestions guys, in my defense, I was half-asleep while submitting this PR, it was really early in the morning :)

Dominick1993:
> Should we cache all the URLClassLoader instead of just the last one?

I think that it is safe to store multiples of them as it allows for greater re-usability.

> Aren't we creating a memory leak here?

This is a point I am not quite sure about. And I think there are 2 points to consider:
- Are the classloaders properly closed when all detekt tasks finish?
- Should we limit the size of the (hash)map? Is it possible that someone might run it with hundreds of different classpaths so the size might become problematic?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
```diff
@@ -12,30 +12,18 @@ fun interface ClassLoaderCache {
...
-
-            if (lastClasspathFiles == null) {
-                cache(classpathFiles)
-            } else if (hasClasspathChanged(lastClasspathFiles, classpathFiles)) {
-                DefaultGroovyMethodsSupport.closeQuietly(lastLoader)
-                cache(classpathFiles)
-            }
-
-            return loaderAndClasspathFiles?.first ?: error("Cached or newly created detekt classloader expected.")
+        val hasSuitableClassLoader = classpathFilesWithLoaders.containsKey(classpathFiles)
```
cortinico, line 19: nit: I'm unsure that using a `Set<File>` as key for the `ConcurrentHashMap` is the best idea. Maybe we can use `classpath.asPath` that is a `String` instead?

cortinico (commented):
That's great 🎉 Thanks for taking the time to fix this.
Also don't forget to add yourself to the contributors list in the README

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
cortinico: Yeah also. I'm also unsure if the order of the `File` in the `FileCollection` is relevant or we can ignore it for our specific case.

Dominick1993: IMO the lookup of existing key should be file order agnostic, so we could use something like `HashSet<File>` constructed from `classpath: FileCollection` and use its `.hashCode()` as `HashMap` keys. `HashSet<File>.hashCode()` would trigger `File.hashCode()` on its content, but it should be fairly quick, since the `File` hash code is returned by the Filesystem (OS).

That's almost what it does in the current state as `HashMap<Set<File>, _>`, except for the order agnosticism (which could be done simply by `HashMap<HashSet<File>, _>`) and not storing the whole `File` objects as keys, which is really unnecessary in our case.

Regarding the MD5, I think that hash code should be a sufficient key, or am I missing something?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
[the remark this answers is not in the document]
chao2zhang: .getOrPut may invoke the `defaultValue` multiple times since it does not have any locking. Yeah I have noticed this interesting comment - It may indicate that JVM 8 locks the entire map while the latest JVMs can perform more granular locking on each entry.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
Dominick1993: I've changed the implementation in 21088b99659a470b37a1fc2e928fab21305be184 to match the one I proposed above. I got the same execution times as I did with the original one.

Dominick1993:
> * Are the classloaders properly closed when all detekt tasks finish?

Seems like once all classes loaded by the classloader and the classloader itself are no longer referenced, the classloader is closed and can be garbage collected. Therefore I assume this shoudn't be a problem, once the detekt task finishes. Please correct me if I am wrong.

> * Should we limit the size of the (hash)map? Is it possible that someone might run it with hundreds of different classpaths so the size might become problematic?

This probably shouldn't be an issue and if so OOM on Gradle JVM would be fairly easy to overcome / fix.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
```diff
@@ -12,31 +13,18 @@ fun interface ClassLoaderCache {
 
 internal class DefaultClassLoaderCache : ClassLoaderCache {
 
-    private var loaderAndClasspathFiles: Pair<URLClassLoader, Set<File>>? = null
+    private val classpathFilesHashWithLoaders: ConcurrentMap<Int, URLClassLoader> = ConcurrentHashMap()
```
sschuberth, line 16: Nit: I guess you could simplify this declaration a bit to

    private val classpathFilesHashWithLoaders = ConcurrentHashMap<Int, URLClassLoader>()

And this gets also rid of an import statement.

Dominick1993: I have the habit of explicitly providing types for class properties in Kotlin instead of relying on type inference, which I generally rely on in functions. However I don't mind changing it ;)

Dominick1993:
I've squashed the commits and rebased onto latest state of `main` branch.

cortinico (approved):

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCache.kt
```diff
@@ -12,31 +12,18 @@ fun interface ClassLoaderCache {
...
 
     override fun getOrCreate(classpath: FileCollection): URLClassLoader {
         val classpathFiles = classpath.files
-        synchronized(this) {
-            val lastLoader = loaderAndClasspathFiles?.first
-            val lastClasspathFiles = loaderAndClasspathFiles?.second
-
-            if (lastClasspathFiles == null) {
-                cache(classpathFiles)
-            } else if (hasClasspathChanged(lastClasspathFiles, classpathFiles)) {
```
chao2zhang, line 25: Could we remove `hasClasspathChanged()` method along with its tests?

Dominick1993: Good catch @chao2zhang, thanks! I've tried one better and re-wrote them. As naive as they are, they at least capture the essence of the current behavior.

Dominick1993: Good catch @chao2zhang, thanks! I've tried one better and re-wrote them. As naive as they are, they at least capture the essence of the current behavior.

chao2zhang (approved):

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCacheSpec.kt
```diff
@@ -11,53 +11,40 @@ import java.util.concurrent.CountDownLatch
...
-                setOf(DifferentDateFile("/a/b/c"), DifferentDateFile("/c/b/a")),
-                setOf(DifferentDateFile("/a/b/c"))
-            )
+        it("same classloader is returned for the same files") {
+            val cache = DefaultClassLoaderCache()
+            val initialClassLoader = cache.getOrCreate(TestFileCollection(File("a/b/c")))
+            val secondClassLoader = cache.getOrCreate(TestFileCollection(File("a/b/c")))
 
-            assertThat(changed).isTrue()
+            assertThat(initialClassLoader).isEqualTo(secondClassLoader)
```
BraisGabin, line 24: Should we check `==` or should we check `===`?

BraisGabin (approved):
Thanks for all the research!

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/internal/ClassLoaderCacheSpec.kt
chao2zhang: In theory, this should be equivalent, because `URLClassLoader` does not implement `equals()`. Semantically I believe `===` is better.

cortinico:
Can we merge this? So we can collect some early feedbacks from the latest snapshot?

chao2zhang:
I was actually waiting for all checks to pass before merging.

PR #3638, opened 2021-03-31
Kotlin 1.5.0

3flex:
Smoke test Kotlin 1.5 with new IR backend which is used by default.

Todo:
- [x] update Gradle config now that Java 1.8 bytecode is generated by default
- [ ] Update to Kotlin 1.5.0 final
- [ ] Update languageVersion to "1.5" (or should this be held back?)

PR #3800, opened 2021-05-20
Prepare Detekt 1.17.1

cortinico:
This PR contains the release branch for Detekt `1.17.1`.
I'm pushing the `release/1.17.1` branch to detekt/detekt so others can track it as well.

The CI will fail as the artifacts haven't been published yet.
I'm waiting for a positive review before publishing the assets 👍 

I've included #3792 #3793 #3794 only as discussed

The branch was created from: `git checkout -b release/1.17.1 6a92af522c67b9f3aa9d5965423f00009b2a6cb6`

### DIFF
--- a/buildSrc/src/main/kotlin/Versions.kt
+++ b/buildSrc/src/main/kotlin/Versions.kt
@@ -1,6 +1,6 @@
 object Versions {
 
-    const val DETEKT: String = "1.17.0"
+    const val DETEKT: String = "1.17.1"
     const val SNAPSHOT_NAME: String = "main"
     const val JVM_TARGET: String = "1.8"
 
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -106,6 +106,8 @@ class DetektPlugin : Plugin<Project> {
         const val BASELINE_TASK_NAME = "detektBaseline"
         const val DETEKT_EXTENSION = "detekt"
         private const val GENERATE_CONFIG = "detektGenerateConfig"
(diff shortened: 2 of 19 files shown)


chao2zhang:
Rebase and merge option was enabled. Please let me know if this is the most suitable option for this PR.

chao2zhang (approved):

cortinico:
> Please let me know if this is the most suitable option for this PR.

I would actually opt for creating a merge commit. Mostly because the resulting diff should be equivalent to 40c3330

chao2zhang:
👍  Yes it looks like some of the bugfix commits we want to merge is already on `main`. Merge looks simpler.

BraisGabin:
Careful, we shouldn't merge main here until we release 1.17.1, right?

cortinico:
> Careful, we shouldn't merge main here until we release 1.17.1, right?

Agree. I've reverted @chao2zhang merge of `main` into this branch.
The release of `1.17.1` was successful yesterday. I'll create a tag/release today.

I see two approaches for this PR/branch here:
1. We merge it with a merge commit on `main`. We would have to resolve a conflict introduced by #3638 but that shouldn't be a big deal.
1. We close it and we keep the `release/1.17.1` branch alive. We will have to cherry-pick 40c3330 on `main` as it contains the release notes etc.

chao2zhang:
My bad. I used "Resolve conflicts" on the Github UI.

In my work, approach #2 is the orthodox approach but I also believe #1 should work (Since we are manually editing release notes for each release, and thank you @cortinico for doing a great job).

cortinico:
So I've merged the branch manually and tagged `v1.17.1` on https://github.com/detekt/detekt/commit/40c3330859d13aba6b92b88acc616ef4266c486d

BraisGabin:
I always use the number one. And we never squash the tagged version so it is always in main. I think that we don't need to keep any branch open. The tags should be enough to recover in case we need it.

PR #3803, opened 2021-05-20
Add a page with our compatibility table

cortinico:
Adding a short table with the dependencies we used to build the Gradle Plugin.


docs/pages/compatibility.md
```diff
@@ -0,0 +1,28 @@
...
+sidebar: 
+permalink: compatibility.html
+summary: This page lists the version of the Gradle plugins have been used to build Detekt.
+---
+
+When shipping the Detekt Gradle Plugin, we depend on both the **Kotlin Gradle Plugin** and the **Android Gradle Plugin**.
+
+Those dependencies are applied as `compileOnly` ([see here](https://github.com/detekt/detekt/blob/75622d3ba88b0ae0357aec5f2d82a55aa6c6d157/detekt-gradle-plugin/build.gradle.kts#L17-L18)) to allow our users to pick the version of the Gradle plugin they prefer and don't impose the one we use inside detekt.
+
+We try to support b**ackward compatibility** when possible, although that's not always trivial (especially with AGP or across minor versions of Kotlin).
```
schalkms, line 13:
```suggestion
We try to support **backward compatibility** when possible, although that's not always trivial (especially with AGP or across minor versions of Kotlin).
```

docs/pages/compatibility.md
```diff
@@ -0,0 +1,28 @@
...
+
+This table lists the version of the Gradle plugin we used to compile the Detekt Gradle plugin.
```
schalkms, line 15:
```suggestion
The following table lists the version of the Gradle plugin we used to compile the Detekt Gradle plugin. 
```

docs/pages/gettingstarted/gradle.md
```diff
@@ -10,7 +10,7 @@ redirect_from:
 summary:
 ---
 
-detekt requires Gradle 5.4 or higher.
+Detekt requires **Gradle 5.4** or higher. We however recommend using the version of Gradle that is [listed in this table](https://detekt.github.io/detekt/compatibility.html).
```
schalkms, line 13:
```suggestion
Detekt requires **Gradle 5.4** or higher. We, however, recommend using the version of Gradle that is [listed in this table](https://detekt.github.io/detekt/compatibility.html).
```

BraisGabin (approved):
<3 nice!!

3flex:
This is great! I think the compatibility table should be in the README though, or at least details from the top line of the table showing details for the current version.

For consideration:
1. It sounds like we're declaring support for older detekt versions. Personally I wouldn't want to commit to supporting anything except current detekt (or RCs).
1. I think the project should explicitly support "current minus one major version" for the key dependencies of Gradle and Kotlin - but for Kotlin, only the most recent versions in the series. For example today that would mean Gradle 6.0-7.0.2, and Kotlin 1.5 and 1.4. Ideally we'd do this for Android Gradle Plugin as well but that might not be feasible (until we're ready to recommend the compiler plugin which should address this). I'm working on integration tests that would allow this to be testable, so we could hold off on making that declaration until we have tests.
2. Add info about JDK - we test on all currently supported versions (or will soon, once Java 16 tests pass) so we should declare this in our docs.

cortinico:
> This is great! I think the compatibility table should be in the README though, or at least details from the top line of the table showing details for the current version.

Added 👍 

>     1. It sounds like we're declaring support for older detekt versions. Personally I wouldn't want to commit to supporting anything except current detekt (or RCs).

I've added a short paragraph where I mentioned that.
 
>     2. I think the project should explicitly support "current minus one major version" for the key dependencies of Gradle and Kotlin - but for Kotlin, only the most recent versions in the series. For example today that would mean Gradle 6.0-7.0.2, and Kotlin 1.5 and 1.4. Ideally we'd do this for Android Gradle Plugin as well but that might not be feasible (until we're ready to recommend the compiler plugin which should address this). I'm working on integration tests that would allow this to be testable, so we could hold off on making that declaration until we have tests.

Agree. Also "current minus one major version" it's really trick as every tool has his own versioning scheme :/

>     3. Add info about JDK - we test on all currently supported versions (or will soon, once Java 16 tests pass) so we should declare this in our docs.

I've added Java min/max as well

3flex (approved):
💯

PR #3810, opened 2021-05-21
Update the implementation of ClassOrdering to handle false negatives

### LINKED ISSUE #3809: False negative for `ClassOrdering`
## Expected Behavior
For `ClassOrdering`, code like the following should report violations.
```
class MultipleMisorders(private val x: String) {
    companion object {
        const val IMPORTANT_VALUE = 3
    }
    
    class Nested { }
    
    fun returnX() = x
    
    class Nested2 { }
    
    constructor(z: Int): this(z.toString())
    
    class Nested3 { }
    
    val y = x
}
```

## Observed Behavior
There are no violations reported.

## Steps to Reproduce
Use the provided code snippet.

## Context
There were many violations detected in my company's projects when upgraded to 1.17, however, there are still false negatives remaining in the codebase.

## Your Environment
* Version of detekt used: 1.17

chao2zhang:
Resolves #3809.

With this new implementation, one noticeable changes are that the number of violations is going to be increased:
For example, in a class layout like

```
class MyClass {
    val a: Int
    fun b(): String
    val c: Int
    val d: Int
    val e: Int
    val f: Int
}
```
Previous the violation will be reported only the boundary (a <-> b, b <-> c), now it will be reported on (c, d, e, f).
If the increase in the violations is a concern, we can implement grouping logic to reduce the number of violations.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
@@ -61,59 +61,65 @@ class ClassOrdering(config: Config = Config.empty) : Rule(config) {
         Debt.FIVE_MINS
     )
 
-    private val comparator: Comparator<KtDeclaration> = Comparator { dec1: KtDeclaration, dec2: KtDeclaration ->
-        if (dec1.priority == null || dec2.priority == null) return@Comparator 0
-        compareValues(dec1.priority, dec2.priority)
-    }
-
     override fun visitClassBody(classBody: KtClassBody) {
         super.visitClassBody(classBody)
 
-        val misorders = comparator.findOutOfOrder(classBody.declarations)
-        if (misorders.isNotEmpty()) {
-            report(
-                misorders.map {
-                    CodeSmell(
-                        issue = issue,
-                        entity = Entity.from(it.first),
-                        message = "${it.first.description} should not come before ${it.second.description}",
-                        references = listOf(Entity.from(classBody))
+        var currentSection = Section(0)
+        classBody.declarations.forEach { ktDeclaration ->
+            val section = ktDeclaration.toSection()
+            when {
+                section != null && section < currentSection -> {
+                    val message =
+                        "${ktDeclaration.toDescription()} should be declared before ${currentSection.toDescription()}."
+                    report(
+                        CodeSmell(
+                            issue = issue,
+                            entity = Entity.from(ktDeclaration),
+                            message = message,
+                            references = listOf(Entity.from(classBody))
+                        )
                     )
                 }
-            )
+                section != null && section > currentSection -> currentSection = section
+            }
         }
     }
 }
 
-private fun Comparator<KtDeclaration>.findOutOfOrder(
-    declarations: List<KtDeclaration>
-): List<Pair<KtDeclaration, KtDeclaration>> =
-    declarations
-        .zipWithNext { a, b -> if (compare(a, b) > 0) Pair(a, b) else null }
-        .filterNotNull()
+private fun KtDeclaration.toDescription(): String = when {
+    this is KtProperty -> "property `$name`"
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
```diff
@@ -61,59 +61,65 @@ class ClassOrdering(config: Config = Config.empty) : Rule(config) {
...
+    this is KtProperty -> Section(0)
+    this is KtClassInitializer -> Section(0)
+    this is KtSecondaryConstructor -> Section(1)
+    this is KtNamedFunction -> Section(2)
+    this is KtObjectDeclaration && isCompanion() -> Section(3)
+    else -> null
+}
+
+@JvmInline
+@Suppress("MagicNumber", "ModifierOrder") // TODO Remove ModifierOrder once value class is supported.
```
chao2zhang, line 109: This is pending #3719

BraisGabin:
Time ago I did some research regarging this rule and the number of issue that it raises. I found that we could use ["longest increasing subsequence (LIS)"](https://www.geeksforgeeks.org/minimum-insertions-sort-array/) to reduce the number of issues. But I never implement it. I don't know even if we need to implement it. But as you talk about it I thought it could be usefull.

BraisGabin (approved):

cortinico (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
```diff
@@ -61,59 +61,65 @@ class ClassOrdering(config: Config = Config.empty) : Rule(config) {
...
+                        CodeSmell(
+                            issue = issue,
+                            entity = Entity.from(ktDeclaration),
+                            message = message,
+                            references = listOf(Entity.from(classBody))
+                        )
                     )
                 }
-            )
+                section != null && section > currentSection -> currentSection = section
```
schalkms, line 83: Can we move out the `section != null` check in order to not check it twice?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
```diff
@@ -61,59 +61,65 @@ class ClassOrdering(config: Config = Config.empty) : Rule(config) {
...
+}
+
+@Suppress("MagicNumber")
+private fun KtDeclaration.toSection(): Section? = when {
+    this is KtProperty -> Section(0)
+    this is KtClassInitializer -> Section(0)
+    this is KtSecondaryConstructor -> Section(1)
+    this is KtNamedFunction -> Section(2)
+    this is KtObjectDeclaration && isCompanion() -> Section(3)
+    else -> null
```
schalkms, line 105: Is there any alternative to the `null` return? Then we could spare the null checks above?

schalkms (commented):
Thanks for the fix!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
chao2zhang: Kotlin documentation does not specify where we should put nested classes or objects, so the ordering for nested classes shouldn't matter from a static analysis point of view, hence it should not have any priority order - `null` is a reasonable choice.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
chao2zhang: Refactored using `continue` in a `for loop`.

PR #3811, opened 2021-05-22
[RFC] Autocorrection on native rules

chao2zhang:
## Motivation
When upgraded to Detekt 1.17 across the repo in my company, we have encountered numerous violations of `ClassOrdering` which requires manual fixing. This takes much more effort and should better be automated.

This is probably the first Detekt native rule that is auto-correctible, therefore I am posting this as a draft to collect feedback.

## Implementation

This is built on top of https://github.com/detekt/detekt/issues/3809 and should be merged.

Mutating the PSI is not as easy as it sounds according to the [IntelliJ documentation](https://plugins.jetbrains.com/docs/intellij/modifying-psi.html). This is probably why ktlint opts to mutate on ASTNode instead of PsiElement. However, in this PR, I still managed to perform modifications using only the PSI.

One tricky part is handling braces and whitespace, since `KtClassBody.declarations` and `KtClassBody.children` does not include those leaf elements.

Another tricky part is that adding PSIElement to a parent node requires its parent to be present, otherwise it will produce NPE. This is why I am copying a node and moving the children to the new node to circumvent this problem.

## Expectation
Ultimately, Detekt is a static analysis tool and is not a perfect style formatted. It is safe to assume that the autocorrected code can still violate other rules. The only guarantee which is verified by the test in this PR is that once auto-corrected, the same rule should detect no findings.

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtFileModifier.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtFileModifier.kt
@@ -16,14 +16,14 @@ class KtFileModifier : FileProcessListener {
 
     override fun onFinish(files: List<KtFile>, result: Detektion, bindingContext: BindingContext) {
         files.filter { it.modificationStamp > 0 }
-            .map { it.absolutePath() to it.unnormalizeContent() }
-            .forEach {
-                result.add(ModificationNotification(it.first))
-                Files.write(it.first, it.second.toByteArray())
+            .map { it.absolutePath() to it.denormalizeContent() }
+            .forEach { (path, content) ->
+                result.add(ModificationNotification(path))
+                Files.write(path, content.toByteArray())
             }
     }
 
-    private fun KtFile.unnormalizeContent(): String {
+    private fun KtFile.denormalizeContent(): String {
         val lineSeparator = getUserData(LINE_SEPARATOR)
         require(lineSeparator != null) {
             "No line separator entry for ktFile ${javaFileFacadeFqName.asString()}"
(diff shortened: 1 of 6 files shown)


cortinico:
I really like the opportunities and the PR looks good to me overall. There are a couple of things I'd like to discuss/address:

1. Historically Detekt was always considered a "read-only" tool (modulo the ktlint wrapper). If we shift this, we should clearly state this in the release notes/websites, etc. Also all the usages of the `autoCorrect` property on the CLI/Gradle API should be updated.

2. I'm a bit afraid this will increase complexity of some rules. Specifically while we have a lot of util functions for inspecting the AST, we don't have much support tools for editing it. I'm not saying that we should build all the utilities now, but we should be mindful when adding autoCorrection capabilities to a rule.

3. Editing the source code is a potentially dangerous operation. How do we prevent that we don't "lose" user's code? 

4. How this is going to play with the compilation lifecycle? I suppose the code should be recompiled after the autoCorrection (unless we run before). How does this will interact with the `detekt-gradle-plugin`?

BraisGabin:
I don't know how I feel about this. Autocorrect is great but it is dangerous.

🤔 Maybe, as detekt, we should just create a `.patch` for each fix that we want to do. This will give us more flexibility. Those patch could be applied later in our gradle plugin so for the user it works exactly the same way. But, other tools could read those patchs and create comments in your PR proposing the change. This way we just "suggest" changes but we don't mess with our user's code. Is the user who decide to apply those patches.

I did a proof of concept time ago about that tool that I'm talking about: https://github.com/BraisGabin/ktlint-bot-playground/pull/1#pullrequestreview-387858301

chao2zhang:
Would we recommend auto-correction to limited to ktlint then? ClassOrdering could have been a rule in ktlint but it arrived as a Detekt native rule first and resulted in multiple violations in my codebase. (Given the current activity on ktlint, I don't think it will land too soon).

The other aspect of moving to the compiler plugin is that compilation should not modify the source code. The path idea could be great, but we could also leverage detekt intellij plugin for IDE auto fix suggestions?

BraisGabin:
> The path idea could be great, but we could also leverage detekt intellij plugin for IDE auto fix suggestions?

I suppose that you mean patch, right? I'm not sure how the IntelliJ API is but I think that we should be able to integrate those patchs with the plugin too.

3flex:
> moving to the compiler plugin is that compilation should not modify the source code

Agreed. This is the best argument for not adding autocorrect to existing rules, even if the compiler plugin will not be stable potentially for a long time (until Kotlin has a stable compiler plugin API at least).

PR #3812, opened 2021-05-22
Update default-detekt-config.yml to match generated output

marschwar:
The detekt-config-default.yml that is generated by the gradle task differed from the one in the project. I vaguely remember that the CI used to fail in this case but I could be wrong about that.


cortinico:
This effectively reverts #3801 
I suggests we instead merge this other #3815

PR #3813, opened 2021-05-22
Fix generation issues around (deprecated) list properties

### LINKED ISSUE #3806: Inconsistent use of deprecation information in default-detekt-config.yaml
While looking into #3671 I came across something that I thought was a little inconsistent but was not sure if it is intended or not.

When the `default-detekt-config.yml` is generated we do not include simple deprecated config options. That seems reasonable to me. But we do include those config properties that allow the user to specify a list (at the moment there are none). For rule sets no all config options are always included.

Is there a reason behind all this or should it be _hide when deprecated_?

marschwar:
This fixes #3806 plus a few more minor issues along the way

- deprecated ruleset config options are ignored for config file generation
- fix output of deprecated default value when it is a list
- omit deprecated list property from config
- fix indentation of yml list to 2 spaces

### DIFF
--- a/detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/util/RuleSetPageCreator.kt
+++ b/detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/util/RuleSetPageCreator.kt
@@ -26,6 +26,18 @@ internal fun createRuleSetPage(): RuleSetPage {
                     name = "rulesetconfig2",
                     description = "description rulesetconfig2",
                     defaultValue = "['foo', 'bar']",
+                    deprecated = null
+                ),
+                Configuration(
+                    name = "deprecatedSimpleConfig",
+                    description = "description deprecatedSimpleConfig",
+                    defaultValue = "true",
+                    deprecated = "is deprecated"
+                ),
+                Configuration(
+                    name = "deprecatedListConfig",
+                    description = "description deprecatedListConfig",
+                    defaultValue = "['foo', 'bar']",
                     deprecated = "is deprecated"
                 )
             )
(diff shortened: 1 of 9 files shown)


marschwar:
This will have to be rebased and updated after #3815 is merged.

detekt-generator/src/test/resources/RuleSet.md
```diff
@@ -22,10 +22,16 @@ a wildcard import
 
   deprecated config
 
-* ``conf3`` (default: ``['a', 'b']``)
+* ``conf3`` (default: ``a, b``)
 
   list config
 
+* ~~``conf4``~~ (default: ``a, b``)
```
BraisGabin, line 29: 🤔 I don't think we should change this. `['a', 'b']` this is clearly a list `a, b` isn't that obvious.

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/util/RuleSetPageCreator.kt
```diff
@@ -26,6 +26,18 @@ internal fun createRuleSetPage(): RuleSetPage {
...
+                Configuration(
+                    name = "deprecatedSimpleConfig",
+                    description = "description deprecatedSimpleConfig",
+                    defaultValue = "true",
+                    deprecated = "is deprecated"
+                ),
+                Configuration(
+                    name = "deprecatedListConfig",
+                    description = "description deprecatedListConfig",
+                    defaultValue = "['foo', 'bar']",
```
cortinico, line 40: Should you assert inside `ConfigPrinterSpec` that those two keys are not printed in the default config?

schalkms (approved):

detekt-generator/src/test/resources/RuleSet.md
marschwar: Good catch. That was not intended.

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/util/RuleSetPageCreator.kt
marschwar: Since the entire `RuleSetConfig.yml` is matched it actually is not necessary, but I will add it to make it more explicit

cortinico (approved):

BraisGabin (approved):

PR #3817, opened 2021-05-22
Improve cli documentation

BraisGabin:
Probably we should not be exhaustive here. But this at least is not missleading.

### DIFF
--- a/detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/spec/CompilerSpec.kt
+++ b/detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/spec/CompilerSpec.kt
@@ -11,7 +11,7 @@ interface CompilerSpec {
     val jvmTarget: String?
 
     /**
-     * Kotlin language version (e.g. 1.0, 1.1, 1.2, 1.3, 1.4, ...).
+     * Kotlin language version (e.g. 1.0, 1.1, 1.2, 1.3 ...).
      */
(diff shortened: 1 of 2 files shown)


schalkms (approved):

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/spec/CompilerSpec.kt
```diff
@@ -11,7 +11,7 @@ interface CompilerSpec {
     val jvmTarget: String?
 
     /**
-     * Kotlin language version (e.g. 1.0, 1.1, 1.2, 1.3, 1.4, ...).
+     * Kotlin language version (e.g. 1.0, 1.1, 1.2, 1.3 ...).
```
cortinico, line 14: Shouldn't you add `1.5` instead of removing `1.4`?

BraisGabin: I prefer to not be exhaustive and only have 4 samples as the case above it. Otherwise we need to update this every time.

chao2zhang: `detekt-tooling` is an internal module, so a comment identifying the format of `languageVersion` should suffice.

chao2zhang (approved):

PR #3818, opened 2021-05-22
jvmTarget can't be null

BraisGabin:
There is not need to define this property as nullable.


schalkms (approved):

schalkms:
According to the error message, tests need to be adapted.

`D:\a\detekt\detekt\detekt-tooling\src\main\kotlin\io\github\detekt\tooling\dsl\CompilerSpecBuilder.kt: (16, 29): Type of 'jvmTarget' is not a subtype of the overridden property 'public abstract val jvmTarget: String defined in io.github.detekt.tooling.api.spec.CompilerSpec' `

cortinico (approved):

Issue #3563, opened 2021-03-17
Configuration cache error: cannot serialize object of type 'java.net.URLClassLoader'


yogurtearl:
Detekt is not compatible with config cache: 
https://docs.gradle.org/current/userguide/configuration_cache.html

## Expected Behavior
No config cache errors/warnings

## Observed Behavior

```
2 problems were found storing the configuration cache.
- Task `:app:detekt` of type `io.gitlab.arturbosch.detekt.Detekt`: cannot serialize object of type 'java.net.URLClassLoader', a subtype of 'java.lang.ClassLoader', as these are not supported with the configuration cache.
  See https://docs.gradle.org/6.8.3/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
- Task `:detekt` of type `io.gitlab.arturbosch.detekt.Detekt`: cannot serialize object of type 'java.net.URLClassLoader', a subtype of 'java.lang.ClassLoader', as these are not supported with the configuration cache.
  See https://docs.gradle.org/6.8.3/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
```
with this path: 

```⚠️cannot serialize object of type java.net.URLClassLoader📋, a subtype of java.lang.ClassLoader📋, as these are not supported with the configuration cache. ?
⌄fieldfirst📋 of kotlin.Pair📋
⌄bean of type kotlin.Pair📋
⌄fieldloaderAndClasspathFiles📋 of io.gitlab.arturbosch.detekt.internal.DefaultClassLoaderCache📋
⌄bean of type io.gitlab.arturbosch.detekt.internal.DefaultClassLoaderCache📋
⌄field$$delegate_0📋 of io.gitlab.arturbosch.detekt.internal.GlobalClassLoaderCache📋
⌄bean of type io.gitlab.arturbosch.detekt.internal.GlobalClassLoaderCache📋
⌄fieldclassLoaderCache📋 of io.gitlab.arturbosch.detekt.invoke.DefaultCliInvoker📋
⌄bean of type io.gitlab.arturbosch.detekt.invoke.DefaultCliInvoker📋
⌄fieldinvoker📋 of io.gitlab.arturbosch.detekt.Detekt📋
■task:app:detekt📋 of type io.gitlab.arturbosch.detekt.Detekt📋
■task:detekt📋 of type io.gitlab.arturbosch.detekt.Detekt📋
```

## Steps to Reproduce
- Create a blank kotlin project 
- add detekt 
[...]

PR #3822, opened 2021-05-23
Plugin: Fix Gradle configuration cache by not keeping classloader as a nested task field

mateuszkwiecinski:
Fixes https://github.com/detekt/detekt/issues/3563

I tried to reproduce the issue in tests, however I'm unable to trigger the `cannot serialize object of type 'java.net.URLClassLoader', a subtype of 'java.lang.ClassLoader'` error there :/ I noticed tests are passing the `dry-run` flag which calls simpler `DryRunInvoker`, but even after adding a test case with the `DefaultCliInvoker` the test still doesn't fail.  
I'd love to hear some ideas how to cover the linked scenario in tests 👀 

I verified manually on a failing project changes I'm suggesting work 🤷

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
@@ -3,7 +3,7 @@ package io.gitlab.arturbosch.detekt.invoke
 import io.gitlab.arturbosch.detekt.internal.ClassLoaderCache
 import io.gitlab.arturbosch.detekt.internal.GlobalClassLoaderCache
 import org.gradle.api.GradleException
-import org.gradle.api.Project
+import org.gradle.api.Task
 import org.gradle.api.file.FileCollection
 import org.gradle.api.logging.Logger
 import java.io.PrintStream
@@ -21,14 +21,14 @@ internal interface DetektInvoker {
     companion object {
         private const val DRY_RUN_PROPERTY = "detekt-dry-run"
 
-        fun create(project: Project): DetektInvoker =
-            if (project.isDryRunEnabled()) {
-                DryRunInvoker(project.logger)
+        fun create(task: Task): DetektInvoker =
+            if (task.isDryRunEnabled()) {
+                DryRunInvoker(task.logger)
             } else {
                 DefaultCliInvoker()
             }
 
-        private fun Project.isDryRunEnabled(): Boolean {
+        private fun Task.isDryRunEnabled(): Boolean {
             return hasProperty(DRY_RUN_PROPERTY) && property(DRY_RUN_PROPERTY) == "true"
         }
     }
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
@@ -2,7 +2,6 @@ package io.gitlab.arturbosch.detekt.invoke
 
 import io.gitlab.arturbosch.detekt.internal.ClassLoaderCache
 import io.gitlab.arturbosch.detekt.internal.GlobalClassLoaderCache
-import org.codehaus.groovy.runtime.DefaultGroovyMethods.hasProperty
 import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.Task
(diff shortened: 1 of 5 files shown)


chao2zhang:
We would need to use `DslGradleRunner` which leverages the [TestKit](https://docs.gradle.org/current/userguide/test_kit.html) to pass in the additional `--configuration-cache`.

The change looks good to me though.

mateuszkwiecinski:
The test you mentioned is already there: https://github.com/detekt/detekt/blob/4a89a25c82d35874571270fa1ff32751f2f268a1/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/ConfigurationCacheTest.kt#L8

I expected that removing the `dryRun()` call there, it'd invoke `DefaultCliInvoker` which would reveal the serialization issue, but that doesn't happen. Basically, I can't make the test to fail with linked issue on current `main` branch (I just pushed an updated test which I expected to fail)

chao2zhang:
When inspecting the task output in the aforementioned test, I can even observe `Stored cache entry for task ':detekt' with cache key beef5daf9ca25f7b5c981f92fb0b68d0`, which makes me think if the configuration cache works for write. And test passes mean configuration cache works for read.

On second thought, we could use a real detekt task instead of dry run to verify the configuration cache behavior. We could create another issue for that.

mateuszkwiecinski:
Wouldn't the
>Stored cache entry for task

be a log for _regular_ Gradle Build Cache? For some reason detekt tests use build cache.  
AFAIU configuration cache prints different log messages, like "_Calculating task graph as no configuration cache is available for tasks"_ or _"Configuration cache entry stored"_ or _"Reusing configuration cache"_ (and actually that's what the tests makes assertions on)

>we could use a real detekt task instead of dry run to verify the configuration cache behavior. We could create another issue for that.

That's right, the _real_ detekt task, aka `DefaultCliInvoker` breaks Configuration Cache. That's what gradle report points at in the linked issue. And that's why I added a _real_ detekt task call to [the test](https://github.com/detekt/detekt/pull/3822/commits/438305710419717f13fc0d0286c54784151344ef#diff-cf3e1dea8c434fc3378845cc4058ad164d60610a26396784cd13167d03865710R11).

I expected it to fail without the fix I applied, but it doesn't - and that's my _challenge_ here :/

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
```diff
@@ -19,21 +21,23 @@ internal interface DetektInvoker {
...
 
     companion object {
-        private const val DRY_RUN_PROPERTY = "detekt-dry-run"
 
-        fun create(project: Project): DetektInvoker =
-            if (project.isDryRunEnabled()) {
-                DryRunInvoker(project.logger)
+        fun create(task: Task, isDryRun: Boolean): DetektInvoker =
+            if (isDryRun) {
+                task.logger.quiet("DUPA dryRun")
```
BraisGabin, line 27: Do this help us?

BraisGabin (approved):

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
mateuszkwiecinski: 😬 
No, not at all. I used it for debugging because passing `withDebug(true)` to the runner flag makes the tests to fail here 🤷 
Useless log has been removed ✅

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
```diff
@@ -2,8 +2,10 @@ package io.gitlab.arturbosch.detekt.invoke
 
 import io.gitlab.arturbosch.detekt.internal.ClassLoaderCache
 import io.gitlab.arturbosch.detekt.internal.GlobalClassLoaderCache
+import org.codehaus.groovy.runtime.DefaultGroovyMethods.hasProperty
```
cortinico, line 5: You should use `org.gradle.api.Project.hasProperty` instead.

mateuszkwiecinski: Good catch 👍 

This fortunately wasn't used: 
image
I must have added this by accident

chao2zhang (approved):

PR #3828, opened 2021-05-24
Fix false positive for UnusedPrivateMember - Backtick identifiers

### LINKED ISSUE #3825: False positive "UnusedPrivateMember" on local val `in` (with apostrophes)
## Expected Behavior
A local variable can be named `in` by using backticks.
This is for example interesting for code which is translated/converted from Java.
The name `in` is most likely not the issue of this bug.
The issue is most likely related to local variables being named with backticks in general.

## Observed Behavior
In the following code, the variable `in` was reported as `UnusedPrivateMember`:
```
    @Test
    fun testReadAllUBytes() {
        val `in` = ByteArrayInputStream(ubyteArrayOf("010203CAFEBABE")) as InputStream
        val expected = ubyteArrayOf("010203CAFEBABE")
        val actual = `in`.readAllUBytes()
        assertArrayEquals(expected, actual)
    }
```
Because `in` is clearly used, the expected behavior is that detekt does not report `UnusedPrivateMember` about `in`.
The actual behavior is that detekt does report `UnusedProvateMember` about `in`:
```
> Task :lib-nelkinda-core:detekt FAILED
style - 10min debt
        UnusedPrivateMember - [in] at /home/christian.hujer/git/nelkinda/xxx-xxx/lib-nelkinda-core/src/test/kotlin/com/nelkinda/java/io/InputStreamTest.kt:12:9
```

## Steps to Reproduce
Try the following source code:
```
    @Test
    fun testRead() {
        val `in` = StringReader("foo") as Reader
        val expected = "foo"
        val actual = `in`.readText()
        assertEquals(expected, actual)
    }
```

## Context
n/a

## Workaround
Use `@SuppressWarnings("UnusedPrviateMemger")` before the declaration of `in`.

## Your Environment

* Version of Gradle used (if applicable): 9.8.2
* Operating System and version: Kubuntu 20.04.2 LTS
* Link to your project (if it's a public repository): n/a

cortinico:
Fixes #3825

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
@@ -317,7 +317,7 @@ private class UnusedPropertyVisitor(allowedNames: Regex) : UnusedMemberVisitor(a
     private fun KtProperty.isMemberOrTopLevel() = isMember || isTopLevel
 
     override fun visitReferenceExpression(expression: KtReferenceExpression) {
-        nameAccesses.add(expression.text.removeSurrounding("`"))        
+        nameAccesses.add(expression.text.removeSurrounding("`"))
         super.visitReferenceExpression(expression)
     }
 }
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
@@ -317,7 +317,11 @@ private class UnusedPropertyVisitor(allowedNames: Regex) : UnusedMemberVisitor(a
     private fun KtProperty.isMemberOrTopLevel() = isMember || isTopLevel
 
     override fun visitReferenceExpression(expression: KtReferenceExpression) {
-        nameAccesses.add(expression.text)
+        if (expression.text.startsWith("`") && expression.text.endsWith("`")) {
+            nameAccesses.add(expression.text.removePrefix("`").removeSuffix("`"))
+        } else {
+            nameAccesses.add(expression.text)
+        }
         super.visitReferenceExpression(expression)
     }
 }
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
@@ -317,11 +317,7 @@ private class UnusedPropertyVisitor(allowedNames: Regex) : UnusedMemberVisitor(a
     private fun KtProperty.isMemberOrTopLevel() = isMember || isTopLevel
 
     override fun visitReferenceExpression(expression: KtReferenceExpression) {
-        if (expression.text.startsWith("`") && expression.text.endsWith("`")) {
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -317,7 +317,11 @@ private class UnusedPropertyVisitor(allowedNames: Regex) : UnusedMemberVisitor(a
     private fun KtProperty.isMemberOrTopLevel() = isMember || isTopLevel
 
     override fun visitReferenceExpression(expression: KtReferenceExpression) {
-        nameAccesses.add(expression.text)
+        if (expression.text.startsWith("`") && expression.text.endsWith("`")) {
+            nameAccesses.add(expression.text.removePrefix("`").removeSuffix("`"))
+        } else {
+            nameAccesses.add(expression.text)
+        }
```
marschwar, line 324:
```suggestion
        nameAccesses.add(expression.text.removeSurrounding("`"))        
```

I know I was wrong on how this function is used but in this particular case it really does what it is supposed to do.

marschwar, line 324:
```suggestion
        nameAccesses.add(expression.text.removeSurrounding("`"))        
```

I know I was wrong on how this function is used but in this particular case it really does what it is supposed to do.

BraisGabin (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
```diff
@@ -1214,4 +1214,40 @@ class UnusedPrivateMemberSpec : Spek({
             assertThat(subject.compileAndLintWithContext(env, code)).hasSize(2)
         }
     }
+
+    describe("backtick identifiers - #3825") {
```
schalkms, line 1218: Shall we add a positive test case as well?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
```diff
@@ -1214,4 +1214,40 @@ class UnusedPrivateMemberSpec : Spek({
...
+                fun main() {
+                    val actual = "fee"
+                    val expected = "foo"
+                    println(expected == `actual`)
+                }
+            """
+            assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
+        }
+
+         it("does not report unused variables when declared with backticks") {
```
schalkms, line 1242:
```suggestion
        it("does not report unused variables when declared with backticks") {
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
cortinico: TIL :O

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
cortinico: Yup added

chao2zhang (approved):

PR #3835, opened 2021-05-25
Re-enable warnings as errors for the gradle plugin module

eygraber:
Blocked by https://github.com/gradle/gradle/issues/16345

### DIFF
--- a/buildSrc/src/main/kotlin/module.gradle.kts
+++ b/buildSrc/src/main/kotlin/module.gradle.kts
@@ -90,12 +90,7 @@ tasks.withType<KotlinCompile>().configureEach {
             "-Xopt-in=kotlin.RequiresOptIn"
         )
         // Usage: <code>./gradlew build -PwarningsAsErrors=true</code>.
-        // Note: currently there are warnings for detekt-gradle-plugin that seemingly can't be fixed
-        //       until Gradle releases an update (https://github.com/gradle/gradle/issues/16345)
-        allWarningsAsErrors = when (project.name) {
-            "detekt-gradle-plugin" -> false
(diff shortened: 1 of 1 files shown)


BraisGabin (approved):

eygraber:
@BraisGabin @cortinico it looks like the only way to re-enable warnings as errors for the gradle plugin module is going to be to either:

1. Use the current Gradle's version of Kotlin just for the gradle plugin module (currently 1.4.31)
2. Wait for a new version of Gradle that supports 1.5, although that will change again in the future unless:
3. Gradle finds some way to isolate their version of Kotlin from a plugin's version of Kotlin

See https://github.com/gradle/gradle/issues/16345#issuecomment-861119625 for some more details

BraisGabin:
I think that 1 is not an option. I would keep this PR here until 2 or 3.

cortinico:
> I would keep this PR here until 2 or 3.

Agree. This looks like a long standing issue till Gradle comes up with a roadmap/decison on the Kotlin language version to bundle.

Tapchicoma:
Few cents from my side:
- Kotlin Gradle plugin automatically adds stdlib dependency if none is added by the user. I would advice you to add following line in `detekt-gradle-plugin/build.gradle.kts`:
```
compileOnly(kotlin("stdlib-jdk8"))
```
- You could set `kotlinOptions.languageLevel`/`kotlinOptions.apiLevel` to `1.3`, use Kotlin 1.5 and support Gradle versions with Kotlin 1.3 runtime.

3flex:
I've opened #5037 for tracking, so we can close this PR, as there have been no updates for nearly a year.

PR #3841, opened 2021-05-26
Fix UnnecessaryLet false positive in inner lambdas

xouabita:
When upgrading detekt to 1.7.1, I noticed some false positives for `UnnecessaryLet`.

For example:

```kotlin
fun callMe(callback: () -> Unit) {
    callback()
}
 
fun test(value: Int?) {
    value?.let { 
        callMe {
            println(it)
        }
     }
}
```

would be reported.

This PR should fix the issue

### DIFF
--- a/detekt-psi-utils/api/detekt-psi-utils.api
+++ b/detekt-psi-utils/api/detekt-psi-utils.api
@@ -88,6 +88,7 @@ public final class io/gitlab/arturbosch/detekt/rules/KtCallExpressionKt {
 }
 
 public final class io/gitlab/arturbosch/detekt/rules/KtLambdaExpressionKt {
+	public static final fun firstParameter (Lorg/jetbrains/kotlin/psi/KtLambdaExpression;Lorg/jetbrains/kotlin/resolve/BindingContext;)Lorg/jetbrains/kotlin/descriptors/ValueParameterDescriptor;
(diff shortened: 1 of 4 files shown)


BraisGabin:
That doesn't seem as a false positive. It could be replaced with an if.

xouabita:
@BraisGabin from [the docs](https://github.com/detekt/detekt/blob/main/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLet.kt#L40)

```kotlin
a?.let { print(it) }
```

is compliant

xouabita:
also,

```kotlin
fun callMe(callback: () -> Unit) {
    callback()
}
fun test(value: Int?) {
    value?.let { 
        callMe {
            it.plus(it)
        }
     }
}
```

will also be reported, which is a false positive

BraisGabin:
You are right. For me that's an unnecessary usage but this rule, right now, should not flag it.

cortinico (approved):

BraisGabin (approved):

xouabita:
Hi, thank you very much for reviewing this PR. Is there anything else I need to do before this PR can be merged ?

BraisGabin:
I don't think so. Sorry for the late merge and thanks for the fix :)

Issue #3844, opened 2021-05-28
Random NullPointerException


wkornewald:
We randomly get this exception, sometimes on module X, sometimes module Y, often it just works.

I'll try to reproduce this with --scan, but for now this is with --stacktrace. Do you need more information?

```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':xxx:detekt'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:187)
	at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:268)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:185)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:173)
	at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:109)
	at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
	at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:51)
	at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
	at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:56)
	at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:36)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:52)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:200)
[...]

PR #3848, opened 2021-06-01
Remove Extensions.getRootArea

### LINKED ISSUE #3683: Don't use Extensions.getRootArea()
We should remove the usages of `Extensions.getRootArea()` in our code. That function is deprecated.

Related with #3646

cortinico:
`Extensions.getRootArea` is deprecated and should be replaced with `ComponentManager.getExtensionArea()`

Moreover this will potentially address #3844

Fixes #3683


BraisGabin (approved):

chao2zhang (approved):
Thank you for doing this. I was about to take a look yesterday but I am glad that this is addressed already

PR #3852, opened 2021-06-03
Set apiVersion to 1.4

### LINKED ISSUE #3849: Use apiVersion to keep compatible with previous version Kotlin stdlib
## Expected Behavior

`detekt` using `apiVersion` compilation flag to ensure that used classes and methods backward compatible with previous version of Kotlin stdlib.

## Current Behavior

Currently `detekt` doesn't use `apiVersion` compilation flag that lead to the situation that `detekt` may not be used if old version of Kotlin stdlib on classpath.

### Real world example

We have complex project that using Gradle with many plugins and seems one of them bringing Kotlin stdlib 1.3.x on the classpath, `detekt` since a long time was build with stdlib 1.4.x that not fully backward compatible with stdlib 1.3.x. Therefore we can't use `detekt` in current setup before we find and update all used plugins to the 1.4.x compatible version.

### Ktlint

Some time ago [ktlint made similar move](https://github.com/pinterest/ktlint/pull/848)   

### Docs

- [Kotlin Gradle plugin flags](https://kotlinlang.org/docs/gradle.html#attributes-common-for-jvm-and-js)
- [Kotlin compatibility tools](https://kotlinlang.org/docs/kotlin-evolution.html#compatibility-tools)

cortinico:
This sets the `apiVersion` to `1.4` and forces us to use only stdlib methods that were included in the 1.4 version.

Given that the usage of the 1.5 APIs is limited, I believe we can support this.

Fixes #3849

### DIFF
--- a/buildSrc/src/main/kotlin/module.gradle.kts
+++ b/buildSrc/src/main/kotlin/module.gradle.kts
@@ -84,6 +84,7 @@ tasks.withType<KotlinCompile>().configureEach {
     kotlinOptions {
         jvmTarget = Versions.JVM_TARGET
         languageVersion = "1.5"
+        apiVersion = "1.4"
         freeCompilerArgs = listOf(
             "-progressive",
             "-Xopt-in=kotlin.RequiresOptIn"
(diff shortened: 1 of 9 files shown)


schalkms (approved):
Agreed!

BraisGabin (approved):
Should we add information about this in our compatibility table?

cortinico:
> Should we add information about this in our compatibility table?

Yup I will do it as soon as we release 1.18.0 👍

PR #3858, opened 2021-06-06
UnnecessaryParentheses: fix false positive for delegated expressions

### LINKED ISSUE #3851: UnnecessaryParentheses is invalid for expressions used for delegation.
## Expected Behavior

In Kotlin, an expression used to delegate to an interface by using the `by` keyword sometimes needs to be parenthesized. In particular, this seems to occur when using expressions with tailing lambda arguments (though I am not sure exactly what causes this behavior in the parser). For example:

```
interface MyInterface {
    val someProperty: String
}

fun mkInterface(value: String): MyInterface = 
    object: MyInterface {
        override val someProperty = value
    } 

fun MyInterface.map(f: (String) -> String): MyInterface {
    val orig = this
    return object: MyInterface {
        override val someProperty = f(orig.someProperty)
    }
}

class MyClass: MyInterface by (
    mkInterface("hello").map {
        it + "!"
    }
)
```

If the parenthesis after the "by" are removed, the above will not parse. So the expected behavior is that detekt will not report a UnnecessaryParentheses error for the above. 

## Observed Behavior

Detekt will report an UnnecessaryParentheses error for the expression above, even though the parenthesis are actually nescesary.

## Context

I use this sort of delegation all the time, so I get a ton of mis-reported "UnnecessaryParentheses" in my code-base that are actually necessary.

## Your Environment

cortinico:
Fixes #3851

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParentheses.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParentheses.kt
@@ -7,6 +7,7 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.psi.KtDelegatedSuperTypeEntry
 import org.jetbrains.kotlin.psi.KtParenthesizedExpression
 import org.jetbrains.kotlin.psi.KtPsiUtil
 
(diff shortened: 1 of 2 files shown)


chao2zhang (approved):

schalkms (approved):

BraisGabin:
This rule delegates all the heavy lifting on `KtPsiUtil.areParenthesesUseless(expression)`. Shouldn't we open an issue to JetBrains to fix it?

I'm not sure if this is an issue in `areParenthesesUseless` or even the compiler. I don't see why those parentheses are needed.

cortinico:
> I'm not sure if this is an issue in `areParenthesesUseless` or even the compiler. I don't see why those parentheses are needed.

True. I've created an issue here https://youtrack.jetbrains.com/issue/KT-47159
Will try to ask on Slack if someone has insights on this.

PR #3859, opened 2021-06-06
Remove custom assertions that check kdoc of rules

marschwar:
This is part of the #3670 cleanup. 

- removes checks that are no longer necessary after switching to annotations
- add missing rule sets to checks
- add dependency to formatting rules to be able to check those
- test rule configs more generically

### DIFF
--- a/detekt-core/build.gradle.kts
+++ b/detekt-core/build.gradle.kts
@@ -15,6 +15,7 @@ dependencies {
     implementation(projects.detektReportSarif)
 
     testRuntimeOnly(projects.detektRules)
+    testRuntimeOnly(projects.detektFormatting)
     testImplementation(projects.detektTest)
     testImplementation(testFixtures(projects.detektApi))
     testImplementation(libs.mockk)
--- a/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/ConfigAssert.kt
+++ b/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/ConfigAssert.kt
@@ -17,15 +17,19 @@ class ConfigAssert(
     private val allowedOptions = setOf(
         Config.ACTIVE_KEY,
         Config.EXCLUDES_KEY,
-        Config.INCLUDES_KEY
+        Config.INCLUDES_KEY,
+        Config.AUTO_CORRECT_KEY,
+        "android"
     )
 
     fun assert() {
         val ymlDeclarations = getYmlRuleConfig().properties.filter { it.key !in allowedOptions }
         assertThat(ymlDeclarations).isNotEmpty
+
         val ruleClasses = getRuleClasses()
-        assertThat(ruleClasses).isNotEmpty
-        assertThat(ruleClasses).hasSize(ymlDeclarations.size)
+        val foundRulesClassNames = ruleClasses.map { it.simpleName }
+        val ruleNamesInConfig = ymlDeclarations.keys
+        assertThat(foundRulesClassNames).containsExactlyInAnyOrderElementsOf(ruleNamesInConfig)
 
         checkRules(ruleClasses, ymlDeclarations)
     }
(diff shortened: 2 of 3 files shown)


detekt-core/build.gradle.kts
```diff
@@ -15,6 +15,7 @@ dependencies {
     implementation(projects.detektReportSarif)
 
     testRuntimeOnly(projects.detektRules)
+    testRuntimeOnly(projects.detektFormatting)
```
marschwar, line 18: It took me a while to realize that the module `detekt-formatting` is not part of the `detektRules` configuration. Is that done intentionally?

chao2zhang: I believe so because detekt-formatting is ktlint which has different ownership from detekt native rules.

marschwar: Ok, thank you. Do you agree, that the config for those rules should be checked here?

chao2zhang: Agreed. I believe it is reasonable to define it as testRuntimeOnly to increase our test coverage.

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/ConfigAssert.kt
```diff
@@ -38,33 +40,9 @@ class ConfigAssert(
...
-        var filter = ymlOptions.filterKeys { it !in allowedOptions }
-        if (filter.containsKey(THRESHOLD)) {
-            assertThat(ruleClass.superclass.simpleName).isEqualTo(THRESHOLD_RULE)
-            filter = filter.filterKeys { it != THRESHOLD }
-        }
-        for (ymlOption in filter) {
-            val configField = configFields.singleOrNull { ymlOption.key == it.get(null) }
-            if (configField == null) {
-                fail<String>("${ymlOption.key} option for ${ruleClass.simpleName} rule is not correctly defined")
-            }
```
chao2zhang, line 61: This piece code is providing certain coverage for at least the rule config key "active".

schalkms: Agree

marschwar: I disagree. This code explicitly ignores the `active` key as it is part of `allowedOptions` (see below).

```kotlin
val configFields = ruleClass.declaredFields.filter { isPublicStaticFinal(it) && it.name != "Companion" }
var filter = ymlOptions.filterKeys { it !in allowedOptions }
// ...
for (ymlOption in filter) {
  // ...
}
```

What this code acutally does is make sure there is a constant with the name of every config key (except for "active", "excludes", ...) such as.
```kotlin
const val THRESHOLD = "threshold"
const val FUNCTION_THRESHOLD = "functionThreshold"
const val CONSTRUCTOR_THRESHOLD = "constructorThreshold"
const val IGNORE_DEFAULT_PARAMETERS = "ignoreDefaultParameters"
const val IGNORE_DATA_CLASSES = "ignoreDataClasses"
const val IGNORE_ANNOTATED = "ignoreAnnotated"
```
for
```yaml
  LongParameterList:
    active: true
    functionThreshold: 6
    constructorThreshold: 7
    ignoreDefaultParameters: false
    ignoreDataClasses: true
    ignoreAnnotated: []
```

chao2zhang: Thanks for pointing out the logic, I did run the test in IDE again for the existing code, it looks like most of the rules are hitting the return branch of `if (ruleClass.isConfiguredWithAnnotations()) return`, and what I have observed is that when hitting `for (ymlOption in filter)`, the filter is almost always an empty list.

marschwar: Yes, as all rules have been moved to the configuration by annotation that was pretty much what I expected. Is there any check you would like me to add here?

In a separate PR I was thinking about adding something to the `ConfigurationCollector` that fails when there is a `@configuration` kdoc on a rule. Technically this could be done here as well.

chao2zhang: Let's do it in a separate PR. Since the tests are not covering much as of today.

chao2zhang (approved):

Issue #3488, opened 2021-02-21
Detekt type resolution task not working with Android data binding


chao2zhang:
## Expected Behavior
Run `./gradlew :app:detektMain` should succeed

## Observed Behavior
It failed
```
> Task :app:detektDebug
//Users/cazhang/kotlin-android-template/app/src/main/java/com/ncorti/kotlin/template/app/MainActivity.kt:6:39: error: unresolved reference: databinding
import com.ncorti.kotlin.template.app.databinding.ActivityMainBinding
                                      ^
//Users/cazhang/kotlin-android-template/app/src/main/java/com/ncorti/kotlin/template/app/MainActivity.kt:13:35: error: unresolved reference: ActivityMainBinding
    private lateinit var binding: ActivityMainBinding
                                  ^
//Users/cazhang/kotlin-android-template/app/src/main/java/com/ncorti/kotlin/template/app/MainActivity.kt:17:19: error: unresolved reference: ActivityMainBinding
        binding = ActivityMainBinding.inflate(layoutInflater)
                  ^
//Users/cazhang/kotlin-android-template/app/src/main/java/com/ncorti/kotlin/template/app/MainActivity.kt:24:32: error: variable expected
            binding.textResult.text = result
                               ^
//Users/cazhang/kotlin-android-template/app/src/main/java/com/ncorti/kotlin/template/app/MainActivity.kt:25:32: error: variable expected
            binding.textResult.visibility = View.VISIBLE
                               ^
```

## Steps to Reproduce
Run `./gradlew detektMain` in `https://github.com/cortinico/kotlin-android-template`

## Context
This is found during my local change to move detekt to use `com.android.tools.build:gradle-api` instead of ``com.android.tools.build:gradle`. I believe this is a blocker for type resolution to be stable.

## Your Environment

* Version of detekt used: 1.16.0-RC1
* Version of Gradle used (if applicable): 6.8.2
* Gradle scan link (add `--scan` option when running the gradle task): https://scans.gradle.com/s/v5ihlc5ed3z4c
* Operating System and version: MacOS 11.2
* Link to your project (if it's a public repository): https://github.com/cortinico/kotlin-android-template

BraisGabin:
I think that the only way to fix this is to move to a compiler plugin.

schalkms:
Duplicates of #3272 #3409

chao2zhang:
[...]

PR #3867, opened 2021-06-12
Android: add javac intermediates to classpath

damianw:
This adds the classes from `build/intermediates/javac` to the classpath used to compile the code, in order for the compiler to be able to resolve references to Java code (included that which is generated). This resolves issues such as #3488 among probably others.

A test is included which enables `viewBinding` on a sample project and verifies that the Detekt invocation doesn't produce any errors about unresolved references to the generated classes.

Downside: invoking Detekt (with type resolution) now depends on the full assembly of the variant. Before (I think?) this was not a requirement. This is a bit of a trade-off, although in practice I don't think it makes too much of a difference considering the "expensive" type resolution tasks are typically run on CI (we configure Detekt in this way already in a fairly large codebase and haven't found it to be especially burdensome). It would be possible to make this an opt-in option if needed though. 🙃


chao2zhang (approved):
Thank you for the improvements. I think the long-term goal is to build detekt as a compiler plugin, but we should help improve the existing android integration at the same time.

BraisGabin (approved):

PR #3879, opened 2021-06-17
Improve LongParameterList rule by supporting ignoring annotated parameters

andrepaz:
Add support for ignoring annotated parameters in constructors and functions in addition to the other forms of ignoring already supported by the LongParameterList rule.

It should be useful for annotations such as Spring's @Value that mark something that usually doesn't actually add a dependency, just a custom form of configuration (similar to what a parameter with a default value would do) for a function or class and as such might not violate the Single Responsibility Principle as stated by the rule's own documentation of its motivation.

Users should be able to decide if annotated parameters should count or not, just as they can choose or not to ignore parameters with a default value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,7 @@ If you contributed to detekt but your name is not in the list, please feel free 
 - [Vinicius Montes Munhoz](https://github.com/vfmunhoz) - Documentation improvement
 - [Eliezer Graber](https://github.com/eygraber) - Rule fix: ModifierOrder
 - [Dominik Labuda](https://github.com/Dominick1993) - Gradle plugin improvement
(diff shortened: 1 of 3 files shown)


rmzoni (approved):
Great work @andrepaz

schalkms (approved):
Awesome contribution! Thank you. Could you please add this to the documentation in the rule class in order to make this change transparent for users?

andrepaz:
> Awesome contribution! Thank you. Could you please add this to the documentation in the rule class in order to make this change transparent for users?

Sure. I've changed the text to mention explicit support for parameters as well as the other annotation forms. Is that OK?

cortinico (approved):
Great work 👍

PR #3880, opened 2021-06-17
VarCouldBeVal: don't report assigned property that is defined in anonymous object

### LINKED ISSUE #3805: Wrong VarCouldBeVal report
## Expected Behavior

I shouldn't get a report about `VarCouldBeVal` for a property in an object in a function.

## Observed Behavior

I get a `VarCouldBeVal` issue for a property, that cannot be changed to `val`

## Steps to Reproduce

The following code reproduces the issue
```kotlin
fun test(){
    val wrapper = object {
        var test: Boolean = true
    }
    wrapper.test = false
}
```

## Your Environment

* Version of detekt used: 1.17.0
* Version of Gradle used (if applicable): 7.0.2
* Link to your project (if it's a public repository): https://github.com/EdwarDDay/serialization.kprefs/pull/98

t-kameyama:
Fixes #3805

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
@@ -9,16 +9,18 @@ import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
-import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.KtBinaryExpression
-import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
+import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtNamedDeclaration
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtProperty
-import org.jetbrains.kotlin.psi.KtThisExpression
+import org.jetbrains.kotlin.psi.KtQualifiedExpression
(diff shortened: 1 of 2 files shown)


BraisGabin (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
```diff
@@ -41,6 +43,7 @@ private val unaryAssignmentOperators = setOf(KtTokens.MINUSMINUS, KtTokens.PLUSP
  * </compliant>
  */
 @ActiveByDefault(since = "1.16.0")
+@RequiresTypeResolution
```
chao2zhang, line 46: It seems quite unfortunate that we are now converting plain rules to type resolving rules like such.

I think this PR should still be good to go, but I would want to highlight this for us to favor detekt compiler plugin.

chao2zhang (approved):

schalkms (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
cortinico: I believe we could mention this in the release notes but we're good to go imho.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
```diff
@@ -144,11 +149,34 @@ class VarCouldBeValSpec : Spek({
...
-                // we accept wrong entity reporting here due to no type resolution
-                // false reporting with shadowed vars vs false positives
-                assertThat(entity.ktElement?.text).isEqualTo("private var myVar: String? = null")
+            with(subject.compileAndLintWithContext(env, code)[0]) {
+                assertThat(entity.ktElement?.text).isEqualTo("var myVar = value")
             }
         }
     }
+
+    describe("properties defined in anonymous object - #3805") {
```
cortinico, line 158: Would you be able to add a test case also for:
https://github.com/detekt/detekt/issues/3805#issuecomment-862068331

cortinico (commented):
Great work, thanks for fixing this 🙏

cortinico (approved):

PR #3892, opened 2021-06-22
LongMethod: add 'ignoreAnnotated' configuration option

### LINKED ISSUE #3887: Rule configuration request: [LongMethod] Ignore based on annotations
### Expected Behavior of the rule

```
LongMethod:
    active: true
    threshold: 20
    ignoreAnnotated: ['Composable']
```

### Context

We started recently to refactor our project to use Jetpack Compose. 
For `LongMethod` rule we have strict threshold set for 20 lines. It makes sense for regular Kotlin stuff. 

Jetpack `@Composable` methods tend to be quite big and we need to suppress detect in multiple places for such methods.

I would like to suggest adding a new configuration param for `LongMethod` rule to ignore checks based on annotation.

Please let me know your thoughts.

t-kameyama:
Fixes #3887

### DIFF
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongMethod.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongMethod.kt
@@ -40,7 +40,7 @@ class LongMethod(config: Config = Config.empty) : Rule(config) {
     private val threshold: Int by config(defaultValue = 60)
 
     @Configuration("ignore long methods in the context of these annotation class names")
-    private val ignoreAnnotated: List<String> by config(listOf("Composable"))
+    private val ignoreAnnotated: List<String> by config(emptyList())
 
     private val functionToLinesCache = HashMap<KtNamedFunction, Int>()
     private val functionToBodyLinesCache = HashMap<KtNamedFunction, Int>()
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongMethod.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongMethod.kt
@@ -61,8 +61,6 @@ class LongMethod(config: Config = Config.empty) : Rule(config) {
         }
         for ((function, lines) in functionToLines) {
             if (lines >= threshold) {
-                val annotationExcluder = AnnotationExcluder(function.containingKtFile, ignoreAnnotated)
-                if (annotationExcluder.shouldExclude(function.annotationEntries)) continue
                 report(
                     ThresholdedCodeSmell(
                         issue,
@@ -77,6 +75,9 @@ class LongMethod(config: Config = Config.empty) : Rule(config) {
     }
 
     override fun visitNamedFunction(function: KtNamedFunction) {
+        val annotationExcluder = AnnotationExcluder(function.containingKtFile, ignoreAnnotated)
+        if (annotationExcluder.shouldExclude(function.annotationEntries)) return
+
         val parentMethods = function.getStrictParentOfType<KtNamedFunction>()
         val bodyEntity = function.bodyBlockExpression ?: function.bodyExpression
         val lines = (if (parentMethods != null) function else bodyEntity)?.linesOfCode() ?: 0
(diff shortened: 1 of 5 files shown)


detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongMethod.kt
```diff
@@ -57,6 +61,8 @@ class LongMethod(config: Config = Config.empty) : Rule(config) {
         }
         for ((function, lines) in functionToLines) {
             if (lines >= threshold) {
+                val annotationExcluder = AnnotationExcluder(function.containingKtFile, ignoreAnnotated)
+                if (annotationExcluder.shouldExclude(function.annotationEntries)) continue
```
marschwar, line 65: Just a thought: How about checking if the function is annotated with an ignored function in `visitNamedFunction` and not count the lines and put the function into the map in the first place?

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongMethod.kt
```diff
@@ -38,6 +39,9 @@ class LongMethod(config: Config = Config.empty) : Rule(config) {
     @Configuration("number of lines in a method to trigger the rule")
     private val threshold: Int by config(defaultValue = 60)
 
+    @Configuration("ignore long methods in the context of these annotation class names")
+    private val ignoreAnnotated: List<String> by config(listOf("Composable"))
```
marschwar, line 43: I don't like the idea that the detekt defaults become more and more android related. In my opinion it should focus on the kotlin language and not on frameworks like Spring or Android.

t-kameyama: I've implemented it the same as FunctionNaming.

marschwar: I know, your PR is totally fine. Maybe I should have started a discussion rather than raising this here. Sorry about that.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongMethod.kt
```diff
@@ -71,6 +75,9 @@ class LongMethod(config: Config = Config.empty) : Rule(config) {
     }
 
     override fun visitNamedFunction(function: KtNamedFunction) {
+        val annotationExcluder = AnnotationExcluder(function.containingKtFile, ignoreAnnotated)
```
marschwar, line 78: I don't know how nit picky this is but creating the `Excluder` for every function seems wasteful. I would prefer the way this is implemented in `LongParameterList`.

```kotlin
    private lateinit var annotationExcluder: AnnotationExcluder

    override fun visitKtFile(file: KtFile) {
        annotationExcluder = AnnotationExcluder(file, ignoreAnnotated)
        super.visitKtFile(file)
    }
```

marschwar (approved):

schalkms (approved):

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongMethod.kt
schalkms: I agree that we shouldn’t have such opinionated defaults. Sometimes it slipped through the reviews. In general, we handled it in the past like @marschwar said.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongMethod.kt
schalkms: +1

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongMethod.kt
schalkms: I think we should remove this default value from both rules.

dugsmith:
Thank you for adding this! 👏

PR #3898, opened 2021-06-24
Don't construct an AnnotationExcluder unnecessarily

schalkms:
The creation of objects would a be waste of memory.
The changes made comply with the implementation in other rules.

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/LateinitUsage.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/LateinitUsage.kt
@@ -47,24 +47,22 @@ class LateinitUsage(config: Config = Config.empty) : Rule(config) {
     private val ignoreOnClassesPattern: Regex by config("", String::toRegex)
 
     private var properties = mutableListOf<KtProperty>()
-
-    override fun visitProperty(property: KtProperty) {
-        if (property.isLateinit()) {
-            properties.add(property)
-        }
-    }
+    private lateinit var annotationExcluder: AnnotationExcluder
 
     override fun visit(root: KtFile) {
-        properties = mutableListOf()
-
         super.visit(root)
-
-        val annotationExcluder = AnnotationExcluder(root, excludeAnnotatedProperties)
-
+        properties = mutableListOf()
+        annotationExcluder = AnnotationExcluder(root, excludeAnnotatedProperties)
         properties.filterNot { annotationExcluder.shouldExclude(it.annotationEntries) }
             .filterNot { it.containingClass()?.name?.matches(ignoreOnClassesPattern) == true }
             .forEach {
                 report(CodeSmell(issue, Entity.from(it), "Usages of lateinit should be avoided."))
             }
+    }
+
+    override fun visitProperty(property: KtProperty) {
+        if (property.isLateinit()) {
+            properties.add(property)
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/LateinitUsage.kt
```diff
@@ -47,24 +47,22 @@ class LateinitUsage(config: Config = Config.empty) : Rule(config) {
     private val ignoreOnClassesPattern: Regex by config("", String::toRegex)
 
     private var properties = mutableListOf<KtProperty>()
-
-    override fun visitProperty(property: KtProperty) {
-        if (property.isLateinit()) {
-            properties.add(property)
-        }
-    }
+    private lateinit var annotationExcluder: AnnotationExcluder
```
marschwar, line 50: In this particular case it does not make sense to change it since the object is only create once per file. This change does seem to break the tests.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClass.kt
```diff
@@ -64,14 +64,16 @@ class UseDataClass(config: Config = Config.empty) : Rule(config) {
...
     private val allowVars: Boolean by config(false)
 
+    private lateinit var annotationExcluder: AnnotationExcluder
+
     override fun visit(root: KtFile) {
         super.visit(root)
-        val annotationExcluder = AnnotationExcluder(root, excludeAnnotatedClasses)
-        root.forEachDescendantOfType<KtClass> { visitKlass(it, annotationExcluder) }
+        annotationExcluder = AnnotationExcluder(root, excludeAnnotatedClasses)
+        root.forEachDescendantOfType<KtClass> { visitKlass(it) }
```
marschwar, line 72: Only created once per file. No change needed.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionNaming.kt
```diff
@@ -44,10 +45,17 @@ class FunctionNaming(config: Config = Config.empty) : Rule(config) {
...
+    override fun visitKtFile(file: KtFile) {
+        annotationExcluder = AnnotationExcluder(file, ignoreAnnotated)
+        super.visitKtFile(file)
+    }
+
     override fun visitNamedFunction(function: KtNamedFunction) {
         super.visitNamedFunction(function)
 
-        if (ignoreOverridden && function.isOverride() || shouldAnnotatedFunctionBeExcluded(function)) {
+        if (ignoreOverridden && function.isOverride() || annotationExcluder.shouldExclude(function.annotationEntries)) {
```
marschwar, line 58: :+1:

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClass.kt
schalkms: True. I should have noted that in this case we don't need to pass the instance internally around.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/LateinitUsage.kt
schalkms: In this case it doesn't bring anything to the table, yes. Thanks for the review.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionNaming.kt
schalkms: Unfortunately, this usage scenario isn't easily changeable, since detekt calls this visitor function externally by the `NamingFunction` class.

```kotlin
private fun handleFunction(declaration: KtNamedFunction) {
        functionNamingRule.runIfActive { visitNamedFunction(declaration) }
```

schalkms:
Closed since the one scenario can't be changed that easily, as noted in the comments. The rest of the changes only concern the code arrangement.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionNaming.kt
marschwar: :cry:

PR #3905, opened 2021-06-25
Allow else when {...} in MandatoryBracesIfStatements rule

### LINKED ISSUE #3708: Allow multi-line `when` statement following an `else` without requiring braces
This is related to the `MandatoryBracesIfStatements` rule.

## Expected Behavior
To allow this code snippet without lint issues:
```kotlin
fun getList(quality: QualityEnum, specialFlag: Boolean) =
  if (specialFlag) {
    // ...
  } else when (quality) {
    // exhaustive enum when statement
  }
```

## Current Behavior
For the above snipped, there are lint errors thrown that require it to be written in a less legible manner (imho), like so:
```kotlin
fun getList(quality: QualityEnum, specialFlag: Boolean) =
  if (specialFlag) {
    // ...
  } else {
    when (quality) {
      // exhaustive enum when statement
    }
  }
```

## Context
- To check a flag which requires custom behavior or otherwise default to a `when(..)` statement's result
- This can be combined into a single big `when` statement, however, in doing so the exhaustive `when(quality)` is lost, meaning that we have to add an `else` branch, where it would not be necessary otherwise
- The check for `specialFlag` is necessary to be done first (and thus cannot be in an `else` branch of a larger `when` statement, as even when its true, `quality` may have a valid value, which would mean it would incorrectly use another one of the cases instead.

schalkms:
This allows multi-line when statements following an else without requiring braces.

Fixes #3708

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatements.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatements.kt
@@ -10,6 +10,8 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.psi.KtBlockExpression
 import org.jetbrains.kotlin.psi.KtIfExpression
+import org.jetbrains.kotlin.psi.KtWhenCondition
+import org.jetbrains.kotlin.psi.KtWhenExpression
 import org.jetbrains.kotlin.psi.psiUtil.siblings
 
 private const val DESCRIPTION = "Multi-line if statement was found that does not have braces. " +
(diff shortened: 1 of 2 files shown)


cortinico (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatements.kt
```diff
@@ -56,5 +57,6 @@ class MandatoryBracesIfStatements(config: Config = Config.empty) : Rule(config)
     private fun KtIfExpression.isNotBlockOrIfExpression(): Boolean =
```
marschwar, line 57: The names of both extension functions seem wrong to me. The change has not exactly made it any better ;)

marschwar (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatements.kt
schalkms: Thanks! Those names should really be improved.
I'll change it in a separate PR.

PR #3908, opened 2021-06-26
Refactor + rename util function inside MandatoryBracesIfStatement rule

schalkms:

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatements.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatements.kt
@@ -36,19 +36,20 @@ class MandatoryBracesIfStatements(config: Config = Config.empty) : Rule(config) 
     override val issue = Issue("MandatoryBracesIfStatements", Severity.Style, DESCRIPTION, Debt.FIVE_MINS)
 
     override fun visitIfExpression(expression: KtIfExpression) {
-        if (expression.then !is KtBlockExpression && hasNewLine(expression.rightParenthesis)) {
-            report(CodeSmell(issue, Entity.from(expression.then ?: expression), DESCRIPTION))
+        super.visitIfExpression(expression)
+
+        val thenExpression = expression.then ?: return
+        if (thenExpression !is KtBlockExpression && hasNewLineAfter(expression.rightParenthesis)) {
+            report(CodeSmell(issue, Entity.from(thenExpression), DESCRIPTION))
         }
 
-        val elseExpression = expression.`else`
-        if (elseExpression != null && hasCorrectElseType(elseExpression) && hasNewLine(expression.elseKeyword)) {
+        val elseExpression = expression.`else` ?: return
+        if (mustBeOnSameLine(elseExpression) && hasNewLineAfter(expression.elseKeyword)) {
             report(CodeSmell(issue, Entity.from(elseExpression), DESCRIPTION))
         }
-
-        super.visitIfExpression(expression)
     }
 
-    private fun hasNewLine(element: PsiElement?): Boolean {
+    private fun hasNewLineAfter(element: PsiElement?): Boolean {
         if (element == null) return false
         return element
             .siblings(forward = true, withItself = false)
@@ -56,6 +57,6 @@ class MandatoryBracesIfStatements(config: Config = Config.empty) : Rule(config) 
             .any { it.textContains('\n') }
     }
 
-    private fun hasCorrectElseType(expression: KtExpression): Boolean =
+    private fun mustBeOnSameLine(expression: KtExpression): Boolean =
         expression !is KtIfExpression && expression !is KtBlockExpression && expression !is KtWhenExpression
 }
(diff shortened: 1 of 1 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatements.kt
```diff
@@ -35,28 +36,26 @@ class MandatoryBracesIfStatements(config: Config = Config.empty) : Rule(config)
...
+            .takeWhile { it.text != "else" }
+            .any { it.textContains('\n') }
+    }
 
-    private fun KtIfExpression.isNotBlockOrIfExpression(): Boolean =
-        this.`else` != null &&
-            this.`else` !is KtIfExpression &&
-            this.`else` !is KtBlockExpression &&
-            this.`else` !is KtWhenExpression
+    private fun hasCorrectElseType(expression: KtExpression): Boolean =
```
schalkms, line 59: Maybe you have better ideas for the function name.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatements.kt
```diff
@@ -35,28 +36,26 @@ class MandatoryBracesIfStatements(config: Config = Config.empty) : Rule(config)
...
-        if (expression.isNotBlockExpression() && hasNewLine(expression.rightParenthesis)) {
+        if (expression.then !is KtBlockExpression && hasNewLine(expression.rightParenthesis)) {
             report(CodeSmell(issue, Entity.from(expression.then ?: expression), DESCRIPTION))
         }
 
-        if (expression.isNotBlockOrIfExpression() && hasNewLine(expression.elseKeyword)) {
-            report(CodeSmell(issue, Entity.from(expression.`else` ?: expression), DESCRIPTION))
+        val `else`: KtExpression? = expression.`else`
+        if (`else` != null && hasCorrectElseType(`else`) && hasNewLine(expression.elseKeyword)) {
+            report(CodeSmell(issue, Entity.from(`else`), DESCRIPTION))
```
schalkms, line 45: By doing the null check before, we can safely pass the else variable here.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatements.kt
```diff
@@ -35,28 +36,26 @@ class MandatoryBracesIfStatements(config: Config = Config.empty) : Rule(config)
...
-        if (expression.isNotBlockExpression() && hasNewLine(expression.rightParenthesis)) {
+        if (expression.then !is KtBlockExpression && hasNewLine(expression.rightParenthesis)) {
             report(CodeSmell(issue, Entity.from(expression.then ?: expression), DESCRIPTION))
         }
 
-        if (expression.isNotBlockOrIfExpression() && hasNewLine(expression.elseKeyword)) {
-            report(CodeSmell(issue, Entity.from(expression.`else` ?: expression), DESCRIPTION))
+        val elseExpression = expression.`else`
+        if (elseExpression != null && hasCorrectElseType(elseExpression) && hasNewLine(expression.elseKeyword)) {
+            report(CodeSmell(issue, Entity.from(elseExpression), DESCRIPTION))
```
schalkms, line 45: By doing the null check before, we can safely pass the else variable here.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatements.kt
marschwar: `hasCorrectElseType` is really wrong. It should be `hasIncorrectElseType`.

I played around with it a little bit and realized that we do allow a when expression within the else branch but not in the then branch. It may not be something that is used often, but I don't see a reason why this is compliant

```kotlin
fun f(i: Int) {
    if (true) {
        println()
    } else when(i) {
        1 -> println(1)
        else -> println()
    }
}
```

while this is not
```kotlin
fun f(i: Int) {
    if (true) when(i) {
        1 -> println(1)
        else -> println()
    } else when(i) {
        println()
    }
}
```

I ended up with
```kotlin
val thenExpression = expression.then
if (mustBeOnSameLine(thenExpression) && hasNewLineAfter(expression.rightParenthesis)) {
    report(CodeSmell(issue, Entity.from(thenExpression), DESCRIPTION))
}

val elseExpression = expression.`else`
if (isForbiddenInElse(elseExpression) && hasNewLineAfter(expression.elseKeyword)) {
    report(CodeSmell(issue, Entity.from(elseExpression), DESCRIPTION))
}

@ExperimentalContracts
private fun mustBeOnSameLine(expression: KtExpression?): Boolean {
    contract {
        returns(true) implies (expression != null)
    }
    return expression != null &&
        expression !is KtIfExpression &&
        expression !is KtBlockExpression &&
        expression !is KtWhenExpression
}
```

Ok, the part with the contract is a little bit over the top, but I had never used it before and it seemed to make sense here ;).

schalkms: Thanks for the suggestion. I have the feeling that experimental contracts are a bit overkill in this case. I'm also not sure whether we want experimental features used in detekt's code base.

marschwar: I totally agree. I updated the comment above removing the contract.

cortinico (approved):

BraisGabin (approved):

Issue #3684, opened 2021-04-17
Noisy output of detekt on type solving mode


BraisGabin:
## Expected Behavior
`detektMain` and `detektTest` quietly. At least in `detekt` that we suppose to have it configured correctly.

## Observed Behavior
They don't: https://scans.gradle.com/s/q7fdffszk5aim/console-log/raw

## Steps to Reproduce
Run `./gradlew detektMain detektTest --rerun-tasks` in the main branch of this project

## Issues detected
I think that we have 3 different issues but they could be related.

~~1. Empty lines~~ (fixed by #3909)
```
:detekt-core:detektMain

:detekt-formatting:detektMain
:detekt-generator:detektMain









:detekt-gradle-plugin:detektMain





:detekt-metrics:detektMain
```
2. Errors complaining about `internal` modifiers:
```
:detekt-metrics:detektTest
//Users/brais/projects/detekt/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/AnalyzerSpec.kt:28:28: error: cannot access 'Analyzer': it is internal in 'io.gitlab.arturbosch.detekt.core'
            val analyzer = Analyzer(settings, listOf(StyleRuleSetProvider()), emptyList())
                           ^
```
3. Errors that show that the type solving is not working as expected:
```
//Users/brais/projects/detekt/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFormatSpec.kt:47:67: error: no value passed for parameter 'body'
                assertThat(manuallySuppressedIssues).anySatisfy { it.startsWith("LongParameterList") }
```

```
//Users/brais/projects/detekt/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFormatSpec.kt:47:70: error: unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
public fun File.startsWith(other: File): Boolean defined in kotlin.io
public fun File.startsWith(other: String): Boolean defined in kotlin.io
public fun CharSequence.startsWith(char: Char, ignoreCase: Boolean = ...): Boolean defined in kotlin.text
public fun CharSequence.startsWith(prefix: CharSequence, ignoreCase: Boolean = ...): Boolean defined in kotlin.text
public fun CharSequence.startsWith(prefix: CharSequence, startIndex: Int, ignoreCase: Boolean = ...): Boolean defined in kotlin.text
[...]

PR #3909, opened 2021-06-26
Avoid empty lines when running detekt with type resolution

marschwar:
This relates to #3684 and solves the issue of empty lines when running detekt with type resolution.


schalkms (commented):
Shall we add a test for this behavior in order to avoid this issue in the future?

marschwar:
I don't think there is much that can be tested here other than the `MessageCollector` itself.

schalkms (approved):

PR #3919, opened 2021-06-29
Added `--auto-correct` and `--no-auto-correct` options to `Detekt` task

gmazzo:
Adds Gradle's task option support for fast enabling/disabling `auto correct` feature.

The goal is to have this lightweight switch to avoid extra logic in the build script to support it.

```bash
./gradlew detetkt --auto-correct
```
or
```bash
./gradlew detetkt --no-auto-correct
```

Our use case is to have this command in our CI pipeline.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -48,6 +48,7 @@ import org.gradle.api.tasks.SkipWhenEmpty
 import org.gradle.api.tasks.SourceTask
 import org.gradle.api.tasks.TaskAction
 import org.gradle.api.tasks.VerificationTask
+import org.gradle.api.tasks.options.Option
 import org.gradle.language.base.plugins.LifecycleBasePlugin
 import java.io.File
 import javax.inject.Inject
@@ -210,6 +211,16 @@ open class Detekt @Inject constructor(
     override fun setIgnoreFailures(value: Boolean) = ignoreFailuresProp.set(value)
 
     fun reports(configure: Action<DetektReports>) = configure.execute(reports)
+
+    @Option(option = "auto-correct", description = "Enables auto correct")
+    fun autoCorrect() {
+        autoCorrect = true
+    }
+
+    @Option(option = "no-auto-correct", description = "Disables auto correct")
+    fun noAutoCorrect() {
+        autoCorrect = false
+    }
 
     @TaskAction
     fun check() {


cortinico:
Thanks for your PR

> The goal is to have this lightweight switch to avoid extra logic in the build script to support it.

I'm not necessarily against this, but I think merging this will be awkward in some ways:
- First we will have to maintain a "CLI-like" API for the Gradle tasks, and this should be aligned with the Detekt CLI. In this case Detekt CLI has `--auto-correct` (but also `-ac` that here is not supported), but doesn't have `--no-auto-correct`.

- It feels odd to have just one feature that allows to be configured via Gradle CLI, while all the others are not.

- I'm against having both `--auto-correct` and `--no-auto-correct`. When designing CLI we try to offer defaults for all the values and offers flags that mutate those defaults. In this case we should probably offer only the `--auto-correct` flag.

Happy to discuss what's your view on this @gmazzo. Specifically is this PR coming from your own personal experience? Were you not able to achieve something similar with Gradle Properties?

gmazzo:
> Thanks for your PR
> 
> > The goal is to have this lightweight switch to avoid extra logic in the build script to support it.
> 
> I'm not necessarily against this, but I think merging this will be awkward in some ways:
> 
> * First we will have to maintain a "CLI-like" API for the Gradle tasks, and this should be aligned with the Detekt CLI. In this case Detekt CLI has `--auto-correct` (but also `-ac` that here is not supported), but doesn't have `--no-auto-correct`.
> * It feels odd to have just one feature that allows to be configured via Gradle CLI, while all the others are not.
> * I'm against having both `--auto-correct` and `--no-auto-correct`. When designing CLI we try to offer defaults for all the values and offers flags that mutate those defaults. In this case we should probably offer only the `--auto-correct` flag.
> 
> Happy to discuss what's your view on this @gmazzo. Specifically is this PR coming from your own personal experience? Were you not able to achieve something similar with Gradle Properties?

Let me dig further into our use-case and I'll let the API design decision to you.

One of our engineers decided to support this `auto-update` feature in our core libs. As changing the code can be invasive we decided to let it to each developer's personal criteria to turn it on.

To support this, our engineer added `autoCorrect = project.hasProperty("autoCorrect")` to the DSL configuration. So this will be an opt-in feature either through Gradle user defaults or by command line, but not for the whole project.

That's when I remember Gradle has applied this same pattern in their API and decided to open the PR. For instance, `Build Cache` can be enabled by `org.gradle.caching.enabled=true` or with `--build-cache` option. They also have the `--no-build-cache` as an opt-out (in case it generally enabled).

Finally, `--no-auto-correct` has a justification too. Imagine you have enabled this to the general project, in the standard way trough the DSL. Now you can to run an extra check on the CI but you don't want file to get modified in there, so you can use this option to turn it off.

That's all, feel free to reject if it doesn't fit your API design guidelines ;)

cortinico:
Thanks for your explanation @gmazzo 🙏 

> That's all, feel free to reject if it doesn't fit your API design guidelines ;)

I'd like to have the other maintainers opinion on this as well.

schalkms:
I agree with Nico. We handled such cases in the past as you described, since the same can be achieved in Gradle scripts  with a litte bit more effort but without touching detekt's API.

schalkms (commented):
I agree with Nico. We handled such cases in the past as you described, since the same can be achieved in Gradle scripts with a litte bit more effort but without touching detekt's API.

gmazzo:
> I agree with Nico. We handled such cases in the past as you described, since the same can be achieved in Gradle scripts with a litte bit more effort but without touching detekt's API.

Ok then! I'm closing the PR

chao2zhang:
@gmazzo The other alternative is to define two tasks separately:
- `./gradlew detektCheck` to only check for violations (autoCorrect = false)
- `./gradlew detektFormat` to autocorrect violations (autoCorrect = true)

Meanwhile, there are still limitations of Gradle task options (https://docs.gradle.org/current/userguide/custom_tasks.html#limitations):
`Options cannot be declared globally e.g. on a project-level or as part of a plugin.`
`When assigning an option on the command line then the task exposing the option needs to be spelled out explicitly` 

If you are using Detekt's Jvm, Android, or KMM with type resolution, running `./gradlew detekt --auto-correct` it won't work, because we need to specify the explicit task name (such as `:detektMain`)

PR #3827, opened 2021-05-24
Use multi line format for list options in config

marschwar:
This closes #3671 and generates default list option in multi line format.

PR #3920, opened 2021-06-29
Fix java.lang.ClassCastException is reading default yaml config

chao2zhang:
## Stacktrace
```
Caused by: java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
	at io.gitlab.arturbosch.detekt.rules.style.MagicNumber$ignoreNumbers$2.invoke(MagicNumber.kt:259)
	at io.gitlab.arturbosch.detekt.rules.style.MagicNumber$ignoreNumbers$2.invoke(MagicNumber.kt:86)
	at io.gitlab.arturbosch.detekt.api.internal.TransformedConfigProperty.doGetValue(ConfigProperty.kt:87)
	at io.gitlab.arturbosch.detekt.api.internal.MemoizedConfigProperty.getValue(ConfigProperty.kt:64)
	at io.gitlab.arturbosch.detekt.api.internal.MemoizedConfigProperty.getValue(ConfigProperty.kt:60)
	at io.gitlab.arturbosch.detekt.rules.style.MagicNumber.getIgnoreNumbers(MagicNumber.kt:86)
	at io.gitlab.arturbosch.detekt.rules.style.MagicNumber.visitConstantExpression(MagicNumber.kt:138)
```

## Root cause
With https://github.com/detekt/detekt/pull/3827/, each line in a multi-line YAML array can be unquoted, such as
```
    ignoreNumbers:
      - -1
      - 0
      - 1
      - 2
```
This will lead to YAML config being parsed as `List<Int>`, but `MagicNumber.ignoreValues()` still have a default value of `List<String>`, causing the exception.

## Fix
The pessimistic fix is to always print quotes in multi-line YAML array.

## Testing
Since Gradle integration tests are only testing against latest published version, I have verified the fix through:
```
./gradlew :detekt-generator:generateDocumentation
./gradlew publishToMavenLocal
./gradlew :detekt-gradle-plugin:test
`

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -92,15 +92,15 @@ complexity:
     ignoreSimpleWhenEntries: false
     ignoreNestingFunctions: false
     nestingFunctions:
-      - run
-      - let
-      - apply
-      - with
-      - also
-      - use
-      - forEach
-      - isNotNull
-      - ifNull
+      - 'run'
+      - 'let'
+      - 'apply'
+      - 'with'
+      - 'also'
+      - 'use'
+      - 'forEach'
+      - 'isNotNull'
+      - 'ifNull'
   LabeledExpression:
     active: false
     ignoredLabels: []
@@ -199,10 +199,10 @@ exceptions:
   ExceptionRaisedInUnexpectedLocation:
     active: true
     methodNames:
-      - toString
-      - hashCode
-      - equals
-      - finalize
+      - 'toString'
+      - 'hashCode'
+      - 'equals'
+      - 'finalize'
   InstanceOfCheckForException:
     active: false
     excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by marschwar** [comment]
An alternative approach would be to change the `fun Config.valueOrDefaultCommaSeparated` in PathFilters to assume a `List<Any>` and call `toString()` on every item.

```kotlin
fun Config.valueOrDefaultCommaSeparated(
    key: String,
    default: List<String>
): List<String> {
    fun fallBack() = valueOrDefault(key, default.joinToString(","))
        .trim()
        .commaSeparatedPattern(",", ";")
        .toList()

    return try {
        valueOrDefault<List<Any>>(key, default).map { it.toString() }
    } catch (_: IllegalStateException) {
        fallBack()
    } catch (_: ClassCastException) {
        fallBack()
    }
}
```

marschwar:
On second thought... Your solution is probably the safer bet for now. Since we only support `lists that are parameterized with `String` it makes things easier to understand. Anything I can do to help?

cortinico:
> On second thought... Your solution is probably the safer bet for now.

I'd say let's go for that 👍 

This PR is sort of crucial as it's currently blocking the release of `1.18.0`.
We should also look into why those kind of bugs are happening whenever we're about to release a new version (and not before merging changes).

cortinico (approved):

marschwar:
The expectation in `detekt-generator/src/test/resources/RuleSetConfig.yml` has to be updated as well.

marschwar (approved):
I am so glad that you came across this before the release. Thank you

cortinico (approved):

cortinico (approved):

schalkms (approved):

PR #3924, opened 2021-06-30
Add AvoidReferentialEquality rule

### LINKED ISSUE #3363: Check for === instead of ==
Prevent usage of instance equality checks

## Context
`===` and `==` are both valid in kotlin. We also write a lot of typescript/javascript on our team (sadly no multiplatform kotlin) and a few of us accidently use `===` sometimes instead of `==` for equals. There should be a rule to prevent usage of `===` as you very rarely want instance equality checks anyways.

marschwar:
This closes #3363

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/SimpleGlob.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/SimpleGlob.kt
@@ -1,0 +1,23 @@
+package io.gitlab.arturbosch.detekt.api.internal
+
+/**
+ * This simple globbing implementation allows users to define patterns with `?` (any single character)
+ * and `*` (zero or more characters) wildcards.
+ *
+ */
+class SimpleGlob private constructor(private val regex: Regex) {
+
+    fun matches(input: String) = regex.matches(input)
+
+    companion object {
+        fun of(globPattern: String): SimpleGlob {
+            val regex = globPattern
+                .replace("\\", "\\\\")
+                .replace(".", "\\.")
+                .replace("*", ".*")
+                .replace("?", ".")
+                .toRegex()
+            return SimpleGlob(regex)
+        }
+    }
+}
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/SimpleGlob.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/SimpleGlob.kt
@@ -3,7 +3,6 @@ package io.gitlab.arturbosch.detekt.api.internal
 /**
  * This simple globbing implementation allows users to define patterns with `?` (any single character)
  * and `*` (zero or more characters) wildcards.
- *
  */
 class SimpleGlob private constructor(private val regex: Regex) {
 
--- a/detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/internal/SimpleGlobSpec.kt
+++ b/detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/internal/SimpleGlobSpec.kt
@@ -1,0 +1,197 @@
+package io.gitlab.arturbosch.detekt.api.internal
+
+import org.assertj.core.api.Assertions.assertThat
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+
+object SimpleGlobSpec : Spek({
+    describe("glob") {
+        describe("empty pattern") {
+            val subject by memoized { SimpleGlob.of("") }
+
+            it("matches an empty string") {
+                val actual = subject.matches("")
+                assertThat(actual).isTrue()
+            }
+            it("does not match a blank string") {
+                val actual = subject.matches(" ")
+                assertThat(actual).isFalse()
+            }
+        }
+        describe("blank pattern") {
+            val subject by memoized { SimpleGlob.of(" \t") }
+
+            it("matches an empty string") {
+                val actual = subject.matches(" \t")
+                assertThat(actual).isTrue()
+            }
+            it("does not match a different string") {
+                val actual = subject.matches("  ")
+                assertThat(actual).isFalse()
+            }
+        }
+        describe("static pattern") {
+            val subject by memoized { SimpleGlob.of("abc") }
+
+            it("matches the same string") {
+                val actual = subject.matches("abc")
+                assertThat(actual).isTrue()
+            }
+            it("does not match a other string") {
+                val actual = subject.matches("aaa")
+                assertThat(actual).isFalse()
+            }
+        }
+        describe("* wildcard") {
+            describe("single wildcard") {
+                describe("pattern with wildcard at the beginning") {
+                    val subject by memoized { SimpleGlob.of("*xyz") }
+
+                    it("matches pattern exactly") {
+                        val actual = subject.matches("xyz")
+                        assertThat(actual).isTrue()
+                    }
+                    it("matches pattern with anything before") {
+                        val actual = subject.matches("abcxyz")
+                        assertThat(actual).isTrue()
+                    }
+                    it("does not match with anything after") {
+                        val actual = subject.matches("xyzabc")
+                        assertThat(actual).isFalse()
+                    }
+                }
+                describe("pattern with wildcard at the end") {
+                    val subject by memoized { SimpleGlob.of("xyz*") }
+
+                    it("matches pattern exactly") {
+                        val actual = subject.matches("xyz")
+                        assertThat(actual).isTrue()
+                    }
+                    it("matches pattern with anything after") {
+                        val actual = subject.matches("xyzabc")
+                        assertThat(actual).isTrue()
+                    }
+                    it("does not match with anything before") {
+                        val actual = subject.matches("abcxyz")
+                        assertThat(actual).isFalse()
+                    }
+                }
+                describe("pattern with wildcard at the middle") {
+                    val subject by memoized { SimpleGlob.of("x*yz") }
+
+                    it("matches pattern exactly") {
+                        val actual = subject.matches("xyz")
+                        assertThat(actual).isTrue()
+                    }
+                    it("matches pattern with anything in between") {
+                        val actual = subject.matches("xaaaayz")
+                        assertThat(actual).isTrue()
+                    }
+                    it("does not match with anything before") {
+                        val actual = subject.matches("axyz")
+                        assertThat(actual).isFalse()
+                    }
+                    it("does not match with anything after") {
+                        val actual = subject.matches("xyza")
+                        assertThat(actual).isFalse()
+                    }
+                }
+            }
+            describe("multiple wildcards") {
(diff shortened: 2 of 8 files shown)


detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -554,6 +554,10 @@ potential-bugs:
     active: false
   RedundantElseInWhen:
     active: true
+  ReferentialEquality:
+    active: false
+    forbiddenTypesRegex:
+      - kotlin.String
```
BraisGabin, line 560: If we use Reget this should be `kotlin\.String`. But I think that we should not use regex here and just use globbing. So instead of `kotlin\.collections\..*List` we would write `kotlin.collections.*List`

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ReferentialEqualitySpec.kt
```diff
@@ -0,0 +1,87 @@
...
+                val list = listOf(1)
+                val mutableList = mutableListOf(1)
+                val b = list === listOf(2) || mutableList === mutableListOf(2)
+            """.trimIndent()
+
+            val actual = subject.compileAndLintWithContext(env, code)
+
+            assertThat(actual).hasSize(2)
+        }
+    }
```
BraisGabin, line 86: Could you add some tests that show that `==` is never flagged? And other where the type is `String === Any`. and other with `List === List` (without any new pattern)

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ReferentialEquality.kt
```diff
@@ -0,0 +1,77 @@
...
+import io.gitlab.arturbosch.detekt.rules.fqNameOrNull
+import org.jetbrains.kotlin.lexer.KtTokens.EQEQEQ
+import org.jetbrains.kotlin.lexer.KtTokens.EXCLEQEQEQ
+import org.jetbrains.kotlin.psi.KtBinaryExpression
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getType
+
+/**
+ * Kotlin supports two types of equality: structural equality and referential equality. While there are
+ * use cases for both, checking for referential equality for some types (such as `String` or `List` is
```
cortinico, line 22:
```suggestion
 * use cases for both, checking for referential equality for some types (such as `String` or `List`) is
```

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ReferentialEquality.kt
```diff
@@ -0,0 +1,77 @@
...
+ *     val areNotEqual = "aString" !== otherString
+ * </noncompliant>
+ *
+ * <compliant>
+ *     val areEqual = "aString" == otherString
+ *     val areNotEqual = "aString" != otherString
+ * </compliant>
+ */
+@RequiresTypeResolution
+class ReferentialEquality(config: Config) : Rule(config) {
```
cortinico, line 36: Can we rename this rule to something that would provide more information on the issue? Throwing some ideas `ForbiddenReferentialEquality`, `UseStructuralEquality`, `DoNotUseReferentialEquality`, `WrongEqualityOperator`

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ReferentialEquality.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    override val issue = Issue(
+        "ReferentialEquality",
+        Severity.Warning,
+        "",
+        Debt.FIVE_MINS
+    )
+
+    @Configuration(
+        "specifies those types for which referential equality checks are considered a rule violation. " +
```
cortinico, line 46:
```suggestion
        "Specifies those types for which referential equality checks are considered a rule violation. " +
```

cortinico (commented):
Great work 🚀

detekt-core/src/main/resources/default-detekt-config.yml
marschwar: Is there an example where we use globbing other than `PathMatcher`?

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ReferentialEquality.kt
marschwar: How about `AvoidReferentialEquality`?

detekt-core/src/main/resources/default-detekt-config.yml
marschwar: Would be translate the glob to a regex internally to check for matches?

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ReferentialEquality.kt
cortinico: Sounds great 👍

detekt-core/src/main/resources/default-detekt-config.yml
BraisGabin: > Is there an example where we use globbing other than `PathMatcher`?

`excludeAnnotatedFunction` in `FunctionOnlyReturningConstant` is a sample, but a bad one.

> Would be translate the glob to a regex internally to check for matches?

Yes. I think that the replace would be something like:

- `**` -> `.*`
- `*` -> `[^.]`
- `?` -> `.`

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ReferentialEqualitySpec.kt
marschwar: What do you mean by `String === Any` ?

This does not compile.
```kotlin
val b = "s" === 42
```

detekt-core/src/main/resources/default-detekt-config.yml
marschwar: I added a `SimpleGlob` implementation. The configuration is nicer now but it is also more code that we have to maintain and more that needs to be communicated to the users. I am unsure if that is the right way to got. Regular expressions are known to everyone.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/SimpleGlob.kt
```diff
@@ -0,0 +1,23 @@
+package io.gitlab.arturbosch.detekt.api.internal
+
+/**
+ * This simple globbing implementation allows users to define patterns with `?` (any single character)
+ * and `*` (zero or more characters) wildcards.
+ *
```
cortinico, line 6: empty line here

detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/internal/SimpleGlobSpec.kt
```diff
@@ -0,0 +1,197 @@
...
+                it("does not match a other string") {
+                    val actual = subject.matches("ab5")
+                    assertThat(actual).isFalse()
+                }
+            }
+        }
+
+        describe("invalid pattern") {
+            it("fails") {}
+        }
```
cortinico, line 195: This looks like an unfinished test

cortinico (approved):

cortinico (approved):

marschwar:
After rebase:

```
* What went wrong:
Execution failed for task ':detekt-gradle-plugin:test'.
> Multiple build operations failed.
      Metaspace
      Metaspace
      Metaspace
   > Metaspace
   > Metaspace
   > Metaspace
```

Is there anythig I can do about it?

cortinico:
> Is there anythig I can do about it?

Can you try to `./gradlew --stop` and re-run? 
Also I don't think we merged anything that could have caused this.

marschwar:
Sorry, I should have been more specific. It only happens in the CI.

cortinico:
Are we ready to merge this?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/SimpleGlob.kt
```diff
@@ -0,0 +1,22 @@
...
+    fun matches(input: String) = regex.matches(input)
+
+    companion object {
+        fun of(globPattern: String): SimpleGlob {
+            val regex = globPattern
+                .replace("\\", "\\\\")
+                .replace(".", "\\.")
+                .replace("*", ".*")
+                .replace("?", ".")
+                .toRegex()
```
Marcono1234, line 18: It looks like this does not cover all meta characters. For example `[`, `]`, `$`, `^`, `(`, `)` (and probably more) would still be interpreted as regex meta characters.
This is likely not an issue for valid class names (since they should not contain any of these characters?), but users might misuse these characters which could become a maintenance burden.

A saner (but more complex) implementation might be to use `Regex.escape` for all the literal pieces of the glob pattern.
For example (not tested):
```kotlin
fun buildRegex(globPattern: String): Regex {
    val patternStringBuilder = StringBuilder()
    var nextStart = 0
    
    globPattern.forEachIndexed { index, char ->
        val replacement = when(char) {
            '*' -> ".*"
            '?' -> "."
            else -> null
        }
        
        if (replacement != null) {
            // Append escaped previous text, if any
            if (nextStart < index) {
            	val escaped = Regex.escape(globPattern.substring(nextStart, index))
            	patternStringBuilder.append(escaped)
            }
            // Append meta character replacement
            patternStringBuilder.append(replacement)
            nextStart = index + 1
        }
    }
    
    // Append escaped remainder, if any
    if (nextStart < globPattern.length) {
        val escaped = Regex.escape(globPattern.substring(nextStart))
        patternStringBuilder.append(escaped)
    }
    
    return Regex(patternStringBuilder.toString())
}
```

cortinico: > but users might misuse these characters which could become a maintenance burden.

Is this an edge case or are there a real examples to support this? I'm trying to understand from which prospective you're mentioning users misusing those characters

Marcono1234: It is probably an edge case.

It looks like the intention here was to create a simple glob pattern. However, a user might by accident use a regex pattern such as `java.lang.(String|Integer)` (which the current implementation permits) and then expect that this works in the future as well. This could become a maintenance burden because the intention here was not to support something like this.

But this is probably negligible, so feel free to ignore.

PR #3932, opened 2021-07-02
Verify that the generated default config can be loaded

marschwar:
[Comment](https://github.com/detekt/detekt/pull/3920#issuecomment-871373405) by @cortinico in #3920 
> We should also look into why those kind of bugs are happening whenever we're about to release a new version (and not before merging changes).

With this the pipeline will run the detekt cli with default configuration only. This should in my opinion at least verify, that the default configuration is usable.


marschwar:
Something is weird about the CI. This build should have failed since this branch does not yet contain the fix from #3920. 

When I run the steps locally, I can reproduce the error @chao2zhang mentioned in the PR. 

Where does this cli.jar come from? What worries me a little bit is the `> Task :detekt-cli:shadowJar FROM-CACHE` in the build log. Can someone please enlighten me.

schalkms (approved):

chao2zhang:
> What worries me a little bit is the `> Task :detekt-cli:shadowJar FROM-CACHE` in the build log

I imagine because we have enabled Github action caches, as there is no code change in this PR, `shadowJar`'s output can be extracted from the Gradle cache.

chao2zhang (approved):

marschwar:
Please do not merge just yet. I am looking into different options. I will change the title back when I think it is ready.

marschwar:
My latest try was to add a case to the `RunnerSpec` which (again) locally fails with the expected exception but does not seem to work in the CI. 

Can anyone please tell me why the task `detekt-cli:test` is not run in the pre-merge checks. I can see `Task :detekt-api:test FROM-CACHE` and `Task :detekt-core:test FROM-CACHE`. From the dry run I would expect the cli tests somewhere in between but it is never executed. Is this by design?

cortinico:
> Please do not merge just yet. I am looking into different options. I will change the title back when I think it is ready.

Converting to draft, so we don't accidentally merge.

> Can anyone please tell me why the task `detekt-cli:test` is not run in the pre-merge checks. I can see `Task :detekt-api:test FROM-CACHE` and `Task :detekt-core:test FROM-CACHE`. From the dry run I would expect the cli tests somewhere in between but it is never executed. Is this by design? Since I added a test case, I would expect those tests to be executed.

Sounds like a bug in our CI setup. Ideally those tasks should not result in a `FROM-CACHE` especially if there is a change in the test source.

Are you able to run them locally successfully?

marschwar:
> Sounds like a bug in our CI setup. Ideally those tasks should not result in a FROM-CACHE especially if there is a change in the test source.
I agree that there is something fishy about the CI setup. What puzzles me is the fact, that `detekt-cli:test` does not seem to be part of the execution graph.

> Are you able to run them locally successfully?
When I run `./gradlew build moveJarForIntegrationTest -x detekt` locally, the tests are executed and fail as expected.

marschwar:
Finally the exception [has occurred in the CI](https://github.com/detekt/detekt/runs/2991165639?check_suite_focus=true)

```
       Caused by:
        java.lang.AssertionError: 
        Expecting code not to raise a throwable but caught
          "java.lang.IllegalStateException: Analyzing /home/runner/work/detekt/detekt/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt led to an exception. 
        The original exception message was: class java.lang.Integer cannot be cast to class java.lang.String (java.lang.Integer and java.lang.String are in module java.base of loader 'bootstrap')
```

Issue #3778, opened 2021-05-15
RFC: Test with ProjectBuilder instead of Gradle plugin "dry run"


3flex:
## Expected Behavior
Tests are as fast as they can be.

## Current Behavior
The "dry run" tests aren't integration tests but use Gradle's TestKit, which is slow. The `:detekt-gradle-plugin:test` task takes longer to execute than any other task in the build.

## Context
Gradle's [ProjectBuilder](https://docs.gradle.org/current/javadoc/org/gradle/testfixtures/ProjectBuilder.html) is a speedier alternative to TestKit and can be used for tests that don't execute Gradle tasks.

The "dry run" tests are just checking that when certain options are set on the plugin that the flags passed to the CLI are as expected. A slight refactoring in the plugin combined with switching to ProjectBuilder for the dry run tests will speed things up a lot - individual tests that take 25-30 seconds with TestKit take 15 with ProjectBuilder on my machine.

The test results should also be picked up by JaCoCo so we can measure code coverage in the plugin.

An example (kotlin-dsl was used as the DSL can be used in the `project.run {}` block):

```kotlin
        it("does the test") {
            val project = ProjectBuilder.builder().build()

            project.run {
                apply<KotlinPluginWrapper>()
                apply<DetektPlugin>()
                repositories {
                    mavenCentral()
                    mavenLocal()
                }
                configure<DetektExtension> {
                    allRules = true
                    reports.txt.enabled = false
                }
            }

            project.getTasksByName("detektMain", true) // trigger project evaluation

            val detektTask = project.tasks.getByPath("detektMain") as Detekt
            val argumentString = detektTask.arguments.flatMap(CliArgument::toArgument).joinToString(" ")

            assertThat(argumentString).contains("--report xml:")
            assertThat(argumentString).contains("--report sarif:")
            assertThat(argumentString).doesNotContain("--report txt:")
        }
```

The DSL used won't exactly match what users will use, but for these tests I don't think that's an issue. `apply<KotlinPluginWrapper>()` is functionally identical to `apply("org.jetbrains.kotlin.jvm")`, and applying plugins using `apply<>()` is valid and [documented](https://docs.gradle.org/current/userguide/plugins.html#sec:applying_binary_plugins), even if it's not common. It's the same case for `configure<DetektExtension>`.

[...]

PR #3938, opened 2021-07-05
Increase memory available to gradle integration test daemon

marschwar:
Trying to figure out why there is a OutOfMemoryError on the CI


cortinico (approved):

cortinico:
Do we plan to merge this?

marschwar:
Just increasing the memory does not feel right, but it seems like the only option for the moment as it does not seem to be possible to run the gradle test kit runner without daemon.

Maybe this can bring some live to the discussion in #3778.

marschwar:
> Do we plan to merge this?

But to answer your question. Yes, I would merge this for the time beeing.

schalkms (approved):

PR #3940, opened 2021-07-06
UndocumentedPublicProperty and UndocumentedPublicFunction should include objects

anthonycr:
# Summary
Given an `object` with an undocumented public function or property, I would expect the `UndocumentedPublicFunction` rule or `UndocumentedPublicProperty` rule to fail the build. However, the behavior of the rules is inconsistent in regards to `objects`.

The following code is failed when the rules are enabled and an object is nested in a class, as expected:
```kotlin
class Test {
  object InnerObject {
    // Fails, expected
    fun noComment1() {}
    // Fails, expected
    val a = 1
  }
}
```

However, if the object is promoted to a top level object, then the rules will not fail the function or property.
```kotlin
object Test {
  // Passes, unexpected
  fun noComment1() {}
  // Passes, unexpected
  val a = 1
}
```

## Changes
I discovered that the `containingClass()` function used in both rules to determine if the function or property is part of a public class returns null if the function or property is in an `object`. We can change this behavior to check if the `object` is public or not by replacing the usage of `containingClass()` with `containingClassOrObject`. In this way, we can treat an undocumented function or property in an `object` the same as one in a regular `class`.

I changed both `UndocumentedPublicFunction` and `UndocumentedPublicProperty` to utilize `containingClassOrObject` instead of `containingClass()`. I also added test cases for public and private `object` cases as well as a nested `object` test case. For the test cases I added to the `UndocumentedPublicPropertySpec`, I renamed two of the existing test cases with ambiguous names to avoid confusion with the test cases I added.

### DIFF
--- a/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunction.kt
+++ b/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunction.kt
@@ -9,7 +9,7 @@ import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.rules.isPublicNotOverridden
 import org.jetbrains.kotlin.psi.KtNamedFunction
-import org.jetbrains.kotlin.psi.psiUtil.containingClass
+import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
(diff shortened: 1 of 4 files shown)


cortinico:
Thanks for the fix 🙏 That's really appreciated.

cortinico (approved):

schalkms (approved):
Does the rule also flag interfaces with this implementation?

anthonycr:
@schalkms yes, interfaces were being flagged before with `containingClass()` and are being flagged with this updated implementation, as the interface is returned as the encapulating "class" by the `containingClassOrObject` property. I just added a test case for interfaces to both the property and function test specs as it doesn't look like that case was explicitly being tested for yet.

PR #3947, opened 2021-07-11
Exclude buildSrc from InvalidPackageDeclaration

### LINKED ISSUE #3946: Ignore InvalidPackageDeclaration in buildSrc folder
## Expected Behavior

Rule is not triggered.

## Observed Behavior

Rule is triggered.

## Steps to Reproduce

Run on this file: https://github.com/shadowsocks/shadowsocks-android/blob/f5bc88b83ebbd4d2941a65bea2eaa565f2072db4/buildSrc/src/main/kotlin/Helpers.kt#L1

## Context

This file should be in root package.

## Your Environment

* Version of detekt used: v1.17.1
* Operating System and version: Linux Mint 20
* Link to your project (if it's a public repository): See above

Mygod:
Fixes #3946.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -443,7 +443,7 @@ naming:
     ignoreOverridden: true
   InvalidPackageDeclaration:
     active: false
-    excludes: ['**/*.kts']
+    excludes: ['**/*.kts', 'buildSrc/**']
     rootPackage: ''
   MatchingDeclarationName:
     active: true


marschwar (changes requested):
Unfortunately it is not as easy as that. The file `default-detekt-config.yml` is auto generated and your changes would therefor be overwritten. The exclusions are configured in `detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/rulesetpage/Exclusion.kt` but I don't think this should be changed there as this will be applied to all rules.

You can still put this in your projects config yaml file.

schalkms (approved):
It’s a sensible default value for me! Thanks for the PR!

cortinico:
Playing devil's advocate here, but isn't a `buildSrc` exclusion just another platform specific customization (in this case, a Gradle technicality).

See https://github.com/detekt/detekt/discussions/3894 for further discussions on this topic.

I'd be happy to hear other maintainers opinion on this.

BraisGabin:
I agree with @cortinico.

But I think that the problem here is that a file without package should not trigger this rule. I think that we should split this rule in two different rules.

cortinico:
> But I think that the problem here is that a file without package should not trigger this rule

Imho a file without package AND placed in (`.../src/main/java/` OR `.../src/main/kotlin/`) should not trigger this rule

marschwar:
> I think that we should split this rule in two different rules. 

I agree with @BraisGabin that this rule should probably be split into an `MissingPackageDeclaration` and `InvalidPackageDeclaration` to avoid this problem.


> Imho a file without package AND placed in (.../src/main/java/ OR .../src/main/kotlin/) should not trigger this rule

Agreed. Maybe the `MissingPackageDeclaration` should be configured to exclude files in `src/*/kotlin` and `src/*/java`?

BraisGabin:
> > Imho a file without package AND placed in (.../src/main/java/ OR .../src/main/kotlin/) should not trigger this rule
> 
> Agreed. Maybe the `MissingPackageDeclaration` should be configured to exclude files in `src/*/kotlin` and `src/*/java` and `InvalidPackageDeclaration` should skip all files without a package declaration.

I agree with @cortinico if we don't split the rule. But if we split the rule then `MissingPackageDeclaration` should trigger all the files without package. Even if they are placed in the correct folder. It is a bad practice to have classes and functions without package. You can use them, they work, but you should not use them. Except `.kts`.

BraisGabin:
To sumarize: I think that we agree that this is not the way to fix this issue. And it seems to be two options: Allow missing packages in the root of a source set and/or split this rule. How do we proceed? I vote to split and move this discussion to that new rule.

cortinico:
> I agree with @cortinico if we don't split the rule. But if we split the rule then `MissingPackageDeclaration` should trigger all the files without package. Even if they are placed in the correct folder. It is a bad practice to have classes and functions without package. You can use them, they work, but you should not use them. Except `.kts`.

Totally agree with what said here. Let's split this rule in two and fix `InvalidPackageDeclaration` as a false positive for files without package declaration.

BraisGabin:
Closing in favor of #3957. Thanks for the proposal @Mygod :)

PR #3951, opened 2021-07-12
Rename property in gradle plugin extension from input to source

### LINKED ISSUE #3918: Rename `input` property to `source` in the DetektExtension of the gradle plugin
### Discussed in https://github.com/detekt/detekt/discussions/3917

<div type='discussions-op-text'>

<sup>Originally posted by **peter-budo** June 28, 2021</sup>
... (post shortened)...

However when I try to assemble build I do run in following message
```
* What went wrong:
A problem occurred evaluating script.
> A problem occurred evaluating script.
   > Could not set unknown property 'input' for task ':app:detektAll' of type io.gitlab.arturbosch.detekt.Detekt.

```

Can anybody please explain me what I am doing wrong?</div>

In my opinion we should consider to rename the `input` property to `source` in the extension as well to avoid confusion. Since this will break everyone's builds the change cannot be done before 2.0, but we could add the source property and change the documentation already while supporting both.

marschwar:
This closes #3918 

Since switching to  `org.gradle.api.tasks.SourceTask`, there is a difference how users would specify the input to be checked using the detekt extension vs. creating a custom task which seemed to have confused users.

```groovy
detekt {
    input = files(
        "src/main/kotlin",
        "gensrc/main/kotlin"
    )
}
```

vs.

```groovy
tasks.register(name: myDetekt, type: io.gitlab.arturbosch.detekt.Detekt) {
   source = files("src/main/kotlin", "src/test/kotlin")
}
```

### DIFF
--- a/docs/_posts/2019-03-03-configure-detekt-on-root-project.md
+++ b/docs/_posts/2019-03-03-configure-detekt-on-root-project.md
@@ -41,7 +41,7 @@ repositories {
 }
 
 detekt {
-    input = files(rootProject.rootDir)
+    source = files(rootProject.rootDir)
     buildUponDefaultConfig = true
 }
 ```
--- a/docs/pages/gettingstarted/gradle.md
+++ b/docs/pages/gettingstarted/gradle.md
@@ -191,7 +191,7 @@ detekt {
     
     // The directories where detekt looks for source files. 
     // Defaults to `files("src/main/java", "src/test/java", "src/main/kotlin", "src/test/kotlin")`.
-    source = files(
+    input = files(
         "src/main/kotlin",
         "gensrc/main/kotlin"
     )
@@ -277,7 +277,7 @@ detekt {
     
     // The directories where detekt looks for source files. 
     // Defaults to `files("src/main/java", "src/test/java", "src/main/kotlin", "src/test/kotlin")`.
-    source = files("src/main/java", "src/main/kotlin")     
+    input = files("src/main/java", "src/main/kotlin")     
     
     // Builds the AST in parallel. Rules are always executed in parallel. 
     // Can lead to speedups in larger projects. `false` by default.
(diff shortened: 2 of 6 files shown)


schalkms (commented):
I favor this change compared to the old behavior.

BraisGabin (approved):

docs/_posts/2019-03-03-configure-detekt-on-root-project.md
```diff
@@ -41,7 +41,7 @@ repositories {
 }
 
 detekt {
-    input = files(rootProject.rootDir)
+    source = files(rootProject.rootDir)
```
cortinico, line 44: This is a blogpost and should probably not be udpated

docs/pages/gettingstarted/gradle.md
```diff
@@ -191,7 +191,7 @@ detekt {
     
     // The directories where detekt looks for source files. 
     // Defaults to `files("src/main/java", "src/test/java", "src/main/kotlin", "src/test/kotlin")`.
-    input = files(
+    source = files(
```
cortinico, line 194: I'm unsure about publishing this specific change right now. Maybe the doc update should be delayed till we release a new version that supports `source` in the extension? Otherwise users will find docs that are effectively wrong till a newer version of `detekt-gradle-plugin` is released.

cortinico (approved):

docs/_posts/2019-03-03-configure-detekt-on-root-project.md
marschwar: That was my initial thought as well but saw that it had been updated three times after original creation. Either way is fine for me.

docs/pages/gettingstarted/gradle.md
marschwar: Do we publish the docs automatically after push on main? If that is the case I agree that we should delay this change.

cortinico: > Do we publish the docs automatically after push on main? If that is the case I agree that we should delay this change.

Yes. Docs are updated whenever a PR is merged

cortinico (approved):

cortinico:
Are we ok to merge this?

PR #3962, opened 2021-07-19
UndocumentedPublicFunction: don't report when nested class is inside not public class

### LINKED ISSUE #3959: UndocumentedPublicFunction for functions of an interface inside an internal interface.
I checked for existing (and closed) issues, but there wasn't anything that matched exactly the problem I want to report. 
I think this only started happening in more recent versions of detekt, but I'm not sure when exactly. I am currently using `1.17.1`

Detekt should not raise `UndocumentedPublicFunction` for `functions` of an `interface` which is inside an `internal interface`.

## Context
I use this pattern commonly for defining dagger2 components, for example: 
```
@Component(...)
internal interface MyComponent {
    
    @Component.Builder
    interface Builder {
        fun xyz(...): ... 
    }
}
```
In this scenario, `xyz()` is raising `UndocumentedPublicFunction`.
It's not possible to make anything inside an interface explicitly internal, even though it obviously is internal if the outer interface is internal - that's true for both inner interface and its functions. There is no other way than to suppress the rule.

t-kameyama:
Fixes #3959

### DIFF
--- a/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunction.kt
+++ b/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunction.kt
@@ -44,6 +44,6 @@ class UndocumentedPublicFunction(config: Config = Config.empty) : Rule(config) {
 
     private fun KtNamedFunction.shouldBeDocumented() =
         (isTopLevel || containingClassOrObject?.isPublic == true) &&
-            parents.filterIsInstance<KtClassOrObject>().none { !it.isPublic } &&
+            parents.filterIsInstance<KtClassOrObject>().all { it.isPublic } &&
             isPublicNotOverridden()
 }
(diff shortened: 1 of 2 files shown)


detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunction.kt
```diff
@@ -41,5 +43,7 @@ class UndocumentedPublicFunction(config: Config = Config.empty) : Rule(config) {
     }
 
     private fun KtNamedFunction.shouldBeDocumented() =
-        (isTopLevel || containingClassOrObject?.isPublic == true) && isPublicNotOverridden()
+        (isTopLevel || containingClassOrObject?.isPublic == true) &&
+            parents.filterIsInstance<KtClassOrObject>().none { !it.isPublic } &&
```
cortinico, line 47:
```suggestion
            parents.filterIsInstance<KtClassOrObject>().all { it.isPublic } &&
```

Perhaps is more readable?

cortinico (approved):
Thanks for the fix 🙏

detekt-rules-documentation/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunctionSpec.kt
```diff
@@ -130,5 +130,33 @@ class UndocumentedPublicFunctionSpec : Spek({
             """
             assertThat(subject.compileAndLint(code)).isEmpty()
         }
+
+        context("nested class") {
```
schalkms, line 134: A test case for objects could be helpful as well, since another rule reported a false positive.

schalkms (approved):

schalkms:
Thanks for fixing this false positive! 👍

BraisGabin (approved):

PR #3964, opened 2021-07-19
Create rule set configurations in a safe way

### LINKED ISSUE #3816: Configure rule sets in a safe way
## Expected Behavior
The configuration of rule sets should be as explicit and safe as the rule configuration. It should be impossible for the documentation to diverge from the actual implementation.

## Current Behavior
The `RuleSetProvider` may describe configuration properties in the kdoc section. That information is used to generate the rule set documentation as well as the `detekt-config-default.yml`. 

The configured value can be accessed within a rule of that rule set by `ruleSetConfig.valueOrDefault("propertyname", false)`. There is a lot of implicit knowledge here.

## Context
#3637 added support for rule configurations via annotation making it impossible to generate a documentation with misspelled configuration options and outdated default values. It would be great to have the same for rule sets.

chao2zhang:
This resolves #3816.

Created `ruleSetConfig` function to move `RuleSetProviders`' configuration from kdoc to actual annotation class.
Note that list as the default value is not supported until we actually have a usecase.


detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
```diff
@@ -69,13 +70,12 @@ class RuleSetProviderVisitor : DetektVisitor() {
         super.visitSuperTypeList(list)
     }
 
-    override fun visitClassOrObject(classOrObject: KtClassOrObject) {
-        description = classOrObject.docComment?.getDefaultSection()?.getContent()?.trim() ?: ""
-        if (classOrObject.isAnnotatedWith(ActiveByDefault::class)) {
-            defaultActivationStatus = Active(since = classOrObject.firstAnnotationParameter(ActiveByDefault::class))
+    override fun visitClass(ktClass: KtClass) {
```
schalkms, line 73: May I ask you to explain this change. I'm not sure why this is required to resolve the linked issue.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
```diff
@@ -69,13 +70,12 @@ class RuleSetProviderVisitor : DetektVisitor() {
...
+        description = ktClass.docComment?.getDefaultSection()?.getContent()?.trim() ?: ""
+        if (ktClass.isAnnotatedWith(ActiveByDefault::class)) {
+            defaultActivationStatus = Active(since = ktClass.firstAnnotationParameter(ActiveByDefault::class))
         }
-        configuration.addAll(classOrObject.parseConfigurationTags())
-        super.visitClassOrObject(classOrObject)
+        super.visitClassOrObject(ktClass)
```
schalkms, line 78: The overridden function is named `KtClass` but the base method of `KtClassorObject` is called.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
```diff
@@ -107,4 +120,16 @@ class RuleSetProviderVisitor : DetektVisitor() {
...
+    companion object {
+
+        private const val DOUBLE_QUOTE = '"'
+
+        private fun formatDefaultValue(defaultValueText: String): String =
+            if (defaultValueText.startsWith(DOUBLE_QUOTE) && defaultValueText.endsWith(DOUBLE_QUOTE)) {
+                "'${defaultValueText.removeSurrounding("$DOUBLE_QUOTE")}'"
+            } else {
+                defaultValueText
+            }
```
schalkms, line 133: Is this function duplicated? I think I have seen this in another class.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
chao2zhang: Fixed.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
chao2zhang: `visitClassOrObject` will be invoked for the class, as well as the companion object. The adversity goes that invoking `visitClassOrObject` on companion object will override the `description`.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
chao2zhang: The similar functionality exists in `ConfigurationCollector`. However, the difference is that in `RuleSetProviderCollector`, we cannot easily extract the type information. Therefore I am using a raw check for `startsWith()` and `endsWith()` to check if the default value is a string.

schalkms: Thanks for the explanation -> resolved for me

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
schalkms: Thanks for the explanation -> resolved for me

schalkms (approved):

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleSetProviderCollector.kt
```diff
@@ -69,13 +70,12 @@ class RuleSetProviderVisitor : DetektVisitor() {
...
-        if (classOrObject.isAnnotatedWith(ActiveByDefault::class)) {
-            defaultActivationStatus = Active(since = classOrObject.firstAnnotationParameter(ActiveByDefault::class))
+    override fun visitClass(ktClass: KtClass) {
+        description = ktClass.docComment?.getDefaultSection()?.getContent()?.trim() ?: ""
+        if (ktClass.isAnnotatedWith(ActiveByDefault::class)) {
+            defaultActivationStatus = Active(since = ktClass.firstAnnotationParameter(ActiveByDefault::class))
         }
-        configuration.addAll(classOrObject.parseConfigurationTags())
-        super.visitClassOrObject(classOrObject)
+        super.visitClass(ktClass)
```
marschwar, line 78: Similar to the `RuleVisitor`I think we should do
```kotlin
if (ktClass.hasConfigurationKDocTag()) {
    throw InvalidDocumentationException(
        "Configuration of rule set $name is invalid. Rule configuration via KDoc tag is no longer supported. " +
                "Use rule set config delegate instead."
    )
}
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/RuleSetConfigProperty.kt
```diff
@@ -0,0 +1,29 @@
+package io.gitlab.arturbosch.detekt.api.internal
+
+import io.gitlab.arturbosch.detekt.api.Config
+import kotlin.properties.ReadOnlyProperty
+import kotlin.reflect.KProperty
+
+fun <T : Any> ruleSetConfig(defaultValue: T): ReadOnlyProperty<Any?, RuleSetConfigProperty<T>> =
```
marschwar, line 7: Could you please add a test.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/RuleSetConfigProperty.kt
```diff
@@ -0,0 +1,29 @@
...
+            RuleSetConfigProperty(property.name, defaultValue).also {
+                _value = it
+            }
+        }
+}
+
+class RuleSetConfigProperty<T : Any>(val key: String, val defaultValue: T)
+
+fun <T : Any> Config.value(ruleSetConfigProperty: RuleSetConfigProperty<T>): T =
+    valueOrDefault(ruleSetConfigProperty.key, ruleSetConfigProperty.defaultValue)
```
marschwar, line 29: nit: This is probably a matter of taste, but I think I prefer the `value` function on the property instead of the config.

```suggestion
class RuleSetConfigProperty<T : Any>(private val key: String, private val defaultValue: T) {
    fun value(config: Config): T = config.valueOrDefault(key, defaultValue)
}
```

This would change the caller to
```kotlin
protected val isAndroid
    get() = FormattingProvider.android.value(ruleSetConfig)
```

marschwar (approved):
Thank you for picking this up. I really like your solution.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/RuleSetConfigProperty.kt
chao2zhang: Good reminder.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/RuleSetConfigProperty.kt
chao2zhang: I have no preference so your advice rules.

cortinico (approved):

BraisGabin:
Is there something blocking us to merge this? 3 approvals and all checks are green seems good to me.

PR #3972, opened 2021-07-25
 UnderscoresInNumericLiterals acceptableDecimalLength is off by one

### LINKED ISSUE #3860: UnderscoresInNumericLiterals acceptableDecimalLength is off by one
If I set `acceptableDecimalLength` to 4, then `2021` is flagged as violating the rule.

The description is `Length under which decimal base 10 literals are not required to have underscores`, which means the rule is working correctly. The issue is that the name indicates that anything with this length will be acceptable, when in reality the rule is that anything less than this length will be acceptable.

BraisGabin:
This rule was using the word "decimal" everywhere. It's correct to call a base 10 number "decimal" but it is saying the same thing twice. And the problem with "decimal" is that it have a lot of meanings so it can be missleading. For that reason I changed the documentation a bit removing the word "decimal" (first commit).

After that the only point where "decimal" was used is in the configuration so to avoid the breaking change and because it also improves "readability" I renamed "acceptableDecimalLength" to "acceptableLenght". All this rule is about base 10 numbers so there's is no need to point that out in the configuration too.

But this is a Draft an it's not a ready to merge PR because I found a missing feature in our new configuration system:

```
private val acceptableLength: Int by configWithFallback("acceptableDecimalLength", 4)
```

We can't use the value of `acceptableDecimalLength` as the value of `acceptableLenght` because it would be off by one. Should we care about this? We were talking about adding a "breaking change" to fix this so, maybe, we should not care.

Fix #3860

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiteralsSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiteralsSpec.kt
@@ -8,6 +8,7 @@ import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
 
 private const val ACCEPTABLE_DECIMAL_LENGTH = "acceptableDecimalLength"
+private const val ACCEPTABLE_LENGTH = "acceptableLength"
 
 class UnderscoresInNumericLiteralsSpec : Spek({
 
@@ -19,7 +20,14 @@ class UnderscoresInNumericLiteralsSpec : Spek({
             assertThat(findings).isEmpty()
         }
 
-        it("should be reported if acceptableDecimalLength is 4") {
+        it("should be reported if acceptableLength is 3") {
+            val findings = UnderscoresInNumericLiterals(
+                TestConfig(mapOf(ACCEPTABLE_LENGTH to "3"))
+            ).compileAndLint(code)
+            assertThat(findings).isNotEmpty
+        }
+
+        it("should be reported if deprecated acceptableDecimalLength is 4") {
             val findings = UnderscoresInNumericLiterals(
                 TestConfig(mapOf(ACCEPTABLE_DECIMAL_LENGTH to "4"))
             ).compileAndLint(code)
@@ -44,9 +52,9 @@ class UnderscoresInNumericLiteralsSpec : Spek({
             assertThat(findings).isNotEmpty
         }
 
-        it("should not be reported if acceptableDecimalLength is 8") {
+        it("should not be reported if acceptableLength is 7") {
             val findings = UnderscoresInNumericLiterals(
-                TestConfig(mapOf(ACCEPTABLE_DECIMAL_LENGTH to "8"))
+                TestConfig(mapOf(ACCEPTABLE_LENGTH to "7"))
             ).compileAndLint(code)
             assertThat(findings).isEmpty()
         }
@@ -60,9 +68,9 @@ class UnderscoresInNumericLiteralsSpec : Spek({
             assertThat(findings).isEmpty()
         }
 
-        it("should be reported if acceptableDecimalLength is 4") {
+        it("should be reported if acceptableLength is 3") {
             val findings = UnderscoresInNumericLiterals(
-                TestConfig(mapOf(ACCEPTABLE_DECIMAL_LENGTH to "4"))
+                TestConfig(mapOf(ACCEPTABLE_LENGTH to "3"))
             ).compileAndLint(code)
             assertThat(findings).isNotEmpty
         }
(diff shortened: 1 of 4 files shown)


BraisGabin:
That api doesn't allow me to convert from one value to the other, I mean, do the - 1

marschwar:
:thinking: You could do `val acceptableDecimalLength: Int by config(5) { it - 1 }`.  So whenever the fallback property is used, the value is reduced by one. (I have not tested that yet).

BraisGabin:
> 🤔 You could do `val acceptableDecimalLength: Int by config(5) { it - 1 }`. So whenever the fallback property is used, the value is reduced by one. (I have not tested that yet).

Good point! That have sense.

BraisGabin:
Rebased and now it uses the new version of `configWithFallback`.

marschwar:
Please regenerate the default config to use the the property name.

BraisGabin:
> Please regenerate the default config to use the the property name.

And not only the default config. The deprecation list too! We merged the PR just in time 😂.

> Should we also update the test to use the new config property and keep only one or two with the deprecated value?

You are right! Done :)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiteralsSpec.kt
```diff
@@ -44,9 +52,9 @@ class UnderscoresInNumericLiteralsSpec : Spek({
             assertThat(findings).isNotEmpty
         }
 
-        it("should not be reported if acceptableDecimalLength is 8") {
+        it("should not be reported if acceptableLength is 8") {
             val findings = UnderscoresInNumericLiterals(
-                TestConfig(mapOf(ACCEPTABLE_DECIMAL_LENGTH to "8"))
+                TestConfig(mapOf(ACCEPTABLE_LENGTH to "8"))
```
marschwar, line 57: should both be "7" now

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiteralsSpec.kt
```diff
@@ -101,9 +109,9 @@ class UnderscoresInNumericLiteralsSpec : Spek({
             assertThat(findings).isNotEmpty
         }
 
-        it("should not be reported if ignored acceptableDecimalLength is 8") {
+        it("should not be reported if ignored acceptableLength is 8") {
             val findings = UnderscoresInNumericLiterals(
-                TestConfig(mapOf(ACCEPTABLE_DECIMAL_LENGTH to "8"))
+                TestConfig(mapOf(ACCEPTABLE_LENGTH to "8"))
```
marschwar, line 114: should both be "7" now

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiteralsSpec.kt
```diff
@@ -300,9 +308,9 @@ class UnderscoresInNumericLiteralsSpec : Spek({
             assertThat(findings).isNotEmpty
         }
 
-        it("should not be reported if acceptableDecimalLength is 6") {
+        it("should not be reported if acceptableLength is 6") {
             val findings = UnderscoresInNumericLiterals(
-                TestConfig(mapOf(ACCEPTABLE_DECIMAL_LENGTH to "6"))
+                TestConfig(mapOf(ACCEPTABLE_LENGTH to "6"))
```
marschwar, line 313: should both be "5" now

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiteralsSpec.kt
```diff
@@ -345,9 +353,9 @@ class UnderscoresInNumericLiteralsSpec : Spek({
             assertThat(findings).isNotEmpty
         }
 
-        it("should not be reported if acceptableDecimalLength is 8") {
+        it("should not be reported if acceptableLength is 8") {
             val findings = UnderscoresInNumericLiterals(
-                TestConfig(mapOf(ACCEPTABLE_DECIMAL_LENGTH to "8"))
+                TestConfig(mapOf(ACCEPTABLE_LENGTH to "8"))
```
marschwar, line 358: should both be "7" now

marschwar (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiteralsSpec.kt
BraisGabin: I didn't update those because they were not "adjusted" before. But you are right, is better to update all of those too. Done :)

PR #3973, opened 2021-07-25
Allow factory function names - fix #1639

### LINKED ISSUE #1639: FunctionNaming should not report when the function name is identical to the type of the result.
Currently, [FunctionNaming](https://arturbosch.github.io/detekt/naming.html#functionnaming) reports names which are valid according to the [Kotlin Coding Conventions](https://kotlinlang.org/docs/reference/coding-conventions.html)

I do not expect any issue report for the following code block, since it comes straight from the official conventions:

```kotlin
abstract class Foo

class FooImpl : Foo()

fun Foo(): Foo { return FooImpl() }
```

Source: https://kotlinlang.org/docs/reference/coding-conventions.html#function-names

But, currently, it reports a `FunctionNaming` because `fun Foo(): Foo` starts with a capital.

## Context

As stated in the conventions (see link above): 
>factory functions used to create instances of classes can have the same name as the class being created:

And I personally find that very useful. If we read `val x = Foo()`, we immediately understand the important point: "It creates an instance of type `Foo`". Whether `Foo()` it is an actual constructor or a factory function, is an implementation detail, and the reader does not have to care.

BraisGabin:
The PR #1668 tried to fix this issue but it seems that it forgot to add the `.isEmpty()` so it wasn't actually fixed.

I would like the review of @schalkms because it was the one tring to fixing this so I can be missing something.

### DIFF
--- a/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionNaming.kt
+++ b/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionNaming.kt
@@ -54,7 +54,7 @@ class FunctionNaming(config: Config = Config.empty) : Rule(config) {
         val functionName = function.nameIdentifier?.text ?: return
         if (!function.isContainingExcludedClassOrObject(excludeClassPattern) &&
             !functionName.matches(functionPattern) &&
-            functionName != function.typeReference?.name
+            functionName != function.typeReference?.text
         ) {
             report(
                 CodeSmell(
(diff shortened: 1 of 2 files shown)


schalkms (approved):
Thanks for the fix!

schalkms:
Such things are tricky to catch but easy to create. Ideally, this code does not compile. It makes no sense.
Do you happen to know whether other similar instances exist in detekt's code base?

```
assertThat(FunctionNaming(config).compileAndLint(code)) // missing .is... check
```

BraisGabin:
I think that if we merge #3974 and fix #3975 (I just created both) detekt will spot this mistakes. It is not a compilation error but I think that with an issue in detekt is enough.

schalkms:
Good idea!

PR #3977, opened 2021-07-25
ClassNaming: Don't treat Kotlin syntax ` as part of class name

### LINKED ISSUE #2656: ClassNaming rule treating Kotlin syntax ` as part of class name

BraisGabin:
This tries to fix #2656. But I find this a bit controversial:

Pros:

- It's true that `` ` ``  is not part of the name of a class (or any other entity in the code).

Cons:

- `FunctionNaming` is using this "error" to provide the feature that only allows any type of function name if it uses backticks. That's like that because in the tests we want to allow that kind of naming in the functions.

I vote to merge because I think that the one that is wrong is `FunctionNaming`. But fixing `FuctionNaming` is not easy. A way to fix it would be to allow multiple instances of the same rule with different configurations. This way we could have an instance that checks FunctionNaming in production code and another for test code.

Maybe a solution is to exclude `FunctionNaming` from test source sets and don't allow backtick at all.

### DIFF
--- a/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ClassNaming.kt
+++ b/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ClassNaming.kt
@@ -32,7 +32,7 @@ class ClassNaming(config: Config = Config.empty) : Rule(config) {
     private val classPattern: Regex by config("[A-Z][a-zA-Z0-9]*") { it.toRegex() }
 
     override fun visitClassOrObject(classOrObject: KtClassOrObject) {
-        if (!classOrObject.identifierName().matches(classPattern)) {
+        if (!classOrObject.identifierName().removeSurrounding("`").matches(classPattern)) {
             report(
                 CodeSmell(
                     issue,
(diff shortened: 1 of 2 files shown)


detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ClassNaming.kt
```diff
@@ -47,3 +47,11 @@ class ClassNaming(config: Config = Config.empty) : Rule(config) {
         const val CLASS_PATTERN = "classPattern"
     }
 }
+
+private fun String.cleanBackticks(): String {
```
chao2zhang, line 51: We could simplify use `String.removeSurrounding("`")`

chao2zhang (approved):

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ClassNaming.kt
BraisGabin: Didn't know abot that one! Thanks!

cortinico (approved):

PR #3979, opened 2021-07-26
IgnoredReturnValue: fix false negative when annotation is on the class

### LINKED ISSUE #3975: IgnoredReturnValue false negative when annotation is on the class
If I have this code:

```
org.assertj.core.api.Assertions.assertThat(somerhing)
```

`IgnoredReturnValue` should raise an issue because `org.assertj.core.api.Assertions` has the `@CheckReturnValue` annotation on it. But it doesn't.

The issue is that in the rule we just check for the annotation in the method but not in its class.

## Context
I spotted this issue working on #3973 and #3974

Related code:

- https://github.com/assertj/assertj-core/blob/main/src/main/java/org/assertj/core/api/Assertions.java#L146-L147
- https://github.com/assertj/assertj-core/blob/main/src/main/java/org/assertj/core/util/CheckReturnValue.java
- https://github.com/findbugsproject/findbugs/blob/master/findbugs/src/java/edu/umd/cs/findbugs/detect/BuildCheckReturnAnnotationDatabase.java#L120

t-kameyama:
Fixes #3975

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -534,6 +534,8 @@ potential-bugs:
     returnValueAnnotations:
       - '*.CheckReturnValue'
       - '*.CheckResult'
+    ignoreReturnValueAnnotations:
+      - '*.CanIgnoreReturnValue'
   ImplicitDefaultLocale:
     active: true
   ImplicitUnitReturnType:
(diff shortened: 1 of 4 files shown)


t-kameyama:
We need to take `@CanIgnoreReturnValue` into account.
https://github.com/assertj/assertj-core/blob/main/src/main/java/org/assertj/core/api/Assertions.java#L1483

BraisGabin:
Nice and thanks for the PR! But now we have another problem: https://github.com/detekt/detekt/pull/3979/checks?check_run_id=3159666709#step:5:1836

And as it seems it was a false positive that AssertJ had and they fixed using `@CanIgnoreReturnValue`: https://github.com/assertj/assertj-core/pull/1534/files

It seems like error prone supports that annotation so maybe we should do it too.

BraisGabin (approved):
Thank you so much for this contribution! I can't imagine how much bugs can fix this PR in our users code base.

chao2zhang (commented):
Would you mind adding a test case when the annotation `@CheckReturnValue` is on the parent class but the method on a child class is invoked without checking the return value?

I do not have strong opinion whether we should report a violation or not, but documenting it in the test would help us clarify.

chao2zhang (approved):

cortinico (approved):

PR #3982, opened 2021-07-26
Use reference in fallback property delegate

marschwar:
The fallback property syntax has always been awkward to use. Triggered by a missing feature for #3972 I wanted to find a way to

* define fallback property without relying on a string matching some other property that is never used
* allow for different transformation functions in the new and old property
* reference the fallback property directly and not just by its name 

~**Drawbacks of this solution:**~
* ~The property to fall back onto has to be public~
* ~This solution relies on `kotlin-reflect`to be available at runtime (which looks like is the case since `kotlin-compiler-embeddable` depends on it)~


detekt-rules-empty/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlock.kt
```diff
@@ -22,13 +22,12 @@ import org.jetbrains.kotlin.psi.psiUtil.getParentOfType
 @ActiveByDefault(since = "1.0.0")
 class EmptyFunctionBlock(config: Config) : EmptyRule(config) {
 
-    @Suppress("unused")
     @Configuration("Excludes all the overridden functions")
     @Deprecated("Use `ignoreOverridden` instead")
-    private val ignoreOverriddenFunctions: Boolean by config(false)
+    val ignoreOverriddenFunctions: Boolean by config(false)
```
BraisGabin, line 27: Isn't it a bit ugly to force to expose the deprecated configuration?

detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/ConfigPropertySpec.kt
```diff
@@ -318,13 +320,15 @@ class ConfigPropertySpec : Spek({
                 val fallbackValue = -1
                 val subject by memoized {
                     object : TestConfigAware("present" to configValue, "fallback" to fallbackValue) {
-                        val present: String by configWithFallback("fallback", defaultValue) { v ->
+                        val fallback: String by config(42, Integer::toString)
+                        val missing: String by config(42, Integer::toString)
+                        val present: String by configWithFallback(::fallback, defaultValue) { v ->
```
BraisGabin, line 325: Could you add a test making it to relay on the fallback and the fallback having a transformation? To check that it works as expected.

BraisGabin (commented):
I can't download the code and test it myself but it's strange that you need to expose the deprecated properties. We should be able to avoid that, right?

marschwar:
> I can't download the code and test it myself but it's strange that you need to expose the deprecated properties. We should be able to avoid that, right?

You can checkout to code using the github tooling `gh pr checkout 3982`. 

There are 2 reasons why the deprecated properties must be there:
* The documentation should include the property marked as deprecated.
* In order to retrieve the (transformed) value, the property has to be there and also must be publicly accessible.

I am wondering if it safe to include the `kotlin-reflect` dependency or if there is a way around it. Any assistance on this is welcome. Please also read the PR description (if not already done). Thanks in advance.

detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/ConfigPropertySpec.kt
```diff
@@ -316,24 +318,27 @@ class ConfigPropertySpec : Spek({
                 val configValue = 99
                 val defaultValue = 0
                 val fallbackValue = -1
+                val fallbackOffset = 10
                 val subject by memoized {
                     object : TestConfigAware("present" to configValue, "fallback" to fallbackValue) {
-                        val present: String by configWithFallback("fallback", defaultValue) { v ->
+                        val fallback: String by config(42) { (it + fallbackOffset).toString() }
+                        val missing: String by config(42) { (it + fallbackOffset).toString() }
+                        val present: String by configWithFallback(::fallback, defaultValue) { v ->
```
marschwar, line 326: This is very close to your use case @BraisGabin. Am I right?

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/ConfigurationCollector.kt
```diff
@@ -140,25 +142,34 @@ class ConfigurationCollector {
...
                 invalidDocumentation {
-                    "The fallback property '$fallbackPropertyName' is missing for property '$name'"
+                    "The fallback property '$fallbackPropertyReference' of property '$name' " +
+                        "must also be defined using a config property delegate "
+                }
+            }
+            if (fallbackProperty.isPrivate()) {
+                invalidDocumentation {
+                    "The fallback property '$fallbackPropertyReference' of property '$name' may not be private"
                 }
```
BraisGabin, line 166: I played a bit with your PR and I still don't get why the fallback can't be private. I make one of them private and I commented this code and the tests and the `generateDocumentation` task worked fine.

marschwar: You can change the visibility in line `detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/ConfigPropertySpec.kt:300` to private and then the test will fail with: 

`java.lang.IllegalAccessException: class kotlin.reflect.jvm.internal.calls.CallerImpl$Method cannot access a member of class io.gitlab.arturbosch.detekt.api.ConfigPropertySpec$1$1$4$1$subject$2$1 with modifiers "private final"
`

The reason is that the `FallbackConfigProperty` tries to access the getter of the field via reflection which is not possible when it is private. That code path is only executed if:
* the config does not contain the key of the new property (configured with `configProperty(::xxx, 42)`)
* AND the config contains the key of the old (deprecated) property

detekt-api/api/detekt-api.api
```diff
@@ -99,8 +99,8 @@ public final class io/gitlab/arturbosch/detekt/api/ConfigPropertyKt {
 	public static final fun config (Ljava/lang/Object;Lkotlin/jvm/functions/Function1;)Lkotlin/properties/ReadOnlyProperty;
 	public static final fun configWithAndroidVariants (Ljava/lang/Object;Ljava/lang/Object;)Lkotlin/properties/ReadOnlyProperty;
 	public static final fun configWithAndroidVariants (Ljava/lang/Object;Ljava/lang/Object;Lkotlin/jvm/functions/Function1;)Lkotlin/properties/ReadOnlyProperty;
-	public static final fun configWithFallback (Ljava/lang/String;Ljava/lang/Object;)Lkotlin/properties/ReadOnlyProperty;
-	public static final fun configWithFallback (Ljava/lang/String;Ljava/lang/Object;Lkotlin/jvm/functions/Function1;)Lkotlin/properties/ReadOnlyProperty;
+	public static final fun configWithFallback (Lkotlin/reflect/KProperty;Ljava/lang/Object;)Lkotlin/properties/ReadOnlyProperty;
+	public static final fun configWithFallback (Lkotlin/reflect/KProperty;Ljava/lang/Object;Lkotlin/jvm/functions/Function1;)Lkotlin/properties/ReadOnlyProperty;
```
cortinico, line 103: That's a breaking change basically right?

cortinico (commented):
> Triggered by a missing feature

What is exactly the missing feature?
I don't have a strong opinion on this PR to be frank. 
Just a small note, imho we're making the API safer but also harder to use. 
Maybe we can provide methods with both `String` and `KProperty` instead? ~of introducing a breaking change?~

detekt-api/api/detekt-api.api
cortinico: It's not my bad, as those method got introduced in 1.18.0 that is not released yet.

marschwar:
> What is exactly the missing feature?

Previously it was only possible to use the exact value from the deprecated property but not apply a function to its value.
 
> Just a small note, imho we're making the API safer but also harder to use.
> Maybe we can provide methods with both `String` and `KProperty` instead of introducing a breaking change?

Technically this is not a breaking change as the original version has not been released yet.

detekt-api/api/detekt-api.api
marschwar: Technically this is not a breaking change as the original version has not been released yet.

marschwar:
> Just a small note, imho we're making the API safer but also harder to use.
> Maybe we can provide methods with both String and KProperty instead? 

I am a little bit weary to increase the api surface even further. Especially for such a functionality that is most likely not used very much. I am very much open to suggestions on how to implement the _missing feature_ in a better way. As I wrote earlier, I do not really know if relying on reflection for this feature is such a great idea.

cortinico:
How do we want to proceed? Ideally if we want to merge this, it should go inside `1.18.0`

BraisGabin:
Sorry, I didn't have the time to look at this until now. I did this:

https://github.com/marschwar/detekt/compare/fallback-with-reference...BraisGabin:fallback-with-reference

It seems that now it allows private properties. The key is to use `KtProperty0` instead of `KtProperty` because the first have a `get()` method that allow us to not call the `getter` that seems that is the one that uses reflection. (It may seems that I know what I'm talking about but I just tried and it worked).

marschwar:
Thank you @BraisGabin. Let me look at it tonight. It seems like a good solution.

marschwar:
```kotlin
/**
* Represents a property without any kind of receiver.
* Such property is either originally declared in a receiverless context such as a package,
* or has the receiver bound to it.
*/
public actual interface KProperty0<out V> : KProperty<V>, () -> V 
```

I don't really understand why it is working either but it seems to do the trick. Thank you for looking into it.

BraisGabin (approved):

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/ConfigProperty.kt
```diff
@@ -157,12 +159,19 @@ private class TransformedConfigProperty<T : Any, U : Any>(
...
     private val transform: (T) -> U
 ) : MemoizedConfigProperty<U>() {
     override fun doGetValue(thisRef: ConfigAware, property: KProperty<*>): U {
-        val fallbackValue = getValueOrDefault(thisRef, fallbackPropertyName, defaultValue)
-        return transform(getValueOrDefault(thisRef, property.name, fallbackValue))
+        if (thisRef.isConfigured(property.name)) {
+            return transform(getValueOrDefault(thisRef, property.name, defaultValue))
+        }
+        if (thisRef.isConfigured(fallbackProperty.name)) {
+            return fallbackProperty.get()
```
chao2zhang, line 171: Question: If we specify fallback property and transformer at the same time, would the transformer be disregarded?

chao2zhang (approved):

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/ConfigProperty.kt
marschwar: If the property exists in the config, the transformer is executed. If that is not the case and the fallback property exists, then its transformer (if present) is used.

cortinico (approved):

marschwar:
We could add `@UnstableApi` for now so this decision does not block the next release

BraisGabin:
I'd vote to move this to 1.19.0 so we can release 1.18.0. Add this code after 3 RC doesn't feel the right move.

cortinico:
> I'd vote to move this to 1.19.0 so we can release 1.18.0. Add this code after 3 RC doesn't feel the right move.

Agree. Given the unstable annotation we added to those APIs, we can easily change it in the next release.

BraisGabin:
Can we merge this now?

cortinico:
> Can we merge this now?

@marschwar Can you take a look at the merge conflicts? Then I belive we can merge this.

PR #3992, opened 2021-08-01
Enable more rules for detekt code base

marschwar:
There are quite a few rule that are deactivated by default and therefor are not checked when running checking the detekt code base.  I think it is very helpful to at least run those that are not controversial.

Looking back I should have probably created multiple PR for this. Sorry about that.

There are still some rules I would like to enable but that would blow up the change set even more. So there will be more PR unless you think there is no need for that.

### DIFF
--- a/detekt-metrics/src/main/kotlin/io/github/detekt/metrics/ComplexityMetric.kt
+++ b/detekt-metrics/src/main/kotlin/io/github/detekt/metrics/ComplexityMetric.kt
@@ -7,6 +7,7 @@ import io.github.detekt.metrics.processors.logicalLinesKey
 import io.github.detekt.metrics.processors.sourceLinesKey
 import io.gitlab.arturbosch.detekt.api.Detektion
 
+@Suppress("UseDataClass")
 class ComplexityMetric(detektion: Detektion) {
 
     val mcc = detektion.getData(complexityKey)
--- a/detekt-metrics/src/main/kotlin/io/github/detekt/metrics/ComplexityMetric.kt
+++ b/detekt-metrics/src/main/kotlin/io/github/detekt/metrics/ComplexityMetric.kt
@@ -6,15 +6,28 @@ import io.github.detekt.metrics.processors.linesKey
 import io.github.detekt.metrics.processors.logicalLinesKey
 import io.github.detekt.metrics.processors.sourceLinesKey
 import io.gitlab.arturbosch.detekt.api.Detektion
+import io.gitlab.arturbosch.detekt.api.Finding
+import io.gitlab.arturbosch.detekt.api.RuleSetId
 
-@Suppress("UseDataClass")
-class ComplexityMetric(detektion: Detektion) {
-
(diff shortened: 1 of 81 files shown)


detekt-metrics/src/main/kotlin/io/github/detekt/metrics/ComplexityMetric.kt
```diff
@@ -7,6 +7,7 @@ import io.github.detekt.metrics.processors.logicalLinesKey
 import io.github.detekt.metrics.processors.sourceLinesKey
 import io.gitlab.arturbosch.detekt.api.Detektion
 
+@Suppress("UseDataClass")
```
marschwar, line 10: This feels like a false positive to me. Should we create an issue for that?

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/LicenceHeaderLoaderExtension.kt
```diff
@@ -57,6 +57,7 @@ class LicenceHeaderLoaderExtension : FileProcessListener {
                 """.trimIndent()
             }
 
+            @Suppress("UnnecessaryLet")
             return Files.newBufferedReader(templateFile)
                 .use(BufferedReader::readText)
                 .let(StringUtilRt::convertLineSeparators)
```
marschwar, line 63: This feels like a false positive to me. Should we create an issue for that?

detekt-metrics/src/main/kotlin/io/github/detekt/metrics/ComplexityMetric.kt
```diff
@@ -7,6 +7,7 @@ import io.github.detekt.metrics.processors.logicalLinesKey
...
 class ComplexityMetric(detektion: Detektion) {
 
     val mcc = detektion.getData(complexityKey)
```
marschwar, line 13: This feels like a false positive to me. Should we create an issue for that?

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/LicenceHeaderLoaderExtension.kt
schalkms: Yes, please

detekt-metrics/src/main/kotlin/io/github/detekt/metrics/ComplexityMetric.kt
schalkms: This is also a false-positive.

marschwar: On second thought I believe this should be a data class with a factory method instead.

marschwar:
This change set is way to big. I will close this and start from scratch with multiple PR.

PR #3996, opened 2021-08-01
Enable more rules from naming rule set for detekt code base

marschwar:
This relates to #3993 and enables checks from the `naming` rule set

- Enable more naming rules for detekt code base
- Fix NoNameShowing violations


detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ForbiddenClassName.kt
```diff
@@ -31,19 +31,14 @@ class ForbiddenClassName(config: Config = Config.empty) : Rule(config) {
...
+        if (forbiddenEntries.isEmpty()) {
+            return
         }
-    }
 
-    companion object {
-        const val FORBIDDEN_NAME = "forbiddenName"
+        val message = "Class name $name is forbidden as it contains: ${forbiddenEntries.joinToString(", ")}"
+        report(CodeSmell(issue, Entity.atName(classOrObject), message))
     }
```
marschwar, line 43: This is a "drive-by" fix as the original code did not successfully remove the trailing comma. The result of the `trimEnd` should have been re-assigned to message.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/YamlConfig.kt
```diff
@@ -72,7 +72,9 @@ class YamlConfig internal constructor(
             val map: Map<*, *>? = runCatching {
                 @Suppress("USELESS_CAST") // runtime inference bug
                 Yaml().loadAs(it, Map::class.java) as Map<*, *>?
-            }.getOrElse { throw Config.InvalidConfigurationError(it) }
+            }.getOrElse { cause ->
+                throw Config.InvalidConfigurationError(cause)
+            }
```
BraisGabin, line 77: Shouldn't it be better to name the `.use` variable instead of this? It's easy to understand that here `it` referse to the cause. But it's not that clear that in line 74 the `it` refers to `reader.buffered()`.

cortinico (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/YamlConfig.kt
marschwar: Very good idea

BraisGabin (approved):

PR #3997, opened 2021-08-01
enable more potential-bugs rules for detekt code base

marschwar:
This relates to #3993 and enables checks from the potential-bugs rule set


detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/ProcessingSettingsFactory.kt
```diff
@@ -48,7 +48,7 @@ fun createNullLoggingSpec(init: (ProcessingSpecBuilder.() -> Unit)? = null): Pro
 class DirectExecutor : AbstractExecutorService() {
 
     override fun execute(command: Runnable): Unit = command.run()
-    override fun shutdown() = Unit
+    override fun shutdown(): Unit = Unit
```
marschwar, line 51: Fun fact:

```kotlin
override fun shutdown() = Unit // -> violates ImplicitUnitReturnType 

override fun shutdown(): Unit = Unit // -> violates OptionalUnit

override fun shutdown() {} // -> violates EmptyFunctionBlock
```

BraisGabin:
```kotlin
override fun shutdown() = Unit // -> violates ImplicitUnitReturnType 
```

This feels like a false positive, right?

It's not "Implicit".

Anyway I like this way to implement it:

```kotlin
override fun shutdown() {
  // no-op
}
```

BraisGabin: I'd vote to revert this change (and the one in LicenceHeaderLoaderExtensionSpec.kt). Add them to our baseline and open an issue to fix the false-positive. Those changes doesn't improve the code so they should not be made just to make detekt happy.

cortinico: > Those changes doesn't improve the code so they should not be made just to make detekt happy.

Agree +1. I also prefer the `//no-op` approach but the status quo (` = Unit`) should also be accepted.

marschwar: I added `@Suppress' to those 2 occurrences and raised #4004

cortinico (approved):

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/ProcessingSettingsFactory.kt
BraisGabin: Is it good to add the suppress? For false postives the suppress have the problem that generates debt because we will need to remove it. If we use the baseline instead we can just regenerate it in the next version.

marschwar: Agreed and changed. Unfortunately `gradlew detektBaselineTest` will be run for each module and output the baseline to `config/detekt/baseline-test.xml`. So each module overwrites the baseline from the previous module making it a rather manual process to create a combined version.

BraisGabin (approved):

PR #4002, opened 2021-08-04
Update ktlint to 0.42.0

G00fY2:
This PR updates ktlint to [0.42.0](https://github.com/pinterest/ktlint/releases/tag/0.42.0).

Looks like 0.42.0 is mainly a bugfix release (beside the SARIF output support). So I don't see any necessary changes beside bumping the version number.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,7 +2,7 @@
 dokka = "1.4.32"
 jacoco = "0.8.7"
 kotlin = "1.5.21"
-ktlint = "0.41.0"
+ktlint = "0.42.0"
 spek = "2.0.15"
 
 [libraries]


BraisGabin:
0.42.0 introduced two regressions:

https://github.com/pinterest/ktlint/issues/1188
https://github.com/pinterest/ktlint/issues/1189

I think that we should wait for the next release of ktlint.

G00fY2:
@BraisGabin thanks for pointing out. I think we can close this PR then.

Was actually looking forward to ktlint 0.42.0 because I thought the Indentation rule is finally fixed (pinterest/ktlint#764). We struggle with this formatting rule since 0.37.0. But if there are regressions we should definitely wait for a fix.

cortinico:
@G00fY2 don't you mind reviving this PR now that `0.42.1` has landed? https://github.com/pinterest/ktlint/releases/tag/0.42.1

PR #4009, opened 2021-08-06
Add information about config delegates to release notes

marschwar:
This is my first attempt contributing to the release notes. Any feedback is welcome. 

Is this sort of what you had in mind @cortinico or did you talk about something else in https://github.com/detekt/detekt/pull/3891#issuecomment-893871480 ?


BraisGabin (approved):
<3 Great work!

cortinico (approved):
Yes that's exactly what I meant :)

I'm more wondering if, given the amount of information, we should probably create a "configuration-delegate" page in the official website 🤔 and maybe link it in the release notes?

The risk of having this small snippets only in the release notes will make them harder to find in weeks/months for people that are still wondering what's the best practice.

marschwar:
Thank you. I don't think this needs an extra page but we should probably elaborate more on the "How to write a rufe" page. I can do that but I was going to wait until We decide about the fallback property.

schalkms (approved):

cortinico (approved):

PR #4011, opened 2021-08-07
ImplicitUnitReturnType: don't report when expression body is 'Unit'

### LINKED ISSUE #4004: ImplicitUnitReturnType false positive
## Expected Behavior

```kotlin
fun foo() = Unit
```

should not violate the `ImplicitUnitReturnType`rule as there is nothing implicit about the return type if the function body is simply `Unit`. What makes matters worse is the fact that the `OptionalUnit` rule explicitly suggests `override fun foo() = Unit`.

t-kameyama:
Fixes #4004


cortinico (approved):

BraisGabin (approved):
@t-kameyama I just wanted to say you thanks! Your contributions are so helpful fixing a lot of false positives <3

PR #4013, opened 2021-08-07
Add IllegalArgumentException and IllegalStateException to ThrowingExceptionsWithoutMessageOrCause

### LINKED ISSUE #4012: Add more exceptions to ThrowingExceptionsWithoutMessageOrCause
## Expected Behavior
`ThrowingExceptionsWithoutMessageOrCause` should report `IllegalArgumentException` and `IllegalStateException` by default as before.

## Observed Behavior
In 1.18.0-RC3 detekt doesn't report those two.

## Context
Those two exceptions are too generic if they don't have a cause or a message.

## Your Environment
* Version of detekt used: 1.18.RC-3

BraisGabin:
This two exceptions where removed but I think that we should keep them.

Fix #4012


cortinico (approved):

schalkms (approved):
What was the reason for removing these exceptions in the first place?

marschwar:
Previously the situation was:

```
/**
 * @configuration exceptions - exceptions which should not be thrown without message or cause
 * (default: `- IllegalArgumentException
 *            - IllegalStateException
 *            - IOException`)
**/
...
        private val exceptionsDefaults = listOf(
            "ArrayIndexOutOfBoundsException",
            "Error",
            "Exception",
            "IllegalMonitorStateException",
            "NullPointerException",
            "IndexOutOfBoundsException",
            "RuntimeException",
            "Throwable"
        )
```

So there was quite a mismatch between what was configured and what was documented.

BraisGabin:
Yeah, but if you used the "default config" as "default" you were using the "documentation" instead of the real default values. For that reason I think that it is kind of a regresion.

marschwar (approved):
Please run `gradlew generateDocumentation` to update the default config.

BraisGabin:
> Please run `gradlew generateDocumentation` to update the default config.

Good call!

🤔 we have a problem with CI here, right? Why I'm getting green if I forgot about that?

marschwar:
That check is gone for quite some time. I have no idea when and why.

PR #4014, opened 2021-08-07
Sort config lists

BraisGabin:
In this release we are already changing our default config yaml. For that reason I think that we could use it to sort the list configurations to follow an alphabetic order. I sorted all of them but I'm not sure if it have sense to do it with all of them because maybe there is another "more logical order". These are the rules that maybe a "more logical order" can exist:

- ComplexMethod (keep all the [scope functions](https://kotlinlang.org/docs/scope-functions.html#function-selection) together)
- ExceptionRaisedInUnexpectedLocation (order them for "more common")
- ForbiddenComment (order them for "more common")

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -92,15 +92,15 @@ complexity:
     ignoreSimpleWhenEntries: false
     ignoreNestingFunctions: false
     nestingFunctions:
-      - 'run'
-      - 'let'
-      - 'apply'
-      - 'with'
       - 'also'
-      - 'use'
+      - 'apply'
       - 'forEach'
       - 'isNotNull'
       - 'ifNull'
+      - 'let'
+      - 'run'
+      - 'use'
+      - 'with'
   LabeledExpression:
     active: false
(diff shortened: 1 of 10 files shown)


cortinico (approved):

schalkms (commented):
I'm not sure. We would have to pay attention. Every PR could change the alphabetical order. I don't think this would improve anything.

cortinico:
> Every PR could change the alphabetical order.

I was thinking about the same. Perhaps we should consider adding some pre-commit hooks or PR checks that verifies the YAML files?

schalkms:
A pre-commit hook doesn't help much, since every contributor needs to configure the hook on the local machine.
A PR check or even better a unit test could help, but I am unsure regarding value/effort.

BraisGabin:
> I'm not sure. We would have to pay attention.

Completely agree. This is easy to be broken. I do this change just because we are changing all those lines anyway so, at least, have an alphabetical order for some configurations that, usually, doesn't change. For me the check/unit test is not worth it. I vote to merge because this lines will be changed for all our users anyway. And if a next change "breaks" the order... well, that's not a big deal.

PR #4016, opened 2021-08-07
UseDataClass: fix false positive on value classes

### LINKED ISSUE #4010: UseDataClass triggers on value classes
## Expected Behavior

Detekt does not report "UseDataClass" on value classes.

## Observed Behavior

Detekt reports "UseDataClass" on value classes.

## Steps to Reproduce

```kotlin
@JvmInline
value class PlayerId(val value: Long)
```

## Context

Kotlin 1.5 renamed `inline class` to `value class`. The former was handled in #1450 but the latter is the stable feature.

## Your Environment

cortinico:
Fixes #4010

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
@@ -326,6 +326,14 @@ class UseDataClassSpec : Spek({
             assertThat(subject.lint("inline class A(val x: Int)")).isEmpty()
         }
 
+        it("does not report value classes") {
+            val code = """
+                @JvmInline
+                value class PlayerId(val value: Long)
+            """.trimIndent()
+            assertThat(subject.lint(code)).isEmpty()
+        }
+
         it("does not report a class which has an ignored annotation") {
             val code = """
                 import kotlin.SinceKotlin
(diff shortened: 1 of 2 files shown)


BraisGabin (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
```diff
@@ -326,6 +326,14 @@ class UseDataClassSpec : Spek({
             assertThat(subject.lint("inline class A(val x: Int)")).isEmpty()
         }
 
+        it("does not report value classes") {
+            val code = """
+                @JvmInline
+                value class PlayerId(val value: Long)
```
dlin-ta, line 332: Not that it should make a difference but it looks like other tests in this file use simpler names and types.
```kotlin
class A(val x: Int)
```

dlin-ta (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
schalkms: +1

schalkms (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseDataClassSpec.kt
cortinico: Makes sense, updated.

marschwar (approved):

PR #4021, opened 2021-08-08
enable UseDataClass rule for detekt code base

marschwar:
Belongs to discussion #3993

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
@@ -54,7 +54,8 @@ internal interface Lifecycle {
     }
 }
 
-internal data class DefaultLifecycle(
+@Suppress("UseDataClass")
+internal class DefaultLifecycle(
     override val settings: ProcessingSettings,
     override val parsingStrategy: ParsingStrategy,
     override val bindingProvider: (files: List<KtFile>) -> BindingContext =
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
@@ -54,7 +54,7 @@ internal interface Lifecycle {
     }
 }
 
-internal class DefaultLifecycle(
+internal data class DefaultLifecycle(
     override val settings: ProcessingSettings,
     override val parsingStrategy: ParsingStrategy,
     override val bindingProvider: (files: List<KtFile>) -> BindingContext =
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektReport.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektReport.kt
@@ -2,7 +2,7 @@ package io.gitlab.arturbosch.detekt.extensions
 
 import java.io.File
 
-class DetektReport(val type: DetektReportType) {
+data class DetektReport(val type: DetektReportType) {
 
     var enabled: Boolean? = null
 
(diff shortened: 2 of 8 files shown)


cortinico (approved):

detekt-metrics/src/main/kotlin/io/github/detekt/metrics/ComplexityMetric.kt
```diff
@@ -6,14 +6,30 @@ import io.github.detekt.metrics.processors.linesKey
...
 
-    val mcc = detektion.getData(complexityKey)
-    val cognitiveComplexity = detektion.getData(CognitiveComplexity.KEY)
-    val loc = detektion.getData(linesKey)
-    val sloc = detektion.getData(sourceLinesKey)
-    val lloc = detektion.getData(logicalLinesKey)
-    val cloc = detektion.getData(commentLinesKey)
-    val findings = detektion.findings.entries
+    companion object Factory {
+        fun create(detektion: Detektion): ComplexityMetric = detektion.run {
```
BraisGabin, line 23: I never know if in this case is better to use `run` or `with`. Both does the same but I don't know which one is more semantically correct. 

And, another think is factory vs secundary constructor.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
```diff
@@ -54,7 +54,7 @@ internal interface Lifecycle {
     }
 }
 
-internal class DefaultLifecycle(
+internal data class DefaultLifecycle(
```
BraisGabin, line 57: This class doesn't feel like a data class, don't you think? I don't know if it's a false positive or just a good place to add a suppress.

If this is a false positive how could we fix it? Maybe we could check if all the properties implement a real equals? Or that the properties are data classes?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektReport.kt
```diff
@@ -2,7 +2,7 @@ package io.gitlab.arturbosch.detekt.extensions
 
 import java.io.File
 
-class DetektReport(val type: DetektReportType) {
+data class DetektReport(val type: DetektReportType) {
```
BraisGabin, line 5: And I think that this is a real false positive. This class has other properties so the equals can't be correctly implemented.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
marschwar: I am actually not worried so much about the equals method (maybe I should) but rather what the class does. If it just holds data, it should be a data class. At first glance this seems to be the case. But the class implements an interface with a function and therefore does have behavior. Hence you are right - it should not be a data class. But it really is hard to define a rule for this.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektReport.kt
marschwar: Good catch. But I don't think it is a false positive. I added those properties to the primary constructor. All this class does is hold data.

detekt-metrics/src/main/kotlin/io/github/detekt/metrics/ComplexityMetric.kt
```diff
@@ -6,14 +6,30 @@ import io.github.detekt.metrics.processors.linesKey
 import io.github.detekt.metrics.processors.logicalLinesKey
 import io.github.detekt.metrics.processors.sourceLinesKey
 import io.gitlab.arturbosch.detekt.api.Detektion
+import io.gitlab.arturbosch.detekt.api.Finding
+import io.gitlab.arturbosch.detekt.api.RuleSetId
 
-class ComplexityMetric(detektion: Detektion) {
+data class ComplexityMetric(
```
schalkms, line 12: I'm not sure whether the new code is better overall than the old one.

cortinico:
> So my vote goes to keep just the changes related with those two files and don't enable this rule by default in the detekt code.

Agree. Specifically the `ComplexityMetric` change seems like is not improving readability.

PR #4024, opened 2021-08-08
Enable UnnecessaryLet rule for detekt code base

marschwar:
Yet another one

### DIFF
--- a/config/detekt/baseline-main.xml
+++ b/config/detekt/baseline-main.xml
@@ -1,7 +1,0 @@
-<?xml version='1.0' encoding='UTF-8'?>
-<SmellBaseline>
-  <ManuallySuppressedIssues/>
-  <CurrentIssues>
-    <ID>UnnecessaryLet:LicenceHeaderLoaderExtension.kt$LicenceHeaderLoaderExtension$let(StringUtilRt::convertLineSeparators)</ID>
-  </CurrentIssues>
-</SmellBaseline>
--- a/config/detekt/baseline-main.xml
+++ b/config/detekt/baseline-main.xml
@@ -2,8 +2,6 @@
 <SmellBaseline>
   <ManuallySuppressedIssues/>
   <CurrentIssues>
-    <ID>
-      UnnecessaryLet:LicenceHeaderLoaderExtension.kt$LicenceHeaderLoaderExtension$let(StringUtilRt::convertLineSeparators)
-    </ID>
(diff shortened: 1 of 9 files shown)


config/detekt/baseline-main.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<SmellBaseline>
+  <ManuallySuppressedIssues/>
+  <CurrentIssues>
+    <ID>
+      UnnecessaryLet:LicenceHeaderLoaderExtension.kt$LicenceHeaderLoaderExtension$let(StringUtilRt::convertLineSeparators)
+    </ID>
+  </CurrentIssues>
+</SmellBaseline>
```
marschwar, line 9: This refers to 

```kotlin
return Files.newBufferedReader(templateFile)
    .use(BufferedReader::readText)
    .let(StringUtilRt::convertLineSeparators)
```

I am not sure if it is a false positive or should be written as this instead
```kotlin
return StringUtilRt.convertLineSeparators(
    Files.newBufferedReader(templateFile).use(BufferedReader::readText)
)
```

config/detekt/baseline-main.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<SmellBaseline>
+  <ManuallySuppressedIssues/>
+  <CurrentIssues>
+    <ID>UnnecessaryLet:LicenceHeaderLoaderExtension.kt$LicenceHeaderLoaderExtension$let(StringUtilRt::convertLineSeparators)</ID>
```
marschwar, line 5: This refers to 

```kotlin
return Files.newBufferedReader(templateFile)
    .use(BufferedReader::readText)
    .let(StringUtilRt::convertLineSeparators)
```

I am not sure if it is a false positive or should be written as this instead
```kotlin
return StringUtilRt.convertLineSeparators(
    Files.newBufferedReader(templateFile).use(BufferedReader::readText)
)
```

BraisGabin: I think that this is not a false positive. `convertLineSeparators` could be an extension function. The problem is that it would be an extension function over String so we have the problem of Scope polution... So I woud vote to create the extension funcion or adding the `@Supress` because, event that it's a real unnecessary `let` we want to keep it.

schalkms (approved):

BraisGabin (approved):

Issue #3927, opened 2021-06-30
Provide a way to exclude detekt tasks in multiplatform project


petertrr:
## Expected Behavior
For JVM-only projects excluding detekt is as easy as runnging `./gradlew <task> -x detekt`. It should be also possible for KMP projects, e.g. by using `detekt` or `detektAll` or similar aggregator task instead of using `check` directly.

## Current Behavior
For KMP projects detekt-gradle-plugin creates a number of tasks, one per kotlin compilation target. All these tasks are added as dependencies for `check` task, so, e.g. running `./gradlew build -x detekt` doesn't exclude all `detekt*` tasks.

## Context
No there is no obvious options. One can manually exclude all tasks, but it makes CI scripts grow larger. Or one can try to manually remove dependencies from `check` task, but it overcomplicates gradle configuration.

cortinico:
That is already doable today by adding

```
detekt.multiplatform.disabled=true
```

To your `gradle.properties`.
Can I ask you to confirm that works as intended and eventually close this ticket?

petertrr:
@cortinico Thanks, this is really helpful! However, in project with mixed JVM-only and multiplatform subprojects disabling detekt requires two steps: `detekt.multiplatform.disabled=true` for multiplatform subprojects and `-x detekt` for JVM-only subprojects. Maybe, there is still a possible improvement for this via some sort of aggregating task instead of binding to `check`?

Anyway, with your help my issue is resolved, so I'm closing it.

cortinico:
> in project with mixed JVM-only and multiplatform subprojects disabling detekt requires two steps: `detekt.multiplatform.disabled=true` for multiplatform subprojects and `-x detekt` for JVM-only subprojects.

Just out of curiosity: why are you applying the Detekt gradle plugin then? I'm not sure if your problem is having detekt tasks in the graph or having `check` depend on them.

BraisGabin:
It seems to me that the best way to solve your problem is to remove the dependecy between `check` and the detekt tasks.

petertrr:
> Just out of curiosity: why are you applying the Detekt gradle plugin then? I'm not sure if your problem is having detekt tasks in the graph or having `check` depend on them.

@cortinico I want to check all modules using detekt, but on CI I want two distinct jobs: one with tests only, and one with static analysis. Usually this can be achieved by excluding detekt tasks from the former, and I thought to use the same approach with more complicated project structure.

[...]

Issue #3838, opened 2021-05-25
Create new gradle task that runs all the detekt tasks


mkrussel77:
Your feature request may already be reported!
Please search in the [issues](https://github.com/detekt/detekt/issues) before creating a new one.

## Expected Behavior

Would be nice for there to be a single task that runs all the other Detekt tasks.

So with an Android project it would run

- detekt
- detektMain
- detektTest

For Multiplatform it would run
- detekt
- detektAndroidDebug
- detektAndroidDebugAndroidTest
- detektAndroidDebugUnitTest
- detektAndroidRelease
- detektAndroidReleaseAndroidTest
- detektAndroidReleaseUnitTest
- detektIosArm64Main
- detektIosArm64Test
- detektIosX64Main
- detektIosXcd64Test
- detektJvmMain
- detektJvmTest
...

## Current Behavior

There is no task

## Context

Goal would be to simplify the number of tasks to enter to get the type resolution runs to run on all the source code.

Right now it is easy for Android projects, but Multiplatform projections are complicated.

BraisGabin:
Agree. I propose to use `detektCheck` for that. I think that `detektMain` and `detektTest` should be internal tasks, not intended for the use of the users. And, probably, it should output an unified report of that module instead of two. 

But I'm not sure if we are at the point to provide it without adding breaking changes in the future. 

I recommend you to create a custom task that depend on all those rules so you don't need to write all of those each time. 

Side note: If you run `detektMain` and `detektTest` you don't need to run `detekt`.

cortinico:
I believe we could add a `detektAll` task that depends on all the `detekt*` tasks, but **not** on the `detekt` task (as it could produce conflicting results having Type Resolution enabled/disabled.

mkrussel77:
After creating this ticket, I found out that everything is added to the `check` task.  So that is one option.

I'm currently having problems creating my own task due to #3840

3flex:
> I found out that everything is added to the check task

This is the Gradle convention for code quality tasks - individual tasks are added as dependencies of the `check` lifecycle task.

The problem with adding new tasks with "detekt" prefix that depend on other detekt tasks is that if there's a source set named "check" for example and this plugin had (again, example) a task called "detektCheck" that just ran other detekt tasks then there'd be a collision on task names.

So I don't support adding a new task that just depends on other detekt tasks. But an example of how to configure this and attach tasks to `check`, or create a custom task for that purpose, can be added to the documentation.

[...]

PR #4025, opened 2021-08-08
Multiplatform tasks should not depend on check

cortinico:
This PR is cleaning up our task dependencies a bit. Specifically the only task that should be a dependency of `check` is the plain `detekt` task.

When I worked on the KMP implementation, I accidentally plugged all the KMP tasks to depend on `check`. That's problematic as we still have some rough edges to fix and people might experience problems with their build because of this. See #3927 #3838


detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPlainSpec.kt
```diff
@@ -72,5 +72,11 @@ object DetektPlainSpec : Spek({
                 assertThat(buildResult.output).doesNotContain("--classpath")
             }
         }
+
+        it("detekt plain task will depend on check") {
+            gradleRunner.runTasksAndCheckResult(":check") { buildResult ->
+                assertThat(buildResult.tasks.map { it.path }).contains(":detekt")
+            }
+        }
```
chao2zhang, line 80:
```suggestion
```

I think this is already covered by https://github.com/detekt/detekt/pull/4025/files#diff-449c58285efdbf81c836b65ea85659877bfe46350c808afb25144183041de5f7R32

chao2zhang (approved):

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPlainSpec.kt
cortinico: > already covered by #4025 (files)

Which file are you referring to? I moved this test inside `DetektPlain` as we want to verify this behavior only for the plain task

BraisGabin (approved):

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektPlainSpec.kt
chao2zhang: 😢  The link does not work properly. I intended to mention that this is already covered by line 32 in the same file. But now it seems that we could merge the two `describe()` spek together.

I don't think this is a blocking comment to be addressed though.

cortinico: Yup you're right, this test is unnecessary, as the one above is checking the same condition. I'm removing it.

schalkms (approved):

schalkms:
I assume that this PR fixes both linked issues, right?

cortinico:
> I assume that this PR fixes both linked issues, right?

Nope it does not. It's related to both but the real fix needs to be discussed in the issues.

PR #4030, opened 2021-08-11
Fix TextLocation of Indentation rule

Yundin:
Although Indentation applies to the whole file, it's a local rule. Start of the real location reported by ktlint's offset. And I assume that the end location should be on first non-whitespace character.
This cause detekt plugin to highlight the whole file and html output to highlight everything to the end of snippet.

Relevant issue: https://github.com/detekt/detekt-intellij-plugin/issues/131


schalkms (commented):
Awesome 1st contribution! Thank you for that.
The code changes look fine to me. In my opinion, a test should be added in order to avoid this issue in the future.

cortinico (approved):
great stuff 🎉

Yundin:
Great idea! I will add test for that

schalkms (approved):

PR #4033, opened 2021-08-12
Replace `get_analysis_projects.groovy` with kotlin script

marschwar:
I think the script `get_analysis_projects.groovy` has not been working for quite a while now as the `vcsreader:1.1.1` is nowhere to be found and some of the repos no longer seem to exist. 

I have no opinion on which repos should be included. Feel free to add or remove.


BraisGabin (approved):

BraisGabin:
The problem with this script is, that if your rule uses type solving, you will need to setup detekt gradle plugin. Type solving with the cli is a pain.

cortinico:
> has not been working for quite

I'm more wondering who/where this script should be used. I believe this was part of the release process at some point in the past but I don't think it's used right now.

BraisGabin:
> I just thought it is weird to have something in the repo that is not working.

Completely agree with this. I vote for merge this and try to find it a good use for it. The idea is good and it should allow us to find a lot of false-positives.

cortinico (approved):

Issue #4042, opened 2021-08-14
Repository is not consistent with the DoubleMutabilityForCollection rule


geojakes:
## Expected Behavior
Detekt should be compatible with its DoubleMutabilityForCollection rule

## Observed Behavior
Detekt is not currently in conformance with its DoubleMutabilityForCollection rule as seen here: https://github.com/detekt/detekt/pull/4040#issuecomment-898722564

## Steps to Reproduce
Run `./gradlew detektMain detektTest` on the repository ensuring that the rule is enabled in `config/detekt/detekt.yml`

## Context
Blocks merging pull requests in this repository if the rule is enabled

PR #4047, opened 2021-08-16
Verify at compile time that issue id matches rule name

marschwar:
This relates to #4042 and adds a check that verifies that the issue ID matches the rule name.

### DIFF
--- a/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/ConfigAssert.kt
+++ b/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/ConfigAssert.kt
@@ -2,6 +2,9 @@ package io.gitlab.arturbosch.detekt.core
 
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.RuleSetProvider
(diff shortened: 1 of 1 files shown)


BraisGabin (approved):
Does this do the trick? Does core tests wait for all the rule modules to be compiled? `testRuntimeOnly(projects.detektRules)` does that?

marschwar:
Seems like it. I intentionally broke the build with https://github.com/detekt/detekt/pull/4047/commits/e2d32dba6cb352d0f090ec5e96a23c05a39e639f and the check failed.

cortinico (approved):
I believe we should merge this, but at the same time I'd like those kind of checks to happen "earlier" in the rule lifecycle.

Specifically I believe we should update:
```
data class Issue(
    val id: String,
    val severity: Severity,
    val description: String,
    val debt: Debt
) {
```

to have a ctor that accepts a `Rule` as parameter. Or even we should remove the `id` from `Issue` entirely, as the `Rule` class is using it to expose the  `ruleId` property:

```
final override val ruleId: RuleId get() = issue.id
```

So we're creating a tight couplign between the two classes, and we don't need to duplicate informations with the risk of having bugs like the one we just experienced.

marschwar:
Entirely agree. This is not more that a band-aid as this problem can easily be missed in the PR and the consequences are quite big.

PR #4050, opened 2021-08-17
Add an alias for FunctionMinLength/FunctionMaxLength rules to be more descriptive

mohamed-elmahdi:
Renaming the following rules (FunctionMinLength, FunctionMaxLength) to (FunctionNameMaxLength, FunctionNameMinLength) as these rules validate the length of the function names not the functions themselves making them more descriptive to either a user reading the code or a user reading a report.


cortinico:
This is a breaking change and I don't believe it will land as it is.

I suggest we eventually add an alias for those rules like:

```
    override val defaultRuleIdAliases: Set<String>
        get() = setOf("FunctionNameMaxLength")
```

cortinico:
> or will this just change the interface through which people suppress these rules?

^ This
Plus it should work also for the config file.

mohamed-elmahdi:
Does it usually take this long to get reviewed? Or is there something missing form my side delaying the process?

mohamed-elmahdi:
Valid points. I'll investigate later if I can change the rule name (class name) itself without breaking anything, this will unify the naming and eliminate the confusion problem. Would you be kind enough to give me a hint of what this change would break? the last time I tried to make this change, The failing tests were the ones checking for the rule names (Class names). I figured that these were made for some reason but I couldn't fish it out.

cortinico:
>  I'll investigate later if I can change the rule name (class name) itself without breaking anything, this will unify the naming and eliminate the confusion problem.

Please note that you can also "copy" those rules to your own custom rules. This will allow you to rename them as you wish and to don't have an alias at all (it will create less confusion for your users).

> if I can change the rule name (class name) itself without breaking anything,
> Would you be kind enough to give me a hint of what this change would break?

Renaming a rule inside Detekt is always a breaking change and we never do that (unless we ship a major version). The reason is that this is going to affect all the users that were using that rule and, unless something really critical happened, it's not a cost we're willing to pay.

BraisGabin:
Sorry for the late feedback. I thought that this rule was still about renaming the rules, and I saw already cortinico saying no so there was not point to say the same.

I don't see any problem with the current implementation. This will just improve the suppression experience. You could use `@Supress("FunctionMaxNameLength")` or `@Supress("FunctionMaxLength")` and both will work. But nothing else will change. Is that what we want to archive here? If so, I agree with merging this.

I do agree that the name of these two rules is bad and we should change it but I think that the way to fix this is to create a new issue and mark it as a breaking change and do it for detekt 2.0

Issue #4041, opened 2021-08-13
start path with file:// so that intellij turns it into a link in the console


robstoll:
Your feature request may already be reported!
Please search in the [issues](https://github.com/detekt/detekt/issues) before creating a new one.

## Expected Behavior
one can click in the console on the path related to the issue and can jump there immediately

## Current Behavior
the path is in the console but one has to go there manually

## Context
Following an example:
```
/home/robert/projects/kbox/src/jvmMain/kotlin/ch/tutteli/kbox/JvmSynchronized.kt - 5min debt
        EndOfSentenceFormat - [JvmSynchronized] at /home/robert/projects/kbox/src/jvmMain/kotlin/ch/tutteli/kbox/JvmSynchronized.kt:3:1
```
if it would be the following, then it would be recognised by intellij as link and would be clickable:
```
/home/robert/projects/kbox/src/jvmMain/kotlin/ch/tutteli/kbox/JvmSynchronized.kt - 5min debt
        EndOfSentenceFormat - [JvmSynchronized] at file:///home/robert/projects/kbox/src/jvmMain/kotlin/ch/tutteli/kbox/JvmSynchronized.kt:3:1
```
Much nicer isn't it? :slightly_smiling_face: 

I can also provide a PR if you can give me a pointer where to look

cortinico:
Yes it makes sense 👍 Perhaps all the paths should be updated (not just the second one in your example)?

If you want to contribute you should look into those files:
https://github.com/detekt/detekt/tree/main/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/console

marschwar:
As a side note, running the detekt tasks in the IDE allows me to click on the link directly

![Screenshot_2021-08-17_14-07-45](https://user-images.githubusercontent.com/4807823/129723037-cfcffdf9-839a-4208-b22c-3003873b2b52.png)

Version: IntelliJ Idea 2021.1.3

robstoll:
> Perhaps all the paths should be updated (not just the second one in your example)?

I would only update paths which include a line number. I'll draft a PR.

chao2zhang:
Echoing @marschwar, it works for me in intelliJ console without the additional `file://`

robstoll:
@chao2zhang when you say in the IntelliJ console, then you are also referring to the detekt-plugin right? if you ran `./gradlew detekt` then it does not, isn't it?

BraisGabin:
[...]

PR #4053, opened 2021-08-17
include `file://` in compact version of Location

robstoll:
As you can see in the changes, it looks like there are cases where file:// does not make sense. Not all paths are absolute paths. My guess is that we are dealing with test-data here which does not resemble the real world scenario and in real world they would all be absolute paths. If this is the case then I suggest to change this to set the expectations straight -- otherwise I can check if the path starts with / or [A-Z]:

Would fix 4041

### DIFF
--- a/detekt-core/src/test/resources/reporting/grouped-findings-report.txt
+++ b/detekt-core/src/test/resources/reporting/grouped-findings-report.txt
@@ -1,7 +1,7 @@
 File1.kt - 10min debt
-	TestSmell - [TestEntity] at File1.kt:1:1
-	TestSmell - [TestEntity] at File1.kt:1:1
+	TestSmell - [TestEntity] at file://File1.kt:1:1
+	TestSmell - [TestEntity] at file://File1.kt:1:1
 File2.kt - 5min debt
-	TestSmell - [TestEntity] at File2.kt:1:1
+	TestSmell - [TestEntity] at file://File2.kt:1:1
 
 Overall debt: 15min
--- a/detekt-report-txt/src/test/kotlin/io/github/detekt/report/txt/TxtOutputReportSpec.kt
+++ b/detekt-report-txt/src/test/kotlin/io/github/detekt/report/txt/TxtOutputReportSpec.kt
@@ -20,7 +20,7 @@ class TxtOutputReportSpec : Spek({
         it("renders one") {
             val report = TxtOutputReport()
             val detektion = TestDetektion(createFinding())
-            val renderedText = "TestSmell - [TestEntity] at TestFile.kt:1:1 - Signature=TestEntitySignature\n"
+            val renderedText = "TestSmell - [TestEntity] at file://TestFile.kt:1:1 - Signature=TestEntitySignature\n"
             assertThat(report.render(detektion)).isEqualTo(renderedText)
         }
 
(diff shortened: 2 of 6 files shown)


robstoll:
Metaspace problems, out of luck I guess.  Adding the following to `./gradlew` could help:
`-Dorg.gradle.jvmargs="-Xmx2g -XX:MaxMetaspaceSize=1g"` I don't want to change workflows with this PR but I can add it if you want.

BraisGabin:
> Metaspace problems, out of luck I guess. Adding the following to `./gradlew` could help:
> `-Dorg.gradle.jvmargs="-Xmx2g -XX:MaxMetaspaceSize=1g"` I don't want to change workflows with this PR but I can add it if you want.

Yes... we are having "a lot" of those lately...

detekt-core/src/test/resources/reporting/grouped-findings-report.txt
```diff
@@ -1,7 +1,7 @@
 File1.kt - 10min debt
-	TestSmell - [TestEntity] at File1.kt:1:1
-	TestSmell - [TestEntity] at File1.kt:1:1
+	TestSmell - [TestEntity] at file://File1.kt:1:1
+	TestSmell - [TestEntity] at file://File1.kt:1:1
 File2.kt - 5min debt
-	TestSmell - [TestEntity] at File2.kt:1:1
+	TestSmell - [TestEntity] at file://File2.kt:1:1
```
BraisGabin, line 5: These paths aren't full paths. Is it safe tou use `file://` on those?

detekt-report-txt/src/test/kotlin/io/github/detekt/report/txt/TxtOutputReportSpec.kt
```diff
@@ -32,9 +32,9 @@ class TxtOutputReportSpec : Spek({
                 createFinding(ruleName = "TestSmellC")
             )
             val renderedText = """
-                TestSmellA - [TestEntity] at TestFile.kt:1:1 - Signature=TestEntitySignature
-                TestSmellB - [TestEntity] at TestFile.kt:1:1 - Signature=TestEntitySignature
-                TestSmellC - [TestEntity] at TestFile.kt:1:1 - Signature=TestEntitySignature
+                TestSmellA - [TestEntity] at file://TestFile.kt:1:1 - Signature=TestEntitySignature
+                TestSmellB - [TestEntity] at file://TestFile.kt:1:1 - Signature=TestEntitySignature
+                TestSmellC - [TestEntity] at file://TestFile.kt:1:1 - Signature=TestEntitySignature
```
BraisGabin, line 37: I don't think that the txt output should have `file://`.

BraisGabin (commented):
I don't have a strong opinion here but I think that we should not make this change. I'd like to know the thoughts of other maintaneirs about this.

robstoll:
> but I think that we should not make this change.

Just that you know, totally fine with me, close the PR whenever you want. Please adjust as needed in case you want to make changes, I am out... :wave:

marschwar:
As the current format seems to be picked up by the IDE as a link already, I am leaning towards closing this PR as well as the issue. 
Can someone please describe the circumstances in which there is an issue with the links.

cortinico:
> As the current format seems to be picked up by the IDE as a link already, I am leaning towards closing this PR as well as the issue.

I believe this is the case (at least for me). Ideally having `file://` will make it even more recognizable for tools that consumer this output, but that's probably unnecessary for IntelliJ/Android Studio.

chao2zhang:
Let's close this PR while keeping #4041. In case the file location hyperlink of IntelliJ get buggy in the future, we should definitely revisit this.

PR #4063, opened 2021-08-21
Remove deprecated KotlinExtension

3flex:
This was [deprecated ](https://github.com/detekt/detekt/commit/695a6fc9bc665db0d54ee7151070541e5df7198f)in detekt 1.3.0 - it's time to say goodbye.


chao2zhang (approved):

BraisGabin (approved):

cortinico:
> it's time to say goodbye.

Agree.
As a side note: I'm even unsure of how users would use this API in their gradle file. Was it something like?

```
this.detekt {
    ...
}
```

Asking as we probably want to add something to the release notes about it

3flex:
> Was it something like?

It was just:

```kotlin
detekt {

}
```

The problem is the plugin wasn't being applied correctly by some users, so the [type-safe accessors generated for the Kotlin DSL](https://docs.gradle.org/current/userguide/kotlin_dsl.html#type-safe-accessors) weren't being generated. This is a configuration problem, not something to do with the plugin, but this helper provided a workaround.

The helper isn't necessary if the plugin is applied properly. Users should be familiar with Gradle's [Kotlin DSL primer](https://docs.gradle.org/current/userguide/kotlin_dsl.html). I'm not sure if we need something in the release notes since the deprecation message has had a link to this document since 2019.

PR #4065, opened 2021-08-21
Update dependencies

3flex:


cortinico (approved):

cortinico:
For the future: let's try to update dependencies in separate PRs as they will appear as separate entries in the Changelog, and it's easier to revert some of them if something goes on fire.

PR #4074, opened 2021-08-26
Setup Gradle functional tests

3flex:
This is a step on the way to separating Gradle TestKit tests from the other Gradle plugin tests (see #3778).

This PR sets up the Gradle build to support the new structure, and moves tests that use TestKit but _don't_ use dry-run mode into `src/functionalTest`. It also moves shared code into `src/testFixtures` so they can be used by both functional tests and standard tests while that's still necessary.

There's a new task `functionalTest` which runs all tests in `src/functionalTest`. Both `functionalTest` and `test` are dependencies of `check` task and therefore also the `build` task.


chao2zhang:
IntelliJ IDE does not really like addition test source set like `intTest`. I just applied this PR locally and the problem mentioned in https://github.com/detekt/detekt/pull/3504 still persists. 

I am thinking if we should start experimenting with extracting integration tests into a separate module as discussed in https://github.com/detekt/detekt/issues/3324#issuecomment-751909633.

3flex:
> I just applied this PR locally and the problem mentioned in #3504 still persists.

I'm not seeing that - which IDE specifically (IntelliJ, Android Studio) & which version? Can you try deleting the `.idea` directory and clean IDE cache?

I have no red underline in any files under /src/intTest (IntelliJ IDEA 2021.2.1) with almost all settings at defaults:
image

chao2zhang:
After `./gradlew clean` and removing `.idea`, the IDE does not seem to `intTest` as a test module (With green background).
image
Previously IntelliJ cannot infer the module dependencies correctly, and now after regenerating the `.idea` folder, it seems to work for me for the most part.

3flex:
I think as long as the Spek IDE plugin can pick up the tests, Gradle executes them and IDE indexing works, the fact that it's not specifically identified as green/test module isn't a showstopper (it does bug me as well though! But this setup is simpler to me than having a whole new module for integration tests)

We can have others weigh in of course.

detekt-gradle-plugin/build.gradle.kts
```diff
@@ -9,18 +10,28 @@ repositories {
...
+
+val runtimePlugin: Configuration by configurations.creating
+val intTestImplementation: Configuration by configurations.getting {
+    extendsFrom(configurations.testImplementation.get())
+}
+val intTestRuntimeOnly: Configuration by configurations.getting {
+    extendsFrom(configurations.testRuntimeOnly.get())
+}
+
+configurations.compileOnly { extendsFrom(runtimePlugin) }
```
cortinico, line 25: The name is confusing imho. `compileOnly` is the opposite to `runtime` (i.e. we're not imposing this dependency or in other words that such dependency is not always necessary to run the project).

`pluginCompileOnly` would work better for me.

detekt-gradle-plugin/build.gradle.kts
```diff
@@ -73,4 +85,21 @@ tasks {
...
+    }
+
+    check {
+        dependsOn(integrationTest)
+    }
+}
+
+with(components["java"] as AdhocComponentWithVariants) {
+    withVariantsFromConfiguration(configurations["testFixturesApiElements"]) { skip() }
+    withVariantsFromConfiguration(configurations["testFixturesRuntimeElements"]) { skip() }
```
cortinico, line 104: What is this doing? Maybe add a comment?

cortinico:
> the fact that it's not specifically identified as green/test module isn't a showstopper

Agree. Also I believe you can workaround the coloring issue with something like:

```
idea.module {
    sourceDirs.remove(file("src/intTest/kotlin"))
    testSourceDirs.add(file("src/intTest/kotlin"))
}
```

Sadly I wasn't able to let it work (see https://youtrack.jetbrains.com/issue/IDEA-151925). If you folks look into it let me know 🙏

BraisGabin (approved):

3flex:
Converting back to draft and removing from 1.19.0 milestone (for now) - Gradle 7.3 is introducing support for test suites which I believe supersedes the currently recommended configuration I've used here.

It would be better to wait to see what that looks like and use the newer, hopefully simpler setup.

chao2zhang:
@3flex Would you mind sharing some of the pointers how Gradle 7.3 changes the test suites? (Assuming this is for Gradle integration test only)

cortinico:
Great stuff! What's the status on the IDE support?

detekt-gradle-plugin/build.gradle.kts
```diff
@@ -1,6 +1,8 @@
 plugins {
     id("module")
     `java-gradle-plugin`
+    `java-test-fixtures`
+    idea
```
cortinico, line 5: Why is this needed?

detekt-gradle-plugin/build.gradle.kts
```diff
@@ -9,21 +11,44 @@ repositories {
...
 }
 
-val intTest: Configuration by configurations.creating
+testing {
+    suites {
+        getByName("test", JvmTestSuite::class) {
+            dependencies {
+                implementation(projects.detektTestUtils)
+                implementation(libs.assertj)
+                implementation(libs.spek.dsl)
```
chao2zhang, line 20: Since we already define these implementation dependencies in the project `dependencies` block, we don't need line 19-20 and line 34-35.

detekt-gradle-plugin/build.gradle.kts
chao2zhang: Actually this is listed in the Gradle doc https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures

chao2zhang (approved):

detekt-gradle-plugin/build.gradle.kts
3flex: Sorry I don't follow... removing those lines fails the build on `test` and `functionalTest` tasks

chao2zhang: My bad. I thought there is a way to share test implementation dependencies between the test suite.

PR #4075, opened 2021-08-27
Convert DetektJvmSpec to use ProjectBuilder

3flex:
See #3778. This reduced execution of the DetektJvmSpec test by about 25% on my machine.

### DIFF
--- a/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
+++ b/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
@@ -3,6 +3,7 @@ package io.gitlab.arturbosch.detekt
 import io.gitlab.arturbosch.detekt.invoke.CliArgument
 import io.gitlab.arturbosch.detekt.testkit.DslGradleRunner
 import io.gitlab.arturbosch.detekt.testkit.ProjectLayout
+import io.gitlab.arturbosch.detekt.testkit.triggerEvaluation
 import org.assertj.core.api.Assertions.assertThat
 import org.gradle.kotlin.dsl.apply
 import org.gradle.kotlin.dsl.repositories
@@ -37,7 +38,7 @@ object DetektJvmSpec : Spek({
             it("configures detekt type resolution task main") {
                 val project = gradleRunner.buildProject()
 
-                project.getTasksByName("detektMain", true) // trigger project evaluation
+                project.triggerEvaluation()
 
                 val detektTask = project.tasks.getByPath("detektMain") as Detekt
                 val argumentString = detektTask.arguments.flatMap(CliArgument::toArgument).joinToString(" ")
@@ -52,7 +53,7 @@ object DetektJvmSpec : Spek({
             it("configures detekt type resolution task test") {
                 val project = gradleRunner.buildProject()
 
-                project.getTasksByName("detektTest", true) // trigger project evaluation
+                project.triggerEvaluation()
 
                 val detektTask = project.tasks.getByPath("detektTest") as Detekt
                 val argumentString = detektTask.arguments.flatMap(CliArgument::toArgument).joinToString(" ")
--- a/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
+++ b/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
@@ -1,8 +1,12 @@
 package io.gitlab.arturbosch.detekt
 
+import io.gitlab.arturbosch.detekt.invoke.CliArgument
 import io.gitlab.arturbosch.detekt.testkit.DslGradleRunner
 import io.gitlab.arturbosch.detekt.testkit.ProjectLayout
 import org.assertj.core.api.Assertions.assertThat
+import org.gradle.kotlin.dsl.apply
+import org.gradle.kotlin.dsl.repositories
+import org.jetbrains.kotlin.gradle.plugin.KotlinPluginWrapper
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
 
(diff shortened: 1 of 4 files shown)


3flex:
> Is it fair to assume that the improvement comes from changing from groovy interpreter to Kotlin compiler?

It's from using [`ProjectBuilder`](https://docs.gradle.org/current/javadoc/org/gradle/testfixtures/ProjectBuilder.html) instead of Gradle's TestKit.

TestKit runs a full Gradle build and spins up its own daemon. ProjectBuilder is much more limited, and can't be used for functional tests, because you can't execute tasks. But for checking things like whether task input & output values are set correctly it's suitable, and faster.

chao2zhang (approved):

schalkms (approved):

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
```diff
@@ -13,9 +15,17 @@ class DslGradleRunner @Suppress("LongParameterList") constructor(
...
     val gradleVersionOrNone: String? = null,
     val dryRun: Boolean = false
 ) {
+    constructor(
+        projectLayout: ProjectLayout,
+        buildFileName: String,
+        projectScript: Project.() -> Unit,
+        configFileOrNone: String? = null,
+        baselineFiles: List<String> = emptyList(),
+    ) : this(projectLayout, buildFileName, "", configFileOrNone, baselineFiles, projectScript)
```
cortinico, line 28: This is confusing to me. Why do we need an extra ctor where the new paramters is in 3rd position?

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
```diff
@@ -13,9 +15,17 @@ class DslGradleRunner @Suppress("LongParameterList") constructor(
     val mainBuildFileContent: String,
     val configFileOrNone: String? = null,
     val baselineFiles: List<String> = emptyList(),
+    val projectScript: (Project.() -> Unit)? = null,
```
cortinico, line 18: Why this is not last? You could benefit also of trailing lambda syntax

cortinico: Also does it need to be nullable? Can it just be an empty lambda?

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
```diff
@@ -109,6 +119,18 @@ class DslGradleRunner @Suppress("LongParameterList") constructor(
     private val Submodule.moduleRoot: File
         get() = File(rootDir, name).apply { mkdirs() }
 
+    fun buildProject(): Project {
+        val project = ProjectBuilder.builder()
+            .withProjectDir(rootDir)
+            .build()
+
+        requireNotNull(projectScript)
```
cortinico, line 127: You can remove this if `projectScript` becomes non nullable.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
```diff
@@ -109,6 +119,18 @@ class DslGradleRunner @Suppress("LongParameterList") constructor(
     private val Submodule.moduleRoot: File
         get() = File(rootDir, name).apply { mkdirs() }
 
+    fun buildProject(): Project {
```
cortinico, line 122:
```suggestion
    fun buildProject(): Project = ProjectBuilder.builder()
        .withProjectDir(rootDir)
        .build()
        .apply { projectScript.invoke(this) }
```

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
```diff
@@ -13,45 +19,48 @@ object DetektJvmSpec : Spek({
...
-            gradleRunner.runTasksAndCheckResult(":detektMain") { buildResult ->
-                assertThat(buildResult.output).containsPattern("""--baseline \S*[/\\]detekt-baseline-main.xml """)
-                assertThat(buildResult.output).contains("--report xml:")
-                assertThat(buildResult.output).contains("--report sarif:")
-                assertThat(buildResult.output).doesNotContain("--report txt:")
-                assertThat(buildResult.output).contains("--classpath")
-            }
+            val project = gradleRunner.buildProject()
+
+            project.getTasksByName("detektMain", true) // trigger project evaluation
```
cortinico, line 39: Is the test not working without this? Isn't the `getByPath` below sufficient?

cortinico (approved):
Great work 👍 

Just one side note: now we're moving from actually running Gradle tasks to just setting up a project and asserting that the tasks are configured properly.

Not a big deal but if the task blows up during execution, the former task would have found it. This one not. Something to keep in mind 👍

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
chao2zhang: Good catch

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
3flex: Unfortunately not - currently `project.afterEvaluate` is used throughout the detekt plugin to register tasks, so without triggering the project evaluation in the test, getByPath throws UnknownTaskException.

cortinico: How do you feel about having a:

```
fun Project.triggerEvaluation() { project.getTasksByName("tasks", true) }
```

BraisGabin (approved):

chao2zhang (approved):

PR #4043, opened 2021-08-14
Remove detekt-bom module

3flex:
Closes #3988

PR #4094, opened 2021-09-09
Refactor so detekt-gradle-plugin can be added as an included build

3flex:
Stacks on #4043 - ignore the first commit.

This refactors the build logic so it's possible to include detekt-gradle-plugin as either a composite build or a standard project module. Just replace `include("detekt-gradle-plugin")` with `includeBuild("detekt-gradle-plugin")` in settings.gradle.kts to [define the composite build](https://docs.gradle.org/7.2/userguide/composite_builds.html#settings_defined_composite).


detekt-gradle-plugin/settings.gradle.kts
```diff
@@ -0,0 +1,16 @@
...
+
+dependencyResolutionManagement {
+    versionCatalogs {
+        create("libs") {
+            from(files("../gradle/libs.versions.toml"))
+        }
+    }
+}
+
+include("detekt-test-utils")
```
cortinico, line 13: Are there any problmes in including a module both in the included build and in the external build?

3flex: Not that I'm aware of.

This settings file is only used when the plugin module is used as an included build anyway, so it won't affect the standard build.

There's another option, which means changing `testImplementation(projects.detektTestUtils)` to `testImplementation("io.gitlab.arturbosch.detekt:detekt-test-utils")` in the build file. Then the projects don't have to be included in the settings file. This works whether the middle is a standard module or an included build.

3flex:
Converting to draft as there's more to be done before this works as advertised.

detekt-gradle-plugin/settings.gradle.kts
cortinico: > There's another option, which means changing testImplementation(projects.detektTestUtils) to testImplementation("io.gitlab.arturbosch.detekt:detekt-test-utils") in the build file. Then the projects don't have to be included in the settings file. This works whether the project is a standard module or an included build.

I would opt for this as it gives us a little bit more of flexibility.

settings.gradle.kts
```diff
@@ -4,37 +4,35 @@ pluginManagement {
...
+include("detekt-rules-empty")
+include("detekt-rules-errorprone")
+include("detekt-rules-exceptions")
+include("detekt-rules-naming")
+include("detekt-rules-performance")
+include("detekt-rules-style")
+include("detekt-sample-extensions")
+include("detekt-test")
+include("detekt-test-utils")
+include("detekt-tooling")
```
BraisGabin, line 35: This is just curiosity: Does this change any behaviour?

detekt-gradle-plugin/settings.gradle.kts
BraisGabin: But using `testImplementation("io.gitlab.arturbosch.detekt:detekt-test-utils")` If I change something in this module I can't use that update until the next release of the library, right?

3flex: These `include` declarations have now been dropped from the settings file, and a slight refactor removed the need for `detekt-test-utils` dependency. It didn't seem possible to keep the dependency while allowing an easy toggle between including the Gradle plugin as either a regular module or an included build.

settings.gradle.kts
3flex: No change to behaviour, both call https://docs.gradle.org/current/javadoc/org/gradle/api/initialization/Settings.html#include-java.lang.String...-

3flex:
I'm not loving the fact that some of the build config had to move back to the root project's build file instead of a precompiled script plugin in `build-logic`, but it wouldn't work as a composite build without it.

3flex:
LGTM? I think this is as good as it will get.

I'm using these commits and a composite build myself for a couple of things and it's working really well.

build.gradle.kts
```diff
@@ -2,16 +2,57 @@ import io.gitlab.arturbosch.detekt.Detekt
...
+
+nexusStaging {
+    packageGroup = "io.gitlab.arturbosch"
+    stagingProfileId = "1d8efc8232c5c"
+    username = findProperty("sonatypeUsername")
+        ?.toString()
+        ?: System.getenv("MAVEN_CENTRAL_USER")
+    password = findProperty("sonatypePassword")
+        ?.toString()
+        ?: System.getenv("MAVEN_CENTRAL_PW")
```
BraisGabin, line 20: We are moving this here because we can't apply this in the gradle plugin when we use composite compilation, right?

Could we create another plugin in `build-logic` so we don't need to have this here?

build-logic/src/main/kotlin/detekt.gradle.kts
```diff
@@ -1,36 +0,0 @@
...
-        txt.enabled = true
-        sarif.enabled = true
-    }
-}
-
-dependencies {
-    detekt(project(":detekt-cli"))
-    detektPlugins(project(":custom-checks"))
-    detektPlugins(project(":detekt-formatting"))
-}
```
BraisGabin, line 36: Why we can't keep this here?

BraisGabin (approved):
To be honest I'm a bit lost here. So I give my approve because it seems to work.

build-logic/src/main/kotlin/detekt.gradle.kts
3flex: When `detekt-gradle-plugin` is an included build, and it includes `build-logic`, the `project` dependencies can't be resolved. So they need to be moved out of `build-logic` - and applying them to subprojects from the root build file is the best way to do that from a single location.

build.gradle.kts
3flex: I'd moved it out of `build-logic` as it wasn't shared build config like mostly everything else that remained, but there's a home for it in `releasing` so I've moved it there instead. The config only needs to be applied to the root build file, and `releasing` is only applied in the root file so this makes sense to me.

cortinico:
> Just replace `include("detekt-gradle-plugin")` with `includeBuild("detekt-gradle-plugin")` in settings.gradle.kts to [define the composite build](https://docs.gradle.org/7.2/userguide/composite_builds.html#settings_defined_composite).

Is there a reason why you haven't replaced this? That would allow us to don't have to rely on the preivously released version of `detekt-gradle-plugin` for the release and so on. Or are you afraid of a performance impact of the composite build?

3flex:
I'm not too worried about perf personally, but I think some others are or were.

There's also the issue that running, for example, `gradlew build` in the root project won't execute the `detekt-gradle-plugin:build` task without wiring things up as described here: https://docs.gradle.org/current/userguide/composite_builds.html#included_build_task_dependencies

A separate PR could be opened to link up the lifecycle tasks (build, check, run) to the same tasks in the included build so it works as expected.

Issue #4068, opened 2021-08-23
Implement a more advanced way to ignore/suppress issues


BraisGabin:
Detekt should have an advanced way to exclude/ignore/suppress issues by confiugration. And it should be generic so it could be applied to any rule. Right now we just allow to ignore files completely.

## Context
Right now we have this kind of configurations build-in the rules that clearly need this. Examples:

- LongMethod.ignoreAnnotated
- LongParameterList.ignoreAnnotated
- ForbiddenPublicDataClass.ignorePackages
- ForbiddenVoid.ignoreOverridden
- MagicNumber.ignoreEnums
- ReturnCount.excludedFunctions
- ...

Detekt is getting more and more users so the use cases are growing and we get more and more request to adapt our rules for these use cases, Examples: #4062, #4061 and #2231 (there are lots more)

I don't know exactly how this api should look like but I think that detekt really need this feature.

So, any ideas about how to implement this or how the api should be?

cortinico:
I thought a bit about this issue. I believe a generic `ignoreAnnotated` is not that hard to implement. 

Whenever we report a finding we need to make sure we report the `ktElement` correctly:

https://github.com/detekt/detekt/blob/198c53cdcd61c6004911352bb1855917420e23c1/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Entity.kt#L14-L19

On that element we can easily traverse the AST and check if him or any of the parent is annotated with any of the `ignoreAnnotated` annotation.

I believe annotations exceptions are the more critical to address right now, with the advent of Compose. I'm unsure about the other exclusion mechanism and can't think at anything really generic that would be worth to implement (happy to hear the community opinion on this).

BraisGabin:
I can imagine `ignoreInFunction` `IgnoreInClass`. Same as `ignoreAnnotated`. It traverse the AST and check if it is inside a function with that name or a class with that name. There are rules where some of them will not applied, but that should not be a problem. If you add `ignoreInFunction` in `LongClass` it will do nothing because the class (usually) is not inside a function.

cortinico:
> The required change is to incorporate the rule configuration into the traversal.

Agree.
The point here is deciding what we want to inspect for (annotations, packages, etc.)

BraisGabin:
[...]

PR #4102, opened 2021-09-11
Implement ignoreAnnotated as a core feature

### LINKED ISSUE #2231: UnusedPrivateMember should ignore methods annotated with @PostConstruct / @PreDestroy
## Expected Behavior
UnusedPrivateMember should ignore methods annotated with @PostConstruct / @PreDestroy

## Current Behavior
Currently any method that is private and is annotated with @PostConstruct / @PreDestroy is still reported (even though it will be called via reflection at runtime by Spring anyway).

## Context
I get that I could rename all of the methods to be called like "postConstruct" / "preDestroy" and then add them to the excluded names, but it will harm the readability of the code (methods have a name so that it can define what the method does and not what it should be called).

### LINKED ISSUE #4062: MagicNumber: Support for `ignoreAnnotated`
Add `ignoreAnnotated` property for MagicNumber. 

## Context
Just like FunctionNaming has `ignoreAnnotated` that fits well with Composable function. In Preview (annotated with `@Preview`) functions Jetpack Compose, usually we put Magic numbers for purely preview purposes.

BraisGabin:
Discussion here: #4068

This id the first step for a more advance way of suppress findings. It implements the most common request one and probably the easiest: `ignoreAnnotated`. And I implemented it in a way that adding more should be easy.

The main problem that I see with this implementation is the discoverability. As an user this is an interesting feature but it is really difficult to find it. For sure we should add somerhing in our documentation but I think that it would not be enought. Any ideas here?

In this PR I'm removing some default values. For that reason I created #4101 before. We don't want those default values, they are library specific.

And I changed `LongParameterList` a bit too. It was using the same configuration to do two things: 1) ignore a function completely for an annotation and 2) ignore the annotated parameter for the total count. So I removed the first part of that feature to relay on the common one and changed the second one to use `ignoreAnnotatedParameter` instead of `ignoreAnnotated`. This is a breaking change.

Right now I'm just treating the annotations as a name, I'm not using the full qualifier. Do you think that it is necesary to do that? I think that we could keep it like this for now and we could change it later if someone miss that. We should not have problem extending it.

Closes #2231
Closes #4062

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
@@ -20,6 +20,7 @@ import io.gitlab.arturbosch.detekt.core.config.DisabledAutoCorrectConfig
 import io.gitlab.arturbosch.detekt.core.rules.associateRuleIdsToRuleSetIds
 import io.gitlab.arturbosch.detekt.core.rules.isActive
 import io.gitlab.arturbosch.detekt.core.rules.shouldAnalyzeFile
+import io.gitlab.arturbosch.detekt.core.suppressors.getSuppressors
 import org.jetbrains.kotlin.config.languageVersionSettings
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.resolve.BindingContext
@@ -119,7 +120,7 @@ internal class Analyzer(
         fun executeRules(rules: List<BaseRule>) {
             for (rule in rules) {
                 rule.visitFile(file, bindingContext, compilerResources)
-                for (finding in rule.findings) {
+                for (finding in filterSuppressedFindings(rule)) {
                     val mappedRuleSet = checkNotNull(ruleIdsToRuleSetIds[finding.id]) {
                         "Mapping for '${finding.id}' expected."
                     }
@@ -133,6 +134,15 @@ internal class Analyzer(
         executeRules(otherRules)
 
         return result
+    }
+}
+
+private fun filterSuppressedFindings(rule: BaseRule): List<Finding> {
+    val suppressors = getSuppressors(rule)
+    return if (suppressors.isNotEmpty()) {
+        rule.findings.filter { finding -> !suppressors.any { suppressor -> suppressor(finding) } }
+    } else {
+        rule.findings
     }
 }
 
--- a/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressorSpec.kt
+++ b/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressorSpec.kt
@@ -1,0 +1,281 @@
+package io.gitlab.arturbosch.detekt.core.suppressors
+
+import io.github.detekt.psi.FilePath
+import io.github.detekt.test.utils.compileContentForTest
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.ConfigAware
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Finding
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Location
+import io.gitlab.arturbosch.detekt.api.RuleId
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.SourceLocation
+import io.gitlab.arturbosch.detekt.api.TextLocation
+import io.gitlab.arturbosch.detekt.test.TestConfig
+import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtElement
+import org.jetbrains.kotlin.psi.KtFunction
+import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
+import org.jetbrains.kotlin.psi.psiUtil.findFunctionByName
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+import java.nio.file.Paths
+
+class AnnotationSuppressorSpec : Spek({
+
+    describe("AnnotationSuppressorFactory") {
+        it("Factory returns null if ignoreAnnotated is not set") {
+            val suppressor = annotationSuppressorFactory(buildConfigAware(/* empty */))
+
+            assertThat(suppressor).isNull()
+        }
+
+        it("Factory returns null if ignoreAnnotated is set to empty") {
+            val suppressor = annotationSuppressorFactory(
+                buildConfigAware("ignoreAnnotated" to emptyList<String>())
+            )
+
+            assertThat(suppressor).isNull()
+        }
+
+        it("Factory returns not null if ignoreAnnotated is set to a not empty list") {
+            val suppressor = annotationSuppressorFactory(
+                buildConfigAware("ignoreAnnotated" to listOf("Composable"))
+            )
+
+            assertThat(suppressor).isNotNull()
+        }
+    }
+
+    describe("AnnotationSuppressor") {
+        val suppressor by memoized {
+            annotationSuppressorFactory(buildConfigAware("ignoreAnnotated" to listOf("Composable")))!!
+        }
+
+        it("If KtElement is null it returns false") {
+            assertThat(suppressor(buildFinding(element = null))).isFalse()
+        }
+
+        context("If annotation is at file level") {
+            val root by memoized {
+                compileContentForTest(
+                    """
+                    @file:Composable
+
+                    class OneClass {
+                        fun function(parameter: String) {
+                            val a = 0
+                        }
+                    }
+
+                    fun topLevelFunction() = Unit
+                    """.trimIndent()
+                )
+            }
+
+            it("If reports root it returns true") {
+                assertThat(suppressor(buildFinding(element = root))).isTrue()
+            }
+
+            it("If reports class it returns true") {
+                val ktClass = root.findChildByClass(KtClass::class.java)!!
+
+                assertThat(suppressor(buildFinding(element = ktClass))).isTrue()
+            }
+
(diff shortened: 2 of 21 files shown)


cortinico (approved):
Great work! I believe on the discoverability we can just have a page in the documentation and link it in the release notes

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -136,6 +137,15 @@ internal class Analyzer(
     }
 }
 
+private fun filterSuppressedFindings(rule: BaseRule): List<Finding> {
+    val suppressors = getSuppressors(rule)
```
marschwar, line 141: If I read this correctly, the list of suppressors would be generated for each and every file even though it only depends on the rule. This will create a lot of unnecessary objects in the process.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -119,7 +120,7 @@ internal class Analyzer(
         fun executeRules(rules: List<BaseRule>) {
             for (rule in rules) {
                 rule.visitFile(file, bindingContext, compilerResources)
-                for (finding in rule.findings) {
+                for (finding in filterSuppressedFindings(rule)) {
```
marschwar, line 123: Could you please add a test that verifies that the findings actually are being suppressed. The Analyzer seems rather tricky to test. Maybe it is doing too much (even before your change).

marschwar (commented):
I like where this is headed.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
BraisGabin: Yes. But the same happens right now with the rules. We create all the rules for each file. Changing this is a big refactor. It will be fixed in 2.0

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
BraisGabin: I'll check if that's "possible". I think that this class has really few tests. But as you say this file does far too much. This is the core of detekt and it grew out of control.

BraisGabin: Ok, I managed to add some tests. What do you think? Should I cover more cases?

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressorSpec.kt
```diff
@@ -0,0 +1,282 @@
...
+import org.jetbrains.kotlin.psi.KtElement
+import org.jetbrains.kotlin.psi.KtFunction
+import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
+import org.jetbrains.kotlin.psi.psiUtil.findFunctionByName
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+import java.nio.file.Paths
+
+@Suppress("UnsafeCallOnNullableType")
```
BraisGabin, line 29: Should we disable this rule directly for the tests? I think that it's safe to use `!!` in the tests.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/SupressorsSet.kt
```diff
@@ -0,0 +1,20 @@
+package io.gitlab.arturbosch.detekt.core.suppressors
```
chao2zhang, line 1: A minor issue, but could you explain a bit why the filename is `SupressorsSet` not `Suppressors`?

Additionally, there is a typo of missing `p`.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
```diff
@@ -61,27 +60,20 @@ class LongParameterList(config: Config = Config.empty) : Rule(config) {
...
 
     @Configuration(
-        "ignore long parameters list for constructors, functions or their parameters in the " +
-            "context of these annotation class names; (e.g. ['Inject', 'Module', 'Suppress', 'Value']); " +
-            "the most common cases are for dependency injection where constructors are annotated with `@Inject` " +
-            "or parameters are annotated with `@Value` and should not be counted for the rule to trigger"
+        "ignore the annotated parameters for the count (e.g. `fun foo(@Value bar: Int)` would not be counted"
     )
-    private val ignoreAnnotated: List<String> by config(emptyList<String>()) { list ->
+    private val ignoreAnnotatedParameter: List<String> by config(emptyList<String>()) { list ->
```
chao2zhang, line 65: Would this be a backward-incompatible change since we changed the semantical meaning of `ignoreAnnotated`? We should probably highlight this in the release notes in the next version.

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressorSpec.kt
chao2zhang: I would agree with this.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
chao2zhang: It is scary to learn about that.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/SupressorsSet.kt
```diff
@@ -0,0 +1,20 @@
...
+import io.gitlab.arturbosch.detekt.api.ConfigAware
+import io.gitlab.arturbosch.detekt.api.Finding
+import io.gitlab.arturbosch.detekt.api.internal.BaseRule
+
+/**
+ * Given a Finding it decides if it should be suppressed (`true`) or not (`false`)
+ */
+typealias Suppressor = (Finding) -> Boolean
+
+internal fun getSuppressors(rule: BaseRule): List<Suppressor> {
```
chao2zhang, line 12: We only have one `Suppressor` instance in the returned list - Would you consider using the decorator pattern in case we have more `Suppressor`s or use `List<Suppressor>`?

chao2zhang (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
chao2zhang: Thanks for adding test cases, it looks like we have sufficient coverage here, and we should be good to go once jacoco agrees.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/SupressorsSet.kt
BraisGabin: Good catch! I renamed this classes/functions many times and I forgot to change this. Thanks!

detekt-core/src/test/resources/configs/config-value-type-correct-ignore-annotated.yml
```diff
@@ -0,0 +1,6 @@
+style:
+  MaxLineLength:
+    active: true
+    maxLineLength: 120
+    ignoreAnnotated:
+    - "AnAnnotation"
```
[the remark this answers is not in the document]
BraisGabin: I think that it's valid but I updated just in case.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
BraisGabin: No, this is a "minor" behaviour change but for sure something that we should not in our release notes. I wrote about that in the PR description:

> And I changed `LongParameterList` a bit too. It was using the same configuration to do two things: 1) ignore a function completely for an annotation and 2) ignore the annotated parameter for the total count. So I removed the first part of that feature to relay on the common one and changed the second one to use `ignoreAnnotatedParameter` instead of `ignoreAnnotated`. This is a breaking change.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/SupressorsSet.kt
BraisGabin: I don't know what do you mean by "decorator pattern". 

Right now there is only one suppressor but my idea is to have more once we merge this. There is a short list of examples at #4068

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressorSpec.kt
BraisGabin: #4118 and #4119

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/SupressorsSet.kt
chao2zhang:
```
open abstract class Suppressor : Predicate<Finding> {
}

operator fun Suppressor.plus(suppressor: Suppressor): Suppressor {
    return object : Suppressor() {
        override fun test(t: Finding) = this@plus.test(t) || suppressor.test(t)
    }
}

object DefaultSuppressor : Suppressor() {
    override fun test(t: Finding) = true
}
```

I meant like this where we can compose suppressors in a chain (Probably should be called as a chain of responsibility). The benefit of so is that we can simplify the suppressor implementation that to the `Analyzer`, the suppressor function is always a simple function `(Finding) -> Boolean`.

BraisGabin: I think that I prefer to handle this as a `List`. The good part is that all of these isn't an public API so we could change it if we see problems with the current implementation.

PR #4103, opened 2021-09-11
Fix indent issues with continaution indent

BraisGabin:
I don't know if I'm the only one that when there are some continuation indents in a file, if I autoformat them with IntelliJ they are indented with 8 spaces and when that code is pushed the ktlint rules complain about it.

This PR fixes that. It defines the desired indent so IntelliJ and ktlint are aligined.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,6 +6,7 @@ insert_final_newline = true
 
 [*.{kt, groovy}]
 indent_style = space
+indent_size = 4
 
 [*.yml]
 indent_style = space


3flex:
You're not alone, I've also noticed this - but isn't this just changing the IntelliJ config to make ktlint happy? It should be the other way around. I think ktlint has problems with continuation indents so my preference would be to turn off that rule if it's not working correctly instead of changing the default IntelliJ code style.

cortinico (approved):

BraisGabin:
> You're not alone, I've also noticed this - but isn't this just changing the IntelliJ config to make ktlint happy? It should be the other way around. I think ktlint has problems with continuation indents so my preference would be to turn off that rule if it's not working correctly instead of changing the default IntelliJ code style.

Well, that depends if we want to have continuation indentation of size 8 or 4. I prefer them to be of size 4. And because, right now, all the code uses identation of size 4 (because of ktlint) I think that it's ok to continue with indent of 4.

Issue #3067, opened 2020-09-11
Wrong error:  "error: cannot access 'foo': it is internal in 'com.example.bar'"


yogurtearl:
`internal` in kotlin means it can't be accessed outside the current module.
https://kotlinlang.org/docs/reference/visibility-modifiers.html#modules

In gradle projects, a "module" is a gradle source set, with the exception that the `test` source set can access the internal declarations of `main`

## Expected Behavior
If gradle/kotlinc don't complain about `internal`, then detekt shouldn't complain either.
detekt should match gradle/kotlinc here.

## Observed Behavior
If a `Test` in `/test` accesses a `internal` `foo` extension function defined in `/main`, then detekt complains `error: cannot access 'foo': it is internal in 'com.example.bar'`

## Your Environment

* Version of detekt used: 1.12.0
* Version of Gradle used (if applicable): 6.6.1
* Operating System and version: macOS 10.15

cortinico:
> then detekt complains `error: cannot access 'foo': it is internal in 'com.example.bar'`

Is this reported by a Detekt rule? If yes, which one?
Can you provide more context to help reproduce?

arturbosch:
@yogurtearl could you try adding `languageVersion = "1.4"` to the detekt tasks via `tasks.withType<Detekt> { ... }` and see if it helps? Thanks

chao2zhang:
`error: cannot access 'foo': it is internal in 'com.example.bar'` is a compiler error issued when running detekt with type resolution. It does not affect the result of detekt. This is reproducible when running `./gradlew detektTest` in the detekt repo itself. Hence I would suggest reopening this issue.

However, this means that our rules may not run correctly. There is a compiler option `-Xfriend-paths` and we can also programmatically `CompilerConfiguration.put(JVMConfigurationKeys.FRIEND_PATHS, friendPaths)`. I think this is the right path but it turns out we need more investigation more to set the friend paths correctly.

chao2zhang:
After digging into the kotlin compiler code a little bit, we can conclude the status quo as:

- We need to tell Kotlin compiler about friend paths
- In `https://github.com/chao2zhang/detekt/tree/internal-fix`, I write the friends path to `JVMConfigurationKeys.FRIEND_PATHS`
[...]

Issue #3639, opened 2021-03-31
Gradle-Plugin: Tasks with TypeResolution are not using the same compiler options as configured in the project


mervyn-mccreight:
Hi folks :-)

When using tasks like e.g. `detektMain` I figured out that it displays a lot of compiler warnings.

E.g. I have opted-in kotlin.time by passing a compiler-flag like the following in my `build.gradle.kts`
```kotlin
tasks.withType<KotlinCompile> {
    kotlinOptions.freeCompilerArgs += "-Xuse-experimental=kotlin.time.ExperimentalTime"
}
```
so I do not have to opt-in it all over the source-code.

`detektMain` does not seem to pick this up and is constantly complaining about it with messages like this:

```
<path/to/source/File.kt>: error: this declaration is experimental and its usage must be marked with '@kotlin.time.ExperimentalTime' or '@OptIn(kotlin.time.ExperimentalTime::class)
```

which does not look very nice. Is the validation even working for these files?
The task itself "passes" if no source-file violates any rules, but in my perception it is misleading that the terminal is full of red text containing the term "error" if everything is fine.

Is it intended that detekt is not using the same compiler flag as the project uses to check against rules using type-resolution?
Or am I missing some configuration so that detekt does this?

_Originally posted by @mervyn-mccreight in https://github.com/detekt/detekt/discussions/3629_

chao2zhang:
Hello @mervyn-mccreight, these compiler warnings are issued by kotlin compiler, and detekt internally also uses the kotlin compiler for a lightweight compilation pass. This indicate that detekt may not pick up the compiler flags specified in your projects.

These are purely command-line compiler warnings and errors and should not affect the result of the detekt Gradle task. In the long term, we would need to move Detekt to compiler plugin like https://github.com/detekt/detekt-compiler-plugin but in the short term, we can also expose options to allow addition compiler flags in our detekt gradle task.

cortinico:
This will be fixed by #4106

PR #4106, opened 2021-09-12
Allow passing extra flags to Kotlin compiler environment

3flex:
This PR adds a catch-all ["main parameter"](https://jcommander.org/#_main_parameter) called `freeCompilerArgs` to the CLI. This is a nameless parameter which captures all input on the CLI that isn't matched by one of the existing parameters.

An equivalent task parameter has been added to the `Detekt` task to allow the Gradle plugin to pass those flags to the CLI.

This capability is being used to capture extra flags that detekt doesn't natively support so they can be passed through to the Kotlin compiler environment as required. For this PR, support was added that allows the module friend paths to be used when setting up the Kotlin core environment, which partially resolves issue #3067 (there is still some automation to be added in the Gradle plugin before this can be closed). Also see #3639 - again, not actually fixed by this PR, but it does add some of the framework needed to fix it.

Additionally https://github.com/detekt/detekt/issues/3684 would be fixed once the Gradle plugin is next released and the commented lines added to `build-logic/src/main/kotlin/detekt.gradle.kts` can be uncommented.

I'm looking for feedback before proceeding further.

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
@@ -68,6 +68,7 @@ internal fun CliArgs.createSpec(output: Appendable, error: Appendable): Processi
             jvmTarget = args.jvmTarget.description
             languageVersion = args.languageVersion?.versionString
             classpath = args.classpath?.trim()
+            freeCompilerArgs = args.freeCompilerArgs
         }
     }
 }
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
@@ -6,6 +6,7 @@ import io.github.detekt.tooling.api.spec.CompilerSpec
 import io.github.detekt.tooling.api.spec.ProjectSpec
 import org.jetbrains.kotlin.cli.common.arguments.K2JVMCompilerArguments
 import org.jetbrains.kotlin.cli.common.arguments.parseCommandLineArguments
+import org.jetbrains.kotlin.cli.common.arguments.validateArguments
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.jetbrains.kotlin.com.intellij.openapi.Disposable
 import org.jetbrains.kotlin.com.intellij.openapi.util.Disposer
@@ -35,6 +36,8 @@ internal class EnvironmentFacade(
         val compilerArguments = K2JVMCompilerArguments().apply {
             parseCommandLineArguments(compilerSpec.freeCompilerArgs, this)
         }
+
+        validateArguments(compilerArguments.errors)?.let { throw IllegalStateException(it) }
 
         val compilerConfiguration = createCompilerConfiguration(
             projectSpec.inputPaths.toList(),
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
@@ -4,6 +4,8 @@ import io.github.detekt.parser.createCompilerConfiguration
 import io.github.detekt.parser.createKotlinCoreEnvironment
 import io.github.detekt.tooling.api.spec.CompilerSpec
 import io.github.detekt.tooling.api.spec.ProjectSpec
+import org.jetbrains.kotlin.cli.common.arguments.K2JVMCompilerArguments
+import org.jetbrains.kotlin.cli.common.arguments.parseCommandLineArguments
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.jetbrains.kotlin.com.intellij.openapi.Disposable
 import org.jetbrains.kotlin.com.intellij.openapi.util.Disposer
@@ -29,6 +31,10 @@ internal class EnvironmentFacade(
     override val classpath: List<String> = compilerSpec.classpathEntries()
 
     override val environment: KotlinCoreEnvironment by lazy {
+        val compilerArguments = K2JVMCompilerArguments().apply {
+            parseCommandLineArguments(compilerSpec.freeCompilerArgs, this)
+        }
+
         val compilerConfiguration = createCompilerConfiguration(
             projectSpec.inputPaths.toList(),
             classpath,
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
@@ -33,11 +33,11 @@ internal class EnvironmentFacade(
     override val classpath: List<String> = compilerSpec.classpathEntries()
 
     override val environment: KotlinCoreEnvironment by lazy {
-        val compilerArguments = K2JVMCompilerArguments().apply {
+        val jvmCompilerArguments = K2JVMCompilerArguments().apply {
             parseCommandLineArguments(compilerSpec.freeCompilerArgs, this)
         }
 
-        validateArguments(compilerArguments.errors)?.let { throw IllegalStateException(it) }
+        validateArguments(jvmCompilerArguments.errors)?.let { throw IllegalStateException(it) }
 
         val compilerConfiguration = createCompilerConfiguration(
             projectSpec.inputPaths.toList(),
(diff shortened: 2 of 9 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
```diff
@@ -29,13 +33,25 @@ internal class EnvironmentFacade(
     override val classpath: List<String> = compilerSpec.classpathEntries()
 
     override val environment: KotlinCoreEnvironment by lazy {
+        val compilerArguments = K2JVMCompilerArguments().apply {
+            parseCommandLineArguments(compilerSpec.freeCompilerArgs, this)
+        }
+
+        validateArguments(compilerArguments.errors)?.let { throw IllegalStateException(it) }
```
chao2zhang, line 40: We also need to support non-JVM targeting Kotlin, what do you think about adding the CLI parameter as `jvmFreeCompilerArguments`?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
```diff
@@ -29,13 +33,25 @@ internal class EnvironmentFacade(
...
             classpath,
             compilerSpec.parseLanguageVersion(),
             compilerSpec.parseJvmTarget()
         )
-        createKotlinCoreEnvironment(compilerConfiguration, disposable)
+
+        val env = createKotlinCoreEnvironment(compilerConfiguration, disposable)
+
+        val visibilityManager = ModuleVisibilityManager.SERVICE.getInstance(env.project)
+        compilerArguments.friendPaths?.forEach(visibilityManager::addFriendPath)
```
chao2zhang, line 52: Nice! This is the solution I was looking for in my last attempt.

cortinico (commented):
I'm unsure about the `freeCompilerArgs` naming. I think it might be confusing for users, especially if compared to the Kotlin's `freeCompilerArgs` property. I'm not sure users will understand that the two `freeCompilerArgs` are effectively adding flags to distinct compiler invocations (mostly because not all the users are aware of the Detekt internals).

Potentially I would just call it `compilerArgs`  but I also don't have a strong opinion about it. Looking for feedback.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -242,6 +248,7 @@ open class Detekt @Inject constructor(
             DisableDefaultRuleSetArgument(disableDefaultRuleSetsProp.getOrElse(false))
         )
         arguments.addAll(convertCustomReportsToArguments())
+        arguments.add(FreeArgs(freeCompilerArgs.get()))
```
BraisGabin, line 251: Should we add a test to ensure this?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
```diff
@@ -68,6 +68,7 @@ internal fun CliArgs.createSpec(output: Appendable, error: Appendable): Processi
             jvmTarget = args.jvmTarget.description
             languageVersion = args.languageVersion?.versionString
             classpath = args.classpath?.trim()
+            freeCompilerArgs = args.freeCompilerArgs
```
BraisGabin, line 71: Should we test this?

build-logic/src/main/kotlin/detekt.gradle.kts
```diff
@@ -34,3 +35,12 @@ dependencies {
...
     detektPlugins(project(":detekt-formatting"))
 }
+
+/*
+Can be used from next detekt-gradle-plugin release:
+tasks.withType<Detekt>().matching { it.name == "detektTest" }.configureEach {
+    freeCompilerArgs.addAll(kotlin.target.compilations.getByName("test").friendPaths.flatMap { fileCollection ->
+        fileCollection.files.map { "-Xfriend-paths=${it.absolutePath}" }
+    })
+}
```
BraisGabin, line 45: 🤔 isn't this a good default value for our plugin? I mean, instead of force our users to fill it, add a default value passing all what we could get from the kotlin tasks?

BraisGabin:
> I'm unsure about the `freeCompilerArgs` naming. I think it might be confusing for users, especially if compared to the Kotlin's `freeCompilerArgs` property. I'm not sure users will understand that the two `freeCompilerArgs` are effectively adding flags to distinct compiler invocations (mostly because not all the users are aware of the Detekt internals).

We want that the user set here the same values as it set in the compiler, right? If so, I think that it's safe to use the same name, but I'm not sure that I'm understanding all the implications.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
3flex: Yep - I'll ensure test coverage once the design is finalised.

build-logic/src/main/kotlin/detekt.gradle.kts
3flex: Agree, but I plan to do that in a separate PR. I want to get the other pieces in place first.

3flex:
> We want that the user set here the same values as it set in the compiler, right?

Yes, that's the intention. So I'd like to keep `freeCompilerArgs` naming. It matches the Kotlin Gradle plugin's DSL, like the existing `languageVersion` and `jvmTarget` parameters.

cortinico:
> We want that the user set here the same values as it set in the compiler, right?

Theoretically yes. Practically I'm not sure this will be the case. And that's also the reason why I believe there could be some confusion here.

Say a user has Explicit API mode and an Opt-In specified with:

```
tasks.withType<KotlinCompile> {
	kotlinOptions {
		freeCompilerArgs += "-Xopt-in=kotlin.RequiresOptIn"
	}
}
kotlin {
    explicitApi()
}
```

While we would request users to specify this in our Detekt configuration:

```
detekt {
    freeCompilerArgs.addAll(listOf("-Xopt-in=kotlin.RequiresOptIn","-Xexplicit-api=strict"))
}
```

Which I believe could be confusing as you have two `freeCompilerArgs` in your Gradle files that ends up having slightly different semantic.

BraisGabin:
Oh! I see your point... Would it be possible to have `freeCompilerArgs` as a "power feature" and, by default get the same parameters that the kotlin plugin? This way, by default, the user should not bother about the flags at all because we will get them directly from the kotlin plugin. But, if the user wants to do something "special" it can set `freeCompilerArgs` manually.

3flex:
I think that example shows more about the lack of a nice way to set the explicit API mode rather than a problem with the naming of `freeCompilerArgs` - the detekt Gradle plugin can be improved to just pick up the value set using `explicitAPI()` and add the value to the compiler args easily enough, so it becomes:
```kotlin
kotlin {
    explicitApi()
}

tasks.withType<KotlinCompile> {
	kotlinOptions {
		freeCompilerArgs += "-Xopt-in=kotlin.RequiresOptIn"
	}
}

detekt {
    freeCompilerArgs.add("-Xopt-in=kotlin.RequiresOptIn")
}
```

And the detekt plugin gets the `explicitApi` value from the Kotlin Gradle extension.

> have freeCompilerArgs as a "power feature" and, by default get the same parameters that the kotlin plugin?

That's eventually what I want to get to - and not just for this parameter, but others that are set on the Kotlin compilation, to get detekt closer to zero config for type resolution.

cortinico:
> And the detekt plugin gets the `explicitApi` value from the Kotlin Gradle extension.

From you example, why would we treat explicit API mode and the Requires OptIn differently? IMHO either we "copy" everything from KGP (both the _implicit_ args and the `freeCompilerArgs`) or not. And again we can offer a new property to the users where they can customize it (and here I think having it called `freeCompilerArgs` makes it even more confusing).

3flex:
I agree with your comments - it was a bad example, since you're right, and what I want to get to, is all of the detekt parameters required to setup the detekt compiler environment for analysis auto-populated from whatever was set in KGP.

So let's go back a step and design the end state, then we can work backwards.

This is what I'm picturing:
Add missing KGP equivalent parameters to `Detekt` task (we don't need everything from KGP, only those that affect the analysis in some way):
* apiVersion
* explicitApi
* freeCompilerArgs

Then the following can be wired up automatically for each detekt task:
* apiVersion
* languageVersion
* jvmTarget
* classpath
* explicitApi
* freeCompilerArgs

Ideally none of these need to be manually set by the user in their build files, and everything "just works". Aligning the names of the parameters then (hopefully!) makes more sense, since if someone wants to tweak something, it's hopefully clear (and can be made explicit in the docs) that they have the same effect as the equivalent in KGP.

cortinico:
> Ideally none of these need to be manually set by the user in their build files, and everything "just works". Aligning the names of the parameters then (hopefully!) makes more sense, since if someone wants to tweak something, it's hopefully clear (and can be made explicit in the docs) that they have the same effect as the equivalent in KGP.

Totally agree with this 👍 As long as we make sure those are kept in sync between KGP and our detekt-gradle-plugin we should be fine

BraisGabin:
Is there something missing in this PR? Could we merge this once 1.19 is released?

3flex:
I'm going to rethink the approach to this - I'd like to rework the tests a bit to support a refactor of the plugin which will open up some new opportunities to auto config the detekt tasks based on the configured Kotlin compilations.

It'll take a little time to work through though.

3flex:
Closing, as the approach can first be discussed in #4465.

PR #4112, opened 2021-09-19
Ensure detekt-tooling public API is stable

3flex:
It seems `detekt-tooling` should be considered public API. This ensures the API is not accidentally changed.

Modules now checked with the binary compatibility validator:
* detekt-tooling
* detekt-api
* detekt-psi-utils


BraisGabin (approved):

BraisGabin:
A bit off topic but, should we add detekt-test and detekt-test-utils? For example I use both of those in some custom rules: https://github.com/BraisGabin/detekt-junit-rules/blob/main/build.gradle.kts

I'm not sure because they are used to run the tests. With api compatibility should be enough, right?

3flex:
> should we add detekt-test and detekt-test-utils

I think it's worth having the discussion about which APIs need to be strictly maintained, and which don't. That could be a separate issue/discussion.

cortinico (approved):

PR #4113, opened 2021-09-19
Kotlin 1.5.31

3flex:
Updated Kotlin artifacts have been released, but this should be held back until the official release announcement.

### DIFF
--- a/detekt-report-sarif/src/test/kotlin/io/github/detekt/report/sarif/SarifOutputReportSpec.kt
+++ b/detekt-report-sarif/src/test/kotlin/io/github/detekt/report/sarif/SarifOutputReportSpec.kt
@@ -5,6 +5,7 @@ import io.github.detekt.tooling.api.VersionProvider
 import io.gitlab.arturbosch.detekt.api.CodeSmell
 import io.gitlab.arturbosch.detekt.api.Finding
 import io.gitlab.arturbosch.detekt.api.SeverityLevel
+import io.gitlab.arturbosch.detekt.api.UnstableApi
 import io.gitlab.arturbosch.detekt.api.internal.whichOS
 import io.gitlab.arturbosch.detekt.test.EmptySetupContext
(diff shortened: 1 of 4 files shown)


detekt-report-sarif/src/test/kotlin/io/github/detekt/report/sarif/SarifOutputReportSpec.kt
```diff
@@ -16,6 +17,7 @@ import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
 import java.nio.file.Paths
 
+@OptIn(UnstableApi::class)
```
BraisGabin, line 20: Something that we were using now is it unestable? What is it?

BraisGabin (approved):

BraisGabin:
Why is this blocked?

detekt-report-sarif/src/test/kotlin/io/github/detekt/report/sarif/SarifOutputReportSpec.kt
3flex: See link in second commit message: https://kotlinlang.org/docs/whatsnew1530.html#requiring-opt-in-on-implicit-usages-of-experimental-apis

> In Kotlin 1.5.30, the compiler treats any declaration that has an experimental type in the signature as experimental. Namely, it requires opt-in even for implicit usages of an experimental API. 

So the OptIn annotation is required in a few new places because there is usage of something annotated with `@UnstableApi`, in this case `SarifReportOutput.init` which has `SetupContext` in its signature, which itself is annotated with `UnstableApi`.

3flex:
> Why is this blocked?

Waiting until the official release. The IntelliJ plugin 1.5.31 for example is not yet out so it seems premature to merge until the full Kotlin release process has completed.

3flex:
https://github.com/JetBrains/kotlin/releases/tag/v1.5.31

PR #4118, opened 2021-09-20
Disable UnsafeCallOnNullableType on tests

BraisGabin:
Disable `UnsafeCallOnNullableType` on tests. The use of `!!` is a type of assertion. It should not be a problem to use it in your tests.

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -138,6 +138,9 @@ potential-bugs:
     active: true
   UnreachableCatchBlock:
     active: true
+  UnsafeCallOnNullableType:
+    active: true
+    excludes: ['**/test/**', '**/*.Test.kt', '**/*.Spec.kt']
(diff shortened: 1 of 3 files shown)


config/detekt/detekt.yml
```diff
@@ -138,6 +138,9 @@ potential-bugs:
     active: true
   UnreachableCatchBlock:
     active: true
+  UnsafeCallOnNullableType:
```
cortinico, line 141: You wanted to apply this for all the users of Detekt? Or just for our project?

BraisGabin: In this PR I wanted to apply it just for our project. But I opened #4119 to talk if we should change it for all our users.

cortinico (approved):

BraisGabin:
I'm blocking this PR in favor of #4123. If that one is merged I'll close this one.

Issue #3501, opened 2021-02-25
Add reasons in ForbiddenMethodCall & ForbiddenImport


BraisGabin:
## Expected Behavior
When detekt reports that a call or an import is forbidden it tell my why it is forbidden.

## Current Behavior
Right now detekt just say that it is forbidden so we a new person in the project doesn't know what to do.

## Context

Some coworkers ask me some times "Why is this forbidden?" And I need to explain them. That never bother me but after reading this in: https://saket.me/great-teams-merge-fast/ I realised that this explanation should be done by detekt.

BraisGabin:
I imagine that we will need to change the configuration. Probably we should add a new one and maybe deprecate the old one.

Right now:
```yml
  ForbiddenImport:
    active: true
    imports:
      - 'org.mockito.*'
      - 'kotlinx.android.synthetic.*'
```

Proposal:
```yml
  ForbiddenImport:
    active: true
    imports:
      - import: 'org.mockito.*'
        reason: 'Use `com.nhaarman.mockitokotlin2` instead'
      - import: 'kotlinx.android.synthetic.*'
        reason: 'Use ViewBinding instead'
```

I don't like to much the proposal, any idea is more than welcome.

cortinico:
> I imagine that we will need to change the configuration. Probably we should add a new one and maybe deprecate the old one.

Do we have other rules where this pattern could become useful? I suppose we have some scenarios where you want to either:
- Provide a more meaningful message to a developer
- Instantiate the same rule twice but with distinct purpose (as you presented in the example).

BraisGabin:
Probably. For example the naming rules. You can have different regex for different use cases: "standard", "test", "dsl", etc.

But about the forbidden rules... Does it have sense to create an instance for each reason? If you forbid something you have to have a reason so I think that all of them could be in the same rule.

cortinico:
> Does it have sense to create an instance for each reason?

I was more thinking at scenarios where you might want to re-instantiate the same rule but with different configs (say for different areas of your codebase 🤔 )

BraisGabin:
Yes, yes, and that's what I mean with:

> Probably. For example the naming rules. You can have different regex for different use cases: "standard", "test", "dsl", etc.

You can have different configuration for your dsl files and other for test files.

cortinico:
[...]

Issue #4125, opened 2021-09-23
add customMessage to ForbiddenComment Rule


hmSchuller:
Add configuration option to show a custom error message to the rule ForbiddenComment.

## Context
`ForbiddenComment` has the configuration option `allowedPatterns`.
When using this option for finding comments that have no Ticket ID assigned to them, I get presented with the default error message that is:  
```
"This comment contains '%s' that has been defined as forbidden in detekt."
```

I'd like to be able to change this into a custom message to give new members on the project a meaningful message that will guide them on how to resolve this issue.

Example:
```
This comment is missing a ticket ID!
```

BraisGabin:
Related with #3501

strkkk:
@BraisGabin it is fixed by https://github.com/detekt/detekt/pull/41260 isnt it?

BraisGabin:
You are right! My bad :). Closing.

PR #4126, opened 2021-09-24
Add customMessage to ForbiddenComment Rule

hmSchuller:
tackles #4125 

Changes:
- add configuration `customMessage` in `ForbiddenComment`
- add tests to cover new functionality
- generate documentation in `default-detekt-config.yml` (not sure if this is done correctly)
- update contributor list 🥳


BraisGabin:
I agree 100% with the need of this but I think that this solution isn't generic enough. Each value could need a different message to explain why it is forbidden. And a regex could need multiple ones...

The problem here is how the configuration should look like to support this. 

Maybe we could merge this meanwhile we find that generic solution. We can always deprecate it in the future.

hmSchuller:
Whops, i was oblivious that something like #3501 already exists. 
Naming is hard, they say, and in my mind the word "reason" never popped up. I only searched for message 😅 

I was also thinking that a generic solution would be nice, albeit hard to come up with.

hmSchuller:
@BraisGabin 
Anything I should do here? Should I attempt to go for the more generic approach that you described?

BraisGabin (approved):

BraisGabin:
No really, the follow up work is already tracked in #3501 so I think that we could merge this meanwhile we don't have a more generic solution. But I would like to know what the other maintainers think. Is it worht to merge this meanwhile or should we fix #3501 directly?

Issue #4120, opened 2021-09-21
Support Toolchains for JVM projects


3flex:
## Expected Behavior
If I am using Gradle's toolchain support for my Kotlin project, detekt should, by default, ~~use the version of the JDK specified by the toolchain during execution.~~ set the JDK target and JDK path as defined by the toolchain. The values can be overridden as needed.

I.e. if this configuration is in place:
```kotlin
java {
    toolchain {
        languageVersion.set(JavaLanguageVersion.of(11))
    }
}
```
Then detekt should setup its Kotlin compilation environment and perform its analysis against JDK 11, regardless of which JDK version was used to run Gradle itself.

## Current Behavior
~~detekt runs using the same version of Java that is used to run Gradle.~~
detekt sets its own default for the JDK target, which can be manually overridden on individual tasks. It currently does not accept a custom JDK path as an input parameter.

## Context
As an example, if Gradle is run with JDK 11, but toolchain support is enabled and configured with JDK version 16, there will be unresolved reference errors if the codebase uses any new APIs from Java 12-16. It would also prevent rules that are configured with references to Java 12+ APIs, like `ForbiddenMethodCall`, from working correctly.

Adding support mirrors the [toolchain support added to Kotlin](https://kotlinlang.org/docs/whatsnew1530.html#support-for-java-toolchains) in 1.5.30.

~~I'm very happy to work on this one, but it will mean removing the classloader cache, or at least bypassing it, as detekt will have to execute outside of Gradle's process using Gradle's Worker API with [process isolation](https://docs.gradle.org/current/userguide/worker_api.html#creating_a_worker_daemon) or using [JavaExec](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.JavaExec.html), so I am asking for feedback first.~~

~~My preference is removing the classloader cache for simplicity, and I also believe this will reduce Gradle daemon memory pressure.~~

~~**Edit:** Toolchains will help with #3396 as it would allow running the build on Java 8 while running detekt with Java 11.~~

BraisGabin:
I just read about the Gradle¡s Worker API and I never use it directly. But I understood that we could use it and cache the class loader in it. I remember that it was like we could create our own daemon. Am I right? If we can do that, go ahead.

But I would vote against just removing the cache and going back to the old performance. It was *really* bad.

3flex:
I'll start with a PR to switch to worker API, then if that goes well I'll follow up with a PR introducing toolchain support.

[...]

Issue #2957, opened 2020-08-09
Changing any core class invalidates cached detekt classloader


arturbosch:
## Expected Behavior
Classloader gets invalided and a new gets created.

## Observed Behavior
Confusing error message:
```
> Task :detekt-api:detekt FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':detekt-api:detekt'.
> Provided configuration file is invalid: Structure must be from type Map<String,Any>!
```

## Steps to Reproduce
Add a line like `settings.debug { "" }` in Analyzer class and rerun `gradle detekt`.

## Context
Plugin and core developers.

cortinico:
Is this still relevant? Anyone experienced it recently? If not we can close this

BraisGabin:
I think that this is still valid.

BraisGabin:
I think that #4128 will fix it. But I'm not sure.

cortinico:
> This is a long standing issue that we should fix. Maybe we could ask to our gradle collegues.

Have you seen this error happening again though?

BraisGabin:
I can confirm. This is still an issue. Way to reproduce:

1. Execute `rm config/detekt/detekt.yml && ./gradlew generateConfig`
2. In the class `DefaultConfigProvider` launch an unconditional exception on the function `copy`. Something like `error("hello")`
3. Execute `rm config/detekt/detekt.yml && ./gradlew generateConfig` again. You will expect to see all the tasks fail with the message `hello` but they don't and the file `config/detekt/detekt.yml` is generated.
4. Execute `rm config/detekt/detekt.yml && ./gradlew --stop && ./gradlew generateConfig` now the `hello` message will appear and the file `config/detekt/detekt.yml` will not be generated.

For sure there must be easier ways to reproduce it but this is the one that I just found.

atulgpt:
@BraisGabin I tried with the latest main and was not able to reproduce with your steps at https://github.com/detekt/detekt/issues/2957#issuecomment-1132145945. Could this be fixed?

BraisGabin:
I'm still able to reproduce it. I updated the steps a bit because they were old but the issue is still there.

BraisGabin:
I was working on detekt yesterday all day without needing any `--stop`. The stop is only needed if you run the `detekt` task. In general is not a good idea to run detekt that often, it is slow and it's OK to have code smells while you are coding. They should be gone only when you submit that code.

[...]

PR #4128, opened 2021-09-25
Run detekt's Gradle tasks with Gradle's Worker API

3flex:
Run `Detekt` task, `DetektGenerateConfigTask` and `DetektCreateBaselineTask` with Gradle's [Worker API](https://docs.gradle.org/7.2/userguide/worker_api.html). This uses [process isolation mode](https://docs.gradle.org/7.2/userguide/worker_api.html#creating_a_worker_daemon) which will allow using the toolchains feature in future (#4120). Performance in general seems acceptable though I haven't done proper benchmarks.

Two issues:
1. When used on this project with an included build I see very poor performance on the first run. I don't see this when including the build in another project. Looking forward to others trying this on their projects to see the performance impact.
2. On subsequent runs when used on this project with an included build I often see a CNFE which I think is due to a Gradle bug (I don't believe Gradle is restarting the Worker daemons when classpath file contents change). It's intermittent and can be resolved either by running the task a few times and having it complete successfully, or restarting the Gradle daemon.

Neither of these issues will impact users, but contributors will face this. I see issue 2 as being more annoying. It's similar to https://github.com/detekt/detekt/issues/2957 but can happen when changes are made in any subproject, not just core modules.

**Edit:** I've raised issue 2 at https://github.com/gradle/gradle/issues/13678#issuecomment-927495654. I'll open a new issue in a few days if there's no reply.

### DIFF
--- a/detekt-gradle-plugin/build.gradle.kts
+++ b/detekt-gradle-plugin/build.gradle.kts
@@ -69,6 +69,7 @@ dependencies {
     compileOnly(libs.kotlin.gradle)
     compileOnly(libs.kotlin.gradlePluginApi)
     implementation(libs.sarif4k)
+    compileOnly("io.gitlab.arturbosch.detekt:detekt-cli:1.21.0")
 
     testKitRuntimeOnly(libs.kotlin.gradle)
     testKitJava11RuntimeOnly(libs.android.gradle.maxSupported)
--- a/detekt-gradle-plugin/build.gradle.kts
+++ b/detekt-gradle-plugin/build.gradle.kts
@@ -69,7 +69,7 @@ dependencies {
     compileOnly(libs.kotlin.gradle)
     compileOnly(libs.kotlin.gradlePluginApi)
     implementation(libs.sarif4k)
-    compileOnly("io.gitlab.arturbosch.detekt:detekt-cli:1.21.0")
+    compileOnly("io.gitlab.arturbosch.detekt:detekt-cli:1.22.0")
 
     testKitRuntimeOnly(libs.kotlin.gradle)
     testKitJava11RuntimeOnly(libs.android.gradle.maxSupported)
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
@@ -71,7 +71,7 @@ abstract class DetektGenerateConfigTask @Inject constructor(
 
         Files.createDirectories(configFile.get().asFile.parentFile.toPath())
 
-        DetektInvoker.create(task = this).invokeCli(
+        DetektInvoker.create().invokeCli(
             arguments = arguments,
             classpath = detektClasspath.plus(pluginClasspath),
             taskName = name,
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
@@ -5,11 +5,13 @@ import io.gitlab.arturbosch.detekt.DetektPlugin.Companion.CONFIG_FILE
 import io.gitlab.arturbosch.detekt.invoke.CliArgument
 import io.gitlab.arturbosch.detekt.invoke.ConfigArgument
 import io.gitlab.arturbosch.detekt.invoke.DetektInvoker
+import io.gitlab.arturbosch.detekt.invoke.DetektWorkAction
 import io.gitlab.arturbosch.detekt.invoke.GenerateConfigArgument
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.ProviderFactory
 import org.gradle.api.services.BuildService
 import org.gradle.api.services.BuildServiceParameters
 import org.gradle.api.tasks.CacheableTask
@@ -18,13 +20,16 @@ import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
 import org.gradle.language.base.plugins.LifecycleBasePlugin
+import org.gradle.workers.WorkerExecutor
 import java.io.File
 import java.nio.file.Files
 import javax.inject.Inject
 
 @CacheableTask
 abstract class DetektGenerateConfigTask @Inject constructor(
-    objects: ObjectFactory
+    objects: ObjectFactory,
+    private val workerExecutor: WorkerExecutor,
+    private val providers: ProviderFactory,
 ) : DefaultTask() {
 
     init {
@@ -71,11 +76,24 @@ abstract class DetektGenerateConfigTask @Inject constructor(
 
         Files.createDirectories(configFile.get().asFile.parentFile.toPath())
 
-        DetektInvoker.create().invokeCli(
-            arguments = arguments,
-            classpath = detektClasspath.plus(pluginClasspath),
-            taskName = name,
-        )
+        if (providers.gradleProperty(USE_WORKER_API).getOrElse("false") == "true") {
+            logger.info("Executing $name using Worker API")
+            val workQueue = workerExecutor.processIsolation { workerSpec ->
+                workerSpec.classpath.from(detektClasspath)
+            }
+
+            workQueue.submit(DetektWorkAction::class.java) { workParameters ->
+                workParameters.arguments.set(arguments)
+                workParameters.taskName.set(name)
+            }
+        } else {
+            logger.info("Executing $name using DetektInvoker")
+            DetektInvoker.create().invokeCli(
+                arguments = arguments,
+                classpath = detektClasspath.plus(pluginClasspath),
+                taskName = name,
+            )
+        }
     }
 
     @Suppress("UnnecessaryAbstractClass")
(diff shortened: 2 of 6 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
```diff
@@ -70,19 +113,18 @@ internal class DefaultCliInvoker(
         msg != null && "Build failed with" in msg && "issues" in msg
 }
 
-private class DryRunInvoker(private val logger: Logger) : DetektInvoker {
+private class DryRunInvoker : DetektInvoker {
```
3flex, line 116: `Logger` can't be passed to the Worker as everything in `DetektWorkParameters` must be serializable. Using `println` instead means the tests still pass as anything passed to stdout is logged at QUIET level by Gradle. There's no impact on users as the dry run invoker should only be used internally.

3flex:
Java 8 on Windows kept failing as the build runner didn't have enough memory.

This is somewhat expected - Worker API starts more daemons than before, but moving the detekt classloader out of the Gradle daemon reduces the memory requirements for the Gradle daemon itself.

I've therefore removed the JVM args config that increased Gradle and TestKit memory to compensate which should keep the build within the runner's memory constraints (7GB total).

chao2zhang:
With the current implementation in a multi-module project, we will have worker processes up to `org.gradle.workers.max`.

```
// Before
➜  kotlin-android-template git:(main) ✗ jps
86912 
89369 Jps
86365 
// After
➜  kotlin-android-template git:(main) ✗ jps
86912 
89472 KotlinCompileDaemon
89428 KotlinCompileDaemon
89415 GradleDaemon
89497 GradleWorkerMain
89499 GradleWorkerMain
89498 GradleWorkerMain
89533 Jps
86365 
```

It appears to me that these additional number of gradle workers processes will impact developers using detekt.

3flex:
Is the concern about memory usage?

From https://docs.gradle.org/current/userguide/worker_api.html#creating_a_worker_daemon:

> These worker daemon processes will persist across builds and can be reused during subsequent builds. If system resources get low, however, Gradle will stop any unused worker daemons.

So we rely on Gradle managing its resources appropriately.

cortinico:
> It would be an opt in first, then can upgrade to opt out later, then potentially have it as the only execution mode in future.

Agree on this approach 👍

eygraber:
I've had...not the most fun experience with the worker API for things like this (as a consumer). Specifically Android Lint started using it, and each worker was allowed to grow to 6gb (which was the `Xmx` I specified for the daemon). So Gradle might clean them when memory is needed, but until then all of my RAM was allocated (which also slowed down the machine so that in a lot of cases Gradle couldn't run fast enough to actually kill the processes to free the memory).

I think it's a good idea to do this in theory, but in the real world there are a lot of implications.

3flex:
Just wondering if you tried tweaking org.gradle.workers.max at all?

There are also options for tweaking max memory only for the detekt tasks run using worker API which should help mitigate.

A permanent option for opt in/out might be better than removing the existing setup wholesale.

eygraber:
I did, but that affected parallelism overall. 

Lint had options to limit the worker memory, which worked, but was trial and error to get right on a per project basis. They eventually switched to run in process by default.

3flex:
Thoughts on merging, as an experimental feature? Perhaps for 1.22.

It's behind a flag, so it won't impact anyone's build.

BraisGabin:
Which is our current min gradle version? But, if this is behind a flag AND it only requires the bump if you active it I vote to merge it inside `1.21.0` to get feedback from the users.

If that's not the case I vote to move it to 1.22.0.

cortinico:
> Which is our current min gradle version?

It's 6.1. I'm fine with bumping the min Gradle version to anything on 6.x

3flex:
Let's wait for 1.22.0 when the min Gradle version is bumped in #4964

BraisGabin:
We can merge this now, right?

3flex:
I'd still wait for #4964 but otherwise yes, this should be ready to go.

schalkms:
#4964 is done. Is this PR still planned to be merged?

> I'd still wait for https://github.com/detekt/detekt/pull/4964 but otherwise yes, this should be ready to go.

3flex:
I'll fix the merge conflict soon, then anyone is welcome to have a last look & merge :)

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
```diff
@@ -69,10 +74,23 @@ abstract class DetektGenerateConfigTask @Inject constructor(
...
 
-        DetektInvoker.create(task = this).invokeCli(
-            arguments = arguments.get(),
-            classpath = detektClasspath.plus(pluginClasspath),
-            taskName = name,
-        )
+        if (providers.gradleProperty(USE_WORKER_API).getOrElse("false") == "true") {
+            logger.info("Executing $name using Worker API")
+            val workQueue = workerExecutor.processIsolation { workerSpec ->
+                workerSpec.classpath.from(detektClasspath)
```
cortinico, line 80: I believe you need `pluginClasspath` here as well now that we generate config for those also

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
```diff
@@ -36,6 +39,47 @@ internal fun Project.isDryRunEnabled(): Boolean {
...
+                parameters.arguments.get(),
+                parameters.classpath,
+                parameters.taskName.get(),
+                parameters.ignoreFailures.getOrElse(false)
+            )
+            return
+        }
+
+        try {
+            val runner = io.gitlab.arturbosch.detekt.cli.buildRunner(
```
cortinico, line 64: nit: why the fully qualified import here?

cortinico (approved):
Did another pass 👍 The `pluginClasspath` comment needs to be address, but apart from that it's good to go for me

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
3flex: Oh good pickup, thank you! Fixed.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
3flex: I can't recall the original reason, but I've now suppressed the deprecation on `buildRunner`. If this is an import instead, then the suppression would have to be on the file level, as you can't suppress deprecations in the import list. See most recent commit.

detekt-gradle-plugin/build.gradle.kts
```diff
@@ -66,6 +66,7 @@ pluginCompileOnly.attributes {
 dependencies {
     compileOnly(libs.kotlin.gradlePluginApi)
     implementation(libs.sarif4k)
+    compileOnly("io.gitlab.arturbosch.detekt:detekt-cli:1.21.0")
```
cortinico, line 69: Can we leverage the included build here instead of using version 1.21.0?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt
arturbosch: Please take a look at https://github.com/detekt/detekt/pull/2896/files#diff-63765919cc19f1899b99d724bfe9c613285bf9967a35df542daf4974cac7d991R13
on how to run DetektCli without relying on internal cli module.

arturbosch:
Hi, I was triggered by Gradle closing two linked issues regarding the worker api which I linked in my worker api experiment from 2020 https://github.com/detekt/detekt/pull/2896.
Looking over the changes in this PR, I do not quite understand how the classloader caching problem is solved?
Wouldn't the code create a new classloader on each detekt invocation?

3flex:
> I do not quite understand how the classloader caching problem is solved?

It's not.

> Wouldn't the code create a new classloader on each detekt invocation

It should only do so if there's no available Gradle worker, and number of current workers < `org.gradle.workers.max`.

Workers API performance should fall somewhere between the current (fastest) setup, and the JavaExec (slowest) version. Workers can be reused, so yes, each worker suffers the initial performance hit as it's spun up, but that only happens for new instances and not ones that are reused.

The big advantage of workers is increased parallelism in the build, as tasks using workers can run in parallel within a Gradle project (so e.g. detektMain and detektTest can run in parallel within the same project). Tasks that don't use workers can only run serially within a project. This can speed up the build if the machine has enough resources available.

The other important note is that detekt's use of Worker API is opt in and behind a flag so it can be treated as an experiment for now so this will not change the existing behaviour or use of the existing classloader cache.

detekt-gradle-plugin/build.gradle.kts
3flex: Unfortunately the included build doesn't seem to be working correctly, so we still have to nominate a fixed version here.

BraisGabin:
I moved this PR to `1.23.0`. It's a big change to introduce it after two RC

3flex:
LGTM?

detekt-gradle-plugin/build.gradle.kts
```diff
@@ -69,6 +69,7 @@ dependencies {
     compileOnly(libs.kotlin.gradle)
     compileOnly(libs.kotlin.gradlePluginApi)
     implementation(libs.sarif4k)
+    compileOnly("io.gitlab.arturbosch.detekt:detekt-cli:1.21.0")
```
BraisGabin, line 72:
```suggestion
    compileOnly("io.gitlab.arturbosch.detekt:detekt-cli:1.22.0")
```

BraisGabin (approved):
LGTM

chao2zhang (approved):
LGTM. The only concern is https://github.com/detekt/detekt/pull/4128#discussion_r973728890 but given that this PR is already open for over a year and it is behind a feature flag, this should be merged sooner rather than later.

PR #4133, opened 2021-09-28
Upgrade AGP to 7

chao2zhang:
Used https://github.com/cortinico/kotlin-android-template for local testing. It appears to work fine.

*EDIT*: Now that AGP 7 drops support for Java 8, at some point I think we also need to move towards Java 11.


gradle/libs.versions.toml
```diff
@@ -22,7 +22,7 @@ kotlin-stdlibJdk8 = { module = "org.jetbrains.kotlin:kotlin-stdlib-jdk8", versio
 kotlinx-html = "org.jetbrains.kotlinx:kotlinx-html-jvm:0.7.3"
 kotlinx-coroutines = "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.0"
 
-android-gradle = "com.android.tools.build:gradle:4.2.0"
+android-gradle = "com.android.tools.build:gradle:7.0.2"
```
BraisGabin, line 25: Shouldn't we change this to `com.android.tools.build:gradle-api:7.0.2` I think that they created that in AGP7 to provide more stability for other gradle plugins.

chao2zhang: gradle-api does not fully give what we need, such as the issue reported here https://issuetracker.google.com/issues/158777988

If we change to gradle-api now, detekt won't compile and there is no alternative to handle type resolution for android projects.

chao2zhang:
This PR is pending on the outcome of https://github.com/detekt/detekt/discussions/4130

3flex:
To unblock this can we not just restrict the tests to running on JDK 11? We already limit which JDKs can be used to run the Android tests by blocking JDK 16+ due to GSON incompatibility (used by AGP - I haven't tested if that imcompatibility is fixed in AGP 7):

```kotlin
internal fun skipIfAndroidEnvironmentRequirementsUnmet() = when {
    !isAndroidSdkInstalled() -> Skip.Yes("No android SDK.")
    getJdkVersion() >= 16 -> Skip.Yes("Android 4.1.3 & 4.2.1 don't run on JDK 16 or higher")
    else -> Skip.No
}
```

chao2zhang:
I agree with you, contingent on the assumption that users are building at least JDK 11 for Android Projects. Maybe we should announce that we can guarantee compatibility with the latest AGP release?

cortinico:
> Maybe we should announce that we can guarantee compatibility with the latest AGP release?

It's essentially already described on this page: 
https://detekt.github.io/detekt/compatibility.html

Once we bump AGP to 7.x, we're implicitly dropping support for AGP 4.x
That means that is safe to assume that, if you're using Detekt 1.20.x (or whichever version this PR will land into), you're on AGP 7.x and JDK 11.
If you're not, you're on your own essentially ¯\_(ツ)_/¯

BraisGabin:
6 months since agp 7.0 was launched. I'm ok dropping support for agp 4.

3flex (approved):

chao2zhang (commented):
👍

BraisGabin (approved):

PR #4139, opened 2021-09-30
Fixes for Kotlin 1.6.0-M1

3flex:
Fixing a couple of issues identified with a test build using Kotlin 1.6.0-M1

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Main.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Main.kt
@@ -32,8 +32,8 @@ fun main(args: Array<String>) {
                 else -> cause.printStackTrace()
             }
         }
+        else -> exitProcess(result.exitCode())
     }
-    exitProcess(result.exitCode())
 }
 
 @NotApiButProbablyUsedByUsers
--- a/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/Junk.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/Junk.kt
@@ -37,7 +37,7 @@ fun getIntValueForPsiElement(element: PsiElement): Int? {
 
 fun KtClass.companionObject() = this.companionObjects.singleOrNull { it.isCompanion() }
 
-inline fun <reified T : Any> Any.safeAs(): T? = this as? T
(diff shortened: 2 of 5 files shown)


cortinico (approved):

chao2zhang (approved):

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Main.kt
```diff
@@ -32,8 +32,8 @@ fun main(args: Array<String>) {
                 else -> cause.printStackTrace()
             }
         }
+        else -> exitProcess(result.exitCode())
     }
-    exitProcess(result.exitCode())
```
BraisGabin, line 36: Why? This doesn't seem right. If we are in the case of `HandledArgumentViolation` we want to exit the process with the correct exitCode. Right?

I imagine that this is because of the exhaustive enums. But I think that instead of `else` that doesn't give any information about which other cases go there we could use: `null -> Unit` or `null -> { /* no-op */ }`. I prefer the first one because with the second I always have a bad time:  Does this case do nothing or does it return a lambda that does nothing?

detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/Junk.kt
```diff
@@ -37,7 +37,7 @@ fun getIntValueForPsiElement(element: PsiElement): Int? {
 
 fun KtClass.companionObject() = this.companionObjects.singleOrNull { it.isCompanion() }
 
-inline fun <reified T : Any> Any.safeAs(): T? = this as? T
+inline fun <reified T : Any> Any?.safeAs(): T? = this as? T
```
BraisGabin, line 40: I imagine that you are adding this `?` to match what `org.jetbrains.kotlin.utils.addToStdlib.safeAs` did, right? If it's that I'm fine. But I think that an extension function that when receives `null` always return `null` shouldn't handle that case. That responsability should be handled by the caller.

BraisGabin:
😂 I didn't realised that this was already merged

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Main.kt
chao2zhang: Good catch. This is indeed a miss from my side about the behavioral change.

3flex:
@BraisGabin thanks for your feedback, I've raised #4143 to address it.

Issue #4131, opened 2021-09-27
[Feature Request] ImplicitLambdaParameter rules


QuinnBast:
Is there a rule that forces lambda parameters to be defined? My team would love to enforce something like this and completely require all lambda variables to be named.

```
var.map { it.value } // BAD!
var.map { namedVar -> namedVar.value } // Force a name!
```

I am also looking for a rule for lambda variable length requirements.

```
var.map { a -> a.value } // BAD!
var.map { moreSpecificName -> moreSpecificName.value } // Good.
```

_Originally posted by @QuinnBast in https://github.com/detekt/detekt/issues/3245#issuecomment-926899323_

BraisGabin:
I think that we could change `MultilineLambdaItParameter` to fix the first point. But I think that it would be a Breaking change because we should rename it to `ImplicitItParameter`. And then have a flag (on by default) to "ignore one line lambdas".

I think that the second point should be handled by [VariableMinLength](https://detekt.github.io/detekt/naming.html#variableminlength). That or we should create one for `ParameterMinLength`.

QuinnBast:
An alternative is to use the `VariableNaming` rule within lambdas. It would be possible to send it the following rule to error on variables named `it`:

```yaml
  VariableNaming:
    active: true
    excludes: "**/test/**,**/androidTest/**,**/*.Test.kt,**/*.Spec.kt,**/*.Spek.kt"
    variablePattern: '\b(?!it\b)([a-z][A-Za-z0-9]*)' # Error on variables named "it"
    privateVariablePattern: '\b(?!it\b)((_)?[a-z][A-Za-z0-9]*)' # Error on variables named "it"
    ignoreOverridden: true
```

This would make any variable named `it` fail the `VariableNaming` check. We just need to apply it to lambdas.

chao2zhang:
`VariableNaming` and `VariableMinLength` checking class properties and variables. If it also checks lambda expression then we are probably adding too many responsibilities in the rule.

I would vote for new rules.

cortinico:
> I would vote for new rules.

+1 specifically for the part related to:

> I am also looking for a rule for lambda variable length requirements.

BraisGabin:
I Implemented the rule `LambdaParameterNaming` at #4147 because we wasn't checking those parameter names.

[...]

PR #4147, opened 2021-10-02
Implement LambdaParameterNaming

BraisGabin:
This implements part of #4131

### DIFF
--- a/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/LambdaParameterNaming.kt
+++ b/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/LambdaParameterNaming.kt
@@ -1,0 +1,47 @@
+package io.gitlab.arturbosch.detekt.rules.naming
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import org.jetbrains.kotlin.psi.KtNamedDeclaration
+import org.jetbrains.kotlin.psi.KtParameter
+
+/**
+ * Reports lambda parameter names that do not follow the specified naming convention.
+ */
+class LambdaParameterNaming(config: Config = Config.empty) : Rule(config) {
+
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Style,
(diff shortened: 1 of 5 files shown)


detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NamingRulesSpec.kt
```diff
@@ -13,7 +13,6 @@ class NamingRulesSpec : Spek({
                 data class D(val i: Int, val j: Int)
                 fun doStuff() {
                     val (_, HOLY_GRAIL) = D(5, 4)
-                    emptyMap<String, String>().forEach { _, V -> println(V) }
```
BraisGabin, line 16: Why was this here? Should we support this? Is there any documentation that talks about this case?

chao2zhang: Is the problem about why we allow the capital `V` by default?

It feels like this is a test case that verifying that we do not have any naming requirement for lambda parameters 🙈 . I think your change makes sense.

chao2zhang (approved):

detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NamingRulesSpec.kt
chao2zhang: Is the problem about why we allow the capital V by default?

It feels like this is a test case that verifying that we do not have any naming requirement for lambda parameters 🙈 . I think your change makes sense.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/LambdaParameterNaming.kt
```diff
@@ -0,0 +1,48 @@
...
+        Severity.Style,
+        "Lambda parameter names should follow the naming convention set in the projects configuration.",
+        debt = Debt.FIVE_MINS
+    )
+
+    @Configuration("naming pattern")
+    private val parameterPattern: Regex by config("[a-z][A-Za-z0-9]*", String::toRegex)
+
+    override fun visitParameter(parameter: KtParameter) {
+        val parameters: List<KtNamedDeclaration> = parameter.destructuringDeclaration?.entries ?: listOf(parameter)
```
marschwar, line 31: As far as I can tell there is nothing restricting this to lambda expressions.

Should this fail or pass?

```kotlin
it("ignores invalid function parameters") {
    val code = """
        fun foo(BAR: String) = Unit
    """
    assertThat(LambdaParameterNaming().compileAndLint(code))
        .isEmpty()
}
```

BraisGabin: You are right, there is nothing restrictin this to lambda. But all the naming rules are kind of strange. They aren't called by the normal rule engine. They are called from `detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NamingRules.kt`. And it only cal this function when the parameter is from a lambda. You can check my changes in that file.

Issue #4080, opened 2021-08-30
Add regexp to MagicNumber rule


apachucy:
Right now we can use MagicNumber rule with defined ignoreNumbers that are intigers. Is there a chance to add regexp to ignoreNumbers?

I am interested in case when you need to pass hex (for example to colors).
This case will trigger MagicNumber rule: Color( 0x66000000)
This case will not trigger MagicNumber rule: Color(color = 0x66000000) but I prefer using the previous example 

Cheers,
AP

cortinico:
> Is there a chance to add regexp to ignoreNumbers?

I understand your use case, though the rule has already a lot of ignores.
Plus your color example looks like the classical example where you want to export the number to a constant:

```
val COLOR_ALPHA_40 = 0x66000000
```

schalkms:
I agree with Nico. I don’t think we should add code to this rule for this special case. This should be an @Ignore or a custom rule by copying the detekt one.

BraisGabin:
To me this use case should be "allow magic numbers as a parameter of `Color.constructor`". That could be archived if we implement #4068.

ccjernigan:
Detekt's magic number check is also painful with the Kotlin Duration API.  For example, the following declaration will get flagged as a magic number: `val timeout = Duration.seconds(10)`.  At least in my particular codebase, we've currently got over 100 `@Suppress` annotations on durations alone.

I realize the Duration API is not yet stable, although it should be stabilized with the next Kotlin release.

The ability to tune the MagicNumber rule, either with a regex or just a hard-coded exclusion for the Duration API, would make this much less painful to work with.

Would the Detekt maintainers prefer that I open a new issue specifically for the Duration API?  Or does this use case add weight to the regex solution proposed by this issue?

marschwar:
To me this is the same issue and does not anything. The idea behind the MagicNumber rule is to improve readability. In your particular example this should be either `TEN_SECONDS` or `DEFAULT_TIMEOUT`. Both carry a (different) meaning.

BraisGabin:
Does `Duration.seconds(TEN_SECONDS)` improve the readability over `Duration.seconds(10)`? For things like this I have this rule disabled. But I insist that #4068 should fix that. We should find which type of ignores we need.

marschwar:
> Does Duration.seconds(TEN_SECONDS) improve the readability over Duration.seconds(10)?

No, that would be horrible.

[...]

PR #4148, opened 2021-10-02
Allow to suppress issues if they are found in a function with a determined name

### LINKED ISSUE #3855: Some exclusion mechanism for NullableToStringCall

It will be great to have some method definition/call side exclusion mechanism for NullableToStringCall.

## Context

NullableToStringCall is very useful for us and with its help, we detect a couple of real bugs.

But >90% of reports are useless because it reports the code where a nullable String is ok and intended:
- all kinds of log/trace messages
- toString() implementations itself

Tipical toString() impl:

``` kotlin
    override fun toString(): String =
        "Notification(" +
            "clientId=$clientId, " +
            "message=$message, " +
            "type=$type, " +
            "style=$style, " +
            "createdDate=$createdDate, " +
            "relevanceDate=$relevanceDate, " +
            "showedDate=$showedDate, " +
            ")"
```

Some of the class fields (obviously) can be nullable. As a result, NullableToStringCall is the most frequently detected issue in our codebase regardless of the project. So we are forced to disable this useful rule in most projects because of the very high maintenance cost of suspending reports case by case.

BraisGabin:
There are places where you don't care about an issue inside a determined function. One example is #3855

Fix #3855

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
@@ -1,27 +1,19 @@
 package io.gitlab.arturbosch.detekt.core.suppressors
 
+import io.github.detekt.tooling.api.FunctionMatcher
 import io.gitlab.arturbosch.detekt.api.ConfigAware
 import org.jetbrains.kotlin.psi.KtElement
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 import org.jetbrains.kotlin.resolve.BindingContext
 
-/*
- * Possible improvement: don't only check name but also check for parameters.
- * ```yaml
- * ignoreFunctions:
- *   - 'toString()' # only functions called toString without parameter
- *   - 'compare(String)' # only functions called compare with one parameter of type String
- *   - 'equals' # every function called function (it doesn't matter it's parameters is ignored)
- * ```
- * This would not be a breaking change.
- */
 internal fun functionSuppressorFactory(rule: ConfigAware, bindingContext: BindingContext): Suppressor? {
-    val names = rule.valueOrDefault("ignoreFunction", emptyList<String>())
-    return if (names.isNotEmpty()) {
+    val functionMatchers = rule.valueOrDefault("ignoreFunction", emptyList<String>())
+        .map(FunctionMatcher::fromFunctionSignature)
+    return if (functionMatchers.isNotEmpty()) {
         Suppressor { finding ->
             val element = finding.entity.ktElement
-            element != null && functionSuppressor(element, bindingContext, names)
+            element != null && functionSuppressor(element, bindingContext, functionMatchers)
         }
     } else {
         null
@@ -31,18 +23,18 @@ internal fun functionSuppressorFactory(rule: ConfigAware, bindingContext: Bindin
 private fun functionSuppressor(
     element: KtElement,
     bindingContext: BindingContext,
-    names: List<String>,
+    functionMatchers: List<FunctionMatcher>,
 ): Boolean {
-    return element.isInFunctionNamed(bindingContext, names)
+    return element.isInFunctionNamed(bindingContext, functionMatchers)
 }
 
 private fun KtElement.isInFunctionNamed(
     bindingContext: BindingContext,
-    names: List<String>,
+    functionMatchers: List<FunctionMatcher>,
 ): Boolean {
-    return if (this is KtNamedFunction && name in names) {
+    return if (this is KtNamedFunction && functionMatchers.any { it.match(this, bindingContext) }) {
         true
     } else {
-        getStrictParentOfType<KtNamedFunction>()?.isInFunctionNamed(bindingContext, names) ?: false
+        getStrictParentOfType<KtNamedFunction>()?.isInFunctionNamed(bindingContext, functionMatchers) ?: false
     }
 }
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
@@ -4,6 +4,7 @@ import io.gitlab.arturbosch.detekt.api.ConfigAware
 import org.jetbrains.kotlin.psi.KtElement
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
+import org.jetbrains.kotlin.resolve.BindingContext
 
 /*
  * Possible improvement: don't only check name but also check for parameters.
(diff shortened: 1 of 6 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/Suppressors.kt
```diff
@@ -13,6 +13,7 @@ internal fun getSuppressors(rule: BaseRule): List<Suppressor> {
     return if (rule is ConfigAware) {
         listOfNotNull(
             annotationSuppressorFactory(rule),
+            functionSuppressorFactory(rule),
```
chao2zhang, line 16: Nice! Now we have a second suppressor factory

chao2zhang:
This would address the other part of #4080.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.jetbrains.kotlin.psi.KtElement
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
+
+/*
+ * Possible improvement: don't only check name but also check for parameters.
+ * ```yaml
+ * ignoreFunctions:
+ *   - 'toString()' # only functions called toString without parameter
+ *   - 'compare(String)' # only functions called compare with one parameter of type String
```
chao2zhang, line 13: We haven't implemented function name + param name match - Is this correct? In my opinion, this should be implemented along with this PR.

Additionally, I think we might need to handle the extra `()` as well.

chao2zhang (commented):
Thanks for the test coverage.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
BraisGabin: I was thinking that this would be a lot of work. But I just remembered that `ForbiddenMethodCall` is doing something very similar so I can use that code here. I'll give it a try.

BraisGabin:
I'm moving this PR back to draft because I need #4176 merged first.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
```diff
@@ -0,0 +1,40 @@
...
+private fun KtElement.isInFunctionNamed(
+    bindingContext: BindingContext,
+    functionMatchers: List<FunctionMatcher>,
+): Boolean {
+    return if (this is KtNamedFunction && functionMatchers.any { it.match(this, bindingContext) }) {
+        true
+    } else {
+        getStrictParentOfType<KtNamedFunction>()?.isInFunctionNamed(bindingContext, functionMatchers) ?: false
+    }
+}
```
chao2zhang, line 40:
```suggestion
private fun KtElement.isInFunctionNamed(
    bindingContext: BindingContext,
    functionMatchers: List<FunctionMatcher>,
): Boolean = 
  getParentOfType<KtNamedFunction>(false)?.let { function ->
    functionMatchers.any { it.match(function, bindingContext) }
  } ?: false
```

chao2zhang: What do you think about writing in this way to avoid recursion? We can even inline this function.

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressorSpec.kt
```diff
@@ -0,0 +1,168 @@
...
+                        fun toString(parameter: String): String {
+                            return ""
+                        }
+                    }
+
+                    fun toString() = Unit
+                    """.trimIndent()
+                )
+            }
+            val binding by memoized { env.getContextForPaths(listOf(root)) }
```
chao2zhang, line 71: Just to confirm, we can use BindingContext.EMPTY and the tests would still pass, because we don't require type resolution. Is that correct?

chao2zhang (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
BraisGabin: You can have a function inside another function. I prefer the recursive version. But I should add a test to demostraste that intention or any one could refactor this code later.

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressorSpec.kt
BraisGabin: Good point! It depends. If you only match using the name you don't need the BindingContext but if you want to match the parameters you need the BindingContext. I'm going to add some tests to demostrate this.

BraisGabin: Ups! I just reread my own comment: https://github.com/detekt/detekt/pull/4148/files/bda4de4f0177e4493dbd7af7019d209df2ae29fe#r725492307 That behaviour is tested on `FunctionSignatureSpec`.

I'm conflicted here. I want to write "functional-ish code". But if I do so the tests end up beening a bit crapy. I can't create a fake/mock of FunctionSignature and inject it.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
chao2zhang: I see your intention and it makes sense. It does not feel like the code in the existing implementation addresses the case where there is function nested inside another function though.

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressorSpec.kt
chao2zhang: My assumption is that, _if a feature does not require type resolution, we should write our tests as not requiring type resolution_.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
BraisGabin: It work. I added the tests and it works with nested functions :)

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressorSpec.kt
BraisGabin: I understand your comment now. You are completely right! I'm using now BindingContext.EMPTY

chao2zhang (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
chao2zhang: After a second thought, you are right. My mind couldn't keep up with code today

PR #4150, opened 2021-10-02
Fix issues with Elvis operator in UnconditionalJumpStatementInLoop

### LINKED ISSUE #4134: False positive: UnconditionalJumpStatementInLoop with Elvis operator
## Expected Behavior

The Elvis operator in a conditional code like this should be accepted:

```kotlin
fun function(items: List<Int?>): String? {
    for (item in items) {
        val intItem = item ?: continue
        return intItem.toString()
    }
    return null
}
```

## Observed Behavior

It is reported as `UnconditionalJumpStatementInLoop`.

## Context

Original code is a little bit more complex where there is data processing between `return` and `continue`. It can be generally avoided using sequences or the `let` function.

## Your Environment

* Version of detekt used: `1.18.1`
* Version of Gradle used (if applicable): `7.2`
* Operating System and version: `macOS 11`

MiSikora:
It was hard to determine which features I could break during development, so I split the tests into more granular ones.

I tried my best to cover the rule. Still, suppose anyone has ideas for adding more tests for false-positive or false-negative checks. In that case, I'll gladly add them as this rule is quite tricky, in my opinion: especially given nesting and such.

Closes #4134.


chao2zhang (approved):
Thank you for fixing two bugs (The originally reported one for `isConditionalJumpStatement()` and the additional one for `isFollowedByElvisJump()`)!

BraisGabin (approved):
👏👏👏 great refactor in the tests. It's way easier now. Great first contribution 👍.

PR #4151, opened 2021-10-03
Set source with SourceDirectorySet instead of file list

### LINKED ISSUE #4127: Running detekt in "type resolution" mode ignores "excludes" specified in gradle
## Expected Behavior
When running with type resolution enabled (using `detektMain` or `detektDebug`, etc), Detekt should respect excludes specified with 

```groovy
subprojects {
  detekt { ... }

  tasks.withType(io.gitlab.arturbosch.detekt.Detekt).configureEach {
    exclude("**/generated/**")
  }
}
```

## Observed Behavior
Currently, this `exclude` option is ignored and I have files like
```
module/src/main/kotlin/data/network/entity/generated/AmountModel.kt
```
being checked by detekt and errors are reported for them.
When running without type resolution (`detekt` task), no errors are reported.

Maybe this is somehow caused by the fact that

```
tasks.withType<com....Detect> { exclude(...) }
```

while type resolution has tasks like `detektMain`, `detektDebug`, etc

## Steps to Reproduce
* Create a project with several modules, apply configuration specified above
* Create a folder `src/main/com/example/data/generated` with some file and some detektable error within
* Run detekt and observe that file isn't ignored

## Your Environment

* Version of Gradle used (if applicable): 7.0.2
* Operating System and version: Linux

3flex:
Fixes #4127

Unfortunately I can't say definitively why this fixes the problem, but I suspect it's somewhat related to this: https://github.com/gradle/gradle/issues/3417#issuecomment-379306780

Passing `kotlinSourceSet.kotlin.files` passed a list of files to use as the source for the task, while `kotlinSourceSet.kotlin` passes a `SourceDirectorySet`, so it's possible the filtering based on regex described in the comment I linked won't apply when a full file path is provided, but works when the source directory set is passed in instead.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektJvm.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektJvm.kt
@@ -25,7 +25,7 @@ internal class DetektJvm(private val project: Project) {
         val kotlinSourceSet = (sourceSet as HasConvention).convention.plugins["kotlin"] as? KotlinSourceSet
             ?: throw GradleException("Kotlin source set not found. Please report on detekt's issue tracker")
         registerDetektTask(DetektPlugin.DETEKT_TASK_NAME + sourceSet.name.capitalize(), extension) {
-            setSource(kotlinSourceSet.kotlin.files)
+            source = kotlinSourceSet.kotlin
             classpath.setFrom(sourceSet.compileClasspath.existingFiles(), sourceSet.output.classesDirs.existingFiles())
             // If a baseline file is configured as input file, it must exist to be configured, otherwise the task fails.
             // We try to find the configured baseline or alternatively a specific variant matching this task.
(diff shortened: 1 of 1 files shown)


chao2zhang:
I guess `SourceDirectorySet` is a reactive/live data structure that it would be able to consider filetree operations like `exclude`.
`kotlin.files`, on the other hand, is a `Set<File>` which is definitely one time data structure.

chao2zhang (approved):

BraisGabin (commented):
Could we unit test this somehow?

chao2zhang:
In my opinion, this fix should be in 1.19.0 - @3flex Would you have time to verify this in a repro project? I can help if you don't have time.

3flex:
I tested manually before submitting the PR.

I'm working on adding some functional tests for the Gradle plugin but I'm finding odd issues with TestKit so it will take a bit more time to add an automated test.

If anyone wants to verify manually that might be enough to get it over the line...

BraisGabin:
I didn't check that the excludes work. But I tested it in my project and it doesn't introduce any regression as far as I can see.

BraisGabin (approved):

PR #4157, opened 2021-10-03
Enable binary compatibility validator for detekt-test and detekt-test-api

chao2zhang:
Per discussion in #4117, detekt-test and detekt-test-api are commonly used for custom extensions like custom rules.

This PR enables binary compatibility validator so that we are more conscious about the public API changes - It does not necessarily mean we need to ensure binary compatibility but does provide necessary insights.


detekt-test/api/detekt-test.api
```diff
@@ -0,0 +1,66 @@
...
+	public static final fun yamlConfigFromContent (Ljava/lang/String;)Lio/gitlab/arturbosch/detekt/api/Config;
+}
+
+public final class io/gitlab/arturbosch/detekt/test/RuleExtensionsKt {
+	public static final fun compileAndLint (Lio/gitlab/arturbosch/detekt/api/internal/BaseRule;Ljava/lang/String;)Ljava/util/List;
+	public static final fun compileAndLintWithContext (Lio/gitlab/arturbosch/detekt/api/internal/BaseRule;Lorg/jetbrains/kotlin/cli/jvm/compiler/KotlinCoreEnvironment;Ljava/lang/String;)Ljava/util/List;
+	public static final fun lint (Lio/gitlab/arturbosch/detekt/api/internal/BaseRule;Ljava/lang/String;)Ljava/util/List;
+	public static final fun lint (Lio/gitlab/arturbosch/detekt/api/internal/BaseRule;Ljava/nio/file/Path;)Ljava/util/List;
+	public static final fun lint (Lio/gitlab/arturbosch/detekt/api/internal/BaseRule;Lorg/jetbrains/kotlin/psi/KtFile;)Ljava/util/List;
+	public static final fun lintWithContext (Lio/gitlab/arturbosch/detekt/api/internal/BaseRule;Lorg/jetbrains/kotlin/cli/jvm/compiler/KotlinCoreEnvironment;Ljava/lang/String;[Ljava/lang/String;)Ljava/util/List;
```
3flex, line 34: All functions in `RuleExtensions` are extensions on `BaseRule` which is in `api.internal` package. So we have a public test API for a class that's supposed to be used internally only (at least, that's what the package names currently suggest).

Perhaps `BaseRule` should be moved out of `api.internal` into `api`?

3flex (approved):

detekt-test/api/detekt-test.api
chao2zhang: Yeah this makes sense to me - Extension methods like `BaseRule.compileAndLintWithContext` defined in `RuleExtensions` are also used in https://github.com/BraisGabin/detekt-junit-rules/blob/main/src/test/kotlin/com/braisgabin/detekt/junit/TestFunctionsShouldReturnUnitTest.kt

BraisGabin:
I like the idea of adding this to help us take better decissions. But, in general, I think that it's ok to break binary compatibility for the tests.

detekt-test/api/detekt-test.api
BraisGabin: `BaseRule` is internal because no one should extend from it or use it directly. If someone provides to the core a `BaseRule` directly that is not a `Rule` nor `MultiRule` undefined behaviours could happen. So I'm against this change. I don't think that we should move BaseRule to `api`.

If we don't want to expose this in our public api we should duplicate all these extensions for `Rule` and `MultiRule`.

Issue #4154, opened 2021-10-03
Cannot suppress SwallowedException on catch variable


TWiStErRob:
## Expected Behavior
Below code passes SwallowedException check.

## Observed Behavior
```
exceptions - 20min debt
    SwallowedException - [f] at src\main\kotlin\test.kt:4:4
```

## Steps to Reproduce
```kt
fun f() {
	try {
            // more stuff
	} catch (@Suppress("SwallowedException") ex: AssertionError) {
            // more stuff
	}
}
```

## Context
Only this pattern seems to work which has a very wide effect range. Any code inside try and inside catch is also suppressed. Can't target only this specific catch.

## Your Environment

* Version of detekt used: `io.gitlab.arturbosch.detekt:detekt-gradle-plugin:1.18.1`
* Version of Gradle used (if applicable): 6.9
* Gradle scan link (add `--scan` option when running the gradle task): -
* Operating System and version: Windows 10
* Link to your project (if it's a public repository): -

TWiStErRob:
Might be a feature request. I would class it as a "usability problem".

chao2zhang:
Agreed. Currently, we only support suppressing in the `KtTryExpression`. So the code below would work in the short-term without waiting for the next detekt release.
```
@Suppress("SwallowedException")
try {
  // more stuff
} catch (ex: AssertionError) {
  // more stuff
}
```

blundell:
Would be nice if variables named `ignore` where ignored:

```
} catch (ignore: AssertionError) {
```

blundell:
annd this was from 2021, ignore me, you can configure it:

<img width="647" alt="image" src="https://github.com/detekt/detekt/assets/655860/6955b133-43cf-4a47-b841-f9f4d70110b9">

https://detekt.dev/docs/rules/exceptions/

PR #4158, opened 2021-10-03
Report SwallowedException on catchParameter

chao2zhang:
This improves the ergonomics so that users can minimize the suppression scope.
Addresses #4154

I am using the detekt repo itself as the integration test for the PR change, otherwise I probably need to write an integration test in detekt-core which does not seem like the right place because the change is in `detekt-rules-exceptions`.

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/JCommander.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/JCommander.kt
@@ -10,11 +10,11 @@ fun parseArguments(args: Array<out String>): CliArgs {
     val jCommander = JCommander(cli)
     jCommander.programName = "detekt"
 
-    @Suppress("SwallowedException") // Stacktrace in jCommander is likely irrelevant.
     try {
         @Suppress("SpreadOperator")
         jCommander.parse(*args)
-    } catch (ex: ParameterException) {
+    } catch (@Suppress("SwallowedException") ex: ParameterException) {
+        // Stacktrace in jCommander is likely irrelevant
         throw HandledArgumentViolation(ex.message, jCommander.usageAsString())
     }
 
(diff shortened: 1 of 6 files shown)


detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/TooGenericExceptionCaught.kt
```diff
@@ -53,12 +53,11 @@ class TooGenericExceptionCaught(config: Config) : Rule(config) {
     private val allowedExceptionNameRegex: Regex by config("_|(ignore|expected).*", String::toRegex)
 
     override fun visitCatchSection(catchClause: KtCatchClause) {
-        catchClause.catchParameter?.let {
-            if (isTooGenericException(it.typeReference) &&
-                !catchClause.isAllowedExceptionName(allowedExceptionNameRegex)
-            ) {
-                report(CodeSmell(issue, Entity.from(it), issue.description))
-            }
+        val catchParameter = catchClause.catchParameter ?: return
```
BraisGabin, line 56: Careful here. The return is avoiding the super call

BraisGabin (commented):
> I am using the detekt repo itself as the integration test for the PR change, otherwise I probably need to write an integration test in detekt-core which does not seem like the right place because the change is in `detekt-rules-exceptions`.

No you don't need to write any integration test. The engine that runs the rules in test honor the `@Suppress` so you can unit test it easily.

Or, if you prefer, you can test that the reported text is the correct one.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/JCommander.kt
```diff
@@ -10,11 +10,11 @@ fun parseArguments(args: Array<out String>): CliArgs {
     val jCommander = JCommander(cli)
     jCommander.programName = "detekt"
 
-    @Suppress("SwallowedException") // Stacktrace in jCommander is likely irrelevant.
     try {
         @Suppress("SpreadOperator")
         jCommander.parse(*args)
-    } catch (ex: ParameterException) {
+    } catch (@Suppress("SwallowedException") ex: ParameterException) {
```
TWiStErRob, line 16: Will it still accept suppressing on `try`? because a change shouldn't force everyone to move suppressions.

PR #4184, opened 2021-10-13
Use Codecov GitHub Action to upload coverage

### LINKED ISSUE #4183: Migrate to new Codecov uploader
## Expected Behavior
Code coverage is uploaded using a supported Codecov uploader.

## Observed Behavior
Code coverage is uploaded using the deprecated Bash uploader.

## Steps to Reproduce
N/A

## Context
https://about.codecov.io/blog/codecov-uploader-deprecation-plan/

3flex:
Fixes #4183

### DIFF
--- a/.github/workflows/codecoverage.yaml
+++ b/.github/workflows/codecoverage.yaml
@@ -29,5 +29,4 @@ jobs:
 
       - name: Publish Coverage
         if: success()
-        run: bash <(curl -s https://codecov.io/bash)
-        shell: bash
+        uses: codecov/codecov-action@v2


3flex:
> Merging #4184 (cdd3909) into main (eb62de7) will increase coverage by 0.54%.

This is a little surprising - it might have happened due to a bug in the old uploader that's fixed in the new one, or a bug in the new uploader... but as the Bash uploader is definitely deprecated and needs to be removed I don't think this should be a blocker, as we can consider Codecov and the uploader to be a black box. I did a quick search for info about behaviour changes in the new uploader but didn't uncover anything.

Issue #4169, opened 2021-10-07
[Feature] MatchJavadocParams


QuinnBast:
## Expected Behavior of the rule

Good:
```kt
/**
 * @param someArg Description of arg
 */
class MyClass(someArg: String)
```

Bad:
```kt
/**
 * @param someOutdatedArgument Description of arg
 * @param otherOutdatedArgument Description of arg
 */
class MyClass(someArg: String)
```

## Context
Not sure if detekt can parse comments, but it might be nice if Detekt can find outdated javadoc comments where the commented javadoc parameters don't match the actual parameters.

chao2zhang:
I think this makes sense, some edge cases I can think of according to https://kotlinlang.org/docs/kotlin-doc.html#kdoc-syntax:
- @param is legal for generic type parameters
- @property should be used if the parameter is kept as a property
- Generally, @param and return should be avoided https://kotlinlang.org/docs/coding-conventions.html#documentation-comments. We should may also need to check if each parameter is referenced in the main description body.

BraisGabin:
Sure! I just assigned it to you to make it clear ;)

If you need any help just write here or in the slack channel. And thanks!

deeprim:
To make it even tidier, what about such example:

**Bad:**
```kotlin
/**
 * @param secondArg second arg description
 * @param firstArg first arg description
 */
 class SomeClass(firstArg: String, secondArg: String)
 ```
 
 **Good:**
 ```kotlin
 /**
  * @param firstArg first arg description
  * @param secondArg second arg description
  */
  class SomeClass(firstArg: String, secondArg: String)
  ```
  
  Require the order of docs to be in alignment with the order of params

PR #4185, opened 2021-10-15
#4169 OutdatedDocumentation rule

lukaszosowicki:
Hi, I've prepared implementation of OutdatedDocumentation rule proposed in #4169

When KDoc is present and contains any `@param` or `@property` tag, rule will check if params and properties of class/function/constructor declaration matches the ones  from KDoc.

By default, both type and value parameters are considered, but I've added configuration for turning off checks for type parameters. Also the order of declarations matters in default configuration, but can be turned off as well.

If you want to suggest additional test scenarios or configurable behaviour feel free to comment and I will be happy to enhance this PR.


BraisGabin:
To me this PR is ready to merge. I would like to know your opinion about my comments but both can be done in a follow up PR.

The only thing stopping us to merge this is that CI is complaining. Could you check those issues?

PR #4186, opened 2021-10-16
Run CLI sanity checks with Gradle

3flex:
I find that I occasionally prep a PR, run `gradlew build` successfully then submit. Then there are failures because there are some CI jobs that aren't captured when running the `build` task.

This PR migrates these jobs to Gradle tasks that are run on `check` so a full successful `build` run will also pass on CI.


3flex:
We could, but to me it's neater to have everything configured & run from Gradle directly, and using the same tasks in CI. If we add the tasks to the Gradle build but leave CI alone then it's one method to run this in Gradle, and another in CI... why not keep it consistent?

Issue #1931, opened 2019-09-18
Property API for Gradle plugin


3flex:
## Expected Behavior
Gradle plugin properties are set using a consistent method throughout the plugin.

## Current Behavior
1. Some public properties using Provider API in some tasks but not others (e.g. `parallel` is a `Boolean` in `Detekt` task, but `Property<Boolean>` in `DetektCreateBaselineTask`)
1. Tasks (mostly) use Provider API in the back end while extensions use basic types which is a bit confusing
1. ~~Properties in `IdeaExtension` not using appropriate types, e.g. `path` is a string but should be a file type~~ (no longer relevant, IdeaExtension has been removed)
1. ~~`IdeaExtension` is being used as an extension on the task as well as DetektExtension which doesn't make sense - it should be one or the other. It should be configured via the extension, then the plugin should map extension properties to the task (same as the other tasks).~~ (no longer relevant, IdeaExtension has been removed)

## Context
I would like to use the Provider API (which will be stable in 6.0) wherever possible. This will be a breaking change, but in nearly all cases very easy to fix (e.g. change `parallel = true` to `parallel.set(true)`). The Groovy and Kotlin DSL will be identical.

The compatibility layer that was introduced earlier can be removed - this:
```kotlin
    @get:Internal
    internal val autoCorrectProp: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
    var autoCorrect: Boolean
        @Input
        get() = autoCorrectProp.get()
        set(value) = autoCorrectProp.set(value)
```
Would become:
```kotlin
    @get:Input
    val autoCorrect: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
```

I can include a brief migration guide and include this in the Gradle 6 migration PR.

chao2zhang:
Is it breaking for groovy consumers and kts consumers alike?

cortinico:
> Is it breaking for groovy consumers and kts consumers alike?

AFAIK It's breaking only for kts consumers. Groovy consumers won't be affected by the `Boolean` -> `Property<Boolean>` transition (there is a note about it here https://docs.gradle.org/current/userguide/lazy_configuration.html#lazy_properties)

PR #4190, opened 2021-10-17
Export Gradle extensions to detekt-gradle-plugin-extensions

cortinico:
I'm looking into moving the Gradle extensions to their own module.
The idea behind this is to let detekt-compiler-plugin depend on the newly published `detekt-gradle-plugin-extension`. 

This will make sure that the two extensions are kept in sync: https://github.com/detekt/detekt-compiler-plugin/blob/main/plugin-build/src/main/kotlin/io/github/detekt/gradle/extensions/KotlinCompileTaskDetektExtension.kt

The compiler plugin can still extend from those extensions if we wish to add extra properties or so.


cortinico:
> When you say they can be "kept in sync" what did you intend?

I mean that they should effectively be the same class.

> The extensions in the detekt-compiler-plugin are much simpler as they were created using Gradle's Property API, so I'm not sure if the intention is to change that and use the existing detekt extensions in the compiler plugin Gradle plugin (please no - this is such a good opportunity for a fresh start with the Gradle plugin)

That was actually my intention. Happy to re-discuss. I believe we should:
1. Update `detekt-compiler-plugin` to depend on `detekt-gradle-plugin-extensions` and use those.
2. Update the extensions to use the Property API

I understand that is nicer to have them separate as we can iterate faster, but it's hard for users to know what is the correct configuration for the extension. This also mean that we would have to maintain two documentation page and every change to `detekt-gradle-plugin` extension will have to be backported to the `detekt-compiler-plugin`. 

Ultimately, if we end up sharing the extensions, users could migrate from `detekt-gradle-plugin` to `detekt-compiler-plugin` by just changing the plugin id, and everything will "just work" at least from the Gradle point of view.

cortinico:
I understand your point,

> but it's going to be a breaking change for users using the Kotlin DSL as you pointed out in [#1931 (comment)](https://github.com/detekt/detekt/issues/1931#issuecomment-819812918), so I figured that it would be easier to that breaking change only for those who move to the compiler plugin.

I believe the problem here is that the Gradle plugin is following the same versioning schema as detekt-core, and the result is that we never release breaking changes. If we agree that the current Gradle API is far from optimal, we should deprecate it, and call a 2.x with the new change.

As for this specific PR, I'm fine closing it. I anyway think that keeping the two extensions is sync is a pain as the feature set of the two plugins is not aligned, and it requires a lot of back and forth (both for us and for the users).

BraisGabin:
I think that unlink the gradle plugin releases with the project releases is a good idea.

PR #4212, opened 2021-10-31
Report UntilInsteadOfRangeTo for 'rangeTo' calls

### LINKED ISSUE #4207: UntilInsteadOfRangeTo doesn't report when `rangeTo` is used
Both of these uses of `rangeTo` should be flagged:
```kotlin
        for (i in 0.rangeTo(10 - 1)) {}
        val range = 0.rangeTo(10 - 1)
```

## Context
The example above is semantically equivalent to:
```kotlin
        for (i in 0..10 - 1) {}
        val range = 0..10 - 1
```

Which is flagged as expected.

t-kameyama:
Fixes #4207

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
@@ -7,11 +7,11 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
-import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.KtBinaryExpression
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UntilInsteadOfRangeTo.kt
```diff
@@ -37,23 +37,33 @@ class UntilInsteadOfRangeTo(config: Config = Config.empty) : Rule(config) {
...
         }
-        return false
+        super.visitCallExpression(expression)
+    }
+
+    private fun KtExpression?.isMinusOneExpression() = this is KtBinaryExpression &&
+        left != null && operationToken == KtTokens.MINUS && (right as? KtConstantExpression)?.text == "1"
+
+    private fun report(expression: KtExpression, rageTo: String) {
+        report(CodeSmell(issue, Entity.from(expression), "'$rageTo' call can be replaced with 'until'"))
```
cortinico, line 62:
```suggestion
    private fun report(expression: KtExpression, rangeTo: String) {
        report(CodeSmell(issue, Entity.from(expression), "'$rangeTo' call can be replaced with 'until'"))
```

cortinico (approved):
LGTM, just a minor nit

cortinico (approved):

chao2zhang (approved):

Issue #3866, opened 2021-06-12
Cannot configure Detekt task options individually


damianw:
From what I can tell, it's not currently possible to configure some options on individual `Detekt` tasks, for example:

```kotlin
tasks.withType<Detekt>().configureEach {
    val myCustomDestinationForThisTaskXmlReport = somethingBasedOnTheTask(this)
    val myCustomDestinationForThisTaskCustomReport = somethingElseBasedOnTheTask(this)
    reports {
        xml {
            destination = myCustomDestinationForThisTaskXmlReport
        }
        custom {
            reportId = "MyCustomReport"
            destination = myCustomDestinationForThisTaskCustomReport
        }
        // Other possible customizations
    }
    // Other custom options
}
```

The use case here is specifying custom variant-specific destinations for standard and custom reports.

## Expected Behavior
Configuring tasks in a manner such as the example should configure those tasks as expected.

## Observed Behavior
There seem to be two issues:
1. The configuration order of `Detekt` tasks puts the [Detekt internal configuration block](https://github.com/detekt/detekt/blob/4d4a11a054c35b1f077ef6c326a63e9bbdfc270d/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/SharedTasks.kt#L11-L28) _last_, so it overrides the configuration from `configureEach`. (This was actually surprising to me - I had assumed that the block passed to `register` would be executed before other configuration actions... so TIL I guess?)
2. The [assignment of](https://github.com/detekt/detekt/blob/4d4a11a054c35b1f077ef6c326a63e9bbdfc270d/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt#L119) `reports = extension.reports` re-uses the same `DetektReports` object, thus causing it to be shared across all tasks. Configuring one task will result in them all being configured with the same reports.

## Steps to Reproduce

Configure Detekt tasks in a project with different outputs - for example, based on the task name:

```kotlin
tasks.withType<Detekt>().configureEach {
    reports {
        xml {
			enabled = true
            destination = File(project.buildDir, "$name/detekt.xml")
        }
    }
}
```

Observe that all of the tasks will output to the same location.

## Context
[...]

PR #4214, opened 2021-10-31
Replace lock with stale bot

cortinico:
I'm removing the `lock.yaml` file for the bot which is not working anymore (do we need it still?). Given that it was a 3rd party tool, I've moved to a 1st party Github actions that is doing something similar: marking issues inactive since more than 180 days (should I change it)?

Happy to tune it if it doesn't fit our needs.


chao2zhang (approved):
In my opinion, this number should be close to 2 release cycles - My justification is that in case we ask the PR author to hold off until the next release is published.

It feels like 180 days is on the conservative side, but this looks reasonable to me.

3flex:
> It feels like 180 days is on the conservative side

Agree - maybe 90 days? Or perhaps 90 days for PRs and 180 days for issues. PRs shouldn't hang around open for a long time (and yes, I realise I'm guilty of doing this!)

Also does this lock issues/PRs if they're stale & closed, or does it just close them? If they're just closed then 👍 since they can always be reopened by the reporter later if they still want to work on it or the issue is still something they care about.

cortinico:
> Also does this lock issues/PRs if they're stale & closed, or does it just close them?

This will just close the issues. 

> Agree - maybe 90 days?

Updating to 90

3flex:
Just noticed the message posted when an issue or PR is marked as stale says that the Stale label should be removed, but only those with write access on the repo can do that.

Better to just say that a comment should be posted.

3flex (approved):

3flex:
I just commented on an issue marked stale by the new bot but the Stale label wasn't removed... Does the config need changes?

See #3866

cortinico:
> I just commented on an issue marked stale by the new bot but the Stale label wasn't removed... Does the config need changes?

Not really. The action runs daily so it will be updated tonight. The doc says:

> If an update/comment occur on stale issues or pull requests, the stale label will be removed and the timer will restart

BraisGabin:
I have "internal conflicts" with this. I agree that we have far too much issues opened and it's difficult to focus in the important things. But do we want to auto-close valid but low-priority issues?

cortinico:
We can add some labels that will be filtered out by the action. For instance maybe the "help wanted" labelled should never get stale. Or we can create a "never get stale" label that triggers this.

BraisGabin:
Oh, that would be great. We do need to refactor our tags

PR #4222, opened 2021-11-01
Implement the rule to check if dispatchers are injectable

### LINKED ISSUE #4175: New rule: Always inject dispatchers
When standard dispatchers like `Dispatchers.IO/Default/Unconfined` are being used directly without dependency injection, violations should be reported.

The following scenarios shouldn't be reported.
- Coroutine dispatchers used as a default argument of the constructor.
- `Dispatchers.Main`/`Dispatchers.Main.Immediate` because this is overrideable by [Dispatchers.setMain] (https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/set-main.html) and [Dispatchers.resetMain](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/reset-main.html)

## Context
If a coroutine dispatcher is not injected, it is extremely difficult to write unit tests that can complete.
This is recommended by https://developer.android.com/kotlin/coroutines/coroutines-best-practices#inject-dispatchers
as well as https://youtu.be/KMb0Fs8rCRs?t=839.

chao2zhang:
This resolves #4175.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -156,6 +156,12 @@ coroutines:
   active: true
   GlobalCoroutineUsage:
     active: false
+  InjectDispatcher:
+    active: false
+    dispatcherNames:
+      - 'IO'
+      - 'Default'
+      - 'Unconfined'
   RedundantSuspendModifier:
     active: false
   SleepInsteadOfDelay:
--- a/detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/InjectDispatcher.kt
+++ b/detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/InjectDispatcher.kt
@@ -1,0 +1,76 @@
+package io.gitlab.arturbosch.detekt.rules.coroutines
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
+import io.gitlab.arturbosch.detekt.rules.fqNameOrNull
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.KtSimpleNameExpression
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
+import org.jetbrains.kotlin.resolve.BindingContext.EMPTY
+import org.jetbrains.kotlin.resolve.calls.callUtil.getType
+import org.jetbrains.kotlin.types.typeUtil.supertypes
+
+/**
+ * Always use dependency injection to inject dispatchers for easier testing.
+ * This rule is based on the recommendation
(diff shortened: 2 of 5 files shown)


detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/InjectDispatcher.kt
```diff
@@ -0,0 +1,76 @@
...
+@RequiresTypeResolution
+class InjectDispatcher(config: Config) : Rule(config) {
+
+    @Configuration("The names of dispatchers to detect by this rule")
+    private val dispatcherNames: Set<String> by config(listOf("IO", "Default", "Unconfined")) { it.toSet() }
+
+    override val issue = Issue(
+        "InjectDispatcher",
+        Severity.Defect,
+        "Don't hardcode dispatchers when creating new coroutines or calling withContext. " +
```
cortinico, line 50:
```suggestion
        "Don't hardcode dispatchers when creating new coroutines or calling `withContext`. " +
```

cortinico (approved):
Great stuff :)

cortinico (approved):

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -156,6 +156,12 @@ coroutines:
   active: true
   GlobalCoroutineUsage:
     active: false
+  InjectDispatcher:
+    active: false
+    dispatcherNames:
+      - 'IO'
+      - 'Default'
+      - 'Unconfined'
```
BraisGabin, line 164: I don't understand the use case to allow to configure this? Shouldn't we just check all the usages of `Dispatchers.*`? If this is because you can change the `Main` and the `Immediate` maybe a `boolean` flag (disable by default) could be a better fit for discoberavility.

chao2zhang: Dispatchers.Main is generally okay because we have Dispatchers.setMain()/resetMain() available in test (https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test/)

Additionally, if users define their own Dispatchers, this list is necessary for specifying their customized `dispatcherNames`.

BraisGabin: > Dispatchers.Main is generally okay because we have Dispatchers.setMain()/resetMain()

It's possible to change them, but I think that adding `Main` is a safer default. If someone is OK using those function it can remove it. This kind of test functions aren't very good. If you forget to use `resetMain` you can change the behaviour of other tests.

For example JodaTime had `DateTimeUtils.setCurrentMillisFixed` but it was removed in `java.time`. Now you are forced to inject a `Clock` if you want to test it. I think that we should add `Main` but default for the same reason.

BraisGabin (approved):

detekt-core/src/main/resources/default-detekt-config.yml
chao2zhang: Jetbrain announces its plan to improve kotlin coroutin test https://youtrack.jetbrains.com/issue/KT-46783.

On the other hand, [Dispatchers.Main] is used in common androidx libraries:
- Paging3: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:paging/paging-common/src/main/kotlin/androidx/paging/PagingDataDiffer.kt?q=Dispatchers.Main&ss=androidx
- LiveData: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:lifecycle/lifecycle-livedata-ktx/src/main/java/androidx/lifecycle/CoroutineLiveData.kt?q=Dispatchers.Main&ss=androidx&start=61
- Lifecycle: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:lifecycle/lifecycle-runtime-ktx/src/main/java/androidx/lifecycle/RepeatOnLifecycle.kt?q=Dispatchers.Main&ss=androidx&start=91

Since most UI applications do have a single `Main` thread, we are more likely to get complaints if we add `Main` as the default.

BraisGabin: I don't completely agree but we could move this to an issue. There's no need to block the merging for this.

PR #4225, opened 2021-11-02
Workaround for JDK 8 instability when reading config

davidburstromspotify:
If Detekt is being executed in parallel in the same VM
but through different class loader instances, the configuration file
is sometimes not readable as a class loader might have closed
the zip stream. By telling the JVM to not use cached files readers,
the stability improves.

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/YamlConfig.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/YamlConfig.kt
@@ -61,7 +61,17 @@ class YamlConfig internal constructor(
         /**
          * Factory method to load a yaml configuration from a URL.
          */
-        fun loadResource(url: URL): Config = load(url.openStream().reader())
+        fun loadResource(url: URL): Config = load(
+            url.openConnection()
+                /*
+                 * Due to https://bugs.openjdk.java.net/browse/JDK-6947916 and https://bugs.openjdk.java.net/browse/JDK-8155607,
+                 * it is necessary to disallow caches to maintain stability on JDK 8. Otherwise, simultaneous invocations of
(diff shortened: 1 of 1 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/YamlConfig.kt
```diff
@@ -61,7 +61,16 @@ class YamlConfig internal constructor(
...
-        fun loadResource(url: URL): Config = load(url.openStream().reader())
+        fun loadResource(url: URL): Config = load(
+            url.openConnection()
+                /*
+                 * Due to https://bugs.openjdk.java.net/browse/JDK-6947916 and https://bugs.openjdk.java.net/browse/JDK-8155607,
+                 * it is necessary to disallow caches to maintain stability on JDK 8. Otherwise, simultaneous invocations of
+                 * Detekt in the same VM can fail spuriously. A similar bug is referenced in
+                 * https://github.com/detekt/detekt/issues/3396. The performance regression is likely unnoticeable.
+                 */
+                .apply { useCaches = true }.getInputStream().reader()
```
BraisGabin, line 72: Shoudln't this be `useCaches = false` then? And, if this is a JDK issue, could we do only for JDK8?

And this is probably another reason to deprecate JDK8 #4130

davidburstrom: Doh! It's a remnant from trying to set up a test case but failing.

As for JDK8 only, it could be worth adding the condition, so it is detected and removed while removing support for JDK8.

davidburstrom: I've updated the PR.

BraisGabin (approved):

cortinico (approved):

PR #4227, opened 2021-11-03
Bump KtLint to 0.43.2 and add TrailingComma rule

### LINKED ISSUE #2398: New rule: (Require|Forbid)TrailingCommas
Kotlin 1.3.70 supports trailing commas in various places. Detekt should be able to enforce a consistent style.
https://youtrack.jetbrains.com/issue/KT-34743

## Context
The original request said to support trailing commas in the following places:
- Class parameters
- Function value parameters
- Parameters with optional type (anonymous functions and setters)
- Value arguments
- Multi variable declaration (destructuring assignments)
- Indexing suffix (array access expressions)
- Lambda parameters
- Type parameters
- Function type parameters
- Type arguments
- When entry
- Collection literals (in annotations)

Looking for feedback to determine whether configuration for each type is desired or not, or whether to simply enforce (or ban) trailing commas in all places they're supported or not.

cortinico:
See https://github.com/pinterest/ktlint/releases/tag/0.43.0 for the full release note. I've also wrapped the new experimental `TrailingComma` rule.

I have to mention that I'm not entirely happy with the naming of the config keys:

```
    allowTrailingComma: false
    allowTrailingCommaOnCallSite: false
```

As the rule is doing more than just _allowing_. Setting the value to `true` is effectively enforcing trailing commas, while setting it to `false` (the default) is forbidding it. I've anyway followed the same name that ktlint is having.

Plus there seems to be some regressions on KtLint 0.43.0:
- https://github.com/pinterest/ktlint/issues/1268
- https://github.com/pinterest/ktlint/issues/1267
- https://github.com/pinterest/ktlint/issues/1274

Fixes #2398


3flex:
> I have to mention that I'm not entirely happy with the naming of the config keys

Funnily enough that exact feedback was also provided on the ktlint issue, and they said they followed the terminology used in IntelliJ (which also says "allow" but acts as "require").

So all tooling is going to have a confusing name for this but at least it's consistent :/

IMHO it's better to change the name of the config option to reflect the effect it has.

cortinico:
> IMHO it's better to change the name of the config option to reflect the effect it has.

I would agree on this, if this would be a Detekt first party rule. But as we're wrapping a 3rd party rule, let's stick to the naming already decided by others.

BraisGabin:
I vote to don't merge until those regressions are fixed.

BraisGabin (approved):

ZacSweers:
Another possible issue - fyi https://github.com/pinterest/ktlint/issues/1274

cortinico:
> Another possible issue - fyi [pinterest/ktlint#1274](https://github.com/pinterest/ktlint/issues/1274)

Added to the list 👍

romtsn:
> > I have to mention that I'm not entirely happy with the naming of the config keys
> 
> Funnily enough that exact feedback was also provided on the ktlint issue, and they said they followed the terminology used in IntelliJ (which also says "allow" but acts as "require").
> 
> So all tooling is going to have a confusing name for this but at least it's consistent :/
> 
> IMHO it's better to change the name of the config option to reflect the effect it has.

the reasoning behind that is we want a centralized configuration for both ktlint and IJ (so they don't conflict and confuse people) which appears to be `.editorconfig`, and there it's named like that. I'd happily rename it to something more meaningful if it was under our control. Maybe worth throwing an issue on youtrack.

3flex:
> I vote to don't merge until those regressions are fixed.

It's been 6 weeks and not much movement on those regressions. Generally we are prompt at updating dependencies and we don't directly control what happens in ktlint, perhaps it's best to update but make sure docs make clear how ktlint version can be overridden when using the wrapper?

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/TrailingComma.kt
```diff
@@ -0,0 +1,33 @@
...
+import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.internal.AutoCorrectable
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import io.gitlab.arturbosch.detekt.formatting.FormattingRule
+
+/**
+ * See <a href="https://ktlint.github.io/#rule-spacing">ktlint-website</a> for documentation.
+ */
+@OptIn(FeatureInAlphaState::class)
+@AutoCorrectable(since = "1.19.0")
```
3flex, line 16:
```suggestion
@AutoCorrectable(since = "1.20.0")
```

3flex (approved):

cortinico:
> It's been 6 weeks and not much movement on those regressions. Generally we are prompt at updating dependencies and we don't directly control what happens in ktlint, perhaps it's best to update but make sure docs make clear how ktlint version can be overridden when using the wrapper?

Agree. I think it's safe to merge this. Should a point release on KtLint's end happen, we're going to include it inside `1.20.0` regardless 👍

PR #4239, opened 2021-11-08
Handle MultiRules in Suppressors

### LINKED ISSUE #4230: ignoreAnnotated config doesn't work on FunctionNaming rule
## Expected Behavior
Being able to suppress `FunctionNaming` rule on functions annotated with `@Composable`

## Observed Behavior
All `@Composable` function names reported as violations

## Steps to Reproduce
Run detekt 1.19.0-RC1 analysis on a project with Composable functions

## Context
It seems that naming rules are all run through one MulitRule: [NamingRules](https://github.com/detekt/detekt/blob/28830b8db5adb37c08c65e70f9251d9396681b24/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NamingRules.kt) which is not ConfigAware so [getSuppressors](https://github.com/detekt/detekt/blob/edabb6f2b98cf4021606ff48e6acd13be1ef6207/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/Suppressors.kt#L13) returns empty list

## Your Environment
* Version of detekt used: 1.19.0-RC-1

BraisGabin:
`MultiRule`s are a pain to maintain (more about this here: #4240). But for now this code should do the trick.

~I don't know how to test this properly. I manually tested it.~

Fixes #4230

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/Suppressors.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/Suppressors.kt
@@ -3,18 +3,39 @@ package io.gitlab.arturbosch.detekt.core.suppressors
 import io.gitlab.arturbosch.detekt.api.BaseRule
 import io.gitlab.arturbosch.detekt.api.ConfigAware
 import io.gitlab.arturbosch.detekt.api.Finding
+import io.gitlab.arturbosch.detekt.api.MultiRule
+import io.gitlab.arturbosch.detekt.api.Rule
 
 /**
  * Given a Finding it decides if it should be suppressed (`true`) or not (`false`)
  */
 typealias Suppressor = (Finding) -> Boolean
 
+private fun buildSuppressors(rule: ConfigAware): List<Suppressor> {
+    return listOfNotNull(
+        annotationSuppressorFactory(rule),
+    )
+}
+
 internal fun getSuppressors(rule: BaseRule): List<Suppressor> {
-    return if (rule is ConfigAware) {
-        listOfNotNull(
-            annotationSuppressorFactory(rule),
-        )
-    } else {
-        emptyList()
+    return when (rule) {
+        is MultiRule -> rule.rules.flatMap { innerRule ->
+            buildSuppressors(innerRule).map { suppressor -> InnerSuppressor(innerRule, suppressor) }
+        }
+        is ConfigAware -> buildSuppressors(rule)
+        else -> emptyList()
+    }
+}
+
+private class InnerSuppressor(
+    private val rule: Rule,
+    private val suppressor: Suppressor
+) : Suppressor {
+    override fun invoke(finding: Finding): Boolean {
+        return if (finding.id == rule.issue.id) {
+            suppressor(finding)
+        } else {
+            false
+        }
     }
 }
--- a/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/SuppressorsSpec.kt
+++ b/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/SuppressorsSpec.kt
@@ -1,0 +1,57 @@
+package io.gitlab.arturbosch.detekt.core.suppressors
+
+import io.github.detekt.test.utils.compileContentForTest
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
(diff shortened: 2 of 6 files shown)


cortinico:
> I don't know how to test this properly. I manually tested it.

I believe you should be able to unit test this, no?

BraisGabin:
They are not the best test I ever wrote but they should be enough.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/Suppressors.kt
```diff
@@ -3,18 +3,39 @@ package io.gitlab.arturbosch.detekt.core.suppressors
...
+        is ConfigAware -> buildSuppressors(rule)
+        else -> emptyList()
+    }
+}
+
+private class InnerSuppressor(
+    private val rule: Rule,
+    private val suppressor: Suppressor
+) : Suppressor {
+    override fun invoke(finding: Finding): Boolean {
```
chao2zhang, line 34: One benefit using `fun interface` over `typealias` is that the method name would make more sense without looking at the documentation. `shouldSuppress(finding: Finding): Boolean` is definitely more contextual than `invoke(finding: Finding)`

Additionally, we could take `List<Suppressor>` as a `CompositeSuppressor` to simplify the test code.

I don't think it is necessary to address this in this PR. But would like to know your thoughts.

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/SuppressorsSpec.kt
```diff
@@ -0,0 +1,78 @@
...
+                """
+                fun foo() = Unit
+                """.trimIndent()
+            )
+            CodeSmell(ARule().issue, Entity.from(root), "")
+        }
+        val ignorableCodeSmell by memoized {
+            val root = compileContentForTest(
+                """
+                @file:Composable
```
chao2zhang, line 31: Practically, the @Composable should an annotation on the function, not at the file leve.

chao2zhang (approved):

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/SuppressorsSpec.kt
BraisGabin: Yes, it could. But then I can't report directly `root`. I should find the `KtFunction` inside `root`. For that reason I added the annotation as `@file:`

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/Suppressors.kt
BraisGabin: About `fun interface`, I'm ok. The only "problem" is that for instantiate it instead of doing: `{}` we should need to write `Suppressor {}`. But I agree that invoke isn't a good name. I'm going to change it.

About `CompositeSuppressor` I don't think it will make the code simpler.

BraisGabin: I just did the `fun interface`... And I'm not sure if I like it more or less: https://github.com/detekt/detekt/commit/43573a919e71380ab7fc31a53f3b4f1e284971a4

What do you thinkg? I like the old `suppress(finding)` syntax. But, when implementing a non-anonymous `Suppressor` I agree that the `shouldSupress` is better.

chao2zhang: The commit [43573a9](https://github.com/detekt/detekt/commit/43573a919e71380ab7fc31a53f3b4f1e284971a4) looks better in my opinion.

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/SuppressorsSpec.kt
chao2zhang: Makes sense. I think we still have sufficient coverage with the existing implementation.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/Suppressors.kt
BraisGabin: Added

chao2zhang (approved):

PR #4241, opened 2021-11-08
Support full qualified names in AnnotationSuppressor

### LINKED ISSUE #4233: Support full qualified names for annotations in ignoreAnnotated
## Expected Behavior
```yaml
  LateinitUsage:
    active: true
    ignoreAnnotated:
     - 'javax.inject.Inject'
```

Should ignore properties annotated with `@Inject` the same way as the old configuration:

```yaml
  LateinitUsage:
    active: true
    excludeAnnotatedProperties:
     - 'javax.inject.Inject'
```

## Observed Behavior
It doesn't

## Your Environment
* Version of detekt used: 1.19.0-RC1

BraisGabin:
This add supports to full qualified name matching in `AnnotationSuppresor`. It also allow Glob patterns.

Fixes #4233

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
@@ -10,7 +10,9 @@ import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 import org.jetbrains.kotlin.resolve.BindingContext
 
 internal fun annotationSuppressorFactory(rule: ConfigAware, bindingContext: BindingContext): Suppressor? {
-    val annotations = rule.valueOrDefault("ignoreAnnotated", emptyList<String>())
+    val annotations = rule.valueOrDefault("ignoreAnnotated", emptyList<String>()).map {
+        it.qualifiedNameGlobToRegex()
+    }
     return if (annotations.isNotEmpty()) {
         Suppressor { finding ->
             val element = finding.entity.ktElement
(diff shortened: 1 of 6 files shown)


BraisGabin:
cc @TWiStErRob. This was exactly what we talk about at the Droidcon. As you can see it's not as "clean" as before but now you can use both name or full qualified names.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
```diff
@@ -1,30 +1,65 @@
...
 
-private fun KtElement.isAnnotatedWith(annotationNames: Iterable<String>): Boolean {
-    return if (this is KtAnnotated && annotationEntries.find { it.typeReference?.text in annotationNames } != null) {
-        true
-    } else {
-        getStrictParentOfType<KtAnnotated>()?.isAnnotatedWith(annotationNames) ?: false
+private fun KtElement.isAnnotatedWith(annotationNames: Iterable<Regex>, bindingContext: BindingContext): Boolean {
+    if (this is KtAnnotated) {
+        val references = annotationEntries.mapNotNull { it.typeReference }
+        if (references.find { it.text in annotationNames } != null) {
```
TWiStErRob, line 37: I noticed this pattern multiple times, wouldn't
```suggestion
        if (references.any { it.text in annotationNames }) {
```
be more expressive?

TWiStErRob (commented):
Nice one, thanks for the cc. It looks ok, I agree with doing the best possible based on binding context or not.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
BraisGabin: That's true! Thanks! I edited it manually because I had that issue twice and I need to fix other minor issues.

This could be a good rule to add in detekt. Use `any` instead of `find`.

BraisGabin: I created an issue to track this new rule: #4243

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
```diff
@@ -1,30 +1,65 @@
...
 }
 
-private fun annotationSuppressor(element: KtElement, annotations: List<String>): Boolean {
-    return element.isAnnotatedWith(annotations)
+private fun annotationSuppressor(
+    element: KtElement,
+    annotations: List<Regex>,
+    bindingContext: BindingContext
+): Boolean {
+    return element.isAnnotatedWith(annotations, bindingContext)
```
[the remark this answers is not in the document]
BraisGabin: I think it is here for historical reasons. The code was evolving and now it does nothing at all. Well spotted. I'm going to remove it.

cortinico (approved):

PR #4263, opened 2021-11-14
Replace usage of deprecated ConfigureUtil

3flex:
ConfigureUtil is planned for removal in Gradle 8.0.

Use of `Closure` has been discouraged for some time. It can be replaced with `Action` ; https://docs.gradle.org/current/userguide/kotlin_dsl.html#groovy_closures_from_kotlin

> Gradle plugins written in any language should prefer the type `Action<T>` type in place of closures. Groovy closures and Kotlin lambdas are automatically mapped to arguments of that type.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektReports.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektReports.kt
@@ -1,12 +1,11 @@
 package io.gitlab.arturbosch.detekt.extensions
 
-import groovy.lang.Closure
 import io.gitlab.arturbosch.detekt.extensions.DetektReportType.HTML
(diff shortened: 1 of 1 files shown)


BraisGabin (approved):

schalkms (approved):

cortinico:
> Use of `Closure` has been discouraged for some time. It can be replaced with `Action` ; [docs.gradle.org/current/userguide/kotlin_dsl.html#groovy_closures_from_kotlin](https://docs.gradle.org/current/userguide/kotlin_dsl.html#groovy_closures_from_kotlin)

Is this going to be a breaking change for any of our user (from the Groovy or Kotlin point of view)? I just wanted to know so this can be pointed out in the release notes.

3flex:
> Is this going to be a breaking change for any of our user

There shouldn't be any breaking change, as Gradle will transparently do what's needed in the background so either Groovy closure or Kotlin lambda will be mapped to the `Action` automatically.

PR #4264, opened 2021-11-14
Use appropriate annotations on source properties in Gradle tasks

3flex:


BraisGabin (approved):

chao2zhang (commented):
Was the addition of `@IgnoreEmptyDirectories` to prevent known bugs?

3flex:
> Was the addition of @IgnoreEmptyDirectories to prevent known bugs?

No, no bugs. It just means that when input files are snapshot for up to date checks that empty directories will be ignored, increasing performance if the only change in the source was the inclusion (or removal) of an empty directory.

`getSource` on `SourceTask` has had the annotation since Gradle 6.8, but it needs to be added here to take effect since we're overriding to update the path sensitivity annotation.

chao2zhang (approved):

chao2zhang:
Something I learned is that we override the path sensitivity in https://github.com/detekt/detekt/pull/2182. I guess this can be useful if we are exploring remote build cache in the future.

PR #4265, opened 2021-11-14
AssertJ 3.21.0

3flex:


3flex:
Failing due to: https://github.com/assertj/assertj-core/issues/2357

The workarounds aren't very nice, so it might be best to stay on 3.20.2 for now.

3flex:
KT-17765 should be fixed in Kotlin 1.6.20, so when that is released it should be possible to rebase and merge.

3flex:
Though KT-17765 seems fixed in Kotlin 1.6.20-M1 it requires using experimental language version 1.7, e.g.:

```kotlin
tasks.compileTestKotlin {
    kotlinOptions.languageVersion = "1.7"
}
```

We could enable this for test tasks only, but as it's considered experimental, I think it's better to hold back on this until the fix goes stable in Kotlin 1.7. Using the older AssertJ version doesn't block anything.

BraisGabin (approved):

PR #4271, opened 2021-11-15
Remove afterEvaluate wrapper v2

3flex:
v2 of #4159

The reason this didn't work on the first attempt is that `bootClasspath` on the Android Gradle Plugin's `BaseExtension` is a `List<File>` and was being called during the configuration phase when it was empty. By wrapping in a `provider {}` call the parameter becomes live so the value will be queried at task execution instead, when it's been populated. cc @BraisGabin in case you want to confirm (I have successfully tested on a local Android project).

I've also reversed the test behaviour change introduced in #4262 as the test no longer fails.

I still need to add a functional test for Android. I will do this once #4074 is merged.

Original PR description follows:

---

Removes usage of `afterEvaluate` in the Gradle plugin.

This causes race conditions (evidenced by the double use of `afterEvaluate` in DetektMultiplatform class) and makes it more difficult for users to customise individual tasks created by the plugin.

Generally `afterEvaluate` shouldn't be required when Gradle's lazy configuration and task configuration avoidance APIs are used. The only tests that failed when I removed it is the multiplatform Android tests, but switching to reactive DomainObjectContainer#all instead of DomainObjectContainer#forEach fixes that.

If there are other use cases that require using `afterEvaluate` I'll investigate, and if the test coverage didn't pick up issues that removal will now cause, please let me know so I can address it.

Would fix root issue raised in #3428

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
@@ -68,7 +68,7 @@ internal class DetektAndroid(private val project: Project) {
         // There is not a single Android plugin, but each registers an extension based on BaseExtension,
         // so we catch them all by looking for this one
         project.extensions.findByType(BaseExtension::class.java)?.let { baseExtension ->
-            val bootClasspath = project.files(baseExtension.bootClasspath)
+            val bootClasspath = project.files(project.provider { baseExtension.bootClasspath })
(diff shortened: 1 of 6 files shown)


BraisGabin (approved):

cortinico:
Great stuff 🚀 Do we want this inside `1.19.0-RC2`?

cortinico (approved):

3flex:
> Do we want this inside 1.19.0-RC2?

My preference is to include it if possible, though there's no test coverage yet...

cortinico (approved):

Issue #3147, opened 2020-10-16
CamelCase should be accepted for top level constant names


JavierSegoviaCordoba:
As an alternative to companion object const variable, we can just create top level constants and CamelCase should be accepted too, example:

```
private const val AndroidxSnapshotBuildId = 6911187
```

schalkms:
Is there a reference that highlights whether this okay according to Kotlin's style guides?

JavierSegoviaCordoba:
@schalkms sorry but I read about it in a discussion about Google members in compose and they referenced JetBrains was using it too (at least in companion object). I think there is no difference between private const inside or outside companion object (just performance)

BraisGabin:
https://kotlinlang.org/docs/reference/coding-conventions.html#property-names

> Names of top-level or object properties which hold objects with behavior or mutable data should use camel-case names

It says that only for mutable objects but we can't know if the object is mutable or not so we should allow it.

BraisGabin:
Ups, nope, sorry. I'm wrong here. That documention doesn't apply here. This are `const` and all the `const` are immutable so it should not apply.

JavierSegoviaCordoba:
ktor source code is using CamelCase multiple times for Companion objects, objects and top level const val properties. I think JetBrains will add it to naming conventions in the future.

BraisGabin:
The thing here is if we want to follow Jetbrains/Google or the convention. And I think that the default values in detekt should follow the convention. We allow the user to customize this rule to do exactly that:

```diff
   TopLevelPropertyNaming:
-    constantPattern: '[A-Z][_A-Z0-9]*'
+    constantPattern: '[A-Za-z][_A-Za-z0-9]*'
```

So, my vote is to follow the convention and wait for Jetbrains to update it. I'm not agains it neither, so if the other maintainer/more users think that this change is good we can merge the PR.

schalkms:
I also prefer to keep it, since this config change would be kind of controversial as of now. If the official style guide is updated, we can still adapt the config later on.

JavierSegoviaCordoba:
Compose released its guidelines, one of them is what I mentioned here.

Should Detekt provides docs about what rules are necessary to change to avoid problems with the Compose guidelines?

I understand the default behavior of Detekt should be following the Kotlinlang guidelines, but having a doc which explain how to use specific rules (and what rules) for Compose should be great.

https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md

[...]

PR #4272, opened 2021-11-15
Landing page for Compose-related rule customization

drinkthestars:
### What
Addresses https://github.com/detekt/detekt/discussions/4078#discussioncomment-1256162, https://github.com/detekt/detekt/discussions/4078#discussioncomment-1275658

- Adds a landing page within `User Documentation` to go over rule sets that might need customization for Compose styles/usage
- Where "Compose usage" refers to [Compose API Guidelines](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md), as well as [Compose source](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/) as "canon"
- Adds a few starter entries for customizing `TopLevelPropertyNaming`, `LongParameterList`, & `MagicNumber`

### Why & Related
- One needs to search through discussions/issues to know about customizing detekt rules to accommodate Compose styles; there is no central place within detekt docs
- As Compose introduces new paradigms to Kotlin, there might be deviations from JetBrains Kotlin styles; having some centralized docs might become useful as Compose usage becomes more widespread
- A few discussions indicated a need for such docs:
   - https://github.com/detekt/detekt/discussions/4078
   - https://github.com/detekt/detekt/discussions/4078#discussioncomment-1256162
   - https://github.com/detekt/detekt/discussions/4078#discussioncomment-1275658
   - https://github.com/detekt/detekt/issues/3147#issuecomment-797377600
   - https://github.com/detekt/detekt/issues/4080#issuecomment-932996520

### Open thoughts
- Is `User Documentation` a good place to put this?
- Any ways to make this more maintainable? i.e. a new dicussion/issue around Compose is opened that yields some answers, should this page be updated with that answer?, etc.
- General thoughts around improving formatting/organization etc. of this page

### DIFF
--- a/docs/_data/sidebars/home_sidebar.yml
+++ b/docs/_data/sidebars/home_sidebar.yml
@@ -20,6 +20,9 @@ entries:
     - title: Configuration File
       url: /configurations.html
       output: web
+    - title: Configuration for Compose
+      url: /compose.html
+      output: web
     - title: Reporting
       url: /reporting.html
       output: web
@@ -37,9 +40,6 @@ entries:
       output: web
     - title: Compatibility Table
       url: /compatibility.html
-      output: web
-    - title: Customizing For Compose
-      url: /compose.html
       output: web
   - title: Getting Started
     output: web
--- a/docs/_data/sidebars/home_sidebar.yml
+++ b/docs/_data/sidebars/home_sidebar.yml
@@ -38,6 +38,9 @@ entries:
     - title: Compatibility Table
       url: /compatibility.html
       output: web
+    - title: Customizing For Compose
+      url: /compose.html
+      output: web
   - title: Getting Started
     output: web
     folderitems:
--- a/docs/pages/compose.md
+++ b/docs/pages/compose.md
@@ -1,0 +1,66 @@
+---
+title: Customizing for Compose
+sidebar: home_sidebar
+keywords: rules, compose, jetpack-compose
+permalink: compose.html
+toc: true
+folder: documentation
+---
+
+Relevant rule sets and their configuration options for Compose styles & usage.
+
+### TopLevelPropertyNaming
+
+See [TopLevelPropertyNaming](https://detekt.github.io/detekt/naming.html#toplevelpropertynaming).
+
+Compose [guidelines](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#singletons-constants-sealed-class-and-enum-class-values) prescribe `CamelCase` for top-level constants.
+
+##### Default Style:
+
+```kotlin
+private val FOO_PADDING = 16.dp
+```
+
+##### Compose Style:
+
+```kotlin
+private val FooPadding = 16.dp
+```
+
+#### Configurations:
+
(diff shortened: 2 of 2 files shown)


docs/pages/compose.md
```diff
@@ -0,0 +1,66 @@
...
+folder: documentation
+---
+
+Relevant rule sets and their configuration options for Compose styles & usage.
+
+### TopLevelPropertyNaming
+
+See [TopLevelPropertyNaming](https://detekt.github.io/detekt/naming.html#toplevelpropertynaming).
+
+Compose [guidelines](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#singletons-constants-sealed-class-and-enum-class-values) prescribe `CamelCase` for top-level constants.
```
chao2zhang, line 16: What do you think about moving the references to the guidelines in the top-level summary?

docs/pages/compose.md
```diff
@@ -0,0 +1,66 @@
...
+
+##### Compose Style:
+
+```kotlin
+private val FooPadding = 16.dp
+```
+
+#### Configurations:
+
+* Set ``constantPattern`` to ``'[A-Z][_A-Za-z0-9]*'`` (default: ``'[A-Z][_A-Z0-9]*'``)
```
chao2zhang, line 32:
```suggestion
* Set ``constantPattern`` to ``'[A-Z][A-Za-z0-9]*'`` (default: ``'[A-Z][_A-Z0-9]*'``)
```

docs/_data/sidebars/home_sidebar.yml
```diff
@@ -38,6 +38,9 @@ entries:
     - title: Compatibility Table
       url: /compatibility.html
       output: web
+    - title: Customizing For Compose
```
chao2zhang, line 41: It feels like the content is grouped with `Configuration File`, what do you think about naming the title closer and group them together in adjacent lines?

chao2zhang (commented):
Thanks for the contribution - I think documentation is definitely a first step. Once the compose guidelines have stabilized for over a year, we can definitely discuss if we need to provide different flavors of configs.

_Any way to make this more maintainable? i.e. a new discussion/issue around Compose is opened that yields some answers, should this page be updated with that answer?, etc._
Yes, I think we will need to manually update the documentation. The compose guideline in my opinion is still relatively new to most of the developers.

docs/pages/compose.md
drinkthestars: Sounds good, will do.

docs/_data/sidebars/home_sidebar.yml
drinkthestars: Will do. Are we thinking something like this?
Screen Shot 2021-11-14 at 10 10 50 PM

chao2zhang: Thanks. That's my thought.

drinkthestars: 👍🏼  8369719

docs/pages/compose.md
drinkthestars: 8369719

cortinico (approved):

chao2zhang (approved):

PR #4277, opened 2021-11-16
Kotlin 1.6.10

### LINKED ISSUE #4287: Support jvmTarget 17
Currently setting it to 17 results in `Invalid value passed to --jvm-target`

3flex:
https://blog.jetbrains.com/kotlin/2021/11/kotlin-1-6-0-is-released/

Fixes #4287

### DIFF
--- a/build-logic/src/main/kotlin/module.gradle.kts
+++ b/build-logic/src/main/kotlin/module.gradle.kts
@@ -89,13 +89,6 @@ tasks.withType<KotlinCompile>().configureEach {
             "-progressive",
             "-Xopt-in=kotlin.RequiresOptIn"
         )
-        // Usage: <code>./gradlew build -PwarningsAsErrors=true</code>.
-        // Note: currently there are warnings for detekt-gradle-plugin that seemingly can't be fixed
-        //       until Gradle releases an update (https://github.com/gradle/gradle/issues/16345)
-        allWarningsAsErrors = when (project.name) {
-            "detekt-gradle-plugin" -> false
-            else -> (project.findProperty("warningsAsErrors") == "true" || System.getenv("CI") == "true")
-        }
     }
 }
 
--- a/build-logic/src/main/kotlin/module.gradle.kts
+++ b/build-logic/src/main/kotlin/module.gradle.kts
@@ -83,7 +83,7 @@ configurations.create("coverageDataElements") {
 tasks.withType<KotlinCompile>().configureEach {
     kotlinOptions {
         jvmTarget = Versions.JVM_TARGET
-        languageVersion = "1.5"
+        languageVersion = "1.6"
         apiVersion = "1.4"
         freeCompilerArgs = listOf(
             "-progressive",
--- a/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
+++ b/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
@@ -5,19 +5,11 @@ import io.gitlab.arturbosch.detekt.testkit.DslGradleRunner
 import io.gitlab.arturbosch.detekt.testkit.ProjectLayout
 import io.gitlab.arturbosch.detekt.testkit.triggerEvaluation
 import org.assertj.core.api.Assertions.assertThat
-import org.gradle.api.Project
-import org.gradle.api.internal.project.DefaultProject
-import org.gradle.api.provider.Provider
-import org.gradle.build.event.BuildEventsListenerRegistry
-import org.gradle.internal.service.DefaultServiceRegistry
-import org.gradle.internal.service.scopes.ProjectScopeServices
 import org.gradle.kotlin.dsl.apply
 import org.gradle.kotlin.dsl.repositories
-import org.gradle.tooling.events.OperationCompletionListener
 import org.jetbrains.kotlin.gradle.plugin.KotlinPluginWrapper
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
-import java.util.concurrent.atomic.AtomicReference
 
 object DetektJvmSpec : Spek({
     describe("When applying detekt in a JVM project") {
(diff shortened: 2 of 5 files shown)


cortinico:
Should we hold on till we release 1.19.0 for this one?

BraisGabin:
@3flex #4128 should solve the reason for what you disabled the warning as errors, right?

3flex:
> #4128 should solve the reason for what you disabled the warning as errors, right?

We're keeping the current classloader setup as the default mode, so if we need API version 1.4 today for that, we'll still need it (we should actually use API version 1.3 for compatibility with Gradle 6.x but that's another issue).

BraisGabin (approved):

build-logic/src/main/kotlin/module.gradle.kts
```diff
@@ -83,19 +83,12 @@ configurations.create("coverageDataElements") {
...
+        languageVersion = "1.6"
         apiVersion = "1.4"
         freeCompilerArgs = listOf(
             "-progressive",
             "-Xopt-in=kotlin.RequiresOptIn"
         )
-        // Usage: <code>./gradlew build -PwarningsAsErrors=true</code>.
-        // Note: currently there are warnings for detekt-gradle-plugin that seemingly can't be fixed
-        //       until Gradle releases an update (https://github.com/gradle/gradle/issues/16345)
-        allWarningsAsErrors = when (project.name) {
```
eygraber, line 95: Should I close https://github.com/detekt/detekt/pull/3835/files

eygraber: Maybe relevant - https://blog.mbonnin.net/use-kotlin-15-in-your-gradle-plugins

cortinico: I'm not sure the `allWarningsAsErrors` should be removed for the Gradle plugin

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
```diff
@@ -137,3 +146,28 @@ object DetektJvmSpec : Spek({
...
+
+internal class FakeBuildEventsListenerRegistry : BuildEventsListenerRegistry {
+    override fun onTaskCompletion(provider: Provider<out OperationCompletionListener>) {
+        // Unit
+    }
+}
+
+// https://github.com/gradle/gradle/issues/16774
+@Suppress("TooGenericExceptionThrown")
+private fun Project.addFakeService() {
```
cortinico, line 158: :/ That's black magic to an external reader.
Thanks for adding the link to https://github.com/gradle/gradle/issues/16774, don't you mind adding also a couple of lines of comments.

build-logic/src/main/kotlin/module.gradle.kts
eygraber: Currently it's only off for the Gradle plugin because of the mismatched version warning. 

This change turns it off for all modules (I'm not sure why).

I believe the article I linked above provides a workaround for the mismatched version issue.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
3flex: It's black magic to me too, a copy paste from the Google issue tracker... I can link directly to that but otherwise not sure what else to add.

The root issue fix is slated for Gradle 7.4 so we can wait for that instead of adding this workaround, but that will hold things up a bit.

build-logic/src/main/kotlin/module.gradle.kts
3flex: > This change turns it off for all modules (I'm not sure why).

Because API version 1.4 is deprecated in Kotlin 1.6. All modules currently use API version 1.4, generating a deprecation warning. If warnings as errors is enabled the build then fails, so it has to be disabled.

3flex: I'll just add that if API version used for all modules needs to change, or we make changes to the Gradle plugin build config, we should do that in another PR first. But as it stands the warning needs to be disabled to upgrade to 1.6.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
cortinico: > It's black magic to me too, a copy paste from the Google issue tracker... I can link directly to that but otherwise not sure what else to add.

Yup let's do that. Any kind of context is gold for future reference 👍

cortinico (approved):
Can we land this?

chao2zhang:
Compose just upgraded the kotlin dependency to 1.6.10 - https://developer.android.com/jetpack/androidx/releases/compose-kotlin. So it would be good for us to keep up the release cycle.

BraisGabin:
I can't build the project locally after this change. I get this error:

```sh
./gradlew detektMain
Type-safe dependency accessors is an incubating feature.
Type-safe project accessors is an incubating feature.

> Task :build-logic:jar
:jar: A valid plugin descriptor was found for releasing.properties but the implementation class ReleasingPlugin was not found in the jar.
:jar: A valid plugin descriptor was found for packaging.properties but the implementation class PackagingPlugin was not found in the jar.
:jar: A valid plugin descriptor was found for module.properties but the implementation class ModulePlugin was not found in the jar.

* Where:
Build file '/Users/brais/projects/detekt/build.gradle.kts' line: 4

* What went wrong:
An exception occurred applying plugin request [id: 'releasing']
> Could not find implementation class 'ReleasingPlugin' for plugin 'releasing' specified in jar:file:/Users/brais/.gradle/caches/jars-9/ef0048994d61d2bf5b1eedae8d931857/build-logic.jar!/META-INF/gradle-plugins/releasing.properties.
```

3flex:
Try clearing `/Users/brais/.gradle/caches`

It might be some local problem, I didn't see that issue...

PR #4280, opened 2021-11-16
UnderscoresInNumericLiterals: Allow numbers with non standard groupings

### LINKED ISSUE #4276: UnderscoresInNumericLiterals can't handle non-standard groupings
## Expected Behavior
In my codebase, we store numbers as cents.

As such, we need to be able to write values as `50_000_00` for $50,000.00

The current check does not permit this.

I've seen a couple other cases where it's not a strict grouping of 3, but this is, by far, the biggest annoyance for us right now.

## Observed Behavior
UnderscoresInNumericLiterals check fails.

## Steps to Reproduce
`const val FOO = 50_000_00`

## Your Environment

marschwar:
fixes #4276

I am somewhat unsure how to handle this. In my opinion this rule should have never reported a violation for groups other than three but changing this now will is a breaking change.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiterals.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiterals.kt
@@ -37,7 +37,7 @@ class UnderscoresInNumericLiterals(config: Config = Config.empty) : Rule(config)
         javaClass.simpleName,
         Severity.Style,
         "Report missing or invalid underscores in base 10 numbers. Numeric literals " +
-            "should be underscore separated to increase readability. ",
+            "should be underscore separated to increase readability.",
         Debt.FIVE_MINS
     )
 
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiterals.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiterals.kt
@@ -20,8 +20,7 @@ import java.util.Locale
 
 /**
  * This rule detects and reports long base 10 numbers which should be separated with underscores
- * for readability. Underscores that do not make groups of 3 digits are also reported even if their length is
- * under the configured `acceptableLength`. For `Serializable` classes or objects, the field `serialVersionUID` is
+ * for readability. For `Serializable` classes or objects, the field `serialVersionUID` is
  * explicitly ignored. For floats and doubles, anything to the right of the decimal point is ignored.
  *
  * <noncompliant>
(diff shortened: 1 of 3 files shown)


BraisGabin:
What do you think about adding a flag `forceGroupsOfThree` (we need a better name) that forces the old behaviour? Disabled the rule will allow any type of `_`. And I think that it should be enabled by default.

scottkennedy:
I like that, as it keeps the default behavior the same.

marschwar:
I don't love it but for backward compatibility this is most likely our best choice.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiterals.kt
```diff
@@ -38,8 +37,7 @@ class UnderscoresInNumericLiterals(config: Config = Config.empty) : Rule(config)
         javaClass.simpleName,
         Severity.Style,
         "Report missing or invalid underscores in base 10 numbers. Numeric literals " +
-            "should be underscore separated to increase readability. Underscores that do not make groups of " +
-            "3 digits are also reported.",
+            "should be underscore separated to increase readability. ",
```
BraisGabin, line 40:
```suggestion
            "should be underscore separated to increase readability.",
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiterals.kt
```diff
@@ -107,8 +111,14 @@ class UnderscoresInNumericLiterals(config: Config = Config.empty) : Rule(config)
...
 
+    private fun doReport(expression: KtConstantExpression, message: String) {
+        report(CodeSmell(issue, Entity.from(expression), message))
+    }
+
+    private fun String.hasNonStandardGrouping(): Boolean = contains('_') && !matches(HAS_ONLY_STANDARD_GROUPING)
+
     companion object {
-        private val UNDERSCORE_NUMBER_REGEX = Regex("[0-9]{1,3}(_[0-9]{3})*")
+        private val HAS_ONLY_STANDARD_GROUPING = """\d{1,3}(?:_\d{3})*""".toRegex()
```
BraisGabin, line 121: What does the elvis operator here?

BraisGabin (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnderscoresInNumericLiterals.kt
marschwar: This only marks a non-capture group. I could remove it, if it is confusing.

> If you do not need the group to capture its match, you can optimize this regular expression into Set(?:Value)?. The question mark and the colon after the opening parenthesis are the syntax that creates a non-capturing group.

https://www.regular-expressions.info/brackets.html

PR #4286, opened 2021-11-17
Remove old false-positive from baseline

BraisGabin:
This false positive was fixed at #4249 so we can remove this line from the baseline.


BraisGabin:
Ok... It seems that the fix is not applied to our own codebase yet. I'll rebase this in the next release.

BraisGabin:
Blocked by #4242

chao2zhang (approved):

PR #4307, opened 2021-11-19
Add functional test for type resolution for JVM

3flex:
This is a simple functional test for type resolution on JVM, checking that both Kotlin & Java types are being checked.

I have used the complete file structure instead of using the helpers in the test-fixtures module. Although this test is simple, when we start adding functional tests for more complex project structures, building it out from code like the existing TestKit tests is (IMHO) going to get out of hand. I always get a bit lost when reviewing the existing tests due to the heavy use of parameters & functions to build out the build file & modules.

I intend to add other tests (Android, multiplatform) using the same structure as used here, assuming agreement on the approach.


BraisGabin (approved):

BraisGabin:
I agree with you that maybe, for more complex cases this could explote. But have at least some tests like this would be great.

PR #4315, opened 2021-11-20
Allow plugins to contribute to the default configuration

BraisGabin:
We should allow that the detekt plugins could contribute with the configuration of it's rules ([Slack comment](https://kotlinlang.slack.com/archives/C88E12QH4/p1637323166041900) and #3242. There are two places where we should allow this: when the core request the default configuration and when the user request to generare the configuration file.

With this PR if a plugin author adds a configuration file in their resources at the path `config/config.yml` detekt would concat it at the end of the configuration file.

This PR doesn't give any support to generate nor keep in sync those configurations files but I think that it should be doable in the future.

Related #3242
Closes #1103

PR #4316, opened 2021-11-20
Simplify YamlConfig

BraisGabin:
`YamlConfig.loadResource(URL)` is not needed as method. We can use `YamlConfig.load(Reader)` and convert the URLs to readers when calling.

This is part of the refactor to implement #4315

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/Configurations.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/Configurations.kt
@@ -2,6 +2,7 @@ package io.gitlab.arturbosch.detekt.core.config
 
 import io.github.detekt.tooling.api.spec.ConfigSpec
 import io.github.detekt.tooling.api.spec.ProcessingSpec
+import io.github.detekt.tooling.internal.openSafeStream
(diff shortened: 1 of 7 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/Configurations.kt
```diff
@@ -29,10 +30,10 @@ internal fun ProcessingSpec.loadConfiguration(): Config = with(configSpec) {
 
 private fun parseResourceConfig(urls: Collection<URL>): Config =
     if (urls.size == 1) {
-        YamlConfig.loadResource(urls.first())
+        YamlConfig.load(urls.first().openSafeStream().reader())
```
chao2zhang, line 33: Do we need to use `Closeable.use {}` to ensure that the input stream is closed when out of scope?

This makes this PR beyond pure refactoring, but I think it could be helpful in general.

BraisGabin: Done. It wasn't 100% needed because `YamlConfig.load` call `use` already. But I agree that it's better to add it.

chao2zhang (approved):

PR #4329, opened 2021-11-23
Avoid false positives in MemberNameEqualsClassName

### LINKED ISSUE #4070: `MemberNameEqualsClassName` false positive for a factory function with implicit return type
The following code:

```kotlin
internal class Foo private constructor() {

    companion object {

        fun foo() = Foo()
    }
}
```

will fail with `MemberNameEqualsClassName` rule (`MemberNameEqualsClassName - [foo] at ...`). This error looks like a false positive, because the documentation says

> Factory functions that create an instance of the class are exempt from this rule.

and the _offending_ function does in fact create an instance of the class

## Your Environment

* Version of detekt used: 1.18

BraisGabin:
If we can't infer if it's a factory or not assume that it's ok. False-negatives are better than false-positives.

This rule shoul be moved to type-solving only but that's a huge change that probably can be done for 2.0

fixes #4070

### DIFF
--- a/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassName.kt
+++ b/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassName.kt
@@ -113,12 +113,13 @@ class MemberNameEqualsClassName(config: Config = Config.empty) : Rule(config) {
                 val refName = (typeReference.typeElement as? KtUserType)?.referencedName ?: typeReference.text
                 refName == klass.name
             }
+            function.bodyExpression is KtBlockExpression -> false
             function.bodyExpression !is KtBlockExpression && bindingContext != BindingContext.EMPTY -> {
                 val functionDescriptor = bindingContext[BindingContext.FUNCTION, function]
                 val classDescriptor = bindingContext[BindingContext.DECLARATION_TO_DESCRIPTOR, klass]
                 functionDescriptor?.returnType?.constructor?.declarationDescriptor == classDescriptor
             }
-            else -> false
+            else -> true // We don't know if it is or not a factory. We assume it is a factory to avoid false-positives
         }
     }
 }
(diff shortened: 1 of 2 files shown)


detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MemberNameEqualsClassName.kt
```diff
@@ -103,22 +103,23 @@ class MemberNameEqualsClassName(config: Config = Config.empty) : Rule(config) {
         return klass.companionObjects
             .asSequence()
             .flatMap { getMisnamedMembers(it, klass.name) }
-            .filterNot { it is KtNamedFunction && isFactoryMethod(it, klass) }
+            .filterNot { it is KtNamedFunction && isFactoryMethod(it, klass) != false }
     }
 
-    private fun isFactoryMethod(function: KtNamedFunction, klass: KtClass): Boolean {
+    private fun isFactoryMethod(function: KtNamedFunction, klass: KtClass): Boolean? {
```
cortinico, line 109: `Boolean` -> `Boolean?` sounds like a code smell to me.
Are we able to invert the logic here and make return `true` if is a Factory and `false` if we either don't know or it's not?

BraisGabin: Why do you think this is a code smell? The point of `null` to have a value for "we don't know" and it seems like a good usage of `null`.

> Are we able to invert the logic here and make return `true` if is a Factory and `false` if we either don't know or it's not?

It should be the other way around: I want to the same if we know that it is a factory or if we don't know. So it should be: `false` if it is not a factory and `true` if we don't know or if it is a factory. And return `true` if "we don't know" seems odd. For that reason I introduced the `null`. I can change it for `true` if you think it is better. This is a private function so the correction of what it returns is not that important.

cortinico: > Why do you think this is a code smell?

I think it's maybe more a matter of taste, but I'm in general not a big fan of `Boolean?`. It forces you to use `== false/true` in conditions and is not immediately obvious for the reader why the logic is ternary rather than binary.

> It should be the other way around

Yup you're right. My point was anyway to restrict to be a `Boolean` and treat the absence of value as one of the two cases.

BraisGabin: > I think it's maybe more a matter of taste, but I'm in general not a big fan of `Boolean?`. It forces you to use `== false/true` in conditions and is not immediately obvious for the reader why the logic is ternary rather than binary.

Agree. I changed the code for a `Boolean` and I added a comment in the code to make clear that we don't know but we are assuming that it's a factory to avoid false-positives.

cortinico (approved):

Issue #4330, opened 2021-11-24
Don't trigger "MayBeConst" rule on "actual val" in KMP projects.


gtcompscientist:
## Expected Behavior of the rule
`MayBeConst` rule unexpectedly flags Kotlin Multiplatform files using the `expect`/`actual` pattern at the `actual` site.

Example failure:
File 1:
```
expect val stringVal: String
```
File 2:
```
actual val stringVal: String = "abc123" // detekt says "MayBeConst", but it cannot.
```

## Context

Want to be able to use this rule on KMP projects without turning this rule off, or requiring people suppress it everywhere, as there are still lots of valid places to trigger this rule.

Reference for expect/actual: https://kotlinlang.org/docs/mpp-connect-to-apis.html

PR #4331, opened 2021-11-24
Add helper to check if a KtProperty is an "actual"

gtcompscientist:
Small change to ignore said `actual val` properties when checking `MayBeConst` rule.

Fixes Issue #4330

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MayBeConst.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MayBeConst.kt
@@ -57,6 +57,8 @@ class MayBeConst(config: Config = Config.empty) : Rule(config) {
 
     private val topLevelConstants = HashSet<String?>()
     private val companionObjectConstants = HashSet<String?>()
+    private val KtProperty.isActual
+        get() = hasModifier(KtTokens.ACTUAL_KEYWORD)
 
     override fun visitKtFile(file: KtFile) {
         topLevelConstants.clear()
@@ -110,6 +112,7 @@ class MayBeConst(config: Config = Config.empty) : Rule(config) {
     private fun KtProperty.cannotBeConstant(): Boolean {
         return isLocal ||
             isVar ||
+            isActual ||
             getter != null ||
             isConstant() ||
             isOverride()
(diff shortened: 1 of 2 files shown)


BraisGabin (approved):
Thanks for the report and the fix :)

schalkms (approved):

schalkms:
One thing needs to be changed in order to successfully complete all checks

```
Caused by:
javax.script.ScriptException: Error: error: the feature "multi platform projects" is experimental and should be enabled explicitly
actual val arr = "abc123" 
```

gtcompscientist:
@schalkms I'm happy to fix it either way, what would your preference be?

BraisGabin:
I think that `compileAndLint` would be better.

gtcompscientist:
@BraisGabin and @schalkms Locally I don't get this failure and it's making figuring out if I have a solution difficult. Do you know of the correct way to enable this on the script engine?

schalkms:
@gtcompscientist Please take a look at the following variable. You can set it statically to `true` for testing purposes. 

https://github.com/detekt/detekt/blob/26fd459690635d86f597d9599ad07d852f8db5fe/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt#L17-L18

BraisGabin:
> @BraisGabin and @schalkms Locally I don't get this failure and it's making figuring out if I have a solution difficult. Do you know of the correct way to enable this on the script engine?

You should run the tests like this: `. /gradlew test -Pcompile-test-snippets=true`

schalkms:
> You should run the tests like this: . /gradlew test -Pcompile-test-snippets=true

Thanks @BraisGabin!
Both options are valid. It depends whether you want to execute all tests or just a single one with the installed Spek plugin in IntelliJ.

gtcompscientist:
> Thanks @BraisGabin! Both options are valid. It depends whether you want to execute all tests or just a single one with the installed Spek plugin in IntelliJ.

Unfortunately I'm struggling more with how to enable the multiplatform compilation in the script engine. I can't find any good references online and enabling it as I would for a regular project isn't correct. If there isn't a clear way to do this it may be worth just linting this for now.

schalkms:
I assume you tried to enable MPP in Gradle and then rerun the test, did you?
If yes, I think JSR223 doesn't support this feature yet.
The usage of `lint()` with a comment is fine then.

> Unfortunately I'm struggling more with how to enable the multiplatform compilation in the script engine. I can't find any good references online and enabling it as I would for a regular project isn't correct. If there isn't a clear way to do this it may be worth just linting this for now.

gtcompscientist:
I've updated the PR per the last comment, and everything is ✅ . Good to merge?

schalkms:
Thank you for the comment!
✅

PR #4334, opened 2021-11-25
Enable ForbiddenMethodCall

BraisGabin:
This PR enables `ForbiddenMethodCall` by default and then it adds `Url.openStream` and `Class.getResourceAsStream` to the list.

We need to add support to a description for each forbidden method. If someone in one year find this will now know about the existence of the "safe" version.

---

I'm creating a new module `detekt-utils`. The problem is that `detekt-tooling` are toolings related with the compiler and I couldn't add it anywhere becuase I was introducting circular dependenciees. This `detekt-utils` is a more basic module. Right now it only contains one file to avoid those extrange behaviours when using `Url.openStream()` and `Class.openResourceAsStream()`.

I'm not a big fan to create another module but I don't know how to fix this otherwise.

### DIFF
--- a/detekt-api/build.gradle.kts
+++ b/detekt-api/build.gradle.kts
@@ -10,6 +10,7 @@ plugins {
 dependencies {
     api(libs.kotlin.compilerEmbeddable)
     api(projects.detektPsiUtils)
+    implementation(projects.detektUtils)
 
     testImplementation(projects.detektTest)
     testImplementation(libs.bundles.testImplementation)
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -164,6 +164,11 @@ style:
       - '@author'
       - '@requiresTypeResolution'
     excludes: ['**/detekt-rules-style/**/ForbiddenComment.kt']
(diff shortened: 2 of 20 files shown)


detekt-api/build.gradle.kts
```diff
@@ -10,6 +10,7 @@ plugins {
 dependencies {
     api(libs.kotlin.compilerEmbeddable)
     api(projects.detektPsiUtils)
+    implementation(projects.detektUtils)
```
chao2zhang, line 13: Since `detekt-api` is a public module, do you think we should
1) make this an api dependency
2) add binary compatibility validator

chao2zhang (approved):

detekt-api/build.gradle.kts
BraisGabin: This function is not part of our api. No one should use it. I don't care to break the binary compatibility here. If someone uses it is its own problem.

chao2zhang: We can revisit this later if we add more APIs into `detektUtils` in the future.

PR #4343, opened 2021-11-29
Revert "Use major version tag for github-pages-deploy-action"

3flex:
Reverts detekt/detekt#4342

I learned today that using the major version tag (`v4` in this case) relies on the maintainer of the action tagging a release with the major version tag. That's not being done by this action's maintainer, so version `v4` is not being resolved, so GH pages are not being published as the action isn't running. We will have to continue pinning a specific version of this action.


chao2zhang (approved):

BraisGabin:
Should we open an issue to this action requesting our?

PR #4344, opened 2021-11-30
[UseDataClass] Do not report on `inner` classes

Aksi0m:
The `UseDataClass` rule reports to use the `data` keyword on inner classes. Since `data` is incompatible with `inner` this PR fixes the behaviour.


cortinico (approved):

BraisGabin (approved):

BraisGabin:
@cortinico is that milestone correct? Shouldn't it be 1.20.0?

cortinico:
> Shouldn't it be 1.20.0?

Yup my bad. Let me close 1.19.0 👍

PR #4347, opened 2021-11-30
UnusedPrivateMember should not report external classes/interfaces

### LINKED ISSUE #4340: UnusedPrivateMember should ignore `external` classes
The rule or changes you're looking for might have already been suggested!
Please search in the [issues](https://github.com/detekt/detekt/issues) before creating a new one.

The `UnusedPrivateMember` rule should ignore `external` classes by default.

Example code:

```
@JsModule("@bugsnag/js")
@JsNonModule
external class Bugsnag {
    companion object {
        fun start(bugsnagParams: BugsnagParams)
        fun notify(error: Error)
    }
}

external interface BugsnagParams {
    val apiKey: String
    val appVersion: String
}
```

With Detekt 1.19, the actual behavior is that the Bugsnag companion object methods are flagged as unused.  The expected behavior is that this is not flagged.

## Context
In Kotlin/JS, `external` declarations are just a bridge to an existing Javascript API.  The Detekt rule is not actionable for these, and the only option is to suppress them.

cortinico:
Fixes #4340

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
@@ -217,6 +217,7 @@ private class UnusedParameterVisitor(allowedNames: Regex) : UnusedMemberVisitor(
 
     override fun visitClass(klass: KtClass) {
         if (klass.isInterface()) return
+        if (klass.isExternal()) return
 
         super.visitClass(klass)
     }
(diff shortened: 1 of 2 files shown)


BraisGabin:
🤔 I don't get why things that aren't private are flagged by this rule. I'm wondering if there is an underlying bug here. I mean, why this rule mixes external and private?

cortinico:
> 🤔 I don't get why things that aren't private are flagged by this rule. I'm wondering if there is an underlying bug here. I mean, why this rule mixes external and private?

Yup the tests were confusing, I've fixed them. The rule is reporting private members and parameters that are unused and in my tests I accidentally removed the parameters.

BraisGabin:
I should be missing something here... What is`private` in this snipped?

```kotlin
                external class Bugsnag {
                    companion object {
                        fun start(value: Int)
                        fun notify(error: String)
                    }
                }
```

This rule should only flag code that is marked as `private` right? Why was this code triggering the rule?

cortinico:
> I should be missing something here... What is`private` in this snipped?

The confusion is that `UnusedPrivateMember` is doing more than what its name says:
https://github.com/detekt/detekt/blob/5e346458c2f9d2c22e08fd0ee9b10579bff39e13/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt#L54-L56

Ideally it should be split in `UnusedFunction`, `UnusedParameter` and `UnusedPrivateMember`

BraisGabin:
I get it now! Thanks for the explanation. That have sense.

BraisGabin (approved):

Issue #4206, opened 2021-10-30
Missing test for EmptyElseBlock rule


3flex:
## Expected Behavior of the rule
The empty `else` block is reported in tests:
```kotlin
                fun f() {
                    var i = 0
                    if (i == 0) {
                        println(i)
                    } else {
                        
                    }
                }
```

## Context
~~`EmptyIfBlock` is only reporting empty `then` blocks, not empty `else` blocks.~~
There is no test for EmptyElseBlock rule.

severn-everett:
There wasn't a test for ` EmptyElseBlock`, so I created this:
```
class EmptyElseBlockSpec : Spek({

    val subject by memoized { EmptyElseBlock(Config.empty) }

    describe("EmptyElseBlock rule") {
        it("reports empty else block") {
            val code = """
                fun f() {
                    val i = 0
                    if (i == 0) {
                        println(i)
                    } else {
                        
                    }
                }
            """
            Assertions.assertThat(subject.compileAndLint(code)).hasSize(1)
        }
    }
})
```

The test passes. Has this been resolved in the month since this was posted? In any case, it'd be good to have a spec file for `EmptyElseBlock` in the test suite.

3flex:
Whoops. How embarrassing... the rule indeed exists and is enabled.

If you wouldn't mind opening a PR with your test, that would be great. I was reviewing some other code and saw "else" wasn't handled in the EmptyIfBlock rule, not realising there was a separate rule to check the `else` block.

severn-everett:
Well, it might be necessary to put that original tag back on the ticket, as this fails:
```
        it("reports empty else blocks that are punctuated by a semicolon") {
            val code = """
                fun f() {
                    val i = 0
                    if (i == 0) {
                        println(i)
                    } else ;
                }
            """
            Assertions.assertThat(subject.compileAndLint(code)).hasSize(1)
        }
```
Is this correct, or is it desired behavior for the `EmptyElseBlock` rule to not flag this?

3flex:
Probably a matter of opinion. Strictly speaking there's no else block there, so there's no need for this rule to flag it (since it doesn't exist).

I'm personally ok with that behaviour, though your example should probably be picked up by a rule somewhere, just not sure if that should be this rule or a new one.

chao2zhang:
Just to confirm that this issue is NOT urgent for 1.19.1, the only behavior change in the related PR is `else ;`.

PR #4349, opened 2021-12-01
Testing and rule improvement for EmptyElseBlock

severn-everett:
Resolving issue #4206 by creating tests for `EmptyElseBlock` and standardizing the logic for both empty if- and else-block rules.


cortinico (approved):

chao2zhang (approved):
Thanks for fixing!

PR #4352, opened 2021-12-02
Use the new configure config for formatting

### LINKED ISSUE #3242: Implement ConfigPropertiesProvider
## Expected Behavior
If I autogenerate the detekt config file without the formatting plugin, the generated file will not contain any reference about the formatting ruleset.

## Current Behavior
It appears

## Context
This was a comment in #2081 and Artur propose an interesting solution: 

> [...] introduce a new `ConfigPropertiesProvider` interface which allows plugins to generate additional properties when using -gc.

https://github.com/detekt/detekt/issues/2081#issuecomment-568305348

BraisGabin:
Now that we can concat different configuration files this PR uses it on our first-party-plugin `formatting`.

close #3242


chao2zhang (approved):

chao2zhang:
Many codebases I worked with refer to the URL [detekt-core/src/main/resources/default-detekt-config.yml](https://github.com/detekt/detekt/blob/main/detekt-core/src/main/resources/default-detekt-config.yml)  directly in the config file. So this is one of the most notable changes

BraisGabin:
Why do they that instead of using the generate task? It seems odd because we add new rules there that old versions of detekt don't have.

chao2zhang:
Sorry, I should have clarify - The URL was referred as a link in the comment section. Because people need to compare what has changed between the old detekt version and new detekt version.

BraisGabin:
Oh, I see. I use this script https://github.com/detekt/detekt/discussions/3558#discussioncomment-483027

Update the Detekt configuration between versions is really difficult. We should probably look for a way to improve it. Maybe implement a semantic merge but I don't know.

zdenda-online:
Guys this is not well documented at all.
Now detekt.yml does not work with `formatting` and I suppose we should add a new file in resources. Does it mean that for multi-node repo, we should add it to all subprojects? We would like to use just one configuration for the whole repo.

cortinico:
> Now detekt.yml does not work with `formatting` and I suppose we should add a new file in resources

Nope you should not. Please open an issue and explain your setup as you're most likely facing a bug of some sort

zdenda-online:
It just says that formatting is not allowed to be in detekt.yml (in newer version, in older it worked). Do you think that is really the bug? :thinking: I thought that is due to that change.

BraisGabin:
That seems like an issue in your configuration. To be more precise it seems that you are not applying the formatting plugin. And, because you are not applying it `detekt` doesn't know what `formatting:` is so it complains about that key.

chriswininger:
> That seems like an issue in your configuration. To be more precise it seems that you are not applying the formatting plugin. And, because you are not applying it `detekt` doesn't know what `formatting:` is so it complains about that key.

interesting what is the formatting plugin and how do we use it. Also here trying to upgrade from an older version and seeing the same error as above

BraisGabin:
https://detekt.dev//docs/rules/formatting/

You can find information about how to use it there. It's a wrapper around ktlint.

mairi17:
Here's the correct link in case it's useful:
[https://detekt.dev/docs/rules/formatting/](https://detekt.dev/docs/rules/formatting/)

Issue #3353, opened 2021-01-06
Don't null check a mutable property


BraisGabin:
## Expected Behavior of the rule
non compliant:
```kotlin
class A(
    var foo: Asdf?
) {
  fun bar() {
    if (foo != null) {
      foo!!.toString()
      doThings(foo!!.value)
    }
  }
}
```

compliant:
```kotlin
class A(
    var foo: Asdf?
) {
  fun bar() {
    val foo = foo // or use let, or whatever you want
    if (foo != null) {
      foo.toString()
      doThings(foo.value)
    }
  }
}
```

## Context

#3352 is the example that make me think about this rule. This is a pattern that I see often in the my project. Usually this errors happens when you translates code from Java to Kotlin.

Maybe we could check this kind of things too:
non compliant:
```kotlin
class A {
  fun getActivity(): Activity? = // whatever

  fun bar() {
    if (getActivity() != null) {
      getActivity()!!.doSomething()
    }
  }
}
```

compliant:
```kotlin
class A {
  fun getActivity(): Activity? = // whatever

  fun bar() {
    val activity = getActivity() // or ?.let or whatever you want
    if (activity != null) {
      .doSomething()
    }
  }
}
```

This pattern is kind of usual in Android because the framework give us this kind of APIs. But I don't think that this as an Android specific rule.

cortinico:
> non compliant:
> 
> ```kotlin
> class A(
>     var foo: Asdf?
> ) {
>   fun bar() {
>     if (foo != null) {
>       foo!!.toString()
>       doThings(foo!!.value)
>     }
>   }
> }
> ```

I believe the root of this rule should be "don't null check a mutable property". The multiple `!!` (or whatever null checks) are needed as `foo` is a nullable mutable property, and therefore checking if it's not null won't suffice to smart cast it to a `Asdf`.

atulgpt:
Hi, @BraisGabin I think this issue got fixed by the above PR. right?

PR #4353, opened 2021-12-02
[NullCheckOnMutableProperty] Created rule for when a null-check is performed on a mutable property

severn-everett:
This is for issue #3353. Any advice on how to improve this one is welcome, as this was a bit of a stabbing-in-the-dark solution for me.

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutableProperty.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutableProperty.kt
@@ -115,17 +115,18 @@ class NullCheckOnMutableProperty(config: Config) : Rule(config) {
                         it.fqNameOrNull()?.takeIf(mutableProperties::contains)
                     }
                 }
-                // If a candidate mutable property is present, attach the current
-                // if-expression to it and proceed within the if-expression.
+
                 if (candidateFqName != null) {
+                    // If a candidate mutable property is present, attach the current
+                    // if-expression to it and proceed within the if-expression.
                     candidateProperties.getOrPut(candidateFqName) { ArrayDeque() }.add(expression)
+                    super.visitIfExpression(expression)
+                    // Remove the if-expression after having iterated out of its code block.
+                    candidateProperties[candidateFqName]?.pop()
+                    return
                 }
-                super.visitIfExpression(expression)
-                // Remove the if-expression after having iterated out of its code block.
-                if (candidateFqName != null) candidateProperties[candidateFqName]?.pop()
-            } else {
-                super.visitIfExpression(expression)
             }
+            super.visitIfExpression(expression)
         }
 
         override fun visitReferenceExpression(expression: KtReferenceExpression) {
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutableProperty.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutableProperty.kt
@@ -3,12 +3,16 @@ package io.gitlab.arturbosch.detekt.rules.bugs
 import io.gitlab.arturbosch.detekt.api.CodeSmell
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.DetektVisitor
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
 import io.gitlab.arturbosch.detekt.rules.isNonNullCheck
+import org.jetbrains.kotlin.backend.common.peek
+import org.jetbrains.kotlin.backend.common.pop
+import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.psi.KtBinaryExpression
 import org.jetbrains.kotlin.psi.KtConstantExpression
 import org.jetbrains.kotlin.psi.KtFile
@@ -16,7 +20,7 @@ import org.jetbrains.kotlin.psi.KtIfExpression
 import org.jetbrains.kotlin.psi.KtNameReferenceExpression
 import org.jetbrains.kotlin.psi.KtPrimaryConstructor
 import org.jetbrains.kotlin.psi.KtProperty
-import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
+import org.jetbrains.kotlin.psi.KtReferenceExpression
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
 import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
@@ -34,6 +38,16 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
  *     }
  *   }
  * }
+ *
+ * class A(private var a: Int?) {
+ *   inner class B {
+ *     fun foo() {
+ *       if (a != null) {
+ *         println(a)
+ *       }
+ *     }
+ *   }
+ * }
  * </noncompliant>
  *
  * <compliant>
@@ -44,9 +58,7 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
  *     }
  *   }
  * }
- * </compliant>
  *
- * <compliant>
  * class A(private var a: Int?) {
  *   fun foo() {
  *     val a = a
@@ -60,7 +72,6 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
 
 @RequiresTypeResolution
 class NullCheckOnMutableProperty(config: Config) : Rule(config) {
-    private val mutableProperties = mutableMapOf<String?, MutableSet<String>>()
     override val issue = Issue(
         javaClass.simpleName,
         Severity.Defect,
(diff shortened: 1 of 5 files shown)


BraisGabin (commented):
Could you add these tests too:

```kotlin
class A(private var a: Int?) {
    fun foo() {
        if (a != null) {
            println(2 + 2) // this null check should be fine
        } 
    }
}
```

```kotlin
class A(private val a: Int?) {
    class B(private var a: Int?) {
        fun foo() {
            if (a != null) {
                println(2 + a!!) // should report
            } 
        }
    }

    fun foo() {
        if (a != null) {
            println(2 + a!!) // shouldn't report
        } 
    }
}
```

```kotlin
class A(private var a: Int?) {
    inner class B {
        fun foo() {
            if (a != null) {
                println(2 + a!!) // should report
            } 
        }
    }
}
```

I think that you are handling correctly the last two, but just to have tests that ensure it.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutableProperty.kt
```diff
@@ -0,0 +1,136 @@
...
+ * class A(private val a: Int?) {
+ *   fun foo() {
+ *     if (a != null) {
+ *       println(2 + a)
+ *     }
+ *   }
+ * }
+ * </compliant>
+ *
+ * <compliant>
```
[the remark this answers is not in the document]
severn-everett: The code wouldn't pass the first test case proposed here, as doing so would mean the rule would have to detect both a null-check for the mutable property and a double-bang operation on the variable. This, essentially, would turn the rule into a sub-rule for `UnsafeCallOnNullableType`: anything that `NullCheckOnMutableProperty` would pick up would be picked up by that rule as well. It's not a perfectly symmetrical relationship, but I'm not sure that the use case wherein a developer wants to be warned about double-bang operations only if they're being conducted after a null-check is that prevalent.

BraisGabin:
> The code wouldn't pass the first test case proposed here, as doing so would mean the rule would have to detect both a null-check for the mutable property and a double-bang operation on the variable. This, essentially, would turn the rule into a sub-rule for `UnsafeCallOnNullableType`: anything that `NullCheckOnMutableProperty` would pick up would be picked up by that rule as well. It's not a perfectly symmetrical relationship, but I'm not sure that the use case wherein a developer wants to be warned about double-bang operations only if they're being conducted after a null-check is that prevalent.

I'm going to give more examples because I think that it will be more clear:

```kotlin
class A(private var a: Int?) {
    fun foo() {
        if (a != null) {
            println(a) // report! This could print an unexpected `null`
        }
    }
}
```

```kotlin
class A(private var a: Int?) {
    fun foo() {
        if (a != null) {
            println(2 + requireNotNull(a)) // report! This could crash
        } 
    }
}
```

```kotlin
class A(private var a: Int?) {
    fun foo() {
        if (a != null) {
            println("a is not null") // this code is completely fine
        }
    }
}
```

This code is not related with the usage of `!!`. We just need to check if the property is used or not inside the `if`. If it's used, we should flag the check. If it is not used, we shouldn't flag it.

---

Other missing test:

```kotlin
class A(private var a: Int?) {
    fun foo() {
        if (a != null && true) {
            println("a is not null") // this code is completely fine
        }
    }
}
```

severn-everett:
Gotcha, I'll see what I can do

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutablePropertySpec.kt
```diff
@@ -0,0 +1,277 @@
...
+                    private fun otherFoo(): Int? {
+                        return null
+                    }
+                    fun foo() {
+                        if (otherFoo() != null) {
+                            println(2 + otherFoo()!!)
+                        } 
+                    }
+                }
+                """
```
BraisGabin, line 273: It would be nice to have another rule to check something like this. But it would be difficult and with lots of false positives.

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutablePropertySpec.kt
```diff
@@ -0,0 +1,277 @@
...
+            Assertions.assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
+        }
+
+        it("should not report a null-check on a non-mutable class property") {
+            val code = """
+                class A {
+                    private val a: Int? = 5
```
BraisGabin, line 144:
```kotlin
private val a: Int?
  get() = 5
```

Maybe this could be flagged too.

BraisGabin (approved):
🎉🎉🎉 Great job! I *really* like this rule. This will flag a lot of difficult to spot issues. 👏👏👏

Do you think this is ready to be merged?

severn-everett:
The code should be good to go now. The other two points could be made into other issues/tasks to be tackled later.

BraisGabin:
Ok, I'll wait a bit to get the review form other maintainer.

I think that this case should be handled:

```kotlin
private val a: Int?
  get() = 5
```

But I agree that it could be handled in other PR. Can you do it? Or should we open an issue so we don't forget about it?

severn-everett:
I'd say that it's a different issue, as it wouldn't be "error-prone": the value is going to be non-null, no matter how many times it's checked. Maybe this would be a style issue, e.g. some rule like `UnnecessaryNullable`?

BraisGabin:
Sorry, I wasn't clear with my example. It was far too simplified. I mean something like this:

```
private val a: Int?
  get() = foo()
```

`a` is `val`. But it's value could change between calls so it is a mutable property.

BraisGabin (approved):

BraisGabin:
Why did you closed and removed it?

severn-everett:
My apologies, I thought it was merged - thanks for the heads-up!

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutableProperty.kt
```diff
@@ -0,0 +1,176 @@
...
+ *   fun foo() {
+ *     if (a != null) {
+ *       println(2 + a)
+ *     }
+ *   }
+ * }
+ *
+ * class A(private var a: Int?) {
+ *   fun foo() {
+ *     val a = a
```
cortinico, line 65: That's not the best snippet to put in the compliant block. Is doing name shadowing and should not be suggested.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutableProperty.kt
```diff
@@ -0,0 +1,176 @@
...
+ *     if (a != null) {
+ *       println(2 + a)
+ *     }
+ *   }
+ * }
+ * </compliant>
+ */
+
```
cortinico, line 73: Extra white line?

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutableProperty.kt
```diff
@@ -0,0 +1,176 @@
...
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Defect,
+        "Checking nullability on a mutable property is not useful because the " +
+            "property may be set to null afterwards.",
+        Debt.TEN_MINS
+    )
+
+    override fun visitKtFile(file: KtFile) {
+        if (bindingContext == BindingContext.EMPTY) return
```
cortinico, line 85: You're missing a `super.visitKtFile(file)` here

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutableProperty.kt
```diff
@@ -0,0 +1,176 @@
...
+                        ?.let {
+                            it.fqNameOrNull()?.takeIf(mutableProperties::contains)
+                        }
+                }?.let { candidateFqName ->
+                    // A candidate mutable property is present, so attach the current
+                    // if-expression to it in the property candidates map.
+                    candidateProperties.getOrPut(candidateFqName) { ArrayDeque() }.apply { add(expression) }
+                }
+            }
+            super.visitIfExpression(expression)
```
cortinico, line 132: Commenting here but valid for the whole `NullCheckVisitor`:

Is there a reason why you're calling `super.` methods at arbitrary locations inside the function blocks? Ideally you should always call the super method as first statement and then add your logic.

In this line you're calling `super.visitIfExpression(expression)` in the middle of the block.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutableProperty.kt
```diff
@@ -0,0 +1,176 @@
...
+                        (!expressionParent.isNonNullCheck() && !expressionParent.isNullCheck())
+                    ) {
+                        // If there's an if-expression attached to the candidate property, a null-checked
+                        // mutable property is being referenced.
+                        candidateProperties[fqName]?.peek()?.let { ifExpression ->
+                            report(
+                                CodeSmell(
+                                    issue,
+                                    Entity.from(ifExpression),
+                                    "Null-check is being called on a mutable property."
```
cortinico, line 156: nit: if you could customize the message with the name of the property would be 👌

cortinico (approved):
Great work :)

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutableProperty.kt
severn-everett: The line needs to be in the middle of the block in order to analyze whether the descendant expressions use any candidate properties identified in the if-expression, with the `visitIfExpression()` removing any if-statements added to `candidateProperties` afterwards. I've added a comment that explains as much in the code.

cortinico:
Are we good to merge this?

Issue #4356, opened 2021-12-03
ignoreAnnotated not respected on UnnecessaryAbstractClass


ZacSweers:
## Expected Behavior

After migrating to `ignoreAnnotated` per deprecation warning messages, it's no longer respecting the annotations we specify

Detekt 1.18.0
```yml
UnnecessaryAbstractClass:
  active: true
  excludeAnnotatedClasses: ['dagger.Module']
```

Warns

```
Property 'style>UnnecessaryAbstractClass>excludeAnnotatedClasses' is deprecated. Use `ignoreAnnotated` instead.
```

Detekt 1.19.0
```yml
UnnecessaryAbstractClass:
  active: true
  ignoreAnnotated: ['dagger.Module']
```

## Observed Behavior

It incorrectly warns about classes annotated with `dagger.Module`

## Steps to Reproduce

Here's an example module

```kotlin
@Module
abstract class CallsBaseModule {

  @Binds
  abstract fun CallPresenterImpl.provideCallPresenter(): CallContract.CallPresenter

  @Binds
  abstract fun CallServiceProviderImpl.bindCallServiceProvider(): CallServiceProvider
}
```

## Context

It is a regression

## Your Environment

* Version of detekt used: 1.19.0
* Version of Gradle used (if applicable): N/A
* Gradle scan link (add `--scan` option when running the gradle task): N/A
* Operating System and version: macOS 11
* Link to your project (if it's a public repository): N/A

severn-everett:
I was unable to reproduce the issue in a POC project with the following `CallBaseModule.kt` file:
```
package com.severett.kotlinpoc.model

import dagger.Module

@Module
abstract class CallsBaseModule
```
This used Detekt 1.19.0 with the following configuration:
```
  UnnecessaryAbstractClass:
    active: true
    ignoreAnnotated: ['dagger.Module']
```
In addition, the project uses Kotlin 1.6.0 and Gradle 7.3.1.

Would you be able to give a more detailed description of the code base layout and the Detekt configuration file in full?

leinardi:
Hi @severn-everett, my project is also affected by this issue. You should be able to reproduce it checking out the branch `detekt-1.19.0`: https://github.com/leinardi/Forlago/tree/detekt-1.19.0

if you run `./gradlew detekt` you should get this output:
```
> Task :modules:core-navigation:detekt FAILED
style - 5min debt
[...]

Issue #4355, opened 2021-12-03
LongParameterList no longer recognizes ignored annotations in 1.19.0


ZacSweers:
## Expected Behavior

Prior to 1.19.0, this configuration worked to prevent `LongParameterList` warnings on constructors with a few "known" injection annotations from dagger.

```yml
LongParameterList:
  active: true
  functionThreshold: 7
  constructorThreshold: 7
  ignoreDefaultParameters: false
  ignoreDataClasses: true
  ignoreAnnotated: ['javax.inject.Inject', 'dagger.Provides', 'dagger.assisted.AssistedInject']
```

## Observed Behavior

In 1.19.0, these seem to be ignored now

## Steps to Reproduce

Here are some snippets that repro for us

```kotlin
class ActivityFeedPresenter @Inject constructor(
  private val activityApi: ActivityApi,
  private val userRepository: UserRepository,
  private val botsDataProvider: BotsDataProvider,
  private val mpdmDisplayNameHelper: MpdmDisplayNameHelper,
  private val prefsManager: PrefsManager,
  private val conversationRepo: ConversationRepository,
  private val lockedThreadTransformer: ActivityFeedLockedThreadTransformer,
  private val mentionItemFactoryLazy: Lazy<MentionItemFactory>,
) : ActivityFeedContract.Presenter, LoadingStateProvider {

class AppDialogFragment @AssistedInject constructor(
  private val appDialogsRepository: AppDialogsRepository,
  private val presenter: AppDialogPresenter?,
  private val keyboardHelper: KeyboardHelper,
  private val snackbarHelper: SnackbarHelper,
  private val appBuildConfig: AppBuildConfig,
  private val appDialogMenuBinder: AppDialogMenuBinder,
  private val appDialogMenuViewFactory: AppDialogMenuView.Factory,
  private val appDialogSelectViewFactory: AppDialogSelectView.Factory,
  private val appDialogTextViewFactory: AppDialogTextView.Factory,
  private val platformLogger: PlatformLogger,
  @ForScope(UserScope::class) private val fragmentNavRegistrar: FragmentNavRegistrar
) : ViewBindingFragment(),
  AppDialogContract.View,
  MenuSelectionListener,
  BackPressedListener {
  
class AppActionDelegateImpl @Inject constructor(
  private val appActionsRepository: Lazy<AppActionsRepository>,
  private val attachmentActionHelperLazy: Lazy<AttachmentActionHelper>,
[...]

PR #4358, opened 2021-12-05
Integrated former method for looking up annotation names in AnnotationSuppressor

severn-everett:
Addresses #4355 and #4356

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
@@ -120,7 +120,7 @@ internal class Analyzer(
         fun executeRules(rules: List<BaseRule>) {
             for (rule in rules) {
                 rule.visitFile(file, bindingContext, compilerResources)
-                for (finding in filterSuppressedFindings(rule, bindingContext)) {
+                for (finding in filterSuppressedFindings(file, rule, bindingContext)) {
                     val mappedRuleSet = checkNotNull(ruleIdsToRuleSetIds[finding.id]) {
                         "Mapping for '${finding.id}' expected."
                     }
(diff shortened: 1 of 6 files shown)


BraisGabin:
Thanks for the PR! I'm working on something similar here #4368. The problem with the former method is that it didn't work for all the cases as I expose in that PR. I want to create a new class `FullQualifiedNameGuesser` that given a file and a name it provides you with all the possible FullQualifiedNames that it could have (usually it only returns one but it could return mutliple ones). I think that it will be ideal to use the same class to implement both use cases: `AnnotationExcluder` and `AnnotationSuppressor`.

BraisGabin:
To make more clear what I'm talking about this is the WIP `FullQualifiedNameGuesser` class and it's integration with `AnnotationExcluder`: https://github.com/detekt/detekt/compare/AnnotationExcluder..FullQualifiedNameGuesser

severn-everett:
Ok, so should I close this PR?

BraisGabin:
No, don't close it. If that could end up merged this PR could be reworked and, if it's not we can merge directly this.

chao2zhang:
#4368 is merged so this PR should be unblocked.

BraisGabin:
I just launched #4518. Once we have that implemented we can rebase this one and work with `AnnotationExcluder` when we don't have the `BindingContext` and the current implementation when we have it.

cortinico:
What's the timeline here? I'd like to kickoff the `1.20.0-RC1` process and this is pretty high prio to land. If we wish we can release RC1 without this for now, and postpone to RC2. Having it in though would be a major win.

BraisGabin:
I think that as soon as #4518 is merged we can fix this.

BraisGabin:
@severn-everett sorry for the delay. Finally the needed code is merged. Do you want to continue with this using `AnnotationExcluder` for the case when we don't have type solving and keeping the current implementation for when we have type solving?

cortinico:
@BraisGabin how do you feel about taking this over if @severn-everett doesn't have bandwidth? I'd like to include this into a RC1 so we can reiterate on this feature for a RC2 if needed.

BraisGabin:
Sure, I can. I'll try to do it this weekend

BraisGabin:
Closing in favor of #4570

PR #4369, opened 2021-12-10
Fix false positive in RethrowCaughtException for try with more than one catch (#4367)

### LINKED ISSUE #4367: RethrowCaughtException should not be triggered when there is more general exception below

Following code should be compliant with RethrowCaughtException rule:

```kotlin
try {
    // Do something
} catch (e: CancellationException) {
    // Just throw cancellation
    throw e
} catch (e: Exception) {
    // Log everything except cancellation
    logException(e)
}
```

## Context

In above case I want to log all exceptions except for the Cancellation exception (which I just want to propagate). That's why CancellationException has its own catch that gets activated before the general one, which is then never reached. This cause clause is not useless because it is used to filter more general cause clauses that come after it.

Above snippet is just a more cleaner way to write following code, which is compliant:

```kotlin
try {
    // Do something
} catch (e: Exception) {
    if (e is CancellationException) {
        throw e
    } else {
        logException(e)
    }
}
```

strkkk:
Closes #4367 

1. Fixes false positive
2. One more test input is added (when same exception is thrown but with `this` qualifier)

### DIFF
--- a/detekt-rules-exceptions/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/RethrowCaughtExceptionSpec.kt
+++ b/detekt-rules-exceptions/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/RethrowCaughtExceptionSpec.kt
@@ -22,6 +22,21 @@ class RethrowCaughtExceptionSpec : Spek({
             assertThat(subject.compileAndLint(code)).hasSize(1)
         }
 
+        it("does not report when the other exception is rethrown with same name") {
+            val code = """
+                class A {
+                    private lateinit var e: Exception
+                    fun f() {
+                        try {
+                        } catch (e: IllegalStateException) {
+                            throw this.e
+                        }
+                    }
+                }
+            """
+            assertThat(subject.compileAndLint(code)).isEmpty()
+        }
+
         it("reports when the same exception succeeded by dead code is rethrown") {
             val code = """
                 fun f() {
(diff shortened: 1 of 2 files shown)


detekt-rules-exceptions/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/RethrowCaughtExceptionSpec.kt
```diff
@@ -108,5 +123,34 @@ class RethrowCaughtExceptionSpec : Spek({
...
+            val code = """
+                fun f() {
+                    try {
+                    } catch (e: CancellationException) {
+                        throw e
+                    } catch (e: Exception) {
+                        print(e)
+                    }
+                }
+            """
```
BraisGabin, line 137:
```kotlin
                fun f() {
                    try {
                    } catch (e: CancellationException) {
                        print(e)
                    } catch (e: Exception) {
                        throw e
                    }
                }
```

Do we have a test for this case? To ensure that we flag it?

BraisGabin (commented):
👏 really good first contribution :) Thanks!

detekt-rules-exceptions/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/RethrowCaughtExceptionSpec.kt
strkkk: I believe it is same thing, regardless order of catch blocks.

Anyway, I updated this input with reversed order (handling, than throwing)

BraisGabin: No, the order matter here. In this case the second catch is completely useless. In the previous case it wasn't.

strkkk: yes, you are right. Anyway, both cases produce no violations now. Can you start PR workflows? it needs mantainer rights to run

BraisGabin: But the second one should be a violation. You are catching the exception just to rethrow it. You can remove that `catch`.

cortinico (approved):
LGTM 👍

BraisGabin (changes requested):
Adding request for changes becuase right now it introduces a false-negative as commented in this thread: https://github.com/detekt/detekt/pull/4369/files#r767144893

detekt-rules-exceptions/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/RethrowCaughtExceptionSpec.kt
strkkk: @BraisGabin fixed

BraisGabin (approved):

BraisGabin:
Thank you so much! I'll merge it as soon as CI passes :). Greate job.

BraisGabin:
There is an issue in the code, could you check and fix it? https://github.com/detekt/detekt/runs/4594574549?check_suite_focus=true#step:4:356

strkkk:
@BraisGabin please trigger ci

Issue #4363, opened 2021-12-09
OptionalUnit Rule on suspendCoroutine context


panoramix360:
## Expected Behavior
Don't throw an issue for OptionalUnit when used with suspendCoroutines that return nothing

In the following code, I can't remove Unit from the return because suspendCoroutine needs to known the function return type.
```
override suspend fun doAction(): Unit = suspendCoroutine { cont ->
        Integration.signOut() {
            it.fold(
                {
                    cont.resume(Unit)
                },
                { error ->
                    cont.resumeWithException(error as Exception)
                }
            )
        }
    }
```

## Observed Behavior
Detekt is complaining with a OptionalUnit rule.

## Steps to Reproduce
- Create a function that uses suspendCoroutine and returns nothing (Unit)

## Context
Detekt is not passing in this particular case in my project.

## Your Environment
* Version of detekt used: 1.18.1
* Version of Gradle used (if applicable): 7.0.3
* Operating System and version: Ubuntu 18.04

severn-everett:
A workaround for this would be to specify the type in the generic bracket, e.g.
```
override suspend fun doAction() = suspendCoroutine<Unit> { cont ->
        Integration.signOut() {
            it.fold(
                {
                    cont.resume(Unit)
                },
                { error ->
                    cont.resumeWithException(error as Exception)
                }
            )
        }
    }
```

panoramix360:
Hi @severn-everett, it worked like you suggested!

But I think that it's nice to address this on the other way around too.

Thank you for helping and opening a PR to adjust this.

BraisGabin:
Thanks for the report @panoramix360. Closing because @severn-everett fixed it in #4371

PR #4371, opened 2021-12-12
Allow a function to declare a Unit return type when it uses a generic function initializer

severn-everett:
This addresses issue #4363.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/OptionalUnit.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/OptionalUnit.kt
@@ -15,6 +15,7 @@ import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtIfExpression
 import org.jetbrains.kotlin.psi.KtNameReferenceExpression
 import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.KtTypeReference
 import org.jetbrains.kotlin.psi.KtWhenExpression
 import org.jetbrains.kotlin.psi.psiUtil.siblings
 import org.jetbrains.kotlin.resolve.BindingContext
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/OptionalUnit.kt
```diff
@@ -9,6 +9,7 @@ import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.rules.isOverride
 import org.jetbrains.kotlin.cfg.WhenChecker
+import org.jetbrains.kotlin.js.translate.callTranslator.getReturnType
```
severn-everett, line 12: No idea why this is in a JavaScript package, but it appears to work in JVM code.

severn-everett:
@BraisGabin I'm not sure what I'm supposed to do here. CodeCov is requiring that I write a test where one of the two lines below returns a null value:
```
        val isGenericType = getResolvedCall(bindingContext)?.getReturnType()?.isTypeParameter() == true
        val isNothingType = getType(bindingContext)?.isNothing() == true
```
Otherwise, the PR is not hitting the target rate of 80% code coverage. The problem is that writing that test would require creating a code snippet where, for example, the function initializer isn't recognized (e.g. `fun foo(): Unit = unknownFun { }`. However, doing _that_ will fail the `compile-test-snippets` step, because as the function initializer isn't recognized, the code wouldn't compile IRL. Any ideas on how to write a credible test here that will satisfy both `codecov/patch` and `Pre Merge Checks / compile-test-snippets`?

BraisGabin (approved):

BraisGabin:
> @BraisGabin I'm not sure what I'm supposed to do here. CodeCov is requiring that I write a test where one of the two lines below returns a null value:
> 
> ```
>         val isGenericType = getResolvedCall(bindingContext)?.getReturnType()?.isTypeParameter() == true
>         val isNothingType = getType(bindingContext)?.isNothing() == true
> ```
> 
> Otherwise, the PR is not hitting the target rate of 80% code coverage. The problem is that writing that test would require creating a code snippet where, for example, the function initializer isn't recognized (e.g. `fun foo(): Unit = unknownFun { }`). However, doing _that_ will fail the `compile-test-snippets` step, because as the function initializer isn't recognized, the code wouldn't compile IRL. Any ideas on how to write a credible test here that will satisfy both `codecov/patch` and `Pre Merge Checks / compile-test-snippets`?

Don't care about that. If the tests really cover the main logic it's enough. If you check you are under the target but even with that you are increasing the coverage.

Issue #4376, opened 2021-12-16
New rule: Nullable val can be non-nullable


BraisGabin:
## Expected Behavior of the rule

Detect variables and properties that are declared as Nullable but they never contain a `null` value.

Non-compilant:
```kotlin
val a: String? = "hello"

val a: String?
   get() = "hello"
```

## Context
The tanslator of code from java to kotlin adds a lot of unnecesary `?`. And when you are refactoring maybe changes a property that could hold `null` to something that couldn't hold `null` but you can forget about redeclaring the type.

I don't know if we could handle `var` too.

After reading the last commits of #4353 I relasied that this rule shouldn't be that difficult to implement.

severn-everett:
I think the case where `var` could be flagged is if the analyzer could prove that any values that it gets assigned are non-null (or at least not nullable). Private `var`s - along with `vars` that have private setters - seen pretty feasible for this; `var`s that are settable outside of their containing object/class might balloon the complexity (and/or computational time) quite a bit.

cortinico:
Closing this as it was shipped in #4379

PR #4379, opened 2021-12-17
Added rule CanBeNonNullableProperty

severn-everett:
This addresses issue #4376

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -143,8 +143,8 @@ potential-bugs:
     active: true
 
 style:
-  CanBeNonNullableProperty:
-    active: false
+  CanBeNonNullable:
+    active: true
   ClassOrdering:
     active: true
   CollapsibleIfStatements:
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -144,7 +144,7 @@ potential-bugs:
 
 style:
   CanBeNonNullableProperty:
-    active: true
+    active: false
   ClassOrdering:
     active: true
   CollapsibleIfStatements:
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -143,6 +143,8 @@ potential-bugs:
     active: true
 
 style:
+  CanBeNonNullableProperty:
+    active: true
   ClassOrdering:
     active: true
   CollapsibleIfStatements:
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableProperty.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableProperty.kt
@@ -8,7 +8,6 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
-import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
 import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
 import io.gitlab.arturbosch.detekt.rules.isOpen
 import org.jetbrains.kotlin.descriptors.PropertyDescriptor
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableProperty.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableProperty.kt
@@ -9,6 +9,7 @@ import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
+import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
 import io.gitlab.arturbosch.detekt.rules.isOpen
 import org.jetbrains.kotlin.descriptors.PropertyDescriptor
 import org.jetbrains.kotlin.lexer.KtTokens
@@ -65,6 +66,7 @@ import org.jetbrains.kotlin.types.isNullable
  * }
  * </compliant>
  */
+@RequiresTypeResolution
 @ActiveByDefault(since = "1.20.0")
 class CanBeNonNullableProperty(config: Config = Config.empty) : Rule(config) {
     override val issue = Issue(
@@ -77,6 +79,7 @@ class CanBeNonNullableProperty(config: Config = Config.empty) : Rule(config) {
     override fun visitKtFile(file: KtFile) {
         if (bindingContext == BindingContext.EMPTY) return
         NonNullableCheckVisitor().visitKtFile(file)
+        super.visitKtFile(file)
     }
 
     private inner class NonNullableCheckVisitor : DetektVisitor() {
@@ -97,7 +100,7 @@ class CanBeNonNullableProperty(config: Config = Config.empty) : Rule(config) {
                     CodeSmell(
                         issue,
                         Entity.from(property),
-                        "A nullable property can be made non-nullable."
+                        "The nullable property '${property.name}' can be made non-nullable."
                     )
                 )
             }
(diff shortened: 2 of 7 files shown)


severn-everett:
@BraisGabin The latest Detekt run on the branch had `CanBeNonNullableProperty` flag this call in `DetektExtension.kt`:
```
    @Deprecated("Use reportsDir which is equivalent", ReplaceWith("reportsDir"))
    val customReportsDir: File?
        get() = reportsDir
```
The problem is that Kotlin determines that `reportsDir` is non-nullable, even though it is Java code that could return a nullable. I'd say to have `CanBeNonNullableProperty` disqualify any property that derives its value from Java source code, but I'm not sure how I'd be able to test that in the Spec files. Any suggestions for how to compile Java code alongside Kotlin code in the tests, or should I just suppress the warning on `customReportsDir`?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableProperty.kt
```diff
@@ -0,0 +1,177 @@
...
+ *     }
+ * }
+ *
+ * class A {
+ *     val a: Int
+ *         get() = 5
+ * }
+ * </compliant>
+ */
+@ActiveByDefault(since = "1.20.0")
```
cortinico, line 68: You need to add a `@RequiresTypeResolution` annotation here

cortinico: Also I'm unsure we want to have this enabled by default

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableProperty.kt
```diff
@@ -0,0 +1,177 @@
...
+class CanBeNonNullableProperty(config: Config = Config.empty) : Rule(config) {
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Style,
+        "Property can be changed to non-nullable, as it is never set to null.",
```
cortinico, line 73:
```suggestion
        "Property can be changed to non-nullable, as it is never sets to null.",
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableProperty.kt
```diff
@@ -0,0 +1,177 @@
...
+        Debt.TEN_MINS
+    )
+
+    override fun visitKtFile(file: KtFile) {
```
cortinico, line 77: You're missing a `super.visitKtFile(file)` call here

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableProperty.kt
```diff
@@ -0,0 +1,177 @@
...
+            super.visitKtFile(file)
+            // Any candidate properties that were not removed during the inspection
+            // of the Kotlin file were never assigned nullable values in the code,
+            // thus they can be converted to non-nullable.
+            candidateProps.forEach { (_, property) ->
+                report(
+                    CodeSmell(
+                        issue,
+                        Entity.from(property),
+                        "A nullable property can be made non-nullable."
```
cortinico, line 100: nit: You could improve this to mention the name of the property as you have it

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullablePropertySpec.kt
```diff
@@ -0,0 +1,271 @@
...
+                    fun foo(): Int {
+                        val b = a
+                        a = b + 1
+                        fileA = a + 1
+                        val a = null
+                        return b
+                    }
+                }
+                """
+            Assertions.assertThat(subject.compileAndLintWithContext(env, code)).hasSize(5)
```
cortinico, line 60: Can you split into smaller tests?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullablePropertySpec.kt
```diff
@@ -0,0 +1,271 @@
...
+                        5
+                    }
+                    
+                    init {
+                        b = 5
+                        c = cVal
+                    }
+                }
+                """
+            Assertions.assertThat(subject.compileAndLintWithContext(env, code)).hasSize(5)
```
cortinico, line 175: Same as above

cortinico (commented):
> I'd say to have CanBeNonNullableProperty disqualify any property that derives its value from Java source code

+1 on this

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableProperty.kt
severn-everett: The sentence is grammatically correct as it is; `sets` would work only if the word `is` is removed

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableProperty.kt
severn-everett: Why would this be necessary? The `NonNullableCheckVisitor` instance conducts the actual check of the file for the rule, not `CanBeNonNullableProperty` itself.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableProperty.kt
cortinico: > The sentence is grammatically correct as it is; `sets` would work only if the word `is` is removed

Yup my bad, I was reading it too fast 👍

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableProperty.kt
cortinico: > Why would this be necessary?

That's part of the `Rule` contract and in general follows the visitor design pattern implementation. If by any chance we decide to implement a new feature on detekt that relies on the `visitKtFile` and lives inside the `Rule` top level class (or any of its ancestors), this rule will break this behavior.

BraisGabin:
>  but I'm not sure how I'd be able to test that in the Spec files. Any suggestions for how to compile Java code alongside Kotlin code in the tests, or should I just suppress the warning on `customReportsDir`?

@t-kameyama implemented the option to do that in #4203. I think that you can use the same pattern.

config/detekt/detekt.yml
```diff
@@ -143,6 +143,8 @@ potential-bugs:
     active: true
 
 style:
+  CanBeNonNullableProperty:
+    active: false
```
BraisGabin, line 147: I think that it's safe to enable this. Or, if it isn't, remove these lines

BraisGabin (commented):
Run the tests with `-Pcompile-test-snippets=true` in the gradle command. That will make the tests run slower but it will ensure that your snippet codes compile (It is necessary to merge the PR)

This rule is really good! But I'm wondering if this same rule should check for the same pattern in local variables or should we use other rule. I think that this rule is ready to merge now but, if we want to extend it later with the local variables we should change its name before merging it. I mean, it's not needed to implement the local variable thing now but if we want both things in the same ruel it can't be called "property".

severn-everett:
Should be good to go, but something happened on the Windows Java 8 build. It looks like Gradle crashed for whatever reason, not sure if it's got anything to do with the code.

BraisGabin (approved):

BraisGabin (approved):

DHosseiny:
It seems No doc for this in detekt .dev

cortinico:
> It seems No doc for this in detekt.dev

Doc is here: https://detekt.dev/docs/rules/style/#canbenonnullable

Issue #4359, opened 2021-12-06
Add possibility to estimate tech debt in code


vbuberen:
## Expected Behavior
Detekt allows to configure some rules which would help dev teams see estimated cost in either time or money value of the tech debt they have.

## Current Behavior
Currently not implemented in Detekt

## Context
Idea comes from the new feature introduced recently in [Dart Code Metrics](https://dartcodemetrics.dev/).
Here is a brief visual representation of what the feature is about: https://twitter.com/_incendial/status/1467185701569822720?s=20
I believe such metric could be useful for some orgs/teams, so people could have some estimated numbers for their plannings/prioritisation sessions.

BraisGabin:
Right now we generate an stimated time about how many time it will take to solve it. We know that this value is *really* loose. We could add the option to customize the time of each rule. We talked about that some time ago here: https://github.com/detekt/detekt/pull/3339#issuecomment-753401606 Would this be enough?

As a side note. Detekt already allow you to create your own reports: https://detekt.github.io/detekt/extensions.html#custom-reports

vbuberen:
Yes, having an option to customize times for rules might be enough, I believe. Thanks for sharing.
Currently I am not really involved into some Android project where I could try out things and provide feedback you might want to implementation. Opened issue mainly due to seeing such thing for Flutter projects and thinking that it might be a good addition for Android devs as well.

cortinico:
> We could add the option to customize the time of each rule. We talked about that some time ago here: [#3339 (comment)](https://github.com/detekt/detekt/pull/3339#issuecomment-753401606) Would this be enough?

To add to this: I think in general is a good idea. I still have to find a real use case for those numbers though. I see several tools generating those "complexity" reports, but the number are generally really arbitrary and not really meaningful to the developers.

Potentially having those metrics configurable would help to make it more concrete for each codebase.

severn-everett:
Good idea, I'll start taking a look at this shortly.

severn-everett:
I'm not sure if this is going to be possible without introducing breaking changes. I tried implementing this in `Rule`:
```
abstract class Rule(
    override val ruleSetConfig: Config = Config.empty,
    ruleContext: Context = DefaultContext()
) : BaseRule(ruleContext), ConfigAware {

    open val defaultDebt: Debt = Debt.TEN_MINS

    val issueDebt: Debt
        get() = valueOrNull<Int>("issueDebt")?.let { Debt(mins = it) } ?: defaultDebt
```
[...]

PR #4385, opened 2021-12-19
Created new rule class that allows for configurable debt time

severn-everett:
This is a proposal for issue #4359. The advantage of this implementation is that it removes the run-time restriction that the fields of `Rule.issue` are not configurable due to an already-instantiated `Issue` instance being needed for its `id` field to look up configuration values. If some iteration of this class is approved, the next step could be to proceed to replace the rules in the Detekt codebase section by section with the new class, after which the original `Rule` class could be marked as `@Deprecated` so that third-party developers could replace it with the new rule class.

### DIFF
--- a/detekt-api/api/detekt-api.api
+++ b/detekt-api/api/detekt-api.api
@@ -344,6 +344,30 @@ public final class io/gitlab/arturbosch/detekt/api/Issue {
 	public fun toString ()Ljava/lang/String;
 }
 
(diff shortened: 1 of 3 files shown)


BraisGabin:
This is the idea for the rule in 2.0:

https://github.com/detekt/detekt/blob/new-rule/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/v2/Rule.kt

It isn't definitive but we would like to make it an interface and way simpler as you can see.

I don't think that add a new rule type that will be deprecated in 2.0 is a good call. We would force the rule authors to change their rules twice.

severn-everett:
Based on the discussion in the issue, I'm going to close this; some iteration of the code could be incorporated when the 2.0 version of `Rule` is implemented.

Issue #4384, opened 2021-12-19
Formulate rule descriptions consistently


pnq32:
The majority of rule descriptions incorporated into the current version (v1.19.0) seems to be formulated as a sequence of (one or more) short sentences outlining the issue that triggered the rule, potentially extended by a direct recommendation towards the developer. The sentences are usually ended with a period, while inline code is sometimes surrounded by  `` ` `` characters.

Consider the following descriptions as an example:
- **ForbiddenVoid**: `` `Unit` should be used instead of `Void`.``
- **ComplexMethod**: `Prefer splitting up complex methods into smaller, easier to understand methods.`
- **AnnotationOnSeparateLine** (from `detekt-formatting`): `Multiple annotations should be placed on separate lines.`
- **DuplicateCaseInWhenExpression**: `Duplicated case statements in when expression. Both cases should be merged.`
- **InjectDispatcher**: ``Don't hardcode dispatchers when creating new coroutines or calling `withContext`. Use dependency injection for dispatchers to make testing easier.``

However, there are also various rules that deviate from this style as follows:
- One-sentence descriptions are sometimes formatted without a period at the end...
  - **LibraryEntitiesShouldNotBePublic**: `Library class should not be public`
  - **OptionalWhenBraces**: `Optional braces in when expression`
- In descriptions with multiple sentences, the last sentence is sometimes lacking the period...
  - **FunctionOnlyReturningConstant**: `(...) is misleading. Consider declaring a constant instead`
- Sometimes, sentences are separated from each other using newlines:
  - **ImplicitUnitReturnType**: `(...) return type.\nChanging the type (...)`
- Inline code is sometimes surrounded by `'` characters or not surrounded at all:
  - **NonBooleanPropertyPrefixedWithIs**: `Only boolean property names can start with 'is' prefix.`
  - **EqualsAlwaysReturnsTrueOrFalse**: `(...) implementation. For example this == other`
- Especially in `detekt-formatting`, many texts describe what the rule itself does instead of what the underlying issue is:
  - **MaximumLineLength**: `Reports lines with exceeded length`
  - **NoTrailingSpaces**: `Detects trailing spaces`
 
I am not trying to say that the first set of descriptions is "right" and the section set of descriptions is "wrong". In my opinion, however, it would make (console) reports more readable and pleasing to look at if all descriptions were given in a consistent manner. What do you think, would the introduction of a convention that deals with the description format be possible?

BraisGabin:
[...]

Issue #4163, opened 2021-10-05
Some checks have very verbose "names" in GitHub code scanning


TWiStErRob:
This affects multiple rules.

## Expected Behavior of the rule
When looking at https://github.com/org/repo/security/code-scanning I want to see a short quick summary of the issues, not repeated details and default values which I might've overridden anyway. Sadly I noticed that `Issue` doesn't have a title and a description field like Android-lint, so I think what's missing is a consistent scheme of defining check descriptions. What goes into the Issue definition and what goes into the documentation.

## Context
I did as suggested by https://github.com/detekt/detekt/issues/3045#issuecomment-933033224 and I managed to set up Detekt. When I first ran it (with some issues to test), I received a long list with some very verbose titles. When looking at this list I would want to see a quick summary.
![image](https://user-images.githubusercontent.com/2906988/136093861-bdbf73ea-9dee-42df-bafb-95cad5cc029c.png)

Here's one long one when I open it:
![image](https://user-images.githubusercontent.com/2906988/136093685-11018e37-6624-4eca-8f15-696d430a8165.png)

3flex:
Closing, see https://github.com/detekt/detekt/issues/4750#issuecomment-1109627132

TWiStErRob:
I think agree with @BraisGabin that this is a generic issue, because the issue is systematic. I think the solution is to add a single test (is there a rule base test?) that validates that the description of every rule is <100 (arbitrary) characters and then fix the long ones. No-one should manually read and validate messages... 😅

BraisGabin:
That sounds like a good idea to me.

PR #4386, opened 2021-12-19
Add first draft of a rule description style guide

pnq32:
Based on #4384, this PR contains a draft of a style guide for rule descriptions.


.github/contributing/rule-description-style.md
```diff
@@ -0,0 +1,78 @@
...
+that Detekt identified in the code rather than what the rule itself did to come to come to its conclusion.
+
+**Guideline 1**: Rule descriptions shall be focused on the identified issue rather than what the rule itself does.
+
+- :heavy_check_mark:: ``Duplicated case statements in `when` expression.``
+- :x:: `Detects trailing spaces.`
+
+## b) Components to incorporate
+
+There are three categories of information that might be incoporated into a rule description:
```
BraisGabin, line 23:
```suggestion
There are three categories of information that might be incorporated into a rule description:
```

BraisGabin (commented):
Thank you so much!

I think that we should define the 3 different "descritpions" that a rule have:

- The on that is in the javadoc of the rule. That description is used in the html documentation that detetk have. An example: https://detekt.github.io/detekt/coroutines.html. For example, I think that this description should have links to different sources to explain why something is a bad pattern.
- The rule description (Inside the class `Issue`). That's the description of the issue that is used in the reports like html and sarif. In general I think that it should be small than the one in the javadoc. We canalways add a link to the documentation in the html or sarif report to get the complete description with samples, default configurations etc.
- The issue message. That's the message that tells you what's wrong with your code.

I think that there are points in this guide that are general for all of these descriptions but, in geenral, it is targeting the last one. What do you think about talking a bit about this three descriptions and be more precise about what to write in each description. For example I just saw that some issue messagges directly use the rule description. That doesn't feel right.

pnq32:
Thank you very much for your detailed response. I have to admit that until now, I was not really aware of the internal structure that Detekt uses to distinguish rule documentation, issue descriptions, and code smell descriptions.

If I understand you correctly, you are referring to these three kinds of descriptions:

1. Javadoc comments such as [this one](https://github.com/detekt/detekt/blob/7250c885d7ca35c1b26f0af565067229375c9145/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NonBooleanPropertyPrefixedWithIs.kt#L20). They are used to generate documentation sections such as [this one](https://detekt.github.io/detekt/naming.html#nonbooleanpropertyprefixedwithis).
2. Static `String` values used to populate the `description` parameter of an `Issue`, such as in [this line](https://github.com/detekt/detekt/blob/7250c885d7ca35c1b26f0af565067229375c9145/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NonBooleanPropertyPrefixedWithIs.kt#L40). They are used as 'table headers' in reports (such as an HTML report) and serve as the 'container' for all code smells that the rule has identified. They are used neither from `FindingsReport` nor from `LiteFindingsReport` presented to the user via the console.
3. Dynamically generated `String` values that a rule uses to to label each identified code smell, such as [here](https://github.com/detekt/detekt/blob/7250c885d7ca35c1b26f0af565067229375c9145/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NonBooleanPropertyPrefixedWithIs.kt#L84). In generated reports, they are presented to the user as 'items' belonging to their 'enclosing issue'. Furthermore, the new `LiteFindingsReport` makes use of these messages to generate console reports.

Is this correct or did I get something wrong here? If it is, then I fully agree with your suggestion.

What do you think of the following rough ideas?
1. **Javadoc strings** are focused on the strategy pursued by a rule and should answer questions such as the following:
    - What exactly does this rule check for?
    - How does it try to identify violations?
    - Why (according to which references) is it important to identify such violations?
    - What does the rule do to identify specific recommendations targeted at the developer?
2. **Issue descriptions** are formulated under the assumption that at least one code smell has been identified by the respective rule. They should give a concise description of the 'general violation type' that was found in the codebase and, if applicable, summarize why such a violation is bad practice. If it is not obvious, they should also give a rough summary of what is generally done to tackle such violations.
3. **Code smell messages** are concise extensions of issue messages. They have the goal to tell the developer exactly what the violation is, ideally using specific variable/function names or similar knowledge from the analyzed code. Furthermore, they can give much more specific recommendations (if it makes sense in the specific context).

For instance, assume that `NonBooleanPropertyPrefixedWithIs` is extended to handle additional prefixes (such as `has` or `can`) and therefore renamed to something like `MisleadingBooleanPropertyPrefix`:
1. The [current Javadoc string](https://github.com/detekt/detekt/blob/7250c885d7ca35c1b26f0af565067229375c9145/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/NonBooleanPropertyPrefixedWithIs.kt#L20) would obviously need to be extended for the other prefixes and might benefit from a more detailed explanation of why this naming is bad practice. But in general, I think it is fine and can just remain as it is.
2. The issue description would look something like this: `Non-boolean property with prefix suggesting boolean type. This might mislead clients of the API and facilitate programming errors on their side. Such prefixes should therefore be removed.`.
3. A code smell message would be more specific and focused on the actual violation: ``Non-boolean property `hasXyz` suggests a boolean type.``, potentially with a very short recommendation.

If it makes sense, I think it would be nice to encourage 'code-oriented' recommendations in code smell messages. The rule checking for magic numbers, for instance, could determine whether or not you are passing the magic number to a Java method or a Kotlin function. For the Java case, the message could look like this:
- ``Magic number `4` passed to `Class::method`. Consider introducing a named constant.``

If it is a Kotlin function and the magic number is only used once, it could look like this:
- ``Magic number `4` passed to `abc`. Consider passing it as a named argument.``

Please let me know what you think. If this makes sense, I am happy to update the current draft accordingly.

BraisGabin:
> Is this correct or did I get something wrong here?

100% correct. You explained way better than me.

> Please let me know what you think. If this makes sense, I am happy to update the current draft accordingly.

I like it!

The only thing I'm not sure is if we should explain "How" the rule detects the issues in point 1. I think that we should focus on say what exactly does it detect and why it is important (with references if possible). I think that's the more important information that you need in the documentation. That's what you are going to read to decide if you want to enable/disable a rule. So I think that these two points shouldn't be there:

>        * How does it try to identify violations?
>        * What does the rule do to identify specific recommendations targeted at the developer?

Why do you think those two points are important?

pnq32:
> > ```
> >    * How does it try to identify violations?
> >    * What does the rule do to identify specific recommendations targeted at the developer?
> > ```
> Why do you think those two points are important?

I agree that these two points are much less important than the other two aspects. In fact, I think that they are irrelevant for 95 percent of rules. However, my initial idea was that for rather sophisticated rules (with a lot of configuration options), some knowledge about what the rule does (instead of what it is generally concerned with) would simplify the configuration process.

For example, think of an improved `MagicNumber` rule that allows you (as the person coming up with the config) to specify if Detekt should recommend using named arguments over introducing named constants if the magic number is not passed to a Java function. Then the Javadoc could go something like this:

> This rule recommends to replace each magic number detected in a given Kotlin file with more suitable constructs. For each magic number passed to a Java method, it will issue a recommendation to replace it with a named constant. For magic numbers passed to Kotlin functions, the issued recommendation depends on the value of the the `preferNamedArguments` option. If it is active, the rule will recommend to pass the magic number using a named argument. Otherwise, it will recommend to either replace the magic number with a named constant or to pass it using a named argument.

Nevertheless, I totally get your point. Considering that developers will have the opportunity to just look at the code, it is probably neither necessary not worth the effort. So let's just focus on what the rule checks for and why that might be bad practice.

One last question before I get started: Do you happen to know how this relates to #4163? More specifically, what exactly is it that GitHub shows in the table: issue descriptions or code smell messages? I was thinking that it would probably make sense to encourage short texts for whatever is displayed in this overview.

BraisGabin:
I'm not sure what is shown in each case but I think that in that report we are setting the rule description as the title. In detekt we don't have a title for a rule so probably we fail in that map. Probably we should use the id of the rule there instead of the description. 

>  was thinking that it would probably make sense to encourage short texts for whatever is displayed in this overview.

I think that the only description that should be long is the one in javadoc. The other two should be direct to the point. One telling what the rule checks or what an issue raised by this rule means. And the other saying what was found and (if possible) a recommendation about how to fix it.

pnq32:
Alright, I have now updated the style guide to cover the three categories you mentioned.

Could you please check and see if this reflects what you had in mind?

.github/contributing/rule-description-style.md
```diff
@@ -0,0 +1,143 @@
...
+- :heavy_check_mark: (*violation and recommendation*): ``Public library class. Reduce its visibility to the module or the file.``
+- :x: (*just a rationale*): `A function that only returns a constant is misleading.`
+- :heavy_check_mark: (*violation and rationale*): `Function returns a constant. This is misleading.`
+- :heavy_check_mark: (*all categories*): `Function returns a constant, which is misleading. Use a constant property instead.`
+
+### Components of code smell messages
+A code smell message should be dynamically created to describe a
+*specific violation*. It should be regarded as an extension of the more generic
+violation part of the issue description. More specifically, it should explicitly
+reference identifiers, file names, or similar from the codebase. If it makes
```
BraisGabin, line 125:
```suggestion
reference identifiers or similar from the codebase. If it makes
```

I think that the file name is the only thing that this message shouldn't reference. It is always referenced by detekt itself.

BraisGabin (approved):
👏👏👏 Great work! 👏👏👏

I really like it :) I will keep this PR open for now to get more feedback but for me this is ready to merge.

The only thing is that, probably, we should add a line in this file https://github.com/detekt/detekt/blob/main/.github/CONTRIBUTING.md with a link to this guide.

pnq32:
That is good to hear, thank you very much for your help!

One last remark from my side: I was unsure whether or not we want to encourage configuration-dependent issue descriptions. I have left it unspecified for now, but I think that it might actually be worthwhile to handle this consistently as well. For instance, think of the hypothetical `MagicNumbers` rule with a `preferNamedArguments` option again. Assume that the default issue description goes something like this:
> Numeric literal used as magic number. To avoid confusion on the side of the client, introduce a named constant or pass such literals as a named argument.

If `preferNamedArguments` is activated, should the issue description change to something like this?
> Numeric literal used as magic number. To avoid confusion on the side of the client, try to pass such literals as a named argument. If this is not possible, introduce a named constant.

Or should the issue description be an entirely static text?

BraisGabin:
They could be dynamic but I don't know if that would be too specific for the guide. I mean, if someone wants to improve a rule adding dynamic description would be great but I don't know about encouraging it.

BraisGabin:
@schalkms I know that you editted lots of rules descriptions based on this style guide so: could you review this? I would like to have multiple approvals here before merging.

.github/contributing/rule-description-style.md
```diff
@@ -0,0 +1,150 @@
+# Style guide for authoring rule descriptions
+
+A rule incorporated into the detekt framework is generally accompanied by three
```
schalkms, line 3: In general, we tried to be consistent in the docs by writing detekt in lowercase.
I quickly searched and replaced the occurrences in the last commit on this branch.

```suggestion
A rule incorporated into the detekt framework is generally accompanied by three
```

.github/contributing/rule-description-style.md
```diff
@@ -0,0 +1,150 @@
...
+
+*Note*: `non-null` is a borderline case. While `null` itself can be seen as
+code, the prefix turns it into a conventional word.
+
+The above-mentioned guidelines tell you *how* to formulate issue descriptions
+and code smell messages. To learn *what to include* in each of the texts,
+refer to the following subsections.
+
+### Components of issue descriptions
+An issue description should consist of the following components (in this order):
```
schalkms, line 98: For a consistent styling, we could add a new line between the heading and first sentence.

```suggestion

An issue description should consist of the following components (in this order):
```

.github/contributing/rule-description-style.md
```diff
@@ -0,0 +1,150 @@
...
+The following list contains some positive and negative examples:
+- :heavy_check_mark: (*just a violation*): `Public library class.`
+- :x: (*just a recommendation*): `Library class should not be public.`
+- :heavy_check_mark: (*violation and recommendation*): ``Public library class. Reduce its visibility to the module or the file.``
+- :x: (*just a rationale*): `A function that only returns a constant is misleading.`
+- :heavy_check_mark: (*violation and rationale*): `Function returns a constant. This is misleading.`
+- :heavy_check_mark: (*all categories*): `Function returns a constant, which is misleading. Use a constant property instead.`
+
+### Components of code smell messages
+A code smell message should be dynamically created to describe a
```
schalkms, line 122: For a consistent styling, we could add a new line between the heading and first sentence.

```suggestion

A code smell message should be dynamically created to describe a
```

.github/contributing/rule-description-style.md
```diff
@@ -0,0 +1,150 @@
...
+   individuals reading these texts will usually expect specific references to
+   *their codebase* and what can be done in order to improve it.
+
+## Guidelines for documentation strings
+
+When authoring the documentation string for a rule, adhere to the following
+guidelines:
+1. The documentation string shall be formulated in such a manner that it refers
+   to the rule itself:
+   - :heavy_check_mark:: `Detects trailing spaces and recommends to remove them.`
```
schalkms, line 44: I really like this style with *Dos* and *Don'ts*. This is a great idea!

.github/contributing/rule-description-style.md
```diff
@@ -0,0 +1,150 @@
...
+3. **Recommendation** (*optional*): What are developers usually expected to do
+   in order to eliminate the identified violation? In some cases, this might be
+   entirely obvious. If it is not, however, it makes sense to add a suitable
+   recommendation (or multiple alternative recommendations).
+
+:warning: In any case, try to keep the description as brief and concise as
+possible!
+
+The following list contains some positive and negative examples:
+- :heavy_check_mark: (*just a violation*): `Public library class.`
```
schalkms, line 114: Shouldn't it be annotated with an `:x:` as well?

schalkms:
```suggestion
- :x: (*just a violation*): `Public library class.`
```

.github/CONTRIBUTING.md
```diff
@@ -12,8 +12,8 @@
 
 ### When implementing new rules ...
 
-- ... do not forget to add the new rule to a `RuleSetProvider` (e.g. StyleGuideProvider)
-- ... do not forget to write a description for the issue of the new rule.
+- ... add the new rule to a `RuleSetProvider` (e.g. StyleGuideProvider).
```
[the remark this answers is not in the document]
schalkms: This improvement could also be part of a different PR.

schalkms (commented):
I read this guide and I have to say, I really like your writing style. It's excellent. You have done a great job! 👏
Thanks for spending the time to submit this PR and improve the documentation a lot! 🙏

.github/CONTRIBUTING.md
```diff
@@ -12,8 +12,8 @@
 
 ### When implementing new rules ...
```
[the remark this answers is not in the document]
pnq32: Sure, I have tried to take care of this as part of 4e20631.

.github/CONTRIBUTING.md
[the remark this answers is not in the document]
pnq32: See above, I tried to incorporate your suggestion as part of 4e20631.

pnq32:
@schalkms: Thank you for trying to apply the style guide to the codebase and for your helpful feedback!

I think there is one aspect that we should clarify before finalizing it, though. The way the guidelines are currently formulated, they require you to describe the **violation** of an issue and allow you to (optionally) add a **rationale** and/or a **recommendation**. If you follow this style guide closely, the following issue description would actually be valid:
- :heavy_check_mark:: `Public library class.`

I agree that in this specific case, it would actually be beneficial to extend it a bit, but the current style guide does not require you to do so. After all, most detekt users would probably understand what they are expected to do. At the same time, the style guide does not allow you to omit the **violation** part. Therefore, strictly speaking, the following issue descriptions would not comply to the guide:
- :x:: `Don't try to be smarter than the JVM. Your code should work independently whether the...`
- :x:: ``The `next()` method of an Iterator implementation should throw a NoSuchElementException...``
- :x:: `Function invocations which have many parameters should be named.`
- :x:: `Library classes should not be public.`

I obtained these examples from the changes you have just applied to existing issue descriptions. Combined with your comment above, I can see that you seem to have slightly different idea about how issue descriptions should look like. The set of rules I came up with seems to be both too lax and too strict at the same time. What do you think, is this the case? If it is: Could you maybe try to summarize what your idea of the ideal issue description is? I would then try to incorporate it into the draft.

.github/CONTRIBUTING.md
```diff
@@ -12,8 +12,8 @@
...
 
-- ... do not forget to add the new rule to a `RuleSetProvider` (e.g. StyleGuideProvider)
-- ... do not forget to write a description for the issue of the new rule.
+- ... add the new rule to a `RuleSetProvider` (e.g. StyleGuideProvider).
+- ... compose an issue description and code smell messages according to [this style guide](contributing/rule-description-style.md).
```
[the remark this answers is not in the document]
pnq32: Yes, that would in fact be nice. I suggest to first finalize the content of the style guide itself. After this is done, I am happy to give it a try and merge the style guide into `CONTRIBUTING.md`.

schalkms:
I think we have the same understanding regarding how an issue description should look like with the mandatory violation but optional recommendation part. As I have said, it's quite a tedious task to go over ~230 rules, so I might have missed a few descriptions. I'm happy to adapt and further improve the rule descriptions.

Regarding example 1 and 2, I think it's quite clear what the violation is, because it reads like this in the report, whereas example 3 and 4 should be adapted.

1. :x:: `Don't try to be smarter than the JVM. Your code should work independently whether the...`
> **ExplicitGarbageCollectionCall**: Don't try to be smarter than the JVM. Your code should work independently whether the garbage collector is disabled or not. If you face memory issues, try tuning the JVM options instead of relying on code itself.
2. :x:: ``The `next()` method of an Iterator implementation should throw a NoSuchElementException...``
> **IteratorNotThrowingNoSuchElementException**: The `next()` method of an `Iterator` implementation should throw a `NoSuchElementException` when there are no more elements to return. 
3. :x:: `Function invocations which have many parameters should be named.`
4. :x:: `Library classes should not be public.`

pnq32:
Alright, I have now tried to address all your remarks. Most importantly, I have merged the style guide into `CONTRIBUTING.md`.

While doing so, I was unfortunately unable to make proper use of the `<details>` block. When I try to wrap portions of the style guide into these blocks, some Markdown formatting seems to break. For instance, inline code (wrapped by `` ` ``) does not seem to work anymore and the vertical space between bullet point items seems off. If you know suitable workarounds, I would be happy if you could take a look at how to do this. For now, I have just used subheadings to structure the document.

What do you think, does this go into the direction you had in mind?

schalkms (approved):
I read through it and completely approve it. Good documentation is highly valued.
Sorry for the review and merging delay from our side, so thanks for being patient with us.

> While doing so, I was unfortunately unable to make proper use of the <details> block.

Oh, I didn't know that. Using subheadings is also fine.

> What do you think, does this go into the direction you had in mind?

👍

BraisGabin:
Again, thank you so much @pnq32 :)

pnq32:
> Sorry for the review and merging delay from our side, so thanks for being patient with us.

Don't worry, I did not perceive the iterations as a delay. After all, there were a few decisions to make.

Thank you very much for your help to both of you! :slightly_smiling_face:

Issue #875, opened 2018-04-22
UnnecessaryInnerClass Rule


vanniktech:
Sometimes nested classes are declared as `inner` although they should not. In that case the fix is trivial - removing the `inner` keyword. It would be nice if there's a rule for this.

There's a high risk for false positives though so it should be tested thoroughly on sample projects.

schalkms:
Good idea!

BraisGabin:
This is already implmented in 1.20 🎉

PR #4394, opened 2021-12-23
Added rule UnnecessaryInnerClass

severn-everett:
This addresses issue #875.


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryInnerClassSpec.kt
```diff
@@ -0,0 +1,329 @@
...
+                        
+                        fun printFizz() {
+                            println(fileFoo)
+                            println(fizz)
+                        }
+                    }
+                }
+            """.trimIndent()
+
+            Assertions.assertThat(subject.lintWithContext(env, code)).hasSize(1)
```
schalkms, line 40: The `Assertions.` reference can be imported and thus removed from the `assert` statements for a cleaner test structure.
Other rules show examples on how to use it.

schalkms (commented):
Thanks for tackling this very old issue! I think this rule fits nicely into detekt.
As mentioned in the issue description, there's a high risk for false positives though. Hence, this rule should be tested thoroughly on sample projects. Have you run this rule on other codebases? If not, the following script in detekt might help to achieve that.

https://github.com/detekt/detekt/blob/main/scripts/get_analysis_projects.kts

BraisGabin (approved):

severn-everett:
> Thanks for tackling this very old issue! I think this rule fits nicely into detekt. As mentioned in the issue description, there's a high risk for false positives though. Hence, this rule should be tested thoroughly on sample projects. Have you run this rule on other codebases? If not, the following script in detekt might help to achieve that.
> 
> https://github.com/detekt/detekt/blob/main/scripts/get_analysis_projects.kts

The rule requires the binding context to be present, which would mean configuring the `classpath` option when running Detekt on each of the sample projects. Is there a way to determine the classpaths for these projects programatically? Otherwise, as they are both non-trivial and externally-managed, a manually-maintained list of classpaths for each project manually would be rather fragile. Given that this is not going to be an active rule by default - and thus won't be breaking workflows automatically when users upgrade Detekt - I would say to publish the rule in the next release and address the issues as they come.

schalkms:
I'm still hesitant to merge this, since I feel like the rule could take a few more test runs. Testing it in production is not the route, which I'd like to go here.

severn-everett:
Why not? Releasing experimental and beta code is not an unusual workflow, e.g. the pattern matching preview code that's been released in JDK 16 and 17; and given that the rule will be disabled by default, no user's going to have their code broken by any issues with this rule unless they actively go out of their way to enable it. It'll be a lot more straightforward to be able to test the new rule when it can be integrated via the Gradle plug-in; otherwise, it's going to be a very arduous task to hand-build the classpath for all of these projects to test out the rule manually.

chao2zhang:
Alternatively, we could enable this new rule for the detekt project itself to help us evaluate. I agree with @severn-everett that since this is a new rule disabled by default, we can roll it out to production and gather feedback.

A counterpart is `UnusedPrivateMember`, this is a rule that seems to require constant maintenance. I don't think `UnnecessaryInnerClass` would require as much load as `UnusedPrivateMember`.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryInnerClass.kt
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    override fun visitDotQualifiedExpression(expression: KtDotQualifiedExpression) {
+        super.visitDotQualifiedExpression(expression)
+        checkForOuterUsage { parentClasses ->
+            expression.receiverExpression.belongsToParentClass(parentClasses)
+        }
+    }
+
+    // Replace this "contructor().apply{}" pattern with buildSet() when Kotlin is upgraded to 1.6
```
cortinico, line 146: Can we replace it now? https://github.com/detekt/detekt/pull/4277

severn-everett: Yeah, I'll swing back and fix this tomorrow.

severn-everett: It doesn't look like it's available until the API version is at 1.6; I'll update the note to reflect that

BraisGabin (approved):

cortinico (approved):

chao2zhang (approved):

PR #4397, opened 2021-12-26
Remove unnecessary map lambda in test code

schalkms:


BraisGabin (approved):
Do we have a rule to detect this? Probably worth it.

schalkms:
Nope, there isn't. Otherwise, detekt would have detected this snippet.😉

> Do we have a rule to detect this? Probably worth it.

PR #4399, opened 2021-12-26
Fixes false negatives in `UnnecessaryAbstractClass`

BraisGabin:
This PR fixes a false positive in `UnnecessaryAbstractClass`.

Before this PR the class `B` wasn't flagged:

```kotlin
interface A {
    val i: Int
}
abstract class B : A
```

And the same here:

```kotlin
abstract class A {
    abstract val i: Int
}
abstract class B : A()
```

This PR also improves the messages reported for this rule.

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/IndentingXMLStreamWriter.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/IndentingXMLStreamWriter.kt
@@ -5,9 +5,9 @@ import javax.xml.stream.XMLStreamWriter
 
 @Suppress("TooManyFunctions")
 internal class IndentingXMLStreamWriter(
-    writer: XMLStreamWriter,
+    private val writer: XMLStreamWriter,
     private val indent: String = "  "
-) : DelegatingXMLStreamWriter(writer) {
+) : XMLStreamWriter by writer {
 
     private var currentState = NOTHING
     private val stateStack = Stack<Any>()
@@ -25,7 +25,7 @@ internal class IndentingXMLStreamWriter(
     private fun onEndTag() {
         indentationDepth--
         if (currentState === TAG) {
-            super.writeCharacters("\n")
+            writer.writeCharacters("\n")
             writeIndent()
         }
         currentState = stateStack.pop()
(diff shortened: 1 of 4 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/IndentingXMLStreamWriter.kt
```diff
@@ -5,9 +5,9 @@ import javax.xml.stream.XMLStreamWriter
 
 @Suppress("TooManyFunctions")
 internal class IndentingXMLStreamWriter(
-    writer: XMLStreamWriter,
+    private val writer: XMLStreamWriter,
     private val indent: String = "  "
-) : DelegatingXMLStreamWriter(writer) {
+) : XMLStreamWriter by writer {
```
schalkms, line 10: Is there a reason why the two classes regarding the `XMLStreamWriter` are refactored and part of this PR?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAbstractClass.kt
```diff
@@ -73,12 +73,13 @@ class UnnecessaryAbstractClass(config: Config = Config.empty) : Rule(config) {
...
-            val body = klass.body
-            if (body != null) {
-                val namedMembers = body.children.filterIsInstance<KtNamedDeclaration>()
-                val namedClassMembers = NamedClassMembers(klass, namedMembers)
-                namedClassMembers.detectAbstractAndConcreteType()
-            } else if (klass.superTypeListEntries.isEmpty()) {
+            val namedMembers = klass.body?.children.orEmpty().filterIsInstance<KtNamedDeclaration>()
+            if (namedMembers.isNotEmpty()) {
+                NamedClassMembers(klass, namedMembers)
+                    .detectAbstractAndConcreteType()
```
schalkms, line 79: The formatting looks a bit weird here IMHO.

```suggestion
                NamedClassMembers(klass, namedMembers).detectAbstractAndConcreteType()
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAbstractClass.kt
```diff
@@ -73,12 +73,13 @@ class UnnecessaryAbstractClass(config: Config = Config.empty) : Rule(config) {
...
+            } else if (!klass.hasConstructorParameter()) {
+                report(CodeSmell(issue, Entity.from(klass), noConcreteMember), klass)
```
schalkms, line 81: Where did you improve the message description?

> This PR also improves the messages reported for this rule.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAbstractClassSpec.kt
```diff
@@ -38,28 +33,56 @@ class UnnecessaryAbstractClassSpec : Spek({
...
+                    assertFindingMessage(findings, message)
+                }
+
+                it("that inherits from an interface") {
+                    val code = """
+                        interface A {
+                            val i: Int
+                        }
+                        abstract class B : A
+                    """.trimIndent()
```
schalkms, line 70: I don't think we used the trim statement for the code snippets except for asserts regarding the reported line number.

```suggestion
                    """
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAbstractClassSpec.kt
```diff
@@ -38,28 +33,56 @@ class UnnecessaryAbstractClassSpec : Spek({
...
+                }
+
+                it("that inherits from another abstract class") {
+                    val code = """
+                        @Deprecated("We don't care about this first class")
+                        abstract class A {
+                            abstract val i: Int
+                        }
+                        abstract class B : A()
+                    """.trimIndent()
```
schalkms, line 82:
```suggestion
                    """
```

schalkms (commented):
I guess the mentioned code snippets aren't flagged a second time by the `empty` rule set, right?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/IndentingXMLStreamWriter.kt
BraisGabin: `DelegatingXMLSteamingWriter` was a unnecessary usage of abstract so as soon as I fixed the rule these code was flagged. For that reason it is in the same pr. I can move it to other PR but that one should be merged before this one then.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAbstractClass.kt
BraisGabin: You are right. I didn't "improve the message". Before there were some issues that had the message "An abstract class without an abstract member can be refactored to a concrete class." but it should be "An abstract class without a concrete member can be refactored to an interface." and the improvement was that. I set the correct message to some issues that were getting the wrong one.

BraisGabin:
> I guess the mentioned code snippets aren't flagged a second time by the `empty` rule set, right?

I guess that some cases are flagged by that rule set too. And I think that it's good that both rules flag it because one developer could have one of those issues disabled.

schalkms:
If I remember correctly, we excluded those cases for exactly this reason. Thus, the same code is not flagged twice by two different rules. I could be wrong here, however.

> I guess that some cases are flagged by that rule set too. And I think that it's good that both rules flag it because one developer could have one of those issues disabled.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAbstractClass.kt
schalkms: That's not necessary IMHO.
--> resolved

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/IndentingXMLStreamWriter.kt
schalkms: That's not necessary IMHO.
--> resolved

BraisGabin:
I didn't introduce those cases in this PR. Those cases where there already, they weren't tested but they were flagged. I just extended the tests to ensure that I didn't break anything.

schalkms:
I'm afraid, I didn't understand this statement.  As I understood the description of the rule, you fixed a false-positive in this PR (e.g. `abstract class B : A` as well as `abstract class B : A()`).

> I didn't introduce those cases in this PR. Those cases where there already, they weren't tested but they were flagged.

BraisGabin:
This rule was already flagging code at the same time as the empty rule set. I didn't introduce that issue here. For example:

```kotlin
abstract class A : B {}
```

It was flagged by `UnnecessaryAbstractClass` and by `EmptyClassBlock`.

For that reason I say that I didn't introduce this "double report", it was there. For sure, now there are more "double reports" but that is because now `UnnecessaryAbstractClass` is working as expected.

schalkms:
Thanks for the elaboration, Brais! It's good to merge.

schalkms (approved):

PR #4400, opened 2021-12-26
Fix false-positive on ExplicitCollectionElementAccessMethod

### LINKED ISSUE #4288: ExplicitCollectionElementAccessMethod false positive
## Expected Behavior
`ExplicitCollectionElementAccessMethod` Shouldn't flag this ussage but it does: [`Rectf.set(left, top, right, bottom)`](https://developer.android.com/reference/android/graphics/RectF#set(float,%20float,%20float,%20float))

## Your Environment
* Version of detekt used: main (c66b8d67bb78b3361710e70bd9a060f877ef527d)
* Version of Gradle used (if applicable): 7.3

BraisGabin:
This fixes the issue raised con #4288 implementing the solution proposed in this comment: https://github.com/detekt/detekt/issues/4288#issuecomment-980314558:

> Alternatively, we can report **only** when the set/get has exactly one parameter for the key.

In this case I'm just checking for the `set` case because I think that all the cases for `get` should work as expected.

fixes #4288

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExplicitCollectionElementAccessMethod.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExplicitCollectionElementAccessMethod.kt
@@ -10,6 +10,7 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
 import io.gitlab.arturbosch.detekt.rules.fqNameOrNull
 import org.jetbrains.kotlin.descriptors.FunctionDescriptor
+import org.jetbrains.kotlin.load.java.isFromJava
 import org.jetbrains.kotlin.psi.KtBlockExpression
 import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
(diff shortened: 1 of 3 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExplicitCollectionElementAccessMethod.kt
```diff
@@ -57,21 +58,23 @@ class ExplicitCollectionElementAccessMethod(config: Config = Config.empty) : Rul
...
 
     private fun isIndexableSetter(expression: KtCallExpression): Boolean =
         when (expression.calleeExpression?.text) {
-            "set" -> isOperatorFunction(expression)
+            "set" -> {
+                val function = expression.getFunctionDescriptor()
+                when {
+                    function == null -> false
+                    !function.isOperator -> false
+                    else -> !(function.isFromJava && function.valueParameters.size > 2)
```
schalkms, line 70: Here you check for the parameter list size to be **bigger than 2**, but in the test case underneath you assert the following.

> does not report if the function has **more than 3** arguments and it's defined in java

What system behavior can the user of this rule expect in the mentioned case?

BraisGabin: > Here you check for the parameter list size to be **bigger than 2**, but in the test case underneath you assert the following.
> 
> > does not report if the function has **more than 3** arguments and it's defined in java

I fixed the description of the test to:

`does not report if the function has 3 or more arguments and it's defined in java`

> What system behavior can the user of this rule expect in the mentioned case?

(I'm not sure if I understood what you are asking) The idea is to ignore the cases where the a function `set` is called and if it is defined in java and java 3 or more arguments. The full context is in the linked issue.

schalkms: I'm referring to the logical and numerical difference between the two statements.

```
code:  args > 2
test:  args > 3
```

Thanks for fixing the test. However, I think the code snippet in the test should then have exactly 3 parameters to assert the boundary (e.g. 3).
Excuse me for being a bit picky here.

BraisGabin: Done.

schalkms (approved):

PR #4413, opened 2021-12-27
Formulate rule/exceptions descriptions consistently

schalkms:
reference #4384

### DIFF
--- a/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/InstanceOfCheckForException.kt
+++ b/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/InstanceOfCheckForException.kt
@@ -48,7 +48,7 @@ class InstanceOfCheckForException(config: Config = Config.empty) : Rule(config) 
     override val issue = Issue(
         "InstanceOfCheckForException",
         Severity.CodeSmell,
-        "Instead of catching for a general exception type and checking for a specific exception type " +
+        "Instead of catching for a general exception type and checking for a specific exception type, " +
             "use multiple catch blocks.",
         Debt.TWENTY_MINS
     )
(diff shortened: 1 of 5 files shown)


detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/InstanceOfCheckForException.kt
```diff
@@ -48,7 +48,7 @@ class InstanceOfCheckForException(config: Config = Config.empty) : Rule(config)
     override val issue = Issue(
         "InstanceOfCheckForException",
         Severity.CodeSmell,
-        "Instead of checking for a general exception type and checking for a specific exception type " +
+        "Instead of catching for a general exception type and checking for a specific exception type " +
```
chao2zhang, line 51:
```suggestion
        "Instead of catching for a general exception type and checking for a specific exception type, " +
```

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/NotImplementedDeclaration.kt
```diff
@@ -32,9 +32,9 @@ class NotImplementedDeclaration(config: Config = Config.empty) : Rule(config) {
     override val issue = Issue(
         "NotImplementedDeclaration",
         Severity.CodeSmell,
-        "The NotImplementedDeclaration should only be used when a method stub is necessary. " +
+        "The `NotImplementedDeclaration` should only be used when a method stub is necessary. " +
```
chao2zhang, line 35: NotImplementedDeclaration is not something our users would write, I don't think we should put it under `

chao2zhang (approved):

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/NotImplementedDeclaration.kt
schalkms:
```suggestion
        "The NotImplementedDeclaration should only be used when a method stub is necessary. " +
```

PR #1964, opened 2019-10-02
Extend VarCouldBeVal Rule to Include Class/Object Properties

sowmyav24:
Fixes [#1330 ](https://github.com/arturbosch/detekt/issues/1330)

- VarCouldBeVal would run for the class/object variables also
- Excluded variables having lateinit modifiers as they could be injected. Lateinit modifers are verified in `LateinitUsage` rule already.
- Removed specific check for `function.isSomehowNested` . The corresponding tests still pass. Please let me know if this is needed for some other reason

PR #4424, opened 2021-12-29
Extend VarCouldBeVal to include analysis of file- and class-level properties

severn-everett:
Second poor soul attempting to conquer #1964. Also, I noticed what could be a potential false-positive issue for `var` properties declared in object literals that escape their containing declaration, so this code includes checks for that.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
@@ -20,6 +20,7 @@ import org.jetbrains.kotlin.psi.KtIfExpression
 import org.jetbrains.kotlin.psi.KtNameReferenceExpression
 import org.jetbrains.kotlin.psi.KtNamedDeclaration
 import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.KtObjectDeclaration
 import org.jetbrains.kotlin.psi.KtObjectLiteralExpression
 import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.KtQualifiedExpression
@@ -80,12 +81,12 @@ class VarCouldBeVal(config: Config = Config.empty) : Rule(config) {
 
     private class AssignmentVisitor(private val bindingContext: BindingContext) : DetektVisitor() {
 
-        private val declarations = mutableSetOf<KtNamedDeclaration>()
+        private val declarationCandidates = mutableSetOf<KtNamedDeclaration>()
         private val assignments = mutableMapOf<String, MutableSet<KtExpression>>()
         private val escapeCandidates = mutableMapOf<DeclarationDescriptor, List<KtProperty>>()
 
         fun getNonReAssignedDeclarations(): List<KtNamedDeclaration> {
-            return declarations.filterNot { it.hasAssignments() }
+            return declarationCandidates.filterNot { it.hasAssignments() }
         }
 
         private fun KtNamedDeclaration.hasAssignments(): Boolean {
@@ -107,8 +108,8 @@ class VarCouldBeVal(config: Config = Config.empty) : Rule(config) {
 
         override fun visitProperty(property: KtProperty) {
             super.visitProperty(property)
-            if (property.isEligible()) {
-                declarations.add(property)
+            if (property.isDeclarationCandidate()) {
+                declarationCandidates.add(property)
             }
 
             // Check for whether the initializer contains an object literal.
@@ -149,7 +150,9 @@ class VarCouldBeVal(config: Config = Config.empty) : Rule(config) {
         ) {
             when (rightExpression) {
                 is KtObjectLiteralExpression -> {
-                    escapeCandidates[descriptor] = rightExpression.collectDescendantsOfType()
+                    escapeCandidates[descriptor] = rightExpression.collectDescendantsOfType {
+                        it.isEscapeCandidate()
+                    }
                 }
                 is KtIfExpression -> {
                     evaluateAssignmentExpression(descriptor, rightExpression.then)
@@ -166,11 +169,13 @@ class VarCouldBeVal(config: Config = Config.empty) : Rule(config) {
         private fun evaluateReturnExpression(returnedExpression: KtExpression?) {
             when (returnedExpression) {
                 is KtObjectLiteralExpression -> {
-                    returnedExpression.collectDescendantsOfType<KtProperty>().forEach(declarations::remove)
+                    returnedExpression.collectDescendantsOfType<KtProperty> {
+                        it.isEscapeCandidate()
+                    }.forEach(declarationCandidates::remove)
                 }
                 is KtNameReferenceExpression -> {
                     returnedExpression.getResolvedCall(bindingContext)?.resultingDescriptor?.let { descriptor ->
-                        escapeCandidates[descriptor]?.forEach(declarations::remove)
+                        escapeCandidates[descriptor]?.forEach(declarationCandidates::remove)
                     }
                 }
                 is KtIfExpression -> {
@@ -185,7 +190,7 @@ class VarCouldBeVal(config: Config = Config.empty) : Rule(config) {
             }
         }
 
-        private fun KtProperty.isEligible(): Boolean {
+        private fun KtProperty.isDeclarationCandidate(): Boolean {
             return when {
                 !isVar -> false
                 isLocal || isPrivate() -> true
@@ -197,6 +202,11 @@ class VarCouldBeVal(config: Config = Config.empty) : Rule(config) {
                         ?.containingNonLocalDeclaration() != null
                 }
             }
+        }
+
+        private fun KtProperty.isEscapeCandidate(): Boolean {
+            return !isPrivate() &&
+                (containingClassOrObject as? KtObjectDeclaration)?.isObjectLiteral() == true
         }
 
         private fun visitAssignment(assignedExpression: KtExpression?) {
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
@@ -10,17 +10,29 @@ import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
 import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
+import org.jetbrains.kotlin.descriptors.DeclarationDescriptor
 import org.jetbrains.kotlin.lexer.KtTokens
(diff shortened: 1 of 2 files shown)


severn-everett:
First issue identified:
```
[reports] at /home/runner/work/detekt/detekt/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt:78:5
```
I'm not sure what kind of "tell" there would be for this one, given it's marked as `private` not overtly modified in the class. It seems like this would be a place to add a suppression annotation. Thoughts?

schalkms:
> [reports] at /home/runner/work/detekt/detekt/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt:78:5

I guess `reports` can be made `public` as all other fields in this class or `internal`.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
```diff
@@ -55,30 +68,25 @@ class VarCouldBeVal(config: Config = Config.empty) : Rule(config) {
...
     )
 
-    override fun visitNamedFunction(function: KtNamedFunction) {
-        if (bindingContext == BindingContext.EMPTY || function.isSomehowNested()) {
-            return
-        }
-
+    override fun visitKtFile(file: KtFile) {
+        if (bindingContext == BindingContext.EMPTY) return
+        super.visitKtFile(file)
```
schalkms, line 73:
```suggestion
        super.visitKtFile(file)
        if (bindingContext == BindingContext.EMPTY) return
```

schalkms: If there's no `bindingContext` available, the super method isn't called.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
```diff
@@ -91,25 +99,114 @@ class VarCouldBeVal(config: Config = Config.empty) : Rule(config) {
...
         override fun visitProperty(property: KtProperty) {
-            if (property.isVar) {
-                declarations.add(property)
-            }
             super.visitProperty(property)
+            if (property.isDeclarationCandidate()) {
+                declarationCandidates.add(property)
+            }
+
+            // Check for whether the initializer contains an object literal.
```
schalkms, line 115:
```suggestion
            // Check whether the initializer contains an object literal.
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
```diff
@@ -13,6 +13,82 @@ class VarCouldBeValSpec : Spek({
     val env: KotlinCoreEnvironment by memoized()
     val subject by memoized { VarCouldBeVal() }
 
+    describe("file-level declarations") {
+        it("does not report non-private variables") {
+            val code = """
+                var a = 1
+            """.trimIndent()
```
schalkms, line 20: nitpick: if you don't test for characters or reports at a specific position in the code snippet, it's optional to include the `trimIndent()` call.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
```diff
@@ -13,6 +13,82 @@ class VarCouldBeValSpec : Spek({
     val env: KotlinCoreEnvironment by memoized()
     val subject by memoized { VarCouldBeVal() }
 
+    describe("file-level declarations") {
+        it("does not report non-private variables") {
```
schalkms, line 17:
```suggestion
        it("does not report public variables") {
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
```diff
@@ -13,6 +13,82 @@ class VarCouldBeValSpec : Spek({
...
+                    a = 2
+                }
+            """
+
+            assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
+        }
+    }
+
+    describe("class-level declarations") {
+        it("does not report non-private variables in non-private classes") {
```
schalkms, line 51: I guess the following is a bit easier to read and grasp.

```suggestion
        it("does not report public variables in public classes") {
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
```diff
@@ -13,6 +13,82 @@ class VarCouldBeValSpec : Spek({
...
+            val code = """
+                class A {
+                    var a = 1
+                }
+            """.trimIndent()
+            assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
+        }
+
+        it("does not report non-private variables in non-private objects") {
```
schalkms, line 60: same here

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
```diff
@@ -191,8 +267,138 @@ class VarCouldBeValSpec : Spek({
                     }
                     o.optionEnabled = false
                 }
-           """
+            """
             assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
         }
+
+        context("should not report anonymous objects that escape") {
+            it("when initializing a variable directly") {
```
schalkms, line 275: minor: I think it might be beneficial to apply the same test description style as above. I guess it's easier to grasp the many different cases when reading and editing this rule in the future.
e.g.:

```Kotlin
context("anonymous objects that escape") {
    it("does not report when initializing a variable directly") {
    }

    it("does not report")...
```

schalkms (approved):
Looks good! 👍
Thank you very much for tackling this really rule improvement!

BraisGabin:
> ```
> [reports] at /home/runner/work/detekt/detekt/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt:78:5
> ```

Other option is to add an ingnoreAnnotated and ignore the params that had the `@param` annotation. For sure I'm talking about adding that configuration in our own configuration, not as a default one.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
BraisGabin: Maybe that's desirable, right? Faster rule.

I think that we should even move this checks earlier in the life cycle. But that's another topic.

BraisGabin (approved):
Nice one! The next RC will have a lot a juice! Thanks!!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
schalkms: In this case, you don't gain any performance benefit, because `KtFile`'s are usually not nested.

I agree with the following statement. In the best case, you don't even load the rule if it's annotated with the corresponding annotation. As far as I remember, there's an issue floating around, which discusses this topic. Anyways, as you said, this is out of scope for this PR.

> I think that we should even move this checks earlier in the life cycle. But that's another topic.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
severn-everett: `public` != `non-private` - there are `protected` and `internal` scopes as well. I didn't do it because I felt it'd be a bit excessive, but I could put a few more tests to cover the other scopes.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
severn-everett: It's cargo-culting from other files that I copied from when I started out working on this code base. I think it and the test wording show that there might need to be some more definitions in the contribution guide about style and things to add or avoid.

BraisGabin (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
severn-everett: My IDE automatically appends `.trimIndent()` when I create these string blocks. If that's a setting that's generally better left off, I'll disable it, but I'd rather keep things as-is if they don't cause any harm.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
```diff
@@ -10,19 +10,30 @@ import io.gitlab.arturbosch.detekt.api.Rule
...
+import org.jetbrains.kotlin.psi.psiUtil.isPrivate
+import org.jetbrains.kotlin.psi.psiUtil.lastBlockStatementOrThis
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
-
-private val unaryAssignmentOperators = setOf(KtTokens.MINUSMINUS, KtTokens.PLUSPLUS)
+import org.jetbrains.kotlin.util.containingNonLocalDeclaration
 
 /**
  * Reports var declarations (locally-scoped variables) that could be val, as they are not re-assigned.
```
chao2zhang, line 39:
```suggestion
 * Reports var declarations (both local variables and private class properties) that could be val, as they are not re-assigned.
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
```diff
@@ -91,25 +97,114 @@ class VarCouldBeVal(config: Config = Config.empty) : Rule(config) {
...
+                        .takeIf { it != returnedExpression }
+                        ?.let { evaluateReturnExpression(it) }
+                }
+            }
+        }
+
+        private fun KtProperty.isDeclarationCandidate(): Boolean {
+            return when {
+                !isVar -> false
+                isLocal || isPrivate() -> true
```
chao2zhang, line 194: It might be good to follow up with hanlding expect / actual properties.

chao2zhang (approved):

chao2zhang:
Thanks for the contribution!

Issue #4377, opened 2021-12-16
New rule: Function paramater should not be nullable


BraisGabin:
## Expected Behavior of the rule
Some times I found functions that allows a null parameter to then assert that it's not null. Those functions shouldn't accept a null parameter, they should force the caller to take the decission to force the non-null or don't cal it.

Non-compilant
```kotlin
fun foo(a: String?) {
  println(a!!)
}
```

This function should be declared like this:
```kotlin
fun foo(a: String) {
  println(a)
}
```

## Context
This kind of errors are added with the java to kotlin translator. And some times after a few refactor you can end up with code like this.

severn-everett:
I'll add this in when PR #4379 is integrated, as it looks like it falls within the same purview of that rule.

cortinico:
Closed by #4431

PR #4431, opened 2021-12-30
Extend CanBeNonNullable rule to check function params

severn-everett:
This addresses issue #4377.


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt
```diff
@@ -64,6 +78,14 @@ import org.jetbrains.kotlin.types.isNullable
  *     val a: Int
  *         get() = 5
  * }
+ *
+ * fun foo(a: Int?) {
+ *     val b = (a ?: 0) + 2
+ * }
```
BraisGabin, line 84: I wouldn't add this as compilant code. The rule for sure will not complain but, to me, this Elvis should be decided by the caller of the function. The function shouldn't accept `null`s

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
```diff
@@ -383,5 +383,107 @@ class CanBeNonNullableSpec : Spek({
...
+
+        context("evaluating nullable function parameters") {
+            context("reports when they are never treated as nullable") {
+                it("and use a double-bang de-nullifier") {
+                    val code = """
+                        fun foo(a: Int?) {
+                            val b = a!! + 2
+                        }
+                    """.trimIndent()
+                    Assertions.assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
```
schalkms, line 395:
```suggestion
                   assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
```diff
@@ -383,5 +383,107 @@ class CanBeNonNullableSpec : Spek({
...
+            it("does not report when they are accessed using a null-safe extension function") {
+                val code = """
+                    fun foo(a: List<String>?) {
+                        val b = a.orEmpty()
+                        val c = a!! + "TEST"
+                    }
+                """.trimIndent()
+                Assertions.assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
+            }
+        }
```
BraisGabin, line 487: All these cases seems code that should be flagged by this rule. If you use `!!` on the parameter is that you don't allow it to be `null` so it shouldn't be nullable.

And probably `checkNotNull` should be handled here too.

And this case should be flagged too:

```kotlin
fun foo(a: String?) {
  if (a! = null) {
    println(a)
  }
}
```

This function does nothing when the argument is `null` so there is not point to someone to call it with a null value.

schalkms:
Thanks for reviewing this! 🙂 @BraisGabin

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
severn-everett: Wouldn't flagging all usages of `!!` mean that the rule would essentially be overlapping `UnsafeCallOnNullableType`?

BraisGabin: Not really. There are cases where this doesn't apply:

The usage of `!!` is OK if the argument is not null but it has a nullable property.
Or if you use it over a class / top level property.

For sure this will reduce a lot the usages of `!!` and that's good. It will do it in a smarter way than just forbid all the usages of `!!`. It will make the usage of `!!` unnecessary in lots of cases.

BraisGabin (approved):
🎉 Thanks! The tests are really good!

severn-everett:
Still got a few more to appease the CodeCov gods 😁, plus that case with `override` functions that was uncovered in `BaseContext`.

severn-everett:
@BraisGabin I misread your comment about what to do when checking whether a param isn't null, so this is going to take a bit to rework; I'll let you know when it's ready to be evaluated again. On a side note, is there a way to access the PR labels? It'd be nice to just put a `WIP` label on this for situations like this.

BraisGabin:
You can move it back to draft

severn-everett:
@BraisGabin This should be good for review now.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
```diff
@@ -383,5 +383,448 @@ class CanBeNonNullableSpec : Spek({
...
+                        val code = """
+                            class A {
+                                val foo = "BAR"
+                            }
+                            
+                            fun fizz(aObj: A?): String? {
+                                return aObj?.foo
+                            }
+                        """.trimIndent()
+                        assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
```
BraisGabin, line 540: Same here

BraisGabin (commented):
Thanks for the hard work! I have a cople of comments more. Tell me what do you think.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/SplitPattern.kt
```diff
@@ -47,7 +51,9 @@ open class SplitPattern(
     /**
      * Tests if any part starts with the given [value]
      */
-    fun startWith(value: String?): Boolean = excludes.any { value?.startsWith(it) ?: false }
+    fun startWith(value: String?): Boolean {
+        return if (value != null) excludes.any(value::startsWith) else false
+    }
```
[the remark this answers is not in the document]
severn-everett: I think this is code from an older iteration and can probably be reverted.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
```diff
@@ -383,5 +383,448 @@ class CanBeNonNullableSpec : Spek({
...
+                context("in initializer") {
+                    it("does report when the safe-qualified expression is the only expression of the function") {
+                        val code = """
+                            class A {
+                                val foo = "BAR"
+                            }
+                            
+                            fun foo(a: A?) = a?.foo
+                        """.trimIndent()
+                        assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
```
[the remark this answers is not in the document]
severn-everett: My thought was that it'd move the null-check to outside of the function so that the function call could be skipped entirely - see the code in `UselessPostfixExpression` for how it'd look using the rule as it now is. However, you're right that the code in your example wouldn't be flagged, so it'd be better to revert this rule aspect and keep things as symmetrical as possible.

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ExceptionRaisedInUnexpectedLocation.kt
```diff
@@ -65,5 +65,5 @@ class ExceptionRaisedInUnexpectedLocation(config: Config = Config.empty) : Rule(
     private fun isPotentialMethod(function: KtNamedFunction) = methodNames.any { function.name == it }
 
     private fun hasThrowExpression(declaration: KtExpression?) =
-        declaration?.anyDescendantOfType<KtThrowExpression>() == true
+        declaration?.anyDescendantOfType<KtThrowExpression>() ?: false
```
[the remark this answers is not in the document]
severn-everett: This is probably code from an older iteration and should be reverted.

BraisGabin (approved):

BraisGabin:
To me this is good to go now.

cortinico (approved):

PR #4434, opened 2022-01-01
Add list of functions to skip in IgnoredReturnValue rule

artem-zinnatullin:
Hey Detekt team, I was configuring `IgnoredReturnValue` rule in our project and found it extremely useful — fixed few actual bugs! 

In fact, with this rule my old feature request filed to Kotlin team 6 years ago is now fulfilled heh https://youtrack.jetbrains.com/issue/KT-12719 

However the rule does find quite a lot of false-positives that are outside of our or Detekt control.

Thus I'm proposing a configuration option for the `IgnoredReturnValue` rule that will allow user to list fully qualified names of the functions return values of which they don't expect to be consumed — thus skipped by the rule.

```yaml
potential-bugs:
  IgnoredReturnValue:
    active: true
    restrictToAnnotatedMethods: false
    skipFunctions:
      - io.ktor.routing.post
      - io.ktor.routing.get
```

---

Examples:

In [Ktor](https://ktor.io/) some routing DSL methods return values (idk why, seems like bad design, but oh well) and DSL doesn't rely on them to be used:

```kotlin

embeddedServer(Netty, port = 8000) {
	routing {
		// Here get is a DSL configuration function, you never consume its result..
		get ("/") {
			call.respondText("Hello, world!")
		}
	}
}
```

Or some stdlib functions like `use`:

```kotlin
// Here use() returns a value but often you rely only on lambda..
file.outputStream().use { fos ->
        inputStream.use {
            it.copyTo(fos)
        }
}
```

and so on.

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
@@ -60,11 +60,11 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
     }
 
     @Configuration(
-        "List of fully-qualified function call signatures which should be ignored by this rule. " +
-            "Functions can be defined without full signature (i.e. `java.time.LocalDate.now`) which will ignore " +
-            "calls of all functions with this name or with full signature " +
-            "(i.e. `java.time.LocalDate.now(java.time.Clock)`) which would ignore only call " +
-            "with this concrete signature."
+        "List of function signatures which should be ignored by this rule. " +
+            "Specifying fully-qualified function signature with name only (i.e. `java.time.LocalDate.now`) will ignore " +
+            "all function calls matching the name. Specifying fully-qualified function signature with parameters" +
+            "(i.e. `java.time.LocalDate.now(java.time.Clock)`) will ignore only function calls matching the name " +
+            "and parameters exactly."
     )
     private val ignoreFunctionCall: List<FunctionMatcher> by config(emptyList<String>()) {
         it.map(FunctionMatcher::fromFunctionSignature)
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
@@ -59,7 +59,10 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
         it.map(String::simplePatternToRegex)
     }
 
-    @Configuration("List of fully-qualified function names that should be skipped by this check. Example: 'package.class.fun1'")
+    @Configuration(
+        "List of fully-qualified function names that should be skipped by this check. " +
+            "Example: 'package.class.fun1'"
+    )
     private val skipFunctions: List<String> by config(emptyList())
 
     @Suppress("ReturnCount")
@@ -95,6 +98,4 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
         val fqName = annotation.fqName?.asString() ?: return false
         return any { it.matches(fqName) }
     }
-
-
 }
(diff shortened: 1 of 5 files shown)


schalkms (approved):
This is a nice addition to this rule. I like it. Thanks for submitting!

artem-zinnatullin:
Thanks :)

chao2zhang (changes requested):
I will follow up on completing the docs, but we have `ignoreAnnotated` universal option added.
- [Changelog](https://github.com/detekt/detekt/blob/main/docs/pages/changelog%201.x.x.md#notable-changes)
- [Sample configuration guide](https://detekt.github.io/detekt/compose.html)

I am wondering if this usecase should be standardized into `ignoreAnnotated`. @BraisGabin what do you think?

artem-zinnatullin:
@chao2zhang I don't think `ignoreAnnotated` will work for cases I listed, functions in question are in 3rd-party code beyond my/yours control, thus they don't usually have annotations on them.

schalkms:
@artem-zinnatullin I think @chao2zhang meant the following comment in the changelog. `ignoreAnnotated` doesn't help in this scenario, but `ignoreFunction` should do the job. I also didn't know that detekt already supports that. 😉

> Similarly, we now offer also an ignoreFunction configuration key that you can use to suppress findings if inside a function with a given name - #4148

chao2zhang:
Good callout! I forgot about `ignoreFunction` even it is called out in the release note 😅. It should help in this case.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
```diff
@@ -58,6 +59,12 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
         it.map(String::simplePatternToRegex)
     }
 
+    @Configuration(
+        "List of fully-qualified function names that should be skipped by this check. " +
+            "Example: 'package.class.fun1'"
+    )
+    private val skipFunctions: List<String> by config(emptyList())
```
BraisGabin, line 66: I think that you should use `FunctionMatcher`. It provides more porwerful matching (you can match only by name or name + param types. Here a rule using it:

https://github.com/detekt/detekt/blob/ff7824d52518450e4dbcb69064aff35430a587ba/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt#L46-L58

BraisGabin (commented):
I think that `ignoreFunction` would not work on this scenario. That `Suppressor` ignores justthe issues finded on the **function declaration**. Here the issue is thrown in the function call so I think that it would not work.

But I agree that this feature could be handled by a `Suppressor` so it could be used on any rule. But some open questions:

1. Should we use the same `Suppressor` for both cases (issues in the definition and the call side)?
2. If we decide it should be in different `Suppressor`s how do we call this new one? `ignoreFunctionCall`? Naming is hard.

artem-zinnatullin:
@BraisGabin spot on suggestion with a FunctionMatcher and naming! 

I was wondering on distinguishing function overloads, I've changed the code to use the matcher and renamed the property to `ignoreFunctionCall` which is better in terms of consistency I think.

Regarding the `ignoreFunction` suggested above, from how I understand it:

>Similarly, we now offer also an `ignoreFunction` configuration key that you can use to suppress findings if inside a function with a given name

 I think it won't work, "inside a function with a given name" while I'm trying to ignore external function calls.

BraisGabin (approved):

BraisGabin:
@chao2zhang What do you think about this? It would be nice to have a `Suppressor` for this. But we can add it later without any breaking change.

cortinico (approved):

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
```diff
@@ -58,6 +59,17 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
         it.map(String::simplePatternToRegex)
     }
 
+    @Configuration(
+        "List of fully-qualified function call signatures which should be ignored by this rule. " +
+            "Functions can be defined without full signature (i.e. `java.time.LocalDate.now`) which will ignore " +
+            "calls of all functions with this name or with full signature " +
+            "(i.e. `java.time.LocalDate.now(java.time.Clock)`) which would ignore only call " +
+            "with this concrete signature."
```
chao2zhang, line 67:
```suggestion
        "List of function signatures which should be ignored by this rule. " +
            "Specifying fully-qualified function signature with name only (i.e. `java.time.LocalDate.now`) will ignore " +
            "all function calls matching the name. Specifying fully-qualified function signature with parameters" +
            "(i.e. `java.time.LocalDate.now(java.time.Clock)`) will ignore only function calls matching the name " +
            "and parameters exactly."
```

chao2zhang (approved):

PR #4436, opened 2022-01-02
Less restrictive OutdatedDocumentation rule and add UndocumentedPublicFunctionReturn rule

### LINKED ISSUE #4395: More in-depth "Undocumented*" rules with missing/invalid block tags

It would be nice if the following rules could go a little more in-depth by also checking for missing or invalid block tags.

### [`comments.UndocumentedPublicClass`](https://detekt.github.io/detekt/comments.html#undocumentedpublicclass)

- Report undocumented (constructor) parameter and type parameter. Each one of them should either have a corresponding `@param` tag in the class' KDoc *or* have its own KDoc. Also report `@param` blocks that do not correspond to any parameter.

```kotlin
/**
 * I am a documentation string
 */
class InvalidBecauseMissingParamDocs<T>(something: String)

/**
 * I am a documentation string
 *
 * @param foobar What the hell is this?
 * @param something My favorite parameter
 * @param T My generic parameter
 */
class ValidBecauseNoParameterNamedFoobar<T>(something: String)

/**
 * I am a documentation string
 *
 * @param something My favorite parameter
 * @param T My generic parameter
 */
class ValidBecauseInClassKDoc<T>(something: String)

// The following example does not apply to generic parameters, as they cannot have attached KDoc directly.
/**
 * I am a documentation string
 */
class ValidBecauseHasItsOwnKDoc(
    /**
     * My favorite parameter
     */
    something: String
)
```

### [`comments.UndocumentedPublicFunction`](https://detekt.github.io/detekt/comments.html#undocumentedpublicfunction)

- Report undocumented parameters and type parameters. Same as for class parameters basically.
  - Potential caveat: Only raise an error if the function itself declares a specific type parameter. Type parameters coming from the class containing the function should be ignored.

```kotlin
/**
 * This is some docs
 */
fun <T> invalidBecauseUndocumentedParameter(whatAmI: String) {
    TODO("Am I a cat?")
}

/**
 * This is some docs
 * 
 * @param whatAmI This is some more docs
 * @param wolf I ran out of ideas
 * @param T My generic type
 */
fun <T> invalidBecauseNoParameterNamedWolf(whatAmI: String) {
    TODO("Am I a cat?")
}

/**
 * This is some docs
 * 
 * @param whatAmI This is some more docs
 * @param T My generic type
 */
fun <T> validBecauseInFunctionKDoc(whatAmI: String) {
    TODO("Am I a cat?")
}

// (Same remark as for class generic type parameters)
/**
 * This is some docs
 */
fun validBecauseHasItsOwnKDoc(
    /**
     * This is some more docs
     */
    whatAmI: String
) {
    TODO("Am I a cat?")
}
```

- Report undocumented return value if the function does not return Unit nor Nothing

```kotlin
/**
 * I don't really do anything special
 */
fun validBecauseReturnsUnit() {
    println("""¯\_(ツ)_/¯""")
}

/**
 * I do a bit of a throw
 *
 * @throws NotImplementedError I'm not done yet
 */
fun validBecauseReturnsNothing(): Nothing = TODO("I throw things around")

/**
 * What do I return?
 */
fun invalidBecauseUndocumentedReturnValue(): Int = 42

/**
 * What do I return?
 * 
 * @return A pretty cool answer
 */
fun validBecauseDocumentedReturnValue(): Int = 42
```

- Report undocumented exceptions if the function reports nothing. I don't think this one will be able to do much other than report aa missing `@throws`/`@exception` tag.

```kotlin
/**
 * If you call this, I will literally throw something.
 */
fun invalidBecauseNothingReturnedAndNoExceptionDocumented(): Nothing = throw Exception("YEET!")

/**
 * If you call this, I will literally throw something.
 * 
 * @exception Exception I'll throw that.
 */
fun validBecauseExceptionTag(): Nothing = throw Exception("YEET!")

/**
 * If you call this, I will literally throw something.
 *
 * @throws Exception I'll throw that.
 */
fun validBecauseThrowsTag(): Nothing = throw Exception("YEET!")
```

- Report missing `@receiver` documentation for extension functions
  - This one may not be as necessary as the other ones, but may be a nice addition.

### [`comments.UndocumentedPublicProperty`](https://detekt.github.io/detekt/comments.html#undocumentedpublicproperty)

* Report missing documentation for property generic type parameters

### Remarks

These options should be optional -- not everyone wants to go *this* in-depth when it comes to documentation.

## Context

This would help tremendously for keeping a high standard of documentation quality for projects that want them -- I don't think any tool actually does any of these checks in Kotlin although they would not be *that* difficult to implement.

The main context is public libraries that wish to maintain high documentation standards. It is my personal belief that the Kotlin ecosystem has a recurring problem of lacking high quality KDoc, mostly because of the tooling not being that great. Adding these more in-depth checks to Detekt would help in that regard in my opinion.

I would also be okay with creating more rules instead of adding these constraints to the existing rules.

Relevant documentation:

* `@param <name>`: https://kotlinlang.org/docs/kotlin-doc.html#param-name
* `@return`: https://kotlinlang.org/docs/kotlin-doc.html#return
* `@receiver`: https://kotlinlang.org/docs/kotlin-doc.html#receiver
* `@throws <class>` and `@exception <class>`: https://kotlinlang.org/docs/kotlin-doc.html#throws-class-exception-class

utybo:
This PR:

- adds a flag that lets OutdatedDocumentation run even if no `@param` or `@property` tags are present
- adds an UndocumentedPublicFunctionReturn rule that reports missing `@return` tags for functions that return anything other than Unit. It also handles `Nothing` return types by allowing `@exception` or `@throws` to be used instead of `@return`.

This should fulfill everything I talked about in #4395 (as most of the functionality was already present in OutdatedDocumentation)

Apologies for this "two in one" PR, but the changes to OutdatedDocumentation are very small and I did not feel like making two PRs for such a small change

Fixes #4395 

### TODO

- [ ] Need to add the new rule to a ruleset
- [ ] Also, the change might also report private functions? Need to check this

### DIFF
--- a/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunctionReturn.kt
+++ b/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunctionReturn.kt
@@ -79,7 +79,6 @@ class UndocumentedPublicFunctionReturn(config: Config = Config.empty) : Rule(con
     private val throwsOrExceptionReturnTypes: List<String> by config(listOf("kotlin.Nothing"))
 
     override fun visitNamedFunction(function: KtNamedFunction) {
-        val ctx = bindingContext
         val doc = function.docComment
         if (bindingContext == BindingContext.EMPTY || doc == null) return
         val returnType = function.createTypeBindingForReturnType(bindingContext)?.type?.fqNameOrNull()?.toString()
--- a/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunctionReturn.kt
+++ b/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunctionReturn.kt
@@ -1,0 +1,116 @@
+package io.gitlab.arturbosch.detekt.rules.documentation
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Finding
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
+import io.gitlab.arturbosch.detekt.rules.fqNameOrNull
+import org.jetbrains.kotlin.kdoc.psi.api.KDoc
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.typeBinding.createTypeBindingForReturnType
+
+/**
+ * This rule reports functions that are documented yet miss documentation for their return type.
+ *
+ * There are two kinds of problems this rule reports:
+ *
+ * - Missing `@return` in documentation for functions that do not return Unit.
+ * - Missing `@return`, `@throws` or `@exception` documentation for functions that return Nothing.
+ *
+ * You can configure:
+ *
+ * - which types are considered to be Unit via the `nonDocumentedReturnTypes` configuration.
+ * - which types are considered to be Nothing via the `throwsOrExceptionReturnTypes` configuration.
+ *
+ * <noncompliant>
+ * /**
+ *  * My return type is not documented
+ *  */
+ *  fun foo(): String = /* ... */
+ *
+ *  /**
+ *   * My return type is documented
+ *   */
+ *  fun bar(): Nothing = /* ... */
+ * </noncompliant>
+ *
+ * <compliant>
+ * /**
+ *  * My return type is documented
+ *  *
+ *  * @return Something
+ *  */
+ * fun baz(): String = /* ... */
+ *
+ * fun okBecauseNotDocumented(): String = /* ... */
+ *
+ * /**
+ *  * I am nothing, so I can also document my return type with a throws/exception
+ *  *
+ *  * @throws ExceptionOne if something goes wrong
+ *  * @exception ExceptionTwo if something else goes wrong
+ *  */
+ * fun nothingish(): Nothing = /* ... */
+ * </compliant>
+ */
+@RequiresTypeResolution
+class UndocumentedPublicFunctionReturn(config: Config = Config.empty) : Rule(config) {
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Maintainability,
+        "Non-trivial return types should be documented.",
+        Debt.TEN_MINS
+    )
+
+    @Configuration("Return types that do not need to be documented (e.g. kotlin.Unit)")
+    private val nonDocumentedReturnTypes: List<String> by config(listOf("kotlin.Unit"))
+
+    @Configuration(
+        "Return types that can also be documented with @throws or @exception tags (e.g. kotlin.Nothing)"
+    )
+    private val throwsOrExceptionReturnTypes: List<String> by config(listOf("kotlin.Nothing"))
+
+    override fun visitNamedFunction(function: KtNamedFunction) {
+        val ctx = bindingContext
+        val doc = function.docComment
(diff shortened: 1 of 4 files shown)


BraisGabin:
You need to run `./gradlew generateDocumentation` and commit the changes that it make to make CI happy/happier

And I'd ask you to split this in two PRs. Because the `OutdatedDocumentation` change is easier to merge than the other so we can move those a different pace.

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunctionReturn.kt
```diff
@@ -0,0 +1,115 @@
...
+class UndocumentedPublicFunctionReturn(config: Config = Config.empty) : Rule(config) {
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Maintainability,
+        "Non-trivial return types should be documented.",
+        Debt.TEN_MINS
+    )
+
+    @Configuration("Return types that do not need to be documented (e.g. kotlin.Unit)")
+    private val nonDocumentedReturnTypes: List<String> by config(listOf("kotlin.Unit"))
```
BraisGabin, line 74: Shouldn't we include `kotlin.Nothing` here too? And which is the use case where other values than those two could be here? Do we need this configuration?

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunctionReturn.kt
```diff
@@ -0,0 +1,115 @@
...
+
+    @Configuration(
+        "Return types that can also be documented with @throws or @exception tags (e.g. kotlin.Nothing)"
+    )
+    private val throwsOrExceptionReturnTypes: List<String> by config(listOf("kotlin.Nothing"))
```
BraisGabin, line 79: Same here, what's the use case for using this? Each configuration adds complexity to the rule and it needs to be maintained. For that reason I'm asking for a good use case where to use this, other wise I think that it's better to remove it.

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicFunctionReturn.kt
utybo: I agree this isn't clear in the code, but the idea is that return types fall into three categories:

1. Ones that need to be documented by a `@return` tag.
2. Ones that do not need to be documented at all.
3. Ones that need to be documented by either a `@return`, `@throws` *or* `@exception` tag.

Unit falls into category 2 and Nothing falls into category 3, everything else falling in category 1. The configuration flags are mainly for flexibility around these rules, as I felt that hardcoding the exact types would be a bad idea. For example, it is possible that a project may have return types that are so common and possibly self-documented that it may not make sense to document the return type. Although I guess they could just suppress the rule there, I don't really know.

BraisGabin: If a function returns `Nothing` why should we document it with `@return`? To me the point 3 should force a `@throws` or `@exception`. And probably force that it never has a `@return`. But I think that I'm missing somethig here.

> The configuration flags are mainly for flexibility around these rules, as I felt that hardcoding the exact types would be a bad idea. For example, it is possible that a project may have return types that are so common and possibly self-documented that it may not make sense to document the return type. Although I guess they could just suppress the rule there, I don't really know.

I understand. It may have sense but as I said they introduce a complexity that maybe it's not worth the maintenance. I would vote to remove them.

utybo: It may be that a function that returns Nothing never returns because of an infinite loop. In that case, I feel like the best place to document such behavior is in a `@return` tag, although this is more of a side case that could also be solved by just suppressing the rule in those edge cases

BraisGabin: Yes, I see it as an edge case. And I think that a function that never ends should define it in its description because in genera is a not expected behaviour for a function.

If you can remove the configurations and this last thing I see this good to go.

cortinico:
Thanks for this PR

> Apologies for this "two in one" PR, but the changes to OutdatedDocumentation are very small and I did not feel like making two PRs for such a small change

I would actually advice to split this. The reason is that you're both adding a rule and tweaking another one. Should we revert the tweak to `OutdatedDocumentation`, we would have to manually revert the relevant chunks, and we could not use the auto-revert of GitHub.

Plus there will be a single entry in the changelog, which is harder to read for our users 👍

cortinico:
What's the status here?

PR #4437, opened 2022-01-02
Force that no one uses `Assertions.assertThat()`

BraisGabin:
Usually we need to add comments saying: "don't use `Assertions.assertThat`, use `assertThat()`". With this PR detekt will make that check for us.


schalkms (approved):
👍

PR #4445, opened 2022-01-03
Always update the baseline if it exist

### LINKED ISSUE #3736: Update baseline when there is no more issues
## Expected Behavior
If I remove all the issues in my code and I run `detektBaseline` the baseline file is removed or, at least, updated.

## Observed Behavior
The file is not updated nor removed.

## Context
I think that his issue was introduced in #3450

## Your Environment
* Version of detekt used: `1.16.0-RC2`

BraisGabin:
If you have a module with some issues in the baseline and you fix **all of them**, when you update the baseline detekt does nothing. This PR fixes that. If there is a baseline file, we always update it.

This PR also fixes a really annoying flaky test related with baseline.

Fix #3736

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFacade.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFacade.kt
@@ -23,5 +23,5 @@ class BaselineFacade {
         BaselineFormat().write(baseline, baselineFile)
     }
 
-    internal fun baselineExists(baseline: Path) = baseline.exists() && baseline.isFile()
+    private fun baselineExists(baseline: Path) = baseline.exists() && baseline.isFile()
 }
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFacade.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFacade.kt
@@ -9,8 +9,13 @@ import java.nio.file.Path
 
 class BaselineFacade {
 
-    fun transformResult(baselineFile: Path, result: Detektion): Detektion =
-        BaselineFilteredResult(result, Baseline.load(baselineFile))
+    fun transformResult(baselineFile: Path, result: Detektion): Detektion {
+        return if (baselineExists(baselineFile)) {
+            BaselineFilteredResult(result, Baseline.load(baselineFile))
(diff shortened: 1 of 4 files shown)


schalkms:
@BraisGabin shall we also explain the new behavior in the FAQ section on the homepage?
https://detekt.github.io/detekt/baseline.html

schalkms (approved):

BraisGabin:
I don't think so. There wasn't any explanation about when the baseline was created or not. And what this PR does have more sense than the previous behavior.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFacade.kt
```diff
@@ -9,19 +9,29 @@ import java.nio.file.Path
...
             Baseline.load(baselineFile).copy(currentIssues = ids)
         } else {
             Baseline(emptySet(), ids)
         }
-        baselineFile.parent?.let { Files.createDirectories(it) }
-        BaselineFormat().write(baseline, baselineFile)
+        if (baseline.isNotEmpty() || exists) {
+            baselineFile.parent?.let { Files.createDirectories(it) }
+            BaselineFormat().write(baseline, baselineFile)
+        }
```
chao2zhang, line 31: If the content stays the same, always writing to the file will cause the Detekt Gradle task to always not up-to-date.

Do you think it's worth it to not update the file when the content does not change?

BraisGabin: The content is always the same so it doesn't affect gradle and the `up-to-date`. But I'm going to do the check so we avoid an unnecessary write to disk.

BraisGabin: Done, I think that the code now is even better.

BraisGabin:
Is this a notable change? I feel it as a bug-fix.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/baseline/BaselineFacade.kt
chao2zhang: Ah nice. I had thought the file last access timestamp is considered in the input cache key calculation.

chao2zhang:
I think it is - There could be additional tooling that reads those baseline.xml and generates a customized report (My previous team does that)

PR #4447, opened 2022-01-03
Improve HTML report css

JavierSegoviaCordoba:
The colors are based on detekt branding

And... an image is better than 1000 words:

image


BraisGabin (approved):
Wow! What a change! Thanks!

schalkms (approved):
Great 👍

cortinico (approved):

cortinico:
Great stuff! :) We should also think about exporting it to a custom CSS file so users can customize it as they wish.

PR #4453, opened 2022-01-05
Add option for OutdatedDocumentation to allow param in constructor pr…

### LINKED ISSUE #4366: OutdatedDocumentation should allow us to use `@param` even when constructor properties have `val`

Following code:

```kotlin
/**
 * @param firstParam Documentation for the first param
 * @param secondParam Documentation for the first param
 */
class MyClass(val firstParam: String, val secondParam: String)
```

fails the `OutdatedDocumentation` rule. Rule expects that all `val` constructor parameters are marked as `@property` instead.

But there is really no benefit to using `@property` that I can see.

## Context

Main drawback of using `@property` is that there is no documentation when calling constructor of the class:

image

As opposed to `@param`:

image

However, even when using `@param`, IntellIJ still populates the property javadoc, so using `@param` actually gives us the best of both worlds:

image

Semi-related to the https://github.com/detekt/detekt/issues/4365

matejdro:
This fixes #4366

Ideally, all constructor parameters should be marked as `@property` (since they are properties), but currently IntelliJ does not display parameter docs for these: https://youtrack.jetbrains.com/issue/KTIJ-12948

as a workaround until this is fixed on Jetbrains side, we add a property to the OutdatedDocumentation that allows `@param` on constructor properties.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -66,6 +66,7 @@ comments:
     active: false
     matchTypeParameters: true
     matchDeclarationsOrder: true
+    allowParamOnConstructorProperties: false
   UndocumentedPublicClass:
     active: false
     excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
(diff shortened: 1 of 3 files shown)


cortinico (approved):
LGTM 👍

marschwar (approved):

Issue #4448, opened 2022-01-03
Add rule to use runCatching instead of kotlin.runCatching


gpeal:
## Expected Behavior
Per [KTIJ-11561](https://youtrack.jetbrains.com/issue/KTIJ-11561), Android Studio/IntelliJ will autocomplete `runCatching` to `kotlin.runCatching` which is inconsistent with other Kotlin APIs such as `run` which don't do that.

It would be great to have a rule that would enforce that it is used consistently as just `runCatching` across the codebase.

BraisGabin:
🤔 This should be work for `ForbiddenMethodCall`. But I'm not sure if it can handle something like that right now... I'm going to check.

BraisGabin:
As I was expecting `ForbiddenMethodCall` can't handle it right now this case. For it both functions "are the same".

But the problem is bigger than that. Right now `ForbiddenMethodCall` doesn't suport:
- functions with lambda parameters (if you want to specify the parameters)
- functions with generics (if you want to specify the parameters)
- extension functions

`ForbiddenMethodCall` needs to support those three to be able to handle this case. I have a draft for the first case.

BraisGabin:
🤔 the only way that I can think of handle extension functions with the current "function signature" is to treat the receiver type as the "first paramenter" of the function. This will not create any problem because the compiler doesn't allow you to define this two functions

```kotlin
fun String.hello()
fun hello(world: String)
```

because they have the same JVM signature. It is the only way I can think without adding a breaking change or deprecate the current function signature and create a new one.

In terms of UX it would be really hacky because it's difficult to think about: `fun String.hello()` as `hello(kotlin.String)` but it would work.

gpeal:
@BraisGabin That's great! I tried adding a separate ForbiddenMethodCall for an extension function yesterday and couldn't figure out why it wasn't working!

PR #4458, opened 2022-01-06
Match functions signatures with lambdas on it

BraisGabin:
Related with #4448 but it doesn't fix it yet. The next step is to support extension functions.

*Note:* If you check the tests you will see that the signature `foo(() -> kotlin.String)` matches a functione like this one: `fun foo(a: () -> Unit)`. That's because we don't need to distinguish between this two functions:

```kotlin
fun foo(a: () -> Unit)
fun foo(a: () -> String)
```

The compiler doesn't allow to define both at the same time.

If we want to distinguish between those we would introduce a breaking change because until now this signature: `foo(List)` was matching `fun foo(a: List<String>)` but if we enforce to specify the generics that type should be changed.

For those two reasons combined I think that it's better to ignore the generics.

### DIFF
--- a/detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/FunctionMatcher.kt
+++ b/detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/FunctionMatcher.kt
@@ -58,21 +58,77 @@ sealed class FunctionMatcher {
 
     companion object {
         fun fromFunctionSignature(methodSignature: String): FunctionMatcher {
-            val tokens = methodSignature.split("(", ")")
-                .map { it.trim() }
+            @Suppress("TooGenericExceptionCaught", "UnsafeCallOnNullableType")
+            try {
+                val result = functionSignatureRegex.matchEntire(methodSignature)!!
 
-            val methodName = tokens.first().replace("`", "")
-            val params = if (tokens.size > 1) {
-                tokens[1].split(",").map { it.trim() }.filter { it.isNotBlank() }
-            } else {
-                null
-            }
+                val methodName = result.groups[1]!!.value.replace("`", "")
+                val params = result.groups[2]?.value?.splitParams()
+                    ?.map { changeIfLambda(it) ?: it }
 
-            return if (params == null) {
-                NameOnly(methodName)
-            } else {
-                WithParameters(methodName, params)
+                return if (params == null) {
+                    NameOnly(methodName)
+                } else {
+                    WithParameters(methodName, params)
+                }
+            } catch (ex: Exception) {
+                throw IllegalStateException("$methodSignature doesn't match a method signature", ex)
             }
         }
     }
 }
+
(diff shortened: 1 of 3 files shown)


detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/FunctionMatcher.kt
```diff
@@ -58,21 +58,77 @@ sealed class FunctionMatcher {
...
+            result.setLength(0)
+        } else {
+            if (c == '(') nestingLevel++
+            if (c == ')') nestingLevel--
+            check(nestingLevel >= 0)
+            result.append(c)
+        }
+    }
+    split.add(result.toString())
+    return split.map { it.trim() }.filter { it.isNotBlank() }
```
chao2zhang, line 98: Is `it.isNotBlank()` necessary? Without verifying with actual test code, I assume they are only possible at Line 97. I believe the `isNotBlank()` and `trim()` should be applied when `append()` is invoked, which avoids creating more lists.

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/FunctionMatcher.kt
```diff
@@ -58,21 +58,77 @@ sealed class FunctionMatcher {
...
-                NameOnly(methodName)
+private fun splitLambda(param: String): Pair<String, String>? {
+    if (!param.startsWith("(")) return null
+
+    var nestingLevel = 0
+    val paramsRaw = StringBuilder()
+    val returnValue = StringBuilder()
+
+    param.toCharArray()
+        .drop(1)
```
chao2zhang, line 116: This is quite tricky! I think the code works, but the nestingLevel after dropping the first `(` should be 1. Would you mind adding some comments or fixing the nestingLevel value?

chao2zhang (approved):
Thanks for expanding the supports for Function signatures! My comments were not blocking, they were primarily documenting my time spent on the confusion.

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/FunctionMatcher.kt
BraisGabin: I agree, this code is tricky. Any idea to make it easier to read is more than welcome.

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/FunctionMatcher.kt
BraisGabin: Fixed, what do you think now?

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/FunctionMatcher.kt
BraisGabin: Comment added

PR #4459, opened 2022-01-06
Match extension functions

BraisGabin:
More context here: https://github.com/detekt/detekt/issues/4448#issuecomment-1006521371 I'm implementing the no breaking option. It isn't nice but it works.

Related with #4448 but this PR doesn't fix the issue yet.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
@@ -48,7 +48,9 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
             "Methods can be defined without full signature (i.e. `java.time.LocalDate.now`) which will report " +
             "calls of all methods with this name or with full signature " +
             "(i.e. `java.time.LocalDate(java.time.Clock)`) which would report only call " +
-            "with this concrete signature."
+            "with this concrete signature. If you want to forbid an extension function like" +
+            "`fun String.hello(a: Int)` you should add the receiver parameter as the first parameter like this: " +
+            "`hello(kotlin.String, kotlin.Int)`"
     )
     private val methods: List<FunctionMatcher> by config(
         listOf(
--- a/detekt-tooling/src/test/kotlin/io/github/detekt/tooling/api/FunctionMatcherSpec.kt
+++ b/detekt-tooling/src/test/kotlin/io/github/detekt/tooling/api/FunctionMatcherSpec.kt
@@ -186,6 +186,36 @@ class FunctionMatcherSpec(private val env: KotlinCoreEnvironment) {
             val methodSignature = FunctionMatcher.fromFunctionSignature("toString(String)")
             assertThat(methodSignature.match(function, bindingContext)).isEqualTo(result)
         }
+
+        @DisplayName("When lambdas foo(() -> kotlin.String)")
+        @ParameterizedTest(name = "in case {0} it return {1}")
+        @CsvSource(
+            "fun foo(a: () -> String),          true",
+            "fun foo(a: () -> Unit),            true",
+            "fun foo(a: (String) -> String),    false",
+            "fun foo(a: (String) -> Unit),      false",
+            "fun foo(a: (Int) -> Unit),         false",
+        )
+        fun `When foo(() - kotlin#String)`(code: String, result: Boolean) {
+            val (function, bindingContext) = buildKtFunction(env, code)
+            val methodSignature = FunctionMatcher.fromFunctionSignature("foo(() -> kotlin.String)")
+            assertThat(methodSignature.match(function, bindingContext)).isEqualTo(result)
+        }
+
+        @DisplayName("When lambdas foo((kotlin.String) -> Unit)")
+        @ParameterizedTest(name = "in case {0} it return {1}")
+        @CsvSource(
+            "fun foo(a: () -> String),          false",
+            "fun foo(a: () -> Unit),            false",
+            "fun foo(a: (String) -> String),    true",
+            "fun foo(a: (String) -> Unit),      true",
+            "fun foo(a: (Int) -> Unit),         true",
+        )
+        fun `When foo((kotlin#String) - Unit)`(code: String, result: Boolean) {
+            val (function, bindingContext) = buildKtFunction(env, code)
+            val methodSignature = FunctionMatcher.fromFunctionSignature("foo((kotlin.String) -> Unit)")
+            assertThat(methodSignature.match(function, bindingContext)).isEqualTo(result)
+        }
     }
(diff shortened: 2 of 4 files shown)


BraisGabin:
I have no idea what's going on here: https://github.com/detekt/detekt/runs/4726908840?check_suite_focus=true#step:4:499

I didn't change that class at all. Well, I did change it, but one year ago: 04/01/202**1**

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
```diff
@@ -48,7 +48,9 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
             "Methods can be defined without full signature (i.e. `java.time.LocalDate.now`) which will report " +
             "calls of all methods with this name or with full signature " +
             "(i.e. `java.time.LocalDate(java.time.Clock)`) which would report only call " +
-            "with this concrete signature."
+            "with this concrete signature. If you want to forbid an extension function like" +
+            "`fun String.hello(a: Int)` you should add the receiver parameter as the first parameter like this: " +
```
chao2zhang, line 52: This still feels like a stop gap solution to me. When we disable type resolution, and we only want to forbid extension methods like `String.hello(a: Int)`, we would actually report both extension function `String.hello(a: Int)` and normal function `hello(s: String, a: Int)`.

Additionally, detekt should be compatible with other kotlin platforms.

BraisGabin: > This still feels like a stop gap solution to me.

Agree, I can't think of any other way to fix this without adding breaking changes to the syntax. And, even adding breaking changes, I can only think of more complex syntax, I can't think on any elegant way to fix this. Any idea in this regard is more than welcome.

> When we disable type resolution, and we only want to forbid extension methods like `String.hello(a: Int)`, we would actually report both extension function `String.hello(a: Int)` and normal function `hello(s: String, a: Int)`.

This rule (and feature in general) only works with Type Solving so this is not a real issue.

> Additionally, detekt should be compatible with other kotlin platforms.

Agree. Right now it's noy an issue because we only support type solving for JVM. But in the future maybe in js you could have both functions and that would be a problem.

The biggest issue to me is that it's not intuitive for a user how to use this feature. And I think that it should. But as I said before to handle this case properly we need to add a breaking change in the syntax. And, even then, I don't know how to define the function signature in an easy and elegant way.

cortinico: > The biggest issue to me is that it's not intuitive for a user how to use this feature.

I think that's the major point here. TR being a requirement is not a problem, as if you want to use an advanced capability, you need to opt-in to TR.

> But in the future maybe in js you could have both functions and that would be a problem.

It's already a concern, as the compiler plugin runs with TR on all the platforms.

> But as I said before to handle this case properly we need to add a breaking change in the syntax. And, even then, I don't know how to define the function signature in an easy and elegant way.

What would be your suggested way?

cortinico: > Other option is to use the syntax of `apiDump`:

I was actually thinking about something like that. I know that is not super immediate to grasp. But at the same time it comes from the bytecode representation of a method, and it's sort of "consistent" with the rest of the tools in the ecosystem

BraisGabin: In that one the issues raised by @chao2zhang applies too:

- It's not easy to write
- Only works for JVM (js doesn't need a `ResourcesKt` for the top level functions)
- An extension functions has it's receiver as the first parameter (that's what I'm doing here)

I agree that using something standard would be great but I don't think that syntax would work for us. And I don't know of any other standard one...

BraisGabin: How can we unblock this? To me we should think if this change improves of make worst this rule. I think that it's better than don't support extension functions. An example would like to forbid the usage of `androidx.activity.compose.setContent` in favor of my our own `setContent` and right now it's impossible and #4448 is other reason.

We can open an issue to talk about which should be the function signature syntax that we should use and then we can implement it. And once we take a decission we can implement it.

cortinico: Yup I'm in for merging this. I think having the extension function receiver modelled as first parameter is a good compromise and aligns to the approach taken by other tools as well.

BraisGabin:
Can we take a decission here?

chao2zhang:
Would this work across all Kotlin Multiplatform? I would vote for merge if this is not particularly tailored to JVM

BraisGabin:
> Would this work across all Kotlin Multiplatform? I would vote for merge if this is not particularly tailored to JVM

Yes, it should work for all of them.

cortinico:
> I think having the extension function receiver modelled as first parameter is a good compromise and aligns to the approach taken by other tools as well.

☝️ That's my latest opinion. I'm in for merging this

detekt-tooling/src/test/kotlin/io/github/detekt/tooling/api/FunctionMatcherSpec.kt
```diff
@@ -186,6 +186,66 @@ class FunctionMatcherSpec(private val env: KotlinCoreEnvironment) {
...
+        }
+
+        @DisplayName("When extension functions foo(kotlin.String)")
+        @ParameterizedTest(name = "in case {0} it return {1}")
+        @CsvSource(
+            "fun String.foo(),              true",
+            "fun foo(a: String),            true",
+            "fun Int.foo(),                 false",
+            "fun String.foo(a: Int),        false",
+            "'fun foo(a: String, ba: Int)', false",
```
chao2zhang, line 227: Are those single quotes around `fun foo(a: String, ba: Int)` required?

chao2zhang (approved):

detekt-tooling/src/test/kotlin/io/github/detekt/tooling/api/FunctionMatcherSpec.kt
BraisGabin: Yes they are. This string is parsed as a csv. And this function has two parameters so it needs a `, `. So, I need the single quotes.

marschwar: you could use a different separator though *(Random guy yelling from the side line)* :wink:

PR #4460, opened 2022-01-06
Function generics matcher

### LINKED ISSUE #4448: Add rule to use runCatching instead of kotlin.runCatching

BraisGabin:
Match generic names in signature functions.

Now ForbiddenMethod call allows to forbid `kotlin.runCatching {}` but allowing `runCatching {}`. 

closes #4448

### DIFF
--- a/detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/FunctionMatcher.kt
+++ b/detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/FunctionMatcher.kt
@@ -5,6 +5,8 @@ import org.jetbrains.kotlin.descriptors.CallableDescriptor
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.types.typeUtil.isTypeParameter
 
 sealed class FunctionMatcher {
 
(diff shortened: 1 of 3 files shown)


detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/FunctionMatcher.kt
```diff
@@ -33,11 +35,12 @@ sealed class FunctionMatcher {
         private val parameters: List<String>
     ) : FunctionMatcher() {
         override fun match(callableDescriptor: CallableDescriptor): Boolean {
-            if (callableDescriptor.fqNameOrNull()?.asString() != fullyQualifiedName) return false
+            val descriptor = callableDescriptor.original
```
cortinico, line 38: Q: What is `.original` doing here? Why is it needed?

BraisGabin: That get's the function definition instead of the function call. If I request the type of the parameter of `list("hello")` it will say `String` but if I ask for the type of the `original` I get `T`.

cortinico: Gotcha 👍  Thanks for clarifying

BraisGabin:
I would like to merge this one in 1.20 too if possible. This PR was waiting for #4459 to be merged because it depends on that code.

cortinico (approved):

cortinico:
I'm fine adding this, but we need some follow-up docs for this to the website

BraisGabin:
🤔 I don't think we need to add new documentation here. The current documentation matchs what this matcher does. This one is more a "bug-fix" than a new feature. I mean, for sure it is a new feature but for a user it could seems like the same. The PR that introduced real changes was #4459 and I added the documentation directly in it. https://github.com/detekt/detekt/pull/4459/files#diff-76703cf8efe31f0c7e06926568e0d1e248eb82cb63bd40b0b37698dc35bed70fR52

Issue #4463, opened 2022-01-06
Have a reporter that combines the benefits of FindingsReporter and LiteFindintsReporter


gfreivasc:
It would be good to have a reporter that combines the easy to navigate layout structure of `FindingsReporter` with the extra details found on `LiteFindingsReporter`.

## Expected Behavior

```kotlin
fun main() {
println("Hello World!")
}
```

It would be nice to have the output being as follows:

```
formatting - 5min debt
        Indentation - [Unexpected indentation (0) (should be 4)] at /Users/gabrielfv/Projects/detekt/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt:23:1
```

## Current Behavior

For `FindingsReporter`

```
formatting - 5min debt
        Indentation - [] at /Users/gabrielfv/Projects/detekt/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt:23:1
```

Which omits details about the rule

For `LiteFindingsReporter`

```
/Users/gabrielfv/Projects/detekt/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt:23:1: Unexpected indentation (0) (should be 4) [Indentation]
```

Which displays the information, but in a rather messy format, especially on large reports.

## Context

Based on #4446.

cortinico:
Fixed by #4464

Issue #4446, opened 2022-01-03
Formatting: Add more details from KtLint


hfhbd:
## File
```kt
fun main() {
println("Hello World!")
}
```
## Expected Behavior

```
formatting - 5min debt
	Indentation - [Unexpected indentation (0) (should be 4)] at /Users/philipwedemann/Downloads/ktlint/src/main/kotlin/Main.kt:2:1
```
## Current Behavior

```
formatting - 5min debt
	Indentation - [] at /Users/philipwedemann/Downloads/ktlint/src/main/kotlin/Main.kt:2:1
```
## Context
It would be nice, if detekt could provide more details why this line was not indented correctly.
In complex classes/cases, finding the correct indention is hard without any details.

Sample project: [SampleProject.zip](https://github.com/detekt/detekt/files/7802405/ktlint.zip)

Run `./gradlew detekt` and `./gradlew ktlintCheck` to compare its outputs.

BraisGabin:
The reporter is working as expected. And yes, the expected behaviour is completely useless.

For example, this is what it prints when I creare a empty function named `a`

```
empty-blocks - 5min debt
        EmptyFunctionBlock - [a] at /Users/brais.gabin/Downloads/ktlint/src/main/kotlin/Main.kt:5:9
```

We need to change our default console-report from `FindingsReport` to `LiteFindingsReport`.

The output of that console-report is:

```
/Users/brais.gabin/Downloads/ktlint/src/main/kotlin/Main.kt:5:9: This empty block of code can be removed. [EmptyFunctionBlock]
/Users/brais.gabin/Downloads/ktlint/src/main/kotlin/Main.kt:2:1: Unexpected indentation (0) (should be 4) [Indentation]
```

You can enable now in your configuration, but I'm going to make it the default value for `1.20.0`

gfreivasc:
I find that both reporters have their advantages. I like how the current default reporter lays out the issues in a way they're quick to spot on the console output (say for example on pipeline runs). However there's great value in describing the issue in cases where the rule name might not be too helpful. Does it make sense to have somewhat of a hybrid solution similar to the one suggested here?

BraisGabin:
It is possible to create your own report of you want. I would say that we should only add a new one of the is a high demand for it. So, sure, you can open the issue and see if there is that need.

PR #4464, opened 2022-01-07
Adds issue details to findings on FindingsReport and FileBasedFindingsReporter

gfreivasc:
Addresses #4463

Changes `FindingsReport` to print with the current structure, but with `messageOrDescription() + location.compact()` instead of `entity.compact()` as the message gives more details and usually also mentions the entity. It tries to combine that layout with the degree of information found on `LiteFindingsReport`.

Message changes also apply to `FileBasedFindingsReporter`.

### Example

```kotlin
fun main() {
   println( "Hello World!")
    if(1> 2)
        println("None")
}
```

#### `LiteFindingsReport`

```
/Users/gabrielfv/Projects/detekt/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt:24:7: Missing spacing after "if" [SpacingAroundKeyword]
/Users/gabrielfv/Projects/detekt/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt:23:1: Unexpected indentation (3) (should be 4) [Indentation]
/Users/gabrielfv/Projects/detekt/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt:23:12: Unexpected spacing after "(" [SpacingAroundParens]
/Users/gabrielfv/Projects/detekt/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt:25:9: Missing { ... } [MultiLineIfElse]
```

#### Current `FindingsReport`

```
formatting - 20min debt
	SpacingAroundKeyword - [] at /Users/gabrielfv/Projects/detekt/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt:24:7
	Indentation - [] at /Users/gabrielfv/Projects/detekt/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt:23:1
	SpacingAroundParens - [] at /Users/gabrielfv/Projects/detekt/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt:23:12
	MultiLineIfElse - [] at /Users/gabrielfv/Projects/detekt/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt:25:9

Overall debt: 20min
```

#### Modified `FindingsReport`

```
formatting - 25min debt
        MultiLineIfElse - [Missing { ... }] at /Users/gabrielfv/Projects/detekt/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/ConfigPrinter.kt:18:10
        SpacingAroundKeyword - [Missing spacing after "if"] at /Users/gabrielfv/Projects/detekt/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/ConfigPrinter.kt:17:8
        SpacingAroundParens - [Unexpected spacing after "("] at /Users/gabrielfv/Projects/detekt/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/ConfigPrinter.kt:16:13
        Indentation - [Unexpected indentation (5) (should be 4)] at /Users/gabrielfv/Projects/detekt/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/ConfigPrinter.kt:17:1
        Indentation - [Unexpected indentation (9) (should be 8)] at /Users/gabrielfv/Projects/detekt/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/ConfigPrinter.kt:18:1

Overall debt: 25min
```

#### Observations

These changes make it so that `Entity::compact()` method is no longer called internally. I chose to keep it as it's published API.

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt
@@ -10,6 +10,7 @@ import io.gitlab.arturbosch.detekt.api.Debt
 import io.gitlab.arturbosch.detekt.api.Detektion
 import io.gitlab.arturbosch.detekt.api.Finding
 import io.gitlab.arturbosch.detekt.api.RuleSetId
+import io.gitlab.arturbosch.detekt.api.ThresholdedCodeSmell
 
 internal fun defaultReportMapping(reportId: String) = when (reportId) {
     TxtOutputReport::class.java.simpleName -> "txt"
@@ -30,7 +31,7 @@ internal fun printFindings(findings: Map<String, List<Finding>>): String {
             append("$key - $debt debt\n")
             issues.forEach {
                 append("\t")
-                append(it.compact().yellow())
+                append(it.detailed().yellow())
                 append("\n")
             }
         }
@@ -38,6 +39,11 @@ internal fun printFindings(findings: Map<String, List<Finding>>): String {
         append("\nOverall debt: $overallDebt\n")
         toString()
     }
+}
+
+private fun Finding.detailed() = when (this) {
+    is ThresholdedCodeSmell -> "$id - $metric - [${messageOrDescription()}] at ${location.compact()}"
+    else -> "$id - [${messageOrDescription()}] at ${location.compact()}"
 }
 
 const val BUILD = "build"
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt
@@ -11,6 +11,7 @@ import io.gitlab.arturbosch.detekt.api.Detektion
 import io.gitlab.arturbosch.detekt.api.Finding
 import io.gitlab.arturbosch.detekt.api.RuleSetId
 import io.gitlab.arturbosch.detekt.api.ThresholdedCodeSmell
+import kotlin.text.Typography.ellipsis
 
 internal fun defaultReportMapping(reportId: String) = when (reportId) {
     TxtOutputReport::class.java.simpleName -> "txt"
@@ -41,16 +42,13 @@ internal fun printFindings(findings: Map<String, List<Finding>>): String {
     }
 }
 
-private fun Finding.detailed() = when (this) {
-    is ThresholdedCodeSmell -> "$id - $metric - [${messageOrDescription()}] at ${location.compact()}"
-    else -> "$id - [${messageOrDescription()}] at ${location.compact()}"
-}
-
 const val BUILD = "build"
 const val EXCLUDE_CORRECTABLE = "excludeCorrectable"
 
 const val DETEKT_OUTPUT_REPORT_PATHS_KEY = "detekt.output.report.paths.key"
 const val DETEKT_OUTPUT_REPORT_BASE_PATH_KEY = "detekt.output.report.base.path"
+
+private const val REPORT_MESSAGE_SIZE_LIMIT = 80
 
 fun Config.excludeCorrectable(): Boolean = subConfig(BUILD).valueOrDefault(EXCLUDE_CORRECTABLE, false)
 
@@ -73,4 +71,19 @@ fun Detektion.filterAutoCorrectedIssues(config: Config): Map<RuleSetId, List<Fin
         filteredFindings[ruleSetId] = newFindingsList
     }
     return filteredFindings
+}
+
+private fun Finding.truncatedMessage(): String {
+    val message = messageOrDescription()
+        .replace(Regex("\\s+"), " ")
+        .trim()
+    return when {
+        message.length > REPORT_MESSAGE_SIZE_LIMIT -> "${message.take(REPORT_MESSAGE_SIZE_LIMIT)}($ellipsis)"
+        else -> message
+    }
+}
+
+private fun Finding.detailed(): String = when (this) {
+    is ThresholdedCodeSmell -> "$id - $metric - [${truncatedMessage()}] at ${location.compact()}"
+    else -> "$id - [${truncatedMessage()}] at ${location.compact()}"
 }
--- a/detekt-report-txt/src/test/kotlin/io/github/detekt/report/txt/TxtOutputReportSpec.kt
+++ b/detekt-report-txt/src/test/kotlin/io/github/detekt/report/txt/TxtOutputReportSpec.kt
@@ -23,7 +23,7 @@ class TxtOutputReportSpec {
         fun `renders one`() {
             val report = TxtOutputReport()
             val detektion = TestDetektion(createFinding())
-            val renderedText = "TestSmell - [TestMessage] at TestFile.kt:1:1 - Signature=TestEntitySignature\n"
+            val renderedText = "TestSmell - [TestEntity] at TestFile.kt:1:1 - Signature=TestEntitySignature\n"
             assertThat(report.render(detektion)).isEqualTo(renderedText)
         }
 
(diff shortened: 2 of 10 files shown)


cortinico:
Thanks for the PR @gfreivasc 👍 
I'm just a bit worried that we will have `DetailedFindingsReport`, `LiteFindingsReport` and `FindingsReport` so it would be harder for users to understand what are the differences.

Shouldn't we just extend `FindingsReport` then?

gfreivasc:
While I personally don't think that's too much trouble, I also see that `LiteFindingsReporter` was also a recent addition. The reason I didn't initially consider changing `FindingsReporter` itself is because according to #4446 it's currently working as intended.

If you think we're good to make those changes to `FindingsReporter`, I could change my PR no problem.

BraisGabin:
I agree with @cortinico. I also vote to change `FindingsReporter`.

> Edit: Just in case we would move forward changing `FindingsReporter`, would we also want to change the format of `FileBasedFindingsReport`?

Yes, I think that it would be good to keep both aligned.

gfreivasc:
Changed the PR so now it changes `FindingsReporter` and `FileBasedFindingsReporter` message format. I'll be looking for documentation references that would also require update.

BraisGabin (dismissed):
I never understand why `compact` is part of the `Finding` api... I think that we should deprecate it. But that's another topic.

cortinico:
@gfreivasc Can you update the PR title to reflect the current diff?

gfreivasc:
@cortinico changed the message

gfreivasc:
@cortinico Had to fix tests from modules I had not run the tests for. They're all passing now.

docs/pages/reporting.md
```diff
@@ -17,11 +17,11 @@ Similar to the console output, each line of the txt output represents a finding
...
-NoUnusedImports - [] at /user/home/detekt/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt:9:1 - Signature=io.gitlab.arturbosch.detekt.DetektPlugin.kt:9
-NoUnusedImports - [] at /user/home/detekt/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt:10:1 - Signature=io.gitlab.arturbosch.detekt.DetektPlugin.kt:10
-NoConsecutiveBlankLines - [] at /user/home/detekt/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt:86:1 - Signature=io.gitlab.arturbosch.detekt.DetektPlugin.kt:86
-UnusedPrivateMember - [registerDetektJvmTasks] at /user/home/detekt/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt:17:5 - Signature=DetektPlugin.kt$DetektPlugin$private fun Project.registerDetektJvmTasks(extension: DetektExtension)
+EmptyFunctionBlock - [This empty block of code can be removed.] at /user/home/detekt/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt:14:42 - Signature=DetektPlugin.kt$DetektPlugin${ }
```
cortinico, line 20: One last question: are there scenario where this text can either:
1. Be too long
2. Contain `\n`?

Askign as if so, we should escape it as it would screw up the final reporting.

gfreivasc: As far as I'm aware of, there's no checks or anything to prevent these to be the case. These scenarios could also affect `LiteFindingsReport`. We could run a replace `.replace('\n', ' ')` for the second case but I can't think of what could be done in the first one. Truncate and ellipsis? What could be a threshold?

detekt-report-txt/src/test/kotlin/io/github/detekt/report/txt/TxtOutputReportSpec.kt
```diff
@@ -32,9 +32,9 @@ class TxtOutputReportSpec : Spek({
                 createFinding(ruleName = "TestSmellC")
             )
             val renderedText = """
-                TestSmellA - [TestEntity] at TestFile.kt:1:1 - Signature=TestEntitySignature
-                TestSmellB - [TestEntity] at TestFile.kt:1:1 - Signature=TestEntitySignature
-                TestSmellC - [TestEntity] at TestFile.kt:1:1 - Signature=TestEntitySignature
+                TestSmellA - [TestMessage] at TestFile.kt:1:1 - Signature=TestEntitySignature
+                TestSmellB - [TestMessage] at TestFile.kt:1:1 - Signature=TestEntitySignature
+                TestSmellC - [TestMessage] at TestFile.kt:1:1 - Signature=TestEntitySignature
```
BraisGabin, line 37: I'm not sure if we should change the txt output too... I'm fine adding behaviour changes in the console report. But I don't feel that confortable changing the file reports. People can be using those structures and have their own parsers so we should keep those.

BraisGabin (commented):
I approved this PR because I thought that the only report using `compact()` was FindingsReport but I just saw that it also affect the file reports so I don't think we should change the return of `compact()`. We should find another way to implement this.

gfreivasc:
@BraisGabin your comment makes total sense and I should've found another approach rather than applying these changes elsewhere. Addressed them in the last comments. Will work on the other suggestions yet today.

BraisGabin:
Could we don't change `Finding` at all? I mean, we would lose the "metrics" from the `ThresholdedCodeSmell` but I think that's ok. The message of the `Finding` should point out those metrics anyway. And if it doesn't that's a "bug" and we should change it to follow this: https://github.com/detekt/detekt/blob/main/.github/CONTRIBUTING.md#rule-descriptions

gfreivasc:
So in your opinion we should fallback to having a simple extension method similar to the one we had with the previous iteration of `DetailedFindingsReport`? We could keep metrics that way:

```kotlin
fun Findings.detailed() = when (this) {
    is ThresholdedCodeSmell -> "$id - $metric - [${messageOrDescription()}] at ${location.compact()}"
    else -> "$id - [${messageOrDescription()}] at ${location.compact()}"
}
```

This would eliminate API impact.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt
```diff
@@ -68,3 +72,18 @@ fun Detektion.filterAutoCorrectedIssues(config: Config): Map<RuleSetId, List<Fin
...
+
+private fun Finding.truncatedMessage(): String {
+    val message = messageOrDescription()
+        .replace(Regex("\\s+"), " ")
+        .trim()
+    return when {
+        message.length > REPORT_MESSAGE_SIZE_LIMIT -> "${message.take(REPORT_MESSAGE_SIZE_LIMIT)}($ellipsis)"
+        else -> message
+    }
+}
```
gfreivasc, line 84: @cortinico I applied this treatment to handle the scenarios you've raised. As seen above, I've loosely set `REPORT_MESSAGE_SIZE_LIMIT` to be 80. There is a test that validates both truncation and space characters (`\n`, `\r`, `\t`, ...).

BraisGabin:
> So in your opinion we should fallback to having a simple extension method similar to the one we had with the previous iteration of `DetailedFindingsReport`? We could keep metrics that way:
> 
> ```kotlin
> fun Findings.detailed() = when (this) {
>     is ThresholdedCodeSmell -> "$id - $metric - [${messageOrDescription()}] at ${location.compact()}"
>     else -> "$id - [${messageOrDescription()}] at ${location.compact()}"
> }
> ```
> 
> This would eliminate API impact.

I was thinking about

```kotlin
fun Findings.detailed() = "$id - [${messageOrDescription()}] at ${location.compact()}"
}
```

But sure, the `when` have sense too. I'm not sure if `$metric` is relevant information knowing that the descritption should contain that information too. But I'm fine with both options.

gfreivasc:
About `metric` it's likely that the information would be redundant, but I think that's not really a problem in terms of presentation. I might be wrong, but I think that could be a change in the future in case that's the conclusion. And an easy change too.

BraisGabin (approved):
Thanks!

gfreivasc:
Fixed conflicts related to jUnit 5 migration

docs/pages/reporting.md
cortinico: > Truncate and ellipsis? What could be a threshold?\

I think it's something we can address should we face it. Ideally Truncate and ellipsis sounds like a decent way to clean this up, but it's something we can follow-up on.

cortinico (approved):
Change looks go to me 👍 Thanks for doing that. Are we good to merge?

Issue #4450, opened 2022-01-04
Replace Spek with a new test framework


3flex:
## Expected Behavior
We use a test framework that is properly maintained.

## Current Behavior
About a month since IntelliJ 2021.3 was released the Spek plugin remains incompatible. Also, https://github.com/spekframework/spek/issues/959#issuecomment-997344639

## Context
As a contributor this is a point of friction when using new IDE versions. We also can't expect any improvements to be made to Spek itself anymore as it's effectively abandoned by the original maintainer and unfortunately nobody else has stepped up.

I would suggest a transition away from Spek, as painful as this will be. This could be done in phases:
1. Pick a new test framework. Suggest JUnit 5 as it's the de facto standard and works perfectly well out of the box with Kotlin, Gradle & IntelliJ.
2. Don't accept contributions with new Spek tests - only allow the new framework.
3. Over time, rewrite existing tests using the new test framework (I had experimented with [IntelliJ's structural search and replace](https://www.jetbrains.com/help/idea/structural-search-and-replace.html) to assist with this, but didn't have much luck).

### TODO
* [X] custom-checks (#4467)
* [x] detekt-api (#4521)
* [x] detekt-cli (#4519)
* [x] detekt-core (#4507)
* [X] detekt-formatting (#4508)
* [x] detekt-generator (#4494)
* [X] detekt-gradle-plugin (#4529)
* [X] detekt-metrics (#4511)
* [X] detekt-parser (#4512)
* [X] detekt-psi-utils (#4515)
* [X] detekt-report-html (#4513)
* [X] detekt-report-sarif (#4513)
* [X] detekt-report-txt (#4513)
* [X] detekt-report-xml (#4513)
* [X] detekt-rules (#4514)
* [x] detekt-rules-complexity (#4566)
* [X] detekt-rules-coroutines (#4567)
* [X] detekt-rules-documentation (#4568)
* [x] detekt-rules-empty (#4522)
* [x] detekt-rules-errorprone (#4523)
* [x] detekt-rules-exceptions (#4572)
* [X] detekt-rules-naming (#4574)
* [x] detekt-rules-performance (#4569)
* [X] detekt-rules-style (#4575)
* [x] detekt-sample-extensions (#4525)
* [x] detekt-tooling (#4524)

marschwar:
detekt was my first encounter with spek and over time I have grown to like it. Nevertheless I think the transition to junit5 is a good idea.

[...]

PR #4467, opened 2022-01-07
Migrate custom-checks tests to JUnit 5

3flex:
Migrate `custom-checks` tests to JUnit 5.

See #4450

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -10,7 +10,7 @@
 - We use [JUnit 5](https://junit.org/junit5/docs/current/user-guide/) for testing. Please use the `Spec.kt` suffix on
   new test classes. If your new rule requires type resolution (i.e. it utilises `BindingContext`) then annotate your
   test class with `@KotlinCoreEnvironmentTest` and have the test class accept `KotlinCoreEnvironment` as a parameter.
-  See `SpekDiscoverySpec.kt` in the `custom-checks` module for a complete example.
+  See `SpekTestDiscoverySpec.kt` in the `custom-checks` module for a complete example.
 - We are in the process of migrating existing [Spek](https://github.com/spekframework/spek) tests to JUnit 5. For easier
   testing of these tests you might want to use the [Spek IntelliJ Plugin](https://plugins.jetbrains.com/plugin/10915-spek-framework).
 - Feel free to add your name to the contributors list at the end of the readme file when opening a pull request.
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -7,8 +7,12 @@
   which are enforced by ktlint when running `gradle detekt`.
 - Make sure your IDE uses [ktlint](https://github.com/pinterest/ktlint) formatting rules as well
   as the settings in [.editorconfig](../.editorconfig).
-- We use [Spek](https://github.com/spekframework/spek) for testing. Please use the `Spec.kt`-Suffix.
-  For easier testing you might want to use the [Spek IntelliJ Plugin](https://plugins.jetbrains.com/plugin/10915-spek-framework).
+- We use [JUnit 5](https://junit.org/junit5/docs/current/user-guide/) for testing. Please use the `Spec.kt` suffix on
+  new test classes. If your new rule requires type resolution (i.e. it utilises `BindingContext`) then annotate your
+  test class with `@KotlinCoreEnvironmentTest` and have the test class accept `KotlinCoreEnvironment` as a parameter.
+  See `SpekDiscoverySpec.kt` in the `custom-checks` module for a complete example.
+- We are in the process of migrating existing [Spek](https://github.com/spekframework/spek) tests to JUnit 5. For easier
+  testing of these tests you might want to use the [Spek IntelliJ Plugin](https://plugins.jetbrains.com/plugin/10915-spek-framework).
 - Feel free to add your name to the contributors list at the end of the readme file when opening a pull request.
 - The code in `detekt-api` and any rule in `detekt-rules` must be documented. We generate documentation for our website based on these modules.
 - If some Kotlin code in `resources` folder (like `detekt-formatting`) shows a compilation error, right click on it and use `Mark as plain text`.
--- a/custom-checks/src/test/kotlin/io/github/detekt/custom/SpekTestDiscoverySpec.kt
+++ b/custom-checks/src/test/kotlin/io/github/detekt/custom/SpekTestDiscoverySpec.kt
@@ -1,23 +1,27 @@
 package io.github.detekt.custom
 
-import io.gitlab.arturbosch.detekt.rules.setupKotlinEnvironment
+import io.gitlab.arturbosch.detekt.rules.KotlinCoreEnvironmentTest
 import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
 import org.assertj.core.api.Assertions.assertThat
-import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
-import org.spekframework.spek2.Spek
-import org.spekframework.spek2.style.specification.describe
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+import org.junit.jupiter.api.TestInstance.Lifecycle
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ValueSource
 
-class SpekTestDiscoverySpec : Spek({
-    setupKotlinEnvironment()
+@TestInstance(Lifecycle.PER_CLASS)
+class SpekTestDiscoverySpec : KotlinCoreEnvironmentTest() {
+    val subject = SpekTestDiscovery()
 
-    val subject by memoized { SpekTestDiscovery() }
-    val env: KotlinCoreEnvironment by memoized()
+    @Nested
+    inner class VariableDeclarationsInSpekGroupsShouldOnlyBeSimple {
 
-    describe("variable declarations in spek groups should only be simple") {
+        @Nested
+        inner class TopLevelScope {
 
-        context("top level scope") {
-
-            it("allows strings, paths and files by default") {
+            @Test
+            fun `allows strings, paths and files by default`() {
                 val code = createSpekCode(
                     """
                     val s = "simple"
@@ -29,25 +33,29 @@ class SpekTestDiscoverySpec : Spek({
                 assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
             }
 
-            it("detects disallowed types on top level scope") {
+            @Test
+            fun `detects disallowed types on top level scope`() {
                 val code = createSpekCode("val s = Any()")
 
                 assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
             }
 
-            it("allows memoized blocks") {
+            @Test
+            fun `allows memoized blocks`() {
                 val code = createSpekCode("val s by memoized { Any() }")
 
                 assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
             }
         }
 
-        context("describe and context blocks") {
+        @Nested
+        inner class DescribeAndContextBlocks {
 
-            setOf("describe", "context").forEach { name ->
-                it("allows strings, files and paths by default") {
-                    val code = createSpekCode(
-                        """
+            @ParameterizedTest
+            @ValueSource(strings = ["describe", "context"])
+            fun `allows strings, files and paths by default`(name: String) {
+                val code = createSpekCode(
+                    """
                         $name("group") {
                             val s = "simple"
                             val p = Paths.get("")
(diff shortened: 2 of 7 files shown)


schalkms (approved):

schalkms:
@3flex did you migrate the whole test package manually? Did you use some kind of transformation script to ease the migration? I think for some cases this might work.

custom-checks/src/test/kotlin/io/github/detekt/custom/SpekTestDiscoverySpec.kt
```diff
@@ -1,23 +1,27 @@
...
+import org.junit.jupiter.api.TestInstance
+import org.junit.jupiter.api.TestInstance.Lifecycle
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ValueSource
 
-    val subject by memoized { SpekTestDiscovery() }
-    val env: KotlinCoreEnvironment by memoized()
+@TestInstance(Lifecycle.PER_CLASS)
+class SpekTestDiscoverySpec : KotlinCoreEnvironmentTest() {
+    val subject = SpekTestDiscovery()
```
marschwar, line 15: Thank you for getting the whole thing started.

How would you feel about creating a custom extension instead inheriting from a base class? e.g.

```kotlin
@KotlinCoreEnvironmentTest
class SpekTestDiscoverySpec(private val env: KotlinCoreEnvironment) {
...
```

3flex: I looked at the JUnit extension API but I took the easier route as I'm not familiar with it.

If you can implement the extension I agree it's a little cleaner, but I don't feel strongly either way.

3flex:
> did you migrate the whole test package manually?

Yes, manually. A scripted approach would be great. I have tried a series of regex find/replace rules for the various parts which will definitely help (migrating specs like `it(...)` to `@Test fun ...()` is simple) but there are other parts that are more difficult. `context` and `describe` can't be easily converted to Nested JUnit tests because the context & describe descriptions are typically sentences which can't be used as class names without triggering class naming detekt rules (though that could be turned off for test sources).

custom-checks/src/test/kotlin/io/github/detekt/custom/SpekTestDiscoverySpec.kt
severn-everett: Baeldung has a good article on how to write extensions: https://www.baeldung.com/junit-5-extensions. However, I think it'd be better if the environment setup code were added to a base testing class that would be made "mandatory" for all Detekt tests via the guidelines - one less step that a developer would have to "know" about in order to write tests that require the environment variable.

marschwar: I am not sure I agree. If I understand correctly, the `KotlinCoreEnvironmentTest` is only necessary for rules that require type resolution.

The Baeldung article is misleading when it comes to state. The [Junit user guid](https://junit.org/junit5/docs/snapshot/user-guide/#extensions-keeping-state) endorses a different method.

I wrote a little extension that would allow us to "inject" a `KotlinCoreEnvironment` that is shared by all tests of the same test class. Don't get me wrong, I am fine with a base class as well but I have seen too many times that functionalities that are completely unrelated to each other are stuffed into on giant base class because they were all useful for tests. IMO the junit extension mechanism is a great way to mitigate that. 

```kotlin
@Retention(AnnotationRetention.RUNTIME)
@Target(AnnotationTarget.CLASS)
@ExtendWith(KotlinEnvironmentResolver::class)
annotation class KotlinCoreEnvironmentTest

internal class KotlinEnvironmentResolver : ParameterResolver {
    override fun supportsParameter(parameterContext: ParameterContext, extensionContext: ExtensionContext): Boolean {
        return parameterContext.parameter.type == KotlinCoreEnvironment::class.java
    }

    override fun resolveParameter(parameterContext: ParameterContext, extensionContext: ExtensionContext): Any {
        val closeableWrapper = extensionContext.wrapper
            ?: CloseableWrapper(createEnvironment()).also { extensionContext.wrapper = it }
        return closeableWrapper.wrapper.env
    }

    private var ExtensionContext.wrapper: CloseableWrapper?
        get() = getStore(NAMESPACE).get(WRAPPER_KEY, CloseableWrapper::class.java)
        set(value) = getStore(NAMESPACE).put(WRAPPER_KEY, value)

    companion object {
        private val NAMESPACE = ExtensionContext.Namespace.create("KotlinCoreEnvironment")
        private const val WRAPPER_KEY = "wrapper"
    }

    private class CloseableWrapper(val wrapper: KotlinCoreEnvironmentWrapper) : ExtensionContext.Store.CloseableResource {
        override fun close() {
            wrapper.dispose()
        }
    }
}
```

BraisGabin: I agree with @marschwar here. I prefer to use the extension mechanism. If someone prefers to use the class it can always create that class by ussing the extension. But from the class is way more difficult to create the extension.

severn-everett: Does it harm the tests that don't require the usage of the environment variable if that environment variable is included? Having an extension is going to add to the number of rituals that a developer will have to perform to write code correctly in a code base that already has a high amount of rituals and undocumented functions (mostly due to the underlying Kotlin PSI library, but still). If it comes to the point where the base class becomes too cumbersome with the helper functions and objects, then that could be addressed at that point.

BraisGabin:
I just created a script that ports nearly all the tests of `detekt-core` (I didn't try any other module):

```bash
#!/usr/bin/env bash

replace_all() {
  find detekt-core/src/test/ -name "*Spec.kt" -type f -print0 | xargs -0 gsed -i -E "$1"
}

replace_all '/ (it|test|context|describe)\(.*\)/s/[.:`>]/_/g'
replace_all 's/(object|class) (.*) : Spek\(\{/class \2 {/g'
replace_all 's/^class (.*)Spec/internal class \1Spec/g'
replace_all 's/class (.*)Spec/class \1Test/g'
replace_all 's/^\}\)$/}/g'
replace_all 's/describe\("(.*)"\)/@Nested inner class `\1`/g'
replace_all 's/context\("(.*)"\)/@Nested inner class `\1`/g'
replace_all 's/it\("(.*)"\)/@Test fun `\1`()/g'
replace_all 's/test\("(.*)"\)/@Test fun `\1`()/g'
replace_all 's/beforeEachTest \{/@BeforeEach fun setUp() {/g'
replace_all 's/afterEachTest \{/@AfterEach fun tearDown() {/g'
replace_all '/^package /a \\nimport org.junit.jupiter.api.Nested\nimport org.junit.jupiter.api.Test\nimport org.junit.jupiter.api.BeforeEach\nimport org.junit.jupiter.api.AfterEach'
replace_all '/^import org.spekframework.spek2/d'
```

I know that it's far from optimal but it does the work and it's kind of easy to work with.

The biggest problem to port are the `memoized`. I used the "Replace structurally..." option that intellij give us using this two patterns:
```
val $val$ by memoized {
    $code$
}
```
```
private val $val$ = $code$
```

I couldn't find a way to make it with a script so these part need some "clicks".

And the last thing to note here is `@TestInstance(Lifecycle.PER_CLASS)`. I don't like the idea to add that annotation to al our tests. It's way easy to forget about it. I found this that we could pass this argument when running our tests `-Djunit.jupiter.testinstance.lifecycle.default=per_class` and that will make all the tests `PER_CLASS` by default. And, if we want to use `PER_METHOD` we can do it using `@TestInstance(Lifecycle.PER_METHOD)`.

custom-checks/src/test/kotlin/io/github/detekt/custom/SpekTestDiscoverySpec.kt
BraisGabin: I agree that detekt has a high enter barrier and I don't want to move it even higher. But I just find 2 cases here:

- You want to extend/fix/change a rule: this will be setuped already
- You want to create a new rule: you will need to know that you need to add the extension or extend from one class <- both force you to know one step more. At the end you will copy&paste other test and work from there so we go to the first point.

Anyway, I'm ok keeping a class so they can extend. And more if this reduces the verbosity of our tests. But we should keep the extension behind to avoid future problems and to let third party rule authors to use composition over inheritance.

If we have that extension I'd use it in our rule-template instead of this: https://github.com/detekt/detekt-custom-rule-template/blob/a131231dd986e425c072e3e94ca0c17da7cb299f/src/test/kotlin/org/example/detekt/MyRuleTest.kt#L37-L54

marschwar: > Does it harm the tests that don't require the usage of the environment variable if that environment variable is included? 

I think it does a little bit. I ran some tests and found that it takes about 1.500 ms to create the kotlin test compiler and 100-150 ms to create the type resolution environment. Since the kotlin test compiler is shared between all tests this only hurts during development if you run a single test that does not use the test compiler. Creating the type resolution environment for all tests is a cost that can easily be avoided (it does not have to be via extension).

marschwar:
> I just created a script that ports nearly all the tests of detekt-core

Great job. I ran it on detekt-rules-empty and it seemed to do a good enough job to reduce the monkey work to a minimum.
I can definitely help transforming tests to junit. Should we split up by submodule? I would hate to see duplicate work on this.

> I found that we could pass this argument when running our tests -Djunit.jupiter.testinstance.lifecycle.default=per_class and that will make all the tests PER_CLASS by default. And, if we want to use PER_METHOD we can do it using @TestInstance(Lifecycle.PER_METHOD). But I'm don't know how to configure it with gradle.

I agree, we should use `PER_CLASS` as the default. In `build-logic/src/main/kotlin/module.gradle.kts` we can add this to the task config.
```kotlin
    useJUnitPlatform()
    systemProperty("junit.jupiter.testinstance.lifecycle.default","per_class")
    // ...
}
```

custom-checks/src/test/kotlin/io/github/detekt/custom/SpekTestDiscoverySpec.kt
chao2zhang: My observation of Detekt library has been that it tries to evaluate and adopt many cutting-edge libraries, including Kotlin language and its tooling. JUnit 5 extension is also new to me but this is also a great learning opportunity for me to study new cutting-edge technology.

3flex:
Thanks all - to close this one out I will:
1. Switch to the extension provided by marschwar (but I'll skip the subclass, I think the extension is easy enough to use on its own)
2. Apply the system property `junit.jupiter.testinstance.lifecycle.default=per_class`

Anything else? Any concerns?

Once that's done and this is merged @BraisGabin could adjust his script. Then we can setup a GitHub project to track the work to migrate each module, so as marschwar pointed out we can avoid duplication of effort.

custom-checks/src/test/kotlin/io/github/detekt/custom/SpekTestDiscoverySpec.kt
cortinico: Jumping on the discussion train here, sorry for being late :)

> I wrote a little extension that would allow us to "inject" a `KotlinCoreEnvironment` that is shared by all tests of the same test class. Don't get me wrong, I am fine with a base class as well but I have seen too many times that functionalities that are completely unrelated to each other are stuffed into on giant base class because they were all useful for tests. IMO the junit extension mechanism is a great way to mitigate that.

Totally agree on going with the `Extensions` way. I actually think that this should be land blocking. I'm fine merging this PR using inheritance for now, but I won't kickoff the whole migration process if we end up relying on a top level `KotlinCoreEnvironmentTest`.
From my experience, having a hierarchy in your test class is something that will turn against you at some point in the future.

Moreover, as we expose a `detekt-test-utils` that is framework-dependent, we should try to align with the framework specific extension patterns. In this case being a JUnit extension.

Finally, I think it would be great to have a series of annotation that defines the test behavior such as:

```
@ExtendWith(TypeResolutionExtension.class)
@ExtendWith(CompileSnippetsExtension.class)
```

and so on.

BraisGabin (approved):
<3 really nice api! I didn't know that junit 5 allowed that!!!

schalkms (approved):

.github/CONTRIBUTING.md
```diff
@@ -7,8 +7,12 @@
   which are enforced by ktlint when running `gradle detekt`.
 - Make sure your IDE uses [ktlint](https://github.com/pinterest/ktlint) formatting rules as well
   as the settings in [.editorconfig](../.editorconfig).
-- We use [Spek](https://github.com/spekframework/spek) for testing. Please use the `Spec.kt`-Suffix.
-  For easier testing you might want to use the [Spek IntelliJ Plugin](https://plugins.jetbrains.com/plugin/10915-spek-framework).
+- We use [JUnit 5](https://junit.org/junit5/docs/current/user-guide/) for testing. Please use the `Spec.kt` suffix on
+  new test classes. If your new rule requires type resolution (i.e. it utilises `BindingContext`) then annotate your
+  test class with `@KotlinCoreEnvironmentTest` and have the test class accept `KotlinCoreEnvironment` as a parameter.
+  See `SpekDiscoverySpec.kt` in the `custom-checks` module for a complete example.
```
chao2zhang, line 13:
```suggestion
  See `SpekTestDiscoverySpec.kt` in the `custom-checks` module for a complete example.
```

chao2zhang (approved):

marschwar (approved):

Issue #3603, opened 2021-03-23
HasPlatformType should be more strict


BraisGabin:
## Expected Behavior of the rule

```kotlin
val path = URLEncoder().encode("urlPath")
```

This code should be flagged because `path` have a platform type

this code should be:

```kotlin
val path = URLEncoder().encode("urlPath")!!
```
or
```kotlin
val path: String = URLEncoder().encode("urlPath")
```
or
```kotlin
val path: String? = URLEncoder().encode("urlPath")
```

## Context

#3571

Other code that should be flagged:

```kotlin
"${URLEncoder().encode("urlPath")}"
```

```kotlin
URLEncoder().encode("urlPath").toUpperCase()
```

```kotlin
hello(URLEncoder().encode("urlPath"))
```

severn-everett:
A couple of points:
- Which `URLEncoder` is being referred to here? The `URLEncoder` class from the `java.net` package doesn't have a public constructor.
- I'm not sure that there should be a check on string interpolation in and of itself. String interpolation can accept null values, so it wouldn't matter the `URLEncoder.encode()` function returned null in `"${URLEncoder.encode("urlPath")}"` - it'd just print out ``. It'd not be ideal, but it wouldn't be something that would necessary raise an NPE.

BraisGabin:
> it'd just print out a blank string

That is not true. It would print `"null"`. For that reason we have `NullableToStringCall`. And that rule doesn't want to mess with Platform Types. For that reason I think that this rule should handle them.

severn-everett:
Are all of these new checks only for public function and property declarations like how the rule is currently configured? Otherwise, you'd get situations where assigning a local property a flexible type would be fine, but then using that flexible local property in a chain would be flagged.

BraisGabin:
Nope, it applies to any code that returns an felxible type. You should assert about it's nullabilityat that moment. Even if you are assigning a local property. The local property shouldn't have a flexible type.

severn-everett:
Ok. A warning that this is going to affect a _lot_ of code in our own codebase; I'll go into more detail about it in the PR when the next commit gets pushed.

BraisGabin:
That's good. Then I'll talk about disable `UnsafeCallOnNullableType` on our code base. Or update it so the usage of `!!` for platform types could be allowed (adding a new configuration).

fwh-dc:
What's the status of this? I think this feature would be extremely useful.

cortinico:
[...]

PR #4470, opened 2022-01-08
HasPlatformType now will detect when a flexible-type call has been made for function arguments and in call chains

severn-everett:
This addresses issue #3603 (pending further discussion)


detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/HasPlatformType.kt
```diff
@@ -47,6 +54,116 @@ class HasPlatformType(config: Config) : Rule(config) {
         Debt.FIVE_MINS
     )
 
+    /*
+    override fun visitStringTemplateExpression(expression: KtStringTemplateExpression) {
```
severn-everett, line 58: This code is in place for a potential part of the issue that may or may not be necessary; if it's deemed not necessary, I'll remove it before the merge

BraisGabin (commented):
Thanks! I answered in the issue. But I think that if you enable this you want to be sure that you don't use platform type anywhere. So I think that it should check the String templates too.

severn-everett:
Right now, _all_ Java code is being marked as a flexible type, even those that contain the `@Nullable` and `@NonNullable` JetBrains annotations. I could probably pick out the annotations in the code, but I'm curious as to how IntelliJ is able to conduct the ["Infer nullity annotations" feature](https://www.jetbrains.com/help/idea/inferring-nullity.html#inferred-annotations). I would prefer that this PR be able to integrate that functionality - provided it isn't something proprietary from JetBrains - because otherwise, this rule would make interacting with any Java-based code a real pain.

cortinico:
> I could probably pick out the annotations in the code, but I'm curious as to how IntelliJ is able to conduct the "Infer nullity annotations" feature

I ran into similar situation in the past. You should potentially check if this type of check is performed by either the Kotlin Compiler or IntelliJ IDEA. If so, you should be able to find the  inspector code in its repository. We do have already other rules that are delegating to compiler or IDE inspectors (like `ForbiddenMethod`).

severn-everett:
I believe [this](https://github.com/JetBrains/intellij-community/blob/a9815772d1f73206fd6233a98adb77a8fb2573bf/java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/inference/JavaSourceInference.java) might be the code that conducts the inferred annotation investigation. However, I don't see the `dataFlow/inference/` package within the Kotlin Compiler library, and I'm not sure where to look on Maven for a different package that might need to be downloaded.

BraisGabin:
I think that I'm missing something here. If it's that code what do you need from the compiler library? Can't we use the inspector from intelliJ?

severn-everett:
What I wanted to say was that I couldn't find the code in the compiler library, and I'm not sure whether the code that I _did_ find is accessible for use in 3rd-party code like the Detekt project - it appears to be tightly-coupled to the IntelliJ Community source code. Let me know if I'm mistaken, but I think this might require a request that JetBrains transfer the code to the compiler library for our access.

BraisGabin:
Oh! I see. What we do in these cases is to copy&paste that code (adding proper atribution). The lincense allow us to do it.

cortinico:
What's the status of this PR?

schalkms:
@severn-everett may I kindly ask you if you plan to get this PR ahead?

I think it would be a nice addition to detekt.

severn-everett:
My apologies for dropping off the radar on this. I'll get myself spun back
up in this to try and tie this one off.

El dc., 1 de juny 2022, 20:52, schalkms ***@***.***> va
escriure:

severn-everett:
The relevant code for nullity inference in IntelliJ appears to be [this](https://github.com/JetBrains/intellij-community/blob/908c278e0e14b674cc72a2fe31f6b0b28bc20aa0/java/java-impl/src/com/intellij/codeInspection/inferNullity/NullityInferrer.java). As the classes mentioned in the file imply, this works on Java source code; as far as I was able to tell, there is no instance of the requisite `PsiMethod` found by the Detekt visitors when a Java function is invoked in Kotlin code. This may not be the case, but my schedule has tightened up quite a bit, so if somebody else were able to take a further look into this, it'd be appreciated.

BraisGabin:
Closing as there is not work in progress here for too long. Don't hesitate to open a new pr if someone want to ratake this work.

Issue #4474, opened 2022-01-08
`VarCouldBeVal` false positive


BraisGabin:
## Expected Behavior
If I run plain detekt (no type solving) with this code:

```kotlin
interface I {
    var optionEnabled: Boolean
}
fun test(): I {
    val o = object: I {
        override var optionEnabled: Boolean = false
    }
    return o
}
```

It shouldn't raise a `VarCouldBeVal` issue.

## Observed Behavior
It does

## Context
We shoudl accept the false-negatives in this issue without type-solving: any override `var` should be allowed if we don't have a proper `bindingContext`.

I created a test to demostrate it: 

```kotlin
it("does not report when an object initializes a variable directly - without type solving") {
    val code = """
        interface I {
            var optionEnabled: Boolean
        }
        fun test(): I {
            val o = object: I {
                override var optionEnabled: Boolean = false
            }
            return o
        }
    """
    assertThat(subject.compileAndLint(code)).isEmpty()
}
```
## Your Environment
* Version of detekt used: 1.17.1, 1.19.0 & current main

severn-everett:
How would this rule be run if type resolution isn't enabled? I was under the impression that the `@RequiresTypeResolution` annotation meant that it shouldn't run in this scenario.

BraisGabin:
I didn't check it, but it does. And it found some real cases (I runned it over https://github.com/mozilla-mobile/fenix) I assume that without type solving it just check private or local `var`s

cortinico:
Fixed by 4477

PR #4477, opened 2022-01-10
[VarCouldBeVal] Override vars will not be flagged if bindingContext is not set

severn-everett:
This addresses issue #4474


BraisGabin (approved):

cortinico (changes requested):
I think we should ship this change as it is, but instead add a:

```
if(bindingContext == BindingContext.EMPTY) return`
```

statement inside the `visitKtFile` method on top of this class. The problem is that this rule is a rule annotated with `@RequiresTypeResolution`, but failing to add early return statement is allowing to run this rule without TR.

This is a bit of a bummer as this rule is now a mixed rule (see #2994) which I believe it should have never been (see https://github.com/detekt/detekt/pull/3880)

BraisGabin:
Cortinico is right, I was mistaken here.

We should just disable the rule and honor the annotation. Something like adding

```kotlin
```

should do the trick

BraisGabin (approved):

cortinico (approved):

Issue #4483, opened 2022-01-11
InvalidPackageDeclaration doesn't enforce that the rootPackage is present


mateot1:
The `rootPackage` config for `InvalidPackageDeclaration` allows us to account for a file path that doesn't include the root part of the package, but it doesn't check that the package declaration contains this `rootPackage`. This can lead to inconsistency, for example:

with config `rootPackage: 'my.root.package'`
`file/path` containing a file with `package my.root.package.file.path` passes as expected
`file/path` containing a file with `package file.path` also passes 

## Expected Behavior of the rule
The rule should ensure that all classes in a given file path have the same package declaration. This means that if a `rootPackage` is defined, it should also enforce that it is present in the package declaration. 

## Context
For bettor or worse I work in a codebase that has a file structure that does not contain the root directories for kotlin src files which allows these inconsistencies to arise...

cortinico:
I think this is a nice improvement. However, I would like to point out how this is a behavior change. Specifically, before the `rootPackage` was used just to ignore part of the package declaration.

Please also note how this is explained in the documentation:
https://github.com/detekt/detekt/blob/56e7fa94a47060fefa6c6e7a9e4bb73d1e9d1a16/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/InvalidPackageDeclaration.kt#L28-L29

If we end up merging #4484, we should also update that documentation to clarify this behavioral change.

mateot1:
updated!

cortinico:
Fixed by #4484

PR #4484, opened 2022-01-11
update InvalidPackageDeclaration to report if rootPackage is not present

mateot1:
This should fix the issue I just opened here: https://github.com/detekt/detekt/issues/4483

### DIFF
--- a/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/InvalidPackageDeclaration.kt
+++ b/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/InvalidPackageDeclaration.kt
@@ -34,24 +34,26 @@ class InvalidPackageDeclaration(config: Config = Config.empty) : Rule(config) {
         if (declaredPath.isNotBlank()) {
             val normalizedFilePath = directive.containingKtFile.absolutePath().parent.toNormalizedForm()
             val normalizedRootPackage = packageNameToNormalizedForm(rootPackage)
-            val expectedPath =
-                if (normalizedRootPackage.isBlank()) {
-                    declaredPath
-                } else {
-                    declaredPath.substringAfter(normalizedRootPackage)
+            val expectedPath = when {
+                normalizedRootPackage.isBlank() -> declaredPath
+                normalizedRootPackage !in declaredPath -> {
+                    directive.reportInvalidPackageDeclaration("The package declaration is missing the root package")
+                    return
                 }
+                else -> declaredPath.substringAfter(normalizedRootPackage)
+            }
 
             val isInRootPackage = expectedPath.isBlank()
             if (!isInRootPackage && !normalizedFilePath.endsWith(expectedPath)) {
-                report(
-                    CodeSmell(
-                        issue,
-                        Entity.from(directive),
-                        "The package declaration does not match the actual file location.",
-                    )
+                directive.reportInvalidPackageDeclaration(
+                    "The package declaration does not match the actual file location."
                 )
             }
         }
+    }
+
+    private fun KtElement.reportInvalidPackageDeclaration(message: String) {
+        report(CodeSmell(issue, Entity.from(this), message))
     }
 
     private fun <T> Iterable<T>.toNormalizedForm() = joinToString("|")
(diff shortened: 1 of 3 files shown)


detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/InvalidPackageDeclaration.kt
```diff
@@ -34,26 +34,28 @@ class InvalidPackageDeclaration(config: Config = Config.empty) : Rule(config) {
...
-                } else {
-                    declaredPath.substringAfter(normalizedRootPackage)
+            val expectedPath = when {
+                normalizedRootPackage.isBlank() -> declaredPath
+                !declaredPath.startsWith(normalizedRootPackage) -> {
+                    directive.reportInvalidPackageDeclaration("The package declaration is missing the root package")
+                    return
                 }
+                else -> declaredPath.substringAfter(normalizedRootPackage)
+            }
```
marschwar, line 44: If find a return statement within a when a little bit confusing. I would prefer the more verbose but also more explicit version:

```kotlin
if (normalizedRootPackage.isNotBlank() && !declaredPath.startsWith(normalizedRootPackage)) {
    directive.reportInvalidPackageDeclaration("The package declaration is missing the root package")
    return
}
val expectedPath = if (normalizedRootPackage.isBlank()) {
    declaredPath
} else {
    declaredPath.substringAfter(normalizedRootPackage)
}
```

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/InvalidPackageDeclaration.kt
```diff
@@ -25,7 +25,7 @@ class InvalidPackageDeclaration(config: Config = Config.empty) : Rule(config) {
         debt = Debt.FIVE_MINS
     )
 
-    @Configuration("if specified this part of the package structure is ignored")
+    @Configuration("if specified the declaration must start with this root, but it is ignored in the file structure")
```
cortinico, line 28: Why "it is ignored in the file structure"?

mateot1: I guess I was just trying to reword the existing behavior and used "file" instead of "package"... this change forces you to include the root in the package _declaration_ but you can still have a file structure that doesn't include those root directories. e.g. `package my.root.package.file.path` works with `file/path` as well as `my/root/package/file/path`

marschwar: I guess that is the problem when a single configuration means multiple things. ;) How about:

_If specified the package declaration must start with this base package. In the directory structure this part may or may not be omitted._

Also it may help if we include a link to the [corresponding section in the Kotlin Coding Conventions](https://kotlinlang.org/docs/coding-conventions.html#directory-structure).

BraisGabin: I think that we shouldn't "overload" this configuration. If we want that all the files start with the same package we should create a new configuration for it. And, if you want to ignore the first part of that root path AND enforce it you need to set it twice.

mateot1: okay, so just like a `requireRootInDeclaration` boolean?

BraisGabin: Sorry for the late answer. Yes, I think that a `requireRootInDeclaration` have sense.

mateot1:
sorry for the delay. I updated this to use a new config field and added a default to `default-detekt-config.yml`. let me know if there's anything I missed for adding a new config field.

detekt-rules-naming/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/naming/InvalidPackageDeclarationSpec.kt
```diff
@@ -106,6 +108,37 @@ internal class InvalidPackageDeclarationSpec : Spek({
...
+            val config by memoized { TestConfig(mapOf(ROOT_PACKAGE to "com.example", REQUIRE_ROOT_PACKAGE to true)) }
+
+            it("should pass if declaration starts with root package") {
+                val source = """
+                    package com.example.foo.bar
+
+                    class C
+                """
+
+                val ktFile = compileContentForTest(source, createPath("src/foo/bar/File.kt"))
```
cortinico, line 123: Can you add another test for the file in the following path:
```
src/com/example/foo/bar/File.kt
```

cortinico (approved):

BraisGabin (approved):

PR #4491, opened 2022-01-13
Fix main

BraisGabin:
We have `main` failing. I assume that the reason was a PR adding this new rule and another doing a change that violates that rule. Both merged and the problem was found at `main`.

But I'm wondering if we want to keep `CanBeNonNullable` on detekt. PSI isn't a funcitional API. And `CanBeNonNullable` works better in functionalish environments.


chao2zhang (approved):
I guess keeping a rule enabled for the codebase helps with validating the rule itself, as a side benefit.

PR #4494, opened 2022-01-16
Transform tests in detekt-generator to junit

marschwar:
As part of #4450 this replaces Spek tests in `detekt-generator` to JUnit.

I basically used the script from @BraisGabin and did not change a whole lot.


marschwar:
Should we rename the nested class and test function names or disable naming rule for tests?

BraisGabin:
I vote for the second one

marschwar:
Is there anything we need to discuss about the transformation to junit? If not I would move on to the next module.

BraisGabin:
I would vote to don't rename the files yet. It will help git to understand the change and it doesn't mess with this:

> Please use the `Spec.kt` suffix on new test classes

https://github.com/detekt/detekt/blob/259c9de1ab22c6ad1fec544e8c7479481f1251ab/.github/CONTRIBUTING.md?plain=1#L10-L11

Once we have all the tests in Junit 5 we can rename all of them to `Test.kt` and change the documentation. It should be an easy task.

marschwar:
I agree. The diff is still quite large but at least there is a chance to see what is going on.

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/config/DetektYmlConfigSpec.kt
```diff
@@ -2,58 +2,64 @@ package io.gitlab.arturbosch.detekt.generator.config
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments
+import org.junit.jupiter.params.provider.Arguments.arguments
+import org.junit.jupiter.params.provider.MethodSource
 import java.nio.file.Paths
+import java.util.stream.Stream
 
-class DetektYmlConfigSpec : Spek({
+internal class DetektYmlConfigSpec {
```
chao2zhang, line 13: Is this `internal` modifier added by hand? I am curious why tests classes need to be internal

chao2zhang (approved):
Lgtm. Thanks for spearhead this effort

BraisGabin (approved):

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/config/DetektYmlConfigSpec.kt
marschwar: To be honest I do not know. There is no particular reason and I have removed it. Good catch

PR #4504, opened 2022-01-20
[MaxLineLength] Fix signature in for blank characters in the Baseline

### LINKED ISSUE #4480: Baseline not catching multiline string literals correctly
## Expected Behavior
A MaxLineLength with multiline string literals will be written in the baseline and not causing problems when running `detekt`.

## Observed Behavior
It's written in the baseline file as follows:

```
<ID>MaxLineLength:MainActivity.kt$MainActivity$ </ID>
```
The task fails. I think the signature is incorrect.

## Steps to Reproduce
- Checkout my project: https://github.com/jelleheemskerk/detekt-bug
- Run `./gradlew detekt`

## Context
I'm trying to integrate detekt with our existing codebase. We use multiline string literals with JSON data for our tests. I don't seem to have this issue on **1.18.1**.

## Your Environment
* Version of detekt used: **1.19.0**
* Version of Gradle used (if applicable): **7.0.2**
* Gradle scan link (add `--scan` option when running the gradle task): https://gradle.com/s/ix7enmucwqthe
* Operating System and version: **Mac OS 12.1**
* Link to your project (if it's a public repository): https://github.com/jelleheemskerk/detekt-bug

marschwar:
This fixes #4480

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLengthSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLengthSpec.kt
@@ -8,6 +8,7 @@ import io.gitlab.arturbosch.detekt.test.TestConfig
 import io.gitlab.arturbosch.detekt.test.assertThat
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
+import org.assertj.core.api.Assertions.assertThat as doAssert
 
 private const val MAX_LINE_LENGTH = "maxLineLength"
 private const val EXCLUDE_PACKAGE_STATEMENTS = "excludePackageStatements"
@@ -35,7 +36,15 @@ class MaxLineLengthSpec : Spek({
                 val rule = MaxLineLength()
 
                 rule.visit(fileContent)
-                assertThat(rule.findings).hasSize(6)
+                assertThat(rule.findings).hasSize(7)
+            }
+
+            it("should report meaningful signature for all violations") {
+                val rule = MaxLineLength()
+
+                rule.visit(fileContent)
+                val locations = rule.findings.map { it.signature.substringAfterLast('$') }
+                doAssert(locations).allSatisfy { doAssert(it).isNotBlank() }
             }
         }
 
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/Junk.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/Junk.kt
@@ -8,12 +8,12 @@ import org.jetbrains.kotlin.psi.psiUtil.elementsInRange
 import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
 
 /**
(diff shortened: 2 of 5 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLengthSpec.kt
```diff
@@ -8,6 +8,7 @@ import io.gitlab.arturbosch.detekt.test.TestConfig
 import io.gitlab.arturbosch.detekt.test.assertThat
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.describe
+import org.assertj.core.api.Assertions.assertThat as doAssert
```
marschwar, line 11: @BraisGabin This is a probably unintended side effect of #4437
Do have a better idea on how to fix it?

BraisGabin: I don't understand, what's unintended here? I assume that you are using `doAssert` because you don't want to use `io.gitlab.arturbosch.detekt.test.assertThat`. But I think that you could remove this last one and everything should work the same.

BraisGabin (approved):

marschwar:
The question what is the best way to use 
`io.gitlab.arturbosch.detekt.test.assertThat{Finding)` and 
`org.assertj.core.api.Assertions.assertThat(List<String>)` in the same file?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLengthSpec.kt
marschwar: The question what is the best way to use
io.gitlab.arturbosch.detekt.test.assertThat{Finding) and
org.assertj.core.api.Assertions.assertThat(List<String>) in the same file?

BraisGabin: Isn't `io.gitlab.arturbosch.detekt.test.assertThat` an extension of `org.assertj.core.api.Assertions.assertThat`? I thought that we could always use `io.gitlab.arturbosch.detekt.test.assertThat` instead of `org.assertj.core.api.Assertions.assertThat`.

marschwar: No, the detekt versions support `Finding`, `List<Finding>` and `ThresholdedCodeSmell`. For all other assertions `org.assertj.core.api.Assertions.assertThatXXX` are used. This is only a problem if the test uses both. I see 3 options to mitigate this

* allow the assertJ import again
* rename the detekt functions (or provide aliases) to  something like `assertThatFinding`
* use the alias import the way it was done here in this pr

BraisGabin:
If we end up reverting #4335 would we want to merge this anyway? Because if we don't want this if we revert #4335 probably what we should do is just revert #4335.

cortinico (approved):

marschwar:
I vote to merge this regardless as it allows to differentiate between two similar violations in the same file

PR #4510, opened 2022-01-23
Disable `UnnecessaryInnerClass` for tests

marschwar:
In junit inner classes are a way to structure test contexts.


chao2zhang:
What do you think about https://github.com/detekt/detekt/pull/4509?

marschwar:
#4509 could be a real fix

PR #4515, opened 2022-01-24
Migrate detekt-psi-utils to junit

marschwar:
Belongs to #4450

### DIFF
--- a/detekt-psi-utils/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignatureSpec.kt
+++ b/detekt-psi-utils/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignatureSpec.kt
@@ -1,58 +1,58 @@
 package io.gitlab.arturbosch.detekt.rules
 
 import org.assertj.core.api.Assertions.assertThat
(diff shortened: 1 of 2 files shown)


detekt-psi-utils/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignatureSpec.kt
```diff
@@ -1,58 +1,58 @@
 package io.gitlab.arturbosch.detekt.rules
 
 import org.assertj.core.api.Assertions.assertThat
-import org.spekframework.spek2.Spek
-import org.spekframework.spek2.style.specification.describe
-
-class MethodSignatureSpec : Spek({
-    describe("extractMethodNameAndParams") {
-        listOf(
```
3flex, line 9: I'm planning to use JUnit dynamic tests in detekt-tooling because I'm lazy and it mapped well to the existing test code - just FYI in case this pattern is used elsewhere.

But I'm sure individual static tests are best practice so good on you for going this route :)

3flex (approved):

PR #4368, opened 2021-12-10
Improve `AnnotationExcluder` tests

BraisGabin:
Because of #4355 and #4356 we will need to keep `AnnotationExcluder` class. But that class has a lot of errors as you can check in the tests of this PR. This PR just improves the test and make some minor improvements in the class. The next PR will refactorice it heavily to make all those errors spotted in the test to work.

PR #4518, opened 2022-01-24
Fix `AnnotationExcluder`

BraisGabin:
#4368 showed that `AnnotationExcluder` had some issues: false-positives and false-negatives.

This PR implements a `FullQualifiedNameGuesser` to archive that. It's a guesser so there are some scenarios where we can't know for sure which class we are refering too. But in general it only returns one option. This is a best-effort for the users that doesn't use type-solving. With type-solving we have other implementation that works all the times.


detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/FullQualifiedNameGuesser.kt
```diff
@@ -0,0 +1,108 @@
...
+    private val packageName: String?,
+    imports: List<KtImportDirective>,
+) {
+
+    constructor(root: KtFile) : this(
+        packageName = root.packageDirective?.qualifiedName?.ifBlank { null },
+        imports = root.importList?.imports.orEmpty(),
+    )
+
+    @Suppress("ClassOrdering")
```
BraisGabin, line 17: I can remove this suppress and move `resolvedNames` before the secundary constructor. But, for me, it would make the readability worst.

detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/AnnotationExcluderSpec.kt
```diff
@@ -23,77 +23,61 @@ class AnnotationExcluderSpec {
...
+                "*.Component.*,@dagger.Component.Factory,true",
+                "foo.Component,@Component,false",
+                "foo.Component,@dagger.Component,false",
+                "foo.Component,@Factory,false",
+                "foo.Component,@Component.Factory,false",
+                "foo.Component,@dagger.Component.Factory,false",
+            ]
+        )
+        fun `all cases`(exclusion: String, annotation: String, shouldExclude: Boolean) {
+            val excluder = AnnotationExcluder(file, listOf(exclusion))
```
marschwar, line 77: _Non blocking comment:_ I hate parameterized tests like this. That is why I split it into 2 tests. But I know there are more schools of thought here ;) But that is something for our style guide. How many parameters may a test have before it is considered unreadable.

BraisGabin: I don't like this kind of tests either. I would prefer to have way less. But they were needed. There were really random cases that were not covered. And while I was implementing this I broke some of them while fixing others.

And I think that it's better to have all the cases together because they are in blocks of 5. Maybe I could add a new line between each "block of tests" to make it easier to read. But I can reverse the change and split them in two if you think it's better.

detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/FullQualifiedNameGuesser.kt
cortinico: Yup agree

detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/FullQualifiedNameGuesser.kt
```diff
@@ -0,0 +1,108 @@
...
+        val resolvedName = resolvedNames[searchName]
+        return if (name == searchName) {
+            resolvedName
+        } else {
+            "$resolvedName.${name.substringAfter('.')}"
+        }
+    }
+}
+
+private val kotlinPackageClasses = arrayOf(
```
cortinico, line 58: A couple of things here:
1. I would move it in a separate file.
2. Is it copied from somewhere in the Kotlin repo? If so, please provide a reference so we can keep it up to date easily.

BraisGabin: Both of them done. Thanks to force the point 2. I was missing *lots* of classes. I couldn't find a list with this information so I created a script to extract it from the documentation of detekt.

Basically I download the documentation and parse the html. And I extracted the list of packages to parse from the documentation (all explained in the comment)

detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/KotlinNoImportClasses.kt
```diff
@@ -0,0 +1,199 @@
...
+print_classes() {
+  packages=(kotlin kotlin.annotation kotlin.collections kotlin.comparisons kotlin.io kotlin.ranges kotlin.sequences kotlin.text)
+
+  for package in "${packages[@]}"; do
+    curl -s "https://kotlinlang.org/api/latest/jvm/stdlib/$package/" | gsed -n -E "s|<h4><a href=\".*\">([A-Z].*)</a></h4>|\"\1\" to \"$package.\1\",|p"
+  done
+}
+
+print_classes | sort -u | gsed '/^"[A-Z_]*"/d'
+
```
cortinico, line 19: :O

cortinico (approved):

detekt-psi-utils/src/test/kotlin/io/github/detekt/psi/FullQualifiedNameGuesserSpec.kt
```diff
@@ -0,0 +1,82 @@
...
+import io.github.detekt.test.utils.compileContentForTest
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+class FullQualifiedNameGuesserSpec {
+
+    @Nested
+    class `With package` {
+        private val sut = FullQualifiedNameGuesser(
```
chao2zhang, line 12: What is sut?

detekt-psi-utils/src/test/kotlin/io/github/detekt/psi/FullQualifiedNameGuesserSpec.kt
```diff
@@ -0,0 +1,82 @@
...
+        @Test
+        fun `import with alias but using real name`() {
+            assertThat(sut.getFullQualifiedName("JvmStatic"))
+                .containsExactlyInAnyOrder("foo.JvmStatic")
+        }
+
+        @Test
+        fun `no import but maybe kotlin`() {
+            assertThat(sut.getFullQualifiedName("Result"))
+                .containsExactlyInAnyOrder("foo.Result", "kotlin.Result")
```
chao2zhang, line 44: 👍  This is a good catch

chao2zhang (approved):

detekt-psi-utils/src/test/kotlin/io/github/detekt/psi/FullQualifiedNameGuesserSpec.kt
BraisGabin: Subject under test

detekt-psi-utils/api/detekt-psi-utils.api
```diff
@@ -20,6 +20,11 @@ public final class io/github/detekt/psi/FilePath$Companion {
 	public final fun fromRelative (Ljava/nio/file/Path;Ljava/nio/file/Path;)Lio/github/detekt/psi/FilePath;
 }
 
+public final class io/github/detekt/psi/FullQualifiedNameGuesser {
+	public fun <init> (Lorg/jetbrains/kotlin/psi/KtFile;)V
+	public final fun getFullQualifiedName (Ljava/lang/String;)Ljava/util/Set;
+}
+
```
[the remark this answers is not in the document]
BraisGabin: Agree, we can always make it public in the future. I'm moving this to `:detekt-api` in the `internal` package. I don't like that place but is the only option if I want to use `AnnotationExcluder` and keep it out of out public api. We can always move it later.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/AnnotationExcluder.kt
```diff
@@ -35,8 +29,28 @@ class AnnotationExcluder(
...
+            listOf(annotationText)
+        }.flatMap { fqName ->
+            fqName
+                .split(".")
+                .dropWhile { it.first().isLowerCase() }
+                .reversed()
+                .scan("") { acc, name ->
+                    if (acc.isEmpty()) name else "$name.$acc"
+                }
+                .drop(1) + fqName
```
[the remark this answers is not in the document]
BraisGabin: Done

PR #4521, opened 2022-01-24
migrate tests in detekt-api to junit

marschwar:
This is part on #4450.

Unfortunately it was too late when I realized it will collide with #4518. 

@BraisGabin Feel free to mark this as blocked if #4518 should be merged first.


BraisGabin:
Don't worry, it would be really easy to fix.

detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/SingleAssignSpec.kt
```diff
@@ -2,34 +2,36 @@ package io.gitlab.arturbosch.detekt.api
...
-internal class SingleAssignSpec : Spek({
-    describe("value is unset") {
-        var unassigned: Int by SingleAssign()
-        it("should fail when value is retrieved") {
-            assertThatIllegalStateException().isThrownBy {
-                @Suppress("UNUSED_EXPRESSION")
-                unassigned
-            }
-        }
+@TestInstance(PER_METHOD)
```
BraisGabin, line 9: This is the default, right? We don't need it.

BraisGabin (approved):

detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/SingleAssignSpec.kt
marschwar: No, we changed the setting for All Tests to per-class and have to change it here.

BraisGabin: Oh! You are right! my bad.

PR #4523, opened 2022-01-25
Migrate tests in detekt-rules-errorprone to junit

marschwar:
Part of #4450


BraisGabin:
I'm having this same issue in #4459 (https://github.com/detekt/detekt/runs/4934052976?check_suite_focus=true#step:4:556)

It seems that our compile-test-snippets is a bit flaky. If I change the number of engines here:

https://github.com/detekt/detekt/blob/4a89a25c82d35874571270fa1ff32751f2f268a1/detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinScriptEnginePool.kt#L13

From 8 to 1 a lot of issues appear on `main`.

And, in my case if I change the number from 8 to 9 all works perfectly fine...

That PR is older than our PRs related with JUnit so JUnit is not the reason.

marschwar:
Really strange...

BraisGabin:
Any idea why we have 8 of those? It doesn't seem that we are doing any parallel job here so with one it should be enough.

marschwar:
No idea. It may also be something different entirely. It could well be, that I have some errors in the scripts to compile. I will investigate.

marschwar:
In this case it seems like there is something wrong with the snippets in `IgnoredReturnValueSpec.kt`

cortinico:
There is a merge conflict on this PR now. @marschwar are you able to take a look.
We also have #4541 that depends on this 👍

BraisGabin (approved):

cortinico (approved):

PR #4524, opened 2022-01-25
Migrate detekt-tooling tests to JUnit 5

3flex:
See #4450


BraisGabin (approved):
I must say that JUnit5 is way better than I thought.

PR #4528, opened 2022-01-26
Disable deployment jobs in forked repository

Goooler:
See square/kotlinpoet/pull/1119.


chao2zhang (approved):
👍 . I also learned that we don't need `{{ }}` in `if` https://docs.github.com/en/actions/learn-github-actions/expressions

BraisGabin (approved):

PR #4529, opened 2022-01-26
Migrate detekt-gradle-plugin tests to JUnit

3flex:
See #4450

### DIFF
--- a/detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektReportMergeSpec.kt
+++ b/detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektReportMergeSpec.kt
@@ -4,15 +4,19 @@ import io.gitlab.arturbosch.detekt.testkit.DslGradleRunner
 import io.gitlab.arturbosch.detekt.testkit.DslTestBuilder
 import io.gitlab.arturbosch.detekt.testkit.ProjectLayout
 import org.assertj.core.api.Assertions.assertThat
-import org.spekframework.spek2.Spek
-import org.spekframework.spek2.style.specification.describe
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments
+import org.junit.jupiter.params.provider.Arguments.arguments
+import org.junit.jupiter.params.provider.MethodSource
 
-internal class DetektReportMergeSpec : Spek({
+class DetektReportMergeSpec {
 
-    describe("Sarif merge is configured correctly for multi module project") {
+    @Nested
+    inner class `Sarif merge is configured correctly for multi module project` {
 
-        val groovy by memoized { DslTestBuilder.groovy() }
-        val groovyBuildFileContent by memoized {
+        val groovy = DslTestBuilder.groovy()
+        val groovyBuildFileContent =
             """
             |${groovy.gradlePlugins}
             |
@@ -41,9 +45,8 @@ internal class DetektReportMergeSpec : Spek({
             |  }
             |}
             |""".trimMargin()
-        }
-        val kotlin by memoized { DslTestBuilder.kotlin() }
-        val kotlinBuildFileContent by memoized {
+        val kotlin = DslTestBuilder.kotlin()
+        val kotlinBuildFileContent =
             """
             |${kotlin.gradlePlugins}
             |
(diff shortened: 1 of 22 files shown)


detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslTestBuilder.kt
```diff
@@ -96,9 +99,11 @@ private class GroovyBuilder : DslTestBuilder() {
     override val gradleSubprojectsApplyPlugins = """
         |apply plugin: "io.gitlab.arturbosch.detekt"
         |"""
+
+    override fun toString() = "build.gradle"
 }
 
-private class KotlinBuilder : DslTestBuilder() {
+class KotlinBuilder : DslTestBuilder() {
```
BraisGabin, line 106: Why public?

detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektReportMergeSpec.kt
```diff
@@ -72,45 +75,47 @@ internal class DetektReportMergeSpec : Spek({
...
-                        numberOfCodeSmells = 4
-                    )
-                }
+        fun scenarios(): List<Arguments> = listOf(
+            arguments(groovy, groovyBuildFileContent),
+            arguments(kotlin, kotlinBuildFileContent)
+        )
+
+        @ParameterizedTest(name = "Using {0}")
+        @MethodSource("scenarios")
```
BraisGabin, line 85: Don't commit this! For sure it will not compile.
```suggestion
        class Scenarios : ArgumentsProvider {
           override fun provideArguments(context: ExtensionContext) = Stream.of(
            arguments(DslTestBuilder.groovy(), groovyBuildFileContent),
            arguments(DslTestBuilder.kotlin(), kotlinBuildFileContent),
         )
        }

        @ParameterizedTest(name = "Using {0}")
        @ArgumentsSource(DetektReportMergeSpec.Scenarios::class)
```

But what do you think of something like this? It avoid `String` references that are easier to break. The problem is that it's more verbose. It's just a suggestion.

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslTestBuilder.kt
```diff
@@ -71,6 +71,9 @@ abstract class DslTestBuilder {
     companion object {
         fun kotlin(): DslTestBuilder = KotlinBuilder()
         fun groovy(): DslTestBuilder = GroovyBuilder()
+
+        @JvmStatic
+        fun builders(): Set<DslTestBuilder> = setOf(GroovyBuilder(), KotlinBuilder())
```
BraisGabin, line 76: Maybe you could define a `ArgumentsProvider` that returns the same and call it with `@ArgumentsSource` instead of `@MethodSource`

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslTestBuilder.kt
```diff
@@ -1,11 +1,11 @@
...
+open class DslTestBuilder {
 
-    abstract val gradleBuildConfig: String
-    abstract val gradleBuildName: String
-    abstract val gradlePlugins: String
-    abstract val gradleSubprojectsApplyPlugins: String
+    open val gradleBuildConfig: String = ""
+    open val gradleBuildName: String = ""
+    open val gradlePlugins: String = ""
+    open val gradleSubprojectsApplyPlugins: String = ""
```
BraisGabin, line 8: What do you get from doing this? You aren't using it.

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslTestBuilder.kt
```diff
@@ -1,11 +1,11 @@
 package io.gitlab.arturbosch.detekt.testkit
 
-abstract class DslTestBuilder {
+open class DslTestBuilder {
```
BraisGabin, line 3: Maybe this could be a `sealed class`

detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektReportMergeSpec.kt
3flex: I think the string reference is OK in this case, given the function sits directly above the test method. Also I just tried and IntelliJ will update the string in the annotation if the method is renamed using Refactor -> Rename in the IDE, and also if the string itself is renamed with Refactor -> Rename.

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslTestBuilder.kt
3flex: Ah forgot to change it back after trying a different approach.

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslTestBuilder.kt
3flex: I've reverted this to "abstract", this change was left there by mistake after trying a different approach.

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslTestBuilder.kt
3flex: Is that to avoid the string reference in the annotation? I still think it's low risk. Even though it's a runtime error it will be identified by CI if not before (JUnit raises an exception), and it would only affect the test suite. I prefer the less verbose approach.

BraisGabin (approved):

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslTestBuilder.kt
```diff
@@ -71,6 +71,9 @@ abstract class DslTestBuilder {
     companion object {
         fun kotlin(): DslTestBuilder = KotlinBuilder()
         fun groovy(): DslTestBuilder = GroovyBuilder()
```
cortinico, line 73: I was under the assumption we would remove the Kotlin DSL tests here before this PR?

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslTestBuilder.kt
cortinico: > Is that to avoid the string reference in the annotation?

Agree that doesn't look great. However as you mentioned, let's go for the less verbose solution.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/report/ReportMergeSpec.kt
```diff
@@ -189,4 +195,16 @@ class ReportMergeSpec : Spek({
...
+    val version = System.getProperty("java.version")
+    val majorVersion = if (version.startsWith("1.")) {
+        version.substring(2, 3)
+    } else if (!version.contains('.') && version.contains('-')) {
+        version.substringBefore('-') // early access Java.net versions
+    } else {
+        version.substringBefore('.')
+    }
+    return Integer.parseInt(majorVersion)
+}
```
cortinico, line 210: I have the gut feeling this could be exported to a more common location

cortinico: (why is even here in the first place? 🤔 )

cortinico (approved):
Generally looks good to me 👍 Diff is quite big so I haven't looked into too much details as I believe most of the changes were auto-generated.

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslTestBuilder.kt
3flex: Not for the whole test suite yet. The Groovy tests were removed in https://github.com/detekt/detekt/pull/4563 only for `DetektTaskDslSpec` because it made the JUnit migration simpler in that test class. All other test classes still test with both DSLs and can be addressed in another PR.

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/report/ReportMergeSpec.kt
3flex: This function is now unused, removed.

3flex:
> Diff is quite big

Just wait for detekt-rules-style migration :D

image

(With more to do)

cortinico:
> Just wait for detekt-rules-style migration :D

Ahahah my brain refuses to process whatever is above 1k loc 😅

marschwar (approved):

Issue #4496, opened 2022-01-17
Potential security issue


JamieSlome:
Hey there!

I belong to an open source security research community, and a member (@ready-research) has found an issue, but doesn’t know the best way to disclose it.

If not a hassle, might you kindly add a `SECURITY.md` file with an email, or another contact method? GitHub [recommends](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) this best practice to ensure security issues are responsibly disclosed, and it would serve as a simple instruction for security researchers in the future.

Thank you for your consideration, and I look forward to hearing from you!

(cc @huntr-helper)

chao2zhang:
Hello @JamieSlome! Thanks for sharing about the vulnerability found.

- We are definitely planning to add `SECURITY.md`, but we are still waiting on a few things including our newly setup OpenCollective and our specific domain.
- Which vulnerability management platform would you recommend for us to communicate effectively?

If you think the security vulnerability is urgent, we can definitely find other ways to speed things ups.

JamieSlome:
@chao2zhang - thanks for the response! ⚡ 

Let me know if you need any help establishing the `SECURITY.md` - happy to contribute!

We currently have the report on our platform, which you can view here: [report](https://huntr.dev/bounties/23e37ba7-96d5-4037-a90a-8c8f4a70ce44/).

It is private and only accessible to maintainers with repository write permissions! 👍  Let me know if you need any assistance or have any questions.

cortinico:
Thanks for your support @JamieSlome 

> Let me know if you need any help establishing the `SECURITY.md` - happy to contribute!

Do you happen to have a template that we can use as a starting point?

JamieSlome:
@cortinico - you are welcome!

Happy to work on #4538 with you!

JamieSlome:
Just further to the [report](https://huntr.dev/bounties/23e37ba7-96d5-4037-a90a-8c8f4a70ce44/), if you could confirm the fix once it has been addressed, that would be appreciated! ❤️

BraisGabin:
Addressed, but no released yet.

3flex:
Was this the issue fixed in #4499?

If so we can close this.

JamieSlome:
Would be good to know on the above point, so we can close the report off on our side as well:

https://huntr.dev/bounties/23e37ba7-96d5-4037-a90a-8c8f4a70ce44/

chao2zhang:
[...]

PR #4538, opened 2022-01-28
Create SECURITY.md

JamieSlome:
In response to #4496 - adding the default suggested `SECURITY.md` template.

Usually good to add a point of contact, e-mail or person that should be contacted when potential security issues are discovered.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
+| < 4.0   | :x:                |
+
+## Reporting a Vulnerability
+
+Use this section to tell people how to report a vulnerability.
+
+Tell them where to go, how often they can expect to get an update on a
+reported vulnerability, what to expect if the vulnerability is accepted or
+declined, etc.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,20 +2,11 @@
 
 ## Supported Versions
 
-Use this section to tell people about which versions of your project are
-currently being supported with security updates.
-
-| Version | Supported          |
-| ------- | ------------------ |
-| 5.1.x   | :white_check_mark: |
-| 5.0.x   | :x:                |
-| 4.0.x   | :white_check_mark: |
-| < 4.0   | :x:                |
+| Version | Supported                                                                                                          | Fix   |
+|---------|--------------------------------------------------------------------------------------------------------------------|-------|
+| 1.20.0  | Improper Restriction of XML External Entity Reference - [CWE-611](https://cwe.mitre.org/data/definitions/611.html) | #4499 |
 
 ## Reporting a Vulnerability
 
-Use this section to tell people how to report a vulnerability.
-
-Tell them where to go, how often they can expect to get an update on a
-reported vulnerability, what to expect if the vulnerability is accepted or
-declined, etc.
+Please report vulnerability to security@detekt.dev.
+If you have already reported on vulnerability disclosure platform, please include its link in the email.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,6 +2,10 @@
 
 ## Supported Versions
 
+Generally updating to the latest stable version will have all security issues addressed.
+- Security patches are applied up to the current minor version.
+- Earlier versions are not supported by default, but we will examine them on a case-by-case basis.
+
 | Version | Supported                                                                                                        | Fix                                                 |
 |---------|------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------|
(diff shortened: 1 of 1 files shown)


chao2zhang:
Hello @JamieSlome , would you suggest us updating this PR to fill in the actual information for this repo?

JamieSlome:
@chao2zhang - correct 👍

chao2zhang (approved):

marschwar (approved):

SECURITY.md
```diff
@@ -2,20 +2,11 @@
...
-
-| Version | Supported          |
-| ------- | ------------------ |
-| 5.1.x   | :white_check_mark: |
-| 5.0.x   | :x:                |
-| 4.0.x   | :white_check_mark: |
-| < 4.0   | :x:                |
+| Version | Supported                                                                                                          | Fix   |
+|---------|--------------------------------------------------------------------------------------------------------------------|-------|
+| 1.20.0  | Improper Restriction of XML External Entity Reference - [CWE-611](https://cwe.mitre.org/data/definitions/611.html) | #4499 |
```
BraisGabin: In general we don't give any support for previous versions so, should we add this? I mean, we will only create patch versions for the last version of detekt in case a security issue is found, right? Maybe we should write that in this section.

cortinico: Yes that should be added. Essentially we only support security patches for previous X versions (say 3/4 versions in the past). We provide formal support only for latest Detekt version.

That should be clarified here.

chao2zhang: Do you think we want to start with applying security patches for the latest stable version?

BraisGabin: I'm not sure about that. To me we should only provide security patches for our last version. I mean, maybe if we introduce 2.0 we can give security patches for the las 1.x version but we can talk about that once 2.0 lands.

chao2zhang: We can address 2.0 changes in separate PRs.

SECURITY.md
```diff
@@ -0,0 +1,16 @@
+# Security Policy
+
+## Supported Versions
+
+Generally updating to the latest stable version will have all security issues addressed.
+- Security patches are applied up to the current minor version.
```
cortinico, line 6:
```suggestion
- Security patches are applied up to the **current minor version**.
```

SECURITY.md
```diff
@@ -0,0 +1,16 @@
...
+- Earlier versions are not supported by default, but we will examine them on a case-by-case basis.
+
+| Version | Supported                                                                                                        | Fix                                                 |
+|---------|------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------|
+| 1.20.0  | [CWE-611](https://cwe.mitre.org/data/definitions/611.html) Improper Restriction of XML External Entity Reference | [#4499](https://github.com/detekt/detekt/pull/4499) |
+
+## Reporting a Vulnerability
+
+Please report vulnerability to security@detekt.dev.
```
cortinico, line 15:
```suggestion
Please report vulnerability to [security@detekt.dev](mailto:security@detekt.dev).
```

cortinico (commented):
I would also add the following points:
* We commit to answer within 2 weeks to any email sent to `security@`
* As a point of contact, maintainers of this library can be found also on the `#detekt` channel on KotlinLang Slack (though security vulnerabilities should not be shared in the public channel at first).

Moreover, we should probably move this to detekt/.github as the other tools of this org will be covered by this in some form as well.

chao2zhang:
_Moreover, we should probably move this to detekt/.github as the other tools of this org will be covered by this in some form as well._
I agree on an org level one, but each repository may still list its unique supported versions.

SECURITY.md
```diff
@@ -0,0 +1,17 @@
+# Security Policy
+
+## Supported Versions
+
+Generally updating to the latest stable version will have all security issues addressed.
+- Security patches are applied up to the **current minor version**.
+- Earlier versions are not supported by default, but we will examine them on a case-by-case basis.
+
+| Version | Supported                                                                                                        | Fix                                                 |
```
BraisGabin, line 9: This title is strange. It say "supported" but the value is something like "which security error was fixed in this version"

cortinico: Yes I had the same feeling. Seems like we're mixing which versions we do support with which security issues we found and fixed in the past.

chao2zhang: What do you think about the latest version?

SECURITY.md
```diff
@@ -0,0 +1,17 @@
...
+- Earlier versions are not supported by default, but we will examine them on a case-by-case basis.
+
+| Version | Addressed issues                                                                                                 | Fix                                                 |
+|---------|------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------|
+| 1.20.0  | [CWE-611](https://cwe.mitre.org/data/definitions/611.html) Improper Restriction of XML External Entity Reference | [#4499](https://github.com/detekt/detekt/pull/4499) |
+
+## Reporting a Vulnerability
+
+Please report vulnerability to security@detekt.dev. 
+We commit to respond within 2 weeks. You may also find us in #detekt channel of [kotlinlang Slack](https://kotlinlang.slack.com/).
```
cortinico, line 16:
```suggestion
We commit to respond within 2 weeks. You may also find us in the [#detekt](https://kotlinlang.slack.com/archives/C88E12QH4) channel of [kotlinlang Slack](https://kotlinlang.slack.com/).
```

SECURITY.md
cortinico: LGTM 👍

PR #4541, opened 2022-01-29
Make DoubleMutabilityForCollection configurable and set a DoubleMutability alias

### LINKED ISSUE #4501: Rule suggestion: In Compose, forbid `var` with `mutableStateOf` (whether wrapped in `remember` or not)
In Compose, MutableState can be accessed in a variety of ways, see https://developer.android.com/jetpack/compose/state#state-in-composables.

One of these ways is to declare the `MutableState` as a `var` and then access the state value directly via property delegation:

```
var foo by mutableStateOf(something)
```

However, this introduces the use of `var`s into Composables, which can easily be avoided by deconstructing the getter and setter properties and declaring them as `val`:

```
val (foo, setFoo) = mutableStateOf(something)
```

Examples that would trigger the rule:

```
var foo by mutableStateOf(something)
var foo by remember { mutableStateOf(something) }
```

## Context
I believe this results in cleaner code and better compile-time safety through the use of `val`s rather than `var`s.

rocketraman:
Resolves #4501

### DIFF
--- a/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DoubleMutabilityForCollectionSpec.kt
+++ b/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DoubleMutabilityForCollectionSpec.kt
@@ -4,9 +4,12 @@ import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.rules.KotlinCoreEnvironmentTest
 import io.gitlab.arturbosch.detekt.test.assertThat
 import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
+import io.gitlab.arturbosch.detekt.test.TestConfig
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
+
+private const val MUTABLE_TYPES = "mutableTypes"
 
 @KotlinCoreEnvironmentTest
 class DoubleMutabilityForCollectionSpec(private val env: KotlinCoreEnvironment) {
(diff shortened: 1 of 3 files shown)


detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/DoubleMutabilityForCollectionSpec.kt
```diff
@@ -489,6 +530,26 @@ class DoubleMutabilityForCollectionSpec : Spek({
                     assertThat(result).hasSize(1)
                     assertThat(result).hasSourceLocation(2, 5)
                 }
+
+                it("detects var declaration with MutableState, when configured") {
```
cortinico, line 534: Could we also add a test where you declare and use those functions:

```
fun <T : Any?> mutableStateOf(value: T?): MutableState<T>
fun <T : Any?> remember(calculation: (() -> T)?): T
```

So to make sure everything works correctly in a Compose scenario

cortinico (commented):
Great stuff! 🚀

rocketraman:
@cortinico ready for re-review.

Note, I also tried to add a negative test to validate that property delegates are ignored:

```
            describe("ignores declaration with var and property delegation") {
                val rule = DoubleMutabilityForCollection(
                    TestConfig(
                        mapOf(
                            MUTABLE_TYPES to listOf("MutableState")
                        )
                    )
                )

                val code = """
                    data class MutableState<T>(var state: T)
                    fun <T> mutableStateOf(value: T): MutableState<T>
                    fun <T> remember(calculation: () -> T): T
                    inline operator fun <T> MutableState<T>.getValue(thisObj: Any?, property: KProperty<*>): T = value
                    inline operator fun <T> MutableState<T>.setValue(thisObj: Any?, property: KProperty<*>, value: T) {
                        this.value = value
                    }
                    fun main() {
                        var myState by remember { mutableStateOf("foo") }
                    }
                    """
                val result = rule.compileAndLintWithContext(env, code)
                assertThat(result).isEmpty()
            }
```

but got an error:

```
    java.lang.AssertionError: 'env' can not be accessed in this context.
        at org.spekframework.spek2.runtime.lifecycle.MemoizedValueAdapter.get(MemoizedValueAdapter.kt:33)
        at org.spekframework.spek2.runtime.lifecycle.MemoizedValueAdapter.getValue(MemoizedValueAdapter.kt:22)
        at io.gitlab.arturbosch.detekt.rules.bugs.DoubleMutabilityForCollectionSpec$1.invoke$lambda-0(DoubleMutabilityForCollectionSpec.kt:17)
...
```

Any ideas? Seems like the property delegate breaks it.

cortinico:
> at org.spekframework.spek2.runtime.lifecycle.MemoizedValueAdapter.get(MemoizedValueAdapter.kt:33)
>         at org.spekframework.spek2.runtime.lifecycle.MemoizedValueAdapter.getValue(MemoizedValueAdapter.kt:22)

Please note that we're in the process of migrating tests from Spek to JUnit.
Specifically this PR is going to clash with https://github.com/detekt/detekt/pull/4523. Potentially adapting those tests would fix your failure

rocketraman:
NOTE: GitHub is not letting me set the junit branch as the base of this PR, so the junit commits are here too.

cortinico:
You should now be able to rebase 👍

rocketraman:
> You should now be able to rebase +1

Done

cortinico (approved):
Great contrib 🙏 LGTM

BraisGabin (approved):
I really like this one! Thanks :)

PR #4542, opened 2022-01-29
Code cleanups

Goooler:
- Remove redundant SAM.
- Proper accessors in kotlin.
- Remove unneeded generic type.
- Inline local variable.
- Const values.
- `Sort` in kotlin.
- `Until` in kotlin.

### DIFF
--- a/detekt-gradle-plugin/src/functionalTest/resources/jvm/src/main/kotlin/Errors.kt
+++ b/detekt-gradle-plugin/src/functionalTest/resources/jvm/src/main/kotlin/Errors.kt
@@ -8,6 +8,7 @@ class Errors {
     }
 
     fun javaExit() {
+        @Suppress("ReplaceJavaStaticMethodWithKotlinAnalog")
         System.exit(1)
     }
 }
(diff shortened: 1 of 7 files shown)


detekt-gradle-plugin/src/functionalTest/resources/jvm/src/main/kotlin/Errors.kt
```diff
@@ -8,6 +8,7 @@ class Errors {
     }
 
     fun javaExit() {
+        @Suppress("ReplaceJavaStaticMethodWithKotlinAnalog")
```
cortinico, line 11: nit: Why this? Was it triggering a warning of any form? If so where?

cortinico (approved):
Overall looks good 👍 Just a minor nit

detekt-gradle-plugin/src/functionalTest/resources/jvm/src/main/kotlin/Errors.kt
Goooler: As I see a suggestion by IDEA to replace `System.exit(1)` with `exitProcess(1)`, it's not proper right here?

cortinico: > it's not proper right here?

I think it's ok. This is a test resource, so we're not actually executing it and should be treated as plain-text. However, as the IDE is recognizing that is Kotlin code, it's probably ok to have it.

schalkms (approved):

PR #4546, opened 2022-01-30
Reuse `setReportOutputConventions`

Goooler:
Reuse the same logic.


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektMultiplatform.kt
```diff
@@ -168,6 +143,23 @@ internal class DetektMultiplatform(private val project: Project) {
     }
 }
 
+internal fun Project.outputConventions(reports: DetektReports, extension: DetektExtension, name: String) {
```
cortinico, line 146: The name `outputConventions` here is really unfortunate imho. I would probably rename to `setReportOutputConventions` and `setReportOutputConvention`

Goooler: c9f3cd595f5f6368f126740653ec96bf7de5c86d

cortinico (approved):

PR #4551, opened 2022-02-01
Deprecate continuationIndentSize from the Indentation rule

### LINKED ISSUE #2970: Indentation rule stopped working after upgrading to version 1.10.0
## Expected Behavior
In our project we have Indentation rule turned on with the following rule settings:

```
  Indentation:
    active: true
    autoCorrect: true
    indentSize: 2
    continuationIndentSize: 4
```
It is done in accord with our project settings:
image

It worked prior to version 1.10.0

## Observed Behavior
We tried upgrading detekt to version 1.10.0 and got 13k errors related to Indentation rule. It says we have wrong indent and the auto correction turned on tries to reformat the code.

The example code that fails Indentation rule after upgrading:
```kotlin
  override fun getAddresses() = async {
    addressRepository.getAll()
        .toVO { address -> address.id == id }
  }

```
Detekt says that the problem is on the line `.toVO { address -> address.id == id }`

I know that the new version of KtLint now uses a new Indentation rule that was in experimental state earlier, but I found no mentions that something should break and no guidelines on how to fix it.

## Your Environment

* Version of detekt used: 1.10.0
* Version of Gradle used (if applicable): 6.4
* Operating System and version: MacOSX 10.15.4 Catalina

cortinico:
Users should not use the `continuationIndentSize` as it's essentially ignored by KtLint. Sadly we don't have a way to deprecate properties from rules inside `formatting` (cc @BraisGabin, were you aware of it? Should we offer this capacity?)

Fixes #2970

### DIFF
--- a/detekt-core/src/main/resources/deprecation.properties
+++ b/detekt-core/src/main/resources/deprecation.properties
@@ -7,3 +7,4 @@ style>FunctionOnlyReturningConstant>excludeAnnotatedFunction=Use `ignoreAnnotate
 style>UnderscoresInNumericLiterals>acceptableDecimalLength=Use `acceptableLength` instead
 style>UnnecessaryAbstractClass>excludeAnnotatedClasses=Use `ignoreAnnotated` instead
 style>UseDataClass>excludeAnnotatedClasses=Use `ignoreAnnotated` instead
+formatting>Indentation>continuationIndentSize=`continuationIndentSize` is ignored by KtLint and will have no effect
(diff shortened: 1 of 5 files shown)


BraisGabin:
No, I wasn't aware of it.

A quick workaround would be to add a line here: https://github.com/detekt/detekt/blob/main/detekt-core/src/main/resources/deprecation.properties. We could hardcode it in our `detekt-generator` but it should work and I think that it have sense to have it there for 2 or 3 minor versions.

Then, if we want to provide this option to any developer, we can open an issue and work on a proper solution. It shouldn't be too difficult but I vote to don't do this until a user ask for it.

BraisGabin (approved):

PR #4556, opened 2022-02-03
Config to ignore function/constructor annotations from LongParameterList check

FabioCZ:
In our codebase, we have several common function/constructor annotations that tend to come with long parameter lists. We'd like to add the opportunity to ignore functions/constructors with those annotations from the LongParameterList check.

Adds two new config fields for `LongParameterList`:
* `ignoreAnnotatedFunctions` - functions with the annotations specified in this config array will be excluded from the check
* `ignoreAnnotatedConstructors` - constructors with the annotations specified in this config array will be excluded from the check

Example:
our codebase is an Android application, and we're adding a lot of new functions with `@Composable` annotations. These functions tend to have a lot of parameters, and we'd like to specifically exclude these from the check rather than having to suppress each one manually.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -123,6 +123,8 @@ complexity:
     ignoreDefaultParameters: false
     ignoreDataClasses: true
     ignoreAnnotatedParameter: []
+    ignoreAnnotatedFunctions: []
+    ignoreAnnotatedConstructors: []
   MethodOverloading:
     active: false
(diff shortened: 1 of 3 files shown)


cortinico (commented):
Thanks for the PR @FabioCZ 
I believe this capability is already supported using the `ignoreAnnotated` feature we support for all the rules. @BraisGabin can maybe confirm.

FabioCZ:
@cortinico omg, you're absolutely right! Thanks for the recommendation, I will make sure to read the documentation better next time :) Closing PR.

PR #4562, opened 2022-02-05
Fix: Excludes is not evaluated when includes and excludes both specified

sakuna63:
The comment of `PathFilters#isIgnored` says 

```
If [includes] and [excludes] are both specified, return false iff [path] matches any [includes] and [path] does not match any [excludes].
```

but `excludes` is not evaluated when includes and excludes both specified.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/PathFilters.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/PathFilters.kt
@@ -26,10 +26,14 @@ class PathFilters internal constructor(
      *   return false iff [path] matches any [includes] and [path] does not match any [excludes].
      */
     fun isIgnored(path: Path): Boolean {
-        fun isIncluded() = includes?.any { it.matches(path) }
-        fun isExcluded() = excludes?.any { it.matches(path) }
+        val isIncluded = includes?.any { it.matches(path) }
+        val isExcluded = excludes?.any { it.matches(path) }
 
-        return isIncluded()?.not() ?: isExcluded() ?: true
+        return if (isIncluded != null && isExcluded != null) {
+            return !isIncluded || isExcluded
+        } else {
+            isIncluded?.not() ?: isExcluded ?: true
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


sakuna63:
hmm...some test failed. In tests, current `PathFilters#isIgnored` behavior seems to be used as correct behavior. So, should I fix function comment instead of its implementation?

cortinico:
> hmm...some test failed. Before fixing tests, I want to clarify what should i do.

Can you provide more context on why you submitted this change?
Anyway, you're right, the function looks wrongly implemented at first but it's actually prioritizing `include` over `exclude`. I'm unsure why we decided for this approach. @chao2zhang do you happen to know?

chao2zhang:
I think my last change merely added documentation from my understanding of reading the code statically.

_but excludes is not evaluated when includes and excludes both specified._
I agree with this - This documentation is misleading in this sense. Hence I would suggest updating the documentation before making any behavioral changes. Many repos have been configuring includes and excludes extensively.

sakuna63:
@cortinico 

We have multi-module project like a below and want to apply rule to `module1/package1-1` and `module2/package2-2 ~ module2/package2-N`.

```
project-root
├module1
│  ├package1-1
│  ├package1-2
│  ├package1-3
│  ├...
│  └package1-N
└module2
  　├package2-1
　  ├package2-2
  　├package2-3
  　├...
　  └package2-N
```

First, I defined include/exclude like below but it doesn't behave expectedly because of the problem described in [this PR description](https://github.com/detekt/detekt/pull/4562#issue-1124908469).

```yaml
includes:
  - '**/module1/package1-1/**'
  - '**/module2/**'
excludes:
  - '**/module2/package2-1/**'
```

I know I can define include/exclude rule like below but it is too lengthy and easy to forget to add include if new package added.

```yaml
includes:
  - '**/module1/package1-1/**'
  - '**/module2/package2-2/**'
  - '**/module2/package2-3/**'
  - ...
  - '**/module2/package2-N/**'
```

After some try, I read detekt source code and find `PathFilters#isIgnored` is not followed the commented behavior. So I submit this PR.

cortinico:
> After some try, I read detekt source code and find `PathFilters#isIgnored` is not followed the commented behavior. So I submit this PR.

As @chao2zhang mentioned, the problem here is the documentation not being up to date.

cortinico:
> As @chao2zhang mentioned, the problem here is the documentation not being up to date.

Can we revive this and just update the documentation OR close it?

Issue #4560, opened 2022-02-05
Drop Groovy DSL tests in detekt-gradle-plugin


3flex:
## Expected Behavior
Efficient test suite. We shouldn't run tests that don't provide any extra assurances about the behaviour of the program. I propose dropping all Groovy DSL tests to improve test run time. The detekt-gradle-plugin test suite is one of the slowest in the project.

## Observed Behavior
Almost all of the Gradle plugin tests are run twice - once using the Groovy DSL, again using the Kotlin DSL.

## Steps to Reproduce
N/A

## Context
As far as I'm aware, the Gradle Kotlin DSL is always more strict than the Groovy DSL, with a couple of exceptions. This suggests that testing using Kotlin DSL only would be sufficient.

The exceptions are:
* [Calling Kotlin extension functions from Groovy](https://docs.gradle.org/7.3.3/userguide/kotlin_dsl.html#static_extensions) requires different syntax. This issue can be avoided by avoiding extension functions entirely. The plugin doesn't currently have any as part of its public API.
* [Calling Kotlin functions with named arguments from Groovy](https://docs.gradle.org/7.3.3/userguide/kotlin_dsl.html#calling_kotlin_from_groovy) requires use of positional parameters, and can't be called with named arguments from Groovy. I don't believe there are any such functions as part of the plugin's public API.
* [Calling Kotlin functions with default arguments from Groovy](https://docs.gradle.org/7.3.3/userguide/kotlin_dsl.html#calling_kotlin_from_groovy). Values must be passed for all parameters. Again, I'm not aware of any such public functions in the detekt plugin.

We may be able to add custom detekt rules to prevent this occurring, though I don't think that should be a blocker.

Otherwise, Kotlin is more strict. From https://docs.gradle.org/7.3.3/userguide/migrating_from_groovy_to_kotlin_dsl.html#prepare_your_groovy_scripts:
* Kotlin requires double quotes for all strings. In Groovy single quoted strings are also acceptable.
* Kotlin requires parentheses when invoking functions. In Groovy it's optional.
* Kotlin requires the `=` assignment operator when assigning properties. In Groovy it's optional.

[...]

PR #4563, opened 2022-02-06
Drop Groovy DSL testing in DetektTaskDslSpec

3flex:
See #4560

Because I'm partway through migrating detekt-gradle-plugin tests to JUnit, it's easier for me to update this file first, then finish the JUnit migration for this project, then drop Groovy DSL tests in the remaining files.

First commit is the functional change without modifying whitespace, the second is just a reformat of the file to make it easier to review.


BraisGabin (approved):

cortinico (approved):
Accepting to unblock. However please take a look at my comment on https://github.com/detekt/detekt/issues/4560 as I believe we should at least reintroduce a couple of Groovy tests (in a separate file maybe?).

PR #4575, opened 2022-02-08
Migrate detekt-rules-style tests to JUnit

3flex:
#4450

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExplicitCollectionElementAccessMethodSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExplicitCollectionElementAccessMethodSpec.kt
@@ -1,24 +1,27 @@
 package io.gitlab.arturbosch.detekt.rules.style
 
+import io.github.detekt.test.utils.createEnvironment
 import io.github.detekt.test.utils.resourceAsPath
 import io.gitlab.arturbosch.detekt.api.Config
-import io.gitlab.arturbosch.detekt.rules.setupKotlinEnvironment
+import io.gitlab.arturbosch.detekt.rules.KotlinCoreEnvironmentTest
 import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
 import io.gitlab.arturbosch.detekt.test.lintWithContext
 import org.assertj.core.api.Assertions.assertThat
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
-import org.spekframework.spek2.Spek
-import org.spekframework.spek2.style.specification.describe
+import org.junit.jupiter.api.AfterAll
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
 
-class ExplicitCollectionElementAccessMethodSpec : Spek({
-    setupKotlinEnvironment(additionalJavaSourceRootPath = resourceAsPath("java"))
+@KotlinCoreEnvironmentTest
+class ExplicitCollectionElementAccessMethodSpec(val env: KotlinCoreEnvironment) {
+    val subject = ExplicitCollectionElementAccessMethod(Config.empty)
 
-    val env: KotlinCoreEnvironment by memoized()
-    val subject by memoized { ExplicitCollectionElementAccessMethod(Config.empty) }
+    @Nested
+    inner class `Kotlin map` {
 
-    describe("Kotlin map") {
-
-        it("reports map element access with get method") {
+        @Test
+        fun `reports map element access with get method`() {
             val code = """
                     fun f() {
                         val map = mapOf<String, String>() 
@@ -28,7 +31,8 @@ class ExplicitCollectionElementAccessMethodSpec : Spek({
             assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
         }
 
-        it("does not report safe map element access") {
+        @Test
+        fun `does not report safe map element access`() {
             val code = """
                     fun f() {
                         val map = mapOf<String, String>() 
@@ -38,7 +42,8 @@ class ExplicitCollectionElementAccessMethodSpec : Spek({
             assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
         }
 
-        it("reports map set method usage with unused return value") {
+        @Test
+        fun `reports map set method usage with unused return value`() {
             val code = """
                     fun f() {
                         val map = mutableMapOf<String, String>() 
@@ -48,7 +53,8 @@ class ExplicitCollectionElementAccessMethodSpec : Spek({
             assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
         }
 
-        it("reports map put method usage with unused return value") {
+        @Test
+        fun `reports map put method usage with unused return value`() {
             val code = """
                     fun f() {
                         val map = mutableMapOf<String, String>()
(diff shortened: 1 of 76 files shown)


marschwar:
Is there anything I can do to support? Monkey work is my specialty ;) .

3flex:
That's ok, I'll try and wrap it up this weekend! Thanks though.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambdaSpec.kt
```diff
@@ -360,51 +388,70 @@ class ObjectLiteralToLambdaSpec : Spek({
...
-                import com.example.fromjava.SamWithDefaultMethods
-                
-                fun main() {
-                    val x = object : SamWithDefaultMethods {
-                        override fun foo() {
-                            println()
+            @Nested
+            inner class JavaSourceTests {
+
+                private val environmentWrapper =
```
3flex, line 394: :eyes:

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExplicitCollectionElementAccessMethodSpec.kt
```diff
@@ -374,16 +421,30 @@ class ExplicitCollectionElementAccessMethodSpec : Spek({
...
+        @Nested
+        inner class JavaSourceTests {
 
-                fun foo() {
-                    val rect = Rect()
-                    rect.set(0, 1, 2)
-                }
-            """
-            assertThat(subject.lintWithContext(env, code)).isEmpty()
+            private val environmentWrapper =
```
3flex, line 427: :eyes:

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumberSpec.kt
```diff
@@ -741,74 +869,90 @@ class MagicNumberSpec : Spek({
...
-                it("'$codeWithMagicNumberInRange' reports no finding if ranges are ignored") {
-                    val code = codeWithMagicNumberInRange
-                    assertThat(MagicNumber(TestConfig(mapOf(IGNORE_RANGES to "true"))).lint(code))
-                        .isEmpty()
-                }
+            )
+
+            @ParameterizedTest
+            @MethodSource("cases")
+            fun `'$codeWithMagicNumberInRange' reports a code smell by default`(code: String) {
```
marschwar, line 903:
```suggestion
            fun `reports a code smell by default`(code: String) {
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumberSpec.kt
```diff
@@ -741,74 +869,90 @@ class MagicNumberSpec : Spek({
...
+                assertThat(MagicNumber().lint(code)).hasSize(1)
             }
 
-            it("reports a finding for a parenthesized number if ranges are ignored") {
+            @ParameterizedTest
+            @MethodSource("cases")
+            fun `'$codeWithMagicNumberInRange' reports a code smell if ranges are not ignored`(code: String) {
```
marschwar, line 909:
```suggestion
            fun `reports a code smell if ranges are not ignored`(code: String) {
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MagicNumberSpec.kt
```diff
@@ -741,74 +869,90 @@ class MagicNumberSpec : Spek({
...
+                assertThat(MagicNumber(TestConfig(mapOf(IGNORE_RANGES to "false"))).lint(code))
+                    .hasSize(1)
+            }
+
+            @ParameterizedTest
+            @MethodSource("cases")
+            fun `'$codeWithMagicNumberInRange' reports no finding if ranges are ignored`(code: String) {
```
marschwar, line 916:
```suggestion
            fun `reports no finding if ranges are ignored`(code: String) {
```

marschwar (approved):
Thank you for taking this. I did not realize how many style rules there are.

chao2zhang (approved):
Approve because 
1. Coverage does not seem to change 84.58% -> 84.58%
2. All checks are passing
3. Scroll through the PR briefly and did not find undesired changes.
I don't think it's worth time reading every line of code changes

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExplicitCollectionElementAccessMethodSpec.kt
marschwar: I was going to suggest to extend the annotation to something like `@KotlinCoreEnvironmentTest(additionalJavaSourcePaths = ["java"])`. But something seems to be wrong with the test as it does not fail even when the regular `env` is used.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambdaSpec.kt
marschwar: How about this?

```kotlin
class ObjectLiteralToLambdaSpec() {
    val subject = ObjectLiteralToLambda()

    @Nested
    @KotlinCoreEnvironmentTest
    inner class DefaultSourceTests(val env: KotlinCoreEnvironment) {
      // ...
    }

    @Nested
    @KotlinCoreEnvironmentTest(additionalJavaSourcePaths = ["java"])
    inner class JavaSourceTests(val env: KotlinCoreEnvironment) {
      // ...
    }
}

marschwar: It would be really good, If test authors do not have to create and dispose the environment manually

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExplicitCollectionElementAccessMethodSpec.kt
3flex: > it does not fail even when the regular env is used.

I noticed this too, but I only picked up that that I didn't use the custom environment here when the coverage report showed a change in coverage for the rule under test.

Agree this doesn't seem to be testing what it appears to be at first glance, but fixing that can be done another time.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ObjectLiteralToLambdaSpec.kt
3flex: > It would be really good, If test authors do not have to create and dispose the environment manually

Agree though out of the ~1300 style tests there are only 4 using this setup, so it's very much an exception, so perhaps the manual option is OK?

As for options, I did try what you suggested above, but annotation parameters cannot be nullable, so this would mean every time the annotation is used you'd have to pass an empty list to the annotation.

We could instead have a separate annotation like `@KotlinCoreEnvironmentTestWithJavaRoot` that accepts the parameter. It feels awkward though to use just for these few cases, but it would be good to have a general solution for including Java sources in a test like this.

I'll wait for feedback.

marschwar: Annotation parameters May not be null, but they can have defaults. I did get this to work but the Code is not ready to be commited. I think it is fine to merge this and come back to it later.

marschwar:
Any objections to merging this?

chao2zhang:
Not from my side. Let's do it.

PR #4576, opened 2022-02-08
Enable Gradle's configuration cache by default

3flex:
Gradle 7.4 introduces a feature to declare that individual tasks are incompatible with the configuration cache, allowing the feature to be enabled by default while skipping caching when there are tasks in the task graph that don't support it.

More information available in "Disable configuration caching when incompatible tasks are executed" section of the release notes: https://docs.gradle.org/7.4/release-notes.html#config-cache

This PR enables the configuration cache while declaring tasks known to be incompatible with it. Note that there may be some additional tasks that are discovered to be incompatible in future, so this config might need tweaking.

### DIFF
--- a/.github/workflows/codecoverage.yaml
+++ b/.github/workflows/codecoverage.yaml
@@ -25,7 +25,7 @@ jobs:
       - name: Generate Coverage Report
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: --configuration-cache jacocoMergedReport
+          arguments: jacocoMergedReport
 
       - name: Publish Coverage
         if: success()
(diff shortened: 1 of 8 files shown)


cortinico (approved):

3flex:
I've updated the label to "housekeeping" since I thought the yellow labels, including "gradle", related to the named subprojects? This doesn't change behaviour of the detekt Gradle plugin at all, just the detekt build itself.

Happy to switch back though if that makes sense when it comes to generating release notes.

cortinico:
> I've updated the label to "housekeeping" since I thought the yellow labels, including "gradle", related to the named subprojects? This doesn't change behaviour of the detekt Gradle plugin at all, just the detekt build itself.

Yup that's better. I've initially added `gradle` as I expected some changes on some of our tasks. Give that there are none, it's safe to move to housekeeping

Goooler (approved):

BraisGabin (approved):

PR #4589, opened 2022-02-19
Housekeep Gradle scripts

chao2zhang:
- Move our own Gradle scripts to the new reporting configuration API.
- Use Constants defined in `DetektExtension`


3flex:
> Concatenate all include statements into one.

Can we not do this please? It makes it a bit more annoying to tweak the settings script when using Gradle's composite build feature, where you need to change `include` directives to `includeBuild` and back for example.

Unless there's some advantage to doing this I'm not aware of?

chao2zhang:
The only advantage is the visual appearance - I have reverted the change.

3flex (approved):

PR #4611, opened 2022-03-06
Allow users and rule authors to specify a reason per configured value

marschwar:
This PR is still somewhat WIP but I would like your input on it before giving it the final polish.

**What does this PR address?**
A few times there was the request to be able to specify a custom error message or a reason for certain setting. See #3501 for example. This will allow rule authors to enable this features for all rules that have configurable list of string values.

**What I would like your input and feedback on?**
* I do not really like the nameof the `ExplainedValues` type and `explainedValues` config delegate. Any idea what we could call the combination of a configured value and the reasoning why the value is configured?
* Do you agree on the representation in the `config.yml` (see `detekt-core/src/test/resources/explained-values.yml`)?
* Should the reason be shown in the documentation? If so, any ideas on how to do that? (This would be a separate PR)

*Thanks to @BraisGabin for collaborating on this.*


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/ConfigProperty.kt
```diff
@@ -129,6 +123,38 @@ private fun <T : Any> getValueOrDefault(configAware: ConfigAware, propertyName:
...
+                .map { it as Map<*, *> }
+                .map { dict ->
+                    ExplainedValue(
+                        value = dict["value"].toString(),
+                        reason = dict["reason"]?.toString()
+                    )
+                }
+        )
+    }
+    error("Only lists of strings or maps with keys 'value' and 'reason' are supported. '$propertyName' is invalid. ")
```
BraisGabin, line 155:
```suggestion
    error("Only lists of strings or maps with keys 'value' and 'reason' are supported. '$propertyName' is invalid.")
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/ConfigProperty.kt
```diff
@@ -129,6 +123,38 @@ private fun <T : Any> getValueOrDefault(configAware: ConfigAware, propertyName:
...
+    }
+}
+
+private fun ConfigAware.getExplainedValuesOrDefault(
+    propertyName: String,
+    defaultValue: ExplainedValues
+): ExplainedValues {
+    val valuesAsList: List<*> = valueOrNull(propertyName) ?: return defaultValue
+    if (valuesAsList.all { it is String }) {
+        return ExplainedValues(values = valuesAsList.map { ExplainedValue(it.toString()) })
```
BraisGabin, line 141:
```suggestion
        return ExplainedValues(values = valuesAsList.map { ExplainedValue(it as String) })
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/ConfigProperty.kt
```diff
@@ -129,6 +123,38 @@ private fun <T : Any> getValueOrDefault(configAware: ConfigAware, propertyName:
...
+    }
+    if (valuesAsList.all { it is Map<*, *> }) {
+        return ExplainedValues(
+            valuesAsList
+                .map { it as Map<*, *> }
+                .map { dict ->
+                    ExplainedValue(
+                        value = dict["value"].toString(),
+                        reason = dict["reason"]?.toString()
```
BraisGabin, line 150: `.toString()`  or `as String`? I think that `as String` is better so we can fail fast.

detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/ConfigPropertySpec.kt
```diff
@@ -145,6 +146,56 @@ class ConfigPropertySpec {
...
+                        assertThat(subject.notPresent).isEqualTo(defaultValue)
+                    }
+                }
+
+                @Nested
+                inner class `value defined as list of maps` {
+                    private val subject = object : TestConfigAware(
+                        "present" to listOf(
+                            mapOf("value" to "a", "reason" to "reasonA"),
+                            mapOf("value" to "b", "reason" to null)
```
BraisGabin, line 179: Should we handle something like this too?
```suggestion
                            mapOf("value" to "b"), // missing reason
```
I think that usually we will find this case more often than having a `null` reason.

BraisGabin (commented):
>     * I do not really like the nameof the `ExplainedValues` type and `explainedValues` config delegate. Any idea what we could call the combination of a configured value and the reasoning why the value is configured?

`ValueWithReason`? I don't think it's better... it's the only other option that I find.

>     * Do you agree on the representation in the `config.yml` (see `detekt-core/src/test/resources/explained-values.yml`)?

I like it. 

>     * Should the reason be shown in the documentation? If so, any ideas on how to do that? (This would be a separate PR)

I don't think so. I don't think that it could provide any value for the reader.

detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/ConfigPropertySpec.kt
marschwar: Good call. I added this case

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/ConfigProperty.kt
marschwar: I see your point. I changed it to throw a ConfigurationError in those cases

cortinico:
I haven't looked into the code, but if I understand correctly it relies on a separate [explained-values.yml](https://github.com/detekt/detekt/pull/4611/files#diff-0a7ce5ad1be62e85a6297c2bd7ede3a8e7a8d7f2e5717d8e9a41fdc7badada84) file right?

I would rather move this file inside the config on a top level `explanation:` key or even move it closer to the rule config as a metadata addition

marschwar:
> if I understand correctly it relies on a separate [explained-values.yml](https://github.com/detekt/detekt/pull/4611/files#diff-0a7ce5ad1be62e85a6297c2bd7ede3a8e7a8d7f2e5717d8e9a41fdc7badada84) file right?

No, that is not correct the `explained-values.yml` is only used in the tests to verify that the config parser can read those formats correctly. The idea is, that rule authors may decide that their rule supports values with reasons instead of values only. So instead of 

```yaml
  ForbiddenImport:
    active: true
    imports:
      - 'org.assertj.core.api.Assertions'
```
it could become 

```yaml
  ForbiddenImport:
    active: true
    imports:
      - value: 'org.assertj.core.api.Assertions'
        reason: 'import Assertions.assertThat directly'
```

cortinico:
> The idea is, that rule authors may decide that their rule supports values with reasons instead of values only. So instead of

Gotcha 👍 Looks great then.

marschwar:
Could we add this feature with 1.21?

BraisGabin (approved):
Sorry for the late review! I missed the notification about this.

BraisGabin:
@marschwar I think that we can merge this. Could you fix the conflicts?

marschwar:
Ready to be merged from my point of view unless you would like me to add more tests?

schalkms (approved):
Good to go
Thank you!

PR #4613, opened 2022-03-07
NamedArguments: Ignore when argument values are the same as the parameter name

### LINKED ISSUE #4591: NamedArguments: Ignore when argument values are the same as the parameter names
The idea is to introduce a new setting for `NamedArguments` called something like `ignoreSimpleArgumentValues` (preferably a better name :-) ). When enabled, the rule would not require named arguments in cases where the argument values are variables whose name are the same as the corresponding parameters.

Example:
```kotlin
import java.time.LocalDate

// Obviously made-up function which has parameter names which are roughly as long as the ones in our codebase
fun processCandidateForMission(
    birthName: String,
    dateOfBirth: LocalDate,
    fitessForPurpose: Float,
    willingToParticipate: Boolean,
    licensedToKill: Boolean
): Unit = TODO()
fun calculateFitness(): Float = TODO()

// 1) Should be reported - not obvious what the values are
processCandidateForMission("John Doe", LocalDate.of(2000, 1, 1), calculateFitness(), true, false)

// 2) Should not be reported - we have added the required parameter names which helps a lot in readability
processCandidateForMission(
    birthName = "John Doe",
    dateOfBirth = LocalDate.of(2000, 1, 1),
    fitessForPurpose = calculateFitness(),
    willingToParticipate = true,
    licensedToKill = false
)

// Let's define variables matching the parameters.
// We often do this in our codebase, though the expressions are often more complex.
val birthName = "John Doe"
val dateOfBirth = LocalDate.of(2000, 1, 1)
val fitessForPurpose = calculateFitness()
val willingToParticipate = true
val licensedToKill = false

// 3) Is currently reported but should not IMO - adding parameter names will not add anything
processCandidateForMission(birthName, dateOfBirth, fitessForPurpose, willingToParticipate, licensedToKill)

// 4) Parameter names are just verbose and not really helpful - this duplication is what I want to avoid
//    It turns a simple one-line function call without any complexity into an ugly seven-line monster of 
//    verbosity and duplication. ;-) 
processCandidateForMission(
    birthName = birthName,
    dateOfBirth = dateOfBirth,
    fitessForPurpose = fitessForPurpose,
    willingToParticipate = willingToParticipate,
    licensedToKill = licensedToKill
)
```

## Context
I want to enable the `NamedArguments` rule in our codebase. In many cases it helps making code more readable. However, we have many cases where we define local variables with the same names as the parameters, and then use these variables as parameter values. In those cases adding named parameters is just redundant and makes the code _less_ readable.

The idea is similar to where IntelliJ automatically annotates the code with "virtual" parameter names and/or type declarations when it considers it helpful. With the example above, IntelliJ annotates call (1) with parameter names but it does not annotate (3), presumably because it would not add any information.

I do appreciate that this may be a special case, but at least in our codebase it is common. :-)

t-kameyama:
Fixes #4591

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -129,6 +129,7 @@ complexity:
   NamedArguments:
     active: false
     threshold: 3
+    ignoreArgumentsMatchingNames: false
   NestedBlockDepth:
     active: true
     threshold: 4
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArguments.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArguments.kt
@@ -44,6 +44,9 @@ class NamedArguments(config: Config = Config.empty) : Rule(config) {
     @Configuration("number of parameters that triggers this inspection")
     private val threshold: Int by config(defaultValue = 3)
 
+    @Configuration("ignores when argument values are the same as the parameter names")
+    private val ignoreArgumentsMatchingNames: Boolean by config(defaultValue = false)
+
     override fun visitCallExpression(expression: KtCallExpression) {
         if (bindingContext == BindingContext.EMPTY) return
         val valueArguments = expression.valueArguments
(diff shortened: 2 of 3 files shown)


BraisGabin (approved):

marschwar:
Should those examples below be considered ok or not?

```kotlin
class Baz {
    private var b: Int = 42
    fun foo(a: Int, b: Int, c: Int) {}
    fun bar(a: Int, c: Int) {
        foo(a, this.b, c)
    }
}
```

```kotlin
data class Baz(val b: Int)
fun foo(a: Int, b: Int, c: Int) {}
fun bar(a: Int, c: Int, baz: Baz?) {
    baz?.let { foo(a, it.b, c) }
}
```

marschwar (approved):

cortinico (approved):

Issue #4587, opened 2022-02-18
New rule: suspend fun with CoroutineScope as Receiver


hfhbd:
## Expected Behavior of the rule
```kt
suspend fun CoroutineScope.foo() { }
```
You should not use `suspend` and `CoroutineScope`. Instead use `suspend fun foo() = coroutineScope`.

## Context
IntelliJ warning: `Ambiguous coroutineContext due to CoroutineScope receiver of suspend function `
IntelliJ fix: `Remove receiver & wrap with coroutineScope { ... }`

BraisGabin:
Thanks for the idea of rule! It have a lot of sense. Do you want to give it a try? It shouldn't be a difficult rule. If you need any help just ask for it :)

PR #4616, opened 2022-03-07
Add SuspendFunWithCoroutineScopeReceiver Rule

hfhbd:
Fixes https://github.com/detekt/detekt/issues/4587

### DIFF
--- a/detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunWithCoroutineScopeReceiverSpec.kt
+++ b/detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunWithCoroutineScopeReceiverSpec.kt
@@ -1,0 +1,229 @@
+package io.gitlab.arturbosch.detekt.rules.coroutines
+
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.rules.KotlinCoreEnvironmentTest
+import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
+import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+@KotlinCoreEnvironmentTest
+class SuspendFunWithCoroutineScopeReceiverSpec(val env: KotlinCoreEnvironment) {
+
+    val subject = SuspendFunWithCoroutineScopeReceiver(Config.empty)
+
+    @Nested
+    inner class `SuspendFunWithCoroutineScopeReceiver rule` {
+
+        @Test
+        fun `reports when top-level suspend function has explicit CoroutineScope receiver type`() {
+            val code = """
+                import kotlinx.coroutines.CoroutineScope
+                import kotlinx.coroutines.launch
+                import kotlinx.coroutines.delay
+
+                suspend fun CoroutineScope.foo() {
+                    launch {
+                        delay(timeMillis = 1000)
+                    }
+                }
+            """
+            assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
+        }
+
+        @Test
+        fun `reports when top-level suspend function has explicit CoroutineScope receiver type and star import used`() {
+            val code = """
+                import kotlinx.coroutines.*
+
+                suspend fun CoroutineScope.foo() {
+                    launch {
+                        delay(timeMillis = 1000)
+                    }
+                }
+            """
+            assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
+        }
+
+        @Test
+        fun `reports when top-level suspend function has explicit FQN CoroutineScope receiver type`() {
+            val code = """
+                import kotlinx.coroutines.launch
+                import kotlinx.coroutines.delay
+
+                suspend fun kotlinx.coroutines.CoroutineScope.foo() {
+                    launch {
+                        delay(timeMillis = 1000)
+                    }
+                }
+            """
+            assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
+        }
+
+        @Test
+        fun `reports when suspend function has a receiver which inherits from CoroutineScope`() {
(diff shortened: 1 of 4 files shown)


detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunWithCoroutineScopeReceiverSpec.kt
```diff
@@ -0,0 +1,94 @@
...
+                    launch {
+                        delay(1.seconds)
+                    }
+                }
+                """
+            val findings = subject.compileAndLintWithContext(env, code)
+            assertThat(findings).hasSize(1)
+            assertThat(findings.single().issue.id).isEqualTo(SuspendFunWithCoroutineScopeReceiver.issueID)
+        }
+    }
```
BraisGabin, line 93: Could you add three more tests to prove that this doesn't flag any function?

- A no-suspend function that have `CoroutineScope` as receiver.
- A suspend function without receiver.
- A suspend function with a `Int` (or any other type) as receiver.

To ensure that we don't have false positives.

BraisGabin (commented):
Really nice PR! :)

To make CI happy you need to run `./gradlew generateDocumentation` and commit the changes that it will add in the `default-detekt-config` file.

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunWithCoroutineScopeReceiverSpec.kt
hfhbd: Sure, added

hfhbd:
Ah, there is the default configuration. I've already searched for it :D

BraisGabin (approved):
LGTM. Thanks for the contribution!

hfhbd:
I forgot the embedded Kotlin compiler provided by Gradle uses a different language API than the actual compiler executing the tests. So you can't use `Duration`, a inline class with `compile-test-snippets=true`.

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunWithCoroutineScopeReceiverSpec.kt
```diff
@@ -0,0 +1,136 @@
...
+
+                suspend fun CoroutineScope.foo() {
+                    launch {
+                        delay(timeMillis = 1000)
+                    }
+                }
+                """
+            val findings = subject.compileAndLintWithContext(env, code)
+            assertThat(findings).hasSize(1)
+            assertThat(findings.single().issue.id).isEqualTo(SuspendFunWithCoroutineScopeReceiver.issueID)
```
cortinico, line 34: small nit: those lines are somehow useless as you are declaring a

```
val subject = SuspendFunWithCoroutineScopeReceiver(Config.empty)
```

just above. Therefore the only issue that can be reported are from this rule

hfhbd: Oh, that's right :D

BraisGabin:
I just found this on the project I work: `fun foo(block: suspend TestCoroutineScope.() -> Unit) {}`. This is basically the same issue. Should we handle it in this rule too?

hfhbd:
@BraisGabin Sure, added

BraisGabin:
Oh, it seems that #4630 is making this PR to fail now. I suppose that they are just the indentation of `"""`.

hfhbd:
Yeah, fixed it.

hfhbd:
@BraisGabin What about changing `compile-test-snippets` to true by default, to run this check with `gradlew build` too?

cortinico (approved):

hfhbd:
@BraisGabin BTW are you sure about flagging lambdas too? In IntelliJ, only the function with the receiver is marked, but I don't know if this is a bug of IntelliJ...
```kt
import kotlinx.coroutines.*
import kotlin.time.Duration.Companion.seconds

suspend fun main() {
    coroutineScope {
        receiver()
    }
    block {
        launch {

        }
    }
}

suspend fun CoroutineScope.receiver(block: suspend CoroutineScope.() -> Unit) { // Ambiguous coroutineContext due to CoroutineScope receiver of suspend function
    delay(1.seconds)
    launch { // Ambiguous coroutineContext due to CoroutineScope receiver of suspend function
        delay(1.seconds)
        block()
    }
}

suspend fun block(block: suspend CoroutineScope.() -> Unit) {
    delay(1.seconds)
    coroutineScope {
        delay(1.seconds)
        block()
    }
} 
```

BraisGabin:
```kotlin
block {
  launch { } // Which scope am I using the one from suspend or the receiver? 
}
```
I think that the issue is the same. But can you open an issue to intellij to confirm that is a missing part on their end?

hfhbd:
Sure. 
`CoroutineScope`: the inner one wins
`coroutineContext`: this is the problem, because the outer one wins.

hfhbd:
Honestly, after reading https://youtrack.jetbrains.com/issue/KTIJ-12776 and https://youtrack.jetbrains.com/issue/KTIJ-13692/Ambiguous-calls-to-coroutineContext-should-be-reported, I still don't know, when we should mark every coroutine builder (suspend lambda with coroutinescope).
https://youtrack.jetbrains.com/issue/KT-52042/Clarifying-suspend-lambda-and-function-with-coroutineScope-as-pa

BraisGabin:
We can wait to see what they say in the new issue. But for me the issue is the same. In the body of those lambdas is confusing what `coroutineContext` mean. And that's the main reason to use forbid it in `fun`s.

hfhbd:
@BraisGabin Got an interesting answer. https://youtrack.jetbrains.com/issue/KT-52042/Clarifying-suspend-lambda-and-function-with-coroutineScope-as-pa#focus=Comments-27-6029021.0-0
So I would revert marking the lambda parameter too and focus on `suspend fun CoroutineScope.foo` only.

PR #4645, opened 2022-03-20
Bump ktlint to version 0.45.1 and implement new rules

kvn-stgl:
See https://github.com/pinterest/ktlint/releases/tag/0.45.0 for full release notes. I've also wrapped the new (experimental) rules in this version. 

The PR is a follow-up PR of #4641

I will open this PR as a draft, because currently the build doesn't work (see https://github.com/pinterest/ktlint/issues/1421). Feel free to use this PR for further work on it.


detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
```diff
@@ -89,8 +83,14 @@ abstract class FormattingRule(config: Config) : Rule(config) {
 
         // KtLint 0.44.0 is assuming that KtLint.EDITOR_CONFIG_USER_DATA_KEY is available on all the nodes.
         // If not, it crashes with a NPE. Here we're patching their behavior.
+        // This block is deprecated and will be removed in KtLint 0.46. But we have to suppress the
+        // deprecation warning because the ci runs with -Werror.
+        @Suppress("DEPRECATION")
```
kvn-stgl, line 88: To be honest, I'm not really happy with this solution. Also I wanted to declare the ´overrideEditorConfig()` method as deprecated. But sadly each of deprecation will lead to a warning which is handled as error in the ci pipeline so I was forced to suppress this warning (without this code snippet the build will not succeed).

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
```diff
@@ -89,8 +83,14 @@ abstract class FormattingRule(config: Config) : Rule(config) {
...
         if (node.getUserData(KtLint.EDITOR_CONFIG_USER_DATA_KEY) == null) {
-            node.putUserData(KtLint.EDITOR_CONFIG_USER_DATA_KEY, fromMap(emptyMap()))
+            node.putUserData(
+                KtLint.EDITOR_CONFIG_USER_DATA_KEY,
+                com.pinterest.ktlint.core.EditorConfig.Companion.fromMap(emptyMap())
```
cortinico, line 92: Please keep the import as it was before

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/Wrapping.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.pinterest.ktlint.ruleset.standard.WrappingRule
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
+import io.gitlab.arturbosch.detekt.api.internal.AutoCorrectable
+import io.gitlab.arturbosch.detekt.formatting.FormattingRule
+
+/**
+ * See <a href="https://ktlint.github.io/#rule-indentation">ktlint-website</a> for documentation.
+ */
+@ActiveByDefault(since = "1.20.0")
```
cortinico, line 12: Why is this active while the others are not?

cortinico (approved):
Amazing work! 🎉  Left a couple of comments but we're good to go with this

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/Wrapping.kt
kvn-stgl: Because of the changelog:
> All wrapping logic is moved from the indent rule to the new rule wrapping (as part of the standard ruleset). In case you currently have disabled the indent rule, you may want to reconsider whether this is still necessary or that you also want to disable the new wrapping rule to keep the status quo. Both rules can be run independent of each other. 

So the WrappingRule is in the standard ruleset and active by default (as far as I understand). Each other rule is in the experimental ruleset and should not be active by default.

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
kvn-stgl: But then I have to suppress the warning over the entire file. 
Without, the build will throw a warning like `w: .../FormattingRule.kt: (3, 34): 'EditorConfig' is deprecated. Marked for removal ...` and the code will probably not be able to build on the ci. 
It's not aesthetic but I think `@file:Suppress("DEPRECATION")` is the more dirty way here.

cortinico: Yup makes sense. Thanks for clarifying 👍

cortinico (approved):
This is good for me to go 👍

Issue #4652, opened 2022-03-25
Add ignoreOverridden support for BooleanPropertyNaming rule


mhernand40:
## Expected Behavior of the rule
When overriding a Boolean property from a 3rd party library, the [BooleanPropertyNaming](https://detekt.dev/naming.html#booleanpropertynaming) will flag the `override` if it does not adhere to the rule. Since this is out of the developer's control, there should be a way to set the `ignoreOverridden` flag to `true` for such cases.

## Context
One example of this is when overriding the [PagingSource#keyReuseSupported](https://developer.android.com/reference/kotlin/androidx/paging/PagingSource#keyReuseSupported()) property from the Android Paging 3 Library.

Yes, one could configure the rule to accept the naming conventions of the 3rd party library but that may be too invasive if the code base under the dev's control wishes to adhere as close to the default config as possible.

cortinico:
Yup that's a valid addition 👍 Do you want to send a PR?

mhernand40:
Sure. Please feel free to assign to me.

BraisGabin:
Closing as this was already addressed.

PR #4654, opened 2022-03-30
Add ignoreOverridden support for BooleanPropertyNaming rule

mhernand40:
This PR addresses https://github.com/detekt/detekt/issues/4652.

### DIFF
--- a/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNaming.kt
+++ b/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNaming.kt
@@ -37,7 +37,7 @@ class BooleanPropertyNaming(config: Config = Config.empty) : Rule(config) {
     private val allowedPattern: Regex by config("^(is|has|are)", String::toRegex)
 
     @Configuration("ignores properties that have the override modifier")
-    private val ignoreOverridden: Boolean by config(false)
+    private val ignoreOverridden: Boolean by config(true)
 
     override val issue = Issue(
         javaClass.simpleName, Severity.CodeSmell,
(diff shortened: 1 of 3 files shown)


cortinico (approved):

cortinico:
Great stuff 👍

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNaming.kt
```diff
@@ -35,6 +36,9 @@ class BooleanPropertyNaming(config: Config = Config.empty) : Rule(config) {
     @Configuration("naming pattern")
     private val allowedPattern: Regex by config("^(is|has|are)", String::toRegex)
 
+    @Configuration("ignores properties that have the override modifier")
+    private val ignoreOverridden: Boolean by config(false)
```
BraisGabin, line 40: I would default this to `true`. I know, this is to avoid behaviour change but for me this is more a bug fix than a "behaviour change".

BraisGabin:
Thanks for the PR it looks great :)

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNaming.kt
mhernand40: I'd be happy to address and make that change. Would others care to weigh in what their thoughts are before I do so? Thank you 🙂

cortinico: Agree, this is reasonable to be defaulted to true. Is a "bug-fix on steroids"

mhernand40: Done. Thank you so much. 🙂

PR #4668, opened 2022-04-02
Disable Gradle's configuration cache by default

3flex:
There are issues that don't have workarounds at this stage (see #4669). Check again when Gradle 7.5 released.


BraisGabin (approved):

chao2zhang (approved):

cortinico:
Thanks for doing this. On top of this, running `./gradlew publishToMavenLocal` was broken on `main` and fixing it was non trivial.

3flex:
I tested that task and confirmed it was fixed in the other PR, though it required Gradle 7.5 snapshot.

I believe Gradle intends to stabilise configuration cache in 7.5 so I'm hopeful these issues will be fully resolved and we can leave it on when it's out.

PR #4670, opened 2022-04-03
Drop Spek

### LINKED ISSUE #4450: Replace Spek with a new test framework

3flex:
Tying off some loose ends to close #4450.


marschwar:
To my understanding (or wishful thinking) we decided to stay with the `Spec` suffix for the initial migration to be able to process the change set. Personally I think we should switch to the `Test` suffix. This should probably be a separate PR but the documentation changed here should then reflect that.

BraisGabin (approved):

docs/pages/extensions.md
```diff
@@ -161,12 +161,13 @@ To let detekt know about the new processor, we specify a `resources/META-INF/ser
...
-To test the code we use the `detekt-test` module and write a `Spek` testcase.
+To test the code we use the `detekt-test` module and write a JUnit 5 testcase.
 
 ```kotlin
-class NumberOfLoopsProcessorSpec : Spek({
+class NumberOfLoopsProcessorSpec {
 
-	it("should expect two loops") {
+	@Test
+    fun `should expect two loops`() {
```
Goooler, line 170:
```suggestion
    @Test
    fun `should expect two loops`() {
```

3flex:
> I think we should switch to the Test suffix

I'm not too bothered personally, it seems the JUnit convention is for Tests (plural) to be the suffix, though that would depend on the number of tests in the class... Also this convention makes sense for inner classes too so perhaps it's not just files and top level classes to rename.

I don't think there's anything wrong with Spec though. If the current suffix was Spek that would be different!

detekt-test-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KotlinEnvironmentTestSetup.kt
```diff
@@ -11,6 +11,9 @@ import org.spekframework.spek2.dsl.Root
 import org.spekframework.spek2.lifecycle.CachingMode
 import java.nio.file.Path
 
+@Deprecated(
+    "Use @KotlinCoreEnvironmentTest annotation on test classes instead. See detekt's tests for examples."
```
3flex, line 15: Thoughts on wording? This is only relevant if JUnit 5 is being used. Perhaps this message should be test framework agnostic and/or direct readers to the type resolution testing docs which outlines options for different test frameworks.

BraisGabin (approved):

docs/pages/gettingstarted/type-resolution.md
```diff
@@ -111,6 +111,21 @@ If the `bindingContext` is not `EMPTY`, you are free to use it to resolve types
...
+referencing the parameter specified in the constructor:
+
+```kolin
+@KotlinCoreEnvironmentTest
+class MyRuleSpec(private val env: KotlinCoreEnvironment) {
+    @Test
+    fun `reports cast that cannot succeed`() {
+        val code = """/* The code you want to test */"""
+        assertThat(MyRuleSpec().compileAndLintWithContext(env, code)).hasSize(1)
+    }
```
marschwar, line 125:
```suggestion
    private val subject = MyRuleSpec()
    @Test
    fun `reports cast that cannot succeed`() {
        val code = """/* The code you want to test */"""
        assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
    }
```

detekt-test-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KotlinEnvironmentTestSetup.kt
cortinico: Agree, this shouldn't be a deprecation imho but more of a KDoc

docs/pages/extensions.md
```diff
@@ -161,29 +161,29 @@ To let detekt know about the new processor, we specify a `resources/META-INF/ser
...
-		"""
-
-		val ktFile = compileContentForTest(code)
-		NumberOfLoopsProcessor().onProcess(ktFile)
-
-		assertThat(ktFile.getUserData(NumberOfLoopsProcessor.numberOfLoopsKey)).isEqualTo(2)
-	}
-})
-
+class NumberOfLoopsProcessorSpec {
```
cortinico, line 167:
```suggestion
class NumberOfLoopsProcessorTest {
```

docs/pages/gettingstarted/type-resolution.md
```diff
@@ -111,6 +111,21 @@ If the `bindingContext` is not `EMPTY`, you are free to use it to resolve types
 
 To test a rule that uses type resolution, you can use the [`lintWithContext`](https://github.com/detekt/detekt/blob/d3546ff0d539d57e7a502dacbf66e91587fff098/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt#L40-L44) and [`compileAndLintWithContext`](https://github.com/detekt/detekt/blob/cd659ce8737fb177caf140f46f73a1a86b22be56/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt#L63-L72) extension functions.
 
+If you're using JUnit 5 for testing, you can use the `@KotlinCoreEnvironmentTest` annotation on your test class, and
+accept a parameter of type `KotlinCoreEnvironment` in the class constructor. You can then access the environment by
+referencing the parameter specified in the constructor:
+
+```kolin
+@KotlinCoreEnvironmentTest
+class MyRuleSpec(private val env: KotlinCoreEnvironment) {
```
cortinico, line 120: Same as above, I'd use `Test` here

gradle/libs.versions.toml
```diff
@@ -29,8 +28,7 @@ ktlint-rulesetExperimental = { module = "com.pinterest.ktlint:ktlint-ruleset-exp
 
 dokka-jekyll = { module = "org.jetbrains.dokka:jekyll-plugin", version.ref = "dokka"}
 
-spek-dsl = { module = "org.spekframework.spek2:spek-dsl-jvm", version.ref = "spek" }
-spek-runner = { module = "org.spekframework.spek2:spek-runner-junit5", version.ref = "spek" }
+spek-dsl = { module = "org.spekframework.spek2:spek-dsl-jvm", version = "2.0.18" }
```
cortinico, line 31: Q: What's the reason we need this?

cortinico (approved):

docs/pages/extensions.md
3flex: Did we actually land on `Test` as the agree suffix? But yes, agree this should reflect whatever we decide, will update accordingly.

gradle/libs.versions.toml
3flex: We still have this extension on `org.spekframework.spek2.dsl.Root`:

https://github.com/detekt/detekt/blob/ed8b0890c38f82bc0159ce3acf05a1785db580e8/detekt-test-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KotlinEnvironmentTestSetup.kt#L14-L24

detekt-test-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KotlinEnvironmentTestSetup.kt
3flex: Why isn't the deprecation suitable? Are we intending to support those extending detekt and testing with Spek indefinitely?

My preference is a deprecation with removal in the near future (with a copy of the extension function in the release notes that users can then use in their own test code if they're still using Spek).

docs/pages/extensions.md
cortinico: Yup but this is a piece of documentation, which I believe is something we can standardize to match the framework standards.

detekt-test-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KotlinEnvironmentTestSetup.kt
cortinico: Yup sorry, I haven't realized this was Spek-specific. Therefore I'm in for deprecating this

gradle/libs.versions.toml
cortinico: Gotcha. Let's keep it then 👍

cortinico:
I'd like to merge this and prepare 1.20.0 just after it 👍

docs/pages/gettingstarted/type-resolution.md
```diff
@@ -111,6 +111,21 @@ If the `bindingContext` is not `EMPTY`, you are free to use it to resolve types
...
+    @Test
+    fun `reports cast that cannot succeed`() {
+        val code = """/* The code you want to test */"""
+        assertThat(MyRuleSpec().compileAndLintWithContext(env, code)).hasSize(1)
+    }
+}
+```
+
```
chao2zhang, line 128: Do we want to fully drop support for `Spek` test? Line 129 - 141 still have some code references.

At least we can update them.

chao2zhang (approved):

detekt-test-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KotlinEnvironmentTestSetup.kt
3flex: Take a look at the updated deprecation message - I've directed readers to the type resolution docs and copied the Spek DSL extension code there so it becomes a self-contained example that doesn't rely on `detekt-test-utils`. This will make it easier to remove the deprecated code in a future release.

docs/pages/gettingstarted/type-resolution.md
3flex: Sorry didn't see your note before pushing an update - I've actually expanded the docs to make deprecation of `org.spekframework.spek2.dsl.Root.setupKotlinEnvironment` easier for anyone using it in external projects.

docs/pages/gettingstarted/type-resolution.md
```diff
@@ -111,10 +111,38 @@ If the `bindingContext` is not `EMPTY`, you are free to use it to resolve types
 
 To test a rule that uses type resolution, you can use the [`lintWithContext`](https://github.com/detekt/detekt/blob/d3546ff0d539d57e7a502dacbf66e91587fff098/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt#L40-L44) and [`compileAndLintWithContext`](https://github.com/detekt/detekt/blob/cd659ce8737fb177caf140f46f73a1a86b22be56/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt#L63-L72) extension functions.
 
-If you're using Spek for testing, you can use the `setupKotlinEnvironment()` util function, and get access to the `KotlinCoreEnvironment` by simply calling `val env: KotlinCoreEnvironment by memoized()`:  
+If you're using JUnit 5 for testing, you can use the `@KotlinCoreEnvironmentTest` annotation on your test class, and
+accept a parameter of type `KotlinCoreEnvironment` in the class constructor. You can then access the environment by
+referencing the parameter specified in the constructor:
+
+```kolin
```
Goooler, line 118:
```suggestion
```kotlin
```

3flex:
LGTM? I'll let someone else cock the button after a final review.

PR #4631, opened 2022-03-13
Fix concurrency issue when creating PomModel (#4609)

davidburstrom:
Fixes https://github.com/detekt/detekt/issues/4609

Issue #4609, opened 2022-03-04
Parallel invocation of AnalysisFacade fails spuriously in 1.20.0-RC1


davidburstrom:
## Expected Behavior

It should be possible to run tests and other scenarios in parallel where `AnalysisFacade` is invoked.

## Observed Behavior

```
io.github.detekt.tooling.api.UnexpectedError
	at app//io.gitlab.arturbosch.detekt.core.tooling.AnalysisFacade$runAnalysis$1.invoke(AnalysisFacade.kt:55)
	at app//io.gitlab.arturbosch.detekt.core.tooling.AnalysisFacade$runAnalysis$1.invoke(AnalysisFacade.kt:47)
	at app//io.gitlab.arturbosch.detekt.core.tooling.ProcessingSpecSettingsBridgeKt.withSettings(ProcessingSpecSettingsBridge.kt:26)
	at app//io.gitlab.arturbosch.detekt.core.tooling.AnalysisFacade.runAnalysis$detekt_core(AnalysisFacade.kt:47)
	at app//io.gitlab.arturbosch.detekt.core.tooling.AnalysisFacade.run(AnalysisFacade.kt:25)
        ...
Caused by: java.lang.RuntimeException: Duplicate registration for EP 'org.jetbrains.kotlin.com.intellij.treeCopyHandler': first in FakeIdForTests, second in FakeIdForTests
	at org.jetbrains.kotlin.com.intellij.openapi.components.ComponentManager.createError(ComponentManager.java:167)
	at org.jetbrains.kotlin.com.intellij.openapi.extensions.impl.ExtensionsAreaImpl.checkThatPointNotDuplicated(ExtensionsAreaImpl.java:235)
	at org.jetbrains.kotlin.com.intellij.openapi.extensions.impl.ExtensionsAreaImpl.registerExtensionPoint(ExtensionsAreaImpl.java:240)
	at org.jetbrains.kotlin.com.intellij.openapi.extensions.impl.ExtensionsAreaImpl.doRegisterExtensionPoint(ExtensionsAreaImpl.java:210)
	at org.jetbrains.kotlin.com.intellij.openapi.extensions.impl.ExtensionsAreaImpl.doRegisterExtensionPoint(ExtensionsAreaImpl.java:189)
	at org.jetbrains.kotlin.com.intellij.openapi.extensions.impl.ExtensionsAreaImpl.doRegisterExtensionPoint(ExtensionsAreaImpl.java:183)
[...]

PR #4672, opened 2022-04-04
Use ConTester to prove that the synchronized block is required

davidburstrom:
Relates to PR https://github.com/detekt/detekt/pull/4631

The issue that was fixed was never proven, due to lack of test tooling. Therefore, I wrote the ConTester utility (https://github.com/davidburstrom/contester) which allows for defining breakpoints in the production code.

If the `synchronized` block is removed, the tests start to fail for the reason indicated in https://github.com/detekt/detekt/issues/4609

### DIFF
--- a/detekt-parser/src/main/kotlin/io/github/detekt/parser/DetektPomModel.kt
+++ b/detekt-parser/src/main/kotlin/io/github/detekt/parser/DetektPomModel.kt
@@ -1,5 +1,6 @@
 package io.github.detekt.parser
 
+import io.github.davidburstrom.contester.ConTesterBreakpoint
 import org.jetbrains.kotlin.com.intellij.openapi.extensions.ExtensionPoint
 import org.jetbrains.kotlin.com.intellij.openapi.extensions.Extensions.getRootArea
 import org.jetbrains.kotlin.com.intellij.openapi.project.Project
@@ -26,6 +27,9 @@ class DetektPomModel(project: Project) : UserDataHolderBase(), PomModel {
             // Addresses https://github.com/detekt/detekt/issues/4609
             synchronized(extensionArea) {
                 if (!extensionArea.hasExtensionPoint(extension)) {
+                    ConTesterBreakpoint.defineBreakpoint("DetektPomModel.registerExtensionPoint") {
+                        extensionArea == getRootArea()
+                    }
                     extensionArea.registerExtensionPoint(
                         extension,
                         extensionClass,
--- a/detekt-parser/src/test/kotlin/io/github/detekt/parser/KtCompilerSpec.kt
+++ b/detekt-parser/src/test/kotlin/io/github/detekt/parser/KtCompilerSpec.kt
@@ -1,5 +1,6 @@
 package io.github.detekt.parser
 
+import io.github.davidburstrom.contester.ConTesterDriver
 import io.github.detekt.psi.BASE_PATH
 import io.github.detekt.psi.LINE_SEPARATOR
 import io.github.detekt.psi.RELATIVE_PATH
@@ -8,10 +9,24 @@ import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.assertThatIllegalArgumentException
 import org.jetbrains.kotlin.com.intellij.psi.PsiErrorElement
 import org.jetbrains.kotlin.psi.KtTreeVisitorVoid
+import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 
 class KtCompilerSpec {
+    @AfterEach
+    internal fun tearDown() {
+        ConTesterDriver.cleanUp()
+    }
+
+    @Test
+    fun `parallel construction of KtCompilers should be thread safe`() {
+        val thread1 = ConTesterDriver.thread { KtCompiler() }
+        val thread2 = ConTesterDriver.thread { KtCompiler() }
+        ConTesterDriver.runToBreakpoint(thread1, "DetektPomModel.registerExtensionPoint")
+        ConTesterDriver.runUntilBlockedOrTerminated(thread2)
+        ConTesterDriver.join(thread1)
+    }
 
     @Nested
     inner class `Kotlin Compiler` {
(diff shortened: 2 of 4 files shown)


detekt-parser/src/main/kotlin/io/github/detekt/parser/DetektPomModel.kt
```diff
@@ -26,6 +27,9 @@ class DetektPomModel(project: Project) : UserDataHolderBase(), PomModel {
             // Addresses https://github.com/detekt/detekt/issues/4609
             synchronized(extensionArea) {
                 if (!extensionArea.hasExtensionPoint(extension)) {
+                    ConTesterBreakpoint.defineBreakpoint("DetektPomModel.registerExtensionPoint") {
+                        extensionArea == getRootArea()
```
cortinico, line 31: could you put a named parameter here?

detekt-parser/src/test/kotlin/io/github/detekt/parser/KtCompilerSpec.kt
```diff
@@ -8,10 +9,24 @@ import org.assertj.core.api.Assertions.assertThat
...
+        ConTesterDriver.cleanUp()
+    }
+
+    @Test
+    fun `parallel construction of KtCompilers should be thread safe`() {
+        val thread1 = ConTesterDriver.thread { KtCompiler() }
+        val thread2 = ConTesterDriver.thread { KtCompiler() }
+        ConTesterDriver.runToBreakpoint(thread1, "DetektPomModel.registerExtensionPoint")
+        ConTesterDriver.runUntilBlockedOrTerminated(thread2)
+        ConTesterDriver.join(thread1)
```
cortinico, line 28: nit: That's mostly an API design decision, but they would read better if they're extension function instead.

cortinico (commented):
That's awesome 👍 It would help us catch a lot of concurrency problems we have. As a side note: we probably want to wait for a stable release of Contester before merging and releasing this.

detekt-parser/src/test/kotlin/io/github/detekt/parser/KtCompilerSpec.kt
davidburstrom: I get what you're saying, though I'm not too keen on extension functions, especially since they would be called `join()`, `start()` etc which already overload and would be confusing. How about some sort of `ConTester` context, so the nature of the test is obvious?

davidburstrom:
> That's awesome 👍 It would help us catch a lot of concurrency problems we have. As a side note: we probably want to wait for a stable release of Contester before merging and releasing this.

Thanks :) It'd especially useful to regression test them.

As for the side note: Yeah, I've possibly been too defensive about the naming, wanting to get a little bit of feedback on the tool before going `0.1.0` proper. I'd say its quite release ready as it is.

detekt-parser/src/test/kotlin/io/github/detekt/parser/KtCompilerSpec.kt
cortinico: > especially since they would be called join(), start() etc which already overload and would be confusing. 

Agree. If you end up this way, I would not use those names and be a bit more specific with function naming.

> How about some sort of ConTester context, so the nature of the test is obvious?

That also works eventually :)

detekt-parser/src/main/kotlin/io/github/detekt/parser/DetektPomModel.kt
cortinico:
```
ConTesterBreakpoint.defineBreakpoint(
    name = "DetektPomModel.registerExtensionPoint"
    onlyWhen = { extensionArea == getRootArea() }
)
```

Something along those lines. `onlyWhen` could be `when` or `condition` or whatever you prefer

davidburstrom: Ah. That's not possible unless there are Kotlin bindings for the library. `condition` is preferable, to mimic debugger terminology.

BraisGabin:
It's odd to have test code in production code. But we already had regressions on this code so I imagine that it is worth it.

I assume that the overhead in production is near to 0, right?

davidburstrom:
The overhead is 0 for repeated calls, see https://github.com/davidburstrom/contester#benchmarks
Though, I should benchmark the first call too.

detekt-parser/src/main/kotlin/io/github/detekt/parser/DetektPomModel.kt
cortinico: Sorry. I naively thought you wrote this in Kotlin, my bad :) You could add bindings for that btw.

> condition is preferable, to mimic debugger terminology.

Yup that sounds great to me 👍

davidburstrom:
> It's odd to have test code in production code. But we already had regressions on this code so I imagine that it is worth it.

I get what you mean. Though, it's not much different than usage of `@VisibleForTesting` annotations, and other production constructs that only exist to bolster testability. At least this one doesn't affect the API of the system under test.

> I assume that the overhead in production is near to 0, right?

I've updated the benchmark, the first call is on micro-second duration.

BraisGabin (approved):

davidburstrom:
I've released a 0.1.0 GA, just waiting for it to appear on Maven Central.

detekt-parser/src/main/kotlin/io/github/detekt/parser/DetektPomModel.kt
davidburstrom: Yeah, I would not want to impose a Kotlin runtime dependency for such a specific tool :)

cortinico (approved):

BraisGabin:
I vote to move this one to `1.21`. We can merge it as soon as we release `1.20` but because it adds a really new library in production code and we released two RC already without it I think its safer and we lose nearly nothing by waiting.

davidburstrom:
I'm fine with that :)

davidburstrom:
@BraisGabin @cortinico I've released version 0.2.0 of ConTester which fixes a minor bug. Since Detekt 1.20.0 is released, could we consider merging this? I've rebased on top of latest `main`.

BraisGabin (approved):

davidburstrom:
Thanks for that! :)

cortinico:
> @BraisGabin @cortinico I've released version 0.2.0 of ConTester which fixes a minor bug. Since Detekt 1.20.0 is released, could we consider merging this? I've rebased on top of latest `main`.

Thanks for doing this 🙏

PR #4687, opened 2022-04-06
Drop (most) Groovy DSL tests

### LINKED ISSUE #4560: Drop Groovy DSL tests in detekt-gradle-plugin

3flex:
Closes #4560

There's a new Groovy DSL test in `DetektTaskGroovyDslSpec` which covers the entire DSL for the extension and the three detekt tasks. It doesn't run the task since many files etc. won't be accessible. Instead, the `:help` task is run, which compiles the build files. Using `tasks.create` for these tasks means the tasks will be created & configured even though they're not run.

### DIFF
--- a/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
+++ b/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
@@ -71,7 +71,7 @@ class DetektJvmSpec {
                 buildFileName = "build.gradle.kts",
                 mainBuildFileContent = """
                     plugins {
-                        id("org.jetbrains.kotlin.jvm")
+                        kotlin("jvm")
                         id("io.gitlab.arturbosch.detekt")
                     }
 
@@ -112,7 +112,7 @@ class DetektJvmSpec {
                 buildFileName = "build.gradle.kts",
                 mainBuildFileContent = """
                     plugins {
-                        id("org.jetbrains.kotlin.jvm")
+                        kotlin("jvm")
                         id("io.gitlab.arturbosch.detekt")
                     }
 
--- a/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
+++ b/detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
@@ -19,7 +19,7 @@ class DetektJvmSpec {
 
             val gradleRunner = DslGradleRunner(
                 projectLayout = ProjectLayout(numberOfSourceFilesInRootPerSourceDir = 1),
-                buildFileName = "build.gradle",
+                buildFileName = "build.gradle.kts",
                 baselineFiles = listOf("detekt-baseline.xml", "detekt-baseline-main.xml", "detekt-baseline-test.xml"),
                 projectScript = {
                     apply<KotlinPluginWrapper>()
@@ -68,11 +68,11 @@ class DetektJvmSpec {
         inner class `report location set on extension & task` {
             val gradleRunner = DslGradleRunner(
                 projectLayout = ProjectLayout(numberOfSourceFilesInRootPerSourceDir = 1),
-                buildFileName = "build.gradle",
+                buildFileName = "build.gradle.kts",
                 mainBuildFileContent = """
                     plugins {
-                        id "org.jetbrains.kotlin.jvm"
-                        id "io.gitlab.arturbosch.detekt"
+                        id("org.jetbrains.kotlin.jvm")
+                        id("io.gitlab.arturbosch.detekt")
                     }
 
                     repositories {
@@ -86,7 +86,7 @@ class DetektJvmSpec {
                         }
                     }
 
-                    tasks.withType(io.gitlab.arturbosch.detekt.Detekt).configureEach {
+                    tasks.withType<io.gitlab.arturbosch.detekt.Detekt>().configureEach {
                         reports {
                             txt.destination = file("output-path2.txt")
                         }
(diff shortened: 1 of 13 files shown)


detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/report/ReportMergeSpec.kt
```diff
@@ -20,52 +20,52 @@ class ReportMergeSpec {
...
                         ${builder.gradleSubprojectsApplyPlugins}
-                        |apply plugin: 'java-library'
+                        |plugins.apply("java-library")
                     """.trimMargin()
                 )
             }
             val mainBuildFileContent: String = """
                 |plugins {
-                |    id "io.gitlab.arturbosch.detekt" apply false
+                |    id("io.gitlab.arturbosch.detekt")
```
3flex, line 44: This is the only behaviour change I had to introduce (here, and again in a later test in this class).

Because of the way Gradle generates type-safe accessors for extensions provided by plugins it's necessary to apply the plugin to the root project so that the detekt extension can be used later in the `subprojects` block. This doesn't impact the behaviour of the test itself though.

BraisGabin (approved):

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektMultiplatformSpec.kt
```diff
@@ -148,12 +148,12 @@ class DetektMultiplatformSpec {
                     1,
                     buildFileContent = """
                         plugins {
-                            id "kotlin-multiplatform"
-                            id "com.android.library"
-                            id "io.gitlab.arturbosch.detekt"
+                            id("kotlin-multiplatform")
```
cortinico, line 151: nit: Those could become a `kotlin("multiplatform")` and so on with Kotlin DSL

detekt-gradle-plugin/src/test/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
```diff
@@ -68,11 +68,11 @@ class DetektJvmSpec {
         inner class `report location set on extension & task` {
             val gradleRunner = DslGradleRunner(
                 projectLayout = ProjectLayout(numberOfSourceFilesInRootPerSourceDir = 1),
-                buildFileName = "build.gradle",
+                buildFileName = "build.gradle.kts",
                 mainBuildFileContent = """
                     plugins {
-                        id "org.jetbrains.kotlin.jvm"
-                        id "io.gitlab.arturbosch.detekt"
+                        id("org.jetbrains.kotlin.jvm")
```
cortinico, line 74: Same here with `kotlin("jvm")`

cortinico (approved):
LGTM

PR #4697, opened 2022-04-08
Use jacoco-report-aggregation plugin to generate consolidated coverage report

3flex:
This core Gradle plugin was introduced in Gradle 7.4.

The only change in coverage reporting is the addition of `io.github.detekt.utils` to the coverage report as it's a transitive dependency for some projects. The aggregation plugin includes transitive dependencies when generating reports, while the old configuration excluded it (to avoid issues with external dependencies, which the aggregation plugin deals with differently).

The new behaviour is more correct.

image


cortinico:
Not sure how to review this as I don't have experience with this plugin. Changes looks fine on my end 👍

schalkms (approved):

PR #4703, opened 2022-04-10
Allow renovate to report Ruby dependency updates

3flex:
detekt/detekt#4702 dropped dependabot config, but it wasn't actually redundant because bundler dependency checks were disabled in the renovate config.

This enables checks for dependencies managed by bundler.

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -1,9 +1,5 @@
 {
   "labels": ["dependencies"],
-  "packageRules": [{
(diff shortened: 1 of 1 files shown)


schalkms (approved):

cortinico:
> #4702 dropped dependabot config, but it wasn't actually redundant because bundler dependency checks were disabled in the renovate config.

Thanks for doing that. I missed that dependabot was still being used for the website 👍

PR #4704, opened 2022-04-10
Setup GPG signing on CI

cortinico:
This PR configures the signing of GPG to be done on CI rather than on local machine (is a first step towards improving https://github.com/detekt/detekt/discussions/4696).

I've configured the secrets on the org, so from now on, -SNAPSHOTS version should be signed on CI. The release process will follow the same approach.


schalkms (approved):

chao2zhang (commented):
Do we need to update `.github/workflows/deploy-snapshot.yaml` in order to pass these environment variables?

BraisGabin (approved):

PR #4706, opened 2022-04-11
Make `runWithHelpFlag` adopting incremental build

chao2zhang:
This is part of the practice of Gradle enterprise onboarding

After running the following command
```
./gradlew clean build --no-build-cache -Dscan.tag.exp1
./gradlew build --no-build-cache -Dscan.tag.exp1
```
We should let as many tasks as incremental build (up-to-date) check as possible. In this PR:
- `runWithHelpFlag` will be compatible
- `runWithArgsFile` won't be compatible because it is reading the entire directory as the input source
- `:detekt-gradle-plugin:functionalTest` will be inspected further.

### DIFF
--- a/detekt-cli/build.gradle.kts
+++ b/detekt-cli/build.gradle.kts
@@ -42,12 +42,18 @@ tasks {
     }
 
     val runWithHelpFlag by registering(JavaExec::class) {
+        inputs.files(shadowJar)
+        outputs.upToDateWhen { true }
         classpath = files(shadowJar)
         args = listOf("--help")
     }
 
     val runWithArgsFile by registering(JavaExec::class) {
-        inputs.files(formattingJar) // ensures detekt-formatting JAR is built before this task is executed
+        // The task generating these jar files run first.
+        inputs.files(shadowJar, formattingJar)
+        // This task does not adopt incremental-build (up-to-date) check because it is reading
+        // the entire directory as the input source.
+        outputs.upToDateWhen { false }
         classpath = files(shadowJar)
         workingDir = rootDir
         args = listOf("@./config/detekt/argsfile", "-p", formattingJar.singleFile.path)
--- a/detekt-cli/build.gradle.kts
+++ b/detekt-cli/build.gradle.kts
@@ -42,7 +42,6 @@ tasks {
     }
 
     val runWithHelpFlag by registering(JavaExec::class) {
-        inputs.files(shadowJar)
         outputs.upToDateWhen { true }
         classpath = files(shadowJar)
         args = listOf("--help")
@@ -50,7 +49,7 @@ tasks {
 
     val runWithArgsFile by registering(JavaExec::class) {
         // The task generating these jar files run first.
-        inputs.files(shadowJar, formattingJar)
+        inputs.files(formattingJar)
         // This task does not adopt incremental-build (up-to-date) check because it is reading
         // the entire directory as the input source.
         outputs.upToDateWhen { false }
--- a/detekt-cli/build.gradle.kts
+++ b/detekt-cli/build.gradle.kts
@@ -50,9 +50,7 @@ tasks {
     val runWithArgsFile by registering(JavaExec::class) {
         // The task generating these jar files run first.
         inputs.files(formattingJar)
-        // This task does not adopt incremental-build (up-to-date) check because it is reading
-        // the entire directory as the input source.
-        outputs.upToDateWhen { false }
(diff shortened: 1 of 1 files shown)


detekt-cli/build.gradle.kts
```diff
@@ -42,12 +42,18 @@ tasks {
...
         classpath = files(shadowJar)
         args = listOf("--help")
     }
 
     val runWithArgsFile by registering(JavaExec::class) {
-        inputs.files(formattingJar) // ensures detekt-formatting JAR is built before this task is executed
+        // The task generating these jar files run first.
+        inputs.files(shadowJar, formattingJar)
+        // This task does not adopt incremental-build (up-to-date) check because it is reading
+        // the entire directory as the input source.
```
3flex, line 55: The correct file inputs could still be added using something like:

```kotlin
val filesToAnalyze = fileTree(rootDir).include("**/*.kt", "**/*.kts")
inputs.files(formattingJar, filesToAnalyze, "$rootDir/config/detekt/detekt.yml", "$rootDir/config/detekt/baseline.xml")
```

This untested but wouldn't be too far from working as expected. It's messy, but that's necessary because Gradle doesn't know anything about the task inputs specified in the args file so they have to be specified manually here.

Outputs can similarly be added:
```kotlin
outputs.files("build/detekt-report.html") // and include the other report types
```

detekt-cli/build.gradle.kts
```diff
@@ -42,12 +42,18 @@ tasks {
...
+        inputs.files(shadowJar)
+        outputs.upToDateWhen { true }
         classpath = files(shadowJar)
         args = listOf("--help")
     }
 
     val runWithArgsFile by registering(JavaExec::class) {
-        inputs.files(formattingJar) // ensures detekt-formatting JAR is built before this task is executed
+        // The task generating these jar files run first.
+        inputs.files(shadowJar, formattingJar)
```
3flex, line 53: As in the comment on same change in `runWithHelpFlag`, this is unnecessary. There's already a task dependency on `shadowJar` from `classpath = files(shadowJar)`

detekt-cli/build.gradle.kts
```diff
@@ -42,12 +42,18 @@ tasks {
...
+        // This task does not adopt incremental-build (up-to-date) check because it is reading
+        // the entire directory as the input source.
+        outputs.upToDateWhen { false }
```
3flex, line 56: If outputs are added correctly then this line isn't necessary, and Gradle would check if the report outputs are up to date or not based on the inputs.

detekt-cli/build.gradle.kts
```diff
@@ -42,12 +42,18 @@ tasks {
     }
 
     val runWithHelpFlag by registering(JavaExec::class) {
+        inputs.files(shadowJar)
```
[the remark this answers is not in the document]
chao2zhang: Good to know about that.

detekt-cli/build.gradle.kts
chao2zhang: I have tried a similar effort but couldn't get the up-to-date check to return true. We also need to flatten out the `argsfile.txt` in order to keep consistency - This feels like too much effort for me.

detekt-cli/build.gradle.kts
chao2zhang: This `outputs.upToDateWhen { false }` is unnecessary - But I would want to explicitly put it here so that we do not need to look into further optimization

detekt-cli/build.gradle.kts
3flex: Makes sense - in that case the only change that should be needed for this task is adding `outputs.upToDateWhen { false }` and the rest can be left as is.

detekt-cli/build.gradle.kts
3flex: I just remembered there was recent functionality added to specifically mark a task as unsuitable for up to date checks or caching:

```suggestion
        doNotTrackState("The entire root directory is read as the input source.")
```

There's some more info in the [Gradle 7.3 release notes](https://docs.gradle.org/7.3/release-notes.html#plugin-development-improvements):

> @UntrackedTask and Task.doNotTrackState are a replacement for Task.outputs.upToDateWhen { false } if you want your task to never be up-to-date. It has the advantage that it is faster since Task.outputs.upToDateWhen { false } still spends time on capturing task state.

3flex (approved):

PR #4709, opened 2022-04-11
Integrate detekt junit

BraisGabin:
After see #4699 I implemented the Rule [`MissingTestAnnotation`](https://github.com/BraisGabin/detekt-junit-rules/blob/main/src/main/kotlin/com/braisgabin/detekt/junit/MissingTestAnnotation.kt) to avoid this kind of issues.

The problem that I see si that this rule set is stored on Jitpack (I want to publish it on `mavenCental` but each time I think about all those steps...). What do you think? Is it ok to have a dependency from Jitpack?

### DIFF
--- a/detekt-gradle-plugin/build.gradle.kts
+++ b/detekt-gradle-plugin/build.gradle.kts
@@ -16,6 +16,7 @@ detekt {
 repositories {
     mavenCentral()
     google()
+    maven { url = uri("https://jitpack.io") }
 }
 
 testing {
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -69,5 +69,6 @@ gradleEnterprise {
 dependencyResolutionManagement {
     repositories {
         mavenCentral()
+        maven { url = uri("https://jitpack.io") }
     }
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -29,6 +29,7 @@ allprojects {
         detekt(project(":detekt-cli"))
         detektPlugins(project(":custom-checks"))
         detektPlugins(project(":detekt-formatting"))
+        detektPlugins("com.github.BraisGabin:detekt-junit-rules:0.0.4")
     }
 
     tasks.withType<Detekt>().configureEach {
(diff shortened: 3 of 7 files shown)


3flex:
I really think Jitpack should be a last resort, and this should be published to Maven Central.

You will probably get more users if published there too.

cortinico:
> So I'm 👍 to merging but I'd really like to see it published in a more appropriate repo first.

Same. If you move it under the github.com/detekt org, we could reuse some of the Github Actions and Keys setup to publish this as well

detekt-gradle-plugin/build.gradle.kts
```diff
@@ -16,6 +16,7 @@ detekt {
 repositories {
     mavenCentral()
     google()
+    maven { url = uri("https://jitpack.io") }
```
Goooler, line 19:
```suggestion
    maven("https://jitpack.io")
```

settings.gradle.kts
```diff
@@ -69,5 +69,6 @@ gradleEnterprise {
 dependencyResolutionManagement {
     repositories {
         mavenCentral()
+        maven { url = uri("https://jitpack.io") }
```
Goooler, line 72:
```suggestion
        maven("https://jitpack.io")
```

3flex:
I _think_ that was just a suggestion to make publishing easier if you didn't want to go through the steps of setting it up yourself.

I agree that treating the rule set as a third party rule set is a good idea, and not to host in the detekt org, for reasons you outlined.

BraisGabin:
I know, I know, there was 0 anger or bad vibes on my comment (sorry if it seemed so). I *speak* Kotlin better than English 😂. It was just my point of push the third-party plugins. But I appreciate the proposal of cortinico, it means that he likes the rules :)

settings.gradle.kts
yogurtearl: a note for the future, should use `exclusiveContent` where ever possible. 
https://docs.gradle.org/current/userguide/declaring_repositories.html#declaring_content_exclusively_found_in_one_repository

BraisGabin:
Closing until I upload it to maven central

Issue #3078, opened 2020-09-16
Scoped check: detekt just those files, that have been changed, not all the source


Sadwyn:
Is that possible to do with some setting? I want o launch it through git hook

BraisGabin:
You can use de cli for that. But you will not able to use the rules that need type solving.

schalkms:
Thanks for the feedback!
This topic, however, has been covered intensively in the recent past. Please take a look at the following issues including the links. If you have additional question, don’t hesitate to ask.
- #2175 
- #2320 
- #1057

The detekt slack channel Also contains information regarding this topic.

schalkms:
Closed as duplicate

Sadwyn:
Thank you for response. Is there some tutorial how can I use cli detekt
with some file pathes in pre commit git hook?

schalkms:
Yes, there is a tutorial available.
https://detekt.github.io/detekt/git-pre-commit-hook.html

Issue #2320, opened 2020-02-06
Is it possible to run detekt for a single file?


VictorAlbertos:
We're trying to speed up our build time on the CI so we're considering only run static analytical tools on those files that actually contain changes. Is it possible (and if it's possible, is it advisable?) to run Detekt for single files? If not, what's the smaller scope in which Detekt can run with (Is it a Gradle module)?

Thanks!

schalkms:
Yes it is via scripts. Please take a look at the following link for an example.
https://arturbosch.github.io/detekt/git-pre-commit-hook.html

Please note that the approach of analyzing a single file is not recommend, since rules using type and symbol solving don’t work.

VictorAlbertos:
Thanks for the information!

But I don't understand how to use Detekt for specific files based on what you've shared. Can you put a specific example of how to call Detekt for a group of files, please? 

> Please note that the approach of analyzing a single file is not recommend, since rules using type and symbol solving don’t work.

And what are those rules? I don't see in the [docs about the rules](https://arturbosch.github.io/detekt/comments.html) anything telling if one specif rule falls into that category 🤔

schalkms:
Using a Git commit hook could work in your case.
I just outlined a simple example.
When more rules use type and symbol solving, detekt will need the information from referenced types as well. For more sophisticated analysis, detekt needs more context. Using a single that just changed is in this case not enough.

> Can you put a specific example of how to call Detekt for a group of files, please?

I'm citing #1792.

> You can list of changed files and pipe them to detekt via the `--input x,y,z` parameter.
 See `git ls-files --modified` for example.

> And what are those rules?

All rules that use the `bindingContext` fall into that category.
- MissingWhenCase
- HasPlatformType
- UnsafeCallOnNullableType
- MapGetWithNotNullAssertionOperator
- UnsafeCast
- Deprecation
- RedundantElseInWhen
- ForbiddenVoid
- ExplicitCollectionElementAccessMethod
- RedundantExplicitType
- PreferToOverPairSyntax
- ForbiddenMethodCall
- UselessCallOnNotNull
- RedundantSuspendModifier
- SpreadOperator

I hope that I didn't miss a rule. Furthermore, we plan to have more existing rules that use type and symbol solving (#1915).

HTH

VictorAlbertos:
Thank you very much for the detailed response!

MaksimDmitriev:
> Yes it is possible via scripts. Please take a look at the following link for an example.
[...]

Issue #2175, opened 2019-12-12
How to limit scope for Detekt?


SerG3z:
Now Detekt analyzing all the code in my project, but I want to parse only the files in which the changes occurred.
The problem is that I write custom rules and so many classes will not correspond to them, so as not to rewrite the entire project at once, you need to apply these rules only to files that are currently modified, you can recognize them through git. 

Perhaps someone knows how to solve this problem?

schalkms:
This has been covered in #1792 and #1057. The referenced YouTube video shows exactly how to write a Git commit hook for detekt.

Please note that we don't recommend this. You cannot use rules that need type and symbol solving to work.
I'll probably add this to the documentation on detekt's homepage as well if I have time.

Issue #1792, opened 2019-07-30
Too slow to run in precommit hook


srcreigh:
Hi, is there a way to set up detekt gradle task to run on only staged git files?

I did try my hand at it, subclassing the detekt task. However I couldn't re-use the settings in the subclassed detekt task (IIRC they are configured outside of the task itself) and there may have been some caching issues.

Similar issue #1057, do you strongly recommend SonarQube? This is for my workplace so it would be much higher scope to add that to CI, ideally we would get a quick check locally...

schalkms:
Hi @srcreigh 
Yeah the answer is pretty much the same as in the linked issue.
In your case, you could also use a git commit hook shown in this [recorded talk](https://youtu.be/LT6m5_LO2DQ?t=690) on YouTube.
HTH

schalkms:
This [code post](https://stackoverflow.com/questions/2412450/git-pre-commit-hook-changed-added-files) shows you how to get all staged files.
While we don't recommend this, it's certainly possible.
Detekt runs very fast. Unless you have a very big code base, this shouldn't matter at all.

srcreigh:
Hi, forgive me, of course it is fast, I am just very impatient :-) I have a detekt pre-commit hook set up, but it usually takes 7-10 seconds to run the detekt gradle task.

Is there a particular reason you don't recommend detekting only staged files? For example could that produce false positives?

schalkms:
> Is there a particular reason you don't recommend detekting only staged files?

When more rules use type and symbol solving, detekt will need the information from referenced types as well. Currently, only a very small number of rules (2-3) use type and symbol solving.

> I have a detekt pre-commit hook set up, but it usually takes 7-10 seconds to run the detekt gradle task.

If this is a dealbreaker for you, you can also use detekt's console application.

Anyways, fetching the staged files with the code snippet in the mentioned post should work. You need to feed the received file path list into detekt (via Gradle Groovy or Kotlin DSL or CLI).
If you manage to hack this together, I can also see this pre-commit hook as a nice addition to detekt. It would also help other users with the same use case. So please feel free to submit a PR.

There are so many ways for integrating detekt into a users CI toolchain. We are actively learning from the community. Thank you for bringing this up and for the nice active discussion in this issue!

arturbosch:
You can list of changed files and pipe them to detekt via the `--input x,y,z` parameter.

See `git ls-files --modified` for example.

PR #4711, opened 2022-04-12
Document pre-commit hook for staged files

alyssa-glean:
Relates to: #1057, #1792, #2175, #2320, #3078

I just got finished implementing a "partial" detekt hook for my team, so I'm raising a PR to document it since it has been asked about many times in the past. Our use case is that we want a hook for "fast feedback" (i.e. not having to wait for a CI build to find out you've violated a check) - as such, we want it to run quickly and aren't too fussed if it occasionally misses things (because CI will still pick those cases up for us, by running a full check).

I've included a disclaimer that checks requiring types won't work properly, however I am yet to verify exactly what will happen in this scenario. Are the affected checks simply skipped when using the `--include` flag? That's what I'm hoping is the case... it would be more problematic if they fall over or might produce false positives etc!

Anyhow, hopefully this can be of use to someone. Thanks for developing and maintaining this tool! :smile:

### DIFF
--- a/docs/pages/gettingstarted/git-pre-commit-hook.md
+++ b/docs/pages/gettingstarted/git-pre-commit-hook.md
@@ -68,17 +68,14 @@ fileArray=($@)
 detektInput=$(IFS=,;printf  "%s" "${fileArray[*]}")
 echo "Input files: $detektInput"
 
-OUTPUT="/tmp/detekt-$(date +%s)"
-detekt --input "$detektInput" > $OUTPUT
+OUTPUT=$(detekt --input "$detektInput" 2>&1)
 EXIT_CODE=$?
 if [ $EXIT_CODE -ne 0 ]; then
-  cat $OUTPUT
-  rm $OUTPUT
+  echo $OUTPUT
   echo "***********************************************"
   echo "                 Detekt failed                 "
   echo " Please fix the above issues before committing "
   echo "***********************************************"
   exit $EXIT_CODE
 fi
-rm $OUTPUT
 ```
--- a/docs/pages/gettingstarted/git-pre-commit-hook.md
+++ b/docs/pages/gettingstarted/git-pre-commit-hook.md
@@ -10,7 +10,7 @@ summary:
 Detekt can be integrated into your development workflow by using a Git pre-commit hook.
 For that reason Git supports to run custom scripts automatically, when a specific action occurs.
 The mentioned pre-commit hook can be setup locally on your dev-machine.
-The following client-side detekt hook is triggered by a commit operation.
+The following client-side detekt hook is triggered by a commit operation, and checks all files via the gradle task.
 
 ```bash
 #!/usr/bin/env bash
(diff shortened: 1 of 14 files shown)


docs/pages/gettingstarted/git-pre-commit-hook.md
```diff
@@ -38,3 +38,45 @@ More information about Git hooks and how to install them can be found in
 A special thanks goes to Mohit Sarveiya for providing this shell script.
 You can watch his excellent talk about **Static Code Analysis For Kotlin** on 
 [YouTube](https://www.youtube.com/watch?v=LT6m5_LO2DQ).
+
+## Only run on staged files
+
+It is also possible to use [the CLI](cli.md) to create a hook that only runs on staged files. This has the advantage of speedier execution, by running on fewer files and avoiding the warm-up time of the gradle daemon.
+
+Please note, however, that a handful of checks will not work correctly with this approach. This is because they require full type resolution - see for example [ElseCaseInsteadOfExhaustiveWhen](https://detekt.dev/potential-bugs.html#elsecaseinsteadofexhaustivewhen). If you do adopt a partial hook, it is recommended that you still implement a full `detekt` check as part of your CI pipeline.
```
cortinico, line 46: I would link to the Using Type Resolution page here rather than a specific rule

docs/pages/gettingstarted/git-pre-commit-hook.md
```diff
@@ -38,3 +38,45 @@ More information about Git hooks and how to install them can be found in
...
+
+Hook definition in pre-commit:
```
cortinico, line 48: Link to pre-commit.com.

Plus please mention that this is pre-commit.com specific. Users might want to use other mechanism to handle their pre commits

cortinico (commented):
Nice addition 👍

BraisGabin:
Thanks! The main problem of mixing cli and gradle is the baseline, I think. If you don't use it, all is fine, but when you run the cli, usually you run it just once for all the project but the baseline is usually per-project/module. And if you uses type solving you enve have two baselines: main and test. And hopefully you don't have multiple build variants on Android... It's a mess.

But I agree that this fast loop is good so we can add it and iterate it later. This baseline-mess should be something to fix for 2.0

alyssa-glean:
> Thanks! The main problem of mixing cli and gradle is the baseline, I think. If you don't use it, all is fine, but when you run the cli, usually you run it just once for all the project but the baseline is usually per-project/module. And if you uses type solving you enve have two baselines: main and test. And hopefully you don't have multiple build variants on Android... It's a mess.
> 
> But I agree that this fast loop is good so we can add it and iterate it later. This baseline-mess should be something to fix for 2.0

Happily we're not using the baseline stuff, so I haven't encountered these problems. There is a more minor approach of mixing gradle and CLI, though - I have to pipe a few arguments in our hook so the CLI properly mimics our gradle config (`--build-upon-default-config --config '/foo/bar.yml'). So we're kinda maintaining that in two places - I'll have to update our hook in-step with the gradle side to avoid unexpected behaviour. But I'm not anticipating that to change often now it's set up for our project :crossed_fingers:

Another option would be to try and get the file list to `./gradlew detekt-cli` somehow, but that seemed harder and would have had the daemon warm-up problem :shrug:

docs/pages/gettingstarted/git-pre-commit-hook.md
```diff
@@ -38,3 +38,47 @@ More information about Git hooks and how to install them can be found in
 A special thanks goes to Mohit Sarveiya for providing this shell script.
 You can watch his excellent talk about **Static Code Analysis For Kotlin** on 
 [YouTube](https://www.youtube.com/watch?v=LT6m5_LO2DQ).
+
+## Only run on staged files
+
+It is also possible to use [the CLI](cli.md) to create a hook that only runs on staged files. This has the advantage of speedier execution, by running on fewer files and avoiding the warm-up time of the gradle daemon.
+
+Please note, however, that a handful of checks will not work correctly with this approach - those that require [type resolution](type-resolution.md). If you do adopt a partial hook, it is recommended that you still implement a full `detekt` check as part of your CI pipeline.
```
schalkms, line 46:
```suggestion
Please note, however, that a handful of checks requiring [type resolution](type-resolution.md) will not work correctly with this approach. If you do adopt a partial hook, it is recommended that you still implement a full `detekt` check as part of your CI pipeline.
```

schalkms (approved):

schalkms:
I like having a more comprehensive documentation regarding commit hooks. Thank you very much!

cortinico (approved):

alyssa-glean:
We ran into an issue using the hook so I've just pushed our updated version - the `detekt` CLI produces no output if it's successful (as opposed to the `./gradlew` process, which will at least output some gradle bits). This meant the `rm $OUTPUT` final line failed, because no file was created.

I've simplified it to not use a file at all :smile:

cortinico (approved):

PR #4719, opened 2022-04-14
Prepare Detekt 1.20.0

cortinico:
🐣 Happy Easter 🐣


cortinico:
I think there might be something going on with sonatype as the artifact is not showing up:
https://repo1.maven.org/maven2/io/gitlab/arturbosch/detekt/detekt-core/1.20.0/

Should the artifact not show up, I'll do another publishing in some hours.

Issue #4689, opened 2022-04-06
False positive on UnusedImport with const variable import


dimsuz:
The following code when run with type resolution is reporting an error:

```kotlin
import com.example.BuildConfig.FILE_PROVIDER_AUTHORITY
import com.example.BuildConfig.FLAVOR

class ApplicationDelegate {
  override fun getFlavor(): String = FLAVOR
  override fun getFileProviderAuthority(): String = FILE_PROVIDER_AUTHORITY
}
```
The file `BuildConfig` is a java file, generated by the Android Gradle Plugin:
```java
public final class BuildConfig {
  public static final String FLAVOR = "myflavor";
  public static final String FILE_PROVIDER_AUTHORITY = "com.example.debug.provider";
}
```

The error is:

> * The import 'com.example.BuildConfig.FILE_PROVIDER_AUTHORITY' is unused. [UnusedImports]
> * The import 'com.example.BuildConfig.FLAVOR' is unused. [UnusedImports]

## Your Environment
* Version of detekt used: `1.20.0-RC2`
* Version of Gradle used (if applicable): `7.4.x`

eygraber:
@BraisGabin dupe of https://github.com/detekt/detekt/issues/4676?

BraisGabin:
I wouldn't consider it a dope but the issue is probably the same... I'm wondering what created all this new issues... It seems like a change in the compiler behavior

Issue #4676, opened 2022-04-04
False positive on `UnreachableCode`


eygraber:
Using detekt 1.20.0-RC2

The following code results in 2 `UnreachableCode` violations on lines 2 and 4, but I don't see how those are unreachable:

```
private suspend fun transitionToState(mode: Tab) {
    val (arrow, search) = searchBoxDrawables

    when(binding.root.currentState) {
      mode.stateId -> mode.fraction.let { fraction ->
        search.alpha = fraction.searchInterpolation
        arrow.alpha = fraction.arrowInterpolation
      }

      else -> suspendCancellableCoroutine { continuation ->
        val listener = object : TransitionAdapter() {
          override fun onTransitionChange(motionLayout: MotionLayout?, startId: Int, endId: Int, progress: Float) {
            search.alpha = progress.searchInterpolation
            arrow.alpha = progress.arrowInterpolation
          }

          override fun onTransitionCompleted(motionLayout: MotionLayout?, currentId: Int) {
            binding.root.removeTransitionListener(this)
            continuation.resume(Unit)
          }
        }

        continuation.invokeOnCancellation { binding.root.removeTransitionListener(listener) }
        binding.root.addTransitionListener(listener)
        binding.root.transitionToState(mode.stateId)
      }
    }
  }
```

eygraber:
Same issue with lines 4, 11, 16, 27, and 34:

```
override suspend fun findMapPinCluster(latLng: Point): MapPinCluster? {
    val map = binding.mapView.getMapboxMap()

    val cluster =
      binding
        .mapView
        .findFeatures(latLng, Map.Layers.MapPinClusters.id)
        .firstOrNull { feature -> feature.feature.getBooleanProperty("cluster") == true }
        ?: return null

    val zoom = binding.mapView.findGeoJsonClusterExpansionZoom(
      id = Map.Sources.MapPins.id,
      cluster = cluster.feature
    ) ?: return null

    val features = if(map.cameraState.zoom == map.getBounds().maxZoom) {
      binding.mapView.findGeoJsonClusterLeaves(
        id = Map.Sources.MapPins.id,
        cluster = cluster.feature,
        limit = cluster.feature.getNumberProperty("point_count").toLong()
      ) ?: listOf(cluster.feature)
    }
    else {
      listOf(cluster.feature)
    }

    val point =
      cluster
        .feature
        .geometry()
        ?.cast<Point>()
        ?: return null

    return MapPinCluster(
      features = features,
[...]

Issue #3952, opened 2021-07-13
Generating baseline with Detekt 1.18.0-RC1 + Kotlin 1.5.1 uses all remaining disk then fails


brentwatson:
## Expected Behavior

Generating baseline is currently using 100G+ of available disk space, until disk is full, then failing.
This happens with Detekt 1.8.0-RC1 and Kotlin 1.5.1.
It did not happen with Detekt 1.7.1 and Kotlin 1.4.32

## Observed Behavior

```zsh
❯ cd ~/code/my_android_project

❯ du -hs .
 11G	.

❯ ./gradlew detektBaselineMain detektBaselineTest
...
...
java.io.IOException: No space left on device
...
BUILD FAILED in 2h 47m 55s
16407 actionable tasks: 10060 executed, 5100 from cache, 1247 up-to-date
Publishing build scan...
The build scan publishing request was rejected as the data exceeded the maximum size.

❯ du -hs .
114G	.

❯ ./gradlew clean cleanBuildCache
...

❯ du -hs .
 11G	.
```

## Steps to Reproduce

This is a large android project (~1M lines of Kotlin, 300+ modules, and 20 product flavors), and we use detekt tasks that include type resolution.

## Context

We use the gradle plugin version of detekt.

Running `detektMain detektTest` tasks succeed.
Running `detektBaselineMain detektBaselineTest` task fail.

I've tried adding the following config to reduce the number of targets, but get the same issue (uses up all remaining disk space the fails):
```
      ignoredBuildTypes = ["release"]
      ignoredFlavors = [// all our flavors except for 1]
```

We have the following config:
- parallel = true
- allRules = true
- jvmTarget = "1.8"
- `classpath.setFrom(project.configurations.getByName("detekt"))` on both Detekt and DetektCreateBaselineTask tasks.

## Your Environment

* Version of detekt used: 1.8.0-RC1
* Version of Gradle used (if applicable): Gradle 7.0.2
* Gradle scan link (add `--scan` option when running the gradle task): Scan fails with "The build scan publishing request was rejected as the data exceeded the maximum size." error.
* Operating System and version: macOS Catalina (v10.15.7)
* Link to your project (if it's a public repository): N/A - not open source

cortinico:
First, thanks for your report. This will help us ship a `1.18.0` that is hopefully bug-free.

>     * `classpath.setFrom(project.configurations.getByName("detekt"))` on both Detekt and DetektCreateBaselineTask tasks.

[...]

PR #4723, opened 2022-04-15
Android Gradle: add javac intermediates to classpath

kvn-stgl:
This PR is mainly a rebuild of #3867 from @damianw and would fix a lot of issues together with Android (#4676 and #4689 as example). The description is almost the same:

> This adds the classes from build/intermediates/javac to the classpath used to compile the code, in order for the compiler to be able to resolve references to Java code (included that which is generated). This resolves issues such as https://github.com/detekt/detekt/issues/3488 among probably others.
> 
> A test is included which enables viewBinding on a sample project and verifies that the Detekt invocation doesn't produce any errors about unresolved references to the generated classes.

I simply updated the tests and changed that the task detektIntermediatesJavacJarVARIANT depends on the java compile task and not the full assemble task. This should be enough to create the intermediate *.class files and is way faster as the whole packaging process. 
Also I wasn't able to reproduce #3952 with this configuration. But maybe someone with deeper gradle knowledge (@chao2zhang ?) could take a look to be sure.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
@@ -17,7 +17,6 @@ import org.gradle.api.Task
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.FileCollection
 import org.gradle.api.tasks.TaskProvider
-import java.util.Locale
 
 internal class DetektAndroid(private val project: Project) {
 
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
@@ -16,6 +16,8 @@ import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.file.FileCollection
 import org.gradle.api.tasks.TaskProvider
+import org.gradle.api.tasks.bundling.Jar
+import java.util.Locale
 
 internal class DetektAndroid(private val project: Project) {
 
@@ -71,28 +73,42 @@ internal class DetektAndroid(private val project: Project) {
             baseExtension.variants
                 ?.matching { !extension.matchesIgnoredConfiguration(it) }
                 ?.all { variant ->
-                    project.registerAndroidDetektTask(bootClasspath, extension, variant).also { provider ->
+                    val intermediatesJavacJar = project.registerIntermediatesJavacJarTask(variant)
+                    project.registerAndroidDetektTask(
+                        bootClasspath = bootClasspath,
+                        extension = extension,
+                        variant = variant,
+                        intermediatesJavaCJar = intermediatesJavacJar
+                    ).also { provider ->
                         mainTaskProvider.dependsOn(provider)
                     }
-                    project.registerAndroidCreateBaselineTask(bootClasspath, extension, variant)
-                        .also { provider ->
-                            mainBaselineTaskProvider.dependsOn(provider)
-                        }
+                    project.registerAndroidCreateBaselineTask(
+                        bootClasspath = bootClasspath,
+                        extension = extension,
+                        variant = variant,
+                        intermediatesJavaCJar = intermediatesJavacJar
+                    ).also { provider ->
+                        mainBaselineTaskProvider.dependsOn(provider)
+                    }
                     variant.testVariants
                         .filter { !extension.matchesIgnoredConfiguration(it) }
                         .forEach { testVariant ->
-                            project.registerAndroidDetektTask(bootClasspath, extension, testVariant)
-                                .also { provider ->
-                                    testTaskProvider.dependsOn(provider)
-                                }
+                            project.registerAndroidDetektTask(
+                                bootClasspath = bootClasspath,
+                                extension = extension,
+                                variant = testVariant,
+                                intermediatesJavaCJar = intermediatesJavacJar
+                            ).also { provider ->
+                                testTaskProvider.dependsOn(provider)
+                            }
                             project.registerAndroidCreateBaselineTask(
-                                bootClasspath,
-                                extension,
-                                testVariant
-                            )
-                                .also { provider ->
-                                    testBaselineTaskProvider.dependsOn(provider)
-                                }
+                                bootClasspath = bootClasspath,
+                                extension = extension,
+                                variant = testVariant,
+                                intermediatesJavaCJar = intermediatesJavacJar
+                            ).also { provider ->
+                                testBaselineTaskProvider.dependsOn(provider)
+                            }
                         }
                 }
         }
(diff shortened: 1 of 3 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
```diff
@@ -104,38 +120,71 @@ internal class DetektAndroid(private val project: Project) {
...
+        }
+        dependsOn(intermediatesJavaCJar)
         val variantBaselineFile = extension.baseline?.addVariantName(variant.name)
         baseline.set(project.layout.file(project.provider { variantBaselineFile }))
-        description = "EXPERIMENTAL: Creates detekt baseline for ${variant.name} classes with type resolution"
+        description =
+            "EXPERIMENTAL: Creates detekt baseline for ${variant.name} classes with type resolution"
     }
+
+internal fun Project.registerIntermediatesJavacJarTask(variant: BaseVariant): TaskProvider<Jar> {
```
cortinico, line 175: Q: Why do we need a new task for getting the intermediates classes? Wouldn't it be sufficient to just include them in the classpath?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
```diff
@@ -104,38 +120,71 @@ internal class DetektAndroid(private val project: Project) {
...
+    val jarTaskName = "detektIntermediatesJavacJar${variant.name.capitalize(Locale.ROOT)}"
+    val jarOutputDir = layout.buildDirectory.dir(jarTaskName).get()
+    return tasks.register(jarTaskName, Jar::class.java) { task ->
+        task.from(
+            layout
+                .buildDirectory
+                .dir("intermediates/javac/${variant.name}/classes")
```
cortinico, line 182: This feels a bit fragile 🤔 IIRC the new Variant API from AGP allowed to access the compilation classpath for each variant.

kvn-stgl: Well, you're right. I found a way to find out the directory directly over the javaCompileProvider. I think that's a much cleaner solution.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
kvn-stgl: To be honest, I thought it was necessary because the PR #3867 also implemented it. But using the path directly is also possible and it saves a lot overhead.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
```diff
@@ -114,7 +116,12 @@ internal fun Project.registerAndroidDetektTask(
     registerDetektTask(taskName, extension) {
         setSource(variant.sourceSets.map { it.javaDirectories + it.kotlinDirectories })
         extraInputSource?.let { source(it) }
-        classpath.setFrom(variant.getCompileClasspath(null).filter { it.exists() } + bootClasspath)
+        classpath.apply {
+            setFrom(variant.getCompileClasspath(null).filter { it.exists() })
+            from(bootClasspath)
+            from(javaCompileDestination(variant))
+        }
```
3flex, line 123: You should be able to do this:
```kotlin
classpath.setFrom(
    variant.getCompileClasspath(null).filter { it.exists() },
    bootClasspath,
    javaCompileDestination(variant)
}
```

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
```diff
@@ -114,7 +116,12 @@ internal fun Project.registerAndroidDetektTask(
...
+        dependsOn(variant.javaCompileProvider)
```
3flex, line 124: Manually adding task dependencies should not be required if the new inputs to this task are properly declared. Including `javaCompileDestination(variant)` in the classpath should be enough - if not, that's something that should be investigated instead of adding the manual task dependency.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
```diff
@@ -134,8 +141,24 @@ internal fun Project.registerAndroidCreateBaselineTask(
...
         val variantBaselineFile = extension.baseline?.addVariantName(variant.name)
         baseline.set(project.layout.file(project.provider { variantBaselineFile }))
         description = "EXPERIMENTAL: Creates detekt baseline for ${variant.name} classes with type resolution"
     }
+
+private fun Project.javaCompileDestination(variant: BaseVariant): DirectoryProperty? {
+    val javaCompile = variant.javaCompileProvider.orNull
+    if (javaCompile == null) {
+        logger.warn(
+            "Unable to find Java compiler on variant '{}'. Detekt analysis can show false negatives.",
```
3flex, line 159: Is there a way to rewrite the warning to make it actionable for users? Even if that's just updating the AGP version.

Actually while I think of it, will this cause an exception if an older version of AGP is used, if `javaCompileProvider` was only recently added?

chao2zhang: We have the compatibility table listed [here](https://detekt.dev/compatibility.html), so that backward incompatibility with older AGP versions shouldn't be a major concern.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
chao2zhang: This might be necessary though because DetektTask does not necessarily declare on `.java` or `.class` as the sources input.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
3flex: That doesn't address my feedback. If the correct action is to update AGP then the warning message should recommend the update and/or direct users to the compatibility table so they know what to do and why, so that the user understands that using older AGP versions will lead to false positives.

Or if the action to correct this is something else, then that should be shown instead.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
3flex: It doesn't have to, if the properties are wired up correctly. Task dependencies are added implicitly when the input to `classpath` includes types that carry task dependencies. `DirectoryProperty` (returned by `javaCompileDestination(variant)`) is one of those types, and if that's the property that contains the correct path to the `.java` or `.class` files then Gradle has all the information it needs to do the right thing.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
chao2zhang: Note that the `.getJavaCompileProvider()` was added back in 2018-08 through [this PR](https://cs.android.com/android-studio/platform/tools/base/+/8d9060333ea113bb6efcccb6a7b7426dfd9ba7fe).
and `.destinationDirectoy` was built since Gradle 6.1
https://cs.github.com/gradle/gradle/blob/b593969ee3fe63c50466019722cb847d6cd89f4c/subprojects/language-jvm/src/main/java/org/gradle/api/tasks/compile/AbstractCompile.java?q=AbstractCompile

I can't tell easily which AGP version is incompatible with this call, but I assume the current PR is good enough that if customer's build is broken from detekt, it is likely broken from many other plugins and the AGP itself.

kvn-stgl: `.getJavaCompileProvider` was introduced with AGP 3.3 (https://developer.android.com/studio/releases/gradle-plugin#3-3-0), which means Gradle 4.10.

In the end, I don't know what I should recommend to the user. The `.destinationDirectory` should work since Gradle 6.1 (thanks chao2zhang for the research) and detekt has already the minimum of Gradle 6.1+ in the Readme.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektAndroid.kt
kvn-stgl: Good to know, thanks!

cortinico (approved):
Change looks good to me. Thanks for adding tests

PR #4732, opened 2022-04-18
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/codecoverage.yaml
+++ b/.github/workflows/codecoverage.yaml
@@ -8,6 +8,9 @@ on:
     branches:
       - '**'
 
+permissions:
+  contents: read
+
 jobs:
   publish-code-coverage:
     if: ${{ !contains(github.event.head_commit.message, 'coverage skip') }}
--- a/.github/workflows/deploy-snapshot.yaml
+++ b/.github/workflows/deploy-snapshot.yaml
@@ -5,6 +5,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+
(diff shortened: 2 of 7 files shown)


cortinico (approved):

schalkms:
Thanks for including references regarding the permission settings.
I wonder why read and write permissions are the default setting. Wouldn't it be better to have sensible defaults with the least privilege on GitHub CI?

schalkms (approved):

cortinico:
> I wonder why read and write permissions are the default setting. Wouldn't it be better to have sensible defaults with the least privilege on GitHub CI?

Github Actions have already restricted rights when running from forks, so it assumes that it has full write access if running from within a repo. This change is further restricting it.

However nothing prevents you from removing those line and push something to a branch that will execute with write access.

PR #4740, opened 2022-04-20
Remove Unnecesary @Nested

BraisGabin:


BraisGabin:
Once we merge this I will creta a PR adding the hash of that commit to `.git-blame-ignore-revs` to remove it from `blame`. https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

3flex:
Did codecov provide a report? We'll want the coverage diff to be 0%... Not sure how closely anyone will review given the number of lines changed (!!)

BraisGabin:
To give context. To make this PR I implemented a new rule on my junit rule set and then I executed it over detekt until I fixed all the issues.

I did all the changes manually but in general all of them were to remove the first `@Nested` and reduce by one all the indentation.

I don't know if there is a way with github or git to ignore indentation changes. That way the review would be way easier.

BraisGabin:
I would appreciate to have fast feeback on this PR. If you think that it shouldn't be merged I can close it. But this PR will get a lot of conflicts so I would like to take a decission about it asap.

cortinico (approved):
Thanks for doing this, I'd love to merge it. I got enticed away form the big diff size 😅 

I checked a couple of files and it looks good to me. I trust the CI at this stage ¯\_(ツ)_/¯

PR #4745, opened 2022-04-21
`VarCouldBeVal`: Add configuration flag `ignoreLateinitVar`

### LINKED ISSUE #4731: `VarCouldBeVal` reported on `private lateinit var`
## Steps to Reproduce
```kt
class A {
    private lateinit var label: JLabel
}
```

## Context
Per https://github.com/detekt/detekt/pull/4664#issuecomment-1086835405, I'm not sure our use-case is mainstream enough to warrant a change in detekt, but I'd like to bring it up as a datapoint anyways in case there there is additional discussion in the future.

As part of my work I develop plugins for JetBrains IDEs.
We have a number of older Kotlin source files backing visual forms designed using the JetBrains [GUI Designer](https://www.jetbrains.com/help/idea/gui-designer.html). I'm fuzzy on all the details, but how this generally works is that there's instrumentation at compile time that injects constructors and all the necessary method calls needed to setup the Swing components correctly.

For example, the code snippet above would result in decompiled bytecode that might look like the following:
```kt
class A {
    private lateinit var label: JLabel

    private fun `$$$setupUI$$$`() {
        label = JLabel("some label text")
        label.setIcon(Icon("and maybe we have a /path/to/icon declared here too"))
    }

    init {
        `$$$setupUI$$$`() 
    }
}
```

Obviously since detekt acts on source code and not bytecode, it can't know that the fields actually end up initialized.

The `VarCouldBeVal` rule is still very useful for us, so as a workaround to this regression we'd likely either mark the ~60 or so instances in our codebase with an annotation or throw it in a baseline file and call it a day.

## Your Environment
* Link to your project (if it's a public repository): https://github.com/aws/aws-toolkit-jetbrains

rli:
I'm not entirely convinced that it makes sense to disable this across the entire codebase, but maybe someone else has a use-case.

Fixes #4731

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -675,6 +675,7 @@ style:
     active: true
   VarCouldBeVal:
     active: true
+    ignoreLateinitVar: false
   WildcardImport:
     active: true
     excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeVal.kt
@@ -8,8 +8,11 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.config
 import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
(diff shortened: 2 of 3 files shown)


BraisGabin:
> I'm not entirely convinced that it makes sense to disable this across the entire codebase, but maybe someone else has a use-case.

I agree but how could we fix this otherwise? I mean, do you see any better way to fix this for your use case? In general I don't want to add a feature just because "maybe someone else has a use-case".

In your case, does it seem right to add an `@Suppress` to those files? Or to exclude the path of those files? Or to exclude them by package?

Sorry maybe I over-simplify your use case and I propouse a solution that doesn't fit your use case.

vladimirfx:
I have a myriad of use cases. 90% of all our `lateinit` usages are flagged. Our config now is polluted with `ignoreAnnotated `:

```yaml
  VarCouldBeVal:
    active: true
    ignoreAnnotated:
      - 'javax.persistence.Version'
      - 'javax.ws.rs.core.Context'
      - 'javax.inject.Inject'
      - 'org.springframework.beans.factory.annotation.Autowired'

```

But all these excludes are NOT enough because there are cases where no annotations at field level (see @QuarkusTestResource which is set at class level).

Very waiting for merge.

cortinico (approved):

BraisGabin (approved):

PR #4751, opened 2022-04-23
Move detekt-gradle-plugin to be a composite build

cortinico:
Stacked on top of #4748 

This converts our project to use Gradle Composite builds.
With this we won't need to depend on external, previously published versions of `detekt-gradle-plugin`. It will help us prevent and catch earlier a lot of bugs we discover during the release process.

I've also introduced a `buildAll` top level task that builds everything (all the subprojects and run the build task on the included build).

I've removed the `applySelfAnalysisVersion` task as that's unnecessary now.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -3,7 +3,7 @@ import io.gitlab.arturbosch.detekt.DetektCreateBaselineTask
 
 plugins {
     id("releasing")
-    alias(libs.plugins.detekt)
+    id("io.gitlab.arturbosch.detekt")
     alias(libs.plugins.gradleVersions)
     alias(libs.plugins.sonarqube)
 }
@@ -104,4 +104,8 @@ val detektProjectBaseline by tasks.registering(DetektCreateBaselineTask::class) 
     exclude(resourceFiles)
     exclude(buildFiles)
     baseline.set(baselineFile)
+}
+
+tasks.register("build") {
+    dependsOn(gradle.includedBuild("detekt-gradle-plugin").task(":build"))
 }
--- a/detekt-gradle-plugin/build.gradle.kts
+++ b/detekt-gradle-plugin/build.gradle.kts
@@ -9,14 +9,13 @@ plugins {
     alias(libs.plugins.pluginPublishing)
 }
 
-detekt {
-    source.from("src/functionalTest/kotlin")
-}
-
 repositories {
     mavenCentral()
     google()
 }
+
+group = "io.gitlab.arturbosch.detekt"
+version = Versions.currentOrSnapshot()
 
 testing {
     suites {
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -82,4 +82,4 @@ jobs:
       - name: Build and compile test snippets
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: test -x :detekt-gradle-plugin:test -Pcompile-test-snippets=true
+          arguments: test -x -Pcompile-test-snippets=true
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -82,4 +82,4 @@ jobs:
       - name: Build and compile test snippets
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: test -x -Pcompile-test-snippets=true
+          arguments: test -Pcompile-test-snippets=true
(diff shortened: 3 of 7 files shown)


cortinico:
I'd like to hear @3flex opinion on this before we merge it 👍

build.gradle.kts
```diff
@@ -105,3 +105,11 @@ val detektProjectBaseline by tasks.registering(DetektCreateBaselineTask::class)
     exclude(buildFiles)
     baseline.set(baselineFile)
 }
+
+tasks.register("buildAll") {
```
3flex, line 109: I'd just do

```kotlin
tasks.build {
    dependsOn(gradle.includedBuild("detekt-gradle-plugin").task(":build"))
}
```

I can't think of a good reason to register a task with a new name to wire up any dependent tasks. Calling tasks in a project will execute matching tasks in any subprojects relative to the working directory ([source](https://docs.gradle.org/current/userguide/intro_multi_project_builds.html)) so keeping `build` as the task to do this makes more sense to me.

3flex, line 109: I'd just do

```kotlin
tasks.build {
    dependsOn(gradle.includedBuild("detekt-gradle-plugin").task(":build"))
}
```

I can't think of a good reason to register a task with a new name to wire up any dependent tasks. Calling tasks in a project will execute matching tasks in any subprojects relative to the working directory ([source](https://docs.gradle.org/current/userguide/intro_multi_project_builds.html)) so keeping `build` as the task to do this makes more sense to me.

detekt-gradle-plugin/build.gradle.kts
```diff
@@ -9,15 +9,14 @@ plugins {
     alias(libs.plugins.pluginPublishing)
 }
 
-detekt {
-    source.from("src/functionalTest/kotlin")
```
3flex, line 13: This needs to stay, there's no equivalent config anywhere else in the project to add `src/functionalTest` to the source checked by the `detekt` task.

3flex (commented):
Looks good, there may be a few teething issues but I think this is a win overall!

BraisGabin (approved):

build.gradle.kts
cortinico: I've ended up with:

`tasks.register("build") {`

As there is no build task on the top level project.

detekt-gradle-plugin/build.gradle.kts
cortinico: Sadly, this can't stay. The `detekt-gradle-plugin` build can't use itself.

We would have to use again an older published version of the `detekt-gradle-plugin` which I would like to avoid frankly (as it's the problem we're trying to address here.

I would rather rely on the `runWithArgsFile` as it should handle also the `src/functionalTest/kotlin` folder.

chao2zhang: `runWithArgsFile` checks the entire codebase, so it should cover all sources under `src/functionalTest`

chao2zhang:
This change may impact the Gradle Enterprise trial progress. As we found that `functionalTest` does not conform to the incremental buil. 
https://ge.detekt.dev/s/vue3arlrp3gvw/timeline?anchor=5cf5p64octxus
image
Let's merge this in as soon as we can.

detekt-gradle-plugin/build.gradle.kts
3flex: The reason I added this block to this build file is running `:detekt-gradle-plugin:build` doesn't check `functionalTest` sources without it, because none of the configured Gradle detekt tasks in detekt-gradle-plugin checked it, so CI would fail when I pushed and `runWithArgsFile` ran, and it became quite frustrating. Running `runWithArgsFile` locally is just another thing to remember to do before pushing, and it's poor DX.

> detekt-gradle-plugin build can't use itself.

True, but we can add `id("io.gitlab.arturbosch.detekt") version "1.20.0"` to the plugins block in detekt-gradle-plugin. I assume renovate will notify about updates, and updating that version still won't be tied to the release process.

cortinico: I see, thanks for clarifying. I think it's ok to include the `id("io.gitlab.arturbosch.detekt") version "1.20.0"` and don't bump it as part of the release process (let's renovate do the job as you suggested).

cortinico: Are you fine with me adding this in a separate PR?

3flex: > Are you fine with me adding this in a separate PR?

Yep, LGTM!

PR #4752, opened 2022-04-23
Cleanup Gradle Plugin Publications

cortinico:
Stacked on top of #4751 

I've removed the `isAutomatedPublishing` hack we had for the Gradle Plugin.
This means that we'll rely on the publishing configured by the Gradle Publishing plugin (We might want to configure the POM though).

The side effect of this, is that now we'll have both the Gradle Plugin and the Gradle Plugin marker published, either on Maven Local or Central.

This will simplify local development + will allow users to depend on the plugin from Central only.

I've also added several top level tasks to help facilitate this process.


chao2zhang (approved):

build.gradle.kts
```diff
@@ -113,3 +113,33 @@ tasks.register("buildAll") {
...
+
+tasks.register("publishAllToMavenCentral") {
+    description = "Publish all the projects to Sonatype Staging Repository"
+    subprojects {
+        if (this.plugins.hasPlugin("publishing")) {
+            dependsOn(tasks.named("publishAllPublicationsToMavenCentralRepository"))
+        }
+    }
+    dependsOn(gradle.includedBuild("detekt-gradle-plugin").task(":publishAllPublicationsToMavenCentralRepository"))
+}
```
3flex, line 145: All of these tasks could go into `releasing.gradle.kts` to keep the root project file tidy (and because these tasks are all about releasing/publishing)

3flex (approved):

build-logic/src/main/kotlin/packaging.gradle.kts
```diff
@@ -27,31 +27,34 @@ publishing {
...
-                developer {
-                    id.set("Artur Bosch")
-                    name.set("Artur Bosch")
-                    email.set("arturbosch@gmx.de")
+                developers {
+                    developer {
+                        id.set("Artur Bosch")
+                        name.set("Artur Bosch")
+                        email.set("arturbosch@gmx.de")
+                    }
```
BraisGabin, line 53: I can't care less about this, but, do we want to publish developer? If we keep it as a reference to the great work that Artur did I'm completely fine. Otherwise I would keep this empty because it is impossible to list all the developers that contributed to this project

build.gradle.kts
BraisGabin: And maybe expand the current task instead of creating a new one?

BraisGabin (approved):

build.gradle.kts
cortinico: Moved. 

> expand the current task

There are no publishing tasks registered on the top level project. So they all need to be created. I've updated the `publishToMavenLocal` one, as that might be used by contributors. For the other 2, I've used shorter version as they're easier to understand

build-logic/src/main/kotlin/packaging.gradle.kts
cortinico: Updated with a generic Detekt Developers + mail

BraisGabin (approved):
LGTM

PR #4761, opened 2022-04-24
Fix UP-TO-DATE checks for `functionalTest`

cortinico:
Stacked on top of #4751 

This fixes a failure on a UP-TO-DATE checks for `functionalTest`.
Proof: https://ge.detekt.dev/c/c52dhidy2tdqc/tcq346kqad2pq/task-inputs

The problem was that this specific test was generating output (i.e. a detekt HTML report) in a resource folder. Solved by reading the resource folder and copying the content inside a temp folder.

(cc @chao2zhang @runningcode as they have context).

PR #4763, opened 2022-04-25
Gradle Enterprise Trial

chao2zhang:
This PR is a live branch to bookkeep all Gradle enterprise fixes to facilitate the before-and-after evaluation for Gradle enterprise.

[Gradle Enterprise Trial Navigator for Gradle.pdf](https://github.com/detekt/detekt/files/8850041/Gradle.Enterprise.Trial.Navigator.for.Gradle.pdf)


3flex:
On moving the test out of the `resources` folder - what do you think about future functional test for multiplatform and Android projects? Those could have cases that require multiple modules, each with violations that we would test that detekt picks up, and I feel that generating all of these files by placing the file content in a test script instead of a self-contained project can get quite confusing. IMHO it's easier to reason about a test project if it's files on the system instead of source in strings in a test script.

I was inspired by SQLDelight when setting up the self-contained project. SQLDelight has many Gradle projects which are used for functional testing: https://github.com/cashapp/sqldelight/tree/master/sqldelight-gradle-plugin/src/test

There are pros and cons with each approach of course - the Gradle project itself tends to setup tests in code, as a counter-example to SQLDelight.

chao2zhang:
I agreed that it would be easier to explore and review if we hold these test fixtures in `resources` in an organized way.

The only cons from my exploration in the checkstyle repo are that, when we want to start to test variants, we either need to copy-paste test fixtures or define some template within the test fixtures, the readability, in the end, would be jeopardized.

3flex:
One problem with the tests right now (which I'm actually working on over the next few days as time allows) is that many functional tests can be rewritten to use ProjectBuilder (#3778) which makes testing different use cases simpler, as less setup code is required for each one. I think that's where majority of repetition for different test cases will end up (e.g. checking that task creation and wiring dependencies are as expected for different Kotlin plugins), with the functional tests only testing a few specific cases.

That might tip the balance in favour of keeping test fixtures as self-contained projects.

cortinico:
Should we convert this to a Draft PR?

detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
```diff
@@ -3,87 +3,132 @@ package io.gitlab.arturbosch.detekt
...
+                    mavenCentral()
+                    mavenLocal()
+                }
 
-                    repositories {
-                        mavenCentral()
-                        mavenLocal()
+                detekt {
+                    reports {
+                        txt.destination = file("output-path.txt")
```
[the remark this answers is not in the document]
runningcode: I see, should we remove the other case then?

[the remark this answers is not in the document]
chao2zhang: I believe we were softly deprecating this and move towards hard removal in later versions.

chao2zhang:
> One problem with the tests right now (which I'm actually working on over the next few days as time allows) is that many functional tests can be rewritten to use ProjectBuilder (#3778) which makes testing different use cases simpler, as less setup code is required for each one. I think that's where majority of repetition for different test cases will end up (e.g. checking that task creation and wiring dependencies are as expected for different Kotlin plugins), with the functional tests only testing a few specific cases.
> 
> That might tip the balance in favour of keeping test fixtures as self-contained projects.

I see your point - Totally agreed upon simplifying functional tests and using unit test with `ProjectBuiilder` to cover the varieties.

I am going to base my change on https://github.com/detekt/detekt/pull/4761/files to simplify the resource copy in a more organized way (mostly directory structure and naming).

settings.gradle.kts
```diff
@@ -66,6 +66,19 @@ gradleEnterprise {
     }
 }
 
+buildCache {
+    local { isEnabled = false }
```
3flex, line 70: Do we want local cache disabled?

From https://docs.gradle.org/current/userguide/build_cache.html#sec:build_cache_configure:
> When both build caches are enabled, Gradle tries to load build outputs from the local build cache first, and then tries the remote build cache if no build outputs are found. If outputs are found in the remote cache, they are also stored in the local cache, so next time they will be found locally.

Seems like that's the behaviour we'd want, at least for non CI-builds?
```suggestion
    local { isEnabled = !isCiBuild }
```

What do you think?

chao2zhang: Thank you for watching this thread! I am current at experiment 4 in the following trial now and I would be live that we always want to have local cache enabled at the end of the trial.
[Gradle Enterprise Trial Navigator for Gradle.pdf](https://github.com/detekt/detekt/files/8647982/Gradle.Enterprise.Trial.Navigator.for.Gradle.pdf)

chao2zhang: (I probably will continue using this thread for testing/tweaking purposes).

code-coverage-report/build.gradle.kts
```diff
@@ -63,4 +63,5 @@ val customClassDirectories = configurations.allCodeCoverageReportClassDirectorie
 
 tasks.named("jacocoMergedReport", JacocoReport::class).configure {
     this.classDirectories.setFrom(customClassDirectories.files)
+    mustRunAfter(rootProject.project("detekt-generator").tasks.named("generateDocumentation"))
```
chao2zhang, line 66:
```
Execution optimizations have been disabled for task ':code-coverage-report:jacocoMergedReport' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/Users/chaozhang/detekt/detekt-core/src/main/resources'. Reason: Task ':code-coverage-report:jacocoMergedReport' uses this output of task ':detekt-generator:generateDocumentation' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.4.2/userguide/validation_problems.html#implicit_dependency for more details about this problem.
```

detekt-gradle-plugin/build.gradle.kts
```diff
@@ -46,21 +46,24 @@ testing {
...
-            }
         }
         register("functionalTest", JvmTestSuite::class) {
             useJUnitJupiter(libs.versions.junit.get())
 
             dependencies {
                 implementation(libs.assertj)
             }
+
+            targets {
```
chao2zhang, line 57: This is moved to `functionalTest` because `DetektAndroidSpec` is actually run in the `functionalTest` test suite.

chao2zhang:
This PR should be ready for review as we went through all 5 experiments in the Gradle enterprise trial.

Once merged into `main`, we will verify the scenario of CI build-cache + anonymous local checkout.

build-logic/src/main/kotlin/module.gradle.kts
```diff
@@ -57,12 +57,12 @@ tasks.withType<KotlinCompile>().configureEach {
...
         )
-        // Usage: <code>./gradlew build -PwarningsAsErrors=true</code>.
-        // Note: currently there are warnings for detekt-gradle-plugin that seemingly can't be fixed
+        // Note: Currently there are warnings for detekt-gradle-plugin that seemingly can't be fixed
         //       until Gradle releases an update (https://github.com/gradle/gradle/issues/16345)
+        // To ensure CI ~ Local build cache hit, we shouldn't use system env CI.
         allWarningsAsErrors = when (project.name) {
             "detekt-gradle-plugin" -> false
-            else -> (project.findProperty("warningsAsErrors") == "true" || System.getenv("CI") == "true")
+            else -> true
```
BraisGabin, line 65:
```suggestion
            else -> project.findProperty("warningsAsErrors") == "true"
```

I think that we shouldn't have this enable by default. It's a pain to develop with this enabled on local.

BraisGabin (approved):

build-logic/src/main/kotlin/module.gradle.kts
```diff
@@ -57,12 +57,12 @@ tasks.withType<KotlinCompile>().configureEach {
...
         )
-        // Usage: <code>./gradlew build -PwarningsAsErrors=true</code>.
-        // Note: currently there are warnings for detekt-gradle-plugin that seemingly can't be fixed
+        // Note: Currently there are warnings for detekt-gradle-plugin that seemingly can't be fixed
         //       until Gradle releases an update (https://github.com/gradle/gradle/issues/16345)
+        // To ensure CI ~ Local build cache hit, we shouldn't use system env CI.
         allWarningsAsErrors = when (project.name) {
             "detekt-gradle-plugin" -> false
-            else -> (project.findProperty("warningsAsErrors") == "true" || System.getenv("CI") == "true")
+            else -> project.findProperty("warningsAsErrors") == "true"
```
3flex, line 65: The problem here is, if we don't run with warnings as errors on CI, then we lose that validation step for PRs. If it's enabled on CI, but disabled on local builds, then there won't be a remote cache hit (assuming `allWarningsAsErrors` is annotated with `@Input` or similar?)

We might want to run a build with `-PwarningsAsErrors=true` somewhere on CI, in addition to the standard builds, which won't have it enabled but will help with cache hits.

3flex (commented):
Few comments, looks good overall though - especially if the remote build cache works as advertised.

detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/GradleVersionSpec.kt
```diff
@@ -22,6 +22,6 @@ class GradleVersionSpec {
     }
 
     companion object {
-        const val gradleVersion = "6.1"
+        const val gradleVersion = "6.6.1"
```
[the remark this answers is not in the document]
chao2zhang: Yeah upgrading the gradle version was an failed attempt. For some reason the OOM issue seems to be persistent at least on this PR though.

build-logic/src/main/kotlin/module.gradle.kts
BraisGabin: > We might want to run a build with `-PwarningsAsErrors=true` somewhere on CI, in addition to the standard builds, which won't have it enabled but will help with cache hits.

That's a good idea!


I also want to create a detekt-plugin to mark as issues all the warnings that the compiler find. That should help too. (But I'm short in time lately so I don't know when I'll be able to do that)

chao2zhang: I am going to add an additional Github workflow for such validation. I suppose this is the way moving forward - If there are any experimental flag (like Gradle warnings, configuration cache, or Kotlin K2 alpha), we should run additional jobs with different configuration.

detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/GradleVersionSpec.kt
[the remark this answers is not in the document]
chao2zhang: I cleared the Gradle build cache this time, let's see if this get Win + Java 11 passed.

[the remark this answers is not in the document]
chao2zhang: Looks like it has worked.

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/GradleRunnerExtensions.kt
```diff
@@ -0,0 +1,16 @@
+package io.gitlab.arturbosch.detekt.testkit
+
+import org.gradle.testkit.runner.GradleRunner
+import java.io.File
+import java.nio.file.Files
+
+/**
+ * Copy project files from `resources` to temporary directories for isolation.
+ * This helps with the incremental build (up-to-date checks)
```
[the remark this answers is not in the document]
chao2zhang: Let me document this in the comment for future consideration as this is an optimization (And theoretically shouldn't significantly impact test runtime).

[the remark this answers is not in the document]
chao2zhang: https://github.com/detekt/detekt/issues/4949

chao2zhang:
I am merging this PR for the meeting with Gradle folks tomorrow. I will monitor the trunk status closely.

chao2zhang:
https://github.com/gradle/gradle-enterprise-oss-projects/blob/main/.github/workflows/run-experiments-detekt.yml is where the previous GE experiments are being tracked.

PR #4765, opened 2022-04-25
Self-inspect the detekt-gradle-plugin

cortinico:
Here I'm setting up the self-inspection on `detekt-gradle-plugin` (cc @3flex).

I have to admit that I'm not extremely excited of this, as we might face problems with the config file of detekt being incompatible between the latest released version and `main`.

### DIFF
--- a/detekt-gradle-plugin/build.gradle.kts
+++ b/detekt-gradle-plugin/build.gradle.kts
@@ -7,6 +7,8 @@ plugins {
     `java-test-fixtures`
     idea
     alias(libs.plugins.pluginPublishing)
+    // We use this published version of the Detekt plugin to self analyse this project.
+    id("io.gitlab.arturbosch.detekt") version "1.20.0"
 }
 
 repositories {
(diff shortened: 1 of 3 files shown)


3flex:
> we might face problems with the config file of detekt being incompatible between the latest released version and main

Good point - now that the Gradle plugin is in its own isolated build and running a different version of detekt, this is unavoidable if the same config is used for the plugin as the rest of the project.

So I would suggest giving the Gradle plugin its own config and baseline file (though I'm not sure if there are any issues in the baseline at the moment?)

The repetition might be undesirable but I agree with what I think you're saying, that we should prioritise stability of the build.

cortinico:
> So I would suggest giving the Gradle plugin its own config and baseline file (though I'm not sure if there are any issues in the baseline at the moment?)

Yup that sounds like a good compromise. Added here [8bd1bfd](https://github.com/detekt/detekt/pull/4765/commits/8bd1bfd41cc2f95da29202a21e365e8b3c1c9785)

3flex (approved):

Issue #4462, opened 2022-01-06
KDoc comments should not refer to private member variables


Whathecode:
I just ran into code which in KDoc comments pointed to a private member variable. E.g.

```
/**
  * Uses [answerCalculator] to find the answer to life, the universe, and everything.
  */
class Answer
{
    private val answerCalculator: String = "42"
    public fun getAnswer(): String = answerCalculator
}
```

## Expected Behavior of the rule
Given that `KDoc` is for _documentation_, intended to be useful to _callers_ of the thing being documented so that they don't need to read/understand encapsulated implementation details, wouldn't it make sense to consider it a code smell when documentation points to implementation internals?

## Context
This can help people to have some safeguards in place for writing better documentation. That said, I believe this is the first time I ran into code like this, so it may not be prevalent.

cortinico:
> wouldn't it make sense to consider it a code smell when documentation points to implementation internals?

Potentially yes. However, I've been doing this multiple times in the past and I think it's acceptable. If we end up implementing this I believe it's going to be a rule disabled by default

BraisGabin:
I agree that detekt should flag this.

BraisGabin:
Related with #3657

VitalyVPinchuk:
Hello,
may I try to work on the issue?

cortinico:
Yup I've assigned it to you @[VitalyVPinchuk](https://github.com/VitalyVPinchuk)

PR #4768, opened 2022-04-25
Report KDoc comments that refer to non-public properties of a class

VitalyVPinchuk:
New rule to address the issue (#4462) of breaking encapsulation principle when KDoc comments have references to non-public properties.
Clients do not need to know the implementation details.

Screenshot
Screenshot 2022-04-26 at 22 07 12

### DIFF
--- a/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicProperty.kt
+++ b/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicProperty.kt
@@ -20,8 +20,6 @@ import org.jetbrains.kotlin.psi.psiUtil.isPublic
  * This rule will report any KDoc comments that refer to non-public properties of a class.
  * Clients do not need to know the implementation details.
  *
- * See [Encapsulation](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming))
- *
  * <noncompliant>
  * /**
  *  * Comment
--- a/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicProperty.kt
+++ b/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicProperty.kt
@@ -13,6 +13,7 @@ import org.jetbrains.kotlin.psi.KtObjectDeclaration
 import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
 import org.jetbrains.kotlin.psi.psiUtil.getTopmostParentOfType
+import org.jetbrains.kotlin.psi.psiUtil.isProtected
 import org.jetbrains.kotlin.psi.psiUtil.isPublic
 
 /**
@@ -66,12 +67,12 @@ class KDocReferencesNonPublicProperty(config: Config = Config.empty) : Rule(conf
     }
 
     private fun KtProperty.isNonPublicInherited(): Boolean {
-        if (!isPublic) {
+        if (!isPublic && !isProtected()) {
             return true
         }
         var classOrObject = containingClassOrObject
         while (classOrObject != null && classOrObject is KtObjectDeclaration) {
-            if (!classOrObject.isPublic) {
+            if (!classOrObject.isPublic && !isProtected()) {
                 return true
             }
             classOrObject = classOrObject.containingClassOrObject
--- a/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicProperty.kt
+++ b/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicProperty.kt
@@ -16,31 +16,56 @@ import org.jetbrains.kotlin.psi.psiUtil.getTopmostParentOfType
 import org.jetbrains.kotlin.psi.psiUtil.isPublic
 
 /**
- * This rule will report any KDoc comments that refer to encapsulated properties of a class.
+ * This rule will report any KDoc comments that refer to non-public properties of a class.
  * Clients do not need to know the implementation details.
+ *
  * See [Encapsulation](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming))
+ *
+ * <noncompliant>
+ * /**
+ *  * Comment
+ *  * [prop1] - non-public property
+ *  * [prop2] - public property
+ *  */
+ * class Test {
+ *     private val prop1 = 0
+ *     val prop2 = 0
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * /**
+ *  * Comment
+ *  * [prop2] - public property
+ *  */
+ * class Test {
+ *     private val prop1 = 0
+ *     val prop2 = 0
+ * }
+ * </compliant>
+ *
  */
-class ReferencedEncapsulatedProperty(config: Config = Config.empty) : Rule(config) {
+class KDocReferencesNonPublicProperty(config: Config = Config.empty) : Rule(config) {
 
     override val issue = Issue(
         javaClass.simpleName,
         Severity.Maintainability,
-        "KDoc comments should not refer to encapsulated properties.",
+        "KDoc comments should not refer to non-public properties.",
         Debt.FIVE_MINS
     )
 
     override fun visitProperty(property: KtProperty) {
+        super.visitProperty(property)
+
         val enclosingClass = property.getTopmostParentOfType<KtClass>()
         val comment = enclosingClass?.docComment?.text ?: return
 
-        if (property.isEncapsulatedInherited() && property.isReferencedInherited(comment)) {
+        if (property.isNonPublicInherited() && property.isReferencedInherited(comment)) {
             report(property)
         }
-
-        super.visitProperty(property)
     }
 
-    private fun KtProperty.isEncapsulatedInherited(): Boolean {
+    private fun KtProperty.isNonPublicInherited(): Boolean {
         if (!isPublic) {
             return true
         }
@@ -70,7 +95,7 @@ class ReferencedEncapsulatedProperty(config: Config = Config.empty) : Rule(confi
                 issue,
                 Entity.atName(property),
                 "The property ${property.nameAsSafeName} " +
-                    "is encapsulated and should not be referenced from KDoc comments."
+                    "is non-public and should not be referenced from KDoc comments."
             )
         )
     }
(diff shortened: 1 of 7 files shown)


detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedProperty.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
+import org.jetbrains.kotlin.psi.psiUtil.getTopmostParentOfType
+import org.jetbrains.kotlin.psi.psiUtil.isPublic
+
+/**
+ * This rule will report any KDoc comments that refer to encapsulated properties of a class.
+ * Clients do not need to know the implementation details.
+ * See [Encapsulation](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming))
+ */
+class ReferencedEncapsulatedProperty(config: Config = Config.empty) : Rule(config) {
```
cortinico, line 23: I think we can work on the naming here. 

Ideally `KDocReferencesNonPublicProperty` sounds better to me.
Happy to hear other opinion

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedProperty.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Maintainability,
+        "KDoc comments should not refer to encapsulated properties.",
```
cortinico, line 28: I would refrain from using encapsulated and use `non public` instead.

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedProperty.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    override fun visitProperty(property: KtProperty) {
+        val enclosingClass = property.getTopmostParentOfType<KtClass>()
+        val comment = enclosingClass?.docComment?.text ?: return
+
+        if (property.isEncapsulatedInherited() && property.isReferencedInherited(comment)) {
+            report(property)
+        }
+
+        super.visitProperty(property)
```
cortinico, line 40: this should be first

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedProperty.kt
```diff
@@ -0,0 +1,77 @@
...
+        return comment.contains("[$qualifiedName]")
+    }
+
+    private fun report(property: KtNamedDeclaration) {
+        report(
+            CodeSmell(
+                issue,
+                Entity.atName(property),
+                "The property ${property.nameAsSafeName} " +
+                    "is encapsulated and should not be referenced from KDoc comments."
```
cortinico, line 73: same as above

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedProperty.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtNamedDeclaration
+import org.jetbrains.kotlin.psi.KtObjectDeclaration
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
+import org.jetbrains.kotlin.psi.psiUtil.getTopmostParentOfType
+import org.jetbrains.kotlin.psi.psiUtil.isPublic
+
+/**
+ * This rule will report any KDoc comments that refer to encapsulated properties of a class.
```
cortinico, line 19: You're missing `<compliant>` and non compliant examples.

cortinico (commented):
Thanks for adding this 👍

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedProperty.kt
```diff
@@ -0,0 +1,77 @@
...
+            report(property)
+        }
+
+        super.visitProperty(property)
+    }
+
+    private fun KtProperty.isEncapsulatedInherited(): Boolean {
+        if (!isPublic) {
+            return true
+        }
```
marschwar, line 46: How about `protected` properties? I consider them part of the public api as well. Are they not?

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedProperty.kt
VitalyVPinchuk: fixed

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedProperty.kt
VitalyVPinchuk: fixed

detekt-rules-documentation/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedPropertySpec.kt
```diff
@@ -0,0 +1,98 @@
...
+
+import io.gitlab.arturbosch.detekt.test.compileAndLint
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+class ReferencedEncapsulatedPropertySpec {
+    val subject = ReferencedEncapsulatedProperty()
+
+    @Nested
```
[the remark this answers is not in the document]
VitalyVPinchuk: Fixed, but IntelliJ IDEA doesn't let run tests from IDE without `@Nested`. Is it ok?

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedProperty.kt
VitalyVPinchuk: Sorry, I didn't get it. `protected `class members are not accessible either.

Is this it? [kotlinlang.org](https://kotlinlang.org/docs/visibility-modifiers.html#class-members)
```
open class Outer {
    private val a = 1
    protected open val b = 2
}
class Unrelated(o: Outer) {
    // o.a, o.b are not visible
}
```

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedProperty.kt
VitalyVPinchuk: fixed

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedProperty.kt
marschwar: If  a public class has a protected property, that property should be considered part of the api and therefor should be allowed to be referenced in the documentation such as in 

```kotlin
// Library class
/**
 * Some doc that references [a].
 */
open class Parent {
    protected var a: Int = 0
}

// Application class
class Child : Parent() {
    fun foo() {
        print("foo $a")
    }
}
```

marschwar:
Just curious - Should this rule also check the KDoc of methods and properties to verify it is not leaking private information?

```kotlin
class C {
    private var a: Int = 0

    /**
     * Mentioning [a]
     */
    val b = "string"

    /**
     * Mentioning [a]
     */
    fun foo() {
        print("foo $a")
    }
}
```

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedProperty.kt
VitalyVPinchuk: Oh, I see your point. Interesting! I haven't thought about subclasses.
Probably, it makes sense to check if class is `open` and then allow `protected` members. From the other hand, `protected` doesn't make sense for final classes.

VitalyVPinchuk:
> Just curious - Should this rule also check the KDoc of methods and properties to verify it is not leaking private information?

I think it's a good idea!
Should I add these checks?

marschwar:
> Should I add these checks?

If you think they are valid, you could do so. I am just afraid that the rule will become overly complicated and hard to maintain. 

To be absolutely honest, I am not the biggest fan of documentation rules to begin with. Since you brought up the issue, you seem to have a clear use case.

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/ReferencedEncapsulatedProperty.kt
marschwar: > From the other hand, `protected` doesn't make sense for final classes.

True, but that should be handled by a different rule. For the purpose of this rule we should assume that the protected modifier is use correctly.

VitalyVPinchuk: fixed

VitalyVPinchuk:
By the way, why does codecov report such a small coverage?

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicProperty.kt
```diff
@@ -0,0 +1,103 @@
...
+import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
+import org.jetbrains.kotlin.psi.psiUtil.getTopmostParentOfType
+import org.jetbrains.kotlin.psi.psiUtil.isProtected
+import org.jetbrains.kotlin.psi.psiUtil.isPublic
+
+/**
+ * This rule will report any KDoc comments that refer to non-public properties of a class.
+ * Clients do not need to know the implementation details.
+ *
+ * See [Encapsulation](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming))
```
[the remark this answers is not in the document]
VitalyVPinchuk: fixed

PR #4769, opened 2022-04-25
fix: add test case that fails if environment is not properly set up

marschwar:
This relatest to the comment https://github.com/detekt/detekt/pull/4575#discussion_r841284879 and adds a test case that fails if the java environment is not properly set up. The original test still succeeds because the function cannot be resolved and will not issue a code smell by default.


schalkms (approved):
Shall we add a comment or an extended test description for the reasoning behind the test.

Issue #4705, opened 2022-04-11
Multiline string rules


3flex:
This could be either a single, configurable rule, or a series of rules, that performs checks on the contents of multiline strings.

## Expected Behavior of the rule
* Check for consistent indentation type (could be done in one of two ways - either check where a mix of indentation is used so it's not a mix of tabs & spaces, or have a rule that enforces either tabs or spaces)
* Check for trailing whitespace on each line
* When using `trimIndent()` or `trimMargin()` check that the left-most indentation is indented x characters from the enclosing element
* Check that every multiline string uses either `trimMargin()` or `trimIndent()` - Kotlin Coding Conventions doc says "To maintain indentation in multiline strings, use `trimIndent` when the resulting string does not require any internal indentation, or `trimMargin` when internal indentation is required"

We could also have a rule that checks for use of `\n` in a regular string literal and recommend converting to a multiline string, as the Kotlin Coding Conventions says "Prefer multiline strings to embedding `\n` escape sequences into regular string literals."

## Context
The above rules would identify inconsistencies in use of multiline strings, and were inspired by the current use of multiline strings in detekt's test sources.

I'll also note that KtLint won't address any of the above, as they have said in https://github.com/pinterest/ktlint/pull/1052 that they do not want to change the content of any strings in their rules, so this is not in their scope.

cortinico:
Yup this sounds like a valid rule to add to me 👍

gunishmatta:
@BraisGabin  I want to work on this
Would love to contribute

cortinico:
Assigned to you @gunishmatta 👍

3flex:
Personal preference, but I think these should be individual rules, rather than a single rule with multiple config options.

Anyone else have an opinion on this?

cortinico:
> Personal preference, but I think these should be individual rules, rather than a single rule with multiple config options.
> Anyone else have an opinion on this?

+1 on this. Let's make them separate as it's easier to test/review/maintain

BraisGabin:
@gunishmatta did you get any progress?

BraisGabin:
And I have a question which should be the correct way to indent a multiline string:

```kotlin
"""
    hello!
""".trimIndent()
```
or
```kotlin
"""
hello!
""".trimIndent()
```

If we go with the first one we need a configuration for the indentation size.

marschwar:
[...]

PR #4771, opened 2022-04-25
Allow additionalJavaSourceRootPaths to be defined on @KotlinCoreEnvironmentTest

marschwar:
This relates to #4575 and exposes the `additionalJavaSourceRootPaths`property from `KotlinCoreEnvironmentWrapper.createEnvironment` as `additionalJavaSourcePaths` in the KotlinCoreEnvironmentTest annotation.

There are only very few instances where this feature is needed, so the list of additionalJavaSourcePaths defaults to an empty list.

Changes in the tests are mostly indentation because I introduced two top level nested classes to split into WithDefaultSources and WithAdditionalJavaSources.


BraisGabin (approved):
This PR shows the need of #4705 (and nearly all our tests). We have a lot of multiline strings and I don't even know how should we indent them "properly".

marschwar:
I don't think I understand that comment. I introduced and additional level of nesting and that changes the indentation. I do not see this as evidence that we need the multi-line string rule - even though it might still be a good rule to have.

BraisGabin:
> I don't think I understand that comment. I introduced and additional level of nesting and that changes the indentation. I do not see this as evidence that we need the multi-line string rule - even though it might still be a good rule to have.

Captura de pantalla 2022-04-30 a las 13 02 48

You are adding indentation to your code, but if you see that code inside the multiline didn't change its indentation. The code inside the multiline string has less indentation that the opening multiline operator `"""`.

marschwar:
Ok, I get it now. Is there a correct way to indent multi line strings?

BraisGabin:
I assume that there should be. But I don't know which it is. To me, the important part is to be consistent.

schalkms:
@marschwar @BraisGabin any plans to get this ahead?

marschwar:
From my point of view this is ready.

BraisGabin:
More than 1 month with an approve and no open comments seems fair to merge to me so...

PR #4772, opened 2022-04-25
Silence Vercel commits/PR comments

cortinico:
Turns out that Vercel is pretty noisy. I'm turning the commenting off (on commits & PRs). We will still be able to see the website preview in the "Environments" on a opened PR.

Also if you feel we should keep it, I'm also fine :)


BraisGabin (approved):
Thanks!

PR #4775, opened 2022-04-25
Add dependency on slf4j-nop to silence warning

### LINKED ISSUE #4679: SLF4J warning getting printed when running Detekt 1.20.0-RC2
When I run detekt, some warnings about SLF4J are printed:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

cortinico:
Fixes #4679

See https://ge.detekt.dev/s/lotfhvt2gspyq for a run without the SLF4J warning.

### DIFF
--- a/detekt-formatting/build.gradle.kts
+++ b/detekt-formatting/build.gradle.kts
@@ -13,6 +13,7 @@ dependencies {
     implementation(libs.ktlint.rulesetExperimental) {
         exclude(group = "org.jetbrains.kotlin")
     }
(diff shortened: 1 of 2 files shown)


BraisGabin (approved):

cortinico:
And this won't work :/ We might need to use the shadow plugin for the formatting jar.

PR #4784, opened 2022-04-26
Use correct incremental build property type annotations for reports

3flex:
I found some annotations that were applied incorrectly when investigating usage of unused `customReportFiles` property in the Detekt task in the Gradle plugin.

The unused `customReportFiles` property was removed after correcting the annotations.


cortinico (approved):

3flex:
Actually this might be effectively in use because of the annotations on the getter... I have a more complete solution for that which I'll submit later.

PR #4788, opened 2022-04-28
NestedScopeFunctions - Add rule for nested scope functions

### LINKED ISSUE #4780: Rule for method nesting
Hey. I'm wondering if it would be possible to add a rule, to avoid nesting certain methods/functions.

For example:
```
A().apply {
    B().apply {
        C().apply {
            foo()
        }
    }
}
```
This kind of practice is often confusing, since you don't know where `foo()` belongs.

I've considered using `ComplexMethod`, but it seems like you need to pass certain threshold, in order for detekt to fail. I would like to fail instantly in my example.

WildOrangutan:
Fixes #4780

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -137,9 +137,9 @@ complexity:
     active: false
     threshold: 1
     functions:
-      - 'apply'
-      - 'run'
-      - 'with'
+      - 'kotlin.apply'
+      - 'kotlin.run'
+      - 'kotlin.with'
   ReplaceSafeCallChainWithRun:
     active: false
   StringLiteralDuplication:
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -140,6 +140,8 @@ complexity:
       - 'kotlin.apply'
       - 'kotlin.run'
       - 'kotlin.with'
+      - 'kotlin.let'
+      - 'kotlin.also'
   ReplaceSafeCallChainWithRun:
     active: false
   StringLiteralDuplication:
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -133,6 +133,13 @@ complexity:
   NestedBlockDepth:
     active: true
     threshold: 4
+  NestedScopeFunctions:
+    active: false
+    threshold: 1
+    functions:
+      - 'apply'
+      - 'run'
+      - 'with'
   ReplaceSafeCallChainWithRun:
     active: false
   StringLiteralDuplication:
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NestedScopeFunctions.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NestedScopeFunctions.kt
@@ -1,5 +1,6 @@
 package io.gitlab.arturbosch.detekt.rules.complexity
 
+import io.github.detekt.tooling.api.FunctionMatcher
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Debt
 import io.gitlab.arturbosch.detekt.api.DetektVisitor
@@ -11,9 +12,11 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.ThresholdedCodeSmell
 import io.gitlab.arturbosch.detekt.api.config
 import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
+import org.jetbrains.kotlin.descriptors.CallableDescriptor
 import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtNamedFunction
-import org.jetbrains.kotlin.psi.psiUtil.getCallNameExpression
+import org.jetbrains.kotlin.resolve.calls.util.getResolvedCall
 
 /**
  * Although the scope functions are a way of making the code more concise, avoid overusing them: it can decrease
@@ -40,6 +43,7 @@ import org.jetbrains.kotlin.psi.psiUtil.getCallNameExpression
  * second.c = first.b
  * </compliant>
  */
+@RequiresTypeResolution
 class NestedScopeFunctions(config: Config = Config.empty) : Rule(config) {
 
     override val issue = Issue(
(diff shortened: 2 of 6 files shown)


detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NestedScopeFunctions.kt
```diff
@@ -0,0 +1,95 @@
...
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.psiUtil.getCallNameExpression
+
+/**
+ * Although the scope functions are a way of making the code more concise, avoid overusing them: it can decrease
+ * your code readability and lead to errors. Avoid nesting scope functions and be careful when chaining them:
+ * it's easy to get confused about the current context object and the value of this or it.
+ *
+ * [Reference](https://kotlinlang.org/docs/scope-functions.html)
```
cortinico, line 23: Please add `<compliant>` and <noncompliant>` blocks as the other rules

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NestedScopeFunctions.kt
```diff
@@ -0,0 +1,95 @@
...
+ */
+class NestedScopeFunctions(config: Config = Config.empty) : Rule(config) {
```
cortinico, line 25: nit: I'm not a super big fan of this rule explicitly mentioning `ScopeFunctions` while in reality you can customize the functions. So theoretically you could use this rule to check the nesting of your own non-scope functions.

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -133,6 +133,13 @@ complexity:
   NestedBlockDepth:
     active: true
     threshold: 4
+  NestedScopeFunctions:
+    active: false
+    threshold: 1
+    functions:
+      - 'apply'
+      - 'run'
+      - 'with'
```
BraisGabin, line 142: I would add `also` and `let` too.

BraisGabin (commented):
First of all: Thanks for this new PR! It looks really good.

I'm wondering, should we use here the same function matchers as we use in `ForbiddenMethodCall`?

```kotlin
    private val methods: List<FunctionMatcher> by config(
        listOf(
            "kotlin.io.print",
            "kotlin.io.println",
        )
    ) { it.map(FunctionMatcher::fromFunctionSignature) }
```

So instead of:
```yaml
    functions:
      - 'apply'
      - 'run'
      - 'with'
```

We write:
```yaml
    functions:
      - 'kotlin.apply'
      - 'kotlin.run'
      - 'kotlin.with'
```

This would avoid collisions between functions with the same name.

detekt-core/src/main/resources/default-detekt-config.yml
WildOrangutan: Hm, not sure. I think they are less problematic, since you can rename `it`. I think they maybe better fit to `NestedBlockDepth` or `CognitiveComplexity`?

WildOrangutan:
> I'm wondering, should we use here the same function matchers as we use in `ForbiddenMethodCall`?

Yeah, seems to make sense. Will add.

detekt-core/src/main/resources/default-detekt-config.yml
BraisGabin: For sure this is debatible. The good part is that this is configuratble so anyone could decide if they want to add or to remove some functions.

My 2 cents: If we add `let` and `also` and someone doesn't agree it's easy to remove. If we don't add them and someone want those too added is less likely that he/she add them. For that reason I'm thinking about adding them. But I'm ok about keeping them out. If someone else could share his/her opinion I think it would be great.

WildOrangutan: Maybe smth in the middle - report nesting, when multiple `it` (or same name) are used, since that's problematic from my pov. Maybe in a next MR, so we don't drag this one too much. I can add them for time being.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NestedScopeFunctions.kt
cortinico: Happy to hear what @BraisGabin and others are thinking about this

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NestedScopeFunctions.kt
```diff
@@ -52,8 +56,13 @@ class NestedScopeFunctions(config: Config = Config.empty) : Rule(config) {
...
+    @Configuration(
+        "Set of scope function names which add complexity. " +
+            "Function names have to be fully qualified. For example 'kotlin.apply'."
+    )
+    private val functions: List<FunctionMatcher> by config(DEFAULT_FUNCTIONS) {
+        it.toSet().map(FunctionMatcher::fromFunctionSignature)
+    }
 
     override fun visitNamedFunction(function: KtNamedFunction) {
         function.accept(FunctionDepthVisitor())
```
cortinico, line 68: As you're using Type Resolution now, you should check if bindingContext is empty, and early retunr here

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NestedScopeFunctions.kt
```diff
@@ -65,7 +66,9 @@ class NestedScopeFunctions(config: Config = Config.empty) : Rule(config) {
     }
 
     override fun visitNamedFunction(function: KtNamedFunction) {
-        function.accept(FunctionDepthVisitor())
+        if (bindingContext != BindingContext.EMPTY) {
```
cortinico, line 69:
```suggestion
        if (bindingContext == BindingContext.EMPTY) return
```

Let's keep the same style as all the other rules 👍

BraisGabin (approved):

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NestedScopeFunctions.kt
schalkms: I don't have a better name in mind right now. I'm fine with the current one.

schalkms (approved):

schalkms:
@cortinico do you agree with merging this PR? Do you have further comments for your approval?

Otherwise, I'd say LGTM.

cortinico (approved):
Let's merge it 👍

PR #4793, opened 2022-04-30
Report as warnings the Strings that should be an array

### LINKED ISSUE #4498: Show a warning message when ussing comma separate strings instead of lists
If someone has an old configuration can still be using things like this: 

```yaml
    methodNames: 'toString,hashCode,equals,finalize'
```
instead of 
```yaml
    methodNames: ['toString', 'hashCode', 'equals', 'finalize']
```

An example can be found here: https://github.com/mozilla-mobile/android-components/blob/6f60a8a44b9fa24b62c223130d39bc5ea8095ae6/config/detekt.yml#L129

The support for both types was introduced at 1.8.0. The reason to support both was purely retrocompatibility but keeping this retrocompatibility makes more difficult to maintain our code (Example: https://github.com/detekt/detekt/issues/3501#issuecomment-1015324916)

I think that we should add a warning for the users that still use the old comma separated format so in the future we can remove that retrocompatibility.

BraisGabin:
With this we will show to our users a warning when they are using a comma-separeted String instead of a yaml array. This way, some release later, we can drop the support to the comma-separated String.

Fix #4498

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/ValidateConfig.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/ValidateConfig.kt
@@ -148,6 +148,6 @@ internal fun propertyShouldBeAnArray(
     reportAsError: Boolean,
 ): Notification =
     SimpleNotification(
-        "Property '$prop' should be an array instead of a String.",
+        "Property '$prop' should be a YAML array instead of a comma-separated String.",
         if (reportAsError) Notification.Level.Error else Notification.Level.Warning,
     )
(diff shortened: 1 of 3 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/ValidateConfig.kt
```diff
@@ -140,3 +142,12 @@ internal fun propertyIsDeprecated(
         "Property '$prop' is deprecated. $deprecationDescription.",
         if (reportAsError) Notification.Level.Error else Notification.Level.Warning,
     )
+
+internal fun propertyShouldBeAnArray(
+    prop: String,
+    reportAsError: Boolean,
+): Notification =
+    SimpleNotification(
+        "Property '$prop' should be an array instead of a String.",
```
cortinico, line 151:
```suggestion
        "Property '$prop' should be a YAML array instead of a comma-separated String.",
```

I know this is a bit risky, but it's more clear from the user prespective

cortinico (approved):

Issue #4792, opened 2022-04-30
Add links to the rules documentation in the html report


BraisGabin:
## Expected Behavior
When I read a `html` report I can get the maximum context to fix the issues that detekt found. For example, links to the documentation of each rule that has at least one issue.

## Current Behavior
If you want to know a bit more about one rule and what it does you need to copy the name of the rule, paste it on a web search engine add "detekt" and find that documentation that you need.

## Context
That step of copy&paste to search for a rule is anoying and it should be easily fixed.

This is the current report, as you can see, without any link:
![Captura de pantalla 2022-04-30 a las 12 40 56](https://user-images.githubusercontent.com/721244/166102323-61b0588c-5602-45d3-8fcf-632b28529bfd.png)

cortinico:
Fixed by #4799

PR #4799, opened 2022-05-01
Add documentation link for rules in html report

themkat:
My attempt at #4792.

Adds a documentation link to the HTML reports, so we don't have to find it manually on the Detekt website. Looks like this:
example

Printed once for each rule, at the top before the rule violations are printed. The link in the picture is https://detekt.dev/docs/rules/empty-blocks/#emptyclassblock

### DIFF
--- a/detekt-report-html/src/main/kotlin/io/github/detekt/report/html/HtmlOutputReport.kt
+++ b/detekt-report-html/src/main/kotlin/io/github/detekt/report/html/HtmlOutputReport.kt
@@ -123,7 +123,7 @@ class HtmlOutputReport : OutputReport() {
                 span("description") { text(findings.first().issue.description) }
             }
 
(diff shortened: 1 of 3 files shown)


detekt-report-html/src/main/kotlin/io/github/detekt/report/html/HtmlOutputReport.kt
```diff
@@ -120,6 +121,10 @@ class HtmlOutputReport : OutputReport() {
                 span("description") { text(findings.first().issue.description) }
             }
 
+            a("https://detekt.dev/docs/rules/${group.toLowerCase()}#${rule.toLowerCase()}") {
```
cortinico, line 124: Let's export the `https://detekt.dev/` section to a constant so it's easier to change it should we need to?

chao2zhang (approved):

BraisGabin (approved):
Thank you!

PR #4800, opened 2022-05-02
Fix compiler spec type mismatch

### LINKED ISSUE #4786: JVM 17 & Kotlin 1.6.x Support
The changes made to the final version of 1.20.0 have reverted fixes that addressed the `invalid jvm target` error experienced in this ticket. https://github.com/detekt/detekt/issues/4287

```bash
* What went wrong:
Execution failed for task ':match-api:detekt'.
> Invalid value passed to --jvm-target
```

I'm currently unable to upgrade to resolve the current `XML External Entity (XXE) Injection` vulnerability. Version 1.20.0-RC2 continues to have no issues with JVM 17 and Kotlin 1.6.x, but does not have a resolution for the vulnerability.

rocketraman:
Hopefully fixes #4786 .

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
@@ -65,7 +65,7 @@ internal fun CliArgs.createSpec(output: Appendable, error: Appendable): Processi
         }
 
         compiler {
-            jvmTarget = args.jvmTarget.toString()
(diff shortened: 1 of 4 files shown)


BraisGabin:
This is a breaking change (I'm not sure if we can assume it or not). And this seems related with #4801.

cortinico:
I'll hold off on this one till we find what's the root cause for https://github.com/detekt/detekt/issues/4786

3flex:
Closing as we have resolved the original issue, and have a more robust approach to managing the embeddable compiler dependency going forward, which will hopefully avoid the issue occurring again.

PR #4801, opened 2022-05-03
Add tests for #4786

3flex:
Tests added to show that correctly formatted strings provided on the CLI are parsed as expected, and that the Gradle plugin maps the `jvmTarget` and `languageVersion` properties to those CLI flags as expected.

I've also improved the error message provided by the CLI when these flags cannot be parsed which should help narrow down issues.


BraisGabin (approved):
Shouldn't we test the improved error messages?

3flex:
We don't test the old ones :P

Feel free to do it in a new PR.

PR #4691, opened 2022-04-06
Update gradle enterprise server

chao2zhang:
Because we are officially recognized by Gradle
and offered a gradle enterprise managed instance!

PR #4802, opened 2022-05-03
Upgrade to GE enterprise 3.10

chao2zhang:
Release notes: https://gradle.com/enterprise/releases/2022.2/

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -40,6 +40,7 @@ enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
 
 // build scan plugin can only be applied in settings file
 plugins {
+    // check https://gradle.com/enterprise/releases with new versions. GE plugin version should not lag behind Gradle version
     `gradle-enterprise` version "3.10"
     id("com.gradle.common-custom-user-data-gradle-plugin") version "1.6.5"
 }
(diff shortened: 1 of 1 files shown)


3flex:
I'm not opposed, as long as someone remembers to update it from time to time - Renovate bot doesn't check versions in the settings file. That's why using the default version from Gradle itself made sense, because the wrapper is updated regularly, and Gradle updates the GE plugin version in new Gradle releases.

3flex (approved):

chao2zhang:
I should have given more context: This is an explicit request from the Gradle team (I assume they want to use detekt to stress test their new versions).

chao2zhang:
Linking back to the comment https://github.com/detekt/detekt/pull/4691#discussion_r846171568

3flex:
Where did Gradle Inc. specifically request we don't fall behind? Didn't see it mentioned in that comment you linked. If the request came through other channels then that's OK too.

If it's that important to them then maybe they'll raise PRs when new versions of the plugin are released :)

chao2zhang:
GE 3.10 is coming with Gradle 7.5, which is not released yet. They asked to upgrade the GE plugin version ahead of Gradle. The linked comment was their acknowledgment of the fact that we preferred not to pin the version, but they still asked for pinning the GE version ahead of Gradle.

Issue #4791, opened 2022-04-29
Conflict between ExplicitCollectionElementAccessMethod and UseOrEmpty when using SavedStateHandle in Android


mhernand40:
Please see the following code snippet:
```Kotlin
import androidx.lifecycle.SavedStateHandle

fun repro() {
  val savedStateHandle = SavedStateHandle()
  savedStateHandle["key"] = arrayListOf("a", "b", "c")
  savedStateHandle["key"] ?: emptyArray<String>()
}
```

## Expected Behavior
Detekt should not flag any violations for this because `savedStateHandle["key"]` returns `Any?`. The compiler cannot infer yet that it is an `ArrayList<String>`.

## Observed Behavior
Detekt complains with the following error:
```
This '?: emptyArray<String>()' can be replaced with 'orEmpty()' call [UseOrEmpty]
```
Then when attempting to apply the fix:
```Kotlin
import androidx.lifecycle.SavedStateHandle

fun repro() {
  val savedStateHandle = SavedStateHandle()
  savedStateHandle["key"] = arrayListOf("a", "b", "c")
  savedStateHandle.get<String>("key").orEmpty()
}
```
Detekt complains with the following error:
```
Prefer usage of the indexed access operator [] for map element access or insert methods. [ExplicitCollectionElementAccessMethod]
```

## Context
`SavedStateHandle` is a fairly common component in Android development. And it can store certain `Bundle` friendly types, arrays and array lists being one of them. However, when trying to extract the array or array list from the `Bundle`, these two Detekt rules conflict with one another. The only way I can see around this is to pick either approach from the snippets above and suppress the error.

## Your Environment

* Version of detekt used: 1.20.0
* Version of Gradle used (if applicable):
```
------------------------------------------------------------
Gradle 7.4.2
------------------------------------------------------------

Build time:   2022-03-31 15:25:29 UTC
Revision:     540473b8118064efcc264694cbcaa4b677f61041

Kotlin:       1.5.31
Groovy:       3.0.9
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          11.0.14.1 (Azul Systems, Inc. 11.0.14.1+1-LTS)
OS:           Mac OS X 12.3.1 aarch64
```
* Operating System and version: MacOS Monterey 12.3.1

mhernand40:
I guess there is a workaround:
```Kotlin
import androidx.lifecycle.SavedStateHandle

fun repro() {
[...]

PR #4803, opened 2022-05-03
ExplicitCollectionElementAccessMethod: fix false positive for get operators with type parameters

t-kameyama:
Fix for #4791


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExplicitCollectionElementAccessMethod.kt
```diff
@@ -57,8 +57,11 @@ class ExplicitCollectionElementAccessMethod(config: Config = Config.empty) : Rul
         }
     }
 
-    private fun isIndexableGetter(expression: KtCallExpression): Boolean =
-        expression.calleeExpression?.text == "get" && expression.getFunctionDescriptor()?.isOperator == true
+    private fun isIndexableGetter(expression: KtCallExpression): Boolean {
```
cortinico, line 60: nit: Maybe change the function name here as `IndexableGetter` is not really clear. I'm not able to come up with a better name though

BraisGabin (approved):

schalkms (approved):

PR #4820, opened 2022-05-07
Fix ValShouldBeVar false positive inside unknown type

### LINKED ISSUE #4736: False positive: `VarCouldBeVal`
## Expected Behavior
A property that is reasigned shouldn't be flagged even if it is inside an unknown type.

## Observed Behavior
It is flagged.

## Steps to Reproduce
```kotlin
@Test
fun `does not report private variables that are re-assigned inside an unknown type`() {
    val code = """
        private var a = 1

        fun foo() {
            with(UnknownReference) {
                a = 2
            }
        }
    """

    assertThat(subject.lintWithContext(env, code)).isEmpty()
}
```

## Context
In my code `UnkownReference` is a `ViewBinding` that detekt right now can't handle. Probably #4723 will help

## Your Environment
* Version of detekt used: 1.20.0 & main

themkat:
Attempt at fixing false positives for ValShouldBeVar when the assignment is inside an unknown type. This is the only logical solution I can think of at the moment. It basically assumes that if we can't resolve the type of the assignment, then it is from the inside of an unknown context (like a with-statement). Let me know if there is something I'm missing, and we can close/reject this PR...

Fixes #4736

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
@@ -66,6 +66,24 @@ class VarCouldBeValSpec(val env: KotlinCoreEnvironment) {
 
             assertThat(subject.lintWithContext(env, code)).isEmpty()
         }
+
+        @Test
+        fun `reports private variables that have the same name as those re-assigned within a known type`() {
+            val code = """
+            class MyClass(var a: Int)
+
+            private var a = 1
+            private val myObj = MyClass(1)
+
+            fun foo() {
+                with(myObj) {
+                    a = 2
+                }
+            }
+            """
+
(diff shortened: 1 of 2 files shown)


schalkms (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/VarCouldBeValSpec.kt
```diff
@@ -50,6 +51,39 @@ class VarCouldBeValSpec(val env: KotlinCoreEnvironment) {
...
+            private val myObj = MyClass(1)
+
+            fun foo() {
+                with(myObj) {
+                    a = 2
+                }
+            }
+            """
+
+            assertThat(subject.lintWithContext(env, code)).hasSize(1)
```
BraisGabin, line 85: This case should compile

```suggestion
            assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
```

BraisGabin (approved):

PR #4821, opened 2022-05-07
Kotlin 1.7.10

3flex:
https://blog.jetbrains.com/kotlin/2022/06/kotlin-1-7-0-released/

https://github.com/JetBrains/kotlin/releases/tag/v1.7.0

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 dokka = "1.7.0"
 jacoco = "0.8.8"
-kotlin = "1.6.21"
+kotlin = "1.7.10"
 ktlint = "0.45.2"
 junit = "5.8.2"
 contester = "0.2.0"
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/BindingContext.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/BindingContext.kt
@@ -33,6 +33,7 @@ internal fun generateBindingContext(
     val analyzer = AnalyzerWithCompilerReport(
         messageCollector,
         environment.configuration.languageVersionSettings,
+        false,
     )
     analyzer.analyzeAndReport(files) {
         TopDownAnalyzerFacadeForJVM.analyzeFilesWithJavaIntegration(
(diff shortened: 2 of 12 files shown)


BraisGabin:
Could you explain a bit the changes you did on the tests? I don't untesrtand why those changes are needed.

3flex:
Odd failure with binary-compatibility-validator:
```
> Incompatible version of Kotlin metadata.
  Maximal supported Kotlin metadata version: 1.5.1,
  io/github/detekt/psi/FilePath Kotlin metadata version: 1.7.1.
  As a workaround, it is possible to manually update 'kotlinx-metadata-jvm' version in your project.
```
This is strange because I'd expect a similar error using 1.6.21, as that uses metadata version 1.6.0. So I'm really not sure what the root cause is.

cortinico:
Unless strictly necessary, I would merge this after we release 1.21 so it will go in 1.22. Opinion?

3flex:
> Could you explain a bit the changes you did on the tests?

I've added a reference to each commit in this PR, which links to the "What's new in Kotlin 1.7.0" page, the Kotlin 1.7 compatibility guide, or the individual commit in JetBrains/kotlin as appropriate. Hopefully this provides enough context for each change.

BraisGabin (approved):
Thanks for split the PR and all those links! Really easy to follow.

3flex:
Agree - this should be tracked in a separate issue.

chao2zhang:
Do we need to update `module.gradle.kts` - `kotlinOptions.apiVersion`? The existing value 1.4 is already deprecated since Kotlin 1.6, and I believe we need to bump this further.

chao2zhang (approved):

3flex:
It's set to 1.4 to ensure compatibility with Gradle 6.x which sets the same version.

There's probably a way to bump it up for other modules and just hold it back for the Gradle plugin but I haven't looked into it.

gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 dokka = "1.6.21"
 jacoco = "0.8.8"
-kotlin = "1.6.21"
+kotlin = "1.7.0"
```
schalkms, line 4: Should be 1.7.10 then.

```suggestion
kotlin = "1.7.10"
```

Issue #4830, opened 2022-05-11
Gradle Plugin: Cannot find version resource when another plugin has a version.properties


timothyolt:
## Expected Behavior
When another Gradle plugin that has a `version.properties` resource is added to the same project as Detekt,
Detekt should still be able to find it's version.

## Observed Behavior
Instead, the Gradle configuration crashes with an exception:
```
org.gradle.api.internal.plugins.PluginApplicationException: Failed to apply plugin 'io.gitlab.arturbosch.detekt'.
        at org.gradle.api.internal.plugins.DefaultPluginManager.doApply(DefaultPluginManager.java:173)
        at org.gradle.api.internal.plugins.DefaultPluginManager.apply(DefaultPluginManager.java:146)
        at org.gradle.plugin.use.internal.DefaultPluginRequestApplicator.lambda$applyLegacyPlugin$2(DefaultPluginRequestApplicator.java:160)
        at org.gradle.plugin.use.internal.DefaultPluginRequestApplicator.applyPlugin(DefaultPluginRequestApplicator.java:201)
        ... 155 more
Caused by: org.gradle.api.reflect.ObjectInstantiationException: Could not create an instance of type io.gitlab.arturbosch.detekt.extensions.DetektExtension.
        at org.gradle.internal.instantiation.generator.DependencyInjectingInstantiator.doCreate(DependencyInjectingInstantiator.java:69)
        at org.gradle.internal.instantiation.generator.DependencyInjectingInstantiator.newInstanceWithDisplayName(DependencyInjectingInstantiator.java:50)
        at org.gradle.internal.extensibility.DefaultConvention.instantiate(DefaultConvention.java:216)
        at org.gradle.internal.extensibility.DefaultConvention.create(DefaultConvention.java:125)
        at io.gitlab.arturbosch.detekt.DetektPlugin.apply(DetektPlugin.kt:17)
        at io.gitlab.arturbosch.detekt.DetektPlugin.apply(DetektPlugin.kt:13)
        at org.gradle.api.internal.plugins.ImperativeOnlyPluginTarget.applyImperative(ImperativeOnlyPluginTarget.java:43)
        at org.gradle.api.internal.plugins.RuleBasedPluginTarget.applyImperative(RuleBasedPluginTarget.java:51)
        at org.gradle.api.internal.plugins.DefaultPluginManager.addPlugin(DefaultPluginManager.java:187)
[...]

PR #4831, opened 2022-05-11
Search in all versions.properties, not just the first one #4830

timothyolt:
See details in #4830 about the issue. 

This searches all resources in the Gradle classpath for `versions.properties` instead of using the first one.
It will fail if there are more than one `versions.properties` with different `detektVersion`s.

I considered another approach where the `::class.java.protectionDomain.codeSource?.location` was used to locate the current jarfile, and pick the right resource based on that. But I was less confident that would work in all environments, as it depended on more low-level classpath details.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
@@ -107,9 +107,9 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
     }
 }
 
-internal fun loadDetektVersion(classLoader: ClassLoader): String =
+internal fun loadDetektVersion(classLoader: ClassLoader): String {
     // Other Gradle plugins can also have a versions.properties.
-    classLoader.getResources("versions.properties").toList().mapNotNull { versions ->
+    val distinctVersions = classLoader.getResources("versions.properties").toList().mapNotNull { versions ->
         Properties().run {
             val inputStream = versions.openConnection()
                 /*
@@ -124,4 +124,9 @@ internal fun loadDetektVersion(classLoader: ClassLoader): String =
             load(inputStream)
             getProperty("detektVersion")
         }
-    }.distinct().single()
+    }.distinct()
+    return distinctVersions.singleOrNull() ?: error(
+        "You're importing two Detekt plugins which have different versions. " +
+            "(${distinctVersions.joinToString()}) Make sure to align the versions."
+    )
+}
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
@@ -107,17 +107,21 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
     }
 }
 
-internal fun loadDetektVersion(classLoader: ClassLoader): String = Properties().run {
-    val inputStream = classLoader.getResource("versions.properties")!!.openConnection()
-        /*
-         * Due to https://bugs.openjdk.java.net/browse/JDK-6947916 and https://bugs.openjdk.java.net/browse/JDK-8155607,
-         * it is necessary to disallow caches to maintain stability on JDK 8 and 11 (and possibly more).
-         * Otherwise, simultaneous invocations of Detekt in the same VM can fail spuriously. A similar bug is referenced in
-         * https://github.com/detekt/detekt/issues/3396. The performance regression is likely unnoticeable.
-         * Due to https://github.com/detekt/detekt/issues/4332 it is included for all JDKs.
-         */
-        .apply { useCaches = false }
-        .getInputStream()
-    load(inputStream)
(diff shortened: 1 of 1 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -107,17 +107,21 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
...
+                 * Otherwise, simultaneous invocations of Detekt in the same VM can fail spuriously. A similar bug is referenced in
+                 * https://github.com/detekt/detekt/issues/3396. The performance regression is likely unnoticeable.
+                 * Due to https://github.com/detekt/detekt/issues/4332 it is included for all JDKs.
+                 */
+                .apply { useCaches = false }
+                .getInputStream()
+            load(inputStream)
+            getProperty("detektVersion")
+        }
+    }.distinct().single()
```
cortinico, line 127: Should this be adapted to take the first instead?

timothyolt: Yeah, maybe instead `first` but log when there are more than one?

BraisGabin: I kind of like the idea of single. If there are more than one and they are different the build should crash.

Maybe we should improve the error message but I think that's out of scope here. That could be tracked/done in other issue/PR.

cortinico: > If there are more than one and they are different the build should crash.

Yeah ideally this 👍

schalkms (commented):
Thanks! Looks good 😊

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
cortinico: > Right now it is something like IllegalArgumentException List has more than one element

Yup Ideally I would catch this and `error("You're importing two Detekt plugins which are having different versions and will cause unexpected behavior in your build. Make sure th align the versions")`. You can also print the list of versions if you wish

timothyolt: Sounds great. How's this?

```
Caused by: java.lang.IllegalStateException: You're importing two Detekt plugins which are have different versions. (1.20.1, 1.20.0) Make sure to align the versions.
	at io.gitlab.arturbosch.detekt.extensions.DetektExtensionKt.loadDetektVersion(DetektExtension.kt:130)
	at io.gitlab.arturbosch.detekt.extensions.DetektExtension.<init>(DetektExtension.kt:14)
	at io.gitlab.arturbosch.detekt.extensions.DetektExtension_Decorated.<init>(Unknown Source)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at org.gradle.internal.instantiation.generator.AsmBackedClassGenerator$InvokeConstructorStrategy.newInstance(AsmBackedClassGenerator.java:2078)
	at org.gradle.internal.instantiation.generator.AbstractClassGenerator$GeneratedClassImpl$GeneratedConstructorImpl.newInstance(AbstractClassGenerator.java:488)
	at org.gradle.internal.instantiation.generator.DependencyInjectingInstantiator.doCreate(DependencyInjectingInstantiator.java:64)
	... 197 more
```

```kotlin
internal fun loadDetektVersion(classLoader: ClassLoader): String {
    // Other Gradle plugins can also have a versions.properties.
    val distinctVersions = classLoader.getResources("versions.properties").toList().mapNotNull { versions ->
        Properties().run {
            val inputStream = versions.openConnection()
                /*
                 * Due to https://bugs.openjdk.java.net/browse/JDK-6947916 and https://bugs.openjdk.java.net/browse/JDK-8155607,
                 * it is necessary to disallow caches to maintain stability on JDK 8 and 11 (and possibly more).
                 * Otherwise, simultaneous invocations of Detekt in the same VM can fail spuriously. A similar bug is referenced in
                 * https://github.com/detekt/detekt/issues/3396. The performance regression is likely unnoticeable.
                 * Due to https://github.com/detekt/detekt/issues/4332 it is included for all JDKs.
                 */
                .apply { useCaches = false }
                .getInputStream()
            load(inputStream)
            getProperty("detektVersion")
        }
    }.distinct()
    return distinctVersions.singleOrNull() ?: error(
        "You're importing two Detekt plugins which are have different versions. " +
            "(${distinctVersions.joinToString()}) Make sure to align the versions."
    )
}

```

cortinico: Looks great 👍

cortinico (approved):

PR #4836, opened 2022-05-17
include kotlinx imports in ImportOrdering IDEA list

hbmartin:
IDEA applies special ordering rule (end of import block) for `java`, `javax`, `kotlin`, and `kotlinx` imports. Currently, the ImportOrdering rule is missing `kotlinx` from it's IDEA list, so this PR adds it.

### DIFF
--- a/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/ImportOrdering.kt
+++ b/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/ImportOrdering.kt
@@ -32,6 +32,6 @@ class ImportOrdering(config: Config) : FormattingRule(config) {
 
     companion object {
         const val ASCII_PATTERN = "*"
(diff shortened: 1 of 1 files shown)


hbmartin:
Opened a ktlint PR: https://github.com/pinterest/ktlint/pull/1474

schalkms:
How do we proceed with this PR? @BraisGabin @cortinico 
As seen in the linked issue in the ktlint repo, merging this change would diverge from the official Kotlin conventions for this rule.

BraisGabin:
I never saw my kotlinx imports with a special treatment. I would vote to close this until ktlint decides. We just wrap them so we should just align with their defaults.

cortinico:
> How do we proceed with this PR? @BraisGabin @cortinico

As the KtLint PR hasn't been accepted, I'm leaning towards closing this one.
The configuration is still configurable on user ends so you should be unblocked here.

hbmartin:
Makes complete sense, thanks for the insight on this!

PR #4844, opened 2022-05-19
detektGenerateConfig adds the configuration of plugins

### LINKED ISSUE #4812: Default configuration of custom rulesets isn't merged in the resulting detekt-config.yml
Starting with detekt `1.20.0` we've got a great way to supply a default config from a rule set using `resources/config/config.yml`. While adding this feature to my ruleset I discovered it doesn't seem to work.

By following advice from @BraisGabin I checked this with running `detektGenerateConfig` gradle task: it should append contents of `config.yml` to the resulting output.

But it doesn't do so.

I've attached a sample project where you can run `detektGenerateConfig` and see that `formatting` ruleset doesn't get its config applied to the generated configuration, altough it is explicitly added:

```kotlin
dependencies {
    detektPlugins("io.gitlab.arturbosch.detekt:detekt-formatting:1.20.0")
}
```

## Steps to reproduce

1. Download and import [detekt-ruleset-default-config.zip](https://github.com/detekt/detekt/files/8630561/detekt-ruleset-default-config.zip)
1. Run  `detektGenerateConfig` 
2. Observe that `config/detekt/detekt.yml` doesn't contain `formatting` section

## Your Environment

## Sample project
[detekt-ruleset-default-config.zip](https://github.com/detekt/detekt/files/8630561/detekt-ruleset-default-config.zip)

BraisGabin:
I don't know how to unit-test this. Any guidance here is more than welcome.

The problem was that we were not passing the information about the configured plugins to the cli when we were using executing `detektGenerateConfig`.

Fix #4812

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
@@ -36,6 +36,9 @@ open class DetektGenerateConfigTask @Inject constructor(
     @get:Classpath
     val detektClasspath: ConfigurableFileCollection = project.objects.fileCollection()
 
+    @get:Classpath
+    val pluginClasspath: ConfigurableFileCollection = objects.fileCollection()
+
     @get:InputFiles
     @get:Optional
     @get:PathSensitive(PathSensitivity.RELATIVE)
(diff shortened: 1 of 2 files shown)


cortinico:
> I don't know how to unit-test this. Any guidance here is more than welcome.

For Unit testing, we'll have to restructure a bit the `DetektGenerateConfigTask`. You'll have to use the ProjectBuilder + create this task passing the `pluginClasspath` and exercise the TaskAction.

For E2E testing, it's harder as our test infra doesn't have any support for `detektPlugin` or similar. We'll have to extend what we currently have.

I'm fine with merging this as it is btw as we're sufficiently test covered on this + the other `Detekt` and `DetektCreateBaselineTask` are following the same structure

cortinico (approved):

Issue #4838, opened 2022-05-18
A rule for when multiple collection or iterable operations are chained to suggest that a sequence be used instead


mjsztainbok:
The rule or changes you're looking for might have already been suggested!
Please search in the [issues](https://github.com/detekt/detekt/issues) before creating a new one.

## Expected Behavior of the rule

When operations such as `map`, `filter`, etc. are chained on an `Iterable` or `Collection`, the rule would say that `asSequence()` should be called before the operations to optimize performance.

## Context

Kotlin provides operations like `map`, `filter`, etc. directory on `Collection` and `Iterable` objects through extension methods. The problem with using these is that if you chain the methods e.g. do a `filter` then a `map` multiple iterations are done over the collection/iterable.
However, if a Kotlin sequence is used then only a single iteration is done over the data as the operations are composed and lazily evaluated in a similar way to how Java streams operate
https://kotlinlang.org/docs/sequences.html talks about this

BraisGabin:
That sounds like good rule. Do you want to give it a try? We can help you with it. I assume that `UnnecessaryFilter` could give you a good start point.

themkat:
If @mjsztainbok don't want to, then I sure would love to 🙂  I already started looking a bit into it, but can step aside if @mjsztainbok would prefer looking at it himself

BraisGabin:
I'm using a FIFO here. And you have extra points if you looked a bit into it already ;) Thanks!

BraisGabin:
Closed by #4855. Thanks!

PR #4855, opened 2022-05-23
Add new rule CouldBeSequence

themkat:
My attempt at #4838

Some inspiration taken from `UnnecessaryFilter`, and the `isCalling` method is stolen from there. (why re-invent the wheel?). A little unsure about the documentation parts, so very happy to hear about possible improvements. Would also love to hear if the code could be done in a more functional style that I might have missed.

### DIFF
--- a/detekt-rules-performance/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/CouldBeSequenceSpec.kt
+++ b/detekt-rules-performance/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/CouldBeSequenceSpec.kt
@@ -1,0 +1,54 @@
+package io.gitlab.arturbosch.detekt.rules.performance
+
+import io.gitlab.arturbosch.detekt.rules.KotlinCoreEnvironmentTest
+import io.gitlab.arturbosch.detekt.test.TestConfig
+import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
+import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
+import org.junit.jupiter.api.Test
+
+@KotlinCoreEnvironmentTest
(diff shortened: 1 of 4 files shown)


detekt-rules-performance/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/CouldBeSequenceSpec.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.jupiter.api.Test
+
+@KotlinCoreEnvironmentTest
+class CouldBeSequenceSpec(val env: KotlinCoreEnvironment) {
+    val subject = CouldBeSequence(TestConfig(mapOf("threshold" to 3)))
+
+    @Test
+    fun `long collection chain should be sequence`() {
+        val code = """
+        val myCollection = listOf(1, 2, 3, 4, 5, 6, 7, 8)
```
cortinico, line 17: Commenting here, but valid for the whole rule.

Could you test against:

```
val mySequence : Sequence<Int> = mySequenceMethod()
```

And verify that the rule doesn't trigger?
Specifically I feel you need to resolve the receiver type to verify that it's **not** a sequence. From the code it doesn't look like you're doing it.

cortinico (commented):
Great stuff 🎉  Thanks for sendign this over!

detekt-rules-performance/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/CouldBeSequenceSpec.kt
cortinico: Yup my bad, I haven't realized you were used fully qualified name for method references that will essentially implicitly perform this check 👍 Thanks for clarifying

BraisGabin (approved):

cortinico (approved):

Issue #4826, opened 2022-05-10
Add a Markdown report format (for use with GitHub Job Summaries)


sschuberth:
## Expected Behavior

It would be nice if there was an additional report in Markdown format (maybe even replacing the plain text format) that could be used with GitHub's new [Job Summaries](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/).

## Current Behavior

There is no report in Markdown format available.

## Context

While the SARIF report also works nicely with GitHub PRs, [Job Summaries](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) offer a yet nicer way to get a quick overview of all issues found. The existing plain text format does not make use of Markdown's formatting and linking capabilities, so having a dedicated Markdown report format would be nice.

cortinico:
> (maybe even replacing the plain text format)

Sounds like a good idea, unless we know of any user currently relying on the current text format (which I'm not aware of).

BraisGabin:
:) sure! I assigned it to you.

schalkms:
I would also vote for this suggestion.
The plain text report doesn't provide any additional value, if detekt includes a markdown based report.

> maybe even replacing the plain text format

BraisGabin:
I agree aboit to remove txt but for 2.0

Remove it now could break some setups.

cortinico:
> could break some setups.

Which one? If a user is expecting structured data from a report they would be using XML or other format no?

BraisGabin:
No idea. It just feels like a breaking change that doesn't give us too much, but maybe I'm losing something here.

cortinico:
I feel that we don't have a strong API around the text format, given that it's fully unstructured. Adding some markdown won't hurt IMHO. If a user was relying on the text output to feed it into another tool, they should have used a more structured format maybe?

VitalyVPinchuk:
Hello!
If I get it right, Markdown report should return string similar to this.
If no, please, correct me.

Structure:

---

1. **TestSmell-1**

    1. <details>
       <summary>CodeSmell-1</summary>

       - <details>
           <summary>Issue</summary>
         
           - id='TestSmell'
           - severity=CodeSmell
           - debt=5min
         </details>
         
       - <details>
           <summary>Entity</summary>

           - name=TestEntity
           - signature=TestEntitySignature
           - <details>
               <summary>location</summary>
          
               - source=1:1
               - text=0:0
               - file=TestFile.kt
               - <details>
[...]

PR #4858, opened 2022-05-24
Add Markdown report

VitalyVPinchuk:
For #4826 (Add a Markdown report format)

The report is similar to HTML report and looks like this:

# detekt
## Metrics
   - 10,000 M1
   - 2 M2
## Complexity Report
   - 2,222 lines of code (loc)
   - 20 source lines of code (sloc)
   - 10 logical lines of code (lloc)
   - 2 comment lines of code (cloc)
   - 10 cyclomatic complexity (mcc)
   - 10 cognitive complexity
   - 3 number of total code smells
   - 10% comment source ratio
   - 1,000 mcc per 1,000 lloc
   - 300 code smells per 1,000 lloc
### Section-1, id_a (2)
Description id_a
[Documentation](https://detekt.dev/docs/rules/section-1#id_a)
- src/main/com/sample/Sample1.kt:9:17
  ```
  Message finding 1
  ```
  ```kotlin
  6      val greeting: String = "Hello, World!"
  7  
  8      init {
  9          println(greetings)
  !                  ^ error
  10     }
  11     
  12     fun foo() {
  
  ```
- src/main/com/sample/Sample2.kt:13:17
  ```
  Message finding 2
  ```
  ```kotlin
  10     }
  11     
  12     fun foo() {
  13         println(greetings)
  !!                 ^ error
  14         return this
  15     }
  16 }
  
  ```
### Section-2, id_b (1)
Description id_b
[Documentation](https://detekt.dev/docs/rules/section-2#id_b)
- src/main/com/sample/Sample3.kt:14:16
  ```kotlin
  11     
  12     fun foo() {
  13         println(greetings)
  14         return this
  !!                ^ error
  15     }
  16 }
  
  ```
 
generated with [detekt version 1.20.0](https://detekt.dev/) on 2022-05-25 14:25:27 UTC


detekt-report-md/src/main/kotlin/io/github/detekt/report/md/MdOutputReport.kt
```diff
@@ -0,0 +1,158 @@
...
+import kotlin.math.max
+import kotlin.math.min
+
+private const val DETEKT_WEBSITE_BASE_URL = "https://detekt.dev"
+
+private const val EXTRA_LINES_IN_SNIPPET = 3
+
+/**
+ * Contains rule violations in Markdown format report.
+ * [See:](https://detekt.dev/docs/introduction/configurations/#output-reports)
```
cortinico, line 24:
```suggestion
 * [See](https://detekt.dev/docs/introduction/configurations/#output-reports)
```

detekt-report-md/src/main/kotlin/io/github/detekt/report/md/MdUtils.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.jetbrains.kotlin.backend.common.pop
+import org.jetbrains.kotlin.backend.common.push
+import kotlin.math.min
+
+private const val MAX_INDENT = 3
+
+typealias Block = () -> Unit
+
+@Suppress("TooManyFunctions")
+class MdUtils {
```
cortinico, line 13: I'm wondering if we can reuse some other markdown library instead of having to maintaint this?

I see that there is https://github.com/JetBrains/markdown but I'm not sure if they have generators also

BraisGabin:
I like it! I have a suggestion:

```md
# detekt
## Metrics
...
## Complexity
...
## Findings ($total)
### Rule1 ($total)
- Finding1.1
- Finding1.2
### Rule2 ($total)
- Finding2.1
- Finding2.2
```

What do you thinkg? Basically the change is to use a header for each rule instead of adding it in a list.

detekt-report-md/src/main/kotlin/io/github/detekt/report/md/MdUtils.kt
VitalyVPinchuk: As far as I understood, JetBrains/markdown doesn't offer a possibility to generate Markdown structures, it only parses them to HTML.

Probably, someone knows a well maintained library

So far I've found these generators:
https://github.com/Steppschuh/Java-Markdown-Generator
https://github.com/ap0llo/markdown-generator

VitalyVPinchuk:
> What do you thinkg? Basically the change is to use a header for each rule instead of adding it in a list.

Why not! It would be more compact, I'll fix that.

detekt-report-md/src/main/kotlin/io/github/detekt/report/md/MdUtils.kt
schalkms: Plus, the detekt-generator package already contains a markdown report implementation. Ideally, we shouldn’t reinvent the wheel.

VitalyVPinchuk: Thanks for letting me know, I'll rework it

settings.gradle.kts
```diff
@@ -17,6 +17,7 @@ include("detekt-metrics")
 include("detekt-parser")
 include("detekt-psi-utils")
 include("detekt-report-html")
+include("detekt-report-md")
 include("detekt-report-sarif")
 include("detekt-report-txt")
```
schalkms, line 22: I'd remove this, but would like to get consensus from the other maintainers before doing this. In my opinion, the text report doesn't provide value anymore when introducing the md report.

detekt-report-md/build.gradle.kts
```diff
@@ -0,0 +1,15 @@
+plugins {
+    id("module")
+}
+
+dependencies {
+    implementation(projects.detektMetrics)
+    implementation(projects.detektApi)
+    implementation(projects.detektUtils)
+    implementation(projects.detektGenerator)
```
schalkms, line 9: `detektGenerator` is an internal package for generating config files and the `detekt.yml` file.
I think the `detekt-report-md` should have no dependency to the `detektGenerator`. Ideally, we can move the markdown classes to a shared package.

VitalyVPinchuk: @schalkms that makes sense
Should I pull out `markdown` and `yaml` classes from `detektGenerator` and create shared package?

schalkms: @VitalyVPinchuk the existing **detekt-tooling** package is a good place for the markdown parsing logic. I prefer that over creating a whole new package.

VitalyVPinchuk: Did a bit of refactoring.
Was it the way it meant to be?
Also, I updated `detekt-tooling.api` with gradle recommended command (not sure if this was necessary)

detekt-generator/build.gradle.kts
```diff
@@ -10,10 +10,12 @@ dependencies {
     implementation(projects.detektRulesEmpty)
     implementation(projects.detektFormatting)
     implementation(projects.detektCli)
+    implementation(projects.detektTooling)
     implementation(libs.jcommander)
 
     testImplementation(projects.detektCore)
     testImplementation(projects.detektTestUtils)
+    testImplementation(projects.detektTooling)
```
BraisGabin, line 18: nit: I think that we don't need this dependency.
```suggestion
```

settings.gradle.kts
BraisGabin: I'm ok removing it. I'm also ok keeping it until 2.0 and removing it there. But We can decide that in other PR.

detekt-report-md/build.gradle.kts
BraisGabin: > the existing **detekt-tooling** package is a good place for the markdown parsing logic. I prefer that over creating a whole new package.

The problem of that module is that it is part of our "public API". And I'm not sure if we want to make that API "public".

It would be nice to find a Kotlin DSL to write markdown. I google it but I only found one that is uploaded to jitpack :(.

BraisGabin (approved):

detekt-report-md/build.gradle.kts
schalkms: Hmm, can we use another shared package instead of creating a new one? @BraisGabin

BraisGabin: `detekt-utils`? Maybe that's the best option at least as a "temporal place". To avoid to increase our public API.

schalkms: Agreed @BraisGabin

VitalyVPinchuk: @BraisGabin @schalkms
Fixed

schalkms: Thanks, this PR will be merged when @cortinico approves as well.

BraisGabin:
It seems that it is failing on windows.

VitalyVPinchuk:
> It seems that it is failing on windows.

Yep. But I really don't know why, there's no OS specific code in the PR.
Could it be just Gradle failure?
Sometimes on my mac I have to issue `./gradlew --stop`

cortinico (approved):

PR #4864, opened 2022-05-26
UnnecessaryInnerClass: add test for safe qualified expressions

dzirbel:
Add a unit test for a bug in which safe qualified references to a nullable field
in the parent class (i.e. as `?.`) would not be marked as warnings in release
1.20.0. This issue has been inadvertently fixed by
https://github.com/detekt/detekt/pull/4738, but adding a unit test will guard
against future regressions.

I was able to confirm this test fails by checking out the 1.20.0 tag and running
the unit test; against that tag the fix was to override the
visitSafeQualifiedExpression() function with the same logic as
visitCallExpression().


schalkms (approved):

dzirbel:
Hm, looks like some a checks on windows failed due to gradle running out of memory; is there any easy way to re-run the check?

PR #4865, opened 2022-05-26
UnnecessaryInnerClass: fix false positives labeled expression to outer class

dzirbel:
Fix an issue where inner classes which reference their outer class via a labeled
expression would still be reported as unnecessary. This seems to be due to a
missing visitExpressionWithLabel() check which resolves the label name: if the
label is referencing a parent class, then the nested class must be `inner`.

I wasn't able to find a way to resolve the labeled expression as a ClassId to
match the existing class check from KtReferenceExpression, so I settled for
using the class name instead, which is easy to obtain. This means there's a fair
bit of duplication in overriden checkForOuterUsage() methods to check parent
classes which could probably be removed, but I didn't see a particularly clean
way to do so.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryInnerClass.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryInnerClass.kt
@@ -13,6 +13,7 @@ import org.jetbrains.kotlin.backend.common.pop
 import org.jetbrains.kotlin.descriptors.ClassifierDescriptor
 import org.jetbrains.kotlin.name.ClassId
 import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtExpressionWithLabel
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.KtReferenceExpression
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryInnerClass.kt
```diff
@@ -111,8 +117,24 @@ class UnnecessaryInnerClass(config: Config = Config.empty) : Rule(config) {
...
+    private fun checkForOuterUsage(expressionToResolve: KtExpressionWithLabel) {
+        val currentClass = classChain.peek() ?: return
+        val parentClasses = candidateClassToParentClasses[currentClass] ?: return
+
+        val resolvedContainingClassName = expressionToResolve.getLabelName()
+        /*
+         * If class A -> inner class B -> inner class C, and class C has outer usage of A,
+         * then both B and C should stay as inner classes.
+         */
+        val index = parentClasses.indexOfFirst { it.name == resolvedContainingClassName }
```
schalkms, line 129: I think a test case with 2 inner classes in this scenario is beneficial.

dzirbel: Agreed; added

schalkms (approved):

PR #4870, opened 2022-05-28
CanBeNonNullable: fix false positives for parameterized types

dzirbel:
CanBeNonNullable reports that properties of parameterized types can be marked as non-nullable even when this is impossible, just because the parameterized type is itself nullable (i.e. does not inherit from a non-nullable type). Instead, the rule should only report cases where a property is explicitly marked nullable but does not need to be.

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
@@ -370,6 +370,79 @@ class CanBeNonNullableSpec(val env: KotlinCoreEnvironment) {
             """
             assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
         }
+
+        @Test
+        fun `does not report on non-null properties of a parameterized type`() {
+            val code = """
+                class P<T>(foo: T) {
+                    val a: T = foo
+                    val b: T? = null
+                }
+            """.trimIndent()
+
+            assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
+        }
+
+        @Test
+        fun `reports on properties of an unnecessarily nullable parameterized type`() {
+            val code = """
+                class P<T>(foo: T) {
+                    val bar: T? = foo
+                }
+            """.trimIndent()
+
+            assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
+        }
+
+        @Test
+        fun `reports on properties of an unnecessarily nullable parameterized type extending Any`() {
+            val code = """
+                class P<T : Any>(foo: T) {
+                    val bar: T? = foo
+                }
+            """.trimIndent()
+
+            assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
+        }
+
+        @Test
+        fun `does not report on properties of a parameterized type which must be nullable`() {
+            val code = """
+                class P<T : Any>(private val foo: T) {
+                    val a: T
+                        get() = foo
+                    val b: T? = foo.takeIf { Random.nextBoolean() }
+                    val c: T?
+                        get() = foo.takeIf { Random.nextBoolean() }
+                }
+            """.trimIndent()
+
+            assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
+        }
+
+        @Test
+        fun `does not report on properties of a parameterized type which resolve to Nothing`() {
+            val code = """
+                class P<T> {
+                    val foo: T
+                        get() = error("")
+                }
+            """.trimIndent()
+
+            assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
+        }
+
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
```diff
@@ -370,6 +370,56 @@ class CanBeNonNullableSpec(val env: KotlinCoreEnvironment) {
...
+
+        @Test
+        fun `does not report on properties of a parameterized type which must be nullable`() {
+            val code = """
+                class P<T : Any>(private val foo: T) {
+                    val a: T
+                        get() = foo
+                    val b: T? = foo.takeIf { Random.nextBoolean() }
+                    val c: T?
+                        get() = foo.takeIf { Random.nextBoolean() }
```
BraisGabin, line 405: I miss a test where `T : Any` and this rule find some issues.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
```diff
@@ -370,6 +370,56 @@ class CanBeNonNullableSpec(val env: KotlinCoreEnvironment) {
...
+
+            assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
+        }
+
+        @Test
+        fun `does not report on properties of a parameterized type which resolve to Nothing`() {
+            val code = """
+                class P<T> {
+                    val foo: T
+                        get() = error("")
```
BraisGabin, line 417: 🤔 Should we have a test to check what happen if Foo is declared as `T?`?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
dzirbel: Good point; added

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
dzirbel: Also a good idea; added

dzirbel:
Hm, I forced pushed two additional tests and codecov is now complaining it's under the coverage target? 🤔 Did I do something wrong? I'm not as familiar with PRs in GitHub as other tools

BraisGabin:
> Did I do something wrong?

Yet another caveat on our CI configuration... We usualy don't pay attention to that (unless the number is near to 0). Probably we should do something about it to avoid this type of confusions.

@schalkms What do you think we could do here?

schalkms:
I guess you @BraisGabin and @dzirbel mean the two codecov CI runs.
Codecov provides detekt with code coverage reports. One report yields the code coverage for the whole project. The other report called patch just yields the code coverage for the additions and changed lines of code. The patch report fails when the code coverage is below the 80 % target. 80 % constitutes  a sensible target for well tested code. However, this report is not 100 % accurate for every PR. Usually, I pay attention to this code coverage report during code reviews.
In this case, it's fine to merge with 75 %, I guess. Nevertheless, if you @dzirbel have an idea on how to test a currently uncovered path, we'd ask you to add tests for it.

> Yet another caveat on our CI configuration... We usualy don't pay attention to that (unless the number is near to 0). Probably we should do something about it to avoid this type of confusions.

Maybe, we could spend one or two sentences in the `Contributing.md` file. I don't know whether this addition helps. What do you think?

> @schalkms What do you think we could do here?

dzirbel:
> I guess you @BraisGabin and @dzirbel mean the two codecov CI runs. Codecov provides detekt with code coverage reports. One report yields the code coverage for the whole project. The other report called patch just yields the code coverage for the additions and changed lines of code. The patch report fails when the code coverage is below the 80 % target. 80 % constitutes a sensible target for well tested code. However, this report is not 100 % accurate for every PR. Usually, I pay attention to this code coverage report during code reviews. In this case, it's fine to merge with 75 %, I guess. Nevertheless, if you @dzirbel have an idea on how to test a currently uncovered path, we'd ask you to add tests for it.

Right, if the report can be inaccurate that all makes sense. In this case, coverage was above 80% initially (I believe, since all the checks had passed IIRC) and I [force pushed](https://github.com/detekt/detekt/compare/b852f6559e558fba88c0e51bb858c6c09c917567..0f1f1ac484d54d6051b2e06e11d0875ad8b55d22) two additional tests, after which coverage apparently dropped to 75%, so I wanted to double check that my PR method was correct (maybe I'd messed up rebases or something).

Thanks both of you for the quick and helpful reviews on my PRs this weekend :)

schalkms:
You are welcome 🤗

PR #4872, opened 2022-05-28
Add NullableBooleanCheck rule

dzirbel:
Add a new rule which enforces the recommendation of the Kotlin coding conventions to prefer `==` over `?:` when converting a `Boolean?` into a `Boolean`: https://kotlinlang.org/docs/coding-conventions.html#nullable-boolean-values-in-conditions. This is simple to implement and enforces consistency for the two essentially equivalent methods of coalescing nullable boolean values.

The coding conventions only specify this preference for usages "in a conditional statement" but the same rationale applies to any statement which converts a Boolean? to a Boolean, so I have implemented it for any situation. An alternative is to add a configuration setting to toggle application only in if statements.

Unfortunately, this rule requires type resolution since there are technically valid usages of e.g. `value ?: true` for cases where `value` is not a `Boolean?` but an `Any?`. Running without type resolution will not be able to distinguish between these valid usages and ones which could be converted to `== false`.


schalkms (approved):

schalkms:
The code snippets used in tests are compiled with Jsr223 in this repo to ensure that those code snippets are valid Kotlin code. Some of the code snippets in **NullableBooleanCheckSpec** seem to be invalid according to the CI run for this PR. Please make sure to correct the code.
Otherwise, the code looks very good and is ready to merge!

Thank you for this nice addition resulting from an official Kotlin convention!

dzirbel:
> The code snippets used in tests are compiled with Jsr223 in this repo to ensure that those code snippets are valid Kotlin code. Some of the code snippets in **NullableBooleanCheckSpec** seem to be invalid according to the CI run for this PR. Please make sure to correct the code. Otherwise, the code looks very good and is ready to merge!

Should be all set now, thanks! `compile-snippet-tests` took me by surprise, but it's a great setting since I've previously written tests whose code did not compile and so they had unexpected behavior. I'm curious why it's not enabled by default? I'd find it useful on most/all tests when running locally, and if some CI jobs are disabling it for performance that could still be done via `-Pcompile-test-snippets=false`. The closer running `gradle build` is to ensuring CI checks pass the easier contribution is, in my experience :)

schalkms:
That’s because of the increased time for test runs. If you run all the tests locally with the **compile-snippet-tests** setting enabled, it takes quite some time. Hence, we disabled it by default.

BraisGabin:
I think that keep this disabled by default on local is good. But it's true that a lot of new contributors are "bitten" byt this issue. I'm going to open a new issue to look for a way to fix it.

BraisGabin:
We can continue this discussion here: #4874. Any feedback is more than welcome.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/NullableBooleanCheck.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.jetbrains.kotlin.psi.KtBinaryExpression
+import org.jetbrains.kotlin.resolve.calls.util.getType
+import org.jetbrains.kotlin.types.typeUtil.isBooleanOrNullableBoolean
+
+/**
+ * Detects nullable boolean checks which use an elvis expression `?:` rather than equals `==`.
+ *
+ * Per the [Kotlin coding conventions](https://kotlinlang.org/docs/coding-conventions.html#nullable-boolean-values-in-conditions)
+ * converting a nullable boolean property to non-null should be done via `== true` or `== false`
+ * rather than `?: false` or `?: true` (respectively).
```
BraisGabin, line 23:
```suggestion
 * converting a nullable boolean property to non-null should be done via `== true` or `!= false`
 * rather than `?: false` or `?: true` (respectively).
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/NullableBooleanCheck.kt
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * <noncompliant>
+ * value ?: true
+ * value ?: false
+ * </noncompliant>
+ *
+ * <compliant>
+ * value == false
+ * value == true
```
BraisGabin, line 32:
```suggestion
 * value != false
 * value == true
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/NullableBooleanCheck.kt
```diff
@@ -0,0 +1,62 @@
...
+    override fun visitBinaryExpression(expression: KtBinaryExpression) {
+        if (expression.operationToken == KtTokens.ELVIS &&
+            expression.right?.isBooleanConstant() == true &&
+            expression.left?.getType(bindingContext)?.isBooleanOrNullableBoolean() == true
+        ) {
+            report(
+                CodeSmell(
+                    issue,
+                    Entity.from(expression),
+                    "The nullable boolean check `${expression.text}` should use `==` rather than `?:`",
```
BraisGabin, line 53: nit: I think that we could improve the error message making it more actionable. Something like:

- "This nullable boolean check should use `!= false` instead of `?: true`"
- "This nullable boolean check should use `== true` instead of `?: false`"

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/NullableBooleanCheck.kt
dzirbel: Good catch, thanks! Interestingly the Kotling coding conventions only recommend use of `== true` or `== false`, but I don't see how that can cover the necessary cases 🤔

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/NullableBooleanCheck.kt
dzirbel: Agreed, especially since this is a bit nuanced; added

BraisGabin (approved):

BraisGabin:
Thank you for this rule. It's really nice :)

snyman:
I'm not sure if this is the best place to ask, since the change is merged, but I'm wondering if this rule oversteps the intention of the kotlin style guide. The [kotlin style guideline](https://kotlinlang.org/docs/coding-conventions.html#nullable-boolean-values-in-conditions) linked in the change says

> If you need to use a nullable Boolean in a **conditional statement**, use if (value == true) or if (value == false) checks.

(Emphasis mine.) To me this is fairly clearly talking only about conditional statements such as `if` or `when`, and not about all boolean expressions in general. In particular, when computing a non-null boolean expression that isn't being immediately used inside a conditional, I think the elvis operator (`?:`) is clearer than `==`. For example:

```kotlin
fun nonNullBoolean() = nullableBoolean() ?: false
```
vs
```kotlin
fun nonNullBoolean() = nullableBoolean() == true
```

In the former I can immediately see what the returned value is in the case that `nullableBoolean()` returns `null`. In the latter, I need to do a bit of extra thinking to figure it out.

dzirbel:
These are very good points; the reason why we adopted this style on my team is purely as an anti-bikeshedding measure. I would actually also prefer `?: true` and `?: false` as being more intuitive, but the Kotlin style guide has settled in the other direction for whatever reason. Since it's just a matter of standardization as I see it, I personally like having it enforced everywhere that a nullable boolean is being converted to non-null; I don't see any particular reason why `== true` is more intuitive in a conditional statement than in any other line of code.

BraisGabin:
I agree with both. It's true that the Kotlin style guideline only talks about conditional statments. But I agree with @dzirbel here that use it everywhere makes the code more consistent.

Anyway, if someone is willing to add a flag that, when enabled, only checks inside conditional statments I would be happy to merge it.

artem-zinnatullin:
Nice check, found few violations in our project and `==` is def makes it more clear than `?:`

Thanks for your work @dzirbel!

PR #4875, opened 2022-05-30
Active rules by default

### LINKED ISSUE #4643: Enable rules by default for `1.21`
All the rules that we add to detekt start disable by default. That's a good default because the early-adopters can give us feedback about them and we don't create a lot of noise for those users that prefer more stability.

But from time to time we should promote those rules and enable them by default. An example of rule that we could enable by default:

- [InvalidPackageDeclaration](https://detekt.dev/naming.html#invalidpackagedeclaration)

Which other rules do you think we should enable by default?

Note: There are some rules that are "controversial" and they will never become enable by default.

BraisGabin:
This PR enables the rules as we talk in #4643.

If, to help with the changelog, creating one PR for each rule helps I can create those ~30PRs. I would create a script or somthing to do it. But I prefer to avoid it.

Enabled rules:

- InjectDispatcher
- RedundantSuspendModifier
- SleepInsteadOfDelay
- SuspendFunWithFlowReturnType
- InstanceOfCheckForException
- InvalidPackageDeclaration
- NoNameShadowing
- AvoidReferentialEquality
- DoubleMutabilityForCollection
- HasPlatformType
- IgnoredReturnValue
- MapGetWithNotNullAssertionOperator
- MissingPackageDeclaration
- UnreachableCatchBlock
- UnusedUnaryOperator
- UselessPostfixExpression
- DestructuringDeclarationWithTooManyEntries
- ExplicitItLambdaParameter
- ForbiddenVoid
- ObjectLiteralToLambda
- RedundantHigherOrderMapUsage
- UnnecessaryFilter
- UseAnyOrNoneInsteadOfFind
- UseArrayLiteralsInAnnotations
- UseCheckNotNull
- UseCheckOrError
- UseIsNullOrEmpty
- UseOrEmpty
- UseRequire
- UseRequireNotNull

Closes #4643

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -142,6 +142,11 @@ potential-bugs:
     active: true
   MapGetWithNotNullAssertionOperator:
     active: true
+  MissingPackageDeclaration:
+    active: true
+    excludes:
+      - '**/*.kts'
+      - '**/build-logic/**'
   UnconditionalJumpStatementInLoop:
     active: true
   UnreachableCatchBlock:
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -162,19 +162,19 @@ coroutines:
   GlobalCoroutineUsage:
     active: false
   InjectDispatcher:
-    active: false
+    active: true
     dispatcherNames:
       - 'IO'
       - 'Default'
       - 'Unconfined'
   RedundantSuspendModifier:
-    active: false
+    active: true
   SleepInsteadOfDelay:
-    active: false
+    active: true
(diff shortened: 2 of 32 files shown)


schalkms (approved):

marschwar (approved):
Isn't it beautiful how you just have to add the annotation and the config and documentation are in sync. I :heart: it.

Thanks for doing all the leg work.

cortinico:
I had to disable the `MissingPackageDeclaration` in order for 1.21.0-RC1 to be released.

The problem with this rule is that we're creating files in the default package (`src/main/java`) without a package definition inside our Gradle Plugin Tests. I also think that having files without a package declaration is fine as long as they stay in the correct folder.

Issue #4495, opened 2022-01-17
UnnecessaryParentheses where IDE prompts for clarifying parentheses.


eygraber:
The following code triggers a warning in the IDE, saying `Expression should use clarifying parentheses`:

```
val (adjustedFromItem, fromFragment) =
      validPages
        .firstOrNull { (from, frag) -> from != page && frag.isVisible }
        ?: null to null
```

It wants to correct it to:

```
val (adjustedFromItem, fromFragment) =
      validPages
        .firstOrNull { (from, frag) -> from != page && frag.isVisible }
        ?: (null to null)
```

But detekt flags it as `UnnecessaryParentheses`. I see the value of adding parentheses here, and a few other places that the IDE is suggesting it, but I don't want to have to suppress the detekt check in all of those places.

Can detekt use similar logic to exclude these cases?

BraisGabin:
We should look at the rule that is warning this in IJ and see if we can integrate it in this rule.

cortinico:
> The following code triggers a warning in the IDE

Can we confirm that is a IDE warning and not a compiler warning? A compiler warning would be way easier to integrate if so.

eygraber:
https://youtrack.jetbrains.com/issue/KTIJ-8888

dimsuz:
It seems like this has been fixed by #4881 ?

At least I found the corresponding test case [here](https://github.com/detekt/detekt/blob/d5165629636eb5d37fd10b10761c9363adb808e8/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParenthesesSpec.kt#L296).

BraisGabin:
Yes, these one was fixed. Closing.

eygraber:
@BraisGabin I found two occurrences where the IDE tells me to add clarifying parentheses, but after I do, detekt says it is a `UnnecessaryParentheses` violation. Here they are after adding the clarifying parentheses:

```
string("id") to (floatOrNull("progress") ?: (index.toFloat() / (segments.size - 1)))
```

and

```
progress = segmentJson.floatOrNull("progress") ?: (index.toFloat() / (segmentArray.size - 1)),
```

BraisGabin:
I was thinking that maybe #5684 would fix it. But after checking a bit more it doesn't seem so.

atulgpt:
Hi @BraisGabin / @eygraber I have added the fix at https://github.com/detekt/detekt/pull/5983

Issue #2789, opened 2020-06-12
UnnecessaryParentheses warns about xor-like operations


imanushin:
## Expected Behavior
No detekt warnings is appeared.

## Observed Behavior
The warning is appeared - UnnecessaryParentheses - [XXX] at /Users/zzz/XXX.kt:21:17

## Steps to Reproduce
Code:
```kotlin
data class(val id1: Int?, val id2: Int?) {
   init {
      require((id1 === null && id2 !== null) || (id1 !== null && id2 === null)) {
         "..."       
      }
   }
}
```

## Context
Brackets above are needed for readability, to simulate xor-like behaviour (when one and only one field can have value)

## Your Environment
* Version of detekt used: 1.9.1
* Version of Gradle used (if applicable): 6.5.0
* Operating System and version: n/a
* Link to your project (if it's a public repository): private repo

schalkms:
We won’t invest more time in this rule, since detekt complies with IntelliJ’s rule.
That’s why I’ll close this.

Reference: https://github.com/detekt/detekt/blob/1e712757bb399666e26514740b08e2faa565bf7b/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParentheses.kt#L50

schalkms:
If you prefer readability here, please add a Suppress annotation. We discussed this rule in other issues. It’s just not worth the time to apply more exceptions to this rule. It makes the rule unnecessarily complex.

Issue #1929, opened 2019-09-17
UnnecessaryParentheses should be ignored for bit operations


Mygod:
## Expected Behavior of the rule

Example:

```
        val mask = 256 - (1 shl (i * 8 + 8 - prefixSize))
        return (a[i].toInt() and mask) == (b[i].toInt() and mask)
```

## Context

Bitwise operations in C-like languages have confusing priorities so people usually add parentheses for everything.
This sample code also does not trigger warnings in latest Android Studio (3.6 alpha).

detekt 1.0.1

3flex:
1.1.0 has a change in behaviour for this rule - it should only flag issues that are also flagged by Kotlin plugin for IntelliJ (see #1890).

When 1.1.0 is released please see if the new behaviour helps.

If not it's still unlikely that we'll make changes and instead suggest suppressing the rule in these cases, since there were a few issues with this rule in the past and too many edge cases, so we'll now just use same logic as Kotlin itself does.

schalkms:
This has been fixed by 61f1734.
We won’t invest more time in this rule, since detekt complies with IntelliJ’s rule.

Mygod:
In 1.1.0 this issue is still there. Also IntelliJ/Android Studio does not report this.

3flex:
Sorry I wasn't clearer earlier.

IntelliJ has intentions and inspections. Inspections show warnings or errors, intentions are used to change code but don't show a warning or error. Your example code is picked up by the intention detection in IntelliJ as expected, and detekt does the same:

![image](https://user-images.githubusercontent.com/170028/67002788-5ba47080-f128-11e9-966b-02e2a5b649f0.png)

Mygod:
Oh I see. Thank you!

Issue #4168, opened 2021-10-07
Move to versioned documentation


JpCodeTask:
## Expected Behavior
Config shouldn't complete with error with correct fields.

## Observed Behavior
Config validator completes with error. The errors inform about misspelled or not existed fields, although they are correct according to the documentation. At the same time correctness of fields with config validation turned off has been proved.

## Steps to Reproduce
Add default fields fo rules listed below and turn on config validation.
  LongParameterList:
    excludes: *daggerClasses
    active: true
    functionThreshold: 4
    constructorThreshold: 7
    ignoreDefaultParameters: true
    ignoreDataClasses: true
    ignoreAnnotatedParameter: [ ] <-- default value

  UnderscoresInNumericLiterals:
    active: false
    acceptableLength: 5 <-- default value

## Context
Turning on config validation after detekt plugin version update

## Your Environment

* Version of detekt used: 1.18.1
* Version of Gradle used (if applicable): 6.8.3

cortinico:
This is still a thing. I'm actually looking into it 👍

cortinico:
I'm happy to jump back on this 👍

PR #4881, opened 2022-05-31
UnnecessaryParentheses: add options to allow in ambiguous cases

dzirbel:
One blocker for adopting the UnnecessaryParentheses is that in some common cases parentheses that may not be strictly required by order of operations are still enormously helpful in parsing code, such as for boolean operations like `(x && y) || z`. While it's possible to `@Suppress` each of these, this adds bloat and encourages developers to avoid these (in my opinion) often very helpful indicators of code flow.

Instead, we can add a configuration option to UnnecessaryParentheses to skip over some of the most common cases where unnecessary parentheses can be useful in clarifying the code. This strategy was adopted from IntelliJ's UnclearPrecedenceOfBinaryExpressionInspection inspection: https://github.com/JetBrains/intellij-community/blob/master/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnclearPrecedenceOfBinaryExpressionInspection.kt with a fair amount of trimming and a different set of operators that have unclear precedence (adding numeric and boolean ones).

This idea has been discussed a number of times; most recently in #4495, which suggested that it could integrated into detekt (although perhaps only by porting the functionality from IntelliJ). Previous issues such as #2789 and #1929 concluded that it would be too much of an investment or overcomplicate the rule. I'm happy to stick with that precedent, but as a user who doesn't want to adopt this rule on my team until we can avoid at least some of the more common cases where parentheses can be helpful I thought I'd suggest a concrete implementation with fairly low complexity (although admittedly only addressing some of the lowest hanging fruit). I haven't thought through every edge case or application, so perhaps complexity would explode with a more fleshed-out version.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParentheses.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParentheses.kt
@@ -7,7 +7,17 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.com.intellij.psi.tree.IElementType
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.parsing.KotlinExpressionParsing.Precedence
+import org.jetbrains.kotlin.psi.KtBinaryExpression
+import org.jetbrains.kotlin.psi.KtBinaryExpressionWithTypeRHS
 import org.jetbrains.kotlin.psi.KtDelegatedSuperTypeEntry
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtIsExpression
 import org.jetbrains.kotlin.psi.KtParenthesizedExpression
 import org.jetbrains.kotlin.psi.KtPsiUtil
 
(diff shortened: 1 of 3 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParenthesesSpec.kt
```diff
@@ -184,4 +188,46 @@ class UnnecessaryParenthesesSpec {
...
+
+        @ParameterizedTest
+        @ValueSource(booleans = [true, false])
+        fun `does not report numeric expressions`(allow: Boolean) {
+            val config = TestConfig(mapOf("allowInAmbiguousNumericExpressions" to allow))
+            val code = "val local = (1 * 2) + 3"
+
+            assertThat(UnnecessaryParentheses(config).lint(code)).hasSize(if (allow) 0 else 1)
+        }
+    }
```
BraisGabin, line 232: What would happen with this cases: 

- `val local = (1 * 2)`
- `val local = (1 * 2 + 3)`
- `val local = (a || b)`
- `val local = (a || b && c)`
- `val local = (a || b) || c`

To me all of them should be flagged always.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryParentheses.kt
```diff
@@ -46,17 +51,48 @@ class UnnecessaryParentheses(config: Config = Config.empty) : Rule(config) {
         Debt.FIVE_MINS
     )
 
+    @Configuration("whether to allow for potentially ambiguous boolean operations such as (x && y) || z")
+    private val allowInAmbiguousBooleanExpressions: Boolean by config(defaultValue = false)
+
+    @Configuration("whether to allow for potentially ambiguous numeric operations such as (x * y) + z")
+    private val allowInAmbiguousNumericExpressions: Boolean by config(defaultValue = false)
```
cortinico, line 58: I think naming is (sadly) ambiguous.
It's not immediately clear what's the semantic of `AmbiguousBooleanExpressions` and `AmbiguousNumericExpressions` and we're forcing the user to read the example.

I would try to find a better name if possible.

BraisGabin:
> Sorry for the noise in the meantime.

It's not noise! Thanks for proposing this and create a first draft to give fast feedback.

This is really help us. It doesn't feel good to decline or force someone to change nearly all the code that s/he work for weeks. But we need to do those type of things to keep the project healthy and "managable". So don't worry about sending drafts of ideas we always welcome any help.

cortinico:
> Sorry for the noise in the meantime.

Don't worry about that :) We're more than happy to receive contribs like yours.

dzirbel:
OK, I have a new approach heavily modified from IntelliJ's inspection https://github.com/JetBrains/intellij-community/blob/master/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnclearPrecedenceOfBinaryExpressionInspection.kt which is more generic and should (I believe) cleanly cover cases where parentheses are not strictly required but can be very helpful for clarifying the meaning of the code.

This might even open the door for the inverse warning matching the behavior of IntelliJ's inspection: warning when parentheses could be helpful to clarify precedence but aren't currently present. I don't feel strongly about that idea - probably wouldn't use it myself but perhaps others would.

Is there a standard way of unit testing different rule configurations that I should use here? I wanted to test all the existing cases with the new option applied to make sure it wouldn't change their behavior, but this added a fair bit of boilerplate. Perhaps `@ParameterizedTest` could clean those up but would make it hard for the tests that have different behavior between configurations.

BraisGabin:
I think that to test this properly you could move that code out of the rule. And test the function itself. There you can use a parameterized test to add all the cases you want. Given x code the function should return (true/false). 

Then you integrate that function inside the rule and you don't need to test all the cases there, just some to check that the integration is correct.

dzirbel:
Good ideas - I've refactored a bit and went with a different testing strategy using `@ParameterizedTest` and providing a different instance of the rule while also making it easy to check whether `allowForUnclearPrecedence` is active for this test case. This makes most tests easy and avoids needing the extract the functionality or test the integration indirectly. Happy to change that up if you feel that another way to be preferred though.

BraisGabin (approved):
I like it :)

There are a lot of tests but I think that's good because there are a lot of cases and I think that you are cheking all of them.

Really good job 👍

BraisGabin:
I'm moving this to `1.22.0` because it feels like a big change and we are near to release a stable version but as soon as we release `1.21.0` this will be merged.

TWiStErRob:
This got me, the documentation says the flag exists, but when trying to use it, it fails to execute. Is this normal @BraisGabin that the docs don't match the latest release?

TWiStErRob:
@dzirbel I would like to throw another case at you where clarifying parentheses might help:
```kotlin
private fun fakeClock(fixture: JFixture, minOffset: Long, maxOffset: Long): Clock? {
	val now = System.currentTimeMillis()
	val millis = fixture.buildRange((now - minOffset)..(now + maxOffset))
	val zone = fixture.create().fromList(*ZoneId.getAvailableZoneIds().toTypedArray())
	return Clock.fixed(Instant.ofEpochMilli(millis), ZoneId.of(zone))
}
```
Without parens `fixture.buildRange(now - minOffset..now + maxOffset)` one might expect it's interpreted as `fixture.buildRange(now - (minOffset..now) + maxOffset)`

dzirbel:
> This got me, the documentation says the flag exists, but when trying to use it, it fails to execute. Is this normal @BraisGabin that the docs don't match the latest release?

I believe that is the current behavior, yes - new rules, params, etc appear on the website immediately after they are merged rather than when a new version is released. I've always thought a version dropdown like [Gradle](https://docs.gradle.org/current/userguide/userguide.html) and others use could be helpful.

> @dzirbel I would like to throw another case at you where clarifying parentheses might help:
> 
> ```kotlin
> private fun fakeClock(fixture: JFixture, minOffset: Long, maxOffset: Long): Clock? {
> 	val now = System.currentTimeMillis()
> 	val millis = fixture.buildRange((now - minOffset)..(now + maxOffset))
> 	val zone = fixture.create().fromList(*ZoneId.getAvailableZoneIds().toTypedArray())
> 	return Clock.fixed(Instant.ofEpochMilli(millis), ZoneId.of(zone))
> }
> ```
> 
> Without parens `fixture.buildRange(now - minOffset..now + maxOffset)` one might expect it's interpreted as `fixture.buildRange(now - (minOffset..now) + maxOffset)`

I agree, that's a good case. Opened #5143 to add it.

BraisGabin:
> This got me, the documentation says the flag exists, but when trying to use it, it fails to execute. Is this normal @BraisGabin that the docs don't match the latest release?

Yes, we need a versioned documentation. We know that docusaurus let us handle it but we didn't have the time for it: #4168. Any help there is more than appreciatted.

cortinico:
> > This got me, the documentation says the flag exists, but when trying to use it, it fails to execute. Is this normal @BraisGabin that the docs don't match the latest release?
> 
> Yes, we need a versioned documentation. We know that docusaurus let us handle it but we didn't have the time for it: #4168. Any help there is more than appreciatted.

I can take a stab at #4168

PR #4882, opened 2022-05-31
UnusedImports: fix false positive for unresolved imports

dzirbel:
Fix a false positive where UnusedImports would report extension function imports whose signature cannot be resolved by the BindingContext. This can happen for autogenerated classes, etc. (especially Android UI binding classes and the like).

To do this we need to generate text matchers for all the references which cannot be resolved and check if there are any unresolved matches to them, similar to the approach without type resolution.

Along the way, extract mappings inside `KtImportDirective.isNotUsed()` to lazy properties to avoid recalculating them for each import directive. Also make sure to map to a set for constant rather than linear lookup time.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
@@ -54,26 +54,44 @@ class UnusedImports(config: Config) : Rule(config) {
         private val staticReferences = mutableSetOf<KtReferenceExpression>()
         private val namedReferencesInKDoc = mutableSetOf<String>()
 
+        private val namedReferencesAsString: Set<String> by lazy {
+            namedReferences.mapTo(mutableSetOf()) { it.text.trim('`') }
+        }
+        private val staticReferencesAsString: Set<String> by lazy {
+            staticReferences.mapTo(mutableSetOf()) { it.text.trim('`') }
+        }
+        private val fqNames: Set<FqName> by lazy {
+            namedReferences.mapNotNullTo(mutableSetOf()) { it.fqNameOrNull() }
+        }
+
+        /**
+         * All [namedReferences] whose [KtReferenceExpression.fqNameOrNull] cannot be resolved
+         * mapped to their text. String matches to such references shouldn't be marked as unused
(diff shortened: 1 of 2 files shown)


BraisGabin:
I'm wondering how ktlint does this. I have this rule disabled from detekt because I use ktlint for this issue and I never had any false positive or false negative.

dzirbel:
I assume that ktlint does string usage matching, the same as detekt without type resolution (or this change for when type resolution is enabled but can't resolve an import). The [source](https://github.com/pinterest/ktlint/blob/master/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoUnusedImportsRule.kt#L115) seems to suggest the same - reference identifier text minus backticks is added to `ref` which is eventually compared via `ref.map { it.text }.contains(name)` for imports.

BraisGabin (approved):

BraisGabin:
We should decide what to do with this. My point of view: this is a workaround for other bug: We don't get a complete `ContextBinding` so we have this type of false positives. But that is a long standing bug (it's there since we introduced Type solving) and we don't have a clear plan to fix it in the near future.

So, to me, we should merge this, a false-positive is way worst than a false-negative. And we should also focus on fixing that long stading bug to avoid this type of workarounds.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImports.kt
```diff
@@ -54,26 +54,44 @@ class UnusedImports(config: Config) : Rule(config) {
...
+            staticReferences.mapTo(mutableSetOf()) { it.text.trim('`') }
+        }
+        private val fqNames: Set<FqName> by lazy {
+            namedReferences.mapNotNullTo(mutableSetOf()) { it.fqNameOrNull() }
+        }
+
+        /**
+         * All [namedReferences] whose [KtReferenceExpression.fqNameOrNull] cannot be resolved
+         * mapped to their text. String matches to such references shouldn't be marked as unused
+         * imports since they could match the unknown value being imported.
```
chao2zhang, line 70: 💯  documentation.

chao2zhang (approved):
I believe this issue will be resolved once we are using K2 compiler plugin.

PR #4899, opened 2022-06-02
Adds ForbiddenSuppress rule

gfreivasc:
This rule is based on discussion #4890. It has been added to the "style"
rule set, since I believe it is opinionated and usage would vary across code
bases.

Checks for suppression of one or more rules that have been decided to
never be suppressed within a code base. For example, if we've set
MaximumLineLength to a value, every file must be compliant, and if there
is a special case where compliance is discouraged, we can simply exclude
this file within YML configuration.

This rule cannot prevent itself from being suppressed, but should serve
to discourage the over use of `@Suppress` annotations.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSuppress.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSuppress.kt
@@ -1,0 +1,90 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import org.jetbrains.kotlin.psi.KtAnnotationEntry
+import org.jetbrains.kotlin.psi.KtLiteralStringTemplateEntry
+import org.jetbrains.kotlin.psi.KtValueArgument
+import org.jetbrains.kotlin.psi.KtValueArgumentList
+import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
+
+/**
+ * This rule allows to set a list of rules whose suppression is forbidden.
+ * This can be used to discourage the abuse of the `Suppress` and `SuppressWarnings` annotations.
+ * Detekt will report suppression of all forbidden rules.
+ * This rule is not capable of reporting suppression of itself, as that's a language feature with precedence.
+ *
+ * <noncompliant>
+ * package foo
+ *
+ * // When the rule "MaximumLineLength" is forbidden
+ * @@Suppress("MaximumLineLength", "UNCHECKED_CAST")
+ * class Bar
+ * </noncompliant>
+ *
+ * <compliant>
+ * package foo
+ *
+ * // When the rule "MaximumLineLength" is forbidden
+ * @@Suppress("UNCHECKED_CAST")
+ * class Bar
+ * </compliant>
+ */
+class ForbiddenSuppress(config: Config = Config.empty) : Rule(config) {
+
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Maintainability,
+        "Suppressing a rule which is forbidden in current configuration.",
+        Debt.TEN_MINS
+    )
+
+    @Configuration("Rules whose suppression is forbidden.")
+    private val rules: List<String> by config(emptyList())
+
+    override fun visitAnnotationEntry(annotationEntry: KtAnnotationEntry) {
+        if (rules.isEmpty()) return
+        val shortName = annotationEntry.shortName?.asString()
+        if (shortName == KOTLIN_SUPPRESS || shortName == JAVA_SUPPRESS) {
+            val nonCompliantRules = annotationEntry.children
+                .find { it is KtValueArgumentList }
+                ?.children
+                ?.filterIsInstance<KtValueArgument>()
+                ?.filterForbiddenRules()
+                .orEmpty()
+            if (nonCompliantRules.isNotEmpty()) {
+                report(
+                    CodeSmell(
+                        issue,
+                        Entity.from(annotationEntry),
+                        message = "Cannot @Suppress ${nonCompliantRules.formatMessage()} " +
+                            "due to the current configuration.",
+                    )
+                )
+            }
+        }
+    }
+
+    private fun List<String>.formatMessage(): String = if (size > 1) {
+        "rules "
+    } else {
+        "rule "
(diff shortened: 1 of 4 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSuppress.kt
```diff
@@ -0,0 +1,90 @@
...
+ * class Bar
+ * </noncompliant>
+ *
+ * <compliant>
+ * package foo
+ *
+ * // When the rule "MaximumLineLength" is forbidden
+ * @Suppress("UNCHECKED_CAST")
+ * class Bar
+ * </compliant>
```
cortinico, line 37: Vercel is unhappy with this:

```
Invalid rule documentation for noncompliant and compliant code examples in Rule ForbiddenSuppress contains an incorrect noncompliant code definition.
```

Not sure why this is the case, but if you try to `./gradlew generateDocumentation` locally you should be able to reproduce

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSuppress.kt
```diff
@@ -0,0 +1,90 @@
...
+ */
+class ForbiddenSuppress(config: Config = Config.empty) : Rule(config) {
+
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Maintainability,
+        "",
```
cortinico, line 44: This should not be left empty 👍

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSuppress.kt
```diff
@@ -0,0 +1,90 @@
...
+                .find { it is KtValueArgumentList }
+                ?.children
+                ?.filterIsInstance<KtValueArgument>()
+                ?.findForbiddenRules() ?: emptyList()
+            if (nonCompliantRules.isNotEmpty()) {
+                report(
+                    CodeSmell(
+                        issue,
+                        Entity.from(annotationEntry),
+                        message = "Cannot @Suppress ${nonCompliantRules.formatMessage()}.",
```
cortinico, line 64:
```suggestion
                        message = "Cannot @Suppress ${nonCompliantRules.formatMessage()} due to the current configuration.",
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSuppress.kt
gfreivasc: Will have a look

cortinico (commented):
Great stuff 👍 Thanks for taking the time to send this over 
:)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSuppress.kt
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    private fun List<String>.formatMessage(): String = if (size > 1) {
+        "rules "
+    } else {
+        "rule "
+    } + joinToString(", ") { "\"$it\"" }
+
+    private fun List<KtValueArgument>.findForbiddenRules(): List<String> = mapNotNull { argument ->
```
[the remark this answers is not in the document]
gfreivasc: I figured since this is a private method, the approach would be to test the findings output which leverages this filtering, which has been done in one of the test groups. I do agree the name should be changed since `filter` is more fitting.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSuppress.kt
```diff
@@ -0,0 +1,90 @@
...
+        val stringTemplate = argument.children
+            .find { it is KtStringTemplateExpression }
+            ?.children
+            ?.find { it is KtLiteralStringTemplateEntry }
+        val text = stringTemplate?.text
```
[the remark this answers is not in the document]
gfreivasc: I extracted that because I was following along debugging the code, will add to the chain.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSuppress.kt
gfreivasc: It seems like it is unhappy with the use of the `@Suppress` annotation. Maybe it's conflicting with documentation annotations and causing it break.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSuppress.kt
```diff
@@ -0,0 +1,91 @@
...
+import org.jetbrains.kotlin.psi.KtAnnotationEntry
+import org.jetbrains.kotlin.psi.KtLiteralStringTemplateEntry
+import org.jetbrains.kotlin.psi.KtStringTemplateExpression
+import org.jetbrains.kotlin.psi.KtValueArgument
+import org.jetbrains.kotlin.psi.KtValueArgumentList
+
+/**
+ * This rule allows to set a list of rules whose suppression is forbidden. This can be used to discourage abuse of the
+ * `Suppress` and `SuppressWarnings` annotations. Detekt will report suppression of all forbidden rules. This rule is
+ * not capable of reporting suppression of itself, as that's a language feature with precedence.
```
schalkms, line 21: minor formatting and grammar

```suggestion
 * This rule allows to set a list of rules whose suppression is forbidden.
 * This can be used to discourage the abuse of the `Suppress` and `SuppressWarnings` annotations.
 * Detekt will report the suppression of all forbidden rules.
 * This rule is not capable of reporting the suppression of itself, as that's a language feature with precedence.
```

schalkms (commented):
I second Nicos comments.
Thanks for this awesome contribution! 👍

gfreivasc:
@cortinico @schalkms addressed your comments.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenSuppress.kt
gfreivasc: I found that through trial and error. The stacktrace was not helpful at all, removing the annotation line was my first try and from that it was simple to find the problem. I then attempted to escape the docstring with `\@` and `@@` but only the latter worked as expected.

cortinico: Yup seems like this is the correct way to add `@` in code samples: 

https://github.com/detekt/detekt/blob/ad9ef16bb9b8c43d4dfc65947e5e178877c219db/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/DocumentationCollector.kt#L21

TIL I guess :)

cortinico (approved):
LGTM 👍 Thank you for sending this

schalkms (approved):

PR #4900, opened 2022-06-02
Prepare 1.21.0-RC1

cortinico:
Time to prepare a new release. This time it was harder than expected :(

* d4fcd487b Had to make sure the Gradle Plugin has a valid POM, as we now publish it to Central as well. Without a valid pom the whole publishing can't be performed.
* 87f32e117 Revert enabling MissingPackageDeclaration by default - See #4875
* e79afd7f7 Fix InvalidPackageDeclaration on functionalTests - This is really concerning as it means that our `functionalTest` for the Gradle Plugin are still consuming a version from Maven Central are not effectively testing the latest changes on `main` (cc @3flex for visiblity).

### DIFF
--- a/build-logic/src/main/kotlin/Versions.kt
+++ b/build-logic/src/main/kotlin/Versions.kt
@@ -1,6 +1,6 @@
 object Versions {
 
-    const val DETEKT: String = "1.20.0"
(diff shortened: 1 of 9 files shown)


cortinico:
Can we merge this as I can't create a Tag/Github release otherwise?

PR #4907, opened 2022-06-04
CastToNullableType: allow casting null keyword

dzirbel:
Fix a false positive for CastToNullableType for casting a literal `null` to a nullable type. This can be necessary when providing null as a function argument where the incoming type is parameterized. In this case using e.g. `null as? String` is less meaningful and generates a compiler warning that the cast cannot succeed.


BraisGabin (approved):

schalkms (approved):

BraisGabin:
Thanks for the fix!

PR #4909, opened 2022-06-04
Add option to add a reason to `ForbiddenImport`

BraisGabin:
I'm really happy to be able to introduce this feature :). Thanks @marschwar for creating the necesary tooling to be able to implement it.

Important: This introduces a breaking change in `ForbiddenImport`. It no longer allows comma-separated imports as a String. It should use yml lists. I'm targetting this to `1.22.0` because on `1.21.0` we added make that all those usages will be a warning. Tell me is we should be more conservative.

The reason that this is a draft is just to avoid accidental merges. But the code should be ready to merge.

Related with #3501

PR #4910, opened 2022-06-04
Add option to add a reason to `ForbiddenMethodCall`

BraisGabin:
Same as #4909 but now adding it to `ForbiddenMethodCall`

Important: This introduces a breaking change in `ForbiddenMethodCall`. It no longer allows comma-separated imports as a String. It should use yml lists. I'm targetting this to `1.22.0` because on `1.21.0` we added make that all those usages will be a warning. Tell me is we should be more conservative.

The reason that this is a draft is just to avoid accidental merges. But the code should be ready to merge.

Related with #3501

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
@@ -114,7 +114,7 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
 
         for (descriptor in descriptors) {
             methods.find { it.match(descriptor) }?.let { functionMatcher ->
-                val message = "The method $functionMatcher has been forbidden in the Detekt config."
+                val message = "The method `$functionMatcher` has been forbidden in the Detekt config."
                 report(CodeSmell(issue, Entity.from(expression), message))
             }
         }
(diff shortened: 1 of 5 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
```diff
@@ -87,10 +91,16 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
         } ?: return
 
         for (descriptor in descriptors) {
-            methods.find { it.match(descriptor) }?.let { functionMatcher ->
-                val message = "The method $functionMatcher has been forbidden in the Detekt config."
+            methods.find { it.value.match(descriptor) }?.let { forbidden ->
+                val message = if (forbidden.reason != null) {
+                    "The method `${forbidden.value}` has been forbidden: ${forbidden.reason}"
+                } else {
+                    "The method `${forbidden.value}` has been forbidden in the Detekt config."
```
schalkms, line 98:
```suggestion
                    "The method `${forbidden.value}` has been forbidden in the detekt config."
```

schalkms (approved):
👍

marschwar:
How Do we go about breaking changes in minor releases? Do we need an upgrade gradle task for users to call?

marschwar (approved):

PR #4922, opened 2022-06-05
`IgnoredReturnValue`: add option `returnValueTypes` to enable rule for particular types

### LINKED ISSUE #4640: Add rule to check if Flow collected

```kotlin
val profileFlow: Flow<Profile> = ...

fun observeProfile() {
    profileFlow.onEach { ... } // FlowNotCollected: launchIn, collect or another function is not called
}

fun observeProfile(): Flow<ProfileUi>  {
    return profileFlow.onEach { ... } // Ok: Flow is not collected but returned from function
}
```

## Context

It would be nice to catch cases when flow is not collected because:
1. It is quite easy to forget to collect flow, or you can use `onEach` instead of `collect` by mistake.
2. Even if you never forget to collect flow you can mess with this issue after refactoring:
    Imagine you want to migrate some function from suspend to flow:
    ```diff
    -suspend fun refreshProfile() { ... }
    +fun refreshProfile(): Flow<RefreshState> { ... }
    ```
    After this change all methods used `refreshProfile` become broken because returned flow is not collected.

osipxd:
> @BraisGabin said:
> We could extend IgnoredReturnValue and make it receive a list of types from the configuration. The rule will flag any call to a function that returns one of those types if the returned is ignored.

This PR adds config option `returnValueTypes` (is this option name OK?) to make it possible to enable the rule for functions returning specified types.
Also, I've added `kotlinx.coroutines.flow.Flow` to the default config.

Resolves #4640

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -431,7 +431,7 @@ potential-bugs:
     active: true
   IgnoredReturnValue:
     active: true
-    restrictToAnnotatedMethods: true
+    restrictToConfig: true
     returnValueAnnotations:
       - '*.CheckResult'
       - '*.CheckReturnValue'
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -437,6 +437,10 @@ potential-bugs:
       - '*.CheckReturnValue'
     ignoreReturnValueAnnotations:
       - '*.CanIgnoreReturnValue'
+    returnValueTypes:
+      - 'kotlin.sequences.Sequence'
+      - 'kotlinx.coroutines.flow.*Flow'
+      - 'java.util.stream.*Stream'
     ignoreFunctionCall: []
   ImplicitDefaultLocale:
     active: true
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
@@ -13,11 +13,14 @@ import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
 import io.gitlab.arturbosch.detekt.api.internal.Configuration
 import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
 import io.gitlab.arturbosch.detekt.api.simplePatternToRegex
+import io.gitlab.arturbosch.detekt.rules.fqNameOrNull
 import org.jetbrains.kotlin.descriptors.annotations.AnnotationDescriptor
+import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.bindingContextUtil.isUsedAsExpression
 import org.jetbrains.kotlin.resolve.calls.util.getResolvedCall
+import org.jetbrains.kotlin.types.KotlinType
 import org.jetbrains.kotlin.types.typeUtil.isUnit
 
 /**
@@ -61,6 +64,15 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
         it.map(String::simplePatternToRegex)
     }
 
+    @Configuration("List of return types that should not be ignored")
+    private val returnValueTypes: List<Regex> by config(
+        listOf(
+            "kotlin.sequences.Sequence",
+            "kotlinx.coroutines.flow.*Flow",
+            "java.util.stream.*Stream",
+        ),
+    ) { it.map(String::simplePatternToRegex) }
+
     @Configuration(
         "List of function signatures which should be ignored by this rule. " +
             "Specifying fully-qualified function signature with name only (i.e. `java.time.LocalDate.now`) will " +
@@ -87,6 +99,7 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
         val annotations = resultingDescriptor.annotations
         if (annotations.any { it in ignoreReturnValueAnnotations }) return
         if (restrictToAnnotatedMethods &&
+            resultingDescriptor.returnType !in returnValueTypes &&
             (annotations + resultingDescriptor.containingDeclaration.annotations).none { it in returnValueAnnotations }
         ) return
 
@@ -100,9 +113,11 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
         )
     }
 
-    @Suppress("UnusedPrivateMember")
-    private operator fun List<Regex>.contains(annotation: AnnotationDescriptor): Boolean {
-        val fqName = annotation.fqName?.asString() ?: return false
-        return any { it.matches(fqName) }
+    private operator fun List<Regex>.contains(type: KotlinType?) = contains(type?.fqNameOrNull())
+    private operator fun List<Regex>.contains(annotation: AnnotationDescriptor) = contains(annotation.fqName)
+
+    private operator fun List<Regex>.contains(fqName: FqName?): Boolean {
+        val name = fqName?.asString() ?: return false
+        return any { it.matches(name) }
     }
 }
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
@@ -8,7 +8,9 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.UnstableApi
 import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.configWithFallback
 import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
 import io.gitlab.arturbosch.detekt.api.internal.Configuration
 import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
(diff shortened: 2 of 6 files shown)


detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
```diff
@@ -61,6 +64,11 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
         it.map(String::simplePatternToRegex)
     }
 
+    @Configuration("List of return types that should not be ignored")
+    private val returnValueTypes: List<Regex> by config(listOf("kotlinx.coroutines.flow.Flow")) {
```
osipxd, line 68: Is these name and description are good?

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt
```diff
@@ -92,6 +92,15 @@ class FindingsAssert(actual: List<Finding>) :
 }
 
 class FindingAssert(val actual: Finding?) : AbstractAssert<FindingAssert, Finding>(actual, FindingAssert::class.java) {
+
+    fun hasSourceLocation(line: Int, column: Int) = apply {
```
osipxd, line 96: I've added this function to make it possible to use this assert in asserts chain:
```kotlin
assertThat(findings)
    .singleElement()
    .hasSourceLocation(line = 5, column = 10)
    .hasMessage("The call onEach is returning a value that is ignored.")
```
instead of
```kotlin
assertThat(findings).singleElement()
assertThat(findings).hasSourceLocation(line = 5, column = 10)
assertThat(findings.first()).hasMessage("The call onEach is returning a value that is ignored.")
```

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValueSpec.kt
```diff
@@ -881,4 +881,70 @@ class IgnoredReturnValueSpec(private val env: KotlinCoreEnvironment) {
             assertThat(findings).isEmpty()
         }
     }
+
+    @Nested
+    inner class `return value types default config` {
```
osipxd, line 886: Should I add more tests?

cortinico:
By looking at this rule API, I think we should add also the `restrictToReturnValueTypes` boolean config.

```
  IgnoredReturnValue
    active: true
    restrictToAnnotatedMethods: true
    returnValueAnnotations:
      - '*.CheckResult'
      - '*.CheckReturnValue'
    ignoreReturnValueAnnotations:
      - '*.CanIgnoreReturnValue'
+   restrictToReturnValueTypes: true
+   returnValueTypes:
+     - 'kotlinx.coroutines.flow.Flow'
```

The rule should then check if `restrictToReturnValueTypes` is true and the returned type is one of the listed -> Then report.

It makes for a more explicit configuration and it aligns to `restrictToAnnotatedMethods`/`returnValueAnnotations`

osipxd:
> By looking at this rule API, I think we should add also the `restrictToReturnValueTypes` boolean config.
> The rule should then check if `restrictToReturnValueTypes` is true and the returned type is one of the listed -> Then report.
> 
> It makes for a more explicit configuration and it aligns to `restrictToAnnotatedMethods`/`returnValueAnnotations`

Thank you for API review!
Yes, i thought about it, but prefix `restrictTo` implies "exclusiveness" isn't it? In my mind `restrictToAnnotatedMethods` means "this rule will check only methods annotated with the specified annotations", and `restrictToReturnValueTypes` means "this rule will check only menthods returning the specified types", so it may be confusing if both of these flags are `true`.
But maybe it is only my feeling because of language barrier.

cortinico:
> Yes, i thought about it, but prefix `restrictTo` implies "exclusiveness" isn't it? In my mind `restrictToAnnotatedMethods` means "this rule will check only methods annotated with the specified annotations", and `restrictToReturnValueTypes` means "this rule will check only menthods returning the specified types", so it may be confusing if both of these flags are `true`.

Yes that's correct. If enabled, the rule will flag only function which are returning the values types you listed. If this is not the case, then I misunderstood what was the semantic of your change.

osipxd:
> Yes that's correct. If enabled, the rule will flag only function which are returning the values types you listed. If this is not the case, then I misunderstood what was the semantic of your change.

I mean sentence _"rule will check **only** annotated methods"_ conflicts with _"rule will check **only** methods returning the specified types"_ because both contain word "only". So maybe there may be better naming? Something that means _"always check methods returning specified types even if checks enabled only for annotated methods"_?

cortinico:
> So maybe there may be better naming? Something that means _"always check methods returning specified types even if checks enabled only for annotated methods"_?

I would suggest to don't implement this logic. The two `restrict` should apply on top of each other and further restrict the set of return values we inspect. What you're suggesting instead seems harder to explain and understand for users IMHO

BraisGabin:
🤔 I think that we should reframe the configuration of this rule. To me this issue should flag a function if it returns a `Flow` (or any other class listed on the configuration) or if it is annotated with some of the listed annotations. Then, we should have another flag that would ignore all those configurations and will flag all the cases where the return value is not `Unit` nor `Nothing`. Do that have sense? 

To do that I would deprecate the current boolean flag and rename it to some naming more clear.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
```diff
@@ -49,8 +54,14 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
     )
 
     @Configuration("if the rule should check only annotated methods")
+    @Deprecated("Use `restrictToConfig` instead")
     private val restrictToAnnotatedMethods: Boolean by config(defaultValue = true)
 
+    @Suppress("DEPRECATION")
+    @OptIn(UnstableApi::class)
+    @Configuration("If the rule should check only methods matching to configuration, or all methods")
+    private val restrictToConfig: Boolean by configWithFallback(::restrictToAnnotatedMethods, defaultValue = true)
```
osipxd, line 63: The best name that I came up with. Is there any better variants?

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
```diff
@@ -61,6 +72,15 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
         it.map(String::simplePatternToRegex)
     }
 
+    @Configuration("List of return types that should not be ignored")
+    private val returnValueTypes: List<Regex> by config(
+        listOf(
+            "kotlin.sequences.Sequence",
+            "kotlinx.coroutines.flow.*Flow",
+            "java.util.stream.*Stream",
```
osipxd, line 80: Is it OK to add java Streams here or should I keep only Kotlin types?

osipxd, line 80: Is it OK to add java Streams here or should I keep only Kotlin types?

osipxd (commented):
@cortinico, @BraisGabin, I've just updated this PR, can you take a look?

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
```diff
@@ -86,7 +106,8 @@ class IgnoredReturnValue(config: Config = Config.empty) : Rule(config) {
 
         val annotations = resultingDescriptor.annotations
         if (annotations.any { it in ignoreReturnValueAnnotations }) return
-        if (restrictToAnnotatedMethods &&
+        if (restrictToConfig &&
+            resultingDescriptor.returnType !in returnValueTypes &&
```
cortinico, line 110: Can we keep the older behavior in some form (i.e. check for `restrictToAnnotatedMethods` in another if condition as well)?

osipxd: But `restrictToAnnotatedMethods` will always equal to `restrictToConfig`, isn't it? Maybe I don't quite understand what you mean.

cortinico: Never mind. I realized that you marked `restrictToAnnotatedMethods` as deprecated so it won't be possible for users to use it, they'll be warned to migrate to the new property.

Without that, we could have had users with `restrictToAnnotatedMethods` set to true, in the next version of Detekt. For such users, you would have had to add some backward compatibility logic.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
BraisGabin: It's ok for me. We can always remove it later if we end up thinking of "java" as a "framework".

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
BraisGabin, line 63: Naming is so difficult. I think your naming is good. I'm just proposing other but I'm not sure if it's better. What do you think?

```suggestion
    @Suppress("DEPRECATION")
    @OptIn(UnstableApi::class)
    @Configuration("If the rule should check all methods ignoring the annotations or return types.")
    private val checkAllMethods: Boolean by configWithFallback({ !restrictToAnnotatedMethods }, defaultValue = true)
```

BraisGabin (approved):

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValue.kt
osipxd: Yes that's better

osipxd: But when I try to do it, I get the error:
image

The options I see:
1. Add the new parameter to `configWithFallback` that will transform fallback value
2. Keep the new property equal to `restrictToAnnotatedMethods`

BraisGabin (approved):

osipxd:
Sorry for ping, but should it be merged?

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -431,12 +431,16 @@ potential-bugs:
     active: true
   IgnoredReturnValue:
     active: true
-    restrictToAnnotatedMethods: true
```
cortinico, line 434: Reminder for self: this should be noted in the release notes

cortinico (approved):
I'm fine with merging this 👍

BraisGabin:
Thank you for the PR :)

detekt-core/src/main/resources/default-detekt-config.yml
osipxd: @cortinico this change was not mentioned in release notes 🙈

cortinico: Thanks for the reminder 👍

PR #4931, opened 2022-06-07
Add missing redirect

BraisGabin:
Google right now links to a 404 page when you look for `suppressors detekt`. This should fix it.


BraisGabin:
Man, Vercel is great! Thanks for that @cortinico. I could check that it works :).

cortinico (approved):

PR #4937, opened 2022-06-08
Support disabling config validation via tooling spec

arturbosch:
Disabling config validation is not yet supported in the cli via a flag.

The cli just set the spec value to false and `shouldValidateBeforeAnalysis` was never evaluated in `core`.

With this PR tools can programmatically disable config validation if needed.

For a usecase see https://github.com/detekt/detekt-intellij-plugin/issues/179.

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
@@ -41,7 +41,7 @@ internal fun CliArgs.createSpec(output: Appendable, error: Appendable): Processi
 
         config {
             useDefaultConfig = args.buildUponDefaultConfig
-            shouldValidateBeforeAnalysis = false
+            shouldValidateBeforeAnalysis = null
             knownPatterns = emptyList()
             // ^^ cli does not have these properties yet; specified in yaml config for now
             configPaths = config?.let { MultipleExistingPathConverter().convert(it) }.orEmpty()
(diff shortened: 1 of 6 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
```diff
@@ -41,7 +41,7 @@ internal fun CliArgs.createSpec(output: Appendable, error: Appendable): Processi
 
         config {
             useDefaultConfig = args.buildUponDefaultConfig
-            shouldValidateBeforeAnalysis = false
+            shouldValidateBeforeAnalysis = true
```
BraisGabin, line 44: 🤔 Would it be better to have 3 values here? Enable/disabled/undefined? This way we don't need to use `and`. If this is undefined the yml decides. Or we could just remove the yml value and make it only configurable by the cli.

Anyway, this is a step forward so LGTM.

BraisGabin (approved):

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
arturbosch: Three values would be better yes. That way the code says `spec > config`.
Yes, removing it from the config would be optimal in a future breaking change.
That is also the case for many other configuration flags we now have in cli and removed from config.

What was the procedure in deprecating config properties? Deprecate it, implement a cli alternative and remove the config property after three minor releases?

I will change change the PR to support three values `Enable/disabled/undefined` and in a later PR implement the cli flag and deprecate the config property.

BraisGabin: To deprecate you just need to add it inside `detekt-core/src/main/resources/deprecation.properties`. And yes, we should keep it for at least 3 minor versions.

BraisGabin (approved):

chao2zhang (approved):

PR #4948, opened 2022-06-12
UnnecessaryApply: fix false negative for assignment

dzirbel:
Fix a missed case of UnnecessaryApply where the apply contains a single field assignment but the result of the apply{} is unused. This pattern is actually used in 2/3 of the `<noncompliant>` examples and while it occurred in tests for false positives, no test actually checked that warnings were generated in the simple case.

I only added a few tests for the simple cases to avoid duplicating a number of the ones for function calls; more could be added for completeness but it would bloat the test file considerably.

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
@@ -117,12 +117,12 @@ class UnnecessaryApplySpec(val env: KotlinCoreEnvironment) {
                 class C {
                     var prop = 0
                 }
-                    
+
                 fun main() {
                     val list = ArrayList<C>()
                     list.add(
                         if (true) {
-                            C().apply { 
+                            C().apply {
                                 prop = 1
                             }
                         } else {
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
```diff
@@ -235,6 +235,89 @@ class UnnecessaryApplySpec(val env: KotlinCoreEnvironment) {
...
+                }
+
+                fun main() {
+                    val c = C()
+                    c.apply { 1 to 2 }
+                }
+                    """
+                )
+            ).isEmpty()
+        }
```
BraisGabin, line 298: 🤔 shouldn't this be flagged? There's no need for `.apply` here.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
```diff
@@ -235,6 +235,89 @@ class UnnecessaryApplySpec(val env: KotlinCoreEnvironment) {
...
+                    env,
+                    """
+                class C {
+                    var prop = 0
+                }
+
+                fun main() {
+                    val c = C()
+                    var x = 0
+                    c.apply { x = this.prop }
```
BraisGabin, line 313: This could be changed for `x = c.prop`. Or, at least, recommend `also`.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
dzirbel: Yes... as far as I can tell the existing code isn't really intended to warn for this case, i.e. when the receiver context is unused (or the one below where it could be simplified but not in a trivial way). For example, it also doesn't flag `"".apply { 1.toString() }` since it doesn't match that pattern. Probably the rule could be substantially improved to check whether the receiver is ever used, but I think that might be a very difficult thing to check since the rule would need to be certain that each sub-expression is independent from the receiver. (E.g. in this example, whether `toString()` was a function declared like `class X { fun Int.toString() = "" }` which is sometimes done.)

So for this PR I'm inclined to handle the simple cases without doing a more involved analysis.

BraisGabin: I'm OK not fixing these. But I would not add tests to prove this undesired behavior.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryApplySpec.kt
dzirbel: That's fair; removed these two tests.

BraisGabin (approved):
Thanks!

schalkms (approved):

Issue #4465, opened 2022-01-07
Align inputs for detekt type resolution tasks with Kotlin compilation tasks


3flex:
## Expected Behavior
The detekt `BindingContext` is identical to that used in equivalent Kotlin compilation task.

## Current Behavior
There are multiple existing issues:
1. Compilation classpath differs between detekt & Kotlin compilation task
2. Input files used to generate `BindingContext` differs
    * #3488 
    * https://github.com/detekt/detekt/issues/3272
    * https://github.com/detekt/detekt/issues/3409
    * https://github.com/detekt/detekt/issues/3684
    * https://github.com/detekt/detekt/issues/3913
3. Other compilation options differ (e.g. friend paths, compiler args, API version, target, language version)
    * #3067 
    * https://github.com/detekt/detekt/issues/3639
    * https://github.com/detekt/detekt/issues/3684
    * https://github.com/detekt/detekt/issues/3874

## Context
The main reason detekt-compiler-plugin works with no fuss is because it lets the Kotlin Gradle plugin and compiler infrastructure generate `BindingContext` then provides a separate list of files to analyze using that `BindingContext`. We can mimic this with detekt by configuring the detekt tasks based on the Kotlin compilation tasks and allowing detekt to treat files to be analysed separately from files used to generate `BindingContext`. In addition, support should be added for relevant compiler args that aren't currently recognised by detekt.

The tasks would be automatically setup by the plugin similar to this:
```kotlin
tasks.register<Detekt>("detektTask") {
    // files to analyse. This can be filtered at will.
    source = tasks.compileKotlin.get().source

    // New parameter: files used to generate BindingContext. Java & Kotlin source files, includes generated sources (as used
    // by Android Data Binding library & Android BuildConfig for example). Should not be filtered by users at all, as missing
    // files may lead to unresolved references.
    compilerFiles = tasks.compileKotlin.get().source

    classpath.setFrom(tasks.compileKotlin.get().classpath)

    // existing parameters for Detekt task
    languageVersion = tasks.compileKotlin.get().kotlinOptions.languageVersion
    jvmTarget = tasks.compileKotlin.get().kotlinOptions.jvmTarget

    // new parameters to add to Detekt task
    apiVersion = tasks.compileKotlin.get().kotlinOptions.apiVersion
    freeCompilerArgs = tasks.compileKotlin.get().kotlinOptions.freeCompilerArgs
    friendPaths = tasks.compileKotlin.get().friendPaths
}

detekt {
[...]

PR #4960, opened 2022-06-14
No experimental type solving

BraisGabin:
From: with https://github.com/detekt/detekt/discussions/4959

Right now a lot of our rules use type-solving and it is battle tested. It works! But we still say in the documentation that it is "experimental" but, to me, it is not experimental at all. It's one of the core features of detekt.

So I would like to remove the "experimental tag" from the type-solving, what do you think?

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
@@ -173,7 +173,7 @@ class CliArgs {
     */
     @Parameter(
         names = ["--classpath", "-cp"],
-        description = "EXPERIMENTAL: Paths where to find user class files and depending jar files. " +
+        description = "Paths where to find user class files and depending jar files. " +
             "Used for type resolution."
     )
     var classpath: String? = null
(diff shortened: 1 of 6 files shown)


3flex:
In general I support this, but I'd seen https://github.com/detekt/detekt/issues/4465 as requiring a solution before calling this stable (at least when running detekt from Gradle).

Functional test coverage is also lacking.

BraisGabin:
What does it mean experimental in this context? Those tasks don't have pulished edges but they are stable, right? I mean, could we track anything new as an improvement?

3flex:
I've listed a number of related issues in #4465 which still don't have a proper fix, though there are some workarounds in place now to quiet the console output or offer config workarounds.

Those are the kinds of open issues that make me suggest that this feature is still somewhat experimental.

The feature works in most common cases though, and we're generally aware of the outstanding issues. That's why I'm not totally opposed to merging this, since we're not enabling it by default or calling it stable just yet.

BraisGabin:
> The feature works in most common cases though, and we're generally aware of the outstanding issues. That's why I'm not totally opposed to merging this, since we're not enabling it by default or calling it stable just yet.

@3flex I understand your concerns, I know that we need to pulish some edges but in general it is stable and we should promote it. It finds lots of issues and there is no good reason to don't use it.

BraisGabin:
Closing this as there is not consensous on it. We could continue the discussion on an issue and recreate another PR once we decide.

Issue #4887, opened 2022-06-01
Caused by: java.lang.IllegalArgumentException: Error building function signature with range 414 - 211 for element: @JvmStatic         @Parameterized.Parameters(


waliahimanshu:
## Expected Behavior

Successfully analysis of test class with `@Parameterized.Parameters` with a large function returning a list of arrays of around 27 items

## Observed Behavior

```
Caused by: java.lang.IllegalArgumentException: Error building function signature with range 414 - 211 for element: @JvmStatic
        @Parameterized.Parameters(
            name = "{index}-Case-Source-{0}-FT-{1}-expectedFeature-{2}-expectedFeatureList-{3}"
        )
        fun data() =
              listOf(
                    arrayOf(CTA_PREMIUM_SEARCH, emptyList<FeatureToggle>(), PREMIUM_SEARCH, listOf(AD_FREE)),
                    arrayOf(
                        CTA_PREMIUM_SEARCH,
                        listOf(SAVES_LIMIT_TEST, SAVES_LIMIT_PROMOTION),
                        PREMIUM_SEARCH,
                        listOf(UNLIMITED_SAVED_RECIPES, AD_FREE)
                    ),
                    arrayOf(
                        CTA_PREMIUM_SEARCH,
                        listOf(SAVES_LIMIT_TEST),
                        PREMIUM_SEARCH,
                        listOf(UNLIMITED_SAVED_RECIPES, AD_FREE)
                    ),
                    arrayOf(
                        CTA_PREMIUM_SEARCH,
                        listOf(SAVES_LIMIT_PROMOTION),
                        PREMIUM_SEARCH,
                        listOf(UNLIMITED_SAVED_RECIPES, AD_FREE)
                    ),
                    arrayOf(
                        CTA_PREMIUM_SEARCH,
                        listOf(FeatureToggle.HALL_OF_FAME),
                        PREMIUM_SEARCH,
                        listOf(HALL_OF_FAME, AD_FREE)
                    ),
                    arrayOf(
                        CTA_PREMIUM_SEARCH,
                        listOf(PREMIUM_SEARCH_FILTERS),
                        PREMIUM_SEARCH,
                        listOf(PREMIUM_FILTERS, AD_FREE)
                    ),
                    arrayOf(
                        CTA_PREMIUM_SEARCH,
[...]

Issue #4874, opened 2022-05-29
`compile-test-snippets` is confusing for new contributors


BraisGabin:
### A bit of context:

Our unit tests contain *a lot* of code snippets. Because, in general, we don't compile that code, it's easy to write code that doesn't compile. That makes that our tests doesn't test what we want. To avoid this problem we compile all the snippets to ensure that they are valid.

Problem: This extra step consumes a lot of execution time making our tests even more slow.

Solution: Because this check is testing the test code (and it's not testing production code) it's ok if we don't run it *every time*. So we just check it on CI.

### Current state

But that solution has a problem! The new contributors are bitten by this on CI and they don't know what's going on (an example is #4872 but we had a lot of these in the past).

### One possible solution

We could configure a "bot" to write a comment explaining this when [the workflow `compile-test-snippets` fails](https://github.com/detekt/detekt/blob/2f5b8f593361615dda1f0acb460afdd014ab460d/.github/workflows/pre-merge.yaml#L75). This will give fast feedback to the contributor about what's going on, why we have this "strange" behaviour and how s/he can reproduce it locally.

This is an issue and it's not a PR because the actions that I found to add that comment doesn't work when the PRs come from fork repository. And that's exactly the case when we want to add that comment.

cortinico:
> We could configure a "bot" to write a comment explaining this when [the workflow `compile-test-snippets` fails](https://github.com/detekt/detekt/blob/2f5b8f593361615dda1f0acb460afdd014ab460d/.github/workflows/pre-merge.yaml#L75). This will give fast feedback to the contributor about what's going on, why we have this "strange" behaviour and how s/he can reproduce it locally.

Follow up here: This is yet another thing than can be easily automated with Danger (see more on this here https://github.com/detekt/detekt/discussions/4696#discussioncomment-2619017).

Essentially, once the CI results are complete, this can be one of the message reports on the PR by danger. Something along the line of "hey the `compile-test-snippets` CI job failed. This is because bla bla bla. You can test locally with bla bla bla".

dzirbel:
A near-ideal solution might be to have `compile-snippet-tests` be enabled by default when running a single test but not the entire test suite; that would make individual changes easier to test with minimal impact and wouldn't slowing down the entire build. I've never tried doing something quite like that, so I'm not sure if it's possible - but might be if we can access the JUnit filters at test runtime.

cortinico:
[...]

Issue #4767, opened 2022-04-25
Method signature in baseline changes when adding/removing a doc comment


roomscape:
## Expected Behavior
- Given I have a method that exceeds the allowed method length or method complexity
- And I have suppressed the warning via my `baseline.xml`
- When I add a doc comment (`/** ... */`) to the method
- Then detekt should continue to pass

## Observed Behavior
After adding the doc comment, the suppression no longer works. Re-running `detektBaseline` shows that the signature has actually changed.

## Steps to Reproduce
See https://github.com/roomscape/detekt-baseline-issue, or:

- Create a long method in a class
- Run `detekt` and observe that it fails
- Run `detektBaseline`
- Run `detekt` and observe that it passes
- Add a doc comment to the long method
- Run `detekt` and observe that it fails
- Run `detektBaseline` and observe that the `baseline.xml` file has changed

Suppression for a method without a doc comment:
```xml
    <ID>LongMethod:Example.kt$Example$fun checkFoo()</ID>
```

Suppression for the same method after a doc comment is added:
```xml
    <ID>LongMethod:Example.kt$Example$ fun checkFoo()</ID>
```

## Context
- I tried to add a doc comment to an existing method in a large project
- I had to regenerate the detekt baseline in order to push my changes

## Your Environment
* Version of detekt used: 1.20.0
* Version of Gradle used (if applicable): 7.4
* Operating System and version: macOS 12.2.1

cortinico:
Yeah that's a bug. I think we tried to address it once with #4335 but that had to be reverted as it was causing problems.

VitalyVPinchuk:
It's unreproducible on the latest main.
Seems to be fixed.

BraisGabin:
Closing then, thanks!

PR #4961, opened 2022-06-15
Fix Signatures.kt:buildFunctionSignature

VitalyVPinchuk:
For #4887 ('buildFunctionSignature' fails when function body has KDoc)

### DIFF
--- a/detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/internal/SignaturesSpec.kt
+++ b/detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/internal/SignaturesSpec.kt
@@ -1,0 +1,128 @@
+package io.gitlab.arturbosch.detekt.api.internal
+
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.test.compileAndLint
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.Assertions.assertThatThrownBy
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.junit.jupiter.api.Test
+
+class SignaturesSpec {
+    private var result = ""
+    private val subject = TestRule { result = it }
+
+    @Test
+    fun `function with type reference`() {
+        subject.compileAndLint(functionWithTypeReference())
+        assertThat(result).isEqualTo(signatureWithTypeReference())
+    }
+
+    @Test
+    fun `function without type reference`() {
+        subject.compileAndLint(functionWithoutTypeReference())
+        assertThat(result).isEqualTo(signatureWithoutTypeReference())
+    }
+
+    @Test
+    fun `function throws exception`() {
+        assertThatThrownBy {
+            subject.compileAndLint(functionWontCompile())
+        }
+            .isInstanceOf(IllegalArgumentException::class.java)
+            .hasMessageContaining("Error building function signature")
+    }
+}
+
+private class TestRule(val block: (String) -> Unit) : Rule() {
+    override val issue = Issue(javaClass.simpleName, Severity.CodeSmell, "", Debt.TWENTY_MINS)
+
+    override fun visitNamedFunction(function: KtNamedFunction) {
+        block(function.buildFullSignature())
+    }
+}
+
+private fun signatureWithTypeReference() = """
+    Test.kt${'$'}TestClass.Companion${'$'}@JvmStatic @Parameterized.Parameters(name = "parameterName") /** * Function KDoc */ fun data(): List<Array<out Any?>>
+""".trimIndent()
+
+private fun signatureWithoutTypeReference() = """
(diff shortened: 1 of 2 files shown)


VitalyVPinchuk:
`:detekt-api:test` fails, but locally it passes alright. Why is that?

BraisGabin:
You can find information about that here: #4874. Any feedback there is more than welcome.

detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/internal/SignaturesSpec.kt
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Test
+    fun `function throws exception`() {
+        assertThatThrownBy {
+            subject.compileAndLint(functionWontCompile())
+        }
+            .isInstanceOf(IllegalArgumentException::class.java)
+            .hasMessageContaining("Error building function signature")
+    }
```
BraisGabin, line 36: plase, inline the code in the tests. Those functions make the revision of the code harder because you need to jump up and down.

And, are them the minimal code snippets to make these tests pass? I mean, could we remove some annotations, reduce the size of the body, etc?

detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/internal/SignaturesSpec.kt
```diff
@@ -0,0 +1,128 @@
...
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.test.compileAndLint
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.Assertions.assertThatThrownBy
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.junit.jupiter.api.Test
+
+class SignaturesSpec {
+    private var result = ""
```
BraisGabin, line 14: Avoid the usage of `var` in the tests please. Try to keep all the code inside the `@Test`.

detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/internal/SignaturesSpec.kt
```diff
@@ -0,0 +1,128 @@
...
+    private val subject = TestRule { result = it }
+
+    @Test
+    fun `function with type reference`() {
+        subject.compileAndLint(functionWithTypeReference())
+        assertThat(result).isEqualTo(signatureWithTypeReference())
```
BraisGabin, line 20: This way you can call directly to `buildFullSignature` and you don't need to have a "fake" rule:

```kotlin
compileContentForTest("code")
  .findChildByClass(KtNamedFunction::class.java)!!
  .buildFullSignature()
```

VitalyVPinchuk: Thanks, I'll fix that

BraisGabin (approved):
Thanks!

BraisGabin:
This is a fix but it's also kind of "breaking". It will force to regenerate lots of baselines. I'm ok with that but just a heads up if some maintainer wants to pospone he merge to 1.22.0. cc @cortinico

cortinico:
> This is a fix but it's also kind of "breaking". It will force to regenerate lots of baselines. I'm ok with that but just a heads up if some maintainer wants to pospone he merge to 1.22.0

I think we can merge this, but would be great to include a paragraph in the 'migration' section of the changelog.

BraisGabin:
I think that something like this could do the trick on the changelog:

We updated the baseline creation. Now the kdoc is not part of the function signature. This could create that already suppressed issues in your baseline reappear. To fix that just recreate your baseline.

VitalyVPinchuk:
> I think that something like this could do the trick on the changelog:
> 
> We updated the baseline creation. Now the kdoc is not part of the function signature. This could create that already suppressed issues in your baseline reappear. To fix that just recreate your baseline.

I think KDoc comments were trimmed before that too.
```
element.docComment?.let {
    methodStart = it.endOffset - it.startOffset
}
```

cortinico:
> > I think that something like this could do the trick on the changelog:
> > We updated the baseline creation. Now the kdoc is not part of the function signature. This could create that already suppressed issues in your baseline reappear. To fix that just recreate your baseline.
> 
> I think KDoc comments were trimmed before that too.
> 
> ```
> element.docComment?.let {
>     methodStart = it.endOffset - it.startOffset
> }
> ```

Can we verify this. If we require users to regenerate baselines because of trimming of KDocs, that should be reported clearly in the changelog 👍

VitalyVPinchuk:
@cortinico No, I was wrong!
Previous revision doesn't pass current tests because of leaving comments in the signature!
```
org.opentest4j.AssertionFailedError: 
expected: "Test.kt$fun data()"
 but was: "Test.kt$/* comments */ fun data()"
```
Consider the issue #4767

cortinico:
> Previous revision doesn't pass current tests because of leaving comments in the signature!

Still this doesn't answer my original question: do we need to ask users to regenerate baselines?

VitalyVPinchuk:
> Still this doesn't answer my original question: do we need to ask users to regenerate baselines?

Yes, we do.

PR #4964, opened 2022-06-17
Bump minimum supported Gradle version to 6.7.1

3flex:
Aligns to Kotlin 1.7.0, and allows us to support [Java Toolchains](https://docs.gradle.org/current/userguide/toolchains.html).

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ If you want to use a SNAPSHOT version, you can find more info on [this documenta
 
 #### Requirements
 
-Gradle 6.1+ is the minimum requirement. However, the recommended versions together with the other tools recommended versions are:
+Gradle 6.7.1+ is the minimum requirement. However, the recommended versions together with the other tools recommended versions are:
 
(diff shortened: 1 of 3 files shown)


BraisGabin (approved):

cortinico (approved):

3flex:
Off topic - do we need both a milestone AND the blocked tag?

I think of blockers as being something external that prevents merging for some reason, but this PR isn't blocked by anything, we're just making the choice not to merge it yet, and the milestone has been set appropriately.

cortinico:
> Off topic - do we need both a milestone AND the blocked tag?

Nope not necessarily. But the blocked tag is red and has a better visual cue and is harder to miss. Milestones instead can be confused easily as it's just a matter of a character.

3flex:
Maybe we should have a separate "don't merge yet" or "queued for next release" tag (or something like it)?

cortinico:
> Maybe we should have a separate "don't merge yet" or "queued for next release" tag (or something like it)?

I'm fine either way 👍

BraisGabin (approved):

chao2zhang (approved):

PR #4966, opened 2022-06-18
Refactor Gradle tasks to use Gradle's managed properties

3flex:
Unfortunately we can't use managed properties everywhere without some additional refactoring, but this still helps simplify a few things.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -56,33 +56,33 @@ import java.io.File
 import javax.inject.Inject
 
 @CacheableTask
-open class Detekt @Inject constructor(
+abstract class Detekt @Inject constructor(
     private val objects: ObjectFactory
 ) : SourceTask(), VerificationTask {
 
     @get:Classpath
-    val detektClasspath: ConfigurableFileCollection = objects.fileCollection()
+    abstract val detektClasspath: ConfigurableFileCollection
 
     @get:Classpath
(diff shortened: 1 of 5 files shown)


BraisGabin (approved):

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -61,7 +61,9 @@ class DetektPlugin : Plugin<Project> {
 
     private fun Project.registerGenerateConfigTask(extension: DetektExtension) {
         tasks.register(GENERATE_CONFIG, DetektGenerateConfigTask::class.java) {
-            it.config.setFrom(project.provider { extension.config })
+            val defaultConfigPath = rootDir.toPath().resolve(CONFIG_DIR_NAME).resolve(CONFIG_FILE)
+            it.config.setFrom(defaultConfigPath)
```
chao2zhang, line 65: I believe this is a behavior change: Previously we use the `defaultConfigPath` only if the extension config is not present, now `defaultConfigPath` is always added in the first.

It may or may not affect the user behavior, but I am wondering if you think we should lean towards being conservative here.

cortinico: Agree here. Also having a comment on why this is needed would improve readability

3flex: I've dropped this - I did go through the logic when I made the change and it should have been sound, but in the absence of test coverage and in the interest of just getting the PR merged, it's better to leave it out.

PR #4971, opened 2022-06-18
Simplify signing configuration

3flex:
Now the signing tasks will always be registered, but their execution will be skipped when the signing key or password is missing. This is a change from today where the tasks are only registered when the signing key and password are provided.

This is required to play nicely with com.gradle.plugin-publish v1.0.


schalkms (approved):

chao2zhang:
Is this PR blocked by https://github.com/detekt/detekt/pull/4970?

3flex:
It's the other way around - updating com.gradle.plugin-publish to 1.0.0-rc-3 requires the changes made in this PR.

chao2zhang (approved):

PR #4978, opened 2022-06-20
Give a better error message for --jvm-target

yogurtearl:
This will tell you what the bad value was,
and what all the acceptable values are.


chao2zhang (approved):
Thank you for your contribution!

cortinico (approved):

PR #4980, opened 2022-06-21
Add NestedLambdaOnSameLine style rule

dzirbel:
Add a new rule NestedLambdaOnSameLine which requires that nested lambdas cannot be on a single line, which can make it hard to parse.

Also sort the rules in StyleGuideProvider to make it clear where new rules should be inserted.


BraisGabin:
🤔 I don't know if I agree with this one. I think that both styles are completely fine.

schalkms (commented):
I agree with @BraisGabin here. This would certainly be a controversial rule.

> 🤔 I don't know if I agree with this one. I think that both styles are completely fine.

PR #4946, opened 2022-06-11
Drop dependency-analysis plugin

3flex:
This is blocking update to Kotlin 1.7.0 (#4821)

PR #4982, opened 2022-06-21
Add dependency-analysis plugin

3flex:
The issue causing incompatibility with the Kotlin 1.7.0 update (see https://github.com/detekt/detekt/pull/4946) appears to be resolved with the newer 1.7.0 version of the plugin. I've tested the kotlin-1.7 branch with the new dependency-analysis plugin version applied and it builds successfully.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,6 +5,7 @@ plugins {
     id("releasing")
     id("io.gitlab.arturbosch.detekt")
     alias(libs.plugins.gradleVersions)
+    alias(libs.plugins.dependencyAnalysis)
 }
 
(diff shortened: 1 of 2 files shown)


BraisGabin:
Is it possible that it doesn't work on JVM8?

3flex:
The plugin will not be supported anymore when Gradle itself is running on Gradle 8.

Eventually the detekt project should start using Gradle's toolchains support and we can work around this but that's a longer term thing.

BraisGabin:
One more reason (no a strong one to be honest) to think about https://github.com/detekt/detekt/discussions/4130

PR #4985, opened 2022-06-21
Add MaxChainedCallsOnSameLine style rule

dzirbel:
Add a new rule MaxChainedCallsOnSameLine to limit the number of chained calls on placed on a single line. This works well alongside CascadingCallWrapping in #4979 to make long call chains more readable by wrapping them on new lines.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxChainedCallsOnSameLine.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxChainedCallsOnSameLine.kt
@@ -1,0 +1,79 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtQualifiedExpression
+import org.jetbrains.kotlin.psi.KtUnaryExpression
+
+/**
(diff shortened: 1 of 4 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxChainedCallsOnSameLine.kt
```diff
@@ -0,0 +1,79 @@
...
+class MaxChainedCallsOnSameLine(config: Config = Config.empty) : Rule(config) {
+    override val issue = Issue(
+        id = javaClass.simpleName,
+        severity = Severity.Style,
+        description = "Chained calls beyond the maximum should be wrapped to a new line.",
+        debt = Debt.FIVE_MINS,
+    )
+
+    @Configuration("maximum chained calls allowed on a single line")
+    private val maxChainedCalls: Int by config(defaultValue = 5)
```
BraisGabin, line 37: 🤔 Isn't 5 far too much? I would say 3 or 4. But this is really subjective.

BraisGabin (commented):
What do you think about `?:` should it count? I can imagine a chain of elvis operators too... I'm not 100% sure about this, I'm just asking for more opinions.

BraisGabin:
Again, great rule :)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxChainedCallsOnSameLine.kt
dzirbel: I was leaning toward making it more conservative, since in my experience it's not uncommon to have calls on e.g. network models that can be at least a few references deep. But I don't feel strongly; happy to change to any other value.

dzirbel:
> What do you think about `?:` should it count? I can imagine a chain of elvis operators too... I'm not 100% sure about this, I'm just asking for more opinions.

I agree in practice it contributes to the complexity of a line in the same way; I didn't include just since it seemed more likely to be unintuitive. Also more than happy to add it if others have thoughts, or include it as a configuration parameter (but that's likely to overcomplicate it as well).

BraisGabin:
I approve this PR but I would like to know what others think about the default value. I feel it should be lowered a bit but I would like to have more opinions here.

BraisGabin (approved):

cortinico:
> I feel it should be lowered a bit but I would like to have more opinions here.

I think 5 is a good number. Don't have a strong opinion. As the rule is disabled by default, users can easily customize it once they enable it

Issue #4492, opened 2022-01-13
CanBeNonNullable false-negative


BraisGabin:
## Expected Behavior

`assignedExpression` violates `CanBeNonNullable`.

```kotlin
private fun visitAssignment(assignedExpression: KtExpression?) {
    if (assignedExpression == null) return
    val name = if (assignedExpression is KtQualifiedExpression) {
        assignedExpression.selectorExpression
    } else {
        assignedExpression
    }?.text ?: return
    assignments.getOrPut(name) { mutableSetOf() }.add(assignedExpression)
}
```

## Observed Behavior
detekt is fine with that code

## Context
If a function returns `Unit` any null check to exit the function if the parameter is `null` should be consider as a parameter that could be non-null.

I saw this false-negative while working on this PR: #4491

## Your Environment
* Version of detekt used: main
* Link to your project (if it's a public repository): detekt.

cortinico:
Isn't #4431 supposed to catch this actually?

BraisGabin:
Yes, but it is not.

VitalyVPinchuk:
Hi!
Can I work on this?

BraisGabin:
Please go ahead :)

VitalyVPinchuk:
Should `!is` in `if` / `when` statements also be reported?
e.g.
```
fun foo(a: Int?) {
    if (a !is Int) return
    println(a)
}
```

PR #4993, opened 2022-06-23
Fix false-negative for CanBeNonNullable

VitalyVPinchuk:
For #4492 
Report code smell when null-check expression returns 'Unit'


VitalyVPinchuk:
Hmm... I didn't think of it.
We'll lose part of the side effect if we make parameter non-nullable. Probably, it's better not to report such a case...
Give your thoughts, please

marschwar:
That was exactly my thought. I would not report if there is any side effect before the guard statement. Since detecting if it really is a side effect would be hard, I would not allow anything before the guard.

VitalyVPinchuk:
You're right. I'll fix it.

BraisGabin (approved):

BraisGabin (approved):

marschwar (approved):

PR #5004, opened 2022-06-28
Fix reported text position of wrapping rule

arturbosch:
Before: notice the Wrapping reporting at the file level

Screenshot from 2022-06-28 13-48-48

After this PR: after `fold` and at the `}`
Screenshot from 2022-06-28 15-08-10

Fixes https://github.com/detekt/detekt-intellij-plugin/issues/189


arturbosch:
Ups, already fixed in https://github.com/detekt/detekt/pull/4998

Issue #5014, opened 2022-06-30
MaxChainedCallsOnSameLine fails on import statements.


NickButcher1:
## Expected Behavior
detekt should not report any errors.

## Observed Behavior
detekt reports the following for every import statement with 6 or more dots.
MaxChainedCallsOnSameLine - [6 chained calls on a single line; more than 5 calls should be wrapped to a new l(...)] at

## Steps to Reproduce
Update from 1.21.0-RC1 to 1.21.0-RC2 and add this config
  MaxChainedCallsOnSameLine:
    active: true
    maxChainedCalls: 5

## Context
Trying to enable the new MaxChainedCallsOnSameLine rule.

## Your Environment

* Version of detekt used: 1.21.0-RC2.
* 7.4.2
* Operating System and version: Windows 11.

BraisGabin:
@dzirbel could you check this?

dzirbel:
Thanks for the report - this is definitely an issue (along with `package` directives) which should be easy enough to fix in #5020.

TWiStErRob:
Sorry @BraisGabin, found it, didn't look deep enough 😅. I think this issue can be closed as the PR is merged.

BraisGabin:
You are right! Thanks

PR #5020, opened 2022-06-30
#5014 Fix MaxChainedCallsOnSameLine false positives

dzirbel:


dzirbel:
I've now run on the detekt codebase (with a max of 7 since there are [some](https://github.com/detekt/detekt/blob/main/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgsSpec.kt#L18) violations longer than 5) to verify that there aren't any more obvious false positives.

There are still interesting cases ([example](https://github.com/detekt/detekt/blob/main/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt#L92)) with fully-qualified class names which still violate this rule; thoughts on what to do there? I haven't looked into it yet, but I'm not certain whether it will be possible to differentiate those from normal qualified expressions.

BraisGabin (approved):

BraisGabin:
I think that we will need to use type solving.

amitdash291 (approved):

amitdash291:
> There are still interesting cases ([example](https://github.com/detekt/detekt/blob/main/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt#L92)) with fully-qualified class names which still violate this rule; thoughts on what to do there? I haven't looked into it yet, but I'm not certain whether it will be possible to differentiate those from normal qualified expressions.

Not sure if this example qualifies for `MaxChainedCallsOnSameLine` violation, but it is something that needs cleaning up anyway.

schalkms:
> I think that we will need to use type solving.

How would that rule benefit from type solving? Could you please highlight an example?

schalkms (approved):

dzirbel:
> > I think that we will need to use type solving.
> 
> How would that rule benefit from type solving? Could you please highlight an example?

This was in response to this [example](https://github.com/detekt/detekt/blob/main/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt#L92) of a fully-qualified class name which this rule currently flags. Probably this rule should ignore such cases (although this one can just be imported), but I agree that type resolution is probably necessary to determine whether a chained calls are from a fully qualified class name. I'll open a separate issue to track/discuss that.

BraisGabin (approved):

PR #5032, opened 2022-07-03
Add ending line and column to Location.kt

### LINKED ISSUE #5025: Add ending line and column to `Location.kt`
External tooling, such SARIF or IDE may need ending lines and columns, in addition to the `SourceLocation`. They should be provided in `:detekt-api`.

_Originally posted by @BraisGabin in https://github.com/detekt/detekt/pull/5011#discussion_r911449557_

VitalyVPinchuk:
Close #5025 

Added `endSource` as the last parameter in the `Location` constructor.

I need some guidelines:
 - Is it necessary to change the name of parameters, for example `startSource` and `endSource`?
 - Should the order of the parameters be changed? 
 - Should `compact()`  string representation be updated?
 - Should there be a test for such trivia?

This could trigger changes in many files.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
@@ -24,17 +24,27 @@ data class Location @Deprecated("Consider relative path by passing a [FilePath]"
         )
     )
     val file: String,
-    val filePath: FilePath = FilePath.fromAbsolute(Paths.get(file)),
-    val endSource: SourceLocation = SourceLocation(-1, -1)
+    val filePath: FilePath = FilePath.fromAbsolute(Paths.get(file))
 ) : Compactable {
+    var endSource: SourceLocation = source
+        private set
 
     @Suppress("DEPRECATION")
     constructor(
         source: SourceLocation,
         text: TextLocation,
+        filePath: FilePath
+    ) : this(source, text, filePath.absolutePath.toString(), filePath)
+
+    @Suppress("DEPRECATION")
+    constructor(
+        source: SourceLocation,
+        endSource: SourceLocation,
+        text: TextLocation,
         filePath: FilePath,
-        endSource: SourceLocation = SourceLocation(-1, -1)
-    ) : this(source, text, filePath.absolutePath.toString(), filePath, endSource)
+    ) : this(source, text, filePath.absolutePath.toString(), filePath) {
+        this.endSource = endSource
+    }
 
     @Suppress("DEPRECATION")
     @Deprecated(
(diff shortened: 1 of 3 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
```diff
@@ -24,15 +24,17 @@ data class Location @Deprecated("Consider relative path by passing a [FilePath]"
         )
     )
     val file: String,
-    val filePath: FilePath = FilePath.fromAbsolute(Paths.get(file))
+    val filePath: FilePath = FilePath.fromAbsolute(Paths.get(file)),
+    val endSource: SourceLocation = SourceLocation(-1, -1)
```
BraisGabin, line 28: Is this the best default value that we can think of? Would it be better to copy the same as source? Or maybe make endSource nullable and change it to not-null for 2.0.

BraisGabin (commented):
>  - Is it necessary to change the name of parameters, for example `startSource` and `endSource`?

I don't think so. That would introduce even more breaking changes in the api for little benefit. 

>  - Should the order of the parameters be changed? 

That's up to you. You will need to create a new constructor matching the old one to avoid breaking the api so it's OK don't add it as the first one. 

>  - Should `compact()`  string representation be updated?

No

>  - Should there be a test for such trivia?

I don't think so. You are just adding a parameter.

----

But I would change the code inside core to inform this new value in this same pr. This way we can see that this is the change that we need.

VitalyVPinchuk:
Hey @BraisGabin Thanks for clarifying things.
I made just another variable to data class and assign it in the new constructor.

VitalyVPinchuk:
> But I would change the code inside core to inform this new value in this same pr. This way we can see that this is the change that we need.

Sorry, I'm afraid I didn't get it. What core are talking about?

BraisGabin:
> > But I would change the code inside core to inform this new value in this same pr. This way we can see that this is the change that we need.
> 
> Sorry, I'm afraid I didn't get it. What core are you talking about?

I was on my mobile and clearly I dind't explain myself.

You are adding the ability to store endSource inside `Location.kt`. What I think we should have in this same PR or at least in other PR on top of this one is the code that fills this information. The idea is to be sure that this API works for us. It should, but just to be sure.

----

Open question, mainly for maintainers. What do you think about a `data class` that has a property outside its constructor? (We could even have a rule for something like this). I don't know what to think about it. In first places `Location` sholdn't be a `data class` in fisrt place because it breaks the binary compatibility each time we change it. But now that it is a `data class` what do we prefer:

- Ensure binary compatibility but with a "strange `data class`"
- Keep the `data class` consistent but breaking the binary compatibility a bit
- Remove `data` from `Location` to avoid future problems but with a bigger breaking change now

BraisGabin (approved):

BraisGabin:
Thanks :) this looks great

PR #5033, opened 2022-07-03
Xml Report Merger now merges duplicate smells across input report files

timothyolt:
Xml Report Merger can now merge duplicate smells across multiple input report files. This is useful especially in Android, where running Detekt on multiple flavors identities many of the same errors on different flavors. Also cuts down on the amount of noise when these reports are submitted to Sonar.

I tried to make this as readable as possible, even though this process is complicated. Let me know if there is anything to improve on.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/XmlReportMerger.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/XmlReportMerger.kt
@@ -2,6 +2,7 @@ package io.gitlab.arturbosch.detekt.report
 
 import org.w3c.dom.Document
 import org.w3c.dom.Node
+import org.w3c.dom.NodeList
 import java.io.File
 import javax.xml.parsers.DocumentBuilderFactory
 import javax.xml.transform.OutputKeys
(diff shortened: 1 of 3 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/XmlReportMerger.kt
```diff
@@ -16,42 +17,133 @@ object XmlReportMerger {
...
+                        appendChild(mergedDocument.importNode(it, true))
+                    }
+                }
+            )
+        }
+        return mergedDocument
+    }
+
+    /** A list of checkstyle xml files written by Detekt */
+    private class DetektCheckstyleReports(private val files: Collection<File>) {
```
timothyolt, line 56: Using classes here to help call out preconditons and assumptions about the input data. These functions started out as extension functions, but I didn't like the idea of applying something so specific to broad types such as `Colleciton<File>` and `List<Node>`

timothyolt:
Build failure is `:detekt-report-md:test` in `MdOutputReportSpec.kt asserts that the generated HTML is the same even if we change the order of the findings`

Comparing the differences between the expected/actual reports, the difference I found was on the last line of the MD report:

```md
generated with [detekt version 1.21.0-RC2](https://detekt.dev/) on 2022-07-03 15:04:34 UTC
```
vs 
```md
generated with [detekt version 1.21.0-RC2](https://detekt.dev/) on 2022-07-03 15:04:35 UTC
```

The difference is time? Is this a known flake?

timothyolt:
> 



> Build failure is `:detekt-report-md:test` in `MdOutputReportSpec.kt asserts that the generated HTML is the same even if we change the order of the findings`
> 
> Comparing the differences between the expected/actual reports, the difference I found was on the last line of the MD report:
> 
> ```lisp
> generated with [detekt version 1.21.0-RC2](https://detekt.dev/) on 2022-07-03 15:04:34 UTC
> ```
> 
> vs
> 
> ```lisp
> generated with [detekt version 1.21.0-RC2](https://detekt.dev/) on 2022-07-03 15:04:35 UTC
> ```
> 
> The difference is time? Is this a known flake?

Pushed a commit mocking the time for more consistent MarkdownReportSpec results

cortinico:
> The difference is time? Is this a known flake?

Thanks for spotting this. Nope it was not known, but as the Markdown report was introduced recently, this was most likely the root cause.

BraisGabin (approved):

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/XmlReportMerger.kt
```diff
@@ -16,42 +17,133 @@ object XmlReportMerger {
...
     }
 
     /**
-     * Use code instead of XSLT to exclude whitespaces.
+     * A list of checkstyle `file` nodes that may contain 0 to many `error` nodes
+     *
+     * ```xml
+     * <file>
+     *     <error>
+     * </file>
```
chao2zhang, line 86: :+100: on this documentation

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/XmlReportMerger.kt
```diff
@@ -16,42 +17,133 @@ object XmlReportMerger {
...
+         *
+         * @see CheckstyleSourceFileNodes
+         */
+        fun parseCheckstyleToSourceFileNodes() =
+            CheckstyleSourceFileNodes(
+                files.filter { reportFile -> reportFile.exists() }
+                    .flatMap { existingReportFile ->
+                        val checkstyleNode = documentBuilder.parse(existingReportFile.inputStream())
+                        val sourceFileNodes = checkstyleNode.documentElement.childNodes.asSequence().filterWhitespace()
+                        sourceFileNodes
```
chao2zhang, line 75:
```suggestion
                        checkstyleNode.documentElement.childNodes.asSequence().filterWhitespace()
```

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/XmlReportMerger.kt
chao2zhang: I don't have strong opinions, so I would just conform to the pattern in this file (Wrapping inside a class seems to be the pattern)

chao2zhang (approved):

PR #5036, opened 2022-07-04
MaxChainedCallsOnSameLine: don't count package references as chained calls

### LINKED ISSUE #5028: MaxChainedCallsOnSameLine reports fully-qualified class names
## Expected Behavior
The `MaxChainedCallsOnSameLine` should probably ignore fully-qualified class names (like this [example](https://github.com/detekt/detekt/blob/ddcf77d9d4651bd117ff622180312141f900ac23/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt#L92) from detekt's codebase). These call chains are somewhat different in nature from its intended purpose, and the rule probably shouldn't have a second effect of restricting fully-qualified class names, which may be necessary (or at least preferred over [named imports](https://kotlinlang.org/docs/packages.html#imports) with `as`) when classes with the same name but different packages are used. See discussion in #5020.

A fix would likely require type resolution to determine the role of chained calls, but this complicates the rule and makes it harder to be applied, just to avoid a fairly rare edge case. So I wanted to open this issue for discussion/feedback on how to proceed.

## Observed Behavior
`MaxChainedCallsOnSameLine` reports fully qualified class names above its `maxChainedCalls` limit.

## Your Environment
* Version of detekt used: 1.21.0-RC2

t-kameyama:
Fixes #5028

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxChainedCallsOnSameLineSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxChainedCallsOnSameLineSpec.kt
@@ -3,8 +3,8 @@ package io.gitlab.arturbosch.detekt.rules.style
 import io.gitlab.arturbosch.detekt.rules.KotlinCoreEnvironmentTest
 import io.gitlab.arturbosch.detekt.test.TestConfig
 import io.gitlab.arturbosch.detekt.test.assertThat
+import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
 import io.gitlab.arturbosch.detekt.test.lint
-import io.gitlab.arturbosch.detekt.test.lintWithContext
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.junit.jupiter.api.Test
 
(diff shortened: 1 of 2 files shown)


amitdash291 (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxChainedCallsOnSameLineSpec.kt
```diff
@@ -1,58 +1,61 @@
...
-class MaxChainedCallsOnSameLineSpec {
+@KotlinCoreEnvironmentTest
+class MaxChainedCallsOnSameLineSpec(private val env: KotlinCoreEnvironment) {
     @Test
     fun `does not report 2 calls on a single line with a max of 3`() {
         val rule = MaxChainedCallsOnSameLine(TestConfig(mapOf("maxChainedCalls" to 3)))
         val code = "val a = 0.plus(0)"
 
-        assertThat(rule.compileAndLint(code)).isEmpty()
+        assertThat(rule.lintWithContext(env, code)).isEmpty()
```
BraisGabin, line 18: Is it possible to use `compileLintWithContext` in all this cases?

BraisGabin (approved):
Thanks!

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxChainedCallsOnSameLineSpec.kt
t-kameyama: [0e537f8](https://github.com/detekt/detekt/pull/5036/commits/0e537f8ed1dfd44e640d8667be87bdb64b23113f)

TWiStErRob:
Does this also fix import statements?

https://github.com/TWiStErRob/net.twisterrob.sun/security/code-scanning/275

BraisGabin:
That was fixed on other pr.

PR #5040, opened 2022-07-06
Bump ktlint to version 0.46.1 and implement new rules

kvn-stgl:
See https://github.com/pinterest/ktlint/releases/tag/0.46.0 and https://github.com/pinterest/ktlint/releases/tag/0.46.1 for full release notes. 

What did I do:
- I implemented the new (experimental) rules 
- I changed the rules from experimental ruleset to the standard ruleset which changed with the new version (see https://github.com/pinterest/ktlint/pull/1481). These are also tagged with `@ActiveByDefault`
- I rewrote the KtLintMultiRule a bit to support the `runAfterRule` visitorModifier from ktlint
- There are a lot of filename renamings because of the filename rule checks now single method files https://github.com/pinterest/ktlint/pull/1117


kvn-stgl:
I'm unsure how I should proceed with the `filename` rule. 
Problem: 

> A file containing only one (non private) top level declaration (class, interface, object, type alias or function) must be named after that declaration. The name also must comply with the Pascal Case convention. The same applies to a file containing one single top level class declaration and one ore more extension functions for that class. filename (https://github.com/pinterest/ktlint/pull/1117)
So we have a lot of file naming changes that lead to API changes. Currently, the apiCheck task fails because the changes are not binary compatible. So should I supress the filename warnings or should dump a new API file?

BraisGabin:
Would it be possible to split this PR? It's big to make a proper review.

I'm thinking in somethin like this:

- One updating ktlint and enable by default no new non-experimental rules (but enabling them since `1.22.0` instead of `1.21.0).
- Other with the different fixes/refactors
- And other more adding the new rules.

I don't know how possible is what I'm asking for but it would help a lot.

----

> I'm unsure how I should proceed with the filename rule.

I would disable that rule. We had similar rule on detekt that it's already enabled by default: `MatchingDeclarationName`.

PR #5051, opened 2022-07-08
Implement `TrimMultilineRawString`

BraisGabin:
This implements one of the proposed rules in #4705

To be more precise it implements this one:

> Check that every multiline string uses either `trimMargin()` or `trimIndent()` - Kotlin Coding Conventions doc says "To maintain indentation in multiline strings, use `trimIndent` when the resulting string does not require any internal indentation, or `trimMargin` when internal indentation is required"

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrimMultilineRawString.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrimMultilineRawString.kt
@@ -48,7 +48,7 @@ class TrimMultilineRawString(val config: Config) : Rule(config) {
     override fun visitStringTemplateExpression(expression: KtStringTemplateExpression) {
         super.visitStringTemplateExpression(expression)
 
-        if (expression.text.lines().count() == 1) return
+        if (expression.text.lines().count() <= 1) return
 
         val nextCall = expression.getQualifiedExpressionForSelectorOrThis()
             .getQualifiedExpressionForReceiver()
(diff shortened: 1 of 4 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrimMultilineRawString.kt
```diff
@@ -0,0 +1,74 @@
...
+        javaClass.simpleName,
+        Severity.Style,
+        "Multiline raw strings should be followed by `trimMargin()` or `trimIndent()`.",
+        Debt.FIVE_MINS
+    )
+
+    override fun visitStringTemplateExpression(expression: KtStringTemplateExpression) {
+        super.visitStringTemplateExpression(expression)
+
+        if (expression.text.lines().count() == 1) return
```
schalkms, line 51: What happens if the text of the expression is empty? Would this statement then yield 0?
Should the check be more conservative and flag `<= 1`?

```suggestion
        if (expression.text.lines().count() <= 1) return
```

schalkms (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/TrimMultilineRawString.kt
BraisGabin: The expression should contain at least the `""` or `""""""`. But I agree the `<=` is safer.

PR #5059, opened 2022-07-09
Fix current indentation

BraisGabin:
This PR fixes the issues that #5058 found on our code base. I did to ensure that the rule works properly.

If you think about it these cahnges are really few. We have way more bad formatted raw strings. The point is that the rule implemented on #5058 only flags raw-strings that call `.trimMargin()` or `.trimIndent()` and we don't call those functions. I think that we should change that and always call `.trimIndent()` (#5051 will help with that). Then the rule will flag way more bad formatted raw-strings and we will be able to fix them.


BraisGabin:
I'm moving this to draft until #5060 is fixed.

BraisGabin:
This is ready for review now that #5060 is fixed.

cortinico (approved):

PR #5062, opened 2022-07-11
ReturnCount: Make configuration parameter more explicit

TWiStErRob:
## Background
I want to configure `ReturnCount` to ignore Android's onMenuItemSelected and onOptionsItemSelected because they should have many returns to prevent bugs and strange behavior.

## Problem
In the JSON-schema we have this definition:
```
            "excludedFunctions": {
              "type": "string"
            },
```
Stringly typed items are hard to use, because they don't define structure.
It's not clear if this string is a regex, or a pipe separated list, or a comma separated list, or even a semi-colon separated regex list.

## Investigation
Note: this investigation has to be done by anyone who wants to use this configuration, and they might not know how to navigate the code or not have will.

### Docs
https://detekt.dev/docs/rules/style/#returncount

|  |
| - |

No conclusive result.

### Source code
https://github.com/detekt/detekt/blob/1b330e9e42a385a588c9f61021f2da6ae7495a92/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt#L62-L63

Now I know it's split somehow.

### Tests
https://github.com/detekt/detekt/blob/a2734b18c133020520183292c9d2ff0593040a1c/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt#L350

Now I know it's split by comma.

### Deeper look
https://github.com/detekt/detekt/blob/1b330e9e42a385a588c9f61021f2da6ae7495a92/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt#L93

calls

https://github.com/detekt/detekt/blob/1b330e9e42a385a588c9f61021f2da6ae7495a92/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/SplitPattern.kt#L31

also

https://github.com/detekt/detekt/blob/1b330e9e42a385a588c9f61021f2da6ae7495a92/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/SplitPattern.kt#L65-L70

Now I know it's definitely comma separated (from constructor default parameter), and that it's case insensitive (which is strange in a strongly typed case sensitive language), I also know the comma separated list is free-from because of trim and resilient in case there are multiple or trailing commas.

## Solution
Update docs to reflect this.

## Note
Obviously a more explicit way of doing this would be ideal:
```yaml
    excludedFunctions:
      # Ignore MenuProvider.onMenuItemSelected,
      # because it has a special structure for bug prevention.
      - onMenuItemSelected
      # Ignore Activity.onOptionsItemSelected and Fragment.onOptionsItemSelected,
      # because they has a special structure for bug prevention.
      - onOptionsItemSelected
```
but for now, this will do:
```yaml
    excludedFunctions:
      # Ignore MenuProvider.onMenuItemSelected, Activity.onOptionsItemSelected,
      # and Fragment.onOptionsItemSelected, because they have a special structure for bug prevention.
      '
        onMenuItemSelected,
        onOptionsItemSelected,
      '
```

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
@@ -59,7 +59,7 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
     @Configuration("define the maximum number of return statements allowed per function")
     private val max: Int by config(2)
 
-    @Configuration("define comma separated list of function names to be ignored by this check")
+    @Configuration("define a free-form comma separated list of function names to be ignored by this check")
     private val excludedFunctions: SplitPattern by config("equals") { SplitPattern(it) }
 
     @Configuration("if labeled return statements should be ignored")
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
@@ -59,7 +59,7 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
     @Configuration("define the maximum number of return statements allowed per function")
     private val max: Int by config(2)
 
-    @Configuration("define functions to be ignored by this check")
+    @Configuration("define comma separated list of function names to be ignored by this check")
     private val excludedFunctions: SplitPattern by config("equals") { SplitPattern(it) }
 
     @Configuration("if labeled return statements should be ignored")


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -59,7 +59,7 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
     @Configuration("define the maximum number of return statements allowed per function")
     private val max: Int by config(2)
 
-    @Configuration("define functions to be ignored by this check")
+    @Configuration("define a free-form comma separated list of function names to be ignored by this check")
```
TWiStErRob, line 62: re "free-form": [https://en.wikipedia.org/wiki/Free-form_language](https://en.wikipedia.org/wiki/Free-form_language#:~:text=In%20computer%20programming%2C%20a%20free,ends%20of%20lines%20are%20insignificant.)
> In computer programming, a free-form language is a programming language in which the positioning of characters on the page in program text is insignificant. Program text does not need to be placed in specific columns as on old punched card systems, and frequently ends of lines are insignificant.

BraisGabin: Interensting... this shouldn't be a `String` this should be a `List<String>`.

TWiStErRob: But on a related note: ForbiddenImport has the same problem, but there it's on the happy path, not an exceptional setup.

BraisGabin: Why do you say that `ForbiddenImport` have that issue? It shouldn't.

TWiStErRob: It was changed in v1.18.0-RC1 by https://github.com/detekt/detekt/commit/8e1328a2291d078d9bba6b8c8c2adbf0c1380ba4, and we're on 1.18.1 and it still uses a single string, and it works, but anyway, I see in the JSON schema it's now an array.

Issue #5060, opened 2022-07-09
`MaxLineLength` and raw strings


BraisGabin:
## Expected Behavior
```kotlin
"""
    MaxLineLength shouldn't raise any finding when it is find inside a raw string
"""
```

## Observed Behavior
It does: https://github.com/detekt/detekt/pull/5059#discussion_r917284016

## Context
#5059

## Your Environment
* Version of detekt used: main / 1.21.0-RC2
* Gradle scan link: https://ge.detekt.dev/s/2lpijultktfe2

BraisGabin:
Go for it

PR #5079, opened 2022-07-15
MaxLineLength should not raise any finding inside a raw string

gouri-panda:
closes: #5060


gouri-panda:
@marschwar I'll add some more different tests 👍

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLength.kt
```diff
@@ -97,6 +108,7 @@ class MaxLineLength(config: Config = Config.empty) : Rule(config) {
     companion object {
         private const val DEFAULT_IDEA_LINE_LENGTH = 120
         private val BLANK_OR_QUOTES = """[\s"]*""".toRegex()
+        private const val TQ = "\"\"\""
```
BraisGabin, line 111: You shouldn't need to look for `"""`. You should check the parent of those chars that are "MaxLineLength" and see if its type is `KtStringTemplateExpression`.

Issue #4457, opened 2022-01-05
Create a tool for rule authors to contribute to the default configuration


BraisGabin:
After merging #4315 that allow the rule authors to provide its own configuration we should provide a tool to generate those files and then document the feature.

We could document and then create the tool, but I think that it will be easier to create the tool than explain how to generate the config files by hand.

VitalyVPinchuk:
Hey @BraisGabin I'd like to work on this but need more profound description and some guidelines if you don't mind.

BraisGabin:
@VitalyVPinchuk I just updated the issue to give a bit more of context.

To be honest I don't know exactly what should be done here. I imagine the tool as a `cli`. Something like:

`detekt-generator config path/to/rules >config.yml`
`detekt-generator documentation path/to/rules >documentation.md`

Or even two cli:

`detekt-config-generator path/to/rules >config.yml`
`detekt-documentation-generator path/to/rules >documentation.md`

Once the `cli` is/are done we can wrap it/hem inside a gradle task. But that would be other task.

But this part is just an idea, the important thing is to have something that auto generates the `config.yml` file. The documentation has lower priority. And if it's not a `cli` that's ok too.

cortinico:
> And if it's not a `cli` that's ok too.

A Gradle Task would also work imho

PR #5080, opened 2022-07-15
Add config generator for custom rules

VitalyVPinchuk:
Work in progress
For #4457 

This is just a draft to understand if I'm doing it the right way.
Please, share your thoughts.

Created gradle task `generateCustomConfig` by copying `generateDocumentation` which accepts source path for rules and creates config.yml in the path indicated in config parameter.


cortinico:
> This is just a draft to understand if I'm doing it the right way.
> Please, share your thoughts.

Thanks for doing this.
I think we'll need a bit of re-work. Specifically, my understanding of the problem is:

- We want to expose a custom task to rule authors. So we'll have to create it inside the `detekt-gradle-plugin`.
- I would call the task differently. Like `detektGenerateCustomRuleConfig` or so.
- What the task should do is: output a `config.yml` inside the /resources folder of a custom project, or fail if there is no custom ruleset or so.

BraisGabin:
I agree with @cortinico but should we include it in our current gradle plugin? Or should we create a different one? 99.9% of our users will not use this action.

cortinico:
> but should we include it in our current gradle plugin?

Yes it should be inside the plugin, but as a separate Task. Ideally we should not register the task by default, but let users do so if they wish

VitalyVPinchuk:
Hello everyone!
I ran into problems and need help, please.

What I'm trying to do is to call a function in `detekt-generator` module (for example, `io.gitlab.arturbosch.detekt.generator.Main.main()`) from `detekt-gradle-plugin` package.
So I need a dependency to `detekt-generator`.

* In `DetektPlugin.kt:82` I added `dependencySet.add(project.dependencies.create("io.gitlab.arturbosch.detekt:detekt-generator:$version"))`. And when I try to load class in `DetektInvoker.kt:52` by calling `loader.loadClass("io.gitlab.arturbosch.detekt.generator.Main")` the "Class not found exception" occurs. But calling `loader.loadClass("io.gitlab.arturbosch.detekt.cli.Main")` works fine even though I delete `dependencySet.add(project.dependencies.create("io.gitlab.arturbosch.detekt:detekt-cli:$version"))`! How is that even possible???
And isn't Gradle trying to get the dependency from network repository? Because it requests version 1.20.0 whereas local version is 1.21.0.-RC2.

* I tried to `includeBuild("../detekt-generator")` in `build.gradle.kts`. But in this case applying plugin `id("module")` leads to error "module not found". Why is it so? And the other error is "name detekt-generator already used in main build". I don't see where the duplicate is.

* The only way I managed to do this is to add dependency directly from local binaries like this: `implementation(fileTree("../detekt-generator/build/libs/") { include("*.jar") })`. But I think it's not an option because it will require to manually build `detekt-generator` prior to building `detekt-gradle-plugin`.

cortinico:
Hey @VitalyVPinchuk 👋 Happy to follow up.

> What I'm trying to do is to call a function in `detekt-generator` module

I would suggest against doing this ☝️ 

Which function specifically are you looking into copying?
You should pass through the detekt-cli OR copy over the code you need inside the Gradle Plugin.

The reason you can't depend on `detekt-generator` is that `detekt-gradle-plugin` is an included build. It's fully isolated from the other packages and should remain so in order for better testing and project isolation.

VitalyVPinchuk:
Hi @cortinico 
> I would suggest against doing this ☝️

Okay, I got it, I won't!

> You should pass through the detekt-cli OR copy over the code you need inside the Gradle Plugin.

How do I pass through the `detekt-cli`? Is there a sample code?

> The reason you can't depend on `detekt-generator` is that `detekt-gradle-plugin` is an included build. It's fully isolated from the other packages and should remain so in order for better testing and project isolation.

Okay, we'll leave it as it is :) No dependencies.
But isn't it technically possible? I created for test purposes a standalone build inside detekt (like `build-logic` or `detekt-gradle-plugin`) and was able to  `includeBuild("../my-module")` in `detekt-gradle-plugin/settings.gradle.kts` and call its functions from `DetektPlugin.kt`.

cortinico:
> How do I pass through the `detekt-cli`? Is there a sample code?

I'm on mobile now, but essentially what the `detetkGenerate` task is already doing. I would need to look furhter into this as it could be non trivial 🤔

VitalyVPinchuk:
OMG!
Something went wrong with my squash 🤔

VitalyVPinchuk:
When adding `implementation(projects.detektGenerator)` to `detekt-core` a lot of tests fail because of  `detekt-formatting`.

cortinico:
The current solution seems overly complicated. I would do the following:

In this file:
https://github.com/detekt/detekt/blob/main/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt

I would add an extra parameter like:
```
@get:Input
val generateOnlyFromCustomRules: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType).convention(false)
```

If the user registers a task and sets this to true, you need to pass an extra paramter to the CLI here https://github.com/detekt/detekt/blob/fc235154b8a478ef2381c709b3327379b8341863/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt#L57-L63

This extra parameter should be consumed by the CLI here:
https://github.com/detekt/detekt/blob/fc235154b8a478ef2381c709b3327379b8341863/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt#L43-L48

You will probably need a `generateCustomRuleConfig` or so.

VitalyVPinchuk:
Hey Nicola,
Thanks for your answer.

I did almost the same but in different task file (and I will rework it the way you propose) `detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateCustomConfigTask.kt`
But still I don't know how to properly call `detekt-generator` because it contains the logic for generating config for custom rules.
What I do is preparing the list of arguments (like _-path-to-rule-source-files_ and _-output-config-directory_), then through invoker I get `CustomConfigExporter` object and via service loader get `CustomConfigProvider` which calls `detekt-generator` module to generate config.
Does it seem reasonable?

cortinico:
> But still I don't know how to properly call `detekt-generator` because it contains the logic for generating config for custom rules.

I believe you shouldn't call `detekt-generator` directly, but pass through the `detekt-cli` as we're already doing.

VitalyVPinchuk:
So it should be done like this, right?
```kotlin
val loader = classLoaderCache.getOrCreate(classpath)
val clazz = loader.loadClass("io.gitlab.arturbosch.detekt.cli.Main")
```

How can I add `detekt-generator` jar path to `classpath`?

cortinico:
I'd rather follow the same approach that we have inside `DetektGenerateConfigTask.kt`

https://github.com/detekt/detekt/blob/fc235154b8a478ef2381c709b3327379b8341863/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt#L74-L78

You should just pass an extra argument to the generator.

VitalyVPinchuk:
And based on the extra parameter should we modify
https://github.com/detekt/detekt/blob/eb2fde1cbca0460058a08a68d1fd24faa31cb575/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/DetektInvoker.kt#L50-L55
so that it loads `class io.gitlab.arturbosch.detekt.generator`  the same way 

or should we continue to follow the logic further and create another runner
https://github.com/detekt/detekt/blob/eb2fde1cbca0460058a08a68d1fd24faa31cb575/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Main.kt#L55-L60
which would still load `class io.gitlab.arturbosch.detekt.generator`?

cortinico:
I would follow the second approach. You should be find passing an extra `argument` to the CLI, and the inside the https://github.com/detekt/detekt/blob/main/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/ConfigExporter.kt you should be able to handle it somehow

VitalyVPinchuk:
Need an advice how to call this
https://github.com/detekt/detekt/blob/936edba6c75c4121a7bba6eb4e890d3593ed6357/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/ConfigExporter.kt#L39-L47

VitalyVPinchuk:
Is it ok if output file is set with _--config_ parameter and path to rules is set with _--input_?
Then config.yml would contain all the rules.

Or should there be a separate config.yml for every rule in its resources folder?
> What the task should do is: output a config.yml inside the /resources folder of a custom project, or fail if there is no custom ruleset or so.

In this case we need to figure out path to `/resources` folder (as input could be set to a particular file).

cortinico:
> Or should there be a separate config.yml for every rule in its resources folder?

Yup there should be only one `config.yml` per ruleset

> In this case we need to figure out path to /resources folder (as input could be set to a particular file).

You should be able to get the path with `javaClass.getResource("config.yml")`. You can search around in the codebase as we do something like it in other places.

cortinico:
@VitalyVPinchuk ping me when you're ready for another pass. I also see you have merge conflicts which needs to be addressed.

VitalyVPinchuk:
> @VitalyVPinchuk ping me when you're ready for another pass. I also see you have merge conflicts which needs to be addressed.

Thanks!
I'll fix them and get back to you again 😃

VitalyVPinchuk:
Hey @cortinico 
I need help on this. I pass the project.rootDir to --config parameter, so there are more than 3500 jar files. Also Gradle says that accessing project during runtime is not supported and I don't know how to tackle it (but config.yml is generated fine). 
https://github.com/detekt/detekt/blob/991d2295b9b7945ba9c1246904d96f0324fa72ae/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/runners/ConfigExporter.kt#L38-L44

Also I wasn't able to handle `javaClass.getResource("config.yml")`, there's no way I could get project's resources folder  of a custom rule. So I split the pass to files which seems unstable.
https://github.com/detekt/detekt/blob/991d2295b9b7945ba9c1246904d96f0324fa72ae/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/Generator.kt#L49-L53

cortinico:
> I need help on this. I pass the project.rootDir to --config parameter, so there are more than 3500 jar files. Also Gradle [says](https://docs.gradle.org/7.5.1/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution) that accessing project during runtime is not supported and I don't know how to tackle it (but config.yml is generated fine).

Hey @VitalyVPinchuk 
Thanks for asking. So I checked the branch, and you're right. The CLI doesn't have enough capacity to build a config (as it's just copying over the default config). Sorry for the back and forth.

I suggest that we proceed in this way:

1. We keep all the changes you did inside the `detekt-generator/` folder. We would need to write tests for those also before merging.
2. We revert all the changes to `detekt-cli` and `detekt-gradle-plugin`.
3. We will instruct rule authors to download the jar of ``detekt-generator` and invoke it with `--generate-custom-rule-config --input <path_to_their_rules_folder>`.
4. We add a page on the website on how to do this. Potentially we can offer also a custom Gradle task setup, without having to extend the Detekt Gradle Plugin.

I would also like to have an ack from @BraisGabin on this approach.

BraisGabin:
I like the idea. If we have a cli in the future we can iterate and wrap it inside a gradle plugin or whatever we want. It will not be the more published solution but it's a tool that solves the problem.

VitalyVPinchuk:
Hi!
I reverted unnecessary changes and added a test.

BraisGabin (approved):
LGTM :) Thanks for this work :)

BraisGabin:
It seems that the fails on Windows are "legit". Could you take a look?

VitalyVPinchuk:
> It seems that the fails on Windows are "legit". Could you take a look?

I've tried to fix it but Windows still can't find file `D:\a\detekt\detekt\detekt-rules-complexity\src\main\resources\config\config.yml`
What could be the problem?

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/Generator.kt
```diff
@@ -39,4 +39,35 @@ class Generator(
...
+                .toList()
+                .groupBy { (_, folder) -> folder }
+                .toList()
+                .forEach { (folder, list) ->
+                    val collector = DetektCollector()
+                    list.forEach { (file, _) ->
+                        collector.visit(file)
+                    }
+                    printer.printCustomRuleConfig(collector.items, folder)
+                }
```
BraisGabin, line 68: I think that the problem was that `ktFile.name.split("/src/main/kotlin")` on Windows should be `\src\main\kotlin`. But I think that it should not be necessary to use that. Instead of `flatMap` all the `KtFile`s we can keep them in different `List`s. I think that this could work:

```suggestion
            arguments.inputPath
                .map { parseAll(parser, it.resolve("src/main/kotlin/")) to it }
                .forEach { (list: Collection<KtFile>, folder: Path) ->
                    val collector = DetektCollector()
                    list.forEach { file ->
                        collector.visit(file)
                    }
                    printer.printCustomRuleConfig(
                        collector.items,
                        folder.resolve("src/main/resources/config/").toString()
                    )
                }
```

VitalyVPinchuk:
@BraisGabin thanks, it works!
Now I understand what the problem was.

BraisGabin (approved):

cortinico (approved):
Sorry for the late review, I think we're fine with this and we can merge it 🎉 Sorry for the back and forth but it was also a bit unclear for us how we wanted this feature to look like.

Last thing to do (as a follow-up) would be to write a short paragraph on the website about this.

Issue #5063, opened 2022-07-11
`ReturnCount.excludedFunctions` should be a `List<String>` instead of a `String`


BraisGabin:
## Expected Behavior
`ReturnCount.excludedFunctions` should be a `List<String>`

## Observed Behavior
It is a `String`.

## Context
#5062

## Your Environment
* Version of detekt used: main/1.21.0-RC2

marschwar:
Maybe it should be valueAndReason?

gouri-panda:
@BraisGabin Can I try this one :)

PR #5081, opened 2022-07-15
ReturnCount.excludedFunctions should be a List<String>

gouri-panda:
closes: #5063

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
@@ -59,8 +59,8 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
     @Configuration("define the maximum number of return statements allowed per function")
     private val max: Int by config(2)
 
-    @Configuration("define a free-form comma separated list of function names to be ignored by this check")
-    private val excludedFunctions: SplitPattern by config("equals") { SplitPattern(it) }
+    @Configuration("define a list of function names to be ignored by this check")
+    private val excludedFunctions: SplitPattern by config(listOf("equals")) { SplitPattern(it.joinToString(",")) }
 
     @Configuration("if labeled return statements should be ignored")
     private val excludeLabeled: Boolean by config(false)
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -626,7 +626,8 @@ style:
   ReturnCount:
     active: true
     max: 2
-    excludedFunctions: 'equals'
+    excludedFunctions:
+      - 'equals'
     excludeLabeled: false
     excludeReturnFromLambda: true
     excludeGuardClauses: false
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCountSpec.kt
@@ -347,7 +347,7 @@ class ReturnCountSpec {
                 TestConfig(
                     mapOf(
                         MAX to "2",
-                        EXCLUDED_FUNCTIONS to "test1,test2"
+                        EXCLUDED_FUNCTIONS to listOf("test1", "test2")
                     )
                 )
             ).compileAndLint(code)


cortinico:
`verify-generated-config-file` is broken. Could you re-run `./gradlew generateDocumentation`?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -59,8 +58,8 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
     @Configuration("define the maximum number of return statements allowed per function")
     private val max: Int by config(2)
 
-    @Configuration("define a free-form comma separated list of function names to be ignored by this check")
-    private val excludedFunctions: SplitPattern by config("equals") { SplitPattern(it) }
+    @Configuration("define a list of function names to be ignored by this check")
+    private val excludedFunctions: List<String> by config(emptyList())
```
BraisGabin, line 62: You are introducing a breaking change here. You should keep `"equals"`.

And I don't know why the tests doesn't complain but looking at the code of `SplitPattern` this code doesn't do the same as before, right?

gouri-panda: > You are introducing a breaking change here. You should keep `"equals"

Ok 
> 
> And I don't know why the tests doesn't complain but looking at the code of `SplitPattern` this code doesn't do the same as before, right?

Yes

gouri-panda: @BraisGabin Should we change the `SplitPattern` implementation and leave the signature of `ReturnCount` as before i.e `private val excludedFunctions: SplitPattern by config("equals") { SplitPattern(it) }`?

gouri-panda: Is this the right way to change?
`ReturnCount.kt`
```kt
    private val excludedFunctions: SplitPattern by config(listOf("equals")) { SplitPattern(it.toString()) }
```
`SplitPattern.kt`
```kt
@Suppress("detekt.SpreadOperator")
    private val excludes = text
        .commaSeparatedPattern(*delimiters.toCharArray().map { it.toString() }.toTypedArray())
        .mapIf(removeTrailingAsterisks) { seq ->
            seq.map { it.removePrefix("*") }
                .map { it.removeSuffix("*") }
                .map {it.removePrefix("[")}
                .map { it.removeSuffix("]") }
        }
        .toList()
```

BraisGabin: I would look at what `SplitPattern` does and reproduce it in this rule. I'm sure that I did this time on other classes some time ago but for some reason I missed this one.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -1,14 +1,8 @@
+@file:Suppress("WildcardImport", "NoWildcardImports")
```
BraisGabin, line 1: Don't use wildcard imports.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ReturnCount.kt
```diff
@@ -59,8 +53,8 @@ class ReturnCount(config: Config = Config.empty) : Rule(config) {
     @Configuration("define the maximum number of return statements allowed per function")
     private val max: Int by config(2)
 
-    @Configuration("define a free-form comma separated list of function names to be ignored by this check")
-    private val excludedFunctions: SplitPattern by config("equals") { SplitPattern(it) }
+    @Configuration("define a list of function names to be ignored by this check")
+    private val excludedFunctions: SplitPattern by config(listOf("equals")) { SplitPattern(it.toString()) }
```
BraisGabin, line 57: If you want to go this way you can use `it.joinString(",")` (or something similar, I can't recall the signature). This way you don't need to change `SplitPattern`. That could have unexpected implications.

BraisGabin (approved):

BraisGabin (approved):

PR #5109, opened 2022-07-20
Forbid usage of DiagnosticUtils.getLineAndColumnInPsiFile

BraisGabin:
This is a refactor that I did to implement #5058 but that PR is big and this refactor only adds noise there so I'm moving this quick refactor here to reduce the number of changed files there.

### DIFF
--- a/detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/KtFiles.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/KtFiles.kt
@@ -71,7 +71,7 @@ fun PsiFile.toFilePath(): FilePath {
     }
 }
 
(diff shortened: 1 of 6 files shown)


detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/KtFiles.kt
```diff
@@ -68,6 +71,14 @@ fun PsiFile.toFilePath(): FilePath {
     }
 }
 
+// #3317 If any rule mutates the PsiElement, searching the original PsiElement may throw exception.
```
schalkms, line 74:
```suggestion
// #3317 If any rule mutates the PsiElement, searching the original PsiElement may throw an exception.
```

schalkms (approved):

PR #5116, opened 2022-07-23
Assert end source locations

BraisGabin:
I need this refactor to finish: #5058

The idea behind this PR is in the comment: https://github.com/detekt/detekt/pull/5058/files#r927045468

---

All the changes made outside `FindingsAssertions.kt` where done using the "fix deprecation" feature.

### DIFF
--- a/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt
+++ b/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt
@@ -6,7 +6,7 @@ import io.gitlab.arturbosch.detekt.api.TextLocation
 import org.assertj.core.api.AbstractAssert
 import org.assertj.core.api.AbstractListAssert
 import org.assertj.core.util.CheckReturnValue
-import java.util.Objects
+import java.util.*
 
 @CheckReturnValue
 fun assertThat(findings: List<Finding>) = FindingsAssert(findings)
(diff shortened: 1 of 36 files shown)


detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt
```diff
@@ -6,7 +6,7 @@ import io.gitlab.arturbosch.detekt.api.TextLocation
 import org.assertj.core.api.AbstractAssert
 import org.assertj.core.api.AbstractListAssert
 import org.assertj.core.util.CheckReturnValue
-import java.util.Objects
+import java.util.*
```
marschwar, line 9: I did not even realize that we allow wildcard imports. I would have expected that some check would have failed

marschwar (approved):

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FindingsAssertions.kt
BraisGabin: our current configuration:

```yaml
  WildcardImport:
    active: true
    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
    excludeImports:
      - 'java.util.*'
```

This case hit both excludes. I'm going to change this for us.

PR #5121, opened 2022-07-23
Don't exclude tests in WildcardImport Rule

BraisGabin:
Context: https://github.com/detekt/detekt/pull/5116#discussion_r928118070

This PR is also kind of "housekeeping" because I remove the `'java.util.*'` as a valid Wildcard import as we have in the default values.

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -274,3 +274,6 @@ style:
   VarCouldBeVal:
     active: true
     ignoreAnnotated: ['Parameter']
+  WildcardImport:
+    active: true
+    excludeImports: []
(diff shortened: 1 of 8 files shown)


3flex:
This looks like it has #5130 mixed in, I'm guessing that was unintentional and that this should be focused only on the update to remove the exclusion for test sources and fix the new issues that are reported?

BraisGabin:
> This looks like it has #5130 mixed in, I'm guessing that was unintentional and that this should be focused only on the update to remove the exclusion for test sources and fix the new issues that are reported?

You are right. I have far too many open branches right now >_<

3flex (approved):

3flex:
I've tagged it as a notable change because some users will get new issues reported if they're using the default config.

PR #5129, opened 2022-07-24
Create `detekt-rules-ruleauthors` module

BraisGabin:
This is a new first-party detekt plugin for rules that only rule authors will need. In the next PR I'll add the first rule to this rule-set.

Any idea with the naming is more than welcome. I don't like `detekt-authors` too much but `detekt-detekt` seems even worst.


detekt-authors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/AuthorsProvider.kt
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * The authors ruleset provides rules that ensures good practices when writing detekt rules
+ */
+@ActiveByDefault("1.22.0")
+class AuthorsProvider : RuleSetProvider {
+
+    override val ruleSetId: String = "detekt"
+
+    override fun instance(config: Config) = RuleSet(ruleSetId, @Suppress("UseEmptyCounterpart") listOf())
```
BraisGabin, line 16: `:detekt-generator` relies in the fact that there is a `listOf()` in this file. If I use `emptyList()` it fails. This is an edge-transitory-case so I'm just Suppressing it. In the next PR I'll have a rule here so I'll remove `@Suppress`

cortinico:
> I don't like detekt-authors too much but detekt-detekt seems even worst.

I'm also not a big fan of the name. I would go for either
- detekt-rules-customrules
- detekt-rules-authors
- detekt-rules-ruleauthors

BraisGabin:
I didn't want to start it by `detekt-rules` because `detekt-rules` are the detekt's standard rules. I imagine that's the reason for `detekt-formatting` to be called like that instead of `detekt-rules-formatting`.

Other option is to move these to another repo under the `detekt` organization. That could also have sense. We talked about moving `detekt-formatting` out of these repo so maybe is a good thing to directly create these rules out of the main repo.

cortinico:
> I imagine that's the reason for `detekt-formatting` to be called like that instead of `detekt-rules-formatting`

Since we are at this: IMHO `detekt-formatting` is not the best name. I would have called it `detekt-rules-formatting` OR `detekt-ktlint-wrapper`.

I think that having the assumtion that `detekt-rules-*` contains custom rules is sound, and would make the codebase easier to discover.

> Other option is to move these to another repo under the detekt organization. That could also have sense. We talked about moving detekt-formatting out of these repo so maybe is a good thing to directly create these rules out of the main repo.

That's also doable, but it really depends on how often we expect those rules to change. Maybe they can start off inside the Detekt repo and being moved outside in the future.

BraisGabin:
> > I imagine that's the reason for `detekt-formatting` to be called like that instead of `detekt-rules-formatting`
> 
> Since we are at this: IMHO `detekt-formatting` is not the best name. I would have called it `detekt-rules-formatting` OR `detekt-ktlint-wrapper`.

Agree, `detekt-ktlint-wrapper` seems a better name. Or just `detekt-ktlint`. Because there are other formatting tools.

> I think that having the assumtion that `detekt-rules-*` contains custom rules is sound, and would make the codebase easier to discover.

That have sense, true.

--- 

So I'm renaming it to `detekt-rules-ruleauthors` for the gradle subproject.

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/RuleAuthorsProvider.kt
```diff
@@ -0,0 +1,18 @@
...
+import io.gitlab.arturbosch.detekt.api.RuleSetProvider
+import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
+
+/**
+ * The rule authors ruleset provides rules that ensures good practices when writing detekt rules
+ */
+@ActiveByDefault("1.22.0")
+class RuleAuthorsProvider : RuleSetProvider {
+
+    override val ruleSetId: String = "detekt"
```
cortinico, line 14:
```suggestion
    override val ruleSetId: String = "ruleauthors"
```

detekt-authors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/AuthorsProvider.kt
```diff
@@ -0,0 +1,18 @@
...
+import io.gitlab.arturbosch.detekt.api.RuleSetProvider
+import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
+
+/**
+ * The authors ruleset provides rules that ensures good practices when writing detekt rules
+ */
+@ActiveByDefault("1.22.0")
+class AuthorsProvider : RuleSetProvider {
+
+    override val ruleSetId: String = "detekt"
```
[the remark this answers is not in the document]
BraisGabin: Done

cortinico (approved):

PR #5153, opened 2022-07-31
Bundle slf4j-nop in detekt-formatting JAR

### LINKED ISSUE #4811: SLF4J Warning after updating to 1.20
Hi,

I know an issue was already created on this problem during 1.20-RC2 (and was fixed) but i have this problem actually after upgrading from 1.19 to 1.20

## Expected Behavior
I saw on the old issue this warning should not be displayed

## Observed Behavior
I got this warning when lint + autoCorrect : 
`SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.`

## Steps to Reproduce
Run lint with 1.20 with autoCorrect (CLI)

## Context
I just upgraded from 1.19 to 1.20 

## Your Environment
* Version of detekt used: 1.20
* Version of Gradle used (if applicable): 7.2

3flex:
Fixes #4811

### DIFF
--- a/detekt-formatting/build.gradle.kts
+++ b/detekt-formatting/build.gradle.kts
@@ -2,6 +2,8 @@ plugins {
     id("module")
 }
 
+val extraDepsToPackage: Configuration by configurations.creating
+
 dependencies {
     compileOnly(projects.detektApi)
     implementation(libs.ktlint.rulesetStandard) {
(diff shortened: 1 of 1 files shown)


detekt-cli/build.gradle.kts
```diff
@@ -23,6 +23,7 @@ dependencies {
     }
     runtimeOnly(projects.detektCore)
     runtimeOnly(projects.detektRules)
+    runtimeOnly(libs.slf4j.nop)
```
cortinico, line 26: Not a big fan of this as we're pushing a Ktlint related dependency inside our `cli` package which doesn't really need it. Couldn't we find a better solution?

3flex: As I was writing a long response justifying this approach I realised the better way of doing this - see recent commit.

BraisGabin (approved):

schalkms (approved):

PR #5161, opened 2022-07-31
Mark MultiRule as deprecated

### LINKED ISSUE #4240: Should we deprecate MultiRule?
I'd like to do some benchmarks the actual `NamingRules` that is a `MultiRule` agains all the rules that it handles runned individually. Is it worth the speed boost that we get from it?

The `MultiRule` class create complexity in the code:

- When you test a Rule that is inside a MultiRule you aren't testing exactly what it will be executed (https://github.com/detekt/detekt/pull/4147#discussion_r721468277)
- The multi rule and the configuration does't work nicely and we forget about them frequently (#4230 and #4071)
- We need to add code checking if the rule is a `Rule` or a `MultiRule` to do different things (#4239):

https://github.com/detekt/detekt/blob/f944be01561efd80b35c6b4427434caf63dcf88c/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt#L98-L102

https://github.com/detekt/detekt/blob/f944be01561efd80b35c6b4427434caf63dcf88c/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/rules/RuleSets.kt#L37-L42

For all that reasons I think that we should get some benchmarks so we could take an informed decission. If the performance impact is not really high I'd vote to deprecate it.

marschwar:
This is the first step in removing `MultiRule`. It closes #4240. 

I will create a checklist (or separate issues) for all remaining steps.

_Note:_ I changed the usages to fully qualified as the build was failing with deprecation warnings in the imports and I did not want to suppress deprecation warnings for the entire file.


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -95,9 +94,10 @@ internal class Analyzer(
         bindingContext: BindingContext,
         compilerResources: CompilerResources
     ): Map<RuleSetId, List<Finding>> {
+        @Suppress("DEPRECATION")
         fun isCorrectable(rule: BaseRule): Boolean = when (rule) {
             is Rule -> rule.autoCorrect
-            is MultiRule -> rule.rules.any { it.autoCorrect }
+            is io.gitlab.arturbosch.detekt.api.MultiRule -> rule.rules.any { it.autoCorrect }
```
chao2zhang, line 100: Is this intended that we want to use the full qualified name for `MultiRule`?

chao2zhang (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
marschwar: Is there any other way (besides suppressing deprecation warning for the entire file) to avoid the build to fail because of a deprecated import?

cortinico:
You have a couple of references inside `RuleProviderSpec` which should also be annotated

cortinico (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
BraisGabin: There is not as far as I know. It would be really good to know if there is any other way.

BraisGabin (approved):

PR #5163, opened 2022-08-01
Use the correct source directory set on JVM

Goooler:
Same fixes borrowed from https://github.com/square/wire/pull/2310.


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektJvm.kt
```diff
@@ -47,4 +44,8 @@
     }
 
     private fun FileCollection.existingFiles() = filter { it.exists() }
+
+    private val SourceSet.kotlinSourceSet: SourceDirectorySet
+        get() = ((this as HasConvention).convention.plugins["kotlin"] as? KotlinSourceSet)?.kotlin
+            ?: project.objects.sourceDirectorySet("empty", "Empty kotlin source set")
```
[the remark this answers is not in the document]
Goooler: 453338904bb67c57e54568fa3c4b051a5bf09543

3flex:
This logic doesn't make sense to me - it seems that in the case of a null source set, a detekt task would still be created, but `source` would be set to an empty value - so we'd be left with a task that will be skipped with a `NO-SOURCE` status.

If we need to filter out null source sets (and it would be good to understand what change has been made in the Kotlin Gradle Plugin that led to the change in behaviour) that should be done in a way that skips registration of a task that won't execute. I'd suggest logging a warning would be a good idea in that case as well.

3flex (changes requested):
This logic doesn't make sense to me - it seems that in the case of a null source set, a detekt task would still be created, but `source` would be set to an empty value - so we'd be left with a task that will be skipped with a `NO-SOURCE` status.

If we need to filter out null source sets (and it would be good to understand what change has been made in the Kotlin Gradle Plugin that led to the change in behaviour) that should be done in a way that skips registration of a task that won't execute. I'd suggest logging a warning would be a good idea in that case as well.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektJvm.kt
```diff
@@ -2,15 +2,19 @@ package io.gitlab.arturbosch.detekt.internal
...
 import org.gradle.api.Project
 import org.gradle.api.file.FileCollection
+import org.gradle.api.file.SourceDirectorySet
 import org.gradle.api.internal.HasConvention
 import org.gradle.api.plugins.JavaPluginConvention
 import org.gradle.api.tasks.SourceSet
 import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
 
 internal class DetektJvm(private val project: Project) {
+    private val SourceSet.kotlin: SourceDirectorySet
```
cortinico, line 14: Is there a reason why this property is uppercase?

cortinico (commented):
Thanks for sending this over 🙏

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektJvm.kt
Goooler: It's lowercase.

cortinico: 🤦 Yeah my bad, it's an extension property

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektJvm.kt
```diff
@@ -4,49 +4,55 @@ package io.gitlab.arturbosch.detekt.internal
...
-            source = kotlinSourceSet.kotlin
-            classpath.setFrom(sourceSet.compileClasspath.existingFiles(), sourceSet.output.classesDirs.existingFiles())
+    private fun Project.registerJvmDetektTask(
+        compilation: KotlinCompilation<KotlinCommonOptions>,
+        extension: DetektExtension,
+        inputSource: FileCollection
+    ) {
+        registerDetektTask(DetektPlugin.DETEKT_TASK_NAME + compilation.name.capitalize(), extension) {
+            setSource(inputSource)
+            classpath.setFrom(inputSource, compilation.compileDependencyFiles)
```
Goooler, line 34: Seems we can merge `DetektJvm` into `DetektMultiplatform`, see https://github.com/Goooler/detekt/pull/3.

cortinico:
@3flex how do you feel about this change?

It looks promising on my end buy I'm lacking a bit of context. @Goooler could you expand on the PR description on why this is needed?

3flex:
I support this change, using the new approach - I'd still like to know what's changing in Kotlin 1.7.20 that means we have to make this change now, but putting that aside, we should do this anyway because:

1. The concept of Conventions is deprecated in Gradle, and KGP will drop them soon enough, requiring changes like this
2. As Goooler pointed out, it should allow consolidation of logic between MPP and Java variants at some point
3. Something like this will be needed for #4465 

I just wish we had better test coverage!

3flex (approved):

cortinico:
> The concept of Conventions is deprecated in Gradle, and KGP will drop them soon enough, requiring changes like this

TIL. Is this documented somehwere?

3flex:
This is probably most relevant:

https://docs.gradle.org/current/userguide/upgrading_version_7.html#all_convention_deprecation

BraisGabin:
This is ready to merge, right? I would vote to merge it for 1.22.0-RC2.

Goooler:
https://github.com/detekt/detekt/actions/runs/3196227698/jobs/5217876383

Seems `org.jetbrains.kotlin.gradle.plugin.mpp.pm20.targets` was removed from Kotlin 1.7.20, I'll fix it later.

PR #5170, opened 2022-08-02
Enable Predictive Test Selection for local builds

### LINKED ISSUE #4992: Enable Predictive Test Selection
## Expected Behavior
Fast builds & test execution.

## Current Behavior
All tests impacted directly or indirectly by any production code changes are always executed. See #4908 which describes some of the issues around this.

## Context
This should reduce CI times and improve developer experience. For more: https://gradle.com/gradle-enterprise-solutions/predictive-test-selection/

3flex:
Closes #4992

Example results: https://ge.detekt.dev/s/zw2mley3hbwz6/tests/overview. The build was run simply with `./gradlew test` on the main branch.

This enables PTS only for local builds. Gradle recommends enabling PTS for local builds and PRs, however it's simpler and lower risk to leave disabled on CI entirely. Maybe we can enable on PRs too once we have some experience with it.

This doesn't apply to detekt-gradle-plugin, which is unfortunate because it's the slowest test suite, but currently included builds are not supported so this is a technical limitation. Config can be updated in future if this changes.

This is an `@Internal` property so the different value in CI & non-CI builds doesn't impact remote build cache hits.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -44,6 +44,14 @@ allprojects {
     }
 }
 
+subprojects {
+    tasks.withType<Test> {
+        predictiveSelection {
+            enabled.set(System.getenv("CI") == null)
+        }
+    }
+}
+
 val analysisDir = file(projectDir)
 val baselineFile = file("$rootDir/config/detekt/baseline.xml")
 val configFile = file("$rootDir/config/detekt/detekt.yml")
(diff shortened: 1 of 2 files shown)


schalkms (approved):
Interesting!
I’m curious how the results including false-positives will be.

3flex:
Me too! This report can be used to monitor: https://ge.detekt.dev/scans/test-selection

I haven't delved into it but in theory that will surface any issues... Though the ML model should take those issues and incorrect misses and improve over time.

schalkms:
Yeah, let’s see how it works out. Later on, we can still enable PTS for PRs too.

cortinico (approved):

BraisGabin (approved):

PR #5084, opened 2022-07-16
[WIP] :construction: Setup danger :construction:

BraisGabin:
:construction: :construction: :construction:

I'm using this PR to test how Danger works and set it up.

:construction: :construction: :construction:

PR #5172, opened 2022-08-03
Configure danger.js

cortinico:
Following up on #5084, I'm setting up Danger.js for the repo. I've added a couple of checks. I've tested some of them manually and I'll tweak them if they don't work properly:

- Warn if the PR contains a functional change without a test
- Handle PRs for new Detekt rules.
- If `default-detekt-config.yml` changes without a rule change, warn the user
- Say thank you to Website edits.
- Warn if the PR has been accepted but has no milestone set.

I'm not sure about the last one as I think we'll have to set a trigger for `issue_comment` to re-run danger on every PR review. I'll test it a bit and see if we need to tweak it


bots/dangerfile.js
```diff
@@ -0,0 +1,67 @@
+import { danger, fail, markdown, warn, message } from "danger";
+
+// API reference: https://danger.systems/js/reference.html
+const pr = danger.github.pr;
+
+// Ideas to implement for Detekt
+// - If new rule: add the new rule label
+// - If new rule: should be enable by default false
```
schalkms, line 8:
```suggestion
// - If new rule: should be enabled by default false
```

schalkms (approved):
Thank you! 
The PR looks good to me! It's a very nice starting point.

BraisGabin (approved):
<3 Nice work!

bots/dangerfile.js
cortinico: That was a stale comment, I've removed it 👍

cortinico:
So I ended up switching to use `pull_request_target`. This allows PRs from forks to have access to a write token. The catch here is that the workflow is executed against the base of the Pull Request (as it's a trusted source). The alternative would have been to handle a token, which exposes to more security issues.

The catch is that if you edit the `dangerfile` you won't actually see the changes till you merge it.

I'll keep an eye on Danger and tweak it if needed 👍

PR #5176, opened 2022-08-03
Don't run rules annotated with `@RequiresTypeResolution` when running plain detekt

BraisGabin:
This is an alternative for #5130. To be more precise this was proposed by @marschwar here: https://github.com/detekt/detekt/pull/5130#issuecomment-1197932497

I implemented this in `:detekt-core` because I think it should be the one that decides if a rule should or should not be runned. But to be more consistent with the current code we could move this check to the `Rule` inside `:detekt-api`. What do you think?

---

~This PR isn't finish yet. I need to remove all the `if (bindingContext == BindingContext.EMPTY) return` that we have in our code now but I want feedback to know if this is what we want.~

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
@@ -115,12 +115,8 @@ internal class Analyzer(
 
         val (correctableRules, otherRules) = activeRuleSetsToRuleSetConfigs
             .flatMap { (ruleSet, _) -> ruleSet.rules.asSequence() }
-            .let { sequence ->
-                if (bindingContext == BindingContext.EMPTY) {
-                    sequence.filterNot { rule -> rule::class.hasAnnotation<RequiresTypeResolution>() }
-                } else {
-                    sequence
-                }
+            .filter { rule ->
+                bindingContext != BindingContext.EMPTY || !rule::class.hasAnnotation<RequiresTypeResolution>()
             }
             .partition { isCorrectable(it) }
 
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
@@ -10,6 +10,7 @@ import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.RuleSetId
 import io.gitlab.arturbosch.detekt.api.RuleSetProvider
 import io.gitlab.arturbosch.detekt.api.internal.CompilerResources
+import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
 import io.gitlab.arturbosch.detekt.api.internal.whichDetekt
 import io.gitlab.arturbosch.detekt.api.internal.whichJava
 import io.gitlab.arturbosch.detekt.api.internal.whichOS
@@ -24,6 +25,7 @@ import org.jetbrains.kotlin.config.languageVersionSettings
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.calls.smartcasts.DataFlowValueFactoryImpl
+import kotlin.reflect.full.hasAnnotation
 
 private typealias FindingsResult = List<Map<RuleSetId, List<Finding>>>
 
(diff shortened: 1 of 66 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -113,6 +115,13 @@ internal class Analyzer(
 
         val (correctableRules, otherRules) = activeRuleSetsToRuleSetConfigs
             .flatMap { (ruleSet, _) -> ruleSet.rules.asSequence() }
+            .let { sequence ->
+                if (bindingContext == BindingContext.EMPTY) {
+                    sequence.filterNot { rule -> rule::class.hasAnnotation<RequiresTypeResolution>() }
+                } else {
+                    sequence
+                }
+            }
```
marschwar, line 124:
```suggestion
            .filter { rule ->
                bindingContext != BindingContext.EMPTY || !rule::class.hasAnnotation<RequiresTypeResolution>()
            }
```

marschwar (commented):
To me this would be a good solution. As it was mentioned your other PR this would still require a custom rule to verify that a rule that is not annotated with `@RequiresTypeResolution` does not in fact access the `BindingContext`.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
BraisGabin: I wrote that and then changed it to avoid checking the same thing over and over again. But probably the simplification is better than the optimization.

schalkms (commented):
I think putting the filter in `core` is a good idea.
I want to get rid of the mixed rule behavior rather sooner than later.
This is for sure a breaking change and we should be careful with activating this new behavior.
To avoid confusion, I'd rather change this PR to a draft in the meantime.

BraisGabin:
I'm moved this to Draft and I removed all the `if (bindingContext == BindingContext.EMPTY) return` that were not needed anymore. But this PR should not be consider a breaking change. This PR doesn't change any behaviour of any rule. The mixed rules doesn't have the `@RequiresTypeResolution` annotation so they are not disabled.

This PR just makes the plain-detekt faster. This should be just the first step for #3577 and it is an alternative over #5130.

BraisGabin:
I'm moving this back to "ready for review" because I don't see how this PR could be a breaking change.

cortinico:
> I don't see how this PR could be a breaking change.

Not technically a breaking change, but more like a behavioural change. We should make clear that `@RequiresTypeResolution` is now considered by detekt-core. Previously was just metadata.

BraisGabin:
`@RequiresTypeResolution` is inside the `internal` package so if someone uses it we shouldn't care too much. I agree, we should note it but it shouldn't be a big deal.

cortinico:
> so if someone uses it we shouldn't care too much

We should. As if a user wrote a rule with type resolution and haven't used the `@RequiresTypeResolution` we won't be running it.

The fact that lives inside the `.internal` package now is a bit concerning, as I believe it should be moved to `io.gitlab.arturbosch.detekt.api` instead (maybe we can deprecate and create a new annotation also).

BraisGabin:
> We should. As if a user wrote a rule with type resolution and haven't used the `@RequiresTypeResolution` we won't be running it.

It is be the other way around. If a third party rule was written and it has the annotation `@RequiresTypeResolution` but it also work without it, then we should not be running it.

To me that seems a strange case, and we should be free to introduce that breaking change because no one should be using our internal api.

And I agree, we should think about moving this annotation and friends to our public api. Because of this PR and the one by @VitalyVPinchuk about creating the config file for third party. But we can do that on other pr.

cortinico:
> It is be the other way around. If a third party rule was written and it has the annotation `@RequiresTypeResolution` but it also work without it, then we should not be running it.
> To me that seems a strange case, and we should be free to introduce that breaking change because no one should be using our internal api.

Yup in this case, it makes sense.
Should we decide to implement the `iif` behavior (do not offer BindingContext to rules without `@RequiresTypeResolution` then we should move the annotation probably (and we can do it safely).

> And I agree, we should think about moving this annotation and friends to our public api. Because of this PR and the one by @VitalyVPinchuk about creating the config file for third party. But we can do that on other pr.

Yup. Still worth a note in the release notes.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -113,6 +115,9 @@ internal class Analyzer(
 
         val (correctableRules, otherRules) = activeRuleSetsToRuleSetConfigs
             .flatMap { (ruleSet, _) -> ruleSet.rules.asSequence() }
+            .filter { rule ->
```
chao2zhang, line 118: May I still ask if we could at least print warnings in console with something like `Rule X is skipped because it has @RequiresTypeResolution but it is run as a plain text. See https://detekt.dev/docs/gettingstarted/type-resolution for more details`?

Even if we could call it out in the release notes, this will probably save us some questions in #detekt channel.

BraisGabin: Yes, that was the idea behind this PR. I just implemented the first draft here: #5226 (tests pendding, that's the reason it's a draft)

BraisGabin: #5226 is ready to review now.

BraisGabin:
Is there something that I need to do here to move this PR forward? I think it's ready to merge but maybe I'm missunderstanding some comments that require me to change something.

schalkms:
@BraisGabin no comments or requests for change from my side

cortinico (approved):

PR #5182, opened 2022-08-04
Implement rule to check the correct usage of `@RequiresTypeResolution`

### LINKED ISSUE #5148: New rule: check for @RequiresTypeResolution when a rule accesses BindingContext
If a detekt rules requires type resolution it needs the appropriate annotation applied. We should have a custom rule for this.

## Context
#5128 and #5145

BraisGabin:
Fixes #5148

This rule detects all the mixed behaviour rules that we have on our project so I added them to the baseline meanwhile we don't fix them. But at least with this rule we ensure that we don't add any new mixed and we know all the current mixed rules too.

---

~I know that the test are failing but I have no idea why the tests are failing. I debugged and~

```kotlin
bindingContext[BindingContext.CLASS, this]
        ?.defaultType
        ?.supertypes()
```

~Only returns a `List` with a single element: `kotlin.Any`. But when I run the rule over the code base of detekt it seems to work. Any idea?~

### DIFF
--- a/config/detekt/baseline.xml
+++ b/config/detekt/baseline.xml
@@ -2,5 +2,20 @@
 <SmellBaseline>
   <ManuallySuppressedIssues/>
   <CurrentIssues>
+    <ID>RequiresTypeResolution:FunctionOnlyReturningConstant.kt$FunctionOnlyReturningConstant : Rule</ID>
+    <ID>RequiresTypeResolution:ImplicitDefaultLocale.kt$ImplicitDefaultLocale : Rule</ID>
+    <ID>RequiresTypeResolution:InstanceOfCheckForException.kt$InstanceOfCheckForException : Rule</ID>
+    <ID>RequiresTypeResolution:LateinitUsage.kt$LateinitUsage : Rule</ID>
+    <ID>RequiresTypeResolution:LongParameterList.kt$LongParameterList : Rule</ID>
+    <ID>RequiresTypeResolution:MapGetWithNotNullAssertionOperator.kt$MapGetWithNotNullAssertionOperator : Rule</ID>
+    <ID>RequiresTypeResolution:MemberNameEqualsClassName.kt$MemberNameEqualsClassName : Rule</ID>
+    <ID>RequiresTypeResolution:OptionalUnit.kt$OptionalUnit : Rule</ID>
+    <ID>RequiresTypeResolution:SpreadOperator.kt$SpreadOperator : Rule</ID>
+    <ID>RequiresTypeResolution:UnnecessaryAbstractClass.kt$UnnecessaryAbstractClass : Rule</ID>
+    <ID>RequiresTypeResolution:UnusedImports.kt$UnusedImports : Rule</ID>
+    <ID>RequiresTypeResolution:UnusedPrivateMember.kt$UnusedPrivateMember : Rule</ID>
+    <ID>RequiresTypeResolution:UseCheckOrError.kt$UseCheckOrError : Rule</ID>
+    <ID>RequiresTypeResolution:UseDataClass.kt$UseDataClass : Rule</ID>
+    <ID>RequiresTypeResolution:UseRequire.kt$UseRequire : Rule</ID>
   </CurrentIssues>
 </SmellBaseline>
--- a/config/detekt/baseline.xml
+++ b/config/detekt/baseline.xml
@@ -2,20 +2,5 @@
 <SmellBaseline>
   <ManuallySuppressedIssues/>
   <CurrentIssues>
-    <ID>RequiresTypeResolution:FunctionOnlyReturningConstant.kt$FunctionOnlyReturningConstant : Rule</ID>
-    <ID>RequiresTypeResolution:ImplicitDefaultLocale.kt$ImplicitDefaultLocale : Rule</ID>
-    <ID>RequiresTypeResolution:InstanceOfCheckForException.kt$InstanceOfCheckForException : Rule</ID>
-    <ID>RequiresTypeResolution:LateinitUsage.kt$LateinitUsage : Rule</ID>
-    <ID>RequiresTypeResolution:LongParameterList.kt$LongParameterList : Rule</ID>
-    <ID>RequiresTypeResolution:MapGetWithNotNullAssertionOperator.kt$MapGetWithNotNullAssertionOperator : Rule</ID>
-    <ID>RequiresTypeResolution:MemberNameEqualsClassName.kt$MemberNameEqualsClassName : Rule</ID>
-    <ID>RequiresTypeResolution:OptionalUnit.kt$OptionalUnit : Rule</ID>
-    <ID>RequiresTypeResolution:SpreadOperator.kt$SpreadOperator : Rule</ID>
-    <ID>RequiresTypeResolution:UnnecessaryAbstractClass.kt$UnnecessaryAbstractClass : Rule</ID>
-    <ID>RequiresTypeResolution:UnusedImports.kt$UnusedImports : Rule</ID>
-    <ID>RequiresTypeResolution:UnusedPrivateMember.kt$UnusedPrivateMember : Rule</ID>
-    <ID>RequiresTypeResolution:UseCheckOrError.kt$UseCheckOrError : Rule</ID>
-    <ID>RequiresTypeResolution:UseDataClass.kt$UseDataClass : Rule</ID>
-    <ID>RequiresTypeResolution:UseRequire.kt$UseRequire : Rule</ID>
   </CurrentIssues>
 </SmellBaseline>
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -279,6 +279,6 @@ libraries:
     includes: ['**/detekt-api/src/main/**/api/*.kt']
 
 ruleauthors:
-  RequiresTypeResolution:
+  ViolateTypeResolutionRequirements:
     active: true
     excludes: ['**/test/**', '**/*Test.kt', '**/*Spec.kt']
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -277,3 +277,8 @@ libraries:
     active: true
     excludes: ['**/*.kt']
     includes: ['**/detekt-api/src/main/**/api/*.kt']
+
+ruleauthors:
+  RequiresTypeResolution:
+    active: true
+    excludes: ['**/test/**', '**/*Test.kt', '**/*Spec.kt']
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -279,6 +279,6 @@ libraries:
     includes: ['**/detekt-api/src/main/**/api/*.kt']
 
 ruleauthors:
-  ViolateTypeResolutionRequirements:
+  ViolatesTypeResolutionRequirements:
     active: true
     excludes: ['**/test/**', '**/*Test.kt', '**/*Spec.kt']
--- a/detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolution.kt
+++ b/detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolution.kt
@@ -14,7 +14,9 @@ import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
 import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
 import io.gitlab.arturbosch.detekt.rules.fqNameOrNull
 import org.jetbrains.kotlin.psi.KtClass
-import org.jetbrains.kotlin.psi.KtElement
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtNameReferenceExpression
+import org.jetbrains.kotlin.psi.KtReferenceExpression
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.types.typeUtil.supertypes
 import kotlin.reflect.KClass
@@ -33,11 +35,12 @@ class RequiresTypeResolution(config: Config = Config.empty) : Rule(config) {
         Debt.FIVE_MINS
     )
 
-    override fun visitClass(klass: KtClass) {
-        super.visitClass(klass)
+    private val klasses: MutableList<KtClass> = mutableListOf()
+    private var usesBindingContext: Boolean = false
 
-        if (klass.extendsFrom(BaseRule::class)) {
-            val usesBindingContext = usesBindingContext(klass)
+    override fun visitKtFile(file: KtFile) {
+        super.visitKtFile(file)
+        klasses.forEach { klass ->
             val isAnnotatedWithRequiresTypeResolution = klass.isAnnotatedWith(RequiresTypeResolution::class)
             if (usesBindingContext && !isAnnotatedWithRequiresTypeResolution) {
                 report(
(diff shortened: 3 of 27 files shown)


marschwar:
> Only returns a List with a single element: kotlin.Any. But when I run the rule over the code base of detekt it seems to work. Any idea?

It seems like that `Rule` and `BaseRule` types are not known to the  [test compiler](https://github.com/detekt/detekt/blob/main/detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KtTestCompiler.kt#L47). I will try to figure out a way to add them.

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolution.kt
```diff
@@ -0,0 +1,81 @@
...
+private fun usesBindingContext(element: KtElement): Boolean {
+    return element.containingKtFile.text.contains("bindingContext", ignoreCase = false)
+}
+
+context(BaseRule) private inline fun <reified T : Any> KtClass.extendsFrom(kClass: KClass<T>): Boolean {
+    return bindingContext[BindingContext.CLASS, this]
+        ?.defaultType
+        ?.supertypes()
+        .orEmpty()
+        .any { it.fqNameOrNull()?.toString() == checkNotNull(kClass.qualifiedName) }
```
marschwar, line 72:
```suggestion
    return bindingContext[BindingContext.CLASS, this]
        ?.getAllSuperclassesWithoutAny()
        .orEmpty()
        .any { checkNotNull(it.fqNameOrNull()).toString() == checkNotNull(kClass.qualifiedName) }
```

detekt-rules-ruleauthors/src/test/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolutionSpec.kt
```diff
@@ -0,0 +1,92 @@
...
+                    bindingContext
+                }
+            }
+        """
+        val findings = rule.compileAndLintWithContext(env, code)
+        assertThat(findings).isEmpty()
+    }
+
+    @Test
+    fun `should not report Rules that doesn't use bindingContext and are annotated`() {
```
marschwar, line 78:
```suggestion
    fun `should report Rules that don't use bindingContext and are annotated`() {
```

detekt-rules-ruleauthors/src/test/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolutionSpec.kt
```diff
@@ -0,0 +1,92 @@
...
+
+    @Test
+    fun `should not report Rules that doesn't use bindingContext and are not annotated`() {
+        val code = """
+            import io.gitlab.arturbosch.detekt.api.Config
+            import io.gitlab.arturbosch.detekt.api.Rule
+            
+            class C(config: Config) : Rule(config) {
+                override val issue = error("I don't care")
+            }
```
marschwar, line 51:
```kotlin
/**
* The bindingContext is not used on purpose
*/
class C(config: Config) : Rule(config) {
    override val issue = error("I don't care")
}
```

This would produce a rule violation at the moment. 

I was thinking about something like this:

```kotlin
private var isAnnotatedWithRequiresTypeResolution: Boolean = false
private var usesBindingContext: Boolean = false

override fun visitClass(klass: KtClass) {
    if (klass.extendsFrom(BaseRule::class)) {
        isAnnotatedWithRequiresTypeResolution = klass.isAnnotatedWith(RequiresTypeResolution::class)
        super.visitClass(klass)
        if (usesBindingContext && !isAnnotatedWithRequiresTypeResolution) {
            // report(...)
        } else if (!usesBindingContext && isAnnotatedWithRequiresTypeResolution) {
            // report(...)
        }
    }
}

    super.visitReferenceExpression(expression)
    usesBindingContext = usesBindingContext
            || (expression is KtNameReferenceExpression && expression.text == "bindingContext")
}
```

BraisGabin:
> It seems like that `Rule` and `BaseRule` types are not known to the  [test compiler](https://github.com/detekt/detekt/blob/main/detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KtTestCompiler.kt#L47). I will try to figure out a way to add them.

I thought the same. But they compile successfully.

detekt-rules-ruleauthors/src/test/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolutionSpec.kt
BraisGabin: Thanks for the snipet! I changed it a bit because maybe the `class` doesn't use it but it's used inside the file (that's the case with the this rule, for example). If the extension function is not inside the file that could create a false negatives/positives. But I think that we don't have any of those.

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolution.kt
BraisGabin: I applied this change partially. I don't like the `checkNotNull(it.fqNameOrNull())` because the function sais that it could return `null` and I just found cases in #5195 where that happens.

I have the second `checkNotNull` because there is no way for a `kClass` to not know its `qualifiedName` in the way this code is written. And I don't want to handle what happen if both are `null`.

BraisGabin:
Ok, now that #5275 is merged the tests of this PR finally pass. But, before merging this I need to merge this one: #5129

BraisGabin:
This should be ready to review :)

config/detekt/baseline.xml
```diff
@@ -2,5 +2,20 @@
...
+    <ID>RequiresTypeResolution:LateinitUsage.kt$LateinitUsage : Rule</ID>
+    <ID>RequiresTypeResolution:LongParameterList.kt$LongParameterList : Rule</ID>
+    <ID>RequiresTypeResolution:MapGetWithNotNullAssertionOperator.kt$MapGetWithNotNullAssertionOperator : Rule</ID>
+    <ID>RequiresTypeResolution:MemberNameEqualsClassName.kt$MemberNameEqualsClassName : Rule</ID>
+    <ID>RequiresTypeResolution:OptionalUnit.kt$OptionalUnit : Rule</ID>
+    <ID>RequiresTypeResolution:SpreadOperator.kt$SpreadOperator : Rule</ID>
+    <ID>RequiresTypeResolution:UnnecessaryAbstractClass.kt$UnnecessaryAbstractClass : Rule</ID>
+    <ID>RequiresTypeResolution:UnusedImports.kt$UnusedImports : Rule</ID>
+    <ID>RequiresTypeResolution:UnusedPrivateMember.kt$UnusedPrivateMember : Rule</ID>
+    <ID>RequiresTypeResolution:UseCheckOrError.kt$UseCheckOrError : Rule</ID>
```
chao2zhang, line 17: Do we need to address these detekt violations in our own rule?

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolution.kt
```diff
@@ -0,0 +1,96 @@
...
+        "`@RequiresTypeResolution` should be used if and only if `bindingContext` is used.",
+        Debt.FIVE_MINS
+    )
+
+    private val klasses: MutableList<KtClass> = mutableListOf()
+    private var usesBindingContext: Boolean = false
+
+    override fun visitKtFile(file: KtFile) {
+        super.visitKtFile(file)
+        klasses.forEach { klass ->
```
chao2zhang, line 44: is it possible to move the checking logic inside `visitClass(klass: KtClass)` to avoid maintaining `klass: MutableList<KtClass>`?

BraisGabin: The problem is that we have rules that uses the `BindingContext` inside private extension functions. So if I move it to `visitClass` we would hava false positives. For that reason I moved it to `visitClass`. I don't like the `MutableList` either but i don't know other way to implement it.

config/detekt/baseline.xml
cortinico: We can also `@Supress` each file maybe? It makes more explicit that the rule is "problematic"

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolution.kt
```diff
@@ -0,0 +1,96 @@
...
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtNameReferenceExpression
+import org.jetbrains.kotlin.psi.KtReferenceExpression
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+import org.jetbrains.kotlin.resolve.descriptorUtil.getAllSuperclassesWithoutAny
+import kotlin.reflect.KClass
+
+/**
+ * If a rule uses `bindingContext` should be annotated with `@RequiresTypeResolution`. And if it doesn't it shouldn't
```
cortinico, line 26:
```suggestion
 * If a rule uses [BindingContext] should be annotated with `@RequiresTypeResolution`. And if it doesn't it shouldn't
```

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolution.kt
```diff
@@ -0,0 +1,96 @@
...
+ * be annotated with it.
+ */
+@ActiveByDefault("1.22.0")
+@RequiresTypeResolution
+class RequiresTypeResolution(config: Config = Config.empty) : Rule(config) {
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Defect,
+        "`@RequiresTypeResolution` should be used if and only if `bindingContext` is used.",
```
cortinico, line 35:
```suggestion
        "`@RequiresTypeResolution` should be used if and only if `BindingContext` is used.",
```

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolution.kt
cortinico: I'm wondering if the check here is too sophisticated and we should just do a syntactic check on whether `BindingContext` or `bindingContext` are present as string in the body of the rule.

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolution.kt
```diff
@@ -0,0 +1,96 @@
...
+import org.jetbrains.kotlin.resolve.descriptorUtil.getAllSuperclassesWithoutAny
+import kotlin.reflect.KClass
+
+/**
+ * If a rule uses `bindingContext` should be annotated with `@RequiresTypeResolution`. And if it doesn't it shouldn't
+ * be annotated with it.
+ */
+@ActiveByDefault("1.22.0")
+@RequiresTypeResolution
+class RequiresTypeResolution(config: Config = Config.empty) : Rule(config) {
```
cortinico, line 31: I would rename this rule to either:
- `ViolateRequiresTypeResolution`
- `ViolateTypeResolutionRequirements`

I think it's confusing that we have an annotation that is called as a rule.
Plus having local suppresses will look a bit odd:

```
@RequiresTypeResolution
@Suppress("RequiresTypeResolution")
class ARule(...)
```

cortinico (commented):
Let's make sure we ship this inside RC2 otherwise we'll end up shipping an empty ruleset :)

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolution.kt
BraisGabin: That was my initial implementation but I got this comment: https://github.com/detekt/detekt/pull/5182#discussion_r939226197 so I made it more sophisticated. I think it's better like it's now.

config/detekt/baseline.xml
BraisGabin: I was thinking of this as a "report" of those rules that have this problem. If I move it to each rule we will lose this "report". But when you are checking that rule you will know that have "problems".

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolution.kt
BraisGabin: Here I'm talking about the `bindingContext` property. No the type `BindingContext`. I rephrased it to be more clear.

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/RequiresTypeResolution.kt
BraisGabin: Same as before

BraisGabin:
@cortinico can you re-review this? I adressed some of the issues you pointed out and comment in other that I'm not 100% sure if they should be implemented.

config/detekt/detekt.yml
```diff
@@ -271,3 +271,8 @@ style:
   WildcardImport:
     active: true
     excludeImports: []
+
+ruleauthors:
+  ViolateTypeResolutionRequirements:
```
cortinico, line 276:
```suggestion
  ViolatesTypeResolutionRequirements:
```

cortinico (approved):
Minor typo in the rule name. Other than that, looks good to me 👍

config/detekt/detekt.yml
chao2zhang: We also have `UseX` as imperative sentences though, so I don't think it is a typo.

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/ViolateTypeResolutionRequirements.kt
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    override fun visitReferenceExpression(expression: KtReferenceExpression) {
+        super.visitReferenceExpression(expression)
+        usesBindingContext = usesBindingContext ||
+            (expression is KtNameReferenceExpression && expression.text == "bindingContext")
+    }
+}
+
+context(BaseRule) private inline fun <reified T : Any> KtClass.extendsFrom(kClass: KClass<T>): Boolean {
```
chao2zhang, line 83: 💯

chao2zhang: ktlint might need to consider formatting `context` - Created a new issue here https://github.com/pinterest/ktlint/issues/1672

chao2zhang (approved):

config/detekt/detekt.yml
cortinico: But we also use `Has` (`HasPlatformType`) which is a plural.

To me the `UseX` rules (like `UseRequire`) should be interpreted in a sentence like "You should use require here".

While the one with plurarls should be interpreted in sentences where the subject is "the code" so like "your code has platform type" or "your code violates type resolution".

Also this is a nit, so feel free to skip it :)

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/ViolateTypeResolutionRequirements.kt
cortinico: Love this!

cortinico:
@BraisGabin can we finalize this as I'd like to prepare RC2 tomorrow 👍

BraisGabin:
I don't have a computer near this weekend. I'm OK with keeping the current name or adding the `s`.

If you want to release tomorrow feel free to change the name yourself.

PR #5160, opened 2022-07-31
Add information about exhaustiveness check to documentation

marschwar:
This documents the exhaustiveness check during config validation (#4873).

Should only be merged if #5089 is merged and released.

PR #5187, opened 2022-08-05
Update the website with versioned docs

### LINKED ISSUE #4168: Move to versioned documentation

cortinico:
Fixes #4168 

There is not much to review here, other than keeping a look at the website from the preview and make sure no link got broken (that is likely if we were using relative paths around).

By default, users on the website will see the "current" docs which are the latest released version (1.21.0 in this case). Users can switch to "next" to see the current status of the documentation.


schalkms (approved):

website/versions.json
```diff
@@ -0,0 +1,3 @@
+[
+  "1.21.0"
```
TWiStErRob, line 2: Is there any implication of this for the release process? e.g. need to update a script where instead of just regenerating, you would need to make a copy-paste of the current stuff for backup or future use?

cortinico: Yup we'll have to call `yarn run docusaurus docs:version <VERSION_NUMBER>` while cutting the branch.

cortinico:
The only thing that I'm unsure are the snippets where we have `[detekt_version]` in the source code. It might be that the version is resolved once for the whole website, so once we release 1.22.0, users on 1.21.0 will see the snippets with the old version.

I'll address this if it's the case in the future.

marschwar:
Is it now safe to revert the revert of #5160 @cortinico ?

cortinico:
> Is it now safe to revert the revert of #5160 @cortinico ?

Yes those files are untouched 👍

cortinico:
> Is it now safe to revert the revert of #5160 @cortinico ?

Heads up: this was never unreverted and is now missing. @marschwar can we re-apply it?

Issue #5192, opened 2022-08-06
Remove all usages of `MutliRule` in detekt code base


marschwar:
We decided to remove the `MultiRule` construct for all the reasons mentioned in #4240. As it is part of the API we cannot remove support for it just yet but we can change all those places where we use it for first party rules.

This issue keeps track of all the tasks necessary.

- [x] Replace `EmptyBlocks` mutli rule (#5208)
- [x] Replace `FileParsingRule` mutli rule (#5193)
- [x] Replace `KDocStyle`mutli rule (#5207)
- [x] Replace `KtLintMultiRule` (#6094)
- [x] Replace `NamingRules`multi rule (#5212)
- [x] Remove multi rule support in config and documentation generation (blocked until multi rule is no longer used in code base) (#6094)
- [ ] Remove `MultiRule` (#6288)

chao2zhang:
KtLintMultiRule seems to be the last multi rule in source set. I will take a stab at it in https://github.com/detekt/detekt/pull/5312

cortinico:
> I will probably still perform the ktlint upgrades first, then refactor off `MultiRule`.

This is still a think, as suggested by @chao2zhang.
The last multirule is the KtLint one

marschwar:
Have you started to work on it @chao2zhang? I could pick it up If you like.

PR #5193, opened 2022-08-06
Remove multi rule FileParsingRule

marschwar:
This is part of #5192.

All rules from the `FileParsingRule` multi rule are now run individually.


schalkms (approved):

cortinico (approved):

schalkms:
@cortinico seems to be an issue with danger and the GitHub rate limit. Is there something to adapt in the setup?

```
Request failed [403]: https://api.github.com/repos/detekt/detekt/pulls/5193
[19](https://github.com/detekt/detekt/runs/7707582017?check_suite_focus=true#step:5:20)
Response: {
[20](https://github.com/detekt/detekt/runs/7707582017?check_suite_focus=true#step:5:21)
  "message": "API rate limit exceeded for installation ID 7815900.",
[21](https://github.com/detekt/detekt/runs/7707582017?check_suite_focus=true#step:5:22)
  "documentation_url": "https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"
[22](https://github.com/detekt/detekt/runs/7707582017?check_suite_focus=true#step:5:23)
}
```

cortinico:
> Is there something to adapt in the setup?

Not really. The problem is with codecov though. For some reason it seems like they post a comment that they edit over and over:

Screenshot 2022-08-06 at 23 43 29

This is generating 200+ invocations of the Danger workflow causing the own Github token to go out of quota. The labeller also failed.
This should be reported to the Codecov App I believe.

I'm looking into potentially skipping the workflow if the event trigger was coming from Codecov

schalkms:
Ouch. 🙈

> This is generating 200+ invocations of the Danger workflow causing the own Github token to go out of quota.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/KtFileContent.kt
```diff
@@ -0,0 +1,7 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import org.jetbrains.kotlin.psi.KtFile
+
+internal data class KtFileContent(val file: KtFile, val content: Sequence<String>)
+
+internal fun KtFile.toFileContent() = KtFileContent(this, text.splitToSequence("\n"))
```
chao2zhang, line 7: ❤️ for internal

chao2zhang (approved):

PR #5195, opened 2022-08-07
Forbid constructors with `ForbiddenMethodCall`

### LINKED ISSUE #5185: Forbid constructor calls

Forbid the usage of a constructor like the same way we have `ForbiddenMethodCall`.

## Context
I'd like to prevent the usage of the `java.math.BigDecimal(double)` constructor. See https://stackoverflow.com/a/7186298 for explanation.

BraisGabin:
Fix #5185

---

I have no idea what's the real difference between `fqNameOrNull` and `fqNameSafe` but with the former all works as expected so 🤷

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
@@ -61,7 +61,8 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
             "(i.e. `java.time.LocalDate(java.time.Clock)`) which would report only call " +
             "with this concrete signature. If you want to forbid an extension function like" +
             "`fun String.hello(a: Int)` you should add the receiver parameter as the first parameter like this: " +
-            "`hello(kotlin.String, kotlin.Int)`"
+            "`hello(kotlin.String, kotlin.Int)`. To forbid constructor calls you need to define them with `<init>` " +
+            " for example `java.util.Date.<init>`."
     )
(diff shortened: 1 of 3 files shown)


cortinico (approved):

schalkms:
Interesting question...
Hmm, maybe you might want to ask this in the Kotlin Slack channel.

> I have no idea what's the real difference between fqNameOrNull and fqNameSafe but with the former all works as expected so 🤷

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
```diff
@@ -61,7 +61,8 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
             "(i.e. `java.time.LocalDate(java.time.Clock)`) which would report only call " +
             "with this concrete signature. If you want to forbid an extension function like" +
             "`fun String.hello(a: Int)` you should add the receiver parameter as the first parameter like this: " +
-            "`hello(kotlin.String, kotlin.Int)`"
+            "`hello(kotlin.String, kotlin.Int)`. To forbid constructor calls you need to define them with `<init>` " +
```
schalkms, line 64:
```suggestion
            "`hello(kotlin.String, kotlin.Int)`. To forbid constructor calls you need to define them with `<init>`, " +
```

schalkms (approved):

schalkms:
Maybe, a hint should be given in the description of this rule that constructors are also supported.

BraisGabin:
> Maybe, a hint should be given in the description of this rule that constructors are also supported.

Agree, what do you think about it now?

PR #5198, opened 2022-08-07
Remove old configurations

BraisGabin:
I'm opening this as a PR but because it's easier than an Issue in this case.

I think that no one uses this configuration any more. It had sense when all/nearly all the rules didn't have type-solving. But right now configurations like that one have little sense.

If you use these configurations or you think it is worth to keep them just close this PR :)


schalkms:
Is the file `statistics.yaml` still referenced somewhere in the code base? I know that a dogfooding approach was used to run detekt's statistic features on the detekt code base. I'm not sure whether we want to keep it.

I think the Gradle config could be removed quite safely.

BraisGabin:
> Is the file `statistics.yaml` still referenced somewhere in the code base? I know that a dogfooding approach was used to run detekt's statistic features on the detekt code base. I'm not sure whether we want to keep it.

I couldn't find any reference to `statistics.yml` (except the one that I removed).

cortinico (approved):

3flex (approved):

PR #5199, opened 2022-08-07
Fix errors with `detektGenerateConfig`

### LINKED ISSUE #4843: `detektGenerateConfig` fails
## Expected Behavior
If I run `./gradlew detektGenerateConfig` the configuration file should be created and gradle should not fail.

## Observed Behavior
The file is created, but gradle fails: https://ge.detekt.dev/s/lpgeopyjrd46k/failure?expanded-stacktrace=WyIwIiwiMC03LTktMTEiXQ&focused-stack-frame=0-7-9-11-0

## Steps to Reproduce
In this project:
```
rm config/detekt/detekt.yml
./gradlew detektGenerateConfig
```

## Context
This is a long standing bug. I thought it was reported but I can't find it.

## Your Environment
* Version of detekt used: 1.20.0 / main
* Version of Gradle used (if applicable): 7.4.2
* Gradle scan link (add `--scan` option when running the gradle task): https://ge.detekt.dev/s/lpgeopyjrd46k
* Link to your project (if it's a public repository): https://github.com/detekt/detekt

BraisGabin:
I recomment to read #4843 to understand all the context for this solution.

TL;DR `detektGenerateConfig` runs in a race condition and tries to create the same file multiple times and crashes. With this solution we ansure that this task is always run in serial to avoid it.

This PR also have 2 commits with little refactors over `DetektGenerateConfigTask`. I didn't want to fall into the rabbit hole so I didn't refactor anymore. But if someone wants: It seems that we are doing more work than needed here. The providers and properties are executed multiple times without, I think, a good reason.

Fixes #4843

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
@@ -19,6 +19,7 @@ import org.gradle.api.tasks.PathSensitive
 import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.TaskAction
 import org.gradle.language.base.plugins.LifecycleBasePlugin
+import java.io.File
 import java.nio.file.Files
 import javax.inject.Inject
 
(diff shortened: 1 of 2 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
```diff
@@ -45,34 +44,37 @@ abstract class DetektGenerateConfigTask @Inject constructor(
...
 
-    private val configurationToUse: ConfigurableFileCollection
+    private val configurationToUse: File
         get() = if (config.isEmpty) {
-            objects.fileCollection().from(defaultConfigPath)
+            defaultConfigPath.toFile()
         } else {
-            config
+            config.last()
         }
```
chao2zhang, line 53: Just to confirm that these changes are mainly for refactoring.

chao2zhang (approved):

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektGenerateConfigTask.kt
BraisGabin: Yes

schalkms (approved):

PR #5203, opened 2022-08-08
Implement rule `UnnecessaryPartOfBinaryExpression`

VovaStelmashchuk:
Implement the UnnecessaryPartOfBinaryExpression. The rule works with all binary expression include if and when condition. The rule also works with all predicates.

### DIFF
--- a/detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryPartOfBinaryExpression.kt
+++ b/detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryPartOfBinaryExpression.kt
@@ -7,16 +7,23 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.com.intellij.psi.tree.IElementType
+import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.KtBinaryExpression
+import org.jetbrains.kotlin.psi.KtElement
+import org.jetbrains.kotlin.psi.KtExpression
 
 /**
  * Unnecessary binary expression add complexity to the code and accomplish nothing. They should be removed.
  * The rule works with all binary expression included if and when condition. The rule also works with all predicates.
+ * The rule verify binary expression only in case when the expression use only one type of the following
+ * operators || or &&.
  *
  * <noncompliant>
  * val foo = true
+ * val bar = true
  *
- * if (foo || foo) {
+ * if (foo || bar || foo) {
  * }
  * </noncompliant>
  *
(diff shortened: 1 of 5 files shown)


detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryPartOfBinaryExpression.kt
```diff
@@ -0,0 +1,47 @@
...
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.psi.KtBinaryExpression
+
+/**
+ * Unnecessary binary expression add complexity to the code and accomplish nothing. They should be removed.
+ * The rule works with all binary expression include if and when condition. The rule also works with all predicates.
```
cortinico, line 14:
```suggestion
 * The rule works with all binary expression included if and when condition. The rule also works with all predicates.
```

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryPartOfBinaryExpression.kt
```diff
@@ -0,0 +1,47 @@
...
+        Severity.Performance,
+        "Detects duplicate condition into binary expression and recommends to remove unnecessary checks",
+        Debt.FIVE_MINS
+    )
+
+    override fun visitBinaryExpression(expression: KtBinaryExpression) {
+        super.visitBinaryExpression(expression)
+        val children = expression.children.map { it.text.replace(Regex("\\s"), "") }
+
+        if (children.size != children.distinct().size) {
```
cortinico, line 43: Are we missing any potential false positives here?

VovaStelmashchuk: No. I can see possibility for true negative case. In case when condition has extra brackets. Like a (foo && (foo)). I did not add a test for the case because detekt has a forbidden brackets rule.

We can change the condition, and compare contents or add some preprocessing to the children remove the && and ||.

Do we need this improve ?

cortinico: > Do we need this improve ?

Not necessarily. The rule "looked" too easy and I wanted to collect some feedback around it. Let's wait for other maintainers' opinions but it should be good to go 👍

cortinico (approved):

marschwar:
Don't get me wrong, but is this really a problem in die wild that needs its own rule?

I know this is not a binary expression but this is essentially the same thing and not detected by the rule.

```kotlin
fun bar() {
    val foo = true
    val baz = false
    if (foo || baz || foo) {
        //TODO    
    }
}
```

I know that it does not make much sense but an `xor` would also be flagged but removing the second  part would not yield the same result.

marschwar:
> @marschwar The rule will report an issue in the line `if (foo || baz || foo) { `

I tried and it does not.

VovaStelmashchuk:
Ok, I added the test case, and fix them.

VovaStelmashchuk:
@marschwar I added your test case and several new tests and rewrite the rule.

BraisGabin:
> Don't get me wrong, but is this really a problem in die wild that needs its own rule?

I have same the impression... Does this rule "borns" from a real problem in your code base? Could you explain your use case for this rule?

BraisGabin (approved):

VovaStelmashchuk:
> > Don't get me wrong, but is this really a problem in die wild that needs its own rule?
> 
> I have same the impression... Does this rule "borns" from a real problem in your code base? Could you explain your use case for this rule?

Yes, I started a rule development because I detect  something like `if (foo || foo)` into own code base.

cortinico:
@BraisGabin are you fine merging this?

BraisGabin:
Yes, I'm :)

VovaStelmashchuk:
The check is red after merge the pull request. It happened because the merge request pipeline executed before new rules applied to the repository.

Do I need to create a new MR for fix the issues ?

Is the currenty pipline for pull request correct ? (Looks like the repo need one more check [is branch are up to date with main])

BraisGabin:
Yes, please, could you create a PR solving that?

And about a new check... Yes, we could have it but I think that we should rethink our pipelines in general, we have far too much. They are overwhelming to me so they could be even scary for a new contributor.

VovaStelmashchuk:
will be done today

cortinico:
> (Looks like the repo need one more check [is branch are up to date with main])

This can be enforced at the Github level (inside the branch protection rule). It would create a lot of friction though as it will require every PR to be rebased before merging.

I'm fine with some failures like this one from time to time as the number of PR we receive is not that high to require extra tooling (like merge queues or other checks).

VovaStelmashchuk:
@BraisGabin fixed https://github.com/detekt/detekt/pull/5280

PR #5207, opened 2022-08-09
Run all rules from KDocStyle multi rule individually

marschwar:
This is part of #5192


schalkms (approved):
Feels good to remove some extra code.

cortinico (approved):

Issue #5204, opened 2022-08-08
New Rule: `UnnecessaryRequireNotNull`


pbreault:
## Expected Behavior of the rule
Would there be any appetite for a new `UnnecessaryRequireNotNull` rule?

This would be similar to [UnnecessaryNotNullOperator](https://detekt.dev/docs/rules/potential-bugs/#unnecessarynotnulloperator) but it would flag calls to [requireNotNull](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/require-not-null.html)  with a non-nullable parameter.

```kotlin
// non compliant code 
var string = "foo"
println(requireNotNull(string))

// compliant code
var string : String? = "foo"
println(requireNotNull(string))
```

## Context
We have seen optional field become required after refactorings, leaving behind unnecessary calls to `requireNotNull`. 
We have built a custom rule to detect these cases and we would like to contribute it back.

3flex:
I think this would be a great addition, and should also check for the same case with `checkNotNull`

schalkms:
I second this. This rule would be a great addition to detekt.
Shall the rule be called `UnnecessaryNotNullCheck`?

@pbreault may I ask you to take a stab at this rule?

pbreault:
I think a single `UnnecessaryNotNullCheck ` rule that encompasses checkNotNull and requireNotNull makes sense. Definitely more future-proof. I'll take a stab at it.

PR #5218, opened 2022-08-11
Add `UnnecessaryNotNullCheck` rule

VirtualParticle:
This adds the `UnnecessaryNotNullCheck` rule as discussed in #5204. It detects usages of `requireNotNull` and `checkNotNull` on non-null values.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -470,9 +470,9 @@ potential-bugs:
     active: true
   UnconditionalJumpStatementInLoop:
     active: false
-  UnnecessaryNotNullOperator:
-    active: true
   UnnecessaryNotNullCheck:
+    active: false
+  UnnecessaryNotNullOperator:
     active: true
   UnnecessarySafeCall:
     active: true
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -472,6 +472,8 @@ potential-bugs:
     active: false
   UnnecessaryNotNullOperator:
     active: true
+  UnnecessaryNotNullCheck:
+    active: true
   UnnecessarySafeCall:
     active: true
   UnreachableCatchBlock:
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessaryNotNullCheck.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessaryNotNullCheck.kt
@@ -1,6 +1,7 @@
 package io.gitlab.arturbosch.detekt.rules.bugs
 
 import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Debt
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
@@ -32,7 +33,7 @@ import org.jetbrains.kotlin.types.typeUtil.nullability
 class UnnecessaryNotNullCheck(config: Config = Config.empty) : Rule(config) {
 
     override val issue = Issue(
-        "UnnecessaryNotNullOperator",
+        "UnnecessaryNotNullCheck",
         Severity.Defect,
         "Unnecessary not-null check detected.",
         Debt.FIVE_MINS,
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessaryNotNullCheck.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessaryNotNullCheck.kt
@@ -1,0 +1,61 @@
+package io.gitlab.arturbosch.detekt.rules.bugs
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Debt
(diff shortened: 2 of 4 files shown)


detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessaryNotNullCheck.kt
```diff
@@ -0,0 +1,61 @@
...
+        Debt.FIVE_MINS,
+    )
+
+    override fun visitCallExpression(expression: KtCallExpression) {
+        super.visitCallExpression(expression)
+
+        if (bindingContext == BindingContext.EMPTY) return
+
+        val callName = expression.getCallNameExpression()?.text
+        if (callName == "requireNotNull" || callName == "checkNotNull") {
```
chao2zhang, line 48: Since we have binding context present here, I would recommend match the full qualified name. You may find `UseCheckNotNull` as a good example

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessaryNotNullCheck.kt
```diff
@@ -0,0 +1,61 @@
...
+ * println(requireNotNull(string))
+ * </compliant>
+ */
+@RequiresTypeResolution
+class UnnecessaryNotNullCheck(config: Config = Config.empty) : Rule(config) {
+
+    override val issue = Issue(
+        "UnnecessaryNotNullCheck",
+        Severity.Defect,
+        "Unnecessary not-null check detected.",
```
chao2zhang, line 38:
```suggestion
        "Remove unnecessary not-null checks on non-null types.",
```
I would recommend using a more elaborative description since our description has been historically vague.

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -470,6 +470,8 @@ potential-bugs:
     active: true
   UnconditionalJumpStatementInLoop:
     active: false
+  UnnecessaryNotNullCheck:
```
chao2zhang, line 473: I don't think using `UnnecessaryNotNullCheck` would be catching error-prone bugs. What do you think about `style` rules instead of `potential-bugs` section (Define the rule in `:detekt-rules-style` instead of `:detekt-rules-errorprone)?

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessaryNotNullCheck.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        val callName = expression.getCallNameExpression()?.text
+        if (callName == "requireNotNull" || callName == "checkNotNull") {
+            val type = (expression.valueArguments[0].lastChild as KtExpression).getType(bindingContext)
+            if (type?.nullability() == TypeNullability.NOT_NULL) {
+                report(
+                    CodeSmell(
+                        issue = issue,
+                        entity = Entity.from(expression),
+                        message = "`${expression.text}` contains an unnecessary `$callName`",
```
chao2zhang, line 55: Nitpick: Could we improve the message by
`Using `$callName` on non-null `calleeExpression` is unnecessary` to make it more clear to user?

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessaryNotNullCheckSpec.kt
```diff
@@ -0,0 +1,129 @@
...
+                }
+            """.trimIndent()
+            val findings = subject.lintWithContext(env, code)
+            assertThat(findings).hasSize(1)
+            assertThat(findings).hasTextLocations(16 to 58)
+        }
+    }
+
+    @Nested
+    inner class `check valid not null check usage` {
```
chao2zhang, line 68: Do you think it's worth adding one test case in this PR?
- Platform type - Similar to `shouldDetectWhenCallingPrimitiveJavaMethod` but on a non-primitive
- Definitely not null generic type. [Doc](https://kotlinlang.org/docs/whatsnew17.html#stable-definitely-non-nullable-types`)

If this sounds more expensive, we could definitely follow up.

chao2zhang (commented):
💯  on adding this new rule. Thank you for your contribution

detekt-core/src/main/resources/default-detekt-config.yml
VirtualParticle: It looks like similar rules like `UnnecessaryNotNullOperator` and `UnnecessarySafeCall` are in `potential-bugs` as well. I think the reason for that is probably that treating a non-null value as nullable shows a potential misuse of code and while not terribly likely to cause bugs, it still could. For example, consider the following:

```kotlin
fun getSomeValue(id: Int): Int {
    // ...
}

// somewhere else
val value = requireNotNull(getSomeValue(0)) { "Value not found" }
```

This use of `requireNotNull` is probably assuming that `getSomeValue` returns `null` when it can't find the value, but we can see that isn't the case; it could be returning `-1` or even throw an exception. The former might and the latter definitely would cause a bug as the exception wouldn't be handled.

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessaryNotNullCheckSpec.kt
VirtualParticle: I already have the first one with `shouldIgnoreWhenCallingObjectJavaMethod`, I'll add the second one.

BraisGabin (approved):
Nice one! Thanks

PR #5226, opened 2022-08-16
Warn about enabled rules that are not going to run

### LINKED ISSUE #3577: Detect type resolving rules run without type resolution classpath set

BraisGabin:
Closes #3577

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
@@ -52,6 +52,9 @@ internal class Analyzer(
             } else {
                 runSync(ktFiles, bindingContext, compilerResources)
             }
+        if (bindingContext == BindingContext.EMPTY) {
+            warnAboutEnabledRequiresTypeResolutionRules(settings::info)
+        }
 
         val findingsPerRuleSet = HashMap<RuleSetId, List<Finding>>()
         for (findings in findingsPerFile) {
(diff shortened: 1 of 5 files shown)


schalkms (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -140,6 +143,19 @@ internal class Analyzer(
 
         return result
     }
+
+    private fun warnAboutEnabledRequiresTypeResolutionRules(log: (String) -> Unit) {
+        providers.asSequence()
+            .map { it to config.subConfig(it.ruleSetId) }
+            .filter { (_, ruleSetConfig) -> ruleSetConfig.isActive() }
+            .map { (provider, ruleSetConfig) -> provider.instance(ruleSetConfig) to ruleSetConfig }
```
chao2zhang, line 151: I realized that this is going to violate the goal of https://github.com/detekt/detekt/pull/3555/ that `RuleSetProvider.instance` should be invoked only once. Could we refactor this into `analyze()` function.

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/ProcessingSettingsFactory.kt
```diff
@@ -18,16 +19,36 @@ fun createProcessingSettings(
     inputPath: Path? = null,
     config: Config = Config.empty,
     reportPaths: Collection<ReportsSpec.Report> = emptyList(),
-    spec: ProcessingSpec = createNullLoggingSpec {
+    outputChannel: PrintStream = NullPrintStream(),
```
chao2zhang, line 22: 💯

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
BraisGabin: `RuleSetProvider.instance` isn't invoked once right now. It's invoked n times were n is the number of files that we analice. This implementation make that it's called `n+1`.

If I refactor this into `analyce()` intead of reporting this once I would report it once per file. I think it's better like it is right now but maybe I'm loosing something.

I would love to call `RuleSetProvider.instance` only once but that's a completely different story.

chao2zhang: > If I refactor this into analyze() instead of reporting this once I would report it once per file.
Good catch.

chao2zhang (approved):

PR #5263, opened 2022-08-31
Enable NestedScopeFunctions with custom functions

dzirbel:
This PR demonstrates that when a custom value is specified for the functions
configuration property, there are many false positives from totally unrelated
function calls. This doesn't seem to happen when `functions` is unset in the
configuration, and also seems to be somewhat unreliable in the false positives
it generates.


BraisGabin:
Thanks for a way to reproduce the issue. Now that we found a solution I'm closing this PR ;)

Issue #3396, opened 2021-01-21
Getting "io.gitlab.arturbosch.detekt.api.RuleSetProvider: Error reading configuration file" error sometimes


sergeys-opera:
We are getting this error from time to time when running `./gradlew detekt`:
```
1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':base:detekt'.
> io.gitlab.arturbosch.detekt.api.RuleSetProvider: Error reading configuration file
* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
==============================================================================
```

## Expected Behavior
No error

## Observed Behavior
The build stops with an error. Running `./gradlew detekt --no-daemon` solves the issue (but this is not good).

## Steps to Reproduce
Just run `./gradlew detekt` in a Gradle project with daemon enabled (we also use parallel builds if that matters)

## Context
I have `./gradlew detekt` as a pre-push git hook and want it to fail only when there are errors found by detekt (not by detekt failing itself). Adding `--no-daemon` is a waste of resources. 

## Your Environment

* Version of detekt used:1.14.2
* Version of Gradle used (if applicable): 6.8
* Operating System and version: Linux Mint 20.1 Cinnamon
* Link to your project (if it's a public repository): private

sergeys-opera:
Seems related to https://github.com/detekt/detekt/issues/2411 and https://github.com/detekt/detekt/issues/2582 but they are closed.

chao2zhang:
@sergeys-opera Is it possible for you to share the exception stacktrace with `./gradlew detekt --stacktrace` and also the redacted detekt gradle configuration if possible?

cortinico:
On which JVM are you running? Would you be able to try a different one and let us know if the error happens consistently?

sergeys-opera:
> Is it possible for you to share the exception stacktrace with ./gradlew detekt --stacktrace and also the redacted detekt gradle configuration if possible?
Yes, but I need to reproduce the error first. Will let you updated.

> On which JVM are you running?
```
$ java -version
openjdk version "1.8.0_275"
OpenJDK Runtime Environment (build 1.8.0_275-8u275-b01-0ubuntu1~20.04-b01)
OpenJDK 64-Bit Server VM (build 25.275-b01, mixed mode)
```

chao2zhang:
[...]

PR #5269, opened 2022-09-02
Add toolchains support to Gradle plugin

### LINKED ISSUE #4120: Support Toolchains for JVM projects

3flex:
This will need some tweaking to make sure it works properly with incremental builds and build cache.

This toolchain support does what the Kotlin Gradle Plugin does, but for detekt. If a toolchain is configured for the project (either using [`java`](https://docs.gradle.org/current/userguide/toolchains.html#sec:consuming) or [`kotlin`](https://kotlinlang.org/docs/gradle.html#setting-jdk-version-with-the-task-dsl)) that will then set both `jdkTarget` and `jdkHome` defaults on any `Detekt` or `DetektCreateBaseline` tasks.

See [here](https://blog.jetbrains.com/kotlin/2021/11/gradle-jvm-toolchain-support-in-the-kotlin-plugin/#Toolchain_support_in_the_Kotlin_plugin) for more on the KGP design goals.

> In Kotlin, toolchain support affects only the Kotlin/JVM -jdk-home option’s value and additionally sets the -jvm-target value if it was not set explicitly by the user.

Notably this means that detekt itself will still run on the JDK that Gradle runs on - if Gradle is running on Java 8, then so will detekt. This means that despite [what I said](https://github.com/detekt/detekt/issues/3396#issuecomment-927073465), I expect this PR will have no effect on #3396.

Closes #4120

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
@@ -194,6 +194,13 @@ class CliArgs {
     var jvmTarget: String = JvmTarget.DEFAULT.description
 
     @Parameter(
+        names = ["--jdk-home"],
+        description = "EXPERIMENTAL: Use a custom JDK home directory to include into the classpath",
+        converter = PathConverter::class
+    )
+    var jdkHome: Path? = null
+
+    @Parameter(
         names = ["--version"],
         description = "Prints the detekt CLI version."
     )
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
@@ -68,6 +68,7 @@ internal fun CliArgs.createSpec(output: Appendable, error: Appendable): Processi
             jvmTarget = args.jvmTarget
             languageVersion = args.languageVersion?.versionString
             classpath = args.classpath?.trim()
+            jdkHome = args.jdkHome
         }
     }
 }
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
@@ -33,7 +33,8 @@ internal class EnvironmentFacade(
             projectSpec.inputPaths.toList(),
             classpath,
             compilerSpec.parseLanguageVersion(),
-            compilerSpec.parseJvmTarget()
+            compilerSpec.parseJvmTarget(),
+            compilerSpec.jdkHome,
         )
         createKotlinCoreEnvironment(compilerConfiguration, disposable)
     }
--- a/detekt-gradle-plugin/settings.gradle.kts
+++ b/detekt-gradle-plugin/settings.gradle.kts
@@ -2,6 +2,8 @@ pluginManagement {
     includeBuild("../build-logic")
 }
 
+includeBuild("..")
+
 dependencyResolutionManagement {
     versionCatalogs {
         create("libs") {
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -19,6 +19,7 @@ import io.gitlab.arturbosch.detekt.invoke.DetektInvoker
 import io.gitlab.arturbosch.detekt.invoke.DisableDefaultRuleSetArgument
 import io.gitlab.arturbosch.detekt.invoke.FailFastArgument
 import io.gitlab.arturbosch.detekt.invoke.InputArgument
+import io.gitlab.arturbosch.detekt.invoke.JdkHomeArgument
 import io.gitlab.arturbosch.detekt.invoke.JvmTargetArgument
 import io.gitlab.arturbosch.detekt.invoke.LanguageVersionArgument
 import io.gitlab.arturbosch.detekt.invoke.ParallelArgument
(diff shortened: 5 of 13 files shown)


BraisGabin:
I'm not sure what this PR will provide to our users. Let me explain what I think it will provide and you tell me if I'm wrong.

If we run detekt-cli defining `--jdk-home` the creation of `BindingContext` will have the information related with the classes inside the JDK. For example: `java.time.Instant`. But, weren't we adding those classes already?

And then, if we run detekt using the plugin the `--jdk-home` will be informed automatically taking that information from the configuration of the kotlin gradle plugin. For that reason we don't need to add any new configuration to out plugin, right?

And your last sentence means that the `--jdk-home` only works to provide classes information. That jdk will not be used to execute anything, right?

3flex:
> weren't we adding those classes already?

There's special handling in the Kotlin compiler that takes care of it. detekt calls this to make sure it's configured: https://github.com/JetBrains/kotlin/blob/v1.7.10/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/config/JvmContentRoots.kt#L91-L111

The problem is, if JDK home isn't provided, the compiler will use the JRE it's currently running in to populate the JDK class roots. When detekt is run from Gradle, that will be the JDK Gradle is running under. If the Kotlin compilation is using a different JDK version set through a toolchain, there's a mismatch between detekt's JDK home and the Kotlin compilation JDK home, which affects detekt's ability to analyse the project correctly.

> if we run detekt using the plugin the --jdk-home will be informed automatically taking that information from the configuration of the kotlin gradle plugin. For that reason we don't need to add any new configuration to out plugin, right?

Are you asking if the value will be set by default? Yes, it will be - if there's a toolchain setup. If not, detekt will run using the JDK that Gradle is running under.

> That jdk will not be used to execute anything, right?
Correct. detekt will still run using the JDK that Gradle runs.

---

For an example, you can run Gradle with Java 8, add `java.util.Optional.stream` (added in Java 9) to `ForbiddenMethodCall`, and setup a toolchain with:
```kotlin
java {
    toolchain {
        languageVersion.set(JavaLanguageVersion.of(11))
    }
}
```

Then call the forbidden method somewhere in the project. `assemble` works, because Java 11 is applied to the compilation using the toolchain. `detektMain` works, but analysis is not correct as it uses Java 8 to setup the JDK roots, so it throws these warnings:
```
\C:\Users\3flex\IdeaProjects\toolchains-demo\src\main\kotlin\Main.kt:7:18: error: type mismatch: inferred type is Unit but Stream<out TypeVariable(R)!>! was expected
        .flatMap { it.stream() }
                 ^
\C:\Users\3flex\IdeaProjects\toolchains-demo\src\main\kotlin\Main.kt:7:23: error: unresolved reference: stream
        .flatMap { it.stream() }
                      ^
The BindingContext was created with 2 issues. Run detekt with --debug to see the error messages.
```

The forbidden method call is not picked up as an error.

When using this PR, and the toolchain's JDK path passed to detekt, it works correctly:

```
C:\Users\3flex\IdeaProjects\toolchains-demo\src\main\kotlin\Main.kt:7:23: The method `java.util.Optional.stream` has been forbidden in the detekt config. [ForbiddenMethodCall]
```

Gradle is still running Java 8, so detekt itself also runs on Java 8 but now performs analysis with JDK 11 passed as jdkHome, so the issue is flagged as expected.

BraisGabin (approved):

BraisGabin:
Thanks for the explanation. All clear now :)

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -193,6 +193,13 @@ class CliArgs {
     )
     var jvmTarget: String = JvmTarget.DEFAULT.description
 
+    @Parameter(
+        names = ["--jdk-home"],
+        description = "EXPERIMENTAL: Use a custom JDK home directory to include into the classpath",
+        converter = PathConverter::class
+    )
```
cortinico, line 200: This needs an update to the CLI page on the website as well

cortinico (approved):
Code looks great to me 👍

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
3flex: On this page https://detekt.dev/docs/gettingstarted/cli? Isn't that auto generated now? Or have I misunderstood...

That said, it needs to be added to the Grade docs since they're manually maintained, will do that before merging.

cortinico: > Isn't that auto generated now?

Yes you're right 👍 I forgot we autogenerate that 🚀

PR #4684, opened 2022-04-05
NonBooleanPropertyPrefixedWithIs: Allow boolean function reference

kvn-stgl:
This PR will allow boolean function reference properties for the NonBooleanPropertyPrefixedWithIs rule. (And fix issue #4675).
It also fixes a little issue that unknown types are not null, the string is just empty. This should also fix issue #4674.

PR #5285, opened 2022-09-08
NonBooleanPropertyPrefixedWithIs: Allow boolean functions

brittanyberlanga:
This change builds on top of #4684 which allowed functions with a boolean return type to comply with the `NonBooleanPropertyPrefixedWithIs` rule. That change only allowed functions with zero parameters. With this change, I’m proposing that we open up the exception to any function with a boolean return type.

For example, something like the following would be considered compliant.
```
interface Business {
    val isOpenAt: (DateTime) -> Boolean
}
```

Let me know if there are any reservations around this change. My team found that it was odd that there was this limitation and I thought I’d check out if you all agree as well.


schalkms (approved):
Thank you for this awesome contribution! 🙂

cortinico (approved):

PR #5287, opened 2022-09-08
Smoke test Kotlin 1.7.20-RC

3flex:


3flex:
Looks good... Should be an easy update to the final release when it's out.

PR #5212, opened 2022-08-10
Remove rule from NamingRules multi rule

VovaStelmashchuk:
This is part of #5192

PR #5292, opened 2022-09-10
Fix main

BraisGabin:
Main is failing because of missing `.trimIndentation()`.


BraisGabin:
Edit: I thoght it was just missing `trimIndentation()` but somehow we introduced a regression with #5212. The second commit fixes that regression.

cortinico (approved):

PR #5297, opened 2022-09-14
Fat-Jar version of detekt-generator module

VitalyVPinchuk:
To finalize https://github.com/detekt/detekt/pull/5080 

To be able to call `detekt-generator` through `java -jar detek-generator.jar --other-options` shadowJar plugin was used in build.gradle.kts.

Docs on usage were also added.


cortinico:
I think we need to add the .jar also here:

https://github.com/detekt/detekt/blob/15c19aaf07f19b180180fe5344815d9dca13fada/build-logic/src/main/kotlin/releasing.gradle.kts#L38-L44

BraisGabin (approved):

cortinico (approved):

PR #5307, opened 2022-09-15
Add new rule: use named argument for constants

VovaStelmashchuk:

For more information, see the [CONTRIBUTING guide](https://github.com/detekt/detekt/blob/main/.github/CONTRIBUTING.md).


cortinico:
I suppose you're still working on this @VovaStelmashchuk right?
If so, you have a red CI that needs to be addressed.

BraisGabin:
🤔 Why do you only care about constant values? I mean, something like this seems risky too, right?

```kotlin
Fun Foo(world: String, hello: String) 

foo(hello(), world())
```

VovaStelmashchuk:
@BraisGabin Yes, the function case is risk too. I will resench the way to improve the rule or create a separate rule.

PR #5313, opened 2022-09-18
EndOfSentenceFormat: fix HTML tag heuristic

TWiStErRob:
While doing #5311 I noticed that the code doesn't check for "starts with"


TWiStErRob:
cc @Mauin

schalkms (approved):

chao2zhang (approved):

PR #5315, opened 2022-09-18
MaxLineLength: raw typo and test cleanup

### LINKED ISSUE #5314: MaxLineLength highlights a strange part of raw strings with escape sequences (Regex)
## Expected Behavior
The whole line is highlighted.

## Observed Behavior
image

## Steps to Reproduce
Based on the real example above, I minified a test-case:
MaxLineLengthSpec:
```kotlin
    @Test
    fun repro() {
        val rule = MaxLineLength(
            TestConfig(
                MAX_LINE_LENGTH to "30",
                EXCLUDE_RAW_STRINGS to "false",
            )
        )

        rule.visitKtFile(compileContentForTest(
            """
                // some other content
                val x = Regex(${"\"\"\""}
                    Text (.*?)\(in parens\) this is too long to be valid.
                    The regex/raw string continues down another line.
                ${"\"\"\""}.trimIndent())
                // that is the right length
            """.trimIndent()
        ))
        assertThat(rule.findings).hasSize(2)
    }
```
I also debugged it to see why it's wrong (breakpoint at line 65 in MaxLineLength / `if (ktElement != null)`):
(`ktElement` is the found "parent", the grey-highlighted line in the debugger)
image

## Context and investigation
I was fixing Detekt issues on a project when I found this. It baffled me how it's only highlighting the half of the line. After debugging (see above) I'm still baffled, but I can see the escape sequences are separate and there's no common parent. I also found #5171 by @BraisGabin which hides this issue by default (in 1.22). At this point I'm inclined to say this is a non-issue (consider very few people will observe it), but I wanted to report it in case you think otherwise. At the moment the only possible fix I see is to ignore the found "parent" in raw strings and explicitly highlight the entire line.

Note: this could be an improvement even when there's a parent found, because it's possible it could just highlight `val` for example if that line is too long, which is also strange:
image

## Your Environment

* Link to your project (if it's a public repository): https://github.com/TWiStErRob/net.twisterrob.gradle/blob/1b632fd6e2b13fa9e871a7f46bdb66591e9290e8/quality/src/main/kotlin/net/twisterrob/gradle/quality/report/html/model/LintMessageDetailsSplitter.kt#L56-L62

TWiStErRob:
A typo and simple consistency improvements while investigating #5314. 

(Most of the changes were inspired by similar changes done in #5171 by @BraisGabin.)

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLength.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLength.kt
@@ -45,7 +45,7 @@ class MaxLineLength(config: Config = Config.empty) : Rule(config) {
     @Configuration("if comment statements should be ignored")
     private val excludeCommentStatements: Boolean by config(false)
 
-    @Configuration("if comment statements should be ignored")
+    @Configuration("if raw strings should be ignored")
     private val excludeRawStrings: Boolean by config(true)
 
     override fun visitKtFile(file: KtFile) {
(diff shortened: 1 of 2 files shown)


3flex (approved):

3flex:
nit: in the second commit message the wording is a little misleading, it sounds like both mapOf and trailing commas are being removed, when actually it's only mapOf being removed, and trailing commas were added.

3flex:
Also be careful with the PR description, it says "does not fix #5314" but GitHub only sees "fix #5314" and will (incorrectly) close that issue when this is merged.

TWiStErRob:
Ah, fair about the commit message, missed a word there, I was probably trying to be concise :)
Re GitHub closing "does not fix" - amazing... that feature is really simple then, I'll keep it in mind.

PR #5312, opened 2022-09-18
Upgrade ktlint to 0.47.1

chao2zhang:
Fixes https://github.com/detekt/detekt/issues/5281

Release note: 
- https://github.com/pinterest/ktlint/releases/tag/0.47.1
- https://github.com/pinterest/ktlint/releases/tag/0.47.0

Majority of the changes are:
- Rule lifecycle hooks / deprecate RunOnRootOnly visitor modifier
- Split rule trailing-comma into trailing-comma-on-call-site and trailing-comma-on-declaration-site (https://github.com/pinterest/ktlint/pull/1555)
- Fix multiline if-statements multiline-if-else (https://github.com/pinterest/ktlint/issues/828)

PR #5152, opened 2022-07-31
Get Android variant compile classpath from compileConfiguration

3flex:
Fixes #5150

We don't have good functional test coverage for Android so we're not guarded against introducing new issues. However the implementation of `getCompileConfiguration()` calls `getVariantData().getVariantDependencies().getCompileClasspath()`, so without digging further into AGP internals this should provide equivalent results to the previous `variant.getCompileClasspath(null)` method.

PR #5319, opened 2022-09-19
Prepare Detekt 1.22.0-RC1

cortinico:
Time to cut a new release candidate 🎉

### DIFF
--- a/build-logic/src/main/kotlin/Versions.kt
+++ b/build-logic/src/main/kotlin/Versions.kt
@@ -1,6 +1,6 @@
 object Versions {
 
-    const val DETEKT: String = "1.21.0"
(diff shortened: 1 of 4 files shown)


3flex:
We should definitely include these in RC for early testing:
* #5312
* #5163
* #5152

I don't think there are blockers on any of them so probably just a matter of merging.

Would be good to get #4128 in too, to close it out, but can wait for 1.23.

cortinico:
> I don't think there are blockers on any of them so probably just a matter of merging.

They'll end up in RC2 as RC1 was already released (otherwise this PR can't be created).
Agree that the process is not straightforward (you should prepare the release, merge the PR and trigger the publish process). I'm working on it :)

BraisGabin (approved):

PR #5320, opened 2022-09-19
Simplify release script by replacing "or exit" with "always exit"

TWiStErRob:
Note: GitHub actions actively turns this on as well for all unix shell types:

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

(I have a feeling it could be even put in `#!/usr/bin/env sh -e`, but not sure what would happen if someone did `bash release.sh`.)

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,8 +1,9 @@
 #!/usr/bin/env sh
-gradle publishToMavenLocal || exit
-gradle build || exit
-gradle publishAllToMavenCentral --max-workers 1 || exit
-gradle :detekt-gradle-plugin:publishPlugins || exit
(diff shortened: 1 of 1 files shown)


cortinico (approved):

cortinico:
Thansk for the fix @TWiStErRob 
Just a heads up that we don't actually run that script but we invoke commands manually, specifically because you can't `githubRelease` before the PR with the changelog is merged.

TWiStErRob:
Hmm, then it helps more :) Because it doesn't close your terminal when there was an error 😅, so you can read the message.

schalkms (approved):
Cool 😎

PR #5332, opened 2022-09-22
CanBeNonNullable: explain why the rule does what it does.

### LINKED ISSUE #5331: CanBeNonNullable false positive
## Steps to Reproduce
```kotlin
import javax.xml.stream.XMLStreamWriter

inline fun XMLStreamWriter.attribute(name: String, value: Any) {
	writeAttribute(name, value.toString())
}

inline fun <T : Any> XMLStreamWriter.optionalAttribute(
	name: String,
	value: T?,
	transformation: (T) -> String
) {
	value?.let { attribute(name, transformation(it)) }
}
```

## Expected Behavior
No flagging.

## Observed Behavior
```
XMLStreamWriterDSL.kt:80:70: The nullable parameter 'value' can be made non-nullable. [CanBeNonNullable]
```

## Context
The whole point of this method is to hide the nullability in a named nice fun. The method does the null check, so that it can be just called, and the lambda doesn't have to deal with a null-check. It allows this change:
image
which actually fixes `NestedScopeFunctions`.

I'm not entirely sure this is a false positive, I could be mis-interpreting what `CanBeNonNullable` does.

## Your Environment

* Operating System and version: Windows
* Link to your project (if it's a public repository): code is not live yet.

TWiStErRob:
Based on conversation in https://github.com/detekt/detekt/issues/5331 (closes #5331), @BraisGabin, I borrowed your thoughts.

The original explanation described what the rule does, but gave no clarity on why.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt
@@ -62,6 +62,14 @@ import org.jetbrains.kotlin.types.isNullable
  * This rule inspects variables marked as nullable and reports which could be
  * declared as non-nullable instead.
  *
+ * It's preferred to not have functions that do "nothing".
+ * A function that does nothing when the value is null hides the logic,
+ * so it should not allow null values in the first place.
+ * It is better to move the null checks up around the calls,
+ * instead of having it inside the function.
+ *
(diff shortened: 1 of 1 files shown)


BraisGabin (approved):

BraisGabin:
Thank you! This kind of improvement are very valuable and no that common.

TWiStErRob:
> Thank you! This kind of improvement are very valuable and no that common.

Just need to get into the habit of copy-pasting your comments to the rule kdoc 😉

(It's much simpler to explain something to someone else, than sitting down and writing docs.)

Issue #5316, opened 2022-09-18
LabeledExpression to highlight only label


TWiStErRob:
Currently the whole expression is highlighted, creating a lot of noise in the code:
![image](https://user-images.githubusercontent.com/2906988/190927289-b0e7f987-12b0-4981-ad95-c00c9294b18d.png)

## Expected Behavior of the rule
I think it would be more than enough to highlight the label:
![image](https://user-images.githubusercontent.com/2906988/190927312-2f0ccd7c-30da-4029-a955-8f2fde85fef0.png)

Similar accepted requests: https://github.com/detekt/detekt/issues/4749, https://github.com/detekt/detekt/issues/5310

## Context
Warnings exists in some codebases, and that's ok, but then they make the code harder to read, then it becomes a problem. A warnings should highlight where the problem is, specifically. To draw a parallel: the duplicate code inspection in IDEA looks like this:
![image](https://user-images.githubusercontent.com/2906988/190927355-4aa005cf-8461-4b54-97ee-9eb06c8a5962.png)
Luckily this screenshot also shows that the huge warning hides another warning, see the highlighted "Check" string. I wouldn't have noticed that in the current behavior, only after editing the code / fixing the issue and you never know which severity these hidden other warnings would be, it could've been a NPE, not a duplicate string.

PR #5333, opened 2022-09-22
Implement MisusedAlso Rule

### LINKED ISSUE #518: Rule: detect 'also' library function misuse instead of 'apply'
Lurking codebases came across this kind of pattern (example is obfuscated to not to mess with details):
```kotlin
inline fun foo(block: Buzz.() -> Unit): Buzz =
    Buzz().also {
      it.init()
      it.block()
    }
```
Which is misused `also` pattern and surely can be simplified to:
```kotlin
inline fun foo(block: Buzz.() -> Unit): Buzz =
    Buzz().apply {
      init()
      block()
    }
```
The suggested rule is _"when an `also` block contains only it-started expressions, simplify it to the `apply` block"_

Probably, this pattern also applies to the `let` - `run` pair, although I could not get with the real-world example, so this is the mirrored one:
```kotlin
inline fun foo(block: Buzz.() -> T): T =
    Buzz().let {
      it.init()
      it.block()
    }
```
Correct usage:
```kotlin
inline fun foo(block: Buzz.() -> T): T =
    Buzz().run {
      init()
      block()
    }
```

Tommyten:
This pullrequest adds a rule to detect `also` blocks, which could be simplified by using an `apply` block, as discussed in #518.

Closes #518

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -608,6 +608,8 @@ style:
     excludeRawStrings: true
   MayBeConst:
     active: true
+  MisusedAlso:
+    active: false
   ModifierOrder:
     active: true
   MultilineLambdaItParameter:
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlso.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlso.kt
@@ -1,0 +1,62 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.rules.IT_LITERAL
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
+import org.jetbrains.kotlin.psi.KtLambdaExpression
+import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
+
+/**
+ * Detects when an `also` block contains only `it`-started expressions.
+ *
+ * By refactoring the `also` block to an `apply` block makes it so that all `it`s can be removed
+ * thus making the block more concise and easier to read.
+ *
+ * <noncompliant>
+ * Buzz().also {
+ *   it.init()
+ *   it.block()
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * Buzz().apply {
+ *   init()
+ *   block()
+ * }
+ * </compliant>
+ */
+class MisusedAlso(config: Config = Config.empty) : Rule(config) {
+
+    override val issue = Issue(
+        "MisusedAlso",
+        Severity.Style,
+        "When an `also` block contains only `it`-started expressions, simplify it to the `apply` block.",
+        Debt.FIVE_MINS
+    )
+
+    override fun visitCallExpression(expression: KtCallExpression) {
+        if (expression.calleeExpression?.text == "also") {
+            val lambda = expression.lambdaArguments.singleOrNull() ?: expression.valueArguments.single()
+                .collectDescendantsOfType<KtLambdaExpression>()
+                .single()
+            val dotQualifiedsInLambda = lambda.collectDescendantsOfType<KtDotQualifiedExpression>()
+
+            if (
+                dotQualifiedsInLambda.isNotEmpty() &&
+                dotQualifiedsInLambda.all { it.receiverExpression.textMatches(IT_LITERAL) }
+            ) {
+                report(CodeSmell(issue, Entity.from(expression), issue.description))
+            }
+        } else {
+            super.visitCallExpression(expression)
+        }
+    }
+}
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StyleGuideProvider.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StyleGuideProvider.kt
@@ -104,6 +104,7 @@ class StyleGuideProvider : DefaultRuleSetProvider {
             UseAnyOrNoneInsteadOfFind(config),
             UnnecessaryBackticks(config),
             MaxChainedCallsOnSameLine(config),
+            MisusedAlso(config)
         )
     )
 }
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlsoSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlsoSpec.kt
@@ -1,0 +1,87 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.test.assertThat
+import io.gitlab.arturbosch.detekt.test.compileAndLint
+import org.junit.jupiter.api.Test
+
+class MisusedAlsoSpec {
+    val subject = MisusedAlso(Config.empty)
+
+    @Test
+    fun `does not report when no also is used`() {
+        val code = """
+            inline fun foo(block: Buzz.() -> Unit): Buzz =
+                Buzz().let {
+                  init()
+                  block()
+                }
+        """.trimIndent()
+        assertThat(subject.compileAndLint(code)).isEmpty()
+    }
+
+    @Test
+    fun `reports an also in init of class`() {
+        val code = """
+            class Test {
+                private var a = 5
+                
+                init {
+                    a.also { 
+                        it.plus(5)
+                    }
+                }
+            }
+        """.trimIndent()
+        assertThat(subject.compileAndLint(code)).hasSize(1)
+    }
+
+    @Test
+    fun `reports an also on nullable type`() {
+        val code = """
+            class Test {
+                private var a: Int? = 5
+                
+                init {
+                    a?.also { 
+                        it.plus(5)
+                    }
+                }
+            }
+        """.trimIndent()
+        assertThat(subject.compileAndLint(code)).hasSize(1)
+    }
+
+    @Test
+    fun `reports an also with lambda passed as Argument in parenthesis`() {
+        val code = """
+            class Test {
+                private var a: Int? = 5
+                
+                init {
+                    a?.also({ 
+                        it.plus(5)
+                    })
+                }
+            }
+        """.trimIndent()
+        assertThat(subject.compileAndLint(code)).hasSize(1)
+    }
+
+    @Test
+    fun `does not report if it is not used in also`() {
+        val code = """
+            class Test {
+                private var a: Int? = 5
+                private var b: Int = 0
+                
+                init {
+                    a?.also({ 
+                        b.plus(5)
+                    })
+                }
+            }
+        """.trimIndent()
+        assertThat(subject.compileAndLint(code)).isEmpty()
+    }
+}


BraisGabin (approved):
Thanks to implement this rule! 5 years since it was proposed 😄. LGTM.

3flex:
I think the rule name should be a little more specific, like `AlsoCouldBeApply`. I haven't reviewed, but thank you for the PR, this is a great contribution!

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlsoSpec.kt
```diff
@@ -0,0 +1,92 @@
...
+            class Test {
+                private var a = 5
+                
+                init {
+                    a.let { 
+                        it.plus(5)
+                        it.minus(10)
+                    }
+                }
+            }
```
TWiStErRob, line 23: you can remove most of the test code focusing on the expression only:
```suggestion
                fun f(a: Int) {
                    a.let { 
                        it.plus(5)
                        it.minus(10)
                    }
                }
```
this also makes the tests more realistic IMO.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StyleGuideProvider.kt
```diff
@@ -104,6 +104,7 @@ class StyleGuideProvider : DefaultRuleSetProvider {
             UseAnyOrNoneInsteadOfFind(config),
             UnnecessaryBackticks(config),
             MaxChainedCallsOnSameLine(config),
+            MisusedAlso(config)
```
TWiStErRob, line 107: trailing comma

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlso.kt
```diff
@@ -0,0 +1,62 @@
...
+
+            if (
+                dotQualifiedsInLambda.isNotEmpty() &&
+                dotQualifiedsInLambda.all { it.receiverExpression.textMatches(IT_LITERAL) }
+            ) {
+                report(CodeSmell(issue, Entity.from(expression), issue.description))
+            }
+        } else {
+            super.visitCallExpression(expression)
+        }
```
TWiStErRob, line 60: Do you not need to visit super in the `if`? @BraisGabin @3flex

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlso.kt
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * <noncompliant>
+ * Buzz().also {
+ *   it.init()
+ *   it.block()
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * Buzz().apply {
```
TWiStErRob, line 30: Should it be mentioned that this is also compliant? `return foo.also { println(it) }`

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -608,6 +608,8 @@ style:
     excludeRawStrings: true
   MayBeConst:
     active: true
+  MisusedAlso:
```
TWiStErRob, line 611: > I think the rule name should be a little more specific, like `AlsoCouldBeApply`. I haven't reviewed, but thank you for the PR, this is a great contribution! -- https://github.com/detekt/detekt/pull/5333#issuecomment-1256862454

I agree with @3flex on the naming, "misused" can mean many things, for example there are cases when you don't even need an also: `x.also { f(it) }` as a statement is equivalent to `f(x)`.

TWiStErRob (commented):
I agree with @3flex on the naming.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlsoSpec.kt
```diff
@@ -0,0 +1,92 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.test.assertThat
+import io.gitlab.arturbosch.detekt.test.compileAndLint
+import org.junit.jupiter.api.Test
+
+class MisusedAlsoSpec {
```
TWiStErRob, line 8: add a test case for?
```
x.let { named ->
    named.foo()
    named.bar(y)
}
```
should this be reported? (I think not because we would lose information (the variable name).)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlsoSpec.kt
```diff
@@ -0,0 +1,92 @@
...
+    fun `does not report if it is not used in also`() {
+        val code = """
+            class Test {
+                private var a: Int? = 5
+                private var b: Int = 0
+                
+                init {
+                    a?.also({ 
+                        b.plus(5)
+                    })
```
TWiStErRob, line 86: copy-paste, parentheses are not idiomatic
```suggestion
                    a?.also { 
                        b.plus(5)
                    }
```

Tommyten: I actually wanted to have at least one test case, where this unidiomatic variant is used, because in this case I cannot access the via the `lambdaArguments` but rather by using the `valueArguments` of the expression. Just in case (even though highly unlikely) there is some kind of regression or change here, it will be caught by the unit tests.

TWiStErRob: Yep, I agree, nice one. The reason this one stood out, is because the one above this one is specifically testing for it. "A test should fail for only one reason" - is the usual mantra. This test can fail because the parentheses thing is broken OR because `it` is not used.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlso.kt
Tommyten: I'd rather not have it reported, as the renaming of `it` could have a benefit for understanding the code, especially when manipulating collections multiple times using functions like `map` and the likes.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlsoSpec.kt
Tommyten: I'd rather not have it reported, as the renaming of it could have a benefit for understanding the code, especially when manipulating collections multiple times using functions like map and the likes.

I've added a test case for this scenario :)

detekt-core/src/main/resources/default-detekt-config.yml
Tommyten: The rule has been renamed as per the suggestion of @3flex

TWiStErRob: Fix default configuration and PR title too.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlso.kt
TWiStErRob: Resolved

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlsoSpec.kt
TWiStErRob: Resolved

TWiStErRob (approved):
I'm good, except pending conversations, @BraisGabin can you please look again?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/AlsoCouldBeApplySpec.kt
```diff
@@ -0,0 +1,84 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.test.assertThat
+import io.gitlab.arturbosch.detekt.test.compileAndLint
+import org.junit.jupiter.api.Test
+
+class AlsoCouldBeApplySpec {
```
TWiStErRob, line 8: Does semicolon change the PSI?
`a.let { it.plus(5); it.minus(10) }`

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlso.kt
TWiStErRob: I had a dream about this, I think `it.also { x -> x.also { it.foo() } }` will not be found because of the missing super call in the other branch.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/AlsoCouldBeApply.kt
```diff
@@ -0,0 +1,67 @@
...
+            val lambda = expression.lambdaArguments.singleOrNull() ?: expression.valueArguments.single()
+                .collectDescendantsOfType<KtLambdaExpression>()
+                .single()
+            val dotQualifiedsInLambda = lambda.collectDescendantsOfType<KtDotQualifiedExpression>()
+
+            if (
+                dotQualifiedsInLambda.isNotEmpty() &&
+                dotQualifiedsInLambda.all { it.receiverExpression.textMatches(IT_LITERAL) }
+            ) {
+                report(CodeSmell(issue, Entity.from(expression), issue.description))
```
TWiStErRob, line 61: please focus the report to the `also` function name, otherwise you end up with a problem similar to these: https://github.com/detekt/detekt/issues/5316#event-7412635913 (there are PRs for each)

TWiStErRob (commented):
I had a dream about this, left 3 more comments :)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/AlsoCouldBeApplySpec.kt
Tommyten: It does change the PSI a bit, but not in a way that breaks the rule, added a Testcase for this case.

Concerning `it.also { x -> x.also { it.foo() } }`
I've played with it a bit and it isn't found either way... :/ Not sure, whether I'm doing something wrong here. Help is appreciated :)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlso.kt
Tommyten: Yes, adding a super call in the other branch helped with detecting this violation, thank you :)

Tweaked the code and added a corresponding test case.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/AlsoCouldBeApply.kt
Tommyten: I think it should be focused on the `also` now. I've changed the code to be the following:
`report(CodeSmell(issue, Entity.from(expression.calleeExpression!!), issue.description))`

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/AlsoCouldBeApplySpec.kt
TWiStErRob: Looks good, you can resolve this.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MisusedAlso.kt
TWiStErRob: Nice one!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/AlsoCouldBeApply.kt
TWiStErRob: If you want, you can add assertions for this by checking for the start/end location. (See related issues' PRs)

BraisGabin (approved):
I re-reviewed the PR and it looks great. Thanks for the reviews @TWiStErRob and the changes @Tommyten.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/AlsoCouldBeApply.kt
```diff
@@ -0,0 +1,71 @@
...
+            if (
+                dotQualifiedsInLambda.isNotEmpty() &&
+                dotQualifiedsInLambda.all { it.receiverExpression.textMatches(IT_LITERAL) }
+            ) {
+                report(CodeSmell(issue, Entity.from(alsoExpression), issue.description))
+            }
+
+            super.visitCallExpression(expression)
+        } else {
+            super.visitCallExpression(expression)
```
cortinico, line 68: I would move the `super.` call to be the first statement.

cortinico (approved):
Great stuff! Just a minor nit

cortinico (approved):

PR #5374, opened 2022-10-05
UnusedPrivateMember: fix false negative with named arguments

### LINKED ISSUE #5373: `UnusedPrivateMember` doesn't report an unused parameter when function body has a call with same named parameter
Detekt when run on this code
```kotlin
@Composable
fun Compass(
  modifier: Modifier = Modifier,
) {
  Box(modifier = Modifier) {

  }
}
```
Should report `UnusedPrivateMember` for `modifier` parameter, but it doesn't because it confuses a named argument in the `Box` call with parameter usage. Notice that the value of the Box argument is `Modifier`, not `modifier`.

If above snippet is changed to this:
```kotlin
@Composable
fun Compass(
  modifier: Modifier = Modifier,
) {
  Box {

  }
}
```
then detekt correctly reports `modifier` as unused

---

* Type resolution: enabled

t-kameyama:
Fixes #5373


BraisGabin (approved):
Thanks!

PR #5381, opened 2022-10-06
Obtain KotlinJvmProjectExtension's target on Kotlin 1.7.20

Goooler:
See https://github.com/detekt/detekt/pull/5163#issuecomment-1269738215


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektJvm.kt
```diff
@@ -9,18 +9,15 @@ import org.gradle.api.file.FileCollection
 import org.jetbrains.kotlin.gradle.dsl.KotlinCommonOptions
 import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
 import org.jetbrains.kotlin.gradle.plugin.KotlinCompilation
-import org.jetbrains.kotlin.gradle.plugin.mpp.pm20.targets
 
 internal class DetektJvm(private val project: Project) {
     fun registerTasks(extension: DetektExtension) {
-        project.extensions.getByType(KotlinJvmProjectExtension::class.java).targets.forEach { target ->
```
Goooler, line 16: `KotlinProjectExtension.targets` in Kotlin 1.7.10:

```kotlin
val KotlinProjectExtension.targets: Iterable<KotlinTarget>
    get() = when (this) {
        is KotlinSingleTargetExtension -> listOf(this.target)
        is KotlinMultiplatformExtension -> targets
        else -> error("Unexpected 'kotlin' extension $this")
    }
```

BraisGabin (approved):

Goooler:
If not, I'll address a compatible commit with https://github.com/Goooler/detekt/pull/3.

Goooler:
We can fix this broken change like https://github.com/Goooler/detekt/commit/bb81a270c5aa98218c07ce021cd4520e483d8a42?

Goooler:
And the other solution is using `KotlinJvmProjectExtension`'s target directly like https://github.com/Goooler/detekt/commit/c6b95503f32dd18e3819ef23a55a68c0c4fe654d.

3flex:
I had a look at this and there's no need to use that function there. So I've opened https://github.com/detekt/detekt/pull/5383 to remove it which should avoid any compatibility issues.

PR #5383, opened 2022-10-06
Remove usage of MPP targets function for JVM-only projects

3flex:
`KotlinJvmProjectExtension` is a `KotlinSingleJavaTargetExtension` and only has a single target, so using the MPP `targets` function is unnecessary.

See https://github.com/detekt/detekt/pull/5381#issuecomment-1269979238 for background.


cortinico (approved):

Goooler (commented):
LGTM, safe to land, same fix https://github.com/Goooler/detekt/commit/c6b95503f32dd18e3819ef23a55a68c0c4fe654d.

PR #5386, opened 2022-10-07
Enable Ktlint two TrailingComma rules to default setting in Ktlint

chao2zhang:
From https://github.com/detekt/detekt/pull/5312#discussion_r989658904, this PR demonstrates what it looks like when enabling the two TrailingComma rules with ktlint's default setting.

```
  TrailingCommaOnCallSite:
    active: false
    autoCorrect: true
    useTrailingCommaOnCallSite: false
  TrailingCommaOnDeclarationSite:
    active: false
    autoCorrect: true
    useTrailingCommaOnDeclarationSite: false
```

My vote would be to configure these rules per [Kotlin coding convention](https://kotlinlang.org/docs/coding-conventions.html#trailing-commas)

> Trailing commas are entirely optional – your code will still work without them. The Kotlin style guide encourages the use of trailing commas at the declaration site and leaves it at your discretion for the call site.

```
  TrailingCommaOnCallSite:
    active: false
    autoCorrect: true
    useTrailingCommaOnCallSite: false
  TrailingCommaOnDeclarationSite:
    active: true
    autoCorrect: true
    useTrailingCommaOnDeclarationSite: true
```


3flex:
I've opened https://github.com/pinterest/ktlint/pull/1669 & https://github.com/pinterest/ktlint/pull/1670 to get the defaults changed for ktlint, hopefully maintainers agree. I don't think the current defaults make sense either, but also think that if we apply the ktlint rules, then we should apply the same config by default.

3flex:
Changes have just been made upstream, I think it makes sense to change default config in detekt to align before 1.22.0 is released - otherwise users have to deal with the trailing commas rule split in 1.22 then a change to default config in 1.23. It's better if they can deal with both at once.

Upstream both rules were changed so commas are required for standard Kotlin and forbidden for Android. That wasn't my suggestion (I strongly believe they should be forced on for both by default) so we either override ktlint defaults here, or try and convince ktlint maintainer to reconsider.

chao2zhang:
- I will follow the upstream for the default configurations for ktlint
- I will also add references in the rule description as well for people to refer to the official coding convention for Kotlin and Android, before opening new issues in our repository.

chao2zhang:
By the way I would still like to close this PR so that the history of PR is preserved without force push.

3flex:
I think force push is fine but up to you, no problem if you want to close.

PR #5390, opened 2022-10-08
ObjectPropertyNaming: fix false positive with top level properties

### LINKED ISSUE #5340: ObjectPropertyNaming rule false positive in 1.22.0-RC1
## Expected Behavior

For the Kotlin file
```kotlin
val _invalidNaming = 1
```

I expect only `TopLevelPropertyNaming` to trigger.

## Observed Behavior

Since 1.22.0-RC1, it also triggers `ObjectPropertyNaming`.

## Steps to Reproduce

Add

```kotlin
    @Nested
    inner class `top level properties` {
        @Test
        fun `should not detect top level properties`() {
            val subject = ObjectPropertyNaming()

            val code = """
                val _invalidNaming = 1
            """.trimIndent()

        }
    }
```

to `ObjectPropertyNamingSpec` and you'll see that it triggers. This happens also for Detekt 1.21.0, but I think that the `NamingRules` class used to avoid this issue in production:
https://github.com/detekt/detekt/commit/c435739edfb8d51d98f99d518610e6b73d903fe3#diff-45d7198fea5671bd6903db1a08ff11a4e063b8ca9dadc1be56372449d7b3a3ab 

## Context

Trying to bump Detekt without running into false positives ;)

## Your Environment

* Version of detekt used: 1.22.0-RC1

DJtheRedstoner:
Closes #5340


BraisGabin (approved):
Thanks!

PR #5397, opened 2022-10-09
Fix false negative MultilineLambdaItParameter on complex multiline single statement

### LINKED ISSUE #5393: False negative MultilineLambdaItParameter on complex multiline single statement
## Expected Behavior
Expectation to pass this test:
```kotlin
    @Nested
    inner class `single parameter, multiline lambda with a single statement` {

        @Test
        fun `reports when statement is spanning multiple lines`() {
            val code = """
            fun f() {
                val digits = 1234.let {
                    check(it > 0) {
                        // This can be arbitrarily long or complex.
                        "${'$'}it must be positive."
                    }
                }
            }
            """.trimIndent()
        }
```

## Observed Behavior
```kotlin
fun Feed.verifyHasAllAttributes(attributes: Set<Feed.Attribute>) {
	attributes.forEach { // implicit `it ->` here wasn't flagged.
		assertTrue(this.attributes.contains(it)) {
			"$it not found in ${this.attributes}"
		}
	}
}

fun Feed.verifyAllAttributesAreValid() {
	this.attributes.forEach { // implicit `it ->` here was flagged.
		assertNotNull(it.code)
		assertNotNull(it.title)
	}
}
```

```
> Task :backend:feed:detektTest FAILED
style - 20min debt
        MultilineLambdaItParameter - [The implicit `it` should not be used in a multiline lambda. Consider giving your(...)] at backend\feed\src\test\kotlin\net\twisterrob\cinema\cineworld\sync\syndication\FeedData.kt:80:26
```

## Steps to Reproduce
 * Clone https://github.com/TWiStErRob/net.twisterrob.cinema
 * Check out 92bca07146165cc17693b0b6fa95be7d47baf550
 * `gradlew :backend:feed:detektTest`
 * Look for FeedData.kt

## Context
I think this should be flagged if the single statement spans more than 1 line. At the point `it` is captured in any lambda, the complexity goes way up, and an explicit name helps readability greatly same as any other flagging of `MultilineLambdaItParameter`. I think the example clearly shows that a mutli-line single-statement lambda can be very easily more complex than a multi-statement lambda.

## Your Environment

* Link to your project (if it's a public repository): https://github.com/TWiStErRob/net.twisterrob.cinema/blob/490998a42ecda26515ff761db6f51a224130a993/Heroku/backend/feed/src/test/kotlin/net/twisterrob/cinema/cineworld/sync/syndication/FeedData.kt#L67-L84

t-kameyama:
Fixes #5393

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
@@ -11,7 +11,9 @@ import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
 import io.gitlab.arturbosch.detekt.rules.IT_LITERAL
 import io.gitlab.arturbosch.detekt.rules.hasImplicitParameterReference
 import io.gitlab.arturbosch.detekt.rules.implicitParameter
+import org.jetbrains.kotlin.psi.KtBlockExpression
 import org.jetbrains.kotlin.psi.KtLambdaExpression
+import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
 
 /**
  * Lambda expressions are very useful in a lot of cases, and they often include very small chunks of
@@ -74,8 +76,8 @@ class MultilineLambdaItParameter(val config: Config) : Rule(config) {
 
     override fun visitLambdaExpression(lambdaExpression: KtLambdaExpression) {
         super.visitLambdaExpression(lambdaExpression)
-        val size = lambdaExpression.bodyExpression?.statements?.size
-        if (size == null || size <= 1) return
+        val size = lambdaExpression.collectDescendantsOfType<KtBlockExpression>().flatMap { it.statements }.size
+        if (size <= 1) return
 
         val parameterNames = lambdaExpression.valueParameters.map { it.name }
         // Explicit `it`
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameterSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameterSpec.kt
@@ -114,6 +114,21 @@ class MultilineLambdaItParameterSpec(val env: KotlinCoreEnvironment) {
             val findings = subject.compileAndLintWithContext(env, code)
             assertThat(findings).isEmpty()
         }
+
+        @Test
+        fun `reports when statement is spanning multiple lines`() {
+            val code = """
+            fun f() {
+                val digits = 1234.let {
+                    check(it > 0) {
+                        println(it)
+                    }
+                }
+            }
+            """.trimIndent()
+            val findings = subject.compileAndLintWithContext(env, code)
+            assertThat(findings).hasSize(1)
+        }
     }
 
     @Nested


schalkms (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
```diff
@@ -74,8 +76,8 @@ class MultilineLambdaItParameter(val config: Config) : Rule(config) {
 
     override fun visitLambdaExpression(lambdaExpression: KtLambdaExpression) {
         super.visitLambdaExpression(lambdaExpression)
-        val size = lambdaExpression.bodyExpression?.statements?.size
-        if (size == null || size <= 1) return
+        val size = lambdaExpression.collectDescendantsOfType<KtBlockExpression>().flatMap { it.statements }.size
```
TWiStErRob, line 79: @t-kameyama thanks for the quick fix!

Small note: unnecessary creation of intermediate list...
```
val size = lambdaExpression.collectDescendantsOfType<KtBlockExpression>().sumOf { it.statements.size }
```
looks like we could have a `UseSumInsteadOfSize` similar to `UseAnyOrNoneInsteadOfFind`, I'll open an issue.

TWiStErRob: https://github.com/detekt/detekt/issues/5399

eygraber:
Would this change have caused the following to trigger MultilineLambdaItParameter:

```
?.map { it.seekTieFighters { ForegroundColorSpan(colorLink) } }
```

If yes, I'd argue that I want to keep that as `it`

TWiStErRob:
Example 1 is covered in #5504

---

Example 2: I agree that this is simple enough (more on this later), but I guess maybe the rule needs to be smarter, and either consider this style as a "single line", or detect that the receiver is a simple name. Because it might as well be `foo().bar.baz().z { fux(it) }?.lam?.da?.also { ...`, in which case naming the `it` is paramount. Again, a separate issue will help to discuss this. Here's a repro:

```kotlin
    @Test
    fun `example2`() {
        val code = """
            interface TextView {
                var isVisible: Boolean
                var text: String
            }
            fun phrase(text: String, b1: Boolean, b2: Boolean, block: () -> String?): String = TODO()
            fun f(textView: TextView, text: String?) {
                textView.isVisible = text?.also {
                    textView.text = phrase(it, false, false) { null }
                } != null
            }
        """.trimIndent()
    }
```

---

Now, as for "simple enough". I strongly believe that when a Detekt (or any inspection / static analysis tool) flags a line of code, it is for a reason. But we should never take the suggestions literally as the rules are rules, but the solution might be something totally different. For example in your second example, it took me about a 30 seconds to decypher the data flow and nullabilities, which is unnecessarily long. The code could be way simpler:
```kotlin
val somekindOfPhrase = phrase(text, false, false) { null } // name based on those (false, false, null) behavior flags
textView.isVisible = somekindOfPhrase != null
textView.text = somekindOfPhrase
```
ok, this is not a single statement, but we don't need to take _"in Kotlin, everything is an expression"_ literally, right? 🤓

Note that the two `TextView` interactions lend themselves really nicely for an extension `var TextView.textOrGone: CharSequence?`, which is common enough in my experience to warrant a util. At which point we're down to:
```kotlin
textView.textOrGone = phrase(text, false, false) { null } 
```

TWiStErRob:
Example 1 is covered in #5504

---

Example 2: I agree that this is simple enough (more on this later), but I guess maybe the rule needs to be smarter, and either consider this style as a "single line", or detect that the receiver is a simple name. Because it might as well be `foo().bar.baz().z { fux(it) }?.lam?.da?.also { ...`, in which case naming the `it` is paramount. Again, a separate issue will help to discuss this. Did you open https://github.com/detekt/detekt/issues/5503 as a generalization of this example? Here's a repro:

```kotlin
    @Test
    fun `example2`() {
        val code = """
            interface TextView {
                var isVisible: Boolean
                var text: String
            }
            fun phrase(text: String, b1: Boolean, b2: Boolean, block: () -> String?): String = TODO()
            fun f(textView: TextView, text: String?) {
                textView.isVisible = text?.also {
                    textView.text = phrase(it, false, false) { null }
                } != null
            }
        """.trimIndent()
    }
```

---

Now, as for "simple enough". I strongly believe that when a Detekt (or any inspection / static analysis tool) flags a line of code, it is for a reason. But we should never take the suggestions literally as the rules are rules, but the solution might be something totally different. For example in your second example, it took me about a 30 seconds to decypher the data flow and nullabilities, which is unnecessarily long. The code could be way simpler:
```kotlin
val somekindOfPhrase = phrase(text, false, false) { null } // name based on those (false, false, null) behavior flags
textView.isVisible = somekindOfPhrase != null
textView.text = somekindOfPhrase
```
ok, this is not a single statement, but we don't need to take _"in Kotlin, everything is an expression"_ literally, right? 🤓

Note that the two `TextView` interactions lend themselves really nicely for an extension `var TextView.textOrGone: CharSequence?`, which is common enough in my experience to warrant a util. At which point we're down to:
```kotlin
textView.textOrGone = phrase(text, false, false) { null } 
```

eygraber:
> I  strongly believe that when a Detekt (or any inspection / static analysis tool) flags a line of code, it is for a reason

I strongly agree! In fact, I hate this snippet of code, and we did eventually make something like `textOrGone` but only after noticing this pattern happening a lot. 

#5503 is a generalization of this (and I like my example there better), but this one was failing due to this PR so I included it here.

PR #5405, opened 2022-10-12
Add UseSumOfInsteadOfFlatMapSize rule

### LINKED ISSUE #5399: New rule: UseSumInsteadOfFlatMapSize
## Context / Expected Behavior of the rule
Similar to `UseAnyOrNoneInsteadOfFind`, it would recognize cases when we are doing `flatMap` instead of calculating the end result directly.

```kotlin
val size = lambdaExpression.collectDescendantsOfType<KtBlockExpression>().flatMap { it.statements }.size
```
[could be](https://github.com/detekt/detekt/pull/5397/files#r991027678)
```kotlin
val size = lambdaExpression.collectDescendantsOfType<KtBlockExpression>().sumOf { it.statements.size }
```
which is simpler and more performant.

I've seen this flatMap + size/isEmpty pattern used a lot previously, even by myself, but we can do better!

So far I can only think of
 * `.flatMap{}.size` -> `.sumOf`
 * `.flatMap{}.isEmpty()` -> could be `.sumOf`, but can be even shortcut with `.none{it.isNotEmpty()}`
 * `.flatMap{}.isNotEmpty()` -> could be `.sumOf`, but can be shortcut with `.any{it.isNotEmpty()}`

There might be other cases too, I think the first one would be good enough to flag, as it has a clearly better replacement. The last 2 might be too complex for the potentially micro-optimization.

t-kameyama:
Fixes #5399

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseSumOfInsteadOfFlatMapSizeSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseSumOfInsteadOfFlatMapSizeSpec.kt
@@ -106,4 +106,56 @@ class UseSumOfInsteadOfFlatMapSizeSpec(val env: KotlinCoreEnvironment) {
         val actual = subject.compileAndLintWithContext(env, code)
         assertThat(actual).hasSize(1)
     }
+
+    @Test
+    @DisplayName("Does not report flatMap")
+    fun noReportFlatMap() {
+        val code = """
+            fun test(list: List<Foo>) {
+                list.flatMap { it.foo }
+            }
+            class Foo(val foo: List<Int>)
+        """.trimIndent()
+        val actual = subject.compileAndLintWithContext(env, code)
+        assertThat(actual).isEmpty()
+    }
+
+    @Test
+    @DisplayName("Does not report flatMap and first")
+    fun noReportFlatMapAndFirst() {
+        val code = """
+            fun test(list: List<Foo>) {
+                list.flatMap { it.foo }.first()
+            }
+            class Foo(val foo: List<Int>)
+        """.trimIndent()
+        val actual = subject.compileAndLintWithContext(env, code)
+        assertThat(actual).isEmpty()
+    }
+
+    @Test
+    @DisplayName("Does not report flatten")
+    fun noReportFlatten() {
+        val code = """
+            fun test(list: List<List<Int>>) {
+                list.flatten()
+            }
+            class Foo(val foo: List<Int>)
+        """.trimIndent()
+        val actual = subject.compileAndLintWithContext(env, code)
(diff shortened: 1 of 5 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseSumOfInsteadOfFlatMapSizeSpec.kt
```diff
@@ -0,0 +1,109 @@
...
+    @DisplayName("Reports flatMap and count on Set")
+    fun reportFlatMapAndCountOnSet() {
+        val code = """
+            fun test(set: Set<Bar>) {
+                set.flatMap { it.bar }.count()
+            }
+            class Bar(val bar: Set<Int>)
+        """.trimIndent()
+        val actual = subject.compileAndLintWithContext(env, code)
+        assertThat(actual).hasSize(1)
```
BraisGabin, line 107: Could you add some tests to ensure that this rule doesn't flag all the `flatMap` calls?

BraisGabin (commented):
Thank you, as always, for your valuable contributions :)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseSumOfInsteadOfFlatMapSizeSpec.kt
t-kameyama: dcb4891

BraisGabin (approved):

schalkms (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseSumOfInsteadOfFlatMapSizeSpec.kt
```diff
@@ -0,0 +1,161 @@
...
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Test
+
+@KotlinCoreEnvironmentTest
+class UseSumOfInsteadOfFlatMapSizeSpec(val env: KotlinCoreEnvironment) {
+    val subject = UseSumOfInsteadOfFlatMapSize()
+
+    @Test
+    @DisplayName("Reports flatMap and size")
+    fun reportFlatMapAndSize() {
```
cortinico, line 16: What's the reason why this is not:
```kotlin
fun `reports flatMap and size`() {
```

It seems like this goes a bit against the rest of the codebase convention.

t-kameyama: https://github.com/detekt/detekt/pull/5416

PR #5407, opened 2022-10-12
Report if/else as issue location instead of block

### LINKED ISSUE #5317: MandatoryBracesIfStatement to highlight only if/else keywords
Pretty much the same as #5316, but a different rule.

## Context
image

Warnings exists in some codebases, and that's ok, but then they make the code harder to read, then it becomes a problem. A warnings should highlight where the problem is, specifically, but some rules report on the blocks which could be quite large, therefor hiding potential crashes, which could be quite serious. In the example above the `MandatoryBracesIfStatement` hides an `UnsafeCallOnNullableType` (potential NPE).

Move the highlight of the reported violation to the `if` keyword and the `else` keyword. Note to implementer: mind `else if`s.

dizney:
Report mandatory braces issues on the actual if and/or else keywords instead of the then or else code blocks to prevent hiding potential other issues in those code blocks.

Closes #5317


BraisGabin (approved):
Thanks!!

PR #5423, opened 2022-10-16
Deprecate `TrailingComma` as it's now split in two rules

cortinico:
When we bumped KtLint we effectively deprecated a rule `TrailingComma`, but we never updated the `deprecation.properties` file for it. Here I'm doing it.

### DIFF
--- a/detekt-core/src/main/resources/deprecation.properties
+++ b/detekt-core/src/main/resources/deprecation.properties
@@ -13,6 +13,7 @@ style>UnnecessaryAbstractClass>excludeAnnotatedClasses=Use `ignoreAnnotated` ins
 style>UseDataClass>excludeAnnotatedClasses=Use `ignoreAnnotated` instead
 formatting>Indentation>continuationIndentSize=`continuationIndentSize` is ignored by KtLint and will have no effect
 formatting>ParameterListWrapping>indentSize=`indentSize` is ignored by KtLint and will have no effect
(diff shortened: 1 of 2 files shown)


chao2zhang:
I think I run into [the same problem](https://github.com/detekt/detekt/pull/5312#discussion_r989672213) when adding the rule to `deprecation.properties`. 
```

* What went wrong:
> The default-detekt-config.yml is not up-to-date. You can execute the generateDocumentation Gradle task to update it and commit the changed files.
```

It fails the check because these are not deprecated, but rather removed.
I believe we should be fine without update `deprecation.properties` because `TrailingComma` is highly visible when our users upgrade to newer versions of detekt

cortinico:
> It fails the check because these are not deprecated, but rather removed.
> I believe we should be fine without update `deprecation.properties` because `TrailingComma` is highly visible when our users upgrade to newer versions of detekt

Yup you're right. Still it should be possible to give the user an informative message, no?

BraisGabin:
> Yup you're right. Still it should be possible to give the user an informative message, no?

I would hardcode this inside the generator code. This way it will be added automatically.

@VitalyVPinchuk did this already here: 
https://github.com/detekt/detekt/blob/550035411a90b0e133f08291bd10e1a871e43129/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/DeprecatedPrinter.kt#L38-L43

So it should be as easy as add a new line there.

schalkms (approved):

BraisGabin (approved):

PR #5442, opened 2022-10-19
Separating ComplexMethod rule into CyclomaticComplexMethod and CognitiveComplexMethod

### LINKED ISSUE #5349: Allow the user to choose between cyclomatic and cognitive complexity in ComplexMethod rule
## Expected Behavior

The cyclomatic complexity calculates how many paths there are in a function.
It is a measurement for testability as the number of paths tells us how many test cases we need to write.

For example a class with 10x mcabe=1 functions would accumulate to complexity 10 and a class with just one function which is nested a lot could still accumulate to complexity 10.
Most users will say that both classes are not equal complex but the deeper nested one is. 

Here comes the cognitive complexity into play. It takes nesting into count and would calculate a much higher value for the one function with deep nesting.

https://github.com/detekt/detekt/blob/31520dadf2b7eec04139dee9103645d1f208ccbd/detekt-metrics/src/main/kotlin/io/github/detekt/metrics/CognitiveComplexity.kt#L28-L36

To not break users baselines and CI's, we could let the user choose the complexity metric to calculate:

```yaml
  ComplexMethod:
    active: true
    threshold: 15
    metric: "cognitive-complexity" # cyclomatic-complexity
```

## Current Behavior
Cyclomatic complexity is used to calculate function complexity.

## Context
Cyclomatic complexity does not really measure the complexity but the testability of a function.
Cognitive complexity was described in https://www.sonarsource.com/docs/CognitiveComplexity.pdf to take function nesting into count.

sanggggg:
Fixes #5349

### DIFF
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CognitiveComplexMethod.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CognitiveComplexMethod.kt
@@ -14,7 +14,6 @@ import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
 import io.gitlab.arturbosch.detekt.api.internal.Configuration
 import org.jetbrains.kotlin.psi.KtNamedFunction
 
-
 /**
  * Complex methods are hard to understand and read. It might not be obvious what side-effects a complex method has.
  * Prefer splitting up complex methods into smaller methods that are in turn easier to understand.
@@ -53,7 +52,8 @@ class CognitiveComplexMethod(config: Config = Config.empty) : Rule(config) {
                     issue,
                     Entity.atName(function),
                     Metric("CC", complexity, threshold),
-                    "The function ${function.nameAsSafeName} appears to be too complex based on Cognitive Complexity (complexity: $complexity). " +
+                    "The function ${function.nameAsSafeName} appears to be too complex " +
+                        "based on Cognitive Complexity (complexity: $complexity). " +
                         "Defined complexity threshold for methods is set to '$threshold'"
                 )
             )
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CognitiveComplexMethod.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CognitiveComplexMethod.kt
@@ -21,7 +21,7 @@ import org.jetbrains.kotlin.psi.KtNamedFunction
  * Smaller methods can also be named much clearer which leads to improved readability of the code.
  *
  * This rule measures and restricts the complexity of the method through the [Cognitive Complexity metric of Sonasource](https://www.sonarsource.com/docs/CognitiveComplexity.pdf).
- * Which improves McCabe's Cyclomatic Complexity ({@link ComplexMethod}) considering the programmer's mental model.
+ * Which improves McCabe's Cyclomatic Complexity ({@link CyclomaticComplexMethod}) considering the programmer's mental model.
  *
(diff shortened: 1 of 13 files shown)


BraisGabin:
Did you create the PR to get early reviews or to get early CI checks? Or both? In other words, even the PR is a Draft, do you want me to review it or do you want to work a bir more and make CI happy first?

sanggggg:
@BraisGabin Thank you for checking, I just opened it to get a review including your feedback.
(I opened it to receive an early review of the changes!)

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CognitiveComplexMethod.kt
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Similar to cyclomatic complexity, it is a mathematical model that increases +1 complexity for flow control statements,
+ * but increases additional complexity when the statements are deeply nested.
+ *
+ * The statements that increase the complexity or the nesting level are as follows.
+ * - __Complexity Increments__ - `if`, `when`, `for`, `while`, `do while`, `catch`, `labeled break`, `labeled continue`, `labeled return`, `recursion call`, `&&`, `||`
+ * - __Nesting Level Increments__ - `if`, `when`, `for`, `while`, `do while`, `catch`, `nested function`
+ * - __Additional Complexity Increments by Nesting Level__ - `if`, `when`, `for`, `while`, `do while`, `catch`
+ */
+@ActiveByDefault(since = "1.22.0")
```
BraisGabin, line 33: You should remove this line. And probably run `./gradlew generateDocumentation` because CI is complaining that the documentation is not on sync (probably `detekt-core/src/main/resources/default-detekt-config.yml` or `detekt-core/src/main/resources/deprecation.properties`)

That couls also fix the othere CI issues... I don't understand why this rule is executed...

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CyclomaticComplexMethod.kt
```diff
@@ -38,15 +38,17 @@ import org.jetbrains.kotlin.psi.KtWhenExpression
...
  */
 @ActiveByDefault(since = "1.0.0")
-class ComplexMethod(config: Config = Config.empty) : Rule(config) {
+class CyclomaticComplexMethod(config: Config = Config.empty) : Rule(config) {
 
     override val issue = Issue(
-        "ComplexMethod",
+        "CyclomaticComplexMethod",
         Severity.Maintainability,
         "Prefer splitting up complex methods into smaller, easier to understand methods.",
```
BraisGabin, line 46:
```suggestion
        "Prefer splitting up complex methods into smaller, easier to test methods.",
```

Cyclomatic focus more on testability. This also make it a bit more different from cognitive

BraisGabin (commented):
Good job! Just two minor things and you need to make CI happy. I think that fixint the issues that I comment it will be happy but I'm not 100% sure because I don't fully understand why it is complaining.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CognitiveComplexMethod.kt
sanggggg: removing ActiveByDefault & sync config works fine 😄 
2a0ae60

BraisGabin (approved):
LGTM

cortinico:
> I just added the 1.22 milestone to this one. But I would know what others think about that. We are at RC2. Should we wait until the stable version to merge it?

I think we might have a RC3 as we received sevaral regressions in RC3 so it's probabyl also fine to merge this.

BraisGabin:
I fully agree with a `RC3`. The number of contributions this month is great :) I imagine that hacktoberfest is working as a charm :)

PR #5431, opened 2022-10-17
BooleanPropertyNaming highlight only the name of the variable

TWiStErRob:
... the usual, the warning hides too much, but also deeplinking is broken, e.g. clicking on the this code location:
```
BooleanPropertyNaming - [Boolean property name should match a ^(is|has|are) pattern. Actual name is email(...)] at P:\projects\workspace\net.twisterrob.cinema\Heroku\backend\endpoint\src\main\kotlin\net\twisterrob\cinema\cineworld\backend\endpoint\auth\data\UserInfoOpenID.kt:31:2
```
brings me to (before)
image

(expected)
image

PR #5444, opened 2022-10-19
New ruleauthors rule for Entity.from(x.nameIdentifier ?: x) -> Entity.atName(x)

TWiStErRob:
Motivation: automated "code review" for this:
https://github.com/detekt/detekt/pull/5431#discussion_r999050177


detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/UseNamedLocation.kt
```diff
@@ -0,0 +1,85 @@
...
+import org.jetbrains.kotlin.psi.psiUtil.getCallNameExpression
+import org.jetbrains.kotlin.psi.psiUtil.getReceiverExpression
+import org.jetbrains.kotlin.psi.psiUtil.referenceExpression
+
+/**
+ * If a rule reports issues using [Entity.from] with [KtNamedDeclaration.getNameIdentifier],
+ * then it can be replaced with [Entity.atName] for more semantic code and better baseline support.
+ */
+@ActiveByDefault("1.22.0")
+class UseNamedLocation(config: Config = Config.empty) : Rule(config) {
```
TWiStErRob, line 27: Haven't put too much thought into naming yet. Ideas welcome.

BraisGabin: random ideas:
- UseAtNameInstead
- UseEntityAtName
- UnnecesaryEntityFrom

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/UseNamedLocation.kt
```diff
@@ -0,0 +1,85 @@
...
+                    expression.referenceExpression()
+                } else {
+                    null
+                }
+
+            expression is KtPostfixExpression && expression.operationToken == KtTokens.EXCLEXCL ->
+                findNameIdentifierReceiver(expression.baseExpression!!)
+
+            expression is KtBinaryExpression && expression.operationToken == KtTokens.ELVIS ->
+                findNameIdentifierReceiver(expression.left!!)
```
TWiStErRob, line 74: What should I do with these 4 `!!` bangs? (4x UnsafeCallOnNullableType)

Making things nullable feels weird, because a PostFix and a Binary expression without their left sides doesn't make sense. Similarly an argument without a value, and a recognized `Entity.from` call without a name is also impossible.

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/UseNamedLocation.kt
TWiStErRob: I like UseEntityAtName, renamed.

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/UseNamedLocation.kt
BraisGabin: About this. The API is a bit odd. In general we try to avoid the `!!`. It's odd that they could be `null` but we try to decide what to do if they are null. We have a lot of `?: return` and we just ignore those cases. That's not ideal but this is the API that we need to work with.

TWiStErRob: btw, why are there no Detekt annotations on the PR, Danger/Sarif integration is not there?

TWiStErRob: Figured out why: https://github.com/detekt/detekt/pull/5452

BraisGabin (approved):

PR #5446, opened 2022-10-19
Fix invalid syntaxes in test code

sanggggg:
Revert unintended (probably due to automation...?) test code changes in #4511.

Screen Shot 2022-10-19 at 11 27 35 PM


BraisGabin (approved):

BraisGabin:
Thanks! These seems like something that we should be able to detect automatically...

Issue #5436, opened 2022-10-17
Detekt 1.22.0 is getting noisier and noisier


TWiStErRob:
## Expected Behavior
No output in Gradle console log unless there's an actual problem.
1.21.0:
```
> Task :plugins:detektMain
The BindingContext was created with 36 issues. Run detekt with --debug to see the error messages.

> Task :backend:quickbook:detektTest
The BindingContext was created with 6 issues. Run detekt with --debug to see the error messages.

> Task :backend:database:detektTest
The BindingContext was created with 2 issues. Run detekt with --debug to see the error messages.

> Task :backend:endpoint:detektTest
The BindingContext was created with 40 issues. Run detekt with --debug to see the error messages.
```

## Current Behavior
1.22.0-RC1:
```
> Task :plugins:detekt
The rule 'AvoidReferentialEquality' requires type resolution but it was run without it.
The rule 'Deprecation' requires type resolution but it was run without it.
The rule 'DontDowncastCollectionTypes' requires type resolution but it was run without it.
The rule 'DoubleMutabilityForCollection' requires type resolution but it was run without it.
The rule 'ElseCaseInsteadOfExhaustiveWhen' requires type resolution but it was run without it.
The rule 'ExitOutsideMain' requires type resolution but it was run without it.
The rule 'HasPlatformType' requires type resolution but it was run without it.
The rule 'MissingWhenCase' requires type resolution but it was run without it.
The rule 'NullCheckOnMutableProperty' requires type resolution but it was run without it.
The rule 'RedundantElseInWhen' requires type resolution but it was run without it.
The rule 'UnnecessaryNotNullOperator' requires type resolution but it was run without it.
The rule 'UnnecessarySafeCall' requires type resolution but it was run without it.
The rule 'UnreachableCode' requires type resolution but it was run without it.
The rule 'UnsafeCallOnNullableType' requires type resolution but it was run without it.
The rule 'UnsafeCast' requires type resolution but it was run without it.
The rule 'IgnoredReturnValue' requires type resolution but it was run without it.
The rule 'ImplicitUnitReturnType' requires type resolution but it was run without it.
The rule 'NullableToStringCall' requires type resolution but it was run without it.
The rule 'UnreachableCatchBlock' requires type resolution but it was run without it.
The rule 'UnusedUnaryOperator' requires type resolution but it was run without it.
The rule 'NestedScopeFunctions' requires type resolution but it was run without it.
[...]

PR #5449, opened 2022-10-19
Don't show the number of issues generating the BindingContext

BraisGabin:
This PR fixes the "no regression" part of #5436

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/BindingContext.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/BindingContext.kt
@@ -18,7 +18,6 @@ internal fun generateBindingContext(
     classpath: List<String>,
     files: List<KtFile>,
     debugPrinter: (() -> String) -> Unit,
-    warningPrinter: (String) -> Unit,
 ): BindingContext {
     if (classpath.isEmpty()) {
         return BindingContext.EMPTY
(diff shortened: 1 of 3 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/BindingContext.kt
```diff
@@ -64,15 +59,6 @@ internal class DetektMessageCollector(
             messages++
         }
     }
-
-    fun printIssuesCountIfAny() {
-        if (messages > 0) {
-            warningPrinter(
```
TWiStErRob, line 70: rather than fully killing put it behind the debug? or does that already print the number of problems? I should probably check 😂

TWiStErRob (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/BindingContext.kt
TWiStErRob: I double-checked, when running with debug mode, it doesn't print the number of compilation errors, just lists them all. Since the numbers are high, and they can fluctuate, I think it would be beneficial to print the number of them, although that's probably not in scope here. (javac does print error and apt problem count, but kotlinc doesn't)

BraisGabin: Out of scope as you said. We are removing this because no one cares. If someone wants it I'll be happy to review the pr but otherwise it seems a YAGNI.

schalkms (approved):

PR #5452, opened 2022-10-19
Set up Detekt report merging

TWiStErRob:
Triggered by this failure: https://github.com/detekt/detekt/actions/runs/3282055240/jobs/5404921341#step:4:397

Reasons:
 * no-one wants to read logs
 * the ability is already there, but wasn't used with type resolution
 * dogfood own Gradle task

Changes:
 * Review commit by commit.


BraisGabin (commented):
Aren't we uploading the sarif file twice? One with type solving and another without it? If we upload it with type solving we should remove the one without because it is just a sub set of the other one. (Maybe you removed it already but I didn't spot it)

.github/workflows/detekt-with-type-resolution.yaml
```diff
@@ -49,15 +49,22 @@ jobs:
...
+      - name: Run analysis
+        uses: gradle/gradle-build-action@fd32ae908111fe31afa48827bd1ee909540aa971 # tag=v2
+        with:
+          arguments: detektMain detektTest :detektReportMergeSarif
+
+      - name: Upload SARIF to Github using the upload-sarif action
+        uses: github/codeql-action/upload-sarif@cc7986c02bac29104a72998e67239bb5ee2ee110 # tag=v2
+        if: success() || failure()
+        with:
+          sarif_file: build/reports/detekt/merge.sarif
```
TWiStErRob, line 70: > Aren't we uploading the sarif file twice? One with type solving and another without it? If we upload it with type solving we should remove the one without because it is just a sub set of the other one. (Maybe you removed it already but I didn't spot it)

@BraisGabin yeah, I thought about that too, but then I think it's a good argument to notice what's happening. Is it always a subset (I think that's an assumption), this would highlight it; although no-one might actually notice.

Note that GitHub is ok uploading multiple, each GHA job will get its own unique id ("category" in terms of upload-sarif), for example: https://github.com/TWiStErRob/net.twisterrob.cinema/runs/8979749821

Let me push an example failure to test this. (If it doesn't show the same finding twice, I think it's ok to double-upload.)

BraisGabin: Sounds good. Let me know what you find out.

TWiStErRob: Testing:
```
> Task :detekt-cli:detektMain FAILED
P:\projects\contrib\github-detekt\detekt-cli\src\main\kotlin\io\gitlab\arturbosch\detekt\cli\Spec.kt:76:5: This '?: emptyList()' can be replaced with 'orEmpty()' call [UseOrEmpty]

> Task :detekt-rules-style:detekt FAILED
P:\projects\contrib\github-detekt\detekt-rules-style\src\main\kotlin\io\gitlab\arturbosch\detekt\rules\style\UseIsNullOrEmpty.kt:11:1: Unused import [NoUnusedImports]

> Task :detekt-rules-style:detektMain FAILED
P:\projects\contrib\github-detekt\detekt-rules-style\src\main\kotlin\io\gitlab\arturbosch\detekt\rules\style\UseIsNullOrEmpty.kt:11:1: Unused import [NoUnusedImports]
P:\projects\contrib\github-detekt\detekt-rules-style\src\main\kotlin\io\gitlab\arturbosch\detekt\rules\style\UseIsNullOrEmpty.kt:49:7: `UseIsNullOrEmpty` uses `bindingContext` but is not annotated with `@RequiresTypeResolution` [ViolatesTypeResolutionRequirements]
```

TWiStErRob: How it looks:
2 failed builds in https://github.com/detekt/detekt/actions/runs/3285786512
tons of failed builds in PR (but this is a different issue):
image

Nicely deduplicated annotations:
image

One line I actually changed is flagged on PR (and this needs type resolution):
image

TWiStErRob: after pushing the fix:
image

image

BraisGabin (approved):

TWiStErRob:
Resolved recent conflicts, is there anything else I need to do here?

.github/workflows/detekt-with-type-resolution.yaml
```diff
@@ -49,15 +49,22 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ !contains(github.event.head_commit.message, 'ci skip') }}
     steps:
```
TWiStErRob, line 51: Permissions are missing here, @BraisGabin, opened https://github.com/detekt/detekt/pull/5470 to fix. I guess for `push: main` the only way to test is in "production" 🤦‍♂️.

PR #5458, opened 2022-10-21
CognitiveComplexity: count else/else-if as one complexity

### LINKED ISSUE #5447: CognitiveComplexity should count else if and else as +1 complexity
## Expected Behavior
Count `else if` or `else` as +1 complexity.

## Current Behavior
Currently `else if` is assessed as nested `if`, so it counted as +1 +nesting level complexity, and `else` is not counted as complexity.
```kt
    @Test
    fun `else should counts as complexity`() {
        val code = compileContentForTest(
            """
            fun test(condition: Boolean) {
                if (condition) {
                    // do something
                } else {
                    // do something
                }
            }
            """.trimIndent()
        )
        assertThat(CognitiveComplexity.calculate(code)).isEqualTo(2) // was: 1
    }
```

```kt
    @Test
    fun `else if should counts as 1 complexity`() {
        val code = compileContentForTest(
            """
            fun test(condition: Boolean) {
                if (condition) {
                    // do something
                } else if (condition) {
                    // do something
                }
            }
            """.trimIndent()
        )
        assertThat(CognitiveComplexity.calculate(code)).isEqualTo(2) // but was: 3
    }
```

## Context
According to [paper of Cognitive Complexity](https://www.sonarsource.com/docs/CognitiveComplexity.pdf), `else if` or `else` should be counted as +1 code complexity.

> It assesses hybrid increments for:
> ● else if, elif, else, …
> No nesting increment is assessed for these structures because the mental cost has already been paid when reading the if.

Not including `else` affects the accuracy of complexity, and the explanation of why `else if` should be counted as +1 seems to be logical, so it would be good to follow the implementation of the paper.

t-kameyama:
Fixes #5447

### DIFF
--- a/detekt-metrics/src/main/kotlin/io/github/detekt/metrics/CognitiveComplexity.kt
+++ b/detekt-metrics/src/main/kotlin/io/github/detekt/metrics/CognitiveComplexity.kt
@@ -96,19 +96,21 @@ class CognitiveComplexity private constructor() : DetektVisitor() {
             if (element is KtContainerNodeForControlStructureBody && parent is KtIfExpression) {
                 when (element.node.elementType) {
                     KtNodeTypes.THEN -> {
-                        addComplexity()
                         if (parent.isElseIf()) {
-                            super.visitKtElement(element)
+                            complexity++
                         } else {
-                            nestAround { super.visitKtElement(element) }
+                            addComplexity()
                         }
+                        nestAround { super.visitKtElement(element) }
                     }
 
                     KtNodeTypes.ELSE -> {
-                        if (element.expression !is KtIfExpression) {
-                            addComplexity()
+                        if (element.expression is KtIfExpression) {
+                            super.visitKtElement(element)
+                        } else {
+                            complexity++
+                            nestAround { super.visitKtElement(element) }
                         }
-                        super.visitKtElement(element)
                     }
 
                     else ->
--- a/detekt-metrics/src/main/kotlin/io/github/detekt/metrics/CognitiveComplexity.kt
+++ b/detekt-metrics/src/main/kotlin/io/github/detekt/metrics/CognitiveComplexity.kt
@@ -1,6 +1,7 @@
 package io.github.detekt.metrics
 
 import io.gitlab.arturbosch.detekt.api.DetektVisitor
+import io.gitlab.arturbosch.detekt.rules.isElseIf
 import org.jetbrains.kotlin.com.intellij.openapi.util.Key
 import org.jetbrains.kotlin.com.intellij.psi.tree.IElementType
 import org.jetbrains.kotlin.lexer.KtTokens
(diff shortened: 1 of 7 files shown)


detekt-metrics/src/test/kotlin/io/github/detekt/metrics/CognitiveComplexitySpec.kt
```diff
@@ -282,4 +282,98 @@ class CognitiveComplexitySpec {
...
+                        if (condition) { // +2
+                            while(true) { // +3
+                            }
+                        } else if (condition) { // +2
+                            while(true) { // +3
+                            }
+                        } else { // +2
+                            while(true) { // +3
+                            }
+                        }
```
sanggggg, line 347: From  #5447 and [paper of Cognitive Complexity](https://www.sonarsource.com/docs/CognitiveComplexity.pdf),
> No nesting increment is assessed for these structures because the mental cost has already been paid when reading the if.

`else` and `else if` seem to be counted as +1 complexity only (do not include nesting level increments).

t-kameyama: https://github.com/detekt/detekt/pull/5458/commits/693fa8aa59677411ebfa51040c73294de0aa342e

sanggggg: I'm sorry @t-kameyama, I said something that could be misunderstood.
What I'm trying to say is

`else if` and `else`
- should add +1 `complexity` (Increments)
- should increase `nesting level` (Nesting Level)
- should not add `complexity by nesting level` (**no** Nesting increments)

Screen Shot 2022-10-23 at 1 02 10 AM

---

Thus, the complexity of the test code should be 39
```
                fun test(condition: Boolean) {
                    if (condition) { // +1
                        if (condition) { // +2
                            while(true) { // +3
                            }
                        } else if (condition) { // +1
                            while(true) { // +3
                            }
                        } else if (condition) { // +1
                            while(true) { // +3
                            }
                        } else { // +1
                            while(true) { // +3
                            }
                        }
                    } else if (condition) { // +1
                        if (condition) { // +2
                            while(true) { // +3
                            }
                        } else if (condition) // +1
                            while(true) { // +3
                            }
                    } else { // + 1
                        if (condition) { // +2
                            while(true) { // +3
                            }
                        } else // +1
                            while(true) { // +3
                            }
                    }
                    if (condition) { // +1
                    }
                }
                """
                // total 39 comple
```

The complexity suddenly increased, but I think this is a more real complexity.

t-kameyama: [6b06079](https://github.com/detekt/detekt/pull/5458/commits/6b060793728235cb197d61a0da0815037e2d762b)

BraisGabin (approved):
LGTM. @sanggggg it does follow what you were asking, right?

sanggggg (approved):
Yes, thank you for implementation! 🚀

detekt-metrics/src/main/kotlin/io/github/detekt/metrics/CognitiveComplexity.kt
```diff
@@ -88,9 +91,34 @@ class CognitiveComplexity private constructor() : DetektVisitor() {
...
+                            addComplexity()
+                        }
+                        nestAround { super.visitKtElement(element) }
+                    }
+
+                    KtNodeTypes.ELSE -> {
+                        if (element.expression is KtIfExpression) {
+                            super.visitKtElement(element)
+                        } else {
+                            complexity++
```
chao2zhang, line 111: 💯 since an else branch only increment the complexity without adding nesting

chao2zhang (approved):

chao2zhang (approved):

PR #5461, opened 2022-10-21
Add kure-potlin to marketplace

schalkms:


website/src/data/marketplace.js
```diff
@@ -110,6 +110,27 @@ export const extensions = [
...
+      "LoopDefinition",
+      "ReturnStatement",
+      "VariableDefinition",
+      "ReturnUnit",
+      "ClassDefinition",
+      "AbstractClassDefinition",
+      "ThrowExpression",
+      "MutableCollections",
+      "BranchStatement",
+      "MissingElse"
```
cortinico, line 130: Can we sort them?

website/src/data/marketplace.js
```diff
@@ -110,6 +110,27 @@ export const extensions = [
     usesTypeResolution: false,
     tags: ["ruleset"],
   },
+  {
+    title: "Kure",
+    description:
+      "The purpose of this ruleset for functional programming is to report the potential use of impure language elements in kotlin code.",
+    repo: "https://github.com/neeffect/kure-potlin",
+    tags: ["plugin", "ruleset"],
```
cortinico, line 118:
```suggestion
    tags: ["ruleset"],
```

cortinico (commented):
Thanks for adding this ruleset 👍

website/src/data/marketplace.js
schalkms: Hmm, sure, but all the other rulesets don't apply any sorting order. I copied it them from the readme file in the original repo.

schalkms: Hmm, sure, but all the other rulesets don't apply any sorting order. I copied them from the readme file in the original repo.

schalkms: I could work over the whole file and submit a separate PR. I think that would be easier.

cortinico: Yup that's not a blocker 👍

cortinico (approved):

PR #5474, opened 2022-10-24
Sort marketplace rules alphabetically

schalkms:
see comment: https://github.com/detekt/detekt/pull/5461#discussion_r1002420854


TWiStErRob (approved):
- [x] Clear moves
- [x] Alphabetical every rules array
- [x] Commas are well placed

website/src/data/marketplace.js
```diff
@@ -118,16 +118,16 @@ export const extensions = [
...
       "ReturnStatement",
-      "VariableDefinition",
       "ReturnUnit",
-      "ClassDefinition",
-      "AbstractClassDefinition",
       "ThrowExpression",
-      "MutableCollections",
-      "BranchStatement",
-      "MissingElse"
+      "VariableDefinition"
```
TWiStErRob, line 130: Since this is typescript, maybe add trailing commas to each of the `rules` arrays? so that it's consistent and less confusing on next contributions.

website/src/data/marketplace.js
```diff
@@ -175,7 +175,7 @@ export const extensions = [
       "A ruleset which enables verifying whether concrete classes are implemented as specified according to annotations applied to base types.",
     repo: "https://github.com/cph-cachet/detekt-verify-implementation",
     ruleset: "verify-implementation",
-    rules: ["Immutable", "DataClass"],
+    rules: ["DataClass", "Immutable"],
```
TWiStErRob, line 178: wrap down for consistency?
```suggestion
    rules: [
      "DataClass",
      "Immutable",
    ],
```

website/src/data/marketplace.js
schalkms: DONE

website/src/data/marketplace.js
schalkms: This would require reworking the whole file, not just this line. As long as a single line doesn't become too long, I'm fine with this format.

BraisGabin (approved):

PR #5480, opened 2022-10-26
TrimMultilineRawString: fix false positive when it's expected as constant

### LINKED ISSUE #5476: `TrimMultilineRawString` false-positive on annotation parameters
## Observed Behavior
```kotlin
    @DisplayName(
        """
            Given something
            When something
            Then something
        """,
    )
```

`TrimMultilineRawString` reports raw strings inside annotations. The problem is that the compiler doesn't allow to add `.trimIndent()` on that scope so we should disable this rule for this case.

## Your Environment
* Version of detekt used: 1.22.0-RC2

t-kameyama:
Fixes #5476


BraisGabin (approved):
Thanks as always!

PR #5490, opened 2022-10-31
Stop including root build unnecessarily in detekt-gradle-plugin

3flex:
This isn't currently used for anything.


schalkms (approved):

VitalyVPinchuk:
> This isn't currently used for anything.

Great!
Due to this includeBuild it was impossible to use Gradle option --includeBuild to substitute published dependency with local one.

PR #5492, opened 2022-10-31
Add detekt compiler plugin to main project

### LINKED ISSUE #3520: Add compiler plugin to main project
## Expected Behavior
Recommend detekt-compiler-plugin to enable type resolution.

## Current Behavior

Manual setup or flaky automated creation of Gradle tasks to enable features requiring type resolution.

## Context

Make type resolution easier to activate and faster during builds.

3flex:
This brings the detekt compiler plugin into the primary project. Aside from bringing the code over, this also makes changes to the Gradle plugin so that the standard plugin as well as the compiler plugin can be used in the same project. This wasn't previously possible due to the naming conflict on the `detekt` extension which was applied by both Gradle plugins.

I'd like to get this merged, then we can work on:
* Confirming the versioning scheme for the compiler plugin which will be sensitive to the Kotlin version the target project is using. I propose just copying KSP's version scheme of `$kotlinVersion-$pluginVersion` e.g. `1.7.20-1.22.0-RC2`
* Docs
* Pushing out at least one release (can then skip step 1 below)
* Work on a fix for `InvalidPackageDeclaration` failures (means we can skip steps 3 & 4 below)

To test:
1. Include these builds in the test project's settings.gradle(.kts) file:
```
includeBuild("../detekt/detekt-gradle-plugin")
includeBuild("../detekt")
```
2. Apply the compiler plugin in a target project:
```kotlin
plugins {
    id("io.github.detekt.gradle.compiler-plugin")
}
```
3. Add a `detekt.yml` config file and disable `InvalidPackageDeclaration` (see [detekt-compiler-plugin/issues/15](https://github.com/detekt/detekt-compiler-plugin/issues/15)):
```yaml
naming:
  InvalidPackageDeclaration:
    active: false
```
4. Set `buildUponDefaultConfig = true`, e.g.:
```kotlin
detekt {
    buildUponDefaultConfig = true
}
```

The compiler plugin cannot be applied to the detekt project itself due to circular dependencies, which I couldn't work around, but hope to some time in future.

Closes #3520


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -102,6 +106,7 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
         const val DEFAULT_REPORT_ENABLED_VALUE = true
         const val DEFAULT_ALL_RULES_VALUE = false
         const val DEFAULT_BUILD_UPON_DEFAULT_CONFIG_VALUE = false
+        const val DEFAULT_COMPILER_PLUGIN_ENABLED = true
```
chao2zhang, line 109: We should by default disable the detekt extension, but enable the detekt compiler in our own repo.

chao2zhang (approved):

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
cortinico: +1 to this

detekt-gradle-plugin/src/main/kotlin/io/github/detekt/gradle/Constants.kt
```diff
@@ -0,0 +1,7 @@
+package io.github.detekt.gradle
+
+const val DETEKT_NAME = "detekt"
```
cortinico, line 3: Can we move this inside the plugin as we do for `DetektPlugin.kt`?

detekt-compiler-plugin/build.gradle.kts
```diff
@@ -0,0 +1,117 @@
...
+    mergeServiceFiles()
+    dependencies {
+        include(dependency("io.gitlab.arturbosch.detekt:.*"))
+        include(dependency("io.github.detekt:.*"))
+        include(dependency("org.yaml:snakeyaml"))
+        include(dependency("io.github.davidburstrom.contester:contester-breakpoint"))
+    }
+}
+
+val verifyKotlinCompilerDownload by tasks.registering(Verify::class) {
```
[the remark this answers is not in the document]
cortinico: > What do you think about maintaining the kotlinc in our codebase?

Are you referring to the zipfile? I'm a bit against commit it into the repo due to size implications

cortinico (approved):

detekt-compiler-plugin/build.gradle.kts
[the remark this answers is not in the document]
3flex: GitHub is the official source for downloading the Kotlin command-line compiler:  https://kotlinlang.org/docs/command-line.html

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
3flex: This value is only set to true because it's a no-op unless the compiler plugin is also applied. It then allows you to easily disable the plugin programmatically. There may be another way to make this happen, but with the current setup, leaving it true has the least surprising behaviour.

chao2zhang: If you could add comments to this line to clarify the purpose of this property, it would be great so that future readers don't get confused.

3flex:
I'll rebase this

3flex:
I propose to add docs in a separate PR. I'll open an issue once this is merged for tracking.

BraisGabin (approved):

chao2zhang (approved):

PR #5495, opened 2022-11-01
Document and test edge cases for ForbiddenMethodCall function signatures

dzirbel:
I was stuck trying to match some function signatures (from Timber on Android) for the ForbiddenMethodCall rule and had to test inside detekt to figure out how to do it properly. To make this more clear in the future, I've expanded the description of the methods configuration parameter and added more tests for these cases. Some of this might be obvious to some but it was not to me at the time so I think having some quick tips might be helpful; or perhaps a link to a more comprehensive guide on Kotlin function signatures (if one exists).


schalkms (commented):
You might have produced a name clash, which causes the `compileSnippets` task to fail.
The function named `test` might be the issue here. If this issue is resolved, I'd say LGTM.

Thank you for improving detekt's docs. These improvements are very much appreciated.

dzirbel:
> You might have produced a name clash, which causes the `compileSnippets` task to fail. The function named `test` might be the issue here. If this issue is resolved, I'd say LGTM.

Yep, it was just a copy-paste error between the Array and List tests - should be fixed now :crossed_fingers:

BraisGabin:
Thanks! If you need to investigate further I recommend you to check `FunctionMatcherSpec.kt`. It contains a lot of tests about how to parse the function signatures. It's a really complex topic.

PR #5505, opened 2022-11-04
MultilineLambdaItParameter: fix false positive for one-line statements with a lambda argument

### LINKED ISSUE #5504: MultilineLambdaItParameter: false positive for one-line statements with a lambda argument
## Steps to Reproduce

```kotlin
@Nested
inner class `single parameter, single-line lambda` {
    @Test
    fun `does not report when statement has one-line lambda argument`() {
        val code = """
        fun f() {
            val digits = 1234.let {
                check(it > 0) { println(it) }
            }
        }
        """.trimIndent()
    }
}
```

## Context
https://github.com/detekt/detekt/pull/5397#issuecomment-1302776667

t-kameyama:
Fixes #5504

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
@@ -12,8 +12,10 @@ import io.gitlab.arturbosch.detekt.rules.IT_LITERAL
 import io.gitlab.arturbosch.detekt.rules.hasImplicitParameterReference
 import io.gitlab.arturbosch.detekt.rules.implicitParameter
 import org.jetbrains.kotlin.psi.KtBlockExpression
+import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtLambdaExpression
-import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
+import org.jetbrains.kotlin.psi.psiUtil.anyDescendantOfType
+import org.jetbrains.kotlin.utils.ifEmpty
 
 /**
  * Lambda expressions are very useful in a lot of cases, and they often include very small chunks of
@@ -76,8 +78,10 @@ class MultilineLambdaItParameter(val config: Config) : Rule(config) {
 
     override fun visitLambdaExpression(lambdaExpression: KtLambdaExpression) {
         super.visitLambdaExpression(lambdaExpression)
-        val size = lambdaExpression.collectDescendantsOfType<KtBlockExpression>().sumOf { it.statements.size }
-        if (size <= 1) return
+
+        val statements = lambdaExpression.bodyExpression?.statements.orEmpty().ifEmpty { return }
+        val single = statements.singleOrNull()
+        if (single != null && (single.hasNoLineBreak() || single.hasNoStatements())) return
 
         val parameterNames = lambdaExpression.valueParameters.map { it.name }
         // Explicit `it`
@@ -106,4 +110,9 @@ class MultilineLambdaItParameter(val config: Config) : Rule(config) {
             }
         }
     }
+
+    private fun KtExpression.hasNoLineBreak() = !textContains('\n')
+
+    private fun KtExpression.hasNoStatements() =
+        !anyDescendantOfType<KtBlockExpression> { it.statements.isNotEmpty() }
 }
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameterSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameterSpec.kt
@@ -129,6 +129,24 @@ class MultilineLambdaItParameterSpec(val env: KotlinCoreEnvironment) {
             val findings = subject.compileAndLintWithContext(env, code)
             assertThat(findings).hasSize(1)
         }
+
+        @Test
+        fun `does not report when statement has no nested statements`() {
+            val code = """
+            data class Foo(val x: Int, val y: Int)
+            
+            fun f(i: Int?) {
+                val foo = i?.let {
+                    Foo(
+                        x = it,
+                        y = 2
+                    )
+                }
+            }
+            """.trimIndent()
+            val findings = subject.compileAndLintWithContext(env, code)
+            assertThat(findings).isEmpty()
+        }
     }
 
     @Nested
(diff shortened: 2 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
```diff
@@ -106,4 +110,9 @@ class MultilineLambdaItParameter(val config: Config) : Rule(config) {
             }
         }
     }
+
+    private fun KtExpression.hasNoLineBreak() = !textContains('\n')
```
TWiStErRob, line 114: You're assuming file separators, which will work for most, but some might use `\r` only.
Is it possible to get the beginning/ending line number of the PsiElement and calculate how many lines it is?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameterSpec.kt
```diff
@@ -171,6 +189,19 @@ class MultilineLambdaItParameterSpec(val env: KotlinCoreEnvironment) {
...
+            val findings = subject.compileAndLintWithContext(env, code)
+            assertThat(findings).isEmpty()
+        }
+
+        @Test
+        fun `does not report when statement has one-line lambda argument`() {
+            val code = """
+            fun f() {
+                val digits = 1234.let {
+                    check(it > 0) { println(it) }
```
TWiStErRob, line 203: Which `it` is this test targeting? it's unclear what the goal of the test is, could you please clarify it?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameterSpec.kt
```diff
@@ -129,6 +129,24 @@ class MultilineLambdaItParameterSpec(val env: KotlinCoreEnvironment) {
             val findings = subject.compileAndLintWithContext(env, code)
             assertThat(findings).hasSize(1)
         }
+
+        @Test
+        fun `does not report when statement has no nested statements`() {
```
TWiStErRob, line 134: is there a positive test for this? for example `reports when statement has nested statements`?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
```diff
@@ -76,8 +78,10 @@ class MultilineLambdaItParameter(val config: Config) : Rule(config) {
 
     override fun visitLambdaExpression(lambdaExpression: KtLambdaExpression) {
         super.visitLambdaExpression(lambdaExpression)
-        val size = lambdaExpression.collectDescendantsOfType<KtBlockExpression>().sumOf { it.statements.size }
-        if (size <= 1) return
+
+        val statements = lambdaExpression.bodyExpression?.statements.orEmpty().ifEmpty { return }
+        val single = statements.singleOrNull()
+        if (single != null && (single.hasNoLineBreak() || single.hasNoStatements())) return
```
TWiStErRob, line 84: It feels like semantically this might be cleaner (return early + renames), but up to you:
```suggestion
        val statement = statements.singleOrNull() ?: return
        if (statement.isSingleLine() || statement.hasNoSubStatements()) return
```

TWiStErRob:
Amazingly fast response @t-kameyama 👏🏻!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
t-kameyama: Line separators are converted to `\n`.

https://github.com/detekt/detekt/blob/ec6cd40a6fa013c3f2dc4ce8343c4f9a1bd4d334/detekt-parser/src/main/kotlin/io/github/detekt/parser/KtCompiler.kt#L33

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameterSpec.kt
t-kameyama: https://github.com/detekt/detekt/blob/993cb9c255e5f60e41b075476e670fd4e0e4c63b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameterSpec.kt#L118-L131

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
t-kameyama: > val statement = statements.singleOrNull() ?: return

We can't return here because we need to flag multi-line cases.

TWiStErRob: Oh, have to remember to expand the code below the changed lines. Agreed. Please consider renames anyway.

TWiStErRob (approved):

TWiStErRob:
cc @eygraber

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameterSpec.kt
```diff
@@ -121,14 +121,32 @@ class MultilineLambdaItParameterSpec(val env: KotlinCoreEnvironment) {
...
+            val code = """
+            data class Foo(val x: Int, val y: Int)
+            
+            fun f(i: Int?) {
+                val foo = i?.let {
+                    Foo(
+                        x = it,
```
eygraber, line 141: I'm a little less sure about this case, since it's multiline and the `it` isn't the first line.

Not sure if the example is too generic, but I'm trying to look at this test through the lens of a reader not familiar with the code, and it's not immediately clear to me what the `it` represents here. Although I could think of examples where it is clear.

eygraber (approved):
Thanks for the quick response!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameter.kt
TWiStErRob: Hmm, a related thought: you're assuming here that Detekt compiles the PSI, but what if it's run in a different environment? https://github.com/detekt/detekt/pull/5492 at that point all assumptions about line breaks will fail in Detekt rules. cc @3flex

I think checking the line numbers to detect single line is still the most versatile solution.

BraisGabin (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MultilineLambdaItParameterSpec.kt
chao2zhang: Agree with @eygraber . The following block is non-compliant according to the kDoc.

```
 val digits = 1234.let { it ->
      println(it)
      listOf(it)
 }
```

Let's keep the behavior to be consistent with the spec name `MultilineLambdaItParameter`

cortinico (approved):

Issue #2719, opened 2020-05-22
New Rule: Forbidden Annotation


gtcompscientist:
**ForbiddenAnnotation**

## Expected Behavior of the rule
- Allows for certain Annotations to be disallowed at the project level. This is slightly different from ForbiddenImports as there are some times when an import might not also match to the Annotation.
- If you don't enumerate ALL imports (ie: allow .* imports)
- If an Annotation doesn't require an import (ie: `@SuppressWarnings`)

## Context
Kotlin code should use Kotlin's `@Suppress` annotation, not Java's `@SuppressWarnings` and there are other annotations that we would like to not have in our code.

gtcompscientist:
Does `visitAnnotation()` guarantee that it visits all Annotations? ie:

- Class
- Property
- Function
- Parameter
- File

cortinico:
> Does `visitAnnotation()` guarantee that it visits all Annotations? ie:

It's hard to say as the PSI api is not really well documented. The general suggestion is to play around with the PSI-Viewer plugin and/or write some tests to validate all the scenarios.

arturbosch:
We have a `ForbiddenImport` rule which can be used to forbid annotations.
Does it make sense to offer an specialized version for annotations?

cortinico:
Context from this slack thread: https://kotlinlang.slack.com/archives/C88E12QH4/p1590074412134700

I'm also not entirely sure we need a `ForbiddenAnnotations` rule.

@gtcompscientist scenario is involving annotations (such as `java.lang.SuppressWarning`) that don't have an explicit import and are impossible to catch with the `ForbiddenImport` rule.

BraisGabin:
I never open an issue but I would like to support the same for classes too. For example, I would like to forbid the use of `java.lang.Math` and `java.lang.Runnable`. And there's other case: if the class is in the same package. So maybe the rule could be called `ForbiddenClass` and check all the cases: Classes, annotations and objects.

I know that I would use a rule like that but I don't know how much value could it bring for the community in general.

gtcompscientist:
@BraisGabin I think that's a really interesting adaptation. Definitely the kind of thing that would be very useful for pushing people towards the Kotlin implementations.

cortinico:
> For example, I would like to forbid the use of `java.lang.Math` and `java.lang.Runnable`

Small side note: there is a `ForbiddenMethodCall` rule that will most likely cover the use case.

gtcompscientist:
@cortinico 

[...]

Issue #4750, opened 2022-04-23
UnnecessaryAbstractClass title is too verbose


TWiStErRob:
## Expected Behavior of the rule

UnnecessaryAbstractClass should have a concise description like other rules

## Context
I just updated Detekt 1.19.0 to 1.20.0 and got this failure:
https://github.com/TWiStErRob/net.twisterrob.cinema/security/code-scanning/227

The message is just plain weird to be this long:
![image](https://user-images.githubusercontent.com/2906988/164910306-4b4478c4-06c2-45f4-8b4d-132d11a76623.png)

detail screen:
![image](https://user-images.githubusercontent.com/2906988/164910247-930c8fdf-1dcd-40ad-9ae9-bd8268548bb9.png)

TWiStErRob:
I noticed MagicNumber to have the same problem, but many others look perfectly good. Should there be a check on how long the description can be? and/or introduce: "title" and "detail" descriptions?

BraisGabin:
I'm not sure if this issue is in the description or it is on our impregnation of Sarif where we are using the wrong description. The rules have 3 different descriptions: kdoc, description about the role and description about the issue. It seems that we use the second where we should use the third.

TWiStErRob:
So you're thinking about swapping the two? Because we need both, see the description on the line itself in the screenshots.

BraisGabin:
My bad, I dind't see the screenshots with detail. Yes, I agree that this description is long an it's not that useful. We should change it. We even have a style guide for this: https://github.com/detekt/detekt/blob/main/.github/CONTRIBUTING.md#rule-descriptions

3flex:
Is this a dupe of #4163?

BraisGabin:
🤔 It is but I'm not sure which one I prefer. This issue is easily actionable. The bug is at `UnnecessaryAbstractClass`. The other is really generic and if no one is going to do the reseach to list all the rules that should be fixed I would vote to close it and fix the ones that are reported (or just send a PR if someone find an issue like this in other rule).

3flex:
Makes sense, I'll close the other one

TWiStErRob:
I think this is more like a subtask of the other one.

BraisGabin:
[...]

PR #5515, opened 2022-11-07
ForbiddenAnnotation rule #2719

ov7a:
As discussed in #2719, this rule could be sometimes useful for cases when an annotation does not require import: `java.lang.SuppressWarnings`, `kotlin.jvm.Transient`, etc. These cases can't be caught with the `ForbiddenImport` rule.
Also, this rule can be useful for pushing people towards the Kotlin implementations (see [this issue](https://github.com/jhipster/jhipster-kotlin/issues/143) for an example).

More specifically, forbidding `kotlin.jvm.Transient` can be useful in applications with `kotlinx.serialization`: `kotlin.jvm.Transient` does not affect `@Serializable` classes. `ForbiddenAnnotation` can be a quick-fix for that.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -528,6 +528,11 @@ style:
   ExpressionBodySyntax:
     active: false
     includeLineWrapping: false
+  ForbiddenAnnotation:
+    active: false
+    annotations:
+      - reason: 'it is a java annotation. Use `Suppress` instead.'
+        value: 'java.lang.SuppressWarnings'
   ForbiddenComment:
     active: true
     values:
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -533,6 +533,18 @@ style:
     annotations:
       - reason: 'it is a java annotation. Use `Suppress` instead.'
         value: 'java.lang.SuppressWarnings'
+      - reason: 'it is a java annotation. Use `kotlin.Deprecated` instead.'
+        value: 'java.lang.Deprecated'
+      - reason: 'it is a java annotation. Use `kotlin.annotation.MustBeDocumented` instead.'
+        value: 'java.lang.annotation.Documented'
+      - reason: 'it is a java annotation. Use `kotlin.annotation.Target` instead.'
+        value: 'java.lang.annotation.Target'
+      - reason: 'it is a java annotation. Use `kotlin.annotation.Retention` instead.'
+        value: 'java.lang.annotation.Retention'
+      - reason: 'it is a java annotation. Use `kotlin.annotation.Repeatable` instead.'
+        value: 'java.lang.annotation.Repeatable'
+      - reason: 'Kotlin does not support @Inherited annotation, see https://youtrack.jetbrains.com/issue/KT-22265'
+        value: 'java.lang.annotation.Inherited'
   ForbiddenComment:
     active: true
     values:
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
@@ -50,11 +50,13 @@ class ForbiddenAnnotation(config: Config = Config.empty) : Rule(config) {
         valuesWithReason(
             "java.lang.SuppressWarnings" to "it is a java annotation. Use `Suppress` instead.",
             "java.lang.Deprecated" to "it is a java annotation. Use `kotlin.Deprecated` instead.",
-            "java.lang.annotation.Documented" to "it is a java annotation. Use `kotlin.annotation.MustBeDocumented` instead.",
+            "java.lang.annotation.Documented" to "it is a java annotation. " +
+                "Use `kotlin.annotation.MustBeDocumented` instead.",
             "java.lang.annotation.Target" to "it is a java annotation. Use `kotlin.annotation.Target` instead.",
             "java.lang.annotation.Retention" to "it is a java annotation. Use `kotlin.annotation.Retention` instead.",
             "java.lang.annotation.Repeatable" to "it is a java annotation. Use `kotlin.annotation.Repeatable` instead.",
-            "java.lang.annotation.Inherited" to "Kotlin doesn't support @Inherited annotation, see https://youtrack.jetbrains.com/issue/KT-22265",
+            "java.lang.annotation.Inherited" to "Kotlin does not support @Inherited annotation, " +
+                "see https://youtrack.jetbrains.com/issue/KT-22265",
         )
     ) { list ->
         list.associate { it.value to Forbidden(it.value, it.reason) }
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
@@ -8,6 +8,7 @@ import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Location
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.ValueWithReason
 import io.gitlab.arturbosch.detekt.api.config
 import io.gitlab.arturbosch.detekt.api.internal.Configuration
 import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
@@ -49,7 +50,7 @@ class ForbiddenAnnotation(config: Config = Config.empty) : Rule(config) {
     @Configuration(
         "List of fully qualified annotation classes which are forbidden."
     )
-    private val annotations: Map<String, Forbidden> by config(
+    private val annotations: Map<String, ValueWithReason> by config(
         valuesWithReason(
             "java.lang.SuppressWarnings" to "it is a java annotation. Use `Suppress` instead.",
             "java.lang.Deprecated" to "it is a java annotation. Use `kotlin.Deprecated` instead.",
@@ -62,7 +63,7 @@ class ForbiddenAnnotation(config: Config = Config.empty) : Rule(config) {
                 "see https://youtrack.jetbrains.com/issue/KT-22265",
         )
     ) { list ->
-        list.associate { it.value to Forbidden(it.value, it.reason) }
+        list.associateBy { it.value }
     }
 
     override fun visitAnnotationEntry(annotation: KtAnnotationEntry) {
@@ -92,9 +93,9 @@ class ForbiddenAnnotation(config: Config = Config.empty) : Rule(config) {
 
         if (forbidden != null) {
             val message = if (forbidden.reason != null) {
-                "The annotation `${forbidden.name}` has been forbidden: ${forbidden.reason}"
+                "The annotation `${forbidden.value}` has been forbidden: ${forbidden.reason}"
             } else {
-                "The annotation `${forbidden.name}` has been forbidden in the detekt config."
+                "The annotation `${forbidden.value}` has been forbidden in the detekt config."
             }
             val location = Location.from(element).let { location ->
                 location.copy(
@@ -106,8 +107,6 @@ class ForbiddenAnnotation(config: Config = Config.empty) : Rule(config) {
             report(CodeSmell(issue, Entity.from(element, location), message))
         }
     }
-
-    private data class Forbidden(val name: String, val reason: String?)
 
     private fun KtTypeReference.fqNameOrNull(): FqName? =
         bindingContext[BindingContext.TYPE, this]?.fqNameOrNull()
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
@@ -35,9 +35,7 @@ class ForbiddenAnnotation(config: Config = Config.empty) : Rule(config) {
     override val issue = Issue(
         javaClass.simpleName,
         Severity.Style,
-        "Mark forbidden annotations. A forbidden annotation could be an invocation of an unwanted " +
-            "language annotation which does not require explicit import and hence you might want to mark it " +
-            "as forbidden in order to get warned about the usage.",
+        "Avoid using this annotation.",
         Debt.FIVE_MINS
     )
 
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
@@ -110,19 +110,9 @@ class ForbiddenAnnotation(config: Config = Config.empty) : Rule(config) {
 
     private data class Forbidden(val name: String, val reason: String?)
 
-    private fun KtTypeReference.fqNameOrNull(): FqName? {
-        return if (bindingContext != BindingContext.EMPTY) {
-            bindingContext[BindingContext.TYPE, this]?.fqNameOrNull()
-        } else {
-            null
-        }
-    }
+    private fun KtTypeReference.fqNameOrNull(): FqName? =
+        bindingContext[BindingContext.TYPE, this]?.fqNameOrNull()
 
-    private fun KtExpression.expressionTypeOrNull(): KotlinType? {
-        return if (bindingContext != BindingContext.EMPTY) {
-            bindingContext[BindingContext.EXPRESSION_TYPE_INFO, this]?.type
-        } else {
-            null
-        }
-    }
+    private fun KtExpression.expressionTypeOrNull(): KotlinType? =
+        bindingContext[BindingContext.EXPRESSION_TYPE_INFO, this]?.type
 }
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
@@ -5,6 +5,7 @@ import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Debt
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Location
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.config
@@ -15,6 +16,7 @@ import io.gitlab.arturbosch.detekt.rules.fqNameOrNull
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.psi.KtAnnotationEntry
 import org.jetbrains.kotlin.psi.KtTypeReference
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
 import org.jetbrains.kotlin.resolve.BindingContext
 
 /**
@@ -67,7 +69,14 @@ class ForbiddenAnnotation(config: Config = Config.empty) : Rule(config) {
             } else {
                 "The annotation `${forbidden.name}` has been forbidden in the detekt config."
             }
-            report(CodeSmell(issue, Entity.from(annotation), message))
+            val location = Location.from(annotation).let { location ->
+                location.copy(
+                    text = location.text.copy(
+                        end = annotation.children.firstOrNull()?.endOffset ?: location.text.end
+                    )
+                )
+            }
+            report(CodeSmell(issue, Entity.from(annotation, location), message))
         }
     }
 
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
@@ -1,0 +1,83 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
+import io.gitlab.arturbosch.detekt.api.valuesWithReason
+import io.gitlab.arturbosch.detekt.rules.fqNameOrNull
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtAnnotationEntry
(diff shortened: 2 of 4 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotationSpec.kt
```diff
@@ -0,0 +1,173 @@
...
+    fun `should report nothing when annotations are blank`() {
+        val code = """
+        @SuppressWarnings("unused") 
+        fun main() {}
+        """.trimIndent()
+        val findings = ForbiddenAnnotation(
+            TestConfig(mapOf(ANNOTATIONS to listOf("  ")))
+        ).compileAndLintWithContext(env, code)
+        assertThat(findings).isEmpty()
+    }
```
BraisGabin, line 45: You can remove this test. I imagine this is copied from `ForbiddenImport` and we had it there for "historical reasons". But it is not needed here.

BraisGabin (approved):
Good contribution! I really like this one!

I'm blocking this PR because we are really near to cut a new version and it's a bit late to add a new rule to it after 3 RC but it should be inside the next version :)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotationSpec.kt
ov7a: Yep, exactly! Thank you!

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -528,6 +528,11 @@ style:
   ExpressionBodySyntax:
     active: false
     includeLineWrapping: false
+  ForbiddenAnnotation:
+    active: false
+    annotations:
+      - reason: 'it is a java annotation. Use `Suppress` instead.'
+        value: 'java.lang.SuppressWarnings'
```
TWiStErRob, line 535: Maybe add similar defaults for Java annotation annotations?

i.e. java.lang.annotation.* (Retention/Target for sure, maybe Documented/Inherited/Repeatable) most of these have their corresponding `kotlin.annotation.*` equivalent.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
```diff
@@ -0,0 +1,83 @@
...
+ */
+@RequiresTypeResolution
+class ForbiddenAnnotation(config: Config = Config.empty) : Rule(config) {
+
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Style,
+        "Mark forbidden annotations. A forbidden annotation could be an invocation of an unwanted " +
+            "language annotation which does not require explicit import and hence you might want to mark it " +
+            "as forbidden in order to get warned about the usage.",
```
TWiStErRob, line 38: Careful with lengthy descriptions like this, see https://github.com/detekt/detekt/issues/4750 why it could be a problem. Shorten to a concise description and move other info, like how to action it to the KDoc on the class.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
```diff
@@ -0,0 +1,83 @@
...
+    )
+
+    @Configuration(
+        "List of fully qualified annotation classes which are forbidden. " +
+            "For example, `kotlin.jvm.Transient`."
+    )
+    private val annotations: Map<String, Forbidden> by config(
+        valuesWithReason(
+            "java.lang.SuppressWarnings" to "it is a java annotation. Use `Suppress` instead.",
+        )
```
TWiStErRob, line 49: is this going to be the default reason for any annotation?

detekt-core/src/main/resources/default-detekt-config.yml
TWiStErRob: + Deprecated

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotationSpec.kt
```diff
@@ -0,0 +1,165 @@
...
+import io.gitlab.arturbosch.detekt.test.TestConfig
+import io.gitlab.arturbosch.detekt.test.assertThat
+import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
+import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
+import org.junit.jupiter.api.Test
+
+private const val ANNOTATIONS = "annotations"
+
+@KotlinCoreEnvironmentTest
+class ForbiddenAnnotationSpec(val env: KotlinCoreEnvironment) {
```
TWiStErRob, line 14: add test for nested annotations, e.g. forbid `ReplaceWith` only, and test that it's caught when used inside kotlin's Deprecated.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
```diff
@@ -0,0 +1,83 @@
...
+            annotations[it.asString()]
+        }
+
+        if (forbidden != null) {
+            val message = if (forbidden.reason != null) {
+                "The annotation `${forbidden.name}` has been forbidden: ${forbidden.reason}"
+            } else {
+                "The annotation `${forbidden.name}` has been forbidden in the detekt config."
+            }
+            report(CodeSmell(issue, Entity.from(annotation), message))
```
TWiStErRob, line 70: please make sure you flag only the annotation, not the parameters, i.e.
```
@Deprecated(message = "abc")
^^^^^^^^^^^
```

TWiStErRob:
@BraisGabin we can unblock this, right? Unless we need to keep `main` clean for a few days/weeks for 1.22.1?

BraisGabin:
No need to block it any longer.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
ov7a: I think not. For example, `@Inherited` isn't supported by Kotlin at all https://youtrack.jetbrains.com/issue/KT-22265

TWiStErRob: aaaah, this is the default param for a config :) so the default.yml will be generated from here I guess.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotationSpec.kt
```diff
@@ -0,0 +1,195 @@
...
```
TWiStErRob, line 14: add a test for aliased annotations to make sure it's not a false negative
```kotlin
typealias Dep = java.lang.Deprecated
@Dep
fun f() = Unit
```

TWiStErRob, line 14: add test for blocks
```kotlin
fun f(list: List<String>) {
    list.map @Suppress("x") { it.length }
}
```

TWiStErRob, line 14: add tests for expressions
```
val num = @Suppress("MagicNumber") 42

val x = @Suppress("UnnecessaryParentheses") (((1)+(2))) + 3
```

TWiStErRob: ... I know, there are some weird places where you can place annotations 😅

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotationSpec.kt
ov7a: Not sure if I solved this problem in a best way possible, but at least it works:) Thanks for great comments!

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotationSpec.kt
```diff
@@ -0,0 +1,251 @@
...
+
+    @Test
+    fun `should report SuppressWarnings usages by default`() {
+        val code = """
+        @SuppressWarnings("unused")
```
cortinico, line 19:
```suggestion
        @SuppressWarnings("unused")    
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotationSpec.kt
```diff
@@ -0,0 +1,251 @@
...
+        import java.lang.annotation.Retention 
+        import java.lang.annotation.Documented 
+        import java.lang.annotation.Target
+        import java.lang.annotation.Repeatable
+        import java.lang.annotation.Inherited
+        import java.lang.annotation.RetentionPolicy
+        import java.lang.annotation.ElementType
+        import java.lang.Deprecated
+        @Deprecated
+        @Documented
```
cortinico, line 48:
```suggestion
        @Documented
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
```diff
@@ -0,0 +1,124 @@
...
+import org.jetbrains.kotlin.types.KotlinType
+
+/**
+ * This rule allows to set a list of forbidden annotations. This can be used to discourage the use
+ * of language annotations which do not require explicit import.
+ *
+ * <noncompliant>
+ * @@SuppressWarnings("unused")
+ * class SomeClass()
+ * </noncompliant>
```
cortinico, line 32: I would keep both a `compliant` and a `noncompliant` block here.
Maybe show the `@Deprecated` one or others

cortinico (approved):
LGTM 👍 Mostly minor nits

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotationSpec.kt
```diff
@@ -36,8 +36,8 @@ class ForbiddenAnnotationSpec(val env: KotlinCoreEnvironment) {
     @Test
     fun `should report annotations from java lang annotation package by default`() {
         val code = """
-        import java.lang.annotation.Retention 
-        import java.lang.annotation.Documented
```
TWiStErRob: this will fail CI

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
```diff
@@ -0,0 +1,128 @@
...
+    override val issue = Issue(
+        javaClass.simpleName,
+        Severity.Style,
+        "Avoid using this annotation.",
+        Debt.FIVE_MINS
+    )
+
+    @Configuration(
+        "List of fully qualified annotation classes which are forbidden. " +
+            "For example, `kotlin.jvm.Transient`."
```
TWiStErRob, line 51: This doc looks outdated since there's now `reason` too. Probably no need for `for example` since there are many examples in default config.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
```diff
@@ -0,0 +1,128 @@
...
+                "Use `kotlin.annotation.MustBeDocumented` instead.",
+            "java.lang.annotation.Target" to "it is a java annotation. Use `kotlin.annotation.Target` instead.",
+            "java.lang.annotation.Retention" to "it is a java annotation. Use `kotlin.annotation.Retention` instead.",
+            "java.lang.annotation.Repeatable" to "it is a java annotation. Use `kotlin.annotation.Repeatable` instead.",
+            "java.lang.annotation.Inherited" to "Kotlin does not support @Inherited annotation, " +
+                "see https://youtrack.jetbrains.com/issue/KT-22265",
+        )
+    ) { list ->
+        list.associate { it.value to Forbidden(it.value, it.reason) }
+    }
```
TWiStErRob, line 67: Opinion: I think this can be simplified to `list.associateBy { it.value }`. It took me more to understand the extra private inner class than the win between `forbidden.value` and `forbidden.name`.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
```diff
@@ -0,0 +1,128 @@
...
+            check(annotation, it)
+        }
+    }
+
+    override fun visitExpression(expression: KtExpression) {
+        super.visitExpression(expression)
+
+        if (annotations.isEmpty()) {
+            return
+        }
```
TWiStErRob, line 85: Is this permature optimization? (same in above method) why would this rule be active if there are no configured annotations. @BraisGabin /@cortinico is there a way to easily check this during configuration phase? Can ConfigValidator be on a rule level? (btw. ConfigValidator has no [docs](https://detekt.dev/docs/next/introduction/extensions).)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    private fun KtExpression.expressionTypeOrNull(): KotlinType? {
+        return if (bindingContext != BindingContext.EMPTY) {
+            bindingContext[BindingContext.EXPRESSION_TYPE_INFO, this]?.type
+        } else {
+            null
+        }
+    }
```
TWiStErRob, line 127: `@RequiresTypeResolution` guarantees non-empty binding context. @BraisGabin this looks like another one for the rule-authors ruleset.

```suggestion
    private fun KtTypeReference.fqNameOrNull(): FqName? =
        bindingContext[BindingContext.TYPE, this]?.fqNameOrNull()

    private fun KtExpression.expressionTypeOrNull(): KotlinType? =
        bindingContext[BindingContext.EXPRESSION_TYPE_INFO, this]?.type
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotationSpec.kt
```diff
@@ -0,0 +1,251 @@
...
+        val code = """
+        @Deprecated("unused", ReplaceWith("bar"))
+        fun foo() = "1234"
+        """.trimIndent()
+        val findings = ForbiddenAnnotation(
+            TestConfig(mapOf(ANNOTATIONS to listOf("kotlin.ReplaceWith")))
+        ).compileAndLintWithContext(env, code)
+        assertThat(findings).hasSize(1)
+            .hasStartSourceLocation(1, 23)
+            .hasTextLocations("ReplaceWith")
```
TWiStErRob, line 209: I think this is a redundant suppression, since there's only one occurrence of ReplaceWith

```suggestion
            .hasTextLocations("ReplaceWith")
```

TWiStErRob: Same for most tests, hasTextLocations might be superior to other numeric validations (hasStartSourceLocation/hasSize) if the text is unique. Curious what others think.

TWiStErRob (approved):
Sorry, every time I look, I'm in a different mindset :)

Most comments are preference now, take them as you wish.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotationSpec.kt
ov7a: I thought so too, but it was really frustrating to get an exception that the text from the argument appears more than once in a snippet. I thought it kinda confusing to have half of the tests with text and half with numbers.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
TWiStErRob: leaving it open, curious what @BraisGabin thinks

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
ov7a: I think for the sake of readability/clear types `Map<String, Forbidden>` is a bit better than `Map<String, String>`. The best variant would be something like `Map<String, Reason>` or even `Map<FqName, Reason>`, but this requires more work (unrelated to this PR). But if others think that `Forbidden` is redundant too, I'm ok with deleting it.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotationSpec.kt
TWiStErRob: > half of the tests with text and half with numbers.

IMO that just highlights even more the type of the test.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
TWiStErRob: It's not going to be `<String, String>` you still get the same object, just with a different name:

```patch
Index: detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt	(revision 7ce3ebd0ba0cc58984ac3130847d10d23aa27254)
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt	(date 1669413179980)
@@ -8,6 +8,7 @@
 import io.gitlab.arturbosch.detekt.api.Location
+import io.gitlab.arturbosch.detekt.api.ValueWithReason
 import io.gitlab.arturbosch.detekt.api.internal.Configuration
 import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
@@ -50,7 +51,7 @@
         "List of fully qualified annotation classes which are forbidden. " +
             "For example, `kotlin.jvm.Transient`."
     )
-    private val annotations: Map<String, Forbidden> by config(
+    private val annotations: Map<String, ValueWithReason> by config(
         valuesWithReason(
             "java.lang.SuppressWarnings" to "it is a java annotation. Use `Suppress` instead.",
             "java.lang.Deprecated" to "it is a java annotation. Use `kotlin.Deprecated` instead.",
@@ -63,7 +64,7 @@
                 "see https://youtrack.jetbrains.com/issue/KT-22265",
         )
     ) { list ->
-        list.associate { it.value to Forbidden(it.value, it.reason) }
+        list.associateBy { it.value }
     }
 
     override fun visitAnnotationEntry(annotation: KtAnnotationEntry) {
@@ -93,9 +94,9 @@
 
         if (forbidden != null) {
             val message = if (forbidden.reason != null) {
-                "The annotation `${forbidden.name}` has been forbidden: ${forbidden.reason}"
+                "The annotation `${forbidden.value}` has been forbidden: ${forbidden.reason}"
             } else {
-                "The annotation `${forbidden.name}` has been forbidden in the detekt config."
+                "The annotation `${forbidden.value}` has been forbidden in the detekt config."
             }
             val location = Location.from(element).let { location ->
                 location.copy(
@@ -108,8 +109,6 @@
         }
     }
 
-    private data class Forbidden(val name: String, val reason: String?)
-
     private fun KtTypeReference.fqNameOrNull(): FqName? {
         return if (bindingContext != BindingContext.EMPTY) {
             bindingContext[BindingContext.TYPE, this]?.fqNameOrNull()

```

TWiStErRob: I sometimes do [`private typealias FullyQualifiedAnnotationName = String`](https://github.com/TWiStErRob/net.twisterrob.gradle/blob/5f3a9f7b71d54627dea85ec7f95b7e38d3d63c0c/quality/src/main/kotlin/net/twisterrob/gradle/quality/report/TableGenerator.kt#L6-L9) in my personal code, but that probably goes a bit too far in this project :)

ov7a: Oh, missed that, thanks.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotationSpec.kt
TWiStErRob: Nice one! It actually answers one thing I wasn't sure of, in form of a test. The highlight includes the @ symbol :)

Note: because of the @ symbol, it should be possible to use text matchers everywhere, no?

ov7a: Yes, indeed. Only one place remains where there are actually two of them (annotations for expressions), but I guess it's ok to leave only one expression for test. 

Thank you, sorry for me being so careless:(

TWiStErRob: No worries, I'm learning with you :)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
BraisGabin: I agree, there's not need for this `if`. I don't care that much about it either... but if we have this `if`s we should also have a test to check what happen with `annotations` is empty. So I would say that it's better to remove it. But if you, @ov7a, prefer to keep it just add a test for it.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
BraisGabin: Yes, I think that any check of `bindingContext != BindingContext.EMPTY` or `bindingContext == BindingContext.EMPTY` should be flagged as a code smell. That should be an easy rule to implement.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenAnnotation.kt
ov7a: I just removed it. I agree with @TWiStErRob that it would be nice to have some example of config validation.

PR #5534, opened 2022-11-15
Fix KDocReferencesNonPublicProperty false positive

dzirbel:
KDocReferencesNonPublicProperty has a common false-positive where there is a name overlap with a public and non-public property (e.g. a local variable). This should not be reported as an issue since the kdoc is referencing the public field.

To fix this, we need to add awareness of the public properties and filter out name overlaps when checking if non-public properties are being referenced in the kdoc. This also includes properties declared as constructor parameters.

Also add a test that constructor parameters are allowed to be referenced in the class kdoc even when they are private fields.

### DIFF
--- a/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicProperty.kt
+++ b/detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicProperty.kt
@@ -3,13 +3,16 @@ package io.gitlab.arturbosch.detekt.rules.documentation
 import io.gitlab.arturbosch.detekt.api.CodeSmell
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.DetektVisitor
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtDeclaration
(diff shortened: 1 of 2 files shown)


dzirbel:
An alternative I'm only realizing now is to try to just process the class-level properties and skip local variables entirely (if this is possible). That sounds cleaner but would slightly weaken the rule since it would not warn for references to local variables - but that seems like more of an edge case that the rule probably wasn't intended to cover in the first place.

BraisGabin (approved):
LGTM thanks!

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicProperty.kt
```diff
@@ -53,18 +56,56 @@ class KDocReferencesNonPublicProperty(config: Config = Config.empty) : Rule(conf
...
 
+    private val publicPropertiesByClass = mutableMapOf<KtClass, MutableSet<KtNamedDeclaration>>()
+    private val privatePropertiesByClass = mutableMapOf<KtClass, MutableSet<KtNamedDeclaration>>()
+
+    override fun visitClass(klass: KtClass) {
+        super.visitClass(klass)
+
+        val comment = klass.docComment?.text ?: return
+
+        klass.primaryConstructor?.accept(
```
chao2zhang, line 67: Do you know why we couldn't use `klass.primaryConstructor?.valueParameters`?

chao2zhang (approved):

detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicProperty.kt
chao2zhang: Agreed that this PR is improving by fixing known edge cases. This can be done as a follow-up (Today I learned that `[param]` syntax does not work if `param` is only a parameter not property.)

chao2zhang (approved):

PR #5554, opened 2022-11-22
Migrate to Gradle Nexus Publish Plugin

3flex:
The old Gradle Nexus Staging plugin is no longer updated and has been replaced by Gradle Nexus Publish Plugin.

https://blog.solidsoft.pl/2021/02/26/unified-gradle-projects-releasing-to-maven-central-in-2021-migration-guide/

### DIFF
--- a/build-logic/src/main/kotlin/releasing.gradle.kts
+++ b/build-logic/src/main/kotlin/releasing.gradle.kts
@@ -10,8 +10,6 @@ nexusPublishing {
 
     repositories {
         create("sonatype") {
-            stagingProfileId.set("1d8efc8232c5c")
-
             username.set(
                 findProperty("sonatypeUsername")
                     ?.toString()
--- a/build-logic/src/main/kotlin/releasing.gradle.kts
+++ b/build-logic/src/main/kotlin/releasing.gradle.kts
@@ -10,16 +10,8 @@ nexusPublishing {
 
     repositories {
         create("sonatype") {
-            username.set(
-                findProperty("sonatypeUsername")
-                    ?.toString()
-                    ?: System.getenv("MAVEN_CENTRAL_USER")
-            )
-            password.set(
-                findProperty("sonatypePassword")
-                    ?.toString()
-                    ?: System.getenv("MAVEN_CENTRAL_PW")
-            )
+            System.getenv("ORG_GRADLE_PROJECT_SONATYPE_USERNAME")?.let { username.set(it) }
+            System.getenv("ORG_GRADLE_PROJECT_SONATYPE_PASSWORD")?.let { password.set(it) }
         }
     }
 }
--- a/build-logic/src/main/kotlin/releasing.gradle.kts
+++ b/build-logic/src/main/kotlin/releasing.gradle.kts
@@ -2,18 +2,28 @@ import com.vdurmont.semver4j.Semver
 
 plugins {
     id("com.github.breadmoirai.github-release")
-    id("io.codearte.nexus-staging")
+    id("io.github.gradle-nexus.publish-plugin")
 }
 
-nexusStaging {
-    packageGroup = "io.gitlab.arturbosch"
-    stagingProfileId = "1d8efc8232c5c"
-    username = findProperty("sonatypeUsername")
-        ?.toString()
-        ?: System.getenv("MAVEN_CENTRAL_USER")
-    password = findProperty("sonatypePassword")
-        ?.toString()
-        ?: System.getenv("MAVEN_CENTRAL_PW")
+nexusPublishing {
+    packageGroup.set("io.gitlab.arturbosch")
+
+    repositories {
+        create("sonatype") {
+            stagingProfileId.set("1d8efc8232c5c")
+
+            username.set(
+                findProperty("sonatypeUsername")
+                    ?.toString()
+                    ?: System.getenv("MAVEN_CENTRAL_USER")
+            )
+            password.set(
+                findProperty("sonatypePassword")
+                    ?.toString()
+                    ?: System.getenv("MAVEN_CENTRAL_PW")
+            )
+        }
+    }
 }
 
 project.afterEvaluate {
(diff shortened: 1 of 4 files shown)


build-logic/src/main/kotlin/releasing.gradle.kts
```diff
@@ -2,18 +2,28 @@ import com.vdurmont.semver4j.Semver
...
+
+    repositories {
+        create("sonatype") {
+            stagingProfileId.set("1d8efc8232c5c")
+
+            username.set(
+                findProperty("sonatypeUsername")
+                    ?.toString()
+                    ?: System.getenv("MAVEN_CENTRAL_USER")
+            )
```
TWiStErRob, line 19: `sonatypeUserName` is the default value, how about `System.getenv("MAVEN_CENTRAL_USER")?.let { username.set(it) }`

build-logic/src/main/kotlin/releasing.gradle.kts
```diff
@@ -2,18 +2,28 @@ import com.vdurmont.semver4j.Semver
 
 plugins {
     id("com.github.breadmoirai.github-release")
-    id("io.codearte.nexus-staging")
+    id("io.github.gradle-nexus.publish-plugin")
```
TWiStErRob, line 5: release.sh line 9 needs to be changed - https://blog.solidsoft.pl/2021/02/26/unified-gradle-projects-releasing-to-maven-central-in-2021-migration-guide/#tasks---simple-case

TWiStErRob: and probably some others too: https://github.com/detekt/detekt/search?q=publish&type=code

TWiStErRob: Does anything need to change (or redundant) in https://github.com/detekt/detekt/blob/ecd28cf1a3ba1df906f34dfc4d988b5b93ab1763/build-logic/src/main/kotlin/packaging.gradle.kts#L11-L29 ?

build-logic/src/main/kotlin/releasing.gradle.kts
cortinico: +1 to this

build-logic/src/main/kotlin/releasing.gradle.kts
```diff
@@ -2,18 +2,28 @@ import com.vdurmont.semver4j.Semver
...
-        ?: System.getenv("MAVEN_CENTRAL_USER")
-    password = findProperty("sonatypePassword")
-        ?.toString()
-        ?: System.getenv("MAVEN_CENTRAL_PW")
+nexusPublishing {
+    packageGroup.set("io.gitlab.arturbosch")
+
+    repositories {
+        create("sonatype") {
+            stagingProfileId.set("1d8efc8232c5c")
```
cortinico, line 13: I believe this is unnecessary and the new plugin essentially queries the profile ID given username/password

cortinico (approved):
Thanks for doing this @3flex 🙏 It will also prevent several fragmented staging repositories I noticed in the past.

build-logic/src/main/kotlin/releasing.gradle.kts
3flex: There's a note in the docs saying it will be faster if it's explicitly specified, but if it's literally just an API call then it makes sense to keep the config cleaner.

TWiStErRob: It might be one per module. `InitializeNexusStagingRepository` task uses the value, and retrieves it if not set. However in `NexusPublishPlugin.configurePublishingForAllProjects` they register that task on `rootProject` (something we should do too for some things). So we're back to one :)

That said they even provided a task to get it and print it to console:
```kotlin
val retrieveStagingProfileTask = rootProject.tasks.register<RetrieveStagingProfile>("retrieve${capitalizedName}StagingProfile", rootProject.objects, extension, repository)
```
I think they mean it when they say, "set it" :) but the performance cost is likely negligible even for once a month releases.

build-logic/src/main/kotlin/releasing.gradle.kts
3flex: Are the MAVEN_CENTRAL_* vars actually used when publishing?

Just wondering if it can just be set to just a single value.

chao2zhang: > Are the `MAVEN_CENTRAL_*` vars actually used when publishing?

There are action secrets for our CI job https://github.com/detekt/detekt/settings/secrets/actions
So if we are using the same `MAVEN_CENTRAL_*` environment variable when doing the production release, we should be consolidating them into a single key.

cortinico: > Are the MAVEN_CENTRAL_* vars actually used when publishing?

IMHO we can clean them up and stick to those two properties/env-variable:

```
ORG_GRADLE_PROJECT_SONATYPE_USERNAME
ORG_GRADLE_PROJECT_SONATYPE_PASSWORD
```

Those are set and used in the `packaging.gradle.kts` also when publishing snapshots

TWiStErRob: The secrets are not automatic env vars. You need to expose them with `env: X: ${{ secrets.Y }}` anyway, so their names don't have to match. Maybe even better if it doesn't?

cortinico: > You need to expose them with env: X: ${{ secrets.Y }} anyway, so their names don't have to match. Maybe even better if it doesn't?

Yup they're exposed here https://github.com/detekt/detekt/blob/216b44d5c4d8cf497d973f0849ca3fa876cdad63/.github/workflows/deploy-snapshot.yaml#L36-L39

I would keep them the same to avoid unnecessary confusion

build-logic/src/main/kotlin/releasing.gradle.kts
3flex: I updated the release script (thanks for the pointer!), but I didn't notice anything else that specifically has to be updated.

Not sure if the packaging script can be updated or simplified, but suggest that be done separately.

cortinico (approved):

PR #5562, opened 2022-11-23
Random tiny improvements

TWiStErRob:
In the last few months, I've been noticing things here and there and collating this branch. It's really simple changes, I recommend looking at the individual commits for details.

### DIFF
--- a/detekt-core/build.gradle.kts
+++ b/detekt-core/build.gradle.kts
@@ -23,4 +23,5 @@ dependencies {
     testImplementation(libs.mockk)
     testImplementation(libs.reflections)
     testImplementation(libs.assertj)
+    testRuntimeOnly(libs.slf4j.simple)
 }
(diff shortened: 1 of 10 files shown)


cortinico:
Changes looks generally good to me. I would extract https://github.com/detekt/detekt/pull/5562/commits/670435d071bb127d509a46f1c194f555dfd8582c to a separate PR as it will show up in the changelog as well

detekt-core/build.gradle.kts
```diff
@@ -23,4 +23,5 @@ dependencies {
     testImplementation(libs.mockk)
     testImplementation(libs.reflections)
     testImplementation(libs.assertj)
+    testRuntimeOnly(libs.slf4j.simple)
```
TWiStErRob, line 26: > Changes looks generally good to me. I would extract https://github.com/detekt/detekt/commit/670435d071bb127d509a46f1c194f555dfd8582c to a separate PR as it will show up in the changelog as well

I don't understand, can you please elaborate more. This is an internal test-only change, why would it need to be in changelog?

cortinico: > Yup you're right. I haven't noticed it's test only so no need to surface this externally 👍 My bad

cortinico (approved):

chao2zhang (approved):

PR #5573, opened 2022-11-27
Enable configuration cache by default

3flex:
Third time's the charm (previous attempts: #5095, #4576)

I've tested these commands successfully.
```
./gradlew build -x dokkaHtml
./gradlew build -x dokkaHtml
./gradlew clean
./gradlew clean
```


cortinico (approved):

cortinico:
> I've tested these commands successfully.

Could you try `publishToMavenLocal` as I recall this was one of the offending scenario?

3flex:
No need: running that task will disable CC because it's currently tagged with `notCompatibleWithConfigurationCache`:

https://github.com/detekt/detekt/blob/01db248a8c869fb3a93ccdd6d4af70483383907a/build-logic/src/main/kotlin/module.gradle.kts#L24-26

Issue #5247, opened 2022-08-25
detekt-cli:  '1.21.0' / 1.22.0'  fails on KtLintMultiRule / Indentation when using `--build-upon-default-config`


rowi1de:
The intitial description wasn't correct, please see comment: https://github.com/detekt/detekt/issues/5247#issuecomment-1326212020

---

Detekt installed via brew
```shell
brew install detekt
```
It failed on an M1+intel Macs
```
The original exception message was: java.lang.ExceptionInInitializerError
Running detekt '1.21.0' on Java '17.0.4+8-LTS' on OS 'Mac OS X'
If the exception message does not help, please feel free to create an issue on our GitHub page.
```

It seems to happen when using a wrong value for `--classpath` which fails to resolve dependencies
```shell
detekt --jvm-target 17 --classpath .
```

````
The BindingContext was created with 2668 issues. Run detekt with --debug to see the error messages.
java.util.concurrent.CompletionException: java.lang.IllegalStateException: Analyzing....kt led to an exception. 
Location: java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
The original exception message was: java.lang.ExceptionInInitializerError
Running detekt '1.21.0' on Java '17.0.4+8-LTS' on OS 'Mac OS X'
If the exception message does not help, please feel free to create an issue on our GitHub page.
        at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
        at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320)
        at java.base/java.util.concurrent.CompletableFuture.uniExceptionally(CompletableFuture.java:994)
        at java.base/java.util.concurrent.CompletableFuture$UniExceptionally.tryFire(CompletableFuture.java:974)
        at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
        at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1773)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.lang.IllegalStateException: Analyzing ... kt led to an exception. 
[...]

PR #5579, opened 2022-11-28
Allow access to nonpublic members of public types in java.base/java.lang package

### LINKED ISSUE #5576: detekt CLI fails on Java 17+ in some cases when using detekt-formatting plugin with auto correct
## Expected Behavior
I can run detekt CLI without applying any workarounds.

## Observed Behavior
It can fail in some circumstances, see https://github.com/detekt/detekt/issues/5247#issuecomment-1328168323

## Steps to Reproduce
TBC - opening tracking issue to link a brew issue to.

## Context
Users using CLI and Java 17 without `--add-opens java.base/java.lang=ALL-UNNAMED` applied will see detekt failing in some circumstances.

## Your Environment

* Version of detekt used:

3flex:
Fixes #5576 (also see #5247)


BraisGabin (approved):
Is this really a workaround? Isn't this the proper fix? We should work with JDK17.

cortinico (approved):

3flex:
It's a workaround for an upstream issue, I believe it's the IntelliJ components that Kotlin packages up that's causing this. Eventually I expect this can be removed.

I've left this as draft as this fix doesn't work if calling the auto-generated kotlin-cli script, because that uses the -classpath flag when calling java, and that apparently ignores the flag  in the manifest. It works for the shadow version though as that script calls java with -jar flag.

So this doesn't cover all cases yet.

3flex:
OK, I need feedback on options here, because it's a little messy to get a fully working setup for all cases due to two constraints.
1. The CLI flag `--add-opens=java.base/java.lang=ALL-UNNAMED` can only be used on Java 9+. Java 8 will not run if passed that flag as it's unrecognised (it was only added in Java 9).
2. `Add-Opens` in the manifest is only read if the CLI jar is passed to `java -jar`, not if using `java -classpath`.

We generate two different CLI distributions in the build. One is the shadow version (which uses `java -classpath <classpath> io.gitlab.arturbosch.detekt.cli.Main`), one is the non-Shadow version (which uses `java -jar`). The one currently published to GitHub release is non-Shadow.

Options:
1. Publish the Shadow version for the CLI and stop publishing the current non-Shadow version. The Shadow version works with `Add-Opens` in the manifest file - it's read & applied on Java 9+ and ignored on Java 8.
2. Pass `--add-opens=java.base/java.lang=ALL-UNNAMED` to the non-Shadow version. The CLI release script will stop working on Java 8.
3. Generate a custom script for the non-Shadow version, which uses `java -jar`, but my understanding is that the CLI jar's manifest would then have to include a `Class-Path` declaration which isn't autogenerated, so we'd have to generate as part of the build. I assume Gradle would ignore this when the CLI jar's dependencies are resolved (when using the detekt Gradle plugin) and it would just check the Gradle metadata or the POM, but this would have to be tested.

Option 1 looks best to me. Ping @cortinico since the release setup would have to change a bit (which can be done in a different PR).

BraisGabin:
Option 1 seems the best one.

Degradate the experience for users with jdk8 doesn't feel like a bad option either.

cortinico:
I'm in for publishing a jar that works only on 9+ (we can specify it in the release notes).

> Publish the Shadow version for the CLI and stop publishing the current non-Shadow version

Are you sure that we're publishing the non-shadow for the CLI?
I was under the assumption we were publishign the shadow already:
https://github.com/detekt/detekt/blob/main/detekt-cli/build.gradle.kts#L47-L49

3flex:
> I was under the assumption we were publishign the shadow already

For the JAR, yes, it's the shadow version. But the ZIP, which has a script to execute on the CLI, it's the non-Shadow version.

Open [detekt-cli-1.22.0.zip](https://github.com/detekt/detekt/releases/download/v1.22.0/detekt-cli-1.22.0.zip), look at the scripts in `bin` directory. Both scripts pass `-classpath`. The manifest is ignored.

If, instead, we publish the Shadow version, it will use `-jar` and the manifest will be read. The workaround will be applied.

Basically we need to change this:
https://github.com/detekt/detekt/blob/f081ebe6f99d9069a3a7aa43e354c4aff4a7bc59/build-logic/src/main/kotlin/releasing.gradle.kts#L40

to instead publish `distributions/detekt-cli-shadow-${project.version}.zip`

cortinico (approved):

Issue #417, opened 2017-09-20
Rule: FeatureEnvy


schalkms:
As mentioned in #382 the `FeatureEnvy` needs type resolution.

Mauin:
What is the state of this rule? It's currently in the root package of the `detekt-rules` module but not added to any ruleset. Should we remove it for now and bring it back once it is fully ready?

schalkms:
@Mauin 
Rhis rule needs type resolution. It is not fully functional atm.
Please read #382 for more information.
I just added an issue in order to not forget about this rule.

Mauin:
@schalkms In that case I'd say the current implementation of the rule is dead code. Should we remove it for now? We can revert the commit that removes the rule once we feel that we can fully implement the rule. But for now it's not giving us any benefit having the class around unfortunately.

schalkms:
I agree. I would also remove prior v1.

arturbosch:
I'm ok with that. We have VCS :)

BraisGabin:
Does we still want the `FeatureEnvy` rule? @schalkms could you describe what this rule does so anyone could pick this issue?

schalkms:
I think you can keep this issue open. Basically, the linked PR #706 shows the (removed) prototype rule implementation by Artur. However, the `FeatureEnvy` rule needs type and symbol solving in order to work correctly. Because of that, the rule got removed.

**Feature Envy** is a code smell in which one class uses fields or methods of other classes extensively in order to perform some sort of functionality. For instance, a method makes extensive use of methods and fields provided by a class other than the one it lives in. Methods, which extensively use another class, may belong to another class. Therefore, it is considered a code smell, because this arguably breaks encapsulation. To remove this code smell, programmers should consider moving this method to the class it is so envious of.

Reference:
https://wiki.c2.com/?FeatureEnvySmell

PR #5581, opened 2022-11-28
Feature envy

Tommyten:
This pull request adds a rule to detect functions with feature envy, as discussed in #417.

Since Detekt does not provide the possibility of reference searches, it is limited in several ways:

- Feature envy can't be detected in extension functions
- Feature envy can't be detected in top-level functions (this is more of a technical question of how feature envy would work in top-level functions)
- Feature envy cant be detected in unnamed functions (e.g., functions that are saved to a variable)

There also are some issues that kind of block this PR.
The biggest one is that `decimal` configuration values are not allowed currently.
Because of this, the `localityOfDataAccessThreshold` configuration property is currently hard coded and not in the generated default config.
Is this a design choice, or is it just that somebody has to make these changes? If so, I'm willing to try :)

Another issue I've noticed is that the `generateDocumentation` task currently throws an error if I try to set a custom Debt for the `Issue`. It throws the following error:
```
Exception in thread "main" io.gitlab.arturbosch.detekt.generator.collection.exception.InvalidIssueDeclaration: Invalid issue declaration attribute Debt.
```
I don't think every Issue can fit in the 20/10/5 minute options we currently have. Refactoring Feature Envy, depending on complexity, can take quite a bit longer.

Another issue is the fact, that I don't know which of the three thresholds I should report using the `ThresholdedCodeSmell`.
Are there better ways to report this?

Looking forward to your suggestions :)


detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/FeatureEnvy.kt
```diff
@@ -0,0 +1,177 @@
...
+    private val atfdThreshold: Int by config(defaultValue = 2)
+
+    @Configuration(
+        "The minimum number of foreign data providers which " +
+            "must be used so that no Feature Envy is present."
+    )
+    private val fdpThreshold: Int by config(defaultValue = 2)
+
+    override fun visitNamedFunction(function: KtNamedFunction) {
+        if (bindingContext == BindingContext.EMPTY) {
```
BraisGabin, line 64: This `if` is not needed anymore by detekt. The `RequiresTypeResolution` annotation does the trick.

BraisGabin (commented):
First of all, thnks to grab this old request :)

The tests seems too few for all the logic that this implementation had. Aren't they? I see a lot of `filter`s and `if` that I'm not 100% sure if they are exercised. I'm not looking for a 100% coverage but 3 cases feels like too few.

BraisGabin:
> The biggest one is that `decimal` configuration values are not allowed currently. Because of this, the `localityOfDataAccessThreshold` configuration property is currently hard coded and not in the generated default config. Is this a design choice, or is it just that somebody has to make these changes? If so, I'm willing to try :)

It is not a design choice. It's just that we didn't need a float until now. So please feel free to add support for them.

> Another issue I've noticed is that the `generateDocumentation` task currently throws an error if I try to set a custom Debt for the `Issue`. It throws the following error:
> 
> ```
> Exception in thread "main" io.gitlab.arturbosch.detekt.generator.collection.exception.InvalidIssueDeclaration: Invalid issue declaration attribute Debt.
> ```
> 
> I don't think every Issue can fit in the 20/10/5 minute options we currently have. Refactoring Feature Envy, depending on complexity, can take quite a bit longer.

Interesting... I would mark feature envy as `20` minutes anyway. We have them as "buckets". But if you want to have something more custom you should add support for it in `RuleVisitor`. But again, I would set it to `20`. And once this PR is merged you could work on that to improve it (other wise this PR is going to be big)

> Another issue is the fact, that I don't know which of the three thresholds I should report using the `ThresholdedCodeSmell`. Are there better ways to report this?

Probably `localityOfDataAccessThreshold`. Or don't use `ThresholdedCodeSmell` and create your custom message giving enought information to make the issue actionable.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/FeatureEnvy.kt
Tommyten: Indeed, the tests are too few currently. I am still working on building more tests now. I just wanted to get this PR out earlier rather than later to get an idea if any of the mentioned issues are breaking this rule (e.g., float config not being allowed or similar) so that the effort is not pointless.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/FeatureEnvy.kt
```diff
@@ -0,0 +1,166 @@
...
+import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
+import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
+import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
+import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
+import org.jetbrains.kotlin.resolve.calls.util.getResolvedCall
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+import org.jetbrains.kotlin.resolve.descriptorUtil.getOwnerForEffectiveDispatchReceiverParameter
+
+/**
+ * This rule reports methods with Feature Envy.
```
cortinico, line 31: Can you link to https://refactoring.guru/smells/feature-envy or any other source which explains what Feature Envy is?

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -120,6 +120,11 @@ complexity:
       - 'run'
       - 'use'
       - 'with'
+  FeatureEnvy:
+    active: false
+    localityOfDataAccessThreshold: 0.33
+    atfdThreshold: 2
+    fdpThreshold: 2
```
cortinico, line 127: Can we be a bit more explicit with naming here? This forces the user to read the documentation to understadn those flags

cortinico (commented):
I'm not familiar with this inspection, so I'll leave the review to others 👍  Thanks for bringing it up.

My 2 c: the rule logic looks a bit over complicated, and I'm a bit afraid for us having to maintain this in the long run.

PR #5593, opened 2022-12-01
Fix dependencies to generateDocumentation task

jprinet:
The :detekt-rules-libraries:processResources and :detekt-rules-authors:processResources tasks are having implicit dependency to the :detekt-generator:generateDocumentation task

This PR adds the missing declarations

### DIFF
--- a/detekt-generator/build.gradle.kts
+++ b/detekt-generator/build.gradle.kts
@@ -86,6 +86,14 @@ val generateDocumentation by tasks.registering(JavaExec::class) {
     )
 }
 
+rootProject.project("detekt-rules-libraries").tasks.withType<ProcessResources>().configureEach {
+    dependsOn(generateDocumentation)
+}
+
+rootProject.project("detekt-rules-ruleauthors").tasks.withType<ProcessResources>().configureEach {
+    dependsOn(generateDocumentation)
+}
+
 val verifyGeneratorOutput by tasks.registering(Exec::class) {
     dependsOn(generateDocumentation)
     description = "Verifies that generated config files are up-to-date"


cortinico (approved):

schalkms (approved):

3flex:
Is there a better way of doing this? Projects should not access the configuration of other projects. I know Gradle is working on ways to enforce this in builds, so I'd prefer a more forward compatible method.

See here for more: https://gradle.github.io/configuration-cache/

The project isolation feature itself is incubating but we should try to align to those principles if possible.

If it's too much work then let's merge but if you have a better alternative let's use that. Thanks!

jprinet:
Good point, this being said I'm not sure a project isolation compliant solution will be easy to implement given that the task is having I/O across several modules.
Let me check that and pause this one for the moment

cortinico:
> Let me check that and pause this one for the moment

Should this be a draft PR then?

jprinet:
I didn't have much time to spend on it but I'm concerned about the implementation of the [generateDocumentation task](https://github.com/detekt/detekt/blob/edddb48eac9cd9138b3d8963d9437df3da824dbb/detekt-generator/build.gradle.kts#L43) which has some dependencies on a task which uses its [output](https://github.com/detekt/detekt/blob/edddb48eac9cd9138b3d8963d9437df3da824dbb/detekt-generator/build.gradle.kts#L61)

Having the output as part of the source code makes it difficult to solve.

This is not tied to project isolation but should be addressed first imo

jprinet:
Duplicate of https://github.com/detekt/detekt/pull/5773

Issue #5590, opened 2022-11-30
FunctionMaxLength false positive for overridden methods


TWiStErRob:
## Expected Behavior
`FunctionMaxLength` should give actionable results only.

## Observed Behavior
https://github.com/TWiStErRob/net.twisterrob.cinema/security/code-scanning/425
![image](https://user-images.githubusercontent.com/2906988/204913892-809256f9-3626-426d-9bbe-00d0abbdce7a.png)

## Context
The method in question comes from [org.neo4j.ogm.session.EntityInstantiator](https://github.com/neo4j/neo4j-ogm/blob/v4.0.0/core/src/main/java/org/neo4j/ogm/session/EntityInstantiator.java#L49) which I cannot rename, and I **have to** implement this method, otherwise it doesn't compile.

## Steps to Reproduce
 * Find a long method in a library which cannot be changed
   e.g. the above example in `org.neo4j:neo4j-ogm-core:4.0.0`
 * implement the interface or override the method
 * run `FunctionMaxLength` on it

## Your Environment

* Version of detekt used: 1.22.0
* Version of Gradle used (if applicable): N/A
* Gradle scan link (add `--scan` option when running the gradle task): N/A
* Operating System and version: N/A
* Link to your project (if it's a public repository): https://github.com/TWiStErRob/net.twisterrob.cinema/pull/215

TWiStErRob:
During implementation: review other naming rules with the same eye.

schalkms:
It's rather an improvement, since this rule never considered overridden methods to be ignored.

mmorozkov:
Hi I created a [PR](https://github.com/detekt/detekt/pull/5599) with this improvement.
I took into consideration a note by @TWiStErRob above and introduced `ignoreOverridden` for several other naming rules where it made sense

PR #5599, opened 2022-12-03
FunctionMaxLength false positive for overridden methods #5590

mmorozkov:
By analogy with other naming rules a new `ignoreOverridden` property with default value `true` was introduced for following rules:
 - `FunctionMaxLength`
 - `FunctionMinLength`
 - `VariableMaxLength`
 - `VariableMinLength`

### DIFF
--- a/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/VariableMaxLength.kt
+++ b/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/VariableMaxLength.kt
@@ -10,6 +10,7 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.config
 import io.gitlab.arturbosch.detekt.api.internal.Configuration
 import io.gitlab.arturbosch.detekt.rules.identifierName
+import io.gitlab.arturbosch.detekt.rules.isOverride
 import org.jetbrains.kotlin.psi.KtProperty
 
 /**
@@ -28,6 +29,10 @@ class VariableMaxLength(config: Config = Config.empty) : Rule(config) {
     private val maximumVariableNameLength: Int by config(DEFAULT_MAXIMUM_VARIABLE_NAME_LENGTH)
 
     override fun visitProperty(property: KtProperty) {
+        if (property.isOverride()) {
+            return
+        }
+
         if (property.identifierName().length > maximumVariableNameLength) {
             report(
                 CodeSmell(
--- a/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionMaxLength.kt
+++ b/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionMaxLength.kt
@@ -10,6 +10,7 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.config
 import io.gitlab.arturbosch.detekt.api.internal.Configuration
 import io.gitlab.arturbosch.detekt.rules.identifierName
+import io.gitlab.arturbosch.detekt.rules.isOverride
 import org.jetbrains.kotlin.psi.KtNamedFunction
 
 /**
@@ -31,6 +32,10 @@ class FunctionMaxLength(config: Config = Config.empty) : Rule(config) {
     private val maximumFunctionNameLength: Int by config(30)
 
     override fun visitNamedFunction(function: KtNamedFunction) {
+        if (function.isOverride()) {
+            return
+        }
+
         if (function.identifierName().length > maximumFunctionNameLength) {
             report(
                 CodeSmell(
--- a/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionMinLength.kt
+++ b/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionMinLength.kt
@@ -10,6 +10,7 @@ import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.config
 import io.gitlab.arturbosch.detekt.api.internal.Configuration
 import io.gitlab.arturbosch.detekt.rules.identifierName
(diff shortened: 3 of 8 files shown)


BraisGabin (approved):
Thanks! LGTM

I see that you copied the description of this parameter from other rules so we should merge this as it is. But probably we could improve the description on other pr. More than function with the override modifier I would call them override functions.

BraisGabin:
Fixes #5590

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/VariableMaxLength.kt
```diff
@@ -27,7 +28,14 @@ class VariableMaxLength(config: Config = Config.empty) : Rule(config) {
     @Configuration("maximum name length")
     private val maximumVariableNameLength: Int by config(DEFAULT_MAXIMUM_VARIABLE_NAME_LENGTH)
 
+    @Configuration("ignores member properties that have the override modifier")
+    private val ignoreOverridden: Boolean by config(true)
```
TWiStErRob, line 32: Is there any use case we can think of that warrants this to be configurable for any of the rules?

BraisGabin: Maybe if you don't want to extend long named functions. Anyway I would keep the configuration for consistency with the other rules.

To be honest these rules have from little to no sense to me. The length of a name doesn't measure how good a name is. But we have them so we should maintain them.

mmorozkov: This is an interesting question. I personally think that the vast majority of users will never need to configure `ignoreOverridden` parameter, and this probably applies to all the naming rules, not only this one. And to be honest, I couldn't trace the initial idea to have this ability as a configurable option, so thus far these options for the rules make little sense, if you asked me. So I decided to suggest changes that are consistent with the rest of the rules and their configuraitons.

Going back to initial question, one could probably come up with a couple of weird cases, when it's more important to see the failure, rather than silently ignore. Something wacky like this probably: one is using a 3rd party API as a dependency, but they're not simply consuming it, but also able to provide feedback and tailor that API to their needs. In that case, it might be make a little sense not to just ignore such methods by default, but provide an ability to fail in compile time to provide that aforementioned feedback and ask it to be changed

BraisGabin: I would vote to keep this PR as it is and open a new issue to deprecate all those configurations. Our configuration file is complex enough. If we can remove things we can call it a win. If someone really want to use detekt for something as what @mmorozkov (I really like the imagination to come up with an use case) they can implement its own rule. We can't support every single use case.

TWiStErRob: My original idea was that it's a leading question towards removing the flag. Yes, then this PR would change the behavior of the rule without a fallback, but then again as it stands that use case you came up with is pretty edgy. (I actually thought about something similar, as at my previous company we had internal SDKs that were consumed, so we had power over the lib's APIs too, but the lib had stronger Detekt setup than consumers :) )

> Maybe if you don't want to extend long named functions.

The name wouldn't really decide if I need to override a function, it's because of an implementation need. Just cause the original author of the base class/interface made a choice, doesn't mean that I wouldn't override the method if that helps with my implementation, or in the case I reported in the issue, let me compile my code.

The reason I would go ahead and remove it from here before merging, is that we would be knowingly adding more complications when in the common normal case, there's no way to action the finding. If it's in a different codebase it's unlikely to be fixed, and if it's the same codebase, the the finding would come up on the base class anyway.

Regarding removing these rules in the future:
> The length of a name doesn't measure how good a name is.

I agree with this, BUT, these rules are not showing you what a good name is, they're showing you what commonly constitutes as a bad name. Passing these arbitrary length checks doesn't mean that your name is good, it just means it's not horribly bad. I see them as flexible constraints that pull the naming towards the best name possible in a scenario. If you think about having a variable named `cn` or `thisVariableHoldsTheGivenAndFamilyNameOfThePersonWhoBuysFromOurCompany`, we can immediately agree they're not good names, and these rules helps automate this finding rather than having to rely on people in code review or feedback after someone uses the API. (faster feedback loop). Hopefully after having flagged those above two names, one would have a short think and rename it to `customerName`. So yes, they're arbitrary and not telling you what good is, but give a very good simple baseline guideline to stay within the recommended range.

TWiStErRob:
@BraisGabin putting "fixes" in a comment, doesn't trigger GitHub to pick it up, it has to be in OP, or manually linked in the sidebar.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/VariableMaxLength.kt
chao2zhang: @TWiStErRob Are you suggesting that we do not provide such configuration option while merging this exclusion check?

TWiStErRob: Yes. Overridden functions are usually exempt from many checks (in IDEA inspections too), because you cannot change the signature (including names).

BraisGabin: I agree about removing the configuration all together but keeping the behavior. And we will deprecate the flag in the another PR (#5600)

TWiStErRob: Note that I was only advocating for these newly added ones to not be added, so #5600 doesn't grow.

BraisGabin: Just remove the configuration. If they are override these rules should not apply.

BraisGabin (approved):

chao2zhang (approved):

PR #5610, opened 2022-12-07
Run specific rule with 'detekt' task against external codebase

### LINKED ISSUE #4408: How to test new rules in real world projects easily
## Expected Behavior
It should be easy to test a new rule agains some Open Source sample projects to spot issues in the rule implementation.

## Current Behavior
This is really difficult we have this script: https://github.com/detekt/detekt/blob/main/scripts/get_analysis_projects.kts but it is just usefull if you don't use Type Solving.

## Context
#4033 and https://github.com/detekt/detekt/pull/4394#pullrequestreview-840070075

## Possible solution
We should list projects that already use detekt and use those for this tests using the `--including-build` feature from gradle.

VitalyVPinchuk:
[WIP]
Test new rules against an external codebase (that already uses Detekt) with Gradle's --include-build cli option to substitute online Detekt dependency with local one.

Closes #4408

Usage:
From Open Source project under test run
`./gradlew --include-build /path/to/detekt detekt --run-rule ruleset:Rule`
Ex. from mozilla fenix folder run 
`./gradlew --include-build ../../detekt detekt --run-rule comments:UndocumentedPublicFunction`

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -22,6 +22,7 @@ import io.gitlab.arturbosch.detekt.invoke.JdkHomeArgument
 import io.gitlab.arturbosch.detekt.invoke.JvmTargetArgument
 import io.gitlab.arturbosch.detekt.invoke.LanguageVersionArgument
 import io.gitlab.arturbosch.detekt.invoke.ParallelArgument
+import io.gitlab.arturbosch.detekt.invoke.RunRuleArgument
 import io.gitlab.arturbosch.detekt.invoke.isDryRunEnabled
 import org.gradle.api.Action
 import org.gradle.api.file.ConfigurableFileCollection
@@ -61,6 +62,14 @@ import javax.inject.Inject
 abstract class Detekt @Inject constructor(
     private val objects: ObjectFactory
 ) : SourceTask(), VerificationTask {
+
+    @get:Optional
+    @get:Input
+    @set:Option(
+        option = "run-rule",
+        description = "Rule to run"
+    )
+    var ruleToRun: String = ""
 
     @get:Classpath
     abstract val detektClasspath: ConfigurableFileCollection
@@ -228,7 +237,8 @@ abstract class Detekt @Inject constructor(
             AllRulesArgument(allRulesProp.getOrElse(false)),
             AutoCorrectArgument(autoCorrectProp.getOrElse(false)),
             BasePathArgument(basePathProp.orNull),
-            DisableDefaultRuleSetArgument(disableDefaultRuleSetsProp.getOrElse(false))
+            DisableDefaultRuleSetArgument(disableDefaultRuleSetsProp.getOrElse(false)),
+            RunRuleArgument(ruleToRun + ""),
         ).plus(convertCustomReportsToArguments()).flatMap(CliArgument::toArgument)
 
     @InputFiles
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -238,7 +238,7 @@ abstract class Detekt @Inject constructor(
             AutoCorrectArgument(autoCorrectProp.getOrElse(false)),
             BasePathArgument(basePathProp.orNull),
             DisableDefaultRuleSetArgument(disableDefaultRuleSetsProp.getOrElse(false)),
-            RunRuleArgument(ruleToRun + ""),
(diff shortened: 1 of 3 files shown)


cortinico:
> Closes #4408

The idea behind this issue was more to have some projects on CI that we test against (e.g. we clone the Mozilla app and we run Detekt there just to get a sense of how it performs). I don't think we want to touch the Detekt CLI for this

BraisGabin:
I was thinking more about something like this. Run on our ci all the rules against other projects could create a lot of noise.

I think that the ability to test the rule that you are working on against different projects is the thing that give us more value.

I think that this with a bit of documentation to explain how to use this on development would be a good addition. That and remove the old script. Maybe on that documentation we could add some open source project examples to test against.

cortinico:
> I think that the ability to test the rule that you are working on against different projects is the thing that give us more value.

Thanks for clarifying.

I would not edit the CLI API just to allow this though. This should be done via an internal call (to be figured out how).

My idea behind this was to have some sort of test coverage of external projects.
Say we onboard Firefox, we expect they do have a baseline. We run this job every night and get a sense how much we're disruptive in terms of new issues reported. Moreover, this could help us spot early one problems with the CLI/Gradle Plugin.

Still, having the possibility to run just one rule woudl also be beneficial as part of the development flow

VitalyVPinchuk:
Guys, is there anything I could do to improve this PR?

cortinico:
> Guys, is there anything I could do to improve this PR?

Nope at this moment, as we're still discussing and what is the way forward 👍

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
```diff
@@ -40,6 +41,15 @@ internal data class InputArgument(val fileCollection: FileCollection) : CliArgum
     override fun toArgument() = listOf(INPUT_PARAMETER, fileCollection.joinToString(",") { it.absolutePath })
 }
 
+internal data class RunRuleArgument(val rule: String) : CliArgument() {
+    override fun toArgument() = if (rule.isNotEmpty()) {
+        println(">>> Run the only rule <<<")
```
TWiStErRob, line 46: Don't forget to remove this before merge.

BraisGabin:
> I would not edit the CLI API just to allow this though. This should be done via an internal call (to be figured out how).

🤔 This PR doesn't change anything of our cli. This only touches the Detekt Gradle Plugin. The current cli has the option to only run a rule but few users use the cli. And for development I think that no one uses it.

This PR only enables our plugin to use that cli argument that already exists. I think that the `diff` here is really small and the impact on the rule development could be big.

This plus an update on our [`CONTRIBUTING.md`](https://github.com/detekt/detekt/blob/main/.github/CONTRIBUTING.md) would be a good addition.

---

> My idea behind this was to have some sort of test coverage of external projects.
> Say we onboard Firefox, we expect they do have a baseline. We run this job every night and get a sense how much we're disruptive in terms of new issues reported. Moreover, this could help us spot early one problems with the CLI/Gradle Plugin.

I agree but, to me, that's another task.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -228,7 +237,8 @@ abstract class Detekt @Inject constructor(
             AllRulesArgument(allRulesProp.getOrElse(false)),
             AutoCorrectArgument(autoCorrectProp.getOrElse(false)),
             BasePathArgument(basePathProp.orNull),
-            DisableDefaultRuleSetArgument(disableDefaultRuleSetsProp.getOrElse(false))
+            DisableDefaultRuleSetArgument(disableDefaultRuleSetsProp.getOrElse(false)),
+            RunRuleArgument(ruleToRun + ""),
```
BraisGabin, line 241:
```suggestion
            RunRuleArgument(ruleToRun),
```

`ruleToRun` is a `String` already. You don't need ` + ""`, right?

VitalyVPinchuk: Indeed.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -62,6 +63,14 @@ abstract class Detekt @Inject constructor(
     private val objects: ObjectFactory
 ) : SourceTask(), VerificationTask {
 
+    @get:Optional
+    @get:Input
+    @set:Option(
+        option = "run-rule",
+        description = "Rule to run"
+    )
+    var ruleToRun: String = ""
```
TWiStErRob, line 72: Could this be made a `SetProperty<String>` so that --run-rule can accept multiple?

PR #5625, opened 2022-12-16
Update ktlint to v0.48.0

3flex:
There are three new rules that I haven't added wrappers for, since we seem to have the same ones in detekt already:
* Class naming
* Function naming
* Property naming

Also had to disable `NoSemicolons` rule on detekt's codebase because of https://github.com/pinterest/ktlint/issues/1733 but I'm sure that will be fixed soon. When this is merged I'll open a tracking issue for that.

### DIFF
--- a/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
+++ b/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
@@ -1,9 +1,8 @@
 package io.gitlab.arturbosch.detekt.formatting
 
-import com.pinterest.ktlint.core.KtLint
 import com.pinterest.ktlint.core.Rule.VisitorModifier.RunAsLateAsPossible
-import com.pinterest.ktlint.core.api.DefaultEditorConfigProperties.codeStyleSetProperty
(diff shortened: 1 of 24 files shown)


detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
```diff
@@ -78,21 +77,20 @@ abstract class FormattingRule(config: Config) : Rule(config) {
         this.root = root
         positionByOffset = KtLintLineColCalculator
             .calculateLineColByOffset(KtLintLineColCalculator.normalizeText(root.text))
-        root.node.putUserData(KtLint.FILE_PATH_USER_DATA_KEY, root.name)
```
3flex, line 81: This value isn't read by any ktlint rules that we have wrappers for, so removing this line has no effect.

chao2zhang (approved):

chao2zhang:
> There are three new rules that I haven't added wrappers for, since we seem to have the same ones in detekt already:
> 
> Class naming
> Function naming
> Property naming

I believe Detekt implementation for these rules is more flexible (Since ktlint is mostly on/off with limited configurability), hence I believe we should still add a wrapper rule for them (There can be cases where Detekt users turn off the detekt implementation and use ktlint). This can be a discussion / follow-up

PR #5634, opened 2022-12-22
ElseCaseInsteadOfExhaustiveWhen - Add ignoredSubjectTypes config key (#5623)

### LINKED ISSUE #5623: ElseCaseInsteadOfExhaustiveWhen config to fail when there's only one unused type/value
I tried enabling ElseCaseInsteadOfExhaustiveWhen but I had way too many violations, and most of them were intentional (e.g. there's a Result type that has Success type and a bunch of failure types, and I only want to know if it was a success or failure, and don't care about the specific failure).

However, I found a few cases where I intended to use an exhaustive when, but I either missed a type, or didn't handle null, and used else instead.

I think it would be helpful if there was a config (or separate rule) that would allow an else in a when, unless there was only one more type that has to be handled (or null).

mmorozkov:
A new `ignoredSubjectTypes` property with an empty default value was introduced. It can be configured with a list of subject types fully qualified names which should be ignored by the `ElseCaseInsteadOfExhaustiveWhen` rule.

Fixes #5623

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ElseCaseInsteadOfExhaustiveWhen.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ElseCaseInsteadOfExhaustiveWhen.kt
@@ -7,7 +7,10 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
 import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
+import io.gitlab.arturbosch.detekt.rules.fqNameOrNull
 import org.jetbrains.kotlin.cfg.WhenChecker
 import org.jetbrains.kotlin.psi.KtWhenExpression
 import org.jetbrains.kotlin.resolve.calls.util.getType
@@ -59,6 +62,12 @@ class ElseCaseInsteadOfExhaustiveWhen(config: Config = Config.empty) : Rule(conf
         Debt.FIVE_MINS
     )
 
+    @Configuration(
+        "List of fully qualified types which should be ignored for when expressions with a subject. " +
+            "Example `kotlinx.serialization.json.JsonObject`"
+    )
+    private val ignoredSubjectTypes: List<String> by config(emptyList())
+
     override fun visitWhenExpression(whenExpression: KtWhenExpression) {
         super.visitWhenExpression(whenExpression)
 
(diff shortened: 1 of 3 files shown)


detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ElseCaseInsteadOfExhaustiveWhen.kt
```diff
@@ -59,6 +62,9 @@ class ElseCaseInsteadOfExhaustiveWhen(config: Config = Config.empty) : Rule(conf
         Debt.FIVE_MINS
     )
 
+    @Configuration("List of subject types fully qualified names which should be ignored")
```
cortinico, line 65:
```suggestion
    @Configuration("List of fully qualified types which should be ignored for when expressions with a subject. Example 'kotlinx.serialization.json.JsonObject`")
```

cortinico (approved):
LGTM 👍

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ElseCaseInsteadOfExhaustiveWhen.kt
mmorozkov: Thanks, accepted the suggested changes!

G00fY2:
@mmorozkov I initially created this rule and I really like the addition in this PR. The new config property is especially useful for `when` statements with a sealed class subject that represent the success and specific error states. We have something similar in our project and currently use a lot of suppress annotations.

LGTM

mmorozkov:
@cortinico
What are the next steps? Is there anything on me that blocks this PR to be merged?

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ElseCaseInsteadOfExhaustiveWhen.kt
```diff
@@ -70,6 +79,10 @@ class ElseCaseInsteadOfExhaustiveWhen(config: Config = Config.empty) : Rule(conf
         if (whenExpression.elseExpression == null) return
 
         val subjectType = subjectExpression.getType(bindingContext)
+        if (ignoredSubjectTypes.contains(subjectType?.fqNameOrNull().toString())) {
```
BraisGabin, line 82: nit: 
```suggestion
        if (ignoredSubjectTypes.contains(subjectType?.fqNameOrNull()?.toString())) {
```

I don't know what `contains` thinks about `null`. But checking for `"null"` doesn't seem 100% correct.

BraisGabin:
Could you check comment? Once that is addressed I think that this is good to go.

mmorozkov:
> Could you check comment? Once that is addressed I think that this is good to go.

@BraisGabin Done!

schalkms:
@BraisGabin anything left to do there? Otherwise, we should get this merged for inclusion in the detekt next version.

mmorozkov:
@BraisGabin @schalkms @cortinico guys is there anything that block this one from being merged?

schalkms:
@mmorozkov No blockers left. I'll merge this. If @BraisGabin has anything to add, we will create a new PR.
Btw, sorry for the delay. That shouldn't have happened.

PR #5650, opened 2022-12-30
Don't silently use Kotlin compiler fallback strategy

3flex:
See https://kotlinlang.org/docs/whatsnew18.html#ability-to-disable-the-kotlin-daemon-fallback-strategy


schalkms:
https://kotlinlang.org/docs/gradle-compilation-and-caches.html#kotlin-compiler-fallback-strategy

schalkms (approved):

Issue #3484, opened 2021-02-19
Don't hide nullability problems behind a cast


BraisGabin:
## Expected Behavior of the rule

Noncompliant:
```kotlin
fun foo(bar: Any?) {
  val x = bar as String
}
```

Compliant:
```kotlin
fun foo(bar: Any?) {
  val x = bar!! as String
}
```

## Context
The cast operator can hide null problems in your code. I propose a rule that forbid any cast from a nullable type to a non-null type. If you want to do that you first should cast from nullable to non-null and then do the cast as shown in the example.

The idea is to write code that show that it could fails for two things (nullability and type) and not just one (cast).

I had this idea reading #3479 so they are related but they are not the same.

atulgpt:
Hi @BraisGabin I am willing to implement this issue, should I proceed and will your review the PR?

BraisGabin:
Please, go ahead! I'll review it.

atulgpt:
I @BraisGabin added the https://github.com/detekt/detekt/pull/5653

PR #5653, opened 2022-12-31
Cast nullable to non nullable type

atulgpt:
Fixes https://github.com/detekt/detekt/issues/3484

### DIFF
--- a/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
@@ -98,5 +98,5 @@ fun KtExpression.isNullable(
     return dataFlowTypes.all { it.isNullable() }
 }
 
-fun KtExpression.descriptor(bindingContext: BindingContext): CallableDescriptor? =
+private fun KtExpression.descriptor(bindingContext: BindingContext): CallableDescriptor? =
     getResolvedCall(bindingContext)?.resultingDescriptor
--- a/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
@@ -76,13 +76,10 @@ fun KtExpression.getDataFlowAwareTypes(
 @Suppress("ReturnCount")
 fun KtExpression.isNullable(
     bindingContext: BindingContext,
-    languageVersionSettings: LanguageVersionSettings?,
-    dataFlowValueFactory: DataFlowValueFactory?,
+    languageVersionSettings: LanguageVersionSettings,
+    dataFlowValueFactory: DataFlowValueFactory,
     shouldConsiderPlatformTypeAsNullable: Boolean,
 ): Boolean {
-    languageVersionSettings ?: return false
-    dataFlowValueFactory ?: return false
-
     val safeAccessOperation = safeAs<KtSafeQualifiedExpression>()?.operationTokenNode?.safeAs<PsiElement>()
     if (safeAccessOperation != null) {
         return bindingContext.diagnostics.forElement(safeAccessOperation).none {
--- a/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
@@ -1,13 +1,20 @@
 package io.gitlab.arturbosch.detekt.rules
 
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.config.LanguageVersionSettings
+import org.jetbrains.kotlin.descriptors.CallableDescriptor
+import org.jetbrains.kotlin.diagnostics.Errors
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtSafeQualifiedExpression
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.calls.smartcasts.DataFlowValueFactory
+import org.jetbrains.kotlin.resolve.calls.util.getResolvedCall
 import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
 import org.jetbrains.kotlin.types.KotlinType
 import org.jetbrains.kotlin.types.TypeUtils
+import org.jetbrains.kotlin.types.isFlexible
+import org.jetbrains.kotlin.types.isNullable
 import org.jetbrains.kotlin.util.containingNonLocalDeclaration
 
 fun KotlinType.fqNameOrNull(): FqName? {
@@ -52,3 +59,47 @@ fun KtExpression.getDataFlowAwareTypes(
     return dataFlowInfo.getStableTypes(dataFlowValue, languageVersionSettings)
         .ifEmpty { setOf(originalType) }
 }
+
+/**
+ * Return if overall expression is nullable or not nullable
+ *
+ * ```kotlin
+ * var a: Int? = null
+ * val b = a // RHS expression will be nullable
+ * val c = a!! // RHS expression will be not nullable
+ * val d = (a ?: error("null assertion message")) // RHS expression will be not nullable
+ * val c = 1?.and(2) // RHS expression will be not nullable
+ * ```
+ * [shouldConsiderPlatformTypeAsNullable] determines the behaviour of platform type. Passing true considers
+ * the platform type as nullable otherwise it is not considered nullable in case of false
+ */
+@Suppress("ReturnCount")
+fun KtExpression.isNullable(
+    bindingContext: BindingContext,
+    languageVersionSettings: LanguageVersionSettings?,
+    dataFlowValueFactory: DataFlowValueFactory?,
+    shouldConsiderPlatformTypeAsNullable: Boolean,
+): Boolean {
+    languageVersionSettings ?: return false
+    dataFlowValueFactory ?: return false
+
+    val safeAccessOperation = safeAs<KtSafeQualifiedExpression>()?.operationTokenNode?.safeAs<PsiElement>()
+    if (safeAccessOperation != null) {
+        return bindingContext.diagnostics.forElement(safeAccessOperation).none {
+            it.factory == Errors.UNNECESSARY_SAFE_CALL
+        }
+    }
+    val originalType = descriptor(bindingContext)?.returnType?.takeIf {
+        it.isNullable() && (shouldConsiderPlatformTypeAsNullable || !it.isFlexible())
+    } ?: return false
+    val dataFlowTypes = getDataFlowAwareTypes(
+        bindingContext,
+        languageVersionSettings,
+        dataFlowValueFactory,
+        originalType
+    )
+    return dataFlowTypes.all { it.isNullable() }
+}
+
+fun KtExpression.descriptor(bindingContext: BindingContext): CallableDescriptor? =
+    getResolvedCall(bindingContext)?.resultingDescriptor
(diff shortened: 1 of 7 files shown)


atulgpt:
Hi @BraisGabin I have created this PR for https://github.com/detekt/detekt/issues/3484 but not able to add you as reviewer

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
```diff
@@ -0,0 +1,90 @@
...
+        (typeElement is KtNullableType).ifTrue { return }
+        expression.left.isNullable().ifFalse { return }
+
+        val message =
+            "Use separate `null` assertion and type cast like ('(${expression.left.text} ?: " +
+                "error(\"null assertion message\")) as ${typeElement.text}') instead of '${expression.text}'."
+        report(CodeSmell(issue, Entity.from(operationReference), message))
+    }
+
+    private fun KtExpression.isNullable(): Boolean {
```
atulgpt, line 70: Hi @BraisGabin this code is very similar to the same `isNullable` function present in `NullableToStringCall`. Please let me know if this can be put inside some utility function. If yes, please also guide me where should I put it. Since I am new to this project not much aware of the project structure :)

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
```diff
@@ -0,0 +1,90 @@
...
+class CastNullableToNonNullableType(config: Config = Config.empty) : Rule(config) {
+    override val issue: Issue = Issue(
+        javaClass.simpleName,
+        Severity.Defect,
+        "Nullable type to non-null type cast is found. Consider using two assertions, " +
+            "`null` assertions and type cast",
+        Debt.FIVE_MINS
+    )
+
+    override fun visitBinaryWithTypeRHSExpression(expression: KtBinaryExpressionWithTypeRHS) {
```
atulgpt, line 54: Hi @BraisGabin, should  I use `visitBinaryExpression` here? I couldn't find what is the difference between these two 😅

BraisGabin: To be honest, no idea. If it pass the test seems that this function works for this use case so we can keep it like this :)

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
BraisGabin: `detekt-psi-utils` feels like the proper place to move it.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
atulgpt: Ohhh.. No worries.. 😅

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
atulgpt: Hi @BraisGabin I have made the changes for this

detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
```diff
@@ -52,3 +59,47 @@ fun KtExpression.getDataFlowAwareTypes(
...
+ * val c = 1?.and(2) // RHS expression will be not nullable
+ * ```
+ * [shouldConsiderPlatformTypeAsNullable] determines the behaviour of platform type. Passing true considers
+ * the platform type as nullable otherwise it is not considered nullable in case of false
+ */
+@Suppress("ReturnCount")
+fun KtExpression.isNullable(
+    bindingContext: BindingContext,
+    languageVersionSettings: LanguageVersionSettings?,
+    dataFlowValueFactory: DataFlowValueFactory?,
```
BraisGabin, line 80: These two values shouldn't be nullable. It is the responsibility of our calles to ensure that they are not null and decide what to do if they are.

BraisGabin:
To make ci happy you will need to run `./gradlew apiDump` and commit the changes that this command will generate.

detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
atulgpt: Made those changes. Please re-review

detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
```diff
@@ -52,3 +59,44 @@ fun KtExpression.getDataFlowAwareTypes(
...
+    val dataFlowTypes = getDataFlowAwareTypes(
+        bindingContext,
+        languageVersionSettings,
+        dataFlowValueFactory,
+        originalType
+    )
+    return dataFlowTypes.all { it.isNullable() }
+}
+
+fun KtExpression.descriptor(bindingContext: BindingContext): CallableDescriptor? =
```
BraisGabin, line 101: I'm not 100% sure if we should make this function public. I know, it's used in the other rule too. But maybe it's worth to duplicate this function. Because a ktexpression can have multiple types of descriptors (I think). So this function have sense in this context but it could not have that much sense in others.

atulgpt: Okay.. Let me make this private.
Also do let me know if I need to squash my commits for final merge?

BraisGabin: No need. We always merge squashing :)

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
```diff
@@ -0,0 +1,67 @@
...
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.psi.KtBinaryExpressionWithTypeRHS
+import org.jetbrains.kotlin.psi.KtNullableType
+import org.jetbrains.kotlin.utils.addToStdlib.ifFalse
+import org.jetbrains.kotlin.utils.addToStdlib.ifTrue
+
+/**
+ * Reports cast of nullable variable to non-null type. Cast like this can hide `null`
+ * problems in your code. The compliant code would be that which will correctly check
+ *  for two things (nullability and type) and not just one (cast).
```
cortinico, line 21:
```suggestion
 * problems in your code. The compliant code would either us smart-casting or would check the type before performing the cast.
```

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * <noncompliant>
+ * fun foo(bar: Any?) {
+ *     val x = bar as String
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * fun foo(bar: Any?) {
+ *     val x = (bar ?: error("null assertion message")) as String
```
cortinico, line 31: If would suggest something like:
```
val x = if (bar is String) ... else ...
```

atulgpt: Hi @cortinico wouldn't `val x = if (bar is String) ... else ...` defeat the purpose of the rule? As if type of `bar` is `Any?` then the check `bar is String` can be false due to two reasons

1. `bar` being null 
2. `bar` is not the type of `String`

hence there will be no clear and separate handling of `null` and wrong type. The initial idea was `The idea is to write code that shows that it could fail for two things (nullability and type) and not just one (cast).` at the issue [description](https://github.com/detekt/detekt/issues/3484#issue-811805666)

BraisGabin: I would vote to add two alternatives:

```kotlin
val x = bar!! as String
val x = bar as String?
```

Or if `!!` is *so bad* (I really like this operator):

```kotlin
val x = checkNotNull(bar) as String
val x = bar as String?
```

The idea of this rule is to make clear that there is a cast from nullable to non-nullable. The idea is to avoid it use `as String?` or make it completely clear. For that second part I think that `!!` does its job.

atulgpt: Hi, @BraisGabin I personally use `(bar ?: error("null assertion message")) as String` as I always prefer some custom error message. But I am also okay with `val x = checkNotNull(bar) as String`

I don't like `!!` much also this also gets flagged by `UnsafeCallOnNullableType`

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
atulgpt: Hi, @cortinico changing the message to `* problems in your code. The compliant code would either us smart-casting or would check the type before performing the cast.` doesn't actually reflect the intention of the rule. 
Original code `bar as String` asserts two things(nullability as well as type assertions) at once. And suggested fix is to assert both the cases differently or handle both the cases

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
cortinico: > hence there will be no clear and separate handling of null and wrong type

Which might very much be outside the scope of the rule.
The user receives a `bar: Any?` and wants to know if it can be used as a `String`.

Your code suggestion is too opinionated in failing on `null` state, which the user might not want to do.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
cortinico: I'm unsure I get the point of this rule at this point.

The problem of `bar as String` is that it will fail at runtime if the two types are incompatible. Using a `if (bar is String) ... else ...` would not instead.

Your rule is instead suggesting to fail for two different scenarios (null or non compatible types) with different error messages, right?.

BraisGabin: The point of this rule is that a cast can hide a `!!` behind it. That's a potential error and what this rule does is to point it out to ensure that this is not an error but a decission by the devolver. If you really want to cast directly from `Any?` to `String` you should be fine with `foo!! as String`. If you didn't want to do that, well, then you need to rewrite your code in any other way that doesn't produces a null pointer exception.

To me this rule should not be too opinated. Just point to the potential error and its up to the dev to decide how to fix it. If the problem here are the compilant part we can add 3 different ways to make the code compilant because it depends a lot to what the user wants on each case.

If you are casting your code is smelly. This rule at least ensure that the cast doesn't hide any extra surprise.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
atulgpt: Hi @BraisGabin / @cortinico added below compliant block
```
* <compliant>
 * fun foo(bar: Any?) {
 *     val x = checkNotNull(bar) as String
 * }
 *
 * // Alternative
 * fun foo(bar: Any?) {
 *     val x = (bar ?: error("null assertion message")) as String
 * }
 * </compliant>
```

BraisGabin: That works for me

BraisGabin (approved):

BraisGabin:
LGTM

cortinico (approved):

cortinico:
(sorry for the late review)

Issue #5321, opened 2022-09-19
DataClassContainsFunctions: operator overloading is flagged


TWiStErRob:
## Context
Consider having a mathematical/algebraic class, which might have operators. A really simple example:
```kotlin
data class Vector2(val x: Float, val y: Float) {
	operator fun plus(other: Vector2): Vector2 =
		Vector2(x + other.x, y + other.y)
}
```

## Expected Behavior of the rule
Not exactly sure, but it feels like `DataClassContainsFunctions` should allow these. The only other option is writing an extension function:
```kotlin
data class Vector2(val x: Float, val y: Float)
operator fun Vector2.plus(other: Vector2): Vector2 =
	Vector2(x + other.x, y + other.y)
```
which has some trivial drawbacks, for example having to import each function used in math expression at, every, single, usage.

---

Similar, related use case is non-operator functions that do something algebraic (real use case that triggered this issue):
```kotlin
data class SemanticVersion(
	val major: Int,
	val minor: Int,
	val patch: Int,
) : Comparable<AGPVersion> {
	override fun compareTo(other: AGPVersion): Int = // Nothing wrong with this, it's accepted.
		compareValuesBy(this, other, AGPVersion::major, AGPVersion::minor, AGPVersion::patch)

	infix fun compatible(other: SemanticVersion): Boolean = // this is flagged
		this.major == other.major
}
```
again, this could be extracted as extension function.

---

Note: the action coming out of this might be simply updating the documentation, if I misunderstood the intent. (I understand this rule is opt-in, I'm using `allRules`.)

BraisGabin:
To be honest I don't get why this rule should improve your code in general. As long as your data class is immutable it should be completely fine to have methods on it.

I would rather deprecate this rule than "fix this". But an "allow operators" flag could fix this.

As a workaround you could use `conversionFunctionPrefix` and add `plus` but it doesn't feel right...

TWiStErRob:
> But an "allow operators" flag could fix this.

Immediately following on from the main report, notice that it's quite easy to define non-`operator` operator function, so it's quite tricky to define "operators".

BraisGabin:
Yeah, I don't like the "operator" thing either. For that reason, to me, we should deprecate this rule. Theree are a lot of good reasons to define functions inside a `data class`.

[...]

PR #5658, opened 2023-01-02
Add allowOperators in DataClassContainsFunctions

atulgpt:
Fixes https://github.com/detekt/detekt/issues/5321

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DataClassContainsFunctions.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DataClassContainsFunctions.kt
@@ -52,11 +52,13 @@ class DataClassContainsFunctions(config: Config = Config.empty) : Rule(config) {
         super.visitClass(klass)
     }
 
+    @Suppress("ReturnCount")
     private fun checkFunction(klass: KtClass, function: KtNamedFunction) {
         if (function.isOverride()) return
 
         val functionName = function.name
-        if (functionName != null && (isAllowedConversionFunction(functionName) || checkOperator(function))) return
+        if (functionName != null && conversionFunctionPrefix.any { functionName.startsWith(it) }) return
+        if (allowOperators && function.isOperator()) return
 
         report(
             CodeSmell(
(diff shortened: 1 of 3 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DataClassContainsFunctions.kt
```diff
@@ -52,7 +56,7 @@ class DataClassContainsFunctions(config: Config = Config.empty) : Rule(config) {
         if (function.isOverride()) return
 
         val functionName = function.name
-        if (functionName != null && conversionFunctionPrefix.any { functionName.startsWith(it) }) return
+        if (functionName != null && (isAllowedConversionFunction(functionName) || checkOperator(function))) return
```
BraisGabin, line 59: I would move this check to another `if`. Checking that `functionName != null` and the operator in the same statement is a bit strange.

This way you can keep the old `if` as it was and just add a new one.

BraisGabin:
Oh! And thanks for your contributions :)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/DataClassContainsFunctionsSpec.kt
```diff
@@ -43,6 +44,72 @@ class DataClassContainsFunctionsSpec {
...
+                assertThat(rule.compileAndLint(code)).isEmpty()
+            }
+
+            @Test
+            fun `does not report delegated operators if allowed`() {
+                val config = TestConfig(mapOf(ALLOW_OPERATORS to true))
+                val rule = DataClassContainsFunctions(config)
+                assertThat(rule.compileAndLint(code)).isEmpty()
+            }
+        }
```
BraisGabin, line 110: I'm not sure if I even like to have these tests here... This rule shouldn't care about delegation. And if it should we should handle that in a different issue/PR.

atulgpt:
> Oh! And thanks for your contributions :)

Thanks :)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DataClassContainsFunctions.kt
atulgpt: Made these changes

atulgpt:
Hi @BraisGabin can you guide me why there is one check failing? Thanks

BraisGabin (approved):

Issue #5468, opened 2022-10-24
Don't allow `runCatching` with supend functions inside


BraisGabin:
## Expected Behavior of the rule
Raise any usage of `runCatching` that calls a `suspend` function.

Example:

```kotlin
suspend fun foo(): String { /*...*/ }

suspend fun bar(): String? {
  return runCatching { foo() }.getOrNull()
}
```

## Context
From the documentation:

> While catching `Exception` is an anti-pattern, this issue may surface in more subtle ways, like when using the [`runCatching`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/run-catching.html) function, which does not rethrow [`CancellationException`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellation-exception/index.html).

Extracted from: https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-is-cooperative

If you catch and don't rethrow all the `CancellationException` your coroutines are not cancelled even if you cancel their `CoroutineScope`. And that can create huge issues that are really difficult to fix because the errors and exceptions that you get are not related with that `runCatching`.

At work we had this issue for several months until we realised what was going on. We already have [`TooGenericExceptionCaught`](https://detekt.dev/docs/next/rules/exceptions#toogenericexceptioncaught) that works with `try/catch` but we don't have anything to spot this issue. At work we are considering to forbid `runCatching` all together but I think that at detekt we should have a rule to spot this.

It's true that you can be careful and check the type of the caught exception and rethrow it in case of `CancellationException` but that's far to easy to forget.

I think that this rule should be inside `coroutines` but `potential-bugs` could be also a good fit.

TWiStErRob:
What's the solution to a violation like this?
`runCatchingExceptCancellationExceptionWhichIsRethrown` (with a better name ofc)?

cortinico:
> What's the solution to a violation like this?

I'm also a bit hesitant on forbidding `runCatching` for all sorts of `suspend` fun. It feels like we could limit its usage quite a lot.

BraisGabin:
This is similar to forbid the usage of `lateinit var`. There are ways to use it properly. But it's risky. This would be the same.

`runCacthing` is the same as a `try/catch(Throwable)` and we all agree that the second one is bad.

> What's the solution to a violation like this?

[...]

PR #5666, opened 2023-01-04
Add SuspendFunSwallowedCancellation rule

atulgpt:
Fixes https://github.com/detekt/detekt/issues/5468

More details:

This PR tries to detekt, all the `suspend fun` inside a `runCatching` block. Currently, I am finding each suspension point(`suspend fun`) inside the `runCatching` block and then for each suspension point trying to backpropagate if I can reach back to the original `runCatching` block or not.
Cases, where I am not able to reach, are as follows (these cases will not be reported a code smell for the current `runCatching` block)

1. If the suspension point is called by lambda which is not an inline function
2. If suspension point is called by fun which is inline but lambda, where that suspension point is present, is not inlined lambda(in other words lambda is marked with `noinline` or `crossinline`)

If any of the above points gets encountered, I break the backpropagation and then check if the returned function descriptor is the same as the current `runCatching` block(I am not checking against `fqName` as there could be multiple nested `runCatching` and then deeper suspension point will be reported against all the parent `runCatching`)

I have created this draft PR for early feedback and suggestion on improving the propagation around TCs. Although I agree that I am propagation a `runCatching`..`suspend fun` from top to bottom then bottom to top, I want to know if should I optimise this(do this in single propagation(This will require custom propagation of child or some other logic)) or is it okay. I'm fine either way

Update: 
Now in the [commit](https://github.com/detekt/detekt/pull/5666/commits/2a81f8c7c4d97746fbc4264b8276e5876c5d00ff), I have simplified the traversal logic and only going inside a function for traversal when that function is inline or value parameter of the function is not marked `noinline` or `crossinline`(This helped in preventing the backpropagation from the suspension point)


detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatching.kt
```diff
@@ -0,0 +1,142 @@
...
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.util.getResolvedCall
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+
+/**
+ * Suspend functions should not be called inside `runCatching` as `runCatching` catches
+ * all the exception while for Coroutine cooperative cancellation to work, we have to
+ * never catch the `CancellationException` exception or rethrowing it again if caught
+ *
+ * See https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-is-cooperative
```
TWiStErRob, line 28: Suggestion for wording, please rewrap as necessary:
```suggestion
`suspend` functions should not be called inside `runCatching`'s lambda block, because `runCatching` catches all the `Exception`s. For Coroutines to work in all cases, developers should make sure to propagate `CancellationException` exceptions. This means `CancellationException` should never be:
 * caught and swallowed (even if logged)
 * caught and propagated to external systems
 * caught and shown to the user

they must always be rethrown in the same context.

Using `runCatching` increases this risk of mis-handling cancellation. If you catch and don't rethrow all the `CancellationException`, your coroutines are not cancelled even if you cancel their `CoroutineScope`.

This can very easily lead to:
 * unexpected crashes
 * extremely hard to diagnose bugs
 * memory leaks
 * performance issues
 * battery drain

For reference, see [Kotlin documentation](https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-is-cooperative).

If your project wants to use `runCatching` and `Result` objects, it is recommended to write a `coRunCatching` utility function which immediately re-throws `CancellationException`; and forbid `runCatching` and `suspend` combinations by activating this rule.
```

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatching.kt
```diff
@@ -0,0 +1,142 @@
...
+ */
+@RequiresTypeResolution
+class SuspendFunInsideRunCatching(config: Config) : Rule(config) {
+    override val issue = Issue(
+        id = "SuspendFunInsideRunCatching",
+        severity = Severity.Minor,
+        description = "The `suspend` functions should be called inside `runCatching` block as it also swallows " +
+            "`CancellationException` which is important for cooperative cancellation." +
+            "You should either use specific `try-catch` only catching exception that you are expecting" +
+            " or rethrow the `CancellationException` if already caught",
```
TWiStErRob, line 74: please shorten this to prevent https://github.com/detekt/detekt/issues/4750 -like issues.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatching.kt
```diff
@@ -0,0 +1,142 @@
...
+            if (descendant.getResolvedCall(bindingContext)?.resultingDescriptor?.isSuspend == true && shouldReport(
+                    resultingDescriptor,
+                    descendant,
+                    bindingContext,
+                )
+            ) {
+                val message =
+                    "The suspend function call ${descendant.text} is inside `runCatching`. You should either " +
+                        "use specific `try-catch` only catching exception that you are expecting or rethrow the " +
+                        "`CancellationException` if already caught"
```
TWiStErRob, line 95:
```suggestion
                    "The suspend function call ${descendant.text} is inside `runCatching`. You should either " +
                        "use specific `try-catch` only catching exception that you are expecting or rethrow the " +
                        "`CancellationException` if already caught."
```

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatching.kt
```diff
@@ -0,0 +1,142 @@
...
+
+    override fun visitCallExpression(expression: KtCallExpression) {
+        super.visitCallExpression(expression)
+
+        val resultingDescriptor = expression.getResolvedCall(bindingContext)?.resultingDescriptor
+        resultingDescriptor ?: return
+        if (resultingDescriptor.fqNameSafe != RUN_CATCHING_FQ) return
+
+        expression.forEachDescendantOfType<KtCallExpression> { descendant ->
+            if (descendant.getResolvedCall(bindingContext)?.resultingDescriptor?.isSuspend == true && shouldReport(
```
TWiStErRob, line 86: This looks like a performance issue waiting, but I don't have enough experience to flag it fully. What my gut feeling at first look is that this will traverse nested calls multiple times (up and down and descendant and parents, and Rule will call multiple times too), essentially resulting in $O(n^2)$ runtime. Is it so that because of inline functions it's required to deeply visit all calls to essentially look for any suspension points (=little icons on the IDEA sidebar)?

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatching.kt
```diff
@@ -0,0 +1,142 @@
...
+                        it.index
+                    }
+                val callExpressionIndexInParentCall = parentCallExp.valueArguments.indexOfFirst { valueArgument ->
+                    valueArgument?.findDescendantOfType<KtCallExpression> {
+                        it == callExpression
+                    } != null
+                }
+                isParentRunCatching ||
+                    isInline.not() ||
+                    noInlineAndCrossInlineValueParametersIndex.contains(callExpressionIndexInParentCall)
```
TWiStErRob, line 133: Is it necessary to construct this temporary index, is it not possible to directly ask the same? i.e. something like check `isCrossinline || isNoInline` on `valueArgument`?

TWiStErRob (commented):
Really nice, looks pretty advanced, looking at the issue description I didn't expect this complexity.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatching.kt
atulgpt: Really nicely written documentation. Thanks for suggestion

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatching.kt
atulgpt: Hi @TWiStErRob, visit function will only be called for `fun runCatching` see check `if (resultingDescriptor.fqNameSafe != RUN_CATCHING_FQ) return` so this code will only work inside `runCatching` block

Regarding
> Is it so that because of inline functions it's required to deeply visit all calls to look for any suspension points (=little icons on the IDEA sidebar)?

This is for check if there is `suspend fun` inside `fun runCatching` or not

Now, I do agree this rule then again back traverse to see if that found suspension point was exactly reachable by main `fun runCatching`(either chain of `inline fun` or directly) (This is what I am doing line 127 `valueArgument?.findDescendantOfType<KtCallExpression>`) This seems to be a little bit concerning to me and I am planning to rework(that's why this is draft PR for early review and suggestions 🙂). BTW any resource or existing Rule on how to custom traverse the child nodes?

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatchingSpec.kt
```diff
@@ -0,0 +1,531 @@
...
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Finding
+import io.gitlab.arturbosch.detekt.rules.KotlinCoreEnvironmentTest
+import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
+import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
+import org.junit.jupiter.api.Test
+
+@KotlinCoreEnvironmentTest
+class SuspendFunInsideRunCatchingSpec(private val env: KotlinCoreEnvironment) {
```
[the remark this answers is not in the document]
atulgpt: Can you please elaborate on this point?

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatchingSpec.kt
```diff
@@ -0,0 +1,531 @@
...
+                inlineBlock: () -> Unit,
+                crossinline crossinlineBlock: suspend () -> Unit,
+            ) = inlineBlock().toString() + MainScope().launch {
+                noinlineBlock()
+            }.toString() + runBlocking {
+                crossinlineBlock()
+            }.toString() + noinlineBlock()
+        
+            suspend fun bar()
+            {
```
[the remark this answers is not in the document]
atulgpt: Haha true. My work project follow few coding convention from C

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatchingSpec.kt
```diff
@@ -0,0 +1,531 @@
...
+
+            inline fun foo(
+                noinline noinlineBlock: suspend () -> Unit,
+                inlineBlock: () -> Unit,
+                crossinline crossinlineBlock: suspend () -> Unit,
+            ) = inlineBlock().toString() + MainScope().launch {
+                noinlineBlock()
+            }.toString() + runBlocking {
+                crossinlineBlock()
+            }.toString() + noinlineBlock()
```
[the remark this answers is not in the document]
atulgpt: `toString()` is actually required on only first which converts the expression to `String`. I will double check remove the redundant `toString()` call

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatchingSpec.kt
```diff
@@ -0,0 +1,531 @@
...
+
+    @Test
+    fun `does report in case suspend callable refernce is invoked`() {
+        val code = """
+            import kotlinx.coroutines.delay
+
+            suspend fun bar() = delay(1000)
+            suspend fun foo() {
+                runCatching {
+                    ::bar.invoke()
```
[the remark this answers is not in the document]
atulgpt: Thanks 🙂

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatchingSpec.kt
```diff
@@ -0,0 +1,531 @@
...
+
+            suspend fun foo() {
+                runCatching {
+                    val string = "${'$'}{delay(1000)}"
+                }
+            }
+        """.trimIndent()
+
+        val findings = subject.compileAndLintWithContext(env, code)
+        assertFindings(findings, "delay(1000)")
```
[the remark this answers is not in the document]
atulgpt: Sure. Let me recheck. I am not sure how highlighting works currently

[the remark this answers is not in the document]
TWiStErRob: Have a look at the fix of #5316 and related issues. You might be able to use the same, something like atName.

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatchingSpec.kt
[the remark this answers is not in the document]
TWiStErRob: The PSI is very different in cases I listed above, so to have full coverage, I would add in those explicitly or mix into existing tests. Simple things like:
```
val x = runCatching { y() }
when(x) {
    is Success -> TODO()
    is Error -> TODO()
}
```
or just `return x`, but in some way, use the result.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatching.kt
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * // Alternate
+ * suspend fun foo() {
+ *     bar(1_000L)
+ * }
+ * </compliant>
+ *
+ */
+@RequiresTypeResolution
+class SuspendFunInsideRunCatching(config: Config) : Rule(config) {
```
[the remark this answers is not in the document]
atulgpt: Made this change

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatchingSpec.kt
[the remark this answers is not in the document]
atulgpt: Thanks for suggesting it. I have a few TCs around these. Please review. Thanks

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatchingSpec.kt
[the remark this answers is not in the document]
atulgpt: Done

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatching.kt
atulgpt: Done. Now it is like below
```
"`suspend` functions should not be called inside `runCatching`'s lambda block, because " +
            "`runCatching` catches all the `Exception`s."
```

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatching.kt
atulgpt: Hi, @TWiStErRob I have simplified the logic to have a single traversal. Also, it would be great if @t-kameyama can also review it once.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatching.kt
atulgpt: Simplified the logic here as well

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatching.kt
t-kameyama: LGTM

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInsideRunCatching.kt
TWiStErRob: Thoughts on this shorter, and more explicit version?
``"`runCatcing` does not propagate `CancellationException`, don't use it with `suspend` lambda blocks."``

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
```diff
@@ -0,0 +1,180 @@
...
+import org.jetbrains.kotlin.psi.KtValueArgument
+import org.jetbrains.kotlin.psi.psiUtil.forEachDescendantOfType
+import org.jetbrains.kotlin.psi.psiUtil.getParentOfType
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.util.getParameterForArgument
+import org.jetbrains.kotlin.resolve.calls.util.getResolvedCall
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+
+/**
+`suspend` functions should not be called inside `runCatching`'s lambda block, because `runCatching` catches all the
```
TWiStErRob, line 26: Missing stars for kdoc alignment.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
```diff
@@ -0,0 +1,180 @@
...
+        val resultingDescriptor = expression.getResolvedCall(bindingContext)?.resultingDescriptor
+        resultingDescriptor ?: return
+        if (resultingDescriptor.fqNameSafe != RUN_CATCHING_FQ) return
+
+        expression.forEachDescendantOfType<KtCallExpression>({
+            expression == it || shouldTraverseInside(it, bindingContext)
+        }) { descendant ->
+            val callableDescriptor = descendant.getResolvedCall(bindingContext)?.resultingDescriptor
+            if (callableDescriptor?.isSuspend == true) {
+                report(
```
TWiStErRob, line 110: Won't this flag the same `runCatching` from each suspension point?
```
runCatching { <-\-\-\
    delay(100) -/ | |
    delay(100) ---/ |
    delay(100) -----/
}
```

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
```diff
@@ -0,0 +1,180 @@
...
+            callExpression ?: return false
+            val valueParameterDescriptor =
+                callExpression.getResolvedCall(bindingContext)?.getParameterForArgument(psiElement)
+            valueParameterDescriptor ?: return false
+            if (valueParameterDescriptor.isCrossinline || valueParameterDescriptor.isNoinline) return false
+        }
+        return true
+    }
+
+    private fun KtForExpression.doesLoopHasSuspendingIterators(): Boolean {
```
TWiStErRob, line 154: "loop" is contained in receiver type
```suggestion
    private fun KtForExpression.hasSuspendingIterators(): Boolean {
```
or
```suggestion
    private fun KtForExpression.hasSuspendingIteratorsInLoop(): Boolean {
```

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
```diff
@@ -0,0 +1,180 @@
...
+                report(
+                    message = "The suspend function call ${callableDescriptor.fqNameSafe.shortName()} called inside " +
+                        "`runCatching`. You should either use specific `try-catch` only catching exception that you " +
+                        "are expecting or rethrow the `CancellationException` if already caught.",
+                    expression
+                )
+            }
+        }
+
+        expression.forEachDescendantOfType<KtForExpression> { descendant ->
```
TWiStErRob, line 119: will this go too deep into all for loops, which might be nested within other execution contexts?

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
```diff
@@ -0,0 +1,180 @@
...
+            val callableDescriptor =
+                psiElement.getResolvedCall(bindingContext)?.resultingDescriptor as? FunctionDescriptor
+            callableDescriptor ?: return false
+
+            if (callableDescriptor.fqNameSafe == RUN_CATCHING_FQ) return false
+
+            if (callableDescriptor.isInline.not()) return false
+        }
+
+        if (psiElement is KtValueArgument) {
```
TWiStErRob, line 143: Distinct looking `if`s side by side, maybe use `when(psiElement) { is KTCall -> {} is KtValue -> {} else -> true` }? Might need to change `if-return`s to `return a && b`.

Is the fall-through maybe necessary?

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
```diff
@@ -0,0 +1,180 @@
...
+            }
+        }
+    }
+
+    @Suppress("ReturnCount")
+    private fun shouldTraverseInside(psiElement: PsiElement, bindingContext: BindingContext): Boolean {
+        if (psiElement is KtCallExpression) {
+            val callableDescriptor =
+                psiElement.getResolvedCall(bindingContext)?.resultingDescriptor as? FunctionDescriptor
+            callableDescriptor ?: return false
```
TWiStErRob, line 136: this can be merged together to prevent strange "return on variable":
```suggestion
            val callableDescriptor =
                (psiElement.getResolvedCall(bindingContext)?.resultingDescriptor as? FunctionDescriptor)
                ?: return false
```
alternatively match the `if-return` style below instead of elvis.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
```diff
@@ -0,0 +1,180 @@
...
+
+            if (callableDescriptor.isInline.not()) return false
+        }
+
+        if (psiElement is KtValueArgument) {
+            val callExpression = psiElement.getParentOfType<KtCallExpression>(true)
+            callExpression ?: return false
+            val valueParameterDescriptor =
+                callExpression.getResolvedCall(bindingContext)?.getParameterForArgument(psiElement)
+            valueParameterDescriptor ?: return false
```
TWiStErRob, line 148: same as above, this could be 2 statements with non-null types inferred on val instead of smart cast

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
```diff
@@ -0,0 +1,626 @@
...
+        val code = """
+            import kotlinx.coroutines.delay
+
+            suspend fun bar() = delay(2000)
+
+            suspend fun foo() {
+                runCatching {
+                    delay(1000L)
+                    runCatching { 
+                        bar()
```
TWiStErRob, line 42: is there a way to test explicitly (separate test) which `runCatching` is triggering `bar()` finding? (probably need to change the snippet in some way)

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
atulgpt: Hahah nice catch. This came when applying GitHub suggestion but now I got one detekt rule idea 😅

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
atulgpt: Currently, it will flag thrice for the `delay` method for its three usages. To me, this seemed logical as now, I am pointing out the root causes which raise this warning while in the case of pointing `runCatching`, the user has to search which suspend function caused the warning

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
atulgpt: I have added `does report for in case of nested runCatching with suspend fun call in inner runCatching` and `does report for in case of nested runCatching with suspend fun call in outer runCatching` for this

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
atulgpt: I have added `does report when nested suspending iterator is used`, `does report when nested iterator with one suspending iterator is used`, `does report when suspending iterator is used withing inlined block` and `does not report when suspending iterator is used withing non inlined block`

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
atulgpt: I couldn't get this comment completely. I have made some changes maybe those will address this

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
TWiStErRob: It's trivial to see what suspends when opened in IDE. I agree from GitHub it might be harder. What I would do is start searching down from runCatching and stop at the first suspension point. Then report a finding for runCatching token, but include the name of the suspension point in the message as cause. This would reduce the amount of findings and unify the messaging too. In the end what you need to change when you get a detection is the runCatching.

@BraisGabin thoughts?

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
```diff
@@ -0,0 +1,179 @@
...
+    override fun visitCallExpression(expression: KtCallExpression) {
+        super.visitCallExpression(expression)
+
+        val resultingDescriptor = expression.getResolvedCall(bindingContext)?.resultingDescriptor
+        resultingDescriptor ?: return
+        if (resultingDescriptor.fqNameSafe != RUN_CATCHING_FQ) return
+
+        val shouldTraverseInside: (PsiElement) -> Boolean = {
+            expression == it || shouldTraverseInside(it, bindingContext)
+        }
```
TWiStErRob, line 107: I think a local fun might be easier to read:
```suggestion
        fun shouldTraverseInside(element: PsiElement): Boolean =
            expression == element || shouldTraverseInside(element, bindingContext)
```

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
```diff
@@ -0,0 +1,179 @@
...
+        description = "`runCatching` does not propagate `CancellationException`, don't use it with `suspend` lambda " +
+            "blocks.",
+        debt = Debt.TEN_MINS
+    )
+
+    override fun visitCallExpression(expression: KtCallExpression) {
+        super.visitCallExpression(expression)
+
+        val resultingDescriptor = expression.getResolvedCall(bindingContext)?.resultingDescriptor
+        resultingDescriptor ?: return
```
TWiStErRob, line 102: I think this is simpler, and declares the variable more cleanly (no null-smartcast)
```suggestion
        val resultingDescriptor = expression.getResolvedCall(bindingContext)?.resultingDescriptor ?: return
```

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
```diff
@@ -0,0 +1,179 @@
...
+                report(
+                    message = "The suspend function call ${callableDescriptor.fqNameSafe.shortName()} called inside " +
+                        "`runCatching`. You should either use specific `try-catch` only catching exception that you " +
+                        "are expecting or rethrow the `CancellationException` if already caught.",
+                    expression
+                )
+            }
+        }
+
+        expression.forEachDescendantOfType<KtForExpression>(shouldTraverseInside) { descendant ->
```
TWiStErRob, line 120: btw, why is `for` special?

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
```diff
@@ -0,0 +1,179 @@
...
+                        ?: return false
+
+                callableDescriptor.fqNameSafe != RUN_CATCHING_FQ && callableDescriptor.isInline
+            }
+            is KtValueArgument -> {
+                val callExpression = psiElement.getParentOfType<KtCallExpression>(true) ?: return false
+                val valueParameterDescriptor =
+                    callExpression.getResolvedCall(bindingContext)?.getParameterForArgument(psiElement) ?: return false
+
+                valueParameterDescriptor.isCrossinline.not() && valueParameterDescriptor.isNoinline.not()
```
TWiStErRob, line 147: @BraisGabin What's the preffered style in Detekt `.not()` or `!`?

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
```diff
@@ -0,0 +1,179 @@
...
+            else -> true
+        }
+    }
+
+    private fun KtForExpression.hasSuspendingIterators(): Boolean {
+        val iteratorResolvedCall = bindingContext[BindingContext.LOOP_RANGE_ITERATOR_RESOLVED_CALL, this.loopRange]
+        val loopRangeHasNextResolvedCall =
+            bindingContext[BindingContext.LOOP_RANGE_HAS_NEXT_RESOLVED_CALL, this.loopRange]
+        val loopRangeNextResolvedCall = bindingContext[BindingContext.LOOP_RANGE_NEXT_RESOLVED_CALL, this.loopRange]
+        return listOf(iteratorResolvedCall, loopRangeHasNextResolvedCall, loopRangeNextResolvedCall).any {
```
TWiStErRob, line 158: if you remove `ResolvedCall` suffix, it should still bear the same meaning, but less noisy, wydt?

TWiStErRob (commented):
Feels like the code is getting shorter, but definitely simpler, nice one!

atulgpt:
> Feels like the code is getting shorter, but definitely simpler, nice one!

Thanks

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
atulgpt: `for` is not detected in `KtCallExpression` traversal. I got the inspiration from `RedundantSuspendModifier` rule regarding handling `for` case

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
atulgpt: Currently, in the project, there are few existing usages of `.not()`. To me `.not()` feels more connected to an actual variable which it is negating. But I am okay with both the style 😅 @BraisGabin to suggest

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
atulgpt: Made this change

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
TWiStErRob: It's because the iterators missing the loopRange prefix, while the others have it. Once you include that, all names are sandwiched, so the suffix can be removed as it's essentially Hungarian notation.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
TWiStErRob: That's just weird :) I guess it's because each PSI element has some default "children" and for `for` that's the statements inside, rather than the header. By this account the same problem might apply to all control structures (if-ifelse conditions, when target, when conditions, while condition).

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
TWiStErRob: But there's only one violation for each runCatching. The suspension point is not the problem, people won't change the suspending calls, but hopefully the way the exceptions are handled, i.e. the block around the calls.
```
runCatching { <-\-\-\
    delay(100) -/ | |
    delay(100) ---/ |
    delay(100) -----/
    runCatching { <-\-\
        delay(100) -/ |
        delay(100) ---/
    }
}
```
In this example you have two problematic `runCatching` calls. It is caused by 3 and 2 suspension points respectively, but that's not relevant to the issues. Each of the 5 `suspend` calls cases their parent block to suspend, but the problem itself is that it **is** a "`suspend` block". The reason doesn't matter, just helps to understand why it is suspended.

So if we report the two runCatching only, and they fix those two, no more of this violation will surface from the same code, because changing the parent block fixes the issue.

Hope this helps.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
atulgpt: Sorry to be more clear. `for` with suspending iterator is not detected in `KtCallExpression` obviously all the call expression even inside `for` loop, `while` loop etc will be detected.

For example,
```
class SuspendingItr {
    suspend operator fun iterator(): Iterator<Int> = iterator { yield(1) }
}

suspend fun test() {
    for (i in SuspendingItr()) {
        // here in this for loop as per kotlin PSI goes there is not any call
        // but since `SuspendingItr` is as suspending iterator when kotlin compiler generates the code
        // it invokes the `iterator` function which is suspending in nature
    }
}
```

TWiStErRob: Ooh, `suspend operator fun iterator()`, that's new for me 😅.

atulgpt: BTW I tested the below code 
```
suspend fun bar() = 10

suspend fun foo() {
    runCatching {
        for (i in 1..bar()) {
            println(i)
        }
    }
}
```
And code correctly detects the `bar` call

TWiStErRob: Could the same happen with `in` or `..`/`..<` or other operators?

atulgpt: Haha.. Nice catch.. Those cases are failing

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
atulgpt: Hmmm.. Make sense.. I have updated the code report the `runCatching` function name at the call site location

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
atulgpt: BTW same is handled in the new code and added UTs around the same

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
BraisGabin: I think that report the `runCatching` seems like a good solution.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
BraisGabin: We don't have any style rule about this. I think that as long as it is easy to read we are fine with both. We have a mix right now.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
atulgpt: Hi @TWiStErRob made these renaming changes and added `loopRangeHas` prefix to iterator variable

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
```diff
@@ -296,7 +361,7 @@ class SuspendFunSwallowedCancellationSpec(private val env: KotlinCoreEnvironment
         """.trimIndent()
 
         val findings = subject.compileAndLintWithContext(env, code)
-        assertFindingsForSuspendCall(findings)
+        assertThat(findings).isEmpty()
```
TWiStErRob, line 364: hmm, looking at this code, crossinline doesn't necessarily mean it'll be executed in another context, e.g.
```
suspend inline fun <R> foo(crossinline block: suspend () -> R) = block()
```

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
```diff
@@ -757,26 +868,292 @@ class SuspendFunSwallowedCancellationSpec(private val env: KotlinCoreEnvironment
...
+                listOf(SourceLocation(6, 5)),
+                listOf(SourceLocation(6, 16))
+            )
+        }
+
+        @Test
+        fun `does report in case of suspend plus operator`() {
+            val code = """
+                class OperatorClass {
+                    suspend operator fun plus(test: OperatorClass) = test
```
TWiStErRob, line 983: `plus` can be used as `plusAssign` too.
```
class C
operator fun C.plus(i: Int): C = TODO()

fun f() {
	var x = C()
	x += 1
}
```

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
```diff
@@ -0,0 +1,1159 @@
...
+        val findings = subject.compileAndLintWithContext(env, code)
+        assertFindingsForSuspendCall(
+            findings,
+            listOf(SourceLocation(4, 5)),
+            listOf(SourceLocation(4, 16))
+        )
+    }
+
+    @Nested
+    inner class `Containing suspending operator` {
```
TWiStErRob, line 879: should probably test their explicit call too?
```
operatorClass++
// vs
operatorClass.inc()
```
or is that covered by normal function call tests?

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
TWiStErRob: This opened a can of worms, should we report the same problem for the other rule too? You said you adopted some suspend related code from somewhere, does this have the same problem?

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
atulgpt: There is one TC for this `does report in case of suspend invoked operator`

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
atulgpt: Should I flag it or not? As we can't determine 100% time if the `crossinline` using the `coroutineContext` of the `runCatching`. In one case(current one), we will have false negative, in the other case when we disallow `suspend fun` inside `crossinline` we will have the false positive

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
atulgpt: I tested with below TC
```kotlin
class C
suspend operator fun C.plus(i: Int): C = TODO()

suspend fun f() {
    runCatching {
        var x = C()
        x += 1
    }
    
}
```

It is working fine.

Actually initially I was only checking `KtCallExpression` now I also started checking `KtOperationExpression` that's why this case is also covered.
`x += 1` is `KtBinaryExpression` which is child class of `KtOperationExpression`

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
TWiStErRob: yeah, I meant for all the other operators, because their AST doesn't look like function calls while `invoke` does, in both cases.

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
atulgpt: Actually just checked the existing TCs, I already had one TC for plusAssign use case. Name of the TC `does report in case of suspend plusAssign operator` Bode as below
```kotlin
@Test
fun `does report in case of suspend plusAssign operator`() {
    val code = """
        class OperatorClass {
            suspend operator fun plusAssign(operatorClass: OperatorClass) { }
        }
        
        suspend fun foo() {
            runCatching {
                val operatorClass1 = OperatorClass()
                val operatorClass2 = OperatorClass()
                operatorClass1 += (operatorClass2)
            }
        }
    """.trimIndent()

    assertFindingsForSuspendCall(
        findings,
        listOf(SourceLocation(6, 5)),
        listOf(SourceLocation(6, 16))
    )
}
```

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
TWiStErRob: yep, I'm very aware, and I have no idea how to proceed on this.
 * false positives are annoying, but can be easily suppressed, at least it raises awareness and enforces understanding of one's own code
 * false negatives are even more annoying, because you trust the static analyzer to find all the problems, and if it misses one, your confidence in it reduces, because you relied on it to prevent bugs and you have bugs.

@BraisGabin @cortinico ?

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
atulgpt: In that case, call comes under `KtCallExpression` which I am already checking against suspend operation. Also I have verified the same using below TC
```kotlin
@Test
fun `does report in case of suspend inc operator as function`() {
    val code = """
        class OperatorClass {
            suspend operator fun inc(): OperatorClass = OperatorClass()
            suspend operator fun dec(): OperatorClass = OperatorClass()
        }
        
        suspend fun foo() {
            runCatching {
                var operatorClass = OperatorClass()
                operatorClass.inc()
            }
            runCatching {
                var operatorClass = OperatorClass()
                operatorClass.dec()
            }
        }
    """.trimIndent()

    assertFindingsForSuspendCall(
        findings,
        listOf(SourceLocation(7, 5), SourceLocation(11, 5)),
        listOf(SourceLocation(7, 16), SourceLocation(11, 16))
    )
}
```

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
TWiStErRob: yeah but that's a one-to-one mapping from operator fun to operator, there's way more magic going on with += to plus mapping although if you treat all operators in one block and the resolution is done by JetBrains' code then I guess we're covered.

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
atulgpt: TBH I think this case can also be addressed, by getting the PSI element from the `CallDescriptor`. But that will add a lot more complexity the analysis and I am not in favor of also analysing the method where `crossinline` lamda is used.

Rest I agree with your opinion

TWiStErRob: > I am not in favor of also analysing the method where crossinline lamda is used.

me neither, because you might not even have the source code for that, only the compiled (potentially even obfuscated) body.

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
atulgpt: Hi @TWiStErRob calling an operator as a function is simple extension or normal function which I have already added in my TC(for extension and normal function) so I don't see the return of adding call syntax for all the operators.

But if you just want the above TC to be added then I am completely fine with it 🙂

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
atulgpt: This I will check and verify. If there will be issue I'll create a issue in the project

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
TWiStErRob: I think we're good here, this was an example, I meant the comment for all of them. Let's say that `invoke` TC covers all cases.

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
atulgpt: @BraisGabin / @cortinico any thoughts on this?

atulgpt: Hi @TWiStErRob I had an idea today shouldn't code like 
```
suspend inline fun <R> foo(crossinline block: suspend () -> R) = block()
```
be flagged by the detekt that lambda `block` can be inline. Actually, the code demonstrated by you misuses the `crossinline` modifier. What are your thoughts on this? Is there something that I am missing?

BraisGabin:
I was out for some time and there are a lot of comments on these PR. Is there any blockers to merge this? Can I help to unblock something?

To me the tests feel a bit too long but this rule is a clear win for our users so we can improve that in later PRs. I mean, this implements just "half" issue. The `runCatching` thing. We need to implement the `try/catch` too.

BraisGabin (approved):

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
atulgpt: Thanks @TWiStErRob I am resolving this issue as per the above comment

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
TWiStErRob: I think as long as the code compiles and executes, it's "valid", mis-use or not, and Detekt should be able to handle it. After all, Detekt's job is to flag problematic code.

atulgpt:
Hi @BraisGabin `try-catch` was agreed to be implemented as a separate PR(that's why the generic name `SuspendFunSwallowedCancellation`). Your input Is needed for the [comment](https://github.com/detekt/detekt/pull/5666#discussion_r1071220952)(context: Currently I am allowing false negative while addressing that comment will favour false positive instead of false negative)

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
TWiStErRob: @atulgpt what about this one? Did you address it?

TWiStErRob:
@BraisGabin note: unresolved conversations are unresolved, hence the name ;)

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
atulgpt: Hi, @TWiStErRob not yet. I have one more PR which is dependent on the `crossinline` discussion. There I have to few TCs at that time I will add similar TCs and validate in other coroutines related rules

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
atulgpt: Sure.. I am not against your proposal. I am just throwing one idea for a problematic code which is a good candidate for a rule 🙂

BraisGabin: First of all: A rule for `crossinline` would be great. To be honest I don't feel confortable with that modifier so a rule to help with it would be great! @atulgpt could you create a PR with a proposal of what the rule should do?

BraisGabin: It took me a while to undestand why `crossinline` and `noinline` were relevant here.

> hmm, looking at this code, crossinline doesn't necessarily mean it'll be executed in another context, e.g.
> 
> ```
> suspend inline fun <R> foo(crossinline block: suspend () -> R) = block()
> ```

But if `foo` is supend and `foo` is called inside a `runCatching` we don't care about what `foo` does or if it has a crossinline parameter. We can stop the check, that `runCatching` should be reported.

Maybe I'm missing something but looking at your snippet and applying it to the test where this comment is I just see that.

atulgpt: Hi @BraisGabin, let me make a complete example
```kotlin
import kotlinx.coroutines.MainScope
import kotlinx.coroutines.delay
import kotlinx.coroutines.launch
inline fun <R> foo(crossinline block: suspend () -> R) = MainScope().launch {
    block()
}
suspend fun bar() {
    runCatching {
         foo {
             delay(1000L) // discussion is around this point. Whether this should be flagged by the rule or not
        }
    }
}
```

Currently, I am not flagging this if there are `crossinline` and `noinline` modifiers in the lambda, while in case of no modifier, since the function `foo` is inline `delay` gets called in the context where `runCatching` is called.

And hence no modifier inline function with suspend fun inside lambda is getting reported

But there is a catch in `crossinline`, one can mark a lambda unnecessarily `crossinline` and avoid getting detected by the rule. As lambda passed in `crossinline` can be passed to some other function and hence can be a valid example while unnecessarily adding the modifier and calling the lambda block directly ideally be detected by the rule(But currently I don't have a way to distinguish between these two cases see [comment](https://github.com/detekt/detekt/pull/5666#discussion_r1071330077))

While the above example shows the valid code as per the rule, I am also slightly modifying the example which make the code uncompliant with the rule as below
```kotlin
import kotlinx.coroutines.MainScope
import kotlinx.coroutines.delay
import kotlinx.coroutines.launch
inline fun <R> foo(crossinline block: suspend () -> R) = block()
suspend fun bar() {
    runCatching {
         foo {
             delay(1000L) // This is uncompliant code as even `block` is marked `crossinline` we are using `block` inside `foo` as normal lambda
        }
    }
}
```

BraisGabin: Couldn't we use the `suspend` modifier of `foo` to distingish between them? In the first example is `inline fun...` but in the second is `suspend inline fun`. That `suspend` is key here. For sure, someone could do:

```kotlin
suspend inline fun <R> foo(crossinline block: suspend () -> R) = MainScope().launch {
    block()
}
```

But we don't care. You are calling a `suspend` function inside a `runCatching`. We should flag it.

atulgpt: Hi, @BraisGabin I didn't notice the `suspend` modifier at that one. I though about it and found that the use of `crossinline` is valid and hence my proposal of `crossinline` is not valid. I will update the TCs around it after giving it one more thought.

atulgpt: Hi @BraisGabin, you were right. Having `suspend` in the fun was able to detekt the usage. I have added the same TC and grouped TCs in classes for better maintainability

BraisGabin (approved):
LGTM

atulgpt:
> @BraisGabin note: unresolved conversations are unresolved, hence the name ;)

Hi @TWiStErRob can you check once again? I have added the TC as well for your case and it is passing

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
TWiStErRob: I didn't see a production code change related to this, but I guess it was already doing this check correctly. Let's resolve this and start seeing usage in production.

TWiStErRob:
I think it's sufficiently covered with tests, so let's release it and see if false positives/negatives come up. I would add a note in the issue description something like:

```
* Note: This is a very complex topic, so there are ought to be
* false positive/false negative findings, if you come across one,
* please report in the [detekt issue tracker](https://github.com/detekt/detekt/issues).
```

I know this is the default expected behavior, but in the case of this rule, it's even more important to not just "suppress it in my case", but discuss usage patterns we didn't think of.

cortinico:
> I think it's sufficiently covered with tests, so let's release it and see if false positives/negatives come up.

Merging as we want this included in 1.23 👍

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
atulgpt: Hi @TWiStErRob yes, it was already working fine.

PR #5675, opened 2023-01-06
Aggregate coverage reports for new projects

3flex:


BraisGabin (approved):

BraisGabin:
Would it be possible to tell "depend on everything but..."? It seems like a better configuration because we will forget about this the next time we add a new module.

3flex:
There's probably a way - there usually is when it comes to Gradle, even if it's ugly...

Issue #5568, opened 2022-11-24
Do not apply `UseIfInsteadOfWhen` when variable captured


EdwarDDay:
## Expected Behavior of the rule

Currently the `UseIfInsteadOfWhen` rule complains, that the following code is not compliant:
```kotlin
when (val x = calculateSomething()) {
    is Success -> doSomething(x)
    else -> println(x)
}
```
This would then need to be changed to something like
```kotlin
val x = calculateSomething()
if (x is Success) {
    doSomething(x)
} else {
    println(x)
}
```
I would suggest to not apply the rule to `when` with a captured variable, because the changed version would expose the variable to a greater scope (and not only the conditional block).
 
This could also be added as a configuration for the rule, so that it can be generally enabled, but disabled for the named cases.

## Context

Jetbrains removed the `replace 'when' with 'if'` [quick fix from Intellij](https://youtrack.jetbrains.com/issue/KTIJ-15573/Intention-Replace-when-with-if-produces-wrong-code-if-expression-subject-is-a-variable-declaration), when the `when` expression uses a captured variable. So this makes this rule also harder to apply in these cases.

cortinico:
> This would then need to be changed to something like

That's an interesting case. I would argue that the second snippet is more readable though.
Something to raise to JetBrains as a language proposals for capturing variable inside if conditions maybe?

EdwarDDay:
I think for an `if` block it would be not so clear, for which block the variable would be captured (only the if block or additional `else` or `else if` blocks?).

cortinico:
> would be captured (only the if block or additional `else` or `else if` blocks?).

All the blocks in the same if hierarchy (`if`, all the `if else` and the `else`).
Back to your suggestion:

> I would argue that the second snippet is more readable though.

What do you think about this?

EdwarDDay:
I wouldn't see a big difference in the readability. I think this mainly depends on the developer to decide, what is more readable in this case.

BraisGabin:
I could argue that the first one reduces the scope of the variable and that's nice. So maybe adding that configuration is not a bad idea.

atulgpt:
Hi, @BraisGabin / @cortinico  I have raised the [PR](https://github.com/detekt/detekt/pull/5681) for this. This PR is inspired by JetBrains Ide work at [this PR](https://github.com/JetBrains/kotlin/pull/2921)

PR #5681, opened 2023-01-06
Add config for variable expression in when

atulgpt:
Fixes https://github.com/detekt/detekt/issues/5568


BraisGabin (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIfInsteadOfWhen.kt
```diff
@@ -34,9 +37,15 @@ class UseIfInsteadOfWhen(config: Config = Config.empty) : Rule(config) {
         Debt.FIVE_MINS
     )
 
+    @Configuration("ignores when variable declaration is used in subject")
```
cortinico, line 40:
```suggestion
    @Configuration("ignores when statements with a variable declaration used in the subject")
```

as `when` is ambigous here

cortinico (approved):

Issue #5493, opened 2022-11-01
`MaxLineLength` should ignore Markdown links by default


yevhenii-nadtochii:
## Expected Behavior of the rule
This issue is similar to #1070, but for Markdown links.

Consider the following example:

```kotlin
 * See: [Maven Publish Plugin | Publications](https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:publications)
 * https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:publicationspublicationspublicationspublicationspublications
```

The first line will be reported, while the second one not (thanks to #1087).

I would suggest checking if a line ends with a Markdown link as well, and don't report this case.

## Context
From [Kotlin documentation](https://kotlinlang.org/docs/kotlin-doc.html) (emphasis mine):

>> The language used to document Kotlin code (the equivalent of Java's Javadoc) is called KDoc. In its essence, KDoc combines Javadoc's syntax for block tags (extended to support Kotlin's specific constructs) and **Markdown for inline markup**.

PR #5683, opened 2023-01-07
Fix url and kotlin reference support in kdoc

atulgpt:
Fixes https://github.com/detekt/detekt/issues/5493

### DIFF
--- a/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/StringExtensions.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/StringExtensions.kt
@@ -10,13 +10,19 @@ fun String.lastArgumentMatchesUrl(): Boolean {
 }
 
 fun String.lastArgumentMatchesMarkdownUrlSyntax(): Boolean {
-    val urlNonCapturingRegex = "(?:[^ ]*)"
-    val markdownUrlTitleRegexStr = "\"[^\"\\\\\\n]*(?:\\\\.[^\"\\\\\\n]*)*\""
-    val regex = "\\[.*\\]\\($urlNonCapturingRegex(?: $markdownUrlTitleRegexStr)?\\)$".toRegex()
+    val urlNonCapturingRegex = """(?:(?!\s).)+"""
+    val markdownUrlTitleRegexStrWithBraces = """\([^\(\\\n]*(?:\\.[^\)\\\n]*)*\)"""
+    // Below regex works for both " and ' by using captured group \1
+    // It uses negative lookahead to use captured group \1 as to exclude it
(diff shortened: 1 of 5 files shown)


detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/StringExtensions.kt
```diff
@@ -8,3 +8,15 @@ fun String.lastArgumentMatchesUrl(): Boolean {
         URL(lastArgument).toURI()
     }.isSuccess
 }
+
+fun String.lastArgumentMatchesMarkdownUrlSyntax(): Boolean {
+    val urlNonCapturingRegex = "(?:[^ ]*)"
+    val markdownUrlTitleRegexStr = "\"[^\"\\\\\\n]*(?:\\\\.[^\"\\\\\\n]*)*\""
```
BraisGabin, line 14: could you use `"""` to avoid all those scaping `\`? They make the code really difficult to read. This comment applies to all the strings below too.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLengthSpec.kt
```diff
@@ -245,4 +246,171 @@ class MaxLineLengthSpec {
...
+            """.trimIndent()
+            val rule = MaxLineLength(
+                TestConfig(
+                    MAX_LINE_LENGTH to "60",
+                )
+            )
+
+            assertThat(rule.compileAndLint(code)).hasSize(3)
+        }
+    }
```
BraisGabin, line 415: Could be possible to simplify the tests? I know that these tests cover more code but they are difficult to understand.

For example, there are other test that check that all the comments are handled correctly. So we don't need to repeat the same for test on a class, function and property. I will just keep the ones from the class.

Other option is to add tests to the functions inside `detekt-psi-utils`. Those tests should be way easier to write and maintain. And then in these test you just confirm that you are using those with 2 simple tests.

BraisGabin (commented):
The functionality looks good but I added some comments to make the code easier to maintain.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLengthSpec.kt
atulgpt: Hi @BraisGabin I have moved the test inside `detekt-psi-utils` and kept the minimal test in the MaxLineLengthSpec.kt](https://github.com/detekt/detekt/pull/5683/files/b0626290cb6f355ec9bfb6c99f552d56c26a37e0#diff-e1bd0b72de1a0889c33ed5377e4781617f0ab330374ae2acd606965b21e05d96)

BraisGabin (approved):
These tests are way easy to understand. Thanks!!

PR #5698, opened 2023-01-15
Use integrated "test retry" from Gradle Enterprise Gradle Plugin

3flex:
Functionality from `org.gradle.test-retry` plugin was integrated directly into Gradle Enterprise Gradle Plugin in v3.12


BraisGabin (approved):
Let's see if this time they work...

Issue #5659, opened 2023-01-02
False positive in the SleepInsteadOfDelay rule


atulgpt:
## Expected Behavior
In the code example below
```kotlin
private fun bar(lambda: () -> Unit) {
    lambda()
}

@Suppress("unused", "MagicNumber", "InjectDispatcher")
private fun foo(): Deferred<Unit> {
    return CoroutineScope(Dispatchers.IO).async {
        bar {
            Thread.sleep(1000)
        }
    }
}
```

The above example should not report by the rule `SleepInsteadOfDelay` as the user can't call `delay` coz `lambda` is not suspending lambda

## Observed Behavior
The above example is reported by the rule `SleepInsteadOfDelay` as a false positive

## Steps to Reproduce
Failing TC

```kotlin
@Test
fun `should not report Thread sleep() called in non-suspending block`() {
    @Suppress("DeferredResultUnused")
    val code = """
        import kotlinx.coroutines.CoroutineScope
        import kotlinx.coroutines.Dispatchers
        import kotlinx.coroutines.async
        
        fun bar(lambda: () -> Unit) {
            lambda()
        }
        fun foo() {
            CoroutineScope(Dispatchers.IO).async {
                bar {
                    Thread.sleep(1000L)
                }
            }
        }
    """.trimIndent()
    assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
}
```

## Your Environment

* Version of detekt used: 1.22.0
* Version of Gradle used (if applicable): NA
* Operating System and version: MacBook Pro (16-inch, 2019) Version: 11.5.2 (20G95)

atulgpt:
Hi @BraisGabin do you know the optimal approach to solve this? I was working on the `SuspendFunInsideRunCatching` rule and find similar issue there as well

BraisGabin:
Go for it :)

atulgpt:
Hi, @BraisGabin created a PR for the fix. To fix this slightly changed the traversal logic which fortunately fixed more failing TCs

PR #5699, opened 2023-01-15
Check Thread.sleep for block expression

atulgpt:
Fixes https://github.com/detekt/detekt/issues/5659
Add check of canGoInside when lambda argument is suspend


BraisGabin (approved):
👏👏👏 Good one! 👏👏👏 Thanks :)

atulgpt:
Hi @BraisGabin, updated the PR(also reworked the traversal logic as in the previous logic some new added TCs were failing)

atulgpt:
Hi @BraisGabin , some intermittent error in CI can you retrigger?

PR #5700, opened 2023-01-15
BracesOnIfStatements with configurable parameters

### LINKED ISSUE #5133: Allow consistent brace formatting checks (OptionalWhenBraces triggered)
In many cases consistency is more important than minimalistic code. So I think it would be beneficial to add a flag (and enable by default) to treat curly braces as allowed (dare I say "mandatory"?). This would elevate all `when` cases to the same level of visibility and structure giving the code more consistency. I'm all for removing all braces, iff all cases in the `when` let me remove it, but as soon as there's one with complex logic, the others should be treated equally too.

## Context
Compare these two functions:
```kotlin
private fun OutgoingContent.asString(): String? =
	when (this) {
		is OutgoingContent.NoContent -> ""
		is OutgoingContent.WriteChannelContent -> {
			// Need to consume the asynchronicity to satisfy the contract.
			runBlocking {
				val channel = ByteChannel(true)
				this@asString.writeTo(channel)
				channel.tryReadText(Charsets.UTF_8)
			}
		}
		is TextContent ->
			this.text
		else -> {
			logger.error("Unknown response body content: ${this::class}")
			null
		}
	}
```

```kotlin
private fun OutgoingContent.asString(): String? =
	when (this) {
		is OutgoingContent.NoContent -> {
			""
		}

		is OutgoingContent.WriteChannelContent -> {
			// Need to consume the asynchronicity to satisfy the contract.
			runBlocking {
				val channel = ByteChannel(true)
				this@asString.writeTo(channel)
				channel.tryReadText(Charsets.UTF_8)
			}
		}

		is TextContent -> {
			this.text
		}

		else -> {
			logger.error("Unknown response body content: ${this::class}")
			null
		}
	}
```
and notice how in the first one `TextContent` handling is hidden away because the "braces are optional".

### LINKED ISSUE #5322: MandatoryBracesIfStatements: false negative on else
## Expected Behavior
I can't decide if this is a false positive (for `if`) or a false negative (for `else`). What I'm sure about is that it looks inconsistent.

image

## Steps to Reproduce
```kotlin
fun test(a: Boolean, b: Boolean, c: Boolean) {
	if (a) // flagged:
		if (b) {
			println("")
		} else {
			println("")
		}
	else // not flagged:
		if (c) {
			println("")
		} else {
			println("")
		}
}
```

## Context
I was trying to write a nice repro for another issue, when I noticed that this is happening.

## Investigation
I was suspecting that the control flow would not check the else, because there's a reporting in the `if`, but this code doesn't flag it either:
```kotlin
fun test(a: Boolean, b: Boolean, c: Boolean) {
	if (a) {
		if (b) {
			println("")
		} else {
			println("")
		}
	} else
		if (c) {
			println("")
		} else {
			println("")
		}
}
```

## Your Environment

VitalyVPinchuk:
[WIP]
To close #5133 one day.

As proposed by @TWiStErRob and @BraisGabin [here](https://github.com/detekt/detekt/issues/5133#issuecomment-1336267234) `BracesOnIfStatements` rule was added as the first iteration.

It also may fix #5322 as it does exactly the same except for configurable parameters (`always|never|consistent|necessary`).

Please comment.


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatementsSpec.kt
```diff
@@ -0,0 +1,353 @@
...
+            assertThat(findings).hasTextLocations(14 to 16, 34 to 38)
+        }
+    }
+
+    @Nested
+    inner class `multiline if statements with braces` {
+        val subject = createSubject("never", "always")
+
+        @Test
+        fun `does not report if statements with braces`() {
```
TWiStErRob, line 299: I think this rule might need more coverage to cover all possibilities in general.

For example for this @Nested, it could include tests for multiline `if-else` and `if-elseif` and `if-elseif-else`, etc. examples

TWiStErRob: I would probably create some example code blocks as constants, then in individual nested classes define all the config values, then in individual tests verify each code example block whether it passes or not. This should reduce the amount code, but increase the coverage a lot.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+
+    private fun isMultiStatement(expression: KtExpression): Boolean {
+        if (expression !is KtBlockExpression) return false
+
+        return expression
+            .firstChild
+            .siblings(forward = true, withItself = false)
+            .count { it is KtExpression } > 0
```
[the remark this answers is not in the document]
TWiStErRob: actually, looking wider, `isMultiStatement = expression is KtBlockExpression && expression.statements.size > 1`?

TWiStErRob:
Really nice one so far, left some initial thoughts while reading to understand what you did.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,136 @@
...
+        Debt.FIVE_MINS
+    )
+
+    @Configuration("single-line braces policy")
+    private val singleLine: String by config(BRACES_NEVER)
+
+    @Configuration("multi-line braces policy")
+    private val multiLine: String by config(BRACES_ALWAYS)
+
+    override fun visitIfExpression(expression: KtIfExpression) {
```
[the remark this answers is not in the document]
VitalyVPinchuk: I think it visits every single if-else statement so for `if (a) b else if (c) d` it's just called two times, first with `if (a) b else if (c) d` (we skip `if`/`when` after `else`) and then with `if (c) d`.
So testing multiple `if-else-if-else` cases might be excessive, am I right?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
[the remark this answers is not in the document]
VitalyVPinchuk: Indeed!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,136 @@
...
+
+        return expression
+            .firstChild
+            .siblings(forward = true, withItself = false)
+            .count { it is KtExpression } > 0
+    }
+
+    private fun isMultiline(element: PsiElement?): Boolean {
+        if (element == null) return false
+        return element.siblings().any { it.textContains('\n') }
```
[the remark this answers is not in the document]
VitalyVPinchuk: Correct

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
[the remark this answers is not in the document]
TWiStErRob: Oh, ok, that could be interesting then. Because we need to break down the problem to just looking at one part of it at a time, which could work with always/never, but not with consistent. I would say validate the whole if-elseif-elseif-else chain at the first if, which should be detectable by looking at the parent and be able to shortcut and not validate in the middle of it.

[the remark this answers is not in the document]
VitalyVPinchuk: I got your point. That makes sense.
But I viewed **if** only as 2-branch statement because there's no **elseif** in Kotlin, only simple self-sufficient **if-else**.
That's why if (a) b else if (c) { d } is consistent in this commit.
There could also be **when** instead of **if**. Things could get overcomplicated.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,136 @@
...
+ * || singleLine = 'never' multiLine = 'always'
+ * val x = if (condition) 5 else 4
+ * val x = if (condition) {
+ *     5
+ * } else {
+ *     4
+ * }
+ * </compliant>
+ */
+class BracesOnIfStatements(config: Config = Config.empty) : Rule(config) {
```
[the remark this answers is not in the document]
BraisGabin: Yes, we should.

config/detekt/detekt.yml
```diff
@@ -149,6 +149,10 @@ potential-bugs:
     active: true
 
 style:
```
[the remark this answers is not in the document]
BraisGabin: You are right, if we call all of them `BracesOn` they will be grouped. For me it is OK to keep them on `style`.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,136 @@
...
+        Severity.Style,
+        "Braces do not comply with the specified policy",
+        Debt.FIVE_MINS
+    )
+
+    @Configuration("single-line braces policy")
+    private val singleLine: String by config(BRACES_NEVER)
+
+    @Configuration("multi-line braces policy")
+    private val multiLine: String by config(BRACES_ALWAYS)
```
BraisGabin, line 55:
```suggestion
    private val multiLine: BracePolicy by config<String>(BRACES_ALWAYS) {
        when(it) {
            "never" -> BracePolicy.Never
            "consistent" -> BracePolicy.Consistent
            "always" -> BracePolicy.Always
            else -> error("Unknown value $it, the allowed values are: ...")
        }
    }
```

(I'm writing this from memory so maybe this is not the correct syntax but I know that something like this could be done

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,136 @@
...
+ * val i = 1
+ * if (i > 0)
+ *    println(i)
+ * else {
+ *    println(-i)
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * || singleLine = 'never' multiLine = 'always'
```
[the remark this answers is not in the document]
VitalyVPinchuk: I did

config/detekt/detekt.yml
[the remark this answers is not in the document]
VitalyVPinchuk: So let it be style.
Or we could create `braces` package inside style (like `optional`)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,136 @@
...
+            report(CodeSmell(issue, Entity.from(expression.ifKeyword), "Inconsistent braces"))
+        }
+    }
+
+    private fun isCompliant(expression: KtExpression, isMultiline: Boolean): Boolean = when {
+        isMultiStatement(expression) -> {
+            true
+        }
+
+        isMultiline && (multiLine == BRACES_ALWAYS) || !isMultiline && (singleLine == BRACES_ALWAYS) -> {
```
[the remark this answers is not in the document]
VitalyVPinchuk: Tried to simplify but still wasn't able to fit `always|never|consistent` in one function.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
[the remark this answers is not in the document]
VitalyVPinchuk: So there's total rework with @TWiStErRob ideas and guidance.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatementsSpec.kt
VitalyVPinchuk: Tried to do it

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,178 @@
...
+                }
+            }
+        }
+    }
+
+    private fun report(violator: KtExpression, policy: BracePolicy) {
+        val iff = violator.parent.parent as KtIfExpression
+        val reported = when {
+            iff.then === violator -> iff.ifKeyword
+            iff.`else` === violator && isMultiStatement(violator) -> iff.ifKeyword
```
TWiStErRob, line 140: This feels strange, would this work instead (I think this expresses more `else if`, and it localizes the report more targeted on where the problem is)
```suggestion
            iff.`else` === violator && violator is KtIfExpression -> violator.ifKeyword
```
Test case (not executed, just thought-validated):
`if (true) { println() } else if (false) println()` + always or consistent config. Original highlight on `if` at character 1, and expected highlight would be the second `if`.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,178 @@
...
+                when (arg) {
+                    BRACES_NEVER -> Never
+                    BRACES_ALWAYS -> Always
+                    BRACES_CONSISTENT -> Consistent
+                    else -> error(
+                        "Unknown value $arg, allowed values are: $BRACES_NEVER|$BRACES_ALWAYS|$BRACES_CONSISTENT"
+                    )
+                }
+        }
+    }
```
TWiStErRob, line 177: What do you think about?
```suggestion
    enum class BracePolicy(private val config: String) {
        Always("always"),
        Never("never"),
        Consistent("consistent");

        companion object {
            fun getValue(arg: String): BracePolicy =
                values().singleOrNull { it.config == arg }
                   ?: error("Unknown value ${arg}, allowed values are: ${values().joinToString("|")}")
                }
        }
    }
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
TWiStErRob: I would say the expected highlights are (based on the discussion we had yesterday about context):
```kotlin
// always:
if (true) { println() } else if (false) println()
                             ^^

// always:
if (a) b else { println(); println() }
^^

// never:
if (a) b else { println(); println() }
         ^^^^
```

I think at this point in the logic, if you need to check multi-statement or KtBlock, that means you're making an assumption on the structure or policy.

> It would be the second if, because we compare the first element against all others.

This confirms my assumption theory, I think we were talking about different policies, I meant for most of them, and you're talking about consistent.

This actually highlights a problem, and thanks for that. For consistent it makes sense to highlight the first if. That way we're not opinionated on flagging a missing/extra braces. The problem is that we would need to traverse up to find it. As the first option in the `when` we could add something like:
```kotlin
policy == Consistent ->
  // Walk up to the first `if` in an arbitrary long else-if chain.
  generateSequence(violator) { it.parent.parent as? KtIfExpression }.last()
```
and for other cases in the when we just highlight whatever the violator's branch was.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * This rule detects `if` statements which do not comply with the specified rules.
+ * Keeping braces consistent would improve readability and avoid possible errors.
+ *
+ * SingleLine if-statement has no '\n':
+ * if (a) b else c
+ * MultiLine if-statement has at least one '\n':
+ * if (a) b
+ * else c
```
TWiStErRob, line 24:
```suggestion
 * SingleLine if-statement has no '\n':
 * ```
 * if (a) b else c
 * ```
 * MultiLine if-statement has at least one '\n':
 * ```
 * if (a) b
 * else c
 * ```
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,178 @@
...
+import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import org.jetbrains.kotlin.psi.KtBlockExpression
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtIfExpression
+
+/**
+ * This rule detects `if` statements which do not comply with the specified rules.
+ * Keeping braces consistent would improve readability and avoid possible errors.
+ *
```
TWiStErRob, line 19: add a definition here for each policy option?
```
*
* The available options are:
*  * `always`: forces braces on all `if` and `else` branches in the whole codebase
*  * `never`: forces no braces on any `if` and `else` branches in the whole codebase  
*    (except where necessary for multi-statement branches.)
*  * `consistent`: ensures that braces are consistent within each `if`-`else if`-`else` chain.
*    If there's a brace on one of the branches, all branches should have it.
*
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,178 @@
...
+                    report(violator, policy)
+                }
+            }
+
+            BracePolicy.Never -> {
+                if (!list.all { !hasBraces(it) }) {
+                    val violator = list.first { hasBraces(it) }
+                    report(violator, policy)
+                }
+            }
```
TWiStErRob, line 124: I think we need to account for the multi-statement branches in case of `Never`, so this is probably wrong considering the definition I just wrote above:

> `never`: forces no braces on any `if` and `else` branches in the whole codebase  
  (except where necessary for multi-statement branches.)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,178 @@
...
+        }
+    }
+
+    private fun report(violator: KtExpression, policy: BracePolicy) {
+        val iff = violator.parent.parent as KtIfExpression
+        val reported = when {
+            iff.then === violator -> iff.ifKeyword
+            iff.`else` === violator && isMultiStatement(violator) -> iff.ifKeyword
+            iff.`else` === violator -> iff.elseKeyword
+            else -> error("Violating element ($violator) is not part of this if: $iff")
```
TWiStErRob, line 142: When the tests were failing on this (because of bugs in `walk`), there was a problem with their toString, it showed something like:
> Violating element (IF) is not part of this if: IF

I think we should change it so that it prints the whole code text of each of those variables for easier diagnostics.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,178 @@
...
+        }
+        report(CodeSmell(issue, Entity.from(reported ?: violator), issue.description + ": $policy"))
```
TWiStErRob, line 144: Let's be a bit more dev/user-friendly?

```suggestion
        val message = when(policy) {
            Always -> "Missing braces on this branch, add them."
            Never -> "Extra braces exist on this branch, remove them."
            Consistent -> "Inconsistent braces, make sure all branches either have or don't have braces."
        }
        report(CodeSmell(issue, Entity.from(reported ?: violator), message))
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/MandatoryBracesIfStatements.kt
```diff
@@ -28,6 +28,10 @@ import org.jetbrains.kotlin.psi.psiUtil.siblings
  * val x = if (condition) 5 else 4
  * </compliant>
  */
+@Deprecated(
+    "Use 'BracesOnIfStatements()' instead",
+    ReplaceWith("BracesOnIfStatements(config)")
```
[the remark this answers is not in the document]
TWiStErRob: I would go one more step
```suggestion
    "Use 'BracesOnIfStatements' with 'always' configuration instead"
```
since that's the replacement for **Mandatory**BracesIfStatements

config/detekt/detekt.yml
```diff
@@ -149,6 +149,10 @@ potential-bugs:
...
+  BracesOnIfStatements:
+    active: true
+    singleLine: 'consistent'
+    multiLine: 'consistent'
```
TWiStErRob, line 155: Are there no violations in the whole codebase? :) 👏

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatementsSpec.kt
```diff
@@ -0,0 +1,301 @@
...
+                "singleLine" to singleLine,
+                "multiLine" to multiLine
+            )
+        )
+        return BracesOnIfStatements(config)
+    }
+
+    @Nested
+    inner class `braces consistent` {
+        val subject = createSubject("consistent", "consistent")
```
TWiStErRob, line 177: Create one or two specific tests which verify that the right `singleLine` / `multiLine` config is used for the right type (e.g. singleLine: never multiLine: always). This will test the `policy` method directly without accessing it. Right now if I flip the condition in `policy()` all the tests will still pass.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatementsSpec.kt
```diff
@@ -0,0 +1,301 @@
...
+            3 -> if (true) {
+                    println()
+                } else if (true)
+                    println()
+                else {
+                    println()
+                }
+        }
+    }
+"""
```
TWiStErRob, line 161:
```suggestion
""".trimIndent()
```
on all blocks?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatementsSpec.kt
```diff
@@ -0,0 +1,301 @@
...
+            val findings = subject.compileAndLint(BLOCK_WITHOUT_BRACES_PASS)
+
+            assertThat(findings).isEmpty()
+        }
+
+        @Test
+        fun `reports incorrect block without braces`() {
+            val findings = subject.compileAndLint(BLOCK_WITHOUT_BRACES_FAIL)
+
+            assertThat(findings).hasSize(8)
```
TWiStErRob, line 288: there are 10 `if`s in this code block, why only 8 fail?

TWiStErRob: (Please similarly review the others to make sure the right expectations are met, don't be afraid to have failing tests, the code might not be working fully yet.)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatementsSpec.kt
```diff
@@ -0,0 +1,301 @@
...
+            assertThat(findings).hasTextLocations(
```
TWiStErRob, line 289:
```suggestion
            assertThat(findings).hasSize(8)
            assertThat(findings).hasTextLocations(
```
I think the `hasSize` is redundant in these verificaitons, `hasTextLocations` does a deep equal on the arrays, so if the count differs it'll fail, and also show the missing/extra/wrong locations in the error message.

TWiStErRob (commented):
Just a few more finishing touches.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
VitalyVPinchuk: right

config/detekt/detekt.yml
VitalyVPinchuk: No, there aren't.
Do you think that's because of a bug? ))

TWiStErRob: Suspicious, I would add an intentional violation in this rule temporarily so we can see it show up.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatementsSpec.kt
VitalyVPinchuk: Some of them were correct.
Fixed.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
VitalyVPinchuk: As discussed added 'necessary' option, which skips multi-statements.

VitalyVPinchuk:
How a rule should be deprecated?
There's an import which I removed and used fully qualified name instead, but generateDocumentation crashes.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StyleGuideProvider.kt
```diff
@@ -76,7 +75,9 @@ class StyleGuideProvider : DefaultRuleSetProvider {
             UnnecessaryLet(config),
             MayBeConst(config),
             PreferToOverPairSyntax(config),
-            MandatoryBracesIfStatements(config),
+            @Suppress("DEPRECATION")
+            io.gitlab.arturbosch.detekt.rules.style.optional.MandatoryBracesIfStatements(config),
```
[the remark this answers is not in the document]
TWiStErRob: Huh, sorry, forgot to send this comment.

Related by @VitalyVPinchuk:
> How a rule should be deprecated?
There's an import which I removed and used fully qualified name instead, but generateDocumentation crashes.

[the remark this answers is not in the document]
TWiStErRob: is it possible that the class doesn't need to be deprecated, just manually adding a line to deprecation.properties file?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatementsSpec.kt
```diff
@@ -0,0 +1,410 @@
...
+
+import io.gitlab.arturbosch.detekt.test.TestConfig
+import io.gitlab.arturbosch.detekt.test.assertThat
+import io.gitlab.arturbosch.detekt.test.compileAndLint
+import org.intellij.lang.annotations.Language
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+@Language("kotlin")
+private val BLOCK_WITHOUT_BRACES_PASS = """
```
atulgpt, line 11: What I find useful in these cases is to use `@ParameterizedTest`. You can then also configure each test case's display name. That will be more maintainable and you will writing small unit tests which will validate only one scenario

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,222 @@
...
+    )
+
+    @Configuration("single-line braces policy")
+    private val singleLine: BracePolicy by config("never") { BracePolicy.getValue(it) }
+
+    @Configuration("multi-line braces policy")
+    private val multiLine: BracePolicy by config("always") { BracePolicy.getValue(it) }
+
+    override fun visitIfExpression(expression: KtIfExpression) {
+        super.visitIfExpression(expression)
```
atulgpt, line 139: Here `super` call can be avoided with something like the below. Here we are using the fact that traversal happen top-down manner and if we have reached the root if block then we can avoid the traversal to go and below level
```
if (expression.parent.parent !is KtIfExpression) {
    val branches: List<KtExpression> = walk(expression)
    validate(branches, policy(expression))
}
```

Feel free to add more TCs around this logic to validate this logic 😅

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatementsSpec.kt
TWiStErRob: Yeah, my bad, I think I wasn't clean in https://github.com/detekt/detekt/pull/5700#discussion_r1071178966

@VitalyVPinchuk we let's revisit this together I have an idea.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
TWiStErRob: Let's add some tests for this if not covered yet.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StyleGuideProvider.kt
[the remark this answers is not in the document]
BraisGabin: Sorry for the late answer.

That's other option, right. We could just remove it and add it in `:detekt-generator`. We already have some rules there.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatementsSpec.kt
```diff
@@ -0,0 +1,1917 @@
...
+            "if"(1),
+            "if"(2),
+            "else"(1),
+            "if"(3),
+            "else"(2),
+            "else"(3),
+            "if"(4),
+            "else"(4),
+        ),
+    )
```
TWiStErRob, line 1803: this test fails, can you please check why

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,226 @@
...
+
+        val parent = expression.parent.parent
+        val parentCondition = (parent as? KtIfExpression)?.condition
+        if ((parent is KtIfExpression) && (parentCondition !== expression)) return
```
TWiStErRob, line 143: This seems equivalent and simpler:
```suggestion
        // Ignore `then` and `else` branches, they're handled by the initial `if`'s visit.
        // But let us process conditions, because they might be `if` themselves.
        if (parent is KtIfExpression && parent.condition !== expression) return
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnIfStatements.kt
```diff
@@ -0,0 +1,226 @@
...
+        val branches: List<KtExpression> = walk(expression)
+        validate(branches, policy(expression))
+    }
+
+    private fun walk(expression: KtExpression): List<KtExpression> {
+        val list = mutableListOf<KtExpression>()
+        var current: KtExpression? = expression
+        while (current is KtIfExpression) {
+            current.then?.let { list.add(it) }
+            if (current.then is KtIfExpression) { list.addAll(walk(current.then as KtIfExpression)) }
```
TWiStErRob, line 154: This feels strange, it looks like we're processing `x` in the context of the outer if-else, but they're not the same "if":
```
if (c) {
   if (...) ... // x
} else {
   ...
}
```
but I might be reading it wrong.

TWiStErRob (approved):
LGTM, many iterations and refactors, I hope it's bulletproof by now.

TWiStErRob:
@BraisGabin can you please have a look at the deprecation stuff, I'm unsure if that's correct or not.

BraisGabin (approved):
LGTM. Thanks for this awesome contribution!

Issue #5697, opened 2023-01-14
Rule for encouraging MultilineRawString


atulgpt:
## Expected Behavior of the rule
Encourage usage of Multiline raw string in case lots of escape has been added to the string(which can be removed by using multiline raw string). The rule can also encourage users when there is `\n` in the string usage. 

There should be an option to configure the count of `maxEscapedCharacterCount` and `maxNewLineCharacterCount` so that rule can allow strings which have fewer escapedCharcter/newline character counts. Optionally flags like`ignoreEscapedCharacters` and `ignoreNewLineCharacter` can also be there to suppress the detection of particular characters

See below non compliant and compliant code snippets
Non compliant
```
val windowJson = "{\n" +
        "  \"window\": {\n" +
        "    \"title\": \"Sample Konfabulator Widget\",\n" +
        "    \"name\": \"main_window\",\n" +
        "    \"width\": 500,\n" +
        "    \"height\": 500\n" +
        "  }\n" +
        "}"
```
Compliant
```
val windowJson = """
        {
          "window": {
            "title": "Sample Konfabulator Widget",
            "name": "main_window",
            "width": 500,
            "height": 500
          }
        }
    """.trimIndent()
```

Non compliant
```
val markdownUrlTitleRegexStr = "\"[^\"\\\\\\n]*(?:\\\\.[^\"\\\\\\n]*)*\""
```
Compliant
```
val markdownUrlTitleRegexStr = """"[^"\\\n]*(?:\\.[^"\\\n]*)*""""
```

## Context
This rule implies the string representation of JSON, regexes or SQL queries and will promote the usage of multiline string format for new Kotlin users

[Kotlin doc](https://kotlinlang.org/docs/coding-conventions.html#strings) do talk about this `Prefer multiline strings to embedding \n escape sequences into regular string literals.`

BraisGabin:
I really like this rule idea! This issue happens a lot in the wild.

We should also be aware that the usage of raw strings make it difficult to write a `$`, a tab three consecutive `"`. But I don't think we should care about this edge cases for the first implementation of this rule. I would just start simple and then expand the rule.

The name of the rule could be `StringShouldBeRawString`.

atulgpt:
Hi @BraisGabin yup I would like to take a look into this. Even in my projects(even by myself) has seen the same mistake

PR #5705, opened 2023-01-17
Add StringShouldBeRawString.kt rule

atulgpt:
Fixes https://github.com/detekt/detekt/issues/5697

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawString.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawString.kt
@@ -19,10 +19,11 @@ import org.jetbrains.kotlin.psi.psiUtil.getParentOfTypesAndPredicate
 import org.jetbrains.kotlin.psi2ir.deparenthesize
 
 /**
- * This rule reports when string can be converted to Kotlin raw string.
- *
- * Usage of raw string is preferred as that avoids the need of escaping strings escape characters like `\n`, `\t`, `"`.
- * Raw string also allow us to represent multiline string without the need of `\n`
+ * This rule reports when the string can be converted to Kotlin raw string.
+ * Usage of a raw string is preferred as that avoids the need for escaping strings escape characters like \n, \t, ".
+ * Raw string also allows us to represent multiline string without the need of \n.
+ * Also, see [Kotlin coding convention](https://kotlinlang.org/docs/coding-conventions.html#strings)  for
+ * recommendation on using multiline strings
  *
  * <noncompliant>
  * val windowJson = "{\n" +
@@ -90,11 +91,12 @@ class StringShouldBeRawString(config: Config) : Rule(config) {
             expressionParent !is KtBinaryExpression ||
             (rootElement != null && expression.isPivotElementInTheTree(rootElement))
         ) {
-            val shouldReport = rootElement.buildStringExcludingRawString().flatMap { stringTemplateExpressionText ->
-                REGEX_FOR_ESCAPE_CHARS.findAll(stringTemplateExpressionText).filter {
-                    it.value !in ignoredCharacters
-                }
-            }.take(maxEscapedCharacterCount + 1).toList().size > maxEscapedCharacterCount
+            val shouldReport =
+                rootElement.getStringSequenceExcludingRawString().flatMap { stringTemplateExpressionText ->
+                    REGEX_FOR_ESCAPE_CHARS.findAll(stringTemplateExpressionText).filter {
+                        it.value !in ignoredCharacters
+                    }
+                }.take(maxEscapedCharacterCount + 1).toList().size > maxEscapedCharacterCount
             if (shouldReport) {
                 report(
                     CodeSmell(
@@ -118,7 +120,7 @@ class StringShouldBeRawString(config: Config) : Rule(config) {
         }
     }
 
-    private fun KtElement?.buildStringExcludingRawString(): Sequence<String> {
+    private fun KtElement?.getStringSequenceExcludingRawString(): Sequence<String> {
         this ?: return sequence { yield("") }
 
         fun KtElement?.getStringSequence(): Sequence<KtStringTemplateExpression> = sequence {
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawString.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawString.kt
@@ -81,23 +81,19 @@ class StringShouldBeRawString(config: Config) : Rule(config) {
 
     override fun visitStringTemplateExpression(expression: KtStringTemplateExpression) {
         super.visitStringTemplateExpression(expression)
-        if (maxEscapedCharacterCount == Int.MAX_VALUE) {
-            return
-        }
-        val maxEscapedCharacterCount = this.maxEscapedCharacterCount.coerceAtLeast(0)
         val expressionParent = expression.getParentExpressionAfterParenthesis()
         val rootElement = expression.getRootExpression()
         if (
             expressionParent !is KtBinaryExpression ||
             (rootElement != null && expression.isPivotElementInTheTree(rootElement))
         ) {
-            val shouldReport =
-                rootElement.getStringSequenceExcludingRawString().flatMap { stringTemplateExpressionText ->
-                    REGEX_FOR_ESCAPE_CHARS.findAll(stringTemplateExpressionText).filter {
-                        it.value !in ignoredCharacters
-                    }
-                }.take(maxEscapedCharacterCount + 1).toList().size > maxEscapedCharacterCount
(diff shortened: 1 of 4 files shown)


3flex:
Couple of notes:
- rule name should reference "multiline" in the name, not just "raw string"
- I think there should be an option to ignore single lines with \n. Sometimes it's simpler to have a string like "line 1\nline 2", than to convert to a multiline string (and remember we have other rules that relate to formatting multiline strings and enforcing use of trimIndent and trimMargin). All of a sudden there's a bunch of extra things to get "right". Or there could be an option for single lines for max number of new line escape characters which triggers the violation

BraisGabin:
> rule name should reference "multiline" in the name, not just "raw string" 

The second example on the issue doesn't talk about multi-line. A regex is a good example of a single line raw string that is better than a normal string.

3flex:
Good point. I only looked at this very quickly, but the issue being fixed is "Rule for encouraging MultilineRawString". The proposed rule is a bit different.

Agree that this proposal makes sense as is, so ignore my comment :)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawStringSpec.kt
```diff
@@ -0,0 +1,618 @@
...
+        val code = """
+            fun test() {
+                val size1 = "This rule is awesome \n".length
+            }
+        """.trimIndent()
+        val subject = StringShouldBeRawString(TestConfig(mapOf(MAX_ESCAPED_CHARACTER_COUNT to Int.MIN_VALUE)))
+        subject.compileAndLint(code)
+        assertThat(subject.findings).hasSize(1)
+    }
+
```
BraisGabin, line 419: The things that can be tested are infinite. I would try to keep the tests "as short as possible" because then we need to maintain them too. These tests are valid but they provide little value so I would just remove them.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawString.kt
```diff
@@ -0,0 +1,160 @@
...
+        get() {
+            val leftChild = (this as? KtBinaryExpression)?.right?.deparenthesize() ?: return this
+            return leftChild.rightMostElementOfRightSubtree
+        }
+
+    override fun visitStringTemplateExpression(expression: KtStringTemplateExpression) {
+        super.visitStringTemplateExpression(expression)
+        if (maxEscapedCharacterCount == Int.MAX_VALUE) {
+            return
+        }
```
BraisGabin, line 86: If the user doesn't want this rule s/he should disable it. We don't need to have two different ways to disable it.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawString.kt
```diff
@@ -0,0 +1,160 @@
...
+        if (
+            expressionParent !is KtBinaryExpression ||
+            (rootElement != null && expression.isPivotElementInTheTree(rootElement))
+        ) {
+            val shouldReport =
+                rootElement.getStringSequenceExcludingRawString().flatMap { stringTemplateExpressionText ->
+                    REGEX_FOR_ESCAPE_CHARS.findAll(stringTemplateExpressionText).filter {
+                        it.value !in ignoredCharacters
+                    }
+                }.take(maxEscapedCharacterCount + 1).toList().size > maxEscapedCharacterCount
```
BraisGabin, line 99: It's strange that you need to materialize the sequence to int to get the size. Isn't there a `count`?

Or even better. Instead of `.take(maxEscapedCharacterCount + 1)` you can `.drop(maxEscapedCharacterCount)` and then just ask if it is empty.

An I would suggest to use a chain here:
```kotlin
val shouldReport = rootElement.getStringSequenceExcludingRawString()
  .flatMap {
    ...
  }
  .filter { ... }
  .drop(...)
  .isNotEmpty()
```

I feel it is easier to read the different operator that you apply.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawString.kt
```diff
@@ -0,0 +1,160 @@
...
+            leftMostElementOfLeftSubtree
+        } else {
+            rootElement.rightMostElementOfRightSubtree
+        }
+    }
+
+    private fun KtElement?.getStringSequenceExcludingRawString(): Sequence<String> {
+        this ?: return sequence { yield("") }
+
+        fun KtElement?.getStringSequence(): Sequence<KtStringTemplateExpression> = sequence {
```
BraisGabin, line 126: You have 3 different `this` on this scope. For that reason you need to use the `@` that makes the code more difficult to read. Different ideas here:

- This functions doesn't need to be inside the `class` you can move them outside (but in the same file)
- The inner function doesn't need to be a extension function.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawString.kt
```diff
@@ -0,0 +1,160 @@
...
+    ): Boolean {
+        val leftMostElementOfLeftSubtree = rootElement.leftMostElementOfLeftSubtree
+        return this == if (leftMostElementOfLeftSubtree is KtStringTemplateExpression) {
+            leftMostElementOfLeftSubtree
+        } else {
+            rootElement.rightMostElementOfRightSubtree
+        }
+    }
+
+    private fun KtElement?.getStringSequenceExcludingRawString(): Sequence<String> {
```
BraisGabin, line 123: In general is not a good idea to have extension functions over nullable values. They make the things strange for the reader. On the caller site you can to somehting like `(item?.extension() ?: sequenceOf(""))`.

BraisGabin (commented):
It looks good :)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawString.kt
atulgpt: I was trying to do something like above only but couldn't fine the correct operators. I will update the code here

We can't use `count` as that will result in creation of complete list.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawString.kt
atulgpt: Sure. But currently if we set Int.MAX_VALUE in the `detekt.yml` the working of rule breaks. BTW @BraisGabin is there any sanitisation on user input through `detekt.yml`?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawStringSpec.kt
atulgpt: Done

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawString.kt
atulgpt: Sadly we can't use `isNotEmpty` in sequnece. Although I found more better way. I used the code like below
```
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawString.kt
BraisGabin: Int.MAX_VALUE doesn't seems like a sane value. We shouldn't care about that. If they would like the max value of a long it wouldn't work either 🤷.

BraisGabin: And there is little sanitation on that regard as far as I know.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/StringShouldBeRawString.kt
BraisGabin: Good point. Other option is to capture `this` on a variable to avoid the `@`. But this is good to me.

atulgpt:
Hi @BraisGabin is this PR good for merge? Or some changes are required? cc @cortinico @TWiStErRob for their inputs as well

BraisGabin:
This is good to go! Sorry for the delay :)

BraisGabin (approved):

PR #5706, opened 2023-01-18
IgnoredReturnValue: fix false negative when annotation is on the package

### LINKED ISSUE #5686: IgnoredReturnValue false negative when annotation is on the package.
## Expected Behavior

```kotlin
import com.google.common.truth.Truth.assertThat

    // ... in some testcase ...
    assertThat(fooBar == null) // Note: doesn't actually assert; constructs a BooleanSubject but fails to use it.
```

Should report an error like

    The call assertThat is returning a value that is ignored. [IgnoredReturnValue]

## Observed Behavior

No error is reported.

## Context

Error Prone, and possibly other frameworks, allow the `@CheckReturnValue` to be [put on the package](https://errorprone.info/bugpattern/CheckReturnValue#fn:package-info:~:text=%40CheckReturnValue%20may%20be%20applied%20to%20a%20class%20or%20package%202%20to%20indicate%20that%20all%20methods%20in%20that%20class%20or%20package%20must%20have%20their%20return%20values%20used.). Some libraries make use of this capability, for instance [Truth](https://github.com/google/truth/blob/b997ffb1e6122f3117af8dd9585d0cdc4193f679/core/src/main/java/com/google/common/truth/package-info.java#L31).

This issue seems conceptually similar to https://github.com/detekt/detekt/issues/3975 , but at the package-level rather than the class.

### Workaround

I was able to get reports for this form of error after configuring the rule to explicitly know about the Truth package using `returnValueTypes`.

```diff
  IgnoredReturnValue:
    active: true
    restrictToConfig: true
    returnValueAnnotations:
      - "*.CheckResult"
      - "*.CheckReturnValue"
    ignoreReturnValueAnnotations:
      - "*.CanIgnoreReturnValue"
    returnValueTypes:
      - "kotlin.sequences.Sequence"
      - "kotlinx.coroutines.flow.*Flow"
      - "java.util.stream.*Stream"
+     - "com.google.common.truth.*"
```

t-kameyama:
Fixes #5686

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -439,9 +439,12 @@ potential-bugs:
     active: true
     restrictToConfig: true
     returnValueAnnotations:
+      - 'CheckResult'
       - '*.CheckResult'
+      - 'CheckReturnValue'
       - '*.CheckReturnValue'
     ignoreReturnValueAnnotations:
+      - 'CanIgnoreReturnValue'
       - '*.CanIgnoreReturnValue'
     returnValueTypes:
       - 'kotlin.sequences.Sequence'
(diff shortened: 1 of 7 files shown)


BraisGabin (approved):
Thanks!!

PR #5710, opened 2023-01-19
Fix typo

JIY0UNG:
It is written as 'GitHub' elsewhere. But only here, it is written as 'Github'.
So modified it to 'GitHub' for consistency.


BraisGabin (approved):
Thanks!

Issue #5629, opened 2022-12-19
`CanBeNonNullable` false positive 


eygraber:
This is happening in 1.22.0 and before as well https://github.com/detekt/detekt/issues/4677

I copied it from my project and sanitized it, so I'm not sure what the minimum repro is here.

The error is:

```
CanBeNonNullable - [The nullable parameter 'bar' can be made non-nullable.] at Playground.kt:107:33
```

And the repro is:

```
import kotlinx.coroutines.CoroutineScope
import kotlinx.coroutines.Dispatchers
import kotlinx.coroutines.channels.BufferOverflow
import kotlinx.coroutines.delay
import kotlinx.coroutines.flow.Flow
import kotlinx.coroutines.flow.MutableSharedFlow
import kotlinx.coroutines.launch
import kotlinx.coroutines.withContext
import kotlin.random.Random

private suspend inline fun log(msg: () -> String) {
  delay(1_000)
  println(msg())
}

const val SHARED_FLOW_DEFAULT_BUFFER = 64

fun <T> bufferedSharedFlow(
  capacity: Int = SHARED_FLOW_DEFAULT_BUFFER,
  onBufferOverflow: BufferOverflow = BufferOverflow.SUSPEND
) =
  MutableSharedFlow<T>(extraBufferCapacity = capacity, onBufferOverflow = onBufferOverflow)

sealed class BarType {
  object One : BarType()
  object Two : BarType()
}

class Bar {
  val type
    get() = when(System.currentTimeMillis()) {
      0L -> BarType.One
      else -> BarType.Two
    }
}

internal interface BarHelper {
  suspend fun findNextBar(targetBar: Bar? = null): Bar?
}

internal class BarHelperImpl : BarHelper {
  override suspend fun findNextBar(targetBar: Bar?): Bar? {
    val bars = listOf(Bar(), Bar())
    val nextBar =
      bars
        .find { bar ->
          bar.type is BarType.One && when(val type = bar.type) {
            is BarType.One -> withContext(Dispatchers.IO) {
              type == BarType.One
            }

            else -> true
          }
        }

    val isRandomTrue = Random.nextBoolean()

    return when {
      isRandomTrue -> nextBar

      else -> targetBar ?: nextBar
    }
  }
}

internal interface Foo {
  suspend fun foo(bar: Bar? = null): Bar?
}

internal interface Baz {
  fun baz(): Flow<Bar?>
}

internal class BazImpl : Baz {
[...]

PR #5714, opened 2023-01-20
Fix false positive for `CanBeNonNullable` rule

atulgpt:
Added failing TCs
Fixes https://github.com/detekt/detekt/issues/5629

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt
@@ -395,7 +395,7 @@ class CanBeNonNullable(config: Config = Config.empty) : Rule(config) {
 
         private fun List<KtWhenCondition>.evaluateSubjectWhenExpression(
             expression: KtWhenExpression,
-            subjectDescriptor: List<CallableDescriptor>,
+            subjectDescriptors: List<CallableDescriptor>,
         ) {
             var isNonNullChecked = false
             var isNullChecked = false
@@ -422,7 +422,7 @@ class CanBeNonNullable(config: Config = Config.empty) : Rule(config) {
                     isNullChecked = true
                 }
             }
-            subjectDescriptor.forEach { callableDescriptor ->
+            subjectDescriptors.forEach { callableDescriptor ->
                 nullableParams[callableDescriptor]?.let {
                     if (isNullChecked) it.isNullChecked = true
                     if (isNonNullChecked) it.isNonNullChecked = true
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt
@@ -17,6 +17,8 @@ import io.gitlab.arturbosch.detekt.rules.isOverride
 import org.jetbrains.kotlin.descriptors.CallableDescriptor
 import org.jetbrains.kotlin.descriptors.DeclarationDescriptor
 import org.jetbrains.kotlin.descriptors.PropertyDescriptor
+import org.jetbrains.kotlin.descriptors.ValueParameterDescriptor
+import org.jetbrains.kotlin.js.translate.callTranslator.getReturnType
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.psi.KtBinaryExpression
@@ -231,13 +233,22 @@ class CanBeNonNullable(config: Config = Config.empty) : Rule(config) {
         }
 
         override fun visitWhenExpression(expression: KtWhenExpression) {
-            val subjectDescriptor = expression.subjectExpression
-                ?.let { it as? KtNameReferenceExpression }
-                ?.getResolvedCall(bindingContext)
-                ?.resultingDescriptor
+            val nullCheckedDescriptor =
+                expression.subjectExpression?.collectDescendantsOfType<KtNameReferenceExpression>().orEmpty()
+                    .mapNotNull {
+                        val callDescriptor =
+                            it.getResolvedCall(bindingContext)
+                        val valueParameterDescriptor =
+                            callDescriptor?.resultingDescriptor as? ValueParameterDescriptor
+                        if (valueParameterDescriptor != null && callDescriptor.getReturnType().isNullable()) {
+                            valueParameterDescriptor
+                        } else {
+                            null
+                        }
+                    }
             val whenConditions = expression.entries.flatMap { it.conditions.asList() }
-            if (subjectDescriptor != null) {
-                whenConditions.evaluateSubjectWhenExpression(expression, subjectDescriptor)
+            if (nullCheckedDescriptor.isNotEmpty()) {
+                whenConditions.evaluateSubjectWhenExpression(expression, nullCheckedDescriptor)
             } else {
                 whenConditions.forEach { whenCondition ->
                     if (whenCondition is KtWhenConditionWithExpression) {
@@ -342,7 +353,9 @@ class CanBeNonNullable(config: Config = Config.empty) : Rule(config) {
                 } else {
                     { nullableParam -> nullableParam.isNonNullChecked = true }
                 }
-                nonNullChecks.forEach { nullableParams[it]?.let(nullableParamCallback) }
+                nonNullChecks.forEach {
+                    nullableParams[it]?.let(nullableParamCallback)
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


BraisGabin:
Are these tests related with #5629? The cases are a bit difficult to follow. And I'm not sure if all of them are related with the same issue.

Would it be possible to split them in different PRs? Or all of them target the same flaw?

atulgpt:
Hi, @BraisGabin not all follow the same flaw. Some can be grouped and split into little PRs. Is there any provision to add disabled test in the PR? Or should I completely remove them from here and add in the other PRs?

BraisGabin:
I would prefer to split them up.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
```diff
@@ -720,6 +793,193 @@ class CanBeNonNullableSpec(val env: KotlinCoreEnvironment) {
...
+                }
+
+                @Test
+                fun `does report when parameter is not checked for nullability with inc operator`() {
+                    val code = """
+                        class A {
+                            operator fun not() = this
+                        }
+                        
+
```
BraisGabin, line 876:
```suggestion
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
```diff
@@ -720,6 +793,193 @@ class CanBeNonNullableSpec(val env: KotlinCoreEnvironment) {
...
+                                (b != null) && !b != a -> {
+                                    println("doesn't handle null")
+                                }
+                                (c > 10 || (b != null && !b != a)) -> {
+                                    println("doesn't handle null")
+                                }
+                            }
+                        }
+                    """.trimIndent()
+                    assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
```
BraisGabin, line 892: I dont't get what this test does exactly.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
```diff
@@ -774,6 +1034,36 @@ class CanBeNonNullableSpec(val env: KotlinCoreEnvironment) {
                     """.trimIndent()
                     assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
                 }
+
+                // Failing
+                @Test
+                fun `does not report when else handles null case`() {
```
BraisGabin, line 1040: Isn't this test exactly the same as `does not report when else expression variable is smart casted to null`?

BraisGabin (commented):
Hi @atulgpt which are the next steps for this PR? Is there somerhing blocking you here? Can I help you some how? I left some comments on the tests but they seem fine.

I must say that a function that only does something when a parameter is null seems like a function that should be flagged too. But probably not by this rule but by other one. But this is offtopic for this PR.

atulgpt:
> Hi @atulgpt which are the next steps for this PR? Is there somerhing blocking you here? Can I help you some how? I left some comments on the tests but they seem fine.
> 
> I must say that a function that only does something when a parameter is null seems like a function that should be flagged too. But probably not by this rule but by other one. But this is offtopic for this PR.

Hi @BraisGabin good vibes of Bali is stopping me from completing this PR 😂. I will complete back after I come back 🙂

BraisGabin:
> Hi @BraisGabin good vibes of Bali is stopping me from completing this PR 😂. I will complete back after I come back 🙂

Enjoy!!!

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
atulgpt: The name of the TC was wrong. Actually, it should report as all the branches of `when` handles the non-null case. Here we are checking different types of boolean conditions. I have renamed the TC to `does report when parameter is not checked for nullability in boolean conditions`

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt
```diff
@@ -231,13 +233,22 @@ class CanBeNonNullable(config: Config = Config.empty) : Rule(config) {
         }
 
         override fun visitWhenExpression(expression: KtWhenExpression) {
-            val subjectDescriptor = expression.subjectExpression
-                ?.let { it as? KtNameReferenceExpression }
-                ?.getResolvedCall(bindingContext)
-                ?.resultingDescriptor
+            val nullCheckedDescriptor =
```
atulgpt, line 236: Hi, @BraisGabin this fixes the current issue but this solution is not a robust solution and will give a false negative when Kotlin `when` subject expression contains complex logic.

Consider the below example
```
fun foo(a: Int?, b: Int?) {
    when (a?.also {
        println(b)
    }?.plus(4)) {
        null -> println("We have handled a is null as well")
    }
}
```

for the above example, the rule should ideally give one warning but it gives no warning.

Actually, I am kind of stuck as to how I can determine if `b`(or any variable present in `when` expression) contributes to the nullability of the whole `when` subject expression or not. In this example, `b` is not contributing to the nullability of the below `when` subject expression
```
a?.also {
        println(b)
    }?.plus(4)
```.

But this code handles the issue mentioned in the PR. Either we can completely handle all the case or fix the rule case by case

Comment (BraisGabin) [reply detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt:236]: This rule had some false positives already. Is this one of those or is this a new one?

If this improves the rule and doesn't introduce any regression I would merge it. It is not perfect? OK, we can work on that in next PRs but our users can benefit from this improvements meanwhile.

Also, complex expressions inside when's is not a very common case so we shouldn't expand far too much time there.

Comment (atulgpt) [reply detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt:236]: `This rule had some false positives already. Is this one of those or is this a new one?` These are existing ones but initially they were false negative with more extent now some cases has been fix and some has been converted to false postive

Comment (BraisGabin) [reply detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt:236]: 🤔 The false positive are a problem... We shouldn't add of those. Any way to get rid of them?

Comment (atulgpt) [reply detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt:236]: Edited my comment. Please see the edited comment

File: detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt [hunk 5]
```diff
@@ -231,13 +233,22 @@ class CanBeNonNullable(config: Config = Config.empty) : Rule(config) {
...
+                        val callDescriptor =
+                            it.getResolvedCall(bindingContext)
+                        val valueParameterDescriptor =
+                            callDescriptor?.resultingDescriptor as? ValueParameterDescriptor
+                        if (valueParameterDescriptor != null && callDescriptor.getReturnType().isNullable()) {
+                            valueParameterDescriptor
+                        } else {
+                            null
+                        }
+                    }
```
BraisGabin, line 248:
```suggestion
            val nullCheckedDescriptor = expression.subjectExpression?
                .collectDescendantsOfType<KtNameReferenceExpression>()
                .orEmpty()
                .mapNotNull { it.getResolvedCall(bindingContext) }
                .filter { callDescriptor -> callDescriptor.getReturnType().isNullable() }
                .mapNotNull { callDescriptor -> callDescriptor.resultingDescriptor as? ValueParameterDescriptor }
```

What do you think? Is it easier to read?

BraisGabin (approved):
I left some nit comments but LGTM

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullable.kt
BraisGabin: Then we can merge it. Later we could improve it even more.

BraisGabin:
It seems that my code didn't compile 😅. The Github's suggestions feature is great but working without an IDE is difficult.

atulgpt:
> It seems that my code didn't compile 😅. The Github's suggestions feature is great but working without an IDE is difficult.

Haha true 😅 I have fixed it to make it compliable. Now CI should pass

cortinico (approved):

PR #5718, opened 2023-01-21
Remove ignoreOverridden config from BooleanPropertyNaming etc.

### LINKED ISSUE #5600: Deprecate ignoreOverridden in name rules
The rules:

- BooleanPropertyNaming
- ConstructorParameterNaming
- FunctionNaming
- VariableNaming

Have the ignoreOverridden configuration flag but it have really little sense. There's nothing you can do about those names. So it should not be a configuration. It should be by default in all those rules. 

So my proposal is to deprecate those configurations to simplify a bit our configuration that is long enough. We should also include the ones that will be included with #5590

There are also two other rules were I don't have it so clear:

- `FunctionParameterNaming`: in this case you can rename them but that's a really bad idea. So deprecate it and always ignore deprecate seems like a sane decission. 
- ~`MemberNameEqualsClassName`: I don't fully understand how this rule could help my code. But I imagine that keep it here have sense. Because in this one you can act on the issue. You can change the name of the class.~

drawers:
Fixes #5600 where it was agreed to remove `ignoreOverridden` from 5 rules:
1. `BooleanPropertyNaming`
2. `ConstructorParameterNaming`
3. `FunctionNaming`
4. `VariableNaming`
5. `FunctionParameterNaming`


schalkms (approved):
Looks good! Thanks for tidying up!

schalkms:
@BraisGabin do you want to review this, since you brought up this improvement?

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNaming.kt
```diff
@@ -36,9 +36,6 @@ class BooleanPropertyNaming(config: Config = Config.empty) : Rule(config) {
     @Configuration("naming pattern")
     private val allowedPattern: Regex by config("^(is|has|are)", String::toRegex)
 
-    @Configuration("ignores properties that have the override modifier")
-    private val ignoreOverridden: Boolean by config(true)
```
BraisGabin, line 40: This is a "breaking change". It is not in the "normal way" but we treat the yaml configuration as part of our public API.

Instead of removing the property, just keep it and add it this annotation `@Deprecated("This configuration is ignored")` (I'm not inspired with my descriptions feel free to improve it).

Other option is to add the deprecation at`io.gitlab.arturbosch.detekt.generator.printer.writeMigratedRules`. This way we keep the code clean.

What bothe options does is to keep `detekt-core/src/main/resources/deprecation.properties` up to date with old the deprecations.

BraisGabin (commented):
Thanks for the PR! And sorry for the late review. If you could fix what I point out in my other comment we can merge this.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNaming.kt
drawers: Thanks, that's a clever way to manage deprecation. I have assumed that between the two ways of doing things you have stated, the first way is cleaner because the code is kept in sync with the `deprecation.properties.` 

I just need to rebase this on the latest `main` before we review again. I will ping later today when it is ready :pray:

drawers: Thanks @BraisGabin I think this is ready for you to have another look at now

BraisGabin (approved):
Thanks! LGTM

PR #5722, opened 2023-01-23
Split UnusedPrivateMember

### LINKED ISSUE #3418: Split `UnusedPrivateMember`
This rule has gone way beyond reporting unused member, and covers properties, parameters, and member functions.
I would suggest a split at least for parameters vs (properties + functions).

## Context
This is triggered by reviewing #3417

drawers:
Fixes #3418 where it was decided to split `UnusedPrivateMember` in a non-breaking way so that after the split:

* The original rule is enabled by default
* After the split, the rules are still enabled by default

I was not sure from the issue discussion whether to split into 2 or 3. I went for 3 because:
1. It allows individual configuration of each rule. For instance, a user might want to ignore unused parameters that begin with `_` but not unused properties that begin with `_`. 
2. It was suggested by the design of the visitors in the original version of `UnusedPrivateMember`
3. Detecting unused private functions benefits from type resolution whereas detecting private properties does not 
4. Splitting helps with managing the many test cases

| Diagram |
| --- |
| Before |
| class diagram of before |
| After |
| image |

It's quite hard to review - the splitting of the `UnusedPrivateMember` functionality is mechanical but moving the test cases around is harder and requires more care. I guess I would recommend reviewing the individual commits.


detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -700,11 +700,17 @@ style:
     active: false
   UnusedImports:
     active: false
+  UnusedParameter:
+    active: true
+    allowedNames: '(_|ignored|expected|serialVersionUID)'
```
drawers, line 705: `serialVersionUID` is not likely to be a parameter name we want to ignore but I wanted to make this a non-breaking change. 

I would suggest a follow up issue for all of the breaking changes that will further the clean up here (including things like renaming `UnusedPrivateMember` to `UnusedPrivateFunction`)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedParameter.kt
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * <compliant>
+ * fun foo(used: String) {
+ *   println(used)
+ * }
+ * </compliant>
+ */
+@ActiveByDefault(since = "1.23.0")
+class UnusedParameter(config: Config = Config.empty) : Rule(config) {
+    override val defaultRuleIdAliases: Set<String> = setOf("UNUSED_VARIABLE", "UNUSED_PARAMETER", "unused")
```
drawers, line 49: I kept "UNUSED_VARIABLE" as an alias even though this rule now only handles params. This was in order to preserve existing behavior.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMember.kt
```diff
@@ -125,7 +103,8 @@ private class UnusedFunctionVisitor(
                                 KtTokens.MINUS,
                                 KtTokens.MUL,
                                 KtTokens.DIV,
-                                KtTokens.PERC -> operatorValue?.let { functionReferences["$it="] }.orEmpty()
+                                KtTokens.PERC,
+                                -> operatorValue?.let { functionReferences["$it="] }.orEmpty()
```
drawers, line 107: Change from running `ktlint --format`

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateParameterSpec.kt
```diff
@@ -165,66 +166,6 @@ class UnusedPrivateParameterSpec(val env: KotlinCoreEnvironment) {
...
-                    }
-
-                    // this is actually unused, but clashes with the other constructor
-                    constructor(used: Any)
-                }
-            """.trimIndent()
-            assertThat(subject.lint(code)).hasSize(1)
-        }
-    }
-
```
drawers, line 227: An unfortunate consequence of this refactor is that unused primary and secondary constructor params are now detected by `UnusedPrivateProperty` (they used to be detected by `UnusedPrivateMember` which handled everything together.

I wanted to keep this refactor focused on mechanical splitting and not introduce logic changes that would be even harder to review. I would suggest follow up on this in a tracking issue I talk about in the other comments.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateParameterSpec.kt
```diff
@@ -10,7 +11,7 @@ import org.junit.jupiter.api.Test
 
 @KotlinCoreEnvironmentTest
 class UnusedPrivateParameterSpec(val env: KotlinCoreEnvironment) {
```
drawers, line 13: Can I rename this to `UnusedParameterSpec` to match the new rule? It seems to detect unused params for both public and private.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateParameterSpec.kt
chao2zhang: It might be reasonable given that a property can be defined from constructor's parameters, by adding `val` or `var`. If you can describe the behavior in the KDoc documentation it would be great

drawers: Thank you @chao2zhang - that seems like good suggestion. I added to the KDoc in 5897a31ea

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateParameterSpec.kt
drawers: Thank you, done as b563ece4e

detekt-core/src/main/resources/default-detekt-config.yml
chao2zhang: I guess we don't usually have parameters with names like `ignored`, `expected`, and `serialVersionUID`.

chao2zhang (approved):

schalkms (commented):
I haven't reviewed in detail, but I think after the restructuring the code looks way cleaner than before.
After the merge conflict is resolved, I think we should merge this asap.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
```diff
@@ -383,7 +383,35 @@ class UnusedPrivateMemberSpec(val env: KotlinCoreEnvironment) {
     }
 
     @Nested
-    inner class `operators` {
+    inner class `main methods` {
```
drawers, line 386: Hey @schalkms  thank you for performing the merge :pray:   but I think this `main methods` now needs to be in `UnusedParameterSpec.kt` rather than in `UnusedPrivateMemberSpec.kt`. That was how it was in the original patch I submitted since those test cases now cover `UnusedParameter.kt`

drawers: @schalkms would you prefer me to add a commit to address this?

schalkms: Good catch. Btw, sorry for messing around with the merge.

schalkms: @drawers you were a few seconds faster. 😉

drawers: @schalkms I added a commit ddd19eb to address this. If I take Chao's last commit before the merge as authoritative (1e4bcb8) then I do:

```shell
git diff --name-only 1e4bcb8 | grep Unused
```

I get:
```shell
detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
```

Then if I do:
```shell
git diff 1e4bcb8 detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
```

I get:
```shell
diff --git a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
index 2dbf7673d3..237df91e75 100644
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateMemberSpec.kt
@@ -383,7 +383,7 @@ class UnusedPrivateMemberSpec(val env: KotlinCoreEnvironment) {
     }
 
     @Nested
-    inner class `operators` {
+    inner class Operators {
 
         @Test
         fun `does not report used plus operator - #1354`() {
```

So we end up with only the change from renaming `inner class operators` to `inner class Operators` and nothing else. Which I think is what we want.

schalkms: > _inner class `operators` {_
> to
> inner class Operators {

Yes 🙏

schalkms: We could also rollback the merge and do a rebase. That might be easier to grasp.

drawers: @schalkms yes if you would prefer me to rebase then I don't mind doing it just let me know :+1: 
(sorry about leaving the branch to go stale, I didn't anticipate another PR touching that test)

schalkms: @drawers if we are good to go after the merge, we are fine, no need to do another rebase then, IMHO.

drawers: @schalkms looks okay to me now!

PR #5731, opened 2023-01-27
Fix testPluginKotlinc task

3flex:
The task does not currently run successfully and is also currently incompatible with the configuration cache. With this PR the task will run successfully without error, tested locally on Windows 11.

The task is not currently run in CI, but probably should be...


3flex:
This was merged with #5732

PR #5754, opened 2023-02-04
Use stdlib functions for file & path operations

3flex:
I set out to remove these imports and use replacements from stdlib:
* `java.io.File`
* `java.nio.file.Files`
* `java.nio.file.Paths`

I was mostly successful, held back by the detekt-gradle-plugin being on API version 1.4, and a couple of stdlib functions which still require an opt in. I chose not to use the opt in functions until they're stable, hopefully in Kotlin 1.9 if not before.

I recommend reviewing by individual commit as I've broken up the changes into logical chunks.

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Junk.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Junk.kt
@@ -1,10 +1,3 @@
 package io.gitlab.arturbosch.detekt.core
 
-import java.nio.file.Files
-import java.nio.file.Path
-
-fun Path.exists(): Boolean = Files.exists(this)
-fun Path.isFile(): Boolean = Files.isRegularFile(this)
-fun Path.isDirectory(): Boolean = Files.isDirectory(this)
-
 val NL: String = System.lineSeparator()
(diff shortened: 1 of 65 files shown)


Goooler (approved):
And this one
https://github.com/detekt/detekt/blob/80bae4bdc522f567453cb476f60bce794681c7a4/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ArgumentConverters.kt#L15-L18

3flex:
@Goooler they're already covered?

https://github.com/detekt/detekt/pull/5754/files#diff-1d7a1cd6085118b76e5689c6c2bdd234c831a4bc566dfbb946975b90c1b2585e

Goooler:
You are right, just see it.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Junk.kt
```diff
@@ -1,10 +1,3 @@
 package io.gitlab.arturbosch.detekt.core
 
-import java.nio.file.Files
-import java.nio.file.Path
-
-fun Path.exists(): Boolean = Files.exists(this)
-fun Path.isFile(): Boolean = Files.isRegularFile(this)
-fun Path.isDirectory(): Boolean = Files.isDirectory(this)
```
BraisGabin, line 8: <3

BraisGabin (approved):

BraisGabin:
Great work!

PR #5757, opened 2023-02-04
Add NixOS installation method to doc

schalkms:


website/versioned_docs/version-1.22.0/gettingstarted/cli.mdx
```diff
@@ -37,6 +37,12 @@ unzip detekt-cli-[detekt_version].zip
 ./detekt-cli-[detekt_version]/bin/detekt-cli --help
 ```
 
+### NixOS
+
+```sh
+nix-shell -p detekt -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/f80ac848e3d6f0c12c52758c0f25c10c97ca3b62.tar.gz --command 'detekt --version' [detekt_version]
```
3flex, line 43: I'm not familiar with Nix - does the hash in the filename have to be updated with every detekt release? If so, is there a way to remove that from this command, because having to update this for every release may be a bit of a hassle.

schalkms: I just saw on Slack in our `#detekt` channel that @mdr added detekt as a nix package.
@mdr do you know whether the hash has to be updated with each and every detekt release?

mdr: > @mdr do you know whether the hash has to be updated with each and every detekt release?

(With the caveat I'm a Nix beginner!) Yes - if you specify a specific revision as above then you'd always get the same version of Detekt, so you'd have to update to get a more recent version of Detekt. You can just reference the `detekt` package unqualified, e.g. with `nix-shell -p detekt` - and then you'll get the Detekt version associated with the Nix channel you're using. If you're using the stable channel, that might be a few months out of date (and right now, `detekt` is too new to be in a stable Nix channel).

schalkms: Okay, then I'll change the script in the docs to be more generic, like this.

```sh
nixpkgs=https://github.com/NixOS/nixpkgs/archive/[hash].tar.gz
```

3flex: I guess at some point when it's in stable channel we can simplify the command. But this is good in the meantime.

Where do people get the hash though? Is there a link we can include in the instructions?

mdr: You can grab a commit hash from the [Nix status page](https://status.nixos.org/) (for the `nixpkgs-unstable` channel). It does feel kinda awkward - I don't know if there's a way you can directly reference the latest `nixpkgs-unstable` channel directly in the command.

mdr: If you have added the unstable channel via `nix-channel`, you can run: `nix-shell -I nixpkgs=channel:nixpkgs-unstable -p detekt`

3flex (approved):

cortinico:
I've backported it here https://github.com/detekt/detekt/pull/5776

PR #5763, opened 2023-02-06
Docs: Add missing `File.absolutePath` call in Kotlin DSL sample

YoshiRulz:
Not sure if this is also necessary in Groovy as I've never used it.


3flex:
Good catch, thanks!

3flex (approved):

PR #5766, opened 2023-02-07
Run Kotlin compiler plugin CLI test as part of standard build

3flex:
This will add a little time to the CI build times, but will reduce the risk of the task failing in future.

Follow up to #5731/#5732


TWiStErRob (approved):

schalkms (approved):

schalkms:
Good tradeoff IMHO

Issue #5768, opened 2023-02-07
Detekt Build: missing task dependency between sourcesJar and generateDocumentation


TWiStErRob:
## Expected Behavior
Clean build output and correct build every time.

## Observed Behavior
Most likely:

<details><summary>Execution optimizations have been disabled for task ':detekt-core:sourcesJar' to ensure correctness</summary>

```
> Task :detekt-core:sourcesJar
Execution optimizations have been disabled for task ':detekt-core:sourcesJar' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: 'P:\projects\contrib\github-detekt\detekt-core\src\main\resources'. Reason: Task ':detekt-core:sourcesJar' uses this output of task ':detekt-generator:generateDocumentation' without declari
ng an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.6/userguide/validation_problems.html#implicit_dependency for more de
tails about this problem.

> Task :detekt-generator:generateDocumentation
Picked up JAVA_TOOL_OPTIONS: -Xms32M -Xmx256M -XX:+UseG1GC -D-XX:+PrintCommandLineFlags
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\complexity.md
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\coroutines.md
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\comments.md
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\empty-blocks.md
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\potential-bugs.md
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\exceptions.md
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\libraries.md
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\naming.md
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\performance.md
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\ruleauthors.md
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\style.md
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\libraries.md
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\ruleauthors.md
Wrote: P:\projects\contrib\github-detekt\website\docs\rules\formatting.md
[...]

PR #5773, opened 2023-02-09
Fix generateDocumentation implicit dependencies

jprinet:
The `:detekt-rules-libraries:processResources` and `:detekt-rules-authors:processResources` tasks have an implicit dependency on `:detekt-generator:generateDocumentation` task, the files `detekt-rules-libraries/src/main/resources/config/config.yml` and `detekt-rules-ruleauthors/src/main/resources/config/config.yml` being outputs of `:detekt-generator:generateDocumentation`.

This has a small impact on performances (see [here](https://ge.detekt.dev/s/ltywkmui6n6ka/timeline?cacheability=validation-failure#6q6tf2ndv5xfi) an example on `detekt-rules-authors:processResources`)

This PR fixes this by [sharing the artifacts through a Gradle configuration](https://docs.gradle.org/current/userguide/cross_project_publications.html#sec:simple-sharing-artifacts-between-projects).

Please note that I had to cut some task dependencies in `detekt-generator` build script to avoid a cycle in the task graph (`assemble` => `build` => `shadowJar` => `generateDocumentation` => `assemble`). Please review those carefully as I am not familiar enough with the codebase to measure the impacts.

### DIFF
--- a/detekt-generator/build.gradle.kts
+++ b/detekt-generator/build.gradle.kts
@@ -93,11 +93,22 @@ val generatedRulesConfig: Configuration by configurations.creating {
     isCanBeResolved = false
 }
 
+val generatedCoreConfig: Configuration by configurations.creating {
+    isCanBeConsumed = true
+    isCanBeResolved = false
+}
+
 artifacts {
     add(generatedLibrariesConfig.name, file(librariesConfigFile)) {
         builtBy(generateDocumentation)
     }
     add(generatedRulesConfig.name, file(ruleauthorsConfigFile)) {
+        builtBy(generateDocumentation)
+    }
+    add(generatedCoreConfig.name, file(defaultConfigFile)) {
+        builtBy(generateDocumentation)
+    }
+    add(generatedCoreConfig.name, file(deprecationFile)) {
         builtBy(generateDocumentation)
     }
 }
--- a/detekt-generator/build.gradle.kts
+++ b/detekt-generator/build.gradle.kts
@@ -3,8 +3,6 @@ plugins {
     id("module")
 }
 
-tasks.build { finalizedBy(tasks.shadowJar) }
-
 dependencies {
     implementation(projects.detektParser)
     implementation(projects.detektApi)
@@ -37,7 +35,6 @@ val ruleModules = rootProject.subprojects
 
 val generateDocumentation by tasks.registering(JavaExec::class) {
     dependsOn(
-        tasks.assemble,
         tasks.shadowJar,
         ":detekt-api:dokkaHtml",
         ":detekt-rules-libraries:sourcesJar",
@@ -84,6 +81,25 @@ val generateDocumentation by tasks.registering(JavaExec::class) {
         "--cli-options",
         cliOptionsFile,
     )
+}
+
+val generateDocumentationConfigurationLibraries: Configuration by configurations.creating {
+    isCanBeConsumed = true
+    isCanBeResolved = false
+}
+
+val generateDocumentationConfigurationRules: Configuration by configurations.creating {
+    isCanBeConsumed = true
+    isCanBeResolved = false
+}
+
+artifacts {
+    add(generateDocumentationConfigurationLibraries.name, file(librariesConfigFile)) {
+        builtBy(generateDocumentation)
+    }
+    add(generateDocumentationConfigurationRules.name, file(ruleauthorsConfigFile)) {
+        builtBy(generateDocumentation)
+    }
 }
 
 val verifyGeneratorOutput by tasks.registering(Exec::class) {
--- a/detekt-generator/build.gradle.kts
+++ b/detekt-generator/build.gradle.kts
@@ -83,21 +83,21 @@ val generateDocumentation by tasks.registering(JavaExec::class) {
     )
 }
 
-val generateDocumentationConfigurationLibraries: Configuration by configurations.creating {
+val generatedLibrariesConfig: Configuration by configurations.creating {
     isCanBeConsumed = true
     isCanBeResolved = false
 }
 
-val generateDocumentationConfigurationRules: Configuration by configurations.creating {
+val generatedRulesConfig: Configuration by configurations.creating {
     isCanBeConsumed = true
     isCanBeResolved = false
 }
 
 artifacts {
-    add(generateDocumentationConfigurationLibraries.name, file(librariesConfigFile)) {
+    add(generatedLibrariesConfig.name, file(librariesConfigFile)) {
         builtBy(generateDocumentation)
     }
-    add(generateDocumentationConfigurationRules.name, file(ruleauthorsConfigFile)) {
+    add(generatedRulesConfig.name, file(ruleauthorsConfigFile)) {
         builtBy(generateDocumentation)
     }
 }
--- a/detekt-generator/build.gradle.kts
+++ b/detekt-generator/build.gradle.kts
@@ -88,7 +88,7 @@ val generatedLibrariesConfig: Configuration by configurations.creating {
     isCanBeResolved = false
 }
 
-val generatedRulesConfig: Configuration by configurations.creating {
+val generatedRuleauthorsConfig: Configuration by configurations.creating {
     isCanBeConsumed = true
     isCanBeResolved = false
 }
@@ -102,7 +102,7 @@ artifacts {
     add(generatedLibrariesConfig.name, file(librariesConfigFile)) {
         builtBy(generateDocumentation)
     }
-    add(generatedRulesConfig.name, file(ruleauthorsConfigFile)) {
+    add(generatedRuleauthorsConfig.name, file(ruleauthorsConfigFile)) {
         builtBy(generateDocumentation)
     }
     add(generatedCoreConfig.name, file(defaultConfigFile)) {
--- a/detekt-rules-libraries/build.gradle.kts
+++ b/detekt-rules-libraries/build.gradle.kts
@@ -17,7 +17,7 @@ dependencies {
     )
 }
 
-tasks.withType<ProcessResources>().configureEach {
+tasks.named<ProcessResources>("processResources").configure {
     inputs.files(generatedLibrariesConfig)
         .withPropertyName(generatedLibrariesConfig.name)
         .withPathSensitivity(PathSensitivity.RELATIVE)
--- a/detekt-rules-libraries/build.gradle.kts
+++ b/detekt-rules-libraries/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
     id("module")
 }
 
-val generateDocumentationConfiguration: Configuration by configurations.creating {
+val generatedLibrariesConfig: Configuration by configurations.creating {
     isCanBeConsumed = false
     isCanBeResolved = true
 }
@@ -12,19 +12,13 @@ dependencies {
     testImplementation(projects.detektTest)
     testImplementation(libs.assertj)
 
-    generateDocumentationConfiguration(
-        project(
-            mapOf(
-                "path" to ":detekt-generator",
-                "configuration" to "generateDocumentationConfigurationLibraries"
-            )
-        )
+    generatedLibrariesConfig(
+        project(":detekt-generator", "generatedLibrariesConfig")
     )
 }
 
 tasks.withType<ProcessResources>().configureEach {
-    val generateDocumentationFiles: FileCollection = generateDocumentationConfiguration
-    inputs.files(generateDocumentationFiles)
-        .withPropertyName("generateDocumentationConfiguration")
+    inputs.files(generatedLibrariesConfig)
+        .withPropertyName(generatedLibrariesConfig.name)
         .withPathSensitivity(PathSensitivity.RELATIVE)
 }
--- a/detekt-rules-libraries/build.gradle.kts
+++ b/detekt-rules-libraries/build.gradle.kts
@@ -2,23 +2,14 @@ plugins {
     id("module")
 }
 
-val generatedConfig: Configuration by configurations.creating {
-    isCanBeConsumed = false
-    isCanBeResolved = true
-}
-
 dependencies {
     compileOnly(projects.detektApi)
     testImplementation(projects.detektTest)
     testImplementation(libs.assertj)
-
-    generatedConfig(projects.detektGenerator) {
-        targetConfiguration = "generatedLibrariesConfig"
-    }
 }
 
-tasks.named("processResources").configure {
-    inputs.files(generatedConfig)
-        .withPropertyName(generatedConfig.name)
-        .withPathSensitivity(PathSensitivity.RELATIVE)
-}
+consumeGeneratedConfig(
+    fromProject = projects.detektGenerator,
+    fromConfiguration = "generatedLibrariesConfig",
+    forTask = "processResources"
+)
--- a/detekt-rules-libraries/build.gradle.kts
+++ b/detekt-rules-libraries/build.gradle.kts
@@ -13,7 +13,7 @@ dependencies {
     testImplementation(libs.assertj)
 
     generatedLibrariesConfig(
-        project(":detekt-generator", "generatedLibrariesConfig")
+        project(projects.detektGenerator.path, "generatedLibrariesConfig")
     )
 }
 
--- a/detekt-rules-libraries/build.gradle.kts
+++ b/detekt-rules-libraries/build.gradle.kts
@@ -12,7 +12,9 @@ dependencies {
     testImplementation(projects.detektTest)
     testImplementation(libs.assertj)
 
-    generatedConfig(project(projects.detektGenerator.path, "generatedLibrariesConfig"))
+    generatedConfig(projects.detektGenerator) {
+        targetConfiguration = "generatedLibrariesConfig"
+    }
 }
 
 tasks.named("processResources").configure {
--- a/detekt-rules-libraries/build.gradle.kts
+++ b/detekt-rules-libraries/build.gradle.kts
@@ -2,8 +2,29 @@ plugins {
     id("module")
 }
 
+val generateDocumentationConfiguration: Configuration by configurations.creating {
+    isCanBeConsumed = false
+    isCanBeResolved = true
+}
+
 dependencies {
     compileOnly(projects.detektApi)
     testImplementation(projects.detektTest)
     testImplementation(libs.assertj)
+
+    generateDocumentationConfiguration(
+        project(
+            mapOf(
+                "path" to ":detekt-generator",
+                "configuration" to "generateDocumentationConfigurationLibraries"
+            )
+        )
+    )
+}
+
+tasks.withType<ProcessResources>().configureEach {
+    val generateDocumentationFiles: FileCollection = generateDocumentationConfiguration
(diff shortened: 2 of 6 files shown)


3flex:
Thanks! This should fix https://github.com/detekt/detekt/issues/5768. I (or someone) will review when I get some time.

detekt-generator/build.gradle.kts
```diff
@@ -86,6 +83,25 @@ val generateDocumentation by tasks.registering(JavaExec::class) {
...
+}
+
+val generateDocumentationConfigurationRules: Configuration by configurations.creating {
+    isCanBeConsumed = true
+    isCanBeResolved = false
+}
+
+artifacts {
+    add(generateDocumentationConfigurationLibraries.name, file(librariesConfigFile)) {
+        builtBy(generateDocumentation)
```
TWiStErRob, line 98: @3flex Probably out of the scope of this PR, but realistically maybe these files should be generated by separate tasks? Because they're not documentation.

detekt-generator/build.gradle.kts
```diff
@@ -86,6 +83,25 @@ val generateDocumentation by tasks.registering(JavaExec::class) {
...
+    }
+    add(generateDocumentationConfigurationRules.name, file(ruleauthorsConfigFile)) {
+        builtBy(generateDocumentation)
+    }
+}
```
TWiStErRob, line 103: The path of these files are outside this module, so by default `src/main/resources` input of `processResources` of those modules will pick it up. Which means exposing and adding the configuration dependency only exists to make the tasks work. But I guess this is the original problem.

I wonder if it would make sense to relocate those files into the generator module, and pull the artifacts via configuration from the others. In this case the explicit dependency would make more sense (be more clear) and also there wouldn't be overlap between an input folder (src) and an output file (that should be under `build`, really). Although they're version controlled, and in that case their location that makes sense.

I noticed that `gradlew :detekt-generator:clean` doesn't actually delete these files, while `gradlew :detekt-generator:cleanGenerateDocumentation` does, and also breaks the next build. Anyway, I'm digressing a bit too much.

I think because of the source-controlled files, it makes sense to leave it this way. I would like @3flex to read this and ack it, and then I think this conversation can be resolved, because I convinced myself it's good as it is :)

detekt-generator/build.gradle.kts
```diff
@@ -86,6 +83,25 @@ val generateDocumentation by tasks.registering(JavaExec::class) {
     )
 }
 
+val generateDocumentationConfigurationLibraries: Configuration by configurations.creating {
```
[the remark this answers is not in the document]
TWiStErRob: I see why you used `generateDocumentation` as a prefix (task name), but it's a bit hard to grasp at glance what this actually is, because it's actually the generated configuration file for a specific published maven artifact (detekt ruleset jar), not doc.

Here are two proposals, I would pick the first, because it's a bit more consistent naming with only a prefix, and a varying suffix, while in the second the changing part is in the middle.
```suggestion
val generatedConfigForLibraries: Configuration by configurations.creating {
```
```suggestion
val generatedLibrariesConfig: Configuration by configurations.creating {
```

Also note that the fact that this is builtBy `generateDocumentation` is an implementation detail that shouldn't leak to other modules.

detekt-rules-ruleauthors/build.gradle.kts
```diff
@@ -2,12 +2,33 @@ plugins {
     id("module")
 }
 
+val generateDocumentationConfiguration: Configuration by configurations.creating {
```
TWiStErRob, line 5: Would there be a way to share this logic between the two modules?

For example: `fun Project.consumeGeneratedConfig(configurationName: String)` which does these 3 things (create config, add dep, configure resources task). This can be created in as `build-logic/src/main/kotlin/consumeGeneratedConfig.kt` (no package directive)

Then all the changes in this file can be replaced with
```gradle
consumeGeneratedConfig("generatedConfigForRuleauthors")
```

Before you make this change, let's see if @3flex gives a thumbs up.

detekt-generator/build.gradle.kts
TWiStErRob, line 86: Should this be lazy?
```suggestion
val generateDocumentationConfigurationLibraries by configurations.registering {
```

TWiStErRob (changes requested):
Really nice fix, thank you for noticing and implementing a change!

I left a few comments which are mostly thoughts that came to my head during review. Feel free to discuss if it doesn't improve the code or not clear.

Note: I'm new to Detekt, that's why I tagged @3flex on a few things.

detekt-rules-libraries/build.gradle.kts
```diff
@@ -2,8 +2,29 @@ plugins {
...
+            )
+        )
+    )
+}
+
+tasks.withType<ProcessResources>().configureEach {
+    val generateDocumentationFiles: FileCollection = generateDocumentationConfiguration
+    inputs.files(generateDocumentationFiles)
+        .withPropertyName("generateDocumentationConfiguration")
+        .withPathSensitivity(PathSensitivity.RELATIVE)
```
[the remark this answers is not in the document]
TWiStErRob: seems to be safely hardcoded:
https://github.com/detekt/detekt/blob/36fa0a68ea035d1554f42f14586698e2d36c5013/detekt-utils/src/main/kotlin/io/github/detekt/utils/Yaml.kt#L3-L14

detekt-rules-ruleauthors/build.gradle.kts
```diff
@@ -2,12 +2,27 @@ plugins {
...
+}
+
 dependencies {
     compileOnly(projects.detektApi)
     testImplementation(projects.detektTest)
     testImplementation(libs.assertj)
+
+    generatedRulesConfig(
+        project(":detekt-generator", "generatedRulesConfig")
+    )
```
TWiStErRob, line 17: Probably better on single-line as other dependencies?

Also I saw you were trying to use the DSL... `projects.detektGenerator.dependencyProject.path` works, but a bit wordy.

detekt-rules-libraries/build.gradle.kts
```diff
@@ -2,8 +2,29 @@ plugins {
...
     testImplementation(libs.assertj)
+
+    generateDocumentationConfiguration(
+        project(
+            mapOf(
+                "path" to ":detekt-generator",
+                "configuration" to "generateDocumentationConfigurationLibraries"
+            )
+        )
+    )
```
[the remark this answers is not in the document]
jprinet: Thanks for the hint, let's go with your first proposal 👍

detekt-rules-libraries/build.gradle.kts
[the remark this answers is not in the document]
jprinet: This is required for cache relocatability, the idea is to only rely on the [relative](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/PathSensitivity.html#RELATIVE) part of the file path when computing the cache key. This allow to have cache hits when the project has a different clone location, which is usually the case.

detekt-generator/build.gradle.kts
```diff
@@ -37,7 +35,6 @@ val ruleModules = rootProject.subprojects
 
 val generateDocumentation by tasks.registering(JavaExec::class) {
```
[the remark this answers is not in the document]
jprinet: This one didn't pop in the list of tasks for which optimizations were disabled, not sure why. 
I would only pass artifacts through configuration if required.

detekt-generator/build.gradle.kts
jprinet: Having task outputs in the regular `build` directory is our recommendation.
I think that having them transferred from there to the consumer module (int its `build` directory) would be cleaner.

detekt-rules-ruleauthors/build.gradle.kts
jprinet: Yep, I would have preferred the `path` at the upper level, I reverted to string thus. No strong opinion though.

detekt-rules-libraries/build.gradle.kts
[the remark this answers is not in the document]
TWiStErRob: Yeah, I understand that, I'm proposing to make `\r\n` relocatable (this = `normalizeLineEndings`, sorry that wasn't clear), but it's not necessary with these files.

detekt-generator/build.gradle.kts
TWiStErRob: In my understanding that would need [a separate task](https://github.com/detekt/detekt/pull/5773#discussion_r1101364917), right? So it's easy to move around.

detekt-rules-ruleauthors/build.gradle.kts
TWiStErRob: Is it worth opening a Gradle issue for `.path` on accessor level?
Also it should be quite easy to just write an extension function in default package in `build-logic`: `val ProjectDependency.path: String = dependencyProject.path`, which means it would be trivially visible immediately to all build files.

detekt-generator/build.gradle.kts
TWiStErRob: ~I think that's acceptable, any laziness win counts, as they add up.~ That said, I couldn't prove that its creation is actually delayed. This is because when an item is registered in `DefaultDomainObjectCollection`, it might actually realize the item, if there are "listeners" registered in `eventRegister`. This shows that there are 3 listeners in `project.configurations`:
so all the configurations will be always immediately realized (in Kotlin modules or even where the `base` plugin is applied).

image

This means that `.creating` will be just as good, and we can use the `"`.

---

Note: Gradle base plugin is eagerly creating all configurations here:
`org.gradle.api.plugins.BasePlugin#configureConfigurations`: `configurations.all { }`

TWiStErRob:
Note: there might've been some hidden conversations:
image

detekt-generator/build.gradle.kts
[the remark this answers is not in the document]
TWiStErRob: Here's a consistent repro for the libraries module (shows warning on main, works on PR):
```
gradlew :detekt-generator:cleanGenerateDocumentation :detekt-generator:generateDocumentation :detekt-rules-libraries:cleanProcessResources :detekt-rules-libraries:processResources --no-build-cache
```
I've tried to reproduce the `formatting` one too, but it's very elusive. `formatting:processResources` executes way before `genereateDoc` so it doesn't trigger the warning. even with explicit ordering this is the case:
```
gradlew :detekt-generator:cleanGenerateDocumentation :detekt-generator:generateDocumentation :detekt-formatting:cleanProcessResources :detekt-formatting:processResources --no-build-cache --no-parallel
```
with a build scan (click images below for build scan link) I was able to see the dependency chain:
[image](https://scans.gradle.com/s/ptk3dssf5ve74/timeline?details=obip45c7m6nay&expanded=WyIwLjYwIiwiMC42MC0xLjYxIiwiMC42MC0xLjYxLTIuMTUxIl0&show=successors)
while for libraries there's no such chain:
[image](https://scans.gradle.com/s/fxgtfhzfeau2i/timeline?details=kk3t5ppvxtweg&show=successors)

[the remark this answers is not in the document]
TWiStErRob: See https://github.com/detekt/detekt/pull/5777/files#diff-ea59b2a3d3e92b5338889a89da73aa371f1e84fb6117576240f0f3cc5d1b5c4fR169-R173

@3flex what can we do here? I think breaking the dependency will allow for this PR to apply to the formatting module too, but the root cause is that the config file generation and the documentation generation are bundled together into one module/class.

detekt-rules-libraries/build.gradle.kts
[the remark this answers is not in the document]
jprinet: oh sorry I just missed the extra line in the suggestion. As you said, nothing to add given the implementation 👍
Another way to handle that is to configure it in `Git`.

detekt-generator/build.gradle.kts
jprinet: Thanks for the detailed explanation, that's definitely helpful. I'll revert to the `.creating` approach

detekt-generator/build.gradle.kts
```diff
@@ -86,6 +83,36 @@ val generateDocumentation by tasks.registering(JavaExec::class) {
...
+val generatedCoreConfig: Configuration by configurations.creating {
+    isCanBeConsumed = true
+    isCanBeResolved = false
+}
+
+artifacts {
+    add(generatedLibrariesConfig.name, file(librariesConfigFile)) {
+        builtBy(generateDocumentation)
+    }
+    add(generatedRulesConfig.name, file(ruleauthorsConfigFile)) {
```
TWiStErRob, line 105: would you mind renaming this to
```suggestion
    add(generatedRuleauthorsConfig.name, file(ruleauthorsConfigFile)) {
```
"rulesconfig" sounds like it's all the rules.

detekt-rules-libraries/build.gradle.kts
```diff
@@ -2,8 +2,29 @@ plugins {
...
+}
+
+tasks.withType<ProcessResources>().configureEach {
```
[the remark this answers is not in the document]
jprinet: done 👍

detekt-rules-ruleauthors/build.gradle.kts
jprinet: Added an E`xtensions.kt` in the `build-logic` to support this

detekt-generator/build.gradle.kts
jprinet: done 👍

detekt-rules-libraries/build.gradle.kts
```diff
@@ -2,8 +2,23 @@ plugins {
...
+}
+
 dependencies {
     compileOnly(projects.detektApi)
     testImplementation(projects.detektTest)
     testImplementation(libs.assertj)
+
+    generatedLibrariesConfig(
+        project(projects.detektGenerator.path, "generatedLibrariesConfig")
+    )
```
TWiStErRob, line 17: I think it would be better to keep them similar to other dependencies (it probably fits easily):
```suggestion
    generatedLibrariesConfig(project(projects.detektGenerator.path, "generatedLibrariesConfig"))
```

detekt-rules-libraries/build.gradle.kts
```diff
@@ -2,8 +2,23 @@ plugins {
...
+}
+
+tasks.named<ProcessResources>("processResources").configure {
```
TWiStErRob, line 20: oh, it seems we don't need this, Task.inputs is visible anyway.
```suggestion
tasks.named("processResources").configure {
```

TWiStErRob (approved):
Only 3 tiny nits, waiting for @3flex now.

detekt-rules-ruleauthors/build.gradle.kts
```diff
@@ -2,12 +2,27 @@ plugins {
     id("module")
 }
 
+val generatedRuleauthorsConfig: Configuration by configurations.creating {
```
TWiStErRob, line 5: What do you think about naming these consuming configs the same in every module?
```suggestion
val generatedConfig: Configuration by configurations.creating {
```

Having the same name kind-of suggests that their names have to match, which isn't true. Having the same name also looks confusing at usage, compare:
`generatedRuleauthorsConfig(project(detektGenerator, "generatedRuleauthorsConfig"))`
vs
`generatedConfig(project(detektGenerator, "generatedRuleauthorsConfig"))`

detekt-rules-libraries/build.gradle.kts
jprinet: 👍

detekt-rules-libraries/build.gradle.kts
jprinet: indeed 👍

detekt-rules-ruleauthors/build.gradle.kts
jprinet: Yes, I agree about the confusion and it was tempting to convert the string to the variable name although representing different things.
Let's name the consumer configuration as per your suggestion 👍

detekt-rules-ruleauthors/build.gradle.kts
big-guy: I wouldn't add extensions to do this.  If anything, we should remove `dependencyProject` from `ProjectDependency`. 

I would do this:
```
generatedRulesConfig(projects.detektGenerator) {
        targetConfiguration = "generatedRulesConfig"
}
```

This uses the accessor as it was intended and avoids the Kotlin DSL specific `project(String, String)` method.

TWiStErRob: Nice! Thank you! I agree.

detekt-rules-ruleauthors/build.gradle.kts
TWiStErRob: I think now that there are 3 instances and a 4th one looming (formatter), I would go ahead with this. Considering we need to parameterize on task as well:

```
consumeGeneratedConfig(for = "processResources", from = "generatedRuleauthorsConfig")
```
?

jprinet: This is way cleaner with your suggestion 👍

3flex (approved):

3flex:
Nice work! Thanks for taking the time to do this, and thanks to TWiStErRob and big-guy for reviewing.

3flex:
Unfortunately there's an error that only came up after merging, as none of the PR CI jobs just run a plain `./gradlew build`.

But that's run before deploying a snapshot and it failed with:
```
Circular dependency between the following tasks:
:detekt-formatting:build
\--- :detekt-formatting:check
     \--- :detekt-formatting:detekt
          +--- :detekt-rules-libraries:jar
          |    \--- :detekt-rules-libraries:classes
          |         \--- :detekt-rules-libraries:processResources
          |              \--- :detekt-generator:generateDocumentation
          |                   +--- :detekt-formatting:build (*)
          |                   \--- :detekt-rules:build
          |                        \--- :detekt-rules:check
          |                             \--- :detekt-rules:detekt
          |                                  +--- :detekt-rules-libraries:jar (*)
          |                                  \--- :detekt-rules-ruleauthors:jar
          |                                       \--- :detekt-rules-ruleauthors:classes
          |                                            \--- :detekt-rules-ruleauthors:processResources
          |                                                 \--- :detekt-generator:generateDocumentation (*)
          \--- :detekt-rules-ruleauthors:jar (*)

(*) - details omitted (listed previously)
```
https://ge.detekt.dev/s/2xn35555wzg4q/failure#1
https://github.com/detekt/detekt/actions/runs/4183501703/jobs/7247933889

jprinet:
Commenting those 2 `finalizedBy` instructions helps to get rid of the circular dependency
https://github.com/detekt/detekt/blob/a6d81feef18c4719e810b63e2acbd2810870bf62/detekt-rules/build.gradle.kts#L31
https://github.com/detekt/detekt/blob/a6d81feef18c4719e810b63e2acbd2810870bf62/detekt-formatting/build.gradle.kts#L26

jprinet:
> Unfortunately there's an error that only came up after merging, as none of the PR CI jobs just run a plain `./gradlew build`.

See https://github.com/detekt/detekt/pull/5792 for resolution

PR #5777, opened 2023-02-09
Remove dependency that creates a cycle.

TWiStErRob:
Forked from https://github.com/detekt/detekt/pull/5773#discussion_r1101894177

Since a dependency is removed in this PR, now the warning can trigger as described in the conversation thread on the other PR. This will happen randomly based on parallel ordering. Luckily we got a failure on one of the matrix jobs and [Matthew noticed it](https://github.com/detekt/detekt/pull/5777#issuecomment-1454707004).

A consistent repro is
```
gradlew :detekt-generator:generateDocumentation :detekt-formatting:cleanProcessResources :detekt-formatting:processResources --console=verbose
```
which works on the [Remove dependency that creates a cycle.](https://github.com/detekt/detekt/pull/5777/commits/168678f24d19ece908ef262bf92f48e3004d8e95) commit. After that it should be fixed.


3flex:
I think this can be closed now?

TWiStErRob:
I would still continue this way... the fix in #5792 was not complete, it's just working around the issues.

3flex (approved):

3flex:
I've been thinking about this and agree it should be merged. Was there anything else that you're planning to do for this PR?

TWiStErRob:
I think it might be enough to just do this. Can't think of anything else. Originally I only raised the PR to show the issue, but looking at it, it might be production ready 🤠.

3flex:
I'm good with the PR itself but I thought these errors were dealt with? https://github.com/detekt/detekt/actions/runs/4330440036/jobs/7561663181#step:4:520

Is that new with this change?

TWiStErRob:
Ah, yes! Lucky flaky build. Need to fix that. Now that I broke the circular dependency, the ordering is reversed (randomly). We can now really fix the implicit dependencies. I'll have a look next week.

cortinico:
Do we want this included in 1.23 @TWiStErRob ?

TWiStErRob:
@cortinico No rush, as far as I understand it's more for correctness and tech debt, I don't think users will see anything of it since the change is in `detekt-generator`. That said I'll spend a little time on it now, hopefully finishing it.

cortinico:
Got it. Removing the milestone for now. If we merge it in a couple of weeks it will make it into 1.23 👍

TWiStErRob:
Consistent repro for https://github.com/detekt/detekt/pull/5777#issuecomment-1454707004
```
gradlew :detekt-generator:cleanGenerateDocumentation :detekt-generator:generateDocumentation
```

TWiStErRob:
@3flex, I fixed the missing dependency similar to earlier fixes by @jprinet. Please have a look if this makes sense, I don't think there's anything else to do for now. In the future we definitely need to split up the config generation and somehow prevent referencing source-sets between projects:

https://github.com/detekt/detekt/blob/c3bb324b833f28046664b44be77941c57e76db39/detekt-generator/build.gradle.kts#L48

I'm quite surprised this works with configuration cache. Project isolation all the way! 💪

PR #5785, opened 2023-02-12
Remove Spek support from detekt-test-utils

3flex:
It's been about 10 months since v1.20.0 was released and we deprecated Spek support. I think it's time to drop it completely.

For future readers, if you want to continue using the support that was provided by detekt, just use:

```kotlin
fun Root.setupKotlinEnvironment(additionalJavaSourceRootPath: Path? = null) {
    val wrapper by memoized(
        CachingMode.SCOPE,
        { createEnvironment(additionalJavaSourceRootPaths = listOfNotNull(additionalJavaSourceRootPath?.toFile())) },
        { it.dispose() }
    )

    // name is used for delegation
    @Suppress("UNUSED_VARIABLE")
    val env: KotlinCoreEnvironment by memoized(CachingMode.EACH_GROUP) { wrapper.env }
}
```

You'll need to tweak your imports but otherwise this should continue working for the foreseeable future.

### DIFF
--- a/detekt-test-utils/api/detekt-test-utils.api
+++ b/detekt-test-utils/api/detekt-test-utils.api
@@ -50,8 +50,3 @@ public abstract interface annotation class io/gitlab/arturbosch/detekt/rules/Kot
 	public abstract fun additionalJavaSourcePaths ()[Ljava/lang/String;
 }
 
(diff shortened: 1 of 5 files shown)


schalkms (approved):
Good one!

Issue #5660, opened 2023-01-03
False positive in ProtectedMemberInFinalClass with JVM finalizer


ccjernigan:
## Expected Behavior
JVM finalizer methods shouldn't trigger detekt warnings

## Observed Behavior
`Member with protected visibility in final class is private. Consider using private or internal as modifier. [ProtectedMemberInFinalClass]`

## Steps to Reproduce

Create a class for the JVM that implements the finalize API.  See 

E.g.
```
class MyFinalizable {
    @Throws(Throwable::class)
    protected fun finalize() {
    }
}
```

## Context

https://kotlinlang.org/docs/java-interop.html#finalize

## Your Environment

* Version of detekt used: 1.21.0
* Version of Gradle used (if applicable):
* Gradle scan link (add `--scan` option when running the gradle task):
* Operating System and version:
* Link to your project (if it's a public repository):

PR #5786, opened 2023-02-13
Fixed ProtectedMemberInFinalClass rule reporting valid JVM finalize

lexa-diky:
Fixes https://github.com/detekt/detekt/issues/5660

Added check that declaration does not match JVM finalize signature:
1. name == "finalize"
2. parameters are empty


lexa-diky:
Something odd happened with codecov not updating coverage, will close and reopen new one

PR #5788, opened 2023-02-14
Fixed ProtectedMemberInFinalClass rule reporting valid JVM finalize

lexa-diky:
Fixes https://github.com/detekt/detekt/issues/5660

Added check that declaration does not match JVM finalize signature:

name == "finalize"
parameters are empty

### DIFF
--- a/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignature.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignature.kt
@@ -11,8 +11,8 @@ fun KtFunction.isEqualsFunction() =
 fun KtFunction.isHashCodeFunction() =
     this.name == "hashCode" && this.isOverride() && this.valueParameters.isEmpty()
 
-fun KtDeclaration.isJvmFinalizeFunction() =
-    this.name == "finalize" && this is KtNamedFunction && this.valueParameters.isEmpty()
+fun KtNamedFunction.isJvmFinalizeFunction() =
+    this.name == "finalize" && this.valueParameters.isEmpty()
 
 private val knownAnys = setOf("Any?", "kotlin.Any?")
 fun KtFunction.hasCorrectEqualsParameter() =
--- a/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignature.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignature.kt
@@ -1,9 +1,9 @@
 package io.gitlab.arturbosch.detekt.rules
 
-import org.jetbrains.kotlin.psi.KtDeclaration
 import org.jetbrains.kotlin.psi.KtFunction
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtObjectDeclaration
+import org.jetbrains.kotlin.psi.psiUtil.isPrivate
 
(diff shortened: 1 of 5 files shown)


detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignature.kt
```diff
@@ -10,6 +11,9 @@ fun KtFunction.isEqualsFunction() =
 fun KtFunction.isHashCodeFunction() =
     this.name == "hashCode" && this.isOverride() && this.valueParameters.isEmpty()
 
+fun KtDeclaration.isJvmFinalizeFunction() =
```
schalkms, line 14: Please use `KtFunction.` here and also check for the function being overridden. Then this function is also reusable.

schalkms (commented):
Thanks for this nice 1st contribution and for fixing this scenario!

detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignature.kt
lexa-diky: Ok will make isJvmFinalizeFunction take KtFunction as context. But can't make it check if function is overridden, because official Kotlin documentation state that you don't have to specify `override` keyword explicitly. 

https://kotlinlang.org/docs/java-interop.html#finalize
```kotlin
class C {
    protected fun finalize() {
        // finalization logic
    }
}
```

BraisGabin: Don't check that the function has the keyword. Check that the function doesn't have it, as the documentation sais. And also check that it's not private if possible. And if possible add a comment with the link to the documentation so we have a reference for the future.

With that I think this PR is good to go :) Thanks for the contribution!

detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignature.kt
```diff
@@ -1,5 +1,6 @@
 package io.gitlab.arturbosch.detekt.rules
 
+import org.jetbrains.kotlin.psi.KtDeclaration
```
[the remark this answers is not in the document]
lexa-diky: fixed

detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignature.kt
lexa-diky: Oh, probably was a bit tired writing this comment :). Yes of course we should check for !isOverride. Pushed changes according changes.

P.S. Fun thing `finalize` function is not mentioned in Kotlin Language specification 😞, only in doc

BraisGabin (approved):

detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/MethodSignature.kt
lexa-diky: Looks like `Pre Merge Checks / gradle (windows-latest, 11) (pull_request)` is failed because of flaky test. Could, please, someone restart it.

BraisGabin: We know that we have flaky test on windows. This is ready to merge :) thank you again for the PR!

PR #5797, opened 2023-02-21
Exclude the KMP test folders for android

### LINKED ISSUE #5727: Exclude new Kotlin Test directories in default config
## Expected Behavior
A project using kotlin's new source directory structure does not enforce rules like `lateinitusage` for files in `src/androidUnitTest/**` and `src/androidInstrumentedTest/**` when using the default configuration.

## Current Behavior

rules that are excluded by tests in the default config should also be excluded in the new test folder structure.

## Context

Kotlin 1.8 introduces a new folder structure https://kotlinlang.org/docs/whatsnew18.html#sourcedirectories

The default detekt configuration file excludes a bunch of different variants of test source sets by default.

It would be nice to either add the new `androidUnitTest` and `androidInstrumentedTest` to those defaults as well, or maybe even better, build in support for defaulting all test source sets to be excluded (not sure how feasible that is).

BraisGabin:
Fix #5727 

The old tests were `androidTest` and `androidAndroidTest` (we didn't support this one) and the new ones are `androidUnitTest` and `androidInstrumentedTest`. It's better to just exclude `android*Test`.

### DIFF
--- a/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/Exclusion.kt
+++ b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/Exclusion.kt
@@ -18,12 +18,24 @@ abstract class Exclusions {
     abstract val rules: Set<String>
 
     fun isExcluded(rule: Rule) = rule.name in rules || rule.inMultiRule in rules
+
+    companion object {
+        internal val testFolders = listOf(
+            "test",
+            "androidTest",
+            "commonTest",
+            "jvmTest",
+            "androidUnitTest",
+            "androidInstrumentedTest",
+            "jsTest",
+            "iosTest",
+        )
+    }
 }
 
 private object TestExclusions : Exclusions() {
-
-    override val pattern =
-        "['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']"
+    override val pattern = testFolders.map { "**/$it/**" }
+        .joinToString(prefix = "[", separator = ", ", postfix = "]") { "'$it'" }
     override val ruleSets = emptySet<String>()
     override val rules = setOf(
         "FunctionNaming",
@@ -44,15 +56,12 @@ private object TestExclusions : Exclusions() {
 }
 
 private object KotlinScriptExclusions : Exclusions() {
-
     override val pattern = "['**/*.kts']"
     override val rules = setOf("MissingPackageDeclaration")
 }
 
 private object KotlinScriptAndTestExclusions : Exclusions() {
-
-    override val pattern =
-        "['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**', " +
-            "'**/*.kts']"
+    override val pattern = (testFolders.map { "**/$it/**" } + "**/*.kts")
+        .joinToString(prefix = "[", separator = ", ", postfix = "]") { "'$it'" }
     override val rules = setOf("MagicNumber")
 }
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -67,7 +67,7 @@ comments:
     endOfSentenceFormat: '([.?!][ \t\n\r\f<])|([.?!:]$)'
   KDocReferencesNonPublicProperty:
     active: false
-    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
+    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/androidUnitTest/**', '**/androidInstrumentedTest/**', '**/jsTest/**', '**/iosTest/**']
   OutdatedDocumentation:
     active: false
     matchTypeParameters: true
@@ -75,7 +75,7 @@ comments:
     allowParamOnConstructorProperties: false
   UndocumentedPublicClass:
     active: false
-    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
+    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/androidUnitTest/**', '**/androidInstrumentedTest/**', '**/jsTest/**', '**/iosTest/**']
     searchInNestedClass: true
     searchInInnerClass: true
     searchInInnerObject: true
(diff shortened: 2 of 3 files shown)


detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/Exclusion.kt
```diff
@@ -23,7 +23,7 @@ abstract class Exclusions {
 private object TestExclusions : Exclusions() {
 
     override val pattern =
-        "['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']"
```
TWiStErRob, line 26: I would vote for keeping an explicit list, it's really hard to understand what that star is in the middle of the word otherwise.

BraisGabin: Then we need to add at least 3 keys: `androidTest`, `androidUnitTest` and `androidInstrumentedTest`:

```yaml
  LateinitUsage:
    active: false
    excludes: ['**/test/**', '**/androidTest/**', '**/androidUnitTest/**', '**/androidInstrumentedTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
    ignoreOnClassesPattern: ''
```

It's an easy change, if you still think it's better to add all of them I do the chage. 

I would not add `androidAndroidTest`. It wasn't added before and it's deprecated now.

TWiStErRob: I know it's a lot, but explicitness is always better than magic :)
It's also consistent with existing ignores, which are also from KMP.

Agreed on the deprecated, see already we were setting up to ignore more than intended 🤔.

BraisGabin: Changed ;)

TWiStErRob (approved):

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/Exclusion.kt
```diff
@@ -20,10 +20,20 @@ abstract class Exclusions {
...
+    "androidInstrumentedTest",
+    "jsTest",
+    "iosTest",
+).map { "**/$it/**" }
 
-    override val pattern =
-        "['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']"
+private object TestExclusions : Exclusions() {
+    override val pattern: String
+        get() = testFolders.joinToString(prefix = "[", postfix = "]") { "'$it'" }
```
BraisGabin, line 36: I'm not sure why if I do

```kotlin
    override val pattern = testFolders!!.joinToString()
```

I get `NullPointerException` (but the compiler sais that I don't need the `!!`). I thought that `testFolders` should be defined already but it is not. For that reason I'm implementing them in the lazy way.

@TWiStErRob review this again pleas that the code changed. And I was missing the case of `MagicNumber`.

TWiStErRob: I think if you move `val testFolders` to the companion, then this wouldn't happen. It doesn't make sense though, because simply TestExclusions would just load ExclusionKt beforehand, BUUUUUT ExclusionKt.exclusions is what triggers initialization of TestExclusions, so it's already being loaded. This means that `val testFolders` MUST BE before `val exclusions`.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/Exclusion.kt
```diff
@@ -44,15 +54,12 @@ private object TestExclusions : Exclusions() {
...
     override val rules = setOf("MissingPackageDeclaration")
 }
 
 private object KotlinScriptAndTestExclusions : Exclusions() {
-
-    override val pattern =
-        "['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**', " +
-            "'**/*.kts']"
+    override val pattern: String
+        get() = (testFolders + "**/*.kts").joinToString(prefix = "[", postfix = "]") { "'$it'" }
```
TWiStErRob, line 63: I would do
```suggestion
        get() = (testFolders + "**/*.kts")
            .joinToString(prefix = "[", separator = ",", postfix = "]") { "'$it'" }
```
to remove implicitness, it's complex enough anyway

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/Exclusion.kt
```diff
@@ -20,10 +20,20 @@ abstract class Exclusions {
...
+private val testFolders = listOf(
+    "test",
+    "androidTest",
+    "commonTest",
+    "jvmTest",
+    "androidUnitTest",
+    "androidInstrumentedTest",
+    "jsTest",
+    "iosTest",
+).map { "**/$it/**" }
```
TWiStErRob, line 32: I would move this `.map` to the pattern initializer/getter, because it's specific how to how it's being used. `testFolders` should define the test folders, not the `testFolderPatterns`. Rename would also help, but separation of concerns is preferable: define data, use data (and format data). `**/$it/**` is formatting, that's specific to the `Exclusions` implementation.

TWiStErRob (approved):
Looks ok, just a few nits. I would definitely resolve the ordering problem rather than working around it with a `get()`.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/Exclusion.kt
```diff
@@ -20,10 +20,20 @@ abstract class Exclusions {
...
 
-    override val pattern =
-        "['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']"
+private object TestExclusions : Exclusions() {
```
TWiStErRob, line 34: Silly question: probably not in scope of this PR, but does this abstraction help in keeping it simple? These exclusions classes are very tightly coupled, there could be multiple exclusions objects matching a certain rule, but this is not supported by merging them; instead it's thrown away, so we're not getting two exclusions, nor a warning/error, but nothing! Also `ruleSets` is always empty, so `ruleSetExclusion` is pretty much dead code.

Would a simple map like solve the problem as well?:
```kotlin
mapOf(
        "FunctionNaming" to TEST_PATTERN,
        "LateinitUsage" to TEST_PATTERN,
        ...,
        "MagicNumber" to TEST_AND_SCRIPT_PATTERN,
        "MissingPackageDeclaration" to SCRIPT_PATTERN,
```

BraisGabin: Probably. But in this case I don't care too much. For sure this could be refactored but I prefer not to do it right now.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/Exclusion.kt
BraisGabin: It works. I don't like to declare it inside the abstract class, it doesn't belong there. And I don't like to make it `internal` instead of `private` but 🤷. If it works this part of the code is not that important.

But I didn't like the `get() =` either.

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/Exclusion.kt
BraisGabin: Fixed

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/printer/defaultconfig/Exclusion.kt
BraisGabin: Done

TWiStErRob (approved):
Nice, looks clean :)

Funny about the space in the separator, nice catch.

PR #5798, opened 2023-02-21
Force SerialVerionUID to be private

BraisGabin:
From the documentation:

> It is also strongly advised that explicit serialVersionUID declarations use the private modifier where possible, since such declarations apply only to the immediately declaring class--serialVersionUID fields are not useful as inherited members.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SerialVersionUIDInSerializableClass.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SerialVersionUIDInSerializableClass.kt
@@ -20,7 +20,7 @@ import org.jetbrains.kotlin.psi.psiUtil.isPrivate
 
 /**
  * Classes which implement the `Serializable` interface should also correctly declare a `serialVersionUID`.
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SerialVersionUIDInSerializableClass.kt
```diff
@@ -16,11 +16,14 @@ import org.jetbrains.kotlin.psi.KtConstantExpression
 import org.jetbrains.kotlin.psi.KtObjectDeclaration
 import org.jetbrains.kotlin.psi.KtPrefixExpression
 import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.psi.psiUtil.isPrivate
 
 /**
  * Classes which implement the `Serializable` interface should also correctly declare a `serialVersionUID`.
  * This rule verifies that a `serialVersionUID` was correctly defined.
```
cortinico, line 23:
```suggestion
 * This rule verifies that a `serialVersionUID` was correctly defined and declared as `private`.
```

cortinico (approved):
LGTM 👍

PR #5800, opened 2023-02-21
Reduce configuration of UnusedPrivateMember's split rules

BraisGabin:
Those `allowedNames` were copy&pasted when `UnusedPrivateMember` was splitted. But some of them have no sense in some usages.

This is a follow up of #5722

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -710,7 +710,7 @@ style:
     active: true
   UnusedPrivateMember:
     active: true
-    allowedNames: 'ignored|expected'
+    allowedNames: ''
   UnusedPrivateProperty:
     active: true
     allowedNames: '_|ignored|expected|serialVersionUID'
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -705,12 +705,12 @@ style:
     active: false
   UnusedParameter:
     active: true
-    allowedNames: '(ignored|expected|serialVersionUID)'
+    allowedNames: '(ignored|expected)'
   UnusedPrivateClass:
     active: true
   UnusedPrivateMember:
     active: true
-    allowedNames: '(ignored|expected|serialVersionUID)'
+    allowedNames: '(ignored|expected)'
   UnusedPrivateProperty:
     active: true
     allowedNames: '(_|ignored|expected|serialVersionUID)'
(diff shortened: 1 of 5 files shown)


detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -705,15 +705,15 @@ style:
...
   UnusedPrivateClass:
     active: true
   UnusedPrivateMember:
     active: true
-    allowedNames: '(_|ignored|expected|serialVersionUID)'
+    allowedNames: ''
   UnusedPrivateProperty:
     active: true
-    allowedNames: '(_|ignored|expected|serialVersionUID)'
+    allowedNames: '_|ignored|expected|serialVersionUID'
```
TWiStErRob, line 716: feels like `_` should be a parameter name, i.e.

```kotlin
foo.bar { _, x -> ... }
```

TWiStErRob (approved):
I was wondering why these were set up the the way they are. Nice one 👏

detekt-core/src/main/resources/default-detekt-config.yml
BraisGabin: It does... we could open an issue and move that to the other rule.

also I created this test case and we have a lot of false negatives regarding lambdas:

```kotlin
Pair(1, 2).let { (x, y) -> println(y) }
listOf<Int>().map { x -> 1 }.fold(a) { x, y -> y }
val foo: (String, String) -> Unit = { x, y -> println(y) }
```

And I think that we don't even should configure `_` as an unused parameter. That's exactly the reason `_` exists.

TWiStErRob: Yes, but otherwise it would be a false positive, isn't it?

I didn't know there were so many gaps, my point was that you added `_` as `private property`, but not for parameter names, now I tried and it's not allowed:
image
image
image

so it's pretty much only allowed as a lambda parameter.

TWiStErRob: @BraisGabin based on ^, it looks like we should just remove the `_` exception.

BraisGabin: I didn't remove it from the property rule because, otherwise, we have false positives right now. We should probably take a look at it. I don't think the lambda check should be in the property rule and I think that we should omit those by default. But that's an improvement out of scope for this PR.

TWiStErRob: Ah, I see, the lambda param is a "private property", sorry, I missed that.

PR #5801, opened 2023-02-21
Simplify `TestConfig` usages

BraisGabin:
In nearlly all our tests we use `TestConfig(Map<String, Any>)`. This make us write code like this:

```kotlin
val rule = TooManyFunctions(
    TestConfig(
        mapOf(
            THRESHOLD_IN_CLASSES to "1",
            THRESHOLD_IN_FILES to "1",
            IGNORE_OVERRIDDEN to "false"
        )
    )
)
```

The `mapOf` provide us really little value there and only increase the verbosity. I changed all our test to use `TestConfig(vararg Pair<String, Any>)` it does exactly the same but the code is a bit easier:

```kotlin
val rule = TooManyFunctions(
    TestConfig(
        THRESHOLD_IN_CLASSES to "1",
        THRESHOLD_IN_FILES to "1",
        IGNORE_OVERRIDDEN to "false",
    )
)
```

I also deprecated the old version to promote this new one. I can't remove it because it is part of our public API.

### DIFF
--- a/detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexInterfaceSpec.kt
+++ b/detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexInterfaceSpec.kt
@@ -6,15 +6,14 @@ import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 
-private const val THRESHOLD = 4
-private val defaultConfig = "threshold" to THRESHOLD
-private val staticDeclarationsConfig = TestConfig(defaultConfig, "includeStaticDeclarations" to true)
-private val privateDeclarationsConfig = TestConfig(defaultConfig, "includePrivateDeclarations" to true)
-private val ignoreOverloadedConfig = TestConfig(defaultConfig, "ignoreOverloaded" to true)
+private val defaultThreshold = "threshold" to 4
+private val staticDeclarationsConfig = TestConfig(defaultThreshold, "includeStaticDeclarations" to true)
+private val privateDeclarationsConfig = TestConfig(defaultThreshold, "includePrivateDeclarations" to true)
+private val ignoreOverloadedConfig = TestConfig(defaultThreshold, "ignoreOverloaded" to true)
 
 class ComplexInterfaceSpec {
 
-    private val subject = ComplexInterface(TestConfig(defaultConfig))
+    private val subject = ComplexInterface(TestConfig(defaultThreshold))
 
     @Nested
     inner class `ComplexInterface rule positives` {
--- a/detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexInterfaceSpec.kt
+++ b/detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexInterfaceSpec.kt
@@ -7,20 +7,14 @@ import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 
 private const val THRESHOLD = 4
-private val defaultConfigMap = mapOf("threshold" to THRESHOLD)
-private val staticDeclarationsConfig = TestConfig(
-    defaultConfigMap + ("includeStaticDeclarations" to true)
-)
-private val privateDeclarationsConfig = TestConfig(
-    defaultConfigMap + ("includePrivateDeclarations" to true)
-)
-private val ignoreOverloadedConfig = TestConfig(
-    defaultConfigMap + ("ignoreOverloaded" to true)
-)
+private val defaultConfig = "threshold" to THRESHOLD
+private val staticDeclarationsConfig = TestConfig(defaultConfig, "includeStaticDeclarations" to true)
+private val privateDeclarationsConfig = TestConfig(defaultConfig, "includePrivateDeclarations" to true)
+private val ignoreOverloadedConfig = TestConfig(defaultConfig, "ignoreOverloaded" to true)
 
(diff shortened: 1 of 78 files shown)


detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/TestConfig.kt
```diff
@@ -54,6 +59,7 @@ open class TestConfig(
     }
 
     companion object {
+        @Suppress("DEPRECATION")
         operator fun invoke(vararg pairs: Pair<String, Any>) = TestConfig(mapOf(*pairs))
```
TWiStErRob, line 63: Should we move this to be a constructor? and deprecate the invoke too?

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/TestConfig.kt
```diff
@@ -6,10 +6,15 @@ import io.gitlab.arturbosch.detekt.core.config.tryParseBasedOnDefault
 import io.gitlab.arturbosch.detekt.core.config.valueOrDefaultInternal
 
 @Suppress("UNCHECKED_CAST")
-open class TestConfig(
+open class TestConfig
+@Deprecated("Use Detekt instead TestConfig(vararg Pair<String, Any>) instead")
```
TWiStErRob, line 10: can we add ReplaceWith for automatic replacement?

TWiStErRob (approved):
Really nice one! Pushed a small commit (it was easier to fix them than to call them out on GitHub)

Some questions as conversations, please review them.

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/TestConfig.kt
TWiStErRob, line 10: after merging this we'll have to make sure to update all branches otherwise we'll have to whack-a-mole breaking main again.

TWiStErRob: should we enable the requirement to be up to date with main considering we squash and the update is a click of a button + automerge?

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/TestConfig.kt
BraisGabin: 🤷 I don't know what benefits would provide us that. But I think that could be done in a different PR.

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/TestConfig.kt
BraisGabin: We could, but that would make the code worst, I think.

```kotlin
// from
TestConfig(mapOf("threshold" to 1))

// to
TestConfig(*mapOf("threshold" to 1).entries.map { it.toPair() }.toTypedArray())
```

The idea of this deprecation is to make the code simpler.

detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexInterfaceSpec.kt
```diff
@@ -7,20 +7,14 @@ import org.junit.jupiter.api.Nested
...
-private val staticDeclarationsConfig = TestConfig(
-    defaultConfigMap + ("includeStaticDeclarations" to true)
-)
-private val privateDeclarationsConfig = TestConfig(
-    defaultConfigMap + ("includePrivateDeclarations" to true)
-)
-private val ignoreOverloadedConfig = TestConfig(
-    defaultConfigMap + ("ignoreOverloaded" to true)
-)
+private val defaultConfig = "threshold" to THRESHOLD
```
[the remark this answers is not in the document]
BraisGabin: Done in the last commit (I rebased the branch to avoid conflicts with main)

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/TestConfig.kt
BraisGabin: > should we enable the requirement to be up to date with main considering we squash and the update is a click of a button + automerge?

For sure that's a good thing to avoid break `main`. But our pipelines are slow and flaky so I don't know if enabling the merge queue is a good idea.

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/TestConfig.kt
TWiStErRob: Just thinking about the future, since the `map` constructor is deprecated now, I would imagine it will be hidden eventually, which means that there won't be a way to construct `TestConfig` directly, only through the factory method in the companion object, which is strange considering the same could could be just a constructor. Since the constructor gets deprecated now, I think maybe it would be good to deprecated the companion invoke as well, and create a constructor with the same signature. Ideally the overload resolution would just pick the new constructor and give us source compatibility and keeping the invoke method as a deprecated one would give us binary compatibility for public users.

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/TestConfig.kt
TWiStErRob: Hehe, that reminds me of There's no such thing as not compatible

TWiStErRob: I was thinking if there's a way to say, replace `TestConfig(mapOf(x))` with `TestConfig(x)`, like we can say in structural replace.

PR #5802, opened 2023-02-21
Fix main - Don't use Stream

BraisGabin:
After merging #5705 main is broken. This PR fixes it.


TWiStErRob (approved):
Poor GitHub can't handle this change... here's the IDEA diff for reference

image

PR #5804, opened 2023-02-21
Remove trailing spaces in the project + level trimIndent() indentations

TWiStErRob:
See individual commits, they should be very trivial to review just scrolling through.

Happy to break up into a PR per commit if necessary.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -95,7 +95,7 @@ types of descriptions:
    underneath the issue description. Built-in console reports such as
    `LiteFindingsReport` use these messages to display identified code smells to
    the user.
-   
+
 It is important that you provide a documentation string, an issue description,
 and a message for each code smell that the new rule generates. When authoring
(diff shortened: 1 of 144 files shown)


BraisGabin (approved):
👏 Great Job!

I'm not sure about the commit `Level .trimIndent() tri-quote indentations.`. I see the reason and I kind of agree with it. Butm if we want to maintain this, we should add that to `MultilineRawStringIndentation`. Right now this rule accepts lines without any space or lines with at least as many spaces as the indentation.

Options:
- Don't allow empty lines
- Create a flag on the rule to turn on/off if the empy lines should be accepted.

TWiStErRob:
Agreed on the rule enablement and adjustment, it's just to get the consistency up right now, from here it'll be easier to enable the rule. There's a ton of places where the indentation levels are off, because of adding/removing blocks/nested classes around a code string, but this is big enough as it is already.

TWiStErRob:
Opened https://github.com/detekt/detekt/issues/5810

Issue #5767, opened 2023-02-07
Function suppressor does not work as stated in docs


vyndor:
## Expected Behavior
Function suppressor works with fully qualified name according to [docs](https://detekt.dev/docs/introduction/suppressors#function-suppressor).

## Observed Behavior
Function suppressor only works with bare method names.

## Steps to Reproduce
I have an interface defined like this
```
package com.example

interface Reducer<State, Action> {
    fun reduce(state: State, action: Action): State
}
```
And several implementations of it.

My detect config looks like this:
```
  LongMethod:
    active: true
    ignoreFunction: ['com.example.Reducer.reduce']
```
Detekt still warns me about `reduce(...)` method being too long.
However suppression works if I use bare method name `ignoreFunction: ['reduce']`, but it is controversy to documentation.

## Context
I'm trying to suppress LongMethod for specific method in the interface (`Reducer.reduce(...)`).

## Your Environment

* Version of detekt used: 1.21
* Version of Gradle used (if applicable): 7.3.3
* Operating System and version: macOS 13.1

lexa-diky:
Hi @vyndor you might misinterpreted the documentation, It says:

> ... You can ignore all the **overloads** of a function ...

Overloading = multiple methods of _class / interface_ have the same name. 

In your case you have overriding, which is a different thing. So behavior you specified is true to documentation. 

P.S. Would be cool to have an ability to specify suppressor for overridden methods. But It will require type resolution, which is not fully supported by detekt (this last to my knowledge)

vyndor:
Hi!
Thanks for your reply, but I'm not mixed up overriding and overloading. I just wanted to provide enough details for my use case. Actually there are some code in the rule checking parents so it at least seems to be designed with type resolution in mind. But that is not what this bug about.

The issue is that function suppressor check bare method names instead of fully qualified. This is what stated in docs:
> Note: you need to write all the types with fully qualified names

So in my case I can change my config to this and it will not work
```
  LongMethod:
    active: true
    ignoreFunction: ['com.example.MyReducerImpl.reduce']
```

And this will work
```
  LongMethod:
    active: true
    ignoreFunction: ['reduce']
```

lexa-diky:
Oh, ok I see now, interface sample threw me off. Seams like issue is with `FunctionMatcher.NameOnly.match(KtNamedFunction, BindingContext)` it should check `fqName` instead of just `name`.

BraisGabin:
[...]

PR #5812, opened 2023-02-22
FunctionMatcher support for fully qualified function names

lexa-diky:
Now matcher will check both simple name e.g. `foo` and fully qualified one e.g. `com.github.detekt.FooBar.foo`.

Fixes https://github.com/detekt/detekt/issues/5767

### DIFF
--- a/detekt-tooling/src/test/kotlin/io/github/detekt/tooling/api/FunctionMatcherSpec.kt
+++ b/detekt-tooling/src/test/kotlin/io/github/detekt/tooling/api/FunctionMatcherSpec.kt
@@ -5,7 +5,9 @@ import io.gitlab.arturbosch.detekt.rules.KotlinCoreEnvironmentTest
 import io.gitlab.arturbosch.detekt.test.getContextForPaths
 import org.assertj.core.api.Assertions.assertThat
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
+import org.jetbrains.kotlin.psi.KtClass
 import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.psiUtil.findFunctionByName
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.junit.jupiter.api.DisplayName
 import org.junit.jupiter.api.DynamicTest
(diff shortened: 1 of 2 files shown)


BraisGabin:
This looks good! You didn't relate this with the issue because right now it doesn't fix that issue, right?

lexa-diky:
> This looks good! You didn't relate this with the issue because right now it doesn't fix that issue, right?

I want to go through tests one more time, I have a suspension that they dont check behavior correctly

lexa-diky:
> > This looks good! You didn't relate this with the issue because right now it doesn't fix that issue, right?
> 
> I want to go through tests one more time, I have a suspension that they dont check behavior correctly

Now it should be correct.

Previously most tests assumed that function is in the root package. So if we remove name check inside matcher tests would still pass. Not a case anymore

BraisGabin (approved):
Good one :) Thanks!

BraisGabin:
I'm re-running CI because we just enabled a rule regarding the identation on the tests (#5819). I think that you need to change something to make the code complaint. Apart from that this is ready to merge :)

detekt-tooling/src/test/kotlin/io/github/detekt/tooling/api/FunctionMatcherSpec.kt
```diff
@@ -261,6 +295,48 @@ class FunctionMatcherSpec(private val env: KotlinCoreEnvironment) {
...
+            result: Boolean
+        ) {
+            val ktFile = compileContentForTest(
+                """
+                package io.github.detekt
+
+                $classSignature {
+                    $functionSignature {}
+                }
+                """.trimIndent()
```
BraisGabin, line 331:
```suggestion
                """
                    package io.github.detekt

                    $classSignature {
                        $functionSignature {}
                    }
                """.trimIndent()
```

detekt-tooling/src/test/kotlin/io/github/detekt/tooling/api/FunctionMatcherSpec.kt
```diff
@@ -270,8 +346,17 @@ private class TestCase(
...
+private fun buildKtFunction(
+    environment: KotlinCoreEnvironment,
+    code: String,
+    includePackage: Boolean = true
+): Pair<KtNamedFunction, BindingContext> {
+    val ktFile = compileContentForTest(
+        """
+        ${if (includePackage) "package io.github.detekt" else ""}
+        $code
+        """.trimIndent()
```
BraisGabin, line 358:
```suggestion
        """
            ${if (includePackage) "package io.github.detekt" else ""}
            $code
        """.trimIndent()
```

cortinico (approved):

PR #5772, opened 2023-02-09
Avoid configuring Detekt tasks unnecessarily

3flex:

PR #5813, opened 2023-02-22
Stop configuring report merge tasks while configuring Detekt tasks

3flex:
See point 2: https://docs.gradle.org/8.0.1/userguide/task_configuration_avoidance.html#sec:task_configuration_avoidance_general

> Only mutate the current task inside a configuration action... Mutating anything other than the current task can cause indeterminate behavior in your build.

Also see: https://github.com/detekt/detekt/pull/5772#issuecomment-1431708458.

Docs should be updated to align - that can be done in a new commit once these changes are reviewed & approved.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -36,7 +36,7 @@ allprojects {
         detektPlugins(project(":detekt-rules-ruleauthors"))
     }
 
-    tasks.withType<Detekt>().configureEach detekt@{
+    tasks.withType<Detekt>().configureEach {
         jvmTarget = "1.8"
         reports {
             xml.required.set(true)
@@ -47,9 +47,9 @@ allprojects {
         }
         basePath = rootProject.projectDir.absolutePath
         finalizedBy(detektReportMergeSarif)
-        detektReportMergeSarif.configure {
-            input.from(this@detekt.sarifReportFile)
-        }
+    }
+    detektReportMergeSarif {
+        input.from(tasks.withType<Detekt>().map { it.sarifReportFile })
     }
     tasks.withType<DetektCreateBaselineTask>().configureEach {
         jvmTarget = "1.8"
--- a/detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektReportMergeSpec.kt
+++ b/detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektReportMergeSpec.kt
@@ -30,13 +30,11 @@ class DetektReportMergeSpec {
                 |    reports.sarif.enabled = true
                 |  }
                 |
-                |  plugins.withType(io.gitlab.arturbosch.detekt.DetektPlugin::class) {
-                |    tasks.withType(io.gitlab.arturbosch.detekt.Detekt::class) detekt@{
-                |       sarifReportMerge.configure {
-                |         this.mustRunAfter(this@detekt)
-                |         input.from(this@detekt.sarifReportFile)
-                |       }
-                |    }
+                |  tasks.withType(io.gitlab.arturbosch.detekt.Detekt::class).configureEach {
+                |     finalizedBy(sarifReportMerge)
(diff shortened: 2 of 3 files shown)


build.gradle.kts
```diff
@@ -47,9 +47,9 @@ allprojects {
         }
         basePath = rootProject.projectDir.absolutePath
         finalizedBy(detektReportMergeSarif)
-        detektReportMergeSarif.configure {
-            input.from(this@detekt.sarifReportFile)
-        }
+    }
+    detektReportMergeSarif {
+        input.from(tasks.withType<Detekt>().map { it.sarifReportFile })
```
TWiStErRob, line 52: This `.map` is eager, this means all Detekt tasks will be configured always, even when we're only running one of them.

So running `:a:detekt` will configure `:*:detekt`.

TWiStErRob: oh, wait, in the [original code](https://github.com/detekt/detekt/pull/5772/files#diff-c0dfa6bc7a8685217f70a860145fbdf416d449eaff052fa28352c5cec1a98c06L39) we were ALWAYS configuring them, even when running `gradlew jar`. 😅

build.gradle.kts
```diff
@@ -47,9 +47,9 @@ allprojects {
         }
         basePath = rootProject.projectDir.absolutePath
         finalizedBy(detektReportMergeSarif)
-        detektReportMergeSarif.configure {
-            input.from(this@detekt.sarifReportFile)
-        }
+    }
+    detektReportMergeSarif {
```
TWiStErRob, line 51: Interesting issue https://scans.gradle.com/s/oth2mvn5uamrm/timeline?details=f55zutikdiwjg&show=details, it should be cacheable, right? since that task creates the file. Not blocking this PR though.

TWiStErRob (approved):
LGTM, tested as well, behaviors were as expected. Please update docs, and reply to conversations.

build.gradle.kts
TWiStErRob, line 51: I would keep the explicit `.configure { }` to show the laziness, not everyone knows that `operator invoke` is an alias for `configure`.

TWiStErRob: Never mind, this is only because my local state had the file present, and it wasn't changed.

build.gradle.kts
```diff
@@ -36,7 +36,7 @@ allprojects {
         detektPlugins(project(":detekt-rules-ruleauthors"))
     }
 
-    tasks.withType<Detekt>().configureEach detekt@{
+    tasks.withType<Detekt>().configureEach {
```
TWiStErRob, line 39: what's happening on CI with Windows? these changed can't cause that, right?

build.gradle.kts
3flex: I'll include it in the Groovy version of the docs. But the convention for Kotlin DSL is not to include it. See https://docs.gradle.org/current/userguide/more_about_tasks.html

build.gradle.kts
3flex: The problem was introduced with upgrade to Gradle 8 and fixed by https://github.com/detekt/detekt/pull/5811. I'll rebase and it should go green.

chao2zhang (approved):
TIL

Did you catch this bug manually or there are new warnings from Gradle 8?

3flex:
> Did you catch this bug manually or there are new warnings from Gradle 8?

It was caught after #5772 was merged - I don't think it's related to Gradle 8. That warning from the Gradle docs that I referenced has been there for some time, I just haven't had a closer look until I made the change in #5772, not realising that the tasks had to be eagerly configured with the current way of configuring report merging.

build.gradle.kts
3flex: Yeah this isn't perfect, there's an open bug on Gradle side already: https://github.com/gradle/gradle/issues/14870

BraisGabin (approved):

TWiStErRob (approved):

PR #5820, opened 2023-02-24
Rewrite testPluginKotlinc task to be configuration cache compatible

3flex:


BraisGabin (approved):

BraisGabin:
This comment is not directly related with this PR: 

I saw that this task is registered to execute with `check`. Shouldn't it be better to relate it with `test`?

3flex:
> I saw that this task is registered to execute with check. Shouldn't it be better to relate it with test?

I don't think so. Normal, non-lifecycle tasks shouldn't have explicit dependencies on custom tasks just to get them to run together, and `test` isn't a lifecycle task. Custom tasks should be instead linked to lifecycle tasks if those kinds of dependencies are required. [The doc for `check`](https://docs.gradle.org/8.0.1/userguide/base_plugin.html#sec:base_tasks) says:

> Plugins and build authors should attach their verification tasks, such as ones that run tests, to this lifecycle task

PR #5821, opened 2023-02-24
Remove use of afterEvaluate

3flex:


build-logic/src/main/kotlin/releasing.gradle.kts
```diff
@@ -16,37 +16,33 @@ nexusPublishing {
...
-    }
+githubRelease {
+    token(providers.gradleProperty("github.token"))
+    owner.set("detekt")
+    repo.set("detekt")
+    overwrite.set(true)
+    dryRun.set(false)
+    draft.set(true)
+    targetCommitish.set("main")
+    body(provider {
```
3flex, line 27: This is the key change - wrapping the content in a provider means it will be accessed lazily (though I'm not even sure it has to be, if the changelog isn't generated dynamically).

All the other parameters are set to static values that are available at configuration time.

detekt-gradle-plugin/build.gradle.kts
```diff
@@ -175,31 +175,29 @@ tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach
...
-                    }
-                }
-                developers {
-                    developer {
-                        id.set("Detekt Developers")
-                        name.set("Detekt Developers")
-                        email.set("info@detekt.dev")
-                    }
+publishing {
+    publications.withType<MavenPublication> {
```
3flex, line 179: This is the key change - using `withType` means the publications are configured as soon as they're added to the publications container. It's also "live" so the configuration will be done on any publication regardless of which configuration phase it's added.

BraisGabin (approved):

detekt-gradle-plugin/build.gradle.kts
TWiStErRob: It's also eager, did you want configureEach here?

TWiStErRob (approved):
Looks nice, how did you test this? Did you diff the `publishToMavenLocal` before/after?

cortinico (approved):

detekt-gradle-plugin/build.gradle.kts
3flex: `withType` is just shorthand for `withType.all`, and unlike a TaskContainer, using it on a normal DomainObjectCollection will configure any elements in the collection, and any that are added later. It doesn't eagerly realise them.

3flex:
> Looks nice, how did you test this?

For the Gradle plugin I checked the output of the generatePomFile* tasks to make sure the customisations came through to the generated POM.

For the GitHub release I looked at the output of `tasks.githubRelease.get().body.get().toString()` which looked sane to me.

PR #5822, opened 2023-02-24
Dont reexecute tasks and make them more parallel

BraisGabin:
We were executing the tasks `detekt-cli:runWithHelpFlag` and `:detekt-cli:runWithArgsFile` twice:

https://github.com/detekt/detekt/blob/ad0e7930340e73a64318165ac08f00b4f5e4f757/detekt-cli/build.gradle.kts#L73-L75

And we don't need to execute `publishToMavenLocal` in a different ci task. We can executes them together and use the parallelism of gradle.

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -49,16 +49,6 @@ jobs:
           path: '**.hprof'
           if-no-files-found: ignore
 
-      - name: Run detekt-cli --help
-        uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2
-        with:
-          arguments: :detekt-cli:runWithHelpFlag
-
-      - name: Run detekt-cli with argsfile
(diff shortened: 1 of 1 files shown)


3flex:
I don't even know why we run `publishToMavenLocal` in CI... the output isn't used for anything. Could we remove it?

BraisGabin:
I think it's here because it executes parts of the code that are only executed on the release process. And @cortinico always find those in the worst moment.

3flex (approved):

cortinico:
> I think it's here because it executes parts of the code that are only executed on the release process. And @cortinico always find those in the worst moment.

Basically this. IMHO the CI should `publishToMavenLocal` + we should upload `~/.m2` as an output, so people could theoretically download that zip and depend on it (without having to build detekt from a given PR).

> That task is already executed first in [main/scripts/release.sh](https://github.com/detekt/detekt/blob/main/scripts/release.sh?rgh-link-date=2023-02-24T12%3A08%3A40Z) so if there are failures in that task it would be caught at the first step.

@3flex I run that script manually once every 3 months. I can't be the one to bisect the commit if the pipeline fails. Still we publish snapshots so that sort of covers this logic.

PR #5823, opened 2023-02-24
Never skip ci

BraisGabin:
I can't find any reason to don't want to run ci on every PR. And this could even be sploited by someone sending a PR. For that reason I'm removing this `if`

### DIFF
--- a/.github/workflows/deploy-snapshot.yaml
+++ b/.github/workflows/deploy-snapshot.yaml
@@ -14,7 +14,6 @@ permissions:
 jobs:
   gradle:
     runs-on: ubuntu-latest
-    if: github.repository == 'detekt/detekt' && !contains(github.event.head_commit.message, 'ci skip')
     steps:
       - name: Checkout Repo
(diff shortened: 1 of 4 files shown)


3flex:
Could drop the "coverage skip" one in codecoverage.yaml as well

3flex (approved):

BraisGabin:
> Could drop the "coverage skip" one in codecoverage.yaml as well

Done!

.github/workflows/deploy-snapshot.yaml
```diff
@@ -14,7 +14,6 @@ permissions:
 jobs:
   gradle:
     runs-on: ubuntu-latest
-    if: github.repository == 'detekt/detekt' && !contains(github.event.head_commit.message, 'ci skip')
```
cortinico, line 17: Sorry I was late :)
But this was actually needed: 
- https://github.com/detekt/detekt/pull/5833

PR #5824, opened 2023-02-24
Reduce execution of warningsAsErrors job

BraisGabin:
On this task we just want to execute the kotlin compile to get errors if there is any warning.

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -101,4 +101,4 @@ jobs:
       - name: Run with allWarningsAsErrors
         uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2
         with:
-          arguments: build -x detekt -PwarningsAsErrors=true
+          arguments: compileKotlin compileTestKotlin compileTestFixturesKotlin -PwarningsAsErrors=true


3flex:
This misses `compileTestFixturesKotlin`.

Also - why do we actually need a separate job just for compiling with warnings? We could just add `-PwarningsAsErrors=true` to the "build detekt" task on line 42. It would CI time.

BraisGabin:
> This misses `compileTestFixturesKotlin`.

I'm going to check that, thanks!

> Also - why do we actually need a separate job just for compiling with warnings? We could just add `-PwarningsAsErrors=true` to the "build detekt" task on line 42. It would CI time.

The problem is the remote cache. Locally we never compile with `-PwarningsAsErrors=true` (it is a pain in the ass).

BraisGabin:
> This misses `compileTestFixturesKotlin`.

TL;DR: added

It was executed: https://ge.detekt.dev/s/gohvburcndpyg/timeline#siaukmwc4od7u and for that reason I didn't spot the error. I was diffing the `--dry-run` of both executions to ensure that I wasn't missing anything. But for sure it's better to be exhaustive here.

.github/workflows/pre-merge.yaml
```diff
@@ -101,4 +101,4 @@ jobs:
       - name: Run with allWarningsAsErrors
         uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2
         with:
-          arguments: build -x detekt -PwarningsAsErrors=true
+          arguments: compileKotlin compileTestKotlin compileTestFixturesKotlin -PwarningsAsErrors=true
```
TWiStErRob, line 104: I would go with `classes testClasses jar`, because it's more stable in the long run, it would contain anything that is required for compilation, kapt, ksp, plugin generation etc.
```suggestion
          arguments: classes testClasses jar -PwarningsAsErrors=true
```

TWiStErRob, line 104: The above missed `:detekt-gradle-plugin:compileFunctionalTestKotlin`.

TWiStErRob: \+ `functionalTestClasses`

```suggestion
          arguments: classes testClasses functionalTestClasses jar -PwarningsAsErrors=true
```

TWiStErRob:
> The problem is the remote cache. Locally we never compile with -PwarningsAsErrors=true (it is a pain in the ass).

Sounds like maybe we need to go the other way around. Go with strict CI and create a separate job for "fill developer remote cache"?

.github/workflows/pre-merge.yaml
BraisGabin: Doesn't that just compile more things that I don't care? `warningsAsErrors=true` is a flag that only the kotlin compiler reads.

TWiStErRob: KotlinCompile is not a final class, other tasks might extend it. Calling the `compile*Kotlin` tasks directly relies on implementation details of KGP. `classes` / `jar` are documented Gradle lifecycle methods.

3flex:
> The problem is the remote cache. Locally we never compile with -PwarningsAsErrors=true

Ah yes! I'd forgotten about this.

> Go with strict CI and create a separate job for "fill developer remote cache"?

That would make it clearer about why we run it twice.

3flex:
> It was executed

Oh right, because compileTestKotlin would have a dependency on compileTestFixturesKotlin. Probably don't need it explicitly then. Good idea checking the dry run.

.github/workflows/pre-merge.yaml
3flex: I was going to mention that too however compile with warnings is disabled for the detekt Gradle plugin regardless of the value of this flag so there's no reason to add it at this stage.

cortinico: > Calling the compile*Kotlin tasks directly relies on implementation details of KGP

+1 on this

3flex: My 2c is that `compileKotlin compileTestKotlin compileTestFixturesKotlin` captures exactly what we need. That's detekt code that we have control over that we need to warn about.

> Calling the compile*Kotlin tasks directly relies on implementation details of KGP

This isn't true. [Here](https://kotlinlang.org/docs/gradle-compiler-options.html#how-to-define-options) it explicitly says that the tasks can and should be referenced directly by name - "When targeting the JVM, the tasks are called compileKotlin for production code and compileTestKotlin for test code. The tasks for custom source sets are named according to their compile<Name>Kotlin patterns."

> classes / jar are documented Gradle lifecycle methods.

Just FYI `jar` isn't a lifecycle task (lifecycle tasks are tasks that do not do work themselves, and `jar` task does some work as it assembles the JAR archive).

TWiStErRob: Oh, nice RTFM, thanks. True about jar!

I guess if this is the case we can go with explicit refs. What will prevent though future other warnings from showing up if there are more tasks/source sets added?
Someone noticing locally and raising a PR? (I guess that's acceptable, that's what most projects do, except the raising a PR part 😞.)

TWiStErRob (approved):

PR #5825, opened 2023-02-24
Enable test retry for all our modules

BraisGabin:
This was a point we talked with the Gradle folks. It improves the predictive tests.

Documentation: https://docs.gradle.com/enterprise/predictive-test-selection/#flaky_test_retry

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,3 +1,4 @@
+import com.gradle.enterprise.gradleplugin.testretry.retry
 import io.gitlab.arturbosch.detekt.Detekt
 import io.gitlab.arturbosch.detekt.DetektCreateBaselineTask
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -58,6 +59,13 @@ allprojects {
 
 subprojects {
     tasks.withType<Test>().configureEach {
+        retry {
+            @Suppress("MagicNumber")
```
cortinico, line 63: Perhaps we can really extract those as top level constants?

cortinico (approved):

build.gradle.kts
BraisGabin: To me these are false positives. `maxRetries.set(3)` is clear what `3` does. We don't need a constant to make it easier to understand.

PR #5826, opened 2023-02-24
Disable PTS from local and enable it for PRs

BraisGabin:
The github yaml configuration make this a bit ugly.

This was discussed with the gradle folks on how to improve the usage of PTS.

PTS on local doesn't provide us too much benefits because it breaks the incremental compilariont (`UP-TO-DATE` check).

The idea is that we make our PR checks faster now that we have a trained PTS but we run all our tests once it's merged to `main` to ensure that nothing was broken.

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -37,10 +37,17 @@ jobs:
           java-version: ${{ matrix.jdk }}
           distribution: 'temurin'
 
-      - name: Build detekt
+      - if: ${{ !(github.event_name == 'push' && github.ref == 'refs/heads/main') }}
+        name: Build detekt with PTS
         uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2
         with:
-          arguments: build -x detekt publishToMavenLocal
+          arguments: build -x detekt publishToMavenLocal -PenablePTS=true
+
+      - if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+        name: Build detekt without PTS
+        uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2
+        with:
+          arguments: build -x detekt publishToMavenLocal -PenablePTS=false
 
       - uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3
         with:
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -40,7 +40,7 @@ jobs:
       - name: Build detekt
         uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2
         with:
-          arguments: build -x detekt publishToMavenLocal -PenablePTS=${{ github.event_name != 'push' }}
+          arguments: build -x detekt publishToMavenLocal -PenablePTS=${{ github.event_name == 'pull_request' }}
 
       - uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3
         with:
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -37,17 +37,10 @@ jobs:
           java-version: ${{ matrix.jdk }}
           distribution: 'temurin'
 
-      - if: ${{ !(github.event_name == 'push' && github.ref == 'refs/heads/main') }}
-        name: Build detekt with PTS
+      - name: Build detekt
         uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2
         with:
-          arguments: build -x detekt publishToMavenLocal -PenablePTS=true
-
-      - if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
-        name: Build detekt without PTS
-        uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2
-        with:
-          arguments: build -x detekt publishToMavenLocal -PenablePTS=false
+          arguments: build -x detekt publishToMavenLocal -PenablePTS=${{ github.event_name != 'push' }}
 
       - uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3
         with:
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -59,7 +59,7 @@ allprojects {
 subprojects {
     tasks.withType<Test>().configureEach {
         predictiveSelection {
-            enabled.set(providers.gradleProperty("enablePTS").map(String::toBooleanStrict).orElse(false))
+            enabled.set(providers.gradleProperty("enablePTS").map(String::toBooleanStrict))
         }
     }
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -59,7 +59,7 @@ allprojects {
 subprojects {
     tasks.withType<Test>().configureEach {
         predictiveSelection {
-            enabled.set(System.getenv("CI") == null)
+            enabled.set(providers.gradleProperty("enablePTS").map(String::toBooleanStrict).orElse(false))
         }
     }
 }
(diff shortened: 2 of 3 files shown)


clayburn (approved):
I'm only mildly familiar with the GitHub Actions part. If `github.event_name == 'push' && github.ref == 'refs/heads/main'` executes on main after pull requests are merged to main, then this looks good to me! The intent of this looks correct.

.github/workflows/pre-merge.yaml
```diff
@@ -38,7 +38,14 @@ jobs:
           java-version: ${{ matrix.jdk }}
           distribution: 'temurin'
 
-      - name: Build detekt
+      - if: ${{ !(github.event_name == 'push' && github.ref == 'refs/heads/main') }}
+        name: Build detekt with PTS
+        uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2
+        with:
+          arguments: build -x detekt publishToMavenLocal -PisPTSEnabled
```
TWiStErRob, line 45: Name is a bit confusing `-PisPTSEnabled=false` will still have the same effect as `-PisPTSEnabled=false`.

I would recommend
 * setting `isPTSEnabled=false` in `gradle.properties`
 * `isPTSEnabled=true|false #default` for developers and non-pre-merge checks
 * `-PisPTSEnabled=true` in yaml.

build.gradle.kts
```diff
@@ -59,7 +59,7 @@ allprojects {
 subprojects {
     tasks.withType<Test>().configureEach {
         predictiveSelection {
-            enabled.set(System.getenv("CI") == null)
+            enabled.set(project.hasProperty("isPTSEnabled"))
```
TWiStErRob, line 62: This turns of PTS locally, was that intended?

.github/workflows/pre-merge.yaml
BraisGabin: I copy&pasted it from the documentation and I didn't even read it. What do you think about to call it `enablePTS` and keep it as somethig that you have or you don't have:

`./gradlew build -x detekt publishToMavenLocal -PenabledPTS`

TWiStErRob: Still open for interpretation as a boolean :) + There's no documentation on what it does (other than existing in yaml)

Just to show an example of listing properties with defaults and docs:
https://github.com/TWiStErRob/net.twisterrob.gradle/blob/bebbae7eec76dd65234799f47734bc6e8628f286/gradle.properties#L23-L48
everything that the build can do is listed there.

BraisGabin: 😂 That's exactly the reason I like to have a full detekt configuration but I don't know why I like it for detekt but not that much for gradle. And I think that the sames happend to you but the other way around.

Anyway, changed ;) I like the idea to list those on the gradle.properties.

TWiStErRob: Haha, true that. For me it's the amount... there are max 20 gradle.properties in a project, and they (should) change every year or so, but Detekt has hundreds of rules that change every release.

.github/workflows/pre-merge.yaml
```diff
@@ -37,10 +37,17 @@ jobs:
...
         uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2
         with:
-          arguments: build -x detekt publishToMavenLocal
+          arguments: build -x detekt publishToMavenLocal -PenablePTS=true
+
+      - if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+        name: Build detekt without PTS
+        uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2
+        with:
+          arguments: build -x detekt publishToMavenLocal -PenablePTS=false
```
TWiStErRob, line 50: Might be a silly thing, but wdyt about?
```yaml
      - name: Build detekt
        uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2
        with:
          arguments: build -x detekt publishToMavenLocal -PenablePTS=${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
```
(or equivalent construct where we assign a variable in a similar fashion)

TWiStErRob (approved):

build.gradle.kts
```diff
@@ -59,7 +59,7 @@ allprojects {
 subprojects {
     tasks.withType<Test>().configureEach {
         predictiveSelection {
-            enabled.set(System.getenv("CI") == null)
+            enabled.set(project.property("enablePTS")?.toString()?.toBooleanStrict() == true)
```
TWiStErRob, line 62: oh, wait a minute, wasn't this lazy before? I think the configuration-cache friend version is something like:
```suggestion
            enabled.set(providers.property("enablePTS").map { it?.toString()?.toBooleanStrict() == true})
```

BraisGabin: This was extracted from the gradle documentation: https://docs.gradle.com/enterprise/predictive-test-selection/#conditionally_enabling

Seriusly, it is *really* diffcult to implement something in gradle correctly >_<. So many APIs that does the same but only one is the correct.

Fixed.

.github/workflows/pre-merge.yaml
BraisGabin: I don't know what's the output of `${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}` I'm running it and see what happens.

build.gradle.kts
TWiStErRob: There's a fine line between correctness and easy to understand documentation :) It's a very hard balance. That said it's good feedback for @clayburn ;)

.github/workflows/pre-merge.yaml
TWiStErRob: Looks like it's the right thing.

Note: we're lucky though the event name is always there, so be careful if you use the same pattern elsewhere, nulls (= GHA empty string?) can bite, always test in all contexts.

build.gradle.kts
```diff
@@ -59,7 +59,7 @@ allprojects {
 subprojects {
     tasks.withType<Test>().configureEach {
         predictiveSelection {
-            enabled.set(System.getenv("CI") == null)
+            enabled.set(providers.gradleProperty("enablePTS").map(String::toBooleanStrict).orElse(false))
```
TWiStErRob, line 62: 🥗 probably not necessary, because the default is listed in `gradle.properties` and if the property resolution fails or is empty because for example `-PenablePTS=` (or otherwise unset), that should be a fail.

```suggestion
            enabled.set(providers.gradleProperty("enablePTS").map(String::toBooleanStrict))
```

.github/workflows/pre-merge.yaml
```diff
@@ -40,7 +40,7 @@ jobs:
       - name: Build detekt
         uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2
         with:
-          arguments: build -x detekt publishToMavenLocal
+          arguments: build -x detekt publishToMavenLocal -PenablePTS=${{ github.event_name != 'push' }}
```
TWiStErRob, line 43:
```suggestion
          arguments: build -x detekt publishToMavenLocal -PenablePTS=${{ github.event_name != 'push' }}
```
nice iterations, it looks pretty simple now and more natural for anyone who reads it.

I was thinking about `!= push` vs `== pull_request`, because of our current PR title. But then we kind of want to "disable it for main" so that it runs stricter on main, and uses PTS everywhere else (PRs and potential other stuff that we might add as a trigger). ⚖ 🤔

TWiStErRob (approved):

build.gradle.kts
BraisGabin: I added that exactly for that. If someone removes that line from `gradle.properties` I think it shouldn't fail.

TWiStErRob: But the point of `gradle.properties` is that it is the definition. No-one should be editing gradle.properties directly other than if you intend to change defaults. It's like removing a `val x = true` and then expecting that `foo(x)` will still compile :D

TWiStErRob: Also if we decide to change the default value, we have to change it twice now.

.github/workflows/pre-merge.yaml
BraisGabin: Agree! Done.

build.gradle.kts
BraisGabin: Agree, done

TWiStErRob (approved):
Looks minimally correct to me :)

BraisGabin:
I would like to have more eyes on this to ensure that we are ok with the change that this implies. The code is really simple now but I want to be sure that we agree with this. Don't run ALL the tests on each PR is a big deal.

cortinico (approved):
> I would like to have more eyes on this

+1 on my end for this

3flex (approved):

PR #5819, opened 2023-02-24
Activate MultilineRawStringIndentation on detekt

BraisGabin:
Improve the indentation of our tests

PR #5830, opened 2023-02-24
Cleaner merging of Gradle blocks for functionalTests

TWiStErRob:
Follow-up on https://github.com/detekt/detekt/pull/5819#discussion_r1116745695
Related #5827
Closes #5828 

Changes:
 * Fix indentation of fixture code (somewhere 4, somewhere 3, somewhere 2 spaces), yaml kept at 2 everything else is 4.
 * Replace all remaining trimMargin with trimIndent
 * Add some `@Language` annotations, the build code in functionalTests will be highlighted properly (requires trimMargin removal).
 * Re-think how multiple pieces of Gradle script blocks are merged.
   * Previously
     * they were relying on the `|` being present in the input string,
     * some of them didn't have the leading margin marker.
     * simple string concatenation was done with triple-quotes
     * the generated output was still messy, mis-indented, and hard to follow.
   * In this PR
     * Removed all `@Suppress("TrimMultilineRawString")`
     * Each code block is self-contained
     * This is achieved by 2 simple rules:
       1. Apply trimIndent() at the definition of the fragment
       2. Deal with indentation where the merging happens
     * This needed two small functions to be introduced: `reIndent` and `mergeGradleBlocks`.  
       Their usages are pretty localized and make the fragment re-use be at a higher level of abstraction (like `"""$a $b"""` -> `merge(a, b)`).

You can look at the resulting changes by diffing these two files (note: console output trims empty lines so the effect of `\n\n` is not clearly visible here, but it is in the JUnit test report):
[main.txt](https://github.com/detekt/detekt/files/10824885/main.txt)
[pr.txt](https://github.com/detekt/detekt/files/10824887/pr.txt)

### DIFF
--- a/detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
+++ b/detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
@@ -4,6 +4,7 @@ import org.gradle.api.Project
 import org.gradle.testfixtures.ProjectBuilder
 import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.GradleRunner
+import org.intellij.lang.annotations.Language
 import java.io.File
 import java.nio.file.Files
 import java.util.UUID
@@ -13,6 +14,7 @@ class DslGradleRunner @Suppress("LongParameterList")
 constructor(
     val projectLayout: ProjectLayout,
     val buildFileName: String,
+    @Language("gradle.kts")
     val mainBuildFileContent: String = "",
     val configFileOrNone: String? = null,
     val baselineFiles: List<String> = emptyList(),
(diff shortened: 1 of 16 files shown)


detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
```diff
@@ -25,37 +27,40 @@ constructor(
     private val rootDir: File = Files.createTempDirectory("applyPlugin").toFile().apply { deleteOnExit() }
     private val randomString = UUID.randomUUID().toString()
 
+    @Language("gradle.kts")
```
cortinico, line 30: TIL you can do this with `"gradle.kts"` 👀

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/Helpers.kt
```diff
@@ -1,6 +1,32 @@
...
+ *             }
+ *         """.trimIndent()
+ *     }
+ * }
+ * ```
+ * In case the call site doesn't look like the above,
+ * set the [baseIndent] parameter to the number of indentation levels until the inside of the `"""`.
+ */
+fun String.reIndent(level: Int = 0, baseIndent: Int = 3): String =
+    this.replaceIndent("    ".repeat(baseIndent + level)).trim()
```
BraisGabin, line 29: Shouldn't it be better to use `trimStart`?

```suggestion
    this.replaceIndent("    ".repeat(baseIndent + level)).trimStart()
```

🥗😂

BraisGabin (approved):
👏👏👏 Good one! And the diff of those two files looks WAY WAY better.

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/Helpers.kt
TWiStErRob: I was thinking about that, but this way, there's definitely no opportunity for hacking, because even if you add multiple lines at start/end, they get removed so the results are consistent.

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
TWiStErRob: Funny thing: I only learned this 2 days ago while writing this PR too :)

Before that I knew `kotlin` and `gradle` were languages, but recently I was having chats about `.gradle.kts` files as convention plugins (which are way awesome btw), so while I was writing these annotations I just naturally wrote it, and it worked! 😅

PR #5838, opened 2023-02-26
New Rule: BracesOnWhenStatements with configurable parameters

VitalyVPinchuk:
For #5133 

Similar to #5700


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,254 @@
...
+            BracePolicy.Never -> "Extra braces exist on this branch, remove them."
+        }
+        report(CodeSmell(issue, Entity.from(reported), message))
+    }
+
+    enum class BracePolicy(val config: String) {
+        Always("always"),
+        Consistent("consistent"),
+        Necessary("necessary"),
+        Never("never");
```
BraisGabin, line 246: Does `never` have sense? On single line is the same as necessary. And on multi-line: does it forbids whens with multiple statements? There are not tests about that. Could we remove this option?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
```diff
@@ -0,0 +1,902 @@
...
+ *  * Numbers are made relative to the code snippet rather than the whole code passed to Kotlin compiler. [test].
+ *  * To debug a specific test case, remove all other test cases from the same method.
+ *  * Test coverage is added for all possible configuration options, see [NOT_RELEVANT].
+ */
+@Suppress(
+    "ClassName",
+    "LongMethod",
+    "CommentOverPrivateProperty"
+)
+class BracesOnWhenStatementsSpec {
```
BraisGabin, line 31: How many test cases are executed? How much time does them take?

Don't get me wrong. I like this! A lot! But maybe we should think about ways to don't create the same KtFile 5 times (on a different pr)

BraisGabin, line 31: Could a `\n` between cases mess this up?

```kotlin
1->println()

2->println()
```

BraisGabin (commented):
The tests are so easy to read!! 😍

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
VitalyVPinchuk: As we discussed with @TWiStErRob the meaning of `never` is to refactor multiple statements to a function call so that it could be made without braces. No braces are allowed whatsoever.

`necessary` allows braces on multiple statements only.

Yes, the tests are subject to improve.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
VitalyVPinchuk: If I get your point right and you're talking about sharing the same code across rules, we thought that it could be a better idea just to copy&paste it because rules might evolve separately and we don't want tight coupling.
There are about 200 tests now.

VitalyVPinchuk: I don't think so

BraisGabin: Sure! I was thinking about a hash map or something similar on the code that generates the KtFile. But this is out of the scope right now.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
BraisGabin: Oh, I just saw that we have the same on the `if` rule too. So this if fine to me. I'll talk in the issue about this and if then we decide to remove `never` it can be done later.

BraisGabin (approved):

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -505,6 +505,10 @@ style:
     active: false
     singleLine: 'never'
     multiLine: 'always'
+  BracesOnWhenStatements:
+    active: false
+    singleLine: 'consistent'
```
TWiStErRob, line 510: I would go with `necessary` as the default, for single-line whens I think the majority of people prefers:
```
when {
   x -> 1
   y -> 2
   z -> 3
}
```
and necessary would allow for `else -> { foo; bar }` if anyone ever writes that.

All the documentation also uses this style:
https://kotlinlang.org/docs/control-flow.html#when-expression

`consistent` is too allowing, I think we agree that:
```
when {
   x -> { 1 }
   y -> { 2 }
   z -> { 3 }
}
```
is consistent, but too noisy for no reason. This shouldn't be allowed by the default configuration.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
```diff
@@ -0,0 +1,902 @@
...
+                    """
+                        when (1) {
+                            1 -> { println() }
+                            2 -> { println() }
+                            else -> { println() }
+                        }
+                    """.trimIndent(),
+                    "{ println() }"(1),
+                    "{ println() }"(2),
+                    "{ println() }"(3),
```
TWiStErRob, line 234: I think ideally we would highlight the `->` (KtWhenEntry.arrow), because that always exists and it doesn't hide any other warnings. If we highlight the condition or the block any (potentially higher relevance) warnings inside would be hidden behind a big yellow block of code.

See https://github.com/detekt/detekt/issues/5316

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
```diff
@@ -0,0 +1,902 @@
...
+                            else -> println()
+                        }
+                    """.trimIndent(),
+                    "{ println() }"(1),
+                ),
+            )
+        }
+
+        @Nested
+        inner class `=necessary` {
```
TWiStErRob, line 203: Need tests for single-line multiple-statement.
Probably in all single-line `@Nested` tests classes there's a missing `fun`.

This will need some changes in the rule doc too, because
```
 *  // singleLine = 'necessary'
 *  when (a) {
 *      1 -> f1()
 *      2 -> f2()
 *  }
```
doesn't show the multi-statement possibility which is why `necessary` exists.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
```diff
@@ -0,0 +1,902 @@
...
+                                println()
+                            }
+                        }
+                    """.trimIndent(),
+                    "when"(1),
+                ),
+            )
+        }
+    }
+
```
TWiStErRob, line 791: some special-case tests are missing:
 * nested when inside when case block
 * nested when inside when condition
 * nested when inside when subject
 * weird curly formatting for multiline whens  
   (some of these might need to be tested for multiple configurations):
```
1 ->
{
  println()
}
2 -> { println()
}
3 -> {
println() }
4 ->
  { println() }
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,254 @@
...
+/**
+ * This rule detects `when` statements which do not comply with the specified policy.
+ * Keeping braces consistent will improve readability and avoid possible errors.
+ *
+ * Single-line `when` statement is:
+ * a `when` where each of the entries are single-line (has no line breaks `\n`).
+ *
+ * Multi-line `when` statement is:
+ * a `when` where at least one of the entries is multi-line (has a break line `\n`)
+ * or has multiple statements.
```
TWiStErRob, line 26: that doesn't make it multi-line: `1 -> { println(); println() }`

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,254 @@
...
+import org.jetbrains.kotlin.psi.KtBlockExpression
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtWhenExpression
+
+/**
+ * This rule detects `when` statements which do not comply with the specified policy.
+ * Keeping braces consistent will improve readability and avoid possible errors.
+ *
+ * Single-line `when` statement is:
+ * a `when` where each of the entries are single-line (has no line breaks `\n`).
```
TWiStErRob, line 22:
```suggestion
 * a `when` where each of the branches are single-line (has no line breaks `\n`).
```
let's use the official terminology:
https://kotlinlang.org/docs/control-flow.html#when-expression
"entry" is PSI, Kotlin devs call them branches.

Review all of the docs to make sure there are not entry/entries.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,254 @@
...
+ *
+ * Multi-line `when` statement is:
+ * a `when` where at least one of the entries is multi-line (has a break line `\n`)
+ * or has multiple statements.
+ *
+ * Available options are:
+ *  * `never`: forces no braces on any entry.
```
TWiStErRob, line 29:
```suggestion
 *  * `never`: forces no braces on any entry.  
      _Tip_: this is very strict, it will force a simple expression, like a single function call / expression. Extracting a function for "complex" logic is one way to adhere to this policy.
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,254 @@
...
+ *  * `necessary`: forces no braces on any entry except where necessary for multi-statement entries.
+ *  * `consistent`: ensures that braces are consistent within `when` statement.
+ *    If there are braces on one of the entry, all entries should have it.
```
TWiStErRob, line 32: English plural rules: you can only have "one of" something, if there are multiple to begin with.
```suggestion
 *    If there are braces on one of the entries, all entries should have it.
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,254 @@
...
+ *  }
+ *  // singleLine = 'necessary'
+ *  when (a) {
+ *      1 -> { f1() }
+ *      2 -> f2()
+ *  }
+ *  // multiLine = 'necessary'
+ *  when (a) {
+ *      1 -> { f1() }
+ *      2 -> { f2(); f3() }
```
TWiStErRob, line 56: I think this might be confusing, not relevant for the non-compliancy.

TWiStErRob: You can keep it simple with examples, `when`s don't need to have multiple branches, especially not in textual examples which will never be compiled.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,254 @@
...
+
+    override fun visitWhenExpression(expression: KtWhenExpression) {
+        super.visitWhenExpression(expression)
+
+        val branches: List<KtExpression> = walk(expression)
+        validate(branches, policy(expression))
+    }
+
+    private fun walk(expression: KtWhenExpression): List<KtExpression> {
+        return expression.entries.mapNotNull { it.expression }
```
TWiStErRob, line 178: wtf moment for PSI: when is this null?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    private fun validate(list: List<KtExpression>, policy: BracePolicy) {
```
TWiStErRob, line 181:
```suggestion
    private fun validate(branches: List<KtExpression>, policy: BracePolicy) {
```
should improve readability of the function.

(maybe rename in `if` rule too?)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
TWiStErRob: This will probably need to change the logic for `walk` to return `List<KtWhenEntry>` and all the consequences of that. In the `if` rule we were forced to model based on KtExpression, because the `if-elseif-else` is not part of the PSI. Here we have a really good representation of the real-world user-perceived structure, let's use it!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,254 @@
...
+        val entries = parent.entries.mapNotNull { it.expression }
+        val reported = when {
+            violator in entries && policy == BracePolicy.Consistent -> parent.whenKeyword
+            violator in entries -> violator
+            else -> error("Violating element (${violator.text}) is not part of this 'when' (${parent.text})")
+        }
+        val message = when (policy) {
+            BracePolicy.Always -> "Missing braces on this branch, add them."
+            BracePolicy.Consistent -> "Inconsistent braces, make sure all branches either have or don't have braces."
+            BracePolicy.Necessary -> "Extra braces exist on this branch, remove them (ignore multi-statement)."
```
TWiStErRob, line 236:
```suggestion
            BracePolicy.Necessary -> "Extra braces exist on this branch, remove them."
```
you can remove this, multi-statement for necessary shouldn't be flagged, right? That's the reason why `necessary` exists.

TWiStErRob (changes requested):
Nice one, a few changes are needed to make sense of consistent/multi-statement. But the majority is good! Thank you!

detekt-core/src/main/resources/default-detekt-config.yml
VitalyVPinchuk: Agree

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
VitalyVPinchuk: you're right

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
VitalyVPinchuk: Sorry I didn't get your point

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
VitalyVPinchuk: right

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
VitalyVPinchuk: Okay, but in code they call them `entries`.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
VitalyVPinchuk: `it.expression` type is `KtExpression?`

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
VitalyVPinchuk: Fixed

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
VitalyVPinchuk: What kind of `fun` is missing?

Fixed the docs adding multi-statement branches.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
VitalyVPinchuk: Added some test cases.
Could you share an example of "nested when inside when subject"?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
VitalyVPinchuk: Fixed

VitalyVPinchuk:
Also, I can't engage this rule in `detekt.yml` as it fails regardless of any parameter.

BraisGabin:
> Also, I can't engage this rule in `detekt.yml` as it fails regardless of any parameter.

I don't understand the your statment. Can you elavorate?

VitalyVPinchuk:
> > Also, I can't engage this rule in `detekt.yml` as it fails regardless of any parameter.
> 
> I don't understand the your statment. Can you elavorate?

I mean if I turn on this rule, detekt task fails. No matter what parameters I use in the rule: consistent, necessary, etc. We have to fix codebase a bit

BraisGabin:
Oh, sure. No problem about that. We can enable it later in other pr with the proper fixes.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
TWiStErRob: you already wrote a test: ``fun `nested when inside when condition`()``, just rename it.
For `when` a condition is before the arrows, the thing we're branching on in the parentheses is called a subject.
(https://kotlinlang.org/docs/control-flow.html#when-expression)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
```diff
@@ -0,0 +1,1090 @@
...
+                                println()
+                            }
+                        }
+                    """.trimIndent(),
+                    "when"(1),
+                ),
+            )
+
+            @TestFactory
+            fun `nested when inside when case block`() = listOf(
```
TWiStErRob, line 914: I think these 3 edge cases can be extracted to a top-level test (outside of `@Nested class`es. We're specifically testing the rule itself not its configurations.
```
            fun `nested when inside when case block`() = listOf(
            fun `nested when inside when condition`() = listOf(
            fun `weird curly formatting for multiline whens`() = listOf(
```
you can pick a configuration that actually flags one or more nested arrows to make sure it's working. `*NOTHING` could passing tests because the `super` call is missing!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
TWiStErRob: same with the non-compliant example above this one: ` *  // singleLine = 'necessary'`

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,253 @@
...
+        }
+        violators.forEach { report(it, policy) }
+    }
+
+    private fun KtWhenEntry.hasBraces(): Boolean = expression is KtBlockExpression
+
+    private fun KtWhenEntry.parentCandidate(): PsiElement? = this.parent
+
+    private fun KtWhenEntry.isMultiStatement(): Boolean =
+        expression.let { (it is KtBlockExpression) && (it.statements.size > 1) }
```
TWiStErRob, line 216:
```suggestion
        (expression is KtBlockExpression) && (expression.statements.size > 1)
```
it's actually shorter without `let`, and also easier to understand. I would also remove the parens, there's no need for clarification of where the conditions are, because there's no weird precedence or complex expression.:
```suggestion
        expression is KtBlockExpression && expression.statements.size > 1
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
TWiStErRob: ``fun `nested when inside when condition`()`` would be something like:
```
fun f(s: String?) {
when {
    when { s != null -> s.length; else -> 0 } > 5 -> true
    when(s) { "foo" -> 1; "bar" -> 1; else -> 2 } == 1 -> false
    else -> null
}
```

TWiStErRob: you can format these to make sure they would be flagged for braces, it'll look very strange, but it'll do the job

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,253 @@
...
+
+    private fun policy(expression: KtWhenExpression): BracePolicy {
+        val multiLineCandidate = expression.entries.firstOrNull { branch ->
+            branch.text.substringAfter("->", "").contains('\n')
```
TWiStErRob, line 220: in what test there was no arrow in the branch?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,253 @@
...
+                    emptyList()
+                }
+            }
+        }
+        violators.forEach { report(it, policy) }
+    }
+
+    private fun KtWhenEntry.hasBraces(): Boolean = expression is KtBlockExpression
+
+    private fun KtWhenEntry.parentCandidate(): PsiElement? = this.parent
```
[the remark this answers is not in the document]
VitalyVPinchuk: right

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,253 @@
...
+    private val multiLine: BracePolicy by config("consistent") { BracePolicy.getValue(it) }
+
+    override fun visitWhenExpression(expression: KtWhenExpression) {
+        super.visitWhenExpression(expression)
+
+        val branches: List<KtWhenEntry> = walk(expression)
+        validate(branches, policy(expression))
+    }
+
+    private fun walk(expression: KtWhenExpression) = expression.entries
```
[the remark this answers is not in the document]
VitalyVPinchuk: fixed

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
VitalyVPinchuk: But smart cast fails because of custom getter so we need `let` here. Or create a variable.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
VitalyVPinchuk: It's refactored ))

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,253 @@
...
+        return if (multiLineCandidate != null) multiLine else singleLine
+    }
+
+    private fun report(violator: KtWhenEntry, policy: BracePolicy) {
+        val parent = violator.parentCandidate() as KtWhenExpression
+        val reported = when {
+            violator in parent.entries && policy == BracePolicy.Consistent -> parent.whenKeyword
+            violator in parent.entries -> violator.arrow
+            else -> error("Violating element (${violator.text}) is not part of this 'when' (${parent.text})")
+        } ?: return
```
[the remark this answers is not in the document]
VitalyVPinchuk: No, it's because `Entity.from(reported)` doesn't want `reported` to be `null`. The reason for `null` is `else` branch with `error()`. And `detekt` task doesn't like `Entity.from(reported!!)` that's why I decided to add `elvis` with `return`.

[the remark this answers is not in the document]
TWiStErRob: Even the `reported` selection logic could be moved to the enum (similar to `message` below), but that's a bit trickier:
```kotlin
    enum class BracePolicy(val config: String) {
        Always("always") {
            override fun report(violator: KtWhenEntry): PsiElement =
                requireNotNull(violator.arrow) { "When branch ${violator.text} has no arrow!" }
        },
        Consistent("consistent") {
            override fun report(violator: KtWhenEntry): PsiElement =
                (violator.parent as KtWhenExpression).whenKeyword
        },
        Necessary("necessary") {
            override fun report(violator: KtWhenEntry): PsiElement =
                requireNotNull(violator.arrow) { "When branch ${violator.text} has no arrow!" }
        },
        Never("never") {
            override fun report(violator: KtWhenEntry): PsiElement =
                requireNotNull(violator.arrow) { "When branch ${violator.text} has no arrow!" }
        },
        ;
        abstract fun report(violator: KtWhenEntry): PsiElement
```
This results in duplicate code for the arrow, and way way more lines for the win of abstraction? Probably not worthy.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
TWiStErRob: This is a good point, add at least a `specialTestCases` for it.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
TWiStErRob: Tests look good, but single-line necessary examples in kdoc still don't show the use case for necessary.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,247 @@
...
+ *      1 -> { f1() }
+ *      2 -> f2()
+ *  }
+ *  // multiLine = 'never'
+ *  when (a) {
+ *      1 -> {
+ *          f1()
+ *      }
+ *      2 -> f2()
+ *  }
```
TWiStErRob, line 48: For the lack of highlighting on the website, maybe we should add comments to show which part is non-compliant, because it's quite the brain-work to deduce based on the rules/options/context what's wrong with these. For example for these first two:

```suggestion
 *  // singleLine = 'never'
 *  when (a) {
 *      1 -> { f1() } // unnecessary braces
 *      2 -> f2()
 *  }
 *  // multiLine = 'never'
 *  when (a) {
 *      1 -> { // unnecessary braces
 *          f1()
 *      }
 *      2 -> f2()
 *  }
```

TWiStErRob (commented):
Sorry for the delay, move was longer than expected. Thios looks good now, just some final touches / simplifications. Happy to help finishing up, find me on Slack. Please look through all the unresolved conversations on the PR, there might be some hiding behind expanders.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
VitalyVPinchuk: Done

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
VitalyVPinchuk: Removed

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
```diff
@@ -0,0 +1,999 @@
...
+                        when (1) {
+                            1 -> println()
+                            2 -> println()
+                            else -> println()
+                        }
+                    """.trimIndent(),
+                    *NOTHING,
+                ),
+            )
+
```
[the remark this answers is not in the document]
VitalyVPinchuk: You're right. Fixed

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,247 @@
...
+            violator in parent.entries && policy == BracePolicy.Consistent -> parent.whenKeyword
+            violator in parent.entries -> violator.arrow
+            else -> error("Violating element (${violator.text}) is not part of this 'when' (${parent.text})")
+        } ?: return
+        val message = when (policy) {
+            BracePolicy.Always -> "Missing braces on this branch, add them."
+            BracePolicy.Consistent -> "Inconsistent braces, make sure all branches either have or don't have braces."
+            BracePolicy.Necessary -> "Extra braces exist on this branch, remove them."
+            BracePolicy.Never -> "Extra braces exist on this branch, remove them."
+        }
```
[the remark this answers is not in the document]
VitalyVPinchuk: Good idea!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
VitalyVPinchuk: Fixed

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
```diff
@@ -0,0 +1,999 @@
...
+        }
+
+        @Nested
+        inner class `=consistent` {
+
+            private fun flag(code: String, vararg locations: (String) -> Pair<Int, Int>) =
+                testCombinations(NOT_RELEVANT, BracePolicy.Consistent.config, code, *locations)
+
+            @TestFactory
+            fun `no braces are accepted`() = listOf(
```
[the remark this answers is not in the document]
VitalyVPinchuk: Fixed

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
```diff
@@ -0,0 +1,999 @@
...
+                        }
+                        2 -> { println()
+                        }
+                        3 -> { 
+                        println() }
+                        else -> 
+                            { println() }
+                    }
+                """.trimIndent(),
+                *NOTHING,
```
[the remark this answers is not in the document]
VitalyVPinchuk: Should anything be done about it?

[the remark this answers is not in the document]
TWiStErRob: I think it's ok, to each their own style :)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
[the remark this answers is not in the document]
TWiStErRob: I think this got lost in all the messages, I think we should use this last version above with clear behavior for each policy, but not moved to enum.
https://github.com/detekt/detekt/pull/5838#discussion_r1158353012

The simplest and most expressive version, and it would make Codecov very happy too 😌.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,247 @@
...
+
+    private fun KtWhenEntry.isMultiStatement(): Boolean =
+        expression.let { (it is KtBlockExpression) && (it.statements.size > 1) }
+
+    private fun policy(expression: KtWhenExpression): BracePolicy {
+        val isMultiLine = expression.entries.any { branch ->
+            branch.arrow
+                ?.siblings(forward = true, withItself = false)
+                ?.any { it.textContains('\n') }
+                ?: false
```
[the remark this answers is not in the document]
TWiStErRob: Thoughts?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
TWiStErRob: Can you please go through all the `<noncompliant>` snippets in the kdoc and add comments?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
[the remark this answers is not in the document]
VitalyVPinchuk: 100% agree

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatements.kt
```diff
@@ -0,0 +1,242 @@
...
+ *  // singleLine = 'necessary'
+ *  when (a) {
+ *      1 -> f1()
+ *      2 -> { f2(); f3() } // Necessary braces because of multiple statements.
+ *  }
+ *  // multiLine = 'necessary'
+ *  when (a) {
+ *      1 ->
+ *          f1()
+ *      2 -> { // // Necessary braces because of multiple statements.
```
TWiStErRob, line 115:
```suggestion
 *      2 -> { // Necessary braces because of multiple statements.
```

PR #5835, opened 2023-02-25
Exclude super call when generating guard clauses

atulgpt:
Fixes https://github.com/detekt/detekt/issues/5680

PR #5841, opened 2023-02-26
Remove unnecessary `@Suppress("ReturnCount")`

BraisGabin:
This is doable thanks to #5835

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitType.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitType.kt
@@ -58,8 +58,7 @@ class RedundantExplicitType(config: Config) : Rule(config) {
     override fun visitProperty(property: KtProperty) {
         if (!property.isLocal) return
         val typeReference = property.typeReference ?: return
-        val type =
-            (bindingContext[BindingContext.VARIABLE, property])?.type ?: return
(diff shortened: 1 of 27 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitType.kt
```diff
@@ -58,8 +58,7 @@ class RedundantExplicitType(config: Config) : Rule(config) {
     override fun visitProperty(property: KtProperty) {
         if (!property.isLocal) return
         val typeReference = property.typeReference ?: return
-        val type =
-            (bindingContext[BindingContext.VARIABLE, property])?.type ?: return
+        val type = (bindingContext[BindingContext.VARIABLE, property])?.type ?: return
```
3flex, line 61: Should the ReturnCount suppression be removed from this function now? This change looks unrelated if not - also I'm curious if this had to be reformatted so the suppression can be removed?

3flex (approved):

3flex:
Nice!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantExplicitType.kt
BraisGabin: This is just a Boy scout change. It has not relation with what this rule archives.

PR #5844, opened 2023-02-26
Simplify generate documentation

BraisGabin:
A bit of love for `generateDocumentation`. I think that this PR reviewed commit by commit should be easier.

Basically what I did:
- Remove dependencies with other tasks
- Don't hardcode paths, use gradle
- Reduce verbosity
- Some other minor Boy Scouts


cortinico (approved):
LGTM 👍

PR #5846, opened 2023-02-27
Remove dependency between check and jacocoMergedReport

BraisGabin:
`jacocoMergedReport` is in the critical path of my builds: 
https://ge.detekt.dev/s/24zl4n3tj746e/timeline?details=j2hxxhlnrx3wg&show-only-critical-path
https://ge.detekt.dev/s/ok5dejkm7eo34/timeline?details=j2hxxhlnrx3wg&show-only-critical-path

But its output is not relevant. The only place where it is relevant is in the workflow `codecoverage.yaml`. But there we don't execute it using `check` we execute it using its name.

### DIFF
--- a/code-coverage-report/build.gradle.kts
+++ b/code-coverage-report/build.gradle.kts
@@ -44,10 +44,6 @@ dependencies {
     jacocoAggregation(projects.detektUtils)
 }
 
-tasks.check {
-    dependsOn(tasks.named("jacocoMergedReport"))
-}
-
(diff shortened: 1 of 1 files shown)


3flex:
> its output is not relevant

Just FYI there's an aggregated HTML report generated by the report aggregation task, so it's not only used by Codecov.

But reading the definition of `check` from Gradle it says that verification tasks should be attached, but this isn't a verification task as such, so it makes sense to drop it.

3flex (approved):

PR #5852, opened 2023-02-27
Tweak GHA configs a bit

Goooler:
https://github.com/gradle/gradle-build-action/blob/main/README.md#removing-unused-files-from-gradle-user-home-before-saving-to-cache
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - '**'
+  workflow_dispatch:
 
 env:
   GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -20,8 +20,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        jdk: [8, 11, 17]
+        os: [ ubuntu-latest, macos-latest, windows-latest ]
+        jdk: [ 8, 11, 17 ]
         exclude: # windows with JDK8 are *really* flaky
           - os: windows-latest
             jdk: 8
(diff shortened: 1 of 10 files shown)


.github/workflows/pre-merge.yaml
```diff
@@ -5,8 +5,7 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - '**'
+  workflow_dispatch:
```
BraisGabin, line 8: Why do we want this line? Which is the use case?

Goooler: '**' matches any branch? It's the default value.

BraisGabin: I mean `workflow_dispatch:`. Why do we want to launch it manually?

Goooler: We can run pre-merge jobs on any branch, easy to test changes before sending PR.

.github/workflows/website.yaml
```diff
@@ -38,6 +36,7 @@ jobs:
       - name: Build Detekt Documentation
         uses: gradle/gradle-build-action@6095a76664413da4c8c134ee32e8a8ae900f0f1f # v2
         with:
+          gradle-home-cache-cleanup: true
```
BraisGabin, line 39: I'm not sure why this is not their default value.

If in one action we execute just `help` and in other `build`. We'll the `help` action remove a lot of files that are used by the `build` one? My point is, do we know if this is always a net win? Does it have side effects?

I assume that the desired side effect is that if we would like to roll back a gradle upgrade we would need to download some things again. But that is completely reasonable and unlikely to happen. But are there others?

My fear is that this would create far too many misses.

Goooler: I enable this flag on our internal repos, cause the cache size grow up to about 2~3 GB, spend too mang time on restoring cache. Either way, I can revert these lines if you want.

cortinico: > My fear is that this would create far too many misses.

Let's merge it for now, and report back to Gradle if this degrading the GE experience (it's their custom GH Action afterall)

cortinico (approved):

BraisGabin:
Thanks for the cleanup :)

PR #5851, opened 2023-02-27
MatchingDeclarationName. KtFilesSpec also remove .common.kt suffix from kotlin files

lexa-diky:
No changes for rule itself, but MatchingDeclarationName now will threat `.common.kt` suffix as known kotlin file part.

PR #5849, opened 2023-02-27
Create `generateWebsite`

BraisGabin:
`generateDocumentation` is a god task. It generates far too much things:

- Dokka/javadoc
- Default-config
- Config for first party rule sets
- Deprecation list
- Rule documentation

I want to split it. This is the just the first step. I created `generateWebsite`. This task generates the necessary files to create our website. This way when we execute `generateDocumentation` we don't need to run `dokkaHtml`. ~Also, we shouldn't know on `detekt-api` where the documentation should be stored. For that reason I removed it from there and I copy the output of that task to the correct directory.~

As I said, this is just the first step. A quick win.

PR #5853, opened 2023-02-27
Improve `generateDocumentation`

BraisGabin:
Related with #5851

`detekt-cli` should create it's own documentation and `detekt-generator` should consume that and copy it to the correct place inside `website`. That's what this PR does.

It also makes `generateDocumentation` way faster because it no longer depend on all these taks:

```
:detekt-cli:jar
:detekt-core:classes
:detekt-core:compileJava
:detekt-core:compileKotlin
:detekt-core:jar
:detekt-core:processResources
:detekt-metrics:classes
:detekt-metrics:compileJava
:detekt-metrics:compileKotlin
:detekt-metrics:jar
:detekt-metrics:processResources
:detekt-report-html:classes
:detekt-report-html:compileJava
:detekt-report-html:compileKotlin
:detekt-report-html:jar
:detekt-report-html:processResources
:detekt-report-md:classes
:detekt-report-md:compileJava
:detekt-report-md:compileKotlin
:detekt-report-md:jar
:detekt-report-md:processResources
:detekt-report-sarif:classes
:detekt-report-sarif:compileJava
:detekt-report-sarif:compileKotlin
:detekt-report-sarif:jar
:detekt-report-sarif:processResources
:detekt-report-txt:classes
:detekt-report-txt:compileJava
:detekt-report-txt:compileKotlin
:detekt-report-txt:jar
:detekt-report-txt:processResources
:detekt-report-xml:classes
:detekt-report-xml:compileJava
:detekt-report-xml:compileKotlin
:detekt-report-xml:jar
:detekt-report-xml:processResources
:detekt-rules-complexity:classes
:detekt-rules-complexity:compileJava
:detekt-rules-complexity:compileKotlin
:detekt-rules-complexity:jar
:detekt-rules-complexity:processResources
:detekt-rules-coroutines:classes
:detekt-rules-coroutines:compileJava
:detekt-rules-coroutines:compileKotlin
:detekt-rules-coroutines:jar
:detekt-rules-coroutines:processResources
:detekt-rules-documentation:classes
:detekt-rules-documentation:compileJava
:detekt-rules-documentation:compileKotlin
:detekt-rules-documentation:jar
:detekt-rules-documentation:processResources
:detekt-rules-errorprone:classes
:detekt-rules-errorprone:compileJava
:detekt-rules-errorprone:compileKotlin
:detekt-rules-errorprone:jar
:detekt-rules-errorprone:processResources
:detekt-rules-exceptions:classes
:detekt-rules-exceptions:compileJava
:detekt-rules-exceptions:compileKotlin
:detekt-rules-exceptions:jar
:detekt-rules-exceptions:processResources
:detekt-rules-naming:classes
:detekt-rules-naming:compileJava
:detekt-rules-naming:compileKotlin
:detekt-rules-naming:jar
:detekt-rules-naming:processResources
:detekt-rules-performance:classes
:detekt-rules-performance:compileJava
:detekt-rules-performance:compileKotlin
:detekt-rules-performance:jar
:detekt-rules-performance:processResources
:detekt-rules-style:classes
:detekt-rules-style:compileJava
:detekt-rules-style:compileKotlin
:detekt-rules-style:jar
:detekt-rules-style:processResources
:detekt-rules:classes
:detekt-rules:compileJava
:detekt-rules:compileKotlin
:detekt-rules:jar
:detekt-rules:processResources
:detekt-tooling:classes
:detekt-tooling:jar
:detekt-tooling:processResources
```


detekt-cli/build.gradle.kts
```diff
@@ -74,3 +74,27 @@ tasks {
...
+        configurations.compileClasspath,
+        sourceSets.main.map { it.output },
+    )
+    mainClass.set("io.gitlab.arturbosch.detekt.cli.CliUsage")
+    args(cliUsagesOutput.get())
+}
+
+private val generatedCliUsage: Configuration by configurations.creating {
+    isCanBeConsumed = true
+    isCanBeResolved = false
```
cortinico, line 93: Do you need to specify both?

detekt-generator/build.gradle.kts
```diff
@@ -3,24 +3,31 @@ plugins {
     id("module")
 }
 
+val generatedUsage: Configuration by configurations.creating {
```
cortinico, line 6: Same as above. I'm unsure we need extra Gradle configurations to achieve this.

BraisGabin: I think so but I'm not sure at all. I would like to know if there is a less verbose option. I'm kind of noobie in the "gradle realm". This solutions is what I understood from this comment by @3flex https://github.com/detekt/detekt/pull/5849#issuecomment-1445629941. I can't depend directly on the output on the output of `:detekt-cli:cliUsage`. I need to declare an artifact inside `detekt-cli` and consume it inside `detekt-generator`.

If there is an easy/less verbose way to do that I would love to know about it.

BraisGabin: @3flex could you help us here?

detekt-cli/build.gradle.kts
3flex: They're both `true` by default so technically only `isCanBeResolved` needs to be set.

BraisGabin:
I'm still waiting for some feedback on this one. I agree that my solution is verbose but I would like to know if there is any other way to do this with gradle.

detekt-cli/build.gradle.kts
```diff
@@ -74,3 +74,27 @@ tasks {
...
+private val cliUsage by tasks.registering(JavaExec::class) {
+    outputs.file(cliUsagesOutput)
+
+    classpath(
+        configurations.compileClasspath,
+        sourceSets.main.map { it.output },
+    )
+    mainClass.set("io.gitlab.arturbosch.detekt.cli.CliUsage")
+    args(cliUsagesOutput.get())
+}
```
TWiStErRob, line 89: Could we reuse the existing CLI?

something along the lines of
```kotlin
private val cliUsage by tasks.registering(JavaExec::class) {
    outputs.file(cliUsagesOutput)
    classpath = files(shadowJar)
    args = listOf("--help")
    standardOutput = ByteArrayOutputStream()
    doLast {
        cliUsagesOutput.get().apply {
            writeText("") // Clear.
            appendText("```\n")
            appendBytes(standardOutput.toByteArray())
            appendText("```\n")
        }
    }
}
```

detekt-cli/build.gradle.kts
```diff
@@ -74,3 +74,27 @@ tasks {
...
+
+private val cliUsagesOutput = layout.buildDirectory.dir("output/cli-usage.md")
+
+private val cliUsage by tasks.registering(JavaExec::class) {
+    outputs.file(cliUsagesOutput)
+
+    classpath(
+        configurations.compileClasspath,
+        sourceSets.main.map { it.output },
+    )
```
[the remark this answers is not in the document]
TWiStErRob: Never mind, probably even better `classpath = files(shadowJar)`, so the output/env is exactly the same as what the user sees/has.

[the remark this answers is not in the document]
BraisGabin: I got it working with this:

```kotlin

private val cliUsagesOutput = layout.buildDirectory.file("output/cli-usage.md")

private val cliUsage by tasks.registering(JavaExec::class) {
    outputs.file(cliUsagesOutput)
    classpath = files(tasks.shadowJar)
    args = listOf("--help")
    val outputStream = ByteArrayOutputStream()
    standardOutput = outputStream
    doLast {
        cliUsagesOutput.get().asFile.apply {
            writeText("```\n")
            appendBytes(outputStream.toByteArray())
            appendText("```\n")
        }
    }
}
```

The problem is that ByteArrayOutputStream is not compatible with configuration cache: https://ge.detekt.dev/s/al3fm3jtb3glk/failure#1

detekt-cli/build.gradle.kts
BraisGabin: I got it working with this:

```kotlin

private val cliUsagesOutput = layout.buildDirectory.file("output/cli-usage.md")

private val cliUsage by tasks.registering(JavaExec::class) {
    outputs.file(cliUsagesOutput)
    classpath = files(tasks.shadowJar)
    args = listOf("--help")
    val outputStream = ByteArrayOutputStream()
    standardOutput = outputStream
    doLast {
        cliUsagesOutput.get().asFile.apply {
            writeText("```\n")
            appendBytes(outputStream.toByteArray())
            appendText("```\n")
        }
    }
}
```

The problem is that ByteArrayOutputStream is not compatible with configuration cache: https://ge.detekt.dev/s/al3fm3jtb3glk/failure#1

TWiStErRob: Oh, that's a good point. I have an idea, let me try it quick.

detekt-cli/build.gradle.kts
```diff
@@ -74,3 +74,27 @@ tasks {
         dependsOn(runWithHelpFlag, runWithArgsFile)
     }
 }
+
+private val cliUsagesOutput = layout.buildDirectory.dir("output/cli-usage.md")
+
+private val cliUsage by tasks.registering(JavaExec::class) {
```
TWiStErRob, line 80: how are you able to use `private`? I keep getting
```
  Line 80:     private val cliUsagesOutput = ...
               ^ Modifier 'private' is not applicable to 'local variable'
```

TWiStErRob: Ah, this is outside the tasks block... you can move it inside, and in the artifacts you _should be_ able to reference just the task, not the file directly (because there's only one output file of this task).

detekt-cli/build.gradle.kts
TWiStErRob: this doesn't fail and works with/without config cache:
```kotlin
    val cliUsage by registering(JavaExec::class) {
        val cliUsagesOutput = layout.buildDirectory.file("output/cli-usage.md")
        outputs.file(cliUsagesOutput)
        classpath = files(shadowJar)
        args = listOf("--help")
        doFirst {
            standardOutput = ByteArrayOutputStream()
        }
        doLast {
            cliUsagesOutput.get().asFile.apply {
                writeText("```\n")
                appendBytes((standardOutput as ByteArrayOutputStream).toByteArray())
                appendText("```\n")
            }
        }
    }
```

detekt-cli/build.gradle.kts
TWiStErRob: Example: https://github.com/detekt/detekt/pull/6337/files#diff-82ef80a7aac6b0909be0be4077fc848df6e2dcf24be46ff0ee10a7362210e5c5R96

BraisGabin: Thanks for the link of that notation! It looks way better. But even with that I think that it works better as a top level. If I declare it inside `tasks` I need to add `@Suppress("UNUSED_VARIABLE")`. And then, in the artifacts I need to use `artifacts.add(generatedCliUsage.name, tasks.named("cliUsage"))`.

detekt-cli/build.gradle.kts
BraisGabin: Thanks!! That works as a charm :)

detekt-generator/build.gradle.kts
BraisGabin: `consumeGeneratedConfig` doesn't feel like the correct function to use here. As this is not related with configuration. I'll check tomorrow to create a `consumerGeneratedDocumentation` or something similar and see how it looks. But I agree that some refactor could improve this code.

TWiStErRob: Ah, config vs configuration 😅. Have a look if it's reusable, maybe we can simply rename it because seemingly there's no "config" related code in it other than the hardcoded name of the configuration which can be also changed.

detekt-cli/build.gradle.kts
TWiStErRob: You could add the artifact inside the tasks too right after the task declaration and use the "local" variable instead of named(). Also we could just not use the `val by` syntax: `register<JavaExec>("name") { }` does the same thing. You could even totally remove the reference by doing: `artifacts.add(generatedCliUsage.name, tasks.register(...) {...})`. I think the simplest and most readable is to move the `artifacts.add` inside the block.

TWiStErRob: Note: we can also ignore the warning, it should be "fixed" soon:
https://youtrack.jetbrains.com/issue/KTIJ-969

detekt-generator/build.gradle.kts
BraisGabin: I wasn't able to make it work with that function. The `Copy` task doesn't seems to work well with how this function setups the inputs. I agree that this code is a bit verbose and that we could extract it to functions. But this is the first one so create an abstraction is difficult because I have no idea how the next one will look like.

I vote to merge this as it is noew because it is the "gradle way". This was the first step. The next step would be to make each ruleset to generate its own documentation and consume it from this module. Then we will have more examples so we can create that abstraction.

TWiStErRob: Okay. Thanks for trying, good point about having more.

detekt-cli/build.gradle.kts
BraisGabin: I moved the task inside `tasks` on the last commit. What do you think? I'm not sure which one I prefer.

TWiStErRob (approved):
It looks pretty neat in its final form 🎉:
 * A task that uses production code as a user would
 * A configuration to expose the artifact
 * A configuration to consume the artifact
 * A task to put it in its final place
 * \+ all the removed code 🧹

detekt-cli/build.gradle.kts
```diff
@@ -87,4 +88,28 @@ tasks {
...
+                appendBytes((standardOutput as ByteArrayOutputStream).toByteArray())
+                appendText("```\n")
+            }
+        }
+    }
+
+    artifacts.add(generatedCliUsage.name, cliUsage)
+}
+
+private val generatedCliUsage: Configuration by configurations.creating {
```
TWiStErRob, line 112: Huh, this forward usage works? 😮 Didn't think this will happen.

TWiStErRob: Yeah, it doesn't:

> Parameter specified as non-null is null: method org.gradle.kotlin.dsl.NamedDomainObjectCollectionExtensionsKt.getValue, parameter <this>

TWiStErRob: Move the configuration before the tasks. That's what the order is usually roughly: config, deps, exts, tasks

detekt-cli/build.gradle.kts
TWiStErRob: The task is registered with all the other tasks, I think that's a big plus, no surprises.

I consider this resolved, let's continue here: https://github.com/detekt/detekt/pull/5853/files#r1285265216

detekt-generator/build.gradle.kts
```diff
@@ -3,30 +3,43 @@ plugins {
...
 val configDir = "$rootDir/detekt-core/src/main/resources"
-val cliOptionsFile = "$rootDir/website/docs/gettingstarted/_cli-options.md"
 val defaultConfigFile = "$configDir/default-detekt-config.yml"
 val deprecationFile = "$configDir/deprecation.properties"
 val formattingConfigFile = "$rootDir/detekt-formatting/src/main/resources/config/config.yml"
 val librariesConfigFile = "$rootDir/detekt-rules-libraries/src/main/resources/config/config.yml"
 val ruleauthorsConfigFile = "$rootDir/detekt-rules-ruleauthors/src/main/resources/config/config.yml"
 
+val copyDetektCliUsage by tasks.registering(Copy::class) {
+    from(generatedUsage) { rename { "_cli-options.md" } }
```
TWiStErRob, line 36: What this underscore doing?

BraisGabin: No idea, it was like this before: https://github.com/detekt/detekt/pull/5853/files#diff-e82b2d14a65b97fbbbff582f3413e3137477659cc9189e9e50e9516ffde243ffL21

detekt-generator/build.gradle.kts
3flex: > The next step would be to make each ruleset to generate its own documentation

I've started on this

TWiStErRob:
@3flex what do you think of this one?

PR #5856, opened 2023-02-27
Dont depend on jcommander

BraisGabin:
Only `Main` should know about `GeneratorArgs`. The rest of the code shouldn't know about `jcommander`.


cortinico (approved):
Nice one 👍

cortinico:
Removing the milestone as I don't believe this is release blocking right?

PR #5867, opened 2023-03-01
allow for float config values

Tommyten:
This PR allows config values using the `@Configuration` annotation and the `by config` delegates to have `float` values.

Float config values can be helpful when using fractions as threshold values, like in #5581 where a fraction (1/3) is used as a config value.


cortinico (changes requested):
Tests are missing in `ConfigPropertySpec.kt` and `RuleCollectorSpec.kt`. Please add them

PR #5870, opened 2023-03-02
Split UnnecessaryAbstractClass into two rules

### LINKED ISSUE #5778: UnnecessaryAbstractClass: add option to only detect interface, not concrete class migrations
## Current Behavior

UnnecessaryAbstractClass currently serves a double duty:

1. Warn about abstract classes that could be converted to interfaces
2. Warn about abstract classes that could be converted to concrete classes

## Expected Behavior

Split the rule  into two rules or add an option to disable either sub-rule.

## Context

Second rule (converting abstract classes to concrete classes) seems a bit counterintuitive to me. Often I want to mark base classes as abstract, even if they do not have abstract methods, to ensure nobody accidentally constructs them (since they are useless on their own).

With this rule enabled, such abstract classes would get flagged. If I disable the rule, I would also disable the interface part, which I find useful. I could manually suppress every such class, but that would result in a mess.

matejdro:
PR mostly copies UnnecessaryAbstractClass code into a new rule (ClassCanBeInterface) and then removes the overlap between two rules.

I had to add `excludeAnnotatedClasses` to the new rule, because tests depend on being able to exclude classes while directly calling rule. From what I can see, there is no way to apply `ignoreAnnotated` for tests? I would appreciate some feedback on how to avoid this.

this fixes #5778


cortinico:
Can we make the CI green before we review this?

matejdro:
CI is now green.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassCanBeInterface.kt
```diff
@@ -0,0 +1,126 @@
...
+import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.psiUtil.isAbstract
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.lazy.descriptors.LazyClassMemberScope
+import org.jetbrains.kotlin.types.typeUtil.isInterface
+
+/**
+ * This rule inspects `abstract` classes. In case an `abstract class` does not define any
+ * `abstract` members, it should instead be refactored into concrete classes.
```
BraisGabin, line 31: This descripction is not correct for this rule, right? If all the members are abstract it should be an interface.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassCanBeInterface.kt
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * <noncompliant>
+ * abstract class OnlyAbstractMembersInAbstractClass { // violation: no concrete members
+ *
+ *     abstract val i: Int
+ *     abstract fun f()
+ * }
+ * </noncompliant>
```
BraisGabin, line 39: nit: Could you add a compliant to show how this code should be refactored? (this comment doesn't block the pr because it /this wasn't in the original rule)

BraisGabin (commented):
Sorry for the late, late review. I think there is an issue regarding the description. Once that is fixed it is good to go in my opinion.

BraisGabin:
Can you run `generateDocumentation` and commit the changes that it will make? That would make CI happy. Apart from that LGTM.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassCanBeInterface.kt
```diff
@@ -0,0 +1,140 @@
...
+ * <compliant>
+ * interface Interface {
+ *     val i: Int
+ *     fun f()
+ * }
+ * </compliant>
+ * <compliant>
```
BraisGabin, line 46:
```suggestion
 * 
```

BraisGabin (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassCanBeInterface.kt
```diff
@@ -0,0 +1,140 @@
...
+            Severity.Style,
+            "An abstract class is unnecessary. May be refactored to an interface.",
+            Debt.FIVE_MINS
+        )
+
+    @Configuration("Allows you to provide a list of annotations that disable this check.")
+    @Deprecated("Use `ignoreAnnotated` instead")
+    private val excludeAnnotatedClasses: List<Regex> by config(emptyList<String>()) { list ->
+        list.map { it.replace(".", "\\.").replace("*", ".*").toRegex() }
+    }
```
cortinico, line 73: If this is a new rule, why are you adding this deprecated property? I would just remove it as no one would be using it anyway no?

matejdro: Tests depend on this property and I could not find a way to provide a new property to this tests. Should I just remove these tests?

cortinico: I'm not sure I follow, which tests are depending on this? The tests for `ClassCanBeInterface.kt` needs to be updated as well

matejdro: It uses this property to filter out items that tests should not match, see `We don't care about this first class` inside tests. I can refactor tests to remove this, but they would not be as clean.

matejdro: Removed the property in latest config.

What should I update in `ClassCanBeInterface.kt` tests?

matejdro:
I ran `generateDocumentation` and it did not create any changes. Is that okay?

matejdro:
It seems like one check failed due to kotlin daemon crashing:

> java.lang.RuntimeException: Could not connect to Kotlin compile daemon

Can someone restart this check, please?

matejdro:
It seems like codecov failed on some null checks that are not tested. Do you want me to update the tests? Although I'm not even sure how the source code needs to look to result in all of those things being null

cortinico:
@matejdro I'm fine with the code 👍 and the codecov failure can be ignored

matejdro:
Renamed rules.

cortinico:
Thanks for doing it @matejdro 

> Can we rename the two rules to:

@detekt/maintainers can I get an thumbs up that we're fine with the new names?

BraisGabin (approved):

BraisGabin:
LGTM but why are we getting this issue now: https://github.com/detekt/detekt/actions/runs/5832481518/job/15817993033?pr=5870#step:4:280

That rule shouldn't exist now, right?

TWiStErRob:
@BraisGabin chicken and egg

https://github.com/detekt/detekt/blob/main/detekt-gradle-plugin/build.gradle.kts#L15

BraisGabin:
@matejdro could you make CI happy? It seems that we can't change some of the `@Suppress` yet for the chickend and egg problem that TWiStErRob points out. We will fix that once we update the dependency.

matejdro:
I am a bit confused here. It seems like some modules use prepublished detekt, while others use current version? This might be a good idea to improve on in the future as it's confusing.

matejdro:
It looks like mac 11 check failed randomly, can someone restart it please?

cortinico (approved):

cortinico:
@matejdro can you rebase and solve the merge conflict?

Issue #1999, opened 2019-10-08
Cannot Suppress  CommentSpacing, NoMultipleSpaces


radzio:
## Expected Behavior
When use `@Suppress("NoTrailingSpaces", "CommentSpacing", "MaxLineLength", "NoMultipleSpaces")` for class all issues "NoTrailingSpaces", "CommentSpacing", "MaxLineLength", "NoMultipleSpaces" should be suppressed

## Observed Behavior

- `NoMultipleSpaces`
- `CommentSpacing`

Are still being reported

## Steps to Reproduce

Code to test

```
@Suppress("NoTrailingSpaces", "CommentSpacing", "MaxLineLength", "NoMultipleSpaces")
class SomeTestClass{

    @ParameterizedTest
    @CsvSource(value = [
        "2019-09-25T10:15:30, , false",                     // some comment
        "2019-09-26T10:15:30, , true",                      // some comment
        "2019-09-27T10:15:30, 2019-09-26T10:15:30, false",  // some comment
        "2019-10-03T04:15:30, 2019-10-02T03:17:30, true",   // some comment
        "2019-10-03T06:12:30, 2019-10-03T04:15:30, false",  // some comment
        "2019-10-15T06:12:30, 2019-10-03T04:15:30, true",   // some comment
        "2019-10-15T06:15:30, 2019-10-15T06:12:30, false",  // some comment
        "2019-10-16T06:12:30, 2019-10-15T04:15:30, true",   // some comment
        "2019-10-17T06:12:30, 2019-10-15T04:15:30, true",   // some comment
        "2019-10-17T06:12:30, , true",                      // some comment
        "2019-10-17T06:12:30, 2019-10-16T04:15:30, false"   // some comment
    ])
    fun someTestMethod(a: String, b: String?, c: Boolean) {
    }
}
```

## Context
I wanted to add some spaces to improve readability. 

## Your Environment

* Version of detekt used: 1.0.1 and 1.1.0
* Version of Gradle used (if applicable): 5.5.1
* Operating System and version: Windows 10
* Link to your project (if it's a public repository):

elijah-pl:
I spent some time trying to set up test for this issue (can be found [here](https://github.com/elijah-pl/detekt/commit/c2a989dc993a2741185b6bea6c34b9ca64a34bbd)), but it seems like I am missing something.

During the debug session, it seems like `annotationEntries` can not be found in the compiled `KtFile`.

```kt
[...]

PR #5876, opened 2023-03-03
Add support for local suppression inside formatting

### LINKED ISSUE #5547: `EnumEntryNameCase` cannot be suppressed on (enum) class or declaration level
## Expected Behavior

I should be able to suppress `EnumEntryNameCase` via

```
@Suppress("EnumEntryNameCase")
enum class SpdxLicense(
    val id: String,
    val fullName: String,
    val deprecated: Boolean = false
) {
    _0BSD("0BSD", "BSD Zero Clause License");
}
```

or

```
enum class SpdxLicense(
    val id: String,
    val fullName: String,
    val deprecated: Boolean = false
) {
    @Suppress("EnumEntryNameCase")
    _0BSD("0BSD", "BSD Zero Clause License");
}
```

## Observed Behavior

Suppressing this way has no effect, the inspection hint is already being reporter. However, suppressing on the *file*-level works:

```
@file:Suppress("EnumEntryNameCase")

enum class SpdxLicense(
    val id: String,
    val fullName: String,
    val deprecated: Boolean = false
) {
    _0BSD("0BSD", "BSD Zero Clause License");
}
```

## Steps to Reproduce

See code examples above.

## Context

I ran into this while upgrading to Detekt 1.22 in https://github.com/oss-review-toolkit/ort/pull/6096.

## Your Environment

* Version of detekt used: 1.22
* Operating System and version: Windows 11
* Link to your project (if it's a public repository): https://github.com/oss-review-toolkit/ort

### LINKED ISSUE #5566: Cannot suppress `TrailingCommaOnDeclarationSite ` on enum with trailing semicolon
## Expected Behavior
When using `TrailingCommaOnDeclarationSite`, if an enum has a semicolon to separate the last item, KtLint adds a comma before. I reported the error in pinterest/ktlint/issues/1714 and it seems that the KtLint suppress is working, but the suppress for detekt is not.

```kotlin
enum class Example {
    A,
    B;

    fun something() = "blah"
}

enum class Example2 {
    A,
    B;

    companion object {
        val values = listOf(A, B)
    }
}
```

is formatted into

```kotlin
enum class Example {
    A,
    B,
    ;

    fun something() = "blah"
}

enum class Example2 {
    A,
    B,
    ;

    companion object {
        val values = listOf(A, B)
    }
}
```

## Observed Behavior
If I add `@Suppress("TrailingCommaOnDeclarationSite")` before the declaration of the enum, the error won't be suppressed and is still formatting the enum.

## Steps to Reproduce
Using this configuration of detekt

```yaml
TrailingCommaOnCallSite:
  active: true
  autoCorrect: true
  useTrailingCommaOnCallSite: true
TrailingCommaOnDeclarationSite:
  active: true
  autoCorrect: true
  useTrailingCommaOnDeclarationSite: true
```

and

```kotlin
@Suppress("TrailingCommaOnDeclarationSite")
enum class Example {
    A,
    B;

    fun something() = "blah"
}
```

## Context
This happens with detekt 1.22.0. The error should be suppressed, otherwise it will format the enum adding a trailing comma before the semicolon.

cortinico:
Fixes #5547
Fixes #5566
Related to #1999 also

This introduced the possibility to do more fine grained suppression on `detekt-formatting` rules. Specifically I'm dropping the custom logic we apply to the `Entity` we receive and just pass over the `PsiElement` that KtLint passes us.

Some of the location in the tests changed due to this, so this is a disruptive change at least for baselines (as the signature also changes). If we decide to merge it we should flag it in the release notes.

At the same time with KtLint you could only suppress at the `@file:` level so, aside from baseline, this won't disrupt other workflows.

### DIFF
--- a/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
+++ b/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
@@ -4,20 +4,15 @@ import com.pinterest.ktlint.core.Rule.VisitorModifier.RunAsLateAsPossible
 import com.pinterest.ktlint.core.api.EditorConfigProperties
 import com.pinterest.ktlint.core.api.editorconfig.CODE_STYLE_PROPERTY
 import com.pinterest.ktlint.core.api.editorconfig.EditorConfigProperty
-import io.github.detekt.psi.fileName
-import io.github.detekt.psi.toFilePath
 import io.gitlab.arturbosch.detekt.api.CodeSmell
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.CorrectableCodeSmell
 import io.gitlab.arturbosch.detekt.api.Debt
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
-import io.gitlab.arturbosch.detekt.api.Location
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.SingleAssign
-import io.gitlab.arturbosch.detekt.api.SourceLocation
-import io.gitlab.arturbosch.detekt.api.TextLocation
 import org.ec4j.core.model.Property
 import org.jetbrains.kotlin.com.intellij.lang.ASTNode
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.JavaDummyElement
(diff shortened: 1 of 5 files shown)


detekt-formatting/src/test/kotlin/io/gitlab/arturbosch/detekt/formatting/MaximumLineLengthSpec.kt
```diff
@@ -55,14 +56,13 @@ class MaximumLineLengthSpec {
...
 
         // Note that KtLint's MaximumLineLength rule, in contrast to detekt's MaxLineLength rule, does not report
         // exceeded lines in block comments.
-        assertThat(findings).hasSize(2)
-
-        assertThat(findings[0].entity.location.source.line).isEqualTo(8)
-        assertThat(findings[1].entity.location.source.line).isEqualTo(14)
+        io.gitlab.arturbosch.detekt.test.assertThat(findings).hasSize(2)
+        io.gitlab.arturbosch.detekt.test.assertThat(findings[0]).hasSourceLocation(7, 8)
+        io.gitlab.arturbosch.detekt.test.assertThat(findings[1]).hasSourceLocation(13, 12)
```
cortinico, line 65: I have a follow-up PR to clean all the `detekt-formatting` from using non detekt's `assertThat` function here: https://github.com/cortinico/detekt/commit/d9a4ea9a90bf9da6b100aa8a0848cbb8cb3b07c8

cortinico:
Can someone take a look at this one?

BraisGabin (approved):

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
```diff
@@ -41,32 +36,6 @@ abstract class FormattingRule(config: Config) : Rule(config) {
     val runAsLateAsPossible
         get() = RunAsLateAsPossible in wrapping.visitorModifiers
 
-    private val emit = { offset: Int, message: String, canBeAutoCorrected: Boolean ->
-        val (line, column) = positionByOffset(offset)
-        val location = Location(
```
chao2zhang, line 46: Would you mind adding a unit test for the location if such test has not existed

chao2zhang (approved):

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
cortinico: I've updated the tests to follow the logic here (i.e. previously we had a test that was checking file suppression as valid and local suprresion as not. Now the test checks that both suppressions are valid).

PR #5880, opened 2023-03-05
Update Compose headings to prevent duplicate references

TWiStErRob:
The relevance of the actual rule is lower than the compose-specific configuration:
image

I've ran into opening the wrong page multiple times myself, and @cortinico also copied the wrong link; and we know the website 😅, so I guess users will also have problems identifying the right section.


cortinico (approved):
Awesome 🙏

TWiStErRob:
Hmm, this didn't work... it's still only showing "FunctionNaming" for these renamed sections. Is the index per version?

TWiStErRob:
Never mind, might've been browser caching, after PR on v Next:
image

Issue #5607, opened 2022-12-06
Improve the html report


BraisGabin:
I was answering at #5524 but I end up writing far too much so I'm moving out to a new issue. 

In my ideal world, on the html report, I would like to see all the issues that were found (I would also like to see the ones that were suppressed and why they were suppressed). Of course the suppressed ones should be a bit hidden so they don't mess up the default report.

Also, at the bottom of the report, also a bit hidden, I would like to have the list of active rules that report nothing. And even more hidden the disabled rules.

And the final thing is that, every rule should show how it was configured and from where that configuration comes from and a diff with the default values.

The problem with that is that we will need a lot of thinking about how to create a template with all that information and keep the report still easy. We don't want to mess up with the default use case of this report: check why detekt is angry again.

We could split this in little pieces because there are thing here that can not be implemented with our current core.

I think that a report like this could help a lot about why detekt is doing what it is doing. And it should help to spot mud configurations on the configuration.

cortinico:
If someone is up for doing some UI mockups on Figma (or any other prototyping tool) this can be a great contribution <3

lexa-diky:
Should we also use some templating library for html generation? Because even now HTML report generator looks ugly with all in-code templating going on. If we want to add more features it will become even more hard to keep track of what's going on.

I suggest using something like [FreeMarker](https://freemarker.apache.org) as it is seem to be the most popular one (I also like it the most personally 😄 )

3 issues that I see with this approach is:
1. additional library dependency - shouldn't be an issue if we choose something popular. Plus at the end there will not be a lot of work to migrate to something else if we want to for one or another reason.
2. bigger artifact - don't think that it does matter for detekt
3. templating libraries are often full of security vulnerabilities and people tend not to update detekt ever in their projects - do we really care for locally running app? Don't think so

BraisGabin:
I'm not against but I think that before we do that we should know what we want and how we want it. And then decide how to implement it.

The good thing of the current implementation is that it's pure kotlin. You don't need to learn any new language. I know, templating is not that difficult but it is something you need to learn and I'm not sure if that will welcome more people to contribute or less.

lexa-diky:
Just tried to migrate current report to freemarker. 

[...]

Issue #5524, opened 2022-11-09
List currently active rules


haluzpav:
When I run `./gradlew detekt`, it lists only issues in my code.

I'd like to see a list of currently active **rules**, combining the default and my own configs. This would be an useful overview, since I disable many rules in my config, and it's hard to track which are actually enabled.

Bonus points for comparison with the default config, so I see which custom configs I could delete.

I'd like a table such as this in a report for example.

|ruleset| rule | default config | custom config |
|-| ---- | - | - |
|formatting|AnnotationOnSeparateLine|**disabled**|-|
|formatting|ChainWrapping|enabled|**disabled**|
|...||||

chao2zhang:
Hello @haluzpav, 
- 1 This seems like a useful detekt cli parameter to print the active rules. 
- 2 The diff comparison could be something customized in your gradle configuration.

Please let us know if you are willing to contribute to task 1.

haluzpav:
@chao2zhang Thanks for approval. Maybe I could do it.

It basically comes down to adding something like this into [`Analyzer`](https://github.com/detekt/detekt/blob/main/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt) and making it prettier (return instead of `println` etc.), right? 

```kotlin
activeRuleSetsToRuleSetConfigs.forEach { (ruleSet, _) ->
    ruleSet.rules
        .flatMap { rule ->
            when (rule) {
                is ConfigAware -> listOf(rule)
                is MultiRule -> rule.rules
                else -> error("Unknown rule type $rule")
            }
        }
        .filter { it.active }
        .forEach { println("${it.ruleId} is active") }
}
```

cortinico:
We do have work pending on this here:
- https://github.com/detekt/detekt/pull/5885

cortinico:
Reopening as this is still up for grab

schalkms:
Quite similar to issue #4925
Might make sense to merge those two ideas.

PR #5885, opened 2023-03-07
List currently active rules in CLI

haluzpav:
Implementation of the part 1 of https://github.com/detekt/detekt/issues/5524#issuecomment-1336302472. I tried to make it so it's extendable also for #5607.

I'm curious what you guys think. I'll add some docs and fix/add tests when I get some opinion. 🙂

Here are some sample outputs of using `--list-active-rules` flag.

```
Currently active rules:
        <no active rule sets>
```

```
Currently active rules:
        Rule set 'coroutines':
                GlobalCoroutineUsage
                InjectDispatcher
                RedundantSuspendModifier
                SleepInsteadOfDelay
                SuspendFunWithFlowReturnType
                SuspendFunWithCoroutineScopeReceiver
        Rule set 'comments':
                <no active rules in the set>
        Rule set 'empty-blocks':
                EmptyCatchBlock
                EmptyClassBlock
                EmptyDefaultConstructor
```

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/rules/CollectedRules.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/rules/CollectedRules.kt
@@ -1,0 +1,19 @@
+package io.gitlab.arturbosch.detekt.core.rules
+
+import io.gitlab.arturbosch.detekt.api.RuleId
+import io.gitlab.arturbosch.detekt.api.RuleSetId
+
+data class CollectedRules(
+    val ruleSets: List<RuleSet>,
+) {
+    data class RuleSet(
+        val id: RuleSetId,
+        val active: Boolean,
+        val rules: List<Rule>,
+    )
+
+    data class Rule(
+        val id: RuleId,
+        val active: Boolean,
+    )
+}
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/rules/RuleCollector.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/rules/RuleCollector.kt
@@ -1,0 +1,61 @@
+package io.gitlab.arturbosch.detekt.core.rules
+
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.MultiRule
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.RuleSetProvider
+import io.gitlab.arturbosch.detekt.core.ProcessingSettings
+
+internal class RuleCollector(
+    private val settings: ProcessingSettings,
+    private val providers: List<RuleSetProvider>,
+    private val config: Config,
+) {
+    fun run(): CollectedRules? = if (settings.spec.rulesSpec.listActiveRules) {
+        collectActiveRules()
+    } else {
+        null
+    }
+
+    @Suppress("DEPRECATION")
+    private fun collectActiveRules(): CollectedRules = CollectedRules(
+        ruleSets = providers.map { provider ->
+            val ruleSetConfig = config.subConfig(provider.ruleSetId)
+            val ruleSet = provider.instance(ruleSetConfig)
+            val rules = ruleSet.rules
+                .flatMap { rule ->
+                    when (rule) {
+                        is Rule -> listOf(rule)
+                        is MultiRule -> rule.rules
+                        else -> error("Unknown rule type $rule")
+                    }
+                }
+                .map { rule ->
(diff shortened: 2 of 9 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
```diff
@@ -40,9 +43,17 @@ internal interface Lifecycle {
         val (processors, ruleSets) = measure(Phase.LoadingExtensions) {
             processorsProvider.invoke() to ruleSetsProvider.invoke()
         }
+        // TODO move workaroundConfiguration to some existing Phase above or create new one?
+        val config = settings.spec.workaroundConfiguration(settings.config)
```
haluzpav, line 47: Not sure what to do about this.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/rules/RuleCollector.kt
```diff
@@ -0,0 +1,61 @@
...
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.RuleSetProvider
+import io.gitlab.arturbosch.detekt.core.ProcessingSettings
+
+internal class RuleCollector(
+    private val settings: ProcessingSettings,
+    private val providers: List<RuleSetProvider>,
+    private val config: Config,
+) {
+    fun run(): CollectedRules? = if (settings.spec.rulesSpec.listActiveRules) {
```
cortinico, line 14: Can we avoid using `null` here to represent the not enabled state? I would rather use a `sealed class` here if possible

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/rules/CollectedRules.kt
```diff
@@ -0,0 +1,19 @@
+package io.gitlab.arturbosch.detekt.core.rules
+
+import io.gitlab.arturbosch.detekt.api.RuleId
+import io.gitlab.arturbosch.detekt.api.RuleSetId
+
+data class CollectedRules(
```
cortinico, line 6: Do we really need a new data class inside `io.gitlab.arturbosch.detekt.core.rules` called `CollectedRules` to express this concept?

cortinico (commented):
Thanks for sending this 👍 I've left a couple of comments

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/rules/CollectedRules.kt
haluzpav: A few reasons:

1. It provides a clear contract what this "collect rules" feature does.
2. I could do this PR without any intermediate lists, and log to console straight in `RuleCollector.run`. However that's not very nice and extensible, considering we might want to pass it to some HTML reporter (etc.) in future.
3. I kinda need a custom `RuleSet` class - the old/default one (`io.gitlab.arturbosch.detekt.api.RuleSet`) doesn't have `active: Boolean` property. I could get it from `ruleSetConfig` at log/print time, but that's not nice IMO. I also don't want to pollute the namespace with another `RuleSet` class, so I hide it in this `CollectedRules`.

Let me know if you have some suggestion.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/rules/RuleCollector.kt
haluzpav: I'll adjust depending on result of the other comment.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/rules/CollectedRules.kt
cortinico: Why can't we use a `List<Rule>` instead?

haluzpav: I'm sorry, what are you asking? You say we could drop information about rule-sets? I think it's important to know what to which rule-set the listed rules belong to. Take a look at my example output in ticket description.

cortinico:
@haluzpav are you still interested in following up here?

haluzpav:
> @haluzpav are you still interested in following up here?

Sorry, not any time soon, my hands are full.

3flex:
Closing as there's been no activity for some time.

If the original branch is updated in future this can be reopened. If not, someone checking the content history in the original issue will find this and may use it to complete the implementation of the feature.

PR #5887, opened 2023-03-08
Sample freemarker html report

lexa-diky:
This PR is not intended to be ever merged. Just showcasing how freemarker could be used to render html report

NOTE:
1. detekt-report-html/src/main/kotlin/io/github/detekt/report/html/freemarker/FreemarkerReport.kt - just for the demo, should be replaced with adequate entity set and mappers
2. all template files should be moved to resources directory


schalkms (commented):
Cool project! Thanks for showing your project to the community!
I think this might be a good fit for the [detekt marketplace](https://detekt.dev/marketplace/) as custom report.
Did you think about adding this to the marketplace?

PR #5905, opened 2023-03-15
Minor fixes for the UnnecessaryAbstractClass rule

cortinico:
A couple of small fixes for `UnnecessaryAbstractClass`:
- `<compliant>` block was missing
- `visit*` methods were not calling `super.` (do we need a rule about this in the `ruleauthor` ruleset?)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAbstractClass.kt
```diff
@@ -78,8 +90,8 @@ class UnnecessaryAbstractClass(config: Config = Config.empty) : Rule(config) {
     }
```
cortinico, line 90: For some reason, moving the `super.visitKtFile(file)` crashes the Kotlin Compiler Plugin tests 🤔

chao2zhang (approved):

PR #5926, opened 2023-03-22
Add TodoWithoutLink rule

AndreuZaitsev:
Checks whether comment or documentation has TODO without a web-link.

This rule is supposed to be disabled by default, but it can be useful for some teams as mine to reuse it if there is a need.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -89,6 +89,8 @@ comments:
     active: false
     excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/androidUnitTest/**', '**/androidInstrumentedTest/**', '**/jsTest/**', '**/iosTest/**']
     searchProtectedProperty: false
+  TodoWithoutLink:
+    active: false
 
 complexity:
   active: true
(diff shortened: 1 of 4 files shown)


cortinico:
Forcing TODOs to have a specific link seems like maybe too specific. I think we should make the Regex `DEFAULT_LINK_PATTERN = """\bhttps?://(?:www\d?)?[\w/#&?=\-.]+\b"""` configurable so that users can specify how their TODO should look like.

For example people might want to enforce a `TODO (username): text` or a `TODO #githubissuenumber`. If so, the rule should also be renamed

AndreuZaitsev:
> Forcing TODOs to have a specific link seems like maybe too specific. I think we should make the Regex `DEFAULT_LINK_PATTERN = """\bhttps?://(?:www\d?)?[\w/#&?=\-.]+\b"""` configurable so that users can specify how their TODO should look like.
> 
> For example, people might want to enforce a `TODO (username): text` or a `TODO #githubissuenumber`. If so, the rule should also be renamed

Yup, could not agree more on that. 
So, custom pattern can be provided using config, right? 
```Kotlin
@Configuration("custom link regex pattern")
private val linkRegexPattern: String by config(DEFAULT_LINK_PATTERN)
```

As for having username or something else, it also can be configured using more complex regex pattern.

So, yes, since regex can be whatever user wants it to be, this can be renamed a bit no to be specific on web links.

schalkms (changes requested):
Did you check the `ForbiddenImport` rule whether this is already covered by rule configs?
https://detekt.dev/docs/rules/style/#forbiddencomment

BraisGabin:
I feel this rule is far too specific to fit inside the default rule set.

But it would be a great addition for a third party rule set. We could have it listed on our market place.

schalkms:
@AndreuZaitsev may I ask you to publish this rule somewhere? Then submit a new PR, which adds a link to the detekt marketplace. It would be great to have this rule added to the marketplace.

PR #5932, opened 2023-03-24
Make GeneratorSpec use resources

### LINKED ISSUE #5902: GeneratorSpec tries to copy files that may be locked for writing
## Expected Behavior
Build works reliably.

## Observed Behavior
Because of this:
https://github.com/detekt/detekt/blob/0123e2950137daf75840816435bed989820e5f72/detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/printer/GeneratorSpec.kt#L20-L21

When :detekt-generator:test runs around the same time as either detekt-rules-complexity or detekt-rules-coroutines are building, the task can fail because the files are locked.

## Steps to Reproduce
I think this is only a problem on Windows. I've seen this when the tasks are running in parallel.

## Context
The build should be reliable. Tests should also be self-contained and not rely on input from outside its own module.

## Your Environment

* Version of detekt used:

VitalyVPinchuk:
Fixes #5902

### DIFF
--- a/detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/printer/GeneratorSpec.kt
+++ b/detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/printer/GeneratorSpec.kt
@@ -6,48 +6,44 @@ import org.junit.jupiter.api.BeforeAll
 import org.junit.jupiter.api.Test
 import java.io.File
 import kotlin.io.path.Path
-import kotlin.io.path.createTempDirectory
 import kotlin.io.path.readText
 
 class GeneratorSpec {
     private val configPath = "/src/main/resources/config/config.yml"
 
-    private val tempDir1: File = createTempDirectory().toFile()
-    private val tempDir2: File = createTempDirectory().toFile()
+    private val rulePath1: File = Path("../detekt-generator/src/test/resources/ruleset1").toFile()
+    private val rulePath2: File = Path("../detekt-generator/src/test/resources/ruleset2").toFile()
 
     @BeforeAll
     fun init() {
-        Path("../detekt-rules-complexity").toFile().copyRecursively(tempDir1)
-        Path("../detekt-rules-coroutines").toFile().copyRecursively(tempDir2)
-
         val args = arrayOf(
             "--generate-custom-rule-config",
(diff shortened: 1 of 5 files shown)


detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/printer/GeneratorSpec.kt
```diff
@@ -6,48 +6,44 @@ import org.junit.jupiter.api.BeforeAll
...
-import kotlin.io.path.createTempDirectory
 import kotlin.io.path.readText
 
 class GeneratorSpec {
     private val configPath = "/src/main/resources/config/config.yml"
 
-    private val tempDir1: File = createTempDirectory().toFile()
-    private val tempDir2: File = createTempDirectory().toFile()
+    private val rulePath1: File = Path("../detekt-generator/src/test/resources/ruleset1").toFile()
+    private val rulePath2: File = Path("../detekt-generator/src/test/resources/ruleset2").toFile()
```
BraisGabin, line 15: Why is `../detekt-generator` needed? Which is the working dir for these tests? It should be `detekt-generator`, right?

BraisGabin (commented):
LGTM. Just a little nit on the comments.

detekt-generator/src/test/kotlin/io/gitlab/arturbosch/detekt/generator/printer/GeneratorSpec.kt
VitalyVPinchuk: That's because the paths should be consistent with `io.gitlab.arturbosch.detekt.generator.main(args)` call.
I tried it without `../` but it failed.

BraisGabin: And without `../detekt-generator/`? You should be already inside `detekt-generator`, right?

VitalyVPinchuk: Oh yes, you're right. It works this way.

BraisGabin (approved):

Issue #5934, opened 2023-03-24
UnusedPrivateMember rule counts package declarations as "usage references"


luanpotter:
## Expected Behavior

The following code:

```kotlin
package xyz.luan.test

fun main() {
    val xyz = "a"
    val luan = "b"
    val test = "c"
    println("Hello, world!")
}
```

Should fail detekt validation of the `UnusedPrivateMember` for the 3 unused local variables. Note that a similar snippet does fail:

```kotlin
package xyz.luan.test

fun main() {
    val xyz2 = "a"
    val luan = "b"
    val test = "c"
    println("Hello, world!")
}
```

Yields:

```
> Task :detekt FAILED
detekt-testbed/src/main/kotlin/xyz/luan/test/Main.kt:7:9: Private property `xyz2` is unused. [UnusedPrivateMember]
```

## Observed Behavior

The code is not reported and detekt succeeds even though there is an unused local variable.

Note: the same issue can happen with other types of unused private properties, like fields or parameters.

Any "bits" composing the package declaration are considered as variable uses. When in fact, package declarations should not count for referencing variables at all.

## Steps to Reproduce

Please see [my repo](https://github.com/luanpotter/detekt-testbed/tree/luan.unused-private-property-improvements) for a fully reproducible example (just run `./gradlew detekt`). Or run detekt on the code snippets above.  Should be simple to reproduce. I am using version 1.22.0.

## Context

I noticed that I had some private constructor parameters not used on some classes that weren't flagged, and they all had the same name. I had many wrong theories but did some science and turns out it was the package name.

The fix should be fairly simple, I will link a PR to this issue with a proposal.

## Your Environment

* Version of detekt used: 1.22.0
* Version of Gradle used: 8.0.2
* Gradle scan link: https://scans.gradle.com/s/ne6pd7y7cf3t2
* Operating System and version: Arch Linux (kernel 22.3.0)
* Link to your project: https://github.com/luanpotter/detekt-testbed/tree/luan.unused-private-property-improvements

PR #5935, opened 2023-03-24
Improve correctness of UnusedPrivateProperty

luanpotter:
Fixes https://github.com/detekt/detekt/issues/5934

Basically ignore package declarations when counting references for the purposes of identifying Unused Private Properties.

## Followups

I believe there are many other edge cases that could be considered; for example, the right-hand side of a dot-expression is also being considered a use reference:

```kotlin
class A (val a: String = "foo")

fun main() {
    val a = 1
    println(A().a)
}
```

Should cause an error, but it doesn't. I'm not sure if there is a better solution here other than exploring each edge case individually, but I am happy to chip down on at least the more obvious ones on followup PRs.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateProperty.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateProperty.kt
@@ -13,8 +13,10 @@ import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
 import io.gitlab.arturbosch.detekt.api.internal.Configuration
 import io.gitlab.arturbosch.detekt.rules.isActual
 import io.gitlab.arturbosch.detekt.rules.isExpect
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.KtNamedDeclaration
+import org.jetbrains.kotlin.psi.KtPackageDirective
 import org.jetbrains.kotlin.psi.KtParameter
 import org.jetbrains.kotlin.psi.KtPrimaryConstructor
 import org.jetbrains.kotlin.psi.KtProperty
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateProperty.kt
```diff
@@ -131,7 +126,20 @@ private class UnusedPrivatePropertyVisitor(private val allowedNames: Regex) : De
...
     override fun visitReferenceExpression(expression: KtReferenceExpression) {
-        nameAccesses.add(expression.text.removeSurrounding("`"))
+        if (!skipNode(expression)) {
+            nameAccesses.add(expression.text.removeSurrounding("`"))
+        }
         super.visitReferenceExpression(expression)
     }
+
+    private fun skipNode(expression: KtReferenceExpression): Boolean {
+        return when {
```
luanpotter, line 136: note: adding a bit of structure as we can followup with more exceptions and rules for other edge cases

luanpotter: example of a followup: ignore import statements as well

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivatePropertySpec.kt
```diff
@@ -808,4 +808,41 @@ class UnusedPrivatePropertySpec(val env: KotlinCoreEnvironment) {
...
+                fun main() {
+                    val unused = 1
+                    println("foo")
+                }
+            """.trimIndent()
+            assertThat(subject.lint(code)).hasSize(1)
+        }
+
+        @Test
+        fun `package declarations are ignored`() {
```
schalkms, line 837: I do think we only need this test to be added. The other cases are covered by the existing tests.

schalkms (commented):
Thanks for the quick fix!
Besides the redundant two tests, this PR looks very good!

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivatePropertySpec.kt
luanpotter: good point! I had skimmed the file for these tests but initially failed to see them nested in the inner classes.
I removed the other two.

on that note though, lmk of including my 1 test under a new "irrelevant references are ignored" nested class (it seems all tests are nested somewhere)
I was planning on following this up with some other edge cases (eg import statements should be ignored as well)
but I am happy to put the test nested somewhere else if that makes sense, just lmk where it better fits

schalkms: Regarding test under new nested class 👍
Regarding follow up --> please submit a new PR, it's easier to review 👍

schalkms (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivatePropertySpec.kt
luanpotter: yes, def meant followup PRs! thanks for reviewing so swiftly :)

BraisGabin (approved):

PR #5937, opened 2023-03-24
Potential new rule: DoubleNegativeLambda

drawers:
# Potential new rule: DoubleNegativeLambda

## Rationale

There was recently a popular post on Y-Combinator Hacker News arguing against the use of `unless` in Ruby.

https://news.ycombinator.com/item?id=33965933

It is arguing that:
1. Double negatives are hard to read: (`takeUnless { !it.isOdd() }` vs `takeIf { it.isEven() }`)
2. It's easy for a nice simple `takeUnless { it.isEven() }` to become complex when some edge case is found and an extra condition is added so it becomes something like `takeUnless { it.isEven() && !it.isAlreadyAccountedFor() }`. Now the reader needs to unpack this expression using DeMorgan's laws

I am wondering if you would be interested in a rule that checks for double negatives like this using `takeUnless`? (it's okay if 'no', can keep it private)

## Considerations

* Should it check for `filterNot` and `none` by default as well?

To me, negatives in a `filterNot` block seem less harmful than in a `takeUnless` block. I looked for examples and I found this one [in KotlinPoet](https://github.com/square/kotlinpoet/blob/master/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt#L572):

```kotlin
addModifiers(
        flags.modalities
          .filterNot { it == FINAL && !isOverride } // Final is the default
          .filterNot { it == OPEN && isOverride } // Overrides are implicitly open
          .filterNot { it == OPEN && isInInterface }, // interface methods are implicitly open
      )
```
It doesn't seem to benefit a lot from being rewritten in the positive as `filter { it != FINAL || isOverride }` 

* Should it account for functions in backticks? 

One could argue that in addition to detecting simple function names with "not" and "non" in them, it should also try and catch negation in functions with back ticks to report cases like `takeUnless { it.`isn't a word`() }`. I thought it would add a bit of complexity to the rule and detecting negation in a backtick function name would make it a harder problem where we would be trying to solve the problem of detecting negation in English which has so many different ways (isn't, can't, won't etc.)

* Should it use type resolution to find functions using FQN?

To me, functions like `takeUnless` are quite rare. I think it would be unusual to want to distinguish between one or the other.

## Disclosure

I used ChatGPT4 to help me write the rule and was thinking of writing an article about my experience (Chat GPT was not that helpful here despite Detekt code seemingly being in the training data).

Attaching images of prompts and responses for transparency:
https://user-images.githubusercontent.com/11054916/227701241-3e245ea6-bc80-48cb-a7bf-87cd4cbd65fd.png
https://user-images.githubusercontent.com/11054916/227701292-3949810e-af48-4a1d-8e92-e394b4dd39b8.png
https://user-images.githubusercontent.com/11054916/227701325-88c87fd0-ae47-42b6-9eba-bbd84c47b2c9.png
https://user-images.githubusercontent.com/11054916/227701353-d659c1ba-52ab-4823-859f-952f6f5d9155.png
https://user-images.githubusercontent.com/11054916/227701385-5d8c549d-e8b7-4731-92e9-b68ca6187cd1.png
https://user-images.githubusercontent.com/11054916/227701400-c8d409ef-4046-47a7-8e89-3d9e9692dc91.png
https://user-images.githubusercontent.com/11054916/227701424-f4d294d9-2d1e-466e-9492-30406e697d52.png
https://user-images.githubusercontent.com/11054916/227701439-94f86de1-363e-4edc-8979-259ab9393012.png
https://user-images.githubusercontent.com/11054916/227701454-8802f70f-9060-4c04-ba7d-cf8082bbe8fe.png
https://user-images.githubusercontent.com/11054916/227701474-1c2484a1-0ade-467f-b205-8d2f8eec1249.png
https://user-images.githubusercontent.com/11054916/227701485-7cd2e21e-d46d-4e9c-91ac-b18804c5e76e.png
https://user-images.githubusercontent.com/11054916/227701499-49eaa96d-a87c-4383-96ad-89abec25407a.png
https://user-images.githubusercontent.com/11054916/227701509-6895ef2d-9d6e-443b-a9ee-e2ef4418d3e3.png

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
@@ -58,6 +58,7 @@ class DoubleNegativeLambda(config: Config = Config.empty) : Rule(config) {
     private val negativeFunctions: List<NegativeFunction> by config(
         valuesWithReason(
             "takeUnless" to "Use `takeIf` instead.",
+            "none" to "Use `all` instead.",
         )
     ) { list ->
         list.map { NegativeFunction(simpleName = it.value, recommendation = it.reason) }
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
@@ -71,7 +71,7 @@ class DoubleNegativeLambda(config: Config = Config.empty) : Rule(config) {
     override fun visitCallExpression(expression: KtCallExpression) {
         super.visitCallExpression(expression)
         val calleeExpression = expression.calleeExpression?.text ?: return
-        val forbiddenFunction = negativeFunctions.firstOrNull { it.name == calleeExpression } ?: return
+        val negativeFunction = negativeFunctions.firstOrNull { it.name == calleeExpression } ?: return
         val lambdaExpression = expression.lambdaArguments.firstOrNull() ?: return
 
         val forbiddenChildren = lambdaExpression.collectDescendantsOfType<KtExpression> {
@@ -83,7 +83,7 @@ class DoubleNegativeLambda(config: Config = Config.empty) : Rule(config) {
                 CodeSmell(
                     issue,
                     Entity.from(expression),
-                    formatMessage(forbiddenChildren, forbiddenFunction)
+                    formatMessage(forbiddenChildren, negativeFunction)
                 )
             )
         }
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
@@ -1,0 +1,95 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtOperationReferenceExpression
+import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
+
+/**
+ * Detects negation in lambda blocks where the function name is also in the negative (like `takeUnless`).
+ * A double negative is harder to read than a positive. In particular, if there are multiple conditions with `&&` etc. inside
+ * the lambda, then the reader may need to unpack these using DeMorgan's laws. Consider rewriting the lambda to use a positive version
+ * of the function (like `takeIf`).
+ *
+ * <noncompliant>
+ * Random.Default.nextInt().takeUnless { !it.isEven() }
+ * </noncompliant>
+ * <compliant>
+ * Random.Default.nextInt().takeIf { it.isOdd() }
+ * </compliant>
+ */
+class DoubleNegativeLambda(config: Config = Config.empty) : Rule(config) {
+
+    override val issue = Issue(
+        "DoubleNegativeLambdaBlock",
+        Severity.Style,
+        "Double negative from a function name expressed in the negative (like `takeUnless`) with a lambda block " +
+            "that also contains negation. This is more readable when rewritten using a positive form of the function " +
+            "(like `takeIf`).",
+        Debt.FIVE_MINS,
+    )
+
+    private val splitCamelCaseRegex = "(?<=[a-z])(?=[A-Z])".toRegex()
+
+    private val negationTokens = listOf(
+        KtTokens.EXCL,
+        KtTokens.EXCLEQ,
+        KtTokens.EXCLEQEQEQ,
+        KtTokens.NOT_IN,
+        KtTokens.NOT_IS,
+    )
+
+    private val negatingFunctionNameParts = listOf("not", "non")
+
+    @Configuration("Function names expressed in the negative that can form double negatives with their lambda blocks.")
+    private val negativeFunctions: Set<String> by config(listOf("takeUnless")) { it.toSet() }
+
+    override fun visitCallExpression(expression: KtCallExpression) {
+        super.visitCallExpression(expression)
+        val calleeExpression = expression.calleeExpression?.text ?: return
+
+        if (calleeExpression in negativeFunctions) {
+            val lambdaExpression = expression.lambdaArguments.firstOrNull() ?: return
+            val forbiddenChildren = lambdaExpression.collectDescendantsOfType<KtExpression> {
+                it.isForbiddenNegation()
+            }
+
+            if (forbiddenChildren.isNotEmpty()) {
+                report(
+                    CodeSmell(
+                        issue,
+                        Entity.from(expression),
+                        "Double negative through using ${forbiddenChildren.joinInBackTicks()} inside a " +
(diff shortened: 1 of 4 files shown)


schalkms:
This rule is definitely a good addition to the standard style ruleset! I like it. 
Regarding your question, I'd keep the first version of this rule as simple as possible.
Regarding your usage of ChatGPT, I'd like to learn about your experience with PSI and detekt. I'm looking forward to your article. 🙂

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
```diff
@@ -0,0 +1,95 @@
...
+ * Detects negation in lambda blocks where the function name is also in the negative (like `takeUnless`).
+ * A double negative is harder to read than a positive. In particular, if there are multiple conditions with `&&` etc. inside
+ * the lambda, then the reader may need to unpack these using DeMorgan's laws. Consider rewriting the lambda to use a positive version
+ * of the function (like `takeIf`).
+ *
+ * <noncompliant>
+ * Random.Default.nextInt().takeUnless { !it.isEven() }
+ * </noncompliant>
+ * <compliant>
+ * Random.Default.nextInt().takeIf { it.isOdd() }
```
BraisGabin, line 28: The code in the non-compliant and the compliant doesn't behave the same. That doesn't help to understand the rule. 
```suggestion
 * Random.Default.nextInt().takeIf { it.isEven() }
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
```diff
@@ -0,0 +1,95 @@
...
+        KtTokens.EXCLEQ,
+        KtTokens.EXCLEQEQEQ,
+        KtTokens.NOT_IN,
+        KtTokens.NOT_IS,
+    )
+
+    private val negatingFunctionNameParts = listOf("not", "non")
+
+    @Configuration("Function names expressed in the negative that can form double negatives with their lambda blocks.")
+    private val negativeFunctions: Set<String> by config(listOf("takeUnless")) { it.toSet() }
```
BraisGabin, line 55: We could use here valueWithReason so in the reason we could say which is the positive form of that function. For example, for `takeUnless` it would be `takeIf`.

You can see examples of this at forbiddenMethodCall.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    private val negationTokens = listOf(
+        KtTokens.EXCL,
+        KtTokens.EXCLEQ,
+        KtTokens.EXCLEQEQEQ,
+        KtTokens.NOT_IN,
+        KtTokens.NOT_IS,
+    )
+
+    private val negatingFunctionNameParts = listOf("not", "non")
```
BraisGabin, line 52: Should we make this configurable?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
drawers: Thanks, didn't know about this feature. I attempted this in 38a187a6c9

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
drawers: Thanks, I think they should be configurable too. Attempted this in df6a6626

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
```diff
@@ -0,0 +1,130 @@
...
+        KtTokens.NOT_IS,
+    )
+
+    @Configuration(
+        "Function names expressed in the negative that can form double negatives with their lambda blocks. " +
+            "These are grouped together with the positive counterpart of the function, or `null` if this is unknown."
+    )
+    private val negativeFunctions: List<NegativeFunction> by config(
+        valuesWithReason(
+            "takeUnless" to "takeIf",
```
BraisGabin, line 59:
```suggestion
            "takeUnless" to "Use `takeIf` instead.",
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
```diff
@@ -0,0 +1,130 @@
...
+        negativeFunction: NegativeFunction,
+    ) = buildString {
+        append("Double negative through using ${forbiddenChildren.joinInBackTicks()} inside a ")
+        append("`${negativeFunction.simpleName}` lambda. ")
+        append("Rewrite in the positive")
+        if (negativeFunction.positiveCounterpart != null) {
+            append(" with `${negativeFunction.positiveCounterpart}`.")
+        } else {
+            append(".")
+        }
```
BraisGabin, line 113:
```suggestion
        if (negativeFunction.positiveCounterpart != null) {
            append(negativeFunction.positiveCounterpart")
        } else {
            append("Rewrite in the positive.")
        }
```

I think that `reason` should contain a reason, not just the counter part. And that should also make this code a bit simplier. Don't commit exactly my code, I assume that further rework is needed to land this change.

BraisGabin (commented):
I really like the idea behind this rule 👍.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
drawers: Thanks, that makes a lot more sense. I did the change in 2ded6a41c 

I called it a "recommendation" ("Use `takeIf` instead") rather than a "reason" because that seems to be what it's called in the docs:
https://github.com/detekt/detekt/blob/main/.github/CONTRIBUTING.md

but I can rename it back to "reason" if you prefer.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeLambda.kt
```diff
@@ -0,0 +1,130 @@
...
+    )
+
+    @Configuration(
+        "Function names expressed in the negative that can form double negatives with their lambda blocks. " +
+            "These are grouped together with a recommendation to use a positive counterpart, or `null` if this is " +
+            "unknown."
+    )
+    private val negativeFunctions: List<NegativeFunction> by config(
+        valuesWithReason(
+            "takeUnless" to "Use `takeIf` instead.",
```
cortinico, line 60: what's the reason why you haven't included `none` and `filterNot` in the default here?

drawers: Thank you for the review. 

I think there is a good case to add `none` as a default. I searched a few big Kotlin codebases, including Detekt and KotlinPoet. I couldn't find any examples of `none` with a negation in its lambda. So it seems it would be something we would catch in a code review. Adding `none` would help explain the rule to consumers too.

I did not include `filterNot` because it seems there are a few cases where it seems more readable leaving it in the negative. I found [this example](
https://github.com/square/kotlinpoet/blob/master/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt#L572) in KotlinPoet:

```kotlin
addModifiers(
        flags.modalities
          .filterNot { it == FINAL && !isOverride } // Final is the default
          .filterNot { it == OPEN && isOverride } // Overrides are implicitly open
          .filterNot { it == OPEN && isInInterface }, // interface methods are implicitly open
      )
```

Rewriting this, we'd end up with:
```kotlin
addModifiers(
        flags.modalities
          .filter { it != FINAL && isOverride } // Final is the default
          .filterNot { it == OPEN && isOverride } // Overrides are implicitly open
          .filterNot { it == OPEN && isInInterface }, // interface methods are implicitly open
      )
```

What do you think about adding `none` as a default and leaving out `filterNot`?

cortinico: > What do you think about adding none as a default and leaving out filterNot?

Yup makes sense, thanks for clarifying 👍 `none` feels also like the most natural use case for this rule

cortinico (approved):

PR #5944, opened 2023-03-27
Execute tests in parallel

Goooler:
https://docs.gradle.org/8.0/userguide/performance.html#execute_tests_in_parallel

### DIFF
--- a/build-logic/src/main/kotlin/module.gradle.kts
+++ b/build-logic/src/main/kotlin/module.gradle.kts
@@ -24,7 +24,11 @@ jacoco.toolVersion = versionCatalog.findVersion("jacoco").get().requiredVersion
 tasks.withType<Test>().configureEach {
     useJUnitPlatform()
     // https://docs.gradle.org/8.0/userguide/performance.html#execute_tests_in_parallel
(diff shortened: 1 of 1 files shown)


3flex:
I'm not sure why I haven't tried this earlier!

I think we should be a bit less aggressive though and use the suggestion from the Gradle docs which caps forks to half the processor count at most.

Goooler:
I believe we can use `Runtime.getRuntime().availableProcessors()` on CI, which means full cores, there are 2 cores in GHA runners, use `(Runtime.getRuntime().availableProcessors() / 2).takeIf { it > 0 } ?: 1` (result is 1) here seems to be less improvement on CI.

Goooler:
I believe we can use `Runtime.getRuntime().availableProcessors()` on CI, which means full cores, there are 2 cores in GHA runners, use `(Runtime.getRuntime().availableProcessors() / 2).takeIf { it > 0 } ?: 1` (result is 1) here seems to be less improvement on CI.

3flex:
I was thinking about local builds. Maybe make it conditional on `CI` environment variable to use all cores in CI but not when building locally.

3flex (approved):

schalkms (approved):

Issue #5940, opened 2023-03-25
Feature request: Make `Detekt` and `DetektCreateBaselineTask` implement a common interface


ZacSweers:
Your feature request may already be reported!
Please search in the [issues](https://github.com/detekt/detekt/issues) before creating a new one.

## Expected Behavior

Currently, these two tasks are completely independent but have overlapping inputs. This means that if we want to configure Detekt (for example – `jvmTarget`), we must manually configure both types. It would be convenient if this shared configuration was extracted to a common interface (that extends `Task`) that could be used to configure both.

## Current Behavior

```kotlin
project.tasks.configureEach<Detekt> {
  this.jvmTarget = jvmTarget
  exclude("**/build/**")
}
project.tasks.configureEach<DetektCreateBaselineTask> {
  this.jvmTarget = jvmTarget
  exclude("**/build/**")
}
```

Desired, using a name like `DetektTask`

```kotlin
project.tasks.configureEach<DetektTask> {
  // Covers both Detekt and DetektCreateBaselineTask
  this.jvmTarget = jvmTarget
  exclude("**/build/**")
}
```

atulgpt:
Hi @ZacSweers are you talking something similar to https://github.com/detekt/detekt/pull/5946 ?

ZacSweers:
Yup! Note you can use the Task interface rather than DefaultTask and then you don't need to make the base task also a SourceTask (since the baseline task wasn't one before). Then the base task can also be an interface

atulgpt:
`since the baseline task wasn't one before` it was `SourceTask` before as well. See below SS
<img width="1392" alt="Screenshot 2023-03-28 at 1 14 01 AM" src="https://user-images.githubusercontent.com/11026203/228049650-cc867eb8-f508-4e05-8616-cf628790a40a.png">

PR #5946, opened 2023-03-27
Make common DetektBaseSourceTask.kt

atulgpt:
Maybe fixes https://github.com/detekt/detekt/issues/5940

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -59,7 +59,7 @@ abstract class Detekt @Inject constructor(
     private val objects: ObjectFactory,
     private val workerExecutor: WorkerExecutor,
     private val providers: ProviderFactory,
-) : DetektBaseSourceTask(), VerificationTask {
+) : DetektSourceTask(), VerificationTask {
 
     @get:InputFile
     @get:Optional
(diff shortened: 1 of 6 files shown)


3flex:
We'll need to have a think about whether this is the right structure to use, and I suggest it's going to box us in later if we want to make changes.

With this PR we'll have JVM-specific options are put alongside non-JVM-specific options in the extracted abstract class which makes it harder to them move them to a JVM-specific class later without breaking changes. I'd like to agree on which elements we might want to extract and treat differently. This will also let us support cases like #5947.

atulgpt:
> We'll need to have a think about whether this is the right structure to use, and I suggest it's going to box us in later if we want to make changes.
> 
> With this PR we'll have JVM-specific options are put alongside non-JVM-specific options in the extracted abstract class which makes it harder to them move them to a JVM-specific class later without breaking changes. I'd like to agree on which elements we might want to extract and treat differently. This will also let us support cases like #5947.

Hi, @3flex I can also make a few more interfaces(based on frameworks or language they control) and then compose those interfaced in `DetektSourceTask` and make `DetektSourceTask` as `internal` so that it can't be used as public API.

But I agree we should discuss this and close this to avoid any future roadblocs

cortinico (approved):
Generally agree with the direction, but I second what @3flex mentioned. Let's align on this before we making changes to the task class hierarchy.

atulgpt:
Hi @3flex, since we are on the path to 2.0 should we decide on the strategy to be followed here?

3flex:
I'm working on a new version of the plugin which will be ready for initial review in the next couple of weeks depending on how much time I have to spend on it. It will resolve that issue.

atulgpt:
Ohh nice. Closing this PR then

PR #5951, opened 2023-03-29
Add a test for catching undocumented public interfaces

ncteisen:
This PR adds a test in UndocumentedPublicClassSpec that validated behavior for catching uncommented public interfaces, which did not exist before. I wrote this for myself to validate this assumption because I did not see it in the test file already.

I realize that this is already covered in the documentation: "This rule reports public classes, objects and interfaces". But I missed that on initial read and went to the test to check my assumption

### DIFF
--- a/detekt-rules-documentation/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicClassSpec.kt
+++ b/detekt-rules-documentation/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/UndocumentedPublicClassSpec.kt
@@ -236,6 +236,20 @@ class UndocumentedPublicClassSpec {
     }
 
     @Test
+    fun `should report for uncommented interface`() {
+        val code = """
+            fun interface Interface {
+                /**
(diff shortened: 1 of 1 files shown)


BraisGabin (approved):
Thanks!!!

schalkms (approved):

ncteisen:
I updated to fix that test! Thanks for posting that error

I had been running `./gradlew build` locally, which did not show the error. Is there a better way to run all of the tests?

schalkms:
> I updated to fix that test! Thanks for posting that error
> 
> I had been running `./gradlew build` locally, which did not show the error. Is there a better way to run all of the tests?

Detekt's tests compile test snippets to detect invalid Kotlin code being passed to the static analysis rule.
You can add the argument `-Pcompile-test-snippets=true`.
Locally, those tests run without compiling the test snippets for achieving better test-runtimes.

ncteisen:
Ah that makes sense, thanks for explaining

This passes `./gradlew build -Pcompile-test-snippets=true` now, so I think it is good to go

schalkms (approved):

PR #5954, opened 2023-03-31
Update README.md

mjovanc:

For more information, see the [CONTRIBUTING guide](https://github.com/detekt/detekt/blob/main/.github/CONTRIBUTING.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ You can find [other ways to install detekt here](https://detekt.dev/cli.html)
 #### with Gradle
 
 ```kotlin
+import io.gitlab.arturbosch.detekt.*
+
 plugins {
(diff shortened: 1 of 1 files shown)


3flex:
There are a couple of reasons I don't think we should include this:

1. This isn't meant to be a complete Gradle build file, it's just to show a few relevant snippets to get started, and adding this at the top makes it a bit noisier.
2. It uses a star import which is something detekt would warn about.

Can we change the `withType<Detekt>` to use a fully qualified import instead?

cortinico:
> There are a couple of reasons I don't think we should include this:

nit: I think we should not have those examples on the README at all. We do have a versioned website and that should be the source of thruth

PR #5974, opened 2023-04-07
Fix githubRelease picking wrong version

cortinico:
When doing `1.23.0-RC1` I've noticed that assets fail to be uploaded as the version number doesn't get picked correctly (it's "undefined"). I'm updating it here.

### DIFF
--- a/build-logic/src/main/kotlin/releasing.gradle.kts
+++ b/build-logic/src/main/kotlin/releasing.gradle.kts
@@ -16,6 +16,8 @@ nexusPublishing {
     }
 }
 
+val version = Versions.currentOrSnapshot()
+
 githubRelease {
     token(providers.gradleProperty("github.token"))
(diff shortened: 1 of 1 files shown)


build-logic/src/main/kotlin/releasing.gradle.kts
```diff
@@ -16,18 +16,21 @@ nexusPublishing {
...
+val version = Versions.currentOrSnapshot()
+
 githubRelease {
     token(providers.gradleProperty("github.token"))
     owner.set("detekt")
     repo.set("detekt")
     overwrite.set(true)
     dryRun.set(false)
     draft.set(true)
+    prerelease.set(true)
```
cortinico, line 28: I've added this as a safeguard

3flex (approved):

Issue #3508, opened 2021-03-01
ForbiddenComment should allow patterns


TWiStErRob:
Following up on https://github.com/detekt/detekt/issues/959, the resolution of that is not powerful enough / complicates things.

## Context
We have the following configuration:
```yaml
  ForbiddenComment:
    values: ['REVIEW', 'STOPSHIP']
```
and I got this message flagged by Detekt:
```kotlin
// to express in the preview that it's not a normal TextView.
```
the only workaround I found is this
```yaml
  ForbiddenComment:
    values: ['REVIEW', 'STOPSHIP']
    # Allow "preview", "FireView", "DoReView" to occur in comments.
    # So only `// REVIEW explanation` will be flagged.
    # See https://github.com/detekt/detekt/issues/3508
    allowedPatterns: '(?<!^)(review|Review|reView|ReView)'
```

## Expected Behavior of the rule
```kotlin
// REVIEW foo -> flag
// review foo -> flag
// foo review -> allow
// preview -> allow
```
To do this I want to be able to say [`^(?i)REVIEW\b`](https://regex101.com/r/NioVGu/3) (as in case insensitive, starts with the letters "review" and it's not followed by word characters), instead of the current behavior which is https://regex101.com/r/NioVGu/2/

TWiStErRob:
Even this workaround is broken, because it allows `// STOPSHIP breaks preview`, which is not good!

TWiStErRob:
I think we need something like
```yaml
  ForbiddenComment:
    valuePatterns: [ '^(?i)REVIEW\b', '^STOPSHIP\b' ]
```

Which would be exclusively allowed to be present in the configuration only when `values` is not present.

schalkms:
I think the config would be more flexible if allows just one regular expression. I like your idea.
@TWiStErRob may I ask if you have resources to submit a PR and help out?

TWiStErRob:
No time at the moment from my side, maybe later (months). If anyone else is willing, feel free to take it.

TWiStErRob:
@schalkms how come you say one regex is more flexible than multiple? Consider a scenario where we have multiple rules, say:
- comment shouldn't start with a known todo prefix (todo, fixme, stopship)  
but if it contains a Jira ticket number, then it's ok.
- [comments should be sentences](https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html) (i.e. shouldn't end with non-punctuation)
[...]

PR #5981, opened 2023-04-10
Add `comments` with a list of regexes to `ForbiddenComment`

atulgpt:
Fixes https://github.com/detekt/detekt/issues/3508


cortinico:
CI is crashing on rule code here:
```
 Caused by: java.lang.StringIndexOutOfBoundsException: String index out of range: -2
	at java.lang.String.substring(String.java:1931)
	at io.gitlab.arturbosch.detekt.rules.style.ForbiddenComment.checkForbiddenComment(ForbiddenComment.kt:87)
	at io.gitlab.arturbosch.detekt.rules.style.ForbiddenComment.visitKtFile(ForbiddenComment.kt:65)
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -74,6 +80,23 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
                 )
             }
         }
+
+        val strippedText = if (text.startsWith("// ")) {
```
TWiStErRob, line 84: please update the docs how the initial space is handled, some people use it. (e.g. `//Foo` is invalid! so the regex should be able to match that somehow.)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -40,6 +40,12 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
     @Configuration("forbidden comment strings")
     private val values: List<String> by config(listOf("FIXME:", "STOPSHIP:", "TODO:"))
 
+    @Suppress("ExplicitItLambdaParameter")
+    @Configuration("forbidden comment string patterns")
+    private val valuePatterns: List<Regex> by config(emptyList()) { it: List<String> ->
+        it.map(String::toRegex)
```
TWiStErRob, line 46:
```suggestion
    @Configuration("forbidden comment string patterns")
    private val valuePatterns: List<Regex> by config(emptyList<String>()) {
        it.map(String::toRegex)
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
atulgpt: Seeking feedback here... Can I remove the trimming of `// ` part? If yes that regex like `^(gi)REVIEW` will not work as due `^` it will not match? The new regex that we should use will be `^// (gi)REVIEW`

atulgpt: I will update but before that, I want some suggestions should I simply search the regex in the comment text or make some modification to the text and then search it so that `^` honours for every line in the comment? Former is really easy and doesn't involve splitting the comment text into multiple lines.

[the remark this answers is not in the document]
TWiStErRob: Considering how generic this rule is (i.e. it's not named `ForbiddenTodoComment`) I would opt for being really generic and allowing to match the exact comment contents (comment starts after comment delimiter), so space is included in `^...`. That way people can set up `^(?! )` to detect commented code or even misaligned KDoc:
```
/**
 * good
 *bad
 * good
 */
```

TWiStErRob: How does `.text` PSI work for multiline comments?

atulgpt: It gives completed comment content(multiline string if comment spans multiple lines)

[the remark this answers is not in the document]
atulgpt: I have updated the comment rule to match the entire `text`. Anyway people can provide multiple regex for different type of comments(single line, block comment, or documentation)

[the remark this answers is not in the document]
atulgpt: Hi, @TWiStErRob did you get the chance to re-review this PR? Is there anything which is blocking this one?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -38,7 +38,13 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
     )
 
     @Configuration("forbidden comment strings")
-    private val values: List<String> by config(listOf("FIXME:", "STOPSHIP:", "TODO:"))
+    @Deprecated("Use `valuePatterns` instead")
+    private val values: List<String> by config(emptyList())
+
+    @Configuration("forbidden comment string patterns")
+    private val valuePatterns: List<Regex> by config(listOf("FIXME:", "STOPSHIP:", "TODO:")) {
```
BraisGabin, line 45: I would name it `comments`. In the other forbidden rules we use that. Example: on forbidden imports we use `imports`.

BraisGabin (commented):
Looks good. Just a comment about the naming. What do you think?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
```diff
@@ -235,4 +247,53 @@ class ForbiddenCommentSpec {
             assertThat(findings.first().message).isEqualTo(expectedMessage)
         }
     }
+
+    @Nested
+    inner class `custom value pattern is configured` {
+        private val patternStr = """^//( )?(?i)REVIEW\b"""
```
TWiStErRob, line 253: Could you please make these tests a bit more explicit? the `( )?` is really relaxed, you can change the production code to not match spaces and all tests might still pass.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
```diff
@@ -235,4 +247,53 @@ class ForbiddenCommentSpec {
...
+        @Test
+        fun `should not report a finding when review doesn't match the pattern`() {
+            val comment = "// to express in the preview that it's not a normal TextView."
+            val findings = ForbiddenComment(messageConfig).compileAndLint(comment)
+            assertThat(findings).isEmpty()
+        }
+
+        @Test
+        fun `should report a finding when STOPSHIP is present`() {
+            val comment = "// STOPSHIP to express in the preview that it's not a normal TextView."
```
TWiStErRob, line 267: Originally this comment was to ensure that a pattern "(?i)REVIEW\b" doesn't match a normal word:
```
            val comment = "// STOPSHIP to express in the preview that it's not a normal TextView."
                                                          ^^^^^^
```
but since `^` is possible, this might be not necessary?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -38,7 +38,13 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
     )
 
     @Configuration("forbidden comment strings")
-    private val values: List<String> by config(listOf("FIXME:", "STOPSHIP:", "TODO:"))
+    @Deprecated("Use `valuePatterns` instead")
```
TWiStErRob, line 41: Somehow this should be mentioned I feel:
```suggestion
    @Deprecated("Use `valuePatterns` instead, make sure you escape your text for Regular Expressions.")
```
otherwise people might just change `values` to `valuePatterns` and expect it to "just work".

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
```diff
@@ -235,4 +247,53 @@ class ForbiddenCommentSpec {
             assertThat(findings.first().message).isEqualTo(expectedMessage)
         }
     }
+
+    @Nested
+    inner class `custom value pattern is configured` {
```
TWiStErRob, line 252: I would add some tests to verify that the production code is using `containsMatchIn`, not `matchesEntire`, e.g. `// foo STOPSHIP bar` should be flagged for `valuePatterns: [ STOPSHIP ]`

TWiStErRob (approved):
Sorry for the wait. Looks good on my side, I don't think there are blockers, just a few thoughts.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
atulgpt: Hi, @TWiStErRob I didn't get the comment. `( )?` just ensures both `// review` and `//review` gets flagged

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
atulgpt: Are you suggesting to delete this TC?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
atulgpt: Added this TC

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
atulgpt: I have renamed this to `comments` It does make sense as any way we will remove `values`

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
TWiStErRob: Can you please add a test case for my above code and https://github.com/detekt/detekt/pull/5981#discussion_r1166060250 to see what's the UX for this and "document" the behavior.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("ClassName")
```
TWiStErRob, line 1: @BraisGabin should we add `ignoreAnnotation` to `ClassName:` for `@Nested` in the Detekt repo config?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
TWiStErRob: Ah, misread one of your commits, never mind. The full comment is being matched, so the `^//` ensures that the matching is correct.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
TWiStErRob: ah, it's just the comment text that mislead me, this is actually testing `comments: [STOPSHIP]`.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
```diff
@@ -235,4 +247,60 @@ class ForbiddenCommentSpec {
...
+            val findings = ForbiddenComment(messageConfig).compileAndLint(comment)
+            assertThat(findings).isEmpty()
+        }
+
+        @Test
+        fun `should report a finding when STOPSHIP is present`() {
+            val comment = "// STOPSHIP to express in the preview that it's not a normal TextView."
+            val findings = ForbiddenComment(messageConfig).compileAndLint(comment)
+            assertThat(findings).hasSize(1)
+            io.gitlab.arturbosch.detekt.test.assertThat(findings[0])
```
TWiStErRob, line 270: I think you can static import this along with the other one, and the overload resolution will pick the right one.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
atulgpt: Hi @TWiStErRob, I checked the text getting returned here, it returns below string
```
good
bad
good
```

So basically Kotlin handles the misalignment here. Although it's possible to detekt that as well but I think first i have to see if there is equivalent rule in `Ktlint` or not if not then we can implement the same in `detekt`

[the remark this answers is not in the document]
TWiStErRob: Huh, that's good on one level (no need for messing about with ` * ` in regexes), but obviously forces custom processing if needed. It's interesting that for single line comments we get the slashes, but for kdoc we don't get the stars. It's probably worth calling out what to expect in the docs of the rule, so it's easier to write regexes without having to learn PSI debugging.

I think the space matching is resolved then in this thread. There's a way to match `//foo` now, and the `*bad` is less likely/hard to implement. I'll not resolve it to remind you update the docs.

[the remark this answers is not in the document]
atulgpt: Yes @TWiStErRob as for comment we are processing for `commentPsiElement.text` while for `KDocSection` we are processing for `KDocSection.getContent()` rather than `KDocSection.parent.text`.

One more option here is to process for `KDocSection.text` which will also include `/**`, `*` and `*/`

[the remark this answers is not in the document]
TWiStErRob: ah, makes sense, @BraisGabin @cortinico thoughts on this thread?

BraisGabin (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
BraisGabin: I like that idea!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
cortinico: @TWiStErRob which input are you looking for?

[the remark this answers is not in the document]
BraisGabin: I think that this rule shouldn't check if a comment is properly formatted. We should keep it simple. I think that check against what atulgpt show on https://github.com/detekt/detekt/pull/5981#discussion_r1186751170 is good enough.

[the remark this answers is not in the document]
TWiStErRob: @cortinico in general trimming as noted above by Brais and the [original question](https://github.com/detekt/detekt/pull/5981#discussion_r1163249750).

More succinctly: what should `^` mean in regex?

IMO anything between comment delimiters `//`-`\n` and `/*`-`*/` and `/**`-`*/` is part of the comment, but multiline and leading ` * ` complicates things. In the current implementation single line comments (`//`) are not processed at all, allowing for exact matching in whatever form (including leading whitespace):
```
// foo
^
```
AND multiline `/*` comments are processed by the PSI parser removing ` * ` with some black-box logic:
```
/**
 * foo
   ^
 * bar
 */
```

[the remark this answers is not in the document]
BraisGabin: > @BraisGabin "We should keep it simple" - we're talking about regex here 😅

Fair point!

My two cents:
- The comment delemiters should be removed (`//`, `/*` /**`, `*/`)
- The content, as provided by the psi should be trimmed
- That's it.

I don't think this rule needs anything more complex than that. And if it is not possible to make it that simple I would vote to keep it as it is. I think that this rule archive the 99% of the use cases. If we can add patterns easily, go for it. If we need to talk too much about it I would vote to just don't do it.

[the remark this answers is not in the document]
TWiStErRob: I start to agree now, except the trimming part, let's not add more complexity. PSI does trimming already, anything on top of that might be unexpected or unnecessary.

This would mean we can start using `.getContent()` on every type of comment without any post-processing, the simplest there is. That is, if normal comments have the same... This would make it impossible to do [indentation or misaligned matching](https://github.com/detekt/detekt/pull/5981#discussion_r1165642823), but maybe this "ForbiddenComment" can be re-documented to only deal with the contents (human text) of the code, and not style/indentation and the like.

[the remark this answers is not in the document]
BraisGabin: `.getContent()` feels like a good compromise. So we have a decission here then :). Does this have sense to you, @atulgpt?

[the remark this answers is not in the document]
TWiStErRob: Doesn't the current impl include the leading `// `? We need to strip that to emulate `getContent()` for single-line:

`fun PsiComment.getContent(): String = text.removePrefix("//").removePrefix(" ")`

unless this method already exists somewhere as an extension.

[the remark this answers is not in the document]
atulgpt: Hi @TWiStErRob I am not sure of the existing extension. Just stripping `// ` will not work we then also have to handle below as well
```
/* comment */
```,
```
/*
some comment
*/
```
as above two are considered as comment and we take `text` for those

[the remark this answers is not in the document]
Comment (TWiStErRob) [reply detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt]: ugh, so `/*` is also `PsiComment`, and doesn't have a `.getContent()`?

Have a look at `.getContent()` impl then, and let's see if it's easy to mimic for `/*`, because even that can have this style:
```
/*
 * some comment
 */
```
e.g. https://github.com/cashapp/paparazzi/blob/bdd7301db7ed839629d6e813ed955a9f9a9bea93/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt#L1-L16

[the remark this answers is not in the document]
Comment (atulgpt) [reply detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt]: Hi, @TWiStErRob yes `PsiComment` it doesn't have `.getContent()`. And extracting for content is feasible but will be involved and might have some edge cases

File: detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt (hunk 8 above)
Comment (TWiStErRob) [reply detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt:1]: Such a good idea that you already did it in #6019 3 weeks ago 😂 (and I reviewed it 😅)

The root cause is that our detekt rule is named `ClassNaming`, and this is the IJ inspection called `ClassName` with the default setting being: `[A-Z][A-Za-z\d]*`. The problem is that `.idea/inspectionProfiles/Project_Default.xml` is not version controlled, and the default doesn't account for nicely named `@Nested` classes, so anyone who didn't customize their inspection is using the default.

If we had the inspection profile in git, we could turn it off for tests:
image

@cortinico @BraisGabin thoughts on adding that XML and setting up a few things?

File: detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt (hunk 1 above)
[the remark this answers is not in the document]
Comment (TWiStErRob) [reply detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt]: I would say we can accept those edge cases, because it's better detekt takes on the responsibility to parse the comments than have every single user try to figure out a regex that works on all those edge cases or even simple cases. It's probably more involved than is feasible to solve inside the regex language, so we need a bit of Kotlin code to parse this.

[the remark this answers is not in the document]
Comment (atulgpt) [reply detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt]: Ohhhhkay... Let me update the PR

File: detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt [hunk 10]
```diff
@@ -38,7 +38,13 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
...
 
     @Configuration("forbidden comment strings")
-    private val values: List<String> by config(listOf("FIXME:", "STOPSHIP:", "TODO:"))
+    @Deprecated("Use `comments` instead, make sure you escape your text for Regular Expressions.")
+    private val values: List<String> by config(emptyList())
+
+    @Configuration("forbidden comment string patterns")
+    private val comments: List<Regex> by config(listOf("FIXME:", "STOPSHIP:", "TODO:")) {
+        it.map(String::toRegex)
+    }
```
marschwar, line 47: I know this a whole new can of worms...
Should we rather use valueAndReason here to allow the users to provide a custom message why the comment is discouraged?

TWiStErRob: Good idea! I don't think it's a big change.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
BraisGabin: I vote against adding IDEA configurations on the project. We will start getting PRs with people editing those files. I prefer to keep it IDE agnostic (I know, we have the logo. But that's really cool and no one will ever change that by accident).

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
atulgpt: Hi, @TWiStErRob @BraisGabin updated the PR with the `getContent` method which removes `/*`, `/* `, `*/`, ` */`, leading `*` and `* `

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
atulgpt: Hi, @marschwar, @TWiStErRob I have updated the `comments` config to use `ValueWithReason` also I have also deprecated the the `customMessage` config since we are allowing `reason` with each pattern it didn't make sense to have separate global message for the rule

TWiStErRob: Nice catch!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -74,13 +89,75 @@
...
+            String.format(DEFAULT_ERROR_MESSAGE, comment.value.pattern, reason)
+        } ?: run {
+            String.format(DEFAULT_ERROR_MESSAGE_WITH_NO_REASON, comment.value.pattern)
+        }
     }
 
     private fun getErrorMessage(value: String): String =
-        customMessage.takeUnless { it.isEmpty() } ?: String.format(DEFAULT_ERROR_MESSAGE, value)
+        @Suppress("DEPRECATION")
+        customMessage.takeUnless { it.isEmpty() } ?: String.format(DEFAULT_ERROR_MESSAGE_WITH_NO_REASON, value)
```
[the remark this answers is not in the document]
TWiStErRob: I don't get this, based on the İssue name and description, the code looks ok 🤔...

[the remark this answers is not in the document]
atulgpt: Hi, @TWiStErRob what code you didn't get? I just changed the variable name from `DEFAULT_ERROR_MESSAGE` to `DEFAULT_ERROR_MESSAGE_WITH_NO_REASON` to make it more clear as we have two type of message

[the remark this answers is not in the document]
atulgpt: `Expected newline` warning seems to be false +ve to me. If this is what you are referring to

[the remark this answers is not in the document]
TWiStErRob: Good point about ktlint, agree about false positive.

https://github.com/detekt/detekt/blob/main/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/AnnotationOnSeparateLine.kt maps to this rule: https://pinterest.github.io/ktlint/0.49.1/rules/standard/#annotation-formatting but there's no example for expression annotations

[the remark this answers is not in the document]
atulgpt: Added issue at `ktlint` at https://github.com/pinterest/ktlint/issues/2028

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
```diff
@@ -230,9 +285,179 @@ class ForbiddenCommentSpec {
...
+                    */
+                """.trimIndent(),
+                """
+                    
+                    a
+                    b
+
+                    c
+                    
+                """.trimIndent()
```
TWiStErRob, line 453: Does this match KDoc.getContent() behavior? we probably want to drop the empty first/last line when it's multiline and the marker was on a separate line (i.e. first|last line was `""` after removing markers.)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -74,13 +89,75 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
...
+}
+
+internal fun String.getCommentContent(): String {
+    return if (this.startsWith("//")) {
+        this.removePrefix("//").removePrefix(" ")
+    } else {
+        this.split("\n").joinToString("\n") {
+            var hasStartCommentMarker = false
+            var hasEndCommentMarker = false
+            it.run {
```
TWiStErRob, line 138: _(Might be superseded by the comment below this one on `hasEndCommentMarker = this...` line)_

Potential readability improvement, curious what you would think about:
```kotlin
this.split("\n").joinToString("\n") { line ->
  line
    .let { fullLine -> ... }
    .let { lineWithoutStartMarker -> ... }
    .let { lineWithoutStartEndMarkers -> ... }
}
```
( `run` to `let` so that the steps can be named + because it's all explicit `this`s anyway)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -74,13 +89,75 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
...
+                if (this.startsWith("/*")) {
+                    hasStartCommentMarker = true
+                    this.removePrefix("/*")
+                        .removePrefix(" ")
+                } else {
+                    this
+                }
+            }.run {
+                if (this.endsWith("*/")) {
+                    hasEndCommentMarker = this.startsWith("*").not()
```
TWiStErRob, line 148: This looks a bit confusing based on variable name. I think we might be able to get down to two `run`s without variables: one to process start:
```kotlin
if (this.startsWith("/*")) {
    this.removePrefix("/*")
        .removePrefix(" ")
} else if (this.startsWith(" *")) {
    this.removePrefix(" *")
        .removePrefix(" ")
} else if (this.startsWith("*")) {
    this.removePrefix("*")
        .removePrefix(" ")
} else {
    this
}
```
and one to process end marker: same as before.

`hasEndCommentMarker` in the current implementation could be inlined to the last `run`,  
and  
`hasStartCommentMarker` was a simulation of `else if` accross `run`s.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
```diff
@@ -230,9 +285,179 @@ class ForbiddenCommentSpec {
...
+            assertThat(findings).hasSize(1)
+        }
+    }
+
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    @Nested
+    inner class `comment getContent` {
+
+        @Suppress("LongMethod", "UnusedPrivateMember")
+        private fun getCommentsContentArguments() = listOf(
```
TWiStErRob, line 355: I think some tests are missing, e.g.: 
 * empty comments `//`, `/**/`
 * blank comments that collapse into `""`: `// `, `/* */`, `/*\n*/`
 * blank comments that collapse into whitespace: `//  `, `/*  */`
 * star-aligned `/*` comments: `/*\n * foo\n * bar\n */` -> (ideally) `foo\nbar`  
   (+similar cases as the existing multiline tests)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
```diff
@@ -90,6 +117,34 @@ class ForbiddenCommentSpec {
...
         }
+
+        @Test
+        fun `should report violation in KDoc 1`() {
+            @Suppress("MultilineRawStringIndentation")
+            val code = """
+                /**
+                 * TODO: I need to fix this.
+* left shifted asterik
+                            * right shifted astrik
```
TWiStErRob, line 128:
```suggestion
* left shifted asterisk
                            * right shifted asterisk
```
etc.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
TWiStErRob: Nice one! It doesn't look too daunting. Let's close this thread and continue in others, if need be.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -74,13 +89,75 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
...
-        const val DEFAULT_ERROR_MESSAGE = "This comment contains '%s' " +
+        const val DEFAULT_ERROR_MESSAGE_WITH_NO_REASON = "This comment contains '%s' " +
             "that has been defined as forbidden in detekt."
+
+        const val DEFAULT_ERROR_MESSAGE = "This comment contains '%s' " +
+            "that has been forbidden: %s."
+    }
+}
+
+internal fun String.getCommentContent(): String {
```
TWiStErRob, line 131: I'm a bit afraid to ask 😅, but what do you think about using this new method for KDoc too for consistent behavior between `/*` and `/**`?

atulgpt: I am okay with it but the inbuilt `getContent` of `/**` also has the concept of code indentation which my implementation doesn't have. Also, it parses `KDocTag` which is a kind of span in the whole string while I will be using `String` as my input so I will be processing with less info here.

Also to me, it seems we are also digressing from the original issue of this PR 😅

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
atulgpt: This was a debugging TC which got committed by me accidentally 😅. Will remove it

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
atulgpt: TBH I prefer `run` as in my mental model I associate `run` with distinct operation run on the receiver scope. While I use `let` only for scoping a receiver(I personally ignore the return part `let` and hence this approach didn't come to me). Although you are right about using `let` enables us to add names to `it` making it more readable. So I made this change(Bonus points this could also be pitched as a rule in detekt 😛. Although it would be a controversial one)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
atulgpt: Also `/**   */` inbuilt `getContent` for this returns `""`

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
atulgpt: If you start processing for the if block `this.startsWith("*")` since this is processing before end processing you might remove `*` of `*/` marker. That will create the wrong result. 
Take the example of the below comment
```
/*
* a
* /
```
Your example will process all these three lines(unless we again add one more check that `this.startsWith("*") && this.startsWith("*/").not()`)

Which will make it equally complicated

atulgpt: Also I removed `hasEndCommentMarker` that was not required

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
atulgpt: Incorporating your comment at https://github.com/detekt/detekt/pull/5981#discussion_r1194805250 also fixed the star aligned comment issue.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
atulgpt: Hi, @TWiStErRob do you know why this didn't get caught when I ran the ` ./gradlew detektMain detektTest detektTestFixtures apiDump generateDocumentation compileKotlin compileTestKotlin compileTestFixturesKotlin -PwarningsAsErrors=true --parallel` command locally? Running the command successfully passed on my local

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
TWiStErRob: > it seems we are also digressing from the original issue of this PR 😅

Somewhat, but probably still on point: comments are complex, regex is complex, since we're allowing regex on comments, it gets hairy. Regex comes with a lot of extra functionality that user can leverage, but only if the input is in a good form. `^`, `$`, `\A`, `\Z`, `(?m)`, `(?s)`, and probably many more regex constructs are only usable if the input is without any markers or extras. Sadly there's no well-defined description of what is a "comment", so we have to come up with a convenient one.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
TWiStErRob: Unsure, I checked out the commit (https://github.com/detekt/detekt/pull/5981/commits/e2a3af064762518976391edac9de01e590f6f8b4) and ran `gradlew detektMain detekt build -x test -x functionalTest` and it didn't get flagged either. I also tried `runWithArgsFile` from `detekt-with-type-resolution.yaml` and no dice.

@3flex @BraisGabin any idea what's different on CI?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
TWiStErRob: I think we're in agreement here, "slash-slash-space-space" should collapse into a single "space", probably my wording was confusing.

> Also /**   */ inbuilt getContent for this returns ""

That's strange, because it trims something extra. Anyway, that's an edge case the important is when there's contents, in that case extra leading and trailing space is something users might want to flag as you added tests for 👍🏻.

TWiStErRob (approved):
Really nice! Thank you for your patience. Few tiny detail suggestions and some more test coverage and I think we're there! Please review all the unresolved threads.

I'm happy to write the docs if you want, I can do it tomorrow morning.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
atulgpt: Hi @TWiStErRob I didn't get `so the regex should be able to match that somehow.`. Could you explain this further?

TWiStErRob:
FYI: just to show what power this will give to users https://github.com/detekt/detekt/issues/6090#issuecomment-1553156297
Note: having multiple regexes is still a good addition, because of `reason:`

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -74,13 +89,77 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
...
+    return if (this.startsWith("//")) {
+        this.removePrefix("//").removePrefix(" ")
+    } else {
+        val indentedComment = if (this.contains('\n').not()) {
+            this
+        } else {
+            val lines = this.lines()
+            lines.first() + "\n" + lines.drop(1).joinToString("\n").trimIndent()
+        }
```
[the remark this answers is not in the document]
atulgpt: `Can you please add explicit tests for trimming space before //?` `//` can have any indentation as every `//` is treated as single-line separate comments. Also, I already have below TC
```
@Test
fun `should report a finding matching a pattern contained in multiple single line comments`() {
    val comment = """
        // foo STOPSHIP bar
        // foo STOPSHIP bar
        // foo STOPSHIP bar
    """.trimIndent()
    val findings = ForbiddenComment(messageConfig).compileAndLint(comment)
    assertThat(findings).hasSize(3)
}
```

Is this TC handles your case?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
TWiStErRob: Ignore that, I was hoping to be able to distinguish between `// Foo` and `//Foo`, because IMO `//something` just doesn't look as nice, and it's harder to read, and many people (and IDEs) use it to comment out code; so wanted to be able to forbid it using this rule. But after seeing how complex comments are AND having a reasonable `^` behavior, I gave up on this. Left this open so that we can document what form the input the regex is matching on has.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
TWiStErRob:
```
foo
bar
baz
```
(I think that's what the current code does with it.)

`should report a finding matching a pattern contained in multiple single line comments`
doesn't test indentation, because there's a trimindent there, so for Kotlin PSI the `//` starts at 0th character.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
atulgpt: Actually there is a similar ktlint rule for this. So my thinking was that to enforce overall comment structure(structure around `//`, `/*`, `/**`, `*/`) we should rely on separate rules rather than enforcing those from here

TWiStErRob: Makes sense, that's why we ended up trimming the structure and leaving only the content. I'll check out ktlint's rule.

atulgpt: Wrapped `detekt` rule name is `CommentSpacing`

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
```diff
@@ -1,13 +1,21 @@
...
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments
+import org.junit.jupiter.params.provider.MethodSource
 
 private const val VALUES = "values"
+private const val COMMENTS = "comments"
 private const val ALLOWED_PATTERNS = "allowedPatterns"
 private const val MESSAGE = "customMessage"
```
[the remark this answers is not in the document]
atulgpt: Updated. Will push in next commit

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
atulgpt: I have added below TCs
```
@Test
fun `should report a finding start space contained in multiple single line indented comments`() {
    val comment = """
        class a {
            fun test() {
                //  foo
                val a = 0
            }
        }
    """.trimIndent()
    val findings = ForbiddenComment(
        TestConfig(
            COMMENTS to listOf("^ foo"),
        )
    ).compileAndLint(comment)
}

@Test
fun `should not report when not finding start contained in multiple single line indented comments`() {
    val comment = """
        class a {
            fun test() {
                // foo
                val a = 0
            }
        }
    """.trimIndent()
    val findings = ForbiddenComment(
        TestConfig(
            COMMENTS to listOf("^ foo"),
        )
    ).compileAndLint(comment)
}
```

Also `trimIndent` in TC doesn't matter here as Kotlin PSI considers `//` to be the start of single line comment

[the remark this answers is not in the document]
atulgpt: I have added your two TCs as well. The expected output is `"foo\nbar\nbaz\n"`. Notice end line as we are not trimming that away as per the `KDoc` parsing

atulgpt:
> Really nice! Thank you for your patience. Few tiny detail suggestions and some more test coverage and I think we're there! Please review all the unresolved threads.
> 
> I'm happy to write the docs if you want, I can do it tomorrow morning.

Sure if you have something in mind. Do add it. Thanks for giving so much valuable feedbac

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
TWiStErRob: > PSI considers // to be the start of single line comment irrespective of its indentation

That's an assumption worth testing with a unit test :) Thanks for adding!

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -129,34 +129,44 @@ internal fun String.getCommentContent(): String {
...
-                }
+        this
+            .trimIndentIgnoringFirstLine()
+            // Process line by line.
+            .lineSequence()
+            // Remove starting, aligning and ending markers.
+            .map {
+                it
+                    .let { fullLine ->
+                        val trimmedStartFullLine = fullLine.trimStart()
```
TWiStErRob, line 140: I don't think that's a good idea, because
```
/*
       foo
*/
```
would be trimmed to `foo` (i.e. no spaces) at this point.

TWiStErRob: ah, never mind, I see it now :) The line is trimmed, but it's only used for checking the start markers.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
```diff
@@ -230,9 +259,766 @@ class ForbiddenCommentSpec {
...
+        @Test
+        fun `should report all separate findings at once`() {
+            val comment = """
+                /*
+                 * foo baz
+                 * bar qux
+                 */
+            """.trimIndent()
+            assertThat(ForbiddenComment(TestConfig(COMMENTS to listOf("^foo", "^bar"))).compileAndLint(comment)).hasSize(2)
+        }
```
TWiStErRob, line 693: In addition to this I was thinking about this TC
```kotlin
        @Test
        fun `should report all separate findings at once`() {
            val comment = """
                /*
                 * foo baz
                 * bar qux
                 */
            """.trimIndent()
            assertThat(ForbiddenComment(TestConfig(COMMENTS to listOf("^(foo|bar)")).compileAndLint(comment)).hasSize(2)
        }
```
but fixing it might be not worth the time/complexity. And also we don't have a back-mapping from the comment content to the actual code location, so we can't highlight the right place anyway. Curious what you think about this.

TWiStErRob (commented):
Pushed 6 commits, mainly the docs, and some more test cases and minor fixes. Also noticed a few more things while writing the docs, please have a look.

TWiStErRob (commented):
Pushed 6 commits, mainly the docs, and some more test cases and minor fixes. Also noticed a few more things while writing the docs, please have a look.

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -566,12 +566,14 @@ style:
         value: 'java.lang.annotation.Inherited'
   ForbiddenComment:
     active: true
-    values:
-      - 'FIXME:'
-      - 'STOPSHIP:'
-      - 'TODO:'
```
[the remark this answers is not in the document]
atulgpt: Rather three values would be fine `(?i)\bFIXME\b`, `(?i)\bSTOPSHIP\b`, `(?i)\bTODO\b`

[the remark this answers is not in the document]
TWiStErRob: I think that depends on the wording of the reason, but let's see if we get an agreement on removing the `:` first in favor of this. I think we can do this in a future PR to not drag this more.

btw, 1 pattern is easier to copy to your own config when you want to add more, that's why I would prefer the "or"-style pattern.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
atulgpt: TBH this can also be fixed. But again I agree that it's not worth the effort.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -74,13 +89,84 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
...
+
+    private fun getErrorMessage(comment: Comment): String =
+        comment.reason
+            ?.let { reason -> String.format(DEFAULT_ERROR_MESSAGE, comment.value.pattern, reason) }
+            ?: String.format(DEFAULT_ERROR_MESSAGE_WITH_NO_REASON, comment.value.pattern)
+
+    @Suppress("DEPRECATION")
     private fun getErrorMessage(value: String): String =
-        customMessage.takeUnless { it.isEmpty() } ?: String.format(DEFAULT_ERROR_MESSAGE, value)
+        customMessage.takeUnless { it.isEmpty() } ?: String.format(DEFAULT_ERROR_MESSAGE_WITH_NO_REASON, value)
```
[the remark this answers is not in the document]
atulgpt: What about `KDoc tag should have a value. Matched by your forbidden comment <regex>`. `<regex>` will help identify which expression catched this

detekt-core/src/main/resources/default-detekt-config.yml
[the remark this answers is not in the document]
atulgpt: But `or` will not have a different `reason`(Although these terms are pretty much common I have seen folks who were not aware of these 😅)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -74,13 +89,84 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
...
+
+        comments.forEach {
+            if (it.value.containsMatchIn(text)) {
+                report(
+                    CodeSmell(
+                        issue,
+                        Entity.from(comment),
+                        getErrorMessage(it)
+                    )
+                )
```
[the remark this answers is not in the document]
atulgpt: Updated 👍

atulgpt:
> Pushed ~6 7~ 8 commits, mainly the docs, and some more test cases and minor fixes. Also noticed a few more things while writing the docs, please have a look.

Thanks a lot for contributing in this

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
```diff
@@ -507,6 +507,87 @@ class ForbiddenCommentSpec {
         }
     }
 
+    @Nested
+    inner class `mixed code and comment lines` {
+
+        @Test
+        fun `should report a finding in trailing single line comment`() {
+            val comment = """
+                fun f() {} // TODO implement
```
atulgpt, line 502: hI @TWiStErRob thanks for adding these TCs as well. But from a Kotlin parser point of view, it doesn't matter where that comment is located.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -17,15 +17,75 @@ import org.jetbrains.kotlin.kdoc.psi.impl.KDocSection
...
 /**
  * This rule allows to set a list of comments which are forbidden in the codebase and should only be used during
  * development. Offending code comments will then be reported.
  *
+ * The regular expressions in `comments` list will have the following behaviors while matching the comments:
+ *  * Each comment will be handled individually as a whole.
+ *    * single line comments are always separate, conjoint lines are not merged.
+ *    * multi line comments are not split up, the regex will be applied to the whole comment.
+ *    * KDoc comments are not split up, the regex will be applied to the whole comment.
+ *  * The comment markers (`//`, `/​*`, `/​**`, `*` aligners, `*​/`) are removed before applying the regex.
```
atulgpt, line 30: Hi @TWiStErRob should we mention `// `, `/* `, `* ` and ` */` also as comments marker?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -17,15 +17,75 @@ import org.jetbrains.kotlin.kdoc.psi.impl.KDocSection
...
+ *    One leading space is removed from each line of the comment, after starting markers and aligners.
```
atulgpt, line 31: Should we write `One leading space if present` instead of `One leading space`?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -17,15 +17,75 @@ import org.jetbrains.kotlin.kdoc.psi.impl.KDocSection
...
+ *  * The regex is applied as a multiline regex,
+ *    see [Anchors](https://www.regular-expressions.info/anchors.html) for more info.
+ *    To match the start and end of each line, use `^` and `$`.
```
atulgpt, line 34: do we really need to teach regex here 😅

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
TWiStErRob: I think it's easy enough to search for `KDoc tag should have a value.`, since the configuration will have it as literal, even if not found `[ForbiddenComment]` is good hint to look for that in the detekt config. The regex is a strong implementation detail, for end users (devs) it doesn't matter **how** the detection happened, just what's wrong and what to do with it. The shorter the message, the easier to understand, so putting a potentially complex (could be even multiline) regex in someone's face is probably counterproductive.

If you really want to go ahead I would recommend just a short: `%s Pattern: %s` because it'll be appended with `[ForbiddenComment]` anyway so repeating that would be redundant.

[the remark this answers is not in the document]
atulgpt: `for end users (devs) it doesn't matter how the detection happened` I was thinking more in terms of false +ve and false +ve. As in those cases, having regex in msg itself will help in diagnosing the problem

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
TWiStErRob: It could be parsed differently in different contexts, you never know with PSI :)
And also the implementation might change in the future, and we want to keep compatible behavior in all cases.

I also had a weird problem while writing the KDoc comment, that's why I added ZWSPs.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
TWiStErRob: > Hi @TWiStErRob should we mention // , /* , *  and  */ also as comments marker?

Next line covers it, they're not markers, although it might be easier to write than the current two lines :) I'll have a think.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
atulgpt: I just found our 😓. I am afraid having this space might break somewhere else

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
atulgpt: I removed ZWS and ran `generateDocumentation`. It completed successfully

detekt-core/src/main/resources/default-detekt-config.yml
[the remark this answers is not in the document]
TWiStErRob: IMO something like "Todo marker present." is a perfectly valid message for all 3. They're all to-dos, just different severity level.

Recently I use REVIEW, TODEL, STOPSHIP mostly, because TODO/FIXME are too generic.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
TWiStErRob: Extracted https://github.com/detekt/detekt/issues/6111 so this doesn't get lost among all the other messages.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
TWiStErRob: I would say that's an edge case (like 99.9% - 0.01%), where looking up the detekt.yml is acceptable turnaround compared to the "noise" for everyday usage. Think about the perspective of a feature developer who just wrote a random comment and raised a PR and CI flagged the comment true positively. It's enough for them to know what the pattern author indented them to know.

(Sidenote: false negatives won't have message ;)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
TWiStErRob: Assuming you meant https://github.com/detekt/detekt/pull/5981#discussion_r1199262431 to be here.

> I removed ZWS and ran generateDocumentation. It completed successfully

It does not fail, but there's a rendering glitch in the IDE and in Markdown.

> I just found out 😓. I am afraid having this space might break somewhere else

Don't leave me hanging 😲, what? where?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
TWiStErRob: Tried to improve this in https://github.com/detekt/detekt/pull/5981/commits/f4c59c12da88d27b7e96ea250b017beaac3aaacb

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
TWiStErRob: I don't think it hurts, it's relevant info for the documentation here. Also it's not something that would change (Java/Kotlin regex flavor is pretty much set in stone). And it's very useful to those who are not well-versed in regex (try to get the same info from the linked page). Even those who know regex might not know about: "To match the start and end of the whole comment, use `\A` and `\Z`."

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
atulgpt: Hmm.. Make sense. I will use your original proposal to not include regex then. Will iterate if need arise

detekt-core/src/main/resources/default-detekt-config.yml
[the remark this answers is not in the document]
atulgpt: But `TODO/FIXME` is supported by Intellje as they provide highlighting for those(and hence used widely)
Screenshot 2023-05-20 at 3 37 53 AM

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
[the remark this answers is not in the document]
atulgpt: Updated the msg

detekt-core/src/main/resources/default-detekt-config.yml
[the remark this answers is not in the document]
TWiStErRob: IJ highlights what you configure, but agree the defaults help popularize them

image

Also `STOPSHIP` is not highlighted by default, but has Android Lint support (off by default).

Anyway, we're going a bit off-topic :), the point is that they're all the same type of comment:
"something is written down as a comment rather than having the code ready is not ready in some sense"

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -166,23 +166,18 @@ class ForbiddenComment(config: Config = Config.empty) : Rule(config) {
...
 
     private data class Comment(val value: Regex, val reason: String?)
 
     companion object {
-        const val DEFAULT_ERROR_MESSAGE_WITH_NO_REASON = "This comment contains '%s' " +
-            "that has been defined as forbidden in detekt."
-
         const val DEFAULT_ERROR_MESSAGE = "This comment contains '%s' " +
-            "that has been forbidden: %s"
+            "that has been defined as forbidden in detekt."
```
TWiStErRob, line 180: No need to say "detekt", it's usually very clear that the output is from detekt:
 * Gradle task name listed above warning
 * Looking at Detekt a report file
 * GitHub lists tool name from sarif in every violation
 * etc.

TWiStErRob (approved):
Let's aim to resolve the rest of the threads and GO!

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
atulgpt: No, it not breaking as of now. But having these unicode in documentation which gets rendered to HTML via markdown might break somewhere(latter is my assumption and sorry if I created confusion)

Regarding https://github.com/detekt/detekt/pull/5981#discussion_r1199262431 actually, it is breaking if I remove all. But if I only remove it from `# Non-compliant: /*private*/ fun f() { }` and `# Non-compliant: /** ... @see */` it is working fine

TWiStErRob: No worries, I think it's ok, if it breaks, it's an easy fix; although not sure how else to escape these character sequences (`/**/`). Unicode should be supported in most places nowadays, worst case we get some [weird characters](https://stackoverflow.com/questions/5127744).

detekt-core/src/main/resources/default-detekt-config.yml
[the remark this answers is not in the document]
TWiStErRob: I'll extract this as a separate issue.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
atulgpt: There is also option of html escaping

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
atulgpt: Updated.

atulgpt:
> Let's aim to resolve the rest of the threads and GO!

Yesss... It got stretched way too much but yes it was a nice experience to learn about new things. Thanks for your support

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -10,21 +10,83 @@ import io.gitlab.arturbosch.detekt.api.Severity
...
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
 
+// Note: ​ (zero-width-space) is used to prevent the Kotlin parser getting confused by talking about comments in a comment.
 /**
  * This rule allows to set a list of comments which are forbidden in the codebase and should only be used during
  * development. Offending code comments will then be reported.
  *
+ * The regular expressions in `comments` list will have the following behaviors while matching the comments:
+ *  * Each comment will be handled individually as a whole.
```
cortinico, line 26:
```suggestion
 *  * Each comment will be handled individually
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -10,21 +10,83 @@ import io.gitlab.arturbosch.detekt.api.Severity
...
+ *    * single line comments are always separate, conjoint lines are not merged.
```
cortinico, line 27: what does "conjoint lines" mean?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
```diff
@@ -10,21 +10,83 @@ import io.gitlab.arturbosch.detekt.api.Severity
...
+ *
  * <noncompliant>
  * val a = "" // TODO: remove please
- * // FIXME: this is a hack
+ * /**
+ *  * FIXME: this is a hack
+ *  */
  * fun foo() { }
- * // STOPSHIP:
+ * /* STOPSHIP: */
```
cortinico, line 89: Why `// STOPSHIP` was not valid?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
cortinico: > do we really need to teach regex here 

I had this same feeling. I would favour examples over re-explaining regexes here

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
atulgpt: conjoint means for the below comments
```
// comment 1
// comment 2
```

each comment will be treated seperately

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
cortinico: Sorry I was askign why you changed from `//` to `/* ... */`

atulgpt: I think to show all types of supported comments. In the example, we have shown single line, block comment and kdoc comment. @TWiStErRob to add more to it as he contributed towards writing the doc

TWiStErRob: Yeah, just to demonstrate different types of comments.

cortinico: Great thanks for the clarification

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
atulgpt: I agree `individually as a whole.` getting a little bit complex(due to opposite words). @TWiStErRob thoughts?

cortinico: I don't have super strong opinion on this. This whole rule documentation is over complicated though, and will entice people away from using it. Let's try to simplify the language as much as possible

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
TWiStErRob: I can try to revise this, do you want to remove the explanations and just leave key info like marker removal, and which flags are used?

TWiStErRob: Re examples, I would love to add more, but there's only one compliant/noncompliant block possible :(

cortinico: Let's not fixate too much on the content here right now 👍, and merge this as it is. We can refine it afterwards as it's going to be a doc only PR

atulgpt: Offtopic: Hi @cortinico `but there's only one compliant/noncompliant block possible :(` do we have an issue for this? Do we plan to support multiple compliant non-compliant blocks?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenCommentSpec.kt
TWiStErRob: Doesn't work because the escapes would be inside code blocks, also I think HTML escapes are a Java thing, because they used XML-like markup for JavaDoc.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
TWiStErRob: fixed to consecutive

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
TWiStErRob: I made some trivial fixes and removed explanations of examples.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
TWiStErRob: removed.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt
cortinico: > Do we plan to support multiple compliant non-compliant blocks?

Not at this stage (also that's a separate issue)

TWiStErRob (approved):
Everything is resolved, @cortinico merge as you wish for 1.23

detekt-core/src/main/resources/default-detekt-config.yml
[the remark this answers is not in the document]
TWiStErRob: https://github.com/detekt/detekt/issues/6116

PR #5992, opened 2023-04-13
Comment text in the Issue/PR Template

cortinico:
I'm commenting those lines in the template as folks are not removing those lines and we get them in issues/PRs over and over.


schalkms (approved):
👍

Issue #2713, opened 2020-05-21
Flag unnecessary @Suppress annotations


BraisGabin:
## Expected Behavior
When a `@Suppress` annotation is suppressing anything we should flag it.

## Current Behavior
We don't do that.

## Context
As the code evolves there are `@Suppress` that are obsolete. But you don't remember to remove it so there is a change to add the same problem again. It would be great to have a rule/report (I'm not sure which one would fit better here) to remove those.

There are some "caveats" here:
- How do we know if a suppress is related with detekt or not?
- If a rule is disabled (`active = false`) and we have a `@Suppress` should we mark that `@Suppress` as unnecessary?
- What about other `@Suppress` related with the compiler? For example `@Suppress("DEPRECATION")`

schalkms:
I think this is a hard problem in many ways. You would need to get the suppressed scope right and rerun the suppressed rule based on that to make sure the violation isn't present anymore.

> How do we know if a suppress is related with detekt or not?

You would need to query detekt's list of rules. However, you can't be 100 % sure that a detekt rule is suppressed. There might be rules in other projects with the same name.

> If a rule is disabled (`active = false`) and we have a `@Suppress` should we mark that `@Suppress` as unnecessary?

This would make sense.

> What about other `@Suppress` related with the compiler? For example `@Suppress("DEPRECATION")`

These elements shouldn't be touched IMHO.

gtcompscientist:
We have been discussing this a LOT on our team, because it's often ignored. People are afraid or forget to remove Suppression annotations, especially at the Class/file level. My thoughts are:

- Could be a separate task so that you could essentially set a flag that doesn't acknowledge suppressions
- Suggest removal of any suppressions that are off at the top-level
- Determine if a suppression's scope is too wide (Suppress at class/file level that should be more specific

I also think it would be best as a report, because then it can be tracked over time. (For our team we would always want it empty.)

ephemient:
⠀

gtcompscientist:
I was looking at this the other day, and had a thought about how to generate this:

- Add ability to turn off ALL Suppressions
- Run detekt twice under the hood, just compare.

That would likely require a LOT less work than some of the other discussed paths.

cortinico:
I was thinking about a different approach, but I'm unsure if is feasible:
1. Run Detekt once without Suppression. Save to a file similar to the baseline fine.
1. Have a custom rule that checks for all the Suppress annotation and checks if the item they're annotating contains one of the reported

[...]

PR #6004, opened 2023-04-15
Add documentation for all supported suppression formats

### LINKED ISSUE #6000: Suppression Documentation is incomplete
## Expected Behavior
The [documentation about how to suppress findings](https://detekt.dev/docs/introduction/suppressing-rules) should mention all supported ways to do so.

## Observed Behavior
Currently only `RuleId` and `RuleSetId.RuleId` are mentioned. Missing are `all`, `detekt.all`, `detekt:all`,  `detekt.RuleSetId`, `detekt:RuleSetId`, `detekt.RuleSetId.RuleId` and  `detekt:RuleSetId:RuleId` .

Not sure if `all` really must be documented.

Relates to #5976

marschwar:
Fixes #6000 

~~This PR is meant to support the discussion about suppression formats now and in the future. Specifically we need to decide:~~

* ~~do we want to drop support for the `all` suppression~~
* ~~do we want to keep supporting `.` and `:` as separators and if not which one should we drop~~

~~If you prefer a discussion thread to make those decisions, I am glad to open one.~~

### DIFF
--- a/website/docs/introduction/suppressing-rules.md
+++ b/website/docs/introduction/suppressing-rules.md
@@ -19,15 +19,15 @@ object Constants {
 }
 ```
 
-It is also possible to prefix the rule id with either `detekt` and/or the ruleset id such as `@Suppress("detekt:LongMethod")` or `@Suppress("complexity:LongParameterList")`. 
+It is also possible to prefix the rule id with `detekt` and/or the ruleset id such as `@Suppress("detekt:LongMethod")` or `@Suppress("complexity:LongParameterList")`. 
 
 The following table shows the various supported suppression formats.
 
-| Example Suppression                                                                                                     | Description                                               | Notes                                        |
-|-------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------|----------------------------------------------|
-| ~~`all`~~, ~~`detekt:all`~~, ~~`detekt.all`~~                                                                           | Suppresses all detekt findings.                           | These suppressions are deprecated.           |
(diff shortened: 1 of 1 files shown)


cortinico:
My 2 cents

> * do we want to drop support for the `all` suppression

Potentially yes

> * do we want to keep supporting `.` and `:` as separators and if not which one should we drop

I would say we should support only one of the two, as I don't see the reason to support both.
I don't have a strong opinion on which one to support though

schalkms (commented):
> * do we want to drop support for the `all` suppression

A single option for suppressing everything is enough, IMHO. 

> * do we want to keep supporting `.` and `:` as separators and if not which one should we drop

Dot as single separator is enough, no need for having a second separator.

marschwar:
I am sorry for starting a general discussion inside a pull request. I removed all the deprecation information so this should solve the issue of the documentation being incomplete.

While trying to make up my mind if I prefer `. ` or `:` I came across the likely reason why "all" was added in the first place. It seems like many code generators add `@SuppressWarnings("all")`  to the generated code. So there actually is a valid use case for supporting it.

chao2zhang (approved):

chao2zhang:
The code looks good. I am also in favor of supporting single rule suppression or all rules. Suppressing a rule set seems not ideal and would make https://github.com/detekt/detekt/issues/2713 more difficult (If a violation is effectively suppressed by a specific rule, but has `@Suppress` annotation with rule set, narrowing`@Suppress` annotation from a ruleset to a rule feels challenging #2713)

cortinico (approved):

BraisGabin (approved):
Good one!

Issue #5836, opened 2023-02-25
Report the usage unneccessary function literal in case of `inline fun`


atulgpt:
Report usage of unnecessary creation of function literal to make method inline.

Below are a few examples

## Expected Behavior of the rule
// uncompliant code
```kotlin
inline fun test1(crossinline lambda: () -> Unit) {
    thread {
        lambda()
    }.start()
}
```

// compliant code
```kotlin
fun test(lambda: () -> Unit) {
    thread(block = lambda).start()
}
```
// alternate compliant code
```kotlin
inline fun test1(crossinline lambda: () -> Unit) {
    thread {
        println("lambda called")
        lambda()
    }.start()
}
```

// uncompliant code
```kotlin
inline fun test(lambda: (Int) -> Unit) {
    repeat(3) {
        lambda()
    }
}
```

// compliant code
```kotlin
inline fun test(lambda: (Int) -> Unit) {
    repeat(3, lambda)
}
```

## Context
I came up with this issue while following the discussion at the [comment](https://github.com/detekt/detekt/pull/5666#discussion_r1071220952). There could have some cases which I might not think of yet. So creating this issue for discussion on this proposal

BraisGabin:
This should be a rule for the `performance` rule set.

I imagine that this rule should also check for things like:

```kotlin
fun toDomain(i: Int) = i.toString()

// no compliant
listOf(1).map { it.toDomain() }

// compilant
listOf(1).map(::toDomain)
```

```kotlin
// no compliant
listOf(1).map { it.toString() }

// compilant
listOf(1).map(Int::toString)
```

This case is probably the same as the one before.
```kotlin
// no compilant
val a: (Int) -> String = { it.toString() }

// compilant
val a: (Int) -> String = Int::toString
```

atulgpt:
Hi, @BraisGabin is `UnnecessaryFunctionLiteral` name sounds good to you?

BraisGabin:
The problem of this rule is that I'm not sure what's more easy to read:

```kotlin
.map(String::toBooleanStrict)
//or
.map { it.toBooleanString() }
```

But if we add it in the performance rule set this should be good.

> Hi, @BraisGabin is `UnnecessaryFunctionLiteral` name sounds good to you?

What do you think about `UnnecessaryLambda`?

atulgpt:
> What do you think about `UnnecessaryLambda`?

[...]

PR #6005, opened 2023-04-15
Add `UnnecessaryLambda` rule

atulgpt:
Implements https://github.com/detekt/detekt/issues/5836

### DIFF
--- a/detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryLambda.kt
+++ b/detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryLambda.kt
@@ -1,0 +1,337 @@
+package io.gitlab.arturbosch.detekt.rules.performance
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
+import org.jetbrains.kotlin.builtins.getReceiverTypeFromFunctionType
+import org.jetbrains.kotlin.builtins.isExtensionFunctionType
+import org.jetbrains.kotlin.builtins.isFunctionType
+import org.jetbrains.kotlin.builtins.isSuspendFunctionType
+import org.jetbrains.kotlin.config.LanguageFeature
+import org.jetbrains.kotlin.descriptors.CallableDescriptor
+import org.jetbrains.kotlin.descriptors.CallableMemberDescriptor
+import org.jetbrains.kotlin.descriptors.ClassDescriptor
+import org.jetbrains.kotlin.descriptors.FunctionDescriptor
+import org.jetbrains.kotlin.descriptors.PackageViewDescriptor
+import org.jetbrains.kotlin.descriptors.SimpleFunctionDescriptor
+import org.jetbrains.kotlin.descriptors.ValueDescriptor
+import org.jetbrains.kotlin.descriptors.ValueParameterDescriptor
+import org.jetbrains.kotlin.incremental.components.NoLookupLocation
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtLabeledExpression
+import org.jetbrains.kotlin.psi.KtLambdaExpression
+import org.jetbrains.kotlin.psi.KtNameReferenceExpression
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.psi.KtQualifiedExpression
+import org.jetbrains.kotlin.psi.KtReferenceExpression
+import org.jetbrains.kotlin.psi.KtSimpleNameExpression
+import org.jetbrains.kotlin.psi.KtSuperExpression
+import org.jetbrains.kotlin.psi.KtThisExpression
+import org.jetbrains.kotlin.psi.KtValueArgument
+import org.jetbrains.kotlin.psi.psiUtil.anyDescendantOfType
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
+import org.jetbrains.kotlin.psi2ir.deparenthesize
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.model.DefaultValueArgument
+import org.jetbrains.kotlin.resolve.calls.model.VarargValueArgument
+import org.jetbrains.kotlin.resolve.calls.util.getParameterForArgument
+import org.jetbrains.kotlin.resolve.calls.util.getResolvedCall
+import org.jetbrains.kotlin.resolve.lazy.descriptors.LazyPackageDescriptor
+import org.jetbrains.kotlin.synthetic.SyntheticJavaPropertyDescriptor
+import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.types.isDynamic
+import org.jetbrains.kotlin.types.isError
+import org.jetbrains.kotlin.types.typeUtil.isTypeParameter
+
+/**
+ * Avoid creating unnecessary lambda to call a function where function reference can be used. This may have performance
+ * penalty when compared to directly passing lambda variable.
+ *
+ * <noncompliant>
+ * inline fun test1(lambda: () -> Unit) {
+ *     thread {
+ *         // no compliant
+ *         lambda()
+ *     }.start()
+ * }
+ *
+ * // no compliant
+ * val a: (Int) -> String = { it.toString() }
+ *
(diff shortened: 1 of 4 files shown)


detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryLambda.kt
```diff
@@ -0,0 +1,337 @@
...
+import org.jetbrains.kotlin.types.isDynamic
+import org.jetbrains.kotlin.types.isError
+import org.jetbrains.kotlin.types.typeUtil.isTypeParameter
+
+/**
+ * Avoid creating unnecessary lambda to call a function where function reference can be used. This may have performance
+ * penalty when compared to directly passing lambda variable.
+ *
+ * <noncompliant>
+ * inline fun test1(lambda: () -> Unit) {
```
cortinico, line 58:
```suggestion
 * inline fun test(lambda: () -> Unit) {
```

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryLambda.kt
```diff
@@ -0,0 +1,337 @@
...
+ *     thread {
+ *         // no compliant
```
cortinico, line 60: This comment can be removed

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryLambda.kt
```diff
@@ -0,0 +1,337 @@
...
+ *
+ * <compliant>
+ * fun test(lambda: () -> Unit) {
+ *     thread(block = lambda).start()
+ * }
+ *
+ * val a: (Int) -> String = Int::toString
+ *
+ * fun toDomain(i: Int) = i.toString()
+ * // compliant
```
cortinico, line 83: Similarly I will remove all the `//compliant` comment here

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryLambda.kt
```diff
@@ -0,0 +1,337 @@
...
+            }
+            else -> false to null
+        }
+
+        if (isViolation && errorExpression != null) {
+            report(
+                CodeSmell(
+                    issue,
+                    Entity.from(errorExpression),
+                    "Xc"
```
cortinico, line 133: This is missing

cortinico (changes requested):
The rule logic is quite intricated. Can we vendor the `ConvertLambdaToReferenceIntention.kt` file and keep the rule logic smaller?

atulgpt:
> Can we vendor the `ConvertLambdaToReferenceIntention.kt` file and keep the rule logic smaller?

Currently, I have exactly the same code as `ConvertLambdaToReferenceIntention.kt` same is being followed in other rules as well. What benefits will it serve to keep the code to `ConvertLambdaToReferenceIntention.kt` as it doesn't follow the existing convention and later if we want to add configs, or modify the behaviour of the rule it will be difficult to add those in separate `ConvertLambdaToReferenceIntention.kt`

But if you still feel that it should in separate file then I will change this

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryLambda.kt
atulgpt: Done

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryLambda.kt
atulgpt: Added

atulgpt:
Hi @cortinico / @BraisGabin any idea how to run tests against different Kotlin version?

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryLambda.kt
```diff
@@ -0,0 +1,333 @@
...
+ *
+ * <compliant>
+ * fun test(lambda: () -> Unit) {
+ *     thread(block = lambda).start()
+ * }
+ *
+ * val a: (Int) -> String = Int::toString
+ *
+ * fun toDomain(i: Int) = i.toString()
+ * listOf(1).map { toDomain(it) }
```
marschwar, line 80: Why is this code compliant? When I put this snippet into a kotlin file it does not compile.

atulgpt: Did you put
```
fun toDomain(i: Int) = i.toString()
listOf(1).map { toDomain(it) }
```

both the lines?

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/UnnecessaryLambda.kt
```diff
@@ -0,0 +1,333 @@
...
+import org.jetbrains.kotlin.resolve.lazy.descriptors.LazyPackageDescriptor
+import org.jetbrains.kotlin.synthetic.SyntheticJavaPropertyDescriptor
+import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.types.isDynamic
+import org.jetbrains.kotlin.types.isError
+import org.jetbrains.kotlin.types.typeUtil.isTypeParameter
+
+/**
+ * Avoid creating unnecessary lambda to call a function where function reference can be used. This may have performance
+ * penalty when compared to directly passing lambda variable.
```
marschwar, line 55: Do we have documentation, benchmarks that we could link to further explain the issue? I do not like that we state this as a fact and do not show proof. People might blindly follow the advice without second guessing it. I know I would.

I came across [this post](https://dev.to/vtsen/prefer-function-reference-over-lambda-in-kotlin-wrong-1f8k). I have no idea how credible it is. Does the performance penalty depend on the kotlin compiler version?

marschwar: And applies the performance penalty to all output targets?

atulgpt: Hi, @marschwar, I did not run any benchmark on it. My initial motivation was to prohibit below type of use-case
```
fun test(lambda: () -> Unit) {
    thread {
        lambda()
    }.start()
}
```
and prompt below alternate code
```
fun test(lambda: () -> Unit) {
    thread(block = lambda).start()
}
```

But yes I agree that it will not be beneficial for all the use cases. I have also added `This **may** have` implying that it will not be the case always.

Finally, this rule promotes function ref for all the cases for consistency. But I am also open to suggestions

marschwar: I think your original use case is quite different to what this ended up being. I am sorry, I should have made those points before you put in all the work. 

I do not think that we should promote function references as there are many cases where it is not possible to use a function reference and therefor you might end up with a mix of function refs and lambdas in the same map chain.

Just out of curiosity I decompiled the generated kotlin byte code for the compliant and non-compliant version with IntelliJ and ended up with exactly the same java code. I do not have any experience with this feature but at first sight I do not think detekt should claim that this is or may be a performance issue unless we know for a fact in which circumstances that is the case. 

BTW: Your original use case actually does generate an extra anonymous function in the non-compliant version making it a good candidate for a rule.

marschwar: Interestingly the kotlin compiler seems to do some optimizations as both functions below are decompiled into nearly the same java code.

```kotlin
fun compliant(lambda: (Int) -> Unit) {
    repeat(3, lambda)
}
// and
fun nonCompliant(lambda: (Int) -> Unit) {
    repeat(3) { n -> lambda(n) }
}
```

In this case it may just be a matter of style.

atulgpt: It is because `repeat` is `inline fun`

cortinico:
> But if you still feel that it should in a separate file then I will change this

I think that this case might be worth to have the copied logic in a separate file. The reason is that it's quite extensive and hard to isolate. So if we need to update the vendored logic, it should be easier

BraisGabin:
To me this is not a performance rule. It is a style rule. Probably a controversial style rule. But I think like the idea of this rule.

marschwar:
I think it should be moved to the style rule set. But before you put in more work we should decide on the rule focus. should we do that on the original issue?
I do not like us recommending function references for no apparent reason. To me this feels like is own rule.

BraisGabin:
Moving the conversation back to the issue: https://github.com/detekt/detekt/issues/5836#issuecomment-1516328660 so we can talk more about it there.

marschwar:
I believe this PR is obsolete now since we decided to go back to the drawing board. Can we close it @atulgpt ?

Issue #5968, opened 2023-04-05
version.properties in jar root may cause issues with other plugins 


madorb:
Ran into issues when applying both detekt and OpenWrite gradle plugins. Depending on the _order_ the plugins were declared the behavior was different. 

This turned out to be because both expect a `versions.properties` in the root of the classpath, and there were conflicts. Ideally both would scope this a plugin-specific path directory. 

Please see details at https://github.com/openrewrite/rewrite-gradle-plugin/issues/189

## Expected Behavior
{{versions.properties}} from one plugin should not conflict with another

## Observed Behavior
OpenRewrite would fail if applied after detekt, because it would manage to load the `versions.properties` file from detekt, which did not have the information it expected. 

## Steps to Reproduce
See: https://github.com/openrewrite/rewrite-gradle-plugin/issues/189
Create Project, apply detekt followed by openrewrite. 
Run `./gradlew rewriteDry`

## Context
Leads to inconsistent order-dependent bugs. 

## Your Environment

* Version of detekt used: latest
* Version of Gradle used (if applicable): any

cortinico:
Yup we should either move to `detekt/versions.properties` or `detektversion.properties` to prevent this

PR #6006, opened 2023-04-16
Use `detekt-versions.properties` instead of `versions.properties`

atulgpt:
Maybe fixes https://github.com/detekt/detekt/issues/5968

I am new to Gradle plugins and Gradle-related code. So this may be some issues or partial implementation

### DIFF
--- a/detekt-gradle-plugin/build.gradle.kts
+++ b/detekt-gradle-plugin/build.gradle.kts
@@ -136,7 +136,7 @@ tasks {
     val writeDetektVersionProperties by registering(WriteProperties::class) {
         description = "Write the properties file with the detekt version to be used by the plugin."
         encoding = "UTF-8"
-        destinationFile.set(file("$buildDir/versions.properties"))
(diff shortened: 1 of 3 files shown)


atulgpt:
Oops, all checks failed. It is harder than I thought 😅

cortinico:
nit: `detekt-versions.properties` is probably better

atulgpt:
> nit: `detekt-versions.properties` is probably better

I thought in the folder `detekt` we would keep all types of `*.properties` or other autogenerated files. That's why I went for folder.

BTW do you know how to fix the CI?

cortinico:
> BTW do you know how to fix the CI?

Not sure but I'd give the file rename a try

atulgpt:
> Not sure but I'd give the file rename a try

File rename to `detekt-versions.properties`?

atulgpt:
> > BTW do you know how to fix the CI?
> 
> Not sure but I'd give the file rename a try

Yup it worked

cortinico (approved):

Issue #5903, opened 2023-03-14
Improve default configuration for `FunctionNaming`


BraisGabin:
Right now `FunctionNaming` doesn't check any test folder. But that's a bit too much. Not all the functions inside the test source set are test functions. You can have a lot of helper functions and those should follow the normal convention. More about that here: https://kotlinlang.org/docs/coding-conventions.html#names-for-test-methods

For that reason I think that we should do this change:

```diff
   FunctionNaming:
     active: true
-    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/androidUnitTest/**', '**/androidInstrumentedTest/**', '**/jsTest/**', '**/iosTest/**']
+    ignoreAnnotated: ['Test']
     functionPattern: '[a-z][a-zA-Z0-9]*'
     excludeClassPattern: '$^'
```

I know `@Test` sounds to junit and we don't want frameworks inside our default configuration. But let's be clear, junit is by far the most used framework to run tests. And if that doesn't sound like a good argument: The annotation `kotlin.test.Test` exists. And it is not even `kotlinx`. And it has exactly the same semantics as the junit `@Test` annotations. So, for that reason, I think that it would be sane for our documentation to do that change.

Also it would help with thinks like #5896 (but to make it clear, this issue is not because of that other one)

cortinico:
> I know `@Test` sounds to junit and we don't want frameworks inside our default configuration

I'm a bit against this change for this specific reason you already mentioned, but I understand the rationale behind your proposal.

> But let's be clear, junit is by far the most used framework to run tests

We could argue that Dagger is by far the most used DI framework, and so on.

BraisGabin:
@detekt/maintainers what do you think about this? I would like to hear more opinions.

The reason to close this is clear: framework related. And I would not argue against that. But I also would like you to consider about the annotation `kotlin.test.Test` (that I think is created to mimic the junit one) and that even in the [Kotlin Code Convention they talk about the `@Test` annotation](https://kotlinlang.org/docs/coding-conventions.html#names-for-test-methods). Could this be an execption? Or does this just make the things more complex?

TWiStErRob:
TL;DR: I agree, with additional arguments **for** the change, and I would expand the scope.

[...]

PR #6019, opened 2023-04-20
Improve our configuration of `ClassNaming` and `FunctionNaming`

BraisGabin:
Related with #5903

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -90,17 +90,11 @@ formatting:
 
 naming:
   ClassNaming:
-    excludes: ['**/*Spec.kt']
+    ignoreAnnotated: ['org.junit.jupiter.api.Nested']
   FunctionNaming:
     active: true
-    excludes:
-      - '**/test/**'
-      - '**/androidTest/**'
-      - '**/commonTest/**'
-      - '**/functionalTest/**'
-      - '**/jvmTest/**'
-      - '**/jsTest/**'
-      - '**/iosTest/**'
+    excludes: []
+    ignoreAnnotated: ['Test', 'ParameterizedTest', 'RepeatedTest', 'TestFactory']
   TopLevelPropertyNaming:
     constantPattern: '[a-z][_A-Za-z0-9]*|[A-Z][_A-Z0-9]*'
   InvalidPackageDeclaration:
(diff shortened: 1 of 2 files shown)


config/detekt/detekt.yml
```diff
@@ -90,17 +90,11 @@ formatting:
...
-    excludes:
-      - '**/test/**'
-      - '**/androidTest/**'
-      - '**/commonTest/**'
-      - '**/functionalTest/**'
-      - '**/jvmTest/**'
-      - '**/jsTest/**'
-      - '**/iosTest/**'
+    excludes: []
+    ignoreAnnotated: ['Test', 'ParameterizedTest', 'RepeatedTest', 'TestFactory']
```
TWiStErRob, line 97: Let's go with FQCNs like:

https://github.com/TWiStErRob/net.twisterrob.cinema/blob/69111724bf4f828b3b1d9980222d643e58c5aa3a/Heroku/config/detekt/detekt.yml#L90-L101

config/detekt/detekt.yml
```diff
@@ -90,17 +90,11 @@ formatting:
 
 naming:
   ClassNaming:
-    excludes: ['**/*Spec.kt']
+    ignoreAnnotated: ['Nested']
   FunctionNaming:
```
TWiStErRob, line 94: speaking of https://github.com/TWiStErRob/net.twisterrob.cinema/blob/69111724bf4f828b3b1d9980222d643e58c5aa3a/Heroku/config/detekt/detekt.yml#L90-L101, do you want to update that rule too?

BraisGabin: We have `FunctionMaxLength` disabled AFAIK so there is not need to change our configuration for it.

config/detekt/detekt.yml
BraisGabin: In this case I think that I prefer no to use FQCNs. They are easier to read. I don't see how an annotation called `ParametrizedTest` is not related with a test function.

TWiStErRob: What about nested ([Gradle](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Nested.html)) and testfactory? Those could be anything, and only the JUnit 5 variants should be allowed more relaxed names.

It's shorter to read and write, yes, but if you actually want to understand why and what it is, then the implicitness requires way more brainpower. :)

BraisGabin: Agree `Nested` is way too generic. Changed.

cortinico (approved):

Issue #3865, opened 2021-06-11
Require `use` for `Closeable`s


yogurtearl:
## Expected Behavior of the rule
Any time a reference to `Closeable` is created or comes into scope, it should be only used within the lambda of a [`.use` ](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html)

This should be an error or warning. 

Non-compliant code: 
```kotlin
val myCloseable = MyCloseable()
// do stuff with myCloseable
```

Compliant code: 
```kotlin
MyCloseable().use {
    // do stuff with myCloseable
}
```

## Context

Closing `Closeable`s is easy to get wrong. An exception might be thrown, or you might have multiple `Closeable`s. `.use` help make sure they get closed properly.

schalkms:
Good idea!

BraisGabin:
I like it! But we should be carefull with this snipped:

```kotlin
fun foo(bar: Closable) {
  bar.something()
}
```

This code should be compilant.

cortinico:
Can you provide more examples of compliant/non-compliant code. Having some real world scenario will also help. I like the idea behind this rule but have the gut feeling this could be a bit too strict.

yogurtearl:
We start by limited the non-compliant code to only be on constructor invocations of classes that implement `Closable` or functions that returns a type that implements `Closable` (but don't take a Closable as an argument)

Anything else is compliant. Start with by only flagging these 2 cases, but could expand over time

Non-compliant: 
```kotlin
MyClosable().doStuff()
```

Compliant: 
```kotlin
MyClosable().use { it.doStuff() }
```

Non-compliant: 
```kotlin
val closable = MyClosable()
closable.doStuff()
```

Compliant: 
```kotlin
val closable = MyClosable().use {
   it.doStuff()
}
```

Non-compliant: 
```kotlin
functionThatReturnsClosable().doStuff()
```

Compliant: 
```kotlin
functionThatReturnsClosable().use { it.doStuff() }
```

Non-compliant: 
```kotlin
val closable = functionThatReturnsClosable()
closable.doStuff()
```

Compliant: 
```kotlin
val closable = functionThatReturnsClosable().use {
   it.doStuff()
}
```

Compliant
```kotlin
fun foo(bar: Closable) {
  bar.something()
}
```

Compliant (takes a Closable as a param, so it is exempt)
```kotlin
fun modifyClosable(bar: Closable): Closable { ... }

[...]

PR #6026, opened 2023-04-21
Add MissingUseCall.kt rule

atulgpt:
Fixes https://github.com/detekt/detekt/issues/3865

Will need more work for handling more cases


detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * functionThatReturnsClosable().use { it.doStuff() }
+ * </compliant>
+ */
+@RequiresTypeResolution
+class MissingUseCall(config: Config = Config.empty) : Rule(config) {
+    override val issue: Issue = Issue(
+        javaClass.simpleName,
+        Severity.Warning,
+        "Usage of `Closeable` detected with `use` call. Using `Closeable` without `use` can be problematic " +
```
cortinico, line 54:
```suggestion
        "Usage of `Closeable` detected without `use` call. Using `Closeable` without `use` can be problematic " +
```

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
```diff
@@ -0,0 +1,134 @@
...
+            "as closing `Closeable` may throw exception.",
```
cortinico, line 55: Also I'm unsure this sentence is entirely correct, as the point is about users forgetting to close `Closeable`.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    companion object {
+        private val listOfCloseables = listOf(
+            FqName("java.lang.AutoCloseable"),
+            FqName("java.io.Closeable"),
+        )
+
+        private val useFqNames = listOf(
+            FqName("kotlin.io.use"),
+            FqName("kotlin.use"),
```
cortinico, line 131: I believe this is wrong?

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
```diff
@@ -0,0 +1,134 @@
...
+        var expression: PsiElement? = closeableExpression
+        do {
+            expression = expression?.parent
+        } while (expression is KtDotQualifiedExpression && expression.selectorExpression == closeableExpression)
+        return expression
+    }
+
+    companion object {
+        private val listOfCloseables = listOf(
+            FqName("java.lang.AutoCloseable"),
```
cortinico, line 125: I think this is `kotlin.AutoCloseable`

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
```diff
@@ -0,0 +1,134 @@
...
+                CodeSmell(
+                    issue,
+                    Entity.from(expression),
+                    "${expression.text} doesn't call `use` to access the `Closeable`"
+                )
+            )
+        }
+    }
+
+    private fun isChildOfAutoCloseable(calleeReturnType: KotlinType): Boolean {
```
cortinico, line 75:
```suggestion
    private fun isChildOfCloseable(calleeReturnType: KotlinType): Boolean {
```

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
```diff
@@ -0,0 +1,607 @@
...
+                val myCloseable = MyCloseable()
+            }
+        """.trimIndent()
+        val findings = MissingUseCall().compileAndLintWithContext(env, code)
+
+        assertThat(findings).hasSize(1)
+    }
+
+    @Test
+    fun `does report when _AutoCloseable_ is used without _use_`() {
```
cortinico, line 54: I don't recall us using `_` in test names? Have you found this notation somewhere else?

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
```diff
@@ -0,0 +1,607 @@
...
+                functionThatReturnsClosable().use { it.doStuff() }
+            }
+        """.trimIndent()
+        val findings = MissingUseCall().compileAndLintWithContext(env, code)
+
+        assertThat(findings).hasSize(1)
+    }
+
+    @Test
+    fun `does not report when use is used without binding context`() {
```
cortinico, line 569: Let's remove those 2 tests. This rule should not run without TR

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
```diff
@@ -0,0 +1,134 @@
...
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
+import io.gitlab.arturbosch.detekt.rules.fqNameOrNull
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.js.translate.callTranslator.getReturnType
```
cortinico, line 13: This smells. I see we have it in 2 other places so that's fine for now, but we'll have to find a more generic method

atulgpt: Ohh.. you are talking about `kotlin.js` package?

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
atulgpt: `as the point is about users forgetting to close Closeable.` So that's also an issue I think your mentioned issue slightly bigger one. 

But even if you close the `Closeable`, `use` also does more like properly closing it in `finally` block also if you have some exception while using `Closebale` it records that and then in the `finally` block if we get another exception then add that recorded exception as suppressed throwable of the new exception

It also properly checks if Kotlin supports suppressed exceptions or not. If it doesn't then the original exception gets ignored 😅

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
atulgpt: For `AutoClosebale` `use` is defined in `kotlin.use` while for `Closeable` it is present in `kotlin.io.use`

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
atulgpt: Although I have updated the doc

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
cortinico: yes exactly

BraisGabin (commented):
🤦 I reviewed this PR 8 days ago but I forgot to send them.

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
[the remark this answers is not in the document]
atulgpt: This has been removed

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
```diff
@@ -0,0 +1,607 @@
...
+                override fun close() {
+                    // closing the closeable
+                }
+
+                fun doStuff() {
+
+                }
+            }
+
+            fun functionThatReturnsClosable() = MyCloseable()
```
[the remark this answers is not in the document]
atulgpt: Hi, @BraisGabin thanks for the input. To handle these types of cases should I ignore all warnings for functions which return closeable?

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
```diff
@@ -0,0 +1,607 @@
...
+
+        assertThat(findings).hasSize(1)
+    }
+
+    @Test
+    fun `does not report _Closeable_ anonymous object is used with _use_`() {
+        val code = """
+            import java.io.Closeable
+
+            @Suppress("ObjectLiteralToLambda")
```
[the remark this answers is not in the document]
atulgpt: Yes, I know. This I have only added to remove the yellow lines at the IDE. Those lines really disturbs me hence sometimes I suppress/fix all the warnings in the text code as well

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
atulgpt: Although w/o `_` resulting function name will become confusing. Although I can use `DisplayName` to improve the name. Let me know your thoughts

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
```diff
@@ -0,0 +1,607 @@
...
+
+            fun test() {
+                (if (System.currentTimeMillis() % 2 == 0L) MyCloseable(0) else MyCloseable(1)).use { 
+                    /*no-op*/
+                }
+            }
+        """.trimIndent()
+        val findings = MissingUseCall().compileAndLintWithContext(env, code)
+
+        assertThat(findings).hasSize(2)
```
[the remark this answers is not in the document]
atulgpt: Now this is tricky. If let's say I allow this may be based on `selectorExpression` then again there will false positive in the code like below
```
MyCloseable(0).also{ println("using closeable") }.use { // use }
```

In the above code, logic will fail again. Or I can ignore all the chained calls where there is one invocation of `use`. Let me know your thoughts.

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
```diff
@@ -0,0 +1,567 @@
...
+
+            fun test() {
+                if (System.currentTimeMillis() % 2 == 0L) {
+                    MyCloseable(0) 
+                } else {
+                    MyCloseable(1)
+                }.use { 
+                    /*no-op*/
+                }
+            }
```
t-kameyama, line 134: I think the following case should not be reported.

```kotlin
fun test() {
    val closable = if (System.currentTimeMillis() % 2 == 0L) {
        MyCloseable(0)
    } else {
        MyCloseable(1)
    }
    closable.use {
        /*no-op*/
    }
}
```

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
```diff
@@ -0,0 +1,133 @@
...
+                    Entity.from(expression),
+                    "${expression.text} doesn't call `use` to access the `Closeable`"
+                )
+            )
+        }
+    }
+
+    private fun shouldReport(expression: KtExpression): Boolean {
+        val expressionParent = getParentExpression(expression)
+        return if (expressionParent !is KtDotQualifiedExpression) {
```
t-kameyama, line 103: Should not be reported
```kotlin
import java.io.Closeable
fun foo(): Closeable? = TODO()
fun test() {
    foo()?.use {  }
}
```

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
```diff
@@ -0,0 +1,133 @@
...
+                ?.resultingDescriptor
+                ?.fqNameOrNull() !in useFqNames
+        }
+    }
+
+    private fun getParentExpression(closeableExpression: KtExpression): PsiElement? {
+        var expression: PsiElement? = closeableExpression
+        do {
+            expression = expression?.parent
+        } while (expression is KtDotQualifiedExpression && expression.selectorExpression == closeableExpression)
```
t-kameyama, line 118: Should not be reported
```kotlin
import java.io.Closeable
class Bar {
    fun foo(): Closeable? = TODO()
}
fun test(bar: Bar?) {
    bar?.foo()?.use {  }
}
```

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
atulgpt: Hi @t-kameyama what about below TC? Should this rule address below cases or not?
```
fun test() {
    val closable = if (System.currentTimeMillis() % 2 == 0L) {
        MyCloseable(0)
    } else {
        MyCloseable(1)
    }

    if (Random.nextBoolean()) {
        closable.use {
        }
    }
}
```

t-kameyama: Yes. I belive that we can flag it if `use` function is not called immediately after creating/getting `Closable` instance.

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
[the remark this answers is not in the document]
BraisGabin: I think so. I can not think about a good reason to return a closed closable.

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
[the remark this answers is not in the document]
BraisGabin: I'm always amazed that intellij inspect the code inside raw strings.

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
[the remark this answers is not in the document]
BraisGabin: 🤔 Difficult scenario. Multiple options...

- we could say that any call over a closable could generate an exception and keep the closable open because it wasn't inside the `use`. So, it would be OK to raise an issue over the `also` case. 
- we could be more flexible and allow the chains as you suggested.

I think that the idea of this rule is to avoid the case excluded in my first point so I think it should be OK to raise something like that.

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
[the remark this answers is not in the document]
atulgpt: I have updated this logic

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
atulgpt: Hi, @t-kameyama I have updated the logic

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
atulgpt: Updated the chain logic to handle this

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
```diff
@@ -0,0 +1,230 @@
...
+            else -> {
+                false
+            }
+        }
+    }
+
+    @Suppress("ReturnCount")
+    private fun isPartOfIfElseExpressionReturningCloseable(expression: KtExpression): Boolean {
+        val expressionAfterParens = expression.getParentExpressionAfterParenthesis() ?: return false
+        val (ifExpression, containerExpression) = if (expressionAfterParens is KtContainerNodeForControlStructureBody) {
```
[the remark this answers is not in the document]
atulgpt: Hi, @VitalyVPinchuk @BraisGabin is this false +ve?

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
atulgpt: Also let me know if same logic should be build for `when` expressions. Currently I have not added it as that was making the rule more complex

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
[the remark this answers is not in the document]
VitalyVPinchuk: Strange, there are so many tests, and there's definitely a test for `else-if` clause, but I'll have to check it out.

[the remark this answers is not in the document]
VitalyVPinchuk: @atulgpt Hi
The positive check is legit.
That's because first `else` is just `else` and not `else-if`, followed by binary expression: 
```kotlin
if (expressionAfterParens.parent is KtContainerNodeForControlStructureBody) {
                expressionAfterParens.parent.parent to
                    (expressionAfterParens.parent as KtContainerNodeForControlStructureBody).expression
            } else {
                null
            } ?: return false
```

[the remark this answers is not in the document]
atulgpt: Hi, @VitalyVPinchuk what should be the correct code(as per the rule) in this case?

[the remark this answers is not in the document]
VitalyVPinchuk: What do you think about this?
```kotlin
            } else {
                null
            } ?: return false

```
change to
```kotlin
            } else {
                return false
            }
```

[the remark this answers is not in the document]
atulgpt: But is this related to braces rule? W.r.t to the rule both code snippet looks correct to me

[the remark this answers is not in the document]
atulgpt: Ahh got it.. Thanks for the explanation

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCall.kt
atulgpt: I have this TC at `does not report function returning nullable _Closeable_ used with _use_` let me know if I am missing cases here

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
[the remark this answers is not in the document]
atulgpt: Sure.. then resolving this comment.

BraisGabin (commented):
Some comments and I also added a commit changing a bit the tests. They were kind of repetitive and it was a pain to read that code every test with little differences without knowing if they were meaningless or not. Also this removes 200 lines of tests.

Please, review my commit to make sure that I didn't break any assertion. And I don't care if you rebase/remove this commit. They were far too much changes to add them as suggestions.

atulgpt:
> Some comments and I also added a commit changing a bit the tests. They were kind of repetitive and it was a pain to read that code every test with little differences without knowing if they were meaningless or not. Also this removes 200 lines of tests.
> 
> Please, review my commit to make sure that I didn't break any assertion. And I don't care if you rebase/remove this commit. They were far too much changes to add them as suggestions.

I reviewed the changes. Nice refactoring. Thanks

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingUseCallSpec.kt
```diff
@@ -0,0 +1,996 @@
...
+            import java.io.Closeable
+            import java.io.BufferedReader
+            import java.io.FileReader
+
+            val lines = BufferedReader(
+                FileReader("some_file.txt")
+            )
+        """.trimIndent()
+        val findings = subject.compileAndLintWithContext(env, code)
+        assertThat(findings).hasSize(2)
```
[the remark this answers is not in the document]
atulgpt: isn't the below code better at handling the file resource leak?

```kotlin
FileReader("some_file.txt").use { BufferedReader(it).use { it.lines() } }
```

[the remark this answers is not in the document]
BraisGabin: I don't think that that code is better. It's less readable for something that is not needed.

BraisGabin:
@atulgpt can you check my last comments on this PR? Can we do something about them? I would love to merge this rule but I would like to know if we can do something about them.

To me the more important ones are the first two. The other 3 doesn't seem that important. If adding `null` to the "correct part" will reduce the issue count to `1` I think it is worth to change the test. Otherwise I'm good with them like that.

atulgpt:
> Otherwise

Hi @BraisGabin yes I am looking through your comments. And found one more issue(false positive) on which I am working. Will push those changes soon

atulgpt:
> @atulgpt can you check my last comments on this PR? Can we do something about them? I would love to merge this rule but I would like to know if we can do something about them.
> 
> To me the more important ones are the first two. The other 3 doesn't seem that important. If adding `null` to the "correct part" will reduce the issue count to `1` I think it is worth to change the test. Otherwise I'm good with them like that.

Hi @BraisGabin, With the latest change I will also not be able to detekt following violation

```
fun test() {
    MyCloseable(0).also { 
        println("closed is accessed here")
    }.use { 
        /* no-op */
    }

    MyCloseable(0).apply { 
        println("closed is accessed here")
    }.use { 
        /* no-op */
    }
}
```

Also detecting the above code means code like `FileReader(path).buffereed().use { lines() }` will start flagging as well. As `buffereed` is acting same as `also`

BraisGabin:
> With the latest change

Which one?

atulgpt:
> > With the latest change
> 
> Which one?

I have yet to push the changes. Working on some more refinement

atulgpt:
> > With the latest change
> 
> Which one?

Hi @BraisGabin, Pushed the changes at https://github.com/atulgpt/detekt/tree/65f29619aa8ee6d422f9cbe69caf58961730e964

BraisGabin (approved):
It looks good to me now :) Is there any test case were you are not 100% sure and want to talk about it? Or should we just merge this rule?

atulgpt:
> It looks good to me now :) Is there any test case were you are not 100% sure and want to talk about it? Or should we just merge this rule?

Hi @BraisGabin, yes we can merge it. It will have some issues due to its complex nature of detection. But I can monitor those. I will recommend keeping it experimental for a longer duration.

PR #6028, opened 2023-04-22
Update ktlint to v0.49.0

3flex:


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
```diff
@@ -37,7 +37,7 @@ interface Config {
     /**
      * Is thrown when loading a configuration results in errors.
      */
-    class InvalidConfigurationError(throwable: Throwable? = null /* nullable to not break signature */) :
```
cortinico, line 40: This I guess we don't want to lose

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/MultiLineIfElse.kt
```diff
@@ -1,9 +1,13 @@
 package io.gitlab.arturbosch.detekt.formatting.wrappers
 
-import com.pinterest.ktlint.ruleset.standard.MultiLineIfElseRule
+import com.pinterest.ktlint.rule.engine.core.api.editorconfig.EditorConfigProperty
+import com.pinterest.ktlint.rule.engine.core.api.editorconfig.INDENT_SIZE_PROPERTY
+import com.pinterest.ktlint.ruleset.standard.rules.MultiLineIfElseRule
 import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.config
```
cortinico, line 7:
```
```

This is not really cool :( (could also be a rule).
But effectively having the same symbol lowercase and uppercase is really confusing for the reader

gradle/libs.versions.toml
```diff
@@ -2,7 +2,6 @@
 dokka = "1.8.10"
 jacoco = "0.8.9"
 kotlin = "1.8.20"
-ktlint = "0.48.2"
```
cortinico, line 5: I would keep it here. It makes easier to find the KtLint version right away

cortinico (commented):
Great work @3flex left some initial comments 👍

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
3flex: I checked and removing the nullability is caught by binary-compatibility-validator and throws when apiCheck task runs.

We can leave the comment (as a `// end of line comment`) but it seems redundant now we have tooling that maintains ABI compatibility.

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/MultiLineIfElse.kt
3flex: Agreed, but that's not a new issue introduced by this PR - detekt API needs to change. Are you suggesting something needs to change before this is merged?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
cortinico: I see your point, let's remove it then 👍

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/MultiLineIfElse.kt
cortinico: nope, once I left this comment I thought you were introducing `.config` but it turns out it's already part of the API so not really an issue

cortinico (approved):

cortinico:
I'm merging this as I'm about to do 1.23.0-RC2. If there are any follow-ups, we can still do RC3 👍

config/detekt/detekt.yml
```diff
@@ -73,18 +73,10 @@ formatting:
     active: true
   Filename:
     active: false
-  FunctionReturnTypeSpacing:
```
chao2zhang, line 76: I suppose we still want to keep these previously experimental now standard rules in the configs

detekt-formatting/src/test/kotlin/io/gitlab/arturbosch/detekt/formatting/RulesWhichCantBeCorrectedSpec.kt
```diff
@@ -83,11 +82,4 @@ class RulesWhichCantBeCorrectedSpec {
             .isNotEmpty
             .hasExactlyElementsOfTypes(CodeSmell::class.java)
     }
-
-    @Test
```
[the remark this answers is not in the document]
3flex: ktlint behaviour seemed to change and it failed. I didn't dig further assuming that this now might be a rule which can be corrected.

If this was incorrect a test with the new corrected behaviour could be added back in

config/detekt/detekt.yml
3flex: Config has been removed in the past when other rules became active by default, it seems redundant to leave it in when we're using `buildUponDefaultConfig`.

PR #6034, opened 2023-04-23
Consolidate bots to @detekt-ci

cortinico:
Currently, we do have a variety of different bots interacting on our repo: GitHub Actions, Danger and GitHub Code Scanning.

I'm moving all of them to use @detekt-ci's GitHub user token instead.

### DIFF
--- a/.github/workflows/danger.yaml
+++ b/.github/workflows/danger.yaml
@@ -36,4 +36,4 @@ jobs:
         working-directory: bots/
         run: yarn danger ci
         env:
-          DANGER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DANGER_GITHUB_API_TOKEN: ${{ secrets.DETEKT_CI_GITHUB_USER_TOKEN }}
--- a/.github/workflows/detekt-with-type-resolution.yaml
+++ b/.github/workflows/detekt-with-type-resolution.yaml
@@ -41,6 +41,7 @@ jobs:
         if: success() || failure()
         with:
           sarif_file: build/detekt-report.sarif
+          token: ${{ secrets.DETEKT_CI_GITHUB_USER_TOKEN }}
 
 
   gradle:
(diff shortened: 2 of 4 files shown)


.github/workflows/danger.yaml
```diff
@@ -36,4 +36,4 @@ jobs:
         working-directory: bots/
         run: yarn danger ci
         env:
-          DANGER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DANGER_GITHUB_API_TOKEN: ${{ secrets.DETEKT_CI_GITHUB_USER_TOKEN }}
```
TWiStErRob, line 39: Since you're moving them anyway. I would recommend creating a token for each integration. They might require different permissions. E.g. Danger might post comments, while code scanning creates security check, and labeler changes issue metadata. These are very distinct and it would be easier to see who has access to what with separate tokens. It would make rotating in the future easier as well as any other maintenance, since the purpose of each key would be clear.

cortinico: > They might require different permissions.

I'm not entirely sure but I believe they require "no permissions" other than posting comments (which is provided by default).

I.e. they are `(no scope)` as described here:
https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes

TWiStErRob: Looks like more is needed. There are failures. For example upload sarif needs [security events](https://github.com/detekt/detekt/pull/6034/files#diff-8c876a8a0cae7f70b5e719005baefeffa0163b0dc0aeae546938ff184901c36bR50), which were probably given by a previous token. I assume the permissions given by the `permissions:` only affects the built-in token, otherwise we need to explicitly add the same scopes to the token itself.

TWiStErRob: (or I misread the error message and the new secret token was just not observed)

cortinico: > There are failures

The failure is because the job `detekt with type resolution` is running on `pull_request` and not `pull_request_target`. That means that secrets are not exposed so the token we end up passing is empty. We'll have to refactor the thing a bit so that the workflow executes on `pull_request_target` but checks out the user forks to do so.

schalkms (approved):

Issue #5627, opened 2022-12-17
Add wrappers for new rules in ktlint 0.48


3flex:
## Expected Behavior of the rule
I can enable the following ktlint rules:
* Class naming
* Function naming
* Property naming

## Context
We have wrappers for other ktlint rules. We should add wrappers for the rules added in ktlint 0.48.0 (see #5625).

cortinico:
> There are three new rules that I haven't added wrappers for, since we seem to have the same ones in detekt already:

RE To this: I believe we should add those wrappers. The reason is that we've been doing that in the past to keep feature parity between bare Ktlint and our wrapper. We do already have other rules between first-party & our ktlint wrapper that have overlapping features. I won't diverge now

atulgpt:
Hi @cortinico / @3flex so do we want to add the wrapper or not for these new rules?

cortinico:
As mentioned above, I think we should add wrapper for all the rules

3flex:
It's up for grabs @atulgpt if you want to work on it?

atulgpt:
Sure. Will this conflict with https://github.com/detekt/detekt/pull/6028? Or should I create a PR against your branch?

3flex:
Yes, it will conflict. I've just pushed the last changes for that PR and don't expect to make any more, so it should be safe to work off that branch until it's merged.

atulgpt:
Hi, @3flex I have created a https://github.com/3flex/detekt/pull/17 against your PR(https://github.com/detekt/detekt/pull/6028).

Although I have a few queries, 1. We already have `ClassNaming`(at https://github.com/detekt/detekt/blob/main/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ClassNaming.kt) should we support both rules?
2. While running `./gradlew generateDocumentation` it changed `default-detekt-config.yml` the at https://github.com/3flex/detekt/pull/17/files#diff-f4f2976a0cb59a2678eb657bcd337cd6cb115f5543f02a22ca20b928701f5629R312 is that correct behaviour?

PR #6037, opened 2023-04-24
Add wrapper for KtLint rules

atulgpt:
Fixes https://github.com/detekt/detekt/issues/5627

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -309,7 +309,8 @@ naming:
     active: false
     allowedPattern: '^(is|has|are)'
   ClassNaming:
-    active: false
+    active: true
+    classPattern: '[A-Z][a-zA-Z0-9]*'
   ConstructorParameterNaming:
     active: true
     parameterPattern: '[a-z][A-Za-z0-9]*'
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -309,8 +309,7 @@ naming:
     active: false
     allowedPattern: '^(is|has|are)'
   ClassNaming:
-    active: true
-    classPattern: '[A-Z][a-zA-Z0-9]*'
+    active: false
   ConstructorParameterNaming:
     active: true
     parameterPattern: '[a-z][A-Za-z0-9]*'
--- a/detekt-formatting/src/main/resources/config/config.yml
+++ b/detekt-formatting/src/main/resources/config/config.yml
@@ -21,6 +21,8 @@ formatting:
     active: true
     autoCorrect: true
     indentSize: 4
+  ClassNaming:
+    active: false
   CommentSpacing:
     active: true
     autoCorrect: true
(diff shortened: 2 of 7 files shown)


detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -309,8 +309,7 @@ naming:
     active: false
     allowedPattern: '^(is|has|are)'
   ClassNaming:
-    active: true
-    classPattern: '[A-Z][a-zA-Z0-9]*'
+    active: false
```
cortinico, line 312: Why this changed?

detekt-formatting/src/main/resources/config/config.yml
```diff
@@ -21,6 +21,8 @@ formatting:
     active: true
     autoCorrect: true
     indentSize: 4
+  ClassNaming:
```
cortinico, line 24: I would call this differently like `ClassNamingKtlint` or `ClassNamingFormatting` or `ClassNamingWrapper`

cortinico (changes requested):

detekt-core/src/main/resources/default-detekt-config.yml
atulgpt: Yes, that was my [question](https://github.com/detekt/detekt/issues/5627#issuecomment-1519057082). Ideally it should not be changed but automatically changed. This could be a bug as I have also added the `ClassNaming` rule from KtLint and there was one preexisting `ClassNaming` rule as well

detekt-formatting/src/main/resources/config/config.yml
atulgpt: But do we want `ClassNaming` rule from `Ktlint`? Shouldn't the detekt one be preferred over `KtLint` one? Or do we want to deprecate ours in favour of `KtLint`?

3flex: Yeah I don't know about that naming either, it breaks with convention (it would be the only ktlint rule that has a suffix like that).

Could just call it `ClassName`?

It's true we have duplication between rules but this isn't something we need to solve here.

3flex:
I believe these rules are all enabled by default in ktlint so to keep with convention we would enable them by default here too?

Or do we make an exception and keep them disabled when detekt has an equivalent rule

atulgpt:
> Or do we make an exception and keep them disabled when detekt has an equivalent rule

We should not make two different rules enabled. That can potentially cause problems due to conflicting analysis

detekt-core/src/main/resources/default-detekt-config.yml
atulgpt: In other words, in `detekt` what is a unique identifier for a rule? Is this just rule name or combination of rule provider id and name?

cortinico: > how we handle the same name rules, should generateDocumentation fail in cases like that?

I think that's a bug and we should investigate here. Rules with same ID coudl occur, especially if coming from external packages.

detekt-formatting/src/main/resources/config/config.yml
cortinico: Yup I'm fine with any name, as long as it doens't clash with one of our rule we already have

atulgpt: I have changed the name to `ClassName`

3flex (approved):

cortinico (approved):

PR #5996, opened 2023-04-14
Update dependency Gradle to v8.1

3flex:
https://docs.gradle.org/8.1/release-notes.html

PR #6041, opened 2023-04-24
Reduce eager POM task creation

TWiStErRob:
While reading a Gradle issue @3flex opened, I stumbled upon https://github.com/detekt/detekt/pull/5996/files#diff-8a7fc388e89cfbc32780b6887c089aeb9fecbb1e231a51c553e2c87388af7148R60, which as far as I remember is eager. Also swapped `if` with `withType` since there's no need to add the callback on newer versions of Java, JVM running Gradle is not going to change, not even when Toolchains are used.

### DIFF
--- a/build-logic/src/main/kotlin/packaging.gradle.kts
+++ b/build-logic/src/main/kotlin/packaging.gradle.kts
@@ -57,8 +57,8 @@ publishing {
     }
 }
 
-tasks.withType<GenerateMavenPom> {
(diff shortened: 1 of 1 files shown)


3flex:
I can't believe I forgot `configureEach`!

3flex (approved):

TWiStErRob:
No worries 🫣, I'm setting a milestone as 23 is just around the corner.

PR #6042, opened 2023-04-25
BracesOnWhenStatements: fix false positive for necessary braces

### LINKED ISSUE #6036: `BracesOnWhenStatements` with `necessary` false positive when expression yields a lambda
## Expected Behavior
If I have `BracesOnWhenStatements` active, and use the `necessary` flag, and I have a `when` expression that yields a lambda, I must use braces for each clause.

## Observed Behavior
I get a false positive violation of `BracesOnWhenStatements` when using braces for those clauses.

## Steps to Reproduce
```
val foo: () -> Unit = when(bar) {
  true -> { {} }
  false -> { {} }
}
```

## Your Environment
* Version of detekt used: 1.23.0-RC2

### LINKED ISSUE #6039: Allow an empty block for a branch in a when statement with `BracesOnWhenStatements`
## Expected Behavior

For the following configuration of the `BracesOnWhenStatements` rule, I would expect to be allowed to use an empty block for a branch in a when statement:

```yaml
BracesOnWhenStatements:
  active: true
  singleLine: 'necessary'
  multiLine: 'necessary'
```

## Current Behavior

It is not allowed.

```kotlin
enum class Action {
  DoSomething,
  DoNotDoSomething,
  DoNotDoSomethingWithComment
}

fun test(action: Action) {
  when(action) {
    DoSomething -> println("Doing something") // not a violation
    DoNotDoSomething -> {} // is a violation
    DoNotDoSomethingWithComment -> { /* we're not doing something */ } // is a violation
  }
}
```

t-kameyama:
Fixes #6036
Fixes #6039

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
@@ -303,6 +303,48 @@ class BracesOnWhenStatementsSpec {
                     """.trimIndent(),
                     "->"(2),
                 )
+
+            @TestFactory fun `empty blocks are accepted`() =
+                flag(
+                    """
+                        fun test(i: Int) {
+                            when {
+                                i == 1 -> println(1)
+                                i == 2 -> {}
+                                i == 3 -> {}
+                                else -> println(4)
+                            }
+                        }
+                    """.trimIndent(),
+                    *NOTHING,
+                )
+
+            @TestFactory fun `necessary braces for lambda expression are accepted`() =
+                flag(
+                    """
+                        fun test(bool: Boolean) {
+                            val f: () -> Int = when {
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/BracesOnWhenStatementsSpec.kt
```diff
@@ -303,6 +303,48 @@ class BracesOnWhenStatementsSpec {
...
+                    *NOTHING,
+                )
+
+            @TestFactory fun `unnecessary braces for lambda expression are flagged`() =
+                flag(
+                    """
+                        fun test(bool: Boolean) {
+                            val f: () -> Int = when {
+                                bool -> { { -> 1 } }
+                                else -> { { -> 2 } }
```
BraisGabin, line 341: I didn't know that this was valid Kotlin and that it could be used to disambiguate things like this. Nice!

BraisGabin (approved):

PR #6045, opened 2023-04-26
Fix incomplete `requireRootInDeclaration` check in `InvalidPackageDeclaration`

cketti:
The old code used `String.startsWith()` to check if the package declaration is the root package or a subpackage of it. This check erroneously allows "sibling packages" that start with the root package name, e.g. "com.example_extra" when the root package name is "com.example".

This PR adds a test for that case and fixes the check to work as intended.


cortinico (approved):
Nice catch 👍 Thanks for the fix @cketti

Issue #5896, opened 2023-03-12
Test functions on custom test source sets are treated as non test functions


JavierSegoviaCordoba:
## Expected Behavior
Any test function should allow to use backticks and spaces even if it is not on an "official" test source set

## Observed Behavior
I have created custom source sets for different type tests, `testIntegration` and `testFunctional` and the functions reported on those source sets are marked by detekt, for example:

```kotlin
@Test
fun `android configuration cache clean v1_0_0`() { // marked by detekt
    gradleTestKitTest("gradle-features/android configuration cache clean v1_0_0") {
        beforeTest()
        testConfigurationCache(expectTaskOutcome = TaskOutcome.SUCCESS)
    }
}
```

## Steps to Reproduce
The repo with the function mentioned above can be found [here](https://github.com/JavierSegoviaCordoba/semver-gradle-plugin/blob/a46c34ccd5a462c3cf5750bbd2fe6a3180e32713/semver-project-gradle-plugin/testFunctional/kotlin/com/javiersc/semver/project/gradle/plugin/GradleFeaturesTest.kt)

---

I know it is not a good solution, but Detekt should check if a source set contains the word `test` or `Test` in order to apply some rules or others as, I think, it is impossible to know if a custom source set or configuration is for tests or not.

BraisGabin:
I asume that you are using the default configuration. There that rule have this configuration:

```yaml
    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/androidUnitTest/**', '**/androidInstrumentedTest/**', '**/jsTest/**', '**/iosTest/**']
```

As you can see `functionalTest` is not listed there so it doesn't apply. You can just add it and everything should work as expected.

schalkms:
Since there is nothing actionable here, I close this issue. If not, please leave a comment.

JavierSegoviaCordoba:
I missed this notification.

I think the rule should find for *Test instead of hardcoding some of the multiplatform targets.

@schalkms @BraisGabin should I open a different issue or this is valid for this feature request?

schalkms:
@JavierSegoviaCordoba If #5412 does not already fit for your use case, please add your thoughts to this issue.

schalkms:
@andersfischernielsen to my knowledge, there is no merging but rather a overriding mechanism with `buildUponDefaultConfig = true`.

andersfischernielsen:
[...]

PR #6047, opened 2023-04-26
Broaden test exclusions for default rules already excluding tests

andersfischernielsen:
It is quite common to define additional test projects in Kotlin projects for functional tests or integration tests.

These test projects are currently treated as actual implementation projects by the default Detekt ruleset, leading to inconsistent rule detections between unit tests and functional tests.

This fixes this inconsistent behavior by:
- extending the existing `**/test/**` exclusion to include any folder matching `**/*test*/**`, making e.g. folders named `functest`, `testIntegration`, etc. be treated identically by the default ruleset
- simplifying the existing default rule `exclusions` by merging the existing `*Test` exclusions into a single `**/*Test*/**` exclusion for all rules, matching the existing `'**/androidTest/**', '**/commonTest/**'` exclusions as well as e.g. `functionalTest` or `integrationTest`. 

---

Related: https://github.com/detekt/detekt/issues/5896

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -67,7 +67,7 @@ comments:
     endOfSentenceFormat: '([.?!][ \t\n\r\f<])|([.?!:]$)'
   KDocReferencesNonPublicProperty:
     active: false
-    excludes: ['**/*test*/**', '**/*Test*/**']
+    excludes: ['**/*test*/**', '**/*Test/**']
   OutdatedDocumentation:
     active: false
     matchTypeParameters: true
@@ -75,7 +75,7 @@ comments:
     allowParamOnConstructorProperties: false
   UndocumentedPublicClass:
     active: false
-    excludes: ['**/*test*/**', '**/*Test*/**']
+    excludes: ['**/*test*/**', '**/*Test/**']
     searchInNestedClass: true
     searchInInnerClass: true
     searchInInnerObject: true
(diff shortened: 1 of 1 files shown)


cortinico:
`**/*test*/**` this can have a lot of unintended side effects. Say you have a package name `com.legal.protest`, you're basically excluding the whole sourceset from Detekt

schalkms:
> **/*test*/** this can have a lot of unintended side effects. Say you have a package name com.legal.protest, you're basically excluding the whole sourceset from Detekt

Completely agree. It's better to _explicitly_ specify the sources to include or exclude.

andersfischernielsen:
> `**/*test*/**` this can have a lot of unintended side effects. Say you have a package name `com.legal.protest`, you're basically excluding the whole sourceset from Detekt

@cortinico @schalkms That's fair 😄 

What do you think about extending the existing `excludes` with variants of common functional and integration test folder names, then? We're currently using Detekt in our organization across 60+ repositories, which all include functional tests. The behavioral differences of Detekt between unit tests and functional/integration tests is causing a lot of unhappy engineers for us currently and it seems (based on https://github.com/detekt/detekt/issues/5896) like we're not the only ones seeing this. 

Also, thanks for the quick feedback!

andersfischernielsen:
I'd proposed the initial set of extensions to the current `excludes` to be:
- `**/integrationTest/**`
- `**/functionalTest/**`
- `**/functest/**`
- `**/testIntegration/**`
- `**/testFunctional/**`

.. and others we can come up with.

andersfischernielsen:
If you'd rather not extend the current default ruleset; is there another way to extend (not replace) the set of `excludes` on the default ruleset? We really like the default rules and would love to stick as closely to them as possible 🎉

andersfischernielsen:
Thanks a ton! I'll close the PR then until more users request the same changes 👍🏼

Issue #5965, opened 2023-04-05
Add config option to disable dependency on the `check` task


igorwojda:
## Expected Behavior
When running `app:build` task the `detekt` task runs as well. 

## Current Behavior
When running `app:build` task the `detekt` task does not run. 

## Context
At first glance, it looks like running tests and linters with Gradle `check` (defined as check task dependency) task is a great idea. However, this has one major downside - when PR checks build fail you can't instantly determine the reason for the failure.
Did the unit test fail?
Did the integration test fail? 
Did Linter 1 fail?
Did Linter 2 fail?
...

This can be done by going deeper and reviewing PR check log. In practice, we have to manually read through the (usually very long) log file :-( 

The solution for this is to run all checks in parallel, so the PR itself presents a clear reason for the failure:
![image](https://user-images.githubusercontent.com/530122/230060905-0521b261-5f2a-4265-96fd-9ebe70e3befe.png)

This is a quick and convenient way to determine PR status.

## My setup
I am aware that instead of having a dedicated parameter Detekt [proposes](https://detekt.dev/docs/gettingstarted/gradle/#disabling-detekt-from-the-check-task) disabling this check by filtering tasks:
```
tasks.named("check").configure {
    this.setDependsOn(this.dependsOn.filterNot {
        it is TaskProvider<*> && it.name == "detekt"
    })
}
```
This may work fo some simple scenario, but I am using [Gradle conventions plugins](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins) to setup detekt and for me, the above solution does not work.

## Solution

Other liners such as Spotless are solving this issue by introducing configuration parameters:
```
spotless {
    isEnforceCheck = false
}
```

Add a dedicated config option to control this behavior:
```
detekt {
    isEnforceCheck = false
}
```

cortinico:
> At first glance, it looks like running tests and linters with Gradle `check` (defined as check task dependency) task is a great idea. However, this has one major downside - when PR checks build fail you can't instantly determine the reason for the failure.

Just a heads up that this is a problem of Gradle, not a problem of Detekt. Android Lint integrates in our similar way, and your arguments hold true also for them.

>     isEnforceCheck = false

As this is a recurrent ask, I think we can introduce such a property defaulted to true, so that we'll be have a easier way for folks to disable this if they wish.

BraisGabin:
This seems like a really good idea. @igorwojda do you want to give it a try and summit a PR?

[...]

Issue #3122, opened 2020-10-01
Thoughts on hooking detektMain task to Gradle check task?


mhernand40:
According to https://detekt.github.io/detekt/gradle.html#available-plugin-tasks
>Please note that the `detekt` task is automatically run when executing `gradle check`.

The same does not apply for `detektMain`. I understand that `detektMain` is currently experimental but I was wondering if there are any plans to eventually hook it into `gradle check` as well? Another thought would be to swap `detekt` for `detektMain` when `check` is run on an opt-in basis.

cortinico:
I think this can be added to the items for Detekt 2.0 once we promote Type Resolution to be fully stable #2680

TWiStErRob:
Related discussion: https://github.com/detekt/detekt/discussions/4959

PR #6049, opened 2023-04-26
Add config option to disable dependency on the Gradle check task

atulgpt:
Fixes https://github.com/detekt/detekt/issues/5965

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -142,10 +142,6 @@ abstract class Detekt @Inject constructor(
     @get:Internal
     internal abstract val ignoreFailuresProp: Property<Boolean>
 
-    @get:Input
-    @get:Optional
-    abstract val enforceCheck: Property<Boolean>
-
     @get:Internal
     internal abstract val autoCorrectProp: Property<Boolean>
 
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -142,6 +142,10 @@ abstract class Detekt @Inject constructor(
     @get:Internal
     internal abstract val ignoreFailuresProp: Property<Boolean>
 
+    @get:Input
+    @get:Optional
+    abstract val enforceCheck: Property<Boolean>
+
     @get:Internal
     internal abstract val autoCorrectProp: Property<Boolean>
 
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
@@ -79,6 +79,8 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
 
     var allRules: Boolean = DEFAULT_ALL_RULES_VALUE
 
+    var isEnforceCheck: Boolean = DEFAULT_IS_ENFORCE_CHECK
+
     var buildUponDefaultConfig: Boolean = DEFAULT_BUILD_UPON_DEFAULT_CONFIG_VALUE
 
     var disableDefaultRuleSets: Boolean = DEFAULT_DISABLE_RULESETS_VALUE
@@ -119,6 +121,7 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
         const val DEFAULT_DISABLE_RULESETS_VALUE = false
         const val DEFAULT_REPORT_ENABLED_VALUE = true
         const val DEFAULT_ALL_RULES_VALUE = false
+        const val DEFAULT_IS_ENFORCE_CHECK = true
         const val DEFAULT_BUILD_UPON_DEFAULT_CONFIG_VALUE = false
 
         // This flag is ignored unless the compiler plugin is applied to the project
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
@@ -79,7 +79,8 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
 
     var allRules: Boolean = DEFAULT_ALL_RULES_VALUE
 
-    var isEnforceCheck: Boolean = DEFAULT_IS_ENFORCE_CHECK
+    val isEnforceCheck: Property<Boolean> =
+        objects.property(Boolean::class.java).convention(DEFAULT_IS_ENFORCE_CHECK)
 
     var buildUponDefaultConfig: Boolean = DEFAULT_BUILD_UPON_DEFAULT_CONFIG_VALUE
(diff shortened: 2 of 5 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -79,6 +79,8 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
 
     var allRules: Boolean = DEFAULT_ALL_RULES_VALUE
 
+    var isEnforceCheck: Boolean = DEFAULT_IS_ENFORCE_CHECK
```
cortinico, line 82: This is not correct, it should be a `Property<Boolean>` with:

```
    val isEnforceCheck: Property<Boolean> =
      objects.property(Boolean::class.java).convention(DEFAULT_IS_ENFORCE_CHECK)
```

cortinico (changes requested):

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
atulgpt: Hi, @cortinico updated this part. Thanks for pointing out

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -142,6 +142,10 @@ abstract class Detekt @Inject constructor(
     @get:Internal
     internal abstract val ignoreFailuresProp: Property<Boolean>
 
+    @get:Input
+    @get:Optional
+    abstract val enforceCheck: Property<Boolean>
```
atulgpt, line 147: Hi, @cortinico / @3flex can we somehow rename this property to `enforceCheck`? I was trying to do so but `gradle` expects the Java bean convention and fails to compile the task. Using `isEnforceCheck` in Kotlin creates `isEnforceCheck()` getter which violates the bean naming convention

Also `@get:JvmName` doesn't work in abstract properties

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektPlain.kt
```diff
@@ -26,7 +26,7 @@ internal class DetektPlain(private val project: Project) {
         }
 
         tasks.matching { it.name == LifecycleBasePlugin.CHECK_TASK_NAME }.configureEach {
-            if (extension.isEnforceCheck) {
+            if (extension.isEnforceCheck.getOrElse(DetektExtension.DEFAULT_IS_ENFORCE_CHECK)) {
```
cortinico, line 29: I belive here just `.get()` should be fine as teh default is set in the convention

atulgpt: Updated this

3flex (changes requested):
In the past I've considered implementing this not as an on/off option on the extension, but mirroring the way the Gradle code quality extensions allow this to be configured.

They use a [`sourceSets` option](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.quality.CheckstyleExtension.html#org.gradle.api.plugins.quality.CheckstyleExtension:sourceSets) which allows defining:
> The source sets to be analyzed as part of the check and build tasks.

A toggle makes sense for Spotless because it's only adding a dependency for a single task, but our plugin has a task per sourceset for the "plain" detekt tasks, so configuring per sourceset would be much more flexible. That way you can configure to check main source set on check/build but not any of the test source sets for example, or however you'd like to set it up.

The type resolution tasks are different since they're linked to compilations, which often include multiple sourcesets, so we might need a different parameter for those.

cortinico:
> They use a [`sourceSets` option](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.quality.CheckstyleExtension.html#org.gradle.api.plugins.quality.CheckstyleExtension:sourceSets) which allows defining:

If I'm reading this right, it seems like they don't allow to run checkstyle without being connected to the `check`/`build` task (i.e. you specify the `sourceSets` to be inspected and they will run during `check`).

Our case is a bit different as we want to allow to run detekt even when not being hooked to `check` right?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
atulgpt: Hi, @3flex thanks for reviewing this PR. I have removed `enforceCheck` from `Detekt` task input.
I am still curious if renaming `enforceCheck` was possible or not. Let me know if you know some workaround

atulgpt:
> but our plugin has a task per sourceset for the "plain" detekt tasks, so configuring per sourceset would be much more flexible.

Also, I think every single Gradle project only has one `check` task. Currently, only one `detekt` task is added to `check` task. Although I find one [issue](https://github.com/detekt/detekt/issues/3122) replacing `detekt` to `detektMain` but even after that only one task gets added to check

3flex:
> it seems like they don't allow to run checkstyle without being connected to the check/build task

It's just the task dependencies that are affected by that option:

https://github.com/gradle/gradle/blob/v8.1.1/subprojects/code-quality/src/main/groovy/org/gradle/api/plugins/quality/internal/AbstractCodeQualityPlugin.java#L196-L224

The tasks are still created and enabled and can be run directly. They're just not added as dependents of `check`.

> I think every single Gradle project only has one check task. Currently, only one detekt task is added to check task.

Yes, that's how it is today, but I'm working on a proposed change to the plugin that would change this. Also - if you add the `sourceSets` option, then you could add additional detekt source set dependencies to the check task much more easily in the meantime.

atulgpt:
Hi, @3flex I think when `sourceSet.getTaskName(taskBaseName, null);` returns `detekt` here https://github.com/gradle/gradle/blob/1cf537a851c635c364a4214885f8b9798051175b/subprojects/code-quality/src/main/groovy/org/gradle/api/plugins/quality/internal/AbstractCodeQualityPlugin.java#LL217C40-L217C82 then `check`  task will depend on `detekt` o/w not. Let me know if I am correct or not.

So you are suggesting if we want to remove `detekt` from `check` we need to set the `sourceSet` to empty for `detekt`. Right? I am still not clear on what the end-to-end config will look like.

> but I'm working on a proposed change to the plugin that would change this.

If in your work you are also addressing https://github.com/detekt/detekt/issues/5965 then feel free to assign this to yourself. And I will close this PR

3flex:
> Let me know if I am correct or not.

Detekt doesn't extend AbstractCodeQualityPlugin so the behaviour of that class isn't directly relevant to the current Detekt Gradle Plugin behaviour - apologies if I've misunderstood your comment.

The work I'm doing will eventually lead to a fix but it won't be ready that quickly since it's dependent on some changes being made for detekt 2.0. I think this PR is fine in the meantime though with the way the plugin is currently implemented.

cortinico:
> it's just that the flag may get removed again quite quickly in favour of a new parameter.

I would vote to don't add it then as it's going to be yet another breaking change for the user which we'll have to explain how to migrate from

BraisGabin:
I'm not 100% sure if that means closing this PR or not. But most likely it will not be in 1.23.0 so I changed the milestone.

marschwar:
I don't think this PR will be necessary in 2.0 as we will remove the dependency on check altogether if I recall correctly.

atulgpt:
Let me close this PR. Ideally, code from this PR should not be needed in the final solution. if needed @3flex can look into it

Issue #3534, opened 2021-03-10
Always use Named Arguments with booleans


BraisGabin:
## Expected Behavior of the rule

Noncompliant
```kotlin
User("Brais", false)

getUser(id, true)
```

Compliant
```kotlin
User("Brais", isActive = false)

getUser(id, forceRefresh = true)

getUser(id, refresh)
```

## Context
A boolean as a parameter creates difficult code to review:

For example if you have this `data class`:
```kotlin
User(
  val name: String,
  val isActive: Boolean,
  val isPro: Boolean,
)
```

This code is really dificult to understand: `User("Brais", false, true)`. It's kind of easy to understand that `"Brais"` is the name of the user. But it's impossible to know if this user `isActive` or if it `isPro`. If you want to know it you need check the User definition.

With only one parameter the code is a bit better but the problem is still there.

I know that this is a bit subjective and there are other good cases to use argument parameters not adressed here. But I think that the boolean is a real common case that a lot of detekt users can use.

chao2zhang:
Meanwhile, there are good discussions on https://youtrack.jetbrains.com/issue/KT-14934 and I feel like there are a variety of needs for enforcing named parameter usage. I can see why Jetbrains don't want to step in but we could collect more user feedbacks.

cortinico:
> I know that this is a bit subjective and there are other good cases to use argument parameters not adressed here. But I think that the boolean is a real common case that a lot of detekt users can use.

Is this going to be addressing only data classes?
As I can imagine several scenarios where named parameters are not desired:

```
.setWarningAsError(true)
```

vs

```
.setWarningAsError(warningAsError = true)
// or 
.setWarningAsError(enabled = true)
```

I believe the first snipped should be preferred. Plus by enforcing the naming parameters, users would have to explicitly use parameter names that are not always nicely named (as historically they were not part of the method signatures).

BraisGabin:
> We can also build a new rule to enforce named arguments for all adjacent parameters with the same type.

I really like this idea! But, only adjacent? Why is `foo("hi", true, "chao")` better than `foo("hi",  "chao", true)`? Couldn't we just enforce named arguments if your function have two parameters of the same type?

This rule will also promote the usage of `inline class`es as types as described here: https://jakewharton.com/inline-classes-make-great-database-ids/

Down side: I can see cases where this is not desired: `fun plus(a: Int, b: Int)` I don't care the order of the parameters here...

> Is this going to be addressing only data classes?
[...]

PR #6055, opened 2023-04-29
Add UnnamedParameterUse rule

atulgpt:
Fixes https://github.com/detekt/detekt/issues/3534

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnamedParameterUse.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnamedParameterUse.kt
@@ -1,0 +1,153 @@
+package io.gitlab.arturbosch.detekt.rules.bugs
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
+import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
+import org.jetbrains.kotlin.descriptors.CallableMemberDescriptor
+import org.jetbrains.kotlin.descriptors.ValueParameterDescriptor
+import org.jetbrains.kotlin.load.java.isFromJava
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtValueArgument
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.components.isVararg
+import org.jetbrains.kotlin.resolve.calls.util.getParameterForArgument
+import org.jetbrains.kotlin.resolve.calls.util.getResolvedCall
+import org.jetbrains.kotlin.types.typeUtil.isSubtypeOf
+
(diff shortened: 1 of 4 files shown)


BraisGabin (commented):
This issue doesn't treat booleans as any special type, right? I was reading again my initial proposal on the issue and I'm not sure if my initial issue is solved.

Extracted from there:

Noncompliant
```kotlin
User("Brais", false)

getUser(id, true)
```

Compliant
```kotlin
User("Brais", isActive = false)

getUser(id, forceRefresh = true)

getUser(id, refresh)
```

BraisGabin:
Is it feasible? Should we have a configuration to say which types should always use name arguments? Should we only consider booleans if they are the literals?

atulgpt:
> This issue doesn't treat booleans as any special type, right? I was reading again my initial proposal on the issue and I'm not sure if my initial issue is solved.
> 
> Extracted from there:
> 
> Noncompliant
> 
> ```kotlin
> User("Brais", false)
> 
> getUser(id, true)
> ```
> 
> Compliant
> 
> ```kotlin
> User("Brais", isActive = false)
> 
> getUser(id, forceRefresh = true)
> 
> getUser(id, refresh)
> ```

No @BraisGabin this rule is general to any type and doesn't treat boolean as a special type.

atulgpt:
> Is it feasible? Should we have a configuration to say which types should always use name arguments? Should we only consider booleans if they are the literals?

So basically `onlyCheckLiteral` will have no effect on non-primitive data types. Right? Also are `onlyCheckLiteral` and `typesToCheck` correct config names?

atulgpt:
> > Is it feasible? Should we have a configuration to say which types should always use name arguments? Should we only consider booleans if they are the literals?
> 
> So basically `onlyCheckLiteral` will have no effect on non-primitive data types. Right? Also, are `onlyCheckLiteral` and `typesToCheck` correct config names?

^ @BraisGabin

BraisGabin (approved):

schalkms:
I'm not sure whether this rule should be included into the default rule set. I have the feeling that this rule ends up way more controversial than we expect. Let's see what others think about.

Anyhow, the rule is a good fit for the detekt market place. 🙂

atulgpt:
> I'm not sure whether this rule should be included into the default rule set. I have the feeling that this rule ends up way more controversial than we expect. Let's see what others think about.
> 
> Anyhow, the rule is a good fit for the detekt market place. 🙂

Hi, @schalkms I think this can go to `detekt` given we have `NamedArguments` which deals with similar areas but in a more general way. While this rule helps in actually detecting the cases when bugs can occur when names are not provided. There is more discussion in the GitHub comment at https://github.com/detekt/detekt/issues/3534#issuecomment-795863411 which mentions youtrack issue as well. 
But agree let's hear other opinions as well

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnamedParameterUse.kt
```diff
@@ -0,0 +1,153 @@
...
+            valueArgumentList.arguments.associateWith {
+                callDescriptor.getParameterForArgument(it)
+            }
+        if (allowSingleParamUse && paramDescriptorToArgumentMap.values.distinct().size <= 1) {
+            return
+        }
+
+        val namedArgumentList = valueArgumentList.arguments.map {
+            // No name parameter if it is vararg
+            (it.isNamed() || paramDescriptorToArgumentMap[it]?.isVararg == true) to it
```
VovaStelmashchuk, line 115: Can you move `(it.isNamed() || paramDescriptorToArgumentMap[it]?.isVararg == true)` to some new field, like a 
val isParametrNamed = <your expresion>. I think it will make code more clear.

atulgpt: Updated to use the `isNamedOrVararg` variable

cortinico (approved):

DHosseiny:
Is this rule going to add to the detekt rules?

PR #6060, opened 2023-04-30
Remove dependency on mavenLocal() for functionalTests

cortinico:
Our tests should not depend on `mavenLocal()` at all.
The problem is that `detekt-gradle-plugin` is applying a dependency on `detekt-cli` using the version defined here: 
https://github.com/detekt/detekt/blob/2eccd64bd8f8e71e688a222444356c945896c48f/build-logic/src/main/kotlin/Versions.kt#L5

So whenever that version gets bumped, test starts to fail unless you `publishToMavenLocal` as they'll try to find the next version on Maven Central, which won't be there.

That's also the reason why sometimes those PRs:
- https://github.com/detekt/detekt/pull/6056 

fail on CI as they can't find the newer artifact for `detekt-cli` on Maven Central yet.

This is blocking any automation of the release process essentially, and we should move away from it. See:
- https://github.com/detekt/detekt/discussions/4696

In this PR I'm moving away from using the project version to apply the `detekt-cli` to use the latest published stable. This allows those tests to rely just on `mavenCentral()`.
The dependency will be updated by Renovate bot as we release stable versions (we're doing the same thing for the formatting of the `detekt-gradle-plugin` itself).

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -5,6 +5,8 @@ kotlin = "1.8.21"
 ktlint = "0.49.0"
 junit = "5.9.3"
 contester = "0.2.0"
+# Version used to self-inpsect and format the Gradle Plugin.
+detekt-self-inspection = "1.22.0"
 
 [libraries]
 githubRelease-gradle = "com.github.breadmoirai:github-release:2.4.1"
@@ -53,3 +55,4 @@ download = { id = "de.undercouch.download", version = "5.4.0" }
 gradleVersions = { id = "com.github.ben-manes.versions", version = "0.46.0" }
 pluginPublishing = { id = "com.gradle.plugin-publish", version = "1.2.0" }
 shadow = { id = "com.github.johnrengelman.shadow", version = "8.1.1" }
+detekt-self-inspection  = { id = "io.gitlab.arturbosch.detekt", version.ref = "detekt-self-inspection" }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -5,7 +5,7 @@ kotlin = "1.8.21"
 ktlint = "0.49.0"
 junit = "5.9.3"
 contester = "0.2.0"
-# Version used to self-inpsect and format the Gradle Plugin.
+# Version used to self-inspect and format the Gradle Plugin.
 detekt-self-inspection = "1.22.0"
 
 [libraries]
--- a/detekt-gradle-plugin/build.gradle.kts
+++ b/detekt-gradle-plugin/build.gradle.kts
@@ -11,8 +11,7 @@ plugins {
     `java-test-fixtures`
     idea
     alias(libs.plugins.pluginPublishing)
-    // We use this published version of the detekt plugin to self analyse this project.
-    id("io.gitlab.arturbosch.detekt") version "1.22.0"
+    alias(libs.plugins.detekt.self.inspection)
 }
 
 repositories {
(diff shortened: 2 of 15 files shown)


schalkms (approved):
Glad that this pain point gets removed! 👍

gradle/libs.versions.toml
```diff
@@ -5,6 +5,8 @@ kotlin = "1.8.21"
 ktlint = "0.49.0"
 junit = "5.9.3"
 contester = "0.2.0"
+# Version used to self-inpsect and format the Gradle Plugin.
```
3flex, line 8:
```suggestion
# Version used to self-inspect and format the Gradle Plugin.
```

3flex:
Completely agree with the goal of removing mavenLocal usage, however this seems to make it more difficult to iterate on changes to CLI and the plugin and run functional tests.

When doing something like adding a new CLI flag and a complementary property in the Gradle plugin it seems like there's no way to run the functional tests with the updated CLI dependency?

Related (unmerged) PRs:
- #4905
- #5149

cortinico:
> When doing something like adding a new CLI flag and a complementary property in the Gradle plugin it seems like there's no way to run the functional tests with the updated CLI dependency?

Yeah I agree, but I believe this is an improvement from the current status.
Even today, it's not possible to test the Gradle Plugin with the latest version of the CLI, unless you bump version and `publishToMavenLocal`.

If it helps, I can setup a separate CI job that bumps the version to `0.0.0` a local maven repo (I won't use `mavenLocal()` though) and runs `functionalTest` against it.
Still I would keep the current logic (i.e. relying on the latest release by default) as I want to be able to automate the release process, without having to `publishToMavenLocal` to merge the version bump.

3flex:
That would help on CI but not local builds - unless that new task is attached to the `functionalTest` task as a dependency?

Maybe add that and then outline how one would update CLI & Gradle plugin together and run functional tests using that new task. As long as it's not necessary to update test code to get those tests working together when needed.

cortinico:
> That would help on CI but not local builds - unless that new task is attached to the `functionalTest` task as a dependency?

I don't think we test this flow today with local build, though. As of today, local builds fetch `detekt-cli` from `mavenCentral()`.

They will end up using `mavenLocal()` ONLY if you bump the version number here:
https://github.com/detekt/detekt/blob/2eccd64bd8f8e71e688a222444356c945896c48f/build-logic/src/main/kotlin/Versions.kt#L5

3flex:
I recently found a Gradle feature that would fix that: https://docs.gradle.org/current/userguide/declaring_repositories.html#declaring_content_exclusively_found_in_one_repository

Using this means _only_ Maven Local would be searched for the dependency. If the detekt CLI version is declared as a dynamic version then this would avoid the need to do any version bump as well. The only requirement then is pushing the required dependencies to Maven Local which can be easily setup with task dependencies.

Would this resolve your release issues as well?

cortinico:
> I recently found a Gradle feature that would fix that: [docs.gradle.org/current/userguide/declaring_repositories.html#declaring_content_exclusively_found_in_one_repository](https://docs.gradle.org/current/userguide/declaring_repositories.html#declaring_content_exclusively_found_in_one_repository)

Yup we could use this, yet still we will be forcing to depend on `publishToMavenLocal()` before we can run tests. Publishing to another maven local repo (being not `~/.m2`) would probably be better but still suboptimal.

cortinico:
@3flex what's the best course of action for this? I'd really love to get rid of this setup as it's making the release process extremely complicated

BraisGabin:
I vote to merge. If in the future we find better options we can revert this one and implement the better option but we have an issue with our release and this will improve that process.

3flex:
I'd like to take another look at this - I'm not a fan of adding code to the plugin that's only there to support functional tests.

Also, the issue raised in the PR description:

> whenever that version gets bumped, test starts to fail unless you publishToMavenLocal as they'll try to find the next version on Maven Central, which won't be there.

Should have been solved by #6424.

Agree with the goal of not using Maven local repo at all, but I don't think this should be merged.

3flex:
There's a plugin that should help here: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/tree/main/testkit

> it installs your plugin and all of its local project dependencies into a local file repo

Given releases will be a bit more frequent until V2 is out I'll prioritise this. It might be complicated a bit by our use of a composite build but otherwise it should solve this.

cortinico:
Abandoning as stale. I will have to take another stab at this in the future

Issue #5799, opened 2023-02-21
False positive on `CanBeNonNullable`: not differentiating between `&&` and `||` in some scenarios


eygraber:
If my code looks like this, I get the false positive:

```kotlin
import androidx.compose.foundation.layout.Box
import androidx.compose.foundation.layout.ExperimentalLayoutApi
import androidx.compose.foundation.layout.consumeWindowInsets
import androidx.compose.foundation.layout.padding
import androidx.compose.material3.ExperimentalMaterial3Api
import androidx.compose.material3.Scaffold
import androidx.compose.material3.Text
import androidx.compose.material3.TopAppBar
import androidx.compose.runtime.Composable
import androidx.compose.ui.Modifier

data class ScaffoldState(
  val appBarTitle: String? = null
)

@OptIn(ExperimentalLayoutApi::class, ExperimentalMaterial3Api::class)
@Composable
fun DetektReproScaffold(
  state: ScaffoldState,
  shouldNavigate: Boolean,
  modifier: Modifier = Modifier
) {
  Scaffold(
    modifier = modifier,
    topBar = {
      HtTopAppBar(
        shouldNavigate = shouldNavigate,
        title = state.appBarTitle
      )
    }
  ) { innerPadding ->
    Box(
      modifier = Modifier
        .consumeWindowInsets(innerPadding)
        .padding(innerPadding)
    )
  }
}

@OptIn(ExperimentalMaterial3Api::class)
@Composable
private fun HtTopAppBar(
  shouldNavigate: Boolean,
  modifier: Modifier = Modifier,
  title: String? = null
) {
  if(shouldNavigate || title != null) {
    TopAppBar(
      title = {
        if(title != null) {
          Text(
            text = title
          )
        }
      },
      modifier = modifier
    )
  }
}
```

However, if I move the if outside of the function then the false positive does not occur:

```kotlin
@OptIn(ExperimentalLayoutApi::class, ExperimentalMaterial3Api::class)
@Composable
fun DetektReproScaffold(
  state: ScaffoldState,
  shouldNavigate: Boolean,
  modifier: Modifier = Modifier
) {
  Scaffold(
    modifier = modifier,
    topBar = {
      if(shouldNavigate || state.appBarTitle != null) {
        HtTopAppBar(
          title = state.appBarTitle
        )
      }
    }
  ) { innerPadding ->
    Box(
      modifier = Modifier
        .consumeWindowInsets(innerPadding)
        .padding(innerPadding)
    )
  }
[...]

PR #6064, opened 2023-05-01
CanBeNonNullable: Check parent condition for checking if nullability info is used or not

atulgpt:
Fixes: https://github.com/detekt/detekt/issues/5799

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
@@ -768,7 +768,7 @@ class CanBeNonNullableSpec(val env: KotlinCoreEnvironment) {
                 }
 
                 @Test
-                fun `does report when the parameter is only checked on non-nullity with multiple clauses`() {
+                fun `does report when the parameter is only checked on non-nullity with and condition`() {
                     val code = """
                         fun foo(a: Int?) {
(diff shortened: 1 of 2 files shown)


marschwar (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
```diff
@@ -779,6 +779,42 @@ class CanBeNonNullableSpec(val env: KotlinCoreEnvironment) {
                     assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
                 }
 
+                @Test
+                fun `does not report when the parameter is checked on non-nullity with or condition`() {
+                    val code = """
+                        fun foo(a: Int?, other: Int) {
+                            when {
+                                a != null || other % 2 == 0 -> println(2 + other)
```
BraisGabin, line 787: I always have doubts about this. Basically it is a boolean and makes the function do something if it is true. But probably that's probably work for another rule.

BraisGabin (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
atulgpt: Hi, @BraisGabin can you create an issue about the smell here?

Issue #6070, opened 2023-05-03
CastNullableToNonNullableType: false positive on possibly nullable type parameters


eygraber:
## Expected Behavior
The following code should not violate `CastNullableToNonNullableType`:

```kotlin
inline fun <T, R> combine(
  args: Array<*>,
  crossinline transform: (T) -> R
): R = transform(
  args[0] as T
)
```

## Observed Behavior
It does violate it

## Your Environment
* Version of detekt used: 1.23.0-RC3

atulgpt:
Let me look into it

PR #6071, opened 2023-05-03
Check the SimpleType instead of typeElement

atulgpt:
Fixes: https://github.com/detekt/detekt/issues/6070


BraisGabin (approved):
Thanks for the fast fix!

PR #6091, opened 2023-05-11
Add UseLet style rule

tresni:
I asked a question about my rule in the [#detekt channel](https://kotlinlang.slack.com/archives/C88E12QH4/p1683758297010539), and there seemed to be some appetite for the rule. This is what my rule currently looks like.

I found that in a code base I was working with, there were a lot of instances of `if (x != null) { do_something(x) } else null` and wanted to encourage the use of `?.let` in this specific case. The plugin also looks for `if (x == null) { null } else do_something(x)` as we had some of that too.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
@@ -25,8 +25,10 @@ import org.jetbrains.kotlin.types.isNullable
  * https://github.com/JetBrains/kotlin/blob/f5d0a38629e7d2e7017ee645dc4d4bee60614e93/idea/src/org/jetbrains/kotlin/idea/inspections/collections/UselessCallOnNotNullInspection.kt
  */
 /**
- * In many cases the use of an if/else block that simply returns null can be replaced with a `?.let {}` statement for
- * cleaner code.
+ * The Kotlin stdlib provides some functions that are designed to operate on references that may be null. These
+ * functions can also be called on non-nullable references or on collections or sequences that are known to be empty -
+ * the calls are redundant in this case and can be removed or should be changed to a call that does not check whether
+ * the value is null or not.
  *
  * <noncompliant>
  * val testList = listOf("string").orEmpty()
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
@@ -25,10 +25,8 @@ import org.jetbrains.kotlin.types.isNullable
  * https://github.com/JetBrains/kotlin/blob/f5d0a38629e7d2e7017ee645dc4d4bee60614e93/idea/src/org/jetbrains/kotlin/idea/inspections/collections/UselessCallOnNotNullInspection.kt
  */
 /**
- * The Kotlin stdlib provides some functions that are designed to operate on references that may be null. These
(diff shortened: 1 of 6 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
```diff
@@ -25,10 +25,8 @@ import org.jetbrains.kotlin.types.isNullable
  * https://github.com/JetBrains/kotlin/blob/f5d0a38629e7d2e7017ee645dc4d4bee60614e93/idea/src/org/jetbrains/kotlin/idea/inspections/collections/UselessCallOnNotNullInspection.kt
  */
 /**
- * The Kotlin stdlib provides some functions that are designed to operate on references that may be null. These
- * functions can also be called on non-nullable references or on collections or sequences that are known to be empty -
- * the calls are redundant in this case and can be removed or should be changed to a call that does not check whether
- * the value is null or not.
+ * In many cases the use of an if/else block that simply returns null can be replaced with a `?.let {}` statement for
+ * cleaner code.
```
BraisGabin, line 29: I will not include this in this PR. It could be done in another pr.

BraisGabin (commented):
LGTM. Just remove the documentation update so we can talk about it in other pr because I'm not 100% sure about it. But that shouldn't block the rest of the pr.

marschwar:
Thank you for the rule. It looks really good.

I know this style of tests is really popular but to me it feels harder to read than necessary and hard to maintain. But I may be the only one to see it that way.

I agree with @BraisGabin that the documentation change should be its own pr.

BraisGabin:
I agree they are hard to read. But test all the cases possible cases is really nice. All those tests would be a nightmare to maintain too if they are duplicated time over time... It's a difficult compromise.

marschwar:
I think I might have found a false positive.

```kotlin
    fun testCallToCreateTempFile(s: String?) {
        val x = if (s == null) {
            println("foo")
            null
        } else {
            "x"
        }
    }
```

In my opinion this cannot be easily rewritten using `s?.let`.

tresni:
@marschwar I agree that that is the exception case. We are using that as the times to use `@Suppress` as I don't know a meaningful way to exclude that. In our code base, we found only a single time that we were doing that vs ~100 cases where we should have just used let.  That hit rate was acceptable for us.

It was pointed out that you could rewrite that as `s?.let { "X" } ?: run { println("foo") }` but internally we decided that wasn't really better.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNull.kt
tresni: OMG, I totally just put that on the wrong spot. I'm not sure how that happened. I'll remove that. Thanks.

atulgpt:
> @marschwar I agree that that is the exception case. We are using that as the times to use `@Suppress` as I don't know a meaningful way to exclude that. In our code base, we found only a single time that we were doing that vs ~100 cases where we should have just used let. That hit rate was acceptable for us.
> 
> It was pointed out that you could rewrite that as `s?.let { "X" } ?: run { println("foo") }` but internally we decided that wasn't really better.

Hi, @tresni thanks for contributing towards the `detekt` rule. Also regarding the above false positive you can ignore that if `else` branch has more statements other than returning `null`

BraisGabin:
> Also regarding the above false positive you can ignore that if else branch has more statements other than returning null

Yes. And I think that with that, and a test to demonstrate it, this rule is ready to merge.

marschwar (approved):
Great work. 

@detekt/maintainers We could enable it for the detekt codebase right away. There are no findings.

BraisGabin (approved):

marschwar:
@tresni Could you please activate your rule in the `config/detekt/detekt.yml`.

PR #6092, opened 2023-05-12
Remove unnecessary baselines

TWiStErRob:
* added in #3997
* which triggered reporting #4004
* which was fixed in #4011
* but the baseline was not updated at the same time.

cc @marschwar


marschwar (approved):
Thank you. It's about time we add the capability to find unnecessary suppressions and baseline entries.

Issue #5259, opened 2022-08-31
detekt doesn't respect ktlint `RunAfterRule` constraints


3flex:
## Expected Behavior
detekt runs ktlint rules in the order that ktlint expects.

## Observed Behavior
ktlint has a few different visitor modifiers that add constraints to the order rules are run in.

detekt checks and respects `RunAsLateAsPossible` and `RunOnRootNodeOnly` constraints, but does not check `RunAfterRule`. This can affect results when using autocorrect and individual rules aren't run in the correct order.

See https://github.com/pinterest/ktlint/blob/0.46.1/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/Rule.kt#L39-L60

## Steps to Reproduce
I don't have a reproducer, but discovered this when doing some other work on the formatting rule set.

## Context
Can affect autocorrect.

## Your Environment

* Version of detekt used:
* Version of Gradle used (if applicable):
* Gradle scan link (add `--scan` option when running the gradle task):
* Operating System and version:
* Link to your project (if it's a public repository):

schalkms:
Interesting…
I remember a few bug reports regarding auto-correct. It’s good that we finally discovered the reason for it.
Do you have any ideas how we can fix it in detekt?

3flex:
Probably reimplementing some version of https://github.com/pinterest/ktlint/blob/master/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/internal/RuleRunnerSorter.kt (used in 0.47, in 0.46 the class is `RuleSorter`) and sorting rules accordingly.

3flex:
Not stale

PR #6094, opened 2023-05-12
Replace `KtLintMultiRule`

marschwar:
This relates to #5192 and removes the last `MutliRule` usage in the detekt code base.

Unfortunately I had to change the documentation generation mechanism here as well as it did not work when the rule list has a function call attached such as `RuleSet("formatting", listOf(Rule1(config).sorted())`.


marschwar:
Coming back to [your comment](https://github.com/detekt/detekt/issues/5192#issuecomment-1250374044) @chao2zhang 


> The reason behind this is because there are internal orders to execute ktlint rules, hence we need to extract the ordering logic into Detekt's Analyzer.

Looking at the analyzer it seems to me that the execution order of the rules within a rule set is the same as the order returned by the rule set provider. I have not tried to write a test for it though. Do you think there needs to be some more work around that?

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingProvider.kt
```diff
@@ -23,13 +99,94 @@ class FormattingProvider : RuleSetProvider {
...
+            TypeParameterListSpacing(config),
+        ).sortedWith(FormattingRuleComparator)
+    )
 
     companion object {
         @Configuration("if android style guides should be preferred")
         val android by ruleSetConfig(false)
-
-        @Configuration("if rules should auto correct style violation")
-        val autoCorrect by ruleSetConfig(true)
```
marschwar, line 33: This property was seemingly unused. What am I missing?

BraisGabin:
This can be merged after 1.23.0, right? As it is a big change and it doesn't provide any feature for the user.

marschwar:
Yes, it is just a prerequisite for the deprecation removal and for whatever we end up going with the formatting rule set.

BraisGabin (approved):
😍 LGTM!

chao2zhang:
> Coming back to [your comment](https://github.com/detekt/detekt/issues/5192#issuecomment-1250374044) @chao2zhang
> 
> > The reason behind this is because there are internal orders to execute ktlint rules, hence we need to extract the ordering logic into Detekt's Analyzer.
> 
> Looking at the analyzer it seems to me that the execution order of the rules within a rule set is the same as the order returned by the rule set provider. I have not tried to write a test for it though. Do you think there needs to be some more work around that?

- The ordering logic can be found in [Analyzer](https://github.com/detekt/detekt/blob/ebbb8402c856103dc7d95b192d9f03cb034acdd8/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt#L119) - So the order is actually determined by [FormattingProvider.kt](https://github.com/detekt/detekt/pull/6094/files/37c820c8848749c3688799a86242c03991db638e#diff-8f53a55eb4739b329546dd3dba12c13fff801f2a11f9ad97a51f8aa9bca927b3)'s autoCorrect config.

- Before this PR, all ktlint rules are supposed to run ahead of other detekt rules, simply because `is io.gitlab.arturbosch.detekt.api.MultiRule -> rule.rules.any { it.autoCorrect }`
- For maximum compatibility, we should still consider maintaining this behavior though

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingProvider.kt
chao2zhang: See https://github.com/detekt/detekt/pull/6094#issuecomment-1546740912

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRuleComparator.kt
```diff
@@ -0,0 +1,14 @@
+package io.gitlab.arturbosch.detekt.formatting
+
+/**
+ * This serves as weak heuristic to order the wrapped rules according to their visitor modifiers.
+ * Currently only RunAsLateAsPossible is supported.
+ */
+internal object FormattingRuleComparator : Comparator<FormattingRule> {
+    override fun compare(o1: FormattingRule, o2: FormattingRule): Int {
```
chao2zhang, line 8: Ktlint also defines `VisitorModifier.RunAfterRule`, and we should respect that

marschwar: Yap, that is described in #5259

marschwar:
> * The ordering logic can be found in [Analyzer](https://github.com/detekt/detekt/blob/ebbb8402c856103dc7d95b192d9f03cb034acdd8/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt#L119) - So the order is actually determined by [FormattingProvider.kt](https://github.com/detekt/detekt/pull/6094/files/37c820c8848749c3688799a86242c03991db638e#diff-8f53a55eb4739b329546dd3dba12c13fff801f2a11f9ad97a51f8aa9bca927b3)'s autoCorrect config.
> 
> * Before this PR, all ktlint rules are supposed to run ahead of other detekt rules, simply because `is io.gitlab.arturbosch.detekt.api.MultiRule -> rule.rules.any { it.autoCorrect }`
> 
> * For maximum compatibility, we should still consider maintaining this behavior though

Thank you for pointing this out. I can see why rules that might change the source code (autocorrect==true) should run first. That totally makes sense. But why must the other formatting rules run before all other rules? Honestly this feels more like a coincidence to me.

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRuleComparator.kt
chao2zhang: 😭. Would you mind if I contribute to this PR as well?

marschwar: I do not mind at all. But I would not want to fix the `RunAfterRule` issue with this pr or do you see them as strongly connected?

chao2zhang: Either way is fine to me - It seems that we agreed that this PR can be merged post 1.23

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingProvider.kt
chao2zhang: So this Kotlin property is not used in production
- It is used to generate the docs
- It is used to generate the config: https://github.com/detekt/detekt/blob/main/detekt-formatting/src/main/resources/config/config.yml
- Hence the actual config property "autoCorrect" was previously used

chao2zhang (approved):

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingProvider.kt
marschwar: Thank you. I have reverted the change.

cortinico (approved):

Issue #1430, opened 2019-01-17
Support regex for forbidden imports


tasomaniac:
Forbidden imports rule is really powerful to prevent usages of certain classes especially when the names clash. 

It would be amazing to support regex to forbid usages of multiple packages and classes at once.

poldz123:
Yes, I agree. On my end we have libraries that are a wrapper to a library that we used. Now we do not want our developer to use the wrapped library. Would be better to have a check on those forbidden imports.

schalkms:
This leaves the discussion whether the current `imports` config option should be changed.
Otherwise, there's needs to be a new config parameter.

tasomaniac:
I would say a new config parameter. Otherwise that would be a behavior change in existing configurations since dot becomes any char

PR #6095, opened 2023-05-12
ForbiddenImport: Mark forbiddenPatterns as deprecated

### LINKED ISSUE #6090: `ForbiddenImport` rule: Remove `forbiddenPatterns` config property
## Context

The documentation of the `ForbiddenImport` rule could and should be approved upon. This is addressed in #6088. 

But apart from the documentation to me there is no clear value of having two so very similar configuration properties such as:
* `imports`: the list of imports that are not allowed in _glob_ format (plus a reason).
* `forbiddenPatterns`: a single regex that describes all patterns that are not allowed.

Not only is the naming of the second suboptimal it offers little more flexibility over the first. I would argue that everything that can be specified with `forbiddenPatterns` can be expressed with `imports` instead.

## Proposal
* Mark `forbiddenPatterns` as deprecated in 1.x
* Remove `forbiddenPatterns` in 2.x

marschwar:
Closes #6090

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -581,7 +581,6 @@ style:
   ForbiddenImport:
     active: false
     imports: []
-    forbiddenPatterns: ''
   ForbiddenMethodCall:
     active: false
     methods:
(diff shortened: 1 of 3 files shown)


BraisGabin:
I'm not 100% sure about this. I agree that it is really strange but I guess that it's there because someone asked about it.

I agree that this should be fixed. But I wouldn't deprecate it yet. And take the decission about what to do before 2.0

marschwar:
Please go back to #1430. I don't really get the use case. With the current implementation this can easily be achieved by imports. And I am quite certain it could have back then.

TWiStErRob:
Added a few use cases in https://github.com/detekt/detekt/issues/6090

PR #6103, opened 2023-05-15
Dynamically set the ktlint version in rule docs

### LINKED ISSUE #6099: Links to ktlint rule documentation broken
Ktlint seems to have moved to versioned docs as well and therefor all links on https://detekt.dev/docs/rules/formatting pointing to the documentation of the wrapped rule are broken.

404 https://pinterest.github.io/ktlint/rules/standard/#annotation-formatting
200 https://pinterest.github.io/ktlint/0.49.0/rules/standard/#annotation-formatting
 
Additionally: Rules that were promoted from experimental to the standard ruleset [in version 0.49.0](https://github.com/pinterest/ktlint/releases/tag/0.49.0) still have a link to experimental (e.g. `FunctionStartOfBodySpacing` points to https://pinterest.github.io/ktlint/rules/experimental/#function-start-of-body-spacing).

marschwar:
This fixes #6099 

The current ktlint version is injected into to documentation generation process to be able to provide the correct link in the documentation of the formatting ruleset.


build-logic/src/main/kotlin/CopyDirTask.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputDirectory
+import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.TaskAction
+import javax.inject.Inject
+
+abstract class CopyDirTask : DefaultTask() {
+
+    @get:InputDirectory
+    abstract val source: DirectoryProperty
```
TWiStErRob, line 14: Hmm, feels like this should fail or at least warn about PathSensitivity.

detekt-generator/build.gradle.kts
```diff
@@ -69,14 +74,26 @@ val generateDocumentation by tasks.registering(JavaExec::class) {
...
         "--cli-options",
         cliOptionsFile,
     )
 }
 
+val copyRuleDocsToWebsite by tasks.registering(CopyDirTask::class) {
+    dependsOn(doGenerateDocumentation)
+
+    inputs.files(fileTree(tempDocDir))
+    outputs.files(fileTree(documentationDir))
```
TWiStErRob, line 89: No need for explicit listing, the annotations on the task properties are here exactly to do this.

build-logic/src/main/kotlin/CopyDirTask.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    @get:Input
+    abstract val placeholderValue: Property<String>
+
+    @get:Inject
+    abstract val fs: FileSystemOperations
+
+    @TaskAction
+    fun run() {
+        fs.copy {
```
TWiStErRob, line 30: I learned recently that you never want to use `copy` in Gradle, but `sync`. The difference is that `copy` doesn't clean the output, so if there's a file removed in input, it'll still exists after the operation.

detekt-generator/build.gradle.kts
```diff
@@ -69,14 +74,26 @@ val generateDocumentation by tasks.registering(JavaExec::class) {
...
-        documentationDir,
+        tempDocDir,
         "--config",
         configDir,
         "--cli-options",
         cliOptionsFile,
     )
 }
 
+val copyRuleDocsToWebsite by tasks.registering(CopyDirTask::class) {
```
TWiStErRob, line 85: I think the custom task might not be necessary:
```kotlin
val copyRuleDocsToWebsite by tasks.registering(Sync::class) {
	from(tempDocDir)
	into(documentationDir)
	val replacements = mapOf(
		"KTLINT_VERSION" to libs.versions.ktlint.get(),
                // Added benefit: future replacement can be just added here without any other changes.
	)
	// For up to date checks, see https://github.com/gradle/gradle/issues/861.
	replacements.forEach { (name, value) -> inputs.property(name, value) }
	filesMatching("**/*.html") { // or whatever the files are, can be a list of patterns too
		filter(mapOf("tokens" to replacements), org.apache.tools.ant.filters.ReplaceTokens::class.java)
	}
}
```
\+ `__KTLINT_VERSION__` &rarr; `@KTLINT_VERSION@`, because that's how `ReplaceTokens` formats placeholders, OR:
```
mapOf("tokens" to replacements, "beginToken" to "__", "endToken" to "__")
```
but I think the default is KISS-er.

TWiStErRob (commented):
I was about to report the same issue, nice fix with the rewiring!

detekt-generator/build.gradle.kts
marschwar: Thank you, I will give it a try.

3flex: Yeah this was going to be my feedback too, we don't need a custom task for this and should use Copy or Sync.

3flex:
What do you think about using detekt-generator to do the search/replace instead of using a Gradle task for it?

marschwar:
That also was my first thought. This would have to be done either in 
* `RuleSetPagePrinter` and `RuleSetPrinter` or 
* `RuleSetProviderVisitor` and `RuleVisitor`.

Both options seem counter intuitive to include templating support. On top of that, I could not think of a clever way to expose the current ktlint version to the kotlin code other than environment variables which feels very brittle. 

That is how I ended up with the gradle solution.

marschwar:
You are right. That would work. I still like the post processing with gradle better. It is more flexible and neither the printers nor the collectors feel like the right spot.

TWiStErRob:
Yeah, I trust you on that part, not familiar with those classes.

detekt-generator/build.gradle.kts
```diff
@@ -69,14 +74,30 @@ val generateDocumentation by tasks.registering(JavaExec::class) {
...
-        documentationDir,
+        tempDocDir,
         "--config",
         configDir,
         "--cli-options",
         cliOptionsFile,
     )
 }
 
+val copyRuleDocsToWebsite by tasks.registering(Sync::class) {
```
cortinico, line 85: Let's not do this.

I would rather act on the file `DocumentationCollector.kt`.
That's where we extract `<compliant>` tags and the description.

There we simply pass the description as it is. I would rather introduce a new tag called `<ktlintversion/>` that gets processed inside `DocumentationCollector.kt`.

Here you just need to pass a parameter like `--ktlint-version` or you can generify and have a `--replacement-map` which accepts a key value list nd map `ktlintversion` to `libs.versions.ktlint.get()`

cortinico (changes requested):

detekt-generator/build.gradle.kts
cortinico: oops I just saw the other comments above my bad.

marschwar: If you all prefer to do it in the collector, I can do that as well.

TWiStErRob: Had a quick look, I think it could fit there too. Only the dependency injection is tricky as it has to go through a few layers, but that's kind of normal I think.

marschwar:
I removed the custom task added replacements to the docs generation.

marschwar:
@cortinico should I rebase or close this? I removed the implementation that used gradle and added the functionality to the generator.

cortinico (approved):
Sorry for the late review @marschwar 
I think we can merge this one, I like the way how you implemented it on the Gradle layer

3flex (changes requested):
There are a few new rules added by #6253 which will be merged momentarily... can this be rebased please, and the updates made to the four new rules as well? Thanks, and apologies for the last minute request!

marschwar:
> There are a few new rules added by #6253 which will be merged momentarily... can this be rebased please, and the updates made to the four new rules as well? Thanks, and apologies for the last minute request!

No woriies. I saw the PR and knew it was coming ;)

3flex (approved):

PR #6107, opened 2023-05-18
Inline Cases enum and inline other external test code into the test classes

TWiStErRob:
More changes related to https://github.com/detekt/detekt/issues/1089 cc @vanniktech @schalkms 

I wanted to work on `LoopWithTooManyJumpStatements` and then noticed the Cases enum and the comment on it. I wanted some relaxing activity to rest anyway, so I moved all the files into `code` blocks like in newer tests. Then I went ahead by looking at other `resourceAsPath`, and there were some without an enum declaration too, so moved them as well.

I deliberately did not make significant changes like splitting up the code blocks into N tests, because that would be much harder to review. Babysteps.

Tried to make no modifications other than copy-paste, few exceptions:
 * rename variable/field names that host the code
 * remove `package cases` declarations
 * remove unnecessary suppressions
 * remove a few empty lines to make the codes more compact
 * real changes are towards he end of the commit list

Tip: for review I recommend using IDEA, commit by commit + use the "right click in any code editor > Compare with clipboard" feature to compare deleted files (copy to clipboard) with added code blocks.


vanniktech:
Thanks for linking me, but I've stopped using Detekt, so please feel free to continue this in any direction you'd want.

schalkms (approved):
Thanks for doing this tedious work! I like!

cortinico (approved):
Thanks for the cleanup @TWiStErRob

PR #6110, opened 2023-05-19
Reduce LoopWithTooManyJumpStatements finding scope

TWiStErRob:
Similar to other PRs where we reduced the scope.

Real life use case where it hides other warnings, and just plain huge yellow blob:
image

Contains a commit from #6107 + splits up the test cases before changing code.


cortinico:
@TWiStErRob can you rebase?

Issue #6109, opened 2023-05-19
`SerialVersionUIDInSerializableClass` highlights complete class when it finds warning


atulgpt:
## Expected Behavior
`SerialVersionUIDInSerializableClass` In case it finds `serialVersionUID` which is not private or const it should only highlight `serialVersionUID`

In case `serialVersionUID` is missing it should only highlight the class name

## Observed Behavior
The whole class is highlighted. It hides other issues in that class
<img width="589" alt="Screenshot 2023-05-19 at 8 25 00 PM" src="https://github.com/detekt/detekt/assets/11026203/e00cda17-ac6f-43fe-8648-4e707610134c">

PR #6114, opened 2023-05-20
SerialVersionUIDInSerializableClass - Update the error location

atulgpt:
Fixes https://github.com/detekt/detekt/issues/6109

The error will be highlighted only at the class location when `serialVersionUID` is missing. 
In case of the wrong type, or modifier of `serialVersionUID` error will be highlighted at property declaration Add TC asserting location and error msg
Add TC for nested classes

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SerialVersionUIDInSerializableClass.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SerialVersionUIDInSerializableClass.kt
@@ -82,15 +82,13 @@ class SerialVersionUIDInSerializableClass(config: Config = Config.empty) : Rule(
     }
 
     private fun reportFinding(finding: SerialVersionUIDFindings) {
-        if (finding.isValid.not()) {
-            report(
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/SerialVersionUIDInSerializableClass.kt
```diff
@@ -106,4 +135,10 @@ class SerialVersionUIDInSerializableClass(config: Config = Config.empty) : Rule(
         return assignmentText != null && assignmentText.last() == 'L' &&
             assignmentText.substring(0, assignmentText.length - 1).toLongOrNull() != null
     }
+
+    private data class SerialVersionUIDFindings(
+        val isValid: Boolean,
```
cortinico, line 140: this is always false right? Why do we need it then?

atulgpt: Yes, correct. Thanks for catching that 😅

cortinico (approved):

PR #6117, opened 2023-05-21
Replace sarif4k usage with Groovy JSON from gradleApi() in detekt-gradle-plugin

### LINKED ISSUE #5864: Refactor DGP off Sarif4k
Is the remaining action that we should refactor DGP off Sarif4k so that
- Respecting KGP compileOnly dependency
- Not exposing transitive dependency when our client/users depends on KGP

_Originally posted by @chao2zhang in https://github.com/detekt/sarif4k/issues/43#issuecomment-1445413535_

TWiStErRob:
Fixes #5864

Note: I didn't find a clear GSON or Jackson API, so used Groovy's API which was readily on the classpath.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/SarifReportMerger.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/SarifReportMerger.kt
@@ -4,7 +4,7 @@ import groovy.json.JsonOutput
 import groovy.json.JsonSlurper
 import java.io.File
 
-private typealias JsonObject = MutableMap<String, Any>
+private typealias JsonObject = MutableMap<String, Any?>
 
 /**
  * A naive implementation to merge SARIF assuming all inputs are written by detekt.
@@ -16,7 +16,7 @@ object SarifReportMerger {
             @Suppress("UNCHECKED_CAST")
             (JsonSlurper().parse(it) as JsonObject)
         }
-        val mergedResults = sarifs.flatMap { it.runs.single().results }
+        val mergedResults = sarifs.flatMap { it.runs.single().results.orEmpty() }
         val mergedSarif = sarifs[0].apply { this.runs.single().results = mergedResults }
         output.writeText(JsonOutput.prettyPrint(JsonOutput.toJson(mergedSarif)))
     }
@@ -26,9 +26,9 @@ private val JsonObject.runs: List<JsonObject>
     @Suppress("UNCHECKED_CAST")
     get() = this["runs"] as List<JsonObject>
 
-private var JsonObject.results: List<JsonObject>
+private var JsonObject.results: List<JsonObject>?
     @Suppress("UNCHECKED_CAST")
-    get() = this["results"] as List<JsonObject>
+    get() = this["results"] as List<JsonObject>?
     set(value) {
         this["results"] = value
     }
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/SarifReportMerger.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/SarifReportMerger.kt
@@ -1,7 +1,10 @@
 package io.gitlab.arturbosch.detekt.report
 
-import io.github.detekt.sarif4k.SarifSerializer
+import groovy.json.JsonOutput
+import groovy.json.JsonSlurper
 import java.io.File
+
+private typealias JsonObject = MutableMap<String, Any>
 
 /**
  * A naive implementation to merge SARIF assuming all inputs are written by detekt.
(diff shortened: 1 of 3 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/SarifReportMerger.kt
```diff
@@ -1,19 +1,37 @@
...
     fun merge(inputs: Collection<File>, output: File) {
         val sarifs = inputs.filter { it.exists() }.map {
-            SarifSerializer.fromJson(it.readText())
+            @Suppress("UNCHECKED_CAST")
+            (JsonSlurper().parse(it) as JsonObject)
         }
-        val mergedResults = sarifs.flatMap { it.runs.single().results.orEmpty() }
-        val mergedSarif = sarifs[0].copy(runs = listOf(sarifs[0].runs.single().copy(results = mergedResults)))
-        output.writeText(SarifSerializer.toJson(mergedSarif))
+        val mergedResults = sarifs.flatMap { it.runs.single().results }
```
TWiStErRob, line 19: Removed nullability, since it's a "naive" impl, can put it back easily.

Note that detekt sarif always contains an empty array, e.g. `detekt-gradle-plugin/build/reports/detekt/detekt.sarif`
image

chao2zhang (approved):

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/SarifReportMerger.kt
chao2zhang: We could use `as?` for `val JsonObject.runs` and `var JsonObject.results` in case the sarif results are not generated by detekt plugin? (e.g. Ktlint)

TWiStErRob: Interesting, how would that end up in the detekt merger?

Btw, they are guaranteed to be lists of objects by schema, aren't they? So I think we can do `as List<JsonObject>?` so if the file is really invalid, it fails.

detekt-gradle-plugin/src/test/resources/output.sarif.json
```diff
@@ -1,156 +1,157 @@
 {
```
TWiStErRob, line 1: Note: only indentation changed in this file.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/SarifReportMerger.kt
```diff
@@ -1,19 +1,34 @@
...
+        val mergedSarif = sarifs[0].apply { this.runs.single().results = mergedResults }
```
TWiStErRob, line 20: Also, what are your thoughts on mutating? I think it's safe because we're the only ones seeing these instances and they'll be throw away as soon as the output is written and the fun returns.

cortinico: Yeah I see no harm here

3flex:
> so used Groovy's API which was readily on the classpath

Gradle updates Groovy in each release (though will only update major releases when there's a new major Gradle release) so there's a small chance of breakage depending on user's Gradle version because we're not in control of these dependencies anymore.

I have no idea how stable these APIs are in Groovy - it might not be a problem in practice.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/SarifReportMerger.kt
TWiStErRob: https://github.com/detekt/detekt/pull/6117/commits/871c2d3ece1a19323eda67ed9b4d75db614ce324

Note: `runs` is also nullable according to sarif jsonschema, but sarif4k wasn't handling it and it was fine. It stranger to not have a run, because that is the main reason a sarif file exists.

chao2zhang: I tested using `https://sarifweb.azurewebsites.net/Validation`. Missing `results` does not invalidate the sarif file

image

TWiStErRob:
@3flex At my previous workplace we've used `JsonOutput.prettyPrint` and `JsonOutput.toJson` since around AGP 3.x times (Gradle 4+), and I don't remember anything strange.

That said, I don't fully understand the reason for dropping Sarif4k, just wanted to see what else is available out of the box, so I'll let @chao2zhang / @cortinico respond to https://github.com/detekt/detekt/pull/6117#issuecomment-1556299227 and whether we actually want to go this way and merge this PR.

cortinico:
> Gradle updates Groovy in each release (though will only update major releases when there's a new major Gradle release) so there's a small chance of breakage depending on user's Gradle version because we're not in control of these dependencies anymore.

I think this adds less harm than us depending on sarif4k on DCG due to transitive dependencies

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/report/SarifReportMerger.kt
TWiStErRob: Tip: IDEA can also do that.

CMD/Ctrl+Click anything in a sarif file:
image
and it'll show the defintion
image
and if you break the file, it tells you:
image
just make sure the schema is recognized correctly
image

cortinico:
Let's merge this and we can still revert back to sarif4k if needed.

PR #6131, opened 2023-05-24
Changed name of rule from FunctionMaxLength to FunctionNameMaxLength

PoisonedYouth:
Renaming rule name to better match its purpose. Changes done according to https://github.com/detekt/detekt/discussions/6035#discussioncomment-5702473

FunctionMaxLength -> FunctionNameMaxLength


schalkms (approved):

BraisGabin:
🤔 We should fix `FunctionMinLength` too. I'm not sure if this should be done in the same PR (related rules) or not (different rules).

BraisGabin (approved):

PoisonedYouth:
That absolutely makes sense. I will do this with a separate PR.

PoisonedYouth:
> 🤔 We should fix `FunctionMinLength` too. I'm not sure if this should be done in the same PR (related rules) or not (different rules).

Done with https://github.com/detekt/detekt/pull/6161

cortinico (approved):

PR #6139, opened 2023-05-25
PropertyUsedBeforeDeclaration: fix false positive in nested/inner class

### LINKED ISSUE #6125: PropertyUsedBeforeDeclaration false positive nested usage of constant
## Expected Behavior
No finding, the inner class will be initialized after the outer class. Also since this is a constant, it'll be inlined by the compiler, so the property is "unused" in the bytecode.

## Observed Behavior
```console
PropertyUsedBeforeDeclaration - ['DAGGER_NAME' is before declaration.] at ...
```

## Steps to Reproduce
```kotlin
@Module
object MyModule {
	interface Dependencies<Builder> {
		@BindsInstance
		fun prop(@Named(DAGGER_NAME) namedProp: String): Builder
	}
	internal const val DAGGER_NAME = "dagger-name"
}
```
for reference: `annotation class Named(val value: String)`.

## Context
Strangely
```kotlin
object Outer {
    object Inner {
        val inner = outer
    }
    val outer = "value"
}
```
is fine, so there must be something about the annotation probably.

## Your Environment

* Version of Gradle used (if applicable): 8.1
* Operating System and version: Windows

cc @t-kameyama as author of the rule https://github.com/detekt/detekt/pull/6062

t-kameyama:
Fixes #6125

### DIFF
--- a/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/PropertyUsedBeforeDeclarationSpec.kt
+++ b/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/PropertyUsedBeforeDeclarationSpec.kt
@@ -86,4 +86,44 @@ class PropertyUsedBeforeDeclarationSpec(private val env: KotlinCoreEnvironment) 
         val findings = subject.compileAndLintWithContext(env, code)
         assertThat(findings).isEmpty()
     }
+
+    @Test
+    fun `used before declaration in nested object`() {
+        val code = """
+            object Outer {
+                object O {
+                    val inner = outer1
+                }
+            
+                class C {
+                    val inner = outer2
+                }
+            
+                annotation class Ann(val value: String)
+                interface I {
+                    fun f(@Ann(outer3) namedProp: String)
+                }
+            
+                val outer1 = "value1"
+                val outer2 = "value2"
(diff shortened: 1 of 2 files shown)


detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/PropertyUsedBeforeDeclarationSpec.kt
```diff
@@ -86,4 +86,44 @@ class PropertyUsedBeforeDeclarationSpec(private val env: KotlinCoreEnvironment)
...
+    @Test
+    fun `used before declaration in inner class`() {
+        val code = """
+            class A {
+                inner class B {
+                    val inner = outer
+                }
+                val outer = "value"
+            }
+        """.trimIndent()
```
TWiStErRob, line 125: Is it worth adding a similar test where the `outer` is a top-level property, and it is used from a normal class? That sounds like a normal use case.
```
class C {
    val inner = outer
}
val outer = "value"
```
in this case they're essentially not even nested, just sibling classes. Only Kotlin's syntax allows their declaration in one file.

TWiStErRob (approved):

BraisGabin (approved):

schalkms (approved):

Issue #6146, opened 2023-05-25
CastNullableToNonNullableType false positive on platform types


TWiStErRob:
## Observed Behavior

![image](https://github.com/detekt/detekt/assets/2906988/84bd961f-3006-4fbc-b18c-0d7a76e90f24)
`as GradleInternal` is reported

## Expected Behavior
No report, because `Project.gradle` is defined as:
![image](https://github.com/detekt/detekt/assets/2906988/562a7cfa-4256-4063-bcec-d46be4560c58)
Notice that there's no nullability annotation, only javadoc, but there's a package-level annotation:
```
@org.gradle.api.NonNullApi
package org.gradle.api;
```
Looks like this might be a type resolution issue, because the IDE infers non-null:
![image](https://github.com/detekt/detekt/assets/2906988/b0593322-d66a-420a-9587-4590668ce56e)

Anyway, the problem exists on all platform types, even when there's no annotation.

## Steps to Reproduce
Cast any function or property result which has platform type. For example

```java
class C {
    static CharSequence cs() {
        return "";
    }
}
```
```kotlin
fun f() {
    val cs = C.cs()
    val str = cs as String // CastNullableToNonNullableType
}
```
> Use separate `null` assertion and type cast like ('(cs ?: error("null assertion message")) as String') instead of 'cs as String'. [CastNullableToNonNullableType]

## Context
When the inferred type is platform type because there's no nullability defined in Java APIs, by definition we're free to interpret that as nullable or non-nullable. So these should be ignored, or at least ignorable by a flag (and I would vote for that flag to be turned on by default to avoid lots of false positives).

## Your Environment

* Version of detekt used: 1.23.0
* Version of Gradle used (if applicable): 8.1.1
* Gradle scan link (add `--scan` option when running the gradle task): N/A
* Operating System and version: all
* Link to your project (if it's a public repository): not needed, full repro present

cc @atulgpt as [author](https://github.com/detekt/detekt/pull/5653)

TWiStErRob:
Another case: `NumberFormat.getInstance() as DecimalFormat` and if I add `!!`:
```
NumberFormat.getInstance()!! as DecimalFormat
```
I get this compiler warning:

> [UNNECESSARY_NOT_NULL_ASSERTION] Unnecessary non-null assertion (!!) on a non-null receiver of type NumberFormat!

so Kotlin knows it's non-null somehow?

---

[...]

PR #6149, opened 2023-05-25
Assume platform type as non nullable

atulgpt:
Fixes https://github.com/detekt/detekt/issues/6146

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -405,6 +405,7 @@ potential-bugs:
       - 'kotlin.String'
   CastNullableToNonNullableType:
     active: false
+    ignorePlatformTypes: true
   CastToNullableType:
     active: false
   Deprecation:
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
@@ -63,7 +63,7 @@ class CastNullableToNonNullableType(config: Config = Config.empty) : Rule(config
             bindingContext,
             compilerResourcesNonNull.languageVersionSettings,
             compilerResourcesNonNull.dataFlowValueFactory,
-            shouldConsiderPlatformTypeAsNullable = true,
+            shouldConsiderPlatformTypeAsNullable = false,
         ).ifFalse { return }
 
         val message =
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
@@ -7,6 +7,8 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
+import io.gitlab.arturbosch.detekt.api.config
+import io.gitlab.arturbosch.detekt.api.internal.Configuration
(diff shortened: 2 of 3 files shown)


detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
```diff
@@ -63,7 +63,7 @@ class CastNullableToNonNullableType(config: Config = Config.empty) : Rule(config
             bindingContext,
             compilerResourcesNonNull.languageVersionSettings,
             compilerResourcesNonNull.dataFlowValueFactory,
-            shouldConsiderPlatformTypeAsNullable = true,
+            shouldConsiderPlatformTypeAsNullable = false,
```
TWiStErRob, line 66: @cortinico Would it make sense to have this configurable?

atulgpt: Just FYI there is one more rule for explicitly defining the type of platform types. And previously we have ignore platform types for few rules

cortinico: Yes I think this should maybe be configurable.

The reason is that the specific instances that @TWiStErRob reported are annotated with `@org.gradle.api.NonNullApi` which Detekt doesn't understand.

As a rule of thumb, we should assume platform types as nullable types.

TWiStErRob: > annotated with @org.gradle.api.NonNullApi which Detekt doesn't understand.

There are others as well from Java code. It was just the example that triggered me to write the issue :)

> Yes I think this should maybe be configurable.
> As a rule of thumb, we should assume platform types as nullable types.

👍🏻

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
```diff
@@ -48,6 +50,9 @@ class CastNullableToNonNullableType(config: Config = Config.empty) : Rule(config
         Debt.FIVE_MINS
     )
 
+    @Configuration("ignore platform type")
```
cortinico, line 53: Please provide a bit more explanation here

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
```diff
@@ -48,6 +50,9 @@ class CastNullableToNonNullableType(config: Config = Config.empty) : Rule(config
         Debt.FIVE_MINS
     )
 
+    @Configuration("Whether platform types should be considered as non-nullable and ignored by this rule")
+    private val ignorePlatformType: Boolean by config(true)
```
TWiStErRob, line 54: Should this be plural?
```suggestion
    private val ignorePlatformTypes: Boolean by config(true)
```

TWiStErRob (approved):

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastNullableToNonNullableType.kt
atulgpt: Updated

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -405,6 +405,7 @@ potential-bugs:
       - 'kotlin.String'
   CastNullableToNonNullableType:
     active: false
+    ignorePlatformType: true
```
TWiStErRob, line 408: rename didn't propagate

TWiStErRob (approved):

Issue #6142, opened 2023-05-25
False positive in `UnusedPrivateProperty` for private data class fields


lwasyl:
## Steps to Reproduce

```kotlin
data class Foo(
    private val foo: Int,
    private val bar: String,
)
```

will report `private val`s as unused. That's not the case — both are used in `equals` and `hashcode` implementations, and removing them changes semantics

## Your Environment

* Version of detekt used: 1.23.0
* Version of Gradle used (if applicable):
* Gradle scan link (add `--scan` option when running the gradle task):
* Operating System and version:
* Link to your project (if it's a public repository):

cortinico:
Yup that's a valid report. We need `UnusedPrivateProperty` to exclude `data class`es

PR #6151, opened 2023-05-25
UnusedPrivateProperty: Fix false postive by ignoring data classes

atulgpt:
Fixes: https://github.com/detekt/detekt/issues/6142

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateProperty.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateProperty.kt
@@ -14,6 +14,7 @@ import io.gitlab.arturbosch.detekt.api.internal.Configuration
 import io.gitlab.arturbosch.detekt.rules.isActual
 import io.gitlab.arturbosch.detekt.rules.isExpect
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateProperty.kt
```diff
@@ -108,7 +109,8 @@ private class UnusedPrivatePropertyVisitor(private val allowedNames: Regex) : De
         constructor.valueParameters
             .filter {
                 (it.isPrivate() || (!it.hasValOrVar() && !constructor.isActual())) &&
-                    it.containingClassOrObject?.isExpect() == false
+                    it.containingClassOrObject?.isExpect() == false &&
+                    (constructor.parent as? KtClass)?.isData() != true
```
t-kameyama, line 113: The same goes for `value` classes.
inline
```
@JvmInline value class Foo(private val value: String)

inline class Bar(private val value: String)
```

cortinico: Fair point 👍

atulgpt: Updated

schalkms (approved):

BraisGabin (approved):

Issue #6140, opened 2023-05-25
StringLiteralDuplication: have a option to specifiy how many characters to ingore


atulgpt:
## Expected Behavior
There should be one config in the rule to set the number of characters to ignore. Currently, there is `excludeStringsWithLessThan5Characters` that is a limiting config as it assumes `5` characters to be the threshold.

Also, we can have the same behaviour from the regex `ignoreStringsRegex` config as well but having the same with a different config will be easier to use
## Current Behavior
There is no similar config

## Context
In the code base, I found some symbols like `"="`, `"&"` and `"%"` to be used for making queries in the backend. To our project it doesn't make sense to refactor all those usages to `EQUAL_OPERAROR` or in similar names

cortinico:
Are you suggesting we deprecate `excludeStringsWithLessThan5Characters` and add a `charactersThreshold` property?

atulgpt:
Hi, @cortinico yes as `excludeStringsWithLessThan5Characters` will be meaning less after `charactersThreshold`.

PR #6156, opened 2023-05-26
Add `allowedWithLengthLessThan` and remove `excludeStrin…

atulgpt:
Fixes: https://github.com/detekt/detekt/issues/6140

### DIFF
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/StringLiteralDuplication.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/StringLiteralDuplication.kt
@@ -59,12 +59,8 @@ class StringLiteralDuplication(config: Config = Config.empty) : Rule(config) {
     @Configuration("if values in Annotations should be ignored")
     private val ignoreAnnotation: Boolean by config(true)
 
-    @Deprecated("Use `excludeStringsWithLessThanThreshold` instead")
-    @Configuration("if short strings should be excluded")
-    private val excludeStringsWithLessThan5Characters: Boolean by config(true)
-
     @Configuration("if strings is less than the specified length should be excluded")
-    private val excludeStringsWithLessThanThreshold: Int by config(5)
+    private val allowStringWithLength: Int by config(5)
 
     @Configuration("RegEx of Strings that should be ignored")
     private val ignoreStringsRegex: Regex by config("$^", String::toRegex)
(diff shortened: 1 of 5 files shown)


marschwar:
Just a few thoughts on this:

* I am not sure if we currently need to deprecate anything as the next release will be 2.0
* I do not like that we add another property with `threshold` in its name here. We realized that the name threshold is misleading or at least not clear and [decided to come up with a more explicit wording](https://github.com/detekt/detekt/discussions/6035#discussioncomment-5702498).

Other than that I think it is a nice improvement for this rule.

cortinico:
> I am not sure if we currently need to deprecate anything as the next release will be 2.0

Yup correct. If we want to be 100% fair to our users, @atulgpt you should send a PR against `release/1.x` which contains the deprecation and just remove the property from `main`

atulgpt:
> > I am not sure if we currently need to deprecate anything as the next release will be 2.0
> 
> Yup correct. If we want to be 100% fair to our users, @atulgpt you should send a PR against `release/1.x` which contains the deprecation and just remove the property from `main`

Hi, @cortinico `release/1.x` should also contain `excludeStringsWithLessThanThreshold`(obviously with the new name). O/W it will reach end users very late.

Also, what do you think bout the new name `allowStringsCharacterCount`(I know it is not very good 😓. Suggestions are welcome). Or does it need to start with `allow` can it be `minAllowedCharcterCount`?

atulgpt:
> > > I am not sure if we currently need to deprecate anything as the next release will be 2.0
> > 
> > 
> > Yup correct. If we want to be 100% fair to our users, @atulgpt you should send a PR against `release/1.x` which contains the deprecation and just remove the property from `main`
> 
> Hi, @cortinico `release/1.x` should also contain `excludeStringsWithLessThanThreshold`(obviously with the new name). O/W it will reach end users very late.
> 
> Also, what do you think bout the new name `allowStringsCharacterCount`(I know it is not very good 😓. Suggestions are welcome). Or does it need to start with `allow` can it be `maxAllowedCharcterCount`?

Hi, @cortinico, @marschwar updated the code to remove the deprecated method and rename the `excludeStringsWithLessThanThreshold` to `allowStringWithLength`

cortinico:
> I will add the PR in `release/1.x` with deprecation and new property

I don't think this should go in 1.x as it's like a new feature and is not a bugfix

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/StringLiteralDuplication.kt
```diff
@@ -59,8 +59,8 @@ class StringLiteralDuplication(config: Config = Config.empty) : Rule(config) {
     @Configuration("if values in Annotations should be ignored")
     private val ignoreAnnotation: Boolean by config(true)
 
-    @Configuration("if short strings should be excluded")
-    private val excludeStringsWithLessThan5Characters: Boolean by config(true)
+    @Configuration("The maximum allowed string length")
+    private val allowedStringWithLength: Int by config(5)
```
BraisGabin, line 63: I would add something like "less than" to the name of this property. Otherwise it is difficult to know what it does. But I must admit that I can think of any good name. Ideas:
- `allowedStringWithLenghtLessThan` I don't like because we want to include the number. 
- `allowedStringWithLenghtLessOrEqualThan` I don't like because it is too long

atulgpt: Hi, @BraisGabin what about `allowedStringWithLengthLessThan` and the description as `The maximum string length below which duplications are allowed`
We can also use `allowedWithLengthLessThan`(omitting `String`) as that is obvious

BraisGabin (approved):

marschwar (approved):

PR #6158, opened 2023-05-27
Add EnumEntryOrder rule (#6115)

### LINKED ISSUE #6115: New Rule: Enum entries for enums matching a configuration should be in alphabetical order
Searched in issues and could not find this suggested previously.

This rule defines a configuration:

```yaml
EnumEntriesAlphabeticalOrder:
  active: false
  includeAnnotated: ["com.example.annotations.Alphabetical"]
  includeInterfaces: ["com.example.interfaces.Alphabetical"]
  include: ["com.example.MyEnum"]
```
This configuration allows a project to mark enums that they wish to keep in alphabetical order. They can do this in the most generic way by specifying their own annotation (e.g., a `@Alphabetical` annotation within their own project). Detekt will scan the project and look for enums that have that annotation explicitly applied or applied through implementing an interface decorated with the annotation. It will report when these enums have entries that are not in alphabetical order.

Further options for configuration are listing FQN of interfaces that mark enums to be scanned. I am not sure about using regex or wildcard lists to configure this rule

#### With annotation in configuration

```yaml
EnumEntriesAlphabeticalOrder:
  active: true
  includeAnnotated: ["com.example.annotations.Alphabetical"]
```

```kotlin
import com.example.annotations.Alphabetical

@Alphabetical
enum class AnalyticsVariable { // non-compliant (explicitly annotated with configured annotation)
   USER_ID,
   ITEMS_IN_CART,   
}
```

```kotlin
import com.example.annotations.Alphabetical

@Alphabetical
enum class AnalyticsVariable { // compliant 
   ITEMS_IN_CART,
   USER_ID, 
}
```

```kotlin
import com.example.annotations.Alphabetical

@Alphabetical
interface Analytics

// non-compliant (implements interface decorated with @Alphabetical and entries out of order)
enum class AnalyticsVariable : Analytics {
   USER_ID,
   ITEMS_IN_CART,
}

// compliant
enum class AnalyticsVariable: Analytics {
   ITEMS_IN_CART,
   USER_ID
}
```

#### With interfaces
```yaml
EntriesAlphabeticalOrder:
  active: true
  includeInterfaces: ["com.example.interfaces.Alphabetical"]
```

```kotlin
interface Alphabetical
```

```kotlin
// non-compliant (implements interface in config)
enum class AnalyticsVariable : Alphabetical {
   USER_ID,
   ITEMS_IN_CART,
}

// compliant 
enum class AnalyticsVariable: Alphabetical {
   ITEMS_IN_CART,
   USER_ID, 
}
```

## Context
In large Android codebases, the file containing an enum can become chaotic. It's not often not clear for devs where to add their new constant - should it be at the end of the file (to maintain chronological order) or somewhere else? It's not right to say every enum should be in alphabetical order because there are some where the ordinal matters (e.g., an enum where we are using the enum `ordinal` or when we are parsing an Int into an enum). Keeping in alphabetical order allows quick scanning of a file to understand contents but it generates a lot of friction in PRs and it also generates a noisy diff when someone decides to reorder a chaotic enum file. It can also generate potential bugs when people sort in alphabetical order without considering whether this is appropriate

I have worked on codebases where keeping enums entries in alphabetical order was required of devs and we solved it by writing our own Android Lint rule similar to how this proposed rule works. I've also vaguely remember hearing about it in projects at Uber (a difference in style project between the Uber Eats and the Uber Taxi projects?).

drawers:
Closes #6115 where it is proposed to introduce a new rule for checking if enum entries are declared in alphabetical order.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -542,6 +542,10 @@ style:
     negativeFunctionNameParts:
       - 'not'
       - 'non'
+  EnumEntryOrder:
+    active: false
+    includeAnnotations:
+      - 'io.gitlab.arturbosch.detekt.annotations.Alphabetical'
   EqualsNullCall:
     active: true
   EqualsOnSignatureLine:
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/EnumEntryOrder.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/EnumEntryOrder.kt
@@ -8,7 +8,6 @@ import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.config
-import io.gitlab.arturbosch.detekt.api.internal.ActiveByDefault
 import io.gitlab.arturbosch.detekt.api.internal.Configuration
 import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
 import io.gitlab.arturbosch.detekt.rules.identifierName
@@ -60,7 +59,6 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.getSuperInterfaces
  *  </noncompliant>
  */
 @RequiresTypeResolution
-@ActiveByDefault(since = "1.24.0")
 class EnumEntryOrder(config: Config = Config.empty) : Rule(config) {
 
     override val issue: Issue = Issue(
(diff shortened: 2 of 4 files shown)


detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -542,6 +542,10 @@ style:
     negativeFunctionNameParts:
       - 'not'
       - 'non'
+  EnumEntryOrder:
+    active: false
+    includeAnnotations:
+      - 'io.gitlab.arturbosch.detekt.annotations.Alphabetical'
```
drawers, line 548: Not sure about this - we are defaulting to an annotation that doesn't actually exist. Might be better to wait until the `detekt-annotations` artifact proposed here https://github.com/detekt/detekt/issues/6115#issuecomment-1556352994 is ready before merging. I can have a go at building this artifact but it will take me some time

drawers:
Putting this on hold while there is some discussion in the issue tracker for #6115

PR #6159, opened 2023-05-27
Respect `RunAfterRule` constraint from ktlint

### LINKED ISSUE #5259: detekt doesn't respect ktlint `RunAfterRule` constraints

chao2zhang:
Resolves #5259 

Consulted with ktlint in https://github.com/pinterest/ktlint/pull/2062, we are building a consistent sorting logic in detekt to match ktlint.


detekt-formatting/src/test/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingProviderSpec.kt
```diff
@@ -1,19 +1,47 @@
...
-    fun `preferred ktlint rule ordering is applied`() {
+    fun `list is unique`() {
         val subject: RuleSet = FormattingProvider().instance(Config.empty)
+        subject.rules.groupBy { it.ruleId }.forEach {
+            assertThat(it.value).hasSize(1)
+        }
+    }
 
+    @Test
+    fun `run as late as possible is observed`() {
```
chao2zhang, line 20: Shall we disable the test until https://github.com/pinterest/ktlint/pull/2062 is merged and updated in detekt?

marschwar: > The rule sorter first runs all rules without any VisitorModifier.
The next group of rules are the rules marked with RunAsLateAsPossible.
The last group are the rules marked with RunAfter.

Should we update this PR to emulate the same mechanics as ktlint.

BraisGabin (approved):

detekt-formatting/src/test/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingProviderSpec.kt
chao2zhang: We are following the same logic in this PR, let me add some comments for posterity

Issue #6162, opened 2023-05-29
false positive KDocReferencesNonPublicProperty when method and property with same name


robstoll:
## Expected Behavior
KDoc links correctly to the method, hence there should not be an error reported

## Observed Behavior
reports 
```
 KDocReferencesNonPublicProperty - [The property peek is non-public and should not be referenced from KDoc comments.] at kbox/src/commonMain/kotlin/ch/tutteli/kbox/DefaultPeekingIterator.kt:8:17
```

## Steps to Reproduce
1. git clone https://github.com/robstoll/kbox
2. open  DefaultPeekingIterator or run `./gradlew :detektMetadataMain`

## Your Environment

* Version of detekt used: 1.22.0
* Version of Gradle used (if applicable): 7.6.1
* Gradle scan link (add `--scan` option when running the gradle task):
* Operating System and version: ubuntu 22.04
* Link to your project (if it's a public repository): https://github.com/robstoll/kbox

PR #6165, opened 2023-05-29
Allow documenting public fun name when same private variable is present

atulgpt:
Extend the previous logic by traversing to `visitNamedDeclaration`

Fixes: https://github.com/detekt/detekt/issues/6162

### DIFF
--- a/detekt-rules-documentation/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicPropertySpec.kt
+++ b/detekt-rules-documentation/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicPropertySpec.kt
@@ -196,4 +196,29 @@ class KDocReferencesNonPublicPropertySpec {
         """.trimIndent()
         assertThat(subject.compileAndLint(code)).isEmpty()
     }
+
+    @Test
+    fun `does not report public function when same named private var is present - #6162`() {
+        val code = """
+            /**
+             * [peek] - public fun
+             * [Inner.innerPeek] - public fun
+             * [Object.objectPeek] - public fun
+             */
+            class Test {
+                private var peek: Int = 0
+                private var innerPeek: Int = 0
+                private var objectPeek: Int = 0
+                private var companionPeek: Int = 0
+                fun peek() = System.currentTimeMillis()
+                inner class Inner {
+                    fun innerPeek() = System.currentTimeMillis()
+                }
+                object Object {
+                    fun objectPeek() = System.currentTimeMillis()
+                }
+            }
+        """.trimIndent()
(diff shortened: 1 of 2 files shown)


detekt-rules-documentation/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicPropertySpec.kt
```diff
@@ -196,4 +196,29 @@ class KDocReferencesNonPublicPropertySpec {
...
+                private var objectPeek: Int = 0
+                private var companionPeek: Int = 0
+                fun peek() = System.currentTimeMillis()
+                inner class Inner {
+                    fun innerPeek() = System.currentTimeMillis()
+                }
+                object Object {
+                    fun objectPeek() = System.currentTimeMillis()
+                }
+            }
```
schalkms, line 220: I have the feeling this test snippet can be simplified to be easier understood for future readers.
What do you think?

schalkms (approved):
See my comment.
Code adaption looks good.

detekt-rules-documentation/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/KDocReferencesNonPublicPropertySpec.kt
atulgpt: Hi, @schalkms I tried to have multiple similar TCs in a single code as that was possible here. In what respect do you want this to be changed? Will adding comments in the code make it clearer? Or do we want me to split the TC in multiple TCs under one nested class?

schalkms: My intention was to keep the code shorter if possible.

atulgpt: I can replace `System.currentTimeMillis()` with `0` will that make it shorter 😅. O/W code is already short and concise (it represents normal function, `inner class` function, and `object` function)

schalkms: Yep, that helps and removes unnecessary noise from test code. I had to read the test code twice to get the intention of it.

atulgpt: Updated

BraisGabin (approved):

BraisGabin:
Thank you!!

Issue #6168, opened 2023-05-31
`UseDataClass` false positive: "class defines no functionality"


yogurtearl:
Consider this code:

```kotlin
class Bar(
    val param1: String,
    foo: Int,
) {
    val baz = 2 * foo
}
```

## Expected Behavior
`UseDataClass` does not say anything. 

## Observed Behavior
detekt gives this warning with `UseDataClass` is enabled.
```
The class Bar defines no functionality and only holds data. Consider converting it to a data class. [UseDataClass]
```

"defines no functionality" is not true (see `val baz` )

if I follow the advice and add `data` before `class` I will get a compiler error because `foo: Int` is not a `val`: 
```
Data class primary constructor must only have property (val / var) parameters
```

## Context

Following detekt's advice leads to a compile error.

## Your Environment

* Version of detekt used: 1.23.0
* Version of Gradle used (if applicable): 8.1.1

PR #6173, opened 2023-06-01
Don't report `UseDataClass` if class contains non-property parameters

VirtualParticle:
This fixes https://github.com/detekt/detekt/issues/6168. Currently, `UseDataClass` will report an error even if the class has a non-property parameter as seen in the example:
```kotlin
class Bar(
    val param1: String,
    foo: Int,
) {
    val baz = 2 * foo
}
```
Making this a `data class` would not compile, so the rule should not report it. This resolves that issue by checking to make sure that the class does not contain any non-property parameters before reporting it.


schalkms (approved):
This looks good to me.
Thanks for picking this up!

BraisGabin (approved):

PR #6180, opened 2023-06-04
Remove deprecated elements from Gradle plugin

3flex:
Some clean up for 2.0


BraisGabin (approved):

schalkms (approved):

detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektJvmSpec.kt
```diff
@@ -1,83 +0,0 @@
...
-import io.gitlab.arturbosch.detekt.testkit.DslGradleRunner
-import io.gitlab.arturbosch.detekt.testkit.ProjectLayout
-import org.assertj.core.api.Assertions.assertThat
-import org.junit.jupiter.api.Nested
-import org.junit.jupiter.api.Test
-
-class DetektJvmSpec {
-
-    @Nested
-    inner class `report location set on extension & task` {
```
chao2zhang, line 12: Is this spec removed because we have `DetektJvmSpec` under `test` source set?

3flex: It's because these tests are checking for correct behaviour when reports are configured on both the extension and task, but the reports can't be configured on the extension anymore, so the test is not required. Reports can only be configured on the task.

cortinico (approved):

Issue #3679, opened 2021-04-12
Issues raised even when "on" threshold


jotabe87:
Hi everyone! We incorporated detekt in our project a while ago and, while we like it very much, we've been noticing that it seems to report issues even when the checks are "on" the threshold and not after crossing it. 

### Expected Behavior
For this example, imagine any function, with three nested blocks.
```
fun myFunction(): Boolean {
  someNullableVariable?.let {
    if (something) {
      // do something
      if (something else) {
        return true
      }
  }
  return false
}
```

We expect detekt not to report anything, since it's configure like so:
```
  NestedBlockDepth:
    active: true
    threshold: 3
```

### Observed Behavior
The function is reported to have infringed the NestedBlockDepth rule:

`NestedBlockDepth - 3/3 - [applyInternal] at <path>/<file>.kt:16:18`

My team and I believe that this should not be reported, since the number of nested blocks is 3 from the allowed 3 nested blocks. The issue should only be raised, when the set threshold is crossed, for instance at an event where `4/3` could be observed.

## Steps to Reproduce
- Write a similar function 
- Have detekt in your project with the same settings for NestedBlockDepth

## Context
This does not only happen with NestedBlockDepth, but every single rule is treated as such in detekt. While we want to use detekt to ensure code quality, we do not want to have "add one to the real threshold" or suppress the rule in order to get away with it. Also after a while the question could become "Was this a 3 or a 4? The config says 4, so why is 4 not allowed?", mainly if new colleagues come to the project, or the ones that instituted detekt and the rules move to another project.  

## Your Environment
* Version of detekt used: 1.15.0
* Version of Gradle used (if applicable): 6.8
* Operating System and version: macOS Big Sur 11.2.3

BraisGabin:
I do agree that this is missleading but it's impossible to fix right now. We need a major update to make such a change.

jotabe87:
Thanks for the update! I'm looking forward to see this being implemented in the future!

cortinico:
Hey @jotabe87, thanks for reporting. 

> We need a major update to make such a change.

Do we really @BraisGabin? Isn't this considered the same as fixing a false-positive/false-negative bug in a rule? I believe this is fixable provided we add a note in the release notes or so.

schalkms:
This is accepted system behavior. Hence, there's nothing to fix here. This behavior was agreed in the early stages of detekt. Please see #313 for more details.
As of now, this is the first contradicting feedback we got.
[...]

PR #6181, opened 2023-06-04
Replaced "threshold" property of LongMethod rule by "allowedLines" 

PoisonedYouth:
The LongMethod rule reported an issue when the lines of a method is greater or equal the value defined for the  "threshold" property. It is intended that the value defined in the rule is the maximum allowed lines. So the property is renamed to "allowedLines" and the check for reporting an issue is changed to only greater.

The origin issue is #3679


schalkms (approved):

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/config/CompositeConfigSpec.kt
```diff
@@ -25,7 +25,7 @@ class CompositeConfigSpec {
     @Test
     fun `should have code smell sub config with LongMethod threshold 20 from _default_ config`() {
```
marschwar, line 26:
```suggestion
    fun `should have code smell sub config with LongMethod allowedLines 20 from _default_ config`() {
```

marschwar (approved):

PoisonedYouth:
Are these changes to the LongMethod rule the intended ones for all the other rules using "threshold" using a suitable "allowX" property instead?

schalkms:
> Are these changes to the LongMethod rule the intended ones for all the other rules using "threshold" using a suitable "allowX" property instead?

@PoisonedYouth I'd like to help, but I don't quite understand what you mean and refer to with your question. Does it relate to a decision made in the original issue?

PoisonedYouth:
According to [https://github.com/detekt/detekt/discussions/6035#discussioncomment-5702498](Tasks for the 2.0 version) 
all rules using property "threshold" should be updated by replacing the property. 

Before I update step by step all the rules I would ensure that my changes match the correct intention.

marschwar:
You read the comment correctly. That is exactly the intention.
I think for the release notes and for ease of review it would be beneficial to have a one or per rule. Thank you for your contribution.

PoisonedYouth:
Thanks for the clarification. I planned to open a PR per rule to make the review easier for you.

schalkms:
@PoisonedYouth thanks for asking and for the nice contribution. I like!
The approach with separate PRs looks good to me.

cortinico (approved):

PR #6184, opened 2023-06-06
Replaced "threshold" property of CognitiveComplexMethod rule by "allowedComplexity"

PoisonedYouth:
The CognitiveComplexMethod rule reported an issue when the Cognitive Complexity of a method is greater or equal the value defined for the "threshold" property. It is intended that the value defined in the rule is the maximum allowed Cognitive Complexity. So the property is renamed to "allowedComplexity" and the check for reporting an issue is changed to only check for greater, no longer for equal.

### DIFF
--- a/detekt-generator/src/test/resources/ruleset1/src/main/kotlin/CognitiveComplexMethod.kt
+++ b/detekt-generator/src/test/resources/ruleset1/src/main/kotlin/CognitiveComplexMethod.kt
@@ -21,10 +21,10 @@ class CognitiveComplexMethod(config: Config = Config.empty) : Rule(config) {
         Debt.TWENTY_MINS
     )
 
-    @Configuration("Cognitive Complexity number for a method.")
-    private val threshold: Int by config(defaultValue = 15)
+    @Configuration("Maximum Cognitive Complexity allowed for a method.")
+    private val allowedComplexity: Int by config(defaultValue = 15)
 
     override fun visitNamedFunction(function: KtNamedFunction) {
-        if (false) println(threshold)
+        println(allowedComplexity)
     }
 }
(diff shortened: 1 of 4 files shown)


detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CognitiveComplexMethodSpec.kt
```diff
@@ -35,7 +35,21 @@ class CognitiveComplexMethodSpec {
     }
 
     @Test
-    fun `should not report simple function`() {
+    fun `should not report function that has exact the allowed complexity`() {
```
cortinico, line 38:
```suggestion
    fun `should not report function that has exactly the allowed complexity`() {
```

cortinico (approved):

schalkms (approved):

detekt-generator/src/test/resources/ruleset1/src/main/kotlin/CognitiveComplexMethod.kt
```diff
@@ -21,10 +21,10 @@ class CognitiveComplexMethod(config: Config = Config.empty) : Rule(config) {
...
     )
 
-    @Configuration("Cognitive Complexity number for a method.")
-    private val threshold: Int by config(defaultValue = 15)
+    @Configuration("Maximum Cognitive Complexity allowed for a method.")
+    private val allowedComplexity: Int by config(defaultValue = 15)
 
     override fun visitNamedFunction(function: KtNamedFunction) {
-        if (false) println(threshold)
+        println(allowedComplexity)
```
BraisGabin, line 28: This is debug code, right? It should be gone to merge this.

PoisonedYouth: This is part of a test rule. I'm not sure what the purpose of this is.

BraisGabin: Oh! I see! I think that this line was there to avoid unused warning. I would vote to remove this and add a `@Suppress`

BraisGabin (approved):
Oh!

PR #6187, opened 2023-06-06
Replaced "threshold" property of ComplexCondition rule by "allowedComplexity"

PoisonedYouth:
The ComplexCondition rule reported an issue when the complexity is greater or equal the value defined for the "threshold" property. It is intended that the value defined in the rule is the maximum allowed complexity. So the property is renamed to "allowedComplexity" and the check for reporting an issue is changed to check only for greater, no longer for equal.

### DIFF
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
@@ -52,7 +52,7 @@ class ComplexCondition(
     )
 
     @Configuration("Maximum allowed number of conditions.")
-    private val allowedConditions: Int by config(defaultValue = 4)
+    private val allowedConditions: Int by config(defaultValue = 3)
 
     override fun visitIfExpression(expression: KtIfExpression) {
         val condition = expression.condition
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
@@ -51,8 +51,8 @@ class ComplexCondition(
         Debt.TWENTY_MINS
     )
 
-    @Configuration("the number of conditions which will trigger the rule")
-    private val threshold: Int by config(defaultValue = 4)
+    @Configuration("Maximum allowed number of conditions.")
+    private val allowedConditions: Int by config(defaultValue = 4)
 
     override fun visitIfExpression(expression: KtIfExpression) {
         val condition = expression.condition
@@ -81,14 +81,14 @@ class ComplexCondition(
             }
             val conditionString = longestBinExpr.text
             val count = frequency(conditionString, "&&") + frequency(conditionString, "||") + 1
-            if (count >= threshold) {
+            if (count > allowedConditions) {
                 report(
                     ThresholdedCodeSmell(
                         issue,
                         Entity.from(condition),
-                        Metric("SIZE", count, threshold),
+                        Metric("SIZE", count, allowedConditions),
                         "This condition is too complex ($count). " +
-                            "Defined complexity threshold for conditions is set to '$threshold'"
+                            "The defined maximum number of allowed conditions is set to '$allowedConditions'"
                     )
                 )
             }
--- a/detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexConditionSpec.kt
+++ b/detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexConditionSpec.kt
@@ -1,22 +1,60 @@
 package io.gitlab.arturbosch.detekt.rules.complexity
 
+import io.gitlab.arturbosch.detekt.test.TestConfig
+import io.gitlab.arturbosch.detekt.test.assertThat
 import io.gitlab.arturbosch.detekt.test.compileAndLint
-import org.assertj.core.api.Assertions.assertThat
+import io.gitlab.arturbosch.detekt.test.isThresholded
 import org.junit.jupiter.api.Test
 
 class ComplexConditionSpec {
 
-    val code = """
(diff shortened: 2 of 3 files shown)


detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexConditionSpec.kt
```diff
@@ -1,22 +1,64 @@
...
-            do { } while (5 > 4 && 4 < 6 || (3 < 5 || 2 < 5))
-        }
-    """.trimIndent()
+    private val testConfig = TestConfig("allowedComplexity" to 4)
 
     @Test
-    fun `reports some complex conditions`() {
-        assertThat(ComplexCondition().compileAndLint(code)).hasSize(3)
+    fun `should report complex conditions exceeding allowed complexity`() {
+        // given
```
schalkms, line 15: I think the tests are so concise. Hence, we don't need the comments to understand the test..
```kotlin
// given // when // then
```

schalkms (commented):

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
```diff
@@ -51,8 +51,8 @@ class ComplexCondition(
         Debt.TWENTY_MINS
     )
 
-    @Configuration("the number of conditions which will trigger the rule")
-    private val threshold: Int by config(defaultValue = 4)
+    @Configuration("Maximum allowed complexity for conditions.")
+    private val allowedComplexity: Int by config(defaultValue = 4)
```
marschwar, line 55: How about `allowedConditions`? The rule does not check some magical complexity metric but rather counts the number of conditions in an `if` statement.

marschwar, line 55: Leaving the default at `4` will effectively change the behavior. I suggest changing this to `3`.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
```diff
@@ -81,14 +81,14 @@ class ComplexCondition(
...
+            if (count > allowedComplexity) {
                 report(
                     ThresholdedCodeSmell(
                         issue,
                         Entity.from(condition),
-                        Metric("SIZE", count, threshold),
+                        Metric("SIZE", count, allowedComplexity),
                         "This condition is too complex ($count). " +
-                            "Defined complexity threshold for conditions is set to '$threshold'"
+                            "The defined maximum complexity for conditions is set to '$allowedComplexity'"
```
marschwar, line 91:
```suggestion
                            "The defined maximum number of conditions allowed  is set to '$allowedComplexity'"
```

detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexConditionSpec.kt
```diff
@@ -1,22 +1,57 @@
...
+
+        assertThat(actual).isEmpty()
+    }
+
+    @Test
+    fun `should not report conditions that are below the allowed complexity`() {
+        val code = """
+             fun simpleCondition(): Boolean {
+                 return true
+             }
```
marschwar, line 50: The code should contain at least one `if` or `while` expression. Here the rule does not even check anything.

marschwar (changes requested):

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
PoisonedYouth: That's an interesting topic. Until now I didn't change the defaults for the rules I updated. If it is intended to reduce the defaults for all rules I can do this with a separate PR (for the already merged rule updates) and update the existing open PRs.

PoisonedYouth: @cortinico what is the decision about this topic? Leaving the default value the same (which changes the behavior) or reducing the default value by 1 (to keep the behavior the same).

marschwar: > That's an interesting topic. Until now I didn't change the defaults for the rules I updated. 

For params like allowedLines I would not change it from 600 to 599 and the same goes for arbitrary complexity metrics. But for properties like this I would prefer to keep the behavior the same.

PoisonedYouth: I can understand your point but isn't it a general question if the behavior should be the same  (existing code does not report new issues after update) or to have a breaking change?

marschwar: To me this is more about providing sensitive defaults. Ideally most detekt users do not have to customize the rules. 

If you think 4 conditions in a statement should be allowed because the cognitive complexity is still low enough, that's a discussion we can have. Maybe there was a discussion at the time about how to set the default. I have not checked.

marschwar:
Can you please not amend and force push. It makes reviewing much harder. The branch will be squash merged later.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/ComplexCondition.kt
marschwar: Ok, I did check :)

See https://github.com/detekt/detekt/commit/0d45abd756a6daf6cf78c17ae14d299a9081b475

PoisonedYouth: Thanks I will update the default.

marschwar (approved):

BraisGabin (approved):

PR #6188, opened 2023-06-07
Replaced "threshold" property of LargeClass rule by "allowedLines"

PoisonedYouth:
The LargeClass rule reported an issue when the amount of lines is greater or equal the value defined for the "threshold" property. It is intended that the value defined in the rule is the maximum allowed lines. So the property is renamed to "allowedLines" and the check for reporting an issue is changed to check only for greater, no longer for equal.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -125,7 +125,7 @@ complexity:
     ignoredLabels: []
   LargeClass:
     active: true
-    threshold: 600
+    allowedLines: 600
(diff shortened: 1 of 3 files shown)


marschwar (approved):

PoisonedYouth:
> Thanks! I just think we should remove the test code comments as seen in my comment.
> 
> I think the tests are so concise. Hence, we don't need the comments to understand the test.
> 
> ```kotlin
> // given // when // then
> ```

No problem. I can remove this. I'm just used to do this in all my projects.

BraisGabin (approved):

PR #6189, opened 2023-06-07
Replaced "threshold" property of CyclomaticComplexMethod rule by "allowedComplexity"

PoisonedYouth:
The CyclomaticComplexMethod rule reported an issue when the cyclomatic complexity is greater or equal the value defined for the "threshold" property. It is intended that the value defined in the rule is the maximum allowed cyclomatic complexity. So the property is renamed to "allowedComplexity" and the check for reporting an issue is changed to check only for greater, no longer for equal.

### DIFF
--- a/detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CyclomaticComplexMethodSpec.kt
+++ b/detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CyclomaticComplexMethodSpec.kt
@@ -211,7 +211,7 @@ class CyclomaticComplexMethodSpec {
                     if (i == 1) {
                     }
                 }
-            """.trimMargin()
+            """.trimIndent()
 
             val findings = subject.lint(code)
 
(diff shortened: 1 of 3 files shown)


schalkms (commented):

I think the tests are so concise. Hence, we don't need the comments to understand the test.
```kotlin
// given // when // then
```

marschwar (approved):

BraisGabin (approved):

detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CyclomaticComplexMethodSpec.kt
```diff
@@ -192,6 +195,29 @@ class CyclomaticComplexMethodSpec {
...
+                    when (i) {
+                        1 -> print("one")
+                        2 -> print("two")
+                        3 -> print("three")
+                        else -> print(i)
+                    }
+                    if (i == 1) {
+                    }
+                }
+            """.trimMargin()
```
BraisGabin, line 214:
```suggestion
            """.trimIndent()
```

BraisGabin (approved):

BraisGabin:
Thanks!!!

PR #6193, opened 2023-06-07
Remove deprecated rules and tests

LeoColman:
Reference: https://github.com/detekt/detekt/discussions/6035#discussioncomment-5702474 

Suggestion to review this pull request commit by commit.

PR #6195, opened 2023-06-08
Replaced "functionThreshold/constructorThreshold" properties of LongParameterList rule.

PoisonedYouth:
Replaced "functionThreshold" and "constructorThreshold" properties of LongParameterList rule by "allowedFunctionParameters" and "allowedConstructorParameters" (#3679)

The rule reported an issue when the amount of parameters of a method or constructor is greater or equal the value defined for the  "XThreshold" properties.
It is intended that the value defined in the rule is the maximum allowed parameter amount. So the properties are renamed to
"allowedXParameters" and the check for reporting an issue is changed to only check for greater, no longer for equal.


schalkms (commented):

I think the tests are so concise. Hence, we don't need the comments to understand the test.
```kotlin
// given // when // then
```

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
```diff
@@ -42,17 +42,18 @@ class LongParameterList(config: Config = Config.empty) : Rule(config) {
 
     @Deprecated("Use `functionThreshold` and `constructorThreshold` instead")
```
marschwar, line 43: Did you keep thin intentionally in order to remove this with a different or? I know there is an issue that covers deprecated properties.

detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterListSpec.kt
```diff
@@ -8,22 +8,34 @@ import org.junit.jupiter.api.Test
...
-    private val reportMessageForConstructor = "The constructor(a: Int, b: Int) has too many parameters. " +
-        "The current threshold is set to $defaultThreshold."
+    private val reportMessageForFunction = "The function long(a: Int, b: Int, c: Int) has too many parameters. " +
+        "The current maximum allowed parameters are $defaultMaximumParameters."
+    private val reportMessageForConstructor = "The constructor(a: Int, b: Int, c: Int) has too many parameters. " +
+        "The current maximum allowed parameters are $defaultMaximumParameters."
 
     @Test
-    fun `reports too long parameter list`() {
+    fun `does not reports function parameter list that has exactly the allowed amount`() {
```
marschwar, line 25:
```suggestion
    fun `does not report function parameter list that has exactly the allowed amount`() {
```

detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterListSpec.kt
```diff
@@ -79,12 +91,24 @@ class LongParameterListSpec {
...
-        val config = TestConfig("constructorThreshold" to "1")
+        val config = TestConfig("allowedConstructorParameters" to "1")
         val rule = LongParameterList(config)
-        val code = "class LongCtor(a: Int)"
+        val code = "class LongCtor(a: Int, b: Int)"
         assertThat(rule.compileAndLint(code)).hasSize(1)
     }
 
+    @Test
+    fun `does not reports constructor parameter list that has exactly the allowed amount`() {
```
marschwar, line 101:
```suggestion
    fun `does not report constructor parameter list that has exactly the allowed amount`() {
```

marschwar (approved):

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
marschwar: Related to #6193 ?

schalkms:
> * What went wrong:
> At least one generated configuration file is not up-to-date. You can execute the generateDocumentation Gradle task to update generated files then commit the changes.

detekt-core/src/main/resources/deprecation.properties
```diff
@@ -1,5 +1,5 @@
 complexity>ComplexMethod=Rule is renamed to `CyclomaticComplexMethod` to distinguish between Cyclomatic Complexity and Cognitive Complexity
-complexity>LongParameterList>threshold=Use `functionThreshold` and `constructorThreshold` instead
+complexity>LongParameterList>threshold=Use `allowedFunctionParameters` and `allowedConstructorParameters` instead
```
cortinico, line 2: Commenting here but valid for all the PR you sent @PoisonedYouth.

Question for @detekt/maintainers 

Should we just remove this line as `threshold` will be gone?
Or should we instead add all the `threshold` parameters we removed, so users get warned on know how to update their config files?

TWiStErRob: Definitely second, it's major, let's help as much as possible. The facility is already there, let's use it!

cortinico: Reviving this discussion

> (If we're adventurous, introduce a separate parallel removed.properyies?)

Yes I feel we might have to do it as a separate PR for 2.0.0
The problem is that now we're effectively removing a lot of keys from the config file and we won't have an easy way for the user to tell what maps to what.

BraisGabin: I would create that mapping as documentation rather than code. Something similar to what the people of androidx does: https://developer.android.com/guide/topics/media/exoplayer/mappings

The reason is to keep 2.0 deprecation free (at least at launch) to reduce the tech debt.

TWiStErRob: Helping users migrate on major refactors is not tech debt :) It doesn't have to be deprecation, and doesn't have to live long, the first "at launch" release is the one where migration is important. The less "tech debt" we're causing, the faster the adoption and better reception.

BraisGabin: I would vote to merge this PR and move this discussion to an issue to talk about how are we going to provide tools to migrate from 1.x to 2.x

BraisGabin (approved):

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -131,8 +131,8 @@ complexity:
     allowedLines: 60
   LongParameterList:
     active: true
-    functionThreshold: 6
-    constructorThreshold: 7
+    allowedFunctionParameters: 6
+    allowedConstructorParameters: 7
```
marschwar, line 135: Was it a conscious decision to change the rule to be less strict than before? I apologize if the question was already answered before?

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
```diff
@@ -132,7 +133,7 @@ class LongParameterList(config: Config = Config.empty) : Rule(config) {
     }
 
     companion object {
-        private const val DEFAULT_FUNCTION_THRESHOLD = 6
-        private const val DEFAULT_CONSTRUCTOR_THRESHOLD = 7
+        private const val DEFAULT_ALLOWED_FUNCTION_PARAMETERS = 6
+        private const val DEFAULT_ALLOWED_CONSTRUCTOR_PARAMETERS = 7
```
TWiStErRob, line 137: Agreed with @marschwar, these should be decreased by 1
because of https://github.com/detekt/detekt/pull/6195/files#diff-0674ec2b75e040b6bdd727b396d7b18fc19254f9a731f5c0bdfd65c30e467c26R109

detekt-core/src/main/resources/default-detekt-config.yml
TWiStErRob: Good spot!

See also https://github.com/detekt/detekt/pull/6195/files#r1293875657

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
TWiStErRob: And probably all TestConfig setups in tests too.

detekt-core/src/main/resources/default-detekt-config.yml
TWiStErRob: Also in baseline.

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/LongParameterList.kt
PoisonedYouth: You're right, we should decrease the default to not change the behavior. I will do this soon.

detekt-core/src/test/resources/config_validation/baseline.yml
```diff
@@ -16,7 +16,7 @@ complexity:
     allowedLines: 20
   LongParameterList:
     active: false
-    functionThreshold: 5
+    allowedFunctionParameters: 5
```
BraisGabin, line 19:
```suggestion
    allowedFunctionParameters: 4
```

BraisGabin (approved):
LGTM

PR #6207, opened 2023-06-10
Replaced "threshold" property of NestedScopeFunctions rule by "allowedDepth".

PoisonedYouth:
Replaced "threshold" property of NestedScopeFunctions rule by "allowedDepth" (#3679)

The rule already correctly reported an issue if the depth of scope functions ts is greater than the
the number of the "allowedDepth" property. So the property is only renamed, no change
in functionality.


cortinico (approved):

cortinico:
Nit: maybe we should also add a rule inside `ruleauthors` that prevent from using `threshold` in the future at all?

BraisGabin (approved):

PR #6208, opened 2023-06-10
Replaced "threshold" property of StringLiteralDuplication rule by "allowedDuplications".

PoisonedYouth:
Replaced "threshold" property of StringLiteralDuplication rule by "allowedDuplications" (#3679)

The rule reported an issue when the amount of duplications is greater or equal the value defined for the "threshold" property.
It is intended that the value defined in the rule is the maximum allowed duplications. So the property is renamed to "allowedDuplications" and the check for reporting an issue is changed to only use greater and no longer equal.

### DIFF
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/StringLiteralDuplication.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/StringLiteralDuplication.kt
@@ -53,8 +53,8 @@ class StringLiteralDuplication(config: Config = Config.empty) : Rule(config) {
         Debt.FIVE_MINS
     )
 
-    @Configuration("amount of duplications to trigger rule")
-    private val threshold: Int by config(defaultValue = 3)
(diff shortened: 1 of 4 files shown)


detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/StringLiteralDuplication.kt
```diff
@@ -53,8 +53,8 @@ class StringLiteralDuplication(config: Config = Config.empty) : Rule(config) {
         Debt.FIVE_MINS
     )
 
-    @Configuration("amount of duplications to trigger rule")
-    private val threshold: Int by config(defaultValue = 3)
+    @Configuration("The maximum allowed amount of duplications.")
+    private val allowedDuplications: Int by config(defaultValue = 3)
```
BraisGabin, line 57: We should move this value to 2, right? Otherwise we are changing the behavior of the rule.

cortinico (approved):

marschwar (approved):

PR #6210, opened 2023-06-11
CanBeNonNullable: Fix false positive when property is defined after assignment

### LINKED ISSUE #6199: False Positive in CanBeNonNullable
## Expected Behavior
```
class FalsePositive {                        

    fun release() {
        fooBarObj = null
    }

    private var fooBarObj: FooBar? = FooBar()

}

class FooBar {
    // declare member variables and do something
}
```
This tool should not give any issue with CanBeNonNullable.

## Observed Behavior

CanBeNonNullable issue is coming

## Steps to Reproduce

```
@Test
fun failingTc() {
    val code = """
        class FalsePositive {                        

            fun release() {
                fooBarObj = null
            }

            private var fooBarObj: FooBar? = FooBar()

        }

        class FooBar {
            // declare member variables and do something
        }
    """.trimIndent()
}
```

## Your Environment

* Version of detekt used: 1.23.0 (latest) 
* Version of Gradle used (if applicable): 7.x

t-kameyama:
Fixes #6199

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
@@ -1,10 +1,11 @@
 package io.gitlab.arturbosch.detekt.rules.style
 
 import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.SourceLocation
 import io.gitlab.arturbosch.detekt.rules.KotlinCoreEnvironmentTest
+import io.gitlab.arturbosch.detekt.test.assertThat
 import io.gitlab.arturbosch.detekt.test.compileAndLint
 import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
-import org.assertj.core.api.Assertions.assertThat
 import org.intellij.lang.annotations.Language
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/CanBeNonNullableSpec.kt
```diff
@@ -143,6 +143,11 @@ class CanBeNonNullableSpec(val env: KotlinCoreEnvironment) {
                     private fun buzz(bizz: Int): Int? {
                         return if (bizz % 2 == 0) null else bizz
                     }
+
+                    fun baz() {
```
atulgpt, line 147: We can also add TC like below(to prevent any future break around false -ve)
```
fun baz() {
    val r = object: Runnable {
        var g: Int? = null
        override fun run() { g = null }
    }
}

private var g: Int? = 0
```

```
fun baz() {
    val g: Int? = null
    g = null
}
private var g: Int? = 0
```
Potential false +ve TC
```
fun baz(a: Int?) {
    g = a // now this should not be reported
}
private var g: Int? = 0
```
```
fun baz(a: Int?) {
    g = a?.also { println("assigning baz a to g") }
}
private var g: Int? = 0
```

t-kameyama: 241fb90188c908b9a5ea3bfe7477db0cd24c0370

3flex (approved):

PR #6211, opened 2023-06-11
Replaced "thresholdX" properties of TooManyFunctions rule by "allowedFunctionsX".

PoisonedYouth:
Replaced "thresholdX" property of TooManyFunctions rule by "allowedFunctionsX" (#3679)

The rule reported an issue when the number of functions is greater or equal the value defined for the "threshold" property.
It is intended that the value defined in the rule is the maximum allowed number of functions. So the property is renamed to
"allowedFunctionsX" and the check for reporting an issue is changed to only use greater and no longer equal.

### DIFF
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
@@ -39,20 +39,20 @@ class TooManyFunctions(config: Config = Config.empty) : Rule(config) {
         Debt.TWENTY_MINS
     )
 
-    @Configuration("The maximum allowed functions in files")
-    private val allowedFunctionsInFiles: Int by config(DEFAULT_THRESHOLD)
+    @Configuration("The maximum allowed functions per file")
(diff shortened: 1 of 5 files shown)


detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
```diff
@@ -39,20 +39,20 @@ class TooManyFunctions(config: Config = Config.empty) : Rule(config) {
         Debt.TWENTY_MINS
     )
 
-    @Configuration("threshold in files")
-    private val thresholdInFiles: Int by config(DEFAULT_THRESHOLD)
+    @Configuration("The maximum allowed functions in files")
+    private val allowedFunctionsInFiles: Int by config(DEFAULT_THRESHOLD)
```
marschwar, line 43: Question: Should this be `allowedFunctionsInFile` or `allowedFunctionsPerFile` instead?

PoisonedYouth: Having a look in the dictionary "per" is better matching the intention.

BraisGabin (approved):

PR #6212, opened 2023-06-11
Replaced "threshold" property of CouldBeSequence rule by "allowedOperations".

PoisonedYouth:
Replaced "threshold" property of CouldBeSequence rule by "allowedOperations" (#3679)

The rule reported an issue when the numbe of operations on a collection is greater or equal the value defined for the "threshold" property.
It is intended that the value defined in the rule is the maximum allowed number of operations. So the property is renamed to
"allowedOperations" and the check for reporting an issue is changed to only use greater and no longer equal.

### DIFF
--- a/detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/CouldBeSequence.kt
+++ b/detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/CouldBeSequence.kt
@@ -42,7 +42,7 @@ class CouldBeSequence(config: Config = Config.empty) : Rule(config) {
     )
 
     @Configuration("The maximum number of allowed chained collection operations.")
-    private val allowedOperations: Int by config(defaultValue = 3)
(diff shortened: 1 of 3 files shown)


detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/CouldBeSequence.kt
```diff
@@ -41,8 +41,8 @@ class CouldBeSequence(config: Config = Config.empty) : Rule(config) {
         Debt.FIVE_MINS
     )
 
-    @Configuration("the number of chained collection operations required to trigger rule")
-    private val threshold: Int by config(defaultValue = 3)
+    @Configuration("The maximum number of allowed chained collection operations.")
+    private val allowedOperations: Int by config(defaultValue = 3)
```
BraisGabin, line 45: I would low this number to 2 to avoid behavior change on the rule. If we want to change the behavior it should be done in a different PR.

BraisGabin (approved):

PR #6214, opened 2023-06-12
Bring DetektCreateBaselineTask closer to Detekt task

### LINKED ISSUE #6167: `jdkHome` is not passed to CLI from DetektCreateBaselineTask
(This is all in the context of type resolution enabled.)

## Expected Behavior

Specifying `jdkHome` in `DetektCreateBaselineTask` should have the same effect as the `Detekt` task.

## Observed Behavior

Specifying `jdkHome` in `DetektCreateBaselineTask` does nothing. I got the same errors as if I didn't specify `jdkHome` in `Detekt`, e.g. `error: unresolved reference: java`

## Context

We ran into this while trying to generate baselines with type resolution enabled. `detektBaselineMain` never seems to make `detektMain` happy. After some digging, we narrowed down to the `jdkHome` property.

We think this is due to missing plumbing of `jdkHome` in https://github.com/detekt/detekt/blob/9fb43530f67d0f70c818bdfc801e849eeac6a668/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt#L125

Compare it against: https://github.com/detekt/detekt/blob/9fb43530f67d0f70c818bdfc801e849eeac6a668/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt#L218

## Your Environment

* Gradle scan link (add `--scan` option when running the gradle task): (private project)
* Operating System and version: macOS 13.4
* Link to your project (if it's a public repository): (private project)

Hexcles:
* Add missing parameter `languageVersion`
* Fix plumbing of `jdkHome` (fixes #6167 )
* Convert all parameters into getter-setter style (so we can do e.g. `debug=true` in addition to `debug.set(true)`) and reorder them in the same way as `Detekt`.

This makes `DetektCreateBaselineTask.kt` almost a copy-pasta of https://github.com/detekt/detekt/blob/main/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt , but inheritance feels awkward, too. Maybe a base class (in `internal`)?

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
@@ -69,27 +69,63 @@ abstract class DetektCreateBaselineTask @Inject constructor(
     @get:Optional
     abstract val classpath: ConfigurableFileCollection
 
-    @get:Console
-    abstract val debug: Property<Boolean>
+    @get:Input
+    @get:Optional
+    internal abstract val languageVersionProp: Property<String>
+    var languageVersion: String
+        @Internal
+        get() = languageVersionProp.get()
+        set(value) = languageVersionProp.set(value)
+
+    @get:Input
+    @get:Optional
+    internal abstract val jvmTargetProp: Property<String>
+    var jvmTarget: String
+        @Internal
+        get() = jvmTargetProp.get()
+        set(value) = jvmTargetProp.set(value)
 
     @get:Internal
-    abstract val parallel: Property<Boolean>
+    abstract val jdkHome: DirectoryProperty
 
-    @get:Input
-    @get:Optional
-    abstract val disableDefaultRuleSets: Property<Boolean>
+    @get:Internal
+    internal abstract val debugProp: Property<Boolean>
+    var debug: Boolean
+        @Console
+        get() = debugProp.getOrElse(false)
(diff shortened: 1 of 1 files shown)


3flex:
Thanks for the PR, but I'm working on major changes to the plugin for detekt 2.0 which will address this issue but also mean this PR will be redundant.

Thank you for the PR though! Normally we'd be very open to it, but there will be a fair bit of churn as things develop for the new version.

3flex:
It's coming "soon" but there's no date. Probably longer than you'd like.

@cortinico thoughts on a 1.23.1 point release?

cortinico:
> @cortinico thoughts on a 1.23.1 point release?

Yup we'll be having a point release in the next days with JDK 20 and Kotlin 1.9 as those are major blockers (sorry for the late response but I was off)

Hexcles:
Would you be open in taking a minimal PR to fix this in the dot release?

---

cortinico:
> to fix this

@Hexcles what is "this"?

Hexcles:
Sorry I meant https://github.com/detekt/detekt/issues/6167

---

cortinico:
> Sorry I meant #6167

Yup that's doable, as long as it concerns `jdkHome` (and related JDK properties). As @3flex mentioned, we're working on 2.0 so this area will be reworked

Hexcles:
@cortinico see https://github.com/detekt/detekt/pull/6284

PR #6216, opened 2023-06-13
Standardize rule code sample spacing

dzirbel:
Fix some cases where rule kdoc had insufficient spacing (less than 4 spaces), which appears to remove the spacing entirely from the generated documentation; for example: https://detekt.dev/docs/rules/coroutines#injectdispatcher. Most rules already use 4 spaces for their code samples, this brings in line a few others I noticed, but I'm not certain it's an exhaustive fix.


schalkms (approved):
Great! Thanks for submitting such things. Look and feel in the documentation is important too!

PR #6244, opened 2023-07-01
Enable Gradle Predictive Test Selection by default

### LINKED ISSUE #6073: Re-enable Predictive Test Selection for local builds
## Expected Behavior
We use PTS where it speeds things up.

## Current Behavior
PTS isn't run on local builds by default.

## Context
I saw this in Gradle Enterprise Gradle Plugin release notes for 3.13.1:
> [NEW] Predictive Test Selection: Allow loading test task outputs from the Build Cache when Predictive Test Selection is enabled in Gradle 8.2 and later

In theory this will mitigate the issue with PTS breaking incremental builds, which is why it was turned off for local builds in https://github.com/detekt/detekt/pull/5826.

When Gradle 8.2 is released we can look at whether this helps with performance when running local builds.

3flex:
Thanks to changes in Gradle Enterprise Gradle Plugin 3.13.1, we should be able to use Predictive Test Selection and the Build Cache together, and get benefit from both.

From https://docs.gradle.com/enterprise/predictive-test-selection/#build_cache:

> Predictive Test Selection is complementary to Gradle Enterprise Build Cache.
>
> When Predictive Test Selection is enabled, test tasks/goals will store their results in the build cache if build caching is enabled, all tests are selected for execution, and the task/goal was successful. The results from running only a subset of all tests or failing tests are never stored in the build cache. Test tasks/goals will aim to reuse the test results from a prior build invocation if the build cache is enabled, regardless of whether Predictive Test Selection is currently enabled or not (requires Gradle 8.2+ and Gradle Enterprise Gradle plugin 3.13.1+ or Gradle Enterprise Maven extension 1.16+).
>
> On a cache miss, rerunning the same test suite for the same code-under-test with Predictive Test Selection enabled, previously selected tests that passed will not be selected again. Instead, Predictive Test Selection will select different tests, and typically fewer, for subsequent executions.

PTS is now enabled by default, except on the main branch or when generating code coverage statistics for Codecov.

Closes #6073

### DIFF
--- a/.github/workflows/codecoverage.yaml
+++ b/.github/workflows/codecoverage.yaml
@@ -6,6 +6,9 @@ on:
       - main
   pull_request:
 
+env:
+  ORG_GRADLE_PROJECT_enablePTS: false
+
 permissions:
   contents: read
 
--- a/.github/workflows/deploy-snapshot.yaml
+++ b/.github/workflows/deploy-snapshot.yaml
@@ -7,6 +7,7 @@ on:
 
 env:
   GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+  ORG_GRADLE_PROJECT_enablePTS: false
 
 permissions:
   contents: read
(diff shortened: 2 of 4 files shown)


BraisGabin (approved):
Nice! Let's see how this one works :)

chao2zhang (approved):

PR #6251, opened 2023-07-03
Simple property assignments in KTS

Goooler:
Follow up #6241.

https://blog.gradle.org/simpler-kotlin-dsl-property-assignment


BraisGabin (approved):

3flex:
Thanks! I'd thought about doing this but it's incubating so I found it a bit distracting in the IDE. My personal preference is to wait for 8.3 when Gradle plans to make it stable but I don't want to block it if others see the value.

cortinico (approved):

PR #6260, opened 2023-07-07
Kotlin 1.9.0 refactor

3flex:
Implement some Kotlin 1.9 features.


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/util/PerformanceMonitor.kt
```diff
@@ -23,39 +21,19 @@ class PerformanceMonitor {
...
+        finished.add(Entry(phase, timedBlockExecution.duration))
+        return timedBlockExecution.value
     }
 }
 
-typealias StartMillis = Long
-
 internal const val MONITOR_PROPERTY_KEY = "detekt.core.monitor"
 
 @OptIn(UnstableApi::class)
```
schalkms, line 37: Btw, do we still need this annotation?

schalkms (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/util/PerformanceMonitor.kt
3flex: `PropertiesAware` is annotated with `@UnstableApi` so we need to keep it for now.

PR #6269, opened 2023-07-09
Support passing multiple config files to compiler plugin

### LINKED ISSUE #6157: Compiler plugin error on more than one config file: `Wrong plugin option format: null, should be plugin:<pluginId>:<optionName>=<value>`
Create a basic project using the `1.23.0` compiler plugin and add 2 config files.

(in `foo/build.gradle.kts` )
```
detekt {
    parallel = true
    buildUponDefaultConfig = true
    config.setFrom(
           "/Users/me/git/project/detekt-config1.yml", 
           "/Users/me/git/project/detekt-config2.yml", 
    )
}
```

## Expected Behavior
Reads in and applies both config files with no compiler error.

## Observed Behavior

Build fails with:
```
> Task :foo:compileKotlin FAILED
e: Wrong plugin option format: null, should be plugin:<pluginId>:<optionName>=<value>
```

## Your Environment
* Version of detekt used: `1.23.0` (compiler plugin)
* Version of Gradle used (if applicable): `8.1.1`
* Operating System and version: macOS 12.6.5

3flex:
Fixes #6157. Tested locally.

### DIFF
--- a/detekt-compiler-plugin/src/main/kotlin/io/github/detekt/compiler/plugin/DetektCommandLineProcessor.kt
+++ b/detekt-compiler-plugin/src/main/kotlin/io/github/detekt/compiler/plugin/DetektCommandLineProcessor.kt
@@ -22,7 +22,8 @@ class DetektCommandLineProcessor : CommandLineProcessor {
             Options.config,
             "<path|paths>",
             "Comma separated paths to detekt config files.",
-            false
+            false,
+            allowMultipleOccurrences = true,
         ),
         CliOption(
             Options.configDigest,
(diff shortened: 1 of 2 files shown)


detekt-compiler-plugin/src/main/kotlin/io/github/detekt/compiler/plugin/DetektCommandLineProcessor.kt
```diff
@@ -98,7 +99,7 @@ class DetektCommandLineProcessor : CommandLineProcessor {
     override fun processOption(option: AbstractCliOption, value: String, configuration: CompilerConfiguration) {
         when (option.optionName) {
             Options.baseline -> configuration.put(Keys.BASELINE, Path(value))
-            Options.config -> configuration.put(Keys.CONFIG, value.split(",;").map { Path(it) })
```
cortinico, line 101: wasn't this already a list?

3flex: Problem was the separator is currently a comma, and that's also used as a delimiter between options passed to the Kotlin compiler in the CLI argument.

This seems the cleanest way to fix it. Now, you have to specify the config file argument for each config file, and the argument will be a single path, not multiple paths joined with a comma or semicolon.

schalkms (approved):

PR #6271, opened 2023-07-10
Use more appropriate methods to compile KtFiles

### LINKED ISSUE #5747: NullPointerException from `absolutePath()`
```
e: Analyzing 'DbModule.kt' led to an exception.
The original exception message was: root.absolutePath().parent must not be null
Running detekt '1.12.0' on Java '14.0.2+12-46' on OS 'Windows 10'.
If the exception message does not help, please feel free to create an issue on our GitHub page.
e: 

e: io.gitlab.arturbosch.detekt.rules.naming.InvalidPackageDeclaration.postVisit(InvalidPackageDeclaration.kt:45)
```

The problematic line:
https://github.com/detekt/detekt/blob/ab76be9f91aaaab7c0ff7f629d6c55daf03ab08f/detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/InvalidPackageDeclaration.kt#L45

`absolutePath()` extension returns only the filename when run through the compiler plugin, so `absolutePath().parent` is `null` causing the crash.

### LINKED ISSUE #6155: per-rule Path `excludes` don't work with the compiler plugin.
## Expected Behavior

This should ignore test code for the `FunctionNaming` rule. 

```
naming:
  FunctionNaming:
    excludes: ['**/test/**']
```

see https://detekt.dev/docs/introduction/configurations/#path-filters--excludes--includes

## Observed Behavior

Still fails the build in tests, `excludes` is ignored.

```
> Task :foo:compileTestKotlin FAILED
w: Analysis failed with 2 weighted issues.
e: warnings found and -Werror specified
w: file:///Users/me/git/my-project/foo/src/test/kotlin/com/corp/MyTest.kt:38:18 FunctionNaming: Function names should match the pattern: [a-z][a-zA-Z0-9]*
w: file:///Users/me/git/my-project/foo/src/test/kotlin/com/corp/MyTest.kt:44:18 FunctionNaming: Function names should match the pattern: [a-z][a-zA-Z0-9]*
```

to show that the config file is being picked up, this DOES work with the compiler plugin, but of course, it also disables it everywhere and we only want to exclude tests.

```
naming:
  FunctionNaming:
    active: false
```

## Steps to Reproduce

1. Create a simple project with tests and detekt compiler plugin
1. write a test named `this test name is doesnt follow the rules`
1. try to exclude the test sources

## Your Environment

* Version of Gradle used (if applicable): 8.1.1
* Operating System and version: macOS 12.6.5

3flex:
Update detekt core to ensure KtFiles are created using standard methods in the compiler. This means paths and filenames are correctly stored in each KtFile so this fixes #5747 and fixes #6155.

### DIFF
--- a/detekt-compiler-plugin/src/test/resources/hello.kt
+++ b/detekt-compiler-plugin/src/test/resources/hello.kt
@@ -1,6 +1,6 @@
 class KClass {
     fun foo() {
-        val x = 3
-        println(x)
+        var myList = mutableListOf(1,2,3)
+        println(myList)
     }
 }
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/ParsingStrategy.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/ParsingStrategy.kt
@@ -21,6 +21,5 @@ fun pathToKtFile(path: Path): ParsingStrategy = { settings ->
 }
 
 val inputPathsToKtFiles: ParsingStrategy = { settings ->
-    val compiler = KtTreeCompiler(settings, settings.spec.projectSpec)
-    settings.spec.projectSpec.inputPaths.flatMap(compiler::compile)
+    KtTreeCompiler(settings, settings.spec.projectSpec).compiler.environment.getSourceFiles()
 }
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/ParsingStrategy.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/ParsingStrategy.kt
@@ -1,10 +1,11 @@
 package io.gitlab.arturbosch.detekt.core.tooling
 
 import io.github.detekt.parser.KtCompiler
-import io.gitlab.arturbosch.detekt.core.KtTreeCompiler
+import io.gitlab.arturbosch.detekt.api.internal.PathFilters
 import io.gitlab.arturbosch.detekt.core.ProcessingSettings
 import org.jetbrains.kotlin.psi.KtFile
 import java.nio.file.Path
+import kotlin.io.path.Path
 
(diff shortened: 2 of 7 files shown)


detekt-compiler-plugin/src/test/resources/hello.kt
```diff
@@ -1,6 +1,6 @@
 class KClass {
     fun foo() {
-        val x = 3
-        println(x)
+        var myList = mutableListOf(1,2,3)
```
cortinico, line 3: Why are you changign this test?

cortinico (approved):
Code looks good to me 👍 What's the reason why this is in draft?

3flex:
> What's the reason why this is in draft?

Test failures - CI not passing for formatting module which requires further changes to make autocorrect work again.

detekt-compiler-plugin/src/test/resources/hello.kt
3flex: Please see commit messages: https://github.com/detekt/detekt/commit/d0e18968e0d2be52f4e8c421251b74b6541a73bb

schalkms (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/ParsingStrategy.kt
```diff
@@ -1,10 +1,11 @@
 package io.gitlab.arturbosch.detekt.core.tooling
 
 import io.github.detekt.parser.KtCompiler
-import io.gitlab.arturbosch.detekt.core.KtTreeCompiler
+import io.gitlab.arturbosch.detekt.api.internal.PathFilters
```
chao2zhang, line 4: Shall we also remove `KtTreeCompiler` as it becomes unused

3flex:
I couldn't think of a decent solution to support auto correct while using this approach. We might want to consider whether autocorrect should be supported at all by detekt, especially if we start pushing the compiler plugin since that can't reasonably support it.

Closing and replacing with https://github.com/detekt/detekt/pull/6310

PR #6284, opened 2023-07-14
Add missing jdkHome and languageVersion properties to DetektCreateBaselineTask

### LINKED ISSUE #6167: `jdkHome` is not passed to CLI from DetektCreateBaselineTask

Hexcles:
This is a minimal version of https://github.com/detekt/detekt/pull/6214 to fix #6167 only in a point release (1.23.1) given that 2.x will come with major changes in these tasks.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
@@ -108,6 +108,14 @@ abstract class DetektCreateBaselineTask @Inject constructor(
 
     @get:Input
     @get:Optional
+    internal abstract val languageVersionProp: Property<String>
+    var languageVersion: String
+        @Internal
+        get() = languageVersionProp.get()
+        set(value) = languageVersionProp.set(value)
+
+    @get:Input
+    @get:Optional
     internal abstract val jvmTargetProp: Property<String>
     var jvmTarget: String
         @Internal
@@ -122,7 +130,9 @@ abstract class DetektCreateBaselineTask @Inject constructor(
         get() = listOf(
             CreateBaselineArgument,
             ClasspathArgument(classpath),
+            LanguageVersionArgument(languageVersionProp.orNull),
             JvmTargetArgument(jvmTargetProp.orNull),
+            JdkHomeArgument(jdkHome),
             BaselineArgument(baseline.get()),
             InputArgument(source),
             ConfigArgument(config),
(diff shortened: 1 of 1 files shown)


Hexcles:
@cortinico

cortinico (approved):
@3flex thoughts?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
```diff
@@ -106,6 +108,14 @@ abstract class DetektCreateBaselineTask @Inject constructor(
         get() = basePathProp.get()
         set(value) = basePathProp.set(value)
 
+    @get:Input
+    @get:Optional
+    internal abstract val languageVersionProp: Property<String>
+    var languageVersion: String
+        @Internal
+        get() = languageVersionProp.get()
+        set(value) = languageVersionProp.set(value)
```
3flex, line 117: This can be replaced with the following to just expose the property directly:

```kotlin
    @get:Input 
     @get:Optional 
     abstract val languageVersionProp: Property<String>
```

3flex (approved):

cortinico:
> Note that I set the base branch to `release/1.x`, not `main`, with the intention to release this in 1.23.1 _only_ based on my understanding of prior discussions. Please change it if that's not correct.

Let's merge this to `main` and I'll pick it on the release branch

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCreateBaselineTask.kt
Hexcles: Pushed a new commit. Does it look good to you now?

PR #6287, opened 2023-07-15
Ensure paths are not used in tests where plain filenames are required

3flex:
When using KtPsiFactory to create files it uses methods from PsiFileFactory. The methods used to create new files use a filename argument and not a path argument. These methods have been called incorrectly in detekt in various ways - this PR ensures that tests will stop using incorrect values.

Includes a test change that's required as part of the fix for #5747 which can be made independently.


cortinico (approved):
Nice cleanup 👍

PR #6302, opened 2023-07-24
Skip entire subtrees while scanning for source files

### LINKED ISSUE #6240: Excluded files affects scanning speed
## Expected Behavior
Detekt runs fast with skipping excluded folders and files.

## Observed Behavior
Detekt checks all files, even if they are excluded and it takes a very long time

## Steps to Reproduce
-

## Context
I have a big android project with many modules. Each module folder contains **build** folder generated by gradle, which contains huge amount of generated files. All build directories are added to excludes with `**/build/**`.
If project is fresh without **build** directories, detekt task ends in 30 seconds.
When **build**  folders are present, detekt task ends in 5-10 minutes depends on machine.
I looked at the code of detekt and found out the problem is in include-exclude mechanism itself. Detekt takes every file even in and checks it path to include\exclude paths.
So if project contains giant amount of files in excluded directories, it significantly increases scanning time.

## Your Environment
* Version of detekt used:  v1.22.0, v1.23.0
* Version of Gradle used (if applicable): 8.1.1
* Operating System and version: Macos 13,  Windows 10

marschwar:
fixes #6240

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtTreeCompiler.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtTreeCompiler.kt
@@ -3,7 +3,6 @@ package io.gitlab.arturbosch.detekt.core
 import io.github.detekt.parser.KtCompiler
 import io.github.detekt.tooling.api.spec.ProjectSpec
 import io.gitlab.arturbosch.detekt.api.internal.PathFilters
-import io.gitlab.arturbosch.detekt.api.internal.pathMatcher
 import org.jetbrains.kotlin.psi.KtFile
 import java.io.IOException
 import java.nio.file.FileVisitResult
@@ -26,10 +25,6 @@ class KtTreeCompiler(
     private val basePath: Path? = projectSpec.basePath
     private val pathFilters: PathFilters? =
         PathFilters.of(projectSpec.includes.toList(), projectSpec.excludes.toList())
-    private val subtreeExcludes = projectSpec.excludes
-        .filter { it.endsWith("/**") }
-        .map { pathMatcher(it) }
-        .toList()
 
     fun compile(path: Path): List<KtFile> {
         require(path.exists()) { "Given path $path does not exist!" }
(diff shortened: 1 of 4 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtTreeCompiler.kt
```diff
@@ -21,6 +26,10 @@ class KtTreeCompiler(
     private val basePath: Path? = projectSpec.basePath
     private val pathFilters: PathFilters? =
         PathFilters.of(projectSpec.includes.toList(), projectSpec.excludes.toList())
+    private val subtreeExcludes = projectSpec.excludes
+        .filter { it.endsWith("/**") }
+        .map { pathMatcher(it) }
+        .toList()
```
marschwar, line 32: I am not sure that this is the ideal way to check for subtree globs

illuzor: Exceptions don't have to always end with /**
You can specify specific folders. For example `app/build`

marschwar: Thank you. Depending on the outcome of the conflict with the other work that is being done at the moment (see above), I will update this.

3flex:
I've replaced https://github.com/detekt/detekt/pull/6271 with https://github.com/detekt/detekt/pull/6310 which keeps `KtTreeCompiler`, so this won't be blocked after all.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtTreeCompiler.kt
```diff
@@ -57,12 +58,43 @@ class KtTreeCompiler(
...
+                settings.debug { "Ignoring file '$file'" }
+            } else {
+                _collected.add(file)
+            }
+            return FileVisitResult.CONTINUE
+        }
+
+        override fun visitFileFailed(file: Path, exc: IOException): FileVisitResult {
+            settings.error("Error visiting file '$file'.", exc)
+            return FileVisitResult.TERMINATE
```
BraisGabin, line 88: Should we just throw the exception? If we have an exception at this point the issue is really big and if we just ignore the scan will be only partial.

I know that this is a common pattern on detekt: Try to no fail. But this make the work of debugging issues way harder.

BraisGabin (approved):
LGTM.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtTreeCompiler.kt
marschwar: That is a good point. I have changed this.

PR #6311, opened 2023-07-28
Simplify storing metadata with KtFiles

3flex:
Use `UserDataHolder` interface and some of the available extension properties as well as `UserDataHolderBase` base class to simplify storing metadata with KtFiles.

### DIFF
--- a/detekt-psi-utils/api/detekt-psi-utils.api
+++ b/detekt-psi-utils/api/detekt-psi-utils.api
@@ -21,9 +21,12 @@ public final class io/github/detekt/psi/FilePath$Companion {
 }
 
 public final class io/github/detekt/psi/KeysKt {
-	public static final fun getBASE_PATH ()Lorg/jetbrains/kotlin/com/intellij/openapi/util/Key;
-	public static final fun getLINE_SEPARATOR ()Lorg/jetbrains/kotlin/com/intellij/openapi/util/Key;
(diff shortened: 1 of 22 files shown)


BraisGabin (approved):

schalkms (approved):
Way cleaner now 👍

detekt-psi-utils/api/detekt-psi-utils.api
```diff
@@ -21,20 +21,21 @@ public final class io/github/detekt/psi/FilePath$Companion {
 }
 
 public final class io/github/detekt/psi/KeysKt {
-	public static final fun getBASE_PATH ()Lorg/jetbrains/kotlin/com/intellij/openapi/util/Key;
-	public static final fun getLINE_SEPARATOR ()Lorg/jetbrains/kotlin/com/intellij/openapi/util/Key;
-	public static final fun getRELATIVE_PATH ()Lorg/jetbrains/kotlin/com/intellij/openapi/util/Key;
+	public static final fun getBasePath (Lorg/jetbrains/kotlin/com/intellij/psi/PsiFile;)Ljava/nio/file/Path;
```
cortinico, line 24: Is this a breaking change? Do we expect users to depend on those?

3flex: Probably not, but it's part of the API at the moment, so I figured it should be called out.

If it should never have been part of the public API then it should have been given an internal package or something instead - and the new property extensions should go into that internal package too.

PR #6315, opened 2023-07-29
Remove withAutoCorrect from ConfigAware

3flex:
This is not called by detekt at any point, so should be removed from the API.

### DIFF
--- a/detekt-api/api/detekt-api.api
+++ b/detekt-api/api/detekt-api.api
@@ -102,7 +102,6 @@ public abstract interface class io/gitlab/arturbosch/detekt/api/ConfigAware : io
 	public abstract fun subConfig (Ljava/lang/String;)Lio/gitlab/arturbosch/detekt/api/Config;
 	public abstract fun valueOrDefault (Ljava/lang/String;Ljava/lang/Object;)Ljava/lang/Object;
 	public abstract fun valueOrNull (Ljava/lang/String;)Ljava/lang/Object;
-	public abstract fun withAutoCorrect (Lkotlin/jvm/functions/Function0;)V
 }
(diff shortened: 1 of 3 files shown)


schalkms (commented):
> This is not called by detekt at any point, so should be removed from the API.

But custom rules could rely on it without our knowledge.

3flex:
We can break the API in v2 though - or is there some other reason to keep it? I'm not sure how custom rules would be using it.

schalkms:
> We can break the API in v2 though - or is there some other reason to keep it? I'm not sure how custom rules would be using it.

Yes, sure. V2 allows to break the API. I mean it needs to be communicated accordingly, probably in the change log.
Custom rules that want to auto correct code could rely on this API function. However, it's easy to refactor the code from a client perspective.

cortinico:
> We can break the API in v2 though - or is there some other reason to keep it? I'm not sure how custom rules would be using it.

+1 for removing. We might deprecate it in the 1.x branch if we want. Still if this was broken, no one was using it.
We should have an example of a custom rule that modifies the AST though, as I was thinking this function was the entry point to it, but seems like this is not the case.

cortinico (approved):

schalkms:
I created a ticket for this idea in the custom-rule-template repo.
https://github.com/detekt/detekt-custom-rule-template/issues/71

eygraber:
I was using this, and it was working fine. What are my alternatives? Just use `if(autoCorrect)`?

BraisGabin:
> Just use `if(autoCorrect)`?

Yes. Or create your own extension function if you find the other syntax more readable.

PR #6331, opened 2023-07-30
Correctly set scheme for URIs in the SARIF report output

### LINKED ISSUE #4165: Sarif output from detekt is not finding paths on GitHub
## Expected Behavior
image

## Observed Behavior
image

## Steps to Reproduce
<details><summary>High level guidance</summary>

Follow: https://detekt.github.io/detekt/reporting.html#integration-with-github-code-scanning
Then make a change that triggers Detekt and view the annotations.
And view code scanning annotations for the PR:
image
image
</details>

 * Create a repo on GitHub
 * Get https://github.com/TWiStErRob/repros/tree/main/detekt/sarif-github-previews locally
 * `git init .`
 * `git remote add origin <url>`
 * Review CI.yml and choose the branch in checkout (e.g. detekt/detekt v1.23.1)
 * commit and push
 * CI will execute as [shown in the testing setup](https://github.com/detekt/detekt/pull/6331#issuecomment-1660541102)

## Context
I'm trying to set up Code scanning as suggested here: https://github.com/detekt/detekt/issues/3045#issuecomment-933033224

I was trying to make the annotations show the preview, but the default setup didn't let me. Then I looked at the code of the action referenced in the doc:
https://github.com/github/codeql-action/blob/7d56b0cd44f66136d094ee8b67d457c9cec4b8ef/src/upload-lib.ts#L376
this replaces the default `${{ github.workspace }}` file path with a file uri (i.e. prepending with `file://` in case of an absolute path.

then I see this in the documentation of the github endpoint:
https://docs.github.com/en/rest/reference/code-scanning#upload-an-analysis-as-sarif-data

I tried to get a reference from documentation on this, but it's really hard to understand:
https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317430

So finally I understood that the `uri` property must be a URI to work with GitHub, so I added this build step:
```
      - name: Fix detekt paths.
        run: >
          sed -re 's!"uri": "${{ github.workspace }}!"uri": "file://${{ github.workspace }}!g'\
              <build/reports/detekt/detekt.sarif\
              >build/reports/detekt/detekt-fixed.sarif
```
and uploaded the resulting file. Notice that the sarif file still has absolute URIs in them, but in the screenshot they're showing relative paths.

## Your Environment

* Version of Gradle used (if applicable): 6.9
* Gradle scan link (add `--scan` option when running the gradle task): -
* Operating System and version: ubuntu-latest (GitHub actions)
* Link to your project (if it's a public repository): wouldn't see much because probably I will have fixed all issue by the time you see this. Refer to screenshots.

3flex:
Fixes #4165

@TWiStErRob it would be great if you could test this - I've reviewed the output locally but haven't submitted anything to GitHub.

### DIFF
--- a/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/Results.kt
+++ b/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/Results.kt
@@ -58,7 +58,7 @@ private fun Location.toLocation(code: String?): io.github.detekt.sarif4k.Locatio
                     uriBaseID = SRCROOT
                 )
             } else {
-                ArtifactLocation(uri = "file:///${filePath.absolutePath.invariantSeparatorsPathString}")
+                ArtifactLocation(uri = filePath.absolutePath.toUri().toString())
             }
         )
     )
--- a/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/Results.kt
+++ b/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/Results.kt
@@ -58,7 +58,7 @@ private fun Location.toLocation(code: String?): io.github.detekt.sarif4k.Locatio
                     uriBaseID = SRCROOT
                 )
             } else {
-                ArtifactLocation(uri = filePath.absolutePath.invariantSeparatorsPathString)
+                ArtifactLocation(uri = "file:///${filePath.absolutePath.invariantSeparatorsPathString}")
             }
         )
     )
--- a/detekt-report-sarif/src/test/resources/vanilla.sarif.json
+++ b/detekt-report-sarif/src/test/resources/vanilla.sarif.json
@@ -10,7 +10,7 @@
             {
               "physicalLocation": {
                 "artifactLocation": {
-                  "uri": "file:///TestFile.kt"
+                  "uri": "<PREFIX>TestFile.kt"
                 },
                 "region": {
                   "startColumn": 1,
@@ -30,7 +30,7 @@
             {
               "physicalLocation": {
                 "artifactLocation": {
-                  "uri": "file:///TestFile.kt"
+                  "uri": "<PREFIX>TestFile.kt"
                 },
                 "region": {
                   "startColumn": 1,
@@ -50,7 +50,7 @@
             {
               "physicalLocation": {
                 "artifactLocation": {
-                  "uri": "file:///TestFile.kt"
+                  "uri": "<PREFIX>TestFile.kt"
                 },
                 "region": {
                   "startColumn": 1,
--- a/detekt-report-sarif/src/test/resources/vanilla.sarif.json
+++ b/detekt-report-sarif/src/test/resources/vanilla.sarif.json
@@ -10,7 +10,7 @@
             {
               "physicalLocation": {
                 "artifactLocation": {
-                  "uri": "TestFile.kt"
+                  "uri": "file:///TestFile.kt"
                 },
                 "region": {
                   "startColumn": 1,
@@ -30,7 +30,7 @@
             {
               "physicalLocation": {
                 "artifactLocation": {
-                  "uri": "TestFile.kt"
+                  "uri": "file:///TestFile.kt"
                 },
                 "region": {
                   "startColumn": 1,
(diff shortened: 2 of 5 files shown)


detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/Results.kt
```diff
@@ -58,7 +58,7 @@ private fun Location.toLocation(code: String?): io.github.detekt.sarif4k.Locatio
                     uriBaseID = SRCROOT
                 )
             } else {
-                ArtifactLocation(uri = filePath.absolutePath.invariantSeparatorsPathString)
+                ArtifactLocation(uri = "file:///${filePath.absolutePath.invariantSeparatorsPathString}")
```
cortinico, line 61: shouldn't this be `file://`?

TWiStErRob: Chrome also keeps changing `//` to `///`: `file:///P:/projects/foo/bar.baz`.

It's 3 slashes per spec: https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes?

TWiStErRob:
I'll try to test tomorrow.

detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/Results.kt
BraisGabin: But for Unix paths we should just hardcode two because the third will be added by the absolute path, right?

I would create an extension function `Path.toUrl()` and test it (if it doesn't exist already at the default library).

BraisGabin (approved):

detekt-report-sarif/src/test/resources/vanilla.sarif.json
```diff
@@ -10,7 +10,7 @@
             {
               "physicalLocation": {
                 "artifactLocation": {
-                  "uri": "TestFile.kt"
+                  "uri": "<PREFIX>TestFile.kt"
```
TWiStErRob, line 13: Is it possible this is highlighting a regression?
Is it possible without basepath these `uri` will point to a non-existent location?

detekt-report-sarif/src/test/resources/vanilla.sarif.json
TWiStErRob, line 1: The file is versioned (driver) with detekt `1.16.0`, is that normal?

TWiStErRob (approved):
GitHub is now very happy with these changes!

https://github.com/TWiStErRob/detekt-4165/security/code-scanning?query=pr%3A1+tool%3Adetekt+is%3Aopen

<details><summary>Screenshots</summary>

image

image

</details>

Notice:
 * There were 6 reports in the past, linux, mac, windows. 4/6 were not navigable.
 * After your recent force-push, the original 6 issues were closed, and now there are only 2 detected 3 times by separate CI jobs.
 * This means that the paths (including base dirs) cancel each other out, and it doesn't matter which platform you run detekt on, GitHub will see them just the same! 🎉

detekt-report-sarif/src/test/resources/vanilla.sarif.json
3flex: Since https://github.com/detekt/detekt/pull/6323 it shouldn't be possible using CLI or Gradle to run detekt with a null base path, so I'm not too concerned about it.

From memory I started refactoring the SARIF report module to make basePath non-nullable but it became a bit more work than I was prepared to do at the time.

detekt-report-sarif/src/test/resources/vanilla.sarif.json
3flex: There's a MANIFEST file in test resources which has the value set to 1.16.0, so it's expected in this case.

BraisGabin (approved):

3flex:
@TWiStErRob if you could have a final look before merging that would be appreciated, thanks!

TWiStErRob:
Trying to run the integration tests, but need to update the repro with Gradle 8.4 and detekt mainline.

TWiStErRob:
Seems ok https://github.com/TWiStErRob/detekt-4165-2/security/code-scanning/2

The same issue is detected by all 3 platforms and preview is working:
image

hfhbd:
Any chance to get a release containing this fix before detekt 2.0?

cortinico:
> Any chance to get a release containing this fix before detekt 2.0?

I'll include this in the upcoming 1.23.x release

PR #6339, opened 2023-07-30
MaybeConstant: Remove unused alias

TWiStErRob:
Fixes https://github.com/detekt/detekt/issues/3643


marschwar (approved):
Good catch. Interesting side note: `@Suppress("MayBeConstant")` is marked as redundant by IntelliJ, while `@Suppress("MayBeConst")` seems to be accepted.

TWiStErRob:
> marked as redundant by IntelliJ

Could that be because the IJ Detekt plugin doesn't have the rename yet?

cortinico (approved):

marschwar:
> > marked as redundant by IntelliJ
> 
> Could that be because the IJ Detekt plugin doesn't have the rename yet?

No, I do not have the detekt plugin installed. It seems to be part of the checks that are shipped with the IDE.

Issue #2582, opened 2020-04-06
[V1.6.0 -> V1.7.4] Error reading configuration file, java.util.zip.ZipException: invalid code lengths set.


hangduykhiem:
## Expected Behavior
Detekt gradle task finish successfully when there are changes to custom `RuleProvider` or `Rule`.

## Observed Behavior
Detekt gralde task failed with the following stack. 

```
Caused by: java.util.ServiceConfigurationError: io.gitlab.arturbosch.detekt.api.RuleSetProvider: Error reading configuration file
        at io.gitlab.arturbosch.detekt.core.RuleSetLocator.load(RuleSetLocator.kt:29)
        at io.gitlab.arturbosch.detekt.cli.config.DefaultPropertiesConfigValidator$validate$1.invoke(DefaultPropertiesConfigValidator.kt:20)
        at io.gitlab.arturbosch.detekt.cli.config.DefaultPropertiesConfigValidator.validate(DefaultPropertiesConfigValidator.kt:27)
        at io.gitlab.arturbosch.detekt.cli.config.ConfigValidatorsKt.checkConfiguration(ConfigValidators.kt:15)
        at io.gitlab.arturbosch.detekt.cli.runners.Runner.execute(Runner.kt:31)
        at io.gitlab.arturbosch.detekt.invoke.DefaultCliInvoker.invokeCli(DetektInvoker.kt:51)
        ... 89 more
Caused by: java.util.zip.ZipException: invalid code lengths set
        ... 95 more
```

## Steps to Reproduce
1) Remove a rule from a `RuleProvider`
2) Run gradle detekt task.

## Your Environment
* Version of detekt used: v1.7.4
* Version of Gradle used (if applicable): v5.6.4
* Operating System and version: Mac OSX 10.14 Mojave and Arch 5.5.4-arch1-1

hangduykhiem:
### More reproduction:
1) Run `./gradlew detekt` on current master
2) Remove for example the rule `ChainWrapping` from file `io.gitlab.arturbosch.detekt.formatting.KtLintMultiRule`
3) Rerun the gradle task

arturbosch:
This happens due to classloader caching of the newer Gradle plugins (1.7.0+).
What you can do is to stop all daemons with `gradle --stop` and restart with a fresh classloader or start builds without a daemon `gradle build --no-daemon`.

I think we should introduce an additional timestamp comparision ...

hangduykhiem:
[...]

Issue #2411, opened 2020-03-05
RuleSetProvider: Error reading configuration file


r4zzz4k:
## Expected Behavior
Running `./gradlew detekt` on detekt itself should not fail with exception.

## Observed Behavior
Running `./gradlew detekt` sometimes returns the following:
```
Ignoring listeners of task graph ready event, as this build (:detekt-gradle-plugin) has already executed work.
> Task :detekt-api:detekt FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':detekt-api:detekt'.
> io.gitlab.arturbosch.detekt.api.RuleSetProvider: Error reading configuration file
```
Stacktrace provides a bit of additional information:
```
Caused by: java.util.ServiceConfigurationError: io.gitlab.arturbosch.detekt.api.RuleSetProvider: Error reading configuration file
        at io.gitlab.arturbosch.detekt.core.RuleSetLocator.load(RuleSetLocator.kt:29)
        at io.gitlab.arturbosch.detekt.cli.config.DefaultPropertiesConfigValidator$validate$1.invoke(DefaultPropertiesConfigValidator.kt:20)
        at io.gitlab.arturbosch.detekt.cli.config.DefaultPropertiesConfigValidator.validate(DefaultPropertiesConfigValidator.kt:27)
        at io.gitlab.arturbosch.detekt.cli.config.ConfigValidatorsKt.checkConfiguration(ConfigValidators.kt:18)
        at io.gitlab.arturbosch.detekt.cli.runners.Runner.execute(Runner.kt:30)
        at io.gitlab.arturbosch.detekt.invoke.DefaultCliInvoker.invokeCli(DetektInvoker.kt:51)
        ... 93 more
Caused by: java.util.zip.ZipException: invalid distance too far back
        ... 99 more
```

## Steps to Reproduce
I'm not sure it's reproducable 100% of times, but still:
1. Clone PR branch for #2410
2. Checkout it's parrent commit (4ea5873)
3. Run `./gradlew detekt` (should work as expected)
4. Checkout PR branch
5. Run ./gradlew detekt`

## Context
I'm trying to contribute to the project, and I'm unable to walk through steps of contribution guildelines.

## Your Environment
* Version of detekt used: 1.7.0-beta1 (master branch, ref 4ea5873)
* Version of Gradle used (if applicable): 6.2.1
* Operating System and version: Ubuntu 19.10
* Java version just in case:
```
openjdk version "1.8.0_242"
[...]

PR #6340, opened 2023-07-31
Detekt Gradle Plugin should no longer use an invalid cache

### LINKED ISSUE #2957: Changing any core class invalidates cached detekt classloader

### LINKED ISSUE #3396: Getting "io.gitlab.arturbosch.detekt.api.RuleSetProvider: Error reading configuration file" error sometimes

fan87:
Detekt Gradle Plugin uses an invalid Class loader cache and an invalid JarFile cache. Once the Jar is modified, the cache should be invalidated, or some should be disabled.

The Service API uses `URL`, which leads to using `JarFileFactory` on handing jars, with no convenient way of modifying whether the Jar file should be cached. As the code in ServiceLoader is already written for `openStream()` to occur at all times, the practical solution is to disable cache process-wide. 

This only occurs on DetektInvoker and isn't an issue with the Worker API. 

- fixes #2957 ([Way to reproduce](https://github.com/detekt/detekt/issues/2957#issuecomment-1132145945))
- fixes #3396 

#2582 & #2411 re-occurred, and this should fix it again with the `InputChanges` API

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -54,6 +54,8 @@ import org.gradle.api.tasks.TaskAction
 import org.gradle.api.tasks.VerificationTask
 import org.gradle.api.tasks.options.Option
 import org.gradle.language.base.plugins.LifecycleBasePlugin
+import org.gradle.work.Incremental
+import org.gradle.work.InputChanges
 import org.gradle.workers.WorkerExecutor
 import java.io.File
 import javax.inject.Inject
@@ -69,6 +71,7 @@ abstract class Detekt @Inject constructor(
     abstract val detektClasspath: ConfigurableFileCollection
 
     @get:Classpath
+    @get:Incremental
     abstract val pluginClasspath: ConfigurableFileCollection
 
     @get:InputFile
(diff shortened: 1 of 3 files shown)


fan87:
The pr is broken

Issue #6364, opened 2023-08-03
ExplicitItLambdaParameter false positive for un-inferrable lambda


TWiStErRob:
## Expected Behavior
No report when removing would result in compilation error.

## Observed Behavior
The below code in repro flags ``Detekt - ExplicitItLambdaParameter: This explicit usage of `it` as the lambda parameter name can be omitted.``, and if I do omit I get:
![image](https://github.com/detekt/detekt/assets/2906988/4bebc2d1-3299-4e29-aeeb-2ed786b463b3)
because the inference from the return type doesn't pass through `::invoke`, maybe in K2? but more likely not.

## Steps to Reproduce
```kotlin
operator fun <T> KFunction1<T, Boolean>.not(): KFunction1<T, Boolean> =
	{ it: T -> !this(it) }::invoke
```

## Context
https://stackoverflow.com/a/76828982/253468, detekt flags all lambdas before `::invoke`.

## Your Environment

* Version of detekt used:
* Version of Gradle used (if applicable):
* Gradle scan link (add `--scan` option when running the gradle task):
* Operating System and version:
* Link to your project (if it's a public repository):

atulgpt:
Hi @TWiStErRob more simpler example would be as follow
```
fun test(): (Int) -> Int {
    return { it: Int -> it.inc() }::invoke
}
```

PR #6367, opened 2023-08-05
Fix false +ve when lambda is un-inferrable

atulgpt:
Fixes: https://github.com/detekt/detekt/issues/6364

### DIFF
--- a/detekt-psi-utils/api/detekt-psi-utils.api
+++ b/detekt-psi-utils/api/detekt-psi-utils.api
@@ -150,6 +150,8 @@ public final class io/gitlab/arturbosch/detekt/rules/ThrowExtensionsKt {
 }
 
 public final class io/gitlab/arturbosch/detekt/rules/TraversingKt {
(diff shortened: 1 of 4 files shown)


cortinico (approved):

cortinico:
@TWiStErRob do we need this on the 1.x branch?

atulgpt:
Hi @SomeTroglodyte, currently my fix doesn't change the behaviour of your code

Currently as per Detekt below code snippets are compatible

```kotlin
interface IThing {
    val name: String
    fun getComparator(): Comparator<IThing>
}
class Thing(override val name: String) : IThing {
    override fun getComparator() = compareBy<IThing, String>(Collator.getInstance()) { it.name }
}
```

or

```kotlin
interface IThing {
    val name: String
    fun getComparator(): Comparator<IThing>
}
class Thing(override val name: String) : IThing {
    override fun getComparator(): Comparator<IThing> = compareBy(Collator.getInstance()) { it.name }
}
```

TWiStErRob:
or
```kotlin
fun getComparator() = compareBy(Collator.getInstance(), IThing::name)
```
(Sometimes a fix for a detection is using a different language feature, which makes it even more readable IMO.)

---

That said, I think you should report this use case separately if you want it to work.

Issue #6342, opened 2023-07-31
`ImplicitDefaultLocale` should also catch `"%d".format(1)`


marschwar:
## Expected Behavior of the rule
The rule `ImplicitDefaultLocale` catches usages of `String.format` without a locale. Using the format extension function without a locale should also be non-compliant.

```kotlin
// non-compliant
String.format("%d", 1)
"%d".format(1)

// compliant
String.format("%d", Locale.US, 1)
"%d".format(Locale.US, 1)
```

atulgpt:
Hi, @marschwar I think for handling the`"%d".format(1)` case we may need to make this rule dependent on TR what are your thoughts on this?

marschwar:
It already depends on type resolution. It is one of those rules with mixed behavior.
See #6342

PR #6368, opened 2023-08-05
Check format extension function in ImplicitDefaultLocale

atulgpt:
Fixes https://github.com/detekt/detekt/issues/6342

### DIFF
--- a/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ImplicitDefaultLocaleSpec.kt
+++ b/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ImplicitDefaultLocaleSpec.kt
@@ -33,6 +33,52 @@ class ImplicitDefaultLocaleSpec(private val env: KotlinCoreEnvironment) {
     }
 
     @Test
+    fun `does not report custom String_format call`() {
+        val code = """
+            fun String.Companion.format(format: String, value: Int) = format + value.toString()
+
+            fun x() {
+                String.format("%d", 1)
+            }
+        """.trimIndent()
+        assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
+    }
+
+    @Test
+    fun `reports format extension call with template but without explicit locale`() {
+        val code = """
+            fun x() {
+                "%d".format(1)
+            }
+        """.trimIndent()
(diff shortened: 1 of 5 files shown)


atulgpt:
Hi, @3flex one strange observation here, when I run `./gradlew detektMain` build successfully passes(build scan at https://scans.gradle.com/s/ci2o5a2sfqvkq). But when I run `./gradlew detektMain --offline --rerun-tasks` build fails with below error(build scan at https://scans.gradle.com/s/yilo2rorrwnqi)
```
> Task :detekt-report-html:detektMain FAILED
/Users/atulgup/Learning/open_source/detekt/detekt-report-html/src/main/kotlin/io/github/detekt/report/html/HtmlUtils.kt:31:47: "%1$4s ".format(currentLineNumber) uses implicitly default locale for string formatting. [ImplicitDefaultLocale]
```
I have changed the rule and as per the updated rule it should detect the above issue even ruining `./gradlew detektMain`

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/ImplicitDefaultLocaleSpec.kt
```diff
@@ -32,6 +32,52 @@ class ImplicitDefaultLocaleSpec(private val env: KotlinCoreEnvironment) {
         assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
     }
 
+    @Test
+    fun `does not report custom String_format call`() {
+        val code = """
+            fun String.Companion.format(format: String, value: Int) = format + value.toString()
+
+            fun x() {
+                String.format("%d", 1)
```
BraisGabin, line 41: There is a missing test here, right? You aren't testing that this without the extension function should be reported.

atulgpt: Are you talking bout existing `reports String_format call with template but without explicit locale`? This TC is already there. In this TC I am checking our own defined custom method should not report the violation. Earlier we used name matching for detecting the violations and I moved that to `FQName` hence I added this TC

marschwar (commented):
Should we also add those new cases to the compliant/non-compliant examples?

atulgpt:
> Should we also add those new cases to the compliant/non-compliant examples?

Updated the code to include the extension method examples

marschwar (approved):

Issue #6263, opened 2023-07-07
Deprecate or remove UntilInsteadOfRangeTo


3flex:
## Expected Behavior of the rule
It doesn't exist.

## Context
This rule recommends use of `until` infix function, but Kotlin 1.9 makes the `..<` operator stable and the recommendation now is to use that instead. It was introduced in Kotlin 1.7.20, and now it's stable it's a safe recommendation.

See #6262 which requests introduction of a new rule that would replace this one.

cortinico:
> Deprecate or remove UntilInsteadOfRangeTo

I'd say we can effectively "rename" this rule in 2.0.0 to the one you suggested in #6262 and also update the logic

Issue #6262, opened 2023-07-07
Introduce RangeUntilInsteadOfRangeTo


3flex:
## Expected Behavior of the rule
Recommend replacing ranges like `0 .. 10 - 1` with `0..<10`

## Context
The `rangeTo` operator (`..<`) is recommended when creating open-ended ranges as of Kotlin 1.9.

PR #6369, opened 2023-08-05
Suggest `..<` instead of `until`

atulgpt:
Fixes https://github.com/detekt/detekt/issues/6262
Fixes https://github.com/detekt/detekt/issues/6263


atulgpt:
> @atulgpt can we fix `detekt with type resolution`?

Updated the code. It should be green now

BraisGabin (approved):

cortinico:
This sadly can't be picked on the 1.x release branch as it contains a breaking change.

Issue #6371, opened 2023-08-05
False negative on OutdatedDocumentation around private property


atulgpt:
## Expected Behavior
```kotlin
@Test
fun `should report when param which is private property is documented as property`() {
    val code = """
        /**
         * Doc
         * @property a desc
         */
        class A(
            private val a: String,
        )
    """.trimIndent()
    assertThat(subject.compileAndLint(code)).hasSize(1)
}
```
The above TC should pass. As `a` is private property and hence should only be documented as `@param`

## Observed Behavior
The above TC is failing

## Steps to Reproduce
Run the TC against the main code

## Context
While seeing the issue at https://github.com/detekt/detekt/issues/6362, the above TC failed

## Your Environment

* Version of detekt used: latest main
* Version of Gradle used (if applicable): detekt repo
* Gradle scan link (add `--scan` option when running the gradle task):
* Operating System and version: NA
* Link to your project (if it's a public repository): detekt repo

PR #6372, opened 2023-08-05
OutdatedDocumentation - Detect param which private property documented as property

atulgpt:
Fixes https://github.com/detekt/detekt/issues/6371


detekt-rules-documentation/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/documentation/OutdatedDocumentation.kt
```diff
@@ -147,7 +148,7 @@ class OutdatedDocumentation(config: Config = Config.empty) : Rule(config) {
     private fun getDeclarationsForValueParameters(valueParameters: List<KtParameter>): List<Declaration> {
         return valueParameters.mapNotNull {
             it.name?.let { name ->
-                val type = if (it.isPropertyParameter()) {
+                val type = if (it.isPropertyParameter() && it.isPublic) {
```
TWiStErRob, line 151: probably !private? because protected is also public API. `internal` is arguable, but there are still external users outside the component itself, so docs are recomended.

atulgpt: Nice catch.. updated the logic. Thanks

BraisGabin (approved):

PR #6381, opened 2023-08-12
Consider deprecated rules as inactive when running `allRules`

### LINKED ISSUE #6336: Cannot disable OptionalWhenBraces with warningsAsErrors: true
## Expected Behavior
Should be able to disable deprecated rules, or maybe be disabled by default?

## Observed Behavior
Either have many OptionalWhenBraces violations or
```
Execution failed for task ':...:detekt'.
        - Property 'style>OptionalWhenBraces' is deprecated. Same functionality is implemented in BracesOnWhenStatements.
```

## Steps to Reproduce
```gradle.kts
detekt {
    allRules = true
}
```
```yaml
config:
  warningsAsErrors: true
style:
  OptionalWhenBraces:
    active: false
```

## Context
I shouldn't have to choose between having a strict build, or having a short configuration file (allRules + overrides), or having a customized detekt config (active: false). I'm fully sure what's going on here, and why this never happened before, but it's a strange situation and I wonder what you guys think about it.

Can you think of a workaround for this apart from disabling `warningsAsErrors`, or listing all the rules explicitly rather than excluding the ones I don't want?

## Your Environment

* Version of detekt used: 1.23.1
* Version of Gradle used (if applicable): 8.3 rc 2

marschwar:
(quick) fixes #6336 

**Note: This PR targets the `release/1.x` branch instead of `main`.** 

Following [this suggestion](https://github.com/detekt/detekt/issues/6336#issuecomment-1674944509) I hacked something up that should fix the issue short term. The idea is that all deprecated rules are considered as inactive in case detekt is run with the `allRules` option. We should probably find a better solution for the main branch.

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
@@ -209,10 +209,9 @@ internal fun ProcessingSpec.workaroundConfiguration(config: Config): Config = wi
     return declaredConfig ?: getDefaultConfiguration()
 }
 
-internal fun loadDeprecatedRuleIds(): Set<String> {
-    return loadDeprecations().keys
+internal fun loadDeprecatedRuleIds(): Set<String> =
+    loadDeprecations().keys
         .map { it.split(">") }
         .filter { it.size == 2 }
         .map { it.joinToString(" > ") }
         .toSet()
-}
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
@@ -194,7 +194,16 @@ internal fun ProcessingSpec.workaroundConfiguration(config: Config): Config = wi
 
     if (rulesSpec.activateAllRules) {
         val defaultConfig = getDefaultConfiguration()
-        declaredConfig = AllRulesConfig(declaredConfig ?: defaultConfig, defaultConfig)
+        declaredConfig = AllRulesConfig(
+            originalConfig = declaredConfig ?: defaultConfig,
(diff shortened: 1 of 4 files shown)


BraisGabin:
🤔 I'm not sure here. But what will happen if you read the deprecation file? You should be able to extract that list too, right? This way you don't need to hardcode the list of deprecated rules.

marschwar:
>  But what will happen if you read the deprecation file? You should be able to extract that list too, right? This way you don't need to hardcode the list of deprecated rules.

I totally agree. This could be the actual fix for the main branch. This was meant to be a quick fix only in case we want to deliver something quickly. But I guess that still could be done here. I will update the PR.

BraisGabin (approved):

BraisGabin:
Then this PR should go against 2.0, right?

marschwar:
> Then this PR should go against 2.0, right?

No, it is intended for the 1.x only. For the 2.x I would like to  use #6382.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -203,3 +208,11 @@ internal fun ProcessingSpec.workaroundConfiguration(config: Config): Config = wi
 
     return declaredConfig ?: getDefaultConfiguration()
 }
+
+internal fun loadDeprecatedRuleIds(): Set<String> {
+    return loadDeprecations().keys
```
cortinico, line 213:
```suggestion
internal fun loadDeprecatedRuleIds(): Set<String> =
    loadDeprecations().keys
```

cortinico (approved):
Works great for me 👍 Let's merge this so I can prepare `.2`

marschwar:
Fix for 2.0 is #6433.

BraisGabin:
We can merge this, right?

cortinico:
> We can merge this, right?

Yup 👍

PR #6390, opened 2023-08-14
Fix reporting unused imports at file level

### LINKED ISSUE #6105: ktlint violation location in detekt-formatting is broader
## Expected Behavior
Highlight the exact location ktlint reports (│):
```kotlin
private fun getErrorMessage(value: String): String =
        @Suppress("DEPRECATION")
        customMessage.takeUnless { it.isEmpty() }│ ?: String.format(DEFAULT_ERROR_MESSAGE_WITH_NO_REASON, value)
```

## Current Behavior
image

## Context
While working on a PR we found a ktlint violation in the code that was confusing (see above in current screenshot). This was reported at https://github.com/pinterest/ktlint/issues/2028, and investigated and explained there. The reason it was confusing was because the highlighted code (yellow in above screenshot) didn't match up with the error message "Expected newline". The annotation looked correctly wrapped, but the same rule also checks if the annotated element is nicely separated, hence this emission.

In the explanation there was a line that hit me "This LintError contains the offset (line and position)", and a short investigation lead me to
https://github.com/detekt/detekt/blob/948276a8db4578b962bf24c63eee3b8548bac627/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt#L97

showing that the offset reported by the ktlint `emit` function is ignored (`_`). This means that **detekt-wrapped ktlint rules will always report less precise location** than if they were directly executed with another wrapper.

This is just one example, there could be many others where the detekt-wrapped reporting is confusing. I understand this is quite a fundamental issue, and might not fit detekt's model, but wanted to record the problem anyway.

arturbosch:
This PR was triggered by users of the IntelliJ plugin getting every unused import reported at the file level which clutters the editor: https://github.com/detekt/detekt-intellij-plugin/issues/486#issuecomment-1676916223

Screenshot from 2023-08-14 19-57-14

This PR partly reverts the change to the emit logic of https://github.com/detekt/detekt/pull/5876/files#diff-9cf77f9ffcbf036bf5044d1ef99728b56aad97854246f3473f69d54b4b929380.
Emiting at the node is wrong in the case of the NoUnusedImports rule as in `beforeVisitChildNodes` the import usages are calculated and only reported in `afterVisitChildNodes` from the file node (because `fileNode.imports` access is easy I suppose). 

With this revert the lines are again calculated with the offset function and lead to the same results as our MaxLineLength rule:
Screenshot from 2023-08-14 20-16-34

The other changed test cases are also now correct again e,g, the tested code for indentation rule is
`val code = "fun main() {\n println()\n}"` so `.hasStartSourceLocation(1, 13)` could not be correct due to the line breaks.

Fixes #6105

### DIFF
--- a/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
+++ b/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
@@ -5,15 +5,19 @@ import com.pinterest.ktlint.rule.engine.core.api.editorconfig.CODE_STYLE_PROPERT
 import com.pinterest.ktlint.rule.engine.core.api.editorconfig.EditorConfig
 import com.pinterest.ktlint.rule.engine.core.api.editorconfig.EditorConfigProperty
 import com.pinterest.ktlint.rule.engine.core.api.editorconfig.INDENT_STYLE_PROPERTY
+import io.github.detekt.psi.toFilePath
 import io.gitlab.arturbosch.detekt.api.CodeSmell
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.CorrectableCodeSmell
 import io.gitlab.arturbosch.detekt.api.Debt
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Location
 import io.gitlab.arturbosch.detekt.api.Rule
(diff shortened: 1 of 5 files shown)


schalkms (approved):

BraisGabin (approved):

cortinico (approved):
Thanks for fixing this @arturbosch
Do we want to backport this to 1.x?

arturbosch:
> Thanks for fixing this @arturbosch Do we want to backport this to 1.x?

Yes please, this is a regression we introduced in 1.23.0 so I suggest a fix in 1.23.2 :)

PR #6394, opened 2023-08-14
Remove deprecated print function in ConsoleReport

marschwar:
This removes the deprecated `print` function from the `ConsoleReport` class


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/ConsoleReport.kt
```diff
@@ -11,18 +9,6 @@ import java.io.PrintStream
  */
 abstract class ConsoleReport : Extension {
```
marschwar, line 10: Should we change this to an interface?

schalkms: Can we remove this abstract class completely?

schalkms (approved):

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/ConsoleReport.kt
marschwar: I don't think so. This is one of the  [public extension points to create custom reports](https://detekt.dev/docs/introduction/extensions#custom-reports).

schalkms: Yeah. We have the same `render()` function in both reports now.
I don't think the design for the reports fits that well now.

PR #6399, opened 2023-08-16
Remove `Severity` enum

marschwar:
This is part of #6400 and removes the old `Severity` class. Removing the current enum was surprisingly easy. 

## Changes
* severity removed form `Issue`
* Removed severity from documentation generation
* Removed severity from custom rule example in the documentation
* Removed the `Severity` enum  

## Review Notes

Sorry for the big change set, but all rules have to be touched for this. I think it is possible to skip all changes in `detekt-rules-*` and `detekt-formatting`. There I only removed the severity from the `Issue` constructor.


BraisGabin (approved):
<3

PR #6412, opened 2023-08-19
Add warning for versioned docs update on Danger

### LINKED ISSUE #6228: Configure danger to help the PRs that only update the versioned documentation
## Expected Behavior
When someone does a change in the versioned documentation they should also change the not versioned one. 

## Current Behavior
There is no way a contributor could know that. 

## Proposed solution
Configure danger so it adds a comment to inform about this with a proper explanation. A link to which file should be also modified would be great.

I'm not sure if possible but, if we could try to apply the diff to the not versioned one and it does not have conflicts add it as a suggestion would be really neat! This way a maintainer could just click to merge it and close the issue faster.

cortinico:
Fixes #6228

### DIFF
--- a/bots/dangerfile.js
+++ b/bots/dangerfile.js
@@ -15,6 +15,12 @@ const ruleTestChanges = danger.git.fileMatch(
 const detektConfigFileChanges = danger.git.fileMatch(
   "**/default-detekt-config.yml"
 );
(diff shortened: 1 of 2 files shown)


BraisGabin (approved):
Thanks!

schalkms (approved):
👍

Issue #6400, opened 2023-08-16
Implement a new concept for severity


marschwar:
## Expected Behavior

In #6383 we discussed what the severity model for detekt 2.0 should look like. When this issue is resolved, all of the properties below are implemented:

- [x]     The current values of the severity enum is replaced with the severity levels error, warning and info.
- [x]     The default severity of an all rules is error.
- [x]     The severity of a rule instance must be configurable via configuration file.
- [x]     It should be possible to configure to the severity at rule set level to apply it to all rules of the rule set (unless overwritten at rule level)
- [x]     The severity will be moved from the Issue to the Finding as it depends on a configurable rule instance.
- [x]     The concept of maxIssues will be removed. The build fails if (and only if) there is at least one issue with the severity error. #6408 
- [x]     It must be possible to reduce the level that has the build fail. Something similar to warningsAsError but also allowing the threshold to be lowered to info.
- [x]     The concept of weighted issues (build > weights) will be removed. #6389 
- [x]     The SeverityLevel will be removed.

## Current Behavior
The current severity model is confusing because the name suggests something that it really is not.

3flex:
Is this complete now? The last checkbox looks like it can be ticked off as the linked PR is merged.

marschwar:
Yes. Thanks for the reminder

jinia91:
Hi

This feature has been merged into the main branch and has been confirmed to work properly when run on main.

What is the reason it has not yet been released to the release branch?

Is it planned for a version after 2.0?

PR #6417, opened 2023-08-21
Allow the failure severity to be configured

marschwar:
This is supposed to be the final PR for #6400 introducing a new severity concept. It allows users to configure the minimum severity that leads to build failures


detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/spec/RulesSpec.kt
```diff
@@ -28,7 +29,17 @@ interface RulesSpec {
         /**
          * No issues with a severity of error is allowed.
          */
-        data object FailOnError : FailurePolicy()
+        data object DefaultFailurePolicy : FailurePolicy()
```
BraisGabin, line 32: I'm not sure about this one. Should the default value be handled by the core or by the clients? I would vote for the clients. I know that it means duplicate code on our cli and gradle plugin but we keep the flexibility for any other third party client. This case is kind of trivial but I think that our core shouldn't decide the default behaviours.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -16,6 +16,10 @@ open class DetektExtension @Inject constructor(objects: ObjectFactory) : CodeQua
         toolVersion = loadDetektVersion(DetektExtension::class.java.classLoader)
     }
 
+    // TODO: We need to decide what to do with this property.
+    //   Currently we are using it to ignore build failures
+    //   which can also be done with 'failOnSeverity=never'
+    //   But this property is inherited from the CodeQualityExtension
```
BraisGabin, line 22: If it sis inherit I think that we should honor it. But I would remove it from our documentation. If someone sets it to true this will override the other property to value never and that's it.

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/spec/RulesSpec.kt
marschwar: First of all, thank you for looking into this.

> Should the default value be handled by the core or by the clients?

That is exactly the question I was asking myself as well. I am not worried so much about code duplication but rather a consistent behavior across clients. But while implementing this, it already felt kind of strange and I agree now that we should have the clients define the default. That also helps with documentation, since users will not look at the core.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -86,6 +87,14 @@ class CliArgs {
     )
     private var reports: List<String>? = null
 
+    @Parameter(
+        names = ["--fail-on-severity"],
+        description = "Specifies the minimum severity that causes the build to fail. " +
+            "When the value is set to 'NEVER' detekt will not fail regardless of the number " +
+            "of issues and their severities."
+    )
+    var failOnSeverity: FailureSeverity? = null
```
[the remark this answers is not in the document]
marschwar: When the default is removed from the core, it will be set here.  Good point.

marschwar:
@detekt/maintainers: 
Before I finish this up with tests and documentation I would like your take on the `ignoreFailures` parameter in the `DetektExtension` (see https://github.com/detekt/detekt/pull/6417/files#diff-76e96187fb9c234b865786caa99de176c851268f7a284a2fd3d79402dd426f27) It used to be declared by the `CodeQualityExtension` which we no longer extend. This property would collide with the newly introduced `failOnSeverity` parameter. I see 2 options here:

* Keep both parameters and set failOnSeverity to `never` if ignoreFailures is set to true. Then we should also validate that the failOnSeverity parameter is not set. 
* Replace  `ignoreFailures`  with `failOnSeverity`. I do not know if ignoreFailures is some sort of community standard and should be available.

schalkms:
I vote for option 2. It seems to be the better long term option with lower maintenance.

3flex:
There's a third option which is my preference:
* Keep both parameters, use `failOnSeverity` to set the severity level that is considered a failure. If there are any issues raised at this level it won't fail the task/build if `ignoreFailures` is `true`

They're two different flags with two different features. The only reason we no longer extend `CodeQualityExtension` was to unblock https://github.com/detekt/detekt/pull/6496 - it wasn't because the properties in `CodeQualityExtension` were no longer relevant.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -111,6 +113,10 @@ abstract class Detekt @Inject constructor(
     @get:Input
     abstract val ignoreFailures: Property<Boolean>
 
+    @get:Input
+    @get:Optional
+    abstract val failOnSeverity: Property<String>
```
3flex, line 118: This would be better as an enum than a string to make it simpler to choose valid values.

marschwar:
> There's a third option which is my preference:
> 
>     * Keep both parameters, use `failOnSeverity` to set the severity level that is considered a failure. If there are any issues raised at this level it won't fail the task/build if `ignoreFailures` is `true`

That is also an option that is easy to implement but what is a use case? Why would a user for example specify `failOnSeverity=error` and `ignoreFailures=true`?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
marschwar: The goal was to make the configuration case insensitive but I guess this is only really relevant for the groovy dsl. I will introduce an enum.

3flex: It should be fine. It's mentioned as an example here: https://docs.gradle.org/current/userguide/custom_tasks.html#sec:supported_task_option_data_types

3flex:
True that if there's a "never" option for the new property there wouldn't normally be a reason, though it makes it a little simpler to hook up a way to toggle failures on and off if there's a boolean property available.

> I do not know if ignoreFailures is some sort of community standard and should be available.

It's available on all Gradle code quality tasks, test tasks, and the validate plugins task, so it's standard for every task in Gradle core that performs verification.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -86,6 +89,14 @@ class CliArgs {
     )
     private var reports: List<String>? = null
 
+    @Parameter(
+        names = ["--fail-on-severity"],
+        description = "Specifies the minimum severity that causes the build to fail. " +
+            "When the value is set to 'NEVER' detekt will not fail regardless of the number " +
```
BraisGabin, line 95:
```suggestion
            "When the value is set to `never` detekt will not fail regardless of the number " +
```

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/FailOnSeverity.kt
```diff
@@ -0,0 +1,8 @@
+package io.gitlab.arturbosch.detekt.extensions
+
+enum class FailOnSeverity {
+    ERROR,
+    WARNING,
+    INFO,
+    NEVER
```
BraisGabin, line 7: Nit: what do you think about `Never` instead? (same for the others). I usually don't like the snake case for enums. It feels too Java to me.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
```diff
@@ -80,6 +83,14 @@ internal data class BasePathArgument(val basePath: String?) : CliArgument() {
     override fun toArgument() = basePath?.let { listOf(BASE_PATH_PARAMETER, it) }.orEmpty()
 }
 
+internal data class FailOnSeverityArgument(val ignoreFailures: Boolean, val minSeverity: FailOnSeverity) :
+    CliArgument() {
+    override fun toArgument(): List<String> {
+        val effectiveSeverity = if (ignoreFailures) FailOnSeverity.NEVER else minSeverity
+        return listOf(FAIL_ON_SEVERITY_PARAMETER, effectiveSeverity.name.toLowerCase(Locale.ROOT))
```
BraisGabin, line 90:
```suggestion
        return listOf(FAIL_ON_SEVERITY_PARAMETER, effectiveSeverity.name.lowercase())
```
🤔 Isn't this the not deprecated one?

TWiStErRob:
> what is a use case? Why would a user for example specify failOnSeverity=error and ignoreFailures=true?

@marschwar Imagine that the code checked into the repo is error+false, as a default for most developers + CI. But some developers might want to toggle ignoreFailures=true locally, because they have a different process, they could write experimental code, make it work, then clean it up.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
marschwar: Yes it is. Maybe you can help me on this. For some reason the `lowercase` extension function is not available in the detekt-cli main source set. It is available in the test source set though.

marschwar:
> > what is a use case? Why would a user for example specify failOnSeverity=error and ignoreFailures=true?
> 
> @marschwar Imagine that the code checked into the repo is error+false, as a default for most developers + CI. But some developers might want to toggle ignoreFailures=true locally, because they have a different process, they could write experimental code, make it work, then clean it up.

But they would still change the gradle build file. I do not see why changing ignoreFailues to true is any easier than changing failOnSeverity to never. 
But I do not want to argue against the approach. I was just curious if I had been missing something. The question I was asking myself was if it helps the users or will it confuse them to have both.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/FailOnSeverity.kt
marschwar: Fair point. I have been doing way too much java lately.

marschwar: Fun fact: JCommander assumes all-caps enum values. 
https://github.com/cbeust/jcommander/blob/d7ec08c66ee5569089161725f433c7e96ec9403b/src/main/java/com/beust/jcommander/converters/EnumConverter.java#L29-L41

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
BraisGabin: I assume that the gradle plugin code uses an older version of kotlin. Nothing to do here then.

BraisGabin (approved):
LGTM great job :)

BraisGabin:
@marschwar is there something blocking this to be merged? What can we do to make this happen?

marschwar:
Thank you @BraisGabin for the reminder. I did just resolve the merge conflicts. I did not plan to make any more additions to this PR. To me it is good to go.

BraisGabin:
Then, let's merge it! It improves the project :)

PR #6421, opened 2023-08-22
Ignore complexity of local function when calculating cyclomatic complexity by default

### LINKED ISSUE #5344: ComplexMethod: exclude local functions
## Expected Behavior
Complexity of `f()` below is 2.

Maybe behind a toggle? By default?

## Observed Behavior
The function f appears to be too complex (6). Defined complexity threshold for methods is set to '3'

## Steps to Reproduce
```kotlin
    @Test fun f() {
        assertThat(ComplexMethod(TestConfig(mapOf("threshold" to 3))).compileAndLint(
            """
                fun f(): String {
                    fun local1(flag: Boolean) = if (flag) "a" else "b"
                    fun local2(flag: Boolean) = if (flag) "c" else "d"
                    if (Math.random() < 0.5) {
                        return local1(true) + local2(false)
                    } else {
                        return local2(true)
                    }
                }
            """.trimIndent()
        )).isEmpty()
    }
```

## Context
There are legitimate cases for defining local functions, this rule discourages any of those because it reports them as ComplexMethod, by default, this means that those local functions be it however irrelevant in the surrounding class, file, or package, are forced to be one level higher than necessary so they are littering global or local namespace.

@arturbosch put it very nicely:
> Hm, one can argue that local functions help to declutter and structurize your code, removing nestings etc removes complexity. Like an algorithm which implementation is a whole function with many little helper local functions. I would't call this code complex but rather an architectural choice to build it this way :)? -- https://github.com/detekt/detekt/pull/1365#issuecomment-442138661

[Here's an example](https://github.com/TWiStErRob/net.twisterrob.gradle/blob/956016bc8bf5d282c6eb91b8ba64ff199d7a7f76/quality/src/main/kotlin/net/twisterrob/gradle/quality/report/html/ViolationsCollapser.kt#L21-L84), where I argue that `verySimilarProblem` and `merge` are two extracted local functions that make the algorithm below way more readable. I could lift them up to file level as private functions, but at that point I would actually open them up for usage from other functions in the file, even though they're strict implementation details of the `collapseFile` function. Regarding complexity, I don't think those local functions count towards the `collapseFile` complexity in understanding, because they're pure(ish, is new-ing an object pure?). Maybe each invocation of a local function could add 1 complexity?

## Your Environment

* Link to your project (if it's a public repository): https://github.com/TWiStErRob/net.twisterrob.gradle/blob/956016bc8bf5d282c6eb91b8ba64ff199d7a7f76/quality/src/main/kotlin/net/twisterrob/gradle/quality/report/html/ViolationsCollapser.kt#L21-L84

ErdoganSeref:
The change fixes #5344.

When calculating the cyclomatic complexity, the flag `ignoreLocalFunctions` decides how a local function increases the complexity of the parent function.

If the flag is set to `true`, then the complexity of the parent function increases by one for each local function. Otherwise, the complexity of the parent function increases with the complexity of each local function.

### DIFF
--- a/detekt-metrics/src/main/kotlin/io/github/detekt/metrics/CyclomaticComplexity.kt
+++ b/detekt-metrics/src/main/kotlin/io/github/detekt/metrics/CyclomaticComplexity.kt
@@ -27,6 +27,7 @@ class CyclomaticComplexity(private val config: Config) : DetektVisitor() {
     class Config(
         var ignoreSimpleWhenEntries: Boolean = false,
         var ignoreNestingFunctions: Boolean = false,
+        var ignoreLocalFunctions: Boolean = false,
         var nestingFunctions: Set<String> = DEFAULT_NESTING_FUNCTIONS
     )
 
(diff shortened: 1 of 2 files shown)


cortinico:
Can we get the CI green before we can merge this?

BraisGabin (approved):
Thanks!

ErdoganSeref:
How can we update the documentation on the website?

BraisGabin:
The documentation is extracted from the kdoc on the rule. So you just need to update it.

PR #6424, opened 2023-08-23
Fix build issues

3flex:
Fix build issues with the Gradle plugin by:
- Ensuring artifacts are published to Maven Local before running functional tests
- Moving `mavenLocal` to the first position in the `repositories` block in build files in the functional tests
- Fixing issues that weren't caught when CI was run on recent PRs due to the faulty setup

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -37,11 +37,17 @@ jobs:
           java-version: ${{ matrix.jdk }}
           distribution: 'temurin'
 
+      - name: Assemble and publish artifacts to Maven Local
+        uses: gradle/gradle-build-action@243af859f8ca30903d9d7f7936897ca0358ba691 # v2
+        with:
+          gradle-home-cache-cleanup: true
+          arguments: publishToMavenLocal
+
       - name: Build detekt
         uses: gradle/gradle-build-action@243af859f8ca30903d9d7f7936897ca0358ba691 # v2
         with:
           gradle-home-cache-cleanup: true
-          arguments: build -x detekt publishToMavenLocal
+          arguments: build -x detekt
 
       - uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3
         with:
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -21,6 +21,7 @@ class DetektPlugin : Plugin<Project> {
             )
 
         extension.reportsDir = project.extensions.getByType(ReportingExtension::class.java).file("detekt")
+        extension.basePath = project.layout.projectDirectory.asFile.absolutePath
 
         val defaultConfigFile =
             project.file("${project.rootProject.layout.projectDirectory.dir(CONFIG_DIR_NAME)}/$CONFIG_FILE")
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -21,7 +21,7 @@ class DetektPlugin : Plugin<Project> {
             )
 
         extension.reportsDir = project.extensions.getByType(ReportingExtension::class.java).file("detekt")
-        extension.basePath = project.layout.projectDirectory.asFile.absolutePath
+        extension.basePath = project.rootProject.layout.projectDirectory.asFile.absolutePath
 
         val defaultConfigFile =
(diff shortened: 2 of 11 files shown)


marschwar:
I was going to ask this in my next PR but I guess it fits here as well. How should I handle changes that will affect the functional tests (such as the removal of the weights and maxIssues). As far as I understood the functional tests use the latest version of detekt. I did change the tests in the PR but then the build failed.

3flex:
> As far as I understood the functional tests use the latest version of detekt

They should, but they're not in all cases for some reason. I have a POC fix in #6415 which is ugly but works and might be the way forward. We clearly need a reliable solution.

3flex:
@detekt/maintainers please review. CI should be green going forward, and this should also catch missed updates to Gradle plugin functional test updates in future PRs.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -21,6 +21,7 @@ class DetektPlugin : Plugin<Project> {
             )
 
         extension.reportsDir = project.extensions.getByType(ReportingExtension::class.java).file("detekt")
+        extension.basePath = project.layout.projectDirectory.asFile.absolutePath
```
chao2zhang, line 24: This is an actual fix for recent PRs, right?

chao2zhang (approved):

.github/workflows/pre-merge.yaml
```diff
@@ -37,11 +37,17 @@ jobs:
           java-version: ${{ matrix.jdk }}
           distribution: 'temurin'
 
+      - name: Assemble and publish artifacts to Maven Local
+        uses: gradle/gradle-build-action@243af859f8ca30903d9d7f7936897ca0358ba691 # v2
+        with:
+          gradle-home-cache-cleanup: true
+          arguments: publishToMavenLocal
```
BraisGabin, line 44: Is this better than make the `functionalTest` depend on `publishToMavebLocal`?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
3flex: It's a fix, but only for tests on Windows.

When TestKit executes tests it starts the daemon from the Gradle plugin project directory but the build under test is in a different directory. That's ok building locally in most cases but on GitHub runner one of those directories is on C: but the other is on D:

There's no way to get a relative file path from one drive to another, which causes build failures. The base path is determined based on a compiled class's resource path which is on one drive, but the absolute path to the file checked by detekt is on another drive.

The CLI defaults base path to `user.dir` so adding this line makes it more correct when running from Gradle. Technically it should be rootDir to match current behaviour but still fix the build... Can change to projectDir in a different PR if needed.

.github/workflows/pre-merge.yaml
3flex: You're right, but I couldn't find a way to do that while the plugin is an included build, so this seemed the next best option.

3flex:
I'll go ahead and merge to unblock other PRs.

.github/workflows/pre-merge.yaml
TWiStErRob: Could a similar switcheroo be done here? https://github.com/cashapp/paparazzi/pull/924

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
TWiStErRob: Will the absolute path break cache relocation in Gradle when executed on CI vs Local builds on user's setups?

.github/workflows/pre-merge.yaml
3flex: I didn't know there was prior art but there's this draft PR open already to do the same: https://github.com/detekt/detekt/pull/6415

I might look at what they did in cashapp org since they often end up setting unofficial standard ways of doing things.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
3flex: Hmm yes it will due to the usage of `@Input` on that property, I did check the annotation on the task but didn't think of the consequences. I'll submit a PR to fix.

TWiStErRob: Here's a potentially interesting way to validate this (never tried it myself, just read it somewhere):
https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/main/Gradle.md#structure
could run on a fully standalone integration test ([like this](https://github.com/TWiStErRob/net.twisterrob.gradle/tree/main/docs/examples/local)) after the snapshot is published to maven local similar to what you did in this PR?

Issue #6401, opened 2023-08-16
Adjusting MatchingDeclarationName rule for multiplatform expect/actual containing files


mustafaozhan:
Hello 👋 

I saw that same topic covered already with this issue: https://github.com/detekt/detekt/issues/5678

But I just saw a piece from the official kotlin coding convention page which is newly added, and it looks they are suggesting quite the opposite.

```
Multiplatform projects
In multiplatform projects, files with top-level declarations in platform-specific source sets should have a suffix associated with the name of the source set. For example:

jvmMain/kotlin/Platform.jvm.kt

androidMain/kotlin/Platform.android.kt

iosMain/kotlin/Platform.ios.kt

As for the common source set, files with top-level declarations should not have a suffix. For example, commonMain/kotlin/Platform.kt.
```

Link: https://kotlinlang.org/docs/coding-conventions.html#multiplatform-projects

cortinico:
Implemented in #6426

PR #6426, opened 2023-08-24
MatchingDeclarationName now supports platofrm suffixes

lexa-diky:
This is one way of implementing #6401.

We can also consider removing default platform suffixes including 'common' which will break backward compatibility somewhat, but will make this rule much more flexible.


detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MatchingDeclarationName.kt
```diff
@@ -59,6 +59,9 @@ class MatchingDeclarationName(config: Config = Config.empty) : Rule(config) {
     @Configuration("name should only be checked if the file starts with a class or object")
     private val mustBeFirst: Boolean by config(true)
 
+    @Configuration("kotlin multiplatform target file suffixes")
+    private val multiplatformTargetFileSuffixes: List<String> by config(COMMON_KOTLIN_KMP_PLATFORM_TARGET_SUFFIXES)
```
lexa-diky, line 63: maybe `multiplatformTargets`? `multiplatformTargetFileSuffixes` is super explicit, but also confusing whether consumer should put `.jvm.kt` or just `jvm`

BraisGabin: Agree, multiplatformTarget is a better name.

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/MatchingDeclarationName.kt
```diff
@@ -89,4 +92,20 @@ class MatchingDeclarationName(config: Config = Config.empty) : Rule(config) {
...
+            "ios",
+            "android",
+            "js",
+            "jvm",
+            "native",
+            "iosArm64",
+            "iosX64",
+            "macosX64",
+            "mingwX64",
+            "linuxX64"
```
BraisGabin, line 108: These are subtipes of the other ones, right? Would it have sense to remove them from the default configuration?


```suggestion
```

lexa-diky: They are still pretty common, for example IJ has them in template. I suggest leaving them in

BraisGabin (approved):
Thanks!!

PR #6431, opened 2023-08-26
Improve UnnecessaryFilter

### LINKED ISSUE #6429: Improve `UnnecessaryFilter`
Code like this should be flagged by `UnnecessaryFilter`: 
```kotlin
val smells = filterAutoCorrectedIssues(config)
  .flatMap { it.value }
  .filter { it.severity == Severity.ERROR }
return smells.count()
```

As it is the same as 

```kotlin
return filterAutoCorrectedIssues(config)
  .flatMap { it.value }
  .count { it.severity == Severity.ERROR } 
```

## Context
We had this issue on our own code: #6409.

t-kameyama:
Fixes #6429


marschwar:
Thank you for looking into this. Would it be possible to also find cases where the variable is only referenced for the `count()` without a direct return such as this? 

```kotlin
fun test(list: List<Int>): Int {
    val x = list.map { it + 1 }.filter { it > 2 }
    val count = x.count()
    return count + " items"
}
```

t-kameyama:
> Thank you for looking into this. Would it be possible to also find cases where the variable is only referenced for the `count()` without a direct return such as this?
> 
> ```kotlin
> fun test(list: List<Int>): Int {
>     val x = list.map { it + 1 }.filter { it > 2 }
>     val count = x.count()
>     return count + " items"
> }
> ```

https://github.com/detekt/detekt/compare/8327de1b5d371fd04edf9c41da3755bcf1f4539b..de341ccab892b9c1afbad09b1b8dca81f07984f9

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryFilter.kt
```diff
@@ -92,8 +138,8 @@ class UnnecessaryFilter(config: Config = Config.empty) : Rule(config) {
 
         private val secondCalls = listOf(
             SecondCall(FqName("kotlin.collections.List.size"), "count"),
-            SecondCall(FqName("kotlin.collections.List.isEmpty"), "any"),
-            SecondCall(FqName("kotlin.collections.isNotEmpty"), "none"),
+            SecondCall(FqName("kotlin.collections.List.isEmpty"), "none"),
+            SecondCall(FqName("kotlin.collections.isNotEmpty"), "any"),
```
marschwar, line 142: Very good catch :clap:

marschwar (approved):

BraisGabin (approved):

BraisGabin:
@t-kameyama could you check the tests? It seems that the snippets some compile.

PR #6449, opened 2023-09-02
Remove `SingleAssign`

BraisGabin:
Remove `SingleAssign` this class was introduced before kotlin 1.2 because `lateinit` didn't exist but now it is not needed anymore.

This is an effort to reduce the code inside `detekt-api`.


marschwar:
I am not sure about this one because `SingelAssign` also guarantees that the value is written exactly once. I don't know that that feature is particularly useful or necessary within detekt though. I am not against removing it, but wanted to point out that the semantics are different.

BraisGabin:
Yes, they aren't exactly the same but nearly all the usages are in `Extension`s and the properties are set on the `init`. And we are the ones that call that method and we know that we call it just once.

The exception is `detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt`. That uses it on a `visit(KtFile)` that also should be only called once.

Anyway, I agree that this function is safer than `lateinit var`. I thought it wasn't worth the effort to maintain it. But if we want to keep it I would vote to move it outside our public api. But, where? `detekt-utils`?

3flex (approved):
Agree this should be removed. If `init` is called multiple times that effectively breaks the contract of the extension point so that would be a bug that should be resolved if it occurs - consumers shouldn't need to defend against that in their extensions.

PR #6450, opened 2023-09-02
Move `SimpleNotification` to core

BraisGabin:
`SimpleNotification` was at `api-internal` but we don't need it there. It is only used at `core`. For that reason I'm moving it.


chao2zhang (approved):

marschwar (approved):
:+1:

PR #6451, opened 2023-09-02
Move `RuleSetConfigProperty` to formatting

BraisGabin:
These class was at `api-internal` and it was only used by `formatting`. So I moved it to `formatting`


marschwar (approved):
Very nice. This should make it easier to move the formatting ruleset to its new home (hopefully) soon.

PR #6453, opened 2023-09-02
Remove `Metric`s from `Finding`s

BraisGabin:
This code is never used so it can be removed.


marschwar (approved):
Let's wait with this until we have clarity about metrics in #6452

BraisGabin:
I think that this PR is not related with that. To be honest, I created that thread before these set of PRs and I thought that `Metric` and `ProjectMetric` were related somehow. But they are not.

To me, #6452 is about if we should merge #6457.

3flex:

marschwar:
> ... and I thought that `Metric` and `ProjectMetric` were related somehow. But they are not.

That is confusing. Thank you for pointing it out. Then I am good with removing this.

PR #6454, opened 2023-09-02
Remove `type` from `Metric`

BraisGabin:
The value "type" was never used on `Metric` so we can remove it.


3flex (approved):

3flex:
Tagging "breaking" since it impacts public API.

PR #6455, opened 2023-09-02
Simplify `ThresholdedCodeSmell`

BraisGabin:
These values aren't necessary.


marschwar (commented):
I don't quite understand this one. If we remove the only properties this type introduces, should we not remove it instead? Or maybe rename it to `SingleMetricCodeSmell`.

BraisGabin:
No, The property that this type introduces is `val metric: Metric`. And that's what we use on the tests. The properties that I remove are just sintax sugar that only expands the `api`. If someone needs them can create extension properties.

marschwar:
Ok. I still think the name is a little confusing.

marschwar (approved):

BraisGabin:
We can rename it in other PR. Also if we end up merging this set of PRs this class will be an implementation detail inside `detekt-rules-complexity` instead of part of our public API so it will be a big win.

chao2zhang (approved):

PR #6456, opened 2023-09-02
Move `ThresholdedCodeSmell` to complexity

BraisGabin:
These type of CodeSmell is only used on `detekt-complexity` so we can move it and `Metric` to that subproject.


3flex:

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt
```diff
@@ -84,7 +83,4 @@ private fun Finding.truncatedMessage(): String {
     }
 }
 
-private fun Finding.detailed(): String = when (this) {
-    is ThresholdedCodeSmell -> "$id - $metric - [${truncatedMessage()}] at ${location.compact()}"
-    else -> "$id - [${truncatedMessage()}] at ${location.compact()}"
-}
+private fun Finding.detailed(): String = "$id - [${truncatedMessage()}] at ${location.compact()}"
```
BraisGabin, line 86: This PR includes this behaviour change that I think it is acceptable. But I would like feedback about it.

marschwar (approved):

PR #6460, opened 2023-09-03
Remove `Debt`

BraisGabin:
Right now the concept of `Debt` is only used to show it on the docuentation (second commit) and to show it in two different console reports (third commit). We don't even show `Debt` values in the html report. For that reason I think that we should remove it.

If we don't want to remove it I think that we should give it a bit of love:
- #4359 
- Show it on html report (at least)
- Review if the debt values have sense for each rule

And I don't see us doing that. For that reason I vote to remove it. Also, if someone wants to keep this functinality (or even improve it) that's doable with our extension points, so any third party could implement it if they see value on that.

I recommend to review this PR by commits so you can see exactly what "features" I'm removing.


detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
```diff
@@ -132,35 +125,6 @@ internal class RuleVisitor(textReplacements: Map<String, String>) : DetektVisito
         }
     }
 
-    private fun extractIssueSeverityAndDebt(klass: KtClass) {
```
chao2zhang, line 135: Just confirming that this method is only used to extract Debt, not Severity.

chao2zhang (approved):

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
BraisGabin: I was wondering the same. Even with that name this function wasn't reading the severity.

3flex:

cortinico (approved):
> And I don't see us doing that. For that reason I vote to remove it

+1 for removing it

BraisGabin:
Rebased

3flex (approved):

PR #6462, opened 2023-09-03
Setup consistent configuration for modules with public APIs

3flex:
This PR:
* Creates a new convention plugin for modules with a public API
* Configures Dokka to work correctly with a multi module build
* Generates documentation for all modules with public APIs


BraisGabin (approved):
Nice one :)

PR #6463, opened 2023-09-04
Remove safeAs function

3flex:
There's an almost identical function in `org.jetbrains.kotlin.utils.addToStdlib` which comes with this warning:

        Usage of this function is unsafe because it does not have native compiler support
         This means that compiler won't report UNCHECKED_CAST, CAST_NEVER_SUCCEED or similar
         diagnostics in case of error cast (which can happen immediately or after some
         refactoring of class hierarchy)
        Consider using regular `as` and `as?`

https://github.com/JetBrains/kotlin/blob/v1.9.10/core/util.runtime/src/org/jetbrains/kotlin/utils/addToStdlib.kt#L94-L108


3flex:
I've tagged as breaking change as it technically is if anyone happens to use this function in their own code, though I suspect it's not used often.

BraisGabin (approved):

PR #6466, opened 2023-09-04
⚰️ Remove Complexity deprecated properties

LeoColman:
This commit looks at all providers in ComplexityProvider in addition to deprecation.properties to find the only remaining deprecated rule: `threshhold`.

Removing the property seems to be all that is needed. Test code is already adapted to the new version of properties.

Some adaptations were necessary in classes that still used old `ComplexMethod` property, but it wasn't used anymore.

Reference: https://github.com/detekt/detekt/pull/6193 Reference: https://github.com/detekt/detekt/discussions/6035#discussioncomment-5702474

### DIFF
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CyclomaticComplexMethod.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CyclomaticComplexMethod.kt
@@ -47,7 +47,7 @@ class CyclomaticComplexMethod(config: Config = Config.empty) : Rule(config) {
         Debt.TWENTY_MINS
     )
 
-    override val defaultRuleIdAliases: Set<String> = setOf("ComplexMethod")
+    override val defaultRuleIdAliases: Set<String> = emptySet()
 
     @Configuration("The maximum allowed McCabe's Cyclomatic Complexity (MCC) for a method.")
     private val allowedComplexity: Int by config(defaultValue = 14)
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/RunnerSpec.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/RunnerSpec.kt
@@ -11,6 +11,7 @@ import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.assertThatCode
 import org.assertj.core.api.Assertions.assertThatThrownBy
 import org.junit.jupiter.api.BeforeEach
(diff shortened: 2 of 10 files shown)


LeoColman:
And it's already conflicted with main branch.

I'm not sure how to proceed guys, I think deprecations are too unstable to be removed at the moment. What do you think @schalkms @BraisGabin

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CyclomaticComplexMethod.kt
```diff
@@ -47,7 +47,7 @@ class CyclomaticComplexMethod(config: Config = Config.empty) : Rule(config) {
         Debt.TWENTY_MINS
     )
 
-    override val defaultRuleIdAliases: Set<String> = setOf("ComplexMethod")
+    override val defaultRuleIdAliases: Set<String> = emptySet()
```
marschwar, line 50: This can be remove altogether as it is the default.

marschwar (approved):

schalkms:
> And it's already conflicted with main branch.
> 
> I'm not sure how to proceed guys, I think deprecations are too unstable to be removed at the moment. What do you think @schalkms @BraisGabin

Please excuse the late reply.
I'm not sure what you mean by too unstable and what brings you to this conclusion.

LeoColman:
I mean that these files are updated too often to "remove all deprecated properties". 
If we look at PR https://github.com/detekt/detekt/pull/6193 history, we can see that it got conflicted with master (and rebased) many times.

We decided to make simpler changes, but my first simple change is already conflicting with main.

My point is if we should take a different approach, as deprecations are currently too sensitive to be touched, as they'll conflict with many other commits

schalkms:
> I mean that these files are updated too often to "remove all deprecated properties". If we look at PR #6193 history, we can see that it got conflicted with master (and rebased) many times.
> 
> We decided to make simpler changes, but my first simple change is already conflicting with main.
> 
> My point is if we should take a different approach, as deprecations are currently too sensitive to be touched, as they'll conflict with many other commits

I see. I think we could lock the main branch for some time in order to clean up deprecated properties. What do you think?

LeoColman:
I think that could work well. We can also push 'removing deprecations' to the very end of 2.0 migration. What we've tried so far is do both big changes and code removal at once

cortinico:
@LeoColman can we drive this forward?

LeoColman:
I'm still up to the contribution. 
Sorry, I failed to get the notification of your message.

I think the work on this PR is quite stale, I'd like to open new ones.
Could we set a path of 'deliverables' on the discussions thread for this? I think I'm doing some rework and would like to avoid it happening again.

For now, I believe closing this makes more sense than leaving it open

PR #6467, opened 2023-09-04
Remove knownPatterns from tooling API

3flex:
This is not used anywhere in detekt so can be removed from the API.

This is a breaking change as this alters the API but functionally there is no impact.

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
@@ -36,7 +36,6 @@ internal fun CliArgs.createSpec(output: Appendable, error: Appendable): Processi
         config {
             useDefaultConfig = args.buildUponDefaultConfig
             shouldValidateBeforeAnalysis = null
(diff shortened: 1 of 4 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
```diff
@@ -36,7 +36,6 @@ internal fun CliArgs.createSpec(output: Appendable, error: Appendable): Processi
         config {
             useDefaultConfig = args.buildUponDefaultConfig
             shouldValidateBeforeAnalysis = null
-            knownPatterns = emptyList()
             // ^^ cli does not have these properties yet; specified in yaml config for now
```
BraisGabin, line 39:
```suggestion
```

BraisGabin (approved):

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
3flex: Oh good one, didn't even notice that

Issue #6198, opened 2023-06-09
Override the Kotlin Version on Gradle Kotlin DSL


nico-arianto:
## Expected Behavior
__Code:__
```
plugins {
    val kotlinVersion = "1.8.22"
    kotlin("jvm") version kotlinVersion
    kotlin("kapt") version kotlinVersion
    kotlin("plugin.spring") version kotlinVersion
    kotlin("plugin.jpa") version kotlinVersion
    id("io.gitlab.arturbosch.detekt") version "1.23.0"
}

detekt {
    kotlinVersion = "1.8.21"
    source.from("src/main")
    buildUponDefaultConfig = true
    allRules = true
}
```

This line `kotlinVersion = "1.8.21"` will override the Kotlin version and being used by `detekt`

Note: Or other alternative approach also can, as long can overcome the Kotlin compatibility issue.

## Current Behavior
__Code:__
```
plugins {
    val kotlinVersion = "1.8.22"
    kotlin("jvm") version kotlinVersion
    kotlin("kapt") version kotlinVersion
    kotlin("plugin.spring") version kotlinVersion
    kotlin("plugin.jpa") version kotlinVersion
    id("io.gitlab.arturbosch.detekt") version "1.23.0"
}

detekt {
    source.from("src/main")
    buildUponDefaultConfig = true
    allRules = true
}
```

__Gradle Task:__
```
> Task :detekt FAILED
Time elapsed [detekt] : 0s
///// Time Consumption Summary /////
[dependencyCheckAnalyze          ] : 4s
[compileJava                     ] : 1s
[assemble                        ] : 0s
[bootJar                         ] : 0s
[classes                         ] : 0s
[compileKotlin                   ] : 0s
[detekt                          ] : 0s
[jar                             ] : 0s
[kaptGenerateStubsKotlin         ] : 0s
[kaptKotlin                      ] : 0s
[processResources                ] : 0s
[resolveMainClassName            ] : 0s

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':detekt'.
> detekt was compiled with Kotlin 1.8.21 but is currently running with 1.8.22.
  This is not supported. See https://detekt.dev/docs/gettingstarted/gradle#dependencies for more information
```

## Context
`detekt` didn't locked us to use the latest Kotlin version for other Gradle tasks

3flex:
Did you check the link shown in the log for details on how to fix this?

[...]

PR #6472, opened 2023-09-07
Function to return supported Kotlin version

kkocel:
Introduce the function that returns the supported Kotlin version as mentioned in: https://github.com/detekt/detekt/issues/6198#issuecomment-1710122711


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -139,3 +141,15 @@ class DetektPlugin : Plugin<Project> {
 const val CONFIGURATION_DETEKT = "detekt"
 const val CONFIGURATION_DETEKT_PLUGINS = "detektPlugins"
 const val USE_WORKER_API = "detekt.use.worker.api"
+
+fun getSupportedKotlinVersion(): String {
```
3flex, line 145: Can you please add an `@Incubating` annotation here?

3flex (commented):
I just want to note that while this will work in the majority of cases it is possible to change the version of detekt-cli that the plugin calls, and the resolved version of the CLI dependency might depend on a different Kotlin version which will cause failures again.

I think we can get the resolved Kotlin dependency version from Gradle though when I last tried this it didn't seem that easy. But in the meantime annotating with Incubating will at least flag in IDE that this isn't 100% stable or foolproof.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
kkocel: Done

kkocel:
> I just want to note that while this will work in the majority of cases it is possible to change the version of detekt-cli that the plugin calls, and the resolved version of the CLI dependency might depend on a different Kotlin version which will cause failures again.

I thought about adding dependency to detekt-cli and query supported version from there but I assumed that Kotlin versions will be consistent across modules.

kkocel:
@3flex is there anything here that I should change or it's good to be merged?

BraisGabin:
I vote to merge this. They should not be unaligned. And if they do we could always change this implementation without any breaking change.

BraisGabin (approved):

BraisGabin:
@kkocel could you also update the documentation pointing out how to use this workaround? And that it should only be needed if the user applies the spring dependency manager (I can't recall the exact name)

kkocel:
@BraisGabin thank you. I added the relevant part in the _future_ docs and mentioned Spring Dependency Plugin there.

PR #6477, opened 2023-09-10
Stop using data classes in public API

3flex:
https://kotlinlang.org/docs/jvm-api-guidelines-backward-compatibility.html#don-t-use-data-classes-in-an-api

> it's better not to use data classes in your API because almost any change in them breaks source, binary, or behavioral compatibility.

Instead I've introduced the Poko plugin where required because calling code used equality checks. This plugin generates overrides for `equals`,  `hashCode` and `toString` but does not generate the `componentX` or `copy` functions that lead to backwards compatibility issues.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Debt.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Debt.kt
@@ -4,7 +4,7 @@ package io.gitlab.arturbosch.detekt.api
  * Debt describes the estimated amount of work needed to fix a given issue.
  */
 @Suppress("MagicNumber")
-data class Debt(val days: Int = 0, val hours: Int = 0, val mins: Int = 0) {
+class Debt(val days: Int = 0, val hours: Int = 0, val mins: Int = 0) {
 
     init {
         require(days >= 0 && hours >= 0 && mins >= 0)
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Debt.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Debt.kt
@@ -3,7 +3,6 @@ package io.gitlab.arturbosch.detekt.api
 /**
  * Debt describes the estimated amount of work needed to fix a given issue.
  */
-@Suppress("MagicNumber")
 class Debt(val days: Int = 0, val hours: Int = 0, val mins: Int = 0) {
 
     init {
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Entity.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Entity.kt
@@ -11,8 +11,8 @@ import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
 /**
  * Stores information about a specific code fragment.
  */
-data class Entity(
-    private val name: String,
+class Entity(
+    val name: String,
     val signature: String,
     val location: Location,
     val ktElement: KtElement? = null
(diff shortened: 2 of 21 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Entity.kt
```diff
@@ -11,8 +11,8 @@ import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
 /**
  * Stores information about a specific code fragment.
  */
-data class Entity(
-    private val name: String,
+class Entity(
+    val name: String,
```
BraisGabin, line 15: Is this change intended? Shouldn't it remain as private?

BraisGabin:
<3 good one! I had this on my todo list.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Entity.kt
3flex: It's deliberate - to copy to a new instance with changed values this needs to be readable so the value can be copied to the new instance.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Debt.kt
```diff
@@ -4,7 +4,7 @@ package io.gitlab.arturbosch.detekt.api
  * Debt describes the estimated amount of work needed to fix a given issue.
  */
 @Suppress("MagicNumber")
```
schalkms, line 6: Btw, is this `Suppress` statement still needed?

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/CorrectableRulesFirstSpec.kt
```diff
@@ -19,14 +19,14 @@ class CorrectableRulesFirstSpec {
         var actualLastRuleId = ""
 
         class First(config: Config) : Rule(config) {
-            override val issue: Issue = justAnIssue.copy(id = "NonCorrectable")
+            override val issue: Issue = Issue("NonCorrectable", justAnIssue.description, justAnIssue.debt)
```
schalkms, line 22:
```suggestion
            override val issue: Issue = Issue("NonCorrectable", "", Debt.FIVE_MINS)
```

and remove the variable `justAnIssue`

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/CorrectableRulesFirstSpec.kt
```diff
@@ -19,14 +19,14 @@ class CorrectableRulesFirstSpec {
...
             override fun visitClass(klass: KtClass) {
                 actualLastRuleId = issue.id
             }
         }
 
         class Last(config: Config) : Rule(config) {
-            override val issue: Issue = justAnIssue.copy(id = "Correctable")
+            override val issue: Issue = Issue("Correctable", justAnIssue.description, justAnIssue.debt)
```
schalkms, line 29:
```suggestion
            override val issue: Issue = Issue("Correctable", "", Debt.FIVE_MINS)
```

schalkms (approved):
👍

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Entity.kt
BraisGabin: But do we need to copy them?

3flex: I just did a simple refactoring to get rid of the data classes without doing anything more at the call sites. It's possible that copying isn't required. I can take a closer look and see if we can keep this private if that's a concern?

BraisGabin: I think that this one should be even removed (on later refactors) so it will help to keep it private. But if it gives too much work, don't worry about it.

BraisGabin (approved):

Issue #6474, opened 2023-09-08
ForbiddenMethodCall does not trigger on default interface implementations


vatbub:
This bug is related to #4204

## Expected Behavior
Consider the following code:

```kotlin
import javafx.beans.property.SimpleObjectProperty

class DetektForbiddenMethodCallDefaultInterfaceDemo {
    val objectProperty = SimpleObjectProperty("Hello there")
    val mappedProperty = objectProperty.map { it.uppercase() }
}
```

With the following configuration, ForbiddenMethodCall should flag the call to `.map()`.

```yml
  ForbiddenMethodCall:
    active: true
    methods:
      - reason: 'The javafx-built in ObservableValue.map function handles null values in an unusual and in Kotlin undesired way. Use map2() instead.'
        value: 'javafx.beans.value.ObservableValue.map'
```

## Observed Behavior
The call is not flagged. However, the following configuration will flag the call:

```yml
  ForbiddenMethodCall:
    active: true
    methods:
      - reason: 'The javafx-built in ObservableValue.map function handles null values in an unusual and in Kotlin undesired way. Use map2() instead.'
        value: 'javafx.beans.value.ObservableValue.map'
      - reason: 'The javafx-built in ObservableValue.map function handles null values in an unusual and in Kotlin undesired way. Use map2() instead.'
        value: 'javafx.beans.property.SimpleObjectProperty.map'
```

It seems like this is the same bug as #4204 except with one difference: `SimpleObjectProperty` does not override `map`. Instead, the `ObservableValue` interface contains a default implementation, which apparently is not recognized by the patch that fixed #4204 .

Here are the relevant portions of source code for reference: [ObservableValue](https://github.com/openjdk/jfx/blob/eb7de72dafecbedc83c2215b6aed7432d4ec80f9/modules/javafx.base/src/main/java/javafx/beans/value/ObservableValue.java#L178) and [SimpleObjectProperty](https://github.com/openjdk/jfx/blob/eb7de72dafecbedc83c2215b6aed7432d4ec80f9/modules/javafx.base/src/main/java/javafx/beans/property/SimpleObjectProperty.java#L39)

## Steps to Reproduce
Create a project with the following dependency:

```xml
<dependency>
    <groupId>org.openjfx</groupId>
    <artifactId>javafx-base</artifactId>
    <version>19.0.2.1</version>
</dependency>
```

Then paste the above code in the project and run detekt with type resolution enabled on that project.

## Your Environment
[...]

PR #6478, opened 2023-09-11
Handle sequence of overridden methods

atulgpt:
Use `overriddenTreeUniqueAsSequence` of overridden methods which gives all the methods even for the parent classes

Fixes https://github.com/detekt/detekt/issues/6474

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
@@ -12,6 +12,7 @@ import io.gitlab.arturbosch.detekt.api.config
 import io.gitlab.arturbosch.detekt.api.internal.Configuration
 import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
 import io.gitlab.arturbosch.detekt.api.valuesWithReason
+import org.jetbrains.kotlin.descriptors.CallableDescriptor
 import org.jetbrains.kotlin.descriptors.FunctionDescriptor
 import org.jetbrains.kotlin.descriptors.PropertyDescriptor
 import org.jetbrains.kotlin.descriptors.SyntheticPropertyDescriptor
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
```diff
@@ -113,14 +113,20 @@ class ForbiddenMethodCall(config: Config = Config.empty) : Rule(config) {
...
+                        callableDescriptor.unwrappedGetMethod,
+                        callableDescriptor.unwrappedSetMethod
+                    )
+                } else {
+                    listOf(callableDescriptor)
+                }
+                foundDescriptors.flatMap {
+                    it.overriddenTreeUniqueAsSequence(true).toList()
+                }
+            }?.toSet() ?: return
```
BraisGabin, line 129: We could avoid this call to `toSet` if we create sets directly instead of lists inside the `let`.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCallSpec.kt
```diff
@@ -433,6 +433,62 @@ class ForbiddenMethodCallSpec(val env: KotlinCoreEnvironment) {
...
+                }
+            }
+
+            class ComplexComparator : SimpleComparator() {
+                override fun compare(p0: String?, p1: String?): Int {
+                    return 0
+                }
+            }
+            
+            fun foo(i: ComplexComparator) {
```
BraisGabin, line 481:
```suggestion
            fun foo() {
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenMethodCall.kt
atulgpt: Update the code

BraisGabin (approved):

BraisGabin (approved):

PR #6479, opened 2023-09-12
Made all reports have uniform ids

lexa-diky:


BraisGabin:
Shouldn't we align in the other way? Just change the sarif Id? To be honest I don't understand why they got unaligned in the first place.

lexa-diky:
That's also an option, this will be less breaking, but I suggest sill fixing report names in `id` property.

+ now thinking of shorter names in this MR, maybe a bit more cryptic `XmlOutputReport` is better in this case, at least it's not implying that this is some standard `THE XML` report, but just some detekt specific thing.

@BraisGabin could you please summon any third person to this PR, to have final vote?

BraisGabin:
About "the xml" we could even rename it to "checkstyle", because that's the format.

cortinico:
Closing because of #6505

PR #6496, opened 2023-09-17
Convert DGP extension to use Property API

3flex:
#1931 fix part 1 - addressing extension side. Follow up PRs will migrate tasks.

Thanks to the simple property assignment in Kotlin DSL scripts feature [introduced in Gradle 8.1](https://docs.gradle.org/8.1.1/release-notes.html#experimental-simple-property-assignment-in-kotlin-dsl-scripts), this will have a minimal impact on users who are using a modern Gradle version. The feature will be stable from Gradle 8.4.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -21,10 +21,10 @@ allprojects {
 
     detekt {
         source.setFrom(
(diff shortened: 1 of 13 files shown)


BraisGabin (approved):
Nice one.

PR #6497, opened 2023-09-17
Convert DGP tasks to use Property API

### LINKED ISSUE #1931: Property API for Gradle plugin

3flex:
https://github.com/detekt/detekt/issues/1931 fix part 2 (part 1 was done in #6496). Migrates remaining task properties to use Gradle's property API.

Closes #1931

### DIFF
--- a/detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektTaskGroovyDslSpec.kt
+++ b/detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektTaskGroovyDslSpec.kt
@@ -56,7 +56,7 @@ class DetektTaskGroovyDslSpec {
                 buildUponDefaultConfig = true
                 allRules = false
                 autoCorrect = false
-                basePath = projectDir
+                basePath = projectDir.absolutePath
                 reports {
                     xml {
                         required.set(true)
@@ -101,7 +101,7 @@ class DetektTaskGroovyDslSpec {
                 buildUponDefaultConfig = true
                 allRules = false
                 autoCorrect = false
-                basePath = projectDir
+                basePath = projectDir.absolutePath
             }
         """.trimIndent()
         val groovyBuilder = DslTestBuilder.groovy()
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -53,7 +53,6 @@ import org.gradle.api.tasks.TaskAction
 import org.gradle.api.tasks.options.Option
 import org.gradle.language.base.plugins.LifecycleBasePlugin
 import org.gradle.workers.WorkerExecutor
-import java.io.File
 import javax.inject.Inject
 
 @CacheableTask
@@ -127,7 +126,7 @@ abstract class Detekt @Inject constructor(
     var reports: DetektReports = objects.newInstance(DetektReports::class.java)
 
     @get:Internal
-    abstract val reportsDir: Property<File>
+    abstract val reportsDir: DirectoryProperty
 
     val xmlReportFile: Provider<RegularFile>
         @OutputFile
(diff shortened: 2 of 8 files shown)


detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektTaskGroovyDslSpec.kt
```diff
@@ -56,7 +56,7 @@ class DetektTaskGroovyDslSpec {
                 buildUponDefaultConfig = true
                 allRules = false
                 autoCorrect = false
-                basePath = projectDir
+                basePath = projectDir.absolutePath
```
chao2zhang, line 59: What caused you to add `.absolutePath` here?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -50,19 +50,17 @@ import org.gradle.api.tasks.PathSensitivity
...
-import java.io.File
 import javax.inject.Inject
 
 @CacheableTask
 abstract class Detekt @Inject constructor(
     private val objects: ObjectFactory,
     private val workerExecutor: WorkerExecutor,
     private val providers: ProviderFactory,
-) : SourceTask(), VerificationTask {
+) : SourceTask() {
```
chao2zhang, line 63: I thought the verification task can impact the behavior of gradle caching. Is it still possible to add back the interface `VerificationTask`?

3flex: I'm not sure how this would affect Gradle caching? It's just an interface and the `@Input` annotation is carried over to the replacement `ignoreFailures` property.

detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektTaskGroovyDslSpec.kt
3flex: When `basePath` was a `String` then `projectDir` (a `File`) was implicitly converted to a string by Groovy.

Now `basePath` is a `Property<String>` Groovy won't do the implicit conversion so it must be assigned to a `String`.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
chao2zhang: Probably not caching, although some plugin code does rely on `VerificationTask` e.g https://github.com/JetBrains/kotlin/blob/90102ad8b77a393a1367d747bf97b895b18d78f0/build.gradle.kts#L961

3flex: That's not plugin code though, just a build file in the Kotlin project. They must have some reason for having that logic but it doesn't impact the implementation of our plugin and if anyone happens to use detekt and disable verification tasks using that interface they can filter based on a DGP type instead - or if anyone really misses it it can added back later.

chao2zhang (approved):

PR #6501, opened 2023-09-18
Implement rules suggesting to convert statements to expressions

### LINKED ISSUE #6233: Suggest to use `if` and `when` as expression when applicable
Prohibit usage of `if` or `when` like below

Non-compliant code
```
fun test(foo: Int): Int {
   if (foo > 0) {
      return 0
   } else {
      return 1
   }
}
```
(above code is for `if` used for returning value, the same can be applicable for variable assignment or when expression)

Compliant code
```
fun test(foo: Int): Int {
   return if (foo > 0) {
      0
   } else {
      1
   }
}
```

## Context
Many times in my PR reviews in my office project I see these types of usages which generally happens with people who are new to Kotlin. Having this rule will not only improve the code as well as educated the usage of `if` and `when` as an expression

ErdoganSeref:
The rules detect `if` and `when` statements that can be converted to expressions. They check if the statements are exhaustive and if the last statements of each branch allow a conversion.

The change fixes #6233.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -616,6 +616,8 @@ style:
     ignoreOverridableFunction: true
     ignoreActualFunction: true
     excludedFunctions: []
+  IfStatementCouldBeExpression:
+    active: false
   LoopWithTooManyJumpStatements:
     active: true
     maxJumpCount: 1
@@ -789,6 +791,8 @@ style:
   VarCouldBeVal:
     active: true
     ignoreLateinitVar: false
+  WhenStatementCouldBeExpression:
+    active: false
   WildcardImport:
     active: true
     excludeImports:
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/IfStatementCouldBeExpression.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/IfStatementCouldBeExpression.kt
@@ -1,0 +1,91 @@
+package io.gitlab.arturbosch.detekt.rules.style
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtIfExpression
+
+/**
+ * Detects `if` statements that could be expressions.
+ *
+ * An `if` statement can be an expression if it has the following properties:
+ *
+ * 1. The `if` statement is exhaustive i.e. it has an `else` branch.
+ * 2. The last statement of every branch is either a `return` or an assignment to the same variable with the same operator.
+ *
+ * Refactoring the statements to expressions reduces the `return` and assignment count. This improves readability.
+ *
+ * <noncompliant>
+ *     if (a == 1) {
+ *       a = 1
+ *     } else {
+ *       a = 0
+ *     }
+ *
+ *     if (a == 1) {
+ *       return 1
+ *     } else {
+ *       return 0
+ *     }
+ * </noncompliant>
+ *
+ * <compliant>
+ *     a = if (a == 1) {
+ *       1
+ *     } else {
+ *       0
+ *     }
+ *
+ *     return if (a == 1) {
+ *       1
+ *     } else {
+ *       0
+ *     }
+ * </compliant>
+ */
+
+class IfStatementCouldBeExpression(config: Config = Config.empty) : Rule(config) {
+
+    override val issue = Issue(
+        "IfStatementCouldBeExpression",
+        "if statement could be expression. Use an if expression to reduce the return and assignment count.",
+        Debt.FIVE_MINS,
+    )
+
+    override fun visitIfExpression(expression: KtIfExpression) {
+        if (canLiftOutOf(expression)) {
+            report(CodeSmell(issue, Entity.from(expression), issue.description))
+        }
+    }
+
+    private fun canLiftOutOf(expression: KtIfExpression): Boolean {
+        val branches = getBranches(expression)
+        return isExhaustive(expression) && canLiftOut(branches)
+    }
+
+    private fun getBranches(expression: KtIfExpression): List<KtExpression> {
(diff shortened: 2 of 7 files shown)


detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -616,6 +616,8 @@ style:
     ignoreOverridableFunction: true
     ignoreActualFunction: true
     excludedFunctions: []
+  IfStatementCouldBeExpression:
```
cortinico, line 619: nit: we also have a rule called `ExpressionBodySyntax` already, which should probably be renamed to `FunctionBodyCouldBeExpression` if we merge this

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/IfStatementCouldBeExpression.kt
```diff
@@ -0,0 +1,91 @@
...
+class IfStatementCouldBeExpression(config: Config = Config.empty) : Rule(config) {
+
+    override val issue = Issue(
+        "IfStatementCouldBeExpression",
+        "if statement could be expression. Use an if expression to reduce the return and assignment count.",
+        Debt.FIVE_MINS,
+    )
+
+    override fun visitIfExpression(expression: KtIfExpression) {
+        if (canLiftOutOf(expression)) {
```
atulgpt, line 60: we should add `super.visitIfExpression` otherwise in case of nested if-else this will not run

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/IfStatementCouldBeExpression.kt
```diff
@@ -0,0 +1,91 @@
...
+ *     } else {
+ *       0
+ *     }
+ * </compliant>
+ */
+
+class IfStatementCouldBeExpression(config: Config = Config.empty) : Rule(config) {
+
+    override val issue = Issue(
+        "IfStatementCouldBeExpression",
```
atulgpt, line 54:
```suggestion
        this::class.simplename,
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/WhenStatementCouldBeExpression.kt
```diff
@@ -0,0 +1,71 @@
...
+ *       else -> 0
+ *     }
+ * </compliant>
+ */
+
+@RequiresTypeResolution
+class WhenStatementCouldBeExpression(config: Config = Config.empty) : Rule(config) {
+
+    override val issue = Issue(
+        "WhenStatementCouldBeExpression",
```
atulgpt, line 52:
```suggestion
        this::class.simplename
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/WhenStatementCouldBeExpression.kt
```diff
@@ -0,0 +1,71 @@
...
+        "when statement could be expression. Use a when expression to reduce the return and assignment count.",
+        Debt.FIVE_MINS,
+    )
+
+    override fun visitWhenExpression(expression: KtWhenExpression) {
+        if (canLiftOutOf(expression)) {
```
atulgpt, line 58: `super` call should be added

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/IfStatementCouldBeExpressionSpec.kt
```diff
@@ -0,0 +1,144 @@
...
+import org.junit.jupiter.api.Test
+
+class IfStatementCouldBeExpressionSpec {
+    val subject = IfStatementCouldBeExpression(Config.empty)
+
+    @Nested
+    inner class `return in if` {
+        @Test
+        fun `does not report an expression`() {
+            val code = """
```
atulgpt, line 16: can we also add TCs for the below code?

compliant code
```kotlin
fun f(a: Int): Int {
    return if (a > 1) {
        println("a")
        1
    } else if (a < 1) {
        println("b")
        0
    } else {
        println("c")
        -1
    }
}
```

compliant code
```kotlin
fun f() {
    val b: Int
    if (a > 1) {
        println("a")
        b = 1
    } else if (a < 1) {
        println("b")
        b = 0
    } else {
        b = -1
        println("c")
    }
}
```

non-compliant code
```kotlin
fun f() {
    val b: Int
    if (a > 1)  a = 1 else a = 2
}
```

non-compliant code for nested if
```kotlin
fun f(a: Int) {
    if (a > 0) {
        val b: Int
        if (a > 100) { b = 100 } else. { b = 0} 
    } else {
        println("")
    }
}
```

non-compliant code for parent and child if
```kotlin
fun f(a: Int) {\
    val c: Int
    if (a > 0) {
        val b: Int
        if (a > 100) { b = 100 } else. { b = 0}
        c = 0
    } else {
        println("")
       c = -1
    }
}
```


non-compliant code for parent and child if
```kotlin
fun f(a: Int) {\
    val b: Int
    if (a > 0) {
        if (a > 100) { b = 100 } else. { b = 0}
        b
    } else {
        println("")
        b = -1
    }
}
```

non-compliant code last two statements should be pulled out
```kotlin
fun f(a: Int) {\
    if (a > 0) {
        print("if")
        print("a")
        foo()
    } else {
        print("else")
        print("a")
        foo()
    }
}
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/IfStatementCouldBeExpressionSpec.kt
```diff
@@ -0,0 +1,144 @@
...
+                fun f(a: Int): Int {
```
atulgpt, line 17: Should we also report statements like below

```kotlin
fun f(a: Int) {
    if (a > 0) {
       print("a")
        foo()
        print("if")
    } else {
        print("a")
        foo()
        print("else")
    }
}
```

to compliant as below

```kotlin
fun f(a: Int) {
    print("a")
    foo()
    if (a > 0) {
        print("if")
    } else {
        print("else")
    }
}
```

BraisGabin:
@ErdoganSeref Are you willing to continue with this PR? There are some reviews that should be addressed before we can merge this.

ErdoganSeref:
> @ErdoganSeref Are you willing to continue with this PR? There are some reviews that should be addressed before we can merge this.

I"ll continue working on this PR soon.

PR #6511, opened 2023-09-23
Sample implementation of onlyAnnotation Suppressor

lexa-diky:
Split annotation supressor into ignore and only implementations

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
@@ -2,10 +2,25 @@ package io.gitlab.arturbosch.detekt.core.suppressors
 
 import io.gitlab.arturbosch.detekt.api.AnnotationExcluder
 import io.gitlab.arturbosch.detekt.api.ConfigAware
+import io.gitlab.arturbosch.detekt.api.Finding
 import org.jetbrains.kotlin.psi.KtAnnotated
 import org.jetbrains.kotlin.psi.KtElement
 import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 import org.jetbrains.kotlin.resolve.BindingContext
+
+private class IgnoreAnnotationSuppressor(
+    private val ignoreAnnotations: List<Regex>,
+    private val bindingContext: BindingContext,
+    private val shouldContainAnnotation: Boolean
+) : Suppressor {
+
+    override fun shouldSuppress(finding: Finding): Boolean {
+        val element = finding.entity.ktElement
+        return element != null && !shouldContainAnnotation == element.isAnnotatedWith(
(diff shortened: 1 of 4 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
```diff
@@ -2,34 +2,79 @@ package io.gitlab.arturbosch.detekt.core.suppressors
...
+
+/**
+ * Suppress all the issues that are raised under a code that is not annotated with the annotations defined at
+ * `ignoreAnnotated`.
+ *
+ * @config ignoreAnnotated: List<String> The annotations can be defined just by its name or with its fully qualified
+ * name. If you don't run detekt with type solving the fully qualified name does not work.
+ */
+internal fun onlyAnnotationSuppressorFactory(rule: ConfigAware, bindingContext: BindingContext): Suppressor? {
+    return annotationSuppressorFactory(rule, bindingContext, "onlyAnnotated", true)
```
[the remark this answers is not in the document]
cortinico: Nope docs update goes in the same PR when a new config is added

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
```diff
@@ -2,34 +2,79 @@ package io.gitlab.arturbosch.detekt.core.suppressors
...
+}
+
+/**
+ * Order of returned elements is important, `onlyAnnotated` should be applied before `ignoreAnnotated`.
```
TWiStErRob, line 48: Could this be tested somehow in a high level (non-unit) to prevent future regression? This could become a very hidden behavioural breaking change in a future refactoring.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
```diff
@@ -2,34 +2,79 @@ package io.gitlab.arturbosch.detekt.core.suppressors
...
+private class IgnoreAnnotationSuppressor(
+    private val ignoreAnnotations: List<Regex>,
+    private val bindingContext: BindingContext,
+    private val shouldContainAnnotation: Boolean
+) : Suppressor {
+
+    override fun shouldSuppress(finding: Finding): Boolean {
+        val element = finding.entity.ktElement
+        return element != null && !shouldContainAnnotation == element.isAnnotatedWith(
+            AnnotationExcluder(element.containingKtFile, ignoreAnnotations, bindingContext),
```
TWiStErRob, line 20: I think there are too many negations here, I personally found it impossible<sup>*</sup> to reason about these lines without mechanically executing the perceived generated bytecode in my head. Would there be some way to reduce the negations?

The negations I see are `!`, `...Excluder`, `ignore...`, and on top of this there's a boolean `==`.

_<sup>*</sup> I might be just too tired after a conference 😅_

PR #6547, opened 2023-10-15
Create DetektBasePlugin

3flex:
Create DetektBasePlugin for DGP. This is used to apply configuration that's common to the Gradle plugin for the Kotlin compiler plugin as well as the classic plugin, which removes some duplication and should make it easier to reason about the responsibilities of each plugin.

The base plugin is responsible for creating the project extension and configuring it, as well as creating the `detektPlugins` configuration (which is used by both other plugins). The classic plugin sets up the detekt tasks for each compilation. The compiler plugin configures the compiler plugin tasks.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/dev/detekt/gradle/plugin/DetektBasePlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/dev/detekt/gradle/plugin/DetektBasePlugin.kt
@@ -56,20 +56,20 @@ class DetektBasePlugin : Plugin<Project> {
         internal const val CONFIG_DIR_NAME = "config/detekt"
         internal const val CONFIG_FILE = "detekt.yml"
 
-        internal const val DEFAULT_SRC_DIR_JAVA = "src/main/java"
-        internal const val DEFAULT_TEST_SRC_DIR_JAVA = "src/test/java"
-        internal const val DEFAULT_SRC_DIR_KOTLIN = "src/main/kotlin"
-        internal const val DEFAULT_TEST_SRC_DIR_KOTLIN = "src/test/kotlin"
-        internal const val DEFAULT_DEBUG_VALUE = false
-        internal const val DEFAULT_IGNORE_FAILURES = false
-        internal const val DEFAULT_PARALLEL_VALUE = false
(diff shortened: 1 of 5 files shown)


detekt-gradle-plugin/src/main/kotlin/dev/detekt/gradle/plugin/DetektBasePlugin.kt
```diff
@@ -0,0 +1,76 @@
...
+        internal const val DEFAULT_DEBUG_VALUE = false
+        internal const val DEFAULT_IGNORE_FAILURES = false
+        internal const val DEFAULT_PARALLEL_VALUE = false
+        internal const val DEFAULT_AUTO_CORRECT_VALUE = false
+        internal const val DEFAULT_DISABLE_RULESETS_VALUE = false
+        internal const val DEFAULT_ALL_RULES_VALUE = false
+        internal const val DEFAULT_BUILD_UPON_DEFAULT_CONFIG_VALUE = false
+
+        // This flag is ignored unless the compiler plugin is applied to the project
+        internal const val DEFAULT_COMPILER_PLUGIN_ENABLED = true
```
BraisGabin, line 72: Now these constant can be private.

```suggestion
        private const val DEFAULT_SRC_DIR_JAVA = "src/main/java"
        private const val DEFAULT_TEST_SRC_DIR_JAVA = "src/test/java"
        private const val DEFAULT_SRC_DIR_KOTLIN = "src/main/kotlin"
        private const val DEFAULT_TEST_SRC_DIR_KOTLIN = "src/test/kotlin"
        private const val DEFAULT_DEBUG_VALUE = false
        private const val DEFAULT_IGNORE_FAILURES = false
        private const val DEFAULT_PARALLEL_VALUE = false
        private const val DEFAULT_AUTO_CORRECT_VALUE = false
        private const val DEFAULT_DISABLE_RULESETS_VALUE = false
        private const val DEFAULT_ALL_RULES_VALUE = false
        private const val DEFAULT_BUILD_UPON_DEFAULT_CONFIG_VALUE = false

        // This flag is ignored unless the compiler plugin is applied to the project
        private const val DEFAULT_COMPILER_PLUGIN_ENABLED = true
```

BraisGabin (approved):
Great cleanup.

PR #6553, opened 2023-10-18
Better example for UseLet rule

### LINKED ISSUE #6552: Compiant code example for the `UseLet` rule shows non-idiomatic code
## Observed Behavior

The 'compliant' code for the [`UseLet` rule](https://detekt.dev/docs/rules/style#uselet): `x?.let { it.transform() }` (shown as a better alternative to non-compiant `if (x != null) { x.transform() } else null` is itself non-compliant, because it can be better rewritten as a mere

```kotlin
    x?.transform()
```

thus making it not applicable for the rule (because there would be no `let()` call in the resulting most concise code).

The second example `if (x == null) null else y` -> `x?.let { y }` seems to be fine though.

## Expected Behavior

Said example should be removed from rule's documentation.

## Steps to Reproduce

It's just a discrepancy in the documentation.

## Context

Well, I just wanted to decide whether the `UseLet` rule should be activated in my configuration or not.

## Your Environment

* Version of detekt used: latest currently available (**1.23.1**)
* Gradle scan link (add `--scan` option when running the gradle task): --
* Operating System and version: irrelevant

ILikeYourHat:
Fixes #6552

Added a better example for `UseLet` rule, that can't be rewritten in a shorter form.

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseLet.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseLet.kt
@@ -20,12 +20,12 @@ import org.jetbrains.kotlin.psi.KtPsiUtil
  * `null` in the truthy case are better represented as `?.let {}` blocks.
  *
  * <noncompliant>
- * if (x != null) { x.transform() } else null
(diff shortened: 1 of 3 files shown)


schalkms (approved):

schalkms:
Thank you!

Issue #6540, opened 2023-10-10
UnusedImport false positive when extension method import is only used in documentation


matejdro:
## Expected Behavior

UnusedImports should not flag extension method import that is only used in documentation

## Observed Behavior

UnusedImport falsly flags it

## Steps to Reproduce

Analyze following code with `UnusedImports` enabled:

```kotlin
import kotlinx.coroutines.flow.Flow
import kotlinx.coroutines.flow.onStart

/**
 * [Flow.onStart]
 */
class Test
```

## Your Environment

* Version of detekt used: 1.23.1
* Version of Gradle used (if applicable): 8.4

cortinico:
`UnusedImports` is coming from the `detekt-formatting` ruleset which is the KtLint wrapper. So it means that this is a bug in the KtLint code. Have you searched there and/or opened an issue over there?

matejdro:
Ah right, sorry, I mixed up the two rules. Made the ktlint issue here: https://github.com/pinterest/ktlint/issues/2330

Do you want me to close this issue?

BraisGabin:
Yes, we can close it.

matejdro:
Wait, are we sure that UnusedImports is from ktlint?

https://github.com/detekt/detekt/issues/6560#issuecomment-1775900550 I am getting confused now :sweat_smile:

cortinico:
`UnusedImport` is from the `style` ruleset (== from detekt):
https://detekt.dev/docs/next/rules/style/#unusedimport

`NoUnusedImports` is from the `formatti` ruleset (== from ktlint):
https://detekt.dev/docs/next/rules/formatting/#nounusedimports

matejdro:
Can we reopen this ticket then? Since it's not from ktlint

matejdro:
It's not waiting an author's feedback. I might be able to create PR for that at some point, but I cannot give any time lines.

atulgpt:
Hi @matejdro I have created one PR https://github.com/detekt/detekt/pull/6567 this is pending review

matejdro:
Thank you!

matejdro:
@cortinico @BraisGabin It appears this is back in the `2.0.0-alpha.1`. Can we reopen this? Or do you want me to open a new issue?

cortinico:
> Can we reopen this? Or do you want me to open a new issue?

Let's open a new issue 👍

matejdro:
[...]

PR #6567, opened 2023-10-26
Fix by adding last method call to the set

atulgpt:
Fixes https://github.com/detekt/detekt/issues/6540

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImport.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImport.kt
@@ -143,8 +143,9 @@ class UnusedImport(config: Config) : Rule(config) {
             kotlinDocReferencesRegExp.findAll(content, 0)
                 .map { it.groupValues[1] }
                 .forEach {
-                    namedReferencesInKDoc.add(it.split(".")[0])
-                    namedReferencesInKDoc.add(it.split(".").last())
+                    val referenceNames = it.split(".")
+                    namedReferencesInKDoc.add(referenceNames[0])
+                    namedReferencesInKDoc.add(referenceNames.last())
                 }
             kotlinDocBlockTagReferenceRegExp.find(content)?.let {
                 val str = it.groupValues[2].split(whiteSpaceRegex)[0]
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImport.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImport.kt
@@ -142,7 +142,10 @@ class UnusedImport(config: Config) : Rule(config) {
         private fun handleKDoc(content: String) {
             kotlinDocReferencesRegExp.findAll(content, 0)
                 .map { it.groupValues[1] }
-                .forEach { namedReferencesInKDoc.add(it.split(".")[0]) }
+                .forEach {
+                    namedReferencesInKDoc.add(it.split(".")[0])
+                    namedReferencesInKDoc.add(it.split(".").last())
+                }
             kotlinDocBlockTagReferenceRegExp.find(content)?.let {
                 val str = it.groupValues[2].split(whiteSpaceRegex)[0]
                 namedReferencesInKDoc.add(str.split(".")[0])
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportSpec.kt
@@ -142,6 +142,64 @@ class UnusedImportSpec(val env: KotlinCoreEnvironment) {
     }
 
     @Test
+    fun `does not report KDoc references with companion method calls`() {
+        val main = """
+            package com.example
+
+            import android.text.TextWatcher.beforeTextChanged
+
+            class Test {
+                /**
+                 * [beforeTextChanged]
+                 */
+                fun test() {
+                    TODO()
+                }
+            }
+        """.trimIndent()
+        val additional = """
+            package android.text
+
+            object TextWatcher {
+                fun beforeTextChanged() {}
+            }
+        """.trimIndent()
+        assertThat(subject.lintWithContext(env, main, additional)).isEmpty()
+    }
+
+    @Test
+    fun `does not report KDoc references with extension method calls`() {
+        val main = """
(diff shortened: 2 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImport.kt
```diff
@@ -142,7 +142,10 @@ class UnusedImport(config: Config) : Rule(config) {
         private fun handleKDoc(content: String) {
             kotlinDocReferencesRegExp.findAll(content, 0)
                 .map { it.groupValues[1] }
-                .forEach { namedReferencesInKDoc.add(it.split(".")[0]) }
+                .forEach {
+                    namedReferencesInKDoc.add(it.split(".")[0])
+                    namedReferencesInKDoc.add(it.split(".").last())
```
atulgpt, line 147: I have added the last method name in the set which will be matched against the import. This might in some scenarios produce false -ve but those cases will be kind of rare and fixing that might require some juggling with binding context to find the eligible last method call(whose extension are alreade imported)

BraisGabin, line 147: We could store the result of split on a variable to avoid the splitting twice.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportSpec.kt
```diff
@@ -141,6 +141,64 @@ class UnusedImportSpec(val env: KotlinCoreEnvironment) {
...
+        assertThat(subject.lintWithContext(env, main, additional)).isEmpty()
+    }
+
+    @Test
+    fun `does not report KDoc references with extension method calls`() {
+        val main = """
+            package com.example
+
+            import android.text.TextWatcher
+            import android.text.beforeTextChanged
```
BraisGabin, line 176: This is the false negative that you were talking about, right?

atulgpt: Hi @BraisGabin yes it is false positive. There are two which currenlty I can think of

```kolin
@Test
fun `false -ve #1`() {
    val main = """
        package com.example

        import android.text.TextWatcher
        import android.text.beforeTextChanged

        class TestClass {
            /**
             * [TextWatcher.beforeTextChanged]
             */
            fun test() {
                TODO()
            }
        }
    """.trimIndent()
    val additional1 = """
        package android.text

        class TextWatcher {
            fun beforeTextChanged() {}
        }
    """.trimIndent()
    val additional2 = """
        package android.text

        fun TextWatcher.beforeTextChanged() {}
    """.trimIndent()
    assertThat(subject.lintWithContext(env, main, additional1, additional2)).isEmpty()
}

@Test
fun `false -ve #2`() {
    val main = """
        package com.example

        import android.text.TextWatcher
        import android.text.beforeTextChanged

        import test.TextWatcher
        import test.beforeTextChanged
        
        class TestClass {
            /**
             * [TextWatcher.beforeTextChanged]
             */
            fun test() {
                TODO()
            }
        }
    """.trimIndent()
    val additional1 = """
        package android.text

        class TextWatcher
        class TextWatcher2
    """.trimIndent()
    val additional2 = """
        package android.text

        fun TextWatcher2.beforeTextChanged() {}
    """.trimIndent()
    assertThat(subject.lintWithContext(env, main, additional1, additional2)).isEmpty()
}
```

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImport.kt
atulgpt: Updated the code

cortinico:
@atulgpt can we makesure the website preview doesn't fail?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportSpec.kt
BraisGabin: Then I'll remove the import. We don't want a test that enforces a false positive. We could have an ignored test with the false positive. But, probably, it's better to just remove this import.

atulgpt: Yes I have updated the TC to make the TC a valid cass

cortinico (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportSpec.kt
```diff
@@ -141,6 +141,94 @@ class UnusedImportSpec(val env: KotlinCoreEnvironment) {
...
```
BraisGabin, line 176: To make this case valid shouldn't we remove this line? 

```suggestion
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportSpec.kt
```diff
@@ -141,6 +141,94 @@ class UnusedImportSpec(val env: KotlinCoreEnvironment) {
...
+        assertThat(subject.lintWithContext(env, main, additional1, additional2)).isEmpty()
+    }
+
+    @Test
+    fun `does report imported extension method which is not used`() {
+        val main = """
+            package com.example
+
+            import android.text.TextWatcher
+            import android.text.beforeTextChanged
```
BraisGabin, line 206: Same here

```suggestion
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportSpec.kt
atulgpt: No, it needs to be imported otherwise you will get the squiggly line in the Kdoc. See `class TextWatcher` doesn't have this method defined there

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImportSpec.kt
atulgpt: Same comment as above

BraisGabin (approved):

Issue #6106, opened 2023-05-18
Rule to disallow `iterable.any { it == <some_value> }` in favor of `iterable.contains(<some_value>)`


atulgpt:
## Expected Behavior of the rule
`iterable.any { it == <some_value> }` is similar to `iterable.contains(<some_value>)`. Using `iterable.any { it == <some_value> }` complicates the code as the user expects that usage of `any` might follow some complicated `predicate`.

We also have a similar rule in the `detekt` `UseSumOfInsteadOfFlatMapSize` which  enforces similar type of usages

BraisGabin:
I like it! I think it would be a really good addition.

What do you think about: `iterable.find { it == <some-value> } == null`? Should this rule also warn about that?

atulgpt:
Hi @BraisGabin I think there is already a rule `UseAnyOrNoneInsteadOfFind` which suggests `iterable.find { it == <some-value> } == null` to be written as `iterable.any { it == <some-value> }` and which in turn reported by this rule.
Although this would be two step process alternative would be to report this type of structure as well which will result in two violations for a single line of code(of which first one is unnccessary)

BraisGabin:
You are right. Forget about my comment.

atulgpt:
Hi @BraisGabin I have created the PR for the implementing this rule

PR #6571, opened 2023-10-29
UnnecessaryAny - Add rule to prevent usage of `any { }` when simple `contains` can be …

atulgpt:
Fixes https://github.com/detekt/detekt/issues/6106

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAnySpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAnySpec.kt
@@ -121,7 +121,7 @@ class UnnecessaryAnySpec(val env: KotlinCoreEnvironment) {
     @Test
     fun `does not report any with safe call which is used for checking presence of a element`() {
         val code = """
-            fun test(list: List<Int>, value: Int?) {
+            fun test(list: List<Int?>, value: Int?) {
                 list.any { value?.equals(it) == true }
             }
         """.trimIndent()
@@ -141,7 +141,7 @@ class UnnecessaryAnySpec(val env: KotlinCoreEnvironment) {
     }
 
     @Test
-    fun `does not report any when it is not used`() {
+    fun `does report any when it is not used`() {
         val code = """
             fun test(list: List<Int>, value: Int) {
                 list.any { value == 2 }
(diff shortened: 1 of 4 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAnySpec.kt
```diff
@@ -0,0 +1,425 @@
...
+    fun `does not report any with safe call which is used for checking presence of a element`() {
+        val code = """
+            fun test(list: List<Int>, value: Int?) {
+                list.any { value?.equals(it) == true }
+            }
+        """.trimIndent()
+        val actual = subject.compileAndLintWithContext(env, code)
+        assertThat(actual).isEmpty()
+    }
+
```
BraisGabin, line 131: I think that we could remove this test. I think that this code should be flagged but it seems so edge case to me that probably it doesn't need the effort. But we at least should don't have a test that say "this shouldn't be flagged".

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAnySpec.kt
```diff
@@ -0,0 +1,425 @@
...
+        val code = """
+            fun test(list: List<Int>, value: Int) {
+                list.any { value == 2 }
+                list.any { 2 == value }
+            }
+        """.trimIndent()
+        val actual = subject.compileAndLintWithContext(env, code)
+        assertThat(actual).isEmpty()
+    }
+
```
BraisGabin, line 154: I'm not 100% sure how but this is clearly an "UnnecessaryAny". If the lambda is not related with `it` you shouldn't use `any` for sure.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAnySpec.kt
atulgpt: Actually, this should not be reported by the rule. I made below example where usage of `any` and `contain` prints different thing

```kotlin
fun main() {
    val list = listOf(1,2,3, null)
    
    println("${test1(list, null)}, ${test2(list, null)}")
    println("${test1(list, 1)}, ${test2(list, 1)}")
}

fun test1(list: List<Int?>, value: Int?): Boolean {
    return list.any { value?.equals(it) == true }
}

fun test2(list: List<Int?>, value: Int?): Boolean {
    return list.contains(value)
}
```

Here list is of nullable type and the list contains null. So let me instead update the TC and make the `Int` to `Int?`

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAnySpec.kt
atulgpt: I have updated the code to detect when it is not used in the `any`

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryAnySpec.kt
```diff
@@ -0,0 +1,471 @@
...
+    @Test
+    fun `does report any when it is not used`() {
+        val code = """
+            fun test(list: List<Int>, value: Int) {
+                list.any { value == 2 }
+                list.any { 2 == value }
```
BraisGabin, line 148: We should change the error message in this type of errors, don't you think?

I don't know what we should recommend here. Probably just say that `any` shouldn't be used like this?

atulgpt: Hi, I have updated the msg as per msg used in the `UnnecessaryLet`. Updated msg is `any {  }` expression can be omitted`.

Also, fix some bugs which I found while doing this change.

BraisGabin (approved):
LGTM! Thanks!

cortinico:
@BraisGabin I also won't be able to pick this one as it's a new rule

PR #6484, opened 2023-09-13
Fixed documentation on `SarifOutputReport` + docs out `OutputReport::name`

lexa-diky:
Also fixed documentation for legacy versions. This behavior was at most introduced some time in 2020, so old docs where also incorrect

PR #6579, opened 2023-10-31
Fixed documentation on SarifOutputReport + docs out OutputReport::name

lexa-diky:
Also fixed documentation for legacy versions. This behavior was at most introduced some time in 2020, so old docs where also incorrect

copy of https://github.com/detekt/detekt/pull/6484 into main branch.


cortinico:
Could you update also the 1.23.2 version and make sure CI is not red?

BraisGabin (approved):
Maybe that PR already exists but we should also change the id for `2.0.0` to align it with the other reports.

PR #6590, opened 2023-11-01
Fix issues related to kotlin-stdlib's Gradle module metadata in Kotlin 1.9.20

3flex:
Gradle module metadata is published alongside kotlin-stdlib from 1.9.20 which caused a couple of dependency issues due to our use of [`useCompileClasspathVersions`](https://docs.gradle.org/8.4/javadoc/org/gradle/api/plugins/JavaResolutionConsistency.html#useCompileClasspathVersions--) for consistent dependency resolution.

This fixes them and will unblock 1.9.20 update.

### DIFF
--- a/detekt-gradle-plugin/build.gradle.kts
+++ b/detekt-gradle-plugin/build.gradle.kts
@@ -47,7 +47,6 @@ testing {
             useJUnitJupiter(libs.versions.junit.get())
 
             dependencies {
-                compileOnly("org.jetbrains:annotations:24.0.1")
                 implementation(libs.assertj)
                 implementation(testFixtures(project(":")))
             }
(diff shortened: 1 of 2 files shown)


detekt-gradle-plugin/build.gradle.kts
```diff
@@ -74,6 +73,7 @@ dependencies {
     compileOnly(libs.kotlin.gradle)
     compileOnly(libs.kotlin.gradlePluginApi)
     testFixturesCompileOnly("org.jetbrains:annotations:24.0.1")
+    compileOnly("org.jetbrains:annotations:24.0.1")
```
TWiStErRob, line 76: This went from test to production level, is this necessary for [useCompileClasspathVersions](https://docs.gradle.org/8.4/javadoc/org/gradle/api/plugins/JavaResolutionConsistency.html#useCompileClasspathVersions--)? or could it be `functionalTestCompileOnly`? Although I guess the test-suites syntax was doing exactly that.

TWiStErRob (approved):

detekt-gradle-plugin/build.gradle.kts
3flex: kotlin-stdlib pulls in version 13 of the annotations dependency, so if it's not explicitly upgraded by changing the dependency on the compile classpath, Gradle fails the build because it sees 13 on compileClasspath and 24 on functionalTestCompileClasspath.

PR #6592, opened 2023-11-01
Remove redundant association of functional test compilation to main compilation

3flex:
From Kotlin 1.9.20 test fixtures have access to internal declarations in main source set and test sources have access to internal declarations in test fixtures source set: https://kotlinlang.org/docs/whatsnew1920.html#support-for-test-fixtures-to-access-internal-declarations


3flex:
There are issues that require suppressing compiler errors in some files that for some reason didn't occur with the 1.9.20 RC that I'd tried earlier. I don't think that's the right path to go down.

PR #6594, opened 2023-11-01
Remove some unused/unnecessary code

3flex:


detekt-test-utils/api/detekt-test-utils.api
```diff
@@ -18,7 +18,6 @@ public final class io/github/detekt/test/utils/KotlinCoreEnvironmentWrapper {
 public final class io/github/detekt/test/utils/KotlinCoreEnvironmentWrapperKt {
 	public static final fun createEnvironment (Ljava/util/List;Ljava/util/List;)Lio/github/detekt/test/utils/KotlinCoreEnvironmentWrapper;
 	public static synthetic fun createEnvironment$default (Ljava/util/List;Ljava/util/List;ILjava/lang/Object;)Lio/github/detekt/test/utils/KotlinCoreEnvironmentWrapper;
-	public static final fun createPsiFactory ()Lorg/jetbrains/kotlin/psi/KtPsiFactory;
```
3flex, line 21: Technically a breaking change since API surface is different... but it was not documented or used anywhere so probably doesn't warrant a "breaking change" tag.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Signatures.kt
```diff
@@ -10,31 +10,13 @@ import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
...
- * PsiElements of type KtFile use the absolute path as their names. This resulted in
- * having two absolute paths being printed along with the output: one for the name
- * and one for the location. Consequently, IntelliJ was not hyperlinking them because
- * it only hyperlinks one path file per line.
- *
- * This gets the file name only, instead of the entire absolute path, and takes it as the element name.
- *
- * Example: KtFile with name /full/path/to/Test.kt will have its name formatted to be simply Test.kt
- */
-private fun String.formatElementName(): String =
```
marschwar, line 31: Why is this no longer relevant?

3flex: From the commit message:

> Since https://github.com/detekt/detekt/pull/6310 absolute paths are not stored in KtFile's `name` property, so
 this function is no longer required.

marschwar (approved):

PR #6609, opened 2023-11-07
Run build on JDK 21

### LINKED ISSUE #6591: Run CI on JDK 21
## Expected Behavior
CI runs for all LTS Java versions.

## Current Behavior
We currently run on Java 8, 11 and 17.

## Context
Keep CI current.

This is a placeholder issue raised while I remember it - blocked until we update to Kotlin 1.9.20 and Gradle 8.5.

3flex:
Closes #6591

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        jdk: [ 8, 11, 17 ]
+        jdk: [ 8, 11, 17, 21 ]
         exclude: # windows with JDK8 are *really* flaky
           - os: windows-latest
             jdk: 8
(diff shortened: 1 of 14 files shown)


Anksmahajan:
Gradle 8.5 is out now. Any plans on taking this PR forward?

3flex:
Yes

.github/workflows/pre-merge.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        jdk: [ 8, 11, 17 ]
+        jdk: [ 8, 11, 17, 21 ]
```
BraisGabin, line 25: Could we remove any of the others? We already have a lot of tasks I would prefer to don't add 3 more.

3flex: Agree but can we look at that separately?

PR #6611, opened 2023-11-08
Add "excludes" to detekt extension

mrclrchtr:
This PR adds the "excludes"-option to the DetektExtension and fixes the problem described here: https://github.com/detekt/detekt/discussions/6604

I tested it locally with the Kotlin compiler plugin.

Unfortunately, the compiler plugin cannot currently handle the "includes" flag, so I have left it out for now.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
@@ -20,6 +20,11 @@ interface DetektExtension {
 
     val source: ConfigurableFileCollection
 
+    /**
+     * List of glob patterns for files and directories to be excluded from the analysis.
+     */
+    val excludes: ListProperty<String>
+
     val baseline: RegularFileProperty
 
     val basePath: DirectoryProperty
--- a/detekt-gradle-plugin/api/detekt-gradle-plugin.api
+++ b/detekt-gradle-plugin/api/detekt-gradle-plugin.api
@@ -123,6 +123,7 @@ public abstract interface class io/gitlab/arturbosch/detekt/extensions/DetektExt
 	public abstract fun getDebug ()Lorg/gradle/api/provider/Property;
 	public abstract fun getDisableDefaultRuleSets ()Lorg/gradle/api/provider/Property;
 	public abstract fun getEnableCompilerPlugin ()Lorg/gradle/api/provider/Property;
+	public abstract fun getExcludes ()Lorg/gradle/api/provider/ListProperty;
 	public abstract fun getIgnoreFailures ()Lorg/gradle/api/provider/Property;
 	public abstract fun getIgnoredBuildTypes ()Lorg/gradle/api/provider/ListProperty;
 	public abstract fun getIgnoredFlavors ()Lorg/gradle/api/provider/ListProperty;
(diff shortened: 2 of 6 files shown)


schalkms (approved):

3flex:
The base branch should be changed to `main`. It may be cherry picked into the 1.x branch but that would be done separately.

mrclrchtr:
Ah ok, I first implemented it on main, but then realized that the code has changed in such a way that a cherry pick is probably not possible. 

But I'll adjust the PR and we'll see what we do afterwards.

BraisGabin:
Looking at your fix and reading your use case in #6604 probably we don't need to implement anything else. You can use this:

```kotlin
tasks.withType<Detekt>().configureEach {
    exclude("**/your/package/here/**")
}
```

This is not exactly the same as exclude the package because you could have all your files on `main/kotlin` instead of `main/kotlin/your/package/`. But the second is a very common convention so in general it should work fine.

This is documented here: https://detekt.dev/docs/gettingstarted/gradle#leveraging-gradles-sourcetask---excluding-and-including-source-files

mrclrchtr:
Yes, this would work great if I ran detekt with gradle. I tried it, but since I'm using the compiler plugin, these settings aren't applied at all, are they?

mrclrchtr:
The base branch is now `main`.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt
```diff
@@ -20,6 +20,11 @@ interface DetektExtension {
 
     val source: ConfigurableFileCollection
 
+    /**
+     * List of glob patterns for files and directories to be excluded from the analysis.
+     */
+    val excludes: ListProperty<String>
```
cortinico, line 26: I believe we should expose also `includes`

mrclrchtr: That's true, but unfortunately the compiler plugin does not currently support this, as far as I understand it. That's why I omitted it.

BraisGabin:
@3flex what do you think about this one? Should we merge or close? I'm not too familiar with the compiler plugin.

mrclrchtr:
As soon as you confirm that it can be merged, I will resolve the conflicts.

3flex:
I'm a bit torn on this one.

- Includes/excludes don't belong on extension level, since the extension shouldn't be used to configure sources. That's a task responsibility.
- More importantly this is solving a problem that's specific to a single, experimental method of executing detekt. Adding it to the extension makes that config option too high profile.
- There's the ability to configure excludes in the config file. If that's not working, that should be fixed and will resolve the issue, and is more appropriate.
- Compiler plugin won't work day 1 with K2 and will be stuck on language version 1.9 for an unknown amount of time, so fixes for the plugin are lower priority in my mind.

That said it's definitely convenient to exclude things like generated sources for all analysis tasks in one place, and it's a simple fix. I'm also doing rework for 2.0 DGP at the moment so might look at alternatives later. This will fix snapshots though.

mrclrchtr:
So is it currently not a good idea to use the compiler plugin, because of compatibility and low priority? I really liked the idea of using detekt in this way, but I think we are switching back to using gradle then.

3flex:
It's chicken & egg - if more people used it, it would be higher priority, but without some important fixes going in, it won't get used much.

We need to figure out how to make it compatible with K2 as well. When Kotlin 2.0 is released the compiler plugin likely won't be ready. So if upgrading to Kotlin 2.0 is a priority, I would move back to Gradle. I'm making some improvements to the Gradle plugin which will address several issues with it, but performance will never be as good as it could be if it was a compiler plugin.

BraisGabin:
Closing for the reasons explained above.

Anyway #6970 could end up implementing a way to exclude files from the yaml config level. I think that it would fulfil the use case that this PR is trying to fix.

Issue #6623, opened 2023-11-13
False Negative: Useless elvis is not reported by the `UnreachableCode`


atulgpt:
## Expected Behavior
```kotlin
@Test
fun `reports unreachable code after elvis`() {
    val code = """
        fun test(): Int {
            val a = 2 ?: run {
                3
            }
        }
    """.trimIndent()
    assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
}
```

Above TC should pass

## Observed Behavior
Above TC is failing

## Context
In my work project, we are using Detekt and have turned on the `UnreachableCode` but today I found that at some places we are using a useless elvis operator(which might get left there after some refactoring or making some variable from nullable to non null). Although this is also reported by the kotlin complier flag `allWarningsAsErrors = true` but in our project we only rely on detekt so it get slipped.

## Your Environment

* Version of detekt used: 1.23.0
* Version of Gradle used (if applicable):
* Gradle scan link (add `--scan` option when running the gradle task):
* Operating System and version:
* Link to your project (if it's a public repository):

PR #6624, opened 2023-11-13
Also check `USELESS_ELVIS` in `UnreachableCode`

atulgpt:
Fixes https://github.com/detekt/detekt/issues/6623

### DIFF
--- a/detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektAndroidSpec.kt
+++ b/detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektAndroidSpec.kt
@@ -721,7 +721,7 @@ private val SAMPLE_ACTIVITY_USING_VIEW_BINDING = """
     
         override fun onCreate(savedInstanceState: Bundle?) {
             binding = ActivitySampleBinding.inflate(LayoutInflater.from(this))
-            binding.sampleView ?: return
+            binding.sampleView
             setContentView(binding.root)
         }
     }
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnreachableCode.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnreachableCode.kt
@@ -43,7 +43,9 @@ class UnreachableCode(config: Config = Config.empty) : Rule(config) {
 
     override fun visitExpression(expression: KtExpression) {
         super.visitExpression(expression)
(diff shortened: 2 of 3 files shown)


3flex:
I understand why this is added to the existing unreachable code rule but I believe this should be implemented as a separate rule since it's a wrapper of an existing compiler warning. I don't think wrapped compiler warnings should be consolidated in the same rule.

And very technically speaking the Elvis operator itself is not unreachable - it will always be hit.

BraisGabin:
> I understand why this is added to the existing unreachable code rule but I believe this should be implemented as a separate rule since it's a wrapper of an existing compiler warning. I don't think wrapped compiler warnings should be consolidated in the same rule.

I'm not sure about this. It's true, it's easy to think "two different rules" looking at the implementation. But, as a user, both are unreachable code so I think that it is ok to merge them. A new rule will help to make this more configurable but I think that we don't need such configuration.

detekt-gradle-plugin/src/functionalTest/kotlin/io/gitlab/arturbosch/detekt/DetektAndroidSpec.kt
```diff
@@ -721,7 +721,7 @@ private val SAMPLE_ACTIVITY_USING_VIEW_BINDING = """
     
         override fun onCreate(savedInstanceState: Bundle?) {
             binding = ActivitySampleBinding.inflate(LayoutInflater.from(this))
-            binding.sampleView ?: return
+            binding.sampleView
```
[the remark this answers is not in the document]
schalkms: I'm not sure about the return.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnreachableCode.kt
```diff
@@ -43,7 +43,9 @@ class UnreachableCode(config: Config = Config.empty) : Rule(config) {
 
     override fun visitExpression(expression: KtExpression) {
         super.visitExpression(expression)
-        if (bindingContext.diagnostics.forElement(expression).any { it.factory == Errors.UNREACHABLE_CODE }) {
+        if (bindingContext.diagnostics.forElement(expression)
+                .any { it.factory in listOf(Errors.UNREACHABLE_CODE, Errors.USELESS_ELVIS) }
```
schalkms, line 47: We shouldn't create a list each and every time. For two elements it's better to check with an OR `||`.

atulgpt: Although I personally prefer list type of usage as that scales well with more checks. But I am okay with using `||`. I have updated the code

BraisGabin (approved):

PR #6627, opened 2023-11-13
Ignore unresolved classes in UnreachableCatchBlock

### LINKED ISSUE #6626: UnreachableCatchBlock gives false positive for unresolved classes
## Expected Behavior
No UnreachableCatchBlock warnings should be triggered for classes that could for some reason not be resolved.

## Observed Behavior
UnreachableCatchBlock treats each unresolved class as a subclass of each other triggering warnings.

## Steps to Reproduce
```
fun test() {
    try {
    } catch (e: Foo) {
    } catch (e: Bar) {
    }
}
```
Running detekt with type resolution enabled but Foo and Bar missing from classpath.
`detekt-cli -cp . -i test.kt`
Output (I've removed other warnings for clarity):
```
test.kt:10:7: This catch block is unreachable. [UnreachableCatchBlock]
```

## Context
Getting the classpath for type resolution correct for detekt is not easy and getting unrelated warnings because a type doesn't resolve isn't helping.

## Your Environment

* Version of detekt used: 1.23.3
* Operating System and version: Lnux 6.1.61 x86_64

thejk:
Kotlin compiler uses ErrorClassDescriptor for classes that doesn't resolve correctly. They all are subclasses of each other causing the check in UnreachableCatchBlock to consider all them unreachable.

Ignore ErrorClassDescriptor's to not add most likely false warnings as we can't know if the classes are subclasses or not.

Fixes #6626

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnreachableCatchBlock.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnreachableCatchBlock.kt
@@ -14,6 +14,7 @@ import org.jetbrains.kotlin.psi.KtTryExpression
 import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.descriptorUtil.isSubclassOf
+import org.jetbrains.kotlin.types.error.ErrorClassDescriptor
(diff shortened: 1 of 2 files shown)


3flex:
This isn't the right approach to resolving your issue in my opinion.

The rule requires type resolution and this requires the classpaths to be setup correctly. The focus should be on making it simpler to set up the classpath correctly and pass it to detekt, not on excluding findings when that hasn't been done.

cortinico (dismissed):

thejk:
Ok, cleaned up the code before closing.

Issue #4140, opened 2021-09-30
1.18.1: Bazel based test runs fail in the `KtTestCompiler`


davidburstromspotify:
When using Bazel as a test runner for Detekt extensions, the tests fail with

```
    => java.lang.ExceptionInInitializerError
       io.github.detekt.test.utils.CompileExtensionsKt.compileContentForTest(CompileExtensions.kt:14)
       io.github.detekt.test.utils.CompileExtensionsKt.compileContentForTest$default(CompileExtensions.kt:10)
       io.gitlab.arturbosch.detekt.test.RuleExtensionsKt.lint(RuleExtensions.kt:31)
       io.gitlab.arturbosch.detekt.test.RuleExtensionsKt.compileAndLint(RuleExtensions.kt:27)
       [...]
     Caused by: java.nio.file.FileSystemNotFoundException
       jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.getFileSystem(ZipFileSystemProvider.java:169)
       jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.getPath(ZipFileSystemProvider.java:155)
       java.base/java.nio.file.Path.of(Path.java:208)
       java.base/java.nio.file.Paths.get(Paths.java:97)
       io.github.detekt.test.utils.ResourcesKt.resourceAsPath(Resources.kt:21)
       [...]
```

The difference between Gradle and Bazel in this case is that Gradle provides the test classes as file paths, whereas Bazel provides them in a JAR.

Since we're not using resource based tests in our consuming project, it's easy enough to fix this by deferring the `root` path calculation.

```diff
--- a/detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KtTestCompiler.kt
+++ b/detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KtTestCompiler.kt
@@ -23,7 +23,7 @@ import java.nio.file.Path
  */
 internal object KtTestCompiler : KtCompiler() {
 
-    private val root = resourceAsPath("/")
+    private val root by lazy { resourceAsPath("/") }
 
     fun compile(path: Path) = compile(root, path)
 
```

The question is, if we fix this, when might this be available in an upstream release?

cortinico:
> The difference between Gradle and Bazel in this case is that Gradle provides the test classes as file paths, whereas Bazel provides them in a JAR.
[...]

PR #6636, opened 2023-11-19
Follow-up #6572: Re-order test classpath to load correct Logger as workaround for KT-60813

TWiStErRob:
* Follow-up on https://github.com/detekt/detekt/pull/6572/files#r1386261829

This solution replaces the slf4j Logger with the correct one by classpath ordering: c4ef7372bd8bec71c20d864e2d0c170b2ab3c00d

I ran into #4140/#4170 not being implemented fully (current implementation is based on assumptions). The assumption is that the `/` resource on the classpath points to the current module's test sourceSet. By changing the order I changed this and now the slf4j.jar is on the classpath first, which triggers #4140. I was able to workaround this by looking for a specific file that's in the correct folder (ba150dbc3654f4392b06d065f2957a4674770448), but it's not scalable, so I instead changed the logic to insert the new dependency in the middle of the classpath: 68352ceeeb82b9d5c99637718ed9ef6b9797436f

### DIFF
--- a/detekt-formatting/build.gradle.kts
+++ b/detekt-formatting/build.gradle.kts
@@ -3,6 +3,8 @@ plugins {
 }
 
 val extraDepsToPackage: Configuration by configurations.creating
+val testRuntimeOnlyPriority: Configuration = configurations.resolvable("testRuntimeOnlyPriority").get()
+sourceSets.test.configure { runtimeClasspath = testRuntimeOnlyPriority + runtimeClasspath }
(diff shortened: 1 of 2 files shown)


schalkms (approved):

3flex:
Kotlin 1.9.21 fixes KT-60813 so I've opened #6646 to remove the workaround introduced in #6636. This PR isn't required anymore.

PR #6646, opened 2023-11-23
Remove workaround for KT-60813

3flex:
The issue was fixed in Kotlin 1.9.21.


TWiStErRob (approved):
Awesome!

For the record 1.9.21 was merged here: https://github.com/detekt/detekt/pull/6642

PR #6647, opened 2023-11-24
Update documentation on how to pin detekt to a supported kotlin version

mlanglet:


website/docs/gettingstarted/gradle.mdx
```diff
@@ -515,15 +515,14 @@ This happens when the version of `kotlin-compiler-embeddable` is overridden on d
 dependency configuration). This can happen when build scripts use things like this which override all dependency
 configurations (e.g. by [Spring Dependency Management Plugin](https://docs.spring.io/dependency-management-plugin/docs/current/reference/html/):
 
+You can pin the detekt version to a supported one by adding the following to your build gradle file:
+
 ```kotlin
-configurations.all {
-    resolutionStrategy.eachDependency {
+project.afterEvaluate {
+    configurations["detekt"].resolutionStrategy.eachDependency {
```
cortinico, line 522: I'm unsure this is correct, where you got it from?

3flex: Agree with this comment. Using `afterEvaluate` is not recommended on general and should not be used as an example in our documentation.

3flex (changes requested):
See inline comments.

website/docs/gettingstarted/gradle.mdx
mlanglet: I see, thanks for your feedback!

I found the first suggested workaround in [this ](https://github.com/detekt/detekt/issues/6198) issue and applied it to our project because renovatebot was making me go crazy with the failed builds. It fixed it for our project and I thought I'd update the docs linked to from the error during the build since it didn't help me resolve the issue, to help others from not having to dig through issues to find a workaround.

Clearly I should have read the rest of the issue as it is now apparent that documentation for this workaround should be changed and scoped to situation at hand, ie. that the combination of the gradle spring dependencies and spring boot plugin causes an override of the detekt kotlin version. 

I'll update the PR accordingly. My main goal was to make a workaround available at the first site people land on when clicking the link in the failed build log.

Issue #6630, opened 2023-11-16
New rule: all sealed class children must be data classes.


VovaStelmashchuk:
Add a new rule all children of `sealed class` must be data classes.  

The sealed class is usually used to describe the system or feature states. 

Also, the rule prevents the creation of an open class as a child for the `sealed class`, it reduces the count of the potential of unnecessary abstraction in the system.

Also, the rule fixes the potential bug with comparing states because the data classes have auto-generated equals and hashcode methods.

## Expected Behavior of the rule
Correct code:
```kotlin
sealed class Foo {
    object Bar : Foo()
    data class Baz(val i: Int) : Foo()
}

```

Wrong code:
```kotlin
sealed class Foo {
    object Bar : Foo()
    open class Baz : Foo()

    open class Baz2(val test: Int) : Baz()
}
```

BraisGabin:
To be more specific (because its the only way to implement the rule). A `sealed class` can only be extended by a:

- `data class`
- `data object` (Better `toString` than a regular `object`)
- `sealed class` (The "tree" doesn't need to have one single layer)

And, what about the `sealed inerfaces`? Right now I'm not in favor nor against of this rule but we need to make it more specific to decide what to do with it.

VovaStelmashchuk:
Yep, you catch the point.
`value class` is not the case. A value class cannot extend other classes. 

About `sealed interface` not sure we need to handle the case because it looks pretty standard for me to have an open class as a child of a sealed interface. 

Having a normal class as the children for the sealed class does not look good for me because the developers usually use the children of the sealed class to describe the states. The state classes are often a part of some checkers, like an equal or contains, etc.

```kotlin
val queue = PriorityQueue<SealedClass>()

if (queue.contains(foo)) {
//Do something 
}
```

In case foo is an instance of a normal class without implementing equals and hashcode the contains login doesn't work as expected.

cortinico:
> it reduces the count of the potential of unnecessary abstraction in the system.

I'm not super convinced by this rule. What if I'm in a constrained environment and I don't want to use a `data class` but just a `class` to don't have to deal with all the extra methods generated for data classes?

VovaStelmashchuk:
I didn't think to include this rule by default. My idea is to have this rule excluded by default. 

The developers be able to enable it for all classes that are responsible for UiState or just for all sealed classes or whatever. 

If it's not consistent with the detekt rules, it's totally ok for me. I can wrap the rule as a detekt plugin and maintain the project myself.

[...]

PR #6648, opened 2023-11-24
Implement new rule UnsafeSealedClassChild

VovaStelmashchuk:
The rule discussion [here](https://github.com/detekt/detekt/issues/6630)

Rule: UnsafeSealedClassChild

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnsafeSealedClassChild.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnsafeSealedClassChild.kt
@@ -6,6 +6,7 @@ import io.gitlab.arturbosch.detekt.api.Debt
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution
 import org.jetbrains.kotlin.cfg.WhenChecker
 import org.jetbrains.kotlin.descriptors.isInterface
 import org.jetbrains.kotlin.descriptors.isSealed
@@ -35,6 +36,7 @@ import org.jetbrains.kotlin.types.typeUtil.supertypes
  * }
  * </compliant>
  */
+@RequiresTypeResolution
 class UnsafeSealedClassChild(ruleSetConfig: Config = Config.empty) : Rule(ruleSetConfig) {
 
     override val issue: Issue = Issue(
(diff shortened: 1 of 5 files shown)


detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnsafeSealedClassChild.kt
```diff
@@ -0,0 +1,76 @@
...
+ * </noncompliant>
+ *
+ * <compliant>
+ * sealed class Foo {
+ *     object Bar : Foo()
+ *     data class Baz(val i: Int) : Foo()
+ * }
+ * </compliant>
+ */
+class UnsafeSealedClassChild(ruleSetConfig: Config = Config.empty) : Rule(ruleSetConfig) {
```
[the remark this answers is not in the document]
cortinico: Please fix this one

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnsafeSealedClassChildSpec.kt
```diff
@@ -0,0 +1,115 @@
...
+            data class Bar3(override val i : Int) : ChildFoo(i)
+            data object Bar5 : Foo()
+        """.trimIndent()
+
+        val actual = subject.compileAndLintWithContext(env, code)
+        assertThat(actual).hasSize(1)
+    }
+
+    @Test
+    fun `check sealed interface not affect by the rule`() {
```
cortinico, line 104: can you change this into "does not report sealed interface..."

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnsafeSealedClassChildSpec.kt
```diff
@@ -0,0 +1,115 @@
...
+            //The wrong class
+            class Bar2 : Foo()
+        """.trimIndent()
+
+        val actual = subject.compileAndLintWithContext(env, code)
+        assertThat(actual).hasSize(1)
+    }
+
+    @Test
+    fun `non report sealed class has object as child outside sealed body`() {
```
cortinico, line 45:
```suggestion
    fun `does not report sealed class has object as child outside sealed body`() {
```

applies to all the functions

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnsafeSealedClassChild.kt
```diff
@@ -0,0 +1,76 @@
...
+            .any { kotlinType -> isSealedClass(kotlinType) }
+
+        if (isChildOfSealedClass) {
+            val isNormalSealedChild = klass.isData() || klass.isSealed()
+            if (!isNormalSealedChild) {
+                report(
+                    CodeSmell(
+                        issue,
+                        Entity.from(klass),
+                        "The ${klass.name} must be data class or object because its child of sealed class"
```
cortinico, line 62:
```suggestion
                        "The ${klass.name} must be data class or object because it's a subclass of a sealed class"
```

VovaStelmashchuk:
I think the pr is ready to review :)

BraisGabin:
Before reviewing the PR itself I would like to know what other @detekt/maintainers think about adding this rule to the default rule set.

To me this rule leans to the really controversial ones. For that reason I think that this one should be published as a third party rule and add it to our [marketplace](https://detekt.dev/marketplace/) instead of our default rule set.

cortinico:
> Before reviewing the PR itself I would like to know what other @detekt/maintainers think about adding this rule to the default rule set.
> 
> To me this rule leans to the really controversial ones. For that reason I think that this one should be published as a third party rule and add it to our [marketplace](https://detekt.dev/marketplace/) instead of our default rule set.

I've shared my concerns on the linked issue. There might be scenarios where this rule doesn't help the developer but for most of the use cases, is probably valid. Curious to hear other's opinion as well

marschwar:
I agree with the comments above. The example in the [official documentation](https://kotlinlang.org/docs/sealed-classes.html) uses non-data classes.

BraisGabin:
So, I think that we agree that we don't want this rule on the default rule set of detekt. As a rule it has sense but it is best suited on a third party rule set. Closing for that reason.

PR #6655, opened 2023-11-27
Use infix style to apply plugins

Goooler:
It works for both Groovy and Kotlin dsl.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,6 +12,7 @@ indent_size = 4
 indent_style = space
 indent_size = 2
 
-[*.md]
+[{*.md,mdx}]
 indent_style = space
 indent_size = 4
+trim_trailing_whitespace = false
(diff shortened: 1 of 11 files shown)


.editorconfig
```diff
@@ -12,6 +12,7 @@ indent_size = 4
 indent_style = space
 indent_size = 2
 
-[*.md]
+[{*.md,mdx}]
 indent_style = space
 indent_size = 4
+trim_trailing_whitespace = false
```
Goooler, line 18: Maybe we should just apply this for `*.md` only.

3flex: Is this required for this change?

Goooler: Actually nope. I edited these files on Visual Studio Code for the Web, formats changed, so I added this rule.

cortinico (approved):

.editorconfig
3flex: Thanks, I've reverted that commit in this PR.

3flex (approved):

Issue #6676, opened 2023-12-02
False positive CastToNullableType on safe typealias cast


TWiStErRob:
## Expected Behavior
No report, it's a valid and equivalent cast.

## Observed Behavior
```
Use the safe cast ('as? Alias') instead of 'as Alias?'. [CastToNullableType]
```
The cast is very safe, it's so safe in fact, that Kotlin says
```
> Task :compileKotlin
w: ... No cast needed
```
This is why the repro suppresses.

## Steps to Reproduce
```kotlin
typealias Alias = String

fun f(s: String?) {
	@Suppress("USELESS_CAST") // Casts is useful to convert String to typealias.
	val a = s as Alias?
	val l /*List<Alias?>*/ = listOf(a)
	println(l)
}
```
The reason for the cast is to get the hint in the comment. This is a trivial example, the benefit is not easily visible, see below for more.

## Context
I have some complex logic chain where there's a lot of type changes, because of map/filter logic.
I also use some external types which contain string properties representing different concepts.
I created some private aliases to help understand the code and I'm casting where necessary to see the alias types in the type hints.
![image](https://github.com/detekt/detekt/assets/2906988/59689856-67fc-41fc-8567-a45b21a7e082)

## Your Environment

* Version of detekt used: 1.23.4
* Version of Gradle used (if applicable): 8.5
* Gradle scan link (add `--scan` option when running the gradle task): N/A
* Operating System and version: any
* Link to your project (if it's a public repository): N/A, full repro above

atulgpt:
@TWiStErRob fixing this would require converting this rule to `@RequiresTypeResolution`

Also, I think the below failing TCs should be fixed

```kotlin
// https://github.com/detekt/detekt/issues/6676
@Test
fun `cast to same type in alias form allowed`() {
    val code = """
        typealias Alias = String

        fun test(s: String?) {
            @Suppress("USELESS_CAST") // Casts is useful to convert String to typealias.
            val a = s as Alias?
            print(a)
        }
        """.trimIndent()
    val findings = subject.compileAndLint(code)
    assertThat(findings).isEmpty()
}

@Test
fun `cast to parent type in alias form allowed`() {
    val code = """
        typealias Alias = CharSequence

        fun test(s: String?) {
            val a = s as Alias?
            print(a)
        }
        """.trimIndent()
    val findings = subject.compileAndLint(code)
    assertThat(findings).isEmpty()
}

@Test
fun `cast to different type in alias form not allowed`() {
[...]

PR #6677, opened 2023-12-02
Fix `CastToNullableType` warning on same or parent type

atulgpt:
Fixes: https://github.com/detekt/detekt/issues/6676

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastToNullableType.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/CastToNullableType.kt
@@ -5,9 +5,12 @@ import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
 import io.gitlab.arturbosch.detekt.api.Rule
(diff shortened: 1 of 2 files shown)


cortinico:
Do we want this on the 1.x? It's quite of an invasive change as the rule will stop working for users that are nto using TR

cortinico:
> @cortinico I do not have any hard preferences here but speaking from the past data we have made these types of changes even in detekt releases like 1.x to 1.(x+1)

Yup but this will end up in 1.23.x+1

BraisGabin (approved):

atulgpt:
> > @cortinico I do not have any hard preferences here but speaking from the past data we have made these types of changes even in detekt releases like 1.x to 1.(x+1)
> 
> Yup but this will end up in 1.23.x+1

Got your point 😅 . Since it's patch version I think we should not

TWiStErRob:
Why are we limiting to 1.23.x+1? Is there anything wrong with releasing 1.24?

cortinico:
> Why are we limiting to 1.23.x+1? Is there anything wrong with releasing 1.24?

Yes. If so we would have to ship all the new rules in `1.24` and I don't think we have the capacity to do so. Therefore we committed to ship only bugfixes for 1.23.x and version bumps (as those are ecosystem blockers)

TWiStErRob:
Ah, I see, thanks. Although the "have to ship all the new rules" seems like a self-imposed requirement. Anyway, makes sense to focus on backporting important fixes only!

PR #6683, opened 2023-12-04
Remove reports from detekt-core dependencies

3flex:
Refactor to remove the implementation dependency on detekt-reports-* from detekt-core.


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt
```diff
@@ -1,24 +1,15 @@
...
-internal fun defaultReportMapping(reportId: String) = when (reportId) {
-    TxtOutputReport::class.java.simpleName -> "txt"
-    XmlOutputReport::class.java.simpleName -> "xml"
-    HtmlOutputReport::class.java.simpleName -> "html"
-    SarifOutputReport::class.java.simpleName -> "sarif"
-    MdOutputReport::class.java.simpleName -> "md"
-    else -> reportId
-}
+internal fun defaultReportMapping(report: OutputReport) =
+    if (report is BuiltInOutputReport) report.ending else report.id
```
BraisGabin, line 12: As part of 2.0 we can use `report.ending` for all the implementations, right? And if it has a default value or something right now we can just remove it and force it to be defined.

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/reporting/OutputReportsSpec.kt
```diff
@@ -40,46 +39,45 @@ class OutputReportsSpec {
         @Test
         fun `it should properly parse XML report entry`() {
             val xmlReport = reports[0]
-            assertThat(xmlReport.type).isEqualTo(defaultReportMapping(XmlOutputReport::class.java.simpleName))
+            assertThat(xmlReport.type).isEqualTo(defaultReportMapping(XmlOutputReport()))
```
BraisGabin, line 42: These test were here before so no need to change in this PR but they are really strange.

It seems that they are testing the `ReportsSpecBuilder`. But on the expectrd part they also call to the production function ``. So I'm not sure what are we testing here if we are testing anything at all.

Does `defaultReportMapping` have tests? (if we implement my other comment this function will not even exist). And I would change the call here to the hardcoded value that we really expect.

(I'm reviewing on my mobile so I can't check the source code)

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/reporting/Reporting.kt
3flex: I don't know much about the reporting code in detekt so I can't say. A potential issue of using `ending` instead of `id` is that there could be multiple report types with the same extension.

API changes could be done in a follow up, this was just about decoupling the projects a little.

BraisGabin (approved):

PR #6684, opened 2023-12-04
Remove unused dependencies

3flex:


BraisGabin (approved):
<3 should we configure https://github.com/autonomousapps/dependency-analysis-gradle-plugin? Probably we don't want it on ci because it is slow but at least to be able to run it easily and spot this things.

3flex:
The plugin requires JDK 11 last time I checked so we currently can't include it in the build because we run on 8 for CI. We can use tool chains but I don't think it's worth introducing just for this. It's easy enough to apply locally.

Issue #6410, opened 2023-08-18
Enforce runTest in junit test when coroutine is used


tyvsmith:
## Expected Behavior of the rule
I suggest a checker for test code to enforce that runTest is used on a junit @Test method when the underlying code launches a coroutine.

Invalid code
```
@Test
fun test() {
  methodThatLaunchesCoroutine()
}
```

Valid Code
```
@Test
fun test() = runTest {
  methodThatLaunchesCoroutine()
}
```

## Context
Coroutines 1.7 has started propagating previously suppressed exceptions via runTest. During our upgrade from 1.6, we have found several cases where previously swallowed exceptions are now breaking tests. Often this happens on an unrelated test because the previous test didn't use runTest when dealing with underlying code that launched a coroutine, making them more difficult to find and resolve.

- https://github.com/Kotlin/kotlinx.coroutines/pull/3449
- https://github.com/Kotlin/kotlinx.coroutines/issues/3821#issuecomment-1647851085
- https://github.com/Kotlin/kotlinx.coroutines/issues/1205#issuecomment-1238261240

cortinico:
Yup this would make for a great addition to the `coroutines` ruleset

Rai-Sahil:
Hi, If no one's planning on working on it. I could get started

marschwar:
Thank you. I assigned this issue to you.

dmitriev2000:
Any news on this one? :)

BraisGabin:
I unasigned Rai from now so this PR is ready to grab by anyone.

BraisGabin:
Go for it! I just assigned it to you.

PR #6692, opened 2023-12-06
Add CoroutineLaunchedInTestWithoutRunTest rule

paolorotolo:
As described in #6410, this rule enforces that jUnit tests that launch coroutines are run in a `runTest` block to avoid exception swallowing.

To do so:
 * A new rule `CoroutineLaunchedInTestWithoutRunTest.kt` has been created
 * New tests are available in `CoroutineLaunchedInTestWithoutRunTestSpec.kt`
 * `kotlinx.coroutines.test` (required for `kotlinx.coroutines.test.runTest`) and `junit` (required for `org.junit.Test`) has been added as `testRuntimeOnly` dependencies of `detekt-rules-coroutines`.


detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
```diff
@@ -0,0 +1,93 @@
...
+ * }
+ * </compliant>
+ *
+ */
+@RequiresTypeResolution
+class CoroutineLaunchedInTestWithoutRunTest(config: Config) : Rule(config) {
+    override val issue = Issue(
+        id = "CoroutineLaunchedInTestWithoutRunTest",
+        description = "Launching coroutines in tests without a `runTest` block could swallow exceptions" +
+            "You should use `runTest`to avoid altering test results."
```
cortinico, line 53:
```suggestion
        description = "Launching coroutines in tests without a `runTest` block could swallow exceptions. " +
            "You should use `runTest`to avoid altering test results."
```

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    )
+
+    override fun visitNamedFunction(function: KtNamedFunction) {
+        checkAndReport(function)
+    }
+
+    private fun checkAndReport(function: KtNamedFunction) {
```
cortinico, line 60: Is this separate function needed?

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTestSpec.kt
```diff
@@ -0,0 +1,146 @@
...
+                    }
+                    
+                    Unit
+                }
+            }
+        """.trimIndent()
+        assertThat(subject.compileAndLintWithContext(env, code)).hasSize(1)
+    }
+
+/*
```
cortinico, line 60: did you intend to uncomment this one?

detekt-rules-coroutines/build.gradle.kts
```diff
@@ -7,4 +7,6 @@ dependencies {
     testImplementation(projects.detektTest)
     testImplementation(libs.assertj)
     testRuntimeOnly(libs.kotlinx.coroutines)
+    testRuntimeOnly(libs.kotlinx.coroutines.test)
+    testRuntimeOnly(libs.junit)
```
cortinico, line 11: Why do you need this?

paolorotolo: I'm new to Detekt, but I noticed I need to resolve dependencies to make type resolution work in tests. I need the packages `kotlinx.coroutines.test` (required to solve `kotlinx.coroutines.test.runTest`) and `junit` (required for `org.junit.Test` annotation)

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTestSpec.kt
paolorotolo: Yeah, there are some cases not covered by the rule, like the one of this test. I've submitted this as a Draft PR to make sure the base logic passed checks.
I will probably ask on #detekt in KotlinLang, but I didn't find an immediate way to traverse and check all the functions called from the @Test function explored by `visitNamedFunction`.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
paolorotolo: this is WIP, I guess I'll have to check all the functions called from a @Test function in future, so I wanted to extract that logic.

paolorotolo:
This is now ready for review.

detekt-rules-coroutines/build.gradle.kts
cortinico: > and junit (required for org.junit.Test annotation)

This is the part that I believe we should refine a bit. While it makes sense we depend on `libs.kotlinx.coroutines.test` to do TR for those specific tests, I don't think we should depend on `junit`.

Specifically you should check check if the annotation is called `Test`, regardless of its fully qualified name
(and that's what you actually do in your rule, no?). So I believe we should be able to just remove this dependency.

gradle/libs.versions.toml
```diff
@@ -35,6 +37,7 @@ slf4j-simple = { module = "org.slf4j:slf4j-simple", version.ref = "slf4j" }
 slf4j-api = { module = "org.slf4j:slf4j-api", version.ref = "slf4j" }
 
 junit-api = { module = "org.junit.jupiter:junit-jupiter-api", version.ref = "junit" }
+junit = "junit:junit:4.13.2"
```
cortinico, line 40: If we keep it, this should follow the same structure as the other deps

detekt-rules-coroutines/build.gradle.kts
paolorotolo: Yes, I've updated the PR removing the dependency, tests are still passing. Thanks!

paolorotolo:
@cortinico it seems `org.junit.Test` is needed to make test snippets compilable. Sorry, I didn't run `compile-test-snippets` task locally to catch it before.
Should we reintroduce the dependency?

gradle/libs.versions.toml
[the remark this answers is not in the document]
paolorotolo: Nice idea, I've done the latter in the latest commit

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
```diff
@@ -0,0 +1,114 @@
...
+        val functions = mutableSetOf<KtNamedFunction>()
+
+        collectDescendantsOfType<KtExpression>().mapNotNull {
+            it.getResolvedCall(bindingContext)
+                ?.resultingDescriptor
+                ?.source
+                ?.getPsi() as? KtNamedFunction
+        }.forEach {
+            functions.add(it)
+            functions.addAll(it.traverseAndGetAllCalledFunctions())
```
BraisGabin, line 100: A recursive function would create an infinite loop, right? 

Even if we fix that, which is the performance of this rule? This seems like something really expensive. 

Should we create a `Map<KtFunctionNamed, Boolean>` to store if a function (or its children) launch a coroutine? This way multiple tests that call the same function will only traverse the tree once. Also this should help to avoid the infinite loop because you have the information about the already visited functions.

cortinico: Agree on this 👍

paolorotolo:
Thanks for the review, last commit:
- Introduces 2 new test cases agains recursive functions;
- Maintains a cache of explored functions used to:
  - Skip discovering of children function: `traverseAndGetAllCalledFunctionsIfNecessary` stops traversing the function tree if the examined function is already present in the cache. If not, function is added in the cache with the status `FunLaunchesCoroutine.UNKNOWN`
  - Skip checking for coroutine launches if a fun is present in the cache with status `FunLaunchesCoroutine.TRUE` or `FunLaunchesCoroutine.FALSE`.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
```diff
@@ -0,0 +1,148 @@
...
+                    exploredFunctionsCache[function] = TRUE
+                    report(CodeSmell(issue, Entity.from(function), MESSAGE))
+                    return
+                } else {
+                    exploredFunctionsCache[function] = FALSE
+                }
+            }
+            TRUE -> {
+                report(CodeSmell(issue, Entity.from(function), MESSAGE))
+                return
```
BraisGabin, line 89: In general I don't care too much about code coverage but I think that in this case a test to test this code would be great. To ensure that the `cache` works.

BraisGabin (commented):
If you can add a test to ensure that the `cache` works (comment below) I think this is ready to merge.

paolorotolo:
Actually, I've found another way to simplify caching.
There's no need to store if a function launches a coroutine or not. We can just check only unexplored functions.
Last commit contains a `FunTraverseHelper` that starts from an `initialFunction` and returns all the unexplored nested function calls. If another `@Test` calls already explored functions, `FunTraverseHelper` will return only the new unexplored calls (there's a new test for this).

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
```diff
@@ -0,0 +1,136 @@
...
+        if (!initialFunction.hasAnnotation(TEST_ANNOTATION_NAME)) return
+        if (initialFunction.runsInRunTestBlock()) return
+
+        checkAndReport(initialFunction)
+
+        funTraverseHelper
+            .getAllUnexploredCalledFunctions(initialFunction, bindingContext)
+            .forEach {
+                checkAndReport(it)
+            }
```
BraisGabin, line 70: This implemantation now could report the same function more than once, right?

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private fun checkAndReport(function: KtNamedFunction) {
+        if (function.isLaunchingCoroutine()) {
+            report(CodeSmell(issue, Entity.from(function), MESSAGE))
+            return
```
BraisGabin, line 76: This `return` does nothing here. Probably this is related with my previous comment.

BraisGabin (commented):
🤔 Now if I have two `@Test` functions that call the same `function` only the first `@Test` function will have the issue, right? Shouldn't we report both?

paolorotolo:
> 🤔 Now if I have two `@Test` functions that call the same `function` only the first `@Test` function will have the issue, right? Shouldn't we report both?

The issue will be reported once: even previous implementations of this rule reported issues at suspend function launch site. So in this case it makes sense to flag a function launching a coroutine only once.
If this isn't intended behaviour, we should rework part of the reporting logic to report `@Test` functions instead and yes, in this case, I would expect to have both tests reported. Probably we'll need to rollback to the previous caching implementation.

@BraisGabin what happens behind the scenes when I call `report` from a rule? Does the PSI traversing continue? Can I report other subsequent issues?

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
paolorotolo: no, because `getAllUnexploredCalledFunctions` will return a function only once.

BraisGabin:
> @BraisGabin what happens behind the scenes when I call `report` from a rule? Does the PSI traversing continue? Can I report other subsequent issues?

Behind the scenes a `Finding` is added to a list of `List<Finding>`. And that's it. If you call multiple times `report` multiple `Finding`s are added and all of them are added :). Later detekt will take that list and process the findings.

---

About the issues. I think that we should always point to the `@Test` function. Because the issue is there. Even if we find the usage of `launch` in other funtion the issue is not in that function. The issue is in the `@Test` function that doesn't use `runTest`, right? At this moment this is not the case. If the `launch` is inside the `@Test` we report the issue there, but if it is not we report the issue on the other function.

So yes, I think that just with a "visited functions" we can't implement this.

paolorotolo:
Please wait reviewing this, I've found that all cache iterations fail with function deeper than two levels.
That's because parent functions are not updated once a child has been found launching a coroutine.

We may need to migrate the cache data structure to something like a tree to keep track of function relationship.

paolorotolo:
Now while exploring the functions, a list of parents is maintained which gets updated each time a child is flagged.

BraisGabin (approved):

BraisGabin:
@detekt/maintainers it would be good if someone else can review this rule. It has a complex traversal logic so I would like someone else's eyes on it.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
```diff
@@ -0,0 +1,147 @@
...
+    override val issue = Issue(
+        javaClass.simpleName,
+        description = "Launching coroutines in tests without a `runTest` block could swallow exceptions. " +
+            "You should use `runTest` to avoid altering test results."
+    )
+
+    override fun visitNamedFunction(initialFunction: KtNamedFunction) {
+        if (!initialFunction.hasBody()) return
+        if (!initialFunction.hasAnnotation(TEST_ANNOTATION_NAME)) return
+        if (initialFunction.runsInRunTestBlock()) return
```
cortinico, line 63:
```suggestion
        if (initialFunction.isRunTestFunction()) return
```

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
```diff
@@ -0,0 +1,147 @@
...
+
+        funCoroutineLaunchesTraverseHelper
+            .isFunctionOrChildrenLaunchingCoroutines(initialFunction, bindingContext)
+            .ifTrue { report(CodeSmell(issue, Entity.from(initialFunction), MESSAGE)) }
```
cortinico, line 67:
```suggestion
        // By this point we know we're inside a test function that is not a `runTest` function.
        if (funCoroutineLaunchesTraverseHelper.isFunctionLaunchingCoroutines(initialFunction, bindingContext)) {
            report(CodeSmell(issue, Entity.from(initialFunction), MESSAGE))
        }
```

This improves the readability here

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
```diff
@@ -0,0 +1,147 @@
...
+
+        private const val TEST_ANNOTATION_NAME = "Test"
+        private val RUN_TEST_FQ = FqName("kotlinx.coroutines.test.runTest")
+    }
+}
+
+class FunCoroutineLaunchesTraverseHelper {
+    val exploredFunctionsCache = mutableMapOf<KtNamedFunction, Boolean>()
+
+    fun isFunctionOrChildrenLaunchingCoroutines(
```
cortinico, line 88:
```suggestion
    fun isFunctionLaunchingCoroutines(
```

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
```diff
@@ -0,0 +1,147 @@
...
+        initialFunction: KtNamedFunction,
+        bindingContext: BindingContext
+    ): Boolean {
+        val traversedFunctions = mutableSetOf<KtNamedFunction>()
```
cortinico, line 92: You don't need both `exploredFunctionsCache` and `traversedFunctions`.
Let's move `exploredFunctionsCache` inside the function and use the map to also keep track of the visited function.

At the end of the day, you just want to do an exhaustive visit of all the `KtNamedFunction` and return true if any of the visited function fired a coroutine.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
```diff
@@ -0,0 +1,147 @@
...
+            if (isLaunching) {
+                parents.forEach { exploredFunctionsCache[it] = true }
+            }
+        }
+
+        fun getChildFunctionsOf(
+            function: KtNamedFunction,
+            parents: List<KtNamedFunction> = emptyList()
+        ): Set<KtNamedFunction> {
+            function.collectDescendantsOfType<KtExpression>().mapNotNull {
```
cortinico, line 110: Q: Isn't this finding all the descendants to the n-th level? I don't think we need to recurse (but I might be wrong).

paolorotolo: Unfortunately not :(
For example, this is what happens running it with the first function in test `reports correctly coroutine launches in deep functions`.

Code:
```kotlin
// function.name -> "test that launches a coroutine"

function.collectDescendantsOfType<KtExpression>().mapNotNull {
    it.getResolvedCall(bindingContext)
        ?.resultingDescriptor
        ?.source
        ?.getPsi() as? KtNamedFunction
}.map { it.name }
```

Result:
```
0: "launchCoroutineOne"
1: "launchCoroutineOne"
```

While we expected:
```
0: "launchCoroutineOne"
1: "launchCoroutineTwo"
2: "launchCoroutineThree"
3: "launchCoroutineFour"
```

So it seems `collectDescendantsOfType` only return the first degree descendants.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/CoroutineLaunchedInTestWithoutRunTest.kt
paolorotolo: Sorry, I didn't get the suggestion: `exploredFunctionsCache` is outside the function to persist a global cache of explored functions when running the rule on multiple `@Test` functions.

`isFunctionLaunchingCoroutines` will return true if any of the descendants functions starting from the `initialFunction` launches a coroutine.
If the rule runs on another `@Test` function subsequently, which at some level of depth calls an already explored function, it will get the result from the global `exploredFunctionsCache`. So that map needs to survive multiple function calls.

cortinico: Oh I see. Ok then we need both. THanks for clarifying

cortinico (approved):

paolorotolo:
Rebasing the branch, last commits in `main` should fix failed tests on Windows

cortinico (approved):
Thanks for the contrib @paolorotolo, sorry for the lenghty review

PR #6693, opened 2023-12-06
Keep build cache configuration consistent in all builds

3flex:
It is recommended to keep the build cache configuration of the root build consistent with the build cache configuration of early evaluated included builds.

Included builds applied using `pluginManagement` do not inherit build cache configuration from the top-level build so were not saving task output to the remote cache.

### DIFF
--- a/build-logic/settings.gradle.kts
+++ b/build-logic/settings.gradle.kts
@@ -7,3 +7,20 @@ dependencyResolutionManagement {
         }
     }
 }
+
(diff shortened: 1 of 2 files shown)


3flex:
This should probably be applied with a settings plugin instead of copy pasting the config into each settings file, but there's very little documentation on settings plugins, and it might not be very practical to do within our project. Good candidate for a follow up though.

BraisGabin (approved):

PR #6700, opened 2023-12-08
Disable local build cache on CI

3flex:
From https://docs.gradle.org/8.5/userguide/build_cache.html#sec:build_cache_configure:
> Gradle tries to load build outputs from the local build cache first, and
> then tries the remote build cache if no build outputs are found. If
> outputs are found in the remote cache, they are also stored in the local
> cache, so next time they will be found locally.

Inversely this means that that when local cache is hit, the outputs will not be stored in the remote cache. This is not what we want on CI nodes. Forcing remote build cache only on CI means cache will be updated for every task that does not have outputs already stored in the remote cache. This ensures they're available for local dev builds.

This recent build result is from CI: https://ge.detekt.dev/s/55usta3jzxkya/timeline?toggled=WyI5Il0&view=by-type#9
The first two tasks listed are `:detekt-gradle-plugin:compileFunctionalTestKotlin` and `:detekt-gradle-plugin:compileTestKotlin`. They were both restored from (local) cache which was restored by gradle-build-action.

This result is from my machine hours later: https://ge.detekt.dev/s/cphvzlb5tnkti/timeline?toggled=WyIzOCJd&view=by-type#38
The same two tasks listed 2nd and 3rd were not restored from (remote) cache because they weren't stored there by the CI run.

### DIFF
--- a/build-logic/settings.gradle.kts
+++ b/build-logic/settings.gradle.kts
@@ -16,7 +16,7 @@ val isCiBuild = providers.environmentVariable("CI").isPresent
 
 buildCache {
     local {
-        isEnabled = true
+        isEnabled = !isCiBuild
     }
     remote<HttpBuildCache> {
         isPush = isCiBuild
--- a/detekt-gradle-plugin/settings.gradle.kts
+++ b/detekt-gradle-plugin/settings.gradle.kts
@@ -21,7 +21,7 @@ val isCiBuild = providers.environmentVariable("CI").isPresent
 
 buildCache {
     local {
-        isEnabled = true
+        isEnabled = !isCiBuild
     }
     remote<HttpBuildCache> {
         isPush = isCiBuild
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -63,7 +63,7 @@ develocity {
 
 buildCache {
     local {
-        isEnabled = true
+        isEnabled = !isCiBuild
     }
     remote(develocity.buildCache) {
         isEnabled = true


cortinico:
cc @runningcode

BraisGabin:
For sure that get hits on dev end is more important that the hits on ci. But wouldn't this slow down our ci biuilds a lot? And, if we don't use the local cache on ci, shouldn't we do something to don't store and restore it every time? Otherwise we are losing time in all the steps: restoring cache, building and storing new cache.

3flex:
> For sure that get hits on dev end is more important that the hits on ci. But wouldn't this slow down our ci biuilds a lot?

It shouldn't slow it down much; restoring from remote cache is quick on my machine. It also saves some time since there's no need to store or restore from local cache.

> And, if we don't use the local cache on ci, shouldn't we do something to don't store and restore it every time? Otherwise we are losing time in all the steps: restoring cache, building and storing new cache.

I didn't really understand this? Disabling the local cache will avoid storing or restoring local cache.

BraisGabin:
> I didn't really understand this? Disabling the local cache will avoid storing or restoring local cache.

I didn't know that. I thought that it will be always stored. But it have sense that it will not be stored too. 🤐

BraisGabin (approved):

schalkms (approved):

3flex:
@cortinico what do you think about this? I know you wanted Gradle to review.

I can also ask that this specific use case is covered in the `setup-gradle` documentation. Most CI builds don't have remote cache enabled so this use case isn't covered in those docs.

3flex:
Actually scratch that - I'll create an issue in that repo to request documentation updates to cover this use case.

runningcode:
The Detekt project benefitted from a savings of 14min on average on CI from the local build cache in the last 7 days. [Here is the data](https://ge.detekt.dev/scans/trends?search.startTimeMax=1707300683783&search.startTimeMin=1706655600000&search.tags=ci&search.timeZoneId=Europe%2FBerlin&trends.activeMetrics=%7B%22avoidanceSavings%22:%22localBuildCache%22%7D&trends.section=avoidanceSavings)

3flex:
Sure, but if CI was configured to prime the remote cache exclusively so all builds could benefit, and never store build outputs in local cache, would that be a sensible thing to do?

It seems some build outputs are being cached locally only on CI and aren't available on the remote cache. That's the problem I was trying to solve. But maybe there's a config issue elsewhere. I could validate whether it's still an issue too.

3flex:
I've opened https://github.com/gradle/actions/issues/102 to see if this can be addressed in the setup-gradle README and will reopen this PR later depending on the response.

3flex:
gradle-setup action now includes this in the REAMDE:

> If you have a remote build-cache available for your build, then it is recommended to do the following:
> - Enable [remote build-cache push](https://docs.gradle.org/current/userguide/build_cache.html#sec:build_cache_configure_use_cases) for your GitHub Actions builds
> - Disable [local build-cache]() for your GitHub Actions build
> 
> As well as reducing the content that needs to be saved to the GitHub Actions cache, 
> this setup will ensure that your CI builds populate the remote cache and keep the cache entries fresh by reading these entries.
> Local builds can then benefit from the remote cache.

PR #6701, opened 2023-12-08
Demo false negative for `UnusedPrivateProperty`

xouabita:
```kt
class A(private val foo: Any) {
    fun myFoo() = foo
}
class B(private val foo: Any) // <-- foo should be reported here
```


cortinico:
Is there an issue opened for this one @xouabita ?

BraisGabin:
This is a demostration of #6702 and we have #7089 that tries to fix this so I'm closing this PR.

PR #6706, opened 2023-12-11
Use consistent JDK for Kotlin compilation

3flex:
If toolchains are not used the version of Java that Gradle is running with is passed as an input to Kotlin compilation tasks. Using toolchains means compilation tasks have a consistent JDK version as the input. Builds can then run on any Java version and hit remote cache primed with builds run on different Java versions.

### DIFF
--- a/build-logic/src/main/kotlin/module.gradle.kts
+++ b/build-logic/src/main/kotlin/module.gradle.kts
@@ -1,5 +1,6 @@
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 import org.gradle.api.tasks.testing.logging.TestLogEvent
+import org.jetbrains.kotlin.gradle.tasks.UsesKotlinJavaToolchain
 
 plugins {
     id("packaging")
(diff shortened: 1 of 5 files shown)


build-logic/src/main/kotlin/module.gradle.kts
```diff
@@ -53,6 +54,14 @@ kotlin {
     }
 }
 
+val java8Launcher = javaToolchains.launcherFor {
+    languageVersion = JavaLanguageVersion.of(8)
+}
+
+project.tasks.withType<UsesKotlinJavaToolchain>().configureEach {
+    kotlinJavaToolchain.toolchain.use(java8Launcher)
+}
```
cortinico, line 63: Q: Why do we need this, can't just specify `jvmToolchain(8)` in the `kotlin{}`?

cortinico (approved):

build-logic/src/main/kotlin/module.gradle.kts
3flex: Not here because that would make tests only use 8, but with our current setup on CI we want test tasks to inherit the version Gradle is using.

`build-logic` has no tests so we can configure that differently.

PR #6717, opened 2023-12-13
Reuse testKitDir

Goooler:
Use this to share test dependencies cache for speeding up clean build.

https://docs.gradle.org/8.5/javadoc/org/gradle/testkit/runner/GradleRunner.html#withTestKitDir-java.io.File-

### DIFF
--- a/detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
+++ b/detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
@@ -143,6 +143,7 @@ constructor(
             withProjectDir(rootDir)
             withPluginClasspath()
             withArguments(args)
+            withTestKitDir(testKitDir)
             gradleVersionOrNone?.let(::withGradleVersion)
         }
     }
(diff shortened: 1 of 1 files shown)


3flex (changes requested):
This doesn't look right to me. The docs for `withTestKitDir` say:

>  This directory is not deleted by the runner after the test build.
>
> You may wish to specify a location that is within your project and regularly cleaned, such as the project's build directory.

If directory location is set to a directory that's not cleaned regularly then it will grow unchecked.

Goooler:
> If directory location is set to a directory that's not cleaned regularly then it will grow unchecked.

Yeah, it is possible to encounter such a situation, although I haven't experienced it yet. But if we put this folder under `build` or temp dir like the default, there will be no more speedup in clean builds. Maybe we can exclude `.gradle/testKit` on `gradle-build-action`, and keep the location on local devices, if something happends, we'll clean it manually.

See https://github.com/gradle/gradle-build-action?tab=readme-ov-file#exclude-content-from-gradle-user-home-cache.

3flex:
Also found this:

> Note that TestKit re-uses a Gradle User Home between tests (see GradleRunner.withTestKitDir(java.io.File)) which contains the default location for the local build cache. For testing with the build cache, the build cache directory should be cleaned between tests. The easiest way to accomplish this is to configure the local build cache to use a temporary directory.

Source: https://docs.gradle.org/8.5/userguide/test_kit.html#sub:test-kit-build-cache

We're probably not clearing enough from that directory as it is between tests, so I wouldn't want to move it from a temp directory at this stage.

The other thing is, the contents of the directory is an implementation detail, so there are no guarantees that it's actually storing data that will speed things up if persisting between tests or tests runs.

PR #6725, opened 2023-12-18
Move test to their correct location

BraisGabin:
These tests should be in the tests of each rule.


3flex:
I haven't looked, but is this how suppressions are tested on other rules? Or are there only a couple of suppression tests, which are the ones that have now been moved?

BraisGabin:
There are other tests that check that the suppression works as expected. This are just tests to ensure that these rules can be suppressed.

3flex (approved):

PR #6733, opened 2023-12-18
RuleSet provides a Factory of rules intead of the rules themselfs

BraisGabin:
This is the first step to allow detekt to instantiate the same rule more than once. Right now `RuleSet` provides a list of already configured rules. And that make impossible for us to instantiate the same rule more than once.

My proposal is that instead of give us a `List<BaseRule>` it should give us a `Map<RuleId, (Config) -> BaseRule>` This way we could instantiate the same rule id multiple times with different configurations.

Be aware! This PR doesn't implement the option to instantiate the same rule twice. This is only the first step in that direction.

I tried to do this kind of refactor multiple times in the past and I failed because I was trying to make big chunks and always have nice APIs. So, for sure, I don't like to expose a `Map<RuleId, (Config) -> BaseRule>`. The lambda is terrible too. But this is a step forward and it can be merged without breaking anything.

~Review the last commit only. The other commits are there to make all this compile but those changes are in other PRs (#6729, #6730, #6731, #6732 and #6734)~

Related with #6738


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSet.kt
```diff
@@ -7,9 +7,15 @@ typealias RuleSetId = String
 /**
  * A rule set is a collection of rules and must be defined within a rule set provider implementation.
  */
-class RuleSet(val id: RuleSetId, val rules: List<BaseRule>) {
+class RuleSet(val id: RuleSetId, val rules: Map<String, (Config) -> BaseRule>) {
```
BraisGabin, line 10: `(Config) -> BaseRule` is basically a `Factory`. You give it a configuration and it provides you a Rule with that configuration. Using a lambda is cool because then on the instantiation you can write this like: `::NameOfTheRule`. So, in the end, we will just be using the constructor of each rule as its factory. But I agree that when you read this code the lambda feels a bit strange.

I could create a `typealias` to don't use the lambda directly but lately I'm not a big fan of them because they just hide the type behind it. I like the `value class` because they enforce that type (in contrast to the `typealias`) but for sure a `value class` is not the good call here. Other option would be `fun interface` but I'm not sure if it will provide us too much.

marschwar (approved):
I like this iterative approach. This makes it at least possible to review the PR. Not sure

marschwar:
I know you are juggling many branches at the moment, but could you please whenever possible not  squash and force push. It makes it much harder to review whenever there are additional changes.

BraisGabin:
Sorry, I'll try. I'm use to fix the conflicts with rebases.

PR #6734, opened 2023-12-18
Mandatory config on Rule

BraisGabin:
If you don't pass a configuration to the rule your rule will not be able to be configured. It will seem that everything works but in reality it will not. If we force to pass a configuration it's less likely to have that problem.

There is only one place where this default value is used: in the tests. And we shouldn't have a default value on our production code just to make our tests a bit easier (you can see how few changes on tests this PR has).

TL;DR: api less error prone and the benefits to have it are minimal.

### DIFF
--- a/detekt-api/api/detekt-api.api
+++ b/detekt-api/api/detekt-api.api
@@ -338,7 +338,6 @@ public final class io/gitlab/arturbosch/detekt/api/ReportingExtension$DefaultImp
 }
 
 public abstract class io/gitlab/arturbosch/detekt/api/Rule : io/gitlab/arturbosch/detekt/api/BaseRule, io/gitlab/arturbosch/detekt/api/ConfigAware {
-	public fun <init> ()V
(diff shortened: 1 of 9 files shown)


cortinico:
Can you elaborate why you want to make this change? It seems like that offering a default empty ctor will make for easier integration from the user point of view

BraisGabin:
Sure, let me elaborate. If you don't pass a configuration to the rule your rule will not be able to be configured. It will seem that everything works but in reality it will not. If we force to pass a configuration it's less likely to have that problem.

There is only one place where this default value is used: in the tests. And we shouldn't have a default value on our production code just to make our tests a bit easier (you can see how few changes on tests this PR has).

TL;DR: api less error prone and the benefits to have it are minimal.

cortinico:
Ah I see. I guess it's because if we default to `Config.empty` then any configuration the user will provide in the config file will be always ignored right?

BraisGabin:
> Ah I see. I guess it's because if we default to `Config.empty` then any configuration the user will provide in the config file will be always ignored right?

Yes, exactly that!

cortinico (approved):

PR #6736, opened 2023-12-18
`ConfigAware` doesn't need to implement `Config`

BraisGabin:
`ConfigAware` impelments `Config` and `Rule` implements `ConfigAware` so, at the end a `Rule` is a `Config` class. That's really odd. We were even configuring a rule with another rule on our tests!! This PR changes that.


schalkms (approved):
👍

PR #6741, opened 2023-12-20
Enable MultilineLoop & ConditionWrapping rules

3flex:
Stacks on #6739

Enables two new rules on detekt code base and fixes related violations.


detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KtModifierList.kt
```diff
@@ -14,7 +14,8 @@ fun KtModifierListOwner.isPublicNotOverridden(considerProtectedAsPublic: Boolean
         isPublic || isProtected()
     } else {
         isPublic
-    } && !isOverride()
+    } &&
+        !isOverride()
```
BraisGabin, line 18: I like this rule in general but in these cases it looks a bit strange. But a nice improvement overall for sure.

BraisGabin (approved):

PR #6744, opened 2023-12-21
Add `BigDecimal`'s double constructor to `ForbiddenMethodCall`'s default configuration

### LINKED ISSUE #6704: Add rule to prevent `BigDecimal.valueOf(<some_double_value/some_float_value>)`
As per the comment at https://github.com/detekt/detekt/issues/6668#issuecomment-1834718270 it carried the issue of precision to `BigDecimal` which might be unintended.

We should create the rule to prevent this

VirtualParticle:
This forbids usages of `BigDecimal` double constructor since it will almost never be the intended behavior of the user. Instead of calling `new BigDecimal(double)`, users should use `BigDecimal.valueOf(double)` or just `BigDecimal(string)`.
 
Resolves #6704

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -608,6 +608,8 @@ style:
         value: 'kotlin.io.print'
       - reason: 'println does not allow you to configure the output stream. Use a logger instead.'
         value: 'kotlin.io.println'
+      - reason: 'using `BigDecimal.<init>(kotlin.Double)` can result in unexpected float point precision behavior. Use `BigDecimal.valueOf(kotlin.Double)` or `BigDecimal.<init>(kotlin.String)` instead.'
+        value: 'java.math.BigDecimal.<init>(kotlin.Double)'
   ForbiddenSuppress:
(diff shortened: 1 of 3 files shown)


detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -608,6 +608,8 @@ style:
         value: 'kotlin.io.print'
       - reason: 'println does not allow you to configure the output stream. Use a logger instead.'
         value: 'kotlin.io.println'
+      - reason: 'using `BigDecimal.<init>(kotlin.Double)` can result in unexpected float point precision behavior. Use `BigDecimal.valueOf(kotlin.Double)` or `BigDecimal.<init>(kotlin.String)` instead.'
+        value: 'java.math.BigDecimal.<init>(kotlin.Double)'
```
VirtualParticle, line 612: Not sure if the message should use the same matching syntax as the rule or if it should say "BigDecimal double constructor" or something like that.

BraisGabin (approved):

cortinico (approved):

schalkms (approved):

PR #6754, opened 2023-12-22
Bump Docusarus to 3.x

cortinico:
I'm bumping the website to the new major of Docusarus.
This required a bit of work as the tooling now is a bit stricter (i.e. `{% include links.html %}` was invalid, so I had to remove it).

Please take a look at the preview and let me know if we're missing anything


BraisGabin (approved):
It looks good

Issue #5855, opened 2023-02-27
Refactor `detekt-generator`


BraisGabin:
I would like to reduce the code inside `detekt-generator`. I would like to have a gradle plugin with 3 different tasks: `generateDocumentation`, `genetareDefaultConfiguration` and `generateDeprecation`. And apply that plugin to all our rule modules.

Then `detekt-generator` would:

- Grab all the `generateDocumentation` outputs and copy them to the `webstie/docs/rules/`.

And `detekt-core` would:
- Grab all the `generateDefaultConfiguration` outputs from the defaut rule sets and merge them in `detekt-core/src/main/resources/default-detekt-config.yml`
- Grab all the `generateDeprecation` outputs and merge them in `detekt-core/src/main/resources/deprecation.properties`.

This should remove the strange dependencies that we have.

The problem is that the plugin that I talk about would have dependencies with `detekt-api` and other modules. And I'm not sure if that's possible to archive without an explosion of gradle complexity.

What do you think?

---

Bonus: we could even stop committing those files. But I'm not sure if that would be worth it.

cortinico:
I think this is going to be quite complicated to achieve 🤔 

At first I thought you could just kick off a:
```
build-logic/src/main/kotlin/documentation.gradle.kts
```
and use that.

The problem is that your module-based `generateDocumentation` will have dependency on detekt-api. Why is that the case. At the end of the day it should just need to create Markdown from .kt source file no? So maybe we should start by refactoring the generator away from `detekt-cli` (I know it was done this was as the generator is also a CLI, but maybe we need to split some responsibilities here).

Also please note that the generator is published on Maven Central as a fat jar that users can invoke:
https://repo1.maven.org/maven2/io/gitlab/arturbosch/detekt/detekt-generator/1.22.0/detekt-generator-1.22.0-all.jar

3flex:
> it should just need to create Markdown from .kt source file no? So maybe we should start by refactoring the generator away from detekt-cli

That was my though too. If nobody else is currently working on this I have an approach I'd like to try.

cortinico:
> That was my though too. If nobody else is currently working on this I have an approach I'd like to try.

Please do 🙏

BraisGabin:
Please, go ahead. #5853 is my try to start this job. It is the easiest one, I think. But I'm not sure if it is the way of gradle

[...]

PR #6763, opened 2023-12-27
Remove input and output file registration for generateDocumentation task

3flex:
This task has caching disabled. When enabling with outputs.cacheIf {true} there is an error about overlapping task outputs because the fileTree output location includes `_category_.json` file.

Using a filter like `include("**/*.md")` on the fileTree does not work due to https://github.com/gradle/gradle/issues/9131.

There's no way to enumerate the output file names and use file declarations instead as the output filename is based on the value of `ruleSetId` in the rule set provider.

Based on the above it is not currently feasible to cache this task's outputs, so it makes sense to remove the redundant input and output file registrations from the task declaration.

### DIFF
--- a/detekt-generator/build.gradle.kts
+++ b/detekt-generator/build.gradle.kts
@@ -66,17 +66,6 @@ val generateDocumentation by tasks.registering(JavaExec::class) {
         .filter { it.exists() }
         .toList()
 
-    inputs.files(ruleModules)
-
-    outputs.files(
-        fileTree(documentationDir),
-        file(defaultConfigFile),
-        file(formattingConfigFile),
-        file(librariesConfigFile),
-        file(ruleauthorsConfigFile),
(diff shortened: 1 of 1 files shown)


BraisGabin (approved):

detekt-generator/build.gradle.kts
```diff
@@ -66,17 +66,6 @@ val generateDocumentation by tasks.registering(JavaExec::class) {
         .filter { it.exists() }
         .toList()
 
-    inputs.files(ruleModules)
```
cortinico, line 69: Maybe let's just comment it out for now instead?

3flex: Let's just take it out - generator module needs rework anyway (https://github.com/detekt/detekt/issues/5855) and this task will need to change

PR #6766, opened 2023-12-27
Add parent config property

BraisGabin:
Blocked by #6736

Right now the rules get the configuration of its parent and then look for its configuration there because the rules need information of their configuration and their parent's configuration too. With this change we allow (next PRs) that the rule could receive its configuration and then, when needed, look at its parent configuration.

### DIFF
--- a/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/config/DisabledAutoCorrectConfigSpec.kt
+++ b/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/config/DisabledAutoCorrectConfigSpec.kt
@@ -6,12 +6,19 @@ import org.junit.jupiter.api.Test
 
 class DisabledAutoCorrectConfigSpec {
     private val rulesetId = "style"
-    private val rulesetConfig = yamlConfig("/configs/single-rule-in-style-ruleset.yml").subConfig(rulesetId)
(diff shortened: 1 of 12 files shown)


detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/config/DisabledAutoCorrectConfigSpec.kt
```diff
@@ -6,12 +6,19 @@ import org.junit.jupiter.api.Test
 
 class DisabledAutoCorrectConfigSpec {
     private val rulesetId = "style"
-    private val rulesetConfig = yamlConfig("/configs/single-rule-in-style-ruleset.yml").subConfig(rulesetId)
+    private val rulesetConfig = yamlConfig("/configs/single-rule-in-style-ruleset.yml")
```
marschwar, line 9: [nit] now the name `rulesetConfig` could be a little misleading.

marschwar (approved):

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/config/DisabledAutoCorrectConfigSpec.kt
BraisGabin: You are right, fixed!

PR #6767, opened 2023-12-27
Rule receives its own configuration

BraisGabin:
This PR contains multiple commits from different PRs so it can compile. But the actual PR are just the last 2 commits.

Until now a rule receives the configuration of its parent and then the rule looks for its own configuration. This is what this PR changes. Now the core provides the configuration of the rule.

Thanks to #6766 the rule can access the parent (rulset) configuration if needed.


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSetProvider.kt
```diff
@@ -19,5 +19,5 @@ interface RuleSetProvider {
      * Make sure to pass the configuration to each rule to allow rules
      * to be self configurable.
```
TWiStErRob, line 20: Outdated KDoc.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSetProvider.kt
```diff
@@ -19,5 +19,5 @@ interface RuleSetProvider {
...
      */
-    fun instance(config: Config): RuleSet
+    fun instance(): RuleSet
```
TWiStErRob, line 22: I have a question: does this mean that the ruleset is not configurable any more? Wasn't there a way to dynamically decide which rules to include for example, or apply some kind of global configuration/decision on the ruleset level? This seems to be gone now.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSetProvider.kt
BraisGabin: You are right! I updated it at #6733 (and I'll rebase it here soon)

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSetProvider.kt
BraisGabin: No, this only means that to create a `RuleSet` now you don't need to pass a configuration. But this change is related with #6733. Can we move the conversation there?

> Wasn't there a way to dynamically decide which rules to include for example, or apply some kind of global configuration/decision on the ruleset level?

That's still there. That's done by `detekt-core`. I didn't change anything related with that afaik.

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
```diff
@@ -31,8 +31,8 @@ abstract class FormattingRule(config: Config) : Rule(config) {
      * Should the android style guide be enforced?
      * This property is read from the ruleSet config.
      */
-    protected val isAndroid
-        get() = FormattingProvider.android.value(ruleSetConfig)
+    protected val isAndroid: Boolean
+        get() = config.parent?.let { FormattingProvider.android.value(it) } == true
```
marschwar, line 35: Looking at this I realized I do not understand why we are doing this weird indirection. Can someone please refresh my memory.

Anyway, I then tried to follow the flow and realized, that #6733 might have broken android specific config properties. (I think) 

```kotlin
private class TransformedConfigPropertyWithAndroidVariants<T : Any, U : Any>(
    private val defaultValue: T,
    private val defaultAndroidValue: T,
    private val transform: (T) -> U
) : MemoizedConfigProperty<U>() {
    override fun doGetValue(thisRef: Rule, property: KProperty<*>): U {
        val isAndroid = getValueOrDefault(thisRef.config, "android", false)
        val value = if (isAndroid) defaultAndroidValue else defaultValue
        return transform(getValueOrDefault(thisRef.config, property.name, value))
    }
}
```
Since the android property is on the ruleset config and now we read it from the rule config, I think this no longer works. I have not confirmed that. I could try to create a test to verify.

marschwar (approved):
I like how this new approach is starting to come together.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/Suppressor.kt
```diff
@@ -21,7 +21,7 @@ private fun buildSuppressors(rule: ConfigAware, bindingContext: BindingContext):
 
 internal fun getSuppressors(rule: BaseRule, bindingContext: BindingContext): List<Suppressor> {
     return when (rule) {
-        is ConfigAware -> buildSuppressors(rule, bindingContext)
+        is Rule -> buildSuppressors(rule, bindingContext)
         else -> emptyList()
     }
 }
```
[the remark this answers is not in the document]
BraisGabin: Completely agree, we should merge the code from `BaseRule` inside `Rule` and remove `BaseRule`. It provides us nothing. I created a PR to track that: #6832

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
BraisGabin: > Looking at this I realized I do not understand why we are doing this weird indirection. Can someone please refresh my memory.

I imagine that the reason is that, as a user, I only want to define if `android: true` once in the configuration. But every single rule needs to know that. For that reason this rule reads this configuration from the parent.

-----

Could you elaborate how #6733 broke this? I can not see it.

Also, I didn't know that we had `configWithAndroidVariants` on `detekt-api`. That's something that we should move to `detekt-formatting`. I created #6833 to track it.

BraisGabin: Ok, I see what's strange on the code that you copy&pasted.

```
val isAndroid = getValueOrDefault(thisRef.config, "android", false)
```

should be somrthing like

```
val isAndroid = getValueOrDefault(thisRef.config.parent ?: Config.empty, "android", false)
```

Otherwise we always receive `false`.

But I can't find when that issue was added. It doesn't seem related with any of my PRs. I mean, we should check if it is really broken and fix it but I think that the issue is not related with this PR, right?

BraisGabin: Ok, I'm merging this meanwhile. If you find anything just let me know if I can help.

PR #6769, opened 2023-12-27
Use entries instead of values() on enum class

BraisGabin:


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
```diff
@@ -4,9 +4,6 @@ package io.gitlab.arturbosch.detekt.api
  * A code smell indicates any possible design problem inside a program's source code.
  * The type of a code smell is described by an [Issue].
  *
- * If the design problem results from metric violations, a list of [Metric]'s
- * can describe further the kind of metrics.
- *
```
3flex, line 9: This removal looks unrelated?

BraisGabin: You are right. I usually develop all this code in a single branch and then I cut it in different PRs. This change is in #6768 but on the rebase I forget to remove that commit from this PR. I will remove it as soon as I'm near a computer.

cortinico (approved):

Issue #6738, opened 2023-12-19
Add the option to instantiate the same rule more than once


BraisGabin:
This is a feature that we want to have for detekt 2.0

## Needed changes
There are multiple changes that should be made to make this possible. Here a list with all the changes that we need:

1. `RuleSet`s shouldn't instantiate the rules. They should provider factories to pass a confguration to create that rule.
    - **State:** #6733
1. `Rule`s shouldn't pick their own configuration from its parent configuration. We should provide their configuration directly.
    - **State:** #6767
1. The `Rule` id can't be hardcoded. If a rule can have multiple instances the id should change between those instances.
    - **State:** No started
1. The `Config` class should provide an API to read all the subconfigurations inside it.
    - **State:** #6842
1. The core should read the `Config` and instantiate the rules from there.
    - **State:** This should be the easiest. We need to align on the new yaml key to find the rule but the change should be as easy as this:
```kotlin
.flatMap { (ruleSet, ruleSetConfig) ->
    ruleSetConfig.subconfigs
        .map { (configId, config) ->
            val ruleId = config.valueOrNull("rule") ?: configId
            val ruleProvider = checkNotNull(ruleSet.rules[ruleId]) {
                "Unknown rule ${ruleSet.id}:$ruleId"
            }
            ruleProvider.invoke(config)
        }
}
```

----

This is a huge change. There is code that assumes that the rules have a hardcoded id or that the there is only one instance of each. And this is a breaking change, we need to change the API for 3 main classes/interfaces: `RuleSet`, `Rule` and `Config`. This will impact every single third party rule out there.

Also, those are the big "chunks" that should be done, then a lot of minor changes should be made on the way to be able to make those possible.

BraisGabin:
You are correct, that's exactly the use case. Could I ask you how do you imagine the configuration should look like to make your use case work?

BraisGabin:
I need to write down all my thoughts about Rule/Issue/Finding and all that part of the detekt API:

We need to talk about rule ids. Right now it's easy! Every `Rule` has an `Issue`. The `Issue` stores the `id` and the `description` of the `Rule`. And `Rule.id` is just syntax sugar for `Rule.issue.id`.

That works when you only have one instance per rule. But once you have multiple instance of the same rule the thing gets messy really quick.

First of all the rules need 2 `id`s:
- One to identify the type of the rule (needed to link for its documentation, for example)
- One to identify the instance of the rule (as a user I want to know which configuration of the rule raised that issue)

[...]

PR #6774, opened 2023-12-27
Remove Finding.id

BraisGabin:
The `id` of `Finding` wasn't a real id. It didn't identify the `Finding` it was just a shortcut for the `issue.id`.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
@@ -8,7 +8,6 @@ package io.gitlab.arturbosch.detekt.api
  * an entity. Metrics and entity references can also considered for deeper characterization.
  */
 interface Finding : Compactable, HasEntity {
-    val id: String
     val issue: Issue
     val references: List<Entity>
     val message: String
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
@@ -1,11 +1,10 @@
 package io.gitlab.arturbosch.detekt.api
 
 /**
- * Base interface of detection findings. Inherits a bunch of useful behaviour
- * from sub interfaces.
+ * Represents a problem detected by a rule on the source code
  *
- * Basic behaviour of a finding is that is can be assigned to an id and a source code position described as
- * an entity. Metrics and entity references can also considered for deeper characterization.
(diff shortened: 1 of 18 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
```diff
@@ -8,7 +8,6 @@ package io.gitlab.arturbosch.detekt.api
  * an entity. Metrics and entity references can also considered for deeper characterization.
  */
 interface Finding : Compactable, HasEntity {
-    val id: String
```
cortinico, line 11: Perhaps keeping this as:

```kotlin
val id: String get() = issue.id
```

is maybe better?

BraisGabin: We could. I even had it as an extension function of `Finding`:

```kotlin
val Finding.id: String get() = issue.id
```

But it is not an `id`. The issue.id doesn't identify the finding. It's more, we even have a baseline finding id to identify a `Finding` because the `id` of `Finding` doesn't do the job.

These refactors try to simplify our `Finding` and `Rule` API because I need to break the 1 to 1 relation that we have right now between `Issue` and `Rule` and, to be honest, I'm not sure how to do it yet. I'll write about this on #6738 because it is not easy.

BraisGabin: Also the problem is that if we go to something like I described at https://github.com/detekt/detekt/issues/6738#issuecomment-1871283168 which should be the Finding id: the rule id or the instance id?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
```diff
@@ -8,7 +8,6 @@ package io.gitlab.arturbosch.detekt.api
  * an entity. Metrics and entity references can also considered for deeper characterization.
  */
```
marschwar, line 9: Apart from grammar and spelling, now the type comment also references a non existent `id`. :smile:

marschwar (approved):
I think you are right. This suggests that a finding has a unique id which it does not.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Findings.kt
BraisGabin: I rewrite the kdoc.

marschwar: Thank you. LGTM

PR #6789, opened 2024-01-02
Inline yaml samples in tests

BraisGabin:
Inline some configuration to make the tests easier to read.


cortinico:
Seems like the website preview failed for this PR due to insufficient JVM memory. Is this the first time that it happens?

BraisGabin:
No, I see this issue a lot lately. I'm not sure if this happens due to changes at `core` or `api` that force a lot of compilation (the build cache gives only misses) or any other reason. But, as I said this is not the first.

marschwar (approved):

PR #6793, opened 2024-01-02
Documentation improvements

BraisGabin:


3flex (approved):

cortinico:
> It looks like you're editing the un-versioned copy of our website. This affects only users on the 'next' version of detekt, and it's correct only if you intend to document a future change or feature. If you intended to make a change also for the current version of detekt, please make sure you edit the equivalent file inside website/versioned_docs/ as well.

I think we should update this also for the 1.23.4 version of the website @BraisGabin

BraisGabin:
Done at #6801

PR #6794, opened 2024-01-02
Remove `abstract` from `Rule.issue`

BraisGabin:
Change how to implement a `Rule`:

- Before: override a property with an `id` and a `description`
- After: add `description` at the constructor level.

This makes the API less errorprone as the `id` should always be `javaClass.simpleName`. And also make prepares the code to add the new `id` to `Issue`.


detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/ConfigPropertySpec.kt
```diff
@@ -14,13 +14,15 @@ class ConfigPropertySpec {
     inner class `simple property` {
         @Nested
         inner class `String property` {
-            private val configValue = "value"
-            private val defaultValue = "default"
-            private val subject = object : TestRule("present" to configValue) {
+            inner class Subject : TestRule("present" to configValue) {
```
BraisGabin, line 17: I like this PR in general but these tests are ugly as hell now. `object` doesn't return an empty String when call `Rule.javaClass.simpleName` so the creation of `Issue` fails because the `id` of an `Issue` can't be empty.

Also this make me think if this hard link between `Rule.javaClass.simpleName` and `Issue.id` is a good idea. Should I add an `id` parameter in the constructor of `Rule` like this:

```kotlin
abstract class Rule(
    val config: Config,
    description: String,
    id: String = javaClass.simpleName, // this doesn't compile but just to give the general idea
)
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
```diff
@@ -17,12 +17,13 @@ import org.jetbrains.kotlin.psi.KtFile
...
 abstract class Rule(
     val config: Config,
+    description: String,
 ) : BaseRule() {
 
     /**
      * A rule is motivated to point out a specific issue in the code base.
      */
-    abstract val issue: Issue
+    val issue: Issue = Issue(javaClass.simpleName, description)
```
BraisGabin, line 26: THIS is the change of this PR. The rest of the PR is just a refactor to make the code compile to THIS change.

detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/ConfigPropertySpec.kt
marschwar: How about using a field instead of the constructor.

```kotlin
    open val ruleId: String = javaClass.simpleName
```

BraisGabin: Done! I'm not 100% sure if this the best solution but for sure it is way better. And we can revisit this in the future. In another PR with a cleaner diff will be easier.

marschwar (approved):

BraisGabin:
> Are you actively working on the one @BraisGabin at the moment or can I fix the pipeline?

Thanks for the fixes :)

marschwar:
I you are ok with the changes I made, I would say this can be merged.

PR #6798, opened 2024-01-03
Properly handle nullable type

3flex:
`actual` is a nullable type but the K1 compiler ignores that when referencing `actual` in functions in the class. This is a compilation failure in K2. Backporting the fix as the new behaviour is correct.


cortinico (approved):

cortinico:
@3flex I'm re-adding this to the 2.0.0 milestone. Was there as a specific reason why you added and then removed it?

3flex:
Not really... It's housekeeping with no external impact so I'm not really fussed if it's in the changelog.

cortinico:
Yup you're right. I haven't realised this is only `/test/` source code 👍

PR #6799, opened 2024-01-03
Fix test snippets for K2

3flex:
In language version 2.0 it is not permitted to use `protected` on a class that is in a file. As test snippets are compiled as KTS scripts this will otherwise cause these tests to fail.


cortinico (approved):

BraisGabin:
Removing the `protected` those tests have no sense anymore. Would them work if instead of `compileAndLint` we use just `lint`? Otherwise I would vote to remove them instead.

BraisGabin:
My bad!

PR #6804, opened 2024-01-04
Remove UnstableApi annotation

3flex:
In 2.0 final we should not have any unstable APIs. See first commit for the elements which were declared unstable API in the past which can now be considered stable.

~~Tagged breaking change as a public annotation was removed from detekt-api but that might be unnecessary to feature in release notes.~~

Edit: changed my mind and untagged as breaking change - nobody would be using this in their own code.


BraisGabin (approved):
<3

PR #6805, opened 2024-01-04
Move Configuration annotation from internal to public API

3flex:


BraisGabin (approved):
This would be another annotation to be moved to the `detekt-generated-api` as I described at https://github.com/detekt/detekt/pull/6807#pullrequestreview-1803713512. Again, probably it is not worth it and it doesn't block this PR.

marschwar:
Is this really the case? What are the benefits for 3rd party rule authors to use the annotation. It is only used to generate the documentation for the 1st party rules. Or am I missing something?

marschwar (changes requested):
Is this really the case? What are the benefits for 3rd party rule authors to use the annotation. It is only used to generate the documentation for the 1st party rules. Or am I missing something?

3flex:
> It is only used to generate the documentation for the 1st party rules

It should be available for third party rules to use as well so they can also generate documentation.

marschwar (approved):

PR #6806, opened 2024-01-04
Move RequiresTypeResolution annotation from internal to public API

3flex:

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RequiresTypeResolution.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RequiresTypeResolution.kt
@@ -1,4 +1,4 @@
-package io.gitlab.arturbosch.detekt.api.internal
+package io.gitlab.arturbosch.detekt.api
 
 /**
  * Annotated [io.gitlab.arturbosch.detekt.api.Rule] requires type resolution to work.
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RequiresTypeResolution.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RequiresTypeResolution.kt
@@ -1,7 +1,7 @@
 package io.gitlab.arturbosch.detekt.api
 
 /**
- * Annotated [io.gitlab.arturbosch.detekt.api.Rule] requires type resolution to work.
+ * Annotated [Rule] requires type resolution to work.
  *
  * The detekt core will honor this annotation and it will not run any rule with this annotation if the bindingContext
  * is empty.
--- a/detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/ViolatesTypeResolutionRequirements.kt
+++ b/detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/ViolatesTypeResolutionRequirements.kt
@@ -6,8 +6,8 @@ import io.gitlab.arturbosch.detekt.api.CodeSmell
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.RequiresTypeResolution
 import io.gitlab.arturbosch.detekt.api.Rule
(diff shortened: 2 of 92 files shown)


BraisGabin (approved):
Most likely this annotation is used by third party rules. Otherwise they will not be able to use type resolution.

Should we keep a copy of the annotation on `internal` with a `@Deprecate` and set the deprecation level to `Error` and a `replaceWith` with the new annotation?

I'm not 100% sure about this and if this would make us fall in a deep rabbit hall but this seems an easy deprecation to maintain (at least for some versions) and we would make the transition to `2.0` easier.

I mean, I wouldn't want to maintain hundreds of this `@Deprecate` because some of them will be really difficult ones. But this one seems kind of easy.

3flex:
Yeah I did think about that. But we have so many changes already in 2.0 and it's a very simple change for consumers that it didn't seem worth it to keep the old one even if it's simple.

@detekt/maintainers - thoughts? This applies to a couple of other open PRs.

schalkms (approved):

schalkms:
The annotation should have been public from the beginning IMHO.

marschwar:
Since we will change our base package to `dev.detekt` in 2.0, I do not really see much benefit. Users will have to touch the import not matter what.

marschwar (approved):

TWiStErRob:
@marschwar assuming you're replying to Brais's comment for migration help, there's a difference between `s/io.gitlab.arturbosch.detekt.api./dev.detekt./g`, and manually fixing errors afterwards when the IDE could help migrate based on a deprecation.

We don't need a full annotation, because the annotation wouldn't do anything, instead we could try to add a deprecated `typealias` [a'la IntelliJ plugin](https://github.com/JetBrains/gradle-intellij-plugin/blob/95c66816ce182feb271e1eaadcb1c82e6002a645/src/main/kotlin/org/jetbrains/intellij/tasks/InstrumentCodeTask.kt#L27-L28).

TWiStErRob:
For reference: https://github.com/search?q=%22io.gitlab.arturbosch.detekt.api.internal.RequiresTypeResolution%22+-is%3Afork+-repo%3Adetekt%2Fdetekt+-repo%3Adiffblue%2Fdetekt&type=code

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RequiresTypeResolution.kt
```diff
@@ -1,4 +1,4 @@
-package io.gitlab.arturbosch.detekt.api.internal
+package io.gitlab.arturbosch.detekt.api
 
 /**
  * Annotated [io.gitlab.arturbosch.detekt.api.Rule] requires type resolution to work.
```
TWiStErRob, line 4:
```suggestion
 * Annotated [Rule] requires type resolution to work.
```

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/ViolatesTypeResolutionRequirements.kt
TWiStErRob, line 1: I'm surprised this doesn't need adjustment.

detekt-rules-ruleauthors/src/test/kotlin/io/gitlab/arturbosch/detekt/authors/ViolatesTypeResolutionRequirementsSpec.kt
TWiStErRob, line 1: Organize imports doesn't work in code blocks, order is swapped after relocation.

TWiStErRob (approved):

3flex:
After reflection (and the approvals) I'm merging as is without a deprecation because:
1. I personally don't think it adds enough value
2. It was an `internal` package anyway so shouldn't have been used as is with an expectation it would always be supported as it was never part of the public API.

I'll merge related PRs as well. If we decide to add a deprecation to smooth migration it can always be added later, perhaps just for the RC phase for 2.0.

detekt-rules-ruleauthors/src/main/kotlin/io/gitlab/arturbosch/detekt/authors/ViolatesTypeResolutionRequirements.kt
3flex: The rule checks that the annotation RequiresTypeResolution is applied to a class using BindingContext, and the import for the annotation was changed in this file, so it did get need & get adjustment.

PR #6807, opened 2024-01-04
Move ActiveByDefault annotation from internal to public API

3flex:

### DIFF
--- a/detekt-api/api/detekt-api.api
+++ b/detekt-api/api/detekt-api.api
@@ -1,3 +1,7 @@
+public abstract interface annotation class io/gitlab/arturbosch/detekt/api/ActiveByDefault : java/lang/annotation/Annotation {
+	public abstract fun since ()Ljava/lang/String;
+}
+
 public final class io/gitlab/arturbosch/detekt/api/AnnotationExcluder {
 	public fun <init> (Lorg/jetbrains/kotlin/psi/KtFile;Ljava/util/List;Lorg/jetbrains/kotlin/resolve/BindingContext;)V
 	public final fun shouldExclude (Ljava/util/List;)Z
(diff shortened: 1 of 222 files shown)


BraisGabin (approved):
This annotation is only used by `detekt-generator`. I was wondering if `detekt-generator` should have it's own "api" library and let the third party authors decide if they depend on it or not.

Probably it is not worth the effort but I just wanted to share my thoughts.

Anyway this doesn't block this PR. For sure this shouldn't be `internal`, we want third party authors to use `detekt-generator`.

3flex:
>  I was wondering if detekt-generator should have it's own "api" library and let the third party authors decide if they depend on it or not.

I had the same thought though it's convenient to have a single API module to maintain. If we did this it would be best to do it before 2.0 is released.

BraisGabin:
> it's convenient to have a single API module to maintain.

Agree, so let's forget about it for now. If we have problems with this "mixing" in the future we can take measure later. YAGNI.

schalkms (approved):

marschwar:
Does this annotation even make sense for 3rd party rule authors? To me this suggests that we interpret it at runtime. Why would we want an author to be able to define that with detekt version 1.x.x the rule is active by default?

If we want the `detekt-formatting` rule set in its own repository to be able to use this, we also have to make the `detekt-generator` publicly available as mentioned above. If that is the route we want to go, I am ok with this as an intermediate step. The same is true for #6805.

BraisGabin:
> If we want the `detekt-formatting` rule set in its own repository to be able to use this, we also have to make the `detekt-generator` publicly available as mentioned above. If that is the route we want to go, I am ok with this as an intermediate step. The same is true for #6805.

I think that this is the route we should follow.

3flex:
> Does this annotation even make sense for 3rd party rule authors?

I think it does, there will be third party rulesets that have rules that are disabled by default. The annotation doesn't make sense if it's only related to the detekt version it became active by default, but I changed the doc to remove detekt-specific references. That way the third party author can declare which version of their own artifact the rule was enabled by default.

PR #6808, opened 2024-01-04
Remove deprecated XML tag support

3flex:


detekt-core/src/test/resources/baseline_feature/deprecated-baseline.xml
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <SmellBaseline>
-    <Blacklist>
+    <ManuallySuppressedIssues>
```
BraisGabin, line 3: I think that we can remove this file and the test that uses it. If I recall it correctly it was there just to ensure that we support both options correctly.

schalkms (approved):

detekt-core/src/test/resources/baseline_feature/deprecated-baseline.xml
3flex: Thanks for the suggestion, you're right - the change to the content made it identical to `valid-baseline.xml` and the associated test became the same as another existing test. Removed.

PR #6811, opened 2024-01-04
Move CompilerResources from internal to public API

3flex:
This type is used in `BaseRule` so it should be part of the public API.


BraisGabin (approved):

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantVisibilityModifierRuleSpec.kt
```diff
@@ -2,7 +2,7 @@ package io.gitlab.arturbosch.detekt.rules.style
 
 import io.github.detekt.test.utils.compileContentForTest
 import io.gitlab.arturbosch.detekt.api.Config
-import io.gitlab.arturbosch.detekt.api.internal.CompilerResources
+import io.gitlab.arturbosch.detekt.api.CompilerResources
```
schalkms, line 5: > RedundantVisibilityModifierRuleSpec.kt:3:1: Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end [ImportOrdering]

Imports shall be reordered.

schalkms (approved):
Rest is fine for me.

PR #6812, opened 2024-01-04
Don't use Config.empty on rule initialisation

3flex:
Comment on `Config.empty` says "This config should only be used in test cases". Unfortunately it's been added over time to many other places in detekt, particularly as a default value in rule constructors.

This PR does the bulk of the work to remove those usages in production code (by removing the default value from the rule constructors) leaving only a handful of other instances which can hopefully be removed later. I'd like to move the empty config code to detekt-api test fixtures.


BraisGabin:
Instead of moving it to test I would vote to remove it directly. We have `TestConfig` already. And if you call the constructor without arguments you get an empty config.

This doesn't mean to not merge this PR. This one does the heavy work. In other pr we can just change `Config.empty` for `TestConfig()`.

BraisGabin (approved):

PR #6814, opened 2024-01-04
Move PathFilters from internal to public API

3flex:
This type is used in `Rule` so it should be part of the public API.


BraisGabin:
Can we hold this PR? I have in mind some refactors in this area where I move all these code from `api` to `core`. The idea is that the `Rule` will not need to handle if it should run over a file or not. The `core` will check the configuration and only call the rule when it should check a file.

I can't advance right now on that because I have a lot of open PRs. They depend between each other and I can't continue because I'm on a branch hell right now. Reviews are more than welcome 🙏, I'm stuck.

3flex:
Sure, I'll mark as a draft and come back to it later.

3flex:
Closing in favour of #6866

PR #6817, opened 2024-01-05
Make test pass on JVM 8

BraisGabin:
I'm not sure how this went through CI but it is there on `main`: https://github.com/detekt/detekt/commit/a03e3be950e29663f14bf036f1a1f77014dfe21b

On the JVM 8 the `simpleName` of a class contains the function name where it was declared. With this fix I move the rule declaration to the top level so the problem should be gone.


schalkms (approved):
Really strange. CI passed for your PR here…
https://github.com/detekt/detekt/pull/6791/commits

3flex:
> Really strange. CI passed for your PR here…
> https://github.com/detekt/detekt/pull/6791/commits

Might be due to predictive test selection? I'll have a look

3flex:
@schalkms all PRs should have a milestone of 2.0.0 for now since `main` branch is the V2 codeline. Changes still only be made in point releases by exception.

BraisGabin:
I don't care about the count of PRs in the changelog. I usually have that question. Maybe we shouldn't add the housekeeping PRs in the changelog. This way adding a milestone or not would not be important.

PR #6823, opened 2024-01-06
Remove unnecessary functions from detekt-psi-utils

3flex:
Remove two functions which had single usages that could be simplify replaced with functions available in `kotlin-compiler-embeddable`.


BraisGabin (approved):

TWiStErRob:
Recommended replacement for `element.parentsOfTypeUntil<A, B>()` and `element.parentsOfTypeUntil<A, B>(true)`:
```kotlin
element
	.parents
	.takeWhile { it !is B }
	.filterIsInstance<A>()
```

Recommended replacement for `element.parentsOfTypeUntil<A, B>(false)`:
```kotlin
element
	.parentsWithSelf
	.takeWhile { it !is B }
	.filterIsInstance<A>()
```

PR #6835, opened 2024-01-08
Remove the baserule, migrate code to rule

VovaStelmashchuk:
#6832 Remove base rule.

PR #6837, opened 2024-01-08
Remove base rule

### LINKED ISSUE #6832: Remove `BaseRule`
Right now we have `BaseRule` and `Rule`. But the only implementation of `BaseRule` is `Rule`. [This creates really odd code on our core](https://github.com/detekt/detekt/pull/6767#discussion_r1444068161). There is no reason to have the code splitted in two different abstract classes.

In the future, probably, we want to have an `interface Rule` and the `core` will use that. But that's something for the future. For now I think that we should merge `BaseRule` and `Rule` inside `Rule` and remove `BaseRule` to simplify the code.

VovaStelmashchuk:
Recreating of #6835 , fixes #6832


VovaStelmashchuk:
Need squash merge :)

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -124,11 +118,13 @@ internal class Analyzer(
             .filter { rule ->
                 bindingContext != BindingContext.EMPTY || !rule::class.hasAnnotation<RequiresTypeResolution>()
             }
-            .partition { isCorrectable(it) }
+            .partition { rule ->
+                rule.autoCorrect
+            }
```
BraisGabin, line 123:
```suggestion
            .partition { rule -> rule.autoCorrect }
```

BraisGabin (approved):
Looks great, thanks!

PR #6841, opened 2024-01-09
useDefaultConfig on tests

BraisGabin:
Right now the tests of `detekt-compiler-plugin` relies in an "undefined behavior" of the core. They don't define a configuration nor use `useDefaultConfig=true`, so no rule should be run. Every rule is disabled by default unless you define the configuration or have `useDefaultConfig=true`.

I want to refactor how the configuration is created on `detekt-core` and remove that "undefined behaviour". For that reason I need to fix these tests to pass `useDefaultConfig=true`.

### DIFF
--- a/detekt-compiler-plugin/build.gradle.kts
+++ b/detekt-compiler-plugin/build.gradle.kts
@@ -85,7 +85,6 @@ val testPluginKotlinc by tasks.registering(Task::class) {
     outputs.dir(outputDir)
 
     val baseExecutablePath = "${unzipKotlinCompiler.get().destinationDir}/kotlinc/bin/kotlinc"
-    val pluginParameters = "plugin:detekt-compiler-plugin:debug=true"
 
     val kotlincExecution = providers.exec {
         workingDir = outputDir.get().asFile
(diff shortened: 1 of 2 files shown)


BraisGabin:
@3flex do you have any idea how to pass those arguments?

detekt-compiler-plugin/build.gradle.kts
```diff
@@ -95,14 +94,15 @@ val testPluginKotlinc by tasks.registering(Task::class) {
             sourceFile.path,
             "-Xplugin=${tasks.shadowJar.get().archiveFile.get().asFile.absolutePath}",
             "-P",
+            "plugin:detekt-compiler-plugin:debug=true".toArg(),
+            "-P",
+            "plugin:detekt-compiler-plugin:useDefaultConfig=true".toArg(),
```
cortinico, line 99: Is this necessary? As it seems you're hardcoding the value to true inside `CompilerTestUtils`

BraisGabin: Yes, it is. This gradle task is "test" task. but not related at all with the unit tests. So I need this too.

cortinico: Ahh I see, thanks for clarifying

BraisGabin:
@detekt/maintainers a review here would help a lot I have other PRs pending to this.

schalkms (approved):

PR #6845, opened 2024-01-10
Remove `findings` property from `Rule`

BraisGabin:
There is no reason to have a `findings` property to get the findings that a rule have. Instead of that each time you execute a rule you get the findings as its return value.

We still depend on a private `_findings` that make this function no multithread. But this function wasn't multithread safe before so no changes there. Future work could fix that. Step by step.


detekt-formatting/src/test/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRuleSpec.kt
```diff
@@ -85,11 +84,8 @@ class FormattingRuleSpec {
...
         val rule = ChainWrapping(Config.empty)
-        rule.visitFile(compileForTest(expectedPath))
-        assertThat(rule.findings).isNotNull()
-
-        val findings = rule.findings
-
-        assertJThat(findings.first().location.filePath.absolutePath.toString()).isEqualTo(expectedPath.toString())
+        val findings = rule.visitFile(compileForTest(expectedPath))
+        assertThat(findings).isNotNull()
+        assertThat(findings.first().location.filePath.absolutePath.toString()).isEqualTo(expectedPath.toString())
```
marschwar, line 89:
```suggestion
        assertThat(findings).anySatisfy { finding ->
            assertThat(finding.location.filePath.absolutePath.toString()).isEqualTo(expectedPath.toString())
        }
```
There is no need to check for findings to be `null`.

marschwar (approved):
I like how things start to become easier and feel more natural

PR #6848, opened 2024-01-11
Simplify the creation of `Config`

BraisGabin:
This PR just cleans the configuration code and moves the functions to where they should be declared. The next step would be to make those `internal` functions `private` and test all that code as a one unit. But I will leave that for later or for anyone willing to contribute.

~Waiting for #6844~


BraisGabin:
Blocked by #7054

schalkms (approved):
Looks good!
Since this PR is around 1 month old, I would prefer rebasing to latest and if everything runs fine, merge it to main.

Issue #6822, opened 2024-01-05
False negative: `ExplicitItLambdaParameter`


SomeTroglodyte:
## Expected Behavior
Typing a lambda `it` to allow inference is sometimes the simplest way to enable shorthands like [SAM conversion](https://kotlinlang.org/docs/fun-interfaces.html).
```kotlin
        val dir = com.badlogic.gdx.files.FileHandle(".")
        val subdirs = dir.list { it: File -> it.isDirectory }
```
...is calling [this list() function](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/files/FileHandle.java#L441), successfully, and succinctly. Without typing the `it`, inference fails.

A rewrite that doesn't trip detekt would definitely be far less readable:
```kotlin
        val dir = com.badlogic.gdx.files.FileHandle(".")
        @Suppress("ObjectLiteralToLambda")
        val subdirs = dir.list(object : FileFilter { override fun accept(file: File) = file.isDirectory })
```

## Observed Behavior
Analysis fails with `This explicit usage of `it` as the lambda parameter name can be omitted. [ExplicitItLambdaParameter]`

## Steps to Reproduce
see above

## Context
I encountered a slightly more complex example earlier in https://github.com/detekt/detekt/pull/6367#issuecomment-1687561697 - and rewrote that source.
Job showing this one: https://github.com/yairm210/Unciv/actions/runs/7426175374/job/20209368340?pr=10881
Issue search turns up a few, but none comparable, most fixed - but this is not, as 1.23.4 already contains PR #6367, right?

## Your Environment

* Version of detekt used: 1.23.4
* Github workflow definition: https://github.com/yairm210/Unciv/blob/cfff0aa66a2bb05d3da002bad54695bae4951d40/.github/workflows/detektAnalysis.yml
* Link to project fork (if it's a public repository): https://github.com/SomeTroglodyte/Unciv

atulgpt:
Hi, @SomeTroglodyte why can't you use the below code? In my case removing `it` and directly using `val subdirs2 = dir.list { it.isDirectory }` working fine to me.

```
fun interface FileFilter {
    fun accept(file: File): Boolean
}

class FileHandle {
    fun list(filter: FileFilter): Array<FileHandle?> {
        return emptyArray()
    }
}

fun main() {
    val dir = FileHandle()
    val subdirs1 = dir.list { it: File -> it.isDirectory } // this is problematic code
[...]

PR #6850, opened 2024-01-12
Fix false positive on `it` usages when type parameter is specified

atulgpt:
Fixes https://github.com/detekt/detekt/issues/6822

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExplicitItLambdaParameterSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExplicitItLambdaParameterSpec.kt
@@ -1,8 +1,8 @@
 package io.gitlab.arturbosch.detekt.rules.style
 
 import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.test.assertThat
 import io.gitlab.arturbosch.detekt.test.compileAndLint
-import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExplicitItLambdaParameterSpec.kt
```diff
@@ -35,20 +38,32 @@ class ExplicitItLambdaParameterSpec {
...
+            assertThat(findings).isEmpty()
+        }
+
         @Test
         fun `does not report when parameter type is declared explicitly for un-inferrable lambda`() {
             val findings = subject.compileAndLint(
                 """
                     fun f1(): (Int) -> Int {
-                        return { it: Int -> it.inc() }::invoke
+                        return { value: Int -> value.inc() }::invoke
```
BraisGabin, line 58: This change seems really odd. Wasn't this test asserting exactly what the issue say that doesn't work?

atulgpt: No, there was one more thing that this rule was checking no parameter should be named `it`. when I fixed the issue https://github.com/detekt/detekt/pull/6367 I accidentally bypassed that check. Which is fixed now and hence needed to fix the TC

`return { it: Int -> it.inc() }::invoke ` this will fail with error `"`it` should not be used as name for a lambda parameter."`

And the problematic code that I added was `if (lambdaExpression.getParentExpressionRemovingParenthesis() is KtCallableReferenceExpression) return`

atulgpt: Actually the fixes of https://github.com/detekt/detekt/pull/6367/files is no more needed, as explicit type is anyway skipping the check

BraisGabin (approved):

BraisGabin:
I didn't see your answer @atulgpt .sorry for the late review 😅

PR #6851, opened 2024-01-12
Create a value class for the id of a `Rule`

BraisGabin:
A continuation of #6849. Add a little of type safety to all the ids that we have in the project so we don't mess up. Now the `id` of a `Rule` has it's own type. It is a `value class` to avoid the Boxing problem.

### DIFF
--- a/detekt-api/api/detekt-api.api
+++ b/detekt-api/api/detekt-api.api
@@ -213,7 +213,7 @@ public final class io/gitlab/arturbosch/detekt/api/HasEntity$DefaultImpls {
 public final class io/gitlab/arturbosch/detekt/api/Issue {
 	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun getDescription ()Ljava/lang/String;
-	public final fun getId-2Rbgges ()Ljava/lang/String;
+	public final fun getId ()Ljava/lang/String;
 	public fun toString ()Ljava/lang/String;
 }
 
--- a/detekt-api/api/detekt-api.api
+++ b/detekt-api/api/detekt-api.api
@@ -211,9 +211,9 @@ public final class io/gitlab/arturbosch/detekt/api/HasEntity$DefaultImpls {
 }
 
 public final class io/gitlab/arturbosch/detekt/api/Issue {
-	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Lio/gitlab/arturbosch/detekt/api/Rule$Id;Ljava/lang/String;)V
 	public final fun getDescription ()Ljava/lang/String;
-	public final fun getId ()Ljava/lang/String;
+	public final fun getId ()Lio/gitlab/arturbosch/detekt/api/Rule$Id;
 	public fun toString ()Ljava/lang/String;
 }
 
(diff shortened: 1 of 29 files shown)


detekt-api/api/detekt-api.api
```diff
@@ -310,7 +310,7 @@ public class io/gitlab/arturbosch/detekt/api/Rule : io/gitlab/arturbosch/detekt/
 	public fun getDefaultRuleIdAliases ()Ljava/util/Set;
 	public fun getFilters ()Lio/gitlab/arturbosch/detekt/api/internal/PathFilters;
 	public final fun getIssue ()Lio/gitlab/arturbosch/detekt/api/Issue;
-	public fun getRuleId ()Ljava/lang/String;
+	public fun getRuleId-2Rbgges ()Ljava/lang/String;
```
3flex, line 313: I don't like that this hash becomes part of the API due to https://kotlinlang.org/docs/inline-classes.html#mangling

It's transparent to Kotlin consumers but to mitigate please use `@JavaName` to disable mangling and give the function a stable name.

detekt-tooling/api/detekt-tooling.api
```diff
@@ -224,7 +224,7 @@ public final class io/github/detekt/tooling/api/spec/RulesSpec$RunPolicy$NoRestr
 
 public final class io/github/detekt/tooling/api/spec/RulesSpec$RunPolicy$RestrictToSingleRule : io/github/detekt/tooling/api/spec/RulesSpec$RunPolicy {
 	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
-	public final fun getRuleId ()Ljava/lang/String;
+	public final fun getRuleId-2Rbgges ()Ljava/lang/String;
 	public final fun getRuleSetId-7o6PJhY ()Ljava/lang/String;
```
3flex, line 228: Unrelated to this PR but we should give this a stable `@JvmName` as well

BraisGabin: Handled at #6855

detekt-api/api/detekt-api.api
BraisGabin: I was able to remove the other instance of this problem but I can't fix this one. It seems that `@get:JvmName` can't be applied to an `open` property. If the proposal at #6854 is accepted this property will be gone from `Rule` so that would "fix" the issue.

So, if there is not other way to fix this one I would vote to merge this. But it would be great to fix it.

cortinico: How about we just use a regular (data) class instead of a `value classfor the id?

BraisGabin: That would work too. And probably we shouldn't care that much about allocating that "extra" class. I'm going to do that change for this and for the `RuleSet.Id` too.

BraisGabin: I just implemented it on this PR and also created #6851 for `RuleSet.Id`.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
```diff
@@ -154,9 +155,27 @@ open class Rule(
             findings.add(finding)
         }
     }
-}
 
-/**
- * The type to use when referring to rule ids giving it more context then a String would.
- */
-typealias RuleId = String
+    class Id(val value: String) {
```
3flex, line 159: Annotate with `@Poko` to generate `hashCode`, `equals` and `toString`

3flex (approved):

cortinico (approved):

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
BraisGabin: I need to override `toString` but the automatic implementation of `equals` and `hashCode` is great :).

PR #6855, opened 2024-01-12
Make the abi a bit cleaner

BraisGabin:
#6849 introduced some hashes on our ABIs. This PR remove some of them.

I couldn't remove all of them because I can't add a `@JvmName` annotation inside an interface so I couldn't fix this one:

```
public abstract interface class io/gitlab/arturbosch/detekt/api/RuleSetProvider {
	public abstract fun getRuleSetId-7o6PJhY ()Ljava/lang/String;
	public abstract fun instance ()Lio/gitlab/arturbosch/detekt/api/RuleSet;
}
```

Any idea with that is more than welcome.


3flex (approved):
Thanks, it's good to clean up the ones that we can while it's easier.

PR #6857, opened 2024-01-13
Import custom assertions together with AssertJ assertions

3flex:
This allows assertThat to be used consistently regardless of which set of assertions is being used.


BraisGabin (approved):

marschwar:
@3flex Why (and how) does this work? Is it a compiler function to determine the best matching version?

```kotlin
import io.gitlab.arturbosch.detekt.test.assertThat

assertThat(listOf()).isNotNull    # uses the detekt custom asserter
assertThat(listOf("")).isNotNull  # uses the one from assertJ
```

3flex:
Yes, the compiler selects the most specific candidate function.

marschwar:
That is awesome. I was not aware of that. Thanks for the clarification.

PR #6860, opened 2024-01-13
Remove all the usages of mockk

cortinico:
While going through the codebase, I realized that we do use mockk only for two external types, `KtElement` and `PsiFile`.

Here I'm cleaning up the tests to instead use Fake versions of those that I have created. The implementation is mostly stubbed, and we can tweak it to fit our needs.

Removing mockk means our tests will be faster + it's a dependency less to maintain.

### DIFF
--- a/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/DetektMessageCollectorSpec.kt
+++ b/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/DetektMessageCollectorSpec.kt
@@ -1,10 +1,5 @@
 package io.gitlab.arturbosch.detekt.core
 
-import io.mockk.Called
-import io.mockk.every
-import io.mockk.mockk
-import io.mockk.slot
-import io.mockk.verify
 import org.assertj.core.api.Assertions.assertThat
 import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSeverity
 import org.junit.jupiter.api.BeforeEach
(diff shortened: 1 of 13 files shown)


detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/DetektMessageCollectorSpec.kt
```diff
@@ -13,12 +8,12 @@ import org.junit.jupiter.api.Test
 
 class DetektMessageCollectorSpec {
 
-    private lateinit var debugPrinter: (() -> String) -> Unit
+    private var debugPrinter = FakePrinter()
```
BraisGabin, line 11:
```suggestion
    private val debugPrinter = FakePrinter()
```

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/DetektMessageCollectorSpec.kt
```diff
@@ -64,7 +55,14 @@ class DetektMessageCollectorSpec {
 
         @Test
         fun `ignores the message`() {
-            verify { debugPrinter wasNot Called }
+            assertThat(debugPrinter.messages).isEmpty()
+        }
+    }
+
+    inner class FakePrinter : (() -> String) -> Unit {
```
BraisGabin, line 62:
```suggestion
   class FakePrinter : (() -> String) -> Unit {
```

BraisGabin (approved):

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/DetektMessageCollectorSpec.kt
```diff
@@ -64,7 +55,14 @@ class DetektMessageCollectorSpec {
 
         @Test
         fun `ignores the message`() {
-            verify { debugPrinter wasNot Called }
+            assertThat(debugPrinter.messages).isEmpty()
+        }
+    }
+
+   class FakePrinter : (() -> String) -> Unit {
```
cortinico, line 62:
```suggestion
    class FakePrinter : (() -> String) -> Unit {
```

PR #6861, opened 2024-01-13
Don't use value class for RuleSet.Id

BraisGabin:
This PR is to follow this idea: https://github.com/detekt/detekt/pull/6851#discussion_r1451588053

Right now the `value class` give us little value and make our ABI messy. So for now we can just use a normal `class`. Once we have our API "settled" we can rethinkg about this and go back to `value class` if needed.


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSet.kt
```diff
@@ -5,22 +5,33 @@ import io.gitlab.arturbosch.detekt.api.internal.validateIdentifier
...
     companion object {
-        @JvmName("invoke")
         operator fun invoke(id: Id, rules: List<(Config) -> Rule>): RuleSet {
             return RuleSet(id, rules.associateBy { it(Config.empty).ruleId })
         }
     }
 
-    @JvmInline
-    value class Id(val value: String) {
+    class Id(val value: String) {
```
3flex, line 15: Annotate with `@Poko` to generate `hashCode`, `equals` and `toString`

3flex (approved):

cortinico (approved):

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSet.kt
BraisGabin: Done :) (keep `toString`)

schalkms (approved):

PR #6866, opened 2024-01-14
Move `PathFilters` from detekt-api to detekt-utils

BraisGabin:
Also see #6814

~We had a little mess around this class. I splitted some tests and move them to the correct locations and then move `PathFilters` to `:detekt-tooling`.~

~We have far too much "utils" subprojects and I'm not 100% sure the responsability of which one, but this one seems the correct one. We don't want this at `:detekt-api` because it is only used by `:detekt-core` and by `:detekt-cli`. And it would be really nice to move it to `:detekt-core` but, as I said it is also used by `:detekt-cli`.~

~If someone create some documentation about which is the responsabilities of each subproject it would be great.~

Move `PathFilters` to `:detekt-utils`. This class is used on `detekt-cli` and `detekt-core` but we don't want it to be part of our public API so I'm moving it to `:detekt-utils`.

~Waiting for #7009~

### DIFF
--- a/detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/internal/PathFiltersSpec.kt
+++ b/detekt-api/src/test/kotlin/io/gitlab/arturbosch/detekt/api/internal/PathFiltersSpec.kt
@@ -1,46 +1,42 @@
 package io.gitlab.arturbosch.detekt.api.internal
 
 import org.assertj.core.api.Assertions.assertThat
-import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import kotlin.io.path.Path
 
 class PathFiltersSpec {
 
-    @Nested
-    inner class `parsing with different nullability combinations of path filters` {
-        @Test
-        fun `returns an empty path filter when includes are empty and excludes are empty`() {
-            val pathFilter = PathFilters.of(emptyList(), emptyList())
-            assertThat(pathFilter).isNull()
-        }
+    @Test
+    fun `returns an empty path filter when includes are empty and excludes are empty`() {
+        val pathFilter = PathFilters.of(emptyList(), emptyList())
+        assertThat(pathFilter).isNull()
+    }
 
-        @Test
-        fun `parses includes correctly`() {
-            val pathFilter = PathFilters.of(listOf("**/one/**", "**/two/**"), emptyList())
-            assertThat(pathFilter).isNotNull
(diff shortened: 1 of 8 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -214,12 +214,4 @@ class CliArgs {
                 FailureSeverity.Info -> FailOnSeverity(minSeverity.toSeverity())
             }
         }
-
-    companion object {
```
3flex, line 218: I agree to remove this, but it should be in a different PR, this is unrelated to moving PathFilters around

3flex:
Thinking about this, I don't think the includes/excludes belongs in detekt-tooling at all, and it makes sense to remove them from `ProjectSpec`. Calls to start an analysis session should include the list of files to scan, but the filtering of those should be done by whatever is calling tooling. Different clients will have different methods of filtering file lists.

As an example, the Gradle plugin doesn't pass exclude rules. It already filters and just sends the file list to the CLI. I think it makes more sense to move the path filters to CLI, and remove includes/excludes from tooling and bring everything up a level.

What do you think?

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
BraisGabin: I moved this to #6997. Right now it doesn't compiles because it needs the changes on this PR, I'll rebase it once this is merged.

BraisGabin:
> What do you think?

That you are completely right. I'm moving this PR to draft to block it an I'll create another PR removing includes/excludes from `ProjectSpec`.

BraisGabin:
I rebased this on top of #7009.

Now instead of moving `PathFilters` to `detekt-tooling` I move it to `detekt-utils`. We still use `PathFilters` on `detetk-cli` and `detekt-core` but now it is not part of any API, it is just an implementation detail.

I'm not a big fan of `detekt-utils` but it is handy to share code between parts of detekt without making it public API.

BraisGabin:
This PR changed a bit since the last review. @3flex can you take a look again?

3flex:
> I'm not a big fan of detekt-utils but it is handy to share code between parts of detekt without making it public API.

Completely agree - it's useful though unless someone comes up with a better idea. LGTM after merge conflict is addressed.

3flex (approved):

Issue #6854, opened 2024-01-12
Split what a `Rule` reports and what a `Report` receives


BraisGabin:
Right now the class that the `Rule` reports is what the `Report` receives. That class is `Finding`. I think that we should change that. The reason behind this is that the `Rule` needs to receive a lot of information just for that. And the core knows all that information. My idea:

- The `Rule` reports a `CodeSmell`, this "just" contains critical things that only the `Rule` can knows. For example: The error message and the `Entity`.
- The core transformate that `CodeSmell` in a `Finding`. That Finding will include all the `CodeSmell` information but it will also include things like:
   - RuleId
   - "InstaceRuleId"
   - RuleSetId
   - If the Finding was suppressed and the reason (no yet on detekt but this is somethign that we want on 2.0)
   - ...

This will make `Rule` way simpler and move the logic to `detekt-core` the place where it belongs.

I would love to get some feedback about this idea.

BraisGabin:
I created a fast draft to show how this change would be beneficial to detekt. This is just a Draft, the names are TERRIBLE and I didn't care about anything else than make the tests pass.

1. 17c5ef3320176247af8a4dc0b02d0d8b8d7512ae Not `CodeSmell` doesn't need to implement `Compactable`. That's not related with how we report findigns. They are just util functions for the reporting.
2. 1c0d1fb4bf4e7779e1840ce4609d1b8381d25267 We don't need to pass the `issue` isntance each time to the `CodeSmell`. Easier an less error prone API. Also this changes opens the door to remove the `issue` property from `Rule` directly. The rule just needs to provide it's description. We can construct the `Issue` at `detekt-core`.
3. 1398a71d542de5d6dbd466e77e9521a528e77b57 Remove all the code related with the `Severity` from `Rule` and move it to the core. The `Rule` doesn't care about the `Severity`. It reports the Finding, then is the core that reads the configuration and decides which is the `Severity` of that Finding. This reduces the code inside the api and also makes the code look better I think.

And, as I said, this is just a Draft. If you agree with this change we can move even more things.

To give context, this is related with the point 3 of #6738. It is difficult to add the new `id` to the `Rule` (ruleId vs ruleInstanceId). With this changes the `Rule` will not need to know those two ids. The `core` will handle them.

marschwar:
I like the direction. Do you also mean to convert `CodeSmell` into a data class or poko not implementing the `Finding` interface? I think that would help a lot.

[...]

PR #6871, opened 2024-01-15
Split `Finding` in two different classes

BraisGabin:
This is the first part to implement #6854

In this part I just split the classes and clean a bit them. We had a lot of things that we didn't need and hacks to avoid compatibility.

For now, we have the normal `Finding` that is what a `Rule` reports and then `Finding2` that is what detekt reports after collect all the necessary metadata.

`Finding2` is a temporal name! It will be changed to something else in future PRs. Probably `Issue` but for now I will keep it as `Finding2` meanwhile I do the other refactors that this PR allow us.


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -214,3 +220,22 @@ internal fun ProcessingSpec.workaroundConfiguration(config: Config): Config = wi
...
+
+private fun Finding.toFinding2(): Finding2 {
+    return when (this) {
+        is CorrectableCodeSmell -> Finding2Impl(issue, entity, message, references, severity, autoCorrectEnabled)
+
+        is CodeSmell -> Finding2Impl(issue, entity, message, references, severity)
+
+        else -> error("wtf?")
+    }
+}
```
BraisGabin, line 232: This code is *really* bad. I know. This will change in the next PRs.

marschwar:
To be absolutely honest I don't know if this is such a helpful next step forwards. Would it be possible to have the rule return a `CodeSmell` instead of a `Finding` and model the `CodeSmell` to only hold the necessary information and do the aggregation to a finding in the core?

marschwar (approved):
I think we have the same general understanding what the whole think should look like once it is finished. It just feels weird to approve a PR that adds throw away code. But the alternative of a giant PR that changes everything at once is even less ideal.

BraisGabin:
> I think we have the same general understanding what the whole think should look like once it is finished. It just feels weird to approve a PR that adds throw away code. But the alternative of a giant PR that changes everything at once is even less ideal.

Lately (probably this last year or so) I found out that I'm way faster refactoring writing code that I know that I'm going to remove in the next 4 commits than try to be smart and avoid that step.

PR #6872, opened 2024-01-15
Remove severity from the Rule

BraisGabin:
The Rule doesn't need to know its `Severity`. That can be handled by the core.


marschwar (approved):
I really like this.

PR #6873, opened 2024-01-15
Remove `Issue` from `Finding`

BraisGabin:
This can be handled by the core.


BraisGabin:
@detekt/maintainers could I get a review here? I know that the diff seems scary but the changes in all the rule files is really simple. Just remove the `Issue` from the `CodeSmell`. Also, if you find it difficult to review just tell me. I could split this up in two PRs. One with the change but keeping the `CodeSmell` constructor as `@Deprecated` and another just removing that deprecation constructor and all their usages. But I'm not sure if that would help to speed up the reviews or just slow them down.

Also I created this chain of PRs that they can be reviewed too. They are "draft" because they doesn't depend directly on `main`, but they are ready to review (also these ones are smaller): #6874, #6877, #6878, #6883 and #6884

schalkms (approved):
LGTM 👍
Sorry for the delay

PR #6875, opened 2024-01-16
Add ignoreAnnotatedFunctions to TooManyFunctions

### LINKED ISSUE #6516: Add ignoreAnnotatedFunctions to TooManyFunctions

Add `ignoreAnnotatedFunctions` (better name suggestions appreciated) config entry to `TooManyFunctions` rule that would exclude functions with specific annotations from the function count

## Context

We would like to exclude Compose preview functions from this rule. It is generally a good thing to have many Compose previews in file (to cover as many cases as possible), so we usually have several functions with actual UI code, then 10+ `@Preview` functions that just render preview for above code. We feel like those functions should not count into this rule as they do not contain any logic and do not actually make the file more complex.

I am prepared to offer a PR for that, if accepted.

matejdro:
This fixes #6516

### DIFF
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
@@ -159,7 +159,8 @@ class TooManyFunctions(config: Config) : Rule(
         ignoreDeprecated && function.hasAnnotation(DEPRECATED) -> true
         ignorePrivate && function.isPrivate() -> true
         ignoreOverridden && function.isOverride() -> true
-        else -> function.hasAnnotation(*ignoreAnnotatedFunctions.toTypedArray())
+        ignoreAnnotatedFunctions.any { function.hasAnnotation(it) } -> true
+        else -> false
     }
 
     companion object {
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
@@ -54,6 +54,9 @@ class TooManyFunctions(config: Config) : Rule(
     @Configuration("ignore overridden functions")
     private val ignoreOverridden: Boolean by config(false)
 
+    @Configuration("ignore functions annotated with these annotations")
+    private val ignoreAnnotatedFunctions: List<String> by config(emptyList())
+
     private var amountOfTopLevelFunctions: Int = 0
 
     override fun visitKtFile(file: KtFile) {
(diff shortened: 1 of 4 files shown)


detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
```diff
@@ -156,6 +159,7 @@ class TooManyFunctions(config: Config) : Rule(
         ignoreDeprecated && function.hasAnnotation(DEPRECATED) -> true
         ignorePrivate && function.isPrivate() -> true
         ignoreOverridden && function.isOverride() -> true
+        ignoreAnnotatedFunctions.any { function.hasAnnotation(it) } -> true
```
cortinico, line 162: Can we make sure that this work with both simple name, and fully qualified names?

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
```diff
@@ -54,6 +54,9 @@ class TooManyFunctions(config: Config) : Rule(
     @Configuration("ignore overridden functions")
     private val ignoreOverridden: Boolean by config(false)
 
+    @Configuration("ignore functions annotated with these annotations")
+    private val ignoreAnnotatedFunctions: List<String> by config(emptyList())
```
atulgpt, line 58: Adding composable `@Preview` by default in `config` will also solve many use cases of this rule by default config only

cortinico: In general we don't hardcode defaults that are framework specific.
Instead we should document this in `website/docs/introduction/compose.md`

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
matejdro: Can we get fully qualified name without type resolution? As far as I can see, this rule does not use it and we should not make hybrid rules (#2994)

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
matejdro: Good point, added a new commit with updated documentation

detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/TooManyFunctions.kt
cortinico: Yup you're right, let's keep it as it is 👍

BraisGabin (approved):

website/docs/introduction/compose.md
```diff
@@ -100,3 +100,15 @@ private fun FooLazyColumnPreview() { // Violation for FooLazyColumnPreview()
 #### Recommended configuration
 
 * Set `ignoreAnnotated` to `['Preview']`
+*
```
cortinico, line 103: This `*` should be removed

PR #6878, opened 2024-01-16
Remove Issue

BraisGabin:
We don't need `Issue` any more. Or you can think of this PR like `Issue` is renamed to `RuleInfo`. This frees this concept so in next PRs we can rename `Finding2` to `Issue`.


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding2.kt
```diff
@@ -3,13 +3,18 @@ package io.gitlab.arturbosch.detekt.api
 /**
  * Represents a problem detected by detekt on the source code
  *
- * A finding has an issue (information about the rule that detected the problem), a severity and a source code position
+ * A finding has the information about the rule that detected the problem, a severity and a source code position
  * described as an entity. Entity references can also be considered for deeper characterization.
  */
 interface Finding2 : HasEntity {
-    val issue: Issue
+    val rule: RuleInfo
```
3flex, line 10: Should this be called `ruleInfo`? There are other places where it should perhaps be renamed as well.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding2.kt
```diff
@@ -3,13 +3,18 @@ package io.gitlab.arturbosch.detekt.api
 /**
  * Represents a problem detected by detekt on the source code
  *
- * A finding has an issue (information about the rule that detected the problem), a severity and a source code position
+ * A finding has the information about the rule that detected the problem, a severity and a source code position
```
3flex, line 6:
```suggestion
 * A finding has information about the rule that detected the problem, a severity and a source code position
```

Reading this now it might make sense to remove the reference to "source code position" since it doesn't hold that directly, instead it's a list of entities that caused the finding.

3flex (approved):
Makes sense!

3flex (commented):
Actually I'll remove approval until this is out of draft

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding2.kt
BraisGabin: I'm not 100% sure to name it `ruleInfo`. It's true that is not a `Rule` but in this context you shouldn't look for one anyway. The main reason to call it `rule` is to how it looks and feels in their usages: `finding.rule.id` vs `finding.ruleInfo.id` or `finding.rule.description` vs `finding.ruleInfo.description`.

Does that have sense? What do you think?

schalkms (approved):

PR #6883, opened 2024-01-18
Add ruleSetId into `Finding2`

BraisGabin:
Add `RuleSet.Id` inside `Finding2`. This change by itself has little sense. But this unlocks a great simplification on the `Deteketion` implementations. This simplification is implemented in the draft #6884

### DIFF
--- a/detekt-api/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/test/TestFactory.kt
+++ b/detekt-api/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/test/TestFactory.kt
@@ -5,6 +5,7 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Finding2
 import io.gitlab.arturbosch.detekt.api.Location
 import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.RuleSet
 import io.gitlab.arturbosch.detekt.api.Severity
 import io.gitlab.arturbosch.detekt.api.SourceLocation
 import io.gitlab.arturbosch.detekt.api.TextLocation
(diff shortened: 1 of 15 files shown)


detekt-api/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/test/TestFactory.kt
```diff
@@ -18,7 +19,7 @@ fun createFinding(
     severity: Severity = Severity.Error,
     autoCorrectEnabled: Boolean = false,
 ): Finding2 = Finding2Impl(
-    rule = Finding2Impl.RuleInfo(Rule.Id(ruleName), "Description $ruleName"),
+    rule = Finding2Impl.RuleInfo(Rule.Id(ruleName), RuleSet.Id("ruleSetId"), "Description $ruleName"),
```
cortinico, line 22: Q: why this field is called `rule` and not `ruleInfo`

cortinico (approved):

detekt-api/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/test/TestFactory.kt
BraisGabin: I thought that the api would be easier. `issue.rule.id` I thought it would be easier than `issue.ruleInfo.id`. And that it would not be confusing because at an issue level it have little sense to have a real rule instance.

But 3flex and you think that `ruleInfo` is clearer or asked about the same. So I'm going to rename it :).

cortinico: Yeah I'd go for `ruleInfo` as we do have a class called `rule` and makes things confusing if the name/type are not aligned

BraisGabin: Done :)

PR #6884, opened 2024-01-18
Make `findings` a `List<Finding2>` instead of a `Map<RuleSet.Id, List<Finding2>>`

BraisGabin:
This change simplifies a lot `Detektion` and all their related code. We were using `Map<RuleSet.Id, List<Finding2>>` as the only way to carry the `RuleSet.Id` of a finding. Now we have that information inside `Finding2` so we don't need this Map anymore.

You can see that a lot of reports were just flattening that map because they don't need it. And the ones that really need it now it is as simple as add a `findings.groupBy { it.rule.ruleSetId }`.

We should even check if we need to `sort` the findings at the `core` level. But I didn't want to touch that part right now.

### DIFF
--- a/detekt-compiler-plugin/src/main/kotlin/io/github/detekt/compiler/plugin/internal/MessageCollectorExtensions.kt
+++ b/detekt-compiler-plugin/src/main/kotlin/io/github/detekt/compiler/plugin/internal/MessageCollectorExtensions.kt
@@ -21,7 +21,7 @@ fun MessageCollector.error(msg: String) {
 }
 
 fun MessageCollector.reportFindings(result: Detektion) {
-    for (finding in result.findings.values.flatten()) {
+    for (finding in result.findings) {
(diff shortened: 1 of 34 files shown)


detekt-compiler-plugin/src/main/kotlin/io/github/detekt/compiler/plugin/internal/MessageCollectorExtensions.kt
```diff
@@ -21,7 +21,7 @@ fun MessageCollector.error(msg: String) {
 }
 
 fun MessageCollector.reportFindings(result: Detektion) {
-    for (finding in result.findings.values.flatten()) {
+    for (finding in result.findings) {
```
cortinico, line 24:
```suggestion
    result.findings.forEach { finding ->
```

cortinico (approved):
Amazing cleanup! Just one minor nit

PR #6886, opened 2024-01-18
Reduce JDK coverage

### LINKED ISSUE #6682: Reduce JDK coverage on CI
## Expected Behavior

No unnecessary work is done when running the test suite on CI.

## Current Behavior

There are a total of 11 runs of the test suite in CI - one for each combination of LTS JDK and major OS (except JDK 8 on Windows).

## Context

With Java 21 adding another LTS version to test against we are likely wasting a lot of resources and time on redundant tests as pointed out by @BraisGabin.

Perhaps we can test only on the oldest and newest LTS JDK that detekt supports, or only on oldest + the current min JDK version supported by AGP, or some other defined strategy that reduces the size of the test matrix.

BraisGabin:
Closes #6682

Just to move this thing forward. What do you think about this?

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -21,11 +21,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, macos-latest, windows-latest ]
+        os: [ ubuntu-latest ]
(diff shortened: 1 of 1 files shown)


3flex (approved):
I think this is reasonable, thanks for pushing it along!

cortinico:
Great stuff!

Issue #5501, opened 2022-11-03
Running detektMain in a project with Gradle precompiled script plugins fails


eygraber:
## Your Environment

* Version of detekt used: 1.22.0-RC2
* Version of Gradle used (if applicable): 7.5.1

Running `detektMain` on [this repro project](https://github.com/detekt/detekt/files/9932092/detekt-repro.zip) results in the following error:

```
Execution failed for task ':detektMain'.
> Front-end Internal error: Failed to analyze declaration Com_eygraber_validate_gradle_properties_gradle
  File being compiled: (2,44) in //home/eli/workspace/detekt-repro/src/main/kotlin/com.eygraber-validate-gradle-properties.gradle.kts
  The root cause org.jetbrains.kotlin.resolve.lazy.NoDescriptorForDeclarationException was thrown at: org.jetbrains.kotlin.resolve.lazy.BasicAbsentDescriptorHandler.diagnoseDescriptorNotFound(AbsentDescriptorHandler.kt:18)

```

eygraber:
Any plans to address this?

eygraber:
Ping

ankushg:
Also running into this issue!

Also for anyone who hasn't seen the relevant Slack thread: https://slack-chats.kotlinlang.org/t/4963811/running-into-issues-when-running-detektmain-on-a-project-tha

BraisGabin:
Could you check #6198? Today there were some progress on that issue and it seems related.

ankushg:
> Could you check #6198? Today there were some progress on that issue and it seems related.

Our project is a multi-module Android project, and it looks like most of the comments in there are around Spring dependency management

Either way, I tried adding this, but had no luck:

```kt
    configurations.matching { it.name == "detekt" }.all {
        resolutionStrategy.eachDependency {
            if (requested.group == "org.jetbrains.kotlin") {
                useVersion("1.9.0") // detekt 1.23.1
            }
        }
    }

```

eygraber:
@BraisGabin I tried this in a project with Kotlin 1.9.0 and Gradle 8.3 (which uses Kotlin 1.9.0) and still have this issue.

cortinico:
> @BraisGabin I tried this in a project with Kotlin 1.9.0 and Gradle 8.3 (which uses Kotlin 1.9.0) and still have this issue.

Is this a blank project?

eygraber:
No I ran it on https://github.com/eygraber/gradle-conventions

[...]

PR #6892, opened 2024-01-22
Add Gradle pre-commit hook to docs

matejdro:
Currently, pre-commit hook page only mentions CLI steps. This requires user to maintain separate detekt binary in the project and it's not compatible with type resolution.

With this PR, I have also added instructions on how to only scan staged files with Gradle.

Setup is kinda verbose, since you have to run through several hoops to get it to work with configuration cache. This is also kts only (it can probably be converted to groovy, but I'm not very well versed in groovy).

### DIFF
--- a/website/docs/gettingstarted/git-pre-commit-hook.md
+++ b/website/docs/gettingstarted/git-pre-commit-hook.md
@@ -40,7 +40,7 @@ A special thanks goes to Mohit Sarveiya for providing this shell script.
 You can watch his excellent talk about **Static Code Analysis For Kotlin** on 
 [YouTube](https://www.youtube.com/watch?v=LT6m5_LO2DQ).
 
-## Only run on staged files
+## Only run on staged files - CLI
 
 It is also possible to use [the CLI](/docs/gettingstarted/cli) to create a hook that only runs on staged files. This has the advantage of speedier execution, by running on fewer files and avoiding the warm-up time of the gradle daemon.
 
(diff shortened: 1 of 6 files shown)


matejdro:
Any ideas what Vercel error is and how can I fix it? I've only updated markdown files

website/docs/gettingstarted/git-pre-commit-hook.md
```diff
@@ -80,3 +80,112 @@ if [ $EXIT_CODE -ne 0 ]; then
   exit $EXIT_CODE
 fi
 ```
+
+## Only run on staged files - Gradle
```
cortinico, line 84: This part should be moved before the CLI, as the initial script suggests to use `./gradlew`.

So this page will look like:
1. Run `gradlew` on everything
2. Run `gradlew` only on modified files
3. Run CLI only on modified files

website/docs/gettingstarted/git-pre-commit-hook.md
```diff
@@ -80,3 +80,112 @@ if [ $EXIT_CODE -ne 0 ]; then
...
+
+If the CLI is not your cup of tea (you don't want to keep separate CLI binary set up, you want type resolution etc.),
```
cortinico, line 86: This is not the style we use in the rest of the docs, could we simplify a bit?

website/docs/gettingstarted/git-pre-commit-hook.md
```diff
@@ -80,3 +80,112 @@ if [ $EXIT_CODE -ne 0 ]; then
   exit $EXIT_CODE
 fi
 ```
+
+## Only run on staged files - Gradle
+
+If the CLI is not your cup of tea (you don't want to keep separate CLI binary set up, you want type resolution etc.), 
+you can also configure Gradle detekt to only run on staged files.
+
+First, we need to declare `GitPreCommitFilesTask` task - a gradle task that will retrieve list of staged files
```
cortinico, line 89: I think it would be useful to specify where to create this file for folks that are not used to precompiled script plugins

cortinico (commented):
Thanks for adding this @matejdro !

website/docs/gettingstarted/git-pre-commit-hook.md
matejdro: Thanks, updated that part.

website/docs/gettingstarted/git-pre-commit-hook.md
matejdro: Thanks, updated that part.

website/docs/gettingstarted/git-pre-commit-hook.md
matejdro: Moved to the top

schalkms (commented):
@cortinico Is there something left to do in this PR to get it merged?
Otherwise, we should give it a go, since it improves detekt's doc.

cortinico:
@schalkms yes versioned docs were missing for all the 1.23.x versions.
I've udpated them all and we can merge this now 👍

cortinico (approved):

matejdro:
By the way, since making this PR, I have found a couple of improvements to this approach:

* a cleaner way to get git info
* exclude gradle kts files if pre commit hook is running with type resolution (as a workaround for https://github.com/detekt/detekt/issues/5501)

If you want, I can include those on Monday and it can be merged after that.

cortinico:
> If you want, I can include those on Monday and it can be merged after that.

Yes please 👍

matejdro:
Done

website/docs/gettingstarted/git-pre-commit-hook.md
```diff
@@ -40,11 +40,88 @@ A special thanks goes to Mohit Sarveiya for providing this shell script.
...
+        .map { outputText ->
+            outputText.trim()
+                .split("\n")
+                .filter { it.isNotBlank() }
+                .map { File(rootDir, it) }
+        }
+}
+```
+
+Then we need to configure `Detekt` task and change its `source` from the entire `src` foler (by default) to only set of
```
BraisGabin, line 66:
```suggestion
Then we need to configure `detekt` task and change its `source` from the entire `src` folder (by default) to only set of
```

website/docs/gettingstarted/git-pre-commit-hook.md
```diff
@@ -40,11 +40,88 @@ A special thanks goes to Mohit Sarveiya for providing this shell script.
...
+files that have been staged by git. Paste following into your project's `build.gradle.kts`:
+
+```kotlin
+tasks.withType<io.gitlab.arturbosch.detekt.Detekt>().configureEach {
+    if (project.hasProperty("precommit")) {
+        dependsOn(gitPreCommitFileList)
```
BraisGabin, line 72: Is this needed? And, if so, where is this task defined?

BraisGabin (commented):
My comments also apply to the versioned files

website/docs/gettingstarted/git-pre-commit-hook.md
matejdro: Ah, forgot to remove this one. Will delete it.

BraisGabin (approved):

PR #6893, opened 2024-01-22
Report rule's default severity in sarif

matejdro:
Some tools (such as [Github's sarif to markdown](https://github.com/security-alert/security-alert/blob/36dcb0cabe078e81f9b776e9a435b3fe1dfe2dbd/packages/sarif-to-markdown/src/sarif-to-markdown.ts#L204C31-L204C51)) seem to rely on `defaultConfiguration.level` being present on rules in sarif files.

This PR adds that field to detekt's sarif output.

I've copied `computeSeverity` method from [Analyzer.kt](https://github.com/detekt/detekt/blob/36e410b7347569ab8c1f71415e33e4fc2d5b0db6/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt#L251). Since this method is in another module and it is fairly short, it seemed prudent to just create a copy. But I'm open to suggestions on how to reuse it.

### DIFF
--- a/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
+++ b/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
@@ -38,7 +38,7 @@ class SarifOutputReport : BuiltInOutputReport, OutputReport() {
                 if (!it.endsWith("/")) "$it/" else it
             }
 
-        this.config = config
+        this.config = context.config
     }
 
     override fun render(detektion: Detektion): String {
--- a/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
+++ b/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
@@ -28,7 +28,7 @@ class SarifOutputReport : BuiltInOutputReport, OutputReport() {
     override val id: String = "sarif"
 
     private var basePath: String? = null
-    private var config: Config? = null
+    private lateinit var config: Config
 
     override fun init(context: SetupContext) {
         this.basePath = context.getOrNull<Path>(DETEKT_OUTPUT_REPORT_BASE_PATH_KEY)
@@ -42,8 +42,6 @@ class SarifOutputReport : BuiltInOutputReport, OutputReport() {
     }
 
     override fun render(detektion: Detektion): String {
-        val config = config ?: Config.empty
-
         val version = whichDetekt()
         val sarifSchema210 = SarifSchema210(
             schema = "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
--- a/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
+++ b/detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
@@ -7,6 +7,7 @@ import io.github.detekt.sarif4k.SarifSerializer
 import io.github.detekt.sarif4k.Tool
 import io.github.detekt.sarif4k.ToolComponent
 import io.github.detekt.sarif4k.Version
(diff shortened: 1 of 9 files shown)


BraisGabin (commented):
Could you add tests to ensure this change? Also that would help to understand exactly what changed at the output

matejdro:
It seems like this part has not been tested before, so change will not be very well visible. But I've added it to the test now.

detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
```diff
@@ -27,6 +28,7 @@ class SarifOutputReport : BuiltInOutputReport, OutputReport() {
     override val id: String = "sarif"
 
     private var basePath: String? = null
+    private var config: Config? = null
```
BraisGabin, line 31:
```suggestion
    private lateinit var config: Config
```

detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
```diff
@@ -35,9 +37,13 @@ class SarifOutputReport : BuiltInOutputReport, OutputReport() {
             ?.let {
                 if (!it.endsWith("/")) "$it/" else it
             }
+
+        this.config = config
     }
 
     override fun render(detektion: Detektion): String {
+        val config = config ?: Config.empty
+
```
BraisGabin, line 46:
```suggestion
```

detekt-report-sarif/src/test/resources/relative_path.sarif.json
```diff
@@ -83,6 +83,17 @@
           "name": "detekt",
           "organization": "detekt",
           "rules": [
+            {
+              "defaultConfiguration": {
+                "level": "error"
```
BraisGabin, line 88: Can we have a test where this is "warning" (for example).

BraisGabin (commented):
Thanks for the tests. We need to improve the tests of this part of the code but probably that's a different PR. For example what happen with disabled rules and so on.

But, to merge this PR, I think that we just need to handle the comments I left.

detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
matejdro: I found that in some cases init is not called, which would left that variable uninitialized. Probably basePath is nullable for the same reason?

detekt-report-sarif/src/test/resources/relative_path.sarif.json
matejdro: Added a test

detekt-report-sarif/src/main/kotlin/io/github/detekt/report/sarif/SarifOutputReport.kt
BraisGabin: 😱 that's a serious bug. `init` should *always* be called.

`basePath` is nullable because `null` is a valid value for it. Right now we allow to run detekt without defining a `basePath` (maybe we could change that for 2.0 but no plans there right now).

So I insist to make it `lateinit var` and we should fix any case where `init` is not called.

BraisGabin (approved):

matejdro:
Thanks for merging!

Any chance a fix for this is released before 2.0.0? From what I see, 2.0.0 is ways of?

BraisGabin:
I think that's going to be difficult. I'm not sure how easy will be to merge this commit inside the `release/1.x` branch. I imagine that if you create a PR pointing there fixing the conflicts that could be doable but the last word here comes from @cortinico because he is the one that does the not that easy releases of detekt.

matejdro:
I am more than happy to create release PR if necessary.

cortinico:
> I am more than happy to create a PR for relase branch if I get the green light.

Please send over a PR and we'll get this shipped out, as it's more of a bugfix rather than a new feature, so we could have it in `.5`

PR #6894, opened 2024-01-22
Merge sarifs from different tools into different runs

matejdro:
Currently, sarif merger task will only merge `results`. This works when merging results from a single tool, but breaks down if we want to merge results from multiple different tools (for example Detekt and Android Lint).

PR fixes that by merging results from every tool independently and then storing merged result from every tool in their own respective `runs` entry in final sarif.


matejdro:
It seems that some tools (like Android Lint) only report detected rules in their respective serif files, not all rules. So I've also added rules merging.

BraisGabin:
🤔 This has a lot of sense but, should be this inside `detekt`? I mean, we could create a different gradle plugin that what it does is to merge sarif files from any source.

matejdro:
While this is not 100% detekt related, detekt already has this task conveniently in and this PR only makes minor tweaks to this existing task, so I think it's worth it. But I agree, that if this was a big change, it might make more sense to split it into different plugin.

BraisGabin:
My concerns are that adding this could add even more maintenance to this repo. My proposal would be to implement the merge functionality inside [detekt/sarif4k](https://github.com/detekt/sarif4k) and then here we just use that. This will fix both problems.

[detekt/sarif4k](https://github.com/detekt/sarif4k) is not related with [detekt/detekt](https://github.com/detekt/detekt). So it have for [detekt/sarif4k](https://github.com/detekt/sarif4k) to take into account multiple tools. And here we will just use what `sarif4k` provides us so we will support multiple tools "for free".

I would like to know what other @detekt/maintainers think about this.

TWiStErRob:
I agree it doesn't belong here, the task you modified is meant to merge detekt reports. I agree that this could/should exist somewhere, but not detekt repo. Sarif4k is a KMP parser/writer, so adding a Gradle task to it might be also scope creep. However the merging logic could be maintained there. I've also hacked together a [lint merger](https://github.com/TWiStErRob/net.twisterrob.sun/blob/v1.2.2/gradle/plugins/src/main/kotlin/net/twisterrob/sun/plugins/tasks/MergeLintSarifReportsTask.kt) for my project(s).

Note: you don't need a single file to upload sarif to GitHub. You can have multiple steps doing separate uploads.

BraisGabin:
> Sarif4k is a KMP parser/writer, so adding a Gradle task to it might be also scope creep.

My idea was add there just the logic. Given two `SarifSchema210` instances, merge them into one. Nothing related with gradle. Then we can keep (for now at least) this gradle task and get the support for multiple rules for free. Later we could discuss if we want to remove the task all together.

BraisGabin:
I'm closing this in favor of https://github.com/detekt/sarif4k/issues/82. @matejdro if you want go ahead and move the code inside this repo to merge sarifs to the other one. Once that's done we can just remove the code from this repo and just delagate the work to the other library.

PR #6913, opened 2024-01-26
Enable gradle worker api by default on detekt tasks

### LINKED ISSUE #2957: Changing any core class invalidates cached detekt classloader

### LINKED ISSUE #6887: Enable Gradle Worker API by default
TL;DR: I think that we should enable Gradle Worker API by default on detekt.

#4128 introduced the option to run detekt by using the Worker API. This allows gradle to execute `detektMain` and `detektTest` in parallel and also execute `detektMain` in parallel with `compileTestKotlin`. This helps gradle to run more tasks in parallel.

For example, in a normal build as soon as `compileKotlin` finishes it can start `detektMain` and `compileTestKotlin` at the same time. Also, as soon as `compileTestKotlin` ends the task `detektTest` can start. Before these 4 task should run one after the other.

At the beginning of #4128 I did some benchmarks and we didn't saw any improvement but I executed other benchmarks today and the picture is really different (also I'm benchmarking slightly different scenarios, but I think that these ones are "more "real").

New benchmarks:

I created 3 different scenarios:
- Execute `detektMain` and `detektTest` on ~250 modules (this is a real project)
- Execute `detektMain` and `detektTest` on a sinle module (same project)
- Execute `detektMain` on a sinle module (same project)

And those scenarios with and without Gradle Worker API enabled.

Also, between each run I removed the folder inside `build/report/detekt` to ensure that detekt is executed again and also with `--no-build-cache`.

Results:

|                        | `detektMain detektTest` | `:module:detektMain :module:detektTest` | `:module:detektMain` |
|------------------------|--------------------------|------------------------------------------|-----------------------|
| without workers (mean) | 39684,6                  | 1507,7                                   | 1404,0                |
| with workers (mean)    | 29129,8                  | 1573,1                                   | 1669,2                |
| difference             | -27%                     | ~+4%~ (no significant)                                      | +19%                   |

So, we end up that if you executes detekt several times the parallelization wins and makes huge improvements (27%). If you only executes detekt in one single module `main` and `test` it runs more less the same. And if you executes detekt only once it's slower by (19%).

Even we have one case that improves and other that don't I would say that we should also see the total numbers. In one we gain 10 seconds and in the other we lose 265ms. And, in general, I think that no one will only run detekt on `main` and not on `test`.

For that reasons I think that we should enable it by default. As an extra this would close #2957. And that would improve the workflow for us and the developers of third party rules.

BraisGabin:
Closes #6887
Closes #2957

To get the reasoning behind this change read the related issue.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -151,7 +151,8 @@ class DetektPlugin : Plugin<Project> {
 internal const val CONFIGURATION_DETEKT = "detekt"
 
 internal fun ProviderFactory.isWorkerApiEnabled(): Boolean {
-    return gradleProperty("detekt.use.worker.api").getOrElse("false") == "true"
+    val defaultValue = isGradleVersionAtLeast(major = 7, minor = 6).toString()
+    return gradleProperty("detekt.use.worker.api").getOrElse(defaultValue).toBoolean()
 }
 
 @Incubating
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -13,7 +13,6 @@ import org.gradle.api.Incubating
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.provider.ProviderFactory
-import org.jetbrains.kotlin.gradle.utils.isGradleVersionAtLeast
 import java.net.URL
 import java.util.jar.Manifest
 
@@ -151,8 +150,7 @@ class DetektPlugin : Plugin<Project> {
 internal const val CONFIGURATION_DETEKT = "detekt"
 
 internal fun ProviderFactory.isWorkerApiEnabled(): Boolean {
-    val defaultValue = isGradleVersionAtLeast(major = 7, minor = 6).toString()
-    return gradleProperty("detekt.use.worker.api").getOrElse(defaultValue).toBoolean()
+    return gradleProperty("detekt.use.worker.api").getOrElse("true").toBoolean()
 }
 
 @Incubating
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -12,6 +12,8 @@ import io.gitlab.arturbosch.detekt.internal.DetektPlain
 import org.gradle.api.Incubating
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.provider.ProviderFactory
+import org.jetbrains.kotlin.gradle.utils.isGradleVersionAtLeast
 import java.net.URL
 import java.util.jar.Manifest
 
@@ -147,7 +149,10 @@ class DetektPlugin : Plugin<Project> {
 }
 
 internal const val CONFIGURATION_DETEKT = "detekt"
-internal const val USE_WORKER_API = "detekt.use.worker.api"
+
+internal fun ProviderFactory.isWorkerApiEnabled(): Boolean {
+    return gradleProperty("detekt.use.worker.api").getOrElse("false") == "true"
+}
 
 @Incubating
 fun getSupportedKotlinVersion(): String {
--- a/detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
+++ b/detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
@@ -22,7 +22,7 @@ constructor(
     val gradleVersionOrNone: String? = null,
     val dryRun: Boolean = false,
     val jvmArgs: String = "-Xmx2g -XX:MaxMetaspaceSize=1g",
-    gradleProperties: Map<String, String> = emptyMap(),
+    val gradleProperties: Map<String, String> = emptyMap(),
     val customPluginClasspath: List<File> = emptyList(),
     val projectScript: Project.() -> Unit = {}
 ) {
@@ -35,9 +35,6 @@ constructor(
         rootProject.name = "rootDir-project"
         include(${projectLayout.submodules.joinToString(",") { "\"${it.name}\"" }})
     """.trimIndent()
-
-    private val propertiesContent = gradleProperties.toList()
-        .joinToString(separator = "\n") { (key, value) -> "$key=$value" }
 
     @Language("xml")
     private val baselineContent = """
@@ -70,7 +67,6 @@ constructor(
     fun setupProject() {
         writeProjectFile(buildFileName, mainBuildFileContent)
         writeProjectFile(SETTINGS_FILENAME, settingsContent)
-        writeProjectFile(PROPERTIES_FILENAME, propertiesContent)
         configFileOrNone?.let { writeProjectFile(it, configFileContent) }
         baselineFiles.forEach { file -> writeProjectFile(file, baselineContent) }
         projectLayout.srcDirs.forEachIndexed { srcDirIdx, sourceDir ->
@@ -142,6 +138,7 @@ constructor(
             if (dryRun) {
                 add("-Pdetekt-dry-run=true")
             }
+            addAll(gradleProperties.toList().map { (key, value) -> "-P$key=$value" })
             addAll(tasks.toList())
         }
 
@@ -181,7 +178,6 @@ constructor(
 
     companion object {
         private const val SETTINGS_FILENAME = "settings.gradle"
-        private const val PROPERTIES_FILENAME = "gradle.properties"
         private const val DETEKT_TASK = "detekt"
     }
 }
--- a/detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
+++ b/detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
@@ -22,6 +22,7 @@ constructor(
     val gradleVersionOrNone: String? = null,
     val dryRun: Boolean = false,
     val jvmArgs: String = "-Xmx2g -XX:MaxMetaspaceSize=1g",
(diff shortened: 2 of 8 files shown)


3flex (changes requested):
Maybe hold off, see #6964 and https://github.com/gradle/gradle/issues/28034

3flex:
Do you want to update the report merge docs to say that the worker API should be disabled if using the report merge feature? Then we can merge.

BraisGabin:
I was checking thy this failed on jdk11 and jdk8. And I found this: https://ge.detekt.dev/s/6lrnem2j7n3qs/tests/overview?outcome=FAILED

Two tests failed because of, I assume, https://github.com/gradle/gradle/issues/28034. So it seems that https://github.com/gradle/gradle/issues/28034 is related with the JDK version.

And the other one seems that we don't support the worker API on old versions of gradle: https://ge.detekt.dev/s/6lrnem2j7n3qs/tests/task/:detekt-gradle-plugin:functionalTest/details/io.gitlab.arturbosch.detekt.GradleVersionSpec/runsOnOldestSupportedGradleVersion()?top-execution=1 Should we check on that or, because this will be released on 2.0.0, move up our min supported version?

BraisGabin:
I need to find a way to disable Worker API for the merge tests otherwise they are not going to pass.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -147,7 +149,11 @@ class DetektPlugin : Plugin<Project> {
 }
 
 internal const val CONFIGURATION_DETEKT = "detekt"
-internal const val USE_WORKER_API = "detekt.use.worker.api"
+
+internal fun ProviderFactory.isWorkerApiEnabled(): Boolean {
+    val defaultValue = isGradleVersionAtLeast(major = 7, minor = 6).toString()
```
BraisGabin, line 154: Gradle worker API doesn't work on any version below `7.6` so we don't enable it by default for those versions.

3flex: What changed in 7.6? Or did you discover that through testing? There's nothing in the 7.6 release notes about workers: https://docs.gradle.org/7.6/release-notes.html

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
```diff
@@ -65,6 +69,7 @@ constructor(
     fun setupProject() {
         writeProjectFile(buildFileName, mainBuildFileContent)
         writeProjectFile(SETTINGS_FILENAME, settingsContent)
+        writeProjectFile(PROPERTIES_FILENAME, propertiesContent)
```
3flex, line 72: Instead of writing a properties file you should be able to pass `-Pdetekt.use.worker.api=true/false` as a CLI argument which avoids writing another file just for this.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
BraisGabin: Yes, I was testing and 7.6 was the first one that work. No idea why it doesn't work on prior versions.

detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
BraisGabin: Done

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
BraisGabin: Probably we should look at this and fix it. But I think that is a different issue.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -12,6 +12,8 @@ import io.gitlab.arturbosch.detekt.internal.DetektPlain
 import org.gradle.api.Incubating
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.provider.ProviderFactory
+import org.jetbrains.kotlin.gradle.utils.isGradleVersionAtLeast
```
3flex, line 16: Sorry I didn't pick this up earlier, but you should use the Gradle methods for version comparison instead of this utility method in KGP:

`GradleVersion.current() >= GradleVersion.version("7.6")`

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
3flex: I'm 99% sure this is an issue with our test setup and not a problem with this PR. Gradle is passing a Gradle plugin variant that's compatible with Gradle 8.2+ which fails in some scenarios on this earlier Gradle version. I ran into this in #7025 and #6489.

I'll open a new PR to address it.

3flex: Should be fixed by #7036. I've also rebased #6489 on top of #7036 to prove that it works (functional tests used to fail because of a similar issue).

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
3flex: Though as mentioned in the other comment this version check shouldn't be required at all.

BraisGabin: Removed

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
BraisGabin: This code was removed, now Worker API is always enabled by default.

3flex:
Annoying the build is still failing. I saw this in the log:

```
    w: Runtime JAR files in the classpath should have the same version. These files were found in the classpath:
        /home/runner/.gradle/wrapper/dists/gradle-6.8.3-bin/7ykxq50lst7lb7wx1nijpicxn/gradle-6.8.3/lib/kotlin-stdlib-1.4.20.jar (version 1.4)
        /home/runner/.gradle/wrapper/dists/gradle-6.8.3-bin/7ykxq50lst7lb7wx1nijpicxn/gradle-6.8.3/lib/kotlin-stdlib-common-1.4.20.jar (version 1.4)
        /home/runner/.gradle/wrapper/dists/gradle-6.8.3-bin/7ykxq50lst7lb7wx1nijpicxn/gradle-6.8.3/lib/kotlin-stdlib-jdk7-1.4.20.jar (version 1.4)
        /home/runner/.gradle/wrapper/dists/gradle-6.8.3-bin/7ykxq50lst7lb7wx1nijpicxn/gradle-6.8.3/lib/kotlin-stdlib-jdk8-1.4.20.jar (version 1.4)
        /home/runner/.gradle/wrapper/dists/gradle-6.8.3-bin/7ykxq50lst7lb7wx1nijpicxn/gradle-6.8.3/lib/kotlin-reflect-1.4.20.jar (version 1.4)
        /home/runner/work/detekt/detekt/detekt-gradle-plugin/build/tmp/functionalTestMinSupportedGradle/work/.gradle-test-kit/caches/jars-8/06130c2f1c60b901150c43a397642dca/kotlin-stdlib-1.9.20.jar (version 1.9)
        /home/runner/work/detekt/detekt/detekt-gradle-plugin/build/tmp/functionalTestMinSupportedGradle/work/.gradle-test-kit/caches/jars-8/a369b37b4f997a63d005d4a5f51fd41f/kotlin-reflect-1.9.20.jar (version 1.9)
```

So my fix is incomplete but it's strange that this is only occurring in this PR and not on `main`

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
3flex: You might need to restore this check.

To isolate the issue I tried using 1.23.5 in a standalone Gradle project running 6.8.3. That failed. The earliest version that worked for me is 7.5. So there's an existing issue unrelated to the Gradle version test split, and not directly related to this PR.

I think enabling where possible with this check restored (changed to 7.5) makes sense but it's odd that it's not working on older versions. The whole point of the process isolation mode is that the worker's classpath is isolated from Gradle's, so the Gradle version shouldn't matter here.

BraisGabin: I must say that I'm a bit lost on this regard. I don't fully understand all the picture so it's difficult to me to know what's going on.

I'm going to restore it using `7.5` as the minimum version then :)

3flex:
I think I've had a fundamental misunderstanding of what process isolation actually achieves when using workers. The worker still has Gradle runtime in its classpath, which is why there's different behaviour when running with different Gradle versions. In many cases things work, because the Kotlin version on the Gradle classpath doesn't conflict with detekt's, but in other cases there's an issue because of Kotlin runtime behaviour changes and Gradle 6.8 embeds Kotlin 1.4.20.

Worker classloader isolation makes it possible to run tasks with different libraries to the buildscript classpath. Process isolation extends this to allow running tasks with different JDK or system properties to those used for Gradle itself.

Neither of these quite describes what we need to execute detekt.

TL;DR: The current worker implementation includes Gradle's runtime which conflicts with detekt's runtime. The worker implementation needs to change before we can use it.

3flex (changes requested):
I'd unfortunately consider this blocked until https://github.com/gradle/gradle/issues/28478 is fixed in Gradle.

BraisGabin:
After #7060 I'll benchmark these again to see how it behaves.

BraisGabin:
I move to draft this again, I need to perform benchmarks again (Note for myself)

PR #6914, opened 2024-01-27
Remove `autoCorrect` from rule set level

### LINKED ISSUE #1466: Cannot enable autoCorrect only for single formatting rule
## Expected Behavior

Using the configuration

```
formatting:
  autoCorrect: false
  ImportOrdering:
    autoCorrect: true
```

I'd expect `autoCorrect` to be enabled, but *only* for fixes to import ordering.

## Observed Behavior

Nothing is auto-corrected when running the `detekt` Gradle task.

## Steps to Reproduce

Checkout branch *detekt-rc-13-bug* of https://github.com/heremaps/oss-review-toolkit and run `./gradlew detekt`. Compare to the output with

```
formatting:
  autoCorrect: true
  ImportOrdering:
    autoCorrect: true
```

to see there actually are imports that *should* get fixed.

## Your Environment

* Version of detekt used: 1.0.0-RC13 (with `buildUponDefaultConfig = true`)
* Version of Gradle used (if applicable): 5.2.1
* Link to your project (if it's a public repository): https://github.com/heremaps/oss-review-toolkit/tree/detekt-rc-13-bug

BraisGabin:
Fixes #1466

Based on the #1466 discussion we want that a rule with `autoCorrect: true` autocorrects even if its rule set says `autoCorrect: false`. Then there is no reason to keep `autoCorrect` at rule set level. If you want to autoCorrect you enable it at "detekt" level and the the rules decide if they should autoCorrect or not. If you don't want to autoCorrect you just not enable it at detekt level and that's it.

For that reason I'm removing this intermediate flag.

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -67,7 +67,6 @@ exceptions:
 formatting:
   active: true
   android: false
-  autoCorrect: true
   BinaryExpressionWrapping:
     active: true
   BlankLineBeforeDeclaration:
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
@@ -52,8 +52,7 @@ open class Rule(
     private val ruleSetId: RuleSet.Id? get() = config.parent?.parentPath?.let(RuleSet::Id)
 
     val autoCorrect: Boolean
-        get() = config.valueOrDefault(Config.AUTO_CORRECT_KEY, false) &&
-            (config.parent?.valueOrDefault(Config.AUTO_CORRECT_KEY, true) != false)
+        get() = config.valueOrDefault(Config.AUTO_CORRECT_KEY, false)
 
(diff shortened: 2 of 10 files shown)


3flex:
That wasn't my understanding of the discussion - instead the intermediate one should be acting as a default which is overridden by the specific rules if necessary.

See the examples provided at the end of this comment https://github.com/detekt/detekt/issues/1466#issuecomment-462255315

That makes it really easy to say "I want autocorrect enabled, but I don't like it for rule X or y so I'll turn it on at the ruleset level and disable it only for the rules I don't want it for"

BraisGabin:
> That wasn't my understanding of the discussion - instead the intermediate one should be acting as a default which is overridden by the specific rules if necessary.
> 
> See the examples provided at the end of this comment [#1466 (comment)](https://github.com/detekt/detekt/issues/1466#issuecomment-462255315)
> 
> That makes it really easy to say "I want autocorrect enabled, but I don't like it for rule X or y so I'll turn it on at the ruleset level and disable it only for the rules I don't want it for"

I'm going to close this for now and continue the discussion on that issue.

PR #6931, opened 2024-02-01
Change how we decide if a rule should autocorrect or not

### LINKED ISSUE #1466: Cannot enable autoCorrect only for single formatting rule

BraisGabin:
Fixes #1466

Implement what #1466 says has one side effect. The rulsets with rules autocorrectables and other no autocorrectables we should define in all of them if they are auto correctables or not. Otherwise all of them are autocorrectables.

This PR is the successor of #6914

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
@@ -52,8 +52,10 @@ open class Rule(
     private val ruleSetId: RuleSet.Id? get() = config.parent?.parentPath?.let(RuleSet::Id)
 
     val autoCorrect: Boolean
-        get() = config.valueOrDefault(Config.AUTO_CORRECT_KEY, false) &&
-            (config.parent?.valueOrDefault(Config.AUTO_CORRECT_KEY, true) != false)
+        get() = config.valueOrDefault(
+            Config.AUTO_CORRECT_KEY,
+            (config.parent?.valueOrNull<Boolean>(Config.AUTO_CORRECT_KEY)) == true,
+        )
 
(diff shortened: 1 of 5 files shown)


BraisGabin:
To be honest I don't like this too much. Now it seems that you can enable the autoCorrect for `ClassName`. And if you do so... well nothing happens. detekt will finish but they will not be fixed.

One of the main problems here is that detekt doesn't know if a rule implement or not the autocorrect feature. It just know if the configuration actives it.

I would remove the autocorrect feature from detekt or, if someone wants to expend the time in it, look for a better way to implement it. Right now it is really fragile. I have some ideas about how to make this work better but I'm not willing to implement it myself.

BraisGabin:
Closing again. This wouldn't fix #1466 either. If you don't say "autoCorrect: false" at your config file the default configuration will decide. I'm not going to expend more time on this issue.

Issue #6947, opened 2024-02-06
False positive on `UnnecessaryLet`


atulgpt:
## Expected Behavior
```kotlin
@Test
fun `does not report let with which result in nested safe call with invoke`() {
    val findings = subject.compileAndLintWithContext(
        env,
        """
            class ClickContainer(val onClick: () -> Unit)

            fun test(callbacks: List<ClickContainer>) {
                callbacks
                    .firstOrNull()?.let {
                        it.onClick()
                    }
                callbacks
                    .firstOrNull()?.let {
                        it.onClick.invoke()
                    }
            }
        """.trimIndent()
    )
    assertThat(findings).isEmpty()
}
```
The above TC should pass

## Observed Behavior
The above TC is failing

## Context
In the above TC, in the below code section
```kotlin
callbacks
    .firstOrNull()?.let {
        it.onClick()
    }
```
we can remove `let` but the resulting code(`callbacks?.firstOrNull()?.onClick?.invoke()`) will not look good and result in other violations of cascading safe call

As a workaround, we can use `run` which doesn't trigger the rule. But the mentioned case should also pass

BraisGabin:
I'm not 100% sure of this change. What do @detekt/maintainers think?

I agree that `callbacks?.firstOrNull()?.onClick?.invoke()` is not ideal. But `let` adds complexity to read too. But it adds more or less the same complexity as `run` does. Probably even less because it doesn't shadow `this`. So I'm not sure what we should do in this case.

TWiStErRob:
I think `run` is not a workaround, but **a** solution here. Yes it's the same-ish complexity as `let`, but that's just syntax, there's also semantic meaning behind `let`/`also`/`run`/`apply`/`with`, that's why there's so many with different signatures. Sadly [the docs](https://kotlinlang.org/docs/scope-functions.html) focus very heavily (dare I say, only) on syntax.

In the above example you "select the first callback to invoke", i.e. `callbacks?.firstOrNull()?.run { onClick() }` a valid expression of that requirement, you get an object and then call operations on it.

<details><summary>Note: to fix the chain call detection (which is totally reasonable in this complexity)</summary>

Extract a val or fun that does a part of the expression, named!:
```kotlin
private val firstCallback: CB?
    get() = callbacks?.firstOrNull() // or callbacks.orEmpty().firstOrNull()

firstCallback?.onClick?.invoke()
```
or
```kotlin
internal fun CB.onClick() {
   this.onClick.invoke()
}
[...]

PR #6948, opened 2024-02-06
Allow last call to be invoke in case of safe chain

atulgpt:
This fixes https://github.com/detekt/detekt/issues/6947


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryLetSpec.kt
```diff
@@ -283,6 +283,24 @@ class UnnecessaryLetSpec(val env: KotlinCoreEnvironment) {
         assertThat(findings).isEmpty()
     }
 
+    @Test
+    fun `does not report when it has trailing chain calls`() {
```
atulgpt, line 287: Just added this TC as I couldn't fine any TC similar and I thought my code might break this. But this was passing earlier as well

3flex (changes requested):
I'll reply on the issue too, but the original example doesn't fail to compile with K2, so I don't think we should merge this.

PR #6952, opened 2024-02-07
Require a path to be passed to --generate-config

3flex:
I was attempting a refactor but got stuck because of `--generate-config` which motivated a rethink of how that parameter is used.

Currently when that flag is used it changes the meaning of `--config` - in normal usage, `--config` is a set of config paths that are read by detekt, but when `-gc` flag is passed, it means `--config` should contain a single file only which will be written or overwritten.

This PR changes the interface so `-gc` accepts a path directly. This makes detekt CLI easier to use. To generate a config now users can do `detekt -gc <path>` instead of `detekt -gc true -c <path>`.


BraisGabin:
I like the idea :) But you need to check the PR, CI isn't happy.

BraisGabin (approved):

PR #6964, opened 2024-02-11
Use standard Gradle properties to manage built-in report outputs

### LINKED ISSUE #4192: XML report merging doesn't work without renaming reports to `detekt.xml`
## Expected Behavior
[XML Report merging](https://detekt.github.io/detekt/reporting.html#merging-reports) should work out of the box when following the examples in the documentation. 

## Observed Behavior
Report merging fails silently for android modules ( and maybe kotlin libraries? )  because it expects all xml reports to be named `detekt.xml`, while by default they are named   `debug.xml`, `main.xml`, etc.

The name seems to be hardcoded in:  `io.gitlab.arturbosch.detekt.Detekt#getTargetFileProvider`

## Steps to Reproduce

1)  Unzip the attached [Repro Project](https://github.com/detekt/detekt/files/7365640/detekt-report-merge-repro-project.zip)
2)  run `./repro.sh`
3) Notice that there is nothing under `$rootProject/build/reports/detekt`

## Your Environment
* Version of Gradle used (if applicable): 7.2
* Operating System and version: MacOS 11.5.1

### LINKED ISSUE #4856: Detekt task outputs are not following Gradle convention
## Expected Behavior
Using task output property works as expected (automatic dependency on task).

## Steps to Reproduce
/build.gradle.kts:
```kotlin
tasks.register<io.gitlab.arturbosch.detekt.report.ReportMergeTask>("detektReportMergeSarif") {
	output.set(rootProject.buildDir.resolve("reports/detekt/merge.sarif"))
}
```
/module/build.gradle.kts:
```kotlin
val detektReportMergeSarif = rootProject.tasks.named<ReportMergeTask>("detektReportMergeSarif")
tasks.withType<Detekt> {
	detektReportMergeSarif.configure { input.from(this@withType.sarifReportFile) }
}
```
Then execute `gradlew :module:detekt :detektReportMergeSarif`.

## Context
Trying to merge files, and use Gradle best practices.

### Workaround
```
detektReportMergeSarif.configure {
	input.from(this@withType.sarifReportFile)

	mustRunAfter(this@withType)
        // or
	dependsOn(this@withType)
}
```

### Possible fix
Stop creating the property manually without a producer:

https://github.com/detekt/detekt/blob/950554c173cbe09c201107194ca7c2a6f2766549/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt#L280

instead let Gradle create it for us:

```diff
-open class Detekt
+abstract class Detekt

+@get:Optional
+@get:OutputFile
-val sarifReportFile: Provider<RegularFile>
+absract val sarifReportFile: Provider<RegularFile>
-        @OutputFile
-        @Optional
-        get() = getTargetFileProvider(reports.sarif)
```

and then lazily configure it without resolving the location:
```
sarifReportFile.set(report.outputLocation.map/flatMap { })
```

## Your Environment

* Version of Gradle used (if applicable): 7.4.2
* Gradle scan link (add `--scan` option when running the gradle task): https://scans.gradle.com/s/huhzubjvpmsuu
* Operating System and version: N/A
* Link to your project (if it's a public repository): https://github.com/TWiStErRob/net.twisterrob.sun/tree/e2646e3f97de5e9f30abec53a44640c47e4cb9f8

### LINKED ISSUE #6963: reportsDir property omission in Detekt task for android and jvm configurations
## Expected Behavior
- The `reportDir` property in the Gradle task [Detekt](https://github.com/detekt/detekt/blob/main/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt#L135) is equivalent to that set in the [DetektExtension](https://github.com/detekt/detekt/blob/main/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektExtension.kt#L21).
- The `reportDir` property in the Gradle task [Detekt](https://github.com/detekt/detekt/blob/main/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt#L135) is annotated with the `@OutputDirectory` annotation.

## Observed Behavior
- The `reportDir` property from DetektExtension is only used in DetektPlain configuration and is omitted in DetektAndroid and DetektJvm. Instead it is directly used to [set the outputLocation](https://github.com/detekt/detekt/blob/main/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektMultiplatform.kt#L143) for DetektReport. 
- Currently, the `reportDir` property in the Gradle task is marked with the `@Internal` annotation, preventing its use for task linking through inputs/outputs.

## Steps to Reproduce
```kotlin
//build.gradle.kts

abstract class CollectDetektReportsTask : DefaultTask() {

    @get:InputFiles
    abstract val reports: ConfigurableFileCollection
}

val collectReportTask = tasks.register("collectDetektReports", CollectDetektReportsTask::class)
subprojects {
    plugins.withType(DetektPlugin::class.java) {
        val detektTasks = tasks.withType(Detekt::class.java)
        collectReportTask.configure {
            reports.from(detektTasks.map { it.reportsDir })
        }
    }
}

//app/build.gradle.kts

plugins {
    ....
    id("io.gitlab.arturbosch.detekt")
}

detekt {
    reportsDir = project.layout.buildDirectory.dir("detektReport").get().asFile
}
```

Run:
```bash
./gradlew :collectDetektReports
```

Output:
```txt
Cannot query the value of task ':app:detektDebugUnitTest' property 'reportsDir' because it has no value available.
```

## Context
I want to collect all reports from Detekt tasks using the root collect task as [ReportMergeTask](https://detekt.dev/docs/introduction/reporting/#kotlin-dsl) does, but I need to use a custom report format (csv) and use task configuration avoidance and link tasks to each other by inputs/outputs as [Gradle recommends](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html). 
Currently this is not possible for Detekt, as the task outputs are not directly linked to the task itself as described at https://github.com/detekt/detekt/issues/4856. 
I noticed that the task also has a [reportDir](https://github.com/detekt/detekt/blob/main/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt#L135) property, which is an output for the task and could be a solution to the problem, but it is marked as `@Internal`. I also noticed that the task property has no convention value from `DetektExtension` because it is omitted in the DetektAndroid and DetektKotlin classes, but is only set for DetektPlain . 

## Solution

To solve the problem with the reportDir property in Detekt tasks, I suggest setting a convention for the reportDir property within the task based on the value from the DetektExtension, as was done for [DetektPlain](https://github.com/detekt/detekt/blob/main/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/DetektPlain.kt#L25).

## Inquiry

Is there a specific reason for marking the `reportDir` as `@Internal`? Could it be annotated with `@OutputDirectory` to resolve issue and improve Detekt's integration with Gradle's task configuration avoidance and output linking?

## Your Environment

* Version of Gradle used (if applicable): 8.6

3flex:
* Fixes #4192 
* Fixes #4856 
* Fixes #6963 

There's one issue I can see which is there's a problem with using `VerificationException` and the Worker API. We don't enable Worker API by default but this could be considered a blocker for doing that. See https://github.com/gradle/gradle/issues/28034.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -129,7 +129,7 @@ abstract class Detekt @Inject constructor(
     abstract val basePath: Property<String>
 
     @get:Internal
-    var reports: DetektReports = objects.newInstance(DetektReports::class.java)
+    abstract val reports: DetektReports
 
     @get:Internal
     abstract val reportsDir: DirectoryProperty
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -130,9 +130,6 @@ abstract class Detekt @Inject constructor(
      */
     open val reports: DetektReports = objects.newInstance(DetektReports::class.java)
 
-    @get:Internal
-    abstract val reportsDir: DirectoryProperty
-
     internal val customReportFiles: ConfigurableFileCollection
         @OutputFiles
         @Optional
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -44,6 +44,7 @@ import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFile
 import org.gradle.api.tasks.InputFiles
 import org.gradle.api.tasks.Internal
+import org.gradle.api.tasks.Nested
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.OutputFiles
@@ -128,8 +129,12 @@ abstract class Detekt @Inject constructor(
     @get:Optional
     abstract val basePath: Property<String>
 
-    @get:Internal
-    abstract val reports: DetektReports
+    @get:Nested
+    /*
+    Property must be open (as do the @Nested properties in DetektReports), see
+    https://github.com/gradle/gradle/pull/12601 and https://github.com/gradle/gradle/issues/6619
+     */
+    open val reports: DetektReports = objects.newInstance(DetektReports::class.java)
 
     @get:Internal
     abstract val reportsDir: DirectoryProperty
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -1,6 +1,5 @@
 package io.gitlab.arturbosch.detekt
 
-import io.gitlab.arturbosch.detekt.extensions.DetektReport
 import io.gitlab.arturbosch.detekt.extensions.DetektReportType
 import io.gitlab.arturbosch.detekt.extensions.DetektReports
 import io.gitlab.arturbosch.detekt.extensions.FailOnSeverity
@@ -26,16 +25,12 @@ import io.gitlab.arturbosch.detekt.invoke.LanguageVersionArgument
 import io.gitlab.arturbosch.detekt.invoke.ParallelArgument
 import org.gradle.api.Action
 import org.gradle.api.file.ConfigurableFileCollection
-import org.gradle.api.file.Directory
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.FileTree
-import org.gradle.api.file.RegularFile
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Property
-import org.gradle.api.provider.Provider
 import org.gradle.api.provider.ProviderFactory
-import org.gradle.api.reporting.ReportingExtension
 import org.gradle.api.tasks.CacheableTask
 import org.gradle.api.tasks.Classpath
 import org.gradle.api.tasks.Console
@@ -46,7 +41,6 @@ import org.gradle.api.tasks.InputFiles
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Nested
 import org.gradle.api.tasks.Optional
-import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.OutputFiles
 import org.gradle.api.tasks.PathSensitive
 import org.gradle.api.tasks.PathSensitivity
@@ -139,39 +133,10 @@ abstract class Detekt @Inject constructor(
     @get:Internal
     abstract val reportsDir: DirectoryProperty
 
-    val xmlReportFile: Provider<RegularFile>
-        @OutputFile
-        @Optional
-        get() = getTargetFileProvider(reports.xml)
-
-    val htmlReportFile: Provider<RegularFile>
-        @OutputFile
-        @Optional
-        get() = getTargetFileProvider(reports.html)
-
-    val txtReportFile: Provider<RegularFile>
-        @OutputFile
-        @Optional
-        get() = getTargetFileProvider(reports.txt)
-
-    val sarifReportFile: Provider<RegularFile>
-        @OutputFile
-        @Optional
-        get() = getTargetFileProvider(reports.sarif)
-
-    val mdReportFile: Provider<RegularFile>
-        @OutputFile
-        @Optional
-        get() = getTargetFileProvider(reports.md)
-
     internal val customReportFiles: ConfigurableFileCollection
         @OutputFiles
         @Optional
         get() = objects.fileCollection().from(reports.custom.mapNotNull { it.outputLocation.asFile.orNull })
-
-    private val defaultReportsDir: Directory = project.layout.buildDirectory.get()
-        .dir(ReportingExtension.DEFAULT_REPORTS_DIR_NAME)
-        .dir("detekt")
 
     private val isDryRun = project.providers.gradleProperty(DRY_RUN_PROPERTY)
 
@@ -189,11 +154,11 @@ abstract class Detekt @Inject constructor(
             JdkHomeArgument(jdkHome),
             ConfigArgument(config),
             BaselineArgument(baseline.orNull),
-            DefaultReportArgument(DetektReportType.XML, xmlReportFile.orNull),
-            DefaultReportArgument(DetektReportType.HTML, htmlReportFile.orNull),
-            DefaultReportArgument(DetektReportType.TXT, txtReportFile.orNull),
-            DefaultReportArgument(DetektReportType.SARIF, sarifReportFile.orNull),
-            DefaultReportArgument(DetektReportType.MD, mdReportFile.orNull),
+            DefaultReportArgument(reports.xml),
+            DefaultReportArgument(reports.html),
+            DefaultReportArgument(reports.txt),
+            DefaultReportArgument(reports.sarif),
+            DefaultReportArgument(reports.md),
             DebugArgument(debug.getOrElse(false)),
             ParallelArgument(parallel.getOrElse(false)),
             BuildUponDefaultConfigArgument(buildUponDefaultConfig.getOrElse(false)),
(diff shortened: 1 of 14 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -89,10 +90,30 @@ class DetektPlugin : Plugin<Project> {
...
+                report.required.convention(DEFAULT_REPORT_ENABLED_VALUE)
+                report.outputLocation.convention(extension.reportsDir.file("${DetektReport.DEFAULT_FILENAME}.sarif"))
+            }
+            it.reports.txt { report ->
+                report.required.convention(DEFAULT_REPORT_ENABLED_VALUE)
+                report.outputLocation.convention(extension.reportsDir.file("${DetektReport.DEFAULT_FILENAME}.txt"))
+            }
+            it.reports.xml { report ->
+                report.required.convention(DEFAULT_REPORT_ENABLED_VALUE)
+                report.outputLocation.convention(extension.reportsDir.file("${DetektReport.DEFAULT_FILENAME}.xml"))
```
BraisGabin, line 115: How do we rename these files for the main/test tasks?

BraisGabin (approved):

BraisGabin:
> There's one issue I can see which is there's a problem with using `VerificationException` and the Worker API. We don't enable Worker API by default but this could be considered a blocker for doing that. See https://github.com/gradle/gradle/issues/28034.


To me this shouldn't block us. It is an issue from gradle. So, once we enable it by default, we should point out in the documentation that if someone wants to use the `--continue` then they should disable the worker api. Also linking to the issue to get reactions.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
3flex: This uses `convention()`, so it just sets the defaults for Detekt tasks. When individual tasks are registered by the plugin, or someone wants to set a specific path in their build script, they use `set()` instead.

BraisGabin: 🤔 Should we then even have default values for the paths? These values are only used by the plain detekt task, they are not really "default" values, right?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -128,46 +123,21 @@ abstract class Detekt @Inject constructor(
...
-    var reports: DetektReports = objects.newInstance(DetektReports::class.java)
+    @get:Nested
+    /*
+    Property must be open (as do the @Nested properties in DetektReports), see
+    https://github.com/gradle/gradle/pull/12601 and https://github.com/gradle/gradle/issues/6619
+     */
+    open val reports: DetektReports = objects.newInstance(DetektReports::class.java)
 
     @get:Internal
     abstract val reportsDir: DirectoryProperty
```
rwqwr, line 134: I noticed that this property is no longer used. Do you have any plans for its use?

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
3flex: They're used when tasks are setup by users too. Also it gives the opportunity to refactor and set defaults for plugin tasks here instead of in the task registration (in almost every case the report name is the task name with "detekt" stripped from the front - and I have plans for the plain task which would remove that as the only exception to that convention).

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
3flex: Oh thanks, I missed that. It should be removed here.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
BraisGabin: But is it good to have a default value like this? If someone creates a default task that would make it collision with the others.

Anyway, this is not directly related with this PR. And I trust your opinion in anything gradle related.

3flex: Oh I see now, yes you're absolutely right that it will cause collisions.

I'll update so it uses the task name as the default filename which will avoid almost all collisions. I'll make further updates in a separate PR.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -89,10 +89,30 @@ class DetektPlugin : Plugin<Project> {
...
     }
 
-    private fun setTaskDefaults(project: Project) {
+    private fun setTaskDefaults(project: Project, extension: DetektExtension) {
         project.tasks.withType(Detekt::class.java).configureEach {
             it.detektClasspath.setFrom(project.configurations.getAt(CONFIGURATION_DETEKT))
             it.pluginClasspath.setFrom(project.configurations.getAt(CONFIGURATION_DETEKT_PLUGINS))
+            it.reports.html { report ->
+                report.required.convention(DEFAULT_REPORT_ENABLED_VALUE)
+                report.outputLocation.convention(extension.reportsDir.file("${it.name}.html"))
```
BraisGabin, line 98: nit: Probably rename `it` to `task` so it's easier to know which "name" are you getting. There are a lot of moving parts here.

BraisGabin (approved):

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektReports.kt
```diff
@@ -7,22 +7,30 @@ import io.gitlab.arturbosch.detekt.extensions.DetektReportType.TXT
...
+    @get:Nested
+    open val sarif: DetektReport = objects.newInstance(DetektReport::class.java, SARIF)
 
-    val md: DetektReport = objects.newInstance(DetektReport::class.java, MD)
+    @get:Nested
+    open val md: DetektReport = objects.newInstance(DetektReport::class.java, MD)
 
-    val custom = mutableListOf<CustomDetektReport>()
+    @get:Nested
+    open val custom = mutableListOf<CustomDetektReport>()
```
rwqwr, line 33: I checked locally and found that when using custom reports, the parent task information is lost and an exception is thrown:
```txt
Property 'outputLocation' is declared as an output property of CustomDetektReport(reportId=csv, outputLocation=property 'outputLocation') (type CustomDetektReport) but does not have a task associated with it.
```
I think the problem at the line [48](https://github.com/detekt/detekt/pull/6964/files#diff-a4ff11125569eadcdf856d472ffde8682d53121a1f71b92445930adbe62eb60fR48). 
When creating an object via `objects.newInstance`, there is no binding despite the `@Nested` option for `custom`.

You can use the following configuration for testing: 
```kotlin
plugins {
    id("io.gitlab.arturbosch.detekt")
}

allprojects {
    repositories {
        mavenLocal()
        mavenCentral()
        google()
    }
}

val reportMerge by tasks.registering(io.gitlab.arturbosch.detekt.report.ReportMergeTask::class) {
    output.set(project.layout.buildDirectory.file("reports/detekt/merge.xml"))
    outputs.cacheIf { false }
    outputs.upToDateWhen { false }
}

subprojects {
    apply(plugin = "io.gitlab.arturbosch.detekt")

    plugins.withType<io.gitlab.arturbosch.detekt.DetektPlugin> {
        val detektTasks = tasks.withType(io.gitlab.arturbosch.detekt.Detekt::class.java)
        detektTasks.configureEach {
            val task = this
            reports {
                custom {
                    reportId = "csv"
                    outputLocation.set(layout.buildDirectory.map { it.file(task.name + "report.csv") })
                }
            }
        }
        reportMerge.configure {
            val reports = detektTasks.map { it.reports.custom.first { it.reportId == "csv"}.outputLocation }
            input.from(reports)
        }
    }
}
```

3flex: Thanks for checking, I'll take a look - I'll have to add tests too since there must be missing coverage.

3flex: I've looked at a few options that kept the existing DSL for custom reports but none seemed to fix it. There's an additional change I could make but that would affect the DSL too much for this PR.

The good news though is that this PR doesn't seem to have introduced that issue. I created a similar test which reproduced the problem and when running that test on `main` I see the same problem.

Can you please confirm? I can then merge this PR as is and create a separate PR to address that problem.

rwqwr: Yes, you are right, this problem exists on the main and I was hoping this PR could fix it :) 
I suppose it's not a blocker and can be done by a separate PR, thank you!

rwqwr (approved):

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/extensions/DetektReports.kt
3flex: Issue opened: https://github.com/detekt/detekt/issues/6980

Issue #5947, opened 2023-03-27
Feature request: create type resolution and non-type-resolution variants of Detekt tasks to make configuration easier


ZacSweers:
## Expected Behavior
See below

## Current Behavior
See below

## Context

Currently, the only way to know if a `Detekt` task is a type resolution task or not is to know the names are `detekt` or `detektBaseline`, vs `detektRelease`/`detektMain`/`detektMainBaseline`/etc.

It would be ideal if there were subtasks with a `UsesTypeResolution` marker supertype to disambiguate these when configuring tasks. Another alternative would be a constructor parameter/property used when registering the task that could be checked via `task.usesTypeResolution` at configuration-time. This would be for both `Detekt` and `DetektCreateBaselineTask`

cortinico:
> Another alternative would be a constructor parameter/property used when registering the task that could be checked via `task.usesTypeResolution` at configuration-time

We briefly discussed this as part of the 2.x roadmap and we'll most likely opt for a property to toggle Type Resolution (and yes, we all agree that the current task graph is confusing from the Type Resolution point of view).

PR #6973, opened 2024-02-15
Register Detekt task for each source set

### LINKED ISSUE #4036: also create detektMetadataTest for MPP to analyse commonTest
## Expected Behavior
code issues in commonTest are also discovered by detekt

## Observed Behavior
Looks like commonTest is not analysed (only commonMain)

## Steps to Reproduce
1. git clone https://github.com/robstoll/kbox
2. checkout new-MPP
3. there are specs in commonTest with a fake KDoc without full-stop, should be reported (as it is for jsMain and jvmMain)

## Your Environment

* Version of detekt used: 1.18.0
* Version of Gradle used (if applicable): 7.1.1
* Link to your project (if it's a public repository): https://github.com/robstoll/kbox

### LINKED ISSUE #6634: Register tasks for shared source sets using default hierarchy templates
[Kotlin 1.9.20 stabilized the default hierarchy templates to make using shared source sets easier](https://kotlinlang.org/docs/whatsnew1920.html#template-for-configuring-multiplatform-projects). It would be nice if detekt was able to hook into this somehow so that detekt tasks could be registered on the shared source sets (apple, posix, ios, etc...).

3flex:
Registers a Detekt task for every source set. This task does not support type resolution by design.

* Fixes #6634
* Fixes #4036

This is also a step towards #5947

### DIFF
--- a/detekt-gradle-plugin/src/functionalTest/kotlin/dev/detekt/gradle/plugin/DetektBasePluginSpec.kt
+++ b/detekt-gradle-plugin/src/functionalTest/kotlin/dev/detekt/gradle/plugin/DetektBasePluginSpec.kt
@@ -1,0 +1,190 @@
+package dev.detekt.gradle.plugin
+
+import io.gitlab.arturbosch.detekt.testkit.DslGradleRunner
+import io.gitlab.arturbosch.detekt.testkit.ProjectLayout
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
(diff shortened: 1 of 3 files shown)


BraisGabin:
This way we could make `detekt` just an alias and make it depend on all these tasks. Is that the idea?

BraisGabin (approved):

3flex:
Still thinking through what's best for that task - I'm going to leave it as is until a few more changes have been made.

BraisGabin:
Sure, I'm not even sure if `detekt` is even a good name for the task.

PR #6977, opened 2024-02-16
Add MissingSuperCall rule

### LINKED ISSUE #6960: Rule to verify calling super
When `@javax.annotation.OverridingMethodsMustInvokeSuper` or `@androidx.annotation.CallSuper` is present on a method which is overridden, report that a call must be made. Very likely `@RequiresTypeResolution`.

## Context
Android Lint has a check for this, but when writing non-Android code, that is not usable.
Most projects likely have `com.google.code.findbugs:jsr305` which includes this annotation, but at present there's no Kotlin detector I could find other than Android Lint.

This gap could be very useful for rule authors and detekt contributors as well, since we continuously forget calling `super.visit()`.

## References
 * https://errorprone.info/bugpattern/MissingSuperCall
 * https://googlesamples.github.io/android-custom-lint-rules/checks/MissingSuperCall.md.html
 * oddly enough spotbugs (successor of maintainer (findbugs) of jsr305.jar) does not support it

t-kameyama:
Fixes #6960

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -483,6 +483,11 @@ potential-bugs:
   MissingPackageDeclaration:
     active: false
     excludes: ['**/*.kts']
+  MissingSuperCall:
+    active: false
+    mustInvokeSuperAnnotations:
+      - 'androidx.annotation.CallSuper'
+      - 'javax.annotation.OverridingMethodsMustInvokeSuper'
   MissingUseCall:
     active: false
   NullCheckOnMutableProperty:
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCall.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCall.kt
@@ -65,13 +65,15 @@ class MissingSuperCall(config: Config) : Rule(
 
         if (!function.hasModifier(KtTokens.OVERRIDE_KEYWORD)) return
         val functionDescriptor = bindingContext[DECLARATION_TO_DESCRIPTOR, function] as? CallableDescriptor ?: return
-
-        val superFunctionDescriptor = functionDescriptor.overriddenDescriptors.firstOrNull() ?: return
-        if (superFunctionDescriptor.annotations.none { it.fqName in mustInvokeSuperAnnotations }) return
-
+        val superFunctionDescriptor = functionDescriptor.superFunctionWithAnnotation() ?: return
         if (function.hasSuperCall(superFunctionDescriptor)) return
 
         report(CodeSmell(Entity.from(function), "Overriding method is missing a call to overridden super method."))
+    }
+
+    private fun CallableDescriptor.superFunctionWithAnnotation(): CallableDescriptor? {
+        return overriddenDescriptors.firstOrNull { d -> d.annotations.any { it.fqName in mustInvokeSuperAnnotations } }
+            ?: overriddenDescriptors.firstNotNullOfOrNull { it.superFunctionWithAnnotation() }
     }
 
     private fun KtNamedFunction.hasSuperCall(superFunctionDescriptor: CallableDescriptor): Boolean {
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCall.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCall.kt
@@ -33,7 +33,7 @@ import org.jetbrains.kotlin.resolve.calls.util.getResolvedCall
  * </noncompliant>
  *
  * <compliant>
- *  open class ParentClass {
+ * open class ParentClass {
  *     @CallSuper
  *     open fun someMethod(arg: Int) {
  *     }
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCall.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCall.kt
@@ -1,0 +1,82 @@
+package io.gitlab.arturbosch.detekt.rules.bugs
+
+import io.gitlab.arturbosch.detekt.api.CodeSmell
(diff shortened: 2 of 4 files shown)


detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCallSpec.kt
```diff
@@ -0,0 +1,108 @@
...
+            class MyClass : ParentClass() {
+                override fun someMethod(arg: Int) {
+                    super.someMethod(arg)
+                    println()
+                }
+            }
+        """.trimIndent()
+        val actual = subject.compileAndLintWithContext(env, code)
+        assertThat(actual).isEmpty()
+    }
```
BraisGabin, line 107: I'm missing two tests:
- Check what happen in case of `MyClass` is extended. The annotation should be honored.
- Check what happen if the function calls a super method but it is not the same one:
```kotlin
fun someMethod() {
  super.anotherMerhod()
}
```

detekt-core/src/main/resources/default-detekt-config.yml
```diff
@@ -483,6 +483,11 @@ potential-bugs:
   MissingPackageDeclaration:
     active: false
     excludes: ['**/*.kts']
+  MissingSuperCall:
+    active: false
+    mustInvokeSuperAnnotations:
+      - 'androidx.annotation.CallSuper'
+      - 'javax.annotation.OverridingMethodsMustInvokeSuper'
```
BraisGabin, line 490: What do we think about these two? One is clearly Android but the other is just a proposal from Google that never ended as a standard. So we could think that both are "third party". But if we remove both this rule have little sense. Only few people Wil configure it properly.

I would vote to keep both but the `androidx` annotation for sure break some of our rules. What do you think @detekt/maintainers?

TWiStErRob: Agree, that without a default this rule is not useful. The whole idea behind it is that it "just" tells you if something is wrong.

I think there should be an exception to the rule of third party where there's a limited commonly used set. The question is how to define "commonly used" or "well known" :)

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCallSpec.kt
TWiStErRob: One more class of tests:
```
interface A {
    @InvokeSuper
    fun f() {}
}
interface B {
    fun f() {}
}

class X : A, B {
    override fun f() {
        super<A>.f()
    }
}
```

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCall.kt
```diff
@@ -0,0 +1,84 @@
...
+ * }
+ * class MyClass : ParentClass() {
+ *     override fun someMethod(arg: Int) {
+ *         doSomething()
+ *     }
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ *  open class ParentClass {
```
BraisGabin, line 36:
```suggestion
 * open class ParentClass {
```

BraisGabin (approved):
LGTM

TWiStErRob (commented):
Few more tests, otherwise lgtm

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCall.kt
```diff
@@ -0,0 +1,84 @@
...
+        if (!function.hasModifier(KtTokens.OVERRIDE_KEYWORD)) return
+        val functionDescriptor = bindingContext[DECLARATION_TO_DESCRIPTOR, function] as? CallableDescriptor ?: return
+        val superFunctionDescriptor = functionDescriptor.superFunctionWithAnnotation() ?: return
+        if (function.hasSuperCall(superFunctionDescriptor)) return
+
+        report(CodeSmell(Entity.from(function), "Overriding method is missing a call to overridden super method."))
+    }
+
+    private fun CallableDescriptor.superFunctionWithAnnotation(): CallableDescriptor? {
+        return overriddenDescriptors.firstOrNull { d -> d.annotations.any { it.fqName in mustInvokeSuperAnnotations } }
```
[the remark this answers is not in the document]
t-kameyama: https://github.com/detekt/detekt/blob/30a1c7d6c392bdcae7d6596f0ff67f2042b2dbb4/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCallSpec.kt#L88-L114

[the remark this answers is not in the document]
t-kameyama: https://github.com/detekt/detekt/blob/30a1c7d6c392bdcae7d6596f0ff67f2042b2dbb4/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCallSpec.kt#L59-L86

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCall.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    @Configuration("Annotations to require that overriding methods invoke the super method")
+    private val mustInvokeSuperAnnotations: List<FqName> by config(
+        listOf(
+            "androidx.annotation.CallSuper",
+            "javax.annotation.OverridingMethodsMustInvokeSuper",
+        )
+    ) { it.map(::FqName) }
+
+    override fun visitNamedFunction(function: KtNamedFunction) {
```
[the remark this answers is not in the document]
t-kameyama: https://github.com/detekt/detekt/blob/30a1c7d6c392bdcae7d6596f0ff67f2042b2dbb4/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MissingSuperCallSpec.kt#L162-L167

BraisGabin:
Thanks as always @t-kameyama for all your awesome contributions :)

PR #6983, opened 2024-02-21
fix task 'config' property getting overwritten

mgroth0:
`setFrom` overwrites whatever is there before. Changed to `from` which only adds, does not overwrite.

In practice, `setFrom` was causing an issue for me where trying to set the config file for a specific Detekt task was not working becuase it was being overwritten by what I put in the plugin extension `config`. The overwrite only occured for some Detekt tasks and not others, adding to the confusion. I suspect this is because the order of when gradle executed certain blocks of code mattered. My expectation is all configured files to be used, none should replace any other and especially not in a silent, confusing, unpredictable way.

With this change, order of when gradle executes different blocks of code no longer matters and all config files are used reliably both from the plugin extension and from task properties.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/SharedTasks.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/internal/SharedTasks.kt
@@ -37,7 +37,7 @@ internal fun Project.registerDetektTask(
         it.autoCorrect.convention(extension.autoCorrect)
         it.ignoreFailures.convention(extension.ignoreFailures)
         it.failOnSeverity.convention(extension.failOnSeverity)
-        it.config.setFrom(provider { extension.config })
+        it.config.from(provider { extension.config })
         it.basePath.convention(extension.basePath.map { basePath -> basePath.asFile.absolutePath })
         it.allRules.convention(extension.allRules)
         configuration(it)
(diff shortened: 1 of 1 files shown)


3flex:
How are you configuring the task to override the config? Can you provide a code snippet?

The proposed PR is half a fix that leaves an issue - instead of replacing the task's config, it will add the extension's path(s) to the end of the task's config paths.

3flex:
> Detekt tasks should have a way to input configuration without having to use any files at all

This can partially be achieved by using [`TextResourceFactory`](https://docs.gradle.org/current/dsl/org.gradle.api.resources.TextResourceFactory.html) and https://docs.gradle.org/current/dsl/org.gradle.api.resources.TextResource.html#org.gradle.api.resources.TextResource:asFile().

The config will still be passed to detekt as a file but it will simplify your task config a lot.

> config.from(generatedConfigFile)

Why aren't you using `setFrom` to override the default config in the task? `set` adds a new file to the list but if you want to override config you should use `setFrom`.

> I am assuming this is because the order in which different configuration blocks in gradle execute is unpredictable.

I don't think this is correct, it might be unexpected in some cases, but it should be predictable and consistent, and if it wasn't, that sounds like a Gradle issue rather than something we can fix properly in the detekt plugin.

> this PR is just a band-aid and doesn't resolve some bigger issues. If you still are not sure about merging it, maybe we could at least leave the PR open and come back to it another time after working on some of these other issues?

Yes please. Opening individual new issues for each of the issues you're seeing would be best, with reproducers. I'm going to close this PR for now though since recreating it is trivial if it turns out this is the right approach (or it can be reopened), but I'd prefer to leave the current functionality as is and introduce `convention` to set the default on the task once Gradle 8.8 is out which should resolve this too.

3flex:
> introduce convention to set the default on the task once Gradle 8.8 is out

See #7356

PR #6988, opened 2024-02-23
Make Android tests more specific

3flex:


BraisGabin (approved):

BraisGabin:
I was wondering. Would it have sense to move some functional tests to unit tests? A lot of our tests (mainly the android ones) just check that some tasks where executed and that should be easily checked by unit tests and also way faster. Also we should be able to check the inputs of the task too. We will still need the functional ones to ensure that the task does what is intended but probably we could reduce those. Something like this: https://github.com/SchwarzIT/dagger-lightsaber/blob/main/gradle-plugin/src/test/kotlin/schwarz/it/lightsaber/gradle/LightsaberPluginTest.kt

I can move this to an issue if you think it is worth the discussion.

PR #6989, opened 2024-02-23
Add ignoreOperator option in TooManyFunctions rule

xgouchet:
This PR adds an option to ignore operators from the `TooManyFunctions` rule. 

When writing a mathematical class, it's common to create Kotlin `operator fun` to allow using Operator overloading. Some operators can need a lot of implementation to deal with all numeric types accordingly (Double, Long, …), which can make the number of functions grow. 

Adding the `ignoreOperator` (which behave very similarly to `ignorePrivate` or `ignoreOverridden`) allow one to write math classes without raising warnings from Detekt.


BraisGabin:
The use case that you are describing is a good candidate to just add a suppress annotation. In general, I think that an operator shouldn't count as a function. And even if in some classes you need a lot of operstors I think that a suppress is a better solution because in other classes you would want to count the operators.

xgouchet:
Yes, I totally get that a `@Suppress` annotation could solve it. The thing is that I have a project with lots of those operator heavy classes, but I still want to have a check on non operator functions, which I can't have if I completely suppress the rule.

3flex:
I agree with Brais, using `@Suppress` is more suitable. This option seems very specific to certain types of projects and isn't generally useful.

BraisGabin:
I'm closing for the reasons explained above. You can use `@Suppress` or you can create your own version of this rule as a third party rule and expand it to make it suits your need.

PR #6995, opened 2024-02-27
UnusedPrivateClass: don't report if private classes are used for type conversion

### LINKED ISSUE #6982: UnusedPrivateClass Rule does not detect casts as usage
All described behavior below applies to the default detekt configuration (the one that is generated by the `detektGenerateConfig` gradle task)

## Expected Behavior
`UnusedPrivateClass` should only trigger if a private class is not used in any way that is relevant to the code

## Observed Behavior
If a private class is only used for a cast (using `as`), `UnusedPrivateClass` still triggers. If an explicit type annotation is used instead of the cast, `UnusedPrivateClass` does not trigger, even though the two cases are functionally the same.

## Steps to Reproduce
The following code will trigger `UnusedPrivateClass` at `1:1`:
```kotlin
private class Foo(val bar: String)

private val fooCasted = null as Foo?
// private val fooExplicitType: Foo? = null

fun use() {
    println(fooCasted)
    // println(fooExplicitType)
}

```
Commenting in `fooExplicitType` will remove the rule violation.

## Context
The case with casting `null` to a nullable type is crafted to demonstrate the most simple reproduction example. For me, the issue arised because I declared a `private external class` in Kotlin/JS and tried to cast the result of a `js("")` call to this class:

```kotlin
private external class Crypto {
    val subtle: SubtleCrypto
}
private val crypto = js("window ? (window.crypto ? window.crypto : window.msCrypto) : self.crypto") as Crypto
```
When an explicit type is used instead of the cast:
```kotlin
private val crypto: Crypto = js("window ? (window.crypto ? window.crypto : window.msCrypto) : self.crypto")
```
the detekt rule violation goes away but the IDE warns me "Implicit (unsafe) cast from dynamic to Crypto" and suggests adding an explicit cast (which then removes the need to explicitly annotate the type of `crypto`).

## Your Environment

* Version of Kotlin used: Kotlin/JVM 1.9.22
* Version of Gradle used: 8.6
* Gradle scan link: https://scans.gradle.com/s/xewkclerzfekg
* Operating System and version: Windows 11

t-kameyama:
Fixes #6982

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateClass.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateClass.kt
@@ -8,6 +8,7 @@ import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Rule
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.psi.KtAnnotationEntry
+import org.jetbrains.kotlin.psi.KtBinaryExpressionWithTypeRHS
 import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtCallableReferenceExpression
 import org.jetbrains.kotlin.psi.KtClass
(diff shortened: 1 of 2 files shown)


BraisGabin (approved):

BraisGabin:
Thanks!

PR #7006, opened 2024-03-01
Add new rules from ktlint 1.2

3flex:
New rules:
* BackingPropertyNaming
* BlankLineBetweenWhenConditions
* Kdoc
* SpacingAroundSquareBrackets

### DIFF
--- a/detekt-formatting/src/main/resources/config/config.yml
+++ b/detekt-formatting/src/main/resources/config/config.yml
@@ -23,6 +23,10 @@ formatting:
   BlankLineBeforeDeclaration:
     active: true
     autoCorrect: true
+  BlankLineBetweenWhenConditions:
+    active: false
+    autoCorrect: true
+    lineBreakAfterWhenEntries: true
   BlockCommentInitialStarAlignment:
     active: true
     autoCorrect: true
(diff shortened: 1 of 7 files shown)


detekt-formatting/src/main/resources/config/config.yml
```diff
@@ -121,6 +126,8 @@ formatting:
     active: true
     autoCorrect: true
     indentSize: 4
+  KDoc:
```
3flex, line 129: I think this is the right name, since KDoc is what it's called, not Kdoc.

ktlint uses `kdoc` though and the next rule in the list `KdocWrapping` also uses lowercase `d` so maybe we should keep the same naming convention?

schalkms: Yes, we have previously matched the rule names of ktlint. Otherwise, it becomes confusing.

schalkms (approved):
Thanks! 👍

Additional question: do we backport new ktlint rules to the v1.x branch?

cortinico:
> Additional question: do we backport new ktlint rules to the v1.x branch?

I'd say not as 1.x is stuck at KtLint 0.50.0

schalkms:
Agreed, makes no sense. Value gained is too low for making such a version bump.
> I'd say not as 1.x is stuck at KtLint 0.50.0

detekt-formatting/src/main/resources/config/config.yml
3flex: So just to confirm, change to Kdoc with lowercase d?

schalkms: > So just to confirm, change to Kdoc with lowercase d?

👍

PR #7007, opened 2024-03-02
Cli provides list of input files

BraisGabin:
This is just the first step. Now the `cli` provides a list of files instead of directories and rules for include/exclude files.

After this change we can do things like #7009 where we can remove a lot of code from `:detekt-core` and simplify `:detekt-tooling`

Related with #6866

### DIFF
--- a/config/detekt/argsfile
+++ b/config/detekt/argsfile
@@ -3,7 +3,7 @@
 -b
 ./config/detekt/baseline.xml
 -ex
-**/resources/**,**/detekt*/build/**,**/build-logic/build/**,**/build-logic/bin/**
+**/resources/**,detekt*/build/**,build-logic/build/**,build-logic/bin/**
 --build-upon-default-config
 -c
 ./config/detekt/detekt.yml
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
@@ -4,7 +4,14 @@ import io.github.detekt.tooling.api.spec.ProcessingSpec
 import io.github.detekt.tooling.api.spec.RulesSpec
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.RuleSet
+import io.gitlab.arturbosch.detekt.api.internal.PathFilters
+import java.nio.file.Path
+import kotlin.io.path.ExperimentalPathApi
+import kotlin.io.path.absolute
+import kotlin.io.path.extension
+import kotlin.io.path.walk
 
+@Suppress("LongMethod")
 internal fun CliArgs.createSpec(output: Appendable, error: Appendable): ProcessingSpec {
     val args = this
     return ProcessingSpec {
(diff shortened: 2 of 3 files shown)


detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgsSpec.kt
```diff
@@ -14,48 +14,165 @@ import org.junit.jupiter.api.Nested
...
 import kotlin.io.path.absolute
 
 internal class CliArgsSpec {
 
-    private val projectPath: Path = resourceAsPath("/").parent.parent.parent.parent.absolute()
-
     @Nested
     inner class `Parsing the input path` {
+        private val pathBuildGradle = Path("build.gradle.kts").absolute()
+        private val pathCliArgs = Path("src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt").absolute()
```
3flex, line 27: Related to my other comment, I don't think these should be passed as absolute paths in tests.

3flex (changes requested):
See inline comments - particularly the one for path filtering.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
```diff
@@ -16,9 +23,20 @@ internal fun CliArgs.createSpec(output: Appendable, error: Appendable): Processi
...
-            inputPaths = args.inputPaths
-            excludes = args.excludes?.let(::asPatterns).orEmpty()
-            includes = args.includes?.let(::asPatterns).orEmpty()
+            val pathFilters = PathFilters.of(
+                args.excludes?.let(::asPatterns).orEmpty(),
+                args.includes?.let(::asPatterns).orEmpty(),
+            )
+            inputPaths = args.inputPaths.walk()
+                .map { it.absolute().normalize() }
+                .filter { pathFilters?.isIgnored(it) != false }
```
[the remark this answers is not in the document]
BraisGabin: You are right, I should not use the absolute path. I change it. What do you think about this implementation (I did the changes on the last commit to make them easier to review)

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgsSpec.kt
BraisGabin: I don't understand this. The `cli` should pass absolute paths to the core. And these variables are used to help me ensure that the correct files are passed.

BraisGabin:
I broke something... I need to check.

config/detekt/argsfile
```diff
@@ -3,7 +3,7 @@
 -b
 ./config/detekt/baseline.xml
 -ex
-**/resources/**,**/detekt*/build/**,**/build-logic/build/**,**/build-logic/bin/**
+**/resources/**,detekt*/build/**,build-logic/build/**,build-logic/bin/**
```
BraisGabin, line 6: Now that we don't take into account the absolute path when you provide excludes/includes you need to be more careful with the initial `**/`. `**/` match at least one directory. If your directory on the basePath then it doesn't match. For that reason I needed to update these excludes.

3flex: >  If your directory on the basePath then it doesn't match

This will need to be called out loud & clear in the release notes - I also don't recall if basepath is always set but we should make sure it's always set to a reasonable default value by default.

3flex: (it's the right thing to do though!)

BraisGabin: About default values for `bashPath`:

- `cli` uses the working directory.
- `gradlePlugin` as far as I saw it passes the project path.

Both of them seems like good default values to me.

---

And about call out this change, I agree. This is a breaking change on the `cli`. But I'm not sure what can I do now about it.

3flex (approved):

PR #7009, opened 2024-03-02
Remove `KtTreeCompiler`

BraisGabin:
After #7007 all our clients provide a list of files to evaluate, it doesn't provide directories nor include/exclude rules.

For that reason we can remove `KtTreeCompiler` and the code related with `include/exclude` input.

This PR is ready but it goes on top of #7007


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/Spec.kt
```diff
@@ -4,7 +4,14 @@ import io.github.detekt.tooling.api.spec.ProcessingSpec
 import io.github.detekt.tooling.api.spec.RulesSpec
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.RuleSet
+import io.gitlab.arturbosch.detekt.api.internal.PathFilters
+import java.nio.file.Path
+import kotlin.io.path.ExperimentalPathApi
```
3flex, line 9: If we're OK with this there are a few other places I'd switch to using the experimental functions too - but I've held off, since they're experimental :)

Unfortunately they're still experimental in Kotlin master branch so probably will still be in 2.0.0 final.

> Beware using the annotated API especially if you're developing a library, since your library might become binary incompatible with the future versions of the standard library.

This probably isn't actually a concern for us since we don't need to maintain compatibility with different stdlib versions.

BraisGabin: I think that in the cli there is no problem to use these experimental functions. But probably we shouldn't use them on our public APIs.

3flex (approved):

PR #7016, opened 2024-03-04
Refactor compilation analysis task setup

3flex:
This removes a fair bit of duplication and code that had similar purpose but slight behavioural differences. It also stops creating redundant non-type-resolution tasks for KMP since there are now source set analysis tasks created for common, iOS and JS source sets.

This will ease future refactoring.


detekt-gradle-plugin/src/main/kotlin/dev/detekt/gradle/plugin/internal/DetektAndroidCompilations.kt
```diff
@@ -0,0 +1,97 @@
...
+                        it.dependsOn(DetektPlugin.BASELINE_TASK_NAME + variant.name.capitalize())
+                    }
+                    testVariants(variant)
+                        .filter { !extension.matchesIgnoredConfiguration(it) }
+                        .forEach { testVariant ->
+                            testTaskProvider.configure {
+                                it.dependsOn(DetektPlugin.DETEKT_TASK_NAME + testVariant.name.capitalize())
+                            }
+                            testBaselineTaskProvider.configure {
+                                DetektPlugin.BASELINE_TASK_NAME + testVariant.name.capitalize()
```
BraisGabin, line 91: These does nothing, right? Shouldn't we add `it.dependsOn()`?

detekt-gradle-plugin/src/main/kotlin/dev/detekt/gradle/plugin/internal/DetektAndroidCompilations.kt
```diff
@@ -0,0 +1,97 @@
...
+                it.group = "verification"
+                it.description = "EXPERIMENTAL: Creates detekt baseline files for test classes across " +
+                    "all variants with type resolution"
+            }
+
+        fun variants(extension: BaseExtension): DomainObjectSet<out BaseVariant>? = when (extension) {
+            is AppExtension -> extension.applicationVariants
+            is LibraryExtension -> extension.libraryVariants
+            is TestExtension -> extension.applicationVariants
+            else -> null
```
BraisGabin, line 63: Is this expected? Should we just throw an exception instead so if this happen we get faster and easier reports to fix it?

detekt-gradle-plugin/src/main/kotlin/dev/detekt/gradle/plugin/internal/DetektKmpJvmCompilations.kt
```diff
@@ -0,0 +1,21 @@
...
+import io.gitlab.arturbosch.detekt.extensions.DetektExtension
+import org.gradle.api.Project
+import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType.androidJvm
+import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType.jvm
+import org.jetbrains.kotlin.gradle.plugin.KotlinTargetsContainer
+
+internal object DetektKmpJvmCompilations {
+    fun registerTasks(project: Project, extension: DetektExtension) {
+        val kotlinExtension = project.extensions.getByName("kotlin")
+        if (kotlinExtension !is KotlinTargetsContainer) return
```
BraisGabin, line 12: Same here. Is this expected? Should we just throw an exception instead so if this happen we get faster and easier reports to fix it?

detekt-gradle-plugin/src/main/kotlin/dev/detekt/gradle/plugin/internal/SharedTasks.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        setSource(siblingTask.sources)
+        classpath.setFrom(compilation.output.classesDirs, siblingTask.libraries)
+
+        // If a baseline file is configured as input file, it must exist to be configured, otherwise the task fails.
+        // We try to find the configured baseline or alternatively a specific variant matching this task.
+        extension.baseline.asFile.orNull?.existingVariantOrBaseFile(compilation.name)?.let { baselineFile ->
+            baseline.convention(layout.file(provider { baselineFile }))
+        }
+        description = if (target != null) {
```
BraisGabin, line 32: Shouldn't we also set the `group`? (In both this and baseline)

detekt-gradle-plugin/src/main/kotlin/dev/detekt/gradle/plugin/internal/DetektAndroidCompilations.kt
3flex: 🤦

detekt-gradle-plugin/src/main/kotlin/dev/detekt/gradle/plugin/internal/DetektAndroidCompilations.kt
3flex: This is a copy paste from the current logic so I'm leaving it as is - I don't think we've ever had an issue for it. I want to remove all this Android-specific logic anyway so it will go soon enough.

detekt-gradle-plugin/src/main/kotlin/dev/detekt/gradle/plugin/internal/DetektKmpJvmCompilations.kt
3flex: Yeah at this point we already know multiplatform plugin is applied and the multiplatform extension implements `KotlinTargetsContainer` so that will be safe to do. I was being defensive here, if there's an issue with a new Kotlin version our tests will expose it, but there's no harm making it an exception.

BraisGabin (approved):

PR #7030, opened 2024-03-08
Don't allow invalid Source Locations

### LINKED ISSUE #7024: Invalid SARIF on EmptyKtFile
## Expected Behavior

A valid SARIF report with details about the empty file.

## Observed Behavior

SARIF report is invalid and if you are using GitHub SARIF uploading is failing due to that.

## Steps to Reproduce

1. Configure detekt to recognise `EmptyKtFile`
2. Create an Empty.kt file (Encoding or line ending of the file does not matter)
3. Run cli command `detekt-cli-1.23.5\bin\detekt-cli.bat --config ../config.yml --report sarif:..\empty.sarif`
```
MyApplication\app\src\main\java\com\example\myapplication\Empty.kt:-1:0: The empty Kotlin file MyApplication\app\src\main\java\com\example\myapplication\Empty.kt can be removed. [EmptyKtFile]

Analysis failed with 1 weighted issues.
```
5. Validate the SARIF report or if possible try using the upload to github. For outside validation I used [this](https://sarifweb.azurewebsites.net/Validation) online tool I am not aware if there is something more official available.
(Code of MyApplication doesn't really matter it was a dummy android studio project)

## Context

Issue identified when GitHub was failing to display the SARIF results inside the PR reviewing tool trying to identified why happening I narrowed down the issue to the fact that erroneously due to some renamings or movements of files in our repository we had some empty `*.kt` files. 

## Your Environment

* Version of Gradle used (if applicable): cli tool used
* Operating System and version: Windows OS 10, MacOS, Ubuntu
* Link to your project (if it's a public repository): Not public

## Attached
The files I used for this issue report reproduction.
- config.yml : The minimum config that enable the  `EmptyKtFile` rule
- empty.sarif : The produced report by running the command in the preproduction step
- Empty.kt : The empty file I created, you can also just delete all contents of any .kt file you have
[ReproductionFiles.zip](https://github.com/detekt/detekt/files/14514964/ReproductionFiles.zip)

BraisGabin:
Fix #7024

A line or column value less than 1 is not valid. Detekt didn't check this but sarif those so we were generating invalid sarif.

Other way to solve the issue is to move this code to the sarif serializer but I think that it's more correct to fix it in our own code and don't allow to point to a source location that doesn't exist.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
@@ -72,7 +72,7 @@ class Location(
 class SourceLocation(val line: Int, val column: Int) {
     init {
         require(line > 0) { "The source location line must be greater than 0" }
-        require(column > 0) { "The source location column must should be greater than 0" }
+        require(column > 0) { "The source location column must be greater than 0" }
     }
 
     override fun toString(): String = "$line:$column"
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
@@ -71,7 +71,7 @@ class Location(
 @Poko
 class SourceLocation(val line: Int, val column: Int) {
     init {
-        require(line > 0) { "The source location line must should be greater than 0" }
+        require(line > 0) { "The source location line must be greater than 0" }
         require(column > 0) { "The source location column must should be greater than 0" }
     }
 
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
@@ -33,9 +33,9 @@ class Location(
          */
         fun from(element: PsiElement, offset: Int = 0): Location {
             val start = startLineAndColumn(element, offset)
-            val sourceLocation = SourceLocation(start.line, start.column)
+            val sourceLocation = SourceLocation(start.line.coerceAtLeast(1), start.column.coerceAtLeast(1))
             val end = endLineAndColumn(element, offset)
-            val endSourceLocation = SourceLocation(end.line, end.column)
+            val endSourceLocation = SourceLocation(end.line.coerceAtLeast(1), end.column.coerceAtLeast(1))
             val textLocation = TextLocation(element.startOffset + offset, element.endOffset + offset)
             return Location(sourceLocation, endSourceLocation, textLocation, element.containingFile.toFilePath())
         }
@@ -70,6 +70,11 @@ class Location(
  */
 @Poko
 class SourceLocation(val line: Int, val column: Int) {
+    init {
+        require(line > 0) { "The source location line must should be greater than 0" }
+        require(column > 0) { "The source location column must should be greater than 0" }
+    }
+
     override fun toString(): String = "$line:$column"
 }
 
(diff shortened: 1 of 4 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
```diff
@@ -70,6 +70,11 @@ class Location(
  */
 @Poko
 class SourceLocation(val line: Int, val column: Int) {
+    init {
+        require(line > 0) { "The line should be greater than 0" }
```
3flex, line 74:
```suggestion
        require(line > 0) { "The source location line must be greater than 0" }
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
```diff
@@ -70,6 +70,11 @@ class Location(
...
+        require(column > 0) { "The column should be greater than 0" }
```
3flex, line 75:
```suggestion
        require(column > 0) { "The source location column must be greater than 0" } }
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
```diff
@@ -33,9 +33,9 @@ class Location(
          */
         fun from(element: PsiElement, offset: Int = 0): Location {
             val start = startLineAndColumn(element, offset)
-            val sourceLocation = SourceLocation(start.line, start.column)
+            val sourceLocation = SourceLocation(start.line.coerceAtLeast(1), start.column.coerceAtLeast(1))
```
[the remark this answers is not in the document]
BraisGabin: What do you think about the new implementation? If a file is empty then `DiagnosticUtils.getLineAndColumnInPsiFile` doesn't provide a correct value. You can see it here:

```
    @NotNull
    public static LineAndColumn offsetToLineAndColumn(@Nullable Document document, int offset) {
        if (document != null && document.getTextLength() != 0) {
            int lineNumber = document.getLineNumber(offset);
            int lineStartOffset = document.getLineStartOffset(lineNumber);
            int column = offset - lineStartOffset;
            int lineEndOffset = document.getLineEndOffset(lineNumber);
            CharSequence lineContent = document.getCharsSequence().subSequence(lineStartOffset, lineEndOffset);
            return new LineAndColumn(lineNumber + 1, column + 1, lineContent.toString());
        } else {
            return new LineAndColumn(-1, offset, (String)null);
        }
    }
```

So in that case I return `null`. And then the default value that we have for the cases where we don't get the line/column returns a valid line/column.

I don't think I should fix this at the rule level because any rule that reports an empty file would crash.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
```diff
@@ -70,6 +70,11 @@ class Location(
  */
 @Poko
 class SourceLocation(val line: Int, val column: Int) {
+    init {
+        require(line > 0) { "The source location line must should be greater than 0" }
```
3flex, line 74:
```suggestion
        require(line > 0) { "The source location line must be greater than 0" }
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
```diff
@@ -70,6 +70,11 @@ class Location(
...
+        require(column > 0) { "The source location column must should be greater than 0" }
```
3flex, line 75:
```suggestion
        require(column > 0) { "The source location column must be greater than 0" }
```

3flex (approved):

PR #7040, opened 2024-03-12
Allow setting default autocorrect mode with Gradle property

3flex:
This makes it possible to enable autocorrect when running tasks that depend on detekt tasks e.g.
`./gradlew check -Pdetekt.default.autocorrect=true`

Without this, the only way to enable autocorrect on CLI is by declaring every detekt task on the CLI and passing `--auto-correct` to every one.

Tests will be added once https://github.com/detekt/detekt/pull/6913 is merged as that provides some support to pass Gradle properties through to the CLI when setting up tests.


schalkms (commented):
@3flex I intended to do some housekeeping and hence stumbled upon this very old PR. Do we still need this PR open although the blockers seem to still exist?

BraisGabin:
The linked pr was closed but the code related with the tests could be cherry picked.

PR #7041, opened 2024-03-12
Refactor DGP to compile against AGP API

### LINKED ISSUE #3327: Refactor detekt-gradle-plugin off com.android.tools.build:gradle

3flex:
Closes #3327

In AGP 7.1.0 `nestedComponents` was incubating, but it's part of the stable API now.

Requires AGP min version bump to 7.1.0, released in Jan 2022. Marking as a notable change for that reason.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/dev/detekt/gradle/plugin/internal/DetektAndroidCompilations.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/dev/detekt/gradle/plugin/internal/DetektAndroidCompilations.kt
@@ -1,16 +1,9 @@
-@file:Suppress("DEPRECATION")
-
 package dev.detekt.gradle.plugin.internal
 
-import com.android.build.gradle.AppExtension
-import com.android.build.gradle.BaseExtension
-import com.android.build.gradle.LibraryExtension
-import com.android.build.gradle.TestExtension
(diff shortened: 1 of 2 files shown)


BraisGabin (approved):
🎉 But if we want to support the default variant we will need to depend again on AGP :(

3flex:
If we ever need to depend on the full AGP plugin to deliver something then we should not implement it.

AGP API will become the only way to integrate with AGP from AGP 10 according to the roadmap.

PR #7048, opened 2024-03-14
Add DoubleNegativeExpression rule

### LINKED ISSUE #7035: New rule: DoubleNegativeExpression
*This issue is continue of discussion started in https://github.com/detekt/detekt/issues/7019#issuecomment-1983082620*

Expression with two or more calls of operator `not` could be simplified.

```kotlin
// Noncompliant
isValid.not().not()
!isValid.not()
!!isValid
!!!isValid

// Compliant
isValid
!isValid
isValid.not()
```

## Context

When you prefer to use `.not()` over `!`, it is quite easy to get construction like `!isValid.not()` as a result of automatic IDE refactorings (e.g. "Invert 'If' condition", "Replace '||' with '&&'")
Additional context: #7019

t-kameyama:
Fixes #7035

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeExpressionSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeExpressionSpec.kt
@@ -45,7 +45,7 @@ class DoubleNegativeExpressionSpec(private val env: KotlinCoreEnvironment) {
     }
 
     @Test
-    fun `exclamation`() {
+    fun exclamation() {
         val code = """
             fun test(b: Boolean): Boolean {
                 return !b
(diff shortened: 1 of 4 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/DoubleNegativeExpressionSpec.kt
```diff
@@ -0,0 +1,113 @@
...
+            fun test(b: Boolean): Boolean {
+                return b.not().not().not()
+            }
+        """.trimIndent()
+        val actual = subject.compileAndLintWithContext(env, code)
+        assertThat(actual).hasSize(1)
+    }
+
+    @Test
+    fun `!`() {
```
cortinico, line 48: Maybe let's have a longer test name here, especially as they're not `@Nested` (let's add one otherwise).

If those tests fail, it will be really cryptic on console: "DoubleNegativeExpressionSpec !"

cortinico (approved):

PR #7052, opened 2024-03-14
Revert "Improve our Sarif merger task to support other sarif reports"

schalkms:
Reverts detekt/detekt#7043 due to failing test


3flex:
The failure was caused by a different PR - see #7053 for the correct revert.

PR #7055, opened 2024-03-15
Improve logging configs

BraisGabin:
Working on #7054 I had a bad time debugging the differences of two config files because their `toString` where not friendly.

This PR fixes that, an example, before:
```
io.gitlab.arturbosch.detekt.core.config.DisabledAutoCorrectConfig@3fb450d7
```

After:
```
DisabledAutoCorrectConfig(
    YamlConfig(
        {
            style={
                autoCorrect=true,
                MagicNumber={autoCorrect=true},
                MagicString={autoCorrect=false},
            },
            comments={
                autoCorrect=false,
                ClassDoc={autoCorrect=true},
                FunctionDoc={autoCorrect=false},
            },
        },
    ),
)
```

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Config.kt
@@ -68,6 +68,10 @@ interface Config {
             override fun subConfigKeys(): Set<String> = emptySet()
 
             override fun <T : Any> valueOrNull(key: String): T? = null
+
+            override fun toString(): String {
+                return "Config.empty"
+            }
         }
 
         const val ACTIVE_KEY: String = "active"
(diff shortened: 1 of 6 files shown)


config/detekt/detekt.yml
```diff
@@ -86,6 +86,8 @@ formatting:
     active: true
   NoEmptyFirstLineInClassBody:
     active: false
+  StringTemplate:
+    active: false
```
3flex, line 90: Why is this disabled now?

BraisGabin: To allow this line:

```kotlin
            deprecatedRules=${deprecatedRules},
```

Otherwise it force me to remove the `{}`. I don't see a good reason to remove those. They don't make the code harder to read, I think that they help. Anyway, I can remove them and re enable the rule if you disagree, I don't have a strong opinion here.

3flex: I meant, why disable in the config, instead of suppressing where necessary?

BraisGabin: Because I think that if we don't agree with the rule in general it is better to disable it. I think this rule shouldn't be suppressed. If you stick to it you should stick to it. Otherwise is better to disable.

BraisGabin: I reverted it and fix the issue. It is just a preference taste it's not important and completely different that the main objective of the PR.

3flex (approved):

PR #7067, opened 2024-03-17
Remove `HasEntity`

BraisGabin:
Another PR in the line to reduce our public api surface. This time remove `HasEntity`.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding.kt
@@ -3,7 +3,7 @@ package io.gitlab.arturbosch.detekt.api
 /**
  * Represents a problem in the source code detected by a rule.
  *
- * A finding has an issue (information about the rule that detected the problem), a severity and a source code position
+ * A finding has an issue (information about the rule that detected the problem), a severity and a source code position.
  * described as an entity. Entity references can also be considered for deeper characterization.
  */
 interface Finding {
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding.kt
@@ -11,20 +11,3 @@ interface Finding {
     val references: List<Entity>
     val message: String
 }
-
-/**
- * Describes a source code position.
- */
-interface HasEntity {
-    val entity: Entity
-    val location: Location
-        get() = entity.location
-    val startPosition: SourceLocation
-        get() = location.source
-    val charPosition: TextLocation
-        get() = location.text
-    val file: String
-        get() = location.filePath.absolutePath.toString()
-    val signature: String
-        get() = entity.signature
-}
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding.kt
@@ -6,7 +6,8 @@ package io.gitlab.arturbosch.detekt.api
  * A finding has an issue (information about the rule that detected the problem), a severity and a source code position
  * described as an entity. Entity references can also be considered for deeper characterization.
  */
-interface Finding : HasEntity {
+interface Finding {
+    val entity: Entity
     val references: List<Entity>
     val message: String
 }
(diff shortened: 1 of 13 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding2.kt
```diff
@@ -6,13 +6,17 @@ package io.gitlab.arturbosch.detekt.api
...
+interface Finding2 {
     val rule: RuleInfo
+    val entity: Entity
     val references: List<Entity>
     val message: String
     val severity: Severity
     val autoCorrectEnabled: Boolean
 
+    val location: Location
+        get() = entity.location
```
BraisGabin, line 18: Probably this could be removed but I didn't want to take the decision on this PR. This keeps the same implementation that there was before.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding.kt
```diff
@@ -0,0 +1,13 @@
+package io.gitlab.arturbosch.detekt.api
+
+/**
+ * Represents a problem detected by a rule on the source code
```
schalkms, line 4:
```suggestion
 * Represents a problem in the source code detected by a rule.
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding.kt
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * A finding has an issue (information about the rule that detected the problem), a severity and a source code position
```
schalkms, line 6:
```suggestion
 * A finding has an issue (information about the rule that detected the problem), a severity and a source code position.
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding.kt
```diff
@@ -0,0 +1,13 @@
...
+ * described as an entity. Entity references can also be considered for deeper characterization.
```
schalkms, line 7: ... can be considered for more details?

Not quite sure what you mean.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding2.kt
schalkms: Shall be removed I guess.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding.kt
BraisGabin: Fixed

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding2.kt
BraisGabin: I'll take care of this in another PR, I'm working on it right now.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding2.kt
```diff
@@ -6,13 +6,17 @@ package io.gitlab.arturbosch.detekt.api
  * A finding has information about the rule that detected the problem, a severity and an entity with information
  * about the position. Entity references can also be considered for deeper characterization.
  */
-interface Finding2 : HasEntity {
+interface Finding2 {
```
cortinico, line 9: Q: Sorry if this has been answered already and I missed it - What's the plan with renaming/merging this with `Finding`?

cortinico (approved):
Accepting to unblock you 👍

schalkms (approved):

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Finding2.kt
BraisGabin: We talked about it, but I don't remember where. `Finding2` will be renamed to `Issue`. The idea is that the rules report `Finding`s and then the core adds more information to that finding to convert it to an `Issue`. For example, add the information about which rule reported it. This way we simplify the rules.

Future updates will make that the core will handle the suppression (instead of the rules). And the issue will not be removed. It will be marked as a suppressed issue this way the reporters can decide what to do.

cortinico: Great thanks for the clarification 👍

PR #7082, opened 2024-03-23
Prepare Release 1.23.6

cortinico:
Time for a new point release

PR #7083, opened 2024-03-23
Add docs for 1.23.6 on main

cortinico:
Those are the docs for: https://github.com/detekt/detekt/pull/7082


3flex:
There's a build failure for docs which I haven't investigated but I assume that needs fixing before merging?

BraisGabin (approved):

PR #7088, opened 2024-03-25
Use constrained syntax to declare plugins

3flex:
Gradle 8.0 introduced [faster Kotlin DSL build script compilation](https://docs.gradle.org/8.0/release-notes.html#new-features,-performance-and-usability-improvements). Gradle uses its own interpreter to interpret the plugins block, avoiding the need to use the Kotlin compiler, but it supports constrained syntax only. This is defined here: https://docs.gradle.org/8.7/userguide/plugins.html#sec:constrained_syntax

It also says:

> Note that using version catalog aliases for plugins (e.g. alias(libs.plugins.acme)) or type-safe plugin accessors (e.g. `acme-plugin`) will not see a performance improvement. Support for these formats will be added later.

However that support is not currently on the Gradle roadmap for 2024.

The performance improvement of 20% faster script compilation should be worth the loss of type safe accessors (issues are identified at runtime) or version catalog aliases (Renovate will take care of updates).

### DIFF
--- a/build-logic/src/main/kotlin/module.gradle.kts
+++ b/build-logic/src/main/kotlin/module.gradle.kts
@@ -5,8 +5,8 @@ import org.jetbrains.kotlin.gradle.tasks.UsesKotlinJavaToolchain
 plugins {
     id("packaging")
     kotlin("jvm")
-    `maven-publish`
-    jacoco
+    id("maven-publish")
+    id("jacoco")
 }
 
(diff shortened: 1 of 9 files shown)


schalkms (approved):

marschwar:
The mentioned road map entry is https://github.com/gradle/build-tool-roadmap/issues/64

PR #7089, opened 2024-03-25
UnusedPrivateProperty improvements and refactoring out new UnusedVariable rule

### LINKED ISSUE #6702: False Negative: Unused private property not reported in a file if used in another class
## Expected Behavior

```kt
class A(private val foo: Any) {
    fun myFoo() = foo
}
class B(private val foo: Any) // <-- foo should be reported here
```

## Observed Behavior

It is not reprorted because the `unusedParameters` list is per file: https://github.com/detekt/detekt/blob/9b2e1eace3d50a8e19f3ce6b1b43bce0696bbcbd/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedParameter.kt#L68

So `foo` is marked as used because of the usage in class `A`, however it is unused in class `B`

## Steps to Reproduce

I have created a PR to reproduce: https://github.com/detekt/detekt/pull/6701

psuzn:
Fixes #6702 and refactors the unused variable rules into its own separate `UnusedVariable` rule.


psuzn:
This is more or less done, but this breaks the unused variable checks (i.e. failng tests). The previous version of `UnusedPrivateProperty` was also doing checks for the unused variables. I can add those checks but it would be better to put checks for unused variables in on the seperate rule and keep this `UnusedPrivateProperty` rule just for unused private properties and constructor parameters. 

@BraisGabin @3flex 

@

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateProperty.kt
```diff
@@ -56,103 +67,163 @@ class UnusedPrivateProperty(config: Config) : Rule(
         setOf("UNUSED_VARIABLE", "UNUSED_PARAMETER", "unused", "UnusedPrivateMember")
 
     @Configuration("unused property names matching this regex are ignored")
-    private val allowedNames: Regex by config("_|ignored|expected|serialVersionUID", String::toRegex)
+    private val allowedNames: Regex by config(
+        "_|ignored|expected|serialVersionUID",
```
BraisGabin, line 71: If this rule only works with properties then `_` should be removed.

psuzn:
@BraisGabin how should I go with the unused variable rule, I can add it to this PR or do that in a new PR (we can create a new 3rd PR as a base for both). 

Also, what should be the name for the new rule `UnusedVariable`?

cortinico:
> And there is other thing here, we are moving another rule away from plain detekt. I'm ok with that in general. Correctness is better than speed. But I wanted just to make that clear for anyone looking at this PR.

I'm also fine with it 👍
Thanks for taking a stab at this @psuzn 

> Also, what should be the name for the new rule UnusedVariable?

`UnusedVariable` sounds like a good one 👍

BraisGabin (commented):
Top level properties shouldn't be handled by `UnusedVariables` because they aren't variables, they are properties. That should remain on `UnusedPrivateProperty`.

Other than that the PR looks really good!

---

Off-topic: Do we have a rule to check that all the parameters of a lambda are used? 🤔 I thought we have but I can't find it.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedVariable.kt
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * <compliant>
+ * fun foo() {
+ *     val used = "used"
+ *     println(used)
+ * }
+ * </compliant>
+ */
+@RequiresTypeResolution
+@ActiveByDefault(since = "1.23.0")
```
BraisGabin, line 48:
```suggestion
@ActiveByDefault(since = "2.0.0")
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedVariableSpec.kt
```diff
@@ -0,0 +1,341 @@
...
+            """.trimIndent()
+            assertThat(subject.lintWithContext(env, code)).hasSize(2)
+        }
+
+        @Test
+        fun `does not report used loop parameters in indexed array`() {
+            val code = """
+                fun use() {
+                    val array = intArrayOf(1, 2, 3)
+                    for ((index, value) in array.withIndex()) {
```
BraisGabin, line 211: Can you add a test deconstructing outside a `for`?

Something like `val (used, unused) = 1 to 2`.

BraisGabin (commented):
Add one test more and this is good to go for me :) really good job! Thank you!

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedVariableSpec.kt
psuzn: sure

BraisGabin (approved):

BraisGabin:
@psuzn can you rebase/merge your pr with main? It has conflicts. After that this is ready to merge.

psuzn:
should be good to go.

PR #7090, opened 2024-03-25
Rename `Finding2` to `Issue`

BraisGabin:
From now on detekt doesn't report `findings` it reports `issues`.


cortinico (approved):
I suppose this is just an automatic renaming right @BraisGabin I haven't looked the full diff but at a first glance it looks good 👍

BraisGabin:
It's a bit mixed. Because only renaming the `class` a lot of variables were still called `findings` and I renamed all the ones that I found, same with related functions and classes (the related classes are just 3 console reporters)

PR #7092, opened 2024-03-26
Refactor CLI args processing

3flex:
The only functional change is that directories passed to detekt CLI are validated as directories. Other than that, this just makes the code a bit more consistent by using more of JCommander's features, and using stronger typing when args are processed.

### DIFF
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ArgumentConverters.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ArgumentConverters.kt
@@ -20,10 +20,6 @@ class ExistingPathConverter : IStringConverter<Path> {
     }
 }
 
-class PathConverter : IStringConverter<Path> {
-    override fun convert(value: String) = Path(value)
-}
-
 interface DetektInputPathConverter<T> : IStringConverter<List<T>> {
     val converter: IStringConverter<T>
     override fun convert(value: String): List<T> =
--- a/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ArgumentConverters.kt
+++ b/detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ArgumentConverters.kt
@@ -38,21 +38,15 @@ class ClasspathResourceConverter : IStringConverter<URL> {
 }
 
 class FailureSeverityConverter : IStringConverter<FailureSeverity> {
-    override fun convert(value: String): FailureSeverity {
-        return FailureSeverity.fromString(value)
-    }
+    override fun convert(value: String): FailureSeverity = FailureSeverity.fromString(value)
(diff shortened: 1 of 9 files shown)


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/ArgumentConverters.kt
```diff
@@ -74,3 +42,29 @@ class FailureSeverityConverter : IStringConverter<FailureSeverity> {
...
+class ReportPathConverter : IStringConverter<ReportPath> {
+    override fun convert(value: String): ReportPath {
+        return ReportPath.from(value)
+    }
+}
+
+class PathSplitter : IParameterSplitter {
+    override fun split(value: String): List<String> {
+        return value.split(',', ';')
+    }
```
cortinico, line 55: nit: they can be converted to expression returns

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/GeneratorArgs.kt
```diff
@@ -1,33 +1,45 @@
...
     @Parameter(
         names = ["--input", "-i"],
         required = true,
+        converter = PathConverter::class,
+        splitter = PathSplitter::class,
+        validateValueWith = [PathValidator::class],
         description = "Input paths to analyze."
     )
-    private var input: String? = null
+    var inputPath: List<Path> = emptyList()
```
cortinico, line 24: Q: moving from `String?` to `List<Path>` makes an empty input valid? Should we check inside `PathValidator` that the input is not empty?

cortinico (approved):
Mostly nits

detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/GeneratorArgs.kt
3flex: If input is empty JCommander returns an error.

`java -jar detekt-generator-1.23.6-all.jar -c . -d . -i` returns `ParameterException: Expected a value after parameter -i`.

Any other value is validated to check that the paths exist. The only way to pass an empty list is to pass only a separator like `-i ,` or `-i ;` but I don't think it's worth guarding against this.

BraisGabin:
This PR also have another functional change that I think is good but just to point it out. When passing the inputs we allowed things like this `"src/main , foo" and we map that to `src/main` and `foo`. Now we map it to `src/main ` and ` foo`.

Also the new exception doesn't allow to spot those errors easily because the path doesn't quote the path so it is really difficult to spot that space. I'm going to create a new PR to fix this last point.

BraisGabin:
Also we should use the splitter on the `--includes` and `--excludes` too. But I would like to first merge #7007 to avoid more conflicts.

PR #7101, opened 2024-03-27
Move suppressions to `core`

BraisGabin:
All the suppression logic should be handled by the `core`. Rule shouldn't know about that. This PR moves that logic to `core` and also moves the functions that are no longer used in `api` to `core` to keep `api` as clean as possible.


detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt
```diff
@@ -25,12 +27,12 @@ fun Rule.compileAndLint(@Language("kotlin") content: String): List<Finding> {
 
 fun Rule.lint(@Language("kotlin") content: String): List<Finding> {
     val ktFile = compileContentForTest(content)
-    return visitFile(ktFile)
+    return visitFile(ktFile).filterSuppressed(this)
```
BraisGabin, line 30: I don't like to add this here too much but I think that for now it's the simpler solution.

Why do we need this? Right now we have some tests on the rules that check that when an annotation is set the finding is not raised. Those tests are testing two things. From one side they are testing that the suppression works, that's tested already somewhere else so it should not be tested. But they are also testing that the `KtElement` that they provide is correct or that the configured `alias` is correct. We could refactor those tests but the tests would be more difficult to follow (I tried). For that reason I decide to keep this here.

The good part is that, if we want, we could refactor those tests and remove this later.

BraisGabin:
@detekt/maintainers could I get a review here?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
```diff
@@ -101,15 +98,9 @@ open class Rule(
 
     /**
      * Reports a single code smell finding.
```
schalkms, line 100: At this point, detekt does not report, bur rather add it to the findings list.

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt
schalkms: Step-by.step is the way to go here. As you said, I don't think it's ideal approach, though sufficient for now.

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/AnalyzerSpec.kt
```diff
@@ -384,6 +384,29 @@ class AnalyzerSpec(val env: KotlinCoreEnvironment) {
             }
             assertThat(findings).isEmpty()
         }
+
+        @ParameterizedTest
+        @ValueSource(strings = ["MaxLineLength", "detekt.MaxLineLength", "MLL", "custom", "all"])
+        fun `when the ktElement is suppressed the issue is not raised`(suppress: String) {
```
schalkms, line 390: Tests are way cleaner now.

schalkms (approved):
👍
Good to go in my point of view.
Just the function comment needs to be adapted.

PR #7105, opened 2024-03-27
Move classes to psi-utils

BraisGabin:
Move `FunctionMatcher` and `AnnotationExcluder` to `:detekt-psi-utils`.

`FunctionMatcher` was in `:detekt-tooling` that forced some rule sets depend on `:detekt-tooling` and that have no sense. `:detekt-tooling` should be our public API for clients (cli, gradle, intellij, bazel...) it should not be related with how to build rules.

`AnnotationExcluder` was on `:detekt-api` but it doesn't belong there. It's an useful class to write some rules but you can implement anything inside detekt without using it. `:detekt-api` should be only expose the necessary classes/interfaces for integrate things inside detekt.

Those two functions are just "utils" to write rules or similar code, for that reason they belong to `:detekt-psi-utils`.


3flex (approved):

3flex:
> :detekt-api should be only expose the necessary classes/interfaces for integrate things inside detekt.

Agreed, but note that detekt-api has an api dependency on detekt-psi-utils, so any project that depends on detekt-api will pull in psi-utils as a transitive dependency, so moving `AnnotationExcluder` doesn't really change anything. It will be on the classpath before & after this change.

Moving `FunctionMatcher` 100% makes sense though.

BraisGabin:
I'm away from a computer right now. But I would like to remove the dependency between psi-utils and api. They shouldn't depend between each other.

PR #7107, opened 2024-03-28
Refactor `Suppressions.kt`

BraisGabin:
The idea of suppression is to look if the reported element (or any of it's parents) has a `@Suppression` annotation with a matching name (what is a matching name can be discussed later, I didn't touch that part at all).

The new implementation creates a `Sequence` with all the annotations that the element and all it's parents have. And then check if any of them matches. And, because `Sequence` is lazy, as soon as we find one we don't need to continue to traverse the parents so this shouldn't create any performance issue.

With that idea I think that `Suppresson.kt` is easier to read now. I know that it's not really easy now wither but PSI doesn't help.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Suppressions.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Suppressions.kt
@@ -3,58 +3,42 @@ package io.gitlab.arturbosch.detekt.api.internal
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.RuleSet
 import org.jetbrains.kotlin.psi.KtAnnotated
+import org.jetbrains.kotlin.psi.KtAnnotationEntry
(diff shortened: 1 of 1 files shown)


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/internal/Suppressions.kt
```diff
@@ -3,58 +3,42 @@ package io.gitlab.arturbosch.detekt.api.internal
...
+    }
+    acceptedSuppressionIds.addAll(aliases)
 
-private fun KtElement.findAnnotatedSuppressedParent(
-    id: Rule.Id,
-    aliases: Set<String>,
-    ruleSetId: RuleSet.Id? = null
-): Boolean {
-    val parent = getStrictParentOfType<KtAnnotated>()
+    return allAnnotationEntries()
```
chao2zhang, line 22: 💯 Nice usage of sequence. Readability is improved significantly from my opinion.

chao2zhang (approved):

PR #7108, opened 2024-03-28
Rewrite `Suppressions.kt` tests

BraisGabin:
The reason of this rewrite is #7101. The previous tests relay on how the `Rule` works to test what `isSuppressedBy` those. The new implementation depends on nothing else than what they are testing. This helps to unblocks #7101 a bit.

Also these test should be more exhaustive and easier to read and maintain.


detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/SuppressionSpec.kt
```diff
@@ -1,215 +1,282 @@
...
-private class TestLM(config: Config = Config.empty) : Rule(config, "") {
-    override fun visitNamedFunction(function: KtNamedFunction) {
-        val start = Location.from(function.funKeyword!!).source.line
-        val end = Location.from(function.lastBlockStatementOrThis()).source.line
-        val offset = end - start
-        if (offset > 10) report(CodeSmell(Entity.from(function), message = "TestMessage"))
-    }
+private fun KtFile.getMethod(): KtElement {
+    return findChildByClass(KtClass::class.java)!!
+        .findFunctionByName("function")!!
```
chao2zhang, line 269: Minor but should `"function"` be a parameter of this `KtFile.getMethod(functionName: String = "function")`? It reads a bit awkward to me

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/SuppressionSpec.kt
```diff
@@ -1,215 +1,282 @@
...
 }
 
-private class TestLPL(config: Config = Config.empty) : Rule(config, "") {
-    override fun visitNamedFunction(function: KtNamedFunction) {
-        val size = function.valueParameters.size
-        if (size > 5) report(CodeSmell(Entity.from(function), message = "TestMessage"))
-    }
+private fun KtFile.getMethodParameter(): KtElement {
+    return findChildByClass(KtClass::class.java)!!
+        .findFunctionByName("function")!!
```
chao2zhang, line 276: Same here

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/SuppressionSpec.kt
```diff
@@ -1,215 +1,282 @@
...
 
-        private fun assertCodeIsSuppressed(@Language("kotlin") code: String, config: Config) {
-            val findings = TestLPL(config).lint(code)
-            assertThat(findings).isEmpty()
-        }
+    @ParameterizedTest
+    @ValueSource(
+        strings = ["aLL", "ruleid", "RuleId2", "RuleId.RuleSetId", "RuleSetId.alias1", "RuleSetId:alias2"]
+    )
+    fun shouldSuppressNotSuppress(value: String) {
```
chao2zhang, line 255:
```suggestion
    fun `should not suppress with @Suppress annotation with unexpected value`(value: String) {
```

chao2zhang (approved):

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/SuppressionSpec.kt
BraisGabin: It is a bit awkward because this functions only work if the `KtFile` is more less as the ones that I defined in the different tests. If you check them they are very similar, the only change is where is the annotation. Adding a parameter to this function has little sense because, anyway, this function only work with that structure of `KtFile`. I changed this a bit to remove the hardcode `"function"` but the idea is still the same. 

I'm going to merge this now because it unblocks me for other PRs but I you think that it's not good enough tell me and I can look at it and find a better solution on other PR.

PR #7118, opened 2024-04-02
Configure reproducible archives

3flex:
Using this configuration: https://docs.gradle.org/8.7/userguide/working_with_files.html#sec:reproducible_archives

This should help testing build changes that affect JAR output. And as a follow up we can get listed at https://github.com/jvm-repo-rebuild/reproducible-central


BraisGabin (approved):
How difficult can it be to build a proper jar? Impossible! 🤣.

I didn't know that this was even a thing. I keep learning :) thanks for the useful links on the pr.

PR #7122, opened 2024-04-03
Migrate to Develocity Gradle Plugin

3flex:


3flex:
@detekt/maintainers the environment variable for authentication is now named `DEVELOCITY_ACCESS_KEY`. I cannot view the existing value for `GRADLE_ENTERPRISE_ACCESS_KEY` in the GitHub org secrets as secrets are write-only, and I don't seem to have access to generate a new one at https://ge.detekt.dev.

Can someone please generate a new key and copy it to DEVELOCITY_ACCESS_KEY and GRADLE_ENTERPRISE_ACCESS_KEY? We'll need the existing environment variable for a while but it can be deleted in future.

cortinico (approved):

3flex:
Windows build failure should be fixed by #7119

settings.gradle.kts
```diff
@@ -43,39 +41,31 @@ include("detekt-utils")
...
-            this as BuildScanExtensionWithHiddenFeatures
-            publishIfAuthenticated()
+            publishing.onlyIf { it.isAuthenticated }
         }
 
-        isUploadInBackground = !isCiBuild
-
-        capture {
-            isTaskInputFiles = true
-        }
```
TWiStErRob, line 70: @3flex I think this has an equivalent, it wasn't removed:
```gradle
develocity {
	buildScan {
		capture { 
			fileFingerprints = true
		}
	}
}
```

Old:
https://docs.gradle.com/enterprise/gradle-plugin/api/com/gradle/scan/plugin/BuildScanCaptureSettings.html#setTaskInputFiles(boolean)
New:
https://docs.gradle.com/enterprise/gradle-plugin/api/com/gradle/develocity/agent/gradle/scan/BuildScanCaptureConfiguration.html#getFileFingerprints()

3flex: Yes but it's now enabled by default according to the docs so I removed the explicit config.

TWiStErRob: huh, I thought it might be that, but the javadoc says "defaults to false"

TWiStErRob: Can you point to the doc sentence that says it's enabled by default? I can't find it.

3flex: https://docs.gradle.com/enterprise/gradle-plugin/current/#capturing_file_fingerprints

TWiStErRob: Thanks! Raised https://github.com/gradle/gradle/issues/28694

Issue #7140, opened 2024-04-06
Break the dependency of `:detekt-api` with `:detekt-psi-utils`


BraisGabin:
## Expected Behavior
`:detekt-api` shouldn't depend on `:detekt-psi-utils`.

## Current Behavior
It does

## Context
This idea comes from this comment: https://github.com/detekt/detekt/pull/7105#issuecomment-2034125369

I did a fast check and the main reason to have this dependency is because `:detekt-api` uses `FilePath` and that class is declared on `:detekt-psi-utils`. That have little sense, that class should be defined on `:detekt-api`. But once you move that class to `:detekt-api` a lot of `:detekt-psi-utils` doesn't compile because they use `FilePath`. So maybe the solution is to invert the dependency, I'm not sure.

Also this could help to unblock #7123.

cortinico:
> and that class is declared on `:detekt-psi-utils`. That have little sense, that class should be defined on `:detekt-api`. But once you move that class to `:detekt-api` a lot of `:detekt-psi-utils` doesn't compile because they use `FilePath`

Which class are we talking about?

BraisGabin:
`FilePath`

3flex:
I'd love to get rid of that class. We have basePath, which should be required from tooling, or have a sensible default. If other paths are absolute, we can always generate the relative path from the absolute path and basePath. No need to store all versions of the path, nor do we need to convert anything to a FilePath instead of just using a Path.

Relative paths are only needed when outputting results. Absolute paths should always be used internally.

BraisGabin:
That's true... And it shouldn't be difficult to change.

BraisGabin:
My findings:

Right now on a `KtFile` we store as `UserDataProperty` 3 different paths:

- `relativePath`
- `absolutePath`
- `basePath`

For sure we don't need 3 because with 2 of them we can calculate the other. So we could reduce it to:
- `relativePath`
- `basePath`

But we can do it even better. `basePath` is ALWAYS the same for ALL the `KtFile`s. We don't need to store this value on every single `KtFile`. So we can just store `relativePath` and then store `basePath` somewhere else.

Note: because we just have one path now on `KtFile` we could rename it to just `path`. But I'm going to keep calling it `relativePath` here to avoid confusions.

The problem with `relativePath` is that it is a `Path` and if you call `relativePath.absolute()` you expect to get the absolute path of it. But that's not true if `basePath != System.getProperty("user.dir")`. And it is way to easy to fall on that error.

[...]

PR #7147, opened 2024-04-08
Stricter paths

BraisGabin:
Part of #7140

This PR makes our core more strict to how the paths should look like. I know that `KtFiles` will be removed. but that's the reason I'm doing this now. Our tests were a bit "loose" on the paths. With this change our tests are a better representation of what happen in the real life and can spot real bugs once we start to refactor this part of the code to remove `KtFiles`.

### DIFF
--- a/detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/KtFiles.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/KtFiles.kt
@@ -53,6 +53,9 @@ class FilePath(
         ) {
             "Absolute path = $absolutePath much match base path = $basePath and relative path = $relativePath"
         }
+        require(absolutePath.isAbsolute) { "absolutePath should be absolute" }
+        require(basePath?.isAbsolute != false) { "basePath should be absolute" }
+        require(relativePath?.isAbsolute != true) { "relativePath should not be absolute" }
     }
 
     override fun toString(): String =
--- a/detekt-psi-utils/src/test/kotlin/io/github/detekt/psi/KtFilesSpec.kt
+++ b/detekt-psi-utils/src/test/kotlin/io/github/detekt/psi/KtFilesSpec.kt
@@ -5,6 +5,8 @@ import org.assertj.core.api.Assertions.assertThat
 import org.jetbrains.kotlin.com.intellij.psi.PsiFile
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
+import kotlin.io.path.Path
+import kotlin.io.path.absolute
 
 class KtFilesSpec {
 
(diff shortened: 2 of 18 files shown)


detekt-parser/src/main/kotlin/io/github/detekt/parser/KtCompiler.kt
```diff
@@ -22,25 +22,22 @@ open class KtCompiler(
...
+        return createKtFile(path.readText(), basePath, path)
     }
 
     fun createKtFile(content: String, basePath: Path, path: Path): KtFile {
-        val normalizedAbsolutePath = path.absolute().normalize()
-        val lineSeparator = content.determineLineSeparator()
+        require(basePath.isAbsolute) { "Given basePath '$basePath' should be absolute!" }
+        require(path.isAbsolute) { "Given path '$path' should be absolute!" }
+        val relativePath = path.relativeToOrNull(basePath)
+        require(relativePath != null) { "Given basePath '$basePath' should be the a base path of path '$path'!" }
```
BraisGabin, line 32: Main change [1/2]

detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/KtFiles.kt
```diff
@@ -53,6 +53,9 @@ class FilePath(
         ) {
             "Absolute path = $absolutePath much match base path = $basePath and relative path = $relativePath"
         }
+        require(absolutePath.isAbsolute) { "absolutePath should be absolute" }
+        require(basePath?.isAbsolute != false) { "basePath should be absolute" }
+        require(relativePath?.isAbsolute != true) { "relativePath should not be absolute" }
```
BraisGabin, line 58: Main change [2/2]

BraisGabin:
Still working on make Windows tests happy :'(. My main problem right now is that `/` is not an absolute path on Windows... So to get the/a root I need to do `Path("").absolute.root`. This gets me `/` on Unix and `D:\` (for example) on Windows.

The main problem is that I don't have windows so I need to use CI to debug the problems :'(

detekt-psi-utils/src/test/kotlin/io/github/detekt/psi/KtFilesSpec.kt
```diff
@@ -47,23 +48,24 @@ class KtFilesSpec {
...
-
-        @Test fun `toString of absolute path`() {
-            val filePath = FilePath.fromAbsolute(File("/a/b/c").toPath())
+        @Test
+        fun `toString of absolute path`() {
+            val absolutePath = Path("/").absolute().resolve("a/b/c")
+            val filePath = FilePath.fromAbsolute(absolutePath)
 
             assertThat(filePath.toString())
-                .isEqualTo("FilePath(absolutePath=${ps}a${ps}b${ps}c, basePath=null, relativePath=null)")
```
cortinico, line 58: nice :)

cortinico (approved):
Looks good to me 👍 although I'm not really expert on paths so you probably want to wait anothe review.

Also there are conflicts to resolve

PR #7151, opened 2024-04-09
Align build cache config across builds

3flex:
Somehow things got out of sync between the builds. Fixes so they're aligned and added a reminder so this doesn't happen again.

After merging the GRADLE_CACHE_PASSWORD and GRADLE_CACHE_USERNAME secrets can be safely deleted. They're actually not used in any actions right now but better safe than sorry :)

### DIFF
--- a/build-logic/settings.gradle.kts
+++ b/build-logic/settings.gradle.kts
@@ -1,6 +1,7 @@
 rootProject.name = "build-logic"
 
 plugins {
+    id("com.gradle.develocity") version "3.17"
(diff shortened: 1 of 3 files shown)


BraisGabin (approved):

cortinico (approved):

cortinico:
> After merging the GRADLE_CACHE_PASSWORD and GRADLE_CACHE_USERNAME secrets can be safely deleted. They're actually not used in any actions right now but better safe than sorry :)

I've removed them 👍

PR #7152, opened 2024-04-09
Enable and encrypt Gradle configuration cache on CI

### LINKED ISSUE #6934: Save/restore Gradle's configuration cache on CI
## Expected Behavior
CC is saved & restored on CI, speeding up all jobs that don't modify build files.

## Current Behavior
Gradle must run the full configuration phase on every invocation.

## Context
CI builds should execute as quickly as possible.

Requires:
* Gradle 8.6 (which should be released any day now)
* Update config for `setup-gradle` action following these instructions: https://github.com/gradle/actions/blob/v3.0.0/setup-gradle/README.md#saving-configuration-cache-data
* The previous step involves creating a Gradle configuration cache key for CI that's stored as a GitHub secret

3flex:
Closes #6934

This uses a repo-specific key, so a new key would need to be generated for other repos in the detekt org if this feature was enabled anywhere else.

### DIFF
--- a/.github/workflows/codecoverage.yaml
+++ b/.github/workflows/codecoverage.yaml
@@ -29,6 +29,7 @@ jobs:
         uses: gradle/actions/setup-gradle@1168cd3d07c1876a65e1724114de42ccbdfa7b78 # v3
         with:
           gradle-home-cache-cleanup: true
+          cache-encryption-key: ${{ secrets.GRADLE_ENCRYPTION_KEY }}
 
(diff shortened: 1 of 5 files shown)


BraisGabin (approved):

cortinico:
> so a new key would need to be generated

@3flex QQ how have you generated the key?

3flex:
https://docs.gradle.org/8.7/userguide/configuration_cache.html#config_cache:secrets:generating_encryption_key

PR #7159, opened 2024-04-10
Remove unused functions from tooling API

3flex:
There don't appear to be any clients using this tooling API entry point.

Also added a note about other known clients for other known tooling API entry points - I've had a quick look around for other third party clients to see if any others are using them, but they're difficult to track down. This can be restored though if needed.


cortinico (approved):

cortinico:
@3flex should we mark this as breaking change?

BraisGabin (approved):

BraisGabin (approved):

3flex:
> should we mark this as breaking change?

Maybe to be safe - if any clients use it, it will definitely break!

PR #7174, opened 2024-04-12
Switch to new coordinates for wrapper-validation action

3flex:
https://github.com/gradle/actions/releases/tag/v3.3.0

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -17,4 +17,4 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
 
       - name: Validate Gradle Wrapper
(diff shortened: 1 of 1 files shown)


BraisGabin:
Now we can do this in `setup-gradle`. Do we want to continue to have an extra job for it?

Pro: reduce the number of jobs that we have for ci.
Con: worst parallelism. But it should be negligible.

I think that the number of ci checks that we have are a bit intimidating, so reduce them sound like a good idea to me. This is not a strong opinion.

I'm removing the auto merge to get an answer but I also approve the pr. So if you think that it is better as it is out that we should talk about this somewhere else you can merge it.

BraisGabin (approved):

cortinico (approved):

3flex:
I think it's ok as it is but I don't feel strongly either way.

Eventually setup-gradle will make checking the wrapper the default so when that happens we can just drop the extra check.

PR #7175, opened 2024-04-12
Move generateBindingContext to detekt-parser

3flex:
To me it makes more sense to have functions in detekt-parser that either create KtFiles from paths or directly from a source snippet, or generate the BindingContext for full analysis - or to look at it another way, anywhere we're delegating to the Kotlin compiler to do something for us.

Once we have either a list of KtFiles (for basic analysis) and optionally BindingContext (for full analysis) detekt-core can take of things from then on.

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
@@ -1,5 +1,6 @@
 package io.gitlab.arturbosch.detekt.core.tooling
 
+import io.github.detekt.parser.generateBindingContext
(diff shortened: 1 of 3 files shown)


BraisGabin:
Completely agree with your ideas here.

But I have a question related with this code and #7165. Why do we have two different codes to generate the `ContextBinding` for production and for tests? Should we remove the one for tests? Or, at least, make one use the other?

I imagine that maybe to reduce duplication on the tests we want to change the parameters. But we should want to generate the `ContextBinding` in the same way as production.

BraisGabin (approved):

cortinico (approved):

3flex:
I can't recall why there are different ones, but we do setup the compiler environment differently and then call different compiler methods to create the BindingContext.

They should be as aligned as possible but it requires a bit more research.

PR #7177, opened 2024-04-13
Remove BindingContext from FileProcessListener

3flex:
This is unused in detekt itself and there wouldn't be any use cases where it makes sense to make this available in this API.


BraisGabin:
Does this change help us to simplify something? To be honest I can't imagine why someone would use the ContextBinding here but it doesn't hurt us to provide it, right? I mean, we have it nearly for free, right?

3flex:
> Does this change help us to simplify something?

The API :)

There's no reason at all to keep it there. Sure it's "free" but 2.0.0 should have a rational API and there's no reason to pass it through.

BraisGabin (approved):

BraisGabin:
I think that, in general, we need to reevaluate our extension points. We have that we have too much and they could be simplified.

PR #7206, opened 2024-04-19
Make detekt-formatting autocorrect work with read-only KtFile

3flex:
Breaking change - instead of changing the KtFile contents directly, a rule implementing autocorrect must create a copy of the file text (or the already modified text, in case another rule has applied autocorrect already, so those changes are not overwritten), modify it then save the updated text to `KtFile.modifiedText`.

They will also need to take a copy of the original file path, since the copy created by `createPhysicalFile` only allows saving the file name but not the full path.

This is the only way to make autocorrect work with a read only KtFile - by reading and storing separate copies of the file's text.

Why do this? Because it allows for some simplification and refactoring of the detekt core and easier handling of file paths (for the majority of rules that don't implement autocorrect).

This has been tested against the new & updated tests in #7204.


detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
```diff
@@ -35,14 +39,20 @@ abstract class FormattingRule(config: Config, description: String) : Rule(config
...
+        this.root = fileCopy
+        originalPath = root.absolutePath()
+        positionByOffset = KtLintLineColCalculator.calculateLineColByOffset(fileCopy.text)
 
         wrapping.beforeFirstNode(computeEditorConfigProperties())
-        root.node.visitASTNodes()
+        this.root.node.visitASTNodes()
         wrapping.afterLastNode()
+
+        root.modifiedText = this.root.text
```
BraisGabin, line 55:
```suggestion
        if (this.root.modificationStamp > 0) {
            root.modifiedText = this.root.text
        }
```

Otherwise we will report that all the files were auto-formated.

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
```diff
@@ -35,14 +39,20 @@ abstract class FormattingRule(config: Config, description: String) : Rule(config
...
     private lateinit var root: KtFile
+    private lateinit var originalPath: Path
 
     override fun visit(root: KtFile) {
-        this.root = root
-        positionByOffset = KtLintLineColCalculator.calculateLineColByOffset(root.text)
+        val fileCopy = KtPsiFactory(root.project).createPhysicalFile(root.name, root.modifiedText ?: root.text)
+
+        this.root = fileCopy
+        originalPath = root.absolutePath()
```
BraisGabin, line 48: Shouldn't you store `root.toFilePath()`? I know that this PR aims to remove that in a near future but for now I think it's better to don't break the current behavior.

BraisGabin (commented):
This would also help with #2560. Now we will have the initial state of the file and the after so create a `patch` should be really easy.

The `patch` should help to implement autocorrect for the compiler plugin.

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt
3flex: I've made the update here as suggested but I'll add the test in #7204

BraisGabin (approved):

schalkms:
> This would also help with #2560. Now we will have the initial state of the file and the after so create a `patch` should be really easy.
> 
> The `patch` should help to implement autocorrect for the compiler plugin.

I had the same thought.

schalkms (approved):

PR #7235, opened 2024-04-27
Remove FilePath

3flex:


detekt-report-md/src/main/kotlin/io/github/detekt/report/md/MdOutputReport.kt
```diff
@@ -145,8 +145,8 @@ private fun MarkdownContent.renderIssues(issues: List<Issue>, basePath: Path?) {
 }
 
 private fun MarkdownContent.renderIssue(issue: Issue, basePath: Path?): String {
-    val filePath = basePath?.let { issue.location.filePath.absolutePath.relativeTo(it) }
-        ?: issue.location.filePath.absolutePath
+    val filePath = basePath?.let { issue.location.path.absolute().relativeTo(it) }
+        ?: issue.location.path.absolute()
```
BraisGabin, line 149: I don't like too much these `absolute()` everywhere. Shouldn't we ensure that `issue.location.path` is already absolute? Or the other way around, ensure that it is relative but to get the absolute we should use `basePath.resolve(issue.location.path)` instead of `issue.location.path.absolute()`. Because I think that we are assuming that both thinks are the same but probably they are not.

I was thinking about this and, maybe, we could change the value in `System.getProperty("user.dir")` to have the same value as `basePath` so we can use `absolute()` freely.

detekt-report-sarif/src/test/kotlin/io/github/detekt/report/sarif/SarifOutputReportSpec.kt
```diff
@@ -52,7 +52,7 @@ class SarifOutputReportSpec {
             .render(result)
 
         val expectedReport = readResourceContent("vanilla.sarif.json")
-            .replace("<PREFIX>", Path("/").absolute().toUri().toString())
+            .replace("<PREFIX>", Path(System.getProperty("user.dir")).toUri().toString())
```
BraisGabin, line 55: Probably this change is needed for the issue that I was talking before.

detekt-report-md/src/main/kotlin/io/github/detekt/report/md/MdOutputReport.kt
3flex: I tried adding `require(path.isAbsolute)` to init in Location but it broke too many tests because many of them use in-memory PSI files which don't have a real file path. It's too much trouble to fix it up now so this is the lesser of two evils.

It should be fine to remove the explicit `abosolute()` calls though as they're theoretically redundant.

> we could change the value in System.getProperty("user.dir") to have the same value as basePath

I don't follow? Do you mean System.setProperty("user.dir")? I don't think we should change that property value since it has a specific definition.

BraisGabin: > I tried adding `require(path.isAbsolute)` to init in Location but it broke too many tests because many of them use in-memory PSI files which don't have a real file path. It's too much trouble to fix it up now so this is the lesser of two evils.

I know what you are talking about #7147 was the first part of cleaning the paths of our tests and it was challenging. Then I also tried to put that `require` that you talk about and it was even more difficult.

> I don't follow? Do you mean System.setProperty("user.dir")? I don't think we should change that property value since it has a specific definition.

Yes, I was talking about that. And I feel the same, changing that property feels really problematic.

> It should be fine to remove the explicit `abosolute()` calls though as they're theoretically redundant.

I think that's the second best solution (the best one would be to add the `require` but as you said, a lot of work, we could create an issue and add "help wanted").

3flex: Another thought - the location could store a reference to a PsiFile (or KtFile) instead of a physical file location. That way the path doesn't matter for storing the location and we don't need to care about in-memory files here.

It pushes the problem downstream though so it might not be the best way of handling it, but I do think it's more correct from `Location`'s perspective.

3flex: I've thought about it some more and I'm convinced that storing the PsiFile instead of a file path is the right thing to do. I'll update the PR.

3flex: I've made some changes and I'm happier with the result, and it should resolve your concerns.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
```diff
@@ -18,13 +18,22 @@ class Location(
     val source: SourceLocation,
     val endSource: SourceLocation,
     val text: TextLocation,
-    val filePath: FilePath,
+    val ktFile: KtFile,
```
BraisGabin, line 21: I'm not 100% sure about this. I like how it looks on the rules. But this is a property that no one reads.

I don't want to block this so what do you think about merging this and I'll give it a try again and see if the API can be improved even more. Also this breaks the idea of making the output of detekt serializable.

BraisGabin (approved):

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
3flex: Sounds good, serialization is something I hadn't considered at all but it could block that depending on the approach.

I could also revert to `Path` and remove the `absolute()` calls, as the PRs that stack on this one will ensure that the path is always absolute except in tests. This means a check on `isAbsolute()` is not required. I do still think `KtFile` is most correct here but the API will be simpler with `Path`. I'm happy to make this change if you agree.

BraisGabin: I agree to use pqth

BraisGabin:
LGTM. It's strange the usages of user.dir on tests but that's clearly not important and you have a lot of PRs waiting for this.

3flex:
> It's strange the usages of user.dir on tests

They weren't really needed, I've stripped them out. Thanks for your time on this!

PR #7237, opened 2024-04-28
Remove PsiFile.basePath

3flex:


BraisGabin (approved):

BraisGabin:
Probably we should forbid any usage of `absolute()` or similar functions called from our core or reporters.

Is not doable right now but it would prevent issues in the future.

PR #7238, opened 2024-04-28
Remove PsiFile.relativePath

3flex:


detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/CompileExtensions.kt
```diff
@@ -27,10 +27,10 @@ fun compileContentForTest(
     basePath: Path = Path("/").absolute(),
     path: Path = basePath.resolve("Test.kt"),
```
BraisGabin, line 28: `basePath` should be removed from here too, right?

BraisGabin (approved):
Such a great refactor 👏👏👏 (not only this PR, all this PR chain).

I'm approving to avoid blocking you but I think that my comment should be fixed. It should be really easy (github doesn't let me provide a suggestion 😭)

detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/CompileExtensions.kt
3flex: Good call - was an easy fix, done.

PR #7240, opened 2024-04-28
Remove PsiFile.lineSeparator

3flex:
Waiting on merge of:
* #7234 
* #7235
* #7237 
* #7238


detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/RunnerSpec.kt
```diff
@@ -286,5 +286,35 @@ class RunnerSpec {
                 """.trimIndent()
             )
         }
+
+        @Test
+        fun `keeps LF line endings after autocorrect`() {
```
BraisGabin, line 291: It's strange to test this on cli. I know that there are other tests similar to these here but it's strange. I also must say that I don't care that much about the autocorrect right now, we have more important concerns right now.

BraisGabin (approved):
I reviewed the new commits in this PR (the ones that are not in the previous ones) and it looks really good :)

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtFileModifier.kt
```diff
@@ -27,7 +26,9 @@ class KtFileModifier : FileProcessListener {
     private fun KtFile.unnormalizeContent(): String =
         StringUtilRt.convertLineSeparators(
             checkNotNull(modifiedText),
-            lineSeparator
+            checkNotNull(virtualFile.detectedLineSeparator) {
+                "Line separator was not automatically detected. This is unexpected."
+            }
```
BraisGabin, line 31: I'm not 100% sure about this. I think that all the `KtFile`s will say `\n` because we clean the line separator by using the function `convertLineSeparators` here:

https://github.com/detekt/detekt/blob/b01bc7a5fbc94708d7d12b3e75ca61ad8fcaa898/detekt-parser/src/main/kotlin/io/github/detekt/parser/KtCompiler.kt#L27-L36

c86160b775507dd0d4dbe6aa78aafabaccd36e6c stops that "cleaning" for one of the fuctions but the other one still uses it. I realised of this while reviewing this commit d36b3dcfdb77d2213974a38be65e4dee99c6e172.

I assume that it is ok to stop using `convertLineSeparators` but I guess that we should stop using it in this PR otherwise we are breaking something. And if the tests doesn't spot this we should first check our tests to be sure that we have proper coverage for this.

Or maybe I'm just wrong, that can be also true.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/runners/RunnerSpec.kt
3flex: Autocorrect involves multiple modules so having a few functional tests makes it easy to verify that things aren't broken even when the changes & unit tests look fine. #7256 & #7206 is an example. There are different approaches but I think these provide value.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtFileModifier.kt
3flex: > I think that all the KtFiles will say `\n`

When #7255 gets merged the createKtFile function is moved so it's used by tests only, so neither function will make copies of the file text or try detecting line separators at all. The tests you commented on also ensure that everything works, at least when calling via CLI, which is what the Gradle plugin uses. That covers all bases with autocorrect.

For now though, please check the updates to `createKtFile` in this PR. It now stores the detected line separator using `virtualFile.setDetectedLineSeparator`: https://github.com/detekt/detekt/pull/7240/files#diff-5dc598360678521200286ef06ac96f589ea689b19e935b5f8dd058dfada7850cR32 which could be `\r\n` or `\n` depending on platform.

PR #7250, opened 2024-05-03
Stop creating in-memory KtFile when file path is provided

3flex:
Waiting on merge of:
* #7235
* #7237 
* #7238
* #7240

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtFileModifier.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/KtFileModifier.kt
@@ -16,10 +16,12 @@ class KtFileModifier : FileProcessListener {
 
     override fun onFinish(files: List<KtFile>, result: Detektion) {
         files.filter { it.modifiedText != null }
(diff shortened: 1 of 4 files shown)


BraisGabin (approved):
Is the last commit related with this PR? Any way, LGTM.

3flex:
> Is the last commit related with this PR?

It is, yes. If it's not done in [AutoCorrectLevelSpec.kt](https://github.com/detekt/detekt/pull/7250/files#diff-4741276de125ebf719e19803d2d516ffb3053c70171f864c4624e00fcf7f17c6) then the tests will fail because the same PsiFile is being reused between tests, and the modified text persists between them. This wasn't a problem before because detekt would always create a new copy of every file it analysed, but that doesn't happen after this PR.

There may be scenarios in IDE or other usages where the same PsiFile is also reused so I figured it's safest to defend against that by making a similar change in [KtFileModifier.kt](https://github.com/detekt/detekt/pull/7250/files#diff-f99104f9f6be143fe6b02b3cda4540c8a06ae78e8f06a7d4f869f18f500b9083)

PR #7254, opened 2024-05-03
Remove PsiFile.absolutePath

3flex:
This is now fully handled by the IntelliJ components in the embeddable compiler.

### DIFF
--- a/detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/KtFiles.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/KtFiles.kt
@@ -4,6 +4,7 @@ import org.jetbrains.kotlin.com.intellij.openapi.util.TextRange
 import org.jetbrains.kotlin.com.intellij.psi.PsiFile
 import org.jetbrains.kotlin.diagnostics.DiagnosticUtils
 import org.jetbrains.kotlin.diagnostics.PsiDiagnosticUtils
+import org.jetbrains.kotlin.psi.KtFile
 import java.nio.file.Path
 import kotlin.io.path.Path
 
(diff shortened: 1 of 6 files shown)


detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/KtFiles.kt
```diff
@@ -26,6 +27,9 @@ fun PsiFile.fileNameWithoutSuffix(multiplatformTargetSuffixes: List<String> = em
 
 fun PsiFile.absolutePath(): Path = Path(virtualFile.path)
```
BraisGabin, line 28: We could even remove this one, right? I guess that there should be really few usages of this one. And, even if they are, it should be safe to cast them to `KtFile`. To keep the api as small as possible. This way we don't even need the comment :P

3flex: My idea for keeping it is that 3rd party rule authors will have an easier time finding the right function to call - requiring a cast makes them less discoverable, particularly when using something like KtElement/PsiElement.containingFile which returns PsiFile.

One improvement would be to change this line to 
```kotlin
fun PsiFile.absolutePath(): Path = (this as KtFile).absolutePath()
```

So that way it always queries the cached path. That would provide some additional value.

BraisGabin (approved):

detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/KtFiles.kt
BraisGabin: I agree with make it easier to find. Then maybe we can always cast it and just remove the new extension function. This is not really important, anyway.

PR #7251, opened 2024-05-03
Remove line separator detection logic

3flex:
This is now fully handled by the IntelliJ components in the embeddable compiler.

PR #7255, opened 2024-05-03
Remove ability to create AnalysisResult by passing source code and filename

3flex:
Calling detekt from production code should only ever pass a `KtFile` or a `Path` (from which a KtFile can be created).

Creating `KtFile` from text directly is only required from tests.

This is a breaking change for detekt-intellij-plugin, but a quick look at the source in that project suggests it will be a simple update.

Waiting on merge of:
* #7235
* #7237 
* #7238
* #7240
* #7250 
* #7251 
* #7254


BraisGabin (approved):
Ok, this have a lot of sense. While I was reviewing this PRs I was thinking "Why do we need this in our production code?". It was generating noise and it's great to move it so only the tests uses it.

PR #7259, opened 2024-05-05
Make `Entity.ktElement` no nullable

BraisGabin:
There is not a good reason to keep `Entity.ktElement` nullable. It just make the API messier and provide no value.

### DIFF
--- a/detekt-api/build.gradle.kts
+++ b/detekt-api/build.gradle.kts
@@ -11,6 +11,7 @@ dependencies {
 
     testImplementation(projects.detektTest)
     testImplementation(libs.assertj)
+    testFixturesImplementation(projects.detektTestUtils)
 }
 
 detekt {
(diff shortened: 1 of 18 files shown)


3flex (dismissed):
This might not be a good idea:

1. Not all rule violations relate to a KtElement, for example trailing whitespace (whitespace is represented by a PsiElement)
2. Rule violations like MaxLineLength can include multiple KtElements

I haven't looked into it, but do we actually need to include the KtElement in the Entity at all? A location, signature and location should be enough to report on issues.

BraisGabin:
Rebased to fix conflicts. A review here would be great. This is not a big blocker or anything else but it would be nice to merge it.

cortinico:
CI is really red here @BraisGabin

BraisGabin:
Fixed. The new rule from ktlint always bites be.

cortinico (approved):

PR #7260, opened 2024-05-06
Remove detekt-psi-utils usage from detekt-api

### LINKED ISSUE #7140: Break the dependency of `:detekt-api` with `:detekt-psi-utils`

3flex:
Closes #7140

### DIFF
--- a/detekt-api/build.gradle.kts
+++ b/detekt-api/build.gradle.kts
@@ -7,7 +7,6 @@ plugins {
 
 dependencies {
     api(libs.kotlin.compilerEmbeddable)
-    api(projects.detektPsiUtils)
     implementation(projects.detektUtils)
 
     testImplementation(projects.detektTest)
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Location.kt
@@ -1,7 +1,6 @@
 package io.gitlab.arturbosch.detekt.api
 
 import dev.drewhamilton.poko.Poko
-import io.github.detekt.psi.absolutePath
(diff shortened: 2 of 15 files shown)


BraisGabin:
Who should add `projects.detektPsiUtils` in the classpath now? The clients? Or should we add it as a runtime only dependency on the core?

3flex:
It's already an implementation dependency of detekt-core, so it's on the runtime classpath: https://github.com/detekt/detekt/blob/38651f82c76306a997887df88b22fc19bd3ea7ed/detekt-core/build.gradle.kts#L12

3flex:
> ci was failing exactly for that issue

If only it were that simple... it's failing because this function was catching an IndexOutOfBoundsException due to errors in the implementation of MaxLineLength:

https://github.com/detekt/detekt/blob/38651f82c76306a997887df88b22fc19bd3ea7ed/detekt-psi-utils/src/main/kotlin/io/github/detekt/psi/KtFiles.kt#L38-L41

The above code was catching the exception, and I didn't copy that exactly to the new location because it shouldn't be needed anymore.

The MaxLineLength implementation should be fixed, but it looks like it needs changes to Location and/or Entity... until that's been corrected I'd call this and #7261 blocked.

BraisGabin (approved):

PR #7263, opened 2024-05-06
Run same rule multiple times with different configuration

### LINKED ISSUE #6738: Add the option to instantiate the same rule more than once

BraisGabin:
With this PR we finally are able to run the same run multiple times with different configurations.

The most important part to discuss, I think, is if we like how those new instances are configured.

Right now if you want to create a new instance of the same rule you should add on your `detekt.yml` something like this:

```yaml
styles:
  ForbiddenImport/compose:
    active: true
    ...
```

Close #6738

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
@@ -97,15 +97,23 @@ internal class Analyzer(
             .flatMap { (ruleSet, ruleSetConfig) ->
                 ruleSetConfig.subConfigKeys()
                     .asSequence()
-                    .mapNotNull { runCatching { Rule.Name(it) }.getOrNull() }
-                    .mapNotNull { ruleName ->
-                        ruleSet.rules[ruleName]?.let { it to ruleSetConfig.subConfig(ruleName.value) }
+                    .mapNotNull { ruleId ->
+                        extractRuleName(ruleId)?.let { ruleName -> ruleId to ruleName }
                     }
-                    .filter { (_, config) -> config.isActiveOrDefault(false) }
-                    .filter { (_, config) -> config.shouldAnalyzeFile(file, settings.spec.projectSpec.basePath) }
-                    .map { (ruleProvider, config) ->
+                    .mapNotNull { (ruleId, ruleName) ->
+                        ruleSet.rules[ruleName]?.let { ruleProvider ->
+                            RuleDescriptor(
+                                ruleProvider = ruleProvider,
+                                config = ruleSetConfig.subConfig(ruleId),
+                                ruleId = ruleId,
+                            )
+                        }
+                    }
+                    .filter { (_, config, _) -> config.isActiveOrDefault(false) }
+                    .filter { (_, config, _) -> config.shouldAnalyzeFile(file, settings.spec.projectSpec.basePath) }
+                    .map { (ruleProvider, config, ruleId) ->
                         val rule = ruleProvider(config)
-                        rule.toRuleInstance(rule.ruleName.value, ruleSet.id) to rule
+                        rule.toRuleInstance(ruleId, ruleSet.id) to rule
                     }
             }
             .filterNot { (ruleInstance, rule) ->
@@ -144,6 +152,11 @@ internal class Analyzer(
             }
     }
 }
+
+internal fun extractRuleName(key: String): Rule.Name? =
+    runCatching { Rule.Name(key.split("/", limit = 2).first()) }.getOrNull()
+
+private data class RuleDescriptor(val ruleProvider: (Config) -> Rule, val config: Config, val ruleId: String)
 
 private fun List<Finding>.filterSuppressedFindings(rule: Rule, bindingContext: BindingContext): List<Finding> {
     val suppressors = buildSuppressors(rule, bindingContext)
--- a/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/AnalyzerSpec.kt
+++ b/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/AnalyzerSpec.kt
@@ -136,6 +136,30 @@ class AnalyzerSpec(val env: KotlinCoreEnvironment) {
         }
 
         @Test
+        fun `with multiple instances of same rule`() {
+            val testFile = path.resolve("Test.kt")
+            val output = StringPrintStream()
+            val settings = createProcessingSettings(
+                testFile,
+                yamlConfigFromContent(
+                    """
+                        custom:
+                          MaxLineLength/foo:
+                            active: true
+                            maxLineLength: 30
+                          MaxLineLength/bar:
+                            active: true
+                            maxLineLength: 30
+                    """.trimIndent()
+                ),
+                outputChannel = output,
+            )
+            val analyzer = Analyzer(settings, listOf(CustomRuleSetProvider()), emptyList())
+
+            assertThat(settings.use { analyzer.run(listOf(compileForTest(testFile))) }).hasSize(2)
+        }
+
+        @Test
         fun `with findings and context binding`() {
             val testFile = path.resolve("Test.kt")
             val output = StringPrintStream()
--- a/detekt-core/src/test/resources/config_validation/rule-id.yml
+++ b/detekt-core/src/test/resources/config_validation/rule-id.yml
@@ -1,0 +1,8 @@
+style:
+  active: true
+  WildcardImport:
+    active: true
+  WildcardImport/extra:
+    active: true
+  WildcardImport/extra2:
+    active: true
--- a/detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/spec/RulesSpec.kt
+++ b/detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/spec/RulesSpec.kt
@@ -67,6 +67,6 @@ interface RulesSpec {
         /**
          * Run a single rule.
          */
-        class RestrictToSingleRule(val ruleSetId: RuleSet.Id, val ruleName: Rule.Name) : RunPolicy()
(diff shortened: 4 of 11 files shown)


cortinico:
> And it produces this output:

The output has rule ids like `[ForbiddenImport/AtTestCode]` while in your config you have `ForbiddenImport/2` and `/3`. I assume that's a mistake

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -27,6 +27,19 @@ import org.jetbrains.kotlin.resolve.BindingContext
...
 
+private data class RuleDescriptor(val ruleInstanceId: String, val ruleName: Rule.Name) {
+    companion object {
+        fun fromKey(key: String): RuleDescriptor {
+            val split = key.split("/", limit = 2)
+            return if (split.size == 1) {
+                RuleDescriptor(key, Rule.Name(key))
+            } else {
+                RuleDescriptor(key, Rule.Name(split[0]))
+            }
```
cortinico, line 38:
```suggestion
            return RuleDescriptor(key, Rule.Name(split.first()))
```

config/detekt/detekt.yml
```diff
@@ -1,3 +1,6 @@
+config:
+  validation: false
```
cortinico, line 2: Why this?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
```diff
@@ -101,19 +98,13 @@ open class Rule(
 
     /**
      * Reports a single code smell finding.
-     *
-     * Before adding a finding, it is checked if it is not suppressed
-     * by @Suppress or @SuppressWarnings annotations.
      */
     fun report(finding: Finding) {
-        val ktElement = finding.entity.ktElement
-        if (ktElement == null || !ktElement.isSuppressedBy(ruleId, aliases, ruleSetId)) {
```
cortinico, line 110: Q: what happened to this?

cortinico (commented):
I really like this :D

BraisGabin:
> > And it produces this output:
> 
> The output has rule ids like `[ForbiddenImport/AtTestCode]` while in your config you have `ForbiddenImport/2` and `/3`. I assume that's a mistake

Indeed. Copy&paste error. I fixed it now.

config/detekt/detekt.yml
BraisGabin: The validation code doesn't know about this type of config keys yet. I need to implement that. I disabled it just to make it work for now. As I said this is still not ready to merge. I just wanted to share it to show that this is near to be done.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
BraisGabin: #7101

BraisGabin:
@detekt/maintainers This PR is ready to review and I think that this one should get multiple eyes on it.

cortinico (approved):
LGTM 👍 We need docs for this also :)

schalkms (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -145,6 +154,11 @@ internal class Analyzer(
     }
 }
 
+internal fun extractRuleName(key: String): Rule.Name? =
+    runCatching { Rule.Name(key.split("/", limit = 2).first()) }.getOrNull()
```
TWiStErRob, line 158: Is there a way to get better DX here? In what cases can this throw and return null? If I read it correctly it'll simply ignore the rule. Should we fail when config is incorrect rather than let the user scratch their head why things are "not working". I guess the Rule.Name validation was there before, but now the runCatching can fail due to 2 reasons. Consider to get back "original" behavior:

```suggestion
    val name = key.split("/", limit = 2).first()
    return runCatching { Rule.Name(name) }.getOrNull()
```

detekt-core/src/test/resources/config_validation/rule-id.yml
```diff
@@ -0,0 +1,8 @@
+style:
+  active: true
+  WildcardImport:
+    active: true
+  WildcardImport/extra:
```
TWiStErRob, line 5: (Probably the wrong place to flag it, but wanted to start a thread.)

Is this syntax suppress-compatible, i.e. can I suppress a specific instance of a rule? Can IDEA handle it?

Will this work and only suppress the `extra` instance?
```
@file:Suppress("WildcardImport/extra")
```

Will this work and suppress all the instances?
```
@file:Suppress("WildcardImport")
```

How can I suppress only the default instance, and not the extras?
```
@file:Suppress("WildcardImport/default")
```

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/spec/RulesSpec.kt
```diff
@@ -62,6 +62,6 @@ interface RulesSpec {
         /**
          * Run a single rule.
          */
-        class RestrictToSingleRule(val ruleSetId: RuleSet.Id, val ruleName: Rule.Name) : RunPolicy()
+        class RestrictToSingleRule(val ruleSetId: RuleSet.Id, val ruleId: String) : RunPolicy()
```
TWiStErRob, line 65: Could this be a value-class that is pre-split, or at least hides the complexity of `split("/", 2).first()` in a `val`?
```suggestion
        class RestrictToSingleRule(val ruleSetId: RuleSet.Id, val ruleId: Rule.Id) : RunPolicy()
```

detekt-core/src/test/resources/config_validation/rule-id.yml
BraisGabin: `@file:Suppress("WildcardImport")` This only suppress the default one (if it exists)

BraisGabin: Yes. It you can give them the same alias to those 3 and use that.

I don't see a real scenario where suppressing all the rules of the same name is a good idea. But if someone end up with an issue like that we could implement wildcards `RuleName/*` or something like that. But I prefer not to do that until a real use case appears.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
atulgpt: Are we not allowing `/` in original rule id? Also instead of `runCatching { Rule.Name(key.split("/", limit = 2).first()) }.getOrNull()` can we do `Rule.Name(key.split("/", limit = 2).firstOrNull()`

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/AnalyzerSpec.kt
```diff
@@ -135,6 +135,30 @@ class AnalyzerSpec(val env: KotlinCoreEnvironment) {
...
+        @Test
+        fun `with multiple instances of same rule`() {
+            val testFile = path.resolve("Test.kt")
+            val output = StringPrintStream()
+            val settings = createProcessingSettings(
+                testFile,
+                yamlConfigFromContent(
+                    """
+                        custom:
+                          MaxLineLength/foo:
```
atulgpt, line 147: one TC should be there with `MaxLineLength/foo/bar` and one with `MaxLineLength/`(no value after `/`) I think latter should be error in validation

BraisGabin: I refactored a bit the code related with this on #7468. Now I have unit test for those cases in the creation of `Rule.Id`. So we don't need to ad even more tests to `AnalyzerSpec`. But really good call, thanks!

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/spec/RulesSpec.kt
BraisGabin: I did these changes at #7468

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
BraisGabin: Same, I did these changes at #7468

PR #7269, opened 2024-05-07
Remove hasCommentInside functions

3flex:
These functions are not general purpose (only used in a couple of rules that check for empty blocks) so can be removed from the API.

Technically breaking change but this is very unlikely to be used in 3rd party rules.


BraisGabin (approved):
What a surprise that all these functions were inside a file called junk 🤣

PR #7270, opened 2024-05-08
Rewrite MaxLineLength & TrailingWhitespace to avoid problematic offset calcs

3flex:


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLength.kt
```diff
@@ -48,26 +52,26 @@ class MaxLineLength(config: Config) : Rule(
...
-        for (line in lines) {
-            offset += line.length
-            if (!isValidLine(file, offset, line)) {
+        val sourceFileLinesMapping = KtPsiSourceFileLinesMapping(file)
+
+        file.text.lines().withIndex()
+            .filterNot { isValidLine(file, sourceFileLinesMapping.getLineStartOffset(it.index), it.value) }
+            .forEach {
+                val offset = sourceFileLinesMapping.getLineStartOffset(it.index)
+                val line = it.value
```
BraisGabin, line 62:
```suggestion
            .map { (index, line) -> sourceFileLinesMapping.getLineStartOffset(index) to line }
            .filterNot { (offset, line) -> isValidLine(file, offset, line) }
            .forEach { (offset, line) ->
```

BraisGabin (approved):

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/Junk.kt
```diff
@@ -12,7 +12,7 @@ import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
  * the given [line] from a given offset in a [KtFile].
  */
 internal fun findKtElementInParents(file: KtFile, offset: Int, line: String): Sequence<PsiElement> {
-    return file.elementsInRange(TextRange.create(offset - line.length, offset))
+    return file.elementsInRange(TextRange.create(offset, offset + line.length))
```
BraisGabin, line 15: If we can do this change in this function without breaking any other rule, shouldn't we move this function to `MaxLineLength`? And, otherwise, how is it that we don't break any other rule if they use it?

3flex: Yeah this is one I need to look at - TrailingWhiteSpace is the other one that uses it, and it doesn't cause any issues (yet) because of the `runCatching` block I've mentioned in other comments that masks the problem.

That rule needs to be rewritten too which I'm working on now and will include in this PR.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/MaxLineLength.kt
3flex: It cuts out a line, but I think it's more readable as is, so I'd like to keep the current implementation.

BraisGabin: Can we at least use deconstructed variables (`(inxed, line) ->`)? so instead of `it.index` we have `index`. I don't care that much about the `map`. It was just to avoid the duplicated call to `sourceFileLinesMapping.getLineStartOffset`.

BraisGabin (approved):
LGTM

This PR is really related with #7259. Because these two rules are the more problematic ones related with the null `KtElement`.

PR #7273, opened 2024-05-08
Faster `InvalidPackageDeclaration`

BraisGabin:
We know that visit a `KtFile` is not free at all. But we don't have any values to see if this is really faster or not.

My next project on detekt is going to be a tool to monitor the detekt performance. Once we have that we could take a look at the `Rule` implementation because I think that it pushes to overuse the `visit` pattern everywhere even in cases that is not needed (for example the one of this PR)

I left this PR as draft because I want to know how much faster is `InvalidPackageDeclaration` with this change.


3flex:
I'm a bit unclear on the purpose of this PR? Would it be better to look into tooling to monitor performance first, then PRs can be opened once there's some data available?

If this PR is closed an issue could be opened instead, linking to this PR.

BraisGabin:
We can just close this. I don't think it's worth the issue. We first need to work on monitoring as you said.

PR #7298, opened 2024-05-15
Add `.vscode/settings.json` to `.gitignore`

baseballyama:
This is a small PR, but I would like to add a VSCode setting when modifying documentation, but I don't want to commit this.
Therefore I updated .gitignore.

(Basically, I work with IntelliJ IDEA, but VSCode is sometimes more convenient for document editing.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -164,3 +164,6 @@ target/
 
 # Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
 !gradle-wrapper.jar
+
+# VSCode
+.vscode/settings.json


.gitignore
```diff
@@ -164,3 +164,6 @@ target/
 
 # Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
 !gradle-wrapper.jar
+
+# VSCode
+.vscode/settings.json
```
Goooler, line 169:
```suggestion
.vscode
```

I believe we can ignore the whole `.vscode` path for now.

Goooler (approved):

3flex (approved):

PR #7299, opened 2024-05-17
New rule set to detekt marketplace.js

Kiolk:
This pull request contain information about new rule set that I want to add to marketplace page. General information about set you can find in GitHub repository. In current moment it contains only one rule, I will add more in the future.

### DIFF
--- a/website/src/data/marketplace.js
+++ b/website/src/data/marketplace.js
@@ -230,8 +230,8 @@ export const extensions = [
     usesTypeResolution: true,
   },
   {
-    title: "Kiolk's Detekt Rules",
-    description: "This set contains rule that helps to improve readability and to keep single convention in the whole project.",
+    title: "UseInvokeForOperator",
+    description: "This set contains rules that help to improve readability and to keep a single project-wide convention. UseInvokeForOperater rule detects cases when using invoke expression instead of a direct call with round brackets. The Invoke operator makes code more readable and the rule helps keeping it.",
     repo: "https://github.com/Kiolk/Detekt-rules",
     ruleset: "kiolk-detekt-rules",
     rules: [
--- a/website/src/data/marketplace.js
+++ b/website/src/data/marketplace.js
@@ -230,6 +230,17 @@ export const extensions = [
     usesTypeResolution: true,
   },
   {
+    title: "Kiolk's Detekt Rules",
+    description: "This set contains rule that helps to improve readability and to keep single convention in the whole project.",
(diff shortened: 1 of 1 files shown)


website/src/data/marketplace.js
```diff
@@ -229,6 +229,17 @@ export const extensions = [
     ],
     usesTypeResolution: true,
   },
+  {
+    title: "Kiolk's Detekt Rules",
+    description: "This set contains rule that helps to improve readability and to keep single convention in the whole project.",
```
schalkms, line 234:
```suggestion
    description: "This set contains rules that help to improve readability and to keep a single project-wide convention.",
```

schalkms (commented):
Thanks for adding your rule set to the marketplace!

website/src/data/marketplace.js
```diff
@@ -229,6 +229,17 @@ export const extensions = [
     ],
     usesTypeResolution: true,
   },
+  {
+    title: "Kiolk's Detekt Rules",
```
cortinico, line 233: Can we call this `UseInvokeForOperator` instead?

website/src/data/marketplace.js
cortinico: Also this description is a bit too generic. Can we describe what your rule is doing instead here?

website/src/data/marketplace.js
Kiolk: Changed

website/src/data/marketplace.js
Kiolk: Updated and added addition description.

website/src/data/marketplace.js
```diff
@@ -229,6 +229,17 @@ export const extensions = [
     ],
     usesTypeResolution: true,
   },
+  {
+    title: "UseInvokeForOperator",
+    description: "This set contains rules that help to improve readability and to keep a single project-wide convention. UseInvokeForOperater detects cases when using invoke expression instead direct call with round brackets. Invoke operator makes code more readable and rule helps keeping it.",
```
schalkms, line 234:
```suggestion
    description: "This set contains rules that help to improve readability and to keep a single project-wide convention. UseInvokeForOperater rule detects cases when using invoke expression instead of a direct call with round brackets. The Invoke operator makes code more readable and the rule helps keeping it.",
```

Kiolk: Updated.

cortinico (approved):

PR #7310, opened 2024-05-19
Remove IT_LITERAL & LET_LITERAL

3flex:
LET_LITERAL was unused.

IT_LITERAL is provided by `StandardNames.IMPLICIT_LAMBDA_PARAMETER_NAME` which has resulted in more descriptive code IMHO. There was one usage which I've replaced with string `"it"` because the rule is literally checking for "\`it`-started expressions" so I think this is reasonable.

### DIFF
--- a/detekt-psi-utils/api/detekt-psi-utils.api
+++ b/detekt-psi-utils/api/detekt-psi-utils.api
@@ -46,7 +46,6 @@ public final class io/gitlab/arturbosch/detekt/rules/JunkKt {
 
 public final class io/gitlab/arturbosch/detekt/rules/KeywordsKt {
 	public static final field IT_LITERAL Ljava/lang/String;
-	public static final field LET_LITERAL Ljava/lang/String;
 }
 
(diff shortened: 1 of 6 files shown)


BraisGabin (approved):
Approved because this is clearly better than before but I would use `"it"` directly. At least for me `IMPLICIT_LAMBDA_PARAMETER_NAME` is more complex than `"it"`.

3flex:
I think it's better to use type safe values, and using a `Name` instead of a plain string assists with that. `IMPLICIT_LAMBDA_PARAMETER_NAME` is awkward but it's descriptive. Maybe just my preference but I think we should promote type safety in our rule implementations.

BraisGabin:
Oh, I thought that it was just a plain string too. I'm still not 100% on board but it's good for me. This PR reduces complexity on our public api and that's the important point.

cortinico (approved):

cortinico:
> At least for me `IMPLICIT_LAMBDA_PARAMETER_NAME` is more complex than `"it"`.

If so, we could have our own internal field that maps to `IMPLICIT_LAMBDA_PARAMETER_NAME`

PR #7311, opened 2024-05-19
Remove KtExpression.isUnitExpression()

3flex:
This function is a bit overengineered and it's simple to refactor to remove its usage.


BraisGabin (approved):
`StandardNames.FqNames.unit.shortName().asString()` was so overkill for just `"Unit"`! 🤣

cortinico (approved):

schalkms (approved):
Agreed

PR #7312, opened 2024-05-19
Remove aliases from `Rule` api

BraisGabin:
The aliases are a really cool feature from `detekt`. But we can put that information in the configuration instead of the actual code of the rule.

This PR removes `aliases` and `getDefaultRuleIdAliases` from `Rule` in favor of the annotation `@Alias`. Also this fixes an issue in our documentation. Some rules were not displaying their aliases, for example: `FunctionNameMaxLength`:

Before: https://detekt.dev/docs/rules/naming#functionmaxlength
After: https://detekt-4art7d1zf-detekt.vercel.app/docs/next/rules/naming#functionnamemaxlength

This new implementation fix that and should prevent future errors.


cortinico (approved):
I like this one! 👍 

Perhaps the only thing to add is some documentation to the website as not `aliases` is a configurable key that users can override via the config file.

BraisGabin:
> Perhaps the only thing to add is some documentation to the website as not `aliases` is a configurable key that users can override via the config file.

Good call! I added the documentation, what do you think?

cortinico (approved):
LGTM 🚀

PR #7313, opened 2024-05-19
Add new id for the rules

BraisGabin:
This PR changes some names. The first one is that `Rule` has a `Name` instead of an `Id`. A rule can no longer be identified by a static name.

The second one is to introduce a new `id` that can be different from `Rule.Name`.

And because now `RuleInfo` had a `name` and an `id` I renamed it to `RuleInstance`.

Naming is REALLY difficult so any help here is more than welcome.

This PR is part of #7263 but because that PR contains a lot of decissions I prefer to split it so we can talk about each one in different PRs but I recomment to read #7263 so you get the full context about why this change is needed.


cortinico (approved):
LGTM 🚢

PR #7324, opened 2024-05-20
Rename `UnusedPrivateMember` to `UnusedPrivateFunction`

BraisGabin:
We were calling `UnusedPrivateMember` to a rule that only checked functions. So I renamed it to make more clear what it does. Also, it was called like this because time ago this rule handled lots of cases so for that reason I also reduced the number of `aliases`.

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -273,9 +273,6 @@ style:
     ignoreAnnotated: ['Nested']
   UnusedImport:
     active: false # formatting already have this rule enabled
-  UnusedPrivateMember:
-    active: true
-    allowedNames: '(_|ignored|expected)'
   UseAnyOrNoneInsteadOfFind:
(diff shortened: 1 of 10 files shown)


config/detekt/detekt.yml
```diff
@@ -273,9 +273,6 @@ style:
     ignoreAnnotated: ['Nested']
   UnusedImport:
     active: false # formatting already have this rule enabled
-  UnusedPrivateMember:
-    active: true
-    allowedNames: '(_|ignored|expected)'
```
cortinico, line 278: What happened to this one?

BraisGabin: We don't need to enable this rule on our own configuration, it is enabled by default. And, allowedNames doesn't have sense any longer on this rule.

cortinico (approved):

PR #7331, opened 2024-05-24
Fix parameter name of NamedArguments rule

baseballyama:
According to the document, we can take 2 config options for `NamedArguments` rule. `threshold` and `ignoreArgumentsMatchingNames`.
But according to actual implementation, config option names are `allowedArguments` and `ignoreArgumentsMatchingNames`.

I think `allowedArguments` should be `threshold`.

https://detekt.dev/docs/rules/complexity/#namedarguments

I will fix CI but can you check this is correct fix or not?.

### DIFF
--- a/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArguments.kt
+++ b/detekt-rules-complexity/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/NamedArguments.kt
@@ -36,16 +36,16 @@ class NamedArguments(config: Config) : Rule(
 ) {
 
     @Configuration("The allowed number of arguments for a function.")
-    private val allowedArguments: Int by config(defaultValue = 3)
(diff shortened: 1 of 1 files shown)


BraisGabin:
You are comparing the versioned documentation against the actual code and sure they are different. You should check the "next" version on the documentation to get the matching one.

We are moving away from threshold as a parameter name to names more descriptive. If you think that the actual one is not good please open an issue explaining it and proposing alternatives. We really like this kind of rules and right now is the time for them because we are cooking 2.0 so any breaking change is allowed.

baseballyama:
Thank you for your check. and I understood about the current status.

`allowedArguments` sounds like array of argument name, so `threshold` is easier to understand for me. But anyway I can read the docs and I can understand what is this, so doesn't matter.

This is out of the PR and I may have missed some information, but do you have an estimated shipping date for 2.0?

BraisGabin:
allowedArgumentCount would be better? No eta for 2.0

baseballyama:
Yeah `allowedArgumentCount` is more explicit.

PR #7352, opened 2024-06-07
Add missing config options to formatting rules

3flex:

### DIFF
--- a/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/FunctionName.kt
+++ b/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/FunctionName.kt
@@ -1,7 +1,11 @@
 package io.gitlab.arturbosch.detekt.formatting.wrappers
 
+import com.pinterest.ktlint.rule.engine.core.api.editorconfig.EditorConfigProperty
 import com.pinterest.ktlint.ruleset.standard.rules.FunctionNamingRule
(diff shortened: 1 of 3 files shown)


detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/FunctionName.kt
```diff
@@ -13,4 +17,12 @@ class FunctionName(config: Config) : FormattingRule(
     "Function name should start with a lowercase letter (except factory methods) and use camel case."
 ) {
     override val wrapping = FunctionNamingRule()
+
+    @Configuration("ignore functions annotated with")
+    private val ignoreWhenAnnotatedWith by config("")
```
BraisGabin, line 22: This one is the same as ignoreAnotated that we support by default on all the rules. Does it have sense to give this configuration? Or should we pass an empty string to the ktlint rule and "force" our users to use `ignoreAnotated`

BraisGabin (approved):
I disabled the auto merge just to know your opinion on this. But I'm not against merging this as it is either.

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/wrappers/FunctionName.kt
3flex: Oh good point - yes I think that makes sense.

BraisGabin (approved):
LGTM

PR #7357, opened 2024-06-10
Enable FunctionExpressionBody on detekt codebase

3flex:
I tried the autocorrect but it seemed to make several errors. It did fix most violations but in some cases I had to manually replace the autocorrect suggestions from ktlint. I didn't dig in any further to troubleshoot the ktlint side of things.

This rule is enabled by default in ktlint 1.3.0 and it's one I wanted to see enforced in detekt for a while. I've explicitly enabled it here but I have another PR to enable rules that migrated to ktlint's standard rule set by default in detekt. This PR is big enough that it needs to be done separately.

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -78,6 +78,8 @@ formatting:
     active: true
   Filename:
     active: false
+  FunctionExpressionBody:
+    active: true
   FunctionSignature:
     active: false
   Kdoc:
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
@@ -16,11 +16,10 @@ open class CodeSmell(
         require(message.isNotBlank()) { "The message should not be empty" }
     }
 
-    override fun toString(): String {
-        return "CodeSmell(entity=$entity, " +
+    override fun toString(): String =
+        "CodeSmell(entity=$entity, " +
             "message=$message, " +
             "references=$references)"
-    }
 }
 
 /**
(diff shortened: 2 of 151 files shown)


BraisGabin:
I would vote to not impose this rule on detekt. I agree that in some places the function expression is easier to read but to me that's not true for all the cases. And it also makes the code a bit harder to extend. Let me explain that latest sentence.

We have a rule that ensures that for any `if` that spans more than one line it should contain `{}`. The reason behind it, or at least to me, is that if you want to add something inside that `if` you will not need to add those `{}` nor remove them if you end up removing code and leave the `if` with one single statement. This rule in my opinion pushes exactly in the other direction. If you need to add something to a function then it force you to add the `{}` and the return. And if you remove then it force you to remove those.

So, for that reason, I think that we are better leaving this rule disabled here. I'm also more than happy to read reason about why this is a good practice in general.

3flex:
> I'm also more than happy to read reason about why this is a good practice in general.

Simply that this is the recommendation in the official Kotlin coding conventions: 

 > Prefer using an expression body for functions with the body consisting of a single expression.

https://kotlinlang.org/docs/coding-conventions.html#functions

detekt should align unless there's a good reason not to.

BraisGabin:
> detekt should align unless there's a good reason not to.

For me that's the reason to don't follow it:

> We have a rule that ensures that for any `if` that spans more than one line it should contain `{}`. The reason behind it, or at least to me, is that if you want to add something inside that `if` you will not need to add those `{}` nor remove them if you end up removing code and leave the `if` with one single statement. This rule in my opinion pushes exactly in the other direction. If you need to add something to a function then it force you to add the `{}` and the return. And if you remove then it force you to remove those.

3flex:
That if rule is for clarity when reading multi line statements, no? Not to help make future changes simpler. And you can easily argue the opposite, why keep extra elements in the code around? Code is far more often read than updated so it's better to make it easier to read and remove needless distractions where possible than make it easier to extend in future.

I think it's odd that we would ignore the coding conventions on this - you might not agree with them but following the official coding conventions follows the principle of least surprise.

cortinico:
> I think it's odd that we would ignore the coding conventions on this - you might not agree with them but following the official coding conventions follows the principle of least surprise.

+1 to this.

I have to mention that I'm a big fan of this function expression and I find mindself suggesting it in reviews over and over. I agree with @BraisGabin that sometimes it could decrease readability, but overall it's a net positive change to enable it

BraisGabin:
Then let's merge it :)

3flex:
Thanks! Just need an approval :)

BraisGabin (approved):

PR #7361, opened 2024-06-11
#7360: Add 'assign' to operatorSet of UnusedImport

### LINKED ISSUE #7360: UnusedImport reported for import org.gradle.kotlin.dsl.assign in kotlin-dsl
## Expected Behavior

No report

## Observed Behavior

Unused import is reported

## Steps to Reproduce

Add the following test to `UnusedImportSpec.kt` and see it failing:

```kotlin
@Test
fun `does not report equals operators in kotlin dsl`() {
    val main = """
        import org.gradle.api.Project
        import org.gradle.kotlin.dsl.assign
        import org.gradle.kotlin.dsl.configure
        import org.jetbrains.kotlin.gradle.dsl.JvmTarget
        import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
        
        fun Project.configureKotlinJvm() {
            this.extensions.configure<KotlinJvmProjectExtension> {
                compilerOptions {
                    jvmTarget = JvmTarget.JVM_17
                }
            }
        }
    """.trimIndent()
    assertThat(subject.lintWithContext(env, main)).isEmpty()
}
```

## Context

Using `kotlin-dsl` in a `build-logic` module to for example configure the `jvmTarget` you need the `org.gradle.kotlin.dsl.assign` import to do things like `jvmTarget = JvmTarget.JVM_17` because the `=` operator resolves to the `assign` function.

## Your Environment

* Version of detekt used: 1.23.6

PhilippNowak96:
This PR adds `assign` to the `operatorSet` of the UnusedImport rule so that kotlin-dsl assignments are not reporting unused imports.

Resolves #7360


BraisGabin:
🤔 sorry, but I can't find `assign` as an operator on the kotlin documentation: https://kotlinlang.org/docs/operator-overloading.html where is it documented? Or is this gradle specific?

PhilippNowak96:
@BraisGabin this seems to be some gradle specific stuff. I read this article about it https://blog.gradle.org/simpler-kotlin-dsl-property-assignment and then verified it by the test

3flex:
Yes it's Gradle specific and implemented as a compiler plugin applied by Gradle when it compiles scripts.

Kotlin offers the extension point to customise behaviour of the assign operator (and possibly others though I'm not sure).

Given this is Gradle specific, and in line with our general philosophy of not building library or tool-specific config as defaults, I think it would be better to introduce a config option for `additionalOperators` where things like this can be added for those that need it.

PhilippNowak96:
Thanks @3flex for the clarification ☺️ I like the idea introducing a separate set for non language default operators. Changes have been pushed.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedImport.kt
```diff
@@ -156,11 +157,17 @@ class UnusedImport(config: Config) : Rule(
...
+        // Kotlin language default operators
         private val operatorSet = setOf(
             "unaryPlus", "unaryMinus", "not", "inc", "dec", "plus", "minus", "times", "div",
             "mod", "rangeTo", "rangeUntil", "contains", "get", "set", "invoke",
             "plusAssign", "minusAssign", "timesAssign", "divAssign", "modAssign",
-            "equals", "compareTo", "iterator", "getValue", "setValue", "provideDelegate"
+            "equals", "compareTo", "iterator", "getValue", "setValue", "provideDelegate",
+        )
+
+        // Additional operators from libraries or tools, e.g. compiler plugins for gradle
```
cortinico, line 168:
```suggestion
        // Additional operators from libraries or tools, e.g. compiler plugins for Gradle
```

cortinico (approved):

schalkms (approved):

3flex:
Sorry I didn't check this more closely - in my earlier comment I meant that the `assign` operator should be configurable and something to set in the YAML config. It's been implemented here as a hardcoded value in the rule but it's highly specific to Gradle so should be something that can be configured by users who are using that for Gradle analysis and not part of the standard rule config which should be framework-, tool- and library-agnostic.

I'll open a new issue.

PR #7376, opened 2024-06-15
Add KotlinBuiltIns extensions

3flex:
These would ideally be provided by org.jetbrains.kotlin.types.typeUtil which has wrappers for most KotlinBuiltIns functions. Those that are missing that would be generally useful can be added to detekt's TypeUtils.

### DIFF
--- a/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
@@ -1,5 +1,6 @@
 package io.gitlab.arturbosch.detekt.rules
 
+import org.jetbrains.kotlin.builtins.KotlinBuiltIns
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.config.LanguageVersionSettings
 import org.jetbrains.kotlin.descriptors.CallableDescriptor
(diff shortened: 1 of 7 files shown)


detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
```diff
@@ -21,6 +22,10 @@ fun KotlinType.fqNameOrNull(): FqName? {
     return TypeUtils.getClassDescriptor(this)?.fqNameOrNull()
 }
 
+fun KotlinType?.isString(): Boolean = KotlinBuiltIns.isString(this)
```
BraisGabin, line 25: In general I'm not a big fan of extensions over nullable. Should `null.isString()` be true or false? To me that's something that the caller should decide.

BraisGabin (commented):
Should we forbid the usages of the `KotlinBuildIns`? With a message to point to these functions.

cortinico:
> Should we forbid the usages of the `KotlinBuildIns`? With a message to point to these functions.

+1 to this 👍

cortinico (approved):

detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
schalkms: +1

schalkms (approved):

cortinico:
@3flex this can be rebase and merged

PR #7380, opened 2024-06-17
Remove `extensions.disableDefaultRuleSets` in favor of `RulesSpec.RunPolicy.DisableDefaultRuleSets`

BraisGabin:
Introduce `RulesSpec.RunPolicy.DisableDefaultRuleSets` and remove the boolean `extensions.disableDefaultRuleSets`. This makes our public api a bit easier and the code that implements it too.

### DIFF
--- a/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/rules/RuleSetsSpec.kt
+++ b/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/rules/RuleSetsSpec.kt
@@ -1,6 +1,7 @@
 package io.gitlab.arturbosch.detekt.core.rules
 
 import io.github.detekt.test.utils.resourceAsPath
+import io.github.detekt.tooling.api.spec.RulesSpec.RunPolicy.DisableDefaultRuleSets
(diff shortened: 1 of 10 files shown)


detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/rules/RuleSetsSpec.kt
```diff
@@ -66,7 +67,7 @@ class RuleSetsSpec {
                 .withSettings { createRuleProviders() }
         }
             .isInstanceOf(IllegalArgumentException::class.java)
-            .hasMessage("There was no rule set with id 'foo'.")
+            .hasMessage("There was not rule set with id 'foo'.")
```
cortinico, line 70: This feels wrong, the previous wording was probably better

cortinico (approved):

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/rules/RuleSetsSpec.kt
BraisGabin: Reverted

PR #7393, opened 2024-06-22
Reduce the times that we instantiantiate the rules

BraisGabin:
For each file detekt instantiate (at least) once each rule (and only the default rules are around 200 rules). I say at least because I think that they are twice but I'm not 100%. That seems like a waste of resources and that should make our analysis slower. This PR fixes that. The idea is that the rules shouldn't contain any state so we can instantiate them once and then reuse those instances all the times we need.

This PR is draft for two different reasons:
- This PR aims to improve the performance but I don't have any reliable benchmark to ensure if this is really an improvement or not.
- Right now this breaks the parallelism because `Rule` now uses a MutableList to store the findings. That MutableList is a state inside the `Rule`. But I don't want to focus on this until I check the first point.

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
@@ -26,6 +26,7 @@ import org.jetbrains.kotlin.config.languageVersionSettings
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.calls.smartcasts.DataFlowValueFactoryImpl
+import java.nio.file.Path
 import kotlin.reflect.full.hasAnnotation
 
 internal class Analyzer(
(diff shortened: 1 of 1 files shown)


atulgpt:
Hi, @BraisGabin for `The idea is that the rules shouldn't contain any state so we can instantiate them once and then reuse those instances all the times we need.` won't that break our rules where we are storing some info in the instance variable. I am also curious how we can tests these types of issue in UTs

BraisGabin:
> won't that break our rules where we are storing some info in the instance variable.

You are right. But if the performance improvement is relevant then we should fix that.

> I am also curious how we can tests these types of issue in UTs

I think that it's not possible to "ensure" that this doesn't happen. But if we create only one instance of each rule and execute all the tests using that single instance I think it could be "good enough".

atulgpt:
I think then immediate step would be to make the changes for a single instance of the rules and first fix the broken UTs. Not sure will they break deterministically or do we need to run those multiple times

BraisGabin:
> I think then immediate step would be to make the changes for a single instance of the rules and first fix the broken UTs. Not sure will they break deterministically or do we need to run those multiple times

No idea. Also I want to know if a change like this one would make detekt faster. Because if doesn't make it faster there is no point on doing this.

BraisGabin:
This code changed a lot. I still want to see if we improve performance but I will open a new pr.

PR #7394, opened 2024-06-23
Fix some Gradle Isolated Projects violations caused by detekt Gradle plugin

3flex:
About this feature: https://docs.gradle.org/current/userguide/isolated_projects.html

This PR should resolve all current IP violations for anyone using detekt on Gradle 8.8 and 8.9, except when using:
* Kotlin Mutiplatform ([KT-57279](https://youtrack.jetbrains.com/issue/KT-57279/Support-Gradle-Project-Isolation-Feature-for-Kotlin-Multiplatform))
* detekt report merging. We need a solution that properly shares & aggregates reports without referring to subprojects. I'm sure IP won't allow aggregating subproject task output into another project at any point, so we need to start doing things the "proper" way - https://docs.gradle.org/current/userguide/cross_project_publications.html
* DetektTaskMultiModuleSpec. These test whether setting things in allprojects/subprojects blocks are correctly applied to the relevant projects. Most of this behaviour is currently forbidden in IP. Gradle is making changes, so this situation should improve in future, possibly with changes required to build scripts: https://github.com/gradle/gradle/issues/25179

Our build also cannot use IP yet, there are violations in nexus publishing (https://github.com/gradle-nexus/publish-plugin/issues/343), dokka (https://github.com/Kotlin/dokka/issues/3656) and shadow (https://github.com/johnrengelman/shadow/issues/907) plugins (and a few in our build scripts, but no point fixing those until the dependencies are fixed).

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
@@ -29,10 +29,24 @@ class DetektPlugin : Plugin<Project> {
 
         project.registerDetektPlainTask(extension)
         project.registerDetektJvmTasks(extension)
-        if (!project.providers.gradleProperty(DETEKT_ANDROID_DISABLED_PROPERTY).getOrElse("false").toBoolean()) {
+        @Suppress("DEPRECATION")
+        val enableAndroidTasks =
+            !project.providers
+                .gradleProperty(DETEKT_ANDROID_DISABLED_PROPERTY)
+                .forUseAtConfigurationTime()
+                .getOrElse("false")
+                .toBoolean()
+        if (enableAndroidTasks) {
             project.registerDetektAndroidTasks(extension)
         }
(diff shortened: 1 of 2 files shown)


3flex:
Odd error on Ubuntu with JVM 8 & 11... need to investigate.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektPlugin.kt
```diff
@@ -29,10 +29,24 @@ class DetektPlugin : Plugin<Project> {
 
         project.registerDetektPlainTask(extension)
         project.registerDetektJvmTasks(extension)
-        if (project.findProperty(DETEKT_ANDROID_DISABLED_PROPERTY) != "true") {
+        @Suppress("DEPRECATION")
+        val enableAndroidTasks =
+            !project.providers
+                .gradleProperty(DETEKT_ANDROID_DISABLED_PROPERTY)
+                .forUseAtConfigurationTime()
```
3flex, line 36: `forUseAtConfigurationTime` is required for Gradle 7.3 and earlier. It's a no-op on 7.4 and up.

This will continue working until its scheduled removal in Gradle 9 - we can worry about it again then.

BraisGabin (approved):

PR #7397, opened 2024-06-24
Handle non-existent baseline file handling at task execution

3flex:
This allows use of lazy configuration for the baseline property instead of checking if the file exists at configuration time. It also fixes some config ordering issues that were exposed by #7398.

Doing this is a little awkward due to https://github.com/gradle/gradle/issues/12388. This situation should improve with Gradle 9 when https://github.com/gradle/gradle/issues/24767 is addressed (it's dependent on Kotlin 2.0.0 for Gradle).

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
@@ -70,6 +70,15 @@ internal data class BaselineArgument(val baseline: RegularFile?) : CliArgument()
     override fun toArgument() = baseline?.let { listOf(BASELINE_PARAMETER, it.asFile.absolutePath) }.orEmpty()
 }
 
+internal data class BaselineArgumentOrEmpty(val baseline: RegularFile?) : CliArgument() {
+    override fun toArgument(): List<String> =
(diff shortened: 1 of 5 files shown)


detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
```diff
@@ -70,6 +70,15 @@ internal data class BaselineArgument(val baseline: RegularFile?) : CliArgument()
     override fun toArgument() = baseline?.let { listOf(BASELINE_PARAMETER, it.asFile.absolutePath) }.orEmpty()
 }
 
+internal data class BaselineArgumentOrEmpty(val baseline: RegularFile?) : CliArgument() {
```
BraisGabin, line 73: Do we still use the Class defined above of this ome? Can we remove it?

BraisGabin (approved):

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
3flex: We need both, one is for passing the baseline parameter when creating a baseline (which may or may not exist). It needs to always pass the baseline parameter.

I tried the same initially but fortunately we have pretty good tests!

Issue #4084, opened 2021-09-02
Deprecation rule ignores @DeprecatedSinceKotlin


osipxd:
## Steps to Reproduce

1. Set Kotlin API version to 1.4 in the build script:
   ```kotlin
   tasks.withType<KotlinCompile>().configureEach {
        kotlinOptions {
            apiVersion = "1.4"
            languageVersion = "1.4"
        }
   }
   
   dependencies {
       implementation(kotlin("stdlib-jdk8", version = "1.5.30"))
   }
   ```
2. Use any function with annotation `@DeprecatedSinceKotlin(warningSince = "1.5")`, for example `String.capitalize()`:
   ```kotlin
   fun foo() {
       println("hello".capitalize())
   }
   ```

## Expected Behavior

There are no reports about deprecation

## Observed Behavior

Detekt reports `capitalize()` extension as Deprecation

## Your Environment

* Version of detekt used: 1.18.1
* Version of Gradle used (if applicable): 7.2
* Operating System and version: macOS 11.2.3

BraisGabin:
Nice catch. For sure this is something that the current rule doesn't handle at all.

cortinico:
We can definitely extend the rule to support the `@DeprecatedSinceKotlin` annotation.

The problem that I see here is that we will most likely not be able to correctly get your `apiLevel` and other custom compiler flags you set.

The reason behind this is that Detekt runs its own compilation with different compiler flags.

This is a feature that could be fully implemented once we move to the Detekt compiler plugin.

atulgpt:
Hi, @BraisGabin / @cortinico  is there any way to write TC against a particular Kotlin language version or specific compiler flag?

cortinico:
> is there any way to write TC against a particular Kotlin language version or specific compiler flag?

Nope that's not possible as of today

atulgpt:
Hi, @3flex since we migrated to worker API, is passing the custom compiler flag feasible now?

3flex:
No that's not related to adding support for the other compiler flags. There are changes needed in multiple modules to add this support, whether through the compiler plugin or detekt's own compilation environment.

atulgpt:
Hi @cortinico, I found a way to write a TC against a particular language version

```kotlin
@Test
fun `does not report when language settings is below than DeprecatedSinceKotlin version`() {
    val code = """
        @Deprecated("deprecation message")
        @DeprecatedSinceKotlin(warningSince = "1.4")
        fun foo() {
[...]

PR #7399, opened 2024-06-24
Allow passing API version to compiler config

3flex:
Prerequisite for #4084

### DIFF
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgsSpec.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgsSpec.kt
@@ -8,6 +8,7 @@ import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.assertThatCode
 import org.assertj.core.api.Assertions.assertThatExceptionOfType
 import org.assertj.core.api.Assertions.assertThatIllegalArgumentException
+import org.assertj.core.api.Assertions.assertThatIllegalStateException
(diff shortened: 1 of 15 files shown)


detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgsSpec.kt
```diff
@@ -269,6 +269,12 @@ internal class CliArgsSpec {
             assertThat(spec.compilerSpec.jvmTarget).isEqualTo("1.8")
         }
 
+        @Test
+        fun `--api-version is accepted`() {
```
BraisGabin, line 273: Should we add a test to check what happen if we pass "foo"? Basically to assert that the error message is useful.

BraisGabin (approved):

BraisGabin:
Would it be possible for the plugin to read the default value form the Kotlin gradle plugin (if it is present)? That would simplify the configuration of detekt. If it's complex we can move it to another isuue / pr. I don't want to block this one, it's a great step forward.

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgsSpec.kt
3flex: I didn't think it was worth it, but went back after this comment and it turns out the parse function for ApiVersion actually uses very forgiving logic, so the error message would have been skipped in many if not all instances.

I've updated the logic and added some additional tests for other parameters.

PR #7400, opened 2024-06-24
Align API, language version and JVM target between detekt type resolution tasks and Kotlin compilation tasks

3flex:
Partial fix for #4465


BraisGabin (approved):
I really like this one! This makes the plugin configuration easier and less error prone.

Issue #7404, opened 2024-06-26
Should we use the Analysis API or `FirFile`? 


BraisGabin:
This change is going to be painful. The new compiler K2 comes with a new frontend: `FIR`. 2.0 still supports `PSI` but we should start thinking about moving from `PSI` to `FIR`.

We should find a way to transition from one to another. I see two main chunks of work:
1. The core, all the `detekt` engine that make the rules work.
2. The rules.

It will be great if we can change all the core without touching the rules, and it seems that it can be possible. `FirElement` has a property called `psi` that should help us with this. Later, once we have the core stable would be the time to move the rules from `PSI` to `FIR`. That seems like A LOT of work so probably we will need to ask help to the community to do it. For that reason I think that we should have our core kind of ready and the tests in place so a contributor can focus on the translation of the rules.

I didn't check too much but I think that we are going to be lucky and we will not need to change our rule tests that much.

I'm working on this but any help is more than welcome.

3flex:
Take a look at the Kotlin analysis API if you haven't yet - this might be an option, but the analysis API looks like the path forward with the most longevity.

3flex:
Analysis API readme: https://github.com/JetBrains/kotlin/blob/master/docs/analysis/analysis-api/analysis-api.md

There's an implementation for both the old and new compiler front ends.

We'd likely integrate with the "standalone" module for the CLI (I believe this is what KSP is doing for KSP2).

Also important to note is that the Kotlin plugin for IDEA is using this API. This should make IDEA inspections simple to port if we use this, and also guarantees that we can offer analysis capabilities that are equal to what the Kotlin IDEA plugin offers, i.e. fully comprehensive and taking care of things that are a bit tricky or hacky in detekt at the moment like dealing with smart casts.

3flex:
https://kotlin.github.io/analysis-api/migrating-from-k1.html

I do wonder if `bindingContext` will become unavailable to detekt in future Kotlin releases... I'm actually a bit surprised everything still seems to work!

3flex:
@nebffa has expressed interest in uplifting compiler plugin support for K2. Before any work starts we should discuss the approach a bit further.

Analysis API has a key limitation: it does not support compiler plugins at this stage. I'm not sure if it ever will. Based on [the definitions of different components used by Analysis API](https://github.com/JetBrains/kotlin/blob/v2.1.0-Beta1/analysis/analysis-api-platform-interface/README.md#definitions), this is where the different components fit:
[...]

PR #7406, opened 2024-06-26
[WIP] Start using FIR

BraisGabin:
WIP! This doesn't even work.

I'm trying to get `FirFile`s from the new compiler and then convert it back to `KtFile`. If this works then we can start changing the references of `KtFile` to `FirFile` and start the transition from PSI to FIR.

First step towards #7404


3flex:
Take a look at the Kotlin analysis API if you haven't yet - this might be an option, but the analysis API looks like the path forward with the most longevity.

BraisGabin:
Closing. At #7404 there are the reasons. Basically, we should use a different api instead of this one.

PR #7408, opened 2024-06-27
Allow passing extra compiler arguments when configuring compiler

3flex:
Builds on #7403, adding support for parsing and applying non-standard compiler flags that configure language settings (e.g. `-Xcontext-receivers`, `-Xexplicit-api` and `-opt-in` via `setupLanguageVersionSettings`), setting other Kotlin/JVM arguments (e.g. `-Xfriend-paths` via `setupJvmSpecificArguments`) and advanced Kotlin/JVM options (e.g. `-Xjvm-enable-preview` via `configureAdvancedJvmOptions`).

Some flags require additional code before they're properly supported by the compiler configuration, e.g. friend paths, but that will be handled in follow up PRs.

Required for #4465

### DIFF
--- a/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgsSpec.kt
+++ b/detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgsSpec.kt
@@ -242,12 +242,6 @@ internal class CliArgsSpec {
             assertThatExceptionOfType(HelpRequest::class.java)
                 .isThrownBy { parseArguments(arrayOf("--help")) }
         }
-
-        @Test
-        fun `throws HandledArgumentViolation on wrong options`() {
-            assertThatExceptionOfType(HandledArgumentViolation::class.java)
-                .isThrownBy { parseArguments(arrayOf("--unknown-to-us-all")) }
-        }
     }
 
     @Test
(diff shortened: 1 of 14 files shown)


detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgsSpec.kt
```diff
@@ -242,12 +242,6 @@ internal class CliArgsSpec {
             assertThatExceptionOfType(HelpRequest::class.java)
                 .isThrownBy { parseArguments(arrayOf("--help")) }
         }
-
-        @Test
-        fun `throws HandledArgumentViolation on wrong options`() {
-            assertThatExceptionOfType(HandledArgumentViolation::class.java)
-                .isThrownBy { parseArguments(arrayOf("--unknown-to-us-all")) }
-        }
```
BraisGabin, line 250: It would help, at least to me, to add a test with two free arguments and how are them stored in the spec.

BraisGabin (approved):

detekt-cli/src/test/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgsSpec.kt
3flex: Sorry I don't quite follow, do you mean an additional test with valid args?

BraisGabin: Yes. One that returns a freeArguments different than empty.

PR #7409, opened 2024-06-27
Automatically handle friend paths in JVM compilation analysis tasks

### LINKED ISSUE #3067: Wrong error:  "error: cannot access 'foo': it is internal in 'com.example.bar'"

3flex:
Fixes #3067 

Stacks on #7408

Required for #4465

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
@@ -8,6 +8,8 @@ import io.github.detekt.tooling.api.spec.ProjectSpec
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.jetbrains.kotlin.com.intellij.openapi.Disposable
 import org.jetbrains.kotlin.com.intellij.openapi.util.Disposer
+import org.jetbrains.kotlin.config.JVMConfigurationKeys
+import org.jetbrains.kotlin.load.kotlin.ModuleVisibilityManager
 import java.io.Closeable
 import java.io.File
 import java.io.OutputStream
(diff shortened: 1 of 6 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
```diff
@@ -42,11 +44,16 @@ internal class EnvironmentFacade(
...
         )
-        createKotlinCoreEnvironment(
+        val env = createKotlinCoreEnvironment(
             compilerConfiguration,
             disposable,
             printStream,
         )
+
+        val moduleVisibilityManager = ModuleVisibilityManager.SERVICE.getInstance(env.project)
+        compilerConfiguration.getList(JVMConfigurationKeys.FRIEND_PATHS).forEach(moduleVisibilityManager::addFriendPath)
```
BraisGabin, line 54: Nit: you could move this inside an `apply` to set more clear that this is still env configuration? I usually like to use apply for finish the correct construction of an instance.

BraisGabin (approved):
Tests maybe? On the last 3 PRs you are adding a lot of new features but none of them are tested to avoid regressions.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/settings/EnvironmentAware.kt
3flex: Tried this just now but it's the `ModuleVisibilityManager` that's being configured by this line, and using `apply` then only referencing `it` in `forEach` triggered `UnnecessaryApply` so I reverted and left it as is.

Using `apply` on `env` doesn't seem appropriate since it's not actually the `KotlinCoreEnvironment` (env) being configured by these lines. I've left it as is.

PR #7412, opened 2024-06-28
Validate the gradle wrapper on each gradle setup

BraisGabin:
This is basically to reduce the number of tasks that we have for each PR because it can be overwhelming for new contirbutors. And, to be honest, a bit annoying to have such a big scroll on the checks.

This moves what that workflow did to all the tasks that setup gradle. But I'm not sure if that's a good idea or I should just set it up for the pre-merge workflow. I guess that the time lost is going to be so little that we shouldn't care but I'm open to change it.

Note: I need to remove the "Validation gradle wrapper" check from the required checks but I will do that once this PR is apprved.


3flex:
I'm ok switching to this but why run the check on every Gradle setup? It makes a network call (if it's a recent Gradle update but the action hasn't been updated yet with the latest hardcoded hashes) so we should only do it once, in a single action.

cortinico (approved):

cortinico:
> Note: I need to remove the "Validation gradle wrapper" check from the required checks but I will do that once this PR is apprved.

I've done it and approved the PR so we can merge it 👍

Issue #4647, opened 2022-03-21
OptionalUnit conflicts with explicitApi=strict


pettermahlen:
## Expected Behavior

It should be possible to enable [strict explicit API mode](https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors) without getting conflicts with Detekt rules.

## Observed Behavior

The OptionalUnit rule, if activated, considers an explicitly declared Unit return type as an error, and the compilation fails if the return type is implicit.

## Steps to Reproduce

Enable both explicitApi = "strict" and OptionalUnit.

marschwar:
Thank you for bringing this up. I believe this rule is much older than the `explicitApi` mode and as such unaware of this setting. Personally I do not see many options here other than not enabling the rule as it will - by definition - cause your build to fail. 

I do not think it is a bad rule but it may not make sense for every project. I guess that is why it is not enabled by default.

Do you have any other suggestions? Please let us know.

pettermahlen:
We have a large repository, 1M+ LOC, where we want to have consistent style checks (so a single set of Detekt rules), but where it doesn't make sense to enable strict API everywhere. So ideally, the OptionalUnit rule would be aware of explicitApi setting, disabling itself when it is not 'off'. I'm not sure if that's even possible?

cortinico:
Hi @pettermahlen 👋 

> So ideally, the OptionalUnit rule would be aware of explicitApi setting, disabling itself when it is not 'off'. I'm not sure if that's even possible?

It's possible. The problem is that you'll have to run Detekt with type resolution, and I don't know if this is the case for you (@davidburstromspotify can maybe say?).

Moreover, we're also blocked on this side because of this:
- https://github.com/detekt/detekt/issues/4465

The tl;dr is that the compilation flag you specify once you enable `explicitApi` doesn't get propagate properly to Detekt, so we're effectively compiling your code without that flag. 

More on this also here:
- https://github.com/detekt/detekt/issues/3874

So yeah, it's doable, but it will take a couple of releases for this to even get implemented and reach the public.

For the timebeing, you should resort to locally suppressing inspections, or defining an `exclude` in the config that will exclude an entire folder (which I believe is where your library lives or so).

pettermahlen:
Thanks for the response, for the time being, we've disabled the `OptionalUnit` rule, so no problem for us. :) It's not the most important rule and off by default, so it seemed like an easy way out.

davidburstromspotify:
[...]

Issue #3874, opened 2021-06-15
RedundantVisibilityModifierRule false-positives when `explicitApi()` is defined in the `kotlin{}` extension 


eygraber:
I have `explicitApi()` enabled for my module:

```
kotlin {
  explicitApi()
}
```

but I still get warned about `RedundantVisibilityModifierRule` for public functions and classes.

## Your Environment
* Version of detekt used: 1.17.1
* Version of Kotlin used: 1.5.10
* Version of Gradle used (if applicable): 7.1

cortinico:
Could you provide a small reproducer? I'm not entirely sure what could be the root cause here

eygraber:
It was happening in this project https://github.com/eygraber/ejson-kotlin

The rule is currently disabled, but if you enable it, it should repro.

cortinico:
I've done a quick investigation here.

The problem is that the Kotlin compilation invoked by Detekt is running with different flags. The Explicit API flag is not passed over to detekt, therefore is essentially ignored.

This will be resolved once we migrate to https://github.com/detekt/detekt-compiler-plugin/ as the compiler flags will be the same.

I'm unsure how we want to fix this, but I have the gut feeling that the check for `isExplicitApiModeActive` we introduced in #3358 is not effective (should we remove it @schalkms ?).

schalkms:
Regarding the `explicitApi` modifier, this issue was fixed in a subsequent PR.

cortinico:
> Regarding the `explicitApi` modifier, this issue was fixed in a subsequent PR.

Do you happen to recall which one?

schalkms:
> Do you happen to recall which one?

#3358

cortinico:
> I'm unsure how we want to fix this, but I have the gut feeling that the check for `isExplicitApiModeActive` we introduced in #3358 is not effective (should we remove it @schalkms ?).

Yup so I believe the fix we introduced in #3358 is ineffective.

schalkms:
> Yup so I believe the fix we introduced in #3358 is ineffective.

I had to fiddle around with very low level internal compiler APIs. So maybe something changed there? It's very hard to test this E2E with automated test cases.

schalkms:
It could also be the case that the setting `explicitApi()` isn't picked up in each and every case depending on a variety of possible options to configure this setting.

cortinico:
> It could also be the case that the setting `explicitApi()` isn't picked up in each and every case depending on a variety of possible options to configure this setting.

[...]

PR #7418, opened 2024-06-30
Automatically handle explicit API option in JVM compilation analysis tasks

3flex:
Required for https://github.com/detekt/detekt/issues/4465, #3874 and https://github.com/detekt/detekt/issues/4647

This is a little different to the other options recently added to DGP for new compiler options. The option can only be set in Kotlin in the `kotlin` extension in Gradle scripts and cannot be set on individual Kotlin compilation tasks, so I've used the same convention for the detekt tasks - the compilation tasks will accept & pass the value to the compiler options used by detekt, so it's available for rules to use if they need to, but since it doesn't actually affect the compilation at all, and the Kotlin compilation tasks don't expose it either, I've kept it internal here too.

### DIFF
--- a/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
+++ b/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
@@ -18,6 +18,7 @@ import io.gitlab.arturbosch.detekt.invoke.DefaultReportArgument
 import io.gitlab.arturbosch.detekt.invoke.DetektInvoker
 import io.gitlab.arturbosch.detekt.invoke.DetektWorkAction
 import io.gitlab.arturbosch.detekt.invoke.DisableDefaultRuleSetArgument
+import io.gitlab.arturbosch.detekt.invoke.ExplicitApiArgument
 import io.gitlab.arturbosch.detekt.invoke.FailOnSeverityArgument
 import io.gitlab.arturbosch.detekt.invoke.FreeArgs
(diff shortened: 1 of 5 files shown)


schalkms (approved):

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -150,6 +151,10 @@ abstract class Detekt @Inject constructor(
     @get:Incubating
     abstract val freeCompilerArgs: ListProperty<String>
 
+    @get:Input
+    @get:Optional
+    internal abstract val explicitApi: Property<String>
```
BraisGabin, line 156: nit: String over enum?

BraisGabin (approved):
Nice!

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
3flex: If it's made public I agree that should change but given it's internal I think it's simpler to do it like this.

scottjasso:
Is there any reason this is internal? We write our own detekt plugin, but still using the `Detekt` task class, and this property is the only one we can't see, making this functionality impossible to use.

3flex:
As mentioned in the PR description it's because it's not configurable at task level in Kotlin so there's no reason to make it configurable on the task level in this plugin.

I suggest raising an issue though and describing your use case in more detail.

Issue #6979, opened 2024-02-16
Fail lint for `error(throwable)`


WildOrangutan:
Using `error(throwable)` is kind of bad thing to do. It will result in `throw IllegalStateException(throwable.toString())`.

## Expected Behavior of the rule

Fail lint, if `error()` argument is of `Throwable` type.

atulgpt:
Hi, @BraisGabin/ @WildOrangutan  I think we can add this in the forbidden method. So that using `error(throwable)` will be reported

psuzn:
Been doing some poking around and the issue lies in `FunctionMatcher.WithParameters.match(CallableDescriptor)`, where it compares the `encounteredParamTypes` and `parameters`. 

<img width="850" alt="image" src="https://github.com/detekt/detekt/assets/19205305/eab420ac-f2df-4874-acee-4c555fd07e02">

 If we configure it to forbid method call `kotlin.error(kotlin.Throwable)` the produced: 
  - `encounteredParamTypes` is `[kotlin.Any]` and.
  - `parameters` is `[kotlin.Throwable]` 
  
and it is just doing a string comparison so it doesn't know both parameter types indicate the same function signature. 

We can check their types to verify if one is a supertype of another? would appreciate it if there is a better alternative.

BraisGabin:
Thinking out loud. If I have this:

```kotlin
fun foo(any: Any){}
fun foo(number: Number){}
```

If now I forbid `foo(Any)` then a call like `foo(5)` would be forbidden? That shouldn't be the case. So I'm not sure if we can change this without breaking other things. Do you see a way to make it work? Then we can think about how to implement it and what we should use instead of `String`.

BraisGabin:
The problem is a bit more complex than that. Even ignoring the problem I pointed out before. In the codebease you would have something like `foo(IllegalStateException())`. And you would forbid `foo(Throwable)`.

Also, I would like to know if there are more real use cases to something like this. This change is not going to be easy so, if there are more use cases to forbid something like this we could thing to implement it. Otherwise I would vote for an specific rule.

psuzn:
We might be better off with a specific rule.

PR #7419, opened 2024-06-30
Add rule `ErrorUsageWithThrowable`

atulgpt:
Fixes https://github.com/detekt/detekt/issues/6979

### DIFF
--- a/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ErrorUsageWithThrowable.kt
+++ b/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ErrorUsageWithThrowable.kt
@@ -61,7 +61,7 @@ class ErrorUsageWithThrowable(config: Config) : Rule(
             errorMsgValueArg.getArgumentExpression()
         ]?.type ?: return
         if (errorMsgTypeInfo.isThrowableSubtypeOfThrowable()) {
-            report(CodeSmell(Entity.from(expression), description))
+            report(CodeSmell(Entity.from(errorMsgValueArg), description))
         }
     }
 
--- a/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ErrorUsageWithThrowable.kt
+++ b/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ErrorUsageWithThrowable.kt
@@ -19,8 +19,7 @@ import org.jetbrains.kotlin.types.typeUtil.supertypes
  * This rule reports usage of `error` method with [Throwable] parameter, i.e, `error(throwable)`.
  * The above will result in `throw IllegalStateException(throwable.toString())` which doesn't
  * provide any info about `throwable.message` or `stackTrace`.
- * Instead, use `error(throwable.message ?: "No error message provided")` to rethrow the throwable
- * as `IllegalStateException` with `throwable.message`.
+ * Instead, use `throw IllegalStateException(throwable)` to rethrow the throwable as `IllegalStateException`.
  *
  * <noncompliant>
  * fun foo() {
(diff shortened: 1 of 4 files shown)


detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ErrorUsageWithThrowable.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.util.getResolvedCall
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.types.typeUtil.supertypes
+
+/**
+ * This rule reports usage of `error` method with `throwable` parameter, i.e, `error(throwable)`.
```
cortinico, line 19:
```suggestion
 * This rule reports usage of `error` method with [Throwable] parameter, i.e, `error(throwable)`.
```

cortinico (approved):
LGTM 👍 just a nit

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ErrorUsageWithThrowable.kt
```diff
@@ -0,0 +1,80 @@
...
+ * }
+ * </noncompliant>
+ *
+ * <compliant>
+ * fun foo() {
+ *     try {
+ *         // ... some code
+ *     } catch(e: IOException) {
+ *         // some addition handling
+ *         error(e.message)
```
BraisGabin, line 42: `throw IllegalStateException(e)` should be the way to fix this kind of issue or just `throw e`. Otherwise you are shallowing the exception.

BraisGabin (commented):
The rule I think that is completely fine. The only problem is the solution that it propouse because it is a bad pattern to shallow the exception. For sure this rule doesn't need to check if something is shallowed or not (we already have a rule for that) but we shouldn't advise against that other rule.

So basically just change the rule description and the compliant code.

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ErrorUsageWithThrowable.kt
atulgpt: Nice catch, I was not aware that using error(e.message) will still trigger the swallowed exception warning

atulgpt: Updated at https://github.com/detekt/detekt/pull/7419/commits/165c52eaa6c219b1df40c7ee2df58ef3ecd1611e

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ErrorUsageWithThrowable.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.types.typeUtil.supertypes
+
+/**
+ * This rule reports usage of `error` method with [Throwable] parameter, i.e, `error(throwable)`.
+ * The above will result in `throw IllegalStateException(throwable.toString())` which doesn't
+ * provide any info about `throwable.message` or `stackTrace`.
+ * Instead, use `error(throwable.message ?: "No error message provided")` to rethrow the throwable
+ * as `IllegalStateException` with `throwable.message`.
```
BraisGabin, line 23:
```suggestion
 * Instead, use `throw IllegalStateException(throwable)` to rethrow the throwable as `IllegalStateException`.
```

Issue #7407, opened 2024-06-26
ClassOrdering rule prints incorrect message


spaceracoon:
## Steps to Reproduce

Example class that can triggers this behavior
```
class ExampleClass {

    var var1 = 1

    fun bar() {

    }

    var var2 = 2

    init {
    }

    companion object {
    }
}
```

## Context

Sometimes ClassOrdering Rule print you incorrect messages. For example in code above it types "ClassOrdering: method `bar()` should be declared before companion object." despite bar() already declared before companion object. The correct error here is "method `bar()` should be declared after property declarations and initializer blocks"

## Your Environment

* Version of detekt used: 1.23.6
* Version of Gradle used (if applicable):
* Gradle scan link (add `--scan` option when running the gradle task):
* Operating System and version:
* Link to your project (if it's a public repository):

PR #7423, opened 2024-07-01
Use the anchor which is already present before

atulgpt:
the violating element in case of reporting the `before` msg

Fixes https://github.com/detekt/detekt/issues/7407

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
@@ -12,8 +12,6 @@ import org.jetbrains.kotlin.psi.KtObjectDeclaration
 import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.KtSecondaryConstructor
 
-private typealias DeclarationToSectionPair = Pair<KtDeclaration, Section>
-
 /**
  * This rule ensures class contents are ordered as follows as recommended by the Kotlin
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrderingSpec.kt
```diff
@@ -137,7 +137,7 @@ class ClassOrderingSpec {
         val findings = subject.compileAndLint(code)
         assertThat(findings).hasSize(1)
         assertThat(findings[0].message).isEqualTo(
-            "secondary constructor should be declared before method declarations."
+            "secondary constructor should be declared after property declarations and initializer blocks."
```
atulgpt, line 140: this message is also fixed as earlier warning msg was not valid as `secondary constructor` was already `before method declarations`

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ClassOrdering.kt
```diff
@@ -12,8 +12,6 @@ import org.jetbrains.kotlin.psi.KtObjectDeclaration
 import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.KtSecondaryConstructor
 
-private typealias DeclarationToSectionPair = Pair<KtDeclaration, Section>
```
atulgpt, line 15: made this into data class for easier understandbility

BraisGabin (approved):

Issue #7306, opened 2024-05-19
Remove deprecated configuration items


3flex:
## Expected Behavior of the rule
Configuration items on the rule are all in use.

## Context
Some are deprecated. These should be removed for detekt 2.

cortinico:
For reference, those are the one I found:

```
detekt-rules-empty/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyFunctionBlock.kt
25-    @Configuration("Excludes all the overridden functions")
26:    @Deprecated("Use `ignoreOverridden` instead")
27-    private val ignoreOverriddenFunctions: Boolean by config(false)

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/VariableNaming.kt
37-    @Configuration("ignores member properties that have the override modifier")
38:    @Deprecated("This configuration is ignored and will be removed in the future")
39-    @Suppress("UnusedPrivateMember")

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionParameterNaming.kt
32-    @Configuration("ignores overridden functions with parameters not matching the pattern")
33:    @Deprecated("Use `ignoreOverridden` instead")
34-    private val ignoreOverriddenFunctions: Boolean by config(true)
--
36-    @Configuration("ignores overridden functions with parameters not matching the pattern")
37:    @Deprecated("This configuration is ignored and will be removed in the future")
38-    @Suppress("DEPRECATION", "UnusedPrivateMember")

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/ConstructorParameterNaming.kt
34-    @Configuration("ignores constructor properties that have the override modifier")
35:    @Deprecated("This configuration is ignored and will be removed in the future")
36-    @Suppress("UnusedPrivateMember")

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/FunctionNaming.kt
34-    @Configuration("ignores functions that have the override modifier")
35:    @Deprecated("This configuration is ignored and will be removed in the future")
36-    @Suppress("UnusedPrivateMember")

detekt-rules-naming/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/naming/BooleanPropertyNaming.kt
39-    @Configuration("ignores properties that have the override modifier")
40:    @Deprecated("This configuration is ignored and will be removed in the future")
[...]

PR #7425, opened 2024-07-01
Delete the Deprecated properties

atulgpt:
Fixes https://github.com/detekt/detekt/issues/7306


detekt-core/src/main/resources/deprecation.properties
```diff
@@ -1,25 +1,6 @@
 complexity>ComplexMethod=Rule is renamed to `CyclomaticComplexMethod` to distinguish between Cyclomatic Complexity and Cognitive Complexity
```
BraisGabin, line 1: We could remove this from the `detekt-generator` too so we remove all the deprecation messages.

BraisGabin (commented):
It looks really good.

detekt-core/src/main/resources/deprecation.properties
atulgpt: [Updated at 7ba555d32bea9e74d94cbe2871e0a0cf44d85fc3](https://github.com/detekt/detekt/pull/7425/commits/7ba555d32bea9e74d94cbe2871e0a0cf44d85fc3)

Issue #6101, opened 2023-05-15
New rule: Disallow `expression1?.let { expression2 } ?: run { expression3 }` when `expression2` is nullable


atulgpt:
## Expected Behavior of the rule
The rule should disallow `expression1?.let { expression2 } ?: run { expression3 }` like code as `run` block can be executed either when `expression1` is null or `expression2` returns `null` which is accidental most of the time. Most of new comers replace if else with `null` check with the above expression and they face this issue very later in the cycle

Some example codes as below
```
fun main() {
    val a: String? = "Non null string"
    a?.let {
        println("a is not null case")
        nullable()
    } ?: run {
        println("a is null case")
    }
}

fun nullable(): String? {
    return null
}
```
above code([playground](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS44LjIxIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiIsIm5vbmVNYXJrZXJzIjp0cnVlLCJ0aGVtZSI6ImlkZWEiLCJjb2RlIjoiLyoqXG4gKiBZb3UgY2FuIGVkaXQsIHJ1biwgYW5kIHNoYXJlIHRoaXMgY29kZS5cbiAqIHBsYXkua290bGlubGFuZy5vcmdcbiAqL1xuXG5mdW4gbWFpbigpIHtcbiAgICB2YWwgYTogU3RyaW5nPyA9IFwiTm9uIG51bGwgc3RyaW5nXCJcbiAgICBhPy5sZXQge1xuICAgICAgICBwcmludGxuKFwiYSBpcyBub3QgbnVsbCBjYXNlXCIpXG4gICAgICAgIG51bGxhYmxlKClcbiAgICB9ID86IHJ1biB7XG4gICAgICAgIHByaW50bG4oXCJhIGlzIG51bGwgY2FzZVwiKVxuICAgIH1cbn1cblxuZnVuIG51bGxhYmxlKCk6IFN0cmluZz8ge1xuICAgIHJldHVybiBudWxsXG59In0=)) prints both `a is not null case` and `a is null case` and hence is problematic
[...]

PR #7428, opened 2024-07-01
Add `UnnecessaryRun` rule

atulgpt:
This rule takes inspiration from `UnnecessaryApply` rule

Fixes https://github.com/detekt/detekt/issues/6101


atulgpt:
@cortinico @BraisGabin should we merge this PR?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRun.kt
```diff
@@ -0,0 +1,124 @@
...
+        (
+            lambdaArguments.firstOrNull()?.getLambdaExpression()
+                ?: valueArguments
+                    .firstOrNull()
+                    ?.getArgumentExpression()
+                    ?.unpackFunctionLiteral()
+            )
+}
+
+private val RUN_FQ_NAME = FqName("kotlin.run")
```
cortinico, line 124: Can't this live inside the `UnnecessaryRun` class?

cortinico (approved):
LGTM 👍

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRun.kt
atulgpt: That will create new instance of `RUN_FQ_NAME` for every rull instance. Or will need the companion object, right?

atulgpt: `UnnecessaryApply` also has same style where consts are defined as private file level variables

cortinico: Yup let's keep it there then for consistency with `UnnecessaryApply`

atulgpt:
Hi, @BraisGabin / @cortinico should we merge this PR? It's approved and open for very long time

cortinico:
Can we get a review from another @detekt/maintainers ?

schalkms (commented):
I'm afraid, I'm pretty busy these days, so can't review it thoroughly.
But, as this is pretty similar to the `UnnecessaryApply` rule, it should be good to go.

atulgpt:
Hi, @cortinico should we merge this PR since it follows the existing rule. If there is any issue we can improve around the same.(Keeping this open requires constant rebase from the main 😅)

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRunSpec.kt
```diff
@@ -0,0 +1,493 @@
...
+                                prop = 1
+                            }
+
+                            val a3 = c.run {
+                                prop += 1
+                            }
+
+                            val a4 = c.run {
+                                prop /= 1
+                            }
```
TWiStErRob, line 381: All these `a*` are `: Unit` right? because the assignment expression is not like Java.

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRunSpec.kt
```diff
@@ -0,0 +1,493 @@
...
+                        }
+                    """.trimIndent()
+                )
+            ).hasSize(3)
+        }
+
+        @Test
+        fun `does not report when result is used`() {
```
TWiStErRob, line 357: feels like a similar test is needed for using the value in different ways, like

```kotlin
f(c.run { prop = 1 })
```

```kotlin
a + c.run { prop = 1 }
```

```kotlin
"${c.run { prop = 1 })}"
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRunSpec.kt
```diff
@@ -0,0 +1,493 @@
...
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.rules.KotlinCoreEnvironmentTest
+import io.gitlab.arturbosch.detekt.test.assertThat
+import io.gitlab.arturbosch.detekt.test.compileAndLintWithContext
+import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+@KotlinCoreEnvironmentTest
+class UnnecessaryRunSpec(val env: KotlinCoreEnvironment) {
```
TWiStErRob, line 12: did you consider this edge case, which is near impossible to fix? (apart from changing to other scope functions)

```kotlin
interface Foo {
   fun Bar.baz()
}
fun f(foo: Foo, bar: Bar) {
    foo.run { bar.baz() }
}
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRunSpec.kt
atulgpt: `All these a* are : Unit right?` yes all these are `Unit`

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRunSpec.kt
atulgpt: Yes, this is failing. Problem here is to find the places where scope is used(traverse the AST nodes filter-out the applicable usages and then check if scope if used and if that is same as provided by the `run`)

Adding the complete TC for ref
```kotlin
@Test
fun `do not report when scope by run is used`() {
    assertThat(
        subject.compileAndLintWithContext(
            env,
            """
                class Bar
                interface Foo {
                    fun Bar.baz()
                }
                fun f(foo: Foo, bar: Bar) {
                    foo.run { bar.baz() }
                }
            """.trimIndent()
        )
    ).isEmpty()
}
```

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRunSpec.kt
atulgpt: Add these TC in this commit https://github.com/detekt/detekt/pull/7428/commits/2e6245ba658b6de2432de005b559c5c5ab8de731

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRunSpec.kt
TWiStErRob: Makes sense, it's just a bit weird code. Maybe it's not this rule's job to highlight it though. There could be more realistic examples, like: `fun f(): Unit = c.run { prop += 1 }`

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRunSpec.kt
TWiStErRob: ~I guess we can create a bug and merge this?~ I wonder how often this would be the case, as `apply` and `with` are usually more applicable to this use case (not returning the value), so it might not even be a false positive?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRunSpec.kt
atulgpt: I can also add `fun f(): Unit = c.run { prop += 1 }` to this example. What do you say?

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRunSpec.kt
atulgpt: Agree but TBH when rule will be used out by many other devs then we will find more cases where these type patterns might become useful or require changes in the rule

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRunSpec.kt
atulgpt: Hi @TWiStErRob added the TC at https://github.com/detekt/detekt/pull/7428/commits/8b5e94421702ac1b2713fd0c48d975f1176d8dda

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRunSpec.kt
atulgpt: @TWiStErRob do you agree?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnnecessaryRun.kt
```diff
@@ -0,0 +1,124 @@
...
+ * config.version = "1.2"
+ * config?.environment = "test"
+ * val final = getNullableValue() ?: fallback
+ * stringBuilder.run {
+ *     append("test")
+ *     toString()
+ * }
+ * </compliant>
+ */
+@RequiresFullAnalysis
```
3flex, line 47: Apologies but we are migrating to the Analysis API so it would be much better if this can be implemented using that instead, otherwise we'll have to migrate it

schalkms: @atulgpt do you still plan to push this PR forward by leveraging the new Analysis API? Or shall we close this PR?

atulgpt: Hi @schalkms I am closing this PR as of now.. Will reopen it as some of TCs which were possible earlier are now not possible with AA apis

schalkms: Ok, good to know. Ty for the reply.

PR #7429, opened 2024-07-01
Update minor logic for checking `isExplicitApiModeActive`

atulgpt:
Enhances https://github.com/detekt/detekt/issues/3874


3flex:
Thanks, agree that the intent of both strict and warning modes is that it suppresses reporting from this rule.

This also needs test coverage for the different explicit API modes, it would be great if that was added here.

atulgpt:
Hi @3flex tests is already there at https://github.com/detekt/detekt/blob/2cb8485ae8dba674222105ffb883c4f43f13c9a5/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantVisibilityModifierSpec.kt#L185 

It creates the fake compiler resource with `STRICT` and `DISABLED`. With `WARNING` can be added

atulgpt:
> With WARNING can be added

Added TC for warning as well

schalkms (approved):

PR #7430, opened 2024-07-02
Use `isExplicitApiModeActive` in `OptionalUnit` rule

atulgpt:
Fixes https://github.com/detekt/detekt/issues/4647


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/optional/OptionalUnit.kt
```diff
@@ -103,11 +105,19 @@ class OptionalUnit(config: Config) : Rule(
                 true
         }
 
+    private fun isExplicitApiModeActive(): Boolean {
```
atulgpt, line 108: This method is also present in `RedundantVisibilityModifier` but was not sure where to put this as common utils. Tried at `TypeUtils.kt` but there a module which contains `CompilerResources` is not present. So created one more utils here as of now

cortinico (approved):
LGTM 👍

BraisGabin (approved):

PR #7440, opened 2024-07-04
[WIP] Check the specific config first

atulgpt:
Fixed https://github.com/detekt/detekt/issues/1466

### DIFF
--- a/detekt-formatting/src/test/kotlin/io/gitlab/arturbosch/detekt/formatting/AutoCorrectLevelSpec.kt
+++ b/detekt-formatting/src/test/kotlin/io/gitlab/arturbosch/detekt/formatting/AutoCorrectLevelSpec.kt
@@ -19,11 +19,11 @@ class AutoCorrectLevelSpec {
         "Undefined, Undefined, false",
         "Undefined, True,      true",
         "Undefined, False,     false",
-        "True,      Undefined, false",
+        "True,      Undefined, true",
         "True,      True,      true",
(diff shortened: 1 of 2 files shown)


BraisGabin:
I think that this will not work either. If you set autocorrect to false on the ruleset and nothing else that configuration will do nothing at all. Because you will read the default value of each rule and it will say true so it will be enabled.

detekt-formatting/src/test/kotlin/io/gitlab/arturbosch/detekt/formatting/AutoCorrectLevelSpec.kt
```diff
@@ -19,11 +19,11 @@ class AutoCorrectLevelSpec {
         "Undefined, Undefined, false",
         "Undefined, True,      true",
         "Undefined, False,     false",
-        "True,      Undefined, false",
+        "True,      Undefined, true",
```
BraisGabin, line 22: Also I don't think we want this change either... If you have a ruleset with 2 rules one with auto correct and the other without this would make both of them autocorrect. I'm not sure about the implications of that.

atulgpt: I think a rule without any support of autocorrect will be no-op if you pass the flag

atulgpt:
Closing now.. some more thoughts needed for this PR

PR #7444, opened 2024-07-05
BracesOnIfStatements: fix false-positive when chained

bomiyr:
`BracesOnIfStatements` configured with `multiLine=always/consistent` reports violation in case of if-else-if-else chained with other expressions. For example:
```
if (true) {
    ""
} else if (true) {
    ""
} else {
    ""
}.trim()
```

This MR fixes such issues


schalkms (approved):
Thanks
Looks good

PR #7468, opened 2024-07-21
Indroduce `Rule.Id`

BraisGabin:
The same way we have `Rule.Name` and `RuleSet.Id` we should also have `Rule.Id` to have an api more type safe.

This PR also do another minor improvements in the code around these values.

Thanks to @TWiStErRob because the main idea of this PR came from him and we work on it on a fast pair programming session.

@TWiStErRob tell me if there is something else that I should cover.

~Blocked by:~
- #7263

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
@@ -39,8 +39,9 @@ val Issue.suppressed: Boolean
     get() = suppressReasons.isNotEmpty()
 
 interface RuleInstance {
-    val id: String
-    val name: Rule.Name
+    val id: Rule.Id
     val ruleSetId: RuleSet.Id
     val description: String
 }
+
+val RuleInstance.name: Rule.Name get() = id.ruleName
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
@@ -43,5 +43,3 @@ interface RuleInstance {
     val ruleSetId: RuleSet.Id
     val description: String
 }
-
-val RuleInstance.name: Rule.Name get() = id.ruleName
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
@@ -1,7 +1,6 @@
 package io.gitlab.arturbosch.detekt.api
 
 import dev.drewhamilton.poko.Poko
-import io.gitlab.arturbosch.detekt.api.internal.identifierRegex
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.resolve.BindingContext
 
@@ -91,7 +90,7 @@ open class Rule(
     @Poko
     class Name(val value: String) {
         init {
-            require(value.matches(identifierRegex)) { "Name '$value' must match ${identifierRegex.pattern}" }
+            require(value.matches(nameRegex)) { "Name '$value' must match ${nameRegex.pattern}}" }
         }
 
         override fun toString(): String = value
@@ -100,7 +99,7 @@ open class Rule(
     @Poko
     class Id(val value: String) {
         init {
-            require(value.matches(ruleIdPattern)) { "Id '$value' must match ${identifierRegex.pattern}" }
+            require(value.matches(idRegex)) { "Id '$value' must match ${idRegex.pattern}" }
         }
 
         val ruleName: Name get() = Name(value.split("/", limit = 2).first())
@@ -109,4 +108,5 @@ open class Rule(
     }
 }
 
-private val ruleIdPattern = Regex("${identifierRegex.pattern}(/${identifierRegex.pattern})?")
+private val nameRegex = Regex("[aA-zZ]+(?:[aA-zZ0-9-]+)*[aA-zZ0-9]")
+private val idRegex = Regex("${nameRegex.pattern}(/${nameRegex.pattern})?")
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
@@ -99,8 +99,14 @@ open class Rule(
 
     @Poko
     class Id(val value: String) {
+        init {
+            require(value.matches(ruleIdPattern)) { "Id '$value' must match ${identifierRegex.pattern}" }
+        }
+
         val ruleName: Name get() = Name(value.split("/", limit = 2).first())
 
         override fun toString(): String = value
     }
 }
+
+private val ruleIdPattern = Regex("${identifierRegex.pattern}(/${identifierRegex.pattern})?")
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
@@ -1,7 +1,7 @@
 package io.gitlab.arturbosch.detekt.api
 
 import dev.drewhamilton.poko.Poko
-import io.gitlab.arturbosch.detekt.api.internal.validateIdentifier
+import io.gitlab.arturbosch.detekt.api.internal.identifierRegex
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.resolve.BindingContext
 
@@ -91,7 +91,7 @@ open class Rule(
     @Poko
     class Name(val value: String) {
         init {
-            validateIdentifier(value)
+            require(value.matches(identifierRegex)) { "Name '$value' must match ${identifierRegex.pattern}" }
         }
 
         override fun toString(): String = value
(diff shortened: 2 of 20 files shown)


detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt
```diff
@@ -72,7 +72,7 @@ fun Rule.lint(ktFile: KtFile, compilerResources: CompilerResources = FakeCompile
 
 private fun List<Finding>.filterSuppressed(rule: Rule): List<Finding> =
     filterNot {
-        it.entity.ktElement.isSuppressedBy(rule.ruleName.value, rule.aliases, RuleSet.Id("NoARuleSetId"))
+        it.entity.ktElement.isSuppressedBy(Rule.Id(rule.ruleName.value), rule.aliases, RuleSet.Id("NoARuleSetId"))
```
TWiStErRob, line 75: does this have a typo in it?

```suggestion
        it.entity.ktElement.isSuppressedBy(Rule.Id(rule.ruleName.value), rule.aliases, RuleSet.Id("NotARuleSetId"))
```
```suggestion
        it.entity.ktElement.isSuppressedBy(Rule.Id(rule.ruleName.value), rule.aliases, RuleSet.Id("NoRuleSetId"))
```

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
```diff
@@ -91,9 +90,23 @@ open class Rule(
...
         override fun toString(): String = value
     }
+
+    @Poko
+    class Id(val value: String) {
+        init {
+            require(value.matches(idRegex)) { "Id '$value' must match ${idRegex.pattern}" }
+        }
+
+        val ruleName: Name get() = Name(value.split("/", limit = 2).first())
```
TWiStErRob, line 105: how come we never read the second item in the split?

detekt-api/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/test/TestFactory.kt
```diff
@@ -58,15 +58,11 @@ fun createRuleInstance(
     id: String = "TestSmell/id",
```
TWiStErRob, line 58: is this default value ever used? it looks strange as if a `$` is missing. Should we choose a different "singleton" value?

detekt-api/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/test/TestFactory.kt
```diff
@@ -58,15 +58,11 @@ fun createRuleInstance(
...
-    return RuleInstanceImpl(
-        id = id,
-        name = Rule.Name(split.first()),
-        ruleSetId = RuleSet.Id(ruleSetId),
-        description = description
-    )
-}
+): RuleInstance = RuleInstanceImpl(
+    id = Rule.Id(id),
+    ruleSetId = RuleSet.Id(ruleSetId),
```
TWiStErRob, line 63: should the propagation of the instantiation of these types stop here? feels like the `createRuleInstance` API should expect a `Rule[Set].Id` already?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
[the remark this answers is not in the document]
BraisGabin: Yeah, I think that I have it like that because at the beginning it was an `@JvmInline value class`.

BraisGabin: The `id` is "FunctionNaming/test" and the `name` is "FunctionNaming". We never need to get only the second part. "test" means nothing.

detekt-api/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/test/TestFactory.kt
BraisGabin: Yes, it is used in multiple places. A rule id can have two "structures": "RuleSmell" or "RuleSmell/something". "something" alone is not a valid id.

Probably this default value is misleading. I could change it for "TestSmell/postfix" or something similar to avoid confusions. What do you think?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSet.kt
```diff
@@ -15,9 +14,11 @@ class RuleSet(val id: Id, val rules: Map<Rule.Name, (Config) -> Rule>) {
     @Poko
     class Id(val value: String) {
         init {
-            validateIdentifier(value)
+            require(value.matches(idRegex)) { "Id '$value' must match ${idRegex.pattern}" }
```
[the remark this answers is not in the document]
BraisGabin: Done, I improved the tests and also remove the {} because now they are not needed.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
```diff
@@ -91,9 +90,23 @@ open class Rule(
...
+            require(value.matches(nameRegex)) { "Name '$value' must match ${nameRegex.pattern}}" }
         }
 
         override fun toString(): String = value
     }
+
+    @Poko
+    class Id(val value: String) {
+        init {
+            require(value.matches(idRegex)) { "Id '$value' must match ${idRegex.pattern}" }
```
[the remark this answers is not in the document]
BraisGabin: Same

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSet.kt
```diff
@@ -15,9 +14,11 @@ class RuleSet(val id: Id, val rules: Map<Rule.Name, (Config) -> Rule>) {
...
         }
 
         override fun toString(): String = value
```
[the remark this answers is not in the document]
BraisGabin: > Btw, doesn't Poko do this already, so we could just remove the toString?

Yes they do, I overrided it intentionally to change it.

My idea is that you could use `g(..."$id"...)` freely. And that it will return the same as `g(..."${id.value}"...)`. I don't see what we win if we don't allow that.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
```diff
@@ -39,8 +39,9 @@ val Issue.suppressed: Boolean
...
 
 interface RuleInstance {
-    val id: String
-    val name: Rule.Name
+    val id: Rule.Id
     val ruleSetId: RuleSet.Id
     val description: String
 }
+
+val RuleInstance.name: Rule.Name get() = id.ruleName
```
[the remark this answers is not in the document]
BraisGabin: Removed. It was just used in 2 places on our code.

detekt-api/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/test/TestFactory.kt
BraisGabin: It could but I find that for tests it's easier to hardcode just `String`s instead of `Rule[Set].Id`s. And probably easier to read too.

cortinico:
CI is really red also 😅

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSet.kt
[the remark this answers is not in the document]
TWiStErRob: Relying on implicit toString is not good in complex code. It's not navigable (find usages), and some of these value classes don't just have value but other calculated properties. Having no "usable" toString forces the user to decide which property they actually need.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Rule.kt
TWiStErRob: Is the limit relevant then?

TWiStErRob: Maybe use substringBefore?

BraisGabin: Well, the limit was more a performance thing. But you are right `substringBefore` is WAY better.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RuleSet.kt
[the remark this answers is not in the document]
BraisGabin: Well, that's kind of similar to what happen with `Regex` in this comment: https://github.com/detekt/detekt/pull/7468#discussion_r1788091282, isn't it? I have `${regex.pattern}` but I could use directly `$regex` because the main value of a `Regex` is really clear. I think that in an `Id` the same happens. The `toString` of an `Id` should be the id itself.

Anyway, I could change it if you still think it's better, my position isn't that strong. Should I do it just for `Rule.Id` or also for `RuleSet.Id` and `Rule.Name`? All of them follow the same pattern.

BraisGabin:
> CI is really red also 😅

Fixed! I would like to integrate spotless on detekt. It's way easier to fix this kind of problems with that plugin than detekt.

BraisGabin:
This PR is too old and has too many conflicts so closing. Also with #7775 this PR has less value.

Issue #6879, opened 2024-01-17
Make output of detekt serializable


BraisGabin:
I'm not even sure if `Finding2` should have an `Entity`. I think that it would be great to make `Finding2` serializable some how (json, sarif...). And `Entity` has a refenrece to a `KtElement` and that can not be serializable.

I think that this change would be good because:

- The reporters shouldn't work with the complexity of `Entity`. We should need to expand `Finding` to contain the locations and those things but that's it.
- We could change our gradle plugin and have one task to run detekt and another to generete the reports. This would make really fast the use case of execute detekt and then execute detektBaseline.
- Merge reports would be way, way easier. You would get a `List<Detektion>` and then you decide how to merge them but you have all the information there.

The cons:
- Right now the html report uses `Entity` to generate the code snippets. If we don't provide the `Entity` we should look for other ways to implement that.

marschwar:
> * Right now the html report uses `Entity` to generate the code snippets.

Same for markdown

BraisGabin:
I'm working on this now.

The main blocker for making `Detektion` serializable is that `Issue` contains an `Entity` and `Entity` contains a reference to a `KtElement` and the `KtElement` is not serializable.

The first step is to create a new `Entity` class for `Issue`. Right now `CodeSmell` and `Issue` share the same implementation of `Entity` but that should change because we need to remove `KtElement` from `Issue` but we can't do that for `CodeSmell`. At the `CodeSmell` level we need the `KtElement` to know if we should suppress for a `@Suppress` annotation and other similar use cases.

For that reason I started simplifying `Entity`: #7470

Also we need to change the `Location` class inside `Entity`.

At the level of `CodeSmell` we always want to play with absolute paths. The absolute paths are less error prone when you want to access the files. But at the `Issue` level we want to work only with relative paths. The `Issue` is the element that will be serialized so it should be relative to work well with the gradle cache. Also the reports that we generate (html, xml...) should be also relative paths so this change should make those less error prone. But the console output should still be absolute. There are a lot of things to tune.

You should expect a lot of PRs improving our tests to ensure that the paths that we show are the correct ones.

BraisGabin:
Ok, I have the first part of this ready. I was able to remove KtElement and make `location.path` relative keeping all working as before. I can't launch yet the PRs because there are already a lot of PRs open with lots of dependencies.

Once I get this first part merged the second part is to make `Issue` serializable but that should be easy. I vote to use `json` for that serialization but I accept other ideas.

[...]

PR #7470, opened 2024-07-21
Remove txt report

BraisGabin:
I remember that someone talked about this before. The `txt` report is kind of useless. The information that if give is not easy to read by a human and the information is not that interesting for a script to parse. I can imagine people to use it to count the number of issues based on the number of lines in that file. But I think that's not a good reason to keep a `txt`. If someone wants that they can build their own reporter that just write the count.

The only use case for it is to manually create a `baseline`. But that's a really odd use case. And if you really want to do that you can create it automatically and them hand-craft it from there.

So for those reasons I vote to remove it.


3flex:
One helpful thing is that it's the easiest format to review from a CLI after a detekt run. And it's also easy to concatenate that output into a file instead.

> The txt report is kind of useless. The information that if give is not easy to read by a human and the information is not that interesting for a script to parse.

Instead of removing it, we can improve it so it's more usable for human readers, I wouldn't expect scripts to parse it, that's what #6879 is for.

cortinico:
> Instead of removing it, we can improve it so it's more usable for human readers, I wouldn't expect scripts to parse it, that's what #6879 is for.

Isn't the MD format already serving this same need? If so we could just remove `txt` or "alias" it to output the same output as the md reported

3flex:
> Isn't the MD format already serving this same need?

I guess it does - TXT output is more concise though.

Personally I rely on Gradle output and never look at the reports so I'm not personally affected by this, but it's not exactly a maintenance burden to keep it and others might use it. But we can revert if anyone complains after an RC :)

cortinico (approved):
Yup agree on potentially removing this now

3flex:
LGTM after a rebase

PR #7474, opened 2024-07-22
Add basePath inside `SetupContext`

### LINKED ISSUE #7241: Create a property to store the `basePath`
## Expected Behavior
It's easy to find the basePath on the detekt API.

I would vote to add a `val basePath: Path` inside `SetupContext` and use it instead of `getOrNull(DETEKT_OUTPUT_REPORT_BASE_PATH_KEY)`.

## Current Behavior
It's not, we are storing in an user property and you need to know the constant to read its value.

## Context
 https://github.com/detekt/detekt/pull/7234#discussion_r1581766941

BraisGabin:
The reports need to access the basePath to create the correct outputs. For that reason I'm moving this to a property inside `SetupContext` so it is easier to find. In the next PRs I will increase the usages of `basePath` for that reason I'm doing this change first.

This also ensure that basePath will be always be there. This makes that all our reports now doesn't need to support absolute and relative. It will be always relative as we want.

Fixes #7241


3flex:
Fixes #7241?

BraisGabin:
> Fixes #7241?

Yes! Updated in the description.

BraisGabin:
@detekt/maintainers a review here would help a lot. I have some PRs depending on this and once all of those are merged we can start to implement the serializable output.

schalkms (approved):
👍

PR #7476, opened 2024-07-22
Improve how we create Fakes of `SetupContext`

BraisGabin:


BraisGabin:
@detekt/maintainers a review here would be great. This one should be easy/non controversial and will unlock more interesting PRs.

schalkms (commented):
Isn’t this *just* a refactoring by moving code around in the test package? What’s really getting improved here?

BraisGabin:
Yes, this is a refactor. It improves the code for two reasons: 

- The tests are less verbose and more focused on what you are tesring.
- make the tests less fragile. If you add a new property `SetupContext` the tests that doesn't care about that new property doesn't need to be changed now. And that's exactly what happen in the next PRs.

(Sorry this should be written when I created this PR)

schalkms (approved):
Thanks for the explanation! Approved ✅

Issue #7123, opened 2024-04-03
detekt-tooling should not depend on detekt-api


3flex:
## Expected Behavior
detekt-tooling should be a stand alone end point that is called by clients (CLI, Gradle, IDE plugin etc). It should not depend on detekt-api.

## Current Behavior
It does.

## Context
detekt-api dependency pulls kotlin-compiler-embeddable transitively which should not be necessary to interface with detekt-tooling. If the detekt-api dependencies were cleaned up then this wouldn't be an issue, but as it stands, this would be a blocker for migrating the Gradle plugin to call detekt-tooling instead of the CLI.

BraisGabin:
I agree, remove `kotlin-compiler-embeddable` from `:detekt-api` right now is really difficult (maybe impossible).

I see 3 main dependencies:
- [x] `KtFile` and `BindingContext` in `Detekt.run(Collection<KtFile>, BindingContext)`
- [ ] `Detektion`
- [ ] `Config` in `DefaultConfigurationProvider`

---

The problem with `Detekt.run(Collection<KtFile>, BindingContext)` is probably the easiest. I imagine that it is ok if `:detekt-tooling` depend on `kotlin-compiler-embeddable` only for compilation (`comileOnly(libs.kotlin.compilerEmbeddable)`). If that's not an option I don't know how to fix it.

`Detektion` is a class defined on `detekt-api` that contains the results of the run. This one shouldn't be difficult. We can basically move `Detektion` to `detekt-tooling` and that's it. We should change some other classes on `detekt-api` that depend on that class but it shouldn't be a big problem. `Detektion` has a transitive dependency with `KtElement` so same solution as before: `comileOnly(libs.kotlin.compilerEmbeddable)`. Also, #6879 should break that transitive dependency.

About `Config`, I like the idea that the client should provide a `Config` instance. This will move classes like `YamlConfig` or `DisabledAutoCorrectConfig` outside `:detekt-core` and that feels good. The problem is that the `Rule` `constructor` requires a `Config` instance to be instantiated so moving `Config` outside `:detekt-api` is more difficult. I can imagine different solutions here:
1. Create a `:common-api` (we should work on the naming). It will contain `Config` (and maybe `Detektion`). `:detekt-api` and `detekt-tooling` would depend on it.
2. Create a `Config2` (same, name pending) on `:detekt-tooling` and `:detekt-core` will tranform `Config2` to `Config` so it can be passed to a `Rule` to be configured.
[...]

PR #7490, opened 2024-07-27
Add flag to enable and disable type resolution mode

### LINKED ISSUE #7173: Introduce "analysis mode" flag to CLI
## Expected Behavior
I can use rules that use type resolution without passing a classpath explicitly.

## Current Behavior
Type resolution is only enabled when a classpath is passed in.

## Context
detekt assumes that if no classpath is passed in, that type resolution & full analysis isn't possible. This isn't true though - we call `configureJdkClasspathRoots` unconditionally:
https://github.com/JetBrains/kotlin/blob/v1.9.23/compiler/cli/cli-base/src/org/jetbrains/kotlin/cli/jvm/config/JvmContentRoots.kt#L91-L103

It's possible to have a module that has no Kotlin dependencies at all, but depends on the JRE, or it might not have any dependencies at all. This is rare but detekt should properly handle this.

We could have a switch or flag to enable or disable full analysis mode so the intention is clear.

3flex:
Let's be very sure about the naming of this @detekt/maintainers since these terms should be used consistently in code, docs and rule annotations.

Some options:
* light/full (used in this PR - the terms aren't opposites which irks me but, but I guess they don't have to be)
* simple/complete
* simple/full (this one is the most descriptive that I could think of for the two existing modes)

Agreed these are not good options:
* fast
* slow

The Gradle plugin flag is not exposed in the API yet but will be exposed in a future PR.

Closes #7173


detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -24,6 +25,14 @@ class CliArgs {
     )
     var inputPaths: List<Path> = listOf(Path(System.getProperty("user.dir")))
 
+    @Parameter(
+        names = ["--analysis-mode"],
+        description = "Analysis mode used by detekt. 'Full' analysis mode is comprehensive but requires the correct " +
```
TWiStErRob, line 30: Is default shown when running help? If not, please put in description.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
```diff
@@ -159,6 +160,9 @@ abstract class Detekt @Inject constructor(
     @get:Optional
     internal abstract val explicitApi: Property<String>
 
+    @get:Input
+    internal val analysisMode = providers.provider { if (classpath.isEmpty) "light" else "full" }
```
TWiStErRob, line 164: Why not use the enum on these?

TWiStErRob, line 164: Should this be maybe an extension function on classpath, so that there's no code duplication?

I'm thinking like this:

```kotlin
fun ConfigurableFileCollection.asInferredAnalysisMode(): Provider<String> =
	elements.map { if (it.isEmpty()) "light" else "full" }
```

and then replace this and CreateBaseLineTask lines with

```kotlin
internal val analysisMode = classpath.asInferredAnalysisMode()
```

Note: `FileCollection.isEmpty()` resolves the collection immediately (via `.getFiles()`) and requires the explicit wrapping in `provider { }`. `getElements()` returns a provider, so there's no need to use `providers`. I think there might be an added benefit of also [propagating `builtBy` correctly](https://docs.gradle.org/current/javadoc/org/gradle/api/file/FileCollection.html#getElements()), which is lost when doing `.provider { }`

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/AnalysisMode.kt
```diff
@@ -0,0 +1,6 @@
+package io.github.detekt.tooling.api
+
+enum class AnalysisMode {
+    FULL,
+    LIGHT,
```
TWiStErRob, line 5: have you considered [full/lite and weak/strong](https://github.com/detekt/detekt/issues/7173#issuecomment-2186453738)?

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/spec/ProjectSpec.kt
```diff
@@ -16,4 +17,11 @@ interface ProjectSpec {
      * Paths to analyze. Works with files and directories.
      */
     val inputPaths: Collection<Path>
+
+    /**
+     * The analysis mode used by detekt. 'light' mode means detekt will analyze code in individual files but cannot use
+     * compiler information like types, symbols and smart casts. 'full' mode allows rules to use additional information
+     * from the compiler when analyzing code.
```
TWiStErRob, line 24:
```suggestion
     * The analysis mode used by detekt.
     *  * 'light' mode means detekt will analyze code in individual files
     *    but cannot use compiler information like types, symbols and smart casts.
     *  * 'full' mode allows rules to use additional information from the compiler when analyzing code.
```
also, maybe these bullet points are redundant? and should be instead documented in 1 place, on the enum!?

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
```diff
@@ -59,7 +60,13 @@ internal class DefaultLifecycle(
     override val settings: ProcessingSettings,
     override val parsingStrategy: ParsingStrategy,
     override val bindingProvider: (files: List<KtFile>) -> BindingContext =
-        { generateBindingContext(settings.environment, settings.classpath, it, settings::debug) },
```
TWiStErRob, line 62: Is `val EnvironmentAware.classpath` still needed? It looks like this was the only usage.

TWiStErRob (commented):
Nice one!

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
3flex: Yes, it is. Run `./gradlew detekt-cli:runWithHelpArg` in this project to see the help output.

Excerpt:
```
    --analysis-mode
      Analysis mode used by detekt. 'Full' analysis mode is comprehensive but 
      requires the correct classpath, API version and language version to be 
      set. 'Light' analysis cannot utilise compiler information and some rules 
      cannot be run in this mode.
      Default: LIGHT
      Possible Values: [FULL, LIGHT]
```

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
3flex: Because DGP doesn't depend on detekt-tooling, and doing that is blocked (in my opinion) by https://github.com/detekt/detekt/issues/7123.

I don't think we should add any detekt dependencies to DGP unless it's absolutely necessary.

3flex: I'm changing this in follow up PRs anyway so let's just leave the duplication for now if that's OK.

Good call on isEmpty vs getElements, I've added a new commit to fix it.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
TWiStErRob: hmmm, nice. An interesting problem though: based on that I would call `detekt --analysis-mode FULL`. The unit tests use all lower-case ("full"), and the wording above uses camel-case ("Full"), this is confusing.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
3flex: Or did you mean we should create a duplicate enum in DGP module?

TWiStErRob: You can still call `.elements` eagerly on the `Task.val` I think, because it's a provider, and .map is lazy too. But maybe it needs to be `@Internal @Input internal val`.

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/AnalysisMode.kt
3flex: I prefer not to use "lite" over "light". I'm not a big fan of "weak" and "strong". To me, that sounds more like detekt is being told to take a stronger stance on rules, like the way the "fail on severity" feature works.

I think we should keep naming discussion in the main comments on this issue instead of inlined here.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
3flex: It's still required, it's passed to `createCompilerConfiguration`

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/spec/ProjectSpec.kt
3flex: Yeah we could move it - the KDoc here would say "analysis mode used by detekt" then refer to the enum, and these descriptions go on each enum option. I think that makes sense.

I won't change it yet so others can weigh in.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
TWiStErRob: No, I thought it was visible. I would say create an enum iff it's a public DSL, which it is not, because it's implied from classpath.

detekt-tooling/src/main/kotlin/io/github/detekt/tooling/api/spec/ProjectSpec.kt
TWiStErRob: Yep that's what I meant.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
3flex: It's case insensitive so it doesn't impact anything. I could rename the enum so it's lower case (or probably capitalised like the enum entries in `FailureSeverity`).

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
3flex: I've taken a different simpler approach for now, after attempting to use classpath.elements there was some ordering issue with the provider accessing `classpath` before `compileJava` had run (compileJava output is an input to `classpath`).

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
BraisGabin: I would vote for all lowercase. Even in the description.

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/Detekt.kt
TWiStErRob: it was probably config cache resolving it before running tasks

schalkms:
I agree 👍

> I think that light/full is the best option.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
```diff
@@ -24,6 +25,14 @@ class CliArgs {
...
+            "classpath, API version and language version to be set. 'Light' analysis cannot utilise compiler " +
```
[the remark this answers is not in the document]
3flex: Yes they can use language & API version and we can default them to the current embedded Kotlin's version. Or the rule could use that logic itself.

I'll change the "full" description to say that the correct compiler options have to be set and remove references to specific parameter names. More comprehensive documentation should then be provided on our website.

detekt-cli/src/main/kotlin/io/gitlab/arturbosch/detekt/cli/CliArgs.kt
3flex: Updated:
```
    --analysis-mode
      Analysis mode used by detekt. 'full' analysis mode is comprehensive but 
      requires the correct classpath, API version and language version to be 
      set. 'light' analysis cannot utilise compiler information and some rules 
      cannot be run in this mode.
      Default: light
      Possible Values: [full, light]
```

TWiStErRob (approved):

BraisGabin (approved):

OksiBlack:
Just as a variants

deep vs quick
in depth vs quick
complete vs quick

detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/invoke/CliArgument.kt
```diff
@@ -56,10 +57,12 @@ internal data class ClasspathArgument(val fileCollection: FileCollection) : CliA
...
         listOf(
             CLASSPATH_PARAMETER,
-            fileCollection.files.filter(File::exists).joinToString(File.pathSeparator) { it.absolutePath }
+            fileCollection.files.filter(File::exists).joinToString(File.pathSeparator) { it.absolutePath },
+            ANALYSIS_MODE,
+            "full",
         )
     } else {
-        emptyList()
+        listOf(ANALYSIS_MODE, "light")
```
OksiBlack, line 65: Maybe it is better to make full the default? And explicitly opt out of it in case it not needed?

TWiStErRob: Not possible because full has a hard dependency on classpath, sadly that can't be made up from thin air.

TWiStErRob: Depending on which task `detekt` (no classpath) / `detekt<Variant>` (has classpath) is called by the user, the right mode will be selected automatically, at least in Gradle.

cortinico:
I arrived late here (my bad), but I wanted to mention how maybe `lite` plays better than `light`. 

It's 4 char, same as `full` so it will "align" better in the codebase (you know what I mean), and it just sounds better to me. Not a strong opinion though @3flex

TWiStErRob:
That's why I was proposing it, I thought it was obvious :)

Add to benefits: how many types have we typed ligth/weihgt/heihgt/depht and the like?

cortinico:
> Add to benefits: how many types have we typed ligth/weihgt/heihgt/depht and the like?

So true!

PR #7496, opened 2024-07-29
Duplicate `Entity` so it can be different in `CodeSmell` and `Issue`

BraisGabin:
This PR is just a refactor duplicating the `Entity` class. Right now it doesn't provide any value. But it will! In next PRs this classes that I'm duplicating will start to change. Examples:

- Remove `KtElement` from it
- The path inside Issue.Entity.Location will be relative (instead of absolute). And that will be ensure in the class construction.


BraisGabin:
@detekt/maintainers a review here would help a lot. I have some PRs depending on this and once all of those are merged we can start to implement the serializable output.

schalkms (approved):

PR #7500, opened 2024-07-29
Create `FakeCompilerResources` and add `CompilerResources` as parameter on `Rule.lint` extensions

### LINKED ISSUE #7485: Easily test functions that depend on language version (and similar)
## Expected Behavior
It's easy to test rules that depends on language version and similar values

## Current Behavior
It is not. Just check #7430 for an example

## Context 
This cases should be easy to test and ensure that the tests are focused on what they want to test. Right now nearly all the code just implement infrastructure to tests what they want to test.

BraisGabin:
Closes #7485

Probably this is not still perfect but it's a step forward to have an easier way to test a Rule that uses the `CompilerResources`.

I think that it's not perfect because the `LanguageVersionSettings` from an `env` and the one that we pass as parameter can be different instances and probably they should be the same. But right now I don't know how to configure `env`.

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantVisibilityModifierSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/RedundantVisibilityModifierSpec.kt
@@ -1,18 +1,11 @@
 package io.gitlab.arturbosch.detekt.rules.style
 
 import io.github.detekt.test.utils.compileContentForTest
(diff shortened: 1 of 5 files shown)


BraisGabin:
> Closes #7485
> 
> Probably this is not perfect but it's a step forward to have an easier way to test a Rule that uses the `CompilerResources`.
> 
> I think that it's not perfect because the `LanguageVersionSettings` from en `env` and the one that we pass as parameter are different instances and probably they should be the same. But right now I don't know how to configure `env`

cortinico (approved):

PR #7503, opened 2024-07-30
`LongParameterList` uses `BindingContext` so it needs to add `@RequiresTypeResolution`

BraisGabin:
I'm not 100% sure about this. `LongParameterList` uses the `bindingContext` but only for the excluder so I'm not sure about how to procede. I see two options:

- This one: `LongParameterList` only works with type resolution
- Remove the Type Resolution from the Excluder. This will make the support of `excludeAnnotation` a bit worst but I think that it will still be good enough.


cortinico (approved):
> This one: LongParameterList only works with type resolution

+1 for this approach 👍

PR #7504, opened 2024-07-31
Use correct lint function

BraisGabin:
There were places where we were using `compileAndLint` that we should use `compileAndLintWithContext`, other that use `lint` that could use `compileAndLint` and similar cases. This PR fixes a lot of them (probably this is not exhaustive but at least it's better than before).

This PR will also help to move from `@RequireTypeSolving` as an annotation to an interface.


schalkms (approved):
Great housekeeping!! 👍

cortinico:
Thanks for doing this @BraisGabin

PR #7508, opened 2024-08-02
RequireTypeResolution now is an interface instead of an annotation

### LINKED ISSUE #7492: Ensure that all the rules define correctly if they need type solving/full mode
## Expected Behavior
It's easy to find out if your rule is missing the requires type solving annotation. 

## Current Behavior
Right now it's really difficult to know that. We have a rule to detect it but it's not bullet proof. 

A solution for this would be that the get of the property `bindingContext` in the rule will crash at any attempt to read its value if the rule doesn't have the annotation. To be clear, I'm talking about the behavior here. We doesn't need to check that for each read.

There is another solution: change the annotation to an interface. The rule will not have a `bindingContext` property. The interface will have that property. This way you would find out this limitation at compilation time instead of run time. Also this will avoid to use reflection to check if the rule has or not an annotation (now I'm talking at the `detrkt-core` level). A simple `is` will work. The downside is that we lose discoberability. With the first solution, a user with little experience will find the `bindingContext`. S/he will see an exception telling her/him to add the annotation as soon as the code is executed. With this solution you need to read documentation to find out that there is a magic annotation that gives you access to the `bindingContext`.

BraisGabin:
Closes #7492

For full context please read the linked issue.


config/detekt/argsfile
```diff
@@ -17,3 +17,4 @@ txt:./build/detekt-report.txt
 sarif:./build/detekt-report.sarif
 -bp
 .
+--auto-correct
```
cortinico, line 20: Why this?

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RequiresTypeResolution.kt
```diff
@@ -1,11 +1,7 @@
...
- * The detekt core will honor this annotation and it will not run any rule with this annotation if the bindingContext
- * is empty.
- */
-@Target(AnnotationTarget.CLASS)
-@Retention(AnnotationRetention.RUNTIME)
-annotation class RequiresTypeResolution
+import org.jetbrains.kotlin.resolve.BindingContext
+
+interface RequiresTypeResolution {
+    var bindingContext: BindingContext
```
cortinico, line 6: How do you feel about moving this to a `val` and instead of letting users use `lateinit` var do something like this:

```
override val bindingContext: BindingContext by lazy { bindingContextProvider() }
```

The `bindingContextProvider` can then be a function inside the `Rule` class with a field that we populate before we `visitFile`, only if the rule implements `RequiresTypeResolution`

config/detekt/argsfile
BraisGabin: My bad. This shouldn't be committed.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RequiresTypeResolution.kt
BraisGabin: I didn't implement exactly your idea but this solution doesn't need `lateinit var` either.

The problem of this solution is that it's not that clear that if you implements `RequiresTypeSolving` you will get access to `bindingContext`.

cortinico: Yeah I see. Still a step in the right direction rather than `var bindingContext`. Curious to hear what other thinks about this

BraisGabin: I also tried your option with 
```kotlin
override val bindingContext: BindingContext by lazy { bindingContextProvider() }
```

The problem is that anyone could use `bindingContextProvider()` without implementing `RequiresTypeResolution` and that would break the main idea behind this PR.

cortinico: > without implementing RequiresTypeResolution and that would break the main idea behind this PR.

We won't be populating the backing field of that function though no? So calling `bindingContextProvider()` on a Rule that does not implement `RequiresTypeResolution` results in a crash.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/RedundantSuspendModifier.kt
```diff
@@ -52,13 +52,13 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
...
-    config,
-    "The `suspend` modifier is only needed for functions that contain suspending calls."
-) {
-
+class RedundantSuspendModifier(config: Config) :
+    Rule(
+        config,
+        "The `suspend` modifier is only needed for functions that contain suspending calls."
+    ),
+    RequiresTypeResolution {
```
3flex, line 61: nit: Should have a blank line between opening brace and class body, or put opening brace in its own line (see the last couple of examples here: https://kotlinlang.org/docs/coding-conventions.html#class-headers)

This prompted a request for a rule enforcing this in ktlint.

It's not a must to adjust in this PR but it looked a bit off so did some sleuthing for recommendations.

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RequiresTypeResolution.kt
```diff
@@ -1,11 +1,3 @@
 package io.gitlab.arturbosch.detekt.api
 
-/**
- * Annotated [Rule] requires type resolution to work.
- *
- * The detekt core will honor this annotation and it will not run any rule with this annotation if the bindingContext
- * is empty.
- */
```
3flex, line 8: We should keep some documentation here.

BraisGabin:
I'll address the comments soon but I'm moving it to draft as this is clearly not ready yet to be merge.

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/RedundantSuspendModifier.kt
BraisGabin: Fixed

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RequiresTypeResolution.kt
BraisGabin: Done

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/RequiresTypeResolution.kt
BraisGabin: What do you think about the current implementation? I think that it have the good parts of both sides. To read the `contextBinding` you need to be inside the context of `RequiresFullAnalysis`. So you need to implement `RequiresFullAnalysis` to access it (or do really odd tricks like `with(object : RequiresFullAnalysis {}) { contextBinding }` that will compile but will crash.

I think that this option keeps the type safety without forcing the rules to have any new `var`

BraisGabin:
I fixed the conflicts (there were a ton) and the comments. It is ready to review again :)

cortinico (approved):
I haven't checked each rule one by one, but the code inside core looks legit 👍

cortinico:
This is awesome btw :)

PR #7513, opened 2024-08-03
Remove `KtElement` from `Issue`

BraisGabin:
After all the refactors we can remove `KtElement` form `Issue`. This was a blocker for #6879. Now `Issue` should be easily serailizable (but we need to implement it).


BraisGabin:
I would like to have a review here. This will unlock the option to make the output of detekt serializable.

cortinico (approved):

PR #7514, opened 2024-08-03
Remove exclude autocorrect

BraisGabin:
This was never fully implemented and no one complained about it. Also I plan to change the autocorrect information in the `CodeSmell` for a new `suppress` that tells you why an issue is suppressed, for example, because it is autocorrected.

For now I'm just removing the `excludeAutocorrect` and enable it by default. You will not see autocorrected issues on the reports. As soon as I implement the `suppress` a custom report could decide to show those.


BraisGabin:
My bad, those tests are usually flaky on my device (I don't know why) so I didn't care about them. But this time I was changing exactly what they test.

Now it's green :)

cortinico (approved):

Issue #3665, opened 2021-04-08
There is no built-in way to use the detekt Gradle plugin on common sources in Kotlin multi-platform projects


utybo:
## Expected Behavior

Any of the following:

* Running a platform specific Detekt task (e.g. `detektJvmMain`, `detektJsMain`...) also checks the common sources (or other project source code the platform code requires)
* There is a `detektCommonMain` task
* The `detekt` task checks the `commonMain` sources

## Observed Behavior

None of the above.

* Platform specific tasks only run checks on their source code (e.g. [this run](https://github.com/utybo/detekt-kotlin-mpp/runs/2295387257#step:5:20) and [this run](https://github.com/utybo/detekt-kotlin-mpp/runs/2295387373#step:5:14), both only report 1 error in their own code but they do not report [this empty-bodied function in the common sources](https://github.com/utybo/detekt-kotlin-mpp/blob/0721c33e626d07c90a7880a79413dc2f1253c928/src/commonMain/kotlin/CommonCode.kt#L11))
* There is no `detektCommonMain` task
* The `detekt` task does not do anything (see also #3663)

## Steps to Reproduce

* Create a Kotlin MPP project
* Try to run detekt only on common source code

## Context

This makes using Detekt with no user-configuration useless in multi-platform projects as most code will be in the `commonMain` source set. This forces users to manually create a task that also checks for those, which is not fantastic.

A [repro repo (that also highlights other issues)](https://github.com/utybo/detekt-kotlin-mpp) is available. I encountered this while trying to set up Detekt for [one of my projects](https://github.com/utybo/Lixy).

A possible workaround is to manually configure the `detekt` task or a custom task and give it the inputs we want (e.g. JVM + common or JS + common).

## Your Environment

* Version of detekt used: 1.16.0
* Version of Gradle used (if applicable): 6.8.3
* Gradle scan link (add `--scan` option when running the gradle task): See the CI links mentioned above.
* Operating System and version: Windows 10, 1903
* Link to your project (if it's a public repository): https://github.com/utybo/detekt-kotlin-mpp

cortinico:
Thanks for the reproducer.

>     * There is a `detektCommonMain` task

There is actually a `detektMetadataMain` task that will run the analysis only on the common module.
The name is unfortunate, but it turns out that the KMP plugin exposes a target called `metadata` with a compilation called `main`, therefore the task name gets composed from those.

I see that in your project the task works correctly and finds an issue:

```
gw detektMetadataMain

[...]

PR #7532, opened 2024-08-11
Add KMP note to type resolution docs

hbmartin:
Follow up to https://github.com/detekt/detekt/issues/3665

As an alternative, this seems interesting: https://github.com/detekt/detekt/issues/6634#issuecomment-1996012038


schalkms (approved):

BraisGabin (commented):
Should this say something about android? It's similar than jvm, I guess, but it is not exactly the same.

BraisGabin (approved):

BraisGabin:
Thanks for the contribution!

PR #7535, opened 2024-08-12
Add tests for isolated projects

technoir42:


3flex (changes requested):
Please add assertions for each task execution that checks that the build is successful, thanks!

technoir42:
@3flex as far as I can tell, `runTasks` delegates to `GradleRunner.build()` which internally asserts that the build is successful. I followed the same pattern as in `ConfigurationCacheSpec`.

3flex (approved):

PR #7574, opened 2024-08-25
Update Kotlin and poko at the same time

BraisGabin:
I did 3 different things here but the important one is related with this comment: https://github.com/detekt/detekt/pull/7568#issuecomment-2308213784

- Move `.renovaterc.json` to `.github/` we have way too many files inside our root so anything that we can move outside is a win ([doc here](https://docs.renovatebot.com/configuration-options/))
- Add the `$schema`, it helps to autocomplete the configuration file.
- If there are updates of kotlin AND poko at the same time they will be merged in the same PR. This will fix problems like #7568. If there is only a update on poko or on kotlin that's not a problem. Those PRs will appear the same as before but now if they happen at the same time they will be merged.


3flex (approved):
This will definitely make things easier going forward, thanks!

Issue #7038, opened 2024-03-11
Allow ForbiddenSuppress to disallow suppressing ForbiddenSuppress


mgroth0:
According to the [docs](https://detekt.dev/docs/next/rules/style/#forbiddensuppress) and [PR](https://github.com/detekt/detekt/pull/4899), `ForbiddenSuppress` cannot disallow supression of itself. 

## Expected Behavior

This should be allowed, because otherwise there is no way to strictly enforce a rule.

## Current Behavior

There is no way to strictly enforce a rule.

## Context

If `ForbiddenSuppress` cannot disallow supressing itself, then as the PR says this rule does not really forbid any suppression, it just discourages it.

I do not know if this limitation is by design, because of a technical limitation, or both.

If it is only by design, I would argue that the developer should be allowed to strictly enforce a rule in their project by forbidding suppression of it completely. Its a decision for the developer, not for this library to make.

If it is a technical limitation, then I would propose that a workaround is implemented. One idea for a workaround, if required, is that this could be a special option for the detekt engine itself, and not considered a regular rule. That way, the suppresion mechaism doesn't take precedence.

One idea for how it could look in the yaml is this: 

```yaml
build:
    forbidSuppressingForbiddenSuppression: true
```

I personally find the tongue twister fun, but it could of course be replaced by something simpler such as `forbidSuppressionsStrictly`.

The point is that if there is a technical limitation, a special configuration option could be added that is not technically part of any rule, but is an option on detekt itself. This special option would report issues just like a normal detekt rule, but it would not itself be suppressible. It would still be easy to disable by setting `forbidSuppressingForbiddenSuppression: false`, but this would be up to the developer. While it might sound radical to bake a issue-reporter into detekt itself without being in the form of a rule, I think this would be the single case where it might be justified.

cortinico:
> I do not know if this limitation is by design, because of a technical limitation, or both.

I believe this was due to a technical limitation.
We would have to add an exception inside the Detekt's core suppression logic to ignore `ForbiddenSuppress` suppressions at all

BraisGabin:
I see two options to implement this:

1. Add the option on the configuration to make a rule "no-suppressable" so it will skip the `@Suppress()` step check if it is enabled.
2. Nico's idea: hardcode inside the core of detekt that `ForbiddenSuppress` should skip that step.

I vote for the first option. With the current state of `core` this should be kind of easy to accomplish. All the logic that needs to change should be in the `Analyzer`.

[...]

PR #7590, opened 2024-08-28
Make ForbiddenSuppress un-suppressable

mgroth0:
Fixes https://github.com/detekt/detekt/issues/7038

Of the [5 options](https://github.com/detekt/detekt/issues/7038#issuecomment-2027003939) listed by @BraisGabin, this is an implementation of option 4. In the issue, there were basically two votes for option 3 and two votes for option 4. If I added my vote, I would have said any of options 1-4 would work (just not 5, since it doesn't fix this issue). I gave option 4 a shot because it seemed the easiest for me.

This PR gets the job done and resolves the issue, but there are some questionable aspects of the implementation. I would ask about these in particular:

- `isForbiddenSuppress` is possibly not robust in that it checks the class based on the class name. This is because I needed to implement the logic in `core`, which can't do an `is` type check for `ForbiddenSuppress`. One alternative I though of would be to add an additional `suppressable` property to the `Rule` class, but I though that would be bad because this would allow other rules to become unsurpassable outside of user configuration. Given that the only alternative is bad, I think that checking based on classname might ok. At least it is reinforced by the tests in `ForbiddenSuppressSpec`. If the classname or package of `ForbiddenSuppress` was ever refactored, those tests should fail.

- The documentation edit is now a bit wordy and can be confusing. Once we finalize the implementation, maybe it can be re-worded in a simpler way.

- Suppressors now do not work at all on ForbiddenSuppress. This matches the original goal, however the annotation and function suppresser configuration properties are still present in the configuration spec and are valid yaml. Not sure the best way to handle this, but maybe the verification of the yaml could do a special check to ensure that these properties are not defined for ForbiddenSuppress in the yaml file. Alternatively, maybe we could somehow remove the suppresser properties completely for the ForbiddenSuppress configuration, but I am not sure how.

- I tried to run `detekt` and `test` tasks before submitting this PR, but got failures in both that were unrelated to this PR. I read the CONTRIBUTING guide and the first step says to run `./gradlew publishToMavenLocal`, but I don't want to do this because that might affect other projects on my machine. So I'm unsure if that is why my `detekt` and `test` tasks are failing in other ways.

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
@@ -19,6 +19,10 @@ internal fun annotationSuppressorFactory(rule: Rule, bindingContext: BindingCont
         it.qualifiedNameGlobToRegex()
     }
     return if (annotations.isNotEmpty()) {
+        if (rule.isForbiddenSuppress()) {
+            warnForbiddenSuppressCannotBeSuppressed()
+            return null
+        }
         Suppressor { finding ->
             val element = finding.entity.ktElement
             element.isAnnotatedWith(AnnotationExcluder(element.containingKtFile, annotations, bindingContext))
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
@@ -20,7 +20,6 @@ internal fun annotationSuppressorFactory(rule: Rule, bindingContext: BindingCont
     }
     return if (annotations.isNotEmpty()) {
         if (rule.isForbiddenSuppress()) {
-            warnForbiddenSuppressCannotBeSuppressed()
             return null
         }
         Suppressor { finding ->
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
@@ -19,9 +19,6 @@ internal fun annotationSuppressorFactory(rule: Rule, bindingContext: BindingCont
         it.qualifiedNameGlobToRegex()
     }
     return if (annotations.isNotEmpty()) {
-        if (rule.isForbiddenSuppress()) {
-            return null
-        }
         Suppressor { finding ->
             val element = finding.entity.ktElement
             element.isAnnotatedWith(AnnotationExcluder(element.containingKtFile, annotations, bindingContext))
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
@@ -25,7 +25,6 @@ internal fun functionSuppressorFactory(rule: Rule, bindingContext: BindingContex
         .map(FunctionMatcher::fromFunctionSignature)
     return if (functionMatchers.isNotEmpty()) {
         if (rule.isForbiddenSuppress()) {
-            warnForbiddenSuppressCannotBeSuppressed()
             return null
         }
         Suppressor { finding ->
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
@@ -24,6 +24,10 @@ internal fun functionSuppressorFactory(rule: Rule, bindingContext: BindingContex
     val functionMatchers = rule.config.valueOrDefault("ignoreFunction", emptyList<String>())
         .map(FunctionMatcher::fromFunctionSignature)
     return if (functionMatchers.isNotEmpty()) {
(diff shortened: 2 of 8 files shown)


cortinico:
Thanks for the PR @mgroth0. The CI is still red, can you follow up here?

cortinico:
> @cortinico I don't know how to solve this `println` issue. I tried to pass a reference to the logger (in `ProcessingSettings`) but adding this parameter is a massive change that requires it to be passed in hundreds of places in the project.

You should definitely not use the `println` function as we want to use the logger. 
I would not create an extension function on `Rule` though and just handle the print inside the core logic (where you already have a hold of a logger).

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/Suppressions.kt
```diff
@@ -1,30 +1,44 @@
...
 
+import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.RuleSet
 import org.jetbrains.kotlin.psi.KtAnnotated
 import org.jetbrains.kotlin.psi.KtAnnotationEntry
 import org.jetbrains.kotlin.psi.KtElement
 import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 import kotlin.text.RegexOption.IGNORE_CASE
 
+internal fun Rule.isForbiddenSuppress() = this::class.qualifiedName == FORBIDDEN_SUPPRESS_QNAME
```
BraisGabin, line 11: You can implement this using the ruleDescriptor. The rule name inside it gives you the name of the rule so you can implement the same. This way you don't need to pass a new parameter.

BraisGabin:
I like option 1. And I will probably don't show any warning. Or at least not for now. If we receive a lot of issues about this we could add it but I think that, by default, we don't need the warning.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/AnnotationSuppressor.kt
```diff
@@ -19,6 +19,9 @@ internal fun annotationSuppressorFactory(rule: Rule, bindingContext: BindingCont
         it.qualifiedNameGlobToRegex()
     }
     return if (annotations.isNotEmpty()) {
+        if (rule.isForbiddenSuppress()) {
+            return null
+        }
```
BraisGabin, line 24:
```suggestion
```

No need. If someone say on the detekt configuration fail that they want to suppress all the errors of `ForbiddenSuppress` on the functions annotated with `@Composable` that's their problem. I agree that adding an `@Suppress("ForbiddenSuppress")` can be "accidental" or could not be cached when doing a review. But if you change the configuration of a rule we should honor that.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/FunctionSuppressor.kt
```diff
@@ -24,6 +24,9 @@ internal fun functionSuppressorFactory(rule: Rule, bindingContext: BindingContex
     val functionMatchers = rule.config.valueOrDefault("ignoreFunction", emptyList<String>())
         .map(FunctionMatcher::fromFunctionSignature)
     return if (functionMatchers.isNotEmpty()) {
+        if (rule.isForbiddenSuppress()) {
+            return null
+        }
```
BraisGabin, line 29:
```suggestion
```

Same as before.

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/Suppressions.kt
```diff
@@ -1,12 +1,18 @@
...
+import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.RuleSet
+import io.gitlab.arturbosch.detekt.core.extractRuleName
 import org.jetbrains.kotlin.psi.KtAnnotated
 import org.jetbrains.kotlin.psi.KtAnnotationEntry
 import org.jetbrains.kotlin.psi.KtElement
 import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 import kotlin.text.RegexOption.IGNORE_CASE
 
+internal fun Rule.isForbiddenSuppress() = this.ruleName == FORBIDDEN_SUPPRESS_NAME
```
BraisGabin, line 12:
```suggestion
```

Not used

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/suppressors/Suppressions.kt
```diff
@@ -18,13 +24,19 @@ fun KtElement.isSuppressedBy(id: String, aliases: Set<String>, ruleSetId: RuleSe
...
         .mapNotNull { it.getArgumentExpression()?.text }
         .map { it.replace(detektSuppressionPrefixRegex, "") }
         .map { it.replace(QUOTES, "") }
         .any { it in acceptedSuppressionIds }
+
+    if (r && isForbiddenSuppressById(id)) {
+        return false
+    }
+
+    return r
```
BraisGabin, line 39:
```suggestion
    if (isForbiddenSuppressById(id)) return false

    return allAnnotationEntries()
        .filter { it.typeReference?.text in suppressionAnnotations }
        .flatMap { it.valueArguments }
        .mapNotNull { it.getArgumentExpression()?.text }
        .map { it.replace(detektSuppressionPrefixRegex, "") }
        .map { it.replace(QUOTES, "") }
        .any { it in acceptedSuppressionIds }
```

BraisGabin (approved):

BraisGabin:
Sorry for the slow review. I did some changes (see comments above) and I'm going to merge this directly. If you don't agree with any of my changes we can talk about it on other issue or you can open a new PR with the changes that you will add now. I'm doing this just to get the things moving and merge this fast.

Issue #3749, opened 2021-05-06
Allow configuration to force processing of rules on methods marked @Suppress or @SuppressWarning


JeffHoover:
## Expected Behavior
If `ignoreSupress: true` is in the `config:` section of the configuration file, _process rules on all methods_, even methods marked with @ Suppress

if `ignoreSupressWarning: true` is in the `config:` section of the configuration file, _process rules on all methods_, even methods marked with @ SuppressWarning

## Current Behavior
(As designed) when a method is marked with @ Suppress or @ SuppressWarning, the specified rule is not reported for that method

## Context
I've been asked to analyze code that I don't own. I'd like a complete picture of results, not just the results that the author(s) want to see.

BraisGabin:
I like this idea, it have sense. Even for your own code you may want to revisit all those supresed warnings from time to time.

I think that it should be only one configuration instead of two and it should disable all the suppression: annotations and baseline too.

cortinico:
Agree, that's really valuable to track.
I was also thinking that right now we just "discard" a report if a `@Suppress` is discovered. Instead we could carry over this information on the report, and display a note like "X issues were identified but not reported as they were suppressed in the baseline/locally, etc."

BraisGabin:
Good idea! Probably that's a big breaking change but we should do that for 2.0. I imagine the API like this:

```kotlin
class Finding(
  ...
  val suppressedReasons: List<String>
) {
  val suppressed: Boolean
    get() = suppressedReasons.isNotEmpty()
}
```

BraisGabin:
I was working on this issue and was thinking what to do with the suppressors. Even with that name we should probably don't allow to inspection them. If I configure a rule to don't check functions with `@Composable` annotation we shouldn't consider that like a "suppresion", it's part of the rule configuration. It's similar to say "don't complain about the name of overrided functions". We could consider that a kind of "suppression" but it have really little sense.

BraisGabin:
This is done already. The reports get all the issues found by detekt. Then, depending on the report, it filter out the ones suppressed or not.

PR #7591, opened 2024-08-28
Introduce suppressReasons

BraisGabin:
First step to implement #3749. The next step is to use this for the suppression and the baseline.


detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/CodeSmell.kt
```diff
@@ -10,7 +10,8 @@ package io.gitlab.arturbosch.detekt.api
 open class CodeSmell(
     final override val entity: Entity,
     final override val message: String,
-    final override val references: List<Entity> = emptyList()
+    final override val references: List<Entity> = emptyList(),
+    final override val suppress: List<String> = emptyList(),
```
cortinico, line 14: nit: I think from the name is not immediately clear what this field is about.

I think it should be like `suppressions` or something more descriptive.

BraisGabin: You are completely right. The name doesn't explain what it is. What do you think about `suppressReasons`? I'm completely fine with your proposal. I'm just thinking about other options.

cortinico: Yeah that would work 👍

BraisGabin: Rename done :)

cortinico (approved):

PR #7592, opened 2024-08-28
Remove `ThresholdedCodeSmell`

BraisGabin:
Right now `CodeSmell` is an `open class` and I will like to remove the `open` from it. #7591 removes `CorrectableCodeSmell` and `ThresholdedCodeSmell` was the other extension of `CodeSmell`.

This is the "short" implementation for this. Just remove the "extra" features of `ThresholdedCodeSmell` because, I think, no one uses them. There is another option. I plan to add a way to add extra information inside a `CodeSmell`. Something like:

```kotlin
CodeSmell(
  ...
  extra: Map<String, String>(),
)
```

So I could replace `Metric(value = 5, threshold = 4)` for something like: `extra = mapOf("value" to "5", "threshold" to "4")`. But I think that this change will give little value (replacing Metrics instead of removing it).


detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CyclomaticComplexMethodSpec.kt
```diff
@@ -84,25 +84,25 @@ class CyclomaticComplexMethodSpec {
         @Test
         fun `counts three with nesting function 'forEach'`() {
             val config = TestConfig(defaultAllowedComplexity, "ignoreNestingFunctions" to "false")
-            assertExpectedComplexityValue(code, config, expectedValue = 3)
+            assertExpectedComplexityValue(code, config)
         }
```
cortinico, line 88: Now you have `three` in the test title, but not in the test body.

detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CyclomaticComplexMethodSpec.kt
```diff
@@ -297,13 +297,10 @@ class CyclomaticComplexMethodSpec {
     }
 }
 
-private fun assertExpectedComplexityValue(code: String, config: TestConfig, expectedValue: Int) {
+private fun assertExpectedComplexityValue(code: String, config: TestConfig) {
```
cortinico, line 300: This should probably be renamed now

cortinico (changes requested):

detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CyclomaticComplexMethodSpec.kt
BraisGabin: I added back this assert. I can still assert the same thing using the finding message.

detekt-rules-complexity/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/complexity/CyclomaticComplexMethodSpec.kt
BraisGabin: I reverted this change.

cortinico (approved):

PR #7599, opened 2024-08-30
[1.x] Bump Kotlin to 2.0.20

cortinico:
This is essentially https://github.com/detekt/detekt/pull/7580 but against the `release/1.x` branch.

### DIFF
--- a/detekt-api/api/detekt-api.api
+++ b/detekt-api/api/detekt-api.api
@@ -636,8 +636,6 @@ public final class io/gitlab/arturbosch/detekt/api/ValueWithReason {
 }
 
 public final class io/gitlab/arturbosch/detekt/api/ValuesWithReason : java/lang/Iterable, kotlin/jvm/internal/markers/KMappedMarker {
-	public final fun copy (Ljava/util/List;)Lio/gitlab/arturbosch/detekt/api/ValuesWithReason;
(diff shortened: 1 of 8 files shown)


3flex:
There are multiple issues leading to CI failures on this branch, I see at least three unique issues, and that's not even including the failures that we've seen in the 2.0.20 PR on main branch which aren't resolved (I'm hoping for a 2.0.21 with a fix for the issue I raised but there's been no progress yet).

Suggest we leave this branch on 2.0.10 at least for now. If anyone really wants support to be added they can submit a PR with fixes. I don't plan to work on the 1.x branch.

cortinico:
> There are multiple issues leading to CI failures on this branch, I see at least three unique issues, and that's not even including the failures that we've seen in the 2.0.20 PR on main branch which aren't resolved (I'm hoping for a 2.0.21 with a fix for the issue I raised but there's been no progress yet).
> 
> Suggest we leave this branch on 2.0.10 at least for now. If anyone really wants support to be added they can submit a PR with fixes. I don't plan to work on the 1.x branch.

Yeah I've realized it as well. The problem is that I pick commits on the 1.x branch but we never run CI there till we actually open a PR. Then all the problems we fixed on `main` surface on CI signals.

I'd say let's close this one as you suggested as it looks unlikely we'll be able to merge it now for the same reason this PR is blocked on `main`.

I'll follow-up on making the CI green also on this branch before releasing.

Issue #7598, opened 2024-08-30
UnnecessaryInnerClass false positive when inner class references another inner class 


matejdro:
## Expected Behavior

`UnnecessaryInnerClass` should not be raised

## Observed Behavior

`UnnecessaryInnerClass` is raised 

## Steps to Reproduce
Perform a detekt check on the following code:

```
class A(val input: Int) {
   inner class B() {
      val c = C()
   }

   inner class C() {
      val processedInput = input + 1
   }
}
```

detekt will report that `inner` on `B` is unnecessary.

## Context

Removing `inner` from the class `B` in above example will cause code to not compile, since class `B` would no longer have access to class `C`

## Your Environment

* Version of detekt used: 1.23.6
* Version of Gradle used (if applicable): 8.9
* Operating System and version: Linux 6.10.6-arch1-1

Sanu5:
we can supress the UnnecessaryInnerClass
@Suppress("UnnecessaryInnerClass")
    inner class ReportsUnnecessaryInnerClassRule

atulgpt:
I think more or less this TCs should pass
```kotlin
@Nested
inner class `inner reffering other inner class` {
    @Test
    fun `reports with nested empty classes`() {
        val code = """
        class A {
            inner class B {
                inner class C
            }
        }
    """.trimIndent()

        val findings = subject.compileAndLintWithContext(env, code)
        assertThat(findings)
            .hasSize(2)
            .map(Finding::message)
            .contains(
                tuple("Class 'C' does not require `inner` keyword."),
                tuple("Class 'B' does not require `inner` keyword."),
            )
    }

    @Test
    fun `reports when only outer ctor is used`() {
        val code = """
        class A {
            inner class B {
                val a = A()
            }
        }
    """.trimIndent()

        val findings = subject.compileAndLintWithContext(env, code)
        assertThat(findings)
            .hasSize(1)
    }

    @Test
    fun `does not report when inner class refers other inner class which is required`() {
        val code = """
            class A {
                val foo = "BAR"
                inner class B {
                    val c = C()
                }

                inner class C(val fizz: String = foo)
            }
    """.trimIndent()

        assertThat(subject.compileAndLintWithContext(env, code)).isEmpty()
    }

    @Test
    fun `does not report when inner class refers other inner class in init block`() {
        val code = """
            class A {
                val foo = "BAR"
                inner class B {
                    init {
                        val c = C()
                    }
                }

                inner class C(val fizz: String = foo)
            }
[...]

PR #7626, opened 2024-09-12
fixed the UnnecessaryInnerClass false positive

Sanu5:
This PR fixes the false positive for the UnnecessaryInnerClass rule by suppressing the rule in cases where it incorrectly flags valid inner classes. Specifically, the rule was triggering for inner classes where access to outer class members is necessary for compilation, but removing inner caused compilation errors.

Fix Details:
Suppressed the UnnecessaryInnerClass rule in test cases where it was falsely triggered.


cortinico:
That's not the right fix. You should fix the rule code to fully address https://github.com/detekt/detekt/issues/7598, not just the tests.

Issue #7612, opened 2024-09-08
Fix broken anchors on detekt.dev


cortinico:
## Expected Behavior
Our website builds without warnings/errors.

## Observed Behavior
The following link with anchors are broken:

```
[WARNING] Docusaurus found broken anchors!

Please check the pages of your site in the list below, and make sure you don't reference any anchor that does not exist.
Note: it's possible to ignore broken anchors with the 'onBrokenAnchors' Docusaurus configuration, and let the build pass.

Exhaustive list of all broken anchors found:
- Broken anchor on source page path = /blog/2019/08/04/calling-it-one-zero:
   -> linking to #with-gradle (resolved as: /blog/2019/08/04/calling-it-one-zero#with-gradle)
- Broken anchor on source page path = /marketplace/:
   -> linking to #unpublished (resolved as: /marketplace/#unpublished)
- Broken anchor on source page path = /docs/1.22.0/gettingstarted/gradle:
   -> linking to #closure (resolved as: /docs/1.22.0/gettingstarted/gradle#closure)
- Broken anchor on source page path = /docs/1.22.0/intro:
   -> linking to #with-gradle (resolved as: /docs/1.22.0/intro#with-gradle)
- Broken anchor on source page path = /docs/1.22.0/rules/complexity:
   -> linking to /docs/rules/complexity#cyclomaticcomplexmethod
- Broken anchor on source page path = /docs/1.23.0/gettingstarted/gradle:
   -> linking to #closure (resolved as: /docs/1.23.0/gettingstarted/gradle#closure)
- Broken anchor on source page path = /docs/1.23.0/intro:
   -> linking to #with-gradle (resolved as: /docs/1.23.0/intro#with-gradle)
- Broken anchor on source page path = /docs/1.23.0/rules/complexity:
   -> linking to /docs/rules/complexity#cyclomaticcomplexmethod
- Broken anchor on source page path = /docs/1.23.1/gettingstarted/gradle:
   -> linking to #closure (resolved as: /docs/1.23.1/gettingstarted/gradle#closure)
- Broken anchor on source page path = /docs/1.23.1/intro:
   -> linking to #with-gradle (resolved as: /docs/1.23.1/intro#with-gradle)
- Broken anchor on source page path = /docs/1.23.1/rules/complexity:
   -> linking to /docs/rules/complexity#cyclomaticcomplexmethod
- Broken anchor on source page path = /docs/1.23.3/gettingstarted/gradle:
[...]

PR #7630, opened 2024-09-13
Fix broken anchors on detekt.dev

ahmdaa:
Fixes https://github.com/detekt/detekt/issues/7612

This fixes all the broken anchors on detekt.dev

### DIFF
--- a/website/blog/2019-08-04-calling-it-one-zero.md
+++ b/website/blog/2019-08-04-calling-it-one-zero.md
@@ -20,7 +20,7 @@ programming flaws like unused or too complex constructs. Think of it as *pmd* or
 - Suppression of findings with Kotlin's `@Suppress` and Java's `@SuppressWarnings` annotations
 - Specification of quality gates which will break your build 
 - Code Smell baseline for legacy projects
-- [Gradle plugin](#with-gradle) for code analysis via Gradle builds
+- [Gradle plugin](https://github.com/detekt/detekt/tree/main/detekt-gradle-plugin) for code analysis via Gradle builds
 - Gradle tasks to use local `IntelliJ` distribution for formatting and inspecting Kotlin code
 - [SonarQube integration](https://github.com/arturbosch/sonar-kotlin)
 - Extensibility by enabling incorporation of personal rule sets, `FileProcessListener's` and `OutputReport's`
--- a/website/blog/2019-08-04-calling-it-one-zero.md
+++ b/website/blog/2019-08-04-calling-it-one-zero.md
@@ -20,7 +20,7 @@ programming flaws like unused or too complex constructs. Think of it as *pmd* or
 - Suppression of findings with Kotlin's `@Suppress` and Java's `@SuppressWarnings` annotations
 - Specification of quality gates which will break your build 
 - Code Smell baseline for legacy projects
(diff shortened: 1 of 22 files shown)


website/blog/2019-08-04-calling-it-one-zero.md
```diff
@@ -20,7 +20,7 @@ programming flaws like unused or too complex constructs. Think of it as *pmd* or
 - Suppression of findings with Kotlin's `@Suppress` and Java's `@SuppressWarnings` annotations
 - Specification of quality gates which will break your build 
 - Code Smell baseline for legacy projects
-- [Gradle plugin](#with-gradle) for code analysis via Gradle builds
+- [Gradle plugin](https://github.com/detekt/detekt/tree/main?tab=readme-ov-file#with-gradle) for code analysis via Gradle builds
```
ahmdaa, line 23: Only valid reference I found to `#with-gradle` was in this link to a section on the README so I believe this is the right link but please correct me if I'm wrong.

website/docs/gettingstarted/gradle.mdx
```diff
@@ -54,7 +54,7 @@ complete compile classpath from the Android Gradle Plugin ([upstream ticket](htt
 and can safely be ignored.
 
 Use the Groovy or Kotlin DSL of Gradle to apply the Detekt Gradle Plugin. You can further configure the Plugin
-using the detekt closure as described [here](#closure).
+using the detekt closure as described [here](#options-for-detekt-configuration-closure).
```
ahmdaa, line 57: Docusaurus does not recognize `id`s or `name`s in regular HTML tags I believe. Docusaurus suggests special syntax when we want to set a heading id:
```
### Hello World {#my-explicit-id}
``` 
or we can just use the one Docusaurus automatically generates from the heading text.
See: https://docusaurus.io/docs/markdown-features/toc#heading-ids for more information.

website/src/pages/marketplace/index.jsx
```diff
@@ -31,16 +31,12 @@ function MarketplaceHeader() {
       >
         🙏 Please add your ruleset
       </Link>
-      <Link
-        className={clsx(
-          "button",
-          "button--secondary",
-          styles.marketplaceHeaderButton
-        )}
+      <a
```
ahmdaa, line 34: Similar to [this issue](https://github.com/detekt/detekt/pull/7630/files#r1759215012). It seems Docusaurus does not recognize the `id` assigned to the HTML `section` in `MarketplaceFooter` when it's used in a `Link`. We can just use a regular `a` tag so that Docusaurus is not concerned with whether the undetected `id` exists or not.

website/blog/2019-08-04-calling-it-one-zero.md
cortinico: Let's link it to https://detekt.dev/docs/gettingstarted/gradle/

website/docs/gettingstarted/gradle.mdx
cortinico: > or we can just use the one Docusaurus automatically generates from the heading text which is what I did here.

Let's go this way 👍

cortinico (approved):

cortinico:
Thank you very much! <3

PR #7637, opened 2024-09-14
Sort results in compiler plugin by column:line

### LINKED ISSUE #5749: Sort log output by line then column
All warnings are output in order of detekt provider type. They should be logged in order: by line, then column. This makes it easier to review the list of issues detekt finds.

neugartf:
This fixes #5749 by sorting the violations by line and column.

### DIFF
--- a/detekt-compiler-plugin/src/main/kotlin/io/github/detekt/compiler/plugin/internal/MessageCollectorExtensions.kt
+++ b/detekt-compiler-plugin/src/main/kotlin/io/github/detekt/compiler/plugin/internal/MessageCollectorExtensions.kt
@@ -24,7 +24,7 @@ fun MessageCollector.error(msg: String) {
 fun MessageCollector.reportIssues(result: Detektion) {
     result.issues
         .filterNot { it.suppressed }
-        .sortedWith(compareBy({ it.location.source.line }, { it.location.source.column }))
+        .sortedBy { it.location }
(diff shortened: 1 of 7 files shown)


detekt-compiler-plugin/src/main/kotlin/io/github/detekt/compiler/plugin/internal/MessageCollectorExtensions.kt
```diff
@@ -24,6 +24,7 @@ fun MessageCollector.error(msg: String) {
 fun MessageCollector.reportIssues(result: Detektion) {
     result.issues
         .filterNot { it.suppressed }
+        .sortedWith(compareBy({ it.location.source.line }, { it.location.source.column }))
```
BraisGabin, line 27: 🤔 Shouldn't we sort first by file (path)? And, would it have sense to make location comparable?

neugartf: Whoops, totally missed that. 
`Location` is now `Comparable` and tests have been changed to reflect multiple files.
I put the `compareTo()` in the interface but happy to pull it down.

BraisGabin (approved):

BraisGabin:
Thanks for the PR :)

Issue #7585, opened 2024-08-27
Rule for linting `suspend` call in `finally` block


WildOrangutan:
## Expected Behavior of the rule

Would it make sense adding a rule, that would check, if suspending function is called from `finally` block, without using `NonCancellable`?

For example, in following case, `suspendingCleanup()` will not be executed, if parent coroutine scope was cancelled, while `suspendingWork()` was still running.

```kotlin
try {
    suspendingWork()
} finally {
    suspendingCleanup()
}
```

The rule I'm suggesting, would either enforce using `NonCancellable` coroutine context or require you to suppress it:

```kotlin
try {
    suspendingWork()
} finally {
    withContext(NonCancellable) {
        suspendingCleanup()
    }
}
```

If you want to play around: [Kotlin Playground
](https://pl.kotl.in/WQzzqa6Cz)
## Context

Trying to avoid potential and subtle bugs, when `NonCancellable` is not used.

BraisGabin:
It sound nice.

VitalyVPinchuk:
I'd like to work on the rule if it's still relevant.

BraisGabin:
It is relevant

VitalyVPinchuk:
Should we report all suspend functions in `finally` that aren't wrapped in `NonCancellable`,
or should we attempt to analyze if they are inherently non-cancelable and ignore those? Though it might be challenging to tell if a suspend function is cancelable or not.

BraisGabin:
I think that we should report all of them.

If we could find "inherently non-cancelable" suspend function then we should create another rule to report them.

atulgpt:
Another pattern that is used to perform the non-cancellable task is to use some externalScope or `applicationScope`. We may also avoid flagging those.

VitalyVPinchuk:
> Another pattern that is used to perform the non-cancellable task is to use some externalScope or `applicationScope`. We may also avoid flagging those.

That's correct, but it seems like this might be a concern of a different rule.
Executing `finally` code in a coroutine scope other than the original `try` block feels a bit counterintuitive to me. Is this a common pattern?

atulgpt:
I read this here https://medium.com/androiddevelopers/coroutines-patterns-for-work-that-shouldnt-be-cancelled-e26c40f142ad He suggested a few ways where using application-hosted coroutine scope is also one way. 
` but it seems like this might be a concern of a different rule.` make sense

cortinico:
Closed by https://github.com/detekt/detekt/pull/7664

PR #7664, opened 2024-09-22
SuspendFunInFinallySection - New rule for linting suspend calls in `finally` section

VitalyVPinchuk:
[WIP]

To address #7585

Visits expression call in `finally` section and determines if the call is `suspend`. Then traverses all its parents to see if one of them is `withContext(NonCancellable)`. This is the simplest case.

Should we consider the following?

```kotlin
// Nested calls. I don't see a way to implement it 
// (e.g. call made to external library that provides non-cancellable context)
suspend fun wrapper(b: suspend () -> Unit) = withContext(NonCancellable) { b() }
wrapper { test() }
```
            
```kotlin
// Lambda
// New coroutine still will be cancelled if launched in cancelled context
val lambda: suspend (suspend () -> Unit) -> Unit = { block -> launch { block() } } 
withContext(NonCancellable) { lambda { test() } }
```

### DIFF
--- a/detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInFinallySection.kt
+++ b/detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInFinallySection.kt
@@ -3,7 +3,7 @@ package io.gitlab.arturbosch.detekt.rules.coroutines
 import io.gitlab.arturbosch.detekt.api.CodeSmell
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Entity
-import io.gitlab.arturbosch.detekt.api.RequiresTypeResolution
+import io.gitlab.arturbosch.detekt.api.RequiresFullAnalysis
 import io.gitlab.arturbosch.detekt.api.Rule
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.descriptors.FunctionDescriptor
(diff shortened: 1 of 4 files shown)


detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInFinallySection.kt
```diff
@@ -0,0 +1,94 @@
...
+        return !isNonCancellableArgument(firstArgument, bindingContext)
+    }
+
+    private fun KtCallExpression.isSuspendCall() =
+        (getResolvedCall(bindingContext)?.resultingDescriptor as? FunctionDescriptor)?.isSuspend == true
+
+    private fun KtCallExpression.parentCallsUpTo(topParent: PsiElement) =
+        generateSequence(this as PsiElement) { it.parent }
+            .takeWhile { it != topParent }
+            // .takeIf { !it.any { element -> element is KtLambdaExpression } }
```
cortinico, line 78: why this?

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunInFinallySectionSpec.kt
```diff
@@ -0,0 +1,151 @@
...
+            "wrapper"(2)(code),
+            "test"(1)(code),
+            "block"(2)(code),
+        )
+    }
+
+    companion object {
+        private val NOTHING: Array<Pair<Int, Int>> = emptyArray()
+
+        operator fun String.invoke(ordinal: Int): (String) -> Pair<Int, Int> =
```
cortinico, line 131: Not a huge fan of this as I believe we don't use it anywhere else in our tests. Can you update to use regular functions rather than a string operator?

VitalyVPinchuk: Sure

cortinico (approved):

BraisGabin (approved):

atulgpt:
@BraisGabin https://github.com/detekt/detekt/issues/7585 this should be closed now. Right?

PR #7676, opened 2024-09-27
Remove version references from SARIF output

3flex:
* Version is not correct when using a snapshot version of detekt
* fullName should include version, according to spec, so removing it as adding an incorrect version does not make sense. It's also ignored by GitHub.
* downloadUri URI is broken and leads to 404. It's also ignored by GitHub.
* version & semanticVersion are not required by GitHub.

These properties "MAY" be included in SARIF output reports so removing them does not impact spec compliance ([spec](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html)).

If correct values can be provided then that can be added in a new PR.


BraisGabin:
I'm not sure if removing this is the correct way to fix this issue. What happens now? I agree that broken links shouldn't happen but we should say that the version is `2.0.0-SNAPSHOT`. And with the links we could point to the "next" version.

3flex:
Like I said, if it's important to include in the report output, open a PR after this and add it.

Right now the output is wrong. Removing the incorrect output is the quickest way to fix that, remain spec compatible, and without impacting anyone using GitHub's SARIF support which I bet is the majority of SARIF report users.

BraisGabin (approved):

PR #7712, opened 2024-10-11
Suppress redundant result levels in the SARIF report

### LINKED ISSUE #7684: SARIF: Don't set result level when it doesn't override rule's level
## Expected Behavior
Result `level` is only set in report output if it overrides the rule's `level`. If it doesn't, it's redundant and can be removed from the output to cut down on the size of the report output.

## Current Behavior
`level` is set on every result.

## Context
Improve SARIF output.

VitalyVPinchuk:
Suppresses issue result level output unless it overrides the rule level.

Fixes #7684


3flex:
It shouldn't be necessary to update the API to fix this - a different approach is required.

VitalyVPinchuk:
Alright, I'll fix that

PR #7714, opened 2024-10-12
Make "detekt-sample-extensions" project self-contained

### LINKED ISSUE #7703: Make "detekt-sample-extensions" project self-contained
## Expected Behavior
It should be possible to extract & build the sample project from source independently of the detekt build.

The project should also not be published.

## Current Behavior

The build references detekt's build-logic `module` and version catalog. The project is also published, which doesn't make sense.

## Context
Make it easier to use the extensions sample.

3flex:
Fixes #7703


BraisGabin:
Nothing of these are blockers, just ideas:

- We could execute this module like this: `./gradlew -p detekt-sample-extensions build`. That should work more less the same and we have less files to maintain, right? And we would have less files to maintain.
- I'm not 100% sure if I like that to have the `main` version of `:detekt-sample-extensions` alaigned with the last stable version of detekt instead of the current `main`.
- Do we even need to maintain this? We have https://github.com/detekt/detekt-custom-rule-template and we have already first party plugins that work as examples. And lots of third party plugins. Maybe we can just remove it.

3flex:
Responses to this depends on what the goal of the sample project is.

> We could execute this module like this: ./gradlew -p detekt-sample-extensions build. That should work more less the same and we have less files to maintain, right? And we would have less files to maintain.

We have less files yes, but then it's not a self-contained project that others can pick up and use as is. If that's not a goal and it's more about the code then I agree.

> I'm not 100% sure if I like that to have the main version of :detekt-sample-extensions alaigned with the last stable version of detekt instead of the current main.

If people are building extensions wouldn't they target our last public release? And the comments in the current build file say to substitute dependencies with the current released versions which wouldn't build.

> Do we even need to maintain this? We have https://github.com/detekt/detekt-custom-rule-template and we have already first party plugins that work as examples. And lots of third party plugins. Maybe we can just remove it.

That's a template for a rule, not extensions, so they serve different purposes in some ways... But I agree we could probably just remove it.

BraisGabin (approved):

Issue #7719, opened 2024-10-13
Documentation is not Updated with latest version 


markoeltiger:
## Expected Behavior
I should find the latest detekt version on the documentation website to use it 
## Observed Behavior
i find older version of detekt

## Steps to Reproduce
-Go for detekt.dev getting started for example go for  [Android -> Gradle ](https://detekt.dev/docs/gettingstarted/gradle) you will find the stated version in the documentation is 1.23.3 while the latest is 1.23.7
## Context
-Developers Doesnt use latest version

BraisGabin:
Could add a link?

markoeltiger:
I have updated the issue description

BraisGabin:
Fixed with #7723

PR #7720, opened 2024-10-13
Update DetektVersion in the documentation to the latest

markoeltiger:
The Documentation was using older version due to not updated key

This fixes https://github.com/detekt/detekt/issues/7719


BraisGabin:
I'm not sure why on the deployed version on Vercel this PR doesn't fix the issue. This also made me spot how broke all this part is. I created #7723 that should fix the issue for now (until the next release).

cortinico (approved):

BraisGabin:
Fixed with #7723 for the current doc and #7724 tracks the fix to avoid this issue in the future. Thank you for the issue and the pr! We really appreciate when someone raises issues in our docs. They are really difficult to spot.

BraisGabin:
I was just thinking. That anyway this change is. Correct so we should merge it anyways.

PR #7721, opened 2024-10-13
Report mishandling of CancellationException

jonapoul:
# Summary
This is a functional extension and slight reworking of the behaviour of the `SuspendFunSwallowedCancellation` rule [from this PR](https://github.com/detekt/detekt/pull/5666). That PR mentioned that this kind of thing could be done in future, but I haven't seen anyone else give it a go, so here it is!

In summary, the change is that you now need to explicitly catch CancellationException when using a try-catch block in a suspending context, then immediately re-throw it without doing anything else (logging, etc.) in that catch block.

I can't find a way of making the patch coverage any higher than it currently is, so that step won't give us a nice green tick unless someone else can suggest something. A few parameters are nullable although I don't think Kotlin supports missing them out, which means a few lines are yellow in the coverage report.

# Clarification
One area where I'm specifically interested in feedback is the case where we have a potentially long-running bit of code being called in a `finally` block after catching a cancellation. I'm not sure whether this should be counted as an issue or not, since other catch blocks might want to make use of a finally block E.g.:
```kotlin
suspend fun foo() {
    try {
        delay(1000L)
    } catch (e: CancellationException) {
        throw e
    } catch (e: SomeOtherException {
        // whatever
    } finally {
        // do some work to tidy up - won't finish the coroutine immediately
    }
}
```
Right now this isn't being checked for at all.

# Behavioural change
The behavioural change is that previously, the test case like below would pass because IllegalStateException is a superclass of CancellationException:
```kotlin
suspend fun foo() {
    try {
        check(1 + 2 == 3) { "Some actual illegal state checking" }
        delay(1000L)
        // handle error - which one?
    }
}
```

whereas now (at least in my opinion), handling of CancellationException should be more explicit, like:

```kotlin
suspend fun foo() {
    try {
        check(1 + 2 == 3) { "Some actual illegal state checking" }
        delay(1000L)
    } catch (e: CancellationException) {
        throw e
        // handle other actual error cases
    }
}
```

Happy to take feedback on this though, of course (and the rest too :smile:).

### DIFF
--- a/detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
+++ b/detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
@@ -151,31 +151,33 @@ class SuspendFunSwallowedCancellation(config: Config) : Rule(
             return
         }
 
-        for (catchClause in expression.catchClauses) {
-            val parameter = catchClause.catchParameter ?: continue
-            val parameterFqName = bindingContext[BindingContext.VALUE_PARAMETER, parameter]
-                ?.type
-                ?.constructor
-                ?.declarationDescriptor
-                ?.fqNameOrNull()
+        // CancellationException should be the first case to catch - so we only want to check the first catch clause
+        val catchClause = expression.catchClauses.first()
+        val parameter = catchClause.catchParameter
+        if (parameter == null) {
+            // Not sure exactly what would cause this, but just in case?
+            report(catchClause)
+            return
+        }
 
-            if (parameterFqName !in CANCELLATION_EXCEPTION_FQ_NAMES) {
-                // Should handle CancellationException first - this is a potential problem
+        val parameterFqName = bindingContext[BindingContext.VALUE_PARAMETER, parameter]
+            ?.type
+            ?.constructor
+            ?.declarationDescriptor
+            ?.fqNameOrNull()
+
+        if (parameterFqName !in CANCELLATION_EXCEPTION_FQ_NAMES) {
+            // Should handle CancellationException first - this is a potential problem
+            report(catchClause)
+        } else {
+            // This is a CancellationException - we should make sure that it gets explicitly
+            // re-thrown upwards immediately
+            if (!catchClause.exceptionWasRethrown(parameter)) {
                 report(catchClause)
-                return
-            } else if (parameterFqName in CANCELLATION_EXCEPTION_FQ_NAMES) {
-                // This is a CancellationException - we should make sure that it gets explicitly
-                // re-thrown upwards immediately
-                if (!catchClause.exceptionWasRethrown(parameter)) {
-                    report(catchClause)
-                } else if (catchClause.doesAnythingElseBeforeRethrowing()) {
-                    // it does re-throw, but a potentially long-lasting piece of code is called first. This is still a
-                    // problem!
-                    report(catchClause)
-                }
-
-                // We've found what we're looking for, so no more work to do
-                return
+            } else if (catchClause.doesAnythingElseBeforeRethrowing()) {
+                // it does re-throw, but a potentially long-lasting bit of logic is called first. This is still a
+                // problem!
+                report(catchClause)
             }
         }
     }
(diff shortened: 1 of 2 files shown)


detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
```diff
@@ -16,6 +17,294 @@ class SuspendFunSwallowedCancellationSpec(private val env: KotlinCoreEnvironment
...
+            class SomeOtherException : Exception()
+
+            suspend fun foo() {
+                try {
+                    delay(1000L)
+                } catch (e: SomeOtherException) {
+                    // handle error case
+                } catch (e: CancellationException) {
+                    throw e
+                }
```
BraisGabin, line 229: I'm not sure of this case. `CancellationException` should be on top of any other exception type if that type is a supertype of `CancellationException`. Otherwise we shouldn't care, right?

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
```diff
@@ -109,6 +140,43 @@ class SuspendFunSwallowedCancellation(config: Config) : Rule(
...
+    @Suppress("ReturnCount")
+    override fun visitTryExpression(expression: KtTryExpression) {
+        super.visitTryExpression(expression)
+
+        val function = expression.getParentOfType<KtFunction>(strict = true)
+        val functionDescriptor = bindingContext[BindingContext.FUNCTION, function]
+        if (functionDescriptor?.isSuspend != true) {
+            // Don't care about the try-catch block unless it's in a suspending context
+            return
+        }
```
BraisGabin, line 152: 🤔 what would happen with something like this:

```kotlin
fun foo() {
  GlobalScope.launch {
    try {
       delay(1_000)
    } catch (_: Exception) {}
  }
}
```

`foo` is not a `suspend` function. But the issue is still there. Could you add a test case to ensure that something like this is still catched?

BraisGabin (commented):
Don't worry about the coverage. The test cases seem fine. Look at my comments and tell me what you think.

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
jonapoul: Fair point - updated to use IllegalStateException instead of SomeOtherException

detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellation.kt
jonapoul: Added - this passes because it's checking the context of the try block which is suspending.

detekt-rules-coroutines/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/SuspendFunSwallowedCancellationSpec.kt
BraisGabin: I think that the test itself was good. The problem is the implementation. You are forcing that the first catch should be `CacellationException` but that's too strict. For example something like this should be completely fine:

```kotlin
try {
  delay(5_000)
} catch (e: IllegalArgumentException) {
}
```

That code doesn't catch `CacellationException` so we shouldn't report this code.

And the same happens with this:

```kotlin
try {
  delay(5_000)
} catch (e: IllegalArgumentException) {
} catch (e: CancellationException) {
}
```

Instead of checking that the first `catch` is `CancellationException` we should check that the first catch of `CancellationException` of any of its super types throws the exception directly.

For example: something like this should be fine:

```kotlin
try {
  delay(5_000)
} catch (e: IllegalArgumentException) {
  throw e
} catch (e: Exception) {
}
```

The first one is not a supertype of `CancellationException` so we don't care. And the second is even if it's not `CancellationException` what it does is to throw it directly. So there is nothing wrong there.

jonapoul: Ah right, I misunderstood you! That makes more sense now, thanks - I'll have a go :smile:

BraisGabin (approved):
Sorry for the late, late, review. This looks really good now! Thanks for the contribution!!

PR #7723, opened 2024-10-13
Fix versions on our versioned docs

BraisGabin:
Related with #7720

Now that we have versioned docs the `detektVersionReplace.js` script doesn't work correctly because it always points to the last version when it shouldn't.

We should change the way that we generated the versioned docs to avoid this issue but this PR should fix all our docs for now. I'm going to open an issue about this.


cortinico (approved):

cortinico:
> We should change the way that we generated the versioned docs to avoid this issue but this PR should fix all our docs for now. I'm going to open an issue about this.

Yeah seems like we missed this once we moved to versioned docs

PR #7730, opened 2024-10-15
Use settings file to set up repositories in tests

3flex:
This improves test suite's compatibility with Gradle's Isolated Projects feature.


BraisGabin (commented):
Couldn't we hardcoding the list of repositories on `DSLGradleRunner`? We want that for all our tests, right? That code in the tests only increase the verbosity.

BraisGabin (approved):

3flex:
That's not really part of the scope for this PR, it's just as verbose as it was before, but this might be something to look at later.

PR #7731, opened 2024-10-15
Enable Isolated Projects for Gradle tests by default

3flex:
IP must be disabled on multiple tests at the moment.

For the tests where IP is disabled:
* Kotlin 2.1.0 fixes a couple of multiplatform issues
* We need to rework how report merging works in our plugin
* The tests in DetektTaskMultiModuleSpec will never be compatible with IP. When everything else eventually becomes IP compatible I think we should remove these tests - there should be no need to specifically test that poor practice in build files still works.

### DIFF
--- a/detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
+++ b/detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
@@ -22,6 +22,7 @@ constructor(
     val baselineFiles: List<String> = emptyList(),
     val gradleVersionOrNone: String? = null,
     val dryRun: Boolean = false,
+    var disableIP: Boolean = false,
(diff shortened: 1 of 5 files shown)


detekt-gradle-plugin/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/testkit/DslGradleRunner.kt
```diff
@@ -22,6 +22,7 @@ constructor(
     val baselineFiles: List<String> = emptyList(),
     val gradleVersionOrNone: String? = null,
     val dryRun: Boolean = false,
+    var disableIP: Boolean = false,
```
3flex, line 25: I didn't love making this `var` not `val` but did it because it was the simplest way to implement without having to do a bunch of refactoring of the KMP tests, and because it's intended to only be used temporarily until all tests pass with IP enabled.

cortinico (approved):

PR #7733, opened 2024-10-17
Fix error in paths filtering for detekt-cli

ilim-ablyaev:
Fixed errors in path filtering for gitlab-cli.

Changes made according to comments in [PathFilters.kt](https://github.com/detekt/detekt/blob/main/detekt-utils/src/main/kotlin/io/github/detekt/utils/PathFilters.kt)
```
    /**
     * - If [includes] and [excludes] are not specified,
     *   return false.
     * - If [includes] is specified but [excludes] is not,
     *   return false iff [path] matches any [includes].
     * - If [includes] is not specified but [excludes] is,
     *   return true iff [path] matches any [excludes].
     * - If [includes] and [excludes] are both specified,
     *   return false iff [path] matches any [includes] and [path] does not match any [excludes].
     */
    fun isIgnored(path: Path): Boolean
```


ilim-ablyaev:
> @BraisGabin take a look, please, I guess there was a problem in #7007

@3flex would you mind reviewing the changes as well, please?

BraisGabin (approved):
Wow! Thanks! This was clearly my bad.

Issue #7433, opened 2024-07-03
When setting `config -> excludes -> ''` to empty string misleading error is logged


atulgpt:
## Expected Behavior
Setting `config -> excludes -> ''` should print some error like below
`Value "" set for config parameter "config > excludes" is not of required type List.`

## Observed Behavior
The below error is getting printed
`Value "" set for config parameter "config > excludes" is not of required type EmptyList.`
See `EmptyList`, it should be `List`

## Steps to Reproduce
In latest branch

## Context
Was updating the detekt to latest snapshot

## Your Environment
NA

PR #7753, opened 2024-10-28
Map `EmptyList` to `List`

atulgpt:
Fixes https://github.com/detekt/detekt/issues/7433

### DIFF
--- a/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/BaseConfig.kt
+++ b/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/BaseConfig.kt
@@ -41,14 +41,20 @@ fun Config.valueOrDefaultInternal(
     } catch (_: ClassCastException) {
         error(
             "Value \"$result\" set for config parameter \"${keySequence(key)}\" is not of" +
-                " required type ${default::class.simpleName}."
+                " required type ${default::class.simpleName?.let { getDefaultName(it) }}."
         )
     } catch (_: NumberFormatException) {
         error(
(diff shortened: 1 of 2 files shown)


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/BaseConfig.kt
```diff
@@ -41,15 +43,21 @@ fun Config.valueOrDefaultInternal(
...
     } catch (_: NumberFormatException) {
         error(
             "Value \"$result\" set for config parameter \"${keySequence(key)}\" is not of" +
-                " required type ${default::class.simpleName}."
+                " required type ${default::class.simpleName?.let { getDefaultName(it) }}."
         )
     }
 
+private fun getDefaultName(className: String): String = if (className == EMPTY_LIST_CLASS_NAME) {
+    LIST_CLASS_NAME
```
BraisGabin, line 56: Nit: I would not use constants here. I would inline the values (to me those constants make the code harder to read). That or use something like `empryList()::class.simpleName`.

BraisGabin (approved):

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/config/BaseConfig.kt
atulgpt: I just inlined the variable, I didn't explicitly use the `empryList()::class.simpleName` or `List::class.simpleName` as just using the plaint string is more simpleter and also `simpleName` is nullable so it avoids the use of nullability handling

Issue #7660, opened 2024-09-21
UnusedPrivateClass false positive with callable references


JordanLongstaff:
## Expected Behavior
I expect to see no `UnusedPrivateClass` warning because the private class is, in fact, used.

## Observed Behavior
The private class is marked by Detekt as an unused private class and the warning occurs.

## Steps to Reproduce
[This](https://github.com/JordanLongstaff/ArtemisAgent/security/code-scanning/286) is one of six private classes like this that are declared in my test suites. All of them are flagged as unused by Detekt, but none of them are actually unused, because there are callable references for all of them (namely, their primary constructors).

![Screen Shot 2024-09-21 at 2 22 09 PM](https://github.com/user-attachments/assets/804521d4-e4fa-4b60-b5fa-7e3cc6a03367)

## Context
These false positives are showing up as security alerts in my GitHub repo. I'm trying to clear some of them out today.

## Your Environment

* Version of detekt used: 1.23.7
* Version of Gradle used (if applicable): 8.10.1
* Operating System and version: macOS Monterey 12.7.6
* Link to your project (if it's a public repository): [ArtemisAgent](https://github.com/JordanLongstaff/ArtemisAgent)

cortinico:
> [This](https://github.com/JordanLongstaff/ArtemisAgent/security/code-scanning/286) is one of six private classes

This link is broken.

Could you provide a snippet that would reproduce this detection instead?

PR #7756, opened 2024-10-29
[UnusedPrivateClass] - Check the right of `::` even when LHS is missing

atulgpt:
Fixes: https://github.com/detekt/detekt/issues/7660

Earlier we were only checking the RHS of `::` when there was empty LHS, so case when we use `::Foo` in `Parent class` like `Parent::Foo` we were skipping the usage of `Foo`

### DIFF
--- a/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateClassSpec.kt
+++ b/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateClassSpec.kt
@@ -457,5 +457,71 @@ class UnusedPrivateClassSpec {
             assertThat(findings).hasSize(1)
             assertThat(findings).hasStartSourceLocation(10, 5)
         }
+
+        @Test
+        fun `should not report when callable ref for constructor is used`() {
+            val code = """
+                private class A
+
+                private val listOfConstructors = listOf(::A)
+            """.trimIndent()
+            val findings = UnusedPrivateClass(Config.empty).compileAndLint(code)
+            assertThat(findings).isEmpty()
+        }
+
+        @Test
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateClassSpec.kt
```diff
@@ -457,5 +457,71 @@ class UnusedPrivateClassSpec {
...
+                private class A
+
+                private val listOfConstructors = listOf(::A)
+            """.trimIndent()
+            val findings = UnusedPrivateClass(Config.empty).compileAndLint(code)
+            assertThat(findings).isEmpty()
+        }
+
+        @Test
+        fun `should not report when callable ref for inner private class constructor is used with parent name`() {
```
atulgpt, line 473: this was failing earlier

detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UnusedPrivateClassSpec.kt
```diff
@@ -457,5 +457,71 @@ class UnusedPrivateClassSpec {
...
+
+                    private val list = listOf(Parent::Foo)
+                }
+            """.trimIndent()
+            val findings = UnusedPrivateClass(Config.empty).compileAndLint(code)
+            assertThat(findings).isEmpty()
+        }
+
+        @Test
+        fun `should not report callable ref for inner private child class constructor is used with parent private class name`() {
```
atulgpt, line 486: Also this

cortinico (approved):

BraisGabin (approved):

BraisGabin:
Thanks for all the PRs!

Issue #7765, opened 2024-11-04
IgnoredReturnValue should not flag ignored `Nothing`


koljakube:
Rule: `IgnoredReturnValue`
Change: Ignore unhandled return values of type `Nothing`

## Expected Behavior of the rule

As far as I can tell, only `Unit` is never checked by this rule. I propose adding `Nothing` to that list. In practice, it is even less of a normal return value than `Unit` (which is an object an can be assigned to variables), so I can't think of any downsides. As it stands, this is a warning:

```
# detekt.yml
  IgnoredReturnValue:
    active: true
    restrictToConfig: false
    returnValueAnnotations: []
    returnValueTypes: []
    ignoreReturnValueAnnotations: []
    ignoreFunctionCall: []
```

```
fun foo(): Int {
  TODO() // The call TODO is returning a value that is ignored. [IgnoredReturnValue]
}
```

## Context

I think this should avoid false positives for situations were `IgnoredReturnValue` is in heavy use.

This is basically the opposite of #6661 (requiring specific types to _always_ be checked, but which seems to have been implemented already by defining the `returnValueTypes` config property), so maybe there is some overlap on the implementation side as well.

PR #7772, opened 2024-11-09
[IgnoredReturnValue] Add `returnType?.isNothing()` check

atulgpt:
Fixes https://github.com/detekt/detekt/issues/7765

### DIFF
--- a/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValueSpec.kt
+++ b/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValueSpec.kt
@@ -1056,6 +1056,28 @@ class IgnoredReturnValueSpec {
             val findings = rule.compileAndLintWithContext(env, code)
             assertThat(findings).isEmpty()
         }
+
+        @Test
+        fun `does not report when a function returns Unit`() {
+            val code = """
+                fun foo() {
+                    println(42)
+                }
+            """.trimIndent()
+            val findings = subject.compileAndLintWithContext(env, code)
+            assertThat(findings).isEmpty()
+        }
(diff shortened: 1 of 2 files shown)


detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValueSpec.kt
```diff
@@ -1056,6 +1056,28 @@ class IgnoredReturnValueSpec {
...
+        @Test
+        fun `does not report when a function returns Unit`() {
+            val code = """
+                fun foo() {
+                    println(42)
+                }
+            """.trimIndent()
+            val findings = subject.compileAndLintWithContext(env, code)
+            assertThat(findings).isEmpty()
+        }
```
BraisGabin, line 1069: Didn't we have a test for this case? Good catch then!

BraisGabin (approved):
x"D I was just reviewing the issue and saw "oh there is a PR already". I enter and it was just 1 minute old :P

Thanks for all the contributions (I know that you have more things that need my review but I have not the time :'( )

detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/IgnoredReturnValueSpec.kt
atulgpt: I could only find Unit involving with chain

schalkms (commented):
Thanks for fixing this issue!
I’d also like to have an explicit test case for `Nothing`, e.g.

```kotlin
fun fail(message: String): Nothing {
    throw IllegalArgumentException(message)
}
```

cortinico (approved):

PR #7775, opened 2024-11-09
Add `RuleInstance` url

BraisGabin:
This PR goes in the line to implement #7717 but it's not yet that.

But this simplifies `RuleInstance` a lot. It had `ruleId` and `ruleName`. That was *really* confusing. With this change we don't need `ruleName` in the `RuleInstance` anymore. The only place where we were using the `name` was to construct the `url` of each rule. Now, instead of having the `ruleName` we have the `url` itself. And now that we have the `url` itself on the `RuleInstance` we can implement #7717 and allow each rule to provide it's own `url`. This way we allow to configure this to custom rules.

### DIFF
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
@@ -1,6 +1,7 @@
 package io.gitlab.arturbosch.detekt.api
 
 import dev.drewhamilton.poko.Poko
+import java.net.URI
 import java.nio.file.Path
 
 /**
@@ -46,7 +47,7 @@ val Issue.suppressed: Boolean
 class RuleInstance(
     val id: String,
     val ruleSetId: RuleSet.Id,
-    val url: String?,
+    val url: URI?,
     val description: String,
     val severity: Severity,
 )
--- a/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
+++ b/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
@@ -45,7 +45,7 @@ val Issue.suppressed: Boolean
 @Poko
 class RuleInstance(
     val id: String,
-    val name: Rule.Name,
     val ruleSetId: RuleSet.Id,
+    val url: String?,
     val description: String,
 )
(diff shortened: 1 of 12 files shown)


alpeshvas:
following this.

BraisGabin:
@detekt/maintainers could I ask for a review on this pr? (also, a review in my other open prs would be appreciated too)

And happy new year too :p

detekt-report-md/src/test/kotlin/io/github/detekt/report/md/MdOutputReportSpec.kt
```diff
@@ -181,20 +179,6 @@ class MdOutputReportSpec {
         assertThat(result).contains("Description rule_b")
     }
 
-    @Test
-    fun `renders the right documentation links for the rules`() {
```
schalkms, line 185: Why is this test case removed?

detekt-report-html/src/test/kotlin/io/github/detekt/report/html/HtmlOutputReportSpec.kt
```diff
@@ -101,20 +101,6 @@ class HtmlOutputReportSpec {
         assertThat(result).contains("<span class=\"description\">Description rule_b</span>")
     }
 
-    @Test
-    fun `renders the right documentation links for the rules`() {
```
schalkms, line 105: Why is this test case removed?

BraisGabin: Before the code that was generating the url was in the reporter itself. For that reason we needed this test. Now this logic was moved to the core so this test is not needed.

These asserts were moved to the new test in the analyzer. In next prs that responsability will be moved again. This time from the core to each rule. This way the third party rules can customize it.

detekt-report-md/src/test/kotlin/io/github/detekt/report/md/MdOutputReportSpec.kt
BraisGabin: Same reason: https://github.com/detekt/detekt/pull/7775#discussion_r1898052377

detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Issue.kt
```diff
@@ -45,7 +45,7 @@ val Issue.suppressed: Boolean
 @Poko
 class RuleInstance(
     val id: String,
-    val name: Rule.Name,
     val ruleSetId: RuleSet.Id,
+    val url: String?,
```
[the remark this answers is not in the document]
BraisGabin: Good idea. I'll update it.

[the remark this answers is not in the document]
BraisGabin: Done :)

schalkms (approved):
👍

PR #7776, opened 2024-11-09
Warn for rules used with multiInstance when no bindingContext

BraisGabin:
We have a warning for when a user runs detekt without type solving but with type solving rules enabled. This was broken with the introduction of multiple instances for the same rule. This PR fixees that.

Also this PR goes in the line of reduce the usages of `name` where it is not needed because I think that we can remove it from `Rule`.

### DIFF
--- a/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/AnalyzerSpec.kt
+++ b/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/AnalyzerSpec.kt
@@ -93,7 +93,7 @@ class AnalyzerSpec(val env: KotlinCoreEnvironment) {
                           MaxLineLength:
                             active: true
                             maxLineLength: 120
-                          RequiresFullAnalysisMaxLineLength:
+                          RequiresFullAnalysisMaxLineLength/foo:
                             active: true
                             maxLineLength: 120
                     """.trimIndent()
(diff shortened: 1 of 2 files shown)


detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/AnalyzerSpec.kt
```diff
@@ -93,7 +93,7 @@ class AnalyzerSpec(val env: KotlinCoreEnvironment) {
                           MaxLineLength:
                             active: true
                             maxLineLength: 120
-                          RequiresFullAnalysisMaxLineLength:
+                          RequiresFullAnalysisMaxLineLength/foo:
```
schalkms, line 96: Where is this *special* rule setup with the *RequiresFullAnalysis* annotation?

schalkms (approved):

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/AnalyzerSpec.kt
BraisGabin: At the bottom of this file: https://github.com/detekt/detekt/pull/7776/files#diff-a0f99b53c3b11361cc006e11a276a54057ea1ab65763c4ac5afc8a4ac4a1d31aR483

PR #7797, opened 2024-11-21
Fix InjectDispatcher false positives

### LINKED ISSUE #6200: False Positive InjectDispatcher
## Expected Behavior

Injected dispatchers get recognised as injected and not reported by detekt.

## Observed Behavior

> Dispatcher IO is used without dependency injection. [InjectDispatcher]

## Steps to Reproduce

We use an interface to inject all dispatchers via koin:

```kotlin
import kotlinx.coroutines.CoroutineDispatcher
import kotlinx.coroutines.Dispatchers as KDispatchers

@Suppress("PropertyName", "VariableNaming")
interface IDispatchers {
    val Default: CoroutineDispatcher
    val Main: CoroutineDispatcher
    val Unconfined: CoroutineDispatcher
    val IO: CoroutineDispatcher
}

@Suppress("InjectDispatcher")
object Dispatchers : IDispatchers {
    override val Default = KDispatchers.Default
    override val Main = KDispatchers.Main
    override val Unconfined = KDispatchers.Unconfined
    override val IO = KDispatchers.IO
}
```

Example usage, I also tried to change it to constructor injection but the issue persists:
```kotlin
class Example : KoinComponent {
    private val dispatchers: IDispatchers by inject()
​
    suspend fun loadHtml(url: String): LoadHtmlResponse {
        return withContext(dispatchers.IO) { // <--  Dispatcher IO is used without dependency injection. [InjectDispatcher]
            ...
```
According to the code of the rule and my tries I assume that the issue is that we inject a whole bundle of dispatchers and access a property of this bundle directly.

## Context

Currently I disabled the rule. Would love to enable it again.

## Your Environment

* Version of Gradle used (if applicable): gradle-8.1.1 / androidGradlePlugin = "7.4.2"
* Gradle scan link (add `--scan` option when running the gradle task): -
* Operating System and version: macOS Ventura 13.4

eygraber:
Fixes #6200

### DIFF
--- a/detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/InjectDispatcher.kt
+++ b/detekt-rules-coroutines/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/coroutines/InjectDispatcher.kt
@@ -10,9 +10,14 @@ import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.config
 import io.gitlab.arturbosch.detekt.rules.fqNameOrNull
 import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtClassOrObject
 import org.jetbrains.kotlin.psi.KtConstructorDelegationCall
+import org.jetbrains.kotlin.psi.KtNamedFunction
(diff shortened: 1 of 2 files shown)


cortinico (approved):
Great stuff @eygraber thanks for sending this over

PR #7803, opened 2024-11-22
Fix wrong baseline path in website documentation

Haransis:
When I read the docs, my baseline was generated but not taken into account with `./gradlew detekt`. 
Changing the path fixed my issue.

### DIFF
--- a/website/docs/introduction/baseline.md
+++ b/website/docs/introduction/baseline.md
@@ -75,7 +75,7 @@ task detektProjectBaseline(type: io.gitlab.arturbosch.detekt.DetektCreateBaselin
 subprojects {
     detekt {
         // ...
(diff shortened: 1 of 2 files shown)


scottdfedorov:
This, and the documentation, leads me to believe that the results from multiple modules would be combined into a single file, but that appears to not be the case. Can someone confirm what the expected behavior is for multi-module projects? 

Seems some people are trying to find workarounds, but the comments here are quite old so not sure if there's been an update not tracked?

cortinico (approved):

cortinico:
> This, and the documentation, leads me to believe that the results from multiple modules would be combined into a single file, but that appears to not be the case. Can someone confirm what the expected behavior is for multi-module projects?

We don't do merging of the results or the baselines. There are other issues to track the progress of those features:
- https://github.com/detekt/detekt/issues/1589

PR #7836, opened 2024-12-13
Reduce duplication on Analyzer

BraisGabin:
We had two functions and they did more less the same: select which rules should be executed. But one of them was for execute them and the other was just to log some information. That was bad because it was easy to get them unaligned and because we were executing the same thing twice. This PR removes that. Now we have one single function that provides the rules that should be executed.

Also, now that we have this logic centralized we can use it to export this information to `Detektion` so the reporters can know all the rules that detekt knows about and which ones were executed.


schalkms (commented):
Looks good besides the merge conflict!

schalkms (approved):

PR #7858, opened 2024-12-28
`ExpressionBodySyntax`: don't report if multiple returns in when expressions (Closes #7829)

### LINKED ISSUE #7829: `ExpressionBodySyntax` false positive when more than one return statement
## Expected Behavior
When function has more than one return statement, `ExpressionBodySyntax` should not be triggered.

## Observed Behavior
It is.

## Steps to Reproduce
Example code:

```kotlin
class PromoCodeSummaryMapper {
    fun PromoApplied.toPresentable(currency: Currency): PromoCodeSummary? {
        return PromoCodeSummary(
            code = code,
            type = when {
                promoFreeDelivery == true -> PromoCodeSummary.FreeDelivery
                promoDiscountAmount != null -> PromoCodeSummary.Discount(Amount(-promoDiscountAmount, currency))
                else -> return null
            },
        )
    }
}
```

> .../PromoCodeSummaryMapper.kt:11:9: Functions with exact one statement, the return statement, can be rewritten with ExpressionBodySyntax. [ExpressionBodySyntax]

When I do as suggested by the issue, I get IDE error:

Screenshot 2024-12-10 at 10 22 02

## Your Environment

* Version of detekt used: 1.23.7
* Version of Gradle used (if applicable): 8.10

paolorotolo:
This PR fixes #7829, not reporting a `ExpressionBodySyntax` if there are multiple return expressions in `when` statements.

This code is now compliant:
```kotlin
    fun stuff(): Pair<String, String>? {
        return Pair(
            first = "",
            second = when {
                true -> ""
                else -> return null 
            }
        )
    }
```

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntax.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntax.kt
@@ -14,6 +14,7 @@ import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.KtReturnExpression
+import org.jetbrains.kotlin.psi.KtWhenEntry
 import org.jetbrains.kotlin.psi.psiUtil.anyDescendantOfType
 
 /**
@@ -59,7 +60,9 @@ class ExpressionBodySyntax(config: Config) : Rule(
     private fun KtDeclarationWithBody.checkForExpressionBodySyntax() {
         val stmt = bodyExpression
             ?.singleReturnStatement()
-            ?.takeUnless { it.containsReturnStmtsInNullableArguments() }
+            ?.takeUnless {
+                it.containsReturnStmtsInNullableArguments() || it.containsReturnInWhenExpressions()
+            }
 
         if (stmt != null && (includeLineWrapping || !isLineWrapped(stmt))) {
             report(CodeSmell(Entity.from(stmt), description))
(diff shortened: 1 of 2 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntax.kt
```diff
@@ -69,6 +72,11 @@ class ExpressionBodySyntax(config: Config) : Rule(
     private fun KtExpression.singleReturnStatement(): KtReturnExpression? =
         (this as? KtBlockExpression)?.statements?.singleOrNull() as? KtReturnExpression
 
+    private fun KtReturnExpression.containsReturnInWhenExpressions(): Boolean =
+        anyDescendantOfType<KtReturnExpression> {
+            (it.parent as? KtWhenEntry) != null
```
schalkms, line 77:
```suggestion
            it.parent is KtWhenEntry
```

schalkms (commented):
Thank you for fixing this issue - looks good!
The explicit type cast could be removed to optimise the check.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntax.kt
BraisGabin: Should we have a rule to find this pattern? It should be an easy rule.

schalkms: Good idea! Definitely, such a rule would be of great value.

> Should we have a rule to find this pattern? It should be an easy rule.

paolorotolo: Good catch, thank you!

schalkms (approved):

Issue #3989, opened 2021-07-29
Custom rules in multiple modules project


varunmendiratta:
## Expected Behavior
Custom rule violations should be reported for a project which has multiple modules and a custom Gradle task.

## Observed Behavior
When `detekt` is run on terminal while specifying 'custom rules project JAR', it reports custom rules violations. But it does not report violations when custom Gradle task is run

## Steps to Reproduce
For adding custom rules, I have created another module as mentioned [here](https://github.com/detekt/detekt/issues/3897) `customdetekt`.

Here is the root level build.gradle. I have a custom Gradle task `detektAll `
```
import io.gitlab.arturbosch.detekt.Detekt

plugins {
    id "io.gitlab.arturbosch.detekt" version "1.17.1"
}

tasks.register("detektAll", Detekt) {
    dependsOn(":customdetekt:assemble")
    description = "Custom Detekt build for all modules"
    parallel = true
    setSource(file(projectDir))
    config.setFrom(files("$rootDir/detekt.yml"))
    jvmTarget = "1.8"
    classpath.setFrom(project.configurations.getByName("detekt"))
    include("**/*.kt")
    exclude("**/build/**")
    reports {
        html {
            enabled = true
            destination = file("$rootDir/build/reports/detekt-results.html")
        }
        xml {
            enabled = true
            destination = file("$rootDir/build/reports/detekt-results.xml")
        }
        txt.enabled = false
    }
}
```

Here is the app level `build.gradle`
```
apply plugin: 'io.gitlab.arturbosch.detekt'
dependencies {
	detektPlugins project(':customdetekt')
}
```

Here is the `build.gradle` file of `customdetekt`
```
apply plugin: "kotlin"

repositories {
    mavenCentral()
}

dependencies {
    compileOnly "io.gitlab.arturbosch.detekt:detekt-api:1.17.1"
    compileOnly "org.jetbrains.kotlin:kotlin-stdlib:1.5.21"
}
```

I realised that `customdetekt.jar` file is created & when I call `detekt` from terminal specifying this JAR, it works fine and shows all violations of my custom rules.
`detekt -c detekt.yml -p customdetekt/build/libs/customdetekt.jar -r html:detekt-report.html`

But it does not show custom rule violations when I run `./gradlew detektAll`
What I am doing wrong? Did I miss any step?

## Context
I want to create custom rules on my project which has multiple modules

## Your Environment

* Version of detekt used: 1.17.1
[...]

PR #7862, opened 2024-12-31
Updating custom rules pitfalls.

AdamMc331:
When configuring a custom `detektAll` parallel task at the project root, custom detekt plugins must also be applied as dependencies on the root project, not sub modules.

This was covered in an [issue](https://github.com/detekt/detekt/issues/3989#issuecomment-890331512) that I found in my search, but hopefully adding it to the pitfalls notes means it could save the next person a little bit of time.

### DIFF
--- a/website/docs/introduction/extensions.md
+++ b/website/docs/introduction/extensions.md
@@ -239,3 +239,4 @@ after your extension sub project is built.
 you created a pure kotlin module which has no Android dependencies. `apply plugin: "kotlin"` is enough to make it work.
 - Sometimes when you run detekt task, you may not see the violations detected by your custom rules. In this case open a terminal and run
 `./gradlew --stop` to stop gradle daemons and run the task again.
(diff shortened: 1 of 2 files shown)


AdamMc331:
> Warnings
> ⚠️	
> It looks like you're editing the **un-versioned copy** of our website. This affects only users on the 'next' version of detekt, and it's correct only if you intend to document a **future change or feature**. If you intended to make a change also for the **current** version of detekt, please make sure you edit the equivalent file inside `website/versioned_docs/` as well.

It looks like versioned_docs goes back several versions, I'm happy to make this change in each if necessary, would appreciate a maintainer confirming before I do that though (I'm worried it just makes noise otherwise).

3flex:
If you make the change for the latest version and the unpublished version that will be fine. No need to go back further for this one in my opinion.

Thanks!

BraisGabin (approved):
Thanks!

schalkms (approved):
👍

AdamMc331:
> If you make the change for the latest version and the unpublished version that will be fine. No need to go back further for this one in my opinion.
> 
> Thanks!

Awesome, I updated the 1.23.7 docs as well! Ready to merge now. :)

PR #7865, opened 2025-01-02
Extract getRules outside `Analyzer`

BraisGabin:
The past year I did a lot of refactors to move logic from `Rule` (`:detekt-api`) to `:detekt-core`. Those refactors I think that are still valuable. The problem is that a lot of this code went directly to `Analyzer`.

I think that it's kind of acceptable that `Analyzer` is big. It is the engine of detekt, the class that executes all the rules. But it was far too big. And test it is a nightmare, it does far too many things.

This PR tries to relieve that issue. I'm moving the code that select the rules that need to be executed on each run outside `Analyzer` and adding unitary test for that code. This also helps to reduce the tests inside `Analyzer`.

I know that this PR is big, but nearly everything is to add new tests. You can check it commit by commit, I tried to do a single thing on each commit to make it easier to review.


schalkms (commented):
I'm sorry... I don't have the time to dig deep into this PR and review it at the moment.
Maybe someone else can join in and review.

detekt-api/src/testFixtures/kotlin/io/gitlab/arturbosch/detekt/test/TestFactory.kt
```diff
@@ -53,18 +53,21 @@ fun createIssue(
     suppressReasons = suppressReasons,
 )
 
+@Suppress("LongParameterList")
 fun createRuleInstance(
     id: String = "TestSmell/id",
-    ruleSetId: String = "RuleSet${id.split("/", limit = 2).first()}",
+    ruleSetId: String = "RuleSet${id.substringBefore("/")}",
```
cortinico, line 59: nice :)

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
```diff
@@ -116,14 +117,13 @@ internal class Analyzer(
     }
 }
 
-private fun getActiveRules(
+private fun getRules(
     fullAnalysis: Boolean,
     providers: List<RuleSetProvider>,
     config: Config,
-    log: (() -> String) -> Unit = {},
```
cortinico, line 124: Why have you removed this default parameter here?

cortinico (approved):
I've reviewed it commit by commit, code looks good to me 👍

detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt
BraisGabin: Because that default value wasn't used. But for sure that was not related with that commit.

PR #7866, opened 2025-01-02
Report the rules used by generate a detekt report and use it on the sarif report

### LINKED ISSUE #7779: Add `rules` to `Detektion`
## Expected Behavior
`Detektion` should include a list of the rules so the reports can use that information (for example SARIF)

## Current Behavior
The `core` doesn't provide that information so the SARIF report needs to do the `core`s job of load the rules. 

## Context
With the recent changes on `core` that load mechanic changed. And now the SARIF report doesn't align with the `core` reporting some rules wrongly.

Also, this will reduce the usages of `Rule.Name`, the initial reason for this change.

--- 

We already have a `RuleInstance` that I think is the perfect structure to report the rules. Maybe we need to add some extra information in `RuleInstance` but that doesn't seem any problem (severity is the one that comes to mind).

The problem is that I'm not sure which rules we should report. I see 3 options:

1. Only the rules that were executed
2. Only the active rules (if we don't run `detekt` with full analysis this can be different than the executed ones)
3. All the registered rules.

Also, in the optiosn 2 and 3 we need to add the `active` and/or the `executed` flag to `RuleInstance`.

Personal opinion: the differentiation between "active" and "executed" doesn't have sense in this context. We should treat the not executed rules as not-active. So for me there are two real options:

1. Only the rules that were executed
2. All the registered rules with an `active` property in `RuleInstance` that will mean if the rule was executed or not.

BraisGabin:
Get the rules from the `Detektion` instead of load them again. That code was wrong because we changed how we instantiate the rules. Now the reporter doesn't care about how the rules are created, it receives the information of the configured rules and that's it.

Closes #7779

Blocked by:
- #7865


cortinico (approved):
LGTM 👍

PR #7868, opened 2025-01-03
Remove `open` from `CodeSmell`

BraisGabin:
Now we don't need different types of `CodeSmell`s.

### DIFF
--- a/detekt-api/api/detekt-api.api
+++ b/detekt-api/api/detekt-api.api
@@ -11,13 +11,13 @@ public final class io/gitlab/arturbosch/detekt/api/AutocorrectKt {
 	public static final fun setModifiedText (Lorg/jetbrains/kotlin/psi/KtFile;Ljava/lang/String;)V
 }
 
-public class io/gitlab/arturbosch/detekt/api/CodeSmell : io/gitlab/arturbosch/detekt/api/Finding {
+public final class io/gitlab/arturbosch/detekt/api/CodeSmell : io/gitlab/arturbosch/detekt/api/Finding {
 	public fun <init> (Lio/gitlab/arturbosch/detekt/api/Entity;Ljava/lang/String;Ljava/util/List;Ljava/util/List;)V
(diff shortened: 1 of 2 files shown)


schalkms (commented):
I'm not sure about this change. Yes, the internal detekt code base does not need subtypes of `CodeSmell`s. But what about 3rd party clients? We don't know for sure whether clients derive from `CodeSmell`. It's all part of the `api` though.

BraisGabin:
They can derive it. But for what? This class is mapped by the core to `Issue`. So, if they drive it to add any new parameter that parameter will be lost.

If we want to give that flexibility (and I think it have sense) we should add a property `userData: Map<String, Any>`. Or something like that. But I think that should be done in another pr.

schalkms:
I'm afraid, I don't fully understand. How should the class `CodeSmell` be derived from, if it's final?
A client could implement the `Finding` interface. That's possible for sure.

BraisGabin:
Yes, my explanation was a mess.

Right now anyone could extend from `Finding` or from `CodeSmell`. In this PR I make `CodeSmell` final and in a next PR I would like to remove the interface. I don't know if we should keep the class named `Finding` or `CodeSmell`, but that's a problem for the future.

> A client could implement the Finding interface. That's possible for sure.

Sure! They can do it but the problem is that as soon as this code is executed: 
https://github.com/detekt/detekt/blob/905c5f75b86a7bbd32723568d50b35578942da5f/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt#L187-L194

That instance will be lost. Did they add a new property inside `Finding`? Detekt doesn't care, it will convert `Finding` to `Issue` and any extra property inside `Finding` is ignored. For that reason, you *can* implement `Finding` but it's "pointless".

I think that give the option to add arbitrary information inside `Finding` and propagate that to `Issue` is good. And I think that's something that we should implement. But that's out of the scope of this issue.

schalkms (approved):
Got it. Thanks for the thorough explanation!

PR #7873, opened 2025-01-03
Simplify test api

BraisGabin:
The idea behind this change is that we don't want usages of the old `lint` because it doesn't checks that the code is "real" kotlin. This is bad because we assume that the code is correct to do our analysis so the test could be wrong. But we also know that there are cases where it's easier to just use `lint` with code that doesn't compile for multiple reasons.

This PR what it does is to rename `compileAndLint` to `lint` so the funcion is always the same AND if you in one case doesn't want to compile the snippet of code you can set `.lint(code, compile = false)` to specify that you don't want to verify the snippet. I think that this API is better for this things:

- Less functions, it's better.
- Reviewing the code is easier to spot a `.lint(code, compile = false)` between a lot of `.lint(code)` thatn a `.lint(code)` between a lot of `.compileandLint(code)`. The `compile = false` is easy to spot.
- New users will use the correct function by default.

If this PR is approved I will do the same for `lintWithContext` and `compileAndLintWithContext` but I didn't want to do that refactor too if you don't like this one.


cortinico:
+1 for this approach.

> If this PR is approved I will do the same for `lintWithContext` and `compileAndLintWithContext` but I didn't want to do that refactor too if you don't like this one.

I've skimmed through the review as GitHub UI is not rendering well this diff (it's too big :/)

cortinico (approved):

PR #7874, opened 2025-01-03
Ensure that the correct lint function is called for the passed rule

BraisGabin:
Related with #7873 but they do different things.

This PR ensures that if the rule requires FullAnalysis we call `lintWithContext` and if we don't need it we call `lint`.

If this PR and #7873 are accepted we could even merge `lint` and `lintWithContext`. We could add a parameter to `lint` like this: `environment: KotlinCoreEnvironment = null` and ensure that if the rule requires full analysis it is not null.

This change should make it easier for rule authors to spot that they are not using the correct function. You can even see that we had some tests with this error.

### DIFF
--- a/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutablePropertySpec.kt
+++ b/detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutablePropertySpec.kt
@@ -2,7 +2,6 @@ package io.gitlab.arturbosch.detekt.rules.bugs
 
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.rules.KotlinCoreEnvironmentTest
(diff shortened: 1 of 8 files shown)


detekt-rules-errorprone/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/NullCheckOnMutablePropertySpec.kt
```diff
@@ -250,20 +249,6 @@ class NullCheckOnMutablePropertySpec(private val env: KotlinCoreEnvironment) {
...
-            class A(private var a: Int?) {
-                fun foo() {
-                    if (a != null) {
-                        println(2 + a!!)
-                    }
-                }
-            }
-        """.trimIndent()
-        assertThat(subject.compileAndLint(code)).isEmpty()
-    }
```
BraisGabin, line 265: These tests are not needed anymore. The core is the one that checks if a rule can/should be executed.

schalkms (approved):

Issue #6962, opened 2024-02-10
Create an extension point at rule execution to measure performances


mgroth0:
I would like to measure the performance of rules. I can measure the time it takes to process a file with the `FileProcessListener` extension point. But I do not see an extension point for rules.

I suggest something like:

```kotlin
interface RuleExecutionListener {
    fun startingRuleExecution(file: KtFile,rule: BaseRule)
    fun finishedRuleExecution(file: KtFile, rule: BaseRule, findings: List<Finding>)
}
```

https://github.com/detekt/detekt/blob/bc83921820a81163768ddaa5713d67d949762026/detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/Analyzer.kt#L139

```kotlin
fun executeRules(rules: List<BaseRule>) {
    for (rule in rules) {
        ruleListeners.forEach { it.startingRuleExecution(file,rule) }
        rule.visitFile(file, bindingContext, compilerResources)
        val findings = filterSuppressedFindings(rule, bindingContext)
        ruleListeners.forEach { it.finishedRuleExecution(file,rule,findings) }
        // ...
    }
}
```

For my use case, which is just to measure time, I do not need the list of findings.

Based on idea from @arturbosch here https://github.com/detekt/detekt/discussions/5183. Also might solve that issue.

BraisGabin:
It's great to see someone interested on this topic! I think that the performance on a static analysis tool is key. It is the main reason to don't use it or use it less times.

About your proposal: that should be easy to implement. But I think that a feature like that should be inside detekt. I would love to have to receive issues related with the performance on specific rules so we can fix them.

So, would you be willing to implement that inside detekt?

My problem to implement it is how to handle all that information.

BraisGabin:
Why do you want to do that? From my side, how you develop it is up to you. But to merge your branch you should rebase it back to main or wait for that other branch to be merged (aka wait for kotlin 2.0).


> What is interesting is that ktlint, even with many more rules enabled, runs like 10 times as fast for me. So I am very curious what is happening.

Right now if you enable the debug mode on detekt `detekt { debug = true }` you can get a high level report. I would assume that we expend a lot of time generating the `ContextBinding`.

> I can think of some ideas for handling the information. I will be happy to give this a shot.

If they are ready to imminent you can just go for it. Otherwise I recommend you to share them to see if they align with our ideas. That would probably make the pr review faster.

BraisGabin:
[...]

PR #7882, opened 2025-01-10
Faster MaxLineLength

BraisGabin:
Based on this data: https://github.com/detekt/detekt/issues/6962#issuecomment-2582455298 I checked why `MaxLineLength` was SO slow. And I found out. We were doing the slow checks before the fast ones. If we know that the line is smaller than the threshold we don't need to check anything else, the line is fine. If it's not then we need to do the other checks.

This makes the rule around 6 times faster. I don't have really good measures here, but my "poor" mesurements indicate that it's CLEARLY faster now.


schalkms (approved):
Cool stuff 👍
Did you try it in _release_ mode as well? I suspect that compiler optimisation also play a role here.

BraisGabin:
Do we have a release mode? I don't know how to compile with that. I thought that was just an android thing

schalkms:
`javac -g` should be none, which we don’t set. So everything’s fine.

PR #7934, opened 2025-02-13
Compile detekt with Java 17

3flex:
Two changes:
1. Shift config so the JVM target and compilation toolchain JVM version are set in the version catalog. This seems like the right place for those settings, and also avoids having to recompile the precompiled scripts when the values are updated (though I acknowledge this is rare).
2. Compiles with the much more modern Java 17 instead of Java 8. Java 17 is selected over 21 because Java 21 introduced some new collection utility functions that conflict with Kotlin's, and compiling with 21 without other changes would mean breaking compatibility with Java 20 and below. The proper way to fix this is by setting `-Xjdk-release=1.8` for compilation but we have a unique problem due to a `sun.*` import in detekt-parser. Setting JDK release to 1.8 to restrict API usage to what was available in Java 1.8 uses a file called `ct.sym` which is included in most JDK distributions. `sun.*` was internal API in Java 1.8 so isn't included in the 1.8 public API definition included in `ct.sym`. It's only included in API definition for Java 9+. Until that usage is removed, or we drop Java 1.8 compatibility (which I don't think we should yet) we're left in this awkward position. The only way around it is to include Java 1.8's `rt.jar` on the boot classpath, but that requires additional config that we'd want to avoid in our build.

More info on point 2:
* [JEP 247: Compile for Older Platform Versions](https://openjdk.org/jeps/247)
* [Comment on JEP 247](https://bugs.openjdk.org/browse/JDK-8058150?focusedId=13847552&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13847552)
* [JEP 260: Encapsulate Most Internal APIs](https://openjdk.org/jeps/260)

### DIFF
--- a/build-logic/src/main/kotlin/Versions.kt
+++ b/build-logic/src/main/kotlin/Versions.kt
@@ -1,10 +1,7 @@
-import org.jetbrains.kotlin.gradle.dsl.JvmTarget
-
 object Versions {
 
     const val DETEKT: String = "1.23.7"
     const val SNAPSHOT_NAME: String = "main"
-    val JVM_TARGET: JvmTarget = JvmTarget.JVM_1_8
 
     fun currentOrSnapshot(): String {
         if (System.getProperty("snapshot")?.toBoolean() == true) {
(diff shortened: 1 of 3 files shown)


BraisGabin (approved):
Your pr descriptions are always great! I learn a ton just by reading them.

3flex:
I just like to write things up so I don't forget the details later :D

PR #7938, opened 2025-02-14
Bump Node to v22

3flex:
Node 22 is the current LTS release. Node 18 is EOL in about 2 months (2025-04-30): https://github.com/nodejs/Release/blob/82d65caf1373a8f5c78dd7869a0fc8b796c974f8/README.md#release-schedule


BraisGabin (approved):

website/package.json
```diff
@@ -42,6 +42,6 @@
     ]
   },
   "engines": {
-    "node": ">=18.0"
+    "node": ">=22.0"
```
cortinico, line 45: This is actually unnecessary as we don't depend on anything Node 22 specific

3flex: I realise it's probably not necessary but I mentioned in the commit that Vercel automatically bumps to Node 22 when building so I thought it made sense to align it here so local builds would use the same version.

Feel free to revert though.

PR #7942, opened 2025-02-14
Add RuleInstance inside RuleDescriptor

BraisGabin:
I'm splitting https://github.com/detekt/detekt/pull/7865 in smaller pieces to make the review easier. They have a bit less sense only by them selves but they are simpler this way.

This PR is just a refactor, there is not any behavior change for that reason I don't need to change any test. This will help for future changes.


BraisGabin:
This was merged already with #7865

PR #7943, opened 2025-02-14
Remove PreferToOverPairSyntax

### LINKED ISSUE #7916: PreferToOverPairSyntax should not activate for `a to b to c`
I think the rule should not activate for cases where there is a Pair<*,Pair<*,*>>, e.g. I think this is actually easier to read:

```

                GET to Pair("foo",READ),
```

than the currently enforced:

```
                GET to ("foo" to READ),
```

## Context
This comes up when defining a `mapOf` for a `Map<*,Pair<*,*>>`, e.g.
```
mapOf(
                GET to Pair("foo",READ),
                POST to Pair("bar", UPDATE)
)
```

### LINKED ISSUE #7920: Remove `PreferToOverPairSyntax` for 2.0
`PreferToOverPairSyntax` rule is controversial. There are cases where `Pair()` is better than `to` and it's difficult to add exclusions to those cases. And we even have this rule enabled by default.

An example can be found at #7916

But here is another:

```kotlin
val foo = "hello hello hello hello hello ... hello hello hello hello" to
  "world"

// vs

val foo = Pair(
  "hello hello hello hello hello ... hello hello hello hello",
  "world",
)
```

I vote to remove this rule. I agree that in general `to` is better than `Pair()` but a rule that raises a lot of false positives is not worth to maintain, I think.

BraisGabin:
Fixes #7920, the reason of this PR is there.

also closes #7916


schalkms (approved):

BraisGabin:
I would like, at least, one more approve from the maintainers to merge this because we are "loosing" functionality.

cortinico (approved):

cortinico:
> I would like, at least, one more approve from the maintainers to merge this because we are "loosing" functionality.

I'm fine with removing it 👍 Haven't received many user reports so far related to this rule.

Issue #7939, opened 2025-02-14
Enable additional non-experimental ktlint rules by default


3flex:
## Expected Behavior
All non-experimental ktlint rules are enabled by default (except those that have a duplicate rule in one of detekt's rule sets).

## Current Behavior
Several non-experimental rules are not enabled by default:

* BackingPropertyNaming #7940 
* BinaryExpressionWrapping #7944 
* ChainMethodContinuation #7945 
* ClassSignature #7946
* ConditionWrapping #7947 
* FunctionExpressionBody #7948 
* FunctionLiteral #7949 
* FunctionTypeModifierSpacing #7950 
* MultilineLoop #7951
* TrailingCommaOnCallSite #7956 
* TrailingCommaOnDeclarationSite #7957

## Context
The only experimental rules in ktlint 1.5.0 are:
* Blank line between when-conditions
* KDoc
* Mixed condition operators
* Square brackets spacing
* When-entry bracing

https://pinterest.github.io/ktlint/1.5.0/rules/experimental/

3flex:
Fixed by:

- https://github.com/detekt/detekt/pull/7940
- https://github.com/detekt/detekt/pull/7944
- https://github.com/detekt/detekt/pull/7945
- https://github.com/detekt/detekt/pull/7946
- https://github.com/detekt/detekt/pull/7947
- https://github.com/detekt/detekt/pull/7948
- https://github.com/detekt/detekt/pull/7949
- https://github.com/detekt/detekt/pull/7950
- https://github.com/detekt/detekt/pull/7951
- https://github.com/detekt/detekt/pull/7956
- https://github.com/detekt/detekt/pull/7957

PR #7956, opened 2025-02-15
Enable TrailingCommaOnCallSite by default

3flex:
See #7939

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -95,6 +95,8 @@ formatting:
     active: true
   StringTemplateIndent:
     active: false
+  TrailingCommaOnCallSite:
+    active: false
   ValueArgumentComment:
     active: false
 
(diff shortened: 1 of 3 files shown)


config/detekt/detekt.yml
```diff
@@ -95,6 +95,8 @@ formatting:
     active: true
   StringTemplateIndent:
     active: false
+  TrailingCommaOnCallSite:
+    active: false
```
BraisGabin, line 99: I would like this rule enable on detekt. I think that it helps a lot to reduce git diffs. It could be done in a different pr.

BraisGabin (approved):

config/detekt/detekt.yml
3flex: Completely agree, and also thought it was best to do it in a separate PR - there are hundreds of violations.

PR #7957, opened 2025-02-15
Enable TrailingCommaOnDeclarationSite by default

3flex:
See #7939

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -95,6 +95,8 @@ formatting:
     active: true
   StringTemplateIndent:
     active: false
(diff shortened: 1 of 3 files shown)


config/detekt/detekt.yml
```diff
@@ -95,6 +95,8 @@ formatting:
     active: true
   StringTemplateIndent:
     active: false
+  TrailingCommaOnDeclarationSite:
+    active: false
```
BraisGabin, line 99: I would like to enable this rule on detekt. It helps a lot to reduce git diffs

BraisGabin (approved):

PR #7959, opened 2025-02-18
Add 'code_style' parameter to formatting rule set

### LINKED ISSUE #5991: Support ktlint code styles

Update to ktlint to version `0.49.0` and add support for new features:
- Ktlint code styles

3flex:
Fixes #5991 

This is a notable change as anyone using the `android` config parameter now needs to set `code_style: android_studio`.


detekt-formatting/src/main/resources/config/config.yml
```diff
@@ -1,6 +1,6 @@
 formatting:
   active: true
-  android: false
+  code_style: 'intellij_idea'
```
3flex, line 3: Should this be `ktlint_code_style`? That's the name of the property ktlint reads from editorconfig.

We don't need the `ktlint` prefix though as we don't need to disambiguate, there's only ever going to be one `code_style` property in this location in the config file.

I personally prefer the shorter `code_style`.

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingProvider.kt
```diff
@@ -227,8 +227,8 @@ class FormattingProvider : RuleSetProvider {
     )
 
     companion object {
-        @Configuration("if android style guides should be preferred")
-        val android by ruleSetConfig(false)
+        @Configuration("ktlint code style for formatting rules (ktlint_official, intellij_idea or android_studio)")
+        val code_style by ruleSetConfig("intellij_idea")
```
BraisGabin, line 231: Is this the default value of ktlint? Also, should we enumerate the cake cakes for this field? Or at least link to a page with them?

cortinico (approved):
+1 to this. We'll need to update the docs for this as well though

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingProvider.kt
3flex: > Is this the default value of ktlint?

No, but I was going to create a follow up PR to change to ktlint's default `ktlint_official`. I can do it here though?

> Also, should we enumerate the cake cakes for this field?

Cake cakes 🤣 I don't think we have enum support for rule set config? Is there an example? If an invalid value is passed I'm pretty sure it's marked invalid (I recall testing this but I might double check).

> Or at least link to a page with them?

I can add that - this should work: https://pinterest.github.io/ktlint/latest/rules/code-styles/. Would this link go in the `@Configuration` description?

BraisGabin: > cake cakes

x"D I reviewed from my phone.

My idea was or add something like this on the configuration: [intellij_idea, ktlint_official, android_studio]. No need to use an `enum class`. And/Or adding the link would be great, inside `@Configuration` I think is the correct place.

> No, but I was going to create a follow up PR to change to ktlint's default `ktlint_official`.

Sounds great!

BraisGabin (approved):

detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingProvider.kt
BraisGabin: Oh, the possible values are in the configuration. Ready to merge from my side.

BraisGabin:
@3flex this can be merged, right?

PR #7964, opened 2025-02-20
WIP: Build on Apple Silicon in CI

3flex:

### DIFF
--- a/.github/workflows/codecoverage.yaml
+++ b/.github/workflows/codecoverage.yaml
@@ -14,7 +14,7 @@ permissions:
 
 jobs:
   publish-code-coverage:
-    runs-on: macos-latest
+    runs-on: macos-15
     steps:
       - name: Checkout Repo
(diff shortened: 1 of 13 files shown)


cortinico:
nice :)

3flex:
I wanted to see if this would be faster. Sometimes it is, sometimes it's not, but looking at the pre-merge check total runtime, the difference between a full build without anything precached is [1h 25m 30s](https://github.com/detekt/detekt/actions/runs/13433716936/usage) for the current config, and [1h 25m 8s](https://github.com/detekt/detekt/actions/runs/13435045543/usage) for macOS, making this a pretty pointless change in terms of improving CI runtimes.

I know this is comparing a single run with a single run so we can't consider it a true benchmark but I was still hoping for some better results.

Public builds have M1 chips only, perhaps whenever that gets upgraded to M2-4 we can check again.

PR #7966, opened 2025-02-20
Simplify test api with context

BraisGabin:
This is the follow up that I promised at #7873

The idea is the same. Collapse `lintWithContext` and `compileAndLintWithContext` in a single function with a parameter `compile` that is defaulted to `true`. This way, if someone wants to not compile its snippets s/he needs to add a `compile = false`. That's easy to see on reviews and when you are reading the code. And it's more clear that it's a decision and not just a typo.


BraisGabin:
That's doable. But it's not that simple because we need an `environment: KotlinCoreEnvironment` for the withContext = true. For that reason in this one I'm just squashing those 2 functions.

In future PRs we can try to squash `lint` and `lintWithContext`.

cortinico (approved):

Issue #7867, opened 2025-01-03
Permalink for the current stable version


BraisGabin:
## Expected Behavior
The doc of all the rules for all the versions should have a permalink.

## Current Behavior
Right now the last stable version doesn't have a permalink. 

- https://detekt.dev/docs/1.23.6/rules/exceptions#printstacktrace: `PrintStackTrace` v1.23.6
- https://detekt.dev/docs/rules/exceptions#printstacktrace: `PrintStackTrace` v1.23.7, but unversioned, this content will change
- https://detekt.dev/docs/1.23.7/rules/exceptions#printstacktrace: 404

## Context
Now that we have #7775 we can improve our links. For example add `helpUrl` again in the SARIF report removed at #7686.

Anyway I still think that: https://detekt.dev/docs/rules/exceptions#printstacktrace should point to the last stable version.

cortinico:
> Anyway I still think that: https://detekt.dev/docs/rules/exceptions#printstacktrace should point to the last stable version.

This points to the last stable version (1.23.7), no?

It feels like this is similar to this feature here:
https://github.com/facebook/docusaurus/issues/9049

Practically we'll have to create redirects such as:

`/docs/1.23.7/rules/exception` -> `/docs/rules/exception`

For each page. And we can keep them up to date once we release a new version.

cortinico:
> To be honest I'm clueless about how to create those redirects. I guess that we need this plugin: [docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects). It would be nice to have redirects from server side but I guess that's not doable on Github Pages.

I can take care of those 👍

PR #7970, opened 2025-02-20
Setup redirects /docs/1.23.7/rules/ -> /docs/rules

cortinico:
Fixes https://github.com/detekt/detekt/issues/7867

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -2,7 +2,7 @@
 // Note: type annotations allow type checking and IDEs autocompletion
 
 import { themes as prismThemes } from "prism-react-renderer";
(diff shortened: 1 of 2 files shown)


cortinico:
@BraisGabin I've created redirects only for `/docs/rules/*` as I assume this is mostly needed for rules right? If needed we can also create redirects for the other `/docs/*` pages.

BraisGabin:
> @BraisGabin I've created redirects only for `/docs/rules/*` as I assume this is mostly needed for rules right? If needed we can also create redirects for the other `/docs/*` pages.

For my use case `/docs/rules/*` is enough. I'm not sure if we need it for other pages but if someone needs it can open a new issue.

BraisGabin (approved):

PR #7978, opened 2025-02-21
Rename `CodeSmell` to `Finding`

### LINKED ISSUE #7870: Remove `CodeSmell` in favor of `Finding`
## Expected Behavior
We don't have two names for the same thing.

## Current Behavior
Right now `Finding` is an `interface` implemented by `CodeSmell`. And both of them are in `:detekt-api`

## Context
In detekt 1.0 we had different types of `Findings` so the polymorphism was useful but with the recent changes for `2.0` we end up with a single implementation of `Finding` that is `CodeSmell`. I think that we shouldn't have this two different concepts for the same thing.

Also, this option to expand a `Finding` was great before because the `Detektion` returned exactly what the `Rule`s emitted. But now we have a map `from `Finding` to `Issue` so if someone adds a property to a `Finding` it will get lost. If we want to give this type of flexibility we should add a `Map<String, Any>` to `Finding` but that's another discussion.

## Proposal
Remove the `interface Finding` and rename the `class CodeSmell` to `Finding`. This way we will have single concept and no one can extend from it.

BraisGabin:
Closes #7870 full context on that issue.


cortinico (approved):
I haven't checked the whole diff as it's massive (I did like 50% of it), but it looks good. Great cleanup also :)

PR #7990, opened 2025-02-23
Migrate to Dokka Gradle plugin v2

3flex:
Sample links:
* https://detekt-55r7zy0zk-detekt.vercel.app/kdoc/
* https://detekt-55r7zy0zk-detekt.vercel.app/kdoc/detekt-gradle-plugin/

I followed the migration guide: https://kotlinlang.org/docs/dokka-migration.html. A little trial and error was involved as the migration guide did not cover all required changes.

The main change required was to include the DGP documentation in the main Dokka aggregation task. This was required to maintain the existing documentation URLs, as otherwise the two output tasks from DGP and the main build would overwrite each other. The new config was based on this example project: https://github.com/Kotlin/dokka/tree/0fbf6f5da30f97952f28e25d6a40fb1d580dcbc3/examples/gradle-v2/composite-build-example

The advantage of this new setup is that all modules are documented in the same place, so the left navigation includes DGP + the main build's modules together.

Apart from eliminating the migration warning messages, Dokka v2 allows use of Gradle's build & configuration caches, and I think will not block use of isolated projects as well, but we have some other issues to resolve before we can turn that on for our build.


cortinico (approved):
LGTM 👍

PR #8004, opened 2025-03-02
Add UnnecessaryTypeCasting rule

### LINKED ISSUE #7860: New Rule: Use Type Checking instead of Type Casting

The explicit type cast can be removed to optimise the check.
The compliant case checks for the type only, while the noncompliant case first checks if the instance is of that type and then returns null or casts explicitly to that type.
The rule should not flag cases, where the result of the cast operator, e.g. the casted type variable, is used in subsequent code.

#### Compliant

```Kotlin
val objList: List<Any> = emptyList()
objList.any { it is String }
```

#### Noncompliant

```Kotlin
val objList: List<Any> = emptyList()
objList.any { it as? String != null }
```

They might be more complicated cases, where this rule could be helpful.

## Context

The `performance` rule set should include this new rule.
The idea was initiated in #7858.

paolorotolo:
This PR adds a new Performance rule: `UnnecessaryTypeCasting`.
Context about this has been discussed in #7860.

### Explanation
 * Checks Binary expressions where operation token is `as?`
 * Checks if the previous expression is a `BinaryExpression` with op token `!=` and one of the terms is `null`.

In this case (`variable as? String != null`), it reports a violation.

### Exceptions
Violation is not reported when cast result is stored in another variable, since it could be used later.

We don't detect if the variable is actually used in other parts of the code since it would complicate the rule, and possibly need type resolution. In case we want to pursue this use case as well, we could have 2 variants of the same rule with and without type resolution.

### Example
Non-compliant code:

```kotlin
variable as? String != null
```

Compliant code:
```kotlin
variable is String
```

Close #7860

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -417,6 +417,8 @@ performance:
     active: false
   UnnecessaryTemporaryInstantiation:
     active: true
+  UnnecessaryTypeCasting:
+    active: false
 
 potential-bugs:
   active: true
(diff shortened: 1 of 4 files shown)


BraisGabin (approved):

BraisGabin:
Thanks for the contribution and sorry for the really late review. The rule looks good! Thanks.

PR #8012, opened 2025-03-08
Configure CLI options file generation entirely in detekt-generator project

3flex:
Simplifies things a little bit and keeps the responsibility for generating CLI options file within the detekt-generator project where it belongs.


BraisGabin (approved):
This have a lot of sense 👍

PR #8013, opened 2025-03-10
UnusedParameter: add ignored function/parameter annotation config options

JordanLongstaff:
I have a use case where I want Detekt to ignore unused parameters in functions annotated with my custom annotation, so I've added the option to the config. I also added a similar option for the parameters themselves.

### DIFF
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -797,6 +797,8 @@ style:
     active: true
     aliases: ['UNUSED_PARAMETER', 'unused']
     allowedNames: 'ignored|expected'
+    ignoreAnnotatedFunctions: []
+    ignoreAnnotatedParameters: []
   UnusedPrivateClass:
     active: true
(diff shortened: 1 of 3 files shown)


cortinico:
We do already have `ignoreAnnotated` which is a suppressor and applies to all the rules:
https://detekt.dev/docs/introduction/suppressors/

Have you tried using it before adding 2 custom props?

JordanLongstaff:
Well, no, I hadn't. Looks like that does work, as you say.

But the `LongParameterList` and `TooManyFunctions` rules have their own explicit annotation suppressor config options, so I assumed these two would also be needed.

cortinico:
> But the `LongParameterList` and `TooManyFunctions` rules have their own explicit annotation suppressor config options

That's because they provide more fine grained control that the top level `ignoreAnnotated` won't offer for specific scenarios.

For example `LongParameterList` would allow to ignore only some of the parameters with `ignoreAnnotatedParameter`, while `ignoreAnnotated` will suppress the whole class/function.

We can close this then 👍

Issue #7823, opened 2024-12-05
(Documentation suggestion) Clarify functional type handling of the ForbiddenMethodCall's `methods`


saiya:
## Expected Behavior of the rule

I would like to suggest some clarification on the [`methods` option of the style.ForbiddenMethodCall's document](https://detekt.dev/docs/rules/style/#forbiddenmethodcall) to explain how to write functional parameter type.

It would be something like this:

> To match with function type parameter such as `initializer` of `fun <T> lazy(intializer: () -> T)`, write `kotlin.Function{N}` (N = number of parameters) as a parameter type like this: `kotlin.lazy(kotlin.Function0)`

## Context

Consider a usage like this:

>   ForbiddenMethodCall:
>       - value: "kotlin.lazy(kotlin.Function0)"
>         reason: Must use `lazy(LazyThreadSafetyMode.PUBLICATION)` for performance and also compatibility with JVM virtual thread

Because Kotlin compiler frontend [internally uses "fictitious (synthetic) classes" `kotlin.Function*`](https://github.com/JetBrains/kotlin/blob/master/spec-docs/function-types.md#function0-function1--types) to model function type (that don't appear in the source code or compiled bytecode), users of the `ForbiddenMethodCall` rule need to understand that compiler frontend specific idea to configure ForbiddenMethodCall to match with particular high order function overload.

The [style.ForbiddenMethodCall's document](https://detekt.dev/docs/rules/style/#forbiddenmethodcall) does not have any hint of that matter and also IDE code jump or decompilation never shows `kotlin.Function*` type information, thus it is hard for normal users who don't have knowledge of Kotlin compiler frontend internals.

BraisGabin:
You are right. Can you open a pr with that change?

On the other hand, I think that we can make it even better. We know what those `kotlin.Function0` means. So we could translate `() - > kotlin.Any` to it. That should improve the user experience. This is a bit more complex so we can start with the documentation and later implement this.

PR #8014, opened 2025-03-11
ForbiddenMethodCall: add forbidden method config options

kimmandoo:
I modified the code after looking at issue #7823.

I suggested how to configure `ForbiddenMethodCall` for function type parameters in the documentation and improves its enforcement by ensuring that `lazy {}` is replaced with `lazy(LazyThreadSafetyMode.PUBLICATION) {}`.

This PR only enforces lazy {} to use LazyThreadSafetyMode.PUBLICATION, not other modes.

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -219,6 +219,7 @@ style:
       - 'java.lang.Class.getResourceAsStream'
       - 'java.lang.ClassLoader.getResourceAsStream'
       - 'kotlin.system.measureTimeMillis'
+      - 'kotlin.lazy(() -> kotlin.Any)'
   ForbiddenVoid:
     active: true
     ignoreOverridden: true
--- a/detekt-core/src/main/resources/default-detekt-config.yml
+++ b/detekt-core/src/main/resources/default-detekt-config.yml
@@ -646,6 +646,8 @@ style:
         value: 'java.math.BigDecimal.<init>(kotlin.String)'
       - reason: 'It is marked as obsolete. Use `kotlin.time.measureTime` instead.'
         value: 'kotlin.system.measureTimeMillis'
+      - reason: 'Must use `lazy(LazyThreadSafetyMode.PUBLICATION)` for performance and also compatibility with JVM virtual thread'
+        value: 'kotlin.lazy(() -> kotlin.Any)'
   ForbiddenNamedParam:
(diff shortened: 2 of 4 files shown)


cortinico:
> I modified the code after looking at issue #7823.

The issue was related to the syntax of the `method:` config option.

Here instead you're adding another default value for it, so this is not the right fix.

> by ensuring that lazy {} is replaced with lazy(LazyThreadSafetyMode.PUBLICATION) {}.

On top of this, we would need to clarify why this is a recommendation. Is there an official reference about it?

kimmandoo:
> On top of this, we would need to clarify why this is a recommendation. Is there an official reference about it?

I'm Sorry. I was misunderstanding about the issue comment. And there isn't any an official reference about `LazyThreadSafetyMode`.

I mistakenly understood this issue as needing to modify the `ForbiddenMethodCall` rule.

Could you clarify what was requested in this issue? I'm not entirely sure what PR BraisGabin was referring to.
Should the PR include additional content in the `ForbiddenMethodCall` documentation located at `/website/versioned_docs/version-1.23.8/rules/style.md`?

cortinico:
> Could you clarify what was requested in this issue?

You would have to update the KDoc on top of the `ForbiddenMethodCall` class so that this page documentation is updated: https://detekt.dev/docs/1.23.7/rules/style/#forbiddenmethodcall

kimmandoo:
@cortinico Could you check on my commit? Thank you for your kindness.

cortinico (commented):
cc @BraisGabin can you take a look at this one?

kimmandoo:
@BraisGabin 
Could you check on my commit? Please let me know if there is any issue or what i have to do. Thank you for your kindness.

BraisGabin (commented):
I would move the documentation to the `@configuration` below. There are other explanations similar to this one there. Also, please reduce it reduce it to fit the wording to align with those examples.

I realised that we don't need to talk about `function{N}`. This rule already supports lambda. This is my bad, I thought we didn't. 

 And sorry for the late review. Crazy weeks.

3flex:
@kimmandoo do you intend to update the PR based on the feedback provided?

kimmandoo:
Sorry to late reply, I don't intend to update the docs. Thank you for your feedback. I will close pr myself.

PR #8029, opened 2025-03-19
Migrate from kotlin-compiler-embeddable to kotlin-compiler

### LINKED ISSUE #8027: Migrate from kotlin-compiler-embeddable to kotlin-compiler
## Expected Behavior
We use dependencies that work properly with the Kotlin Analysis API.

## Current Behavior
We use `kotlin-compiler-embeddable` as our compiler dependency which is incompatible with AA.

## Context
`kotlin-compiler-embeddable` is not compatible with the Kotlin Analysis API - AA has IntelliJ dependencies which are relocated in embeddable causing runtime failures unless the standard compiler dependency is used. We could do some trickery with package relocation, but using the standard compiler artifact will be simpler and more performant when building.

detekt migrated to the embeddable compiler in 0de7b72845d715c138945d4bb296def07a0f757a, first released in [M11](https://github.com/detekt/detekt/releases/tag/M11). There had been some complaints about dependencies conflicting with other plugins. Looking at the Gradle plugin from back then, this would have been the problem:

https://github.com/detekt/detekt/blob/93d7cde3720ea5c49ba8b4bd4bfc64fabc143220/detekt-gradle-plugin/src/main/kotlin/io/gitlab/arturbosch/detekt/DetektCheckTask.kt#L20-L22

The problem there was that the detekt-cli dependency was being added as a build(script) configuration dependency. We don't do that anymore, it's added in its own isolated configuration, so even if we switch back to the normal compiler dependency we shouldn't create new conflicts with other plugins.

Required for #8021.

3flex:
Closes #8027 

I suggest reviewing by commit, as the majority of line changes are updates to import declarations.

There are slightly cleaner ways of repackaging ktlint to make it compatible with the `kotlin-compiler` dependency - probably by using [Gradle's artifact transforms](https://docs.gradle.org/current/userguide/artifact_transforms.html) and something like [ASM](https://asm.ow2.io/) or [Byte Buddy](https://bytebuddy.net), but as I'm not familiar with any of these, that would take some time to investigate. In the meantime the shadow plugin should work just fine.

detekt-compiler-plugin tests were disabled. To avoid this, would likely need some kind of packaging relocation trickery similar to what was done for ktlint, but I don't think it's worth it at this stage. Please review my inline comment and linked commit message for more details.


detekt-compiler-plugin/src/test/kotlin/io/github/detekt/compiler/plugin/CompilerTest.kt
```diff
@@ -2,8 +2,10 @@ package io.github.detekt.compiler.plugin
 
 import io.github.detekt.compiler.plugin.util.CompilerTestUtils.compile
 import io.github.detekt.compiler.plugin.util.assertThat
+import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 
+@Disabled("https://github.com/tschuchortdev/kotlin-compile-testing/issues/234")
```
3flex, line 8: Please review the associated commit https://github.com/detekt/detekt/pull/8029/commits/12165bfdc43166ffd5551b22d74a7f1f596d3154

I don't know if this can ever be fixed, except with some tricky packaging workarounds & package relocation. I don't think it's worth the time investment for this PR.

BraisGabin: I'm OK with this.

BraisGabin (approved):
LGTM

PR #8037, opened 2025-03-23
Compile test snippets with Analysis API

3flex:
WIP - should be able to create a separate job for compiling with AA so for a time we'll support the old & new test snippet compilation methods.

### DIFF
--- a/detekt-test-utils/build.gradle.kts
+++ b/detekt-test-utils/build.gradle.kts
@@ -3,9 +3,16 @@ plugins {
     id("public-api")
 }
 
+repositories {
+    mavenCentral()
+    maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/kotlin-ide-plugin-dependencies")
+}
+
 dependencies {
     api(libs.kotlin.stdlib)
     api(libs.junit.jupiterApi)
+    runtimeOnly(projects.detektKotlinAnalysisApi)
+    compileOnly(libs.bundles.kotlin.analysisApi) { isTransitive = false }
     implementation(projects.detektParser)
     implementation(libs.kotlin.mainKts)
     implementation(libs.kotlinx.coroutinesCore)
--- a/detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinAnalysisAAEngine.kt
+++ b/detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinAnalysisAAEngine.kt
@@ -1,0 +1,102 @@
+package io.github.detekt.test.utils
+
+import org.intellij.lang.annotations.Language
+import org.jetbrains.kotlin.analysis.api.KaExperimentalApi
+import org.jetbrains.kotlin.analysis.api.KaImplementationDetail
+import org.jetbrains.kotlin.analysis.api.KaPlatformInterface
+import org.jetbrains.kotlin.analysis.api.analyze
+import org.jetbrains.kotlin.analysis.api.components.KaCompilationResult
+import org.jetbrains.kotlin.analysis.api.components.KaCompilerTarget
+import org.jetbrains.kotlin.analysis.api.diagnostics.KaDiagnosticWithPsi
+import org.jetbrains.kotlin.analysis.api.diagnostics.KaSeverity
+import org.jetbrains.kotlin.analysis.api.projectStructure.KaModule
+import org.jetbrains.kotlin.analysis.api.projectStructure.contextModule
+import org.jetbrains.kotlin.analysis.api.standalone.buildStandaloneAnalysisAPISession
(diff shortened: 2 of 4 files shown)


detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinAnalysisApiEngine.kt
```diff
@@ -0,0 +1,120 @@
...
+            }
+
+            if (result is KaCompilationResult.Failure) {
+                val errors = result.errors.joinToString("\n") {
+                    if (it is KaDiagnosticWithPsi<*>) {
+                        val lineAndColumn = PsiDiagnosticUtils.offsetToLineAndColumn(
+                            it.psi.containingFile.viewProvider.document,
+                            it.psi.textOffset
+                        )
+                        "${it.severity.name} ${it.defaultMessage} (${it.psi.containingFile.name}:${lineAndColumn.line}:${lineAndColumn.column})"
```
BraisGabin, line 106:
```suggestion
```

This line can be removed, right?

3flex: We can probably build a detekt rule for this!

detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinAnalysisApiEngine.kt
```diff
@@ -0,0 +1,119 @@
...
+import org.jetbrains.kotlin.platform.jvm.JvmPlatforms
+import org.jetbrains.kotlin.psi.KtFile
+import java.io.File
+import java.nio.file.Path
+import kotlin.io.path.Path
+
+/**
+ * The object to use the Kotlin Analysis API for code compilation.
+ */
+class KotlinAnalysisApiEngine(@Language("kotlin") code: String, disposable: Disposable) {
```
BraisGabin: Do we want this as a class instead of a top level function? I don't see why I would call `compile()` 0 or more than once. Or we can reuse the class for multiple snippets or we should make this just a top level function.

Also, we don't need to pass the disposable, right? It can be an implementation detail.

3flex: That's way better, updated. I left the `KotlinAnalysisApiEngine` (now an object) around the `compile` function to match the `KotlinScriptEngine` API.

3flex: We can't reuse the class because the analysis API session is immutable and can't reindex files, so reusing the session with new files means some test snippet compilations will interact with others which isn't what we want.

Hopefully this will address that need in future: https://youtrack.jetbrains.com/issue/KT-67151/Analysis-API-Standalone-Support-incremental-re-indexing-of-additional-modified-source-files

3flex:
@BraisGabin on your other comment...

> We could use the old compiler for that, right?

For now, yes, but the old compiler is being retired in a future Kotlin 2.x release. Or did you mean just using an older version of Kotlin to compile snippets? Also not really sure why we'd need or want to do this?

detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinAnalysisApiEngine.kt
```diff
@@ -0,0 +1,123 @@
...
+     * Compiles a given code string using Kotlin's Analysis API.
+     *
+     * @throws IllegalStateException if the given code snippet does not compile
+     */
+    @Suppress("LongMethod")
+    fun compile(@Language("kotlin") code: String) {
+        val disposable = Disposer.newDisposable()
+
+        @OptIn(KaImplementationDetail::class, KaPlatformInterface::class)
+        val session = buildStandaloneAnalysisAPISession(disposable, unitTestMode = true) {
```
BraisGabin, line 49: I don't care that much about long methods but we could move the sesión creation to a private function to avoid it.

BraisGabin (approved):
Now the api looks way better :)

BraisGabin:
> @BraisGabin on your other comment...
> 
> > We could use the old compiler for that, right?
> 
> For now, yes, but the old compiler is being retired in a future Kotlin 2.x release. Or did you mean just using an older version of Kotlin to compile snippets? Also not really sure why we'd need or want to do this?

I was just thinking how critical was this change to move to AA. Then I realised that it was not. It was just critical to support new versions of Kotlin. And last, I remembered than one of the more important points to move to AA is to support new versions of Kotlin. For that reason all that comment has really few sense.

PR #8098, opened 2025-04-22
Kotlin 2.2.0

3flex:


build-logic/src/main/kotlin/module.gradle.kts
```diff
@@ -71,6 +71,7 @@ kotlin {
     compilerOptions {
         progressiveMode = true
         allWarningsAsErrors = providers.gradleProperty("warningsAsErrors").orNull.toBoolean()
+        freeCompilerArgs.add("-Xjvm-default=all-compatibility")
```
BraisGabin, line 74: In 2.2 we should use this dsl, right?

```korling
jvmDefault = JvmDefaultMode.All_COMPATIBILITY
```

3flex: Short answer yes, but because DGP is being compiled with an older Kotlin version so we can continue using API version 1.4 to maintain Gradle 7.6 compatibility, we can't use the new DSL on all our projects. The older compiler flags will be maintained though at least for now.

Alternative is to condition the application of the new DSL to avoid applying it to DGP but I really don't like adding a conditional on a project name to the Gradle convention plugins.

BraisGabin:
I can reproduce these errors locally, that's good. What I found so far:

- When we compile nearly empty files like the ones in `PackageNamingSpec` we get this error: `Compilation produced no matching output files`. I was able to track it down to this line: https://github.com/JetBrains/kotlin/blob/v2.2.0/analysis/analysis-api-fir/src/org/jetbrains/kotlin/analysis/api/fir/components/KaFirCompilerFacility.kt#L627
  - If instead of `package Foo.Bar` the code is `package Foo.Bar; val a = 1` this error goes away. I have no idea why. I will try to find an issue on kotlin about this. Otherwise I will open an issue because it is really strange that this restriction is only applied to MacOS.
- Junit5 fails to load the classes like `NoNameShadowingSpec` and `MemberNameEqualsClassNameSpec`. But this looks like an issue on our setup. Because if I remove all the test classes except one of those they work.
This is part of the stack trace, looks like an issue on the `close`:
```
java.lang.AssertionError: Write access is allowed inside write-action only (see Application.runWriteAction()); see https://jb.gg/ij-platform-threading for details
Current thread: Thread[Test worker,5,main] 1746467716 (EventQueue.isDispatchThread()=false)
SystemEventQueueThread: null
	at com.intellij.openapi.diagnostic.DefaultLogger.error(DefaultLogger.java:83)
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:436)
	at com.intellij.openapi.util.ObjectTree.handleExceptions(ObjectTree.java:195)
	at com.intellij.openapi.util.ObjectTree.runWithTrace(ObjectTree.java:141)
	at com.intellij.openapi.util.ObjectTree.executeAll(ObjectTree.java:162)
	at com.intellij.openapi.util.Disposer.dispose(Disposer.java:205)
	at com.intellij.openapi.util.Disposer.dispose(Disposer.java:193)
	at io.github.detekt.test.utils.KotlinCoreEnvironmentWrapper.dispose(KotlinCoreEnvironmentWrapper.kt:23)
	at io.gitlab.arturbosch.detekt.rules.KotlinEnvironmentResolver$CloseableWrapper.close(KotlinEnvironmentTestSetup.kt:56)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.stream.SortedOps$RefSortingSink.end(SortedOps.java:395)
	at java.base/java.util.stream.Sink$ChainedReference.end(Sink.java:258)
	at java.base/java.util.stream.Sink$ChainedReference.end(Sink.java:258)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:510)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
Caused by: com.intellij.openapi.diagnostic.RuntimeExceptionWithAttachments: Write access is allowed inside write-action only (see Application.runWriteAction()); see https://jb.gg/ij-platform-threading for details
Current thread: Thread[Test worker,5,main] 1746467716 (EventQueue.isDispatchThread()=false)
SystemEventQueueThread: null
	at app//com.intellij.util.concurrency.ThreadingAssertions.createThreadAccessException(ThreadingAssertions.java:180)
	at app//com.intellij.util.concurrency.ThreadingAssertions.throwThreadAccessException(ThreadingAssertions.java:174)
	at app//com.intellij.util.concurrency.ThreadingAssertions.assertWriteAccess(ThreadingAssertions.java:169)
	at app//com.intellij.psi.impl.file.impl.FileManagerImpl.clearViewProviders(FileManagerImpl.java:175)
	at app//com.intellij.psi.impl.file.impl.FileManagerImpl.dispose(FileManagerImpl.java:170)
	at app//com.intellij.psi.impl.PsiManagerImpl.dispose(PsiManagerImpl.java:75)
	at app//com.intellij.openapi.util.ObjectTree.runWithTrace(ObjectTree.java:130)
	... 17 more
```

I'm using this command to test this `./gradlew :detekt-rules-naming:test --rerun -PenablePTS=false -Pcompile-test-snippets-aa=true`. `detekt-rules-style` has the same problems but it is way slower, for that reason I will try to fix `naming` with the idea that the fixes in naming will apply on style.

I keep working on this.

BraisGabin:
Ok, with 6ca1c751909ba5b993d857f12cb8a65b32bec057 we should be fine about the first point. I still have no idea why this only happens on MacOS, I need to look at the kotlin tracker or raise a new issue.

About the second problem. I found out that the tests themselve work. The problem is that when we `dispose` the `KotlinEnvironmentContainer` after we run the tests. Also I know that this happens for some strange side effects. I got it "fixed" on the `naming` rule set with this commit: 6c38d7af20d55789149ff2f55754f3d405e0bc92 I just pushed to show it, I don't think that's a good solution at all. But again, why this only happens on MacOS?

3flex:
Empty files not being generated is just new behaviour and not an issue to be raised, Kotlin's analysis API behaviour has changed there. If there's no bytecode to generate then no file is generated. Rather than working around with code I think we just disable compilation for those, what do you think?

The workaround for the second issue is odd - it's strange that it works. Very strange that it's only macOS too.

BraisGabin:
> Rather than working around with code I think we just disable compilation for those, what do you think?

I don't like the code of this "hack" either. But this is just to test the test. My position is that we shouldn't care about this behaviour change. The code is correct so we shouldn't need `compile = false` because the code is correct. And a "real compiler" would not complain about it. So, to me, force the caller of the lint function to set `compile = false` when the code is correct should not happen. I know that we already have exceptions for this but at least we should try our best to accomplish this.

BraisGabin:
> The workaround for the second issue is odd - it's strange that it works. Very strange that it's only macOS too.

Sooooooo odd. I want to investigate a bit more about it. As I said I pushed it just to show how strange it is. I would prefer not to merge the workaround commit.

BraisGabin:
Wait, on CI this is failing on Linux. So these issues are not related with MacOS. It's just that it doesn't with 2.2.

BraisGabin:
Well... I have no idea why. But this works now.

BraisGabin:
@3flex can you "approve" this PR? To me it looks fine. I can approve it but I did the last changes so it feels like cheating if I approve and merge.

3flex:
About to do some long haul travel long haul so it won't be for a couple of days - happy to merge if you're happy and we can adjust anything if needed later, up to you!

BraisGabin:
> happy to merge if you're happy and we can adjust anything if needed later, up to you!

Then I'm going to merge it. This unlock the context parameters. And if we find any issue we can always fix them later.

BraisGabin (approved):

BraisGabin:
Great! Back to base 1. It fails again. So it's flaky...

3flex:
@BraisGabin I think I have a solution - https://github.com/detekt/detekt/pull/8098/commits/19c8fff3974c73f84101e8faf8b0c79fab341c9f

Works locally and the first CI run passed. Would be good to get some additional testing done on your end.

For a little more context: https://github.com/JetBrains/kotlin/blob/631e9fdbe59cc92ac1e88e6ff3bbf4418b3ab9da/compiler/cli/cli-base/src/org/jetbrains/kotlin/cli/jvm/compiler/KotlinCoreApplicationEnvironment.kt#L133-L139

In our usage we're not actually unit testing Analysis API itself, so I think it's more correct to not enable the unit testing mode. If unit testing mode is not enabled then a standard `MockApplication` is used to create the Kotlin environment, and that has different behaviour which works here. Disabling unit test mode is also the default when creating a new AA session so unless there's a good reason to enable it we should leave the default. I recall the reason for enabling in the first place was in an attempt to reduce memory usage through experimentation but I don't think it actually has any impact.

If you're happy with that I'd like to split these changes into 3 separate PRs:
1. Your workaround for the compilations that don't generate output files - #8294
3. Disabling unit test mode for the AA session - #8293 
4. The Kotlin update itself along with other required changes - will rebase this PR once 1 & 2 are done

Let me know what you think

BraisGabin:
That sounds like a good plan.

3flex:
@BraisGabin can you open a PR for item 1 please? I can do it and cherry-pick but the approval will be a bit funny (I couldn't approve my own PR and it would be odd for you to approve a PR that only contains your own commits - though I suppose if I open it and you approve it then we still have 2 people involved so I might do it later today if you can't get to it).

3flex:
Thanks @BraisGabin !

When I rebase this I'm planning to remove all redundant commits, many are yours, are you happy for me to do so?

If not I'll close this PR and open a fresh new one.

BraisGabin:
I'm completely fine with that. but I'm fixing a new problem that arised. An usage of `com.intellij.util.containers.addIfNotNull` that it looks that it was removed. So give me some minutes to fix that problem.

I'm also running the tests on my local machine to ensure that the issue is gone now.

BraisGabin:
I created #8297. After that this should be good to go. You can squash everything if you want. The commits are a mess. I didn't want to make any `push --force` to avoid problems because it was kind of a "shared" branch.

BraisGabin (approved):

Issue #8059, opened 2025-03-28
Eliminate direct usage of KotlinCoreEnvironment


3flex:
## Expected Behavior
detekt doesn't use KotlinCoreEnvironment directly.

## Current Behavior
This is passed around liberally within detekt.

## Context
This becomes a problem with the Analysis API. Analysis sessions are created by using a `StandaloneAnalysisAPISession` which uses KotlinCoreEnvironment internally to set up the analysis environment but it is not exposed anywhere.

Within detekt this is mostly used as a container for passing around `Project` and `CompilerConfiguration`. This can be done in other ways. `CompilerConfiguration` is created by detekt and can be used as an input into the analysis session as needed.

detekt should only use `Project` from the `StandaloneAnalysisAPISession`.

https://github.com/JetBrains/kotlin/blob/658a2010b15a22583f9841e1a2d4bddf1baac612/analysis/analysis-api-standalone/src/org/jetbrains/kotlin/analysis/api/standalone/StandaloneAnalysisAPISession.kt

PR #8105, opened 2025-04-25
Remove direct usage of KotlinCoreEnvironment in KtCompiler

3flex:
Part of #8059

### DIFF
--- a/detekt-parser/src/main/kotlin/io/github/detekt/parser/KtCompiler.kt
+++ b/detekt-parser/src/main/kotlin/io/github/detekt/parser/KtCompiler.kt
@@ -2,13 +2,18 @@ package io.github.detekt.parser
 
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFileManager
+import com.intellij.pom.PomModel
 import com.intellij.psi.util.PsiUtilCore
+import org.jetbrains.kotlin.analysis.api.standalone.buildStandaloneAnalysisAPISession
+import org.jetbrains.kotlin.cli.common.CliModuleVisibilityManagerImpl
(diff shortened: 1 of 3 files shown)


BraisGabin (approved):

detekt-parser/src/main/kotlin/io/github/detekt/parser/KtCompiler.kt
```diff
@@ -1,21 +1,41 @@
...
+            "$path cannot be retrieved as a VirtualFile"
+        }
+
+        return requireNotNull(PsiUtilCore.getPsiFile(project, virtualFile) as? KtFile) {
             "$path is not a Kotlin file"
         }
     }
 }
+
+private val createDefaultAnalysisAPISession = buildStandaloneAnalysisAPISession {
```
BraisGabin, line 32: Is this expensive? Are we going to use the default value a lot? Maybe this should be lazy to avoid extra work when loading this file.

3flex: I'm not sure how expensive it is, but it would be doing similar things to the old method also created the `Project` instance and created a `KotlinCoreEnvironment` etc.

I don't think it needs to be lazy. This default is only used in tests and by `detekt-generator`.

BraisGabin: That's my point. This property is going to be instantiated on production when we only use it for testing.

3flex: Ah right. I'll make it a function instead.

https://github.com/detekt/detekt/pull/8107

PR #8110, opened 2025-04-26
Use StandaloneAnalysisAPISession to setup analysis environment

3flex:
Last part of #8059 

Note the increased memory requirement for the build. This is due to https://youtrack.jetbrains.com/issue/KT-71706. There is a workaround available. I've created an issue to track and will fix in a separate PR:
* #8113

**Breaking change:** This is the PR that breaks detekt-compiler-plugin - 1.23.8 will be the final working version.

### DIFF
--- a/detekt-compiler-plugin/build.gradle.kts
+++ b/detekt-compiler-plugin/build.gradle.kts
@@ -64,6 +64,8 @@ val unzipKotlinCompiler by tasks.registering(Copy::class) {
 }
 
 val testPluginKotlinc by tasks.registering(Task::class) {
+    enabled = false
+
     val outputDir = layout.buildDirectory.dir("tmp/kotlinc")
     val sourceFile = file("src/test/resources/hello.kt")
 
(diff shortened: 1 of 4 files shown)


BraisGabin (approved):
LGTM.

Should we remove the compiler plugin directly? (in another pr)

BraisGabin:
Also close #8059, right?

3flex:
> Should we remove the compiler plugin directly? (in another pr)

It was late last night when I finished this PR and didn't think through other options - it should be possible to keep it around but perhaps only supporting PSI-based checks in K2 mode, or it should also be possible to refactor to bypass the code that sets up the analysis environment since the compiler already has `BindingContext`, `Project` and `CompilerConfiguration` so there's no need to recreate any of that when the compiler plugin is used.

I'll merge for now and create a new issue for that.

> Also close https://github.com/detekt/detekt/issues/8059, right?

This is linked and should automatically close it. I recently became aware of this feature: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#manually-linking-a-pull-request-to-an-issue-using-the-pull-request-sidebar

And I've linked the 5 PRs that together addressed #8059 using that method.

PR #8136, opened 2025-05-18
Don't generate compatibility bridges and DefaultImpls classes for default implementations in interfaces

3flex:
There is new default behaviour in Kotlin 2.2.0. For details refer to:
* https://kotlinlang.org/docs/whatsnew-eap.html#changes-to-default-method-generation-for-interface-functions
* https://youtrack.jetbrains.com/issue/KT-71768/Enable-Xjvm-defaultall-compatibility-by-default-to-generate-JVM-default-interface-methods
* https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-generating-default-methods-in-interfaces/

Switching to `all` (or with the new flag introduced in 2.2.0, `-jvm-default=no-compatibility`) is a better option for detekt v2 release. We don't need binary compatibility (since lots of API changes are being introduced) and it reduces the number of generated methods & classes.

### DIFF
--- a/build-logic/src/main/kotlin/module.gradle.kts
+++ b/build-logic/src/main/kotlin/module.gradle.kts
@@ -71,6 +71,7 @@ kotlin {
     compilerOptions {
         progressiveMode = true
         allWarningsAsErrors = providers.gradleProperty("warningsAsErrors").orNull.toBoolean()
+        freeCompilerArgs.add("-Xjvm-default=all")
     }
 }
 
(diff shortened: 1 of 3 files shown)


BraisGabin (approved):
Once again, thanks for the links.

I vote to use the no compatibility mode. As you said we are breaking the binary compatibility already.

3flex:
@BraisGabin I've updated the PR per your comment:

> I vote to use the no compatibility mode. As you said we are breaking the binary compatibility already.

BraisGabin (approved):
Nice!! Reduce our abi is great.

PR #8154, opened 2025-06-03
Fully configure standalone Analysis API session

### LINKED ISSUE #8047: Add supporting infrastructure to allow usage of `analyze` entry point in rules
## Expected Behavior
detekt-parser creates a Standalone Analysis API session that can be used by rules that require support for semantic analysis.

## Current Behavior
Semantic analysis is supported through generation of a `BindingContext`.

## Context
Required for migration to Analysis API.

3flex:
Closes #8047


detekt-core/src/main/kotlin/io/gitlab/arturbosch/detekt/core/tooling/Lifecycle.kt
```diff
@@ -40,7 +39,7 @@ internal interface Lifecycle {
 
     fun analyze(): Detektion {
         measure(Phase.ValidateConfig) { checkConfiguration(settings, baselineConfig) }
-        val filesToAnalyze = measure(Phase.Parsing) { parsingStrategy.invoke(settings) }
+        val filesToAnalyze = measure(Phase.Parsing) { settings.ktFiles }
```
BraisGabin, line 42: This measuremwnt have little sense now, right? It will be always "automatic".

3flex: You're right and I had the same thought. Probably what we want to measure now is creating the analysis session though I think that will be very quick and not worth measuring, since AA is lazy when doing analysis so there shouldn't be a large upfront cost to set things up.

Should we just remove it then in this PR?

3flex: I'll merge as is and follow up with a new PR if needed once you have a chance to respond :)

BraisGabin: We need to improve a lot our time measurement in general. I guess that, for now, we can just remove it for now. Or keep it and keep it in mind.

PR #8156, opened 2025-06-03
Add RequiresAnalysisApi interface

### LINKED ISSUE #8155: Add RequiresAnalysisApi marker interface
## Expected Behavior
Adding a marker interface for rules that use the Analysis API will assist with migration and is also helpful for test infrastructure.

## Current Behavior
No such interface exists.

## Context
Supports #8021

3flex:
Closes #8155 

Also adds support for testing rules that have migrated from `RequiresFullAnalysis` to `RequiresAnalysisApi`

Stacks on:
* #8037

### DIFF
--- a/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
+++ b/detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
@@ -28,6 +28,8 @@ internal class RuleVisitor(textReplacements: Map<String, String>) : DetektVisito
     private val configurationCollector = ConfigurationCollector()
     private val classesMap = mutableMapOf<String, Boolean>()
     private var deprecationMessage: String? = null
+    private val fullAnalysisInterfaces =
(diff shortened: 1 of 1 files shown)


detekt-generator/src/main/kotlin/io/gitlab/arturbosch/detekt/generator/collection/RuleVisitor.kt
```diff
@@ -95,7 +96,10 @@ internal class RuleVisitor(textReplacements: Map<String, String>) : DetektVisito
 
         autoCorrect = classOrObject.isAnnotatedWith(AutoCorrectable::class)
         requiresFullAnalysis = classOrObject.superTypeListEntries
-            .any { it.text.substringAfterLast(".") == RequiresFullAnalysis::class.simpleName }
+            .any {
+                it.text.substringAfterLast(".") in
+                    setOf(RequiresFullAnalysis::class.simpleName, RequiresAnalysisApi::class.simpleName)
```
BraisGabin: This should be moved outside the loop to avoid multiple creations of the same set.

We need better benchmarks but I would like to see what happen if we apply R8 or any other optimizer. Maybe that would help us with the performance and also things like this kind of changes would not be needed.

BraisGabin (approved):

Issue #8041, opened 2025-03-24
Migrate detekt-rules-errorprone to use Kotlin AA for semantic analysis


3flex:
## Expected Behavior of the rule
The following rules are migrated to use Kotlin AA for semantic analysis:
* TBC

## Context
Required for the migration to Kotlin's Analysis API.

3flex:
Closed by:
- #8172
- #8191
- #8157
- #8173
- #8174

And these, which are all thanks to @t-kameyama!
- #8232
- #8233
- #8234
- #8235
- #8236
- #8237
- #8435
- #8239
- #8240
- #8415
- #8241
- #8332
- #8316
- #8317
- #8314
- #8321
- #8242
- #8243
- #8244
- #8245
- #8246
- #8247
- #8408

PR #8157, opened 2025-06-03
Migrate MapGetWithNotNullAssertionOperator to Analysis API

3flex:
First rule migrated to Kotlin Analysis API 🥇 

Part of #8041 

Stacks on:
* #8037 
* #8153
* #8154 
* #8156


BraisGabin (approved):

BraisGabin:
Niceeeeee! This week I will rename `Rule.Name` and also fix the full qualified names usages that it is forcing to use.

Issue #8009, opened 2025-03-07
Intermittent build failures with Java 21


3flex:
## Expected Behavior
No fatal errors reported by the JRE when building detekt.

## Observed Behavior
Intermittent failures on Java 21, for example: https://github.com/detekt/detekt/actions/runs/13706931551/job/38334381288#step:6:916

## Steps to Reproduce
Intermittent issue - example provided above.

## Context
We want reliable CI builds.

Root cause is being determined, but currently the best issue to track is https://github.com/jacoco/jacoco/issues/1801. This seems to be a new issue from Java 21.0.5. Downgrading to 21.0.4 in CI should completely avoid the problem until it's fixed - this only impacts builds that use JaCoCo, so we don't have to set a specific version in CI for all jobs, just those that run tests which also use the JaCoCo agent.

Also see https://github.com/adoptium/adoptium-support/issues/1191

## Your Environment
* Version of detekt used: main branch
* Version of Gradle used (if applicable): 8.13
* Gradle scan link (add `--scan` option when running the gradle task): not available
* Operating System and version: I believe it affects multiple OSs
* Link to your project (if it's a public repository): detekt/detekt

3flex:
Still an issue, but there is a JDK bug opened & fix has landed in Java 25/26 branches, hopefully will land in 21 soon enough.

https://bugs.openjdk.org/browse/JDK-8357885
https://bugs.openjdk.org/browse/JDK-8358534
https://github.com/openjdk/jdk/commit/704b5990a750719ca927e156553db7982637e590

Until it's fixed we can work around by downgrading to 21.0.4 to avoid having to keep rerunning CI all the time.

3flex:
Should be fixed by https://github.com/openjdk/jdk21u-dev/commit/f55ff018b2f6a14e7f751d29f56205aba33e8b69

According to https://www.java.com/releases/ the next release will be 21.0.8 on 2025-07-15 which will include that patch. #8169 can then be reverted.

3flex:
Unfortunately this didn't make 21.0.8 and will not be available until 21.0.9. Planned release date is 2025-10-21.

PR #8169, opened 2025-06-07
Use Java 21.0.4 in CI

3flex:
Workaround for https://github.com/detekt/detekt/issues/8009

### DIFF
--- a/.github/workflows/codecoverage.yaml
+++ b/.github/workflows/codecoverage.yaml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup Java
         uses: actions/setup-java@c5195efecf7bdfc987ee8bae7a71cb8b11521c00 # v4
         with:
-          java-version: 21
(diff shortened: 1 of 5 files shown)


BraisGabin (approved):

schalkms (approved):

3flex:
Updating the actions to use "21" instead of "21.0.4" in the name is a bit annoying and I find it frustrating dealing with actions so I'd rather not. I could update the required actions for the repo (and will have to flip it back later). Or we just deal with it. It mostly occurs in the coverage report job, but soon enough we'll have Java 25 and can use that version on the jobs that aren't specifically testing certain versions.

3flex:
Kept the matrix config on Java 21 which will avoid it. This will still avoid the issue occurring on any other jobs.

3flex:
Argh I forgot to update the PR title before merging, sorry about that. I've updated now but the commit has the wrong dscription.

PR #8171, opened 2025-06-08
Add Analysis API version of isNullable utility function

### LINKED ISSUE #8078: Replace TypeUtils in detekt-psi-utils with Analysis API equivalents
## Context
`KtExpression.getDataFlowAwareTypes` and `KtExpression.isNullable` require `BindingContext`. These functions need to be replaced with something compatible with Kotlin's Analysis API.

3flex:
This was the key function replacement required to close #8078

Stacks on #8175

### DIFF
--- a/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
@@ -118,8 +118,7 @@ private fun KtExpression.descriptor(bindingContext: BindingContext): CallableDes
  */
 @OptIn(KaExperimentalApi::class)
 fun KtExpression.isNullable(shouldConsiderPlatformTypeAsNullable: Boolean): Boolean {
-    val safeAccessOperation = (this as? KtSafeQualifiedExpression)?.operationTokenNode as? PsiElement
-    if (safeAccessOperation != null) {
+    if (this is KtSafeQualifiedExpression) {
         analyze(this) {
             return diagnostics(KaDiagnosticCheckerFilter.ONLY_COMMON_CHECKERS)
                 .none { it is KaFirDiagnostic.UnnecessarySafeCall }
--- a/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
+++ b/detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
@@ -1,6 +1,10 @@
 package io.gitlab.arturbosch.detekt.rules
 
 import com.intellij.psi.PsiElement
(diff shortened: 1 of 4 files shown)


detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
```diff
@@ -98,3 +102,34 @@ fun KtExpression.isNullable(
...
+ * val c = a!! // RHS expression will be not nullable
+ * val d = (a ?: error("null assertion message")) // RHS expression will be not nullable
+ * val c = 1?.and(2) // RHS expression will be not nullable
+ * ```
+ * [shouldConsiderPlatformTypeAsNullable] determines the behaviour of platform type. Passing true considers
+ * the platform type as nullable otherwise it is not considered nullable in case of false
+ */
+@OptIn(KaExperimentalApi::class)
+fun KtExpression.isNullable(shouldConsiderPlatformTypeAsNullable: Boolean): Boolean {
+    val safeAccessOperation = (this as? KtSafeQualifiedExpression)?.operationTokenNode as? PsiElement
```
BraisGabin: Maybe use `is` instead of `as`? This way instead of checking nullability you could check a boolean. This could be a detekt rule. But I'm not sure if it would be worth it.

detekt-psi-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/TypeUtils.kt
```diff
@@ -98,3 +102,34 @@ fun KtExpression.isNullable(
...
+ * val b = a // RHS expression will be nullable
+ * val c = a!! // RHS expression will be not nullable
+ * val d = (a ?: error("null assertion message")) // RHS expression will be not nullable
+ * val c = 1?.and(2) // RHS expression will be not nullable
+ * ```
+ * [shouldConsiderPlatformTypeAsNullable] determines the behaviour of platform type. Passing true considers
+ * the platform type as nullable otherwise it is not considered nullable in case of false
+ */
+@OptIn(KaExperimentalApi::class)
+fun KtExpression.isNullable(shouldConsiderPlatformTypeAsNullable: Boolean): Boolean {
```
BraisGabin, line 120: This function has a lot of conditionals. Some unit test would be good. At least to test the code in the documentation.

3flex: I didn't add tests because: 
1. The function being replaced doesn't have tests 
2. The logic will be implicitly tested by the rules that use it

But I suppose I should do better here...

3flex:
@BraisGabin thanks for the feedback, I've addressed it.

I'd like to merge https://github.com/detekt/detekt/pull/8175 before this one, so I'll mark this as draft as I'd like to keep the change to the dependencies out of this PR. 8175 handles that.

PR #8172, opened 2025-06-08
Migrate CastNullableToNonNullableType to Analysis API

3flex:
Part of #8041 

Stacks on
* #8171
* #8175


3flex:
Yep, gives a lot of confidence that things won't break even though the implementation is completely changing under the hood.

PR #8175, opened 2025-06-09
Repackage Kotlin Analysis API & Standalone API separately

### LINKED ISSUE #8102: Split detekt-kotlin-analysis-api into detekt-kotlin-analysis-api and detekt-kotlin-analysis-api-standalone
## Expected Behavior
Projects that depend on `detekt-api` only have access to Kotlin's `analysis-api`, but not `analysis-api-standalone`.

## Current Behavior
Currently the `detekt-kotlin-analysis-api` project repackages both `analysis-api` and `analysis-api-standalone` (plus dependencies).

## Context
If `detekt-kotlin-analysis-api` is added as an `api` dependency to `detekt-api` it will expose `analysis-api-standalone` to any project that depends on `detekt-api`.

Instead, only `analysis-api` should be exposed by projects that depend on `detekt-api`. `analysis-api-standalone` should only be used as an implementation dependency where needed in detekt itself.

3flex:
Also exposes Kotlin Analysis API to detekt-api consumers so the `analyze` entry point is available to rule authors, and `analysis-api-k2` so that specific diagnostics can be filtered as described in https://kotlin.github.io/analysis-api/diagnostics.html

Fixes #8102


detekt-kotlin-analysis-api-standalone/build.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+val javaComponent = components["java"] as AdhocComponentWithVariants
+javaComponent.withVariantsFromConfiguration(configurations["apiElements"]) {
+    skip()
+}
+javaComponent.withVariantsFromConfiguration(configurations["runtimeElements"]) {
+    skip()
+}
```
BraisGabin, line 42: It's amazing how many workarounds need to be on place to make AA work. As always great work with the research and then documenting it.

BraisGabin (approved):

PR #8176, opened 2025-06-09
Make CloseableWrapper implement AutoCloseable

3flex:
Fixes the following JUnit warning:

```
WARNING: Type implements CloseableResource but not AutoCloseable: io.gitlab.arturbosch.detekt.rules.KotlinEnvironmentResolver$CloseableWrapper
```

### DIFF
--- a/detekt-test-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KotlinEnvironmentTestSetup.kt
+++ b/detekt-test-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KotlinEnvironmentTestSetup.kt
@@ -51,7 +51,7 @@ internal class KotlinEnvironmentResolver : ParameterResolver {
 
     @Suppress("DEPRECATION")
     private class CloseableWrapper(val wrapper: KotlinCoreEnvironmentWrapper) :
-        ExtensionContext.Store.CloseableResource {
+        ExtensionContext.Store.CloseableResource, AutoCloseable {
         override fun close() {
             wrapper.dispose()
         }


detekt-test-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KotlinEnvironmentTestSetup.kt
```diff
@@ -51,7 +51,7 @@ internal class KotlinEnvironmentResolver : ParameterResolver {
 
     @Suppress("DEPRECATION")
     private class CloseableWrapper(val wrapper: KotlinCoreEnvironmentWrapper) :
-        ExtensionContext.Store.CloseableResource {
+        ExtensionContext.Store.CloseableResource, AutoCloseable {
```
BraisGabin, line 54: > Older versions only supported `CloseableResource`, which is deprecated but still available for backward compatibility. 

Should we remove `CloseableResource`?

BraisGabin (approved):
I disabled the automerge because I want to approve this pr because I think that this pr improves our code. I just want to know what you think about removing `CloseableResource`.

detekt-test-utils/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/KotlinEnvironmentTestSetup.kt
3flex: It's required for backwards compatibility. I don't think it's necessary to require JUnit 5.13 or higher for anyone using JUnit to test their own rules, though you could argue with a major release, we'll drop backwards compatibility... but it's not difficult to keep it.

JUnit 6 should be out this year and I assume at that point the old interface will be gone for good and we can remove it when detekt upgrades.

BraisGabin: > though you could argue with a major release, we'll drop backwards compatibility... but it's not difficult to keep it.

That was my thinking. If it's easy then it's good to keep.

PR #8184, opened 2025-06-13
Rename Rule to RuleName

### LINKED ISSUE #8092: Rename `Name` in `Rule` to something else
## Expected Behavior
It's easy to call `Name.identifier("x")` to create a `CallableId` when writing rules that use the Analysis API

## Current Behavior
There is a conflict with nested `Name` class in `Rule`

## Context
I have successfully migrated one rule in an experimental WIP branch. Note the usage of the fully qualified import `org.jetbrains.kotlin.name.Name`:

https://github.com/detekt/detekt/blob/d62eac80ef8e2ab5f7cfeab045ac9d62fa733374/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/MapGetWithNotNullAssertionOperator.kt#L65

Refer to https://kotlin.github.io/analysis-api/migrating-from-k1.html#declaration-names for an explanation of why this is being used. Rules will use `Name.identifier` frequently when using the Analysis API, but `Rule` has a nested `Name` class which currently makes this awkward.

cc @BraisGabin as I believe you introduced this class - obviously not being able to see the future! I don't have a strong opinion on what the name of the `Name` class should be. It could be made a top-level class instead of a subclass.

BraisGabin:
Closes #8092

`Rule.Name` conflicted with `org.jetbrains.kotlin.name.Name`. This PR renames `Rule.Name` with `RuleName` to avoid that collision.

### DIFF
--- a/detekt-api/api/detekt-api.api
+++ b/detekt-api/api/detekt-api.api
@@ -238,7 +238,7 @@ public class io/gitlab/arturbosch/detekt/api/Rule : io/gitlab/arturbosch/detekt/
 	protected final fun getCompilerResources ()Lio/gitlab/arturbosch/detekt/api/CompilerResources;
 	public final fun getConfig ()Lio/gitlab/arturbosch/detekt/api/Config;
 	public final fun getDescription ()Ljava/lang/String;
-	public fun getRuleName ()Lio/gitlab/arturbosch/detekt/api/Rule$Name;
+	public fun getRuleName ()Lio/gitlab/arturbosch/detekt/api/RuleName;
 	public final fun getUrl ()Ljava/net/URI;
 	protected fun postVisit (Lorg/jetbrains/kotlin/psi/KtFile;)V
 	protected fun preVisit (Lorg/jetbrains/kotlin/psi/KtFile;)V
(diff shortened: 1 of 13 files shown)


schalkms (approved):

3flex (commented):
Thanks for this! I'm wondering if there's a reason that `RulesetProvider` has `ruleSetId: RuleSet.Id`, wouldn't it make sense to use the same convention for `Rule` and use `val ruleId: Rule.Id`?

The doc for `ruleName` even says "An **id** this rule is identified with"

I can't see a reason for using Id on one and Name on the other when they're doing the same thing. Or, rename `ruleSetId` to `ruleSetName`.

I know it's not strictly required for this PR but probably a good opportunity to fix it (or in a follow up).

BraisGabin:
There are two different concepts when we talk about a rule. It has an id and a name.

The rule itself has a name but the it can have multiple ids. This is like this because we now allow multi instances of the same rule.

Naming is really difficult so I'm open to new ideas.

An implementation of a rule can only have a name. That's the one that you will use in the configuration to refer to it. But then you can instantiate it multiple times for example `ForbiddenImport/foo` and `ForbiddenImport/bar`. Those are the ids. If a rule only has one instance then its name and its id are the same.

Again, nameing things is difficult and this concept is not obvious so any idea is more than welcome.

3flex:
I think I understand. LGTM and I'll have a think. Maybe it's just that KDoc needs to expand on this concept to explain what you explained above?

Issue #8178, opened 2025-06-09
Remove CompilerResources from API


3flex:
## Expected Behavior
`CompilerResources` is removed from the API, replaced with direct usage of `LanguageVersionSettings`.

## Current Behavior
`CompilerResources` becomes an unnecessary wrapper around `LanguageVersionSettings` once #8177 is merged.

## Context
The unnecessary wrapper can be removed as part of the migration to Kotlin Analysis API.

atulgpt:
Hi @3flex I have removed the file at https://github.com/detekt/detekt/pull/8185/files PTAL

PR #8185, opened 2025-06-15
Removes `CompilerResources` wrapper

atulgpt:
Fixes: https://github.com/detekt/detekt/issues/8178


detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FakeLanguageVersionSettings.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.jetbrains.kotlin.config.AnalysisFlags
+import org.jetbrains.kotlin.config.ApiVersion
+import org.jetbrains.kotlin.config.ExplicitApiMode
+import org.jetbrains.kotlin.config.LanguageVersion
+import org.jetbrains.kotlin.config.LanguageVersionSettingsImpl
+
+@Suppress("FunctionName")
+fun FakeLanguageVersionSettings(
+    mode: ExplicitApiMode = ExplicitApiMode.DISABLED,
+): LanguageVersionSettingsImpl = LanguageVersionSettingsImpl(
```
3flex, line 12: I don't see a need for this wrapper now - I think it would be better to remove it. But that can be a follow up if you don't want to do it now, it's not as important as tidying up the main API.

3flex (approved):

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FakeLanguageVersionSettings.kt
BraisGabin: I agree that this is probably not needed. But it's for sure handy. It has sane defaults. And its api is more handy to configure the explicit mode that is the only thing we care about right now.

So +1 to keep this in this pr and decide afterwards if we want to remove it.

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt
```diff
@@ -37,16 +37,16 @@ fun Rule.lint(
         KotlinAnalysisApiEngine.compile(content)
     }
     val ktFile = compileContentForTest(content)
-    return visitFile(ktFile, compilerResources = compilerResources).filterSuppressed(this)
+    return visitFile(ktFile, languageVersionSettings = languageVersionSettings).filterSuppressed(
+        this
+    )
```
BraisGabin: Please, try to avoid automatic formats. They generate noise on the pr.

BraisGabin (approved):

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FakeLanguageVersionSettings.kt
atulgpt: I had two thoughts when dealing with this method
1. Completely remove it but then I have copy the complete `LanguageVersionSettingsImpl(`
2. Make it more generic which can take list of pair of api modes so that it can be configured. But it would have taken more time to implement

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt
atulgpt: Reverted

detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/FakeLanguageVersionSettings.kt
3flex: It's fine, we can look at it separately. I doubt it's used much if at all outside this codebase anyway so changing later won't impact much.

Issue #8114, opened 2025-04-27
Allow mixed configurations for rules that use `valuesWithReason`


3flex:
## Expected Behavior
I can use config like:

```yaml
  ForbiddenMethodCall:
    active: true
    methods:
      - 'kotlin.io.print'
      - value: 'org.jetbrains.kotlin.analysis.project.structure.builder.buildKtLibraryModule'
        reason: 'Not performant due to KT-71706. Use buildKspLibraryModule instead.'
```

## Current Behavior
```
> Analyzing C:\Users\user\IdeaProjects\detekt\detekt-test-utils\src\main\kotlin\io\github\detekt\test\utils\CompileExtensions.kt led to an exception.
  Location: io.gitlab.arturbosch.detekt.api.ConfigPropertyKt.getValuesWithReasonOrDefault(ConfigProperty.kt:160)
  The original exception message was: Only lists of strings or maps with keys 'value' and 'reason' are supported. 'methods' is invalid.
```

To avoid it, I have to do this:
```yaml
  ForbiddenMethodCall:
    active: true
    methods:
      - value: 'kotlin.io.print'
      - value: 'org.jetbrains.kotlin.analysis.project.structure.builder.buildKtLibraryModule'
        reason: 'Not performant due to KT-71706. Use buildKspLibraryModule instead.'
```

But requiring `value: ` for all listed methods should not be necessary if I only want to set the reason for a subset of methods.

I can also just leave out the reason:
```yaml
  ForbiddenMethodCall:
    active: true
    methods:
      - 'kotlin.io.print'
      - 'org.jetbrains.kotlin.analysis.project.structure.builder.buildKtLibraryModule'
```

But this isn't a good workaround because the reason is lost.

## Context
Configure value and reason only for methods that really need it.

PR #8188, opened 2025-06-15
Add support for mixed types(`string` and `ValueWithReason`) in yml

atulgpt:
Fixes: https://github.com/detekt/detekt/issues/8114

### DIFF
--- a/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/config/YamlConfigSpec.kt
+++ b/detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/config/YamlConfigSpec.kt
@@ -221,6 +221,24 @@ class YamlConfigSpec {
                     .isNotNull
                     .isEmpty()
             }
+
+            @Test
+            fun `supports mixed string and dictionary`() {
+                val actualAsMap: List<Map<*, *>>? = config
+                    .subConfig("style")
+                    .subConfig("MixedWithStringAnsMaps")
+                    .valueOrNull("values")
+                assertThat(actualAsMap)
+                    .isNotNull
+                    .isNotEmpty
+                    .hasSize(6)
+                    .elements(0, 1)
+                    .hasOnlyElementsOfType(String::class.java)
+
+                assertThat(actualAsMap)
+                    .elements(2, 3, 4, 5)
+                    .hasOnlyElementsOfType(Map::class.java)
+            }
         }
 
         @Test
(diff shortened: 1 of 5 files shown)


schalkms (approved):

detekt-core/src/test/kotlin/io/gitlab/arturbosch/detekt/core/config/YamlConfigSpec.kt
```diff
@@ -221,6 +221,24 @@ class YamlConfigSpec {
                     .isNotNull
                     .isEmpty()
             }
+
+            @Test
+            fun `supports mixed string and dictionary`() {
+                val actualAsMap: List<Map<*, *>>? = config
+                    .subConfig("style")
+                    .subConfig("MixedWithStringAnsMaps")
```
3flex:
```suggestion
                    .subConfig("MixedWithStringAndMaps")
```

detekt-core/src/test/resources/values-with-reason.yml
```diff
@@ -13,3 +13,13 @@ style:
         reason: reason C
   EmptyListOfMaps:
     values: []
+  MixedWithStringAnsMaps:
```
3flex:
```suggestion
  MixedWithStringAndMaps:
```

BraisGabin:
🤔 I'm not 100% sure about this, I developed this time ago but I think that something in `getValuesWithReasonOrDefault` should change. Maybe, I'm not sure, that's the only thing that needs to be changed and tested. I repeat, I'm not sure.

But I guess that if you do https://github.com/detekt/detekt/pull/8188#pullrequestreview-2933994093 we will have more information.

detekt-core/src/test/resources/values-with-reason.yml
atulgpt: Done

atulgpt:
> 🤔 I'm not 100% sure about this, I developed this time ago but I think that something in `getValuesWithReasonOrDefault` should change. Maybe, I'm not sure, that's the only thing that needs to be changed and tested. I repeat, I'm not sure.
> 
> But I guess that if you do [#8188 (review)](https://github.com/detekt/detekt/pull/8188#pullrequestreview-2933994093) we will have more information.

I think we are reducing the verbosity, so it should be fine.. BTW I am not changing the way detekt generator generate the config file so as an example it will generate the detekt.yml with value: key: pair only

atulgpt:
@BraisGabin @3flex do we need to update the schema elsewhere(like some schemastore?) if this PR got merged? I was getting `Schema validation: Incompatible types.
Required: object. Actual: string.` this error?

BraisGabin:
> I was getting `Schema validation: Incompatible types.
> Required: object. Actual: string.` this error? 

That scheme is auto generated somehow by someone outside of detekt. We have a issue to generate that file ourselves with all the information about our configuration. The current one is just based on the default config and it doesn't have some keys that can also be used to configure a rule. For example the suppressors. If you search for jsonscheme or something similar in our issues you will find it.

BraisGabin:
Also sorry, I didn't realised that you were changing exactly the function I mentioned 😅

I'm not 100% sure if that should be on our public api but that's another topic because that was like that before.

Issue #8046, opened 2025-03-24
Migrate detekt-rules-style to use Kotlin AA for semantic analysis


3flex:
### Rule migration
1. Find a rule that implements `RequiresFullAnalysis`
2. Replace `RequiresFullAnalysis` with `RequiresAnalysisApi`
3. IDE will flag uses of `bindingContext` which is not available when using `RequiresFullAnalysis` as it uses the K1 compiler
4. Refer to Analysis API migration guide: https://kotlin.github.io/analysis-api/migrating-from-k1.html
5. Complete migration

All tests should still pass, unmodified.

## Expected Behavior of the rule
The following rules are migrated to use Kotlin AA for semantic analysis:

- [x] AbstractClassCanBeConcreteClass
- [x] AbstractClassCanBeInterface
- [x] CanBeNonNullable
- [x] DoubleNegativeExpression
- [x] ExplicitCollectionElementAccessMethod
- [x] ForbiddenAnnotation
- [x] ForbiddenMethodCall
- [x] ForbiddenNamedParam
- [x] ForbiddenVoid
- [x] MaxChainedCallsOnSameLine
- [x] MultilineLambdaItParameter
- [x] NullableBooleanCheck
- [x] ObjectLiteralToLambda
- [x] OptionalUnit
- [x] RedundantExplicitType
- [x] RedundantHigherOrderMapUsage
- [x] UnnecessaryAny
- [x] UnnecessaryApply
- [x] UnnecessaryBracesAroundTrailingLambda
- [x] UnnecessaryFilter
- [x] UnnecessaryInnerClass
- [x] UnnecessaryLet
- [x] UnnecessaryReversed
- [x] UnusedImport
- [x] UnusedPrivateFunction
- [x] UnusedPrivateProperty
- [x] UnusedVariable
- [x] UseAnyOrNoneInsteadOfFind
- [x] UseCheckNotNull
- [x] UseCheckOrError
- [x] UseDataClass
- [x] UseEmptyCounterpart
- [x] UseIfEmptyOrIfBlank
- [x] UseIsNullOrEmpty
- [x] UselessCallOnNotNull
- [x] UseOrEmpty
- [x] UseRequire
- [x] UseRequireNotNull
- [x] UseSumOfInsteadOfFlatMapSize
- [x] VarCouldBeVal

## Context
Required for the migration to Kotlin's Analysis API.

inorichi:
[...]

PR #8192, opened 2025-06-16
Migrate UseRequireNotNull to Analysis API

3flex:
Part of #8046


3flex:
> Is the reduction in test coverage a concern?

I don't think so? The rule has 100% coverage in this case, and the reduction overall is going to be due to the new functions, which like the old, don't have full coverage. Removing the now unused function (which I'll do in a new PR) should bring the overall coverage up again slightly.

Issue #8045, opened 2025-03-24
Migrate detekt-rules-performance to use Kotlin AA for semantic analysis


3flex:
## Expected Behavior of the rule
The following rules are migrated to use Kotlin AA for semantic analysis:
* ArrayPrimitive
* CouldBeSequence
* SpreadOperator

## Context
Required for the migration to Kotlin's Analysis API.

marschwar:
I would like to pick this up.

3flex:
Closed by: 
* #8197
* #8199
* #8326

PR #8199, opened 2025-06-16
Migrate ArrayPrimitive to Analysis API

marschwar:
part of #8045


detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/ArrayPrimitive.kt
```diff
@@ -4,25 +4,23 @@ import io.gitlab.arturbosch.detekt.api.ActiveByDefault
...
 import io.gitlab.arturbosch.detekt.api.Finding
-import io.gitlab.arturbosch.detekt.api.RequiresFullAnalysis
+import io.gitlab.arturbosch.detekt.api.RequiresAnalysisApi
 import io.gitlab.arturbosch.detekt.api.Rule
-import org.jetbrains.kotlin.builtins.KotlinBuiltIns
+import org.jetbrains.kotlin.analysis.api.analyze
+import org.jetbrains.kotlin.analysis.api.resolution.singleFunctionCallOrNull
 import org.jetbrains.kotlin.builtins.PrimitiveType
-import org.jetbrains.kotlin.descriptors.CallableDescriptor
 import org.jetbrains.kotlin.name.FqName
```
3flex: Can use of `FqName` be switched to `CallableId`?

marschwar: yes, it can. I don't know if it helps at all, as only the function name is used as a string.

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/ArrayPrimitive.kt
```diff
@@ -39,22 +39,24 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
...
 
-        val descriptor = expression.getResolvedCall(bindingContext)?.resultingDescriptor
-        if (descriptor != null && isArrayPrimitive(descriptor)) {
+        if (expression.returnsArrayPrimitive()) {
             report(Finding(Entity.from(expression), description))
         }
     }
 
+    private fun KtCallExpression.returnsArrayPrimitive(): Boolean = analyze(this) {
+        val callInfo = resolveToCall()?.singleFunctionCallOrNull() ?: return false
```
3flex:
```suggestion
        val functionCall = resolveToCall()?.singleFunctionCallOrNull() ?: return false
```

`resolveToCall` returns `CallInfo` but `singleFunctionCallOrNull` returns `KaFunctionCall<*>?` so this name would be a bit clearer.

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/ArrayPrimitive.kt
```diff
@@ -39,22 +39,24 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
...
+        val returnType = callInfo.partiallyAppliedSymbol.signature.returnType
```
3flex: `partiallyAppliedSymbol` could be useful in other places too - I didn't see this when migrating earlier rules. It sounds like it's a lighter option than using `symbol` directly and should be used when the information that's missing from the partially applied symbol isn't required?

marschwar: I am not sure. My problem here was, that `functionCall.symbol.returnType` returns `kotlin.Array<T>` instead of `kotlin.Array<kotlin.Float>`. Since this is the first time I looked at the Analysis API, I am not sure what is going on here.

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/ArrayPrimitive.kt
3flex: Since Analysis API works natively with CallableId I think they should all be switched over as we go, regardless of how they're being used. Or if it's only being used as a string then it should probably just be a string.

Issue #8190, opened 2025-06-16
`CouldBeSequence` is missing some functions


marschwar:
## Expected Behavior
The `CouldBeSequence` rule should check for all standard lib collection functions that return a new collection instance.

## Observed Behavior
Some functions are not accounted for even though they create a new collection and should be replaced by a sequence.

## Steps to Reproduce
```kotlin
    @Test
    fun `all functions should be accounted for`() {
        val code = """
        val bar = listOf<String?>("bar")
            .filterIsInstance<String>()
            .filterNot { true }
            .filterNotNull()
            .take(1)
            .drop(1)
            .takeWhile { true }
            .mapNotNull { it }
        """.trimIndent()
        assertThat(subject.lintWithContext(env, code)).hasSize(1)
```

atulgpt:
Not stale PR is raised https://github.com/detekt/detekt/pull/8200

PR #8200, opened 2025-06-16
[CouldBeSequence] Handle all the functions from collections which are present in Sequences

atulgpt:
Fixes https://github.com/detekt/detekt/issues/8190


detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/CouldBeSequence.kt
```diff
@@ -7,13 +7,15 @@ import io.gitlab.arturbosch.detekt.api.Finding
 import io.gitlab.arturbosch.detekt.api.RequiresFullAnalysis
 import io.gitlab.arturbosch.detekt.api.Rule
 import io.gitlab.arturbosch.detekt.api.config
+import org.jetbrains.kotlin.incremental.components.NoLookupLocation
```
3flex: I'm not sure why this is being used, maybe there's a good reason, but this is a bit of a smell to me - detekt shouldn't need to reference anything to do with incremental compilation in the compiler.

A better approach may be to convert this rule to use the Analysis API first, then update this PR.

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/CouldBeSequence.kt
```diff
@@ -65,28 +67,36 @@ class CouldBeSequence(config: Config) :
...
-            FqName("kotlin.collections.mapIndexed"),
-            FqName("kotlin.collections.flatMap"),
-            FqName("kotlin.collections.flatMapIndexed"),
-            FqName("kotlin.collections.reduce"),
-            FqName("kotlin.collections.zip")
+        val sequenceFunctions = sequencePackage.memberScope.getContributedFunctions(
+            functionName,
+            NoLookupLocation.FROM_BACKEND
         )
+
```
marschwar: I like the idea to make the list dynamic, but I see 2 potential issues here.
* It "feels" very expensive, to look up the sequence counter part for each usage of a collection function. Do we have any idea of the implications?
* I worry we might introduce false positives. For example the use of `filterTo` in my opinion should not be flagged.

detekt-rules-performance/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/CouldBeSequenceSpec.kt
```diff
@@ -26,6 +26,33 @@ class CouldBeSequenceSpec(val env: KotlinEnvironmentContainer) {
...
+            }.filterIndexed { index, _ ->
+                index % 2 == 0
+            }.filterIsInstance<Int>()
+            .take(10)
+            .takeWhile { true }
+            .drop(1)
+            .toList()
+        """.trimIndent()
+        assertThat(subject.lintWithContext(env, code)).hasSize(1)
+    }
```
marschwar: I am afraid we would need to test this differently. Once the counter is above the threshold, no more checks will take place. So maybe a parameterized checking one function at a time.

atulgpt:
> rule

No worries.. I will take a look after resolving the conflict

marschwar:
@atulgpt Are you planning to pick this up again?

atulgpt:
> @atulgpt Are you planning to pick this up again?

Hi @marschwar I have updated my logic as per the analysis api

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/CouldBeSequence.kt
atulgpt: I have removed this import and now simply using `findTopLevelCallables`

atulgpt:
> Ignore my inline comment - I forgot this rule is already migrated to Analysis API! See #8197. This PR will need to be redone based on the Analysis API instead. Apologies for the rework - perhaps this migration should be mentioned in the contribution guide and/or README...

Migrated as per the analysis api

detekt-rules-performance/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/performance/CouldBeSequenceSpec.kt
atulgpt: Hi @marschwar can you explain bit more?

detekt-rules-performance/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/performance/CouldBeSequence.kt
atulgpt: For pt #1, I am using `findTopLevelCallables(` in AA api which takes the package and name so now it should not have any performance spike

for pt #2 Can you construct the example or TC? Currently all the TC is passing

detekt-rules-performance/src/main/kotlin/dev/detekt/rules/performance/CouldBeSequence.kt
```diff
@@ -66,29 +67,30 @@ class CouldBeSequence(config: Config) :
...
+                ?.symbol
+                ?.callableId
+                ?: return false
+            if (!callableId.toString().startsWith(KOTLIN_COLLECTION_PREFIX)) return false
+            findTopLevelCallables(
+                SEQUENCE_FQ_NAME,
+                callableId.callableName
+            ).any()
+        }
+    }
```
marschwar:
```suggestion
    private fun KtExpression.isCallingKotlinCollectionFunPresentInSequence(): Boolean = analyze(this) {
        val callableId = resolveToCall()
            ?.singleCallOrNull<KaCallableMemberCall<*, *>>()
            ?.symbol
            ?.callableId
        callableId?.packageName == StandardClassIds.BASE_COLLECTIONS_PACKAGE
            && findTopLevelCallables(StandardClassIds.BASE_SEQUENCES_PACKAGE, callableId.callableName).any()
    }
```

detekt-rules-performance/src/test/kotlin/dev/detekt/rules/performance/CouldBeSequenceSpec.kt
```diff
@@ -79,4 +106,76 @@ class CouldBeSequenceSpec(val env: KotlinEnvironmentContainer) {
...
+            .filterIsInstance<String>()
+            .filterNot { true }
+            .filterNotNull()
+            .take(1)
+            .drop(1)
+            .takeWhile { true }
+            .mapNotNull { it }
+        """.trimIndent()
+        assertThat(subject.lintWithContext(env, code)).hasSize(1)
+    }
```
marschwar:
```suggestion
    @ParameterizedTest
    @ValueSource(strings = [
        ".filterIsInstance<String>()",
        ".filterNot { true }",
        ".filterNotNull()",
        ".take(1)",
        ".drop(1)",
        ".takeWhile { true }",
        ".mapNotNull { it }",
    ])
    fun `#8190 - all functions should be accounted for`(additionalCall: String) {
        val code = """
        val bar = listOf<String>("bar")
            .filter { it.length > 1}
            .map { "text" }
            $additionalCall
        """.trimIndent()
        assertThat(subject.lintWithContext(env, code)).hasSize(1)
    }
```

This will check each method individually.

marschwar (changes requested):
Could you please add a test case that verifies that those functions that do not have a sequence equivalent, will not trigger the rule.

atulgpt:
> Could you please add a test case that verifies that those functions that do not have a sequence equivalent, will not trigger the rule.

Hi @marschwar there is already one TC added in this PR `using method not allowed on sequence should not trigger rule` does that not serve the purpose?

marschwar:
You are right. I missed that one.

marschwar (approved):

atulgpt:
@3flex can you also review this?

detekt-rules-performance/src/test/kotlin/dev/detekt/rules/performance/CouldBeSequenceSpec.kt
```diff
@@ -79,4 +108,83 @@ class CouldBeSequenceSpec(val env: KotlinEnvironmentContainer) {
...
+                .asSequence()
+                .filterIsInstance<String>()
+                .filterNot { true }
+                .filterNotNull()
+                .take(1)
+                .drop(1)
+                .takeWhile { true }
+                .toList()
+                .takeLast(2)
+                .mapNotNull { it }
```
BraisGabin, line 147: 🤔 Out of scope of this PR but: How hard would it be to take these 2 as error? It have no sense to call `.toList()` and then keep doing this other functions. I think that in this case the threshold should be 0.

atulgpt: I think this can be done. Can you create a tracking issue?

BraisGabin: I will

marschwar:
@3flex  Were the changes you requested related to the AA transformation? Can this be merged?

atulgpt:
> @3flex Were the changes you requested related to the AA transformation? Can this be merged?

Hi yes it was. You can wait for @3flex confirmation

atulgpt:
Hi @BraisGabin @marschwar should we merge this PR?

marschwar:
Yes, I think we should

atulgpt:
@3flex can you also review it once

atulgpt:
Finally 🥹 thanks everyone

Issue #8042, opened 2025-03-24
Migrate detekt-rules-exceptions to use Kotlin AA for semantic analysis


3flex:
### Rule migration
1. Find a rule that implements `RequiresFullAnalysis`
2. Replace `RequiresFullAnalysis` with `RequiresAnalysisApi`
3. IDE will flag uses of `bindingContext` which is not available when using `RequiresFullAnalysis` as it uses the K1 compiler
4. Refer to Analysis API migration guide: https://kotlin.github.io/analysis-api/migrating-from-k1.html
5. Complete migration

All tests should still pass, unmodified.

## Expected Behavior of the rule
The following rules are migrated to use Kotlin AA for semantic analysis:
- [ ] ErrorUsageWithThrowable
- [ ] InstanceOfCheckForException
- [ ] ObjectExtendsThrowable
- [ ] ReturnFromFinally

## Context
Required for the migration to Kotlin's Analysis API.

BraisGabin:
All these rules are ported now :)

PR #8214, opened 2025-06-27
Refactor ObjectExtendsThrowable with the analysis api

inorichi:
This PR refactors the `ObjectExtendsThrowable` rule to use the Analysis API as part of #8042

### DIFF
--- a/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
+++ b/detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
@@ -3,12 +3,11 @@ package io.gitlab.arturbosch.detekt.rules.exceptions
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Finding
-import io.gitlab.arturbosch.detekt.api.RequiresFullAnalysis
+import io.gitlab.arturbosch.detekt.api.RequiresAnalysisApi
 import io.gitlab.arturbosch.detekt.api.Rule
+import org.jetbrains.kotlin.analysis.api.analyze
+import org.jetbrains.kotlin.name.ClassId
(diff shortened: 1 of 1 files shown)


detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
```diff
@@ -59,10 +58,10 @@ class ObjectExtendsThrowable(config: Config) :
...
     }
 
-    private fun KtObjectDeclaration.isSubtypeOfThrowable(): Boolean =
-        bindingContext[BindingContext.CLASS, this]
-            ?.defaultType
-            ?.supertypes()
-            .orEmpty()
-            .any { it.isNotNullThrowable() }
+    private fun KtObjectDeclaration.isSubtypeOfThrowable(): Boolean {
+        val throwable = ClassId.fromString("kotlin/Throwable")
```
BraisGabin, line 66: We could move to a top level constant (or companion) to avoid creating it every single time.

inorichi: Moved to companion object 👍

detekt-rules-exceptions/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/exceptions/ObjectExtendsThrowable.kt
```diff
@@ -44,7 +43,7 @@ class ObjectExtendsThrowable(config: Config) :
             "type of Throwable consequently introduces a global singleton exception whose instance may be " +
```
harrytmthy, line 43: Nit: there is a small typo at line 41:

> "An `object` should not extend and type of Throwable..."

should be

> "An `object` should not extend any type of Throwable..."

Minor, but worth fixing to keep the description clear and polished 🙂

inorichi: Done

Issue #8039, opened 2025-03-24
Migrate detekt-rules-complexity to use Kotlin AA for semantic analysis


3flex:
## Expected Behavior of the rule
The following rules are migrated to use Kotlin AA for semantic analysis:
* NamedArguments
* NestedScopeFunctions
* ReplaceSafeCallChainWithRun
* LongParameterList

## Context
Required for the migration to Kotlin's Analysis API.

BraisGabin:
To migrate `LongParameterList` we first need to migrate `AnnotationExcluder`.

PR #8220, opened 2025-06-27
Migrate LongParameterList to AA

BraisGabin:
Part of #8039

To migrate `LongParamenterList` I had to migrate `AnnotationExcluder` and with it the other usage of it: `AnnotationsSuppressor`.


3flex:
> I had to migrate AnnotationExcluder and with it the other usage of it: AnnotationsSuppressor

Is it possible to break this out into a separate PR by any chance?

BraisGabin:
Sure I can. But that will increase the lines to review because I'll need to create a copy of `AnnotationExcluder` (and their tests). I'm on holidays right now (away from a computer, I only have my phone) but I will do it once I'm back.

BraisGabin:
Replaced by #8421, #8422 and #8423

PR #8223, opened 2025-06-27
Migrate UseIsNullOrEmpty to Analysis API

t-kameyama:
Part of https://github.com/detekt/detekt/issues/8046

### DIFF
--- a/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIsNullOrEmpty.kt
+++ b/detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIsNullOrEmpty.kt
@@ -4,22 +4,23 @@ import io.gitlab.arturbosch.detekt.api.ActiveByDefault
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Finding
-import io.gitlab.arturbosch.detekt.api.RequiresFullAnalysis
+import io.gitlab.arturbosch.detekt.api.RequiresAnalysisApi
 import io.gitlab.arturbosch.detekt.api.Rule
-import io.gitlab.arturbosch.detekt.rules.fqNameOrNull
-import org.jetbrains.kotlin.builtins.StandardNames
+import org.jetbrains.kotlin.analysis.api.analyze
+import org.jetbrains.kotlin.analysis.api.resolution.singleFunctionCallOrNull
+import org.jetbrains.kotlin.analysis.api.resolution.symbol
+import org.jetbrains.kotlin.analysis.api.types.symbol
 import org.jetbrains.kotlin.lexer.KtTokens
-import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.name.CallableId
+import org.jetbrains.kotlin.name.ClassId
+import org.jetbrains.kotlin.name.Name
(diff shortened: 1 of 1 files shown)


detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIsNullOrEmpty.kt
```diff
@@ -125,47 +131,65 @@ class UseIsNullOrEmpty(config: Config) :
...
 
-    private fun KtExpression?.isCalling(fqNames: List<FqName>): Boolean {
+    private fun KtExpression?.isCalling(callableIds: List<CallableId>): Boolean {
         val callExpression = this as? KtCallExpression
             ?: (this as? KtDotQualifiedExpression)?.selectorExpression as? KtCallExpression
             ?: return false
-        return callExpression.calleeExpression?.text in fqNames.map { it.shortName().asString() } &&
-            callExpression.getResolvedCall(bindingContext)?.resultingDescriptor?.fqNameOrNull() in fqNames
+        if (callExpression.calleeExpression?.text !in callableIds.map { it.asSingleFqName().shortName().asString() }) {
+            return false
```
BraisGabin, line 139: In general I don't care too much about codecov but it is strage that this line is not covered, right? Do we even need it?

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIsNullOrEmpty.kt
```diff
@@ -125,47 +131,65 @@ class UseIsNullOrEmpty(config: Config) :
...
+        }
+        return analyze(callExpression) {
+            callExpression.resolveToCall()?.singleFunctionCallOrNull()?.symbol?.callableId in callableIds
+        }
```
BraisGabin, line 143:
```suggestion
        return callExpression.isCalling(callableIds)
```

(It is an extension function, it will need the import so don't apply directly)

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIsNullOrEmpty.kt
t-kameyama: I believe it's necessary to avoid unnecessary executions of analyze().

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIsNullOrEmpty.kt
t-kameyama: It leads to a circular reference.

BraisGabin: Oh wow, true. This extension function is called the same as the other. 🤷 Well, this is not important change so resolved.

detekt-rules-style/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIsNullOrEmpty.kt
BraisGabin: Oh, I see. This is a shortcut. So it's just that we don't have any test that uses it.

Off topic: We need a way to measure the performance of our rules so much!

BraisGabin (approved):

PR #8235, opened 2025-06-29
Migrate Deprecation to Analysis API

t-kameyama:
Part of https://github.com/detekt/detekt/issues/8041

### DIFF
--- a/detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinAnalysisApiEngine.kt
+++ b/detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinAnalysisApiEngine.kt
@@ -42,7 +42,7 @@ object KotlinAnalysisApiEngine {
      * @throws IllegalStateException if the given code snippet does not compile
      */
     @Suppress("LongMethod")
-    fun compile(@Language("kotlin") code: String): KtFile {
+    fun compile(@Language("kotlin") code: String, dependencyCodes: List<String> = emptyList()): KtFile {
         val disposable = Disposer.newDisposable()
 
         @OptIn(KaImplementationDetail::class, KaPlatformInterface::class)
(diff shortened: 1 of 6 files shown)


BraisGabin:
@t-kameyama can you run `./gradlew apiDump` and commit the changes? That should fix CI.

detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinAnalysisApiEngine.kt
```diff
@@ -76,12 +76,17 @@ object KotlinAnalysisApiEngine {
 
                 val vf = LightVirtualFile("dummy.kt", code)
```
3flex, line 77: I think this should be renamed now. It was left as `dummy.kt` to align to the older behaviour, but it's not very meaningful.

Maybe `code_for_test.kt` or something? (I'm not great at naming sometimes!)

That can be a follow up though, it's not necessary here.

detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinAnalysisApiEngine.kt
```diff
@@ -90,7 +95,7 @@ object KotlinAnalysisApiEngine {
         }
 
         try {
-            val file = session.modulesWithFiles.values.flatMap { it }.single() as KtFile
+            val file = session.modulesWithFiles.values.flatten().first { it.name == "dummy.kt" } as KtFile
```
3flex: This should work and makes the intent slightly clearer IMHO:

```suggestion
            val file = session.modulesWithFiles.values.flatten().single { it.name == "dummy.kt" } as KtFile
```

3flex (approved):
Thanks for this!

t-kameyama:
> @t-kameyama can you run `./gradlew apiDump` and commit the changes? That should fix CI.

[5f04bb4](https://github.com/detekt/detekt/pull/8235/commits/5f04bb49ade57f7194473417067b438b21606dea)

detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinAnalysisApiEngine.kt
t-kameyama: 31dc4e1

BraisGabin (approved):

PR #8243, opened 2025-06-30
Migrate UnnecessarySafeCall to Analysis API

t-kameyama:
Part of https://github.com/detekt/detekt/issues/8041

### DIFF
--- a/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
+++ b/detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
@@ -1,16 +1,16 @@
 package io.gitlab.arturbosch.detekt.rules.bugs
 
-import com.intellij.psi.impl.source.tree.LeafPsiElement
 import io.gitlab.arturbosch.detekt.api.ActiveByDefault
 import io.gitlab.arturbosch.detekt.api.Config
 import io.gitlab.arturbosch.detekt.api.Entity
 import io.gitlab.arturbosch.detekt.api.Finding
-import io.gitlab.arturbosch.detekt.api.RequiresFullAnalysis
+import io.gitlab.arturbosch.detekt.api.RequiresAnalysisApi
 import io.gitlab.arturbosch.detekt.api.Rule
-import org.jetbrains.kotlin.diagnostics.Errors
-import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.analysis.api.KaExperimentalApi
+import org.jetbrains.kotlin.analysis.api.analyze
+import org.jetbrains.kotlin.analysis.api.components.KaDiagnosticCheckerFilter
+import org.jetbrains.kotlin.analysis.api.fir.diagnostics.KaFirDiagnostic
 import org.jetbrains.kotlin.psi.KtSafeQualifiedExpression
-import org.jetbrains.kotlin.psi.psiUtil.getChildOfType
 
 /**
  * Reports unnecessary safe call operators (`?.`) that can be removed by the user.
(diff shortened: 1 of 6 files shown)


detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/RuleExtensions.kt
```diff
@@ -66,8 +66,9 @@ fun <T> T.lintWithContext(
     environment: KotlinEnvironmentContainer,
     @Language("kotlin") content: String,
     @Language("kotlin") vararg dependencyContents: String,
+    allowCompilationErrors: Boolean = false,
```
BraisGabin, line 69: I wrote a comment of 15+ lines about the name of this parameter. I end up removing it all. I think this is a better name than `compile`.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
```diff
@@ -31,22 +31,19 @@ class UnnecessarySafeCall(config: Config) :
...
     ),
-    RequiresFullAnalysis {
+    RequiresAnalysisApi {
 
+    @OptIn(KaExperimentalApi::class)
     override fun visitSafeQualifiedExpression(expression: KtSafeQualifiedExpression) {
         super.visitSafeQualifiedExpression(expression)
 
-        val safeAccessElement = expression.getChildOfType<LeafPsiElement>()
-        if (safeAccessElement == null || safeAccessElement.elementType != KtTokens.SAFE_ACCESS) {
```
3flex, line 40: Without looking into this further, if this is useful for returning before hitting `analyze`, it should be kept. PSI is faster than analysis sessions.

https://kotlin.github.io/analysis-api/case-studies.html#k1-implementation

> Here we check whether the function is named `main` to avoid running further checking logic that depends on semantic code analysis. It is crucial, as PSI-based checks are much faster than semantic ones.

detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinAnalysisApiEngine.kt
```diff
@@ -42,7 +42,11 @@ object KotlinAnalysisApiEngine {
      * @throws IllegalStateException if the given code snippet does not compile
      */
     @Suppress("LongMethod")
-    fun compile(@Language("kotlin") code: String, dependencyCodes: List<String> = emptyList()): KtFile {
+    fun compile(
+        @Language("kotlin") code: String,
+        dependencyCodes: List<String> = emptyList(),
+        allowCompilationErrors: Boolean = false,
```
3flex, line 48: I'm not sure about this. I think it would be better to either change the code snippet so it successfully compiles or to explicitly disable compilation (if compilation errors are permitted then what's the point of compiling it in the first place?)

The snippets for this test can be updated to remove usage of the `com.sample.function.from.outside` for example and instead use a flexible type from JRE like `System.getProperty` so the tests still pass. Using a flexible type instead of an "external" type does change the nature of the test slightly but in this case the rule leverages the built in diagnostics so there's little risk of something being missed if tests are updated in that way.

BraisGabin: The problem with that is that on detekt we some times doesn't get exactly the same classpath. And in those cases we generate false positives. I guess that this test is here to avoid those false positives. It's better a false negative than a false positive.

I agree that the problem shouldn't be fixed on the rule, it's something that we should fix on detekt so that doesn't happen and this code is just not needed. But we aren't there yet.

For that reason we don't print all the type errors that detekt finds and we just hide them behind the debug flag. We shouldn't need that either. We shouldn't have those kind of errors and if they happen we should show them to the user so they are reported and we can fix them.

In my opinion we should focus on those other problems and, for now, allow this. Once we fix that other part we can remove this flag because we don't need to test cases that doesn't happen.

detekt-rules-errorprone/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/bugs/UnnecessarySafeCall.kt
t-kameyama: Thanks for the comment! In this case, the expression is always a `KtSafeQualifiedExpression`, so the early return will never happen. Therefore, this check is not necessary.

detekt-test-utils/src/main/kotlin/io/github/detekt/test/utils/KotlinAnalysisApiEngine.kt
t-kameyama: What do you think about this case?

https://github.com/detekt/detekt/blob/a0f69ac9e370d2d6a159f9964f6d13294fb96a85/detekt-rules-style/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/UselessCallOnNotNullSpec.kt#L237-L247

It was updated in https://github.com/detekt/detekt/pull/8269.

BraisGabin:
Merging for now. If we find that `allowCompilationErrors` was a bad idea we can remove it later.

